PR #3087, opened 2017-12-26
Fixing assertion in default_distribution_policy

### LINKED ISSUE #3054: Assertion failure when using bulk hpx::new_ in asynchronous mode
I stumbled upon a weird assertion failure in `hpx/runtime/components/default_distribution_policy.hpp`.
You can reproduce it by adding the following lines to `tests/regressions/components/new_2848.cpp`
```c++
    {
    auto locs = hpx::find_all_localities();
    
    std::vector<hpx::id_type> ids =
        hpx::new_<test_server[]>(hpx::default_layout(locs), 10, 42).get();

    hpx::future<std::vector<hpx::id_type>> ids_f =
        hpx::new_<test_server[]>(hpx::default_layout(locs), 10, 42);
    }
```
and running `new_2848_test` on more than one locality.
```
new_2848_test: /home/igor/hpx/hpx.git/hpx/runtime/components/default_distribution_policy.hpp:163: hpx::components::default_distribution_policy::bulk_create(std::size_t, Ts&& ...) const::<lambda(std::vector<hpx::lcos::future<std::vector<hpx::naming::id_type> > >&&)> mutable [with Component = test_server; Ts = {int}]: Assertion `localities_.size() == v.size()' failed.
```
The weird part here is that `ids` are returned normally, and only the second call to `hpx::new_` crashes.

hkaiser:
- this fixes #3054

### DIFF
--- a/hpx/components/containers/container_distribution_policy.hpp
+++ b/hpx/components/containers/container_distribution_policy.hpp
@@ -14,6 +14,7 @@
 #include <hpx/runtime/serialization/serialize.hpp>
 #include <hpx/runtime/serialization/shared_ptr.hpp>
 #include <hpx/runtime/serialization/vector.hpp>
+#include <hpx/util/assert.hpp>
 
 #include <algorithm>
 #include <cstddef>
@@ -30,21 +31,20 @@ namespace hpx
     struct container_distribution_policy
       : components::default_distribution_policy
     {
-    private:
-        typedef components::default_distribution_policy base_type;
-
     public:
         container_distribution_policy()
           : num_partitions_(std::size_t(-1))
         {}
 
-        container_distribution_policy operator()(std::size_t num_partitions) const
+        container_distribution_policy operator()(
+            std::size_t num_partitions) const
         {
-            auto localities = this->base_type::ensure_localities();
-            return container_distribution_policy(num_partitions, *localities);
+            return container_distribution_policy(
+                num_partitions, get_localities());
         }
 
-        container_distribution_policy operator()(hpx::id_type const& locality) const
+        container_distribution_policy operator()(
+            hpx::id_type const& locality) const
         {
             return container_distribution_policy(locality);
         }
@@ -53,7 +53,10 @@ namespace hpx
             std::vector<id_type> const& localities) const
         {
             if (num_partitions_ != std::size_t(-1))
-                return container_distribution_policy(num_partitions_, localities);
+            {
+                return container_distribution_policy(
+                    num_partitions_, localities);
+            }
             return container_distribution_policy(localities.size(), localities);
         }
 
@@ -85,15 +88,25 @@ namespace hpx
         ///////////////////////////////////////////////////////////////////////
         std::size_t get_num_partitions() const
         {
-            auto localities = this->base_type::ensure_localities();
-            std::size_t num_parts = (num_partitions_ == std::size_t(-1)) ?
-                localities->size() : num_partitions_;
-            return (std::max)(num_parts, std::size_t(1));
+            if (localities_)
+            {
+                std::size_t num_parts = (num_partitions_ == std::size_t(-1)) ?
+                    localities_->size() : num_partitions_;
+                return (std::max)(num_parts, std::size_t(1));
+            }
+            return std::size_t(1);
         }
 
-        std::vector<hpx::id_type> const& get_localities() const
+        std::vector<hpx::id_type> get_localities() const
         {
-            return *this->base_type::ensure_localities();
+            if (!localities_)
+            {
+                // use this locality, if this object was default constructed
+                return std::vector<id_type>(1, hpx::find_here());
+            }
+
+            HPX_ASSERT(!localities_->empty());
+            return *localities_;
         }
 
     private:
@@ -110,7 +123,6 @@ namespace hpx
           : components::default_distribution_policy(localities),
             num_partitions_(num_partitions)
         {}
-
 
         container_distribution_policy(std::size_t num_partitions,
                 std::vector<id_type> && localities)
--- a/hpx/components/containers/container_distribution_policy.hpp
+++ b/hpx/components/containers/container_distribution_policy.hpp
@@ -1,5 +1,5 @@
 //  Copyright (c) 2014 Bibek Ghimire
-//  Copyright (c) 2014-2015 Hartmut Kaiser
+//  Copyright (c) 2014-2017 Hartmut Kaiser
 //
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
@@ -12,10 +12,12 @@
 #include <hpx/traits/is_distribution_policy.hpp>
 
 #include <hpx/runtime/serialization/serialize.hpp>
+#include <hpx/runtime/serialization/shared_ptr.hpp>
 #include <hpx/runtime/serialization/vector.hpp>
 
 #include <algorithm>
 #include <cstddef>
+#include <memory>
 #include <type_traits>
 #include <utility>
 #include <vector>
@@ -28,6 +30,9 @@ namespace hpx
     struct container_distribution_policy
       : components::default_distribution_policy
     {
+    private:
+        typedef components::default_distribution_policy base_type;
+
     public:
         container_distribution_policy()
           : num_partitions_(std::size_t(-1))
@@ -35,7 +40,8 @@ namespace hpx
 
         container_distribution_policy operator()(std::size_t num_partitions) const
         {
-            return container_distribution_policy(num_partitions, localities_);
+            auto localities = this->base_type::ensure_localities();
+            return container_distribution_policy(num_partitions, *localities);
         }
 
         container_distribution_policy operator()(hpx::id_type const& locality) const
@@ -79,14 +85,15 @@ namespace hpx
         ///////////////////////////////////////////////////////////////////////
         std::size_t get_num_partitions() const
         {
+            auto localities = this->base_type::ensure_localities();
             std::size_t num_parts = (num_partitions_ == std::size_t(-1)) ?
-                localities_.size() : num_partitions_;
+                localities->size() : num_partitions_;
             return (std::max)(num_parts, std::size_t(1));
         }
 
         std::vector<hpx::id_type> const& get_localities() const
         {
-            return localities_;
+            return *this->base_type::ensure_localities();
         }
 
     private:
--- a/hpx/runtime/components/default_distribution_policy.hpp
+++ b/hpx/runtime/components/default_distribution_policy.hpp
@@ -20,6 +20,7 @@
 #include <hpx/runtime/naming/name.hpp>
 #include <hpx/runtime/serialization/serialization_fwd.hpp>
 #include <hpx/runtime/serialization/vector.hpp>
+#include <hpx/runtime/serialization/shared_ptr.hpp>
 #include <hpx/traits/extract_action.hpp>
 #include <hpx/traits/is_distribution_policy.hpp>
 #include <hpx/traits/promise_local_result.hpp>
@@ -27,6 +28,7 @@
 
 #include <algorithm>
 #include <cstddef>
+#include <memory>
 #include <type_traits>
 #include <utility>
 #include <vector>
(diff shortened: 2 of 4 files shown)


sithhell:
Doesn't seem to work as intended. Have a look at this, for example: http://cdash.cscs.ch/testDetails.php?test=11380446&build=71293

hkaiser:
@sithhell I think all is well now. No idea what the 'errors 1' refers to, following the link to pycicle does not show the corresponding build.

hkaiser:
Here is the remaining test failure: http://cdash.cscs.ch/viewTest.php?onlyfailed&buildid=71394, it's unrelated to this PR.

hpx/components/containers/container_distribution_policy.hpp
```diff
@@ -28,14 +30,18 @@ namespace hpx
...
+
     public:
         container_distribution_policy()
           : num_partitions_(std::size_t(-1))
         {}
 
         container_distribution_policy operator()(std::size_t num_partitions) const
         {
-            return container_distribution_policy(num_partitions, localities_);
+            auto localities = this->base_type::ensure_localities();
```
sithhell: do we really need to assign this to a temporary and introduce reference counting here?

hpx/components/containers/container_distribution_policy.hpp
```diff
@@ -79,14 +85,15 @@ namespace hpx
         ///////////////////////////////////////////////////////////////////////
         std::size_t get_num_partitions() const
         {
+            auto localities = this->base_type::ensure_localities();
```
sithhell: same as above.

hpx/runtime/components/default_distribution_policy.hpp
```diff
@@ -105,7 +107,8 @@ namespace hpx { namespace components
         {
             using components::stub_base;
 
-            for (hpx::id_type const& loc: localities_)
+            auto localities = ensure_localities();
```
sithhell: taking a copy here seems to be a little overkill here.

hpx/runtime/components/default_distribution_policy.hpp
```diff
@@ -271,63 +274,85 @@ namespace hpx { namespace components
...
+            if (localities_->size() > 1 && loc == localities_->back())
+            {
                 return items - detail::round_to_multiple(items, locs, locs-1);
+            }
 
             // otherwise just distribute evenly
             return (items + locs - 1) / locs;
         }
+
+        std::shared_ptr<std::vector<id_type>> ensure_localities() const
```
sithhell: This ensure localities is used inconsistently throughout the patch. Why not just create a shared pointer with an empyt vector in the default ctor?

hpx/components/containers/container_distribution_policy.hpp
hkaiser: YEs we need to do that because of the asynchronous functionality here: https://github.com/STEllAR-GROUP/hpx/pull/3087/files#diff-61639ba7144f3b99ed04835f11dd7924R163.

hpx/runtime/components/default_distribution_policy.hpp
hkaiser: This does not take a copy of the vector, just a copy of the shared_ptr - no harm.

hpx/runtime/components/default_distribution_policy.hpp
hkaiser: What would that change? We would have to delay-fill it anyways

hpx/runtime/components/default_distribution_policy.hpp
sithhell: unnecessary reference count.

hpx/runtime/components/default_distribution_policy.hpp
sithhell: What changes is that we get rid of the checks if `localities_` is not pointing to a `nullptr`. Since we require at least a single element vector, this seems like the perfect fit for having in a default ctor, no?

hkaiser: We can't fill the vector with `hpx::find_here()` in the default constructor as this may cause HPX functionality to be invoked during startup. That means we would have to check whether the vector is empty anyways. I don't really see a benefit of what you suggest.

hpx/runtime/components/default_distribution_policy.hpp
hkaiser: Come on - really?

hpx/components/containers/container_distribution_policy.hpp
sithhell: Sure, this particular function however doesn't need it, it could directly pass on the `*localities_`

hpx/runtime/components/default_distribution_policy.hpp
sithhell: Ok. I was confused since we didn't need this before. There are various uses of `localities_` directly. There doesn't seem to be a consistent way of when to ensure that `localities_` exists or not. I think to resemble the old behavior, it would make sense to at least have a `std::make_shared<std::vector<hpx::id_type>()` in the default ctor. Bringing back the old functionality without introducing further complexity.
Or change the implementation such that an empty `shared_ptr` is equivalent to the previous behavior of an empty vector. Right now, the two ways of expressing the same thing lead to confusion.

hpx/runtime/components/default_distribution_policy.hpp
sithhell: Yeah, shared ownership isn't needed in this particular case, so why artificially introduce it here? Sure, it sounds like minor nitpicking... I just think we shouldn't artificially introduce unnecessary overheads when they are avoidable.

hpx/components/containers/container_distribution_policy.hpp
hkaiser: `localities_` might be empty (or even invalid), while using `ensure_localities()` ensures using a properly initialized vector of localities.

hpx/runtime/components/default_distribution_policy.hpp
hkaiser: I don't see how we could avoid this in this case. `ensure_localities()` ensures having an initialized vector of localities in a thread-safe manner.

hpx/runtime/components/default_distribution_policy.hpp
hkaiser: Could you elaborate what you mean by 'two ways of expressing the same thing', please? The use of the `localities_` vector is well defined: it's either pre-defined during construction or it's populated on demand whenever accessed. 

Yes, this adds more functionality than we've initially had: it now allows to adjust the policy to a changing set of available localities. Is that a bad thing?

sithhell: `localities_` essentially has three states:

1. `localities_ == nullptr`
2. `localities_->size() == 0`
3. `localities_->size() > 0`

1. and 2. are currently more or less equivalent. However, some places in the code use `ensure_localities()`, and others don't but check for `localities_ != nullptr`explicitly (for example here: https://github.com/STEllAR-GROUP/hpx/pull/3087/files#diff-61639ba7144f3b99ed04835f11dd7924R301). And other places use `localities_` without a `nullptr` check (for example here: https://github.com/STEllAR-GROUP/hpx/pull/3087/files#diff-61639ba7144f3b99ed04835f11dd7924R277).

I don't see how we are now able to have a changing set of localities. As before, we can create policies on demand with changing sets of localities.

hkaiser: Well, `localities_->size() == 0` is essentially an error condition. I'd be willing to rule that case out.

hkaiser: > and others don't but check for localities_ != nullptrexplicitly (for example here: https://github.com/STEllAR-GROUP/hpx/pull/3087/files#diff-61639ba7144f3b99ed04835f11dd7924R301)

Well, we could use `ensure_localities()` in this case as well. I can change that.

> And other places use localities_ without a nullptr check (for example here: https://github.com/STEllAR-GROUP/hpx/pull/3087/files#diff-61639ba7144f3b99ed04835f11dd7924R277).

Right, this is a bug, will fix.

sithhell: What I am suggesting is to keep the "empty vector" case and prefer it over the empty shared pointer. The shared pointer is needed for the asynchronous dataflow operation. The empty vector means "here". In all cases, a `shared_ptr<vector<id_type>>` is being constructed. If we are in the "default" constructed case, we'll create a new shared pointer and a new one element vector on each access. I don't think that's what we want.

hkaiser: I don't think this is a good idea. It is completely useless to pre-allocate the shared_ptr with an empty vector in the default-constructed case. The (empty) vector can't be used for anything. We would have to delay-fill that vector later on in order to make it useful. This however may turn out to be a thread-safety problem (would require a lock).  

OTOH, if we left the shared_ptr empty in the default-constructed case we would create a new (independent) shared_ptr on demand and fill it as needed (as suggested by this PR). No lock is needed.

hpx/components/containers/container_distribution_policy.hpp
```diff
@@ -79,14 +85,15 @@ namespace hpx
...
-                localities_.size() : num_partitions_;
+                localities->size() : num_partitions_;
             return (std::max)(num_parts, std::size_t(1));
         }
 
-        std::vector<hpx::id_type> const& get_localities() const
+        std::vector<hpx::id_type> get_localities() const
         {
-            return localities_;
+            return *this->base_type::ensure_localities();
```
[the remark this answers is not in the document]
hkaiser: Yes, this is intentional. I believe the old behavior was wrong.

hpx/runtime/components/default_distribution_policy.hpp
```diff
@@ -271,63 +274,96 @@ namespace hpx { namespace components
...
-                    localities_.end() ||
-                (localities_.empty() && loc == hpx::find_here())
+                std::find(localities->begin(), localities->end(), loc) !=
+                    localities->end()
             );
 
             // this distribution policy places an equal number of items onto
             // each locality
-            std::size_t locs = (std::max)(std::size_t(1), localities_.size());
+            std::size_t locs = (std::max)(std::size_t(1), localities->size());
```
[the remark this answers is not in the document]
hkaiser: Yes, this is guaranteed by the post condition of `ensure_localities()`. We can remove the `max` now.

hpx/runtime/components/default_distribution_policy.hpp
```diff
@@ -271,63 +274,96 @@ namespace hpx { namespace components
...
+        std::shared_ptr<std::vector<id_type>> ensure_localities() const
+        {
+            // use this locality, if this object was default constructed
+            auto localities = localities_;
+            if (!localities)
+            {
+                localities = std::make_shared<std::vector<id_type>>();
+                localities->push_back(hpx::find_here());
+            }
+            if (localities->empty())
```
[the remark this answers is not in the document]
hkaiser: It is empty in the case when the user constructs the policy from an empty vector. But this could be detected in the constructors instead.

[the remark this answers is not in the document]
hkaiser: > I am question if the vector needs to be lazily filled in the default constructed case. 

Isn't that what we are discussing for a while now? We _have_ to lazily fill the vector of localities in the default-constructed case, no?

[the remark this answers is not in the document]
hkaiser: > As far as a I can see, the empty vector case has been handled properly before.

Yes, this could be detected in the (non-default) constructors.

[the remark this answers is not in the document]
sithhell: I don't think, that's what I am trying to say all the time. The vector being empty and the vector holding one element that's `hpx::find_here()` should be equivalent. There shouldn't be a need to lazily construct and fill that vector in the first place (that's what the code was doing before). If I am not completely mistaken, the lazy construction of this one element vector solely serves the purpose to distinguish between the single ( == hpx::find_here()) case and when a vector of localities has been passed to the ctor. If this lazy filing would significantly simplify the code, the picture would be different, but the actual implementation stays more or less the same.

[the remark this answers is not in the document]
hkaiser: I'm not going to rewrite the whole thing in order to fix the problem at hand. Please feel free to create a separate PR proposing an alternative implementation of this policy.

PR #3089, opened 2018-01-01
Fix a strange thing in parallel::detail::handle_exception. (Fix #2834.)

### LINKED ISSUE #2834: One suspicion in parallel::detail::handle_exception.
I have one doubt about parallel::detail::handle_exception.
https://github.com/STEllAR-GROUP/hpx/blob/76f7c3bc8eac51d73be7de4146b729335ce03c3f/hpx/parallel/exception_list.hpp#L26-L74
Above, the first ```handle_exception_impl<>::call()``` in Line 31 just throws exception. But, the second and third one return an exceptional future. As I think, this is strange. It seems inconsistent.

I think that the struct of Line 79 is for parallel algorithm called with ```execution::task```. So, in the struct, ```call()``` all both should return ```hpx::future<Result>```. (In other words, should return exceptional future.)
But in case of Line 27, it is different case. So, it should just return ```Result``` as I think. (In other words, should just throw exception.)

How do you think about this issue?

### LINKED ISSUE #3088: A strange thing in parallel::sort.
https://github.com/STEllAR-GROUP/hpx/blob/f63f4b953917386f28fff31c0fde8ce61634247e/hpx/parallel/algorithms/sort.hpp#L210-L214

```result```(hpx::future) is returned from ```execution::async_execute()```. So, if the task is running, ```result.has_exception()``` always returns ```false```.

As I think, that code block(L210-L214) is not necessary because anyhow the exception which is occurred in the task is wrapped into the future(```result```).

### PR DESCRIPTION
**Pull request by taeguk**
Fixes #2834 .
flyby: Fixes #3088 .

### DIFF
--- a/hpx/parallel/algorithms/sort.hpp
+++ b/hpx/parallel/algorithms/sort.hpp
@@ -1,6 +1,7 @@
 //  Copyright (c) 2015 John Biddiscombe
 //  Copyright (c) 2015-2017 Hartmut Kaiser
 //  Copyright (c) 2015 Francisco Jose Tapia
+//  Copyright (c) 2018 Taeguk Kwon
 //
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
@@ -24,7 +25,6 @@
 #include <hpx/parallel/traits/projected.hpp>
 #include <hpx/parallel/util/compare_projected.hpp>
 #include <hpx/parallel/util/detail/algorithm_result.hpp>
-#include <hpx/parallel/util/detail/handle_local_exceptions.hpp>
 #include <hpx/parallel/util/projection_identity.hpp>
 
 #include <algorithm>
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by hkaiser** [comment]
@taeguk Thanks for looking into this! It's overdue that you get access to the repository - I've sent you an invite.

File: hpx/parallel/algorithms/sort.hpp [hunk 1]
```diff
@@ -201,19 +202,22 @@ namespace hpx { namespace parallel { inline namespace v1
                 result = execution::async_execute(policy.executor(),
                         &sort_thread<ExPolicy, RandomIt, Compare>,
                         std::ref(policy), first, last, comp);
+
+                return result;
```
hkaiser: Wouldn't a simple

    return execution::async_execute(policy.executor(),  
        &sort_thread<ExPolicy, RandomIt, Compare>,  
        std::ref(policy), first, last, comp);

do the trick here?

hpx/parallel/algorithms/sort.hpp
```diff
@@ -201,19 +202,22 @@ namespace hpx { namespace parallel { inline namespace v1
...
-            {
-                return detail::handle_exception<ExPolicy, RandomIt>::call(
-                    std::move(result));
+                    // Not reachable.
+                    HPX_ASSERT(false);
+                }
+                catch (...) {
+                    return hpx::make_exceptional_future<RandomIt>(
+                        std::current_exception());
+                }
```
hkaiser: Hmmm, do we really have to use nested try/catch blocks here? `handle_local_exceptions` only rethrows the current exception as an exception list. Couldn't we directly call `make_exceptional_future` with a exception list constructed from the current exception?

hpx/parallel/algorithms/sort.hpp
taeguk: @hkaiser You're right. That code was derived from just moving ```return result;``` into inside of try block. I'll fix that like you said.

hpx/parallel/exception_list.hpp
```diff
@@ -87,11 +81,11 @@ namespace hpx { namespace parallel { inline namespace v1
...
+                    catch(std::bad_alloc const&) {
+                        throw;
                     }
-                    catch (hpx::exception_list const& el) {
-                        throw el;
+                    catch (hpx::exception_list const&) {
+                        throw;
                     }
                     catch (...) {
                         throw hpx::exception_list(std::current_exception());
```
hkaiser: Hmm, this looks really strange (it probably was strange all along). I'm sure this can be done with one level of try/catch blocks.

hpx/parallel/algorithms/sort.hpp
taeguk: @hkaiser ```handle_local_exceptions()`` handles an exception differently when ```std::current_exception()``` means ```std::bad_alloc```. In that case, ```handle_local_exceptions()``` just rethrows ```std::bad_alloc``` without wrapping into ```exception_list```.
We need to handle the exception differently according to what is the exception which we handles. And for that, I used ```handle_local_exception()```.
And nested catch (Line216-219) is needed for wrapping an exception which is threw properly from ```handle_local_exception()```.

hkaiser: @taeguk yes, I understand. Still, there might be a better way.

hpx/parallel/exception_list.hpp
taeguk: @hkaiser Yes, we can do that with only one level of try/catch blocks.

hpx/parallel/algorithms/sort.hpp
taeguk: @hkaiser Aha, okay i got it.
https://github.com/taeguk/hpx/blob/9d72e4de54495987597bc2e6bbe609bc8125a01c/hpx/parallel/algorithms/sort.hpp#L249-L260
If we move that exception handling from Line208 to above lines, we can avoid nested try/catch blocks with just calling ```detail::handle_exception()```.

taeguk: For example, https://github.com/taeguk/hpx/blob/9d72e4de54495987597bc2e6bbe609bc8125a01c/hpx/parallel/algorithms/partition.hpp#L963-L975

PR #3090, opened 2018-01-01
Unify use of yield/yield_k

hkaiser:
- cleanup configuration of TLS and compiler specific fence operation

### DIFF
--- a/hpx/config/compiler_native_tls.hpp
+++ b/hpx/config/compiler_native_tls.hpp
@@ -1,0 +1,33 @@
+//  Copyright (c) 2018 Hartmut Kaiser
+//
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#if !defined(HPX_COMPILER_NATIVE_TLS)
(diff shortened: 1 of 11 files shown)


hpx/config/compiler_native_tls.hpp
```diff
@@ -0,0 +1,33 @@
...
+#if defined(__has_feature)
+#  if __has_feature(cxx_thread_local)
+#    define HPX_NATIVE_TLS thread_local
+#  endif
+#elif defined(HPX_HAVE_CXX11_THREAD_LOCAL)
+#  define HPX_NATIVE_TLS thread_local
+#endif
+
+#if !defined(HPX_NATIVE_TLS)
+#  if defined(_GLIBCXX_HAVE_TLS)
```
sithhell: There needs to be at least one C++ header include to have this define, maybe `#include <cstdint>`?

sithhell: Or even `<ciso646>`...

hkaiser: Fair point, will fix. Thanks!

hkaiser: Done.

PR #3093, opened 2018-01-06
Short-circuit all_of/any_of/none_of instantiations

K-ballo:


hpx/util/detail/pack.hpp
```diff
@@ -86,36 +115,33 @@ namespace hpx { namespace util { namespace detail
     template <bool ...Vs>
     struct any_of<pack_c<bool, Vs...> >
       : std::integral_constant<bool,
-            !all_of<pack_c<bool, !Vs...> >::value
+            !std::is_same<
+                pack_c<bool, Vs...>
+              , pack_c<bool, (Vs && false)...> // false...
```
sithhell: Is this variant really better than what we previously had? If I read the code correctly, the second pack would always expand to `pack_c<bool, false...>`. Where is the benefit with this approach? I don't see  any short circuiting happening here, what do I miss?

K-ballo: No, this variant is the same than what we previously had, `any_of` used to be implemented in terms of `all_of` and now it isn't anymore. These are the original interfaces, which later turned into specializations, and deal entirely with boolean constants: there are no instantiations to short-circuit. The base templates are the ones that now short-circuit instantiations.

That said, it's likely that those original interfaces are not being used anymore, and that it would be possible to remove them altogether.

PR #3094, opened 2018-01-08
Relax atomic operations on performance counter values

hkaiser:
This relaxes the atomic operations on performance counter values

### DIFF
--- a/hpx/util/get_and_reset_value.hpp
+++ b/hpx/util/get_and_reset_value.hpp
@@ -31,10 +31,9 @@ namespace hpx { namespace util
     template <typename T>
     inline T get_and_reset_value(std::atomic<T>& value, bool reset)
     {
-        T result = value.load();
         if (reset)
-            value.store(0);
-        return result;
+            return value.exchange(0, std::memory_order_acq_rel);
+        return value.load(std::memory_order_relaxed);
     }
 }}
 


hpx/util/get_and_reset_value.hpp
```diff
@@ -31,9 +31,9 @@ namespace hpx { namespace util
     template <typename T>
     inline T get_and_reset_value(std::atomic<T>& value, bool reset)
     {
-        T result = value.load();
+        T result = value.load(std::memory_order_acquire);
         if (reset)
-            value.store(0);
+            value.store(0, std::memory_order_release);
```
sithhell: would it make sense to replace the code with something like this:
```
if (reset)
    return value.exchange(0, std:.memory_order_acq_rel);
else
    return value.load(std::memory_order_relaxed);
```
?

hkaiser: I'm not sure, if using `value.load(std::memory_order_relaxed);` would be sufficient, but otherwise this could be done, yes.

sithhell: I'm not sure about using relaxed either. The idea was: when reading the value, we don't require any ordering since it shouldn't matter if we read an old value.

hkaiser: Makes sense. Will fix.

PR #3100, opened 2018-01-14
Fix #3068 (condition_variable deadlock)

### LINKED ISSUE #3068: hpx::lcos::condition_variable could be suspect to deadlocks
The current code for `hpx::lcos::condition_variable` (and the `_any` variant) might lead to deadlocks.
The deadlock situation might, for example, be encountered, when concurrently calling `wait`.
The problem is, that the outer lock (provided from the caller) is being tried to be acquired unconditionally with the inner lock (protecting the `lcos::detail::condition_variable`) is being held.

brjsp:
According o @K-ballo, condition_variable_any has even more problems, and condition_variable unnecessarily duplicates its functionality, but this should fix the deadlock bug described in #3068 by alternating the order of destructors.

### DIFF
--- a/hpx/lcos/local/condition_variable.hpp
+++ b/hpx/lcos/local/condition_variable.hpp
@@ -51,9 +51,12 @@ namespace hpx { namespace lcos { namespace local
         {
             HPX_ASSERT_OWNS_LOCK(lock);
             util::ignore_while_checking<std::unique_lock<mutex> > il(&lock);
-
             std::unique_lock<mutex_type> l(mtx_);
             util::unlock_guard<std::unique_lock<mutex> > unlock(lock);
+            //The following ensures that the inner lock will be unlocked
+            //before the outer to avoid deadlock (fixes issue #3608)
+            std::lock_guard<std::unique_lock<mutex_type> > unlock_next(
+                l, std::adopt_lock);
 
             cond_.wait(l, ec);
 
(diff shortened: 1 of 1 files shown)


hpx/util/unlock_guard.hpp
```diff
@@ -12,6 +12,8 @@
 ///////////////////////////////////////////////////////////////////////////////
 namespace hpx { namespace util
 {
+    struct already_unlocked_t {};
+    constexpr already_unlocked_t already_unlocked = already_unlocked_t{};
     ///////////////////////////////////////////////////////////////////////////
```
hkaiser: Is this type used anywhere? AFAICS, the newly added constructor `unlock_guard(Mutex& m, already_unlocked_t)` are not called from anywhere.

brjsp: Sorry. I at first misread the paper and thought we needed it. It is indeed not used anywhere.

hpx/lcos/local/condition_variable.hpp
```diff
@@ -194,6 +205,10 @@ namespace hpx { namespace lcos { namespace local
             util::ignore_all_while_checking ignore_lock;
             std::unique_lock<mutex_type> l(mtx_);
             util::unlock_guard<Lock> unlock(lock);
+            //The following ensures that the inner lock will be unlocked
+            //before the outer to avoid deadlock (fixes issue #3608)
+            std::lock_guard <std::unique_lock<mutex_type> > unlock_next
+            (l, std::adopt_lock);
```
hkaiser: Please use clang-format to adapt the formatting of the newly added lines to our coding guidelines.

PR #3101, opened 2018-01-15
Revert "Adapting stack overhead limit for gcc 4.9"

msimberg:
Reverts STEllAR-GROUP/hpx#3097


hkaiser:
Why was this reverted?

msimberg:
Sorry, should've added an explanation. It was reverted because it
actually broke most configurations except the one it fixed (see
buildbot). So until that is sorted out at least the old values only
break one configuration. A fix is still needed of course.

@sithhell Did you change something else when moving the values out to
another file, as the first commit looked like it shouldn't have broken
things this badly?

PR #3104, opened 2018-01-16
Local execution of direct actions is now actually performed directly

hkaiser:
## Proposed Changes

 - the test whether a direct action should actually be executed directly was checking for thread priority and thread stack-size. This patch removes those checks as they overly constrain direct execution.

### DIFF
--- a/hpx/lcos/detail/async_implementations.hpp
+++ b/hpx/lcos/detail/async_implementations.hpp
@@ -1,4 +1,4 @@
-//  Copyright (c) 2007-2017 Hartmut Kaiser
+//  Copyright (c) 2007-2018 Hartmut Kaiser
 //
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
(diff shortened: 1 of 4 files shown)


hpx/lcos/detail/async_implementations.hpp
```diff
@@ -155,16 +155,9 @@ namespace hpx { namespace detail
...
+    HPX_CONSTEXPR bool can_invoke_locally()
     {
-        return !traits::action_decorate_function<Action>::value &&
-            this_thread::get_priority() ==
-                static_cast<threads::thread_priority>(
-                    traits::action_priority<Action>::value) &&
-            this_thread::get_stack_size() ==
-                threads::get_stack_size(
-                    static_cast<threads::thread_stacksize>(
-                        traits::action_stacksize<Action>::value));
```
sithhell: I suggest to rather change it to
```
return !traits::action_decorate_function<Action>::value &&
    this_thread::get_stack_size() >=
        threads::get_stack_size(		
            static_cast<threads::thread_stacksize>(		
                traits::action_stacksize<Action>::value));
```
That way, we don't run into the problem if an action should run on a larger stack than what we have currently.
The priority change is correct though, as we intend to run the action immediately anyway.

sithhell: The problem with not checking the priority is that the following test fails then:
http://cdash.cscs.ch/testDetails.php?test=11627793&build=73031
For the stacksize, we get this failure:
http://cdash.cscs.ch/testDetails.php?test=11628196&build=73031

hkaiser: Checking the priority does not make any sense as the function will be executed right away. I'd suggest correcting the test. I agree on the stack size change you suggested, though. That should also make the test pass.

sithhell: The tests don't use direct actions. The actions are just executed locally.

hkaiser: Sure, I didn't say anything about direct actions, though. I still think that the tests are overly constrained.

hkaiser: @sithhell Your comments have been addressed.

PR #3106, opened 2018-01-17
Add cmake test for std::decay_t to fix cuda build

msimberg:
Fixes part of failing cuda compilation.

## Proposed Changes

`hpx::util::optional` uses `std::decay_t<T>` which is a C++14 feature. This adds a cmake test and a macro `HPX_DECAY_T` which uses `typename std::decay<T>::type` instead with C++11.

This specifically addresses [this](http://rostam.cct.lsu.edu/builders/hpx_cuda_8_0_gcc_5_boost_1_62_centos_x86_64_debug/builds/171/steps/build_examples/logs/stdio).

As an alternative, we could also wrap `type_traits` and add the definition for `decay_t`. Or just always use the old `decay<T>::type`.


hpx/util/optional.hpp
```diff
@@ -483,9 +485,9 @@ namespace hpx { namespace util
 
     ///////////////////////////////////////////////////////////////////////////
     template <typename T>
-    constexpr optional<std::decay_t<T>> make_optional(T && v)
+    constexpr optional<HPX_DECAY_T(T)> make_optional(T && v)
     {
-        return optional<std::decay_t<T>>(std::forward<T>(v));
+        return optional<HPX_DECAY_T(T)>(std::forward<T>(v));
     }
```
hkaiser: Wouldn't it have been simpler to just rewrite this code to use `std::decay`?

PR #3112, opened 2018-01-19
Minor fixes to exclusive_scan algorithm

hkaiser:
- removed unneeded temporary variable


hkaiser:
Looks like the change breaks the tests - will investigate...

hkaiser:
Closing as invalid as the change breaks overlapping input and output sequences.

PR #3114, opened 2018-01-19
Avoid using util::function for thread function wrappers

sithhell:
Currently, the non-plain versions of the register_work/register_thread functions
make use of util::function. This leads due unnecessary dynamic memory allocation
since the potential for util::function using the small object optimization is
limited.


hpx/runtime/threads/thread_helpers.hpp
```diff
@@ -770,15 +786,64 @@ namespace hpx { namespace applier
...
+
+        template <typename F>
+        struct thread_function_nullary
+        {
+            typename std::decay<F>::type f;
+
+            inline threads::thread_result_type operator()(threads::thread_arg_type)
+            {
+                // execute the actual thread function
+                f();
```
hkaiser: Shouldn't this be forwarded as well (same as https://github.com/STEllAR-GROUP/hpx/pull/3114/files#diff-f5e3956f6930a2fde1e87c0917f65d6bR799)?

hkaiser:
@sithhell how much of an improvement do you see from this change?

hpx/runtime/threads/thread_helpers.hpp
sithhell: The forward has no effect here. It should rather be removed on the other spot. Good catch.

sithhell:
> @sithhell how much of an improvement do you see from this change?

For a htts like scenario, with a granularity of 10 us, this is the speedup over current master:
```
cores, speedup over current master
1    , 1.00102564
2    , 1.0060325 
3    , 1.00424445
4    , 1.00872037
5    , 1.00511072
6    , 1.00440103
7    , 1.00665017
8    , 1.00742586
9    , 1.01019793
10   , 1.01536148
11   , 1.00443919
12   , 1.01483466
13   , 1.00874158
14   , 1.02881041
15   , 1.14408835
16   , 1.20076919
```

For all other tested grain sizes, there is no noticable difference.

PR #3115, opened 2018-01-19
Fixing race condition in channel test

hkaiser:
## Proposed Changes

  - fix a race in test channel_2916, the counter of pending tasks was relying on another thread being executed __after__ checking the value

- this was exposed by the recent patch fixing direct execution (#3104)


tests/regressions/lcos/CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ set(async_callback_non_deduced_context_PARAMETERS THREADS_PER_LOCALITY 4)
 set(broadcast_unwrap_future_2885_PARAMETERS LOCALITIES 2 THREADS_PER_LOCALITY 4)
 set(broadcast_wait_for_2822_PARAMETERS LOCALITIES 2 THREADS_PER_LOCALITY 4)
 set(channel_2916_FLAGS DEPENDENCIES iostreams_component)
-set(channel_2916_PARAMETERS LOCALITIES 2 THREADS_PER_LOCALITY 4)
+set(channel_2916_PARAMETERS THREADS_PER_LOCALITY 4)
```
sithhell: why did you remove distributed testing here?

hkaiser: Because the test itself does not do any distributed work.

sithhell: Ah, yes sure...

Issue #2980, opened 2017-11-02
Move background thread APEX timer


khuck:
Currently, the APEX timer around the background thread sits in the wrong location.  Every time that the background thread checks for work, it gets timed, which introduces unnecessary overhead to APEX.

The apex_wrapper() object should be moved inside the background thread code, so that only in cases where there is work to do, it is timed.  apex::stop should be called in all cases (not yield), so that we accurately track how many background tasks/parcels there were to process.

See https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/runtime/threads/detail/scheduling_loop.hpp#L356 for the code that should be moved.

victor-ludorum:
Hello @khuck I want to work on this issue . As you have explained that we have to move the apex timer code ( aprex_wrapper() object ) in the background thread . So , as per my understanding we have to move this code
` ` `
#if defined(HPX_HAVE_APEX)
                        // get the APEX data pointer, in case we are resuming the
                        // thread and have to restore any leaf timers from
                        // direct actions, etc.

                        // the address of tmp_data is getting stored
                        // by APEX during this call
                        util::apex_wrapper apex_profiler(
                            background_thread->get_description(),
                            background_thread->get_apex_data());

                        thrd_stat = (*background_thread)();

                        if (thrd_stat.get_previous() == terminated)
                        {
                            apex_profiler.stop();
                        }
                        else
                        {
                            apex_profiler.yield();
                        }
#else

` ` `
inside this **   if (HPX_UNLIKELY(background_thread))**  for proper functioning . 
suggest me if I am correct or not ?? or provide me some guidance for the work I have to do in it

TenGumis:
@victor-ludorum do you work on this issue?

victor-ludorum:
@TenGumis No, I have exam in my college so, I am not working on it. You may continue on it.

TenGumis:
@victor-ludorum So, I want to take this issue. Can somebody answer for question asked by @victor-ludorum above?

victor-ludorum:
@TenGumis Yeah!! you can work on it 😊 . All the best 👍

hkaiser:
@khuck would you care to respond?

TenGumis:
@khuck  ping

msimberg:
Closing because of #3245.

PR #3116, opened 2018-01-20
Moving background thread APEX timer #2980

TenGumis:
According to @khuck issue #2980 I tried move APEX timer.


hkaiser:
Could you elaborate what this patch is supposed to change, please? From what I can see this moves the measurement from wrapping a function call into the function itself.

hkaiser:
@TenGumis ping? Are you still interested in working on this?

TenGumis:
@hkaiser. A few days ago I have been looking in this code but I still don't know how to do that. If you have any suggestions I will try to do that. If someone else know how to do it I can leave this issue ;)

PR #3118, opened 2018-01-21
Adding performance_counter::reinit to allow for dynamically changing counter sets

hkaiser:
- This change adds a new API function for performance counters that instructs a
  counter to re-initialize the set of values it refers to
- This also adds a new `hpx::reinit_active_counters()` API that triggers a
  reinit for all active counters

We will need this functionality for the new performance counters to be created for Phylanx where the list of values to be returned from a performance counter may change dynamically based on the evaluated expressions.


hkaiser:
@parsa This is a functionality you will need to use for the new Phylanx performance counters. After compiling the expression trees an application now should call `hpx::reinit_active_counters()` that will trigger re-scanning of the existing primitive instances inside the counters.

PR #3119, opened 2018-01-21
Disable default executor compatibility with V1 executors

hkaiser:
- this is in preparation with the V1.1. release at which point the V1 executors
  should be disabled by default
- this also sets the default for HPX_HAVE_TRANSFORM_REDUCE_COMPATIBILITY to OFF


sithhell:
We could remove the v1 tests as well, can't we?

hkaiser:
> We could remove the v1 tests as well, can't we?

The only change for now is to disable the capabilities by default. I think we should remove the tests once we remove the functionality for V1.2.

PR #3120, opened 2018-01-22
Don't use boost::intrusive_ptr for thread_id_type

sithhell:
This change gets rid of using boost::intrusive_ptr to refcount thread_data. The
lifetime of thread_data is perfectly managed by the scheduling policy. That is,
once a task terminates, it can be deleted/reused.

refcounting requires atomic operations which may needlessly (in this case) slow down an application.

### DIFF
--- a/hpx/runtime/actions/transfer_base_action.hpp
+++ b/hpx/runtime/actions/transfer_base_action.hpp
@@ -16,6 +16,8 @@
 #include <hpx/runtime/actions/detail/invocation_count_registry.hpp>
 #include <hpx/runtime/components/pinned_ptr.hpp>
 #include <hpx/runtime/get_locality_id.hpp>
+#include <hpx/runtime/threads/thread_data_fwd.hpp>
+#include <hpx/runtime/threads/thread_id_type.hpp>
 #include <hpx/runtime/serialization/base_object.hpp>
 #include <hpx/runtime/serialization/input_archive.hpp>
 #include <hpx/runtime/serialization/output_archive.hpp>
@@ -176,7 +178,7 @@ namespace hpx { namespace actions
           : arguments_(std::forward<Ts>(vs)...),
 #if defined(HPX_HAVE_THREAD_PARENT_REFERENCE)
             parent_locality_(transfer_base_action::get_locality_id()),
-            parent_id_(reinterpret_cast<std::uint64_t>(threads::get_parent_id())),
+            parent_id_(threads::get_parent_id()),
             parent_phase_(threads::get_parent_phase()),
 #endif
             priority_(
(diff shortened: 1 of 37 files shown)


hpx/runtime/threads/coroutines/coroutine_fwd.hpp
```diff
@@ -34,11 +34,6 @@
 
 namespace hpx { namespace threads
 {
-    class HPX_EXPORT thread_data;
```
hkaiser: The `HPX_EXPORT` needs to go somewhere else, now, I think.

hpx/runtime/actions/transfer_base_action.hpp
```diff
@@ -176,7 +176,7 @@ namespace hpx { namespace actions
           : arguments_(std::forward<Ts>(vs)...),
```
hkaiser: This file now needs to `#include thread_id_type.hpp`

hpx/runtime/threads/coroutines/coroutine_fwd.hpp
sithhell: it's in thread_id_type.hpp now.

hkaiser: So do other files now.

hpx/runtime/threads/thread_id_type.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+namespace hpx { namespace threads
+{
+    class HPX_EXPORT thread_data;
+
+    struct thread_id_type
+    {
+        constexpr thread_id_type()
+          : thrd_(nullptr) {}
+        constexpr thread_id_type(thread_data* thrd)
```
hkaiser: We should make this constructor `explicit` to avoid unexpected conversions (we already have an implicit operator `thread_data*()` below).

hpx/runtime/threads/coroutines/coroutine_fwd.hpp
sithhell: You are saying that this PR should fix all includes to not rely on thread_id_type being implicitly defined by some other include?

hkaiser: Not necessarily. OTOH, we're trying to cleanup our header-include situation. So when introducing a new header file, perhaps it might make sense to introduce it properly... We could even add an inspect check enforcing this.

sithhell: Currently, the header exposing everything was thread_data_fwd.hpp, which still holds. Depending on how fine-grained we want to resolve those headers...
I'd do this in a separate PR though to not conflate those more or less unrelated changes with the header fixups.

hkaiser: ok

hkaiser:
Please fix the Windows and the CircleCI builds, then we can go ahead and merge this.

sithhell:
The build errors should be fixed now.

biddisco:
Is this branch now ready for merging - or is there more work to be done on it?

PR #3126, opened 2018-01-24
Cleaning up coroutine implementation

sithhell:
## Proposed Changes

 - Removing dynamic memory allocation
 - Removing intrusive ptr
 - Removing unused functions
 - Simplifying thread functions to be nullary
 - Using thread_data directly to get/set the state_ex value

This patch both simplifies and cleans up current code. This patch gives a rough speedup of about 1.5x for small grain sizes leading to heavy contention in the thread scheduler.

There is a potential breaking change for third party code using thread functions directly. I think this code is minimal and should be mostly under our control.

### DIFF
--- a/hpx/runtime/threads/thread_data.hpp
+++ b/hpx/runtime/threads/thread_data.hpp
@@ -241,7 +241,7 @@ namespace hpx { namespace threads
                 thread_state(new_state, state_ex, tag));
         }
 
-    private:
+    public:
         /// The set_state function changes the extended state of this
         /// thread instance.
         ///
@@ -269,7 +269,6 @@ namespace hpx { namespace threads
             }
         }
 
-    public:
         /// Return the id of the component this thread is running in
         naming::address_type get_component_id() const
         {
@@ -517,7 +516,7 @@ namespace hpx { namespace threads
         coroutine_type::result_type operator()()
         {
             HPX_ASSERT(this == coroutine_.get_thread_id().get());
-            return coroutine_(set_state_ex(wait_signaled));
+            return coroutine_();
         }
 
         thread_id_type get_thread_id() const
--- a/src/runtime/threads/coroutines/detail/coroutine_impl.cpp
+++ b/src/runtime/threads/coroutines/detail/coroutine_impl.cpp
@@ -34,10 +34,6 @@
 #include <hpx/runtime/threads/coroutines/detail/coroutine_self.hpp>
 #include <hpx/runtime/threads/thread_data_fwd.hpp>
 #include <hpx/util/assert.hpp>
-#include <hpx/util/reinitializable_static.hpp>
-
-#include <boost/exception/exception.hpp>
-#include <boost/lockfree/stack.hpp>
 
 #include <cstddef>
 #include <exception>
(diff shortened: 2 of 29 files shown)


hkaiser:
As said on IRC, I'm slightly worried that this PR removes the previously existing symmetry between the functionality provided by yielding a thread and terminating a thread. While yielding still does allow for specifying another thread-id to the scheduler, instructing it to run the given thread first, this is not possible for terminating threads anymore (i.e. terminating threads do not allow to specify a continuation thread anymore).

sithhell:
I'v put the thread function returns back in.

hpx/runtime/threads/thread_data.hpp
```diff
@@ -517,7 +516,7 @@ namespace hpx { namespace threads
         coroutine_type::result_type operator()()
         {
             HPX_ASSERT(this == coroutine_.get_thread_id().get());
-            return coroutine_(set_state_ex(wait_signaled));
+            return coroutine_();
```
hkaiser: Shouldn't `set_state_ex(wait_signaled)` still be called here (before the call to the coroutine)?

src/runtime/threads/coroutines/detail/coroutine_impl.cpp
```diff
@@ -73,173 +69,40 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail
...
-            }
-            catch (exit_exception const&) {
-                status = super_type::ctx_exited_exit;
-                tinfo = std::current_exception();
-                this->reset();            // reset functor
-            }
-            catch (boost::exception const&) {
-                status = super_type::ctx_exited_abnormally;
-                tinfo = std::current_exception();
                 this->reset();
```
hkaiser: I'd suggest adding an assert here making sure the returned `state == terminated`.

tests/performance/local/htts_v2/htts2_hpx.cpp
```diff
@@ -94,7 +92,7 @@ struct hpx_driver : htts2::driver
             using hpx::util::placeholders::_1;
```
hkaiser: The `using` declaration could be removed now.

hpx/runtime/threads/thread_data.hpp
sithhell: There are currently 3 state transitions:
 - (initialized) -> wait_signaled (not a state transition): a new thread is always put into the wait_signaled state
 - wait_signaled -> wait_abort (to cancel a thread): No need to set it back to wait_signaled since the task will terminate
 - wait_timeout -> wait_signaled (when a thread was woken up after a timeout): This is correctly handled here: https://github.com/STEllAR-GROUP/hpx/pull/3126/files#diff-1495f49124e718a7579611b3ba828883L325
 
I don't see a reason to reset the state_ex at each coroutine invocation.

hkaiser: Ok, makes sense. Thanks for the explanations.

src/runtime/threads/coroutines/detail/coroutine_impl.cpp
sithhell: done.

msimberg:
@hkaiser Thanks for the heads up. I'll go ahead and merge this since this would be a very nice improvement to have in the release, but I'll revert if there are any signs of trouble. Please don't merge other PRs while this is being tested on buildbot, thanks.

PR #3127, opened 2018-01-25
Add `hpx::start` nullptr overloads

msimberg:
## Proposed Changes

Adds overloads so that passing `nullptr` to `hpx::start` will only start up the runtime.

This also adds a missing line to the docstrings of some `hpx::start` overloads.

### DIFF
--- a/hpx/hpx_start.hpp
+++ b/hpx/hpx_start.hpp
@@ -734,125 +734,18 @@ namespace hpx
         int argc, char** argv, std::vector<std::string> const& cfg,
         hpx::runtime_mode mode = hpx::runtime_mode_default);
 
-    /// \brief Main non-blocking entry point for launching the HPX runtime system.
-    ///
-    /// This is a simplified main, non-blocking entry point, which can be used
-    /// to set up the runtime for an HPX application without a main function
-    /// (the runtime system will be set up in console mode or worker mode
-    /// depending on the command line settings). It will return immediately
-    /// after that. Use `hpx::wait` and `hpx::stop` to synchronize with the
-    /// runtime system's execution.
-    ///
-    /// \param f            [in] The function to be scheduled as an HPX
-    ///                     thread. Usually this function represents the main
-    ///                     entry point of any HPX application.
-    /// \param app_name     [in] The name of the application.
-    /// \param argc         [in] The number of command line arguments passed
-    ///                     in \p argv. This is usually the unchanged value as
-    ///                     passed by the operating system (to `main()`).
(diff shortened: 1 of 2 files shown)


msimberg:
Added overloads to `hpx::init` as well. Thanks for pointing that out.

hpx/hpx_init.hpp
```diff
@@ -705,6 +707,18 @@ namespace hpx
     inline int init(util::function_nonser<int(int, char**)> const& f,
         int argc, char** argv, std::vector<std::string> const& cfg,
         hpx::runtime_mode mode = hpx::runtime_mode_default);
+
+/// \cond NOINTERNAL
```
hkaiser: What's your rationale not to document the new overloads? Do you intend to deliberately leave them 'undocumented' (in the sense that the user is not supposed to use them)?

msimberg:
I did not mean for users not to use them. My thinking was that the
nullptr overloads are just an implementation detail to get the correct
behaviour for existing overloads, but in that case I should probably
document the behaviour in the affected overloads.

As a user, would you prefer to see the nullptr overloads explicitly
documented with separate entries or to have the nullptr behaviour
documented in the existing entries?

On Fri, Jan 26, 2018, at 13:07, Hartmut Kaiser wrote:
> *@hkaiser* commented on this pull request.


> 
> In hpx/hpx_init.hpp[1]:


> > @@ -705,6 +707,18 @@ namespace hpx inline int
> > init(util::function_nonser<int(int, char**)> const& f, int argc,
> > char** argv, std::vector<std::string> const& cfg, hpx::runtime_mode
> > mode = hpx::runtime_mode_default); + +/// \cond NOINTERNAL
>> What's your rationale not to document the new overloads? Do you intend
> to deliberately leave them 'undocumented' (in the sense that the user
> is not supposed to use them)?> — You are receiving this because you authored the thread. Reply to
> this email directly, view it on GitHub[2], or mute the thread[3].> 


Links:

  1. https://github.com/STEllAR-GROUP/hpx/pull/3127#discussion_r164095671
  2. https://github.com/STEllAR-GROUP/hpx/pull/3127#pullrequestreview-91818090

hkaiser:
> As a user, would you prefer to see the nullptr overloads explicitly documented with separate entries or to have the nullptr behaviour documented in the existing entries?

Having it documented as a special case with the other overloads would be preferred, I think. Something like: if `f` is `nullptr` no hpx_main will be invoked.

msimberg:
I also added a further note about `hpx_main` not being called when using the `start/init(function, ...)` overloads.

Issue #3003, opened 2017-11-13
Suspension of runtime


msimberg:
This is to collect feedback/wishes on suspending the runtime (putting threads to sleep rather than shutting down the whole runtime).

The main use case is situations where HPX cannot be used for the whole application (e.g. replacing OpenMP in an application that already uses MPI & OpenMP). This makes it easier for users to start integrating HPX piecewise into existing applications.

Currently this only considers running and suspending HPX on a single node (as communication between nodes would be handled by e.g. MPI in the cases where this is useful). As this is a big issue it will be split into separate PRs. What needs to be done is:

- [x] suspend single threads (#3046)
- [x] suspend full pools (#3080)
- [x] suspend the full runtime (#3129)
- [x] documentation (#3152)

# Suspending single threads

This can be done using condition variables. What could be useful is a mode for selecting how to remove/suspend a pu: stop scheduler immediately, wait for work queues to be empty, and wait for thread map and queues to be empty. The current `remove_processing_unit` would normally use the "wait for thread map and queues to be empty" mode at shutdown, and the to-be-added `suspend_processing_unit` would normally use the "wait for work queues to be empty" mode.

*Note*: it's not clear to me if it's best to let each pu itself (i.e. `scheduling_loop`) handle when it's okay to stop the loop (like waiting for thread map to be empty), or if this should be handled externally (like waiting for thread count to be `1 + background_thread_count`), or both. The latter is likely a must, but the former I'm not sure about. This applies to suspending pools and the runtime as well.

# Suspending pools

This seems to be mostly wrapping appropriate/already existing calls (is it?) to get something like `suspend_pool("pool-name", suspend_mode)`. The notes about different modes of suspending apply here as well.

# Suspending the runtime

This can most likely reuse the mechanism for waiting for `hpx::finalize`, but rather than stopping and cleaning everything up it would tell the thread manager to suspend all pools and threads. Again, different modes of suspending apply here.

Deciding on the API is not critical right now but it will need some time to brew in people's minds, so to get started below are a few thoughts.

If the current `init/start/stop/finalize` API should not be broken, usage could look something like this:
```
// Start runtime without a function to run
hpx::start(argc, argv);

...

// Run hpx_function_1 on runtime, start runtime if needed
// Suspend when done
hpx::init(hpx_function_1, mode_suspend);

// Run hpx_function_2 on runtime, don't block
hpx::start(hpx_function_2);

// Wait and suspend when done
[...]

PR #3129, opened 2018-01-25
Suspend runtime

msimberg:
Part of #3003.
## Proposed Changes

- Add `hpx::suspend` and `hpx::resume` functions
- `hpx::suspend` blocks waiting for all pools to finish their work
- `hpx::resume` blocks waiting for pools to resume
- Add a `wait` function to `io_service_pool` which waits for the `io_service`s to finish their work. As `io_services` stay idle when they have no work, `hpx::suspend` only waits for `io_service_pool`s to be empty.

- This uses #3110 and #3127. They should be merged first.

The API is now a lot simpler that initially suggested in #3003, to not complicate the use of `hpx::start/init/stop` further. This also makes the implementation simpler. If a "smarter" API is needed in the future it can be built on top of this, but this is enough as a start.

The `resume/suspend` in this PR is faster than `start/stop`, but not significantly. I will still need to optimize that part to make this useful for small blocks of work, but will do that in a separate PR.


hkaiser:
This PR is a high risk change as it touches on the scheduling loop. We have to be very careful with merging it.

PR #3130, opened 2018-01-29
Avoid completion handler allocation

sithhell:
## Proposed Changes

This patch removes `compose_cb` in favor of `boost::container::small_vector<callback_type, 3>`. The main motivation is to avoid memory allocations for cases where two or more completion handlers are attached to a future.

Flyby: avoid reference counting for pack traversals.

### DIFF
--- a/hpx/lcos/detail/future_data.hpp
+++ b/hpx/lcos/detail/future_data.hpp
@@ -28,6 +28,7 @@
 #include <hpx/util/unused.hpp>
 
 #include <boost/intrusive_ptr.hpp>
+#include <boost/container/small_vector.hpp>
 
 #include <chrono>
 #include <cstddef>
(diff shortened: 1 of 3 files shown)


hpx/lcos/detail/future_data.hpp
```diff
@@ -194,75 +196,6 @@ namespace detail
         std::size_t& count_;
```
hkaiser: I think the `handle_continuation_recursion_count` is not needed anymore.

hkaiser: I take that back, sorry.

hkaiser: But it can be moved to the source file now (out of the header) ...

sithhell: good idea.

PR #3131, opened 2018-01-30
Fixing #2325

### LINKED ISSUE #2325: scan_partitioner is not working as expected.
Parallel algorithms using the scan partitioner are spuriously failing when
the continuation in dataflow is executed asynchronously.

It seems that when passing hpx::launch::sync to the dataflow invocation works around that issue.

A fix needs to be found such that it works with (async) executors again.

sithhell:
Fixes #2325

## Proposed Changes

As time went by, we can re-enable the usage of executors again. Some changes
in our code base seem to have led to this not segfaulting anymore

### DIFF
--- a/hpx/lcos/dataflow.hpp
+++ b/hpx/lcos/dataflow.hpp
@@ -57,7 +57,7 @@ namespace hpx { namespace lcos { namespace detail
 #if defined(HPX_HAVE_CXX14_RETURN_TYPE_DEDUCTION)
         static auto error(F f, Args args)
         {
-            hpx::util::invoke_fused(std::move(f), std::move(args));
+            return hpx::util::invoke_fused(std::move(f), std::move(args));
         }
 #else
         static auto error(F f, Args args)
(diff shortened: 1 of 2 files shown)


taeguk:
@sithhell Why are there still ```hpx::launch::sync```?
https://github.com/STEllAR-GROUP/hpx/blob/18ab2432b707c6a03a89cea87903ad0c3a289be1/hpx/parallel/util/scan_partitioner.hpp#L137-L138

taeguk:
https://github.com/STEllAR-GROUP/hpx/blob/3b0ecd76fe7b878197ee0243ddba6f524d371897/hpx/parallel/util/scan_partitioner.hpp#L128-L144
```c++
if (parts & 0x7) 
    p = hpx::launch::sync;
````
In past days, there was the code in ```scan_partitioner```.
This is not applied into this PR.

hkaiser:
@taeguk I think this was an attempt to avoid spawning asynchronous tasks for every partition, the tasks spawned tend to be very small.

Issue #2996, opened 2017-11-09
Build fails with HPX_WITH_HWLOC=OFF


msimberg:
```
> make hello_world_exe
[  0%] Building CXX object src/CMakeFiles/hpx.dir/runtime.cpp.o
In file included from /home/simbergm/src/hpx/hpx/runtime/serialization/detail/polymorphic_nonintrusive_factory.hpp:23:0,
                 from /home/simbergm/src/hpx/hpx/lcos/future.hpp:15,
                 from /home/simbergm/src/hpx/hpx/performance_counters/counters.hpp:10,
                 from /home/simbergm/src/hpx/hpx/performance_counters/counter_creators.hpp:11,
                 from /home/simbergm/src/hpx/src/runtime.cpp:10:
/home/simbergm/src/hpx/hpx/util/static.hpp: In instantiation of ‘static T& hpx::util::static_<T, Tag>::get_reference() [with T = hpx::threads::noop_topology; Tag = hpx::threads::noop_topology::noop_topology_tag; hpx::util::static_<T, Tag>::reference = hpx::threads::noop_topology&]’:
/home/simbergm/src/hpx/hpx/util/static.hpp:50:26:   required from ‘hpx::util::static_<T, Tag>::static_() [with T = hpx::threads::noop_topology; Tag = hpx::threads::noop_topology::noop_topology_tag]’
/home/simbergm/src/hpx/hpx/runtime/threads/policies/noop_topology.hpp:243:68:   required from here
/home/simbergm/src/hpx/hpx/util/static.hpp:76:22: error: cannot declare variable ‘t’ to be of abstract type ‘hpx::threads::noop_topology’
             static T t;
                      ^
In file included from /home/simbergm/src/hpx/hpx/runtime/threads/policies/topology.hpp:12:0,
                 from /home/simbergm/src/hpx/src/runtime.cpp:25:
/home/simbergm/src/hpx/hpx/runtime/threads/policies/noop_topology.hpp:32:8: note:   because the following virtual functions are pure within ‘hpx::threads::noop_topology’:
 struct noop_topology : topology
        ^
In file included from /home/simbergm/src/hpx/hpx/runtime/threads/thread_executor.hpp:14:0,
                 from /home/simbergm/src/hpx/hpx/lcos/detail/future_data.hpp:16,
                 from /home/simbergm/src/hpx/hpx/lcos/future.hpp:12,
[...]

PR #3009, opened 2017-11-15
Fixing 2996

biddisco:
fix #2996

PR #3133, opened 2018-01-30
Make hwloc compulsory

### LINKED ISSUE #2977: Make hwloc compulsory instead of optional
As suggested in #2923 - we should make hwloc a requirement rather than an optional dependency.

Things to do

- Remove abstract topology class and virtual interface to hwloc_topology
- upgrade to hwloc 2.0 API (Due for release end Nov 2017)
- Improve CMake integration to allow hwloc 2.0 to be downloaded and built as part of the hpx compilation step if it is not found on the system. This may not be possible on _all_ platforms, but should be done where supported/possible.

msimberg:
Fixes #2977.

## Proposed Changes

- Replace abstract topology class with the hwloc topology class (I have only moved over the functions, not done any other structural changes)
- Update documentation with some basic instructions on getting hwloc
- Remove cmake options for disabling hwloc
- Remove some unused functions definitions/declarations

hwloc was already required in practice (for performance reasons, and because the build was broken with hwloc disabled), so this just makes it officially required.

@biddisco could you have a look to see if I might've missed some details, please? Do we need to update the recommended/minimum version numbers (current recommended is 1.10, minimum is 1.5, latest is 1.11)?


docs/manual/build_system/prerequisites.qbk
```diff
@@ -193,6 +192,29 @@ detail in the sections __unix_installation__ and __windows_installation__.
...
+
+where: `<where to install hwloc>` is the directory the built binaries will be
+installed to, and `<N>` is the number of cores to use to build hwloc.
+
+After the above sequence of commands has been executed you will need to specify
+the directory where Hwloc was installed as `HWLOC_ROOT` (`<where to install
+hwloc>`) while executing cmake for __hpx__ as explained in detail in the
+sections __unix_installation__ and __windows_installation__.
+
+[endsect]
```
hkaiser: Could you please add a note that the instructions don't refer to Windows based installations. I'm not sure about MacOS and/or Android based systems (@sithhell?). For Windows the user can simply download the 64Bit binaries from the download page.

msimberg: I've added a note. Could you check that it's accurate (I wrote a one-sentence explanation for how to install hwloc on Windows)?

sithhell: I suggest to link to the hwloc documentation for further information:
https://www.open-mpi.org/projects/hwloc/doc/v1.11.9/
(There's no direct link to the installation section)

From a cursory look, it seems that getting hwloc to run on android is possible (but there's no real documentation), for macosx, I guess that autotools just works? Or homebrew or something similar.

msimberg: @sithhell I've added the hwloc link as a note at the end of the installation section.

docs/hpx.qbk
```diff
@@ -90,7 +90,8 @@
 [def __boost_atomic__           [@http://www.boost.org/doc/libs/release/doc/html/atomic.html Boost.Atomic]]
 [def __boost_serialization__    [@http://www.boost.org/doc/libs/release/libs/serialization/doc/index.html Boost.Serialization]]
 
-[def __hwloc_downloads__        [@http://www.open-mpi.org/projects/hwloc/]]
+[def __hwloc_downloads__        [@http://www.open-mpi.org/software/hwloc/v1.11 Hwloc Downloads]]
+[def __hwloc_doc__              [@http://www.open-mpi.org/projects/hwloc/doc/v1.11.9 Hwloc Documentation]]
```
hkaiser: Should we link to https://www.open-mpi.org/projects/hwloc/doc/ instead (no version)?

msimberg: Good point, less updating for us. What about the download link? There doesn't seem to a page with all the downloads like for the docs. I think linking to the 1.11 download page is nicer in this case.

hkaiser: Indeed.

hkaiser:
This also fixes #2996 and #3009.

hkaiser:
@msimberg This patch generates a ton of MSVC specific warnings complaining about dllexport conflicts (see https://ci.appveyor.com/project/hkaiser/hpx/build/master-1819/job/vu2x573ux94f50y3). Those usually can be safely ignored. Could you please wrap the definition of the new topology type using
```
// your type goes here
#include <hpx/config/warnings_suffix.hpp>
```
(see for instance here how it's done: https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/runtime/threads/threadmanager.hpp#L41 and https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/runtime/threads/threadmanager.hpp#L428, those have to be as close to the type itself as possible).

These includes will disable the warnings as needed.

msimberg:
@hkaiser I've added the includes and appveyor seems happy now. Thanks for pointing that out!

hkaiser:
@msimberg Thanks!

PR #3137, opened 2018-02-02
Suspend speedup

msimberg:
This speeds up #3129.

## Proposed Changes

- Adds two member functions, `suspend_direct` and `resume_direct` to `thread_pool_base` which don't spawn a new thread for suspending/resuming but instead call `suspend/resume_internal` directly.
- Use `suspend_direct` and `resume_direct` in `threadmanager::suspend/resume` to speed up suspend/resume.
- First notify all threads that they should suspend/resume before doing the blocking calls to `suspend/resume_processing_unit_internal` which wait for the threads to finish suspending/resuming.

I will post some more detailed benchmarks later, but rough performance numbers on dual-socket 18-core Xeon machine are (depends of course on the number of threads used):
- HPX start/stop: 50 ms
- HPX suspend/resume: 500 µs
- OpenMP enter and exit a `#pragma parallel for` region: 100 µs


hkaiser:
Nice! Thanks!

src/runtime_impl.cpp
```diff
@@ -528,6 +530,88 @@ namespace hpx {
         cond.notify_all();                  // we're done now
     }
 
+    int runtime_impl::suspend()
+    {
+        std::uint32_t initial_num_localities = get_initial_num_localities();
+        if (initial_num_localities > 1)
```
hkaiser: What is the rationale of this limitation? Wouldn't it be perfectly ok to suspend the runtime on one of the nodes at a time?

tests/unit/resource/suspend_runtime.cpp
```diff
@@ -0,0 +1,59 @@
...
+    };
+
+    hpx::start(nullptr, argc, argv, cfg);
+
+    // Wait for runtime to start
+    hpx::runtime* rt = hpx::get_runtime_ptr();
+    hpx::util::detail::yield_while([rt]()
+        {
+            return rt->get_state() < hpx::state_running;
+        }, "");
```
hkaiser: This tells me that a) `yield_while` shouldn't be in detail namespace and b) that its second argument should have a default value (or it has to be exposed through a different API).

src/runtime_impl.cpp
msimberg: Simply that I haven't tested it in those situations, and I did not think it would be useful. But you're right, it *should* be okay in those cases as well. Do you see a use case for it?

tests/unit/resource/suspend_runtime.cpp
msimberg: Fair point. Would it be okay if I move only `yield_while` into `hpx/util/yield_while.hpp` and leave `yield_k` where it is? I see `yield_k` as being a lower level helper function.

hkaiser: Sounds good to me. Could be done in a separate PR, though.

src/runtime_impl.cpp
hkaiser: Providing a migration path for distributed MPI+X applications.

hkaiser: This does not have to be part of this PR, however.

tests/unit/resource/suspend_runtime.cpp
msimberg: Ok, good. I'll do this in a separate PR as soon as this is merged.

src/runtime_impl.cpp
msimberg: Ok, I'd like to keep it separate in that case so that it can go in with appropriate tests/examples showing how it would be used in that case (and to see if there are any corner cases to be aware of).

msimberg:
For the record, this is what the timings look like. This is with the `balanced` thread binding, so one can see bumps at 19 and 37 threads, going to another numa domain and to hyperthreading, respectively.

The openmp benchmark is a `#pragma parallel for` doing a simple addition to avoid it being optimized away. For fairness HPX spawns a number of empty tasks equal to the number of threads before suspending. Plain resume/suspend without any tasks running is at least a factor of two faster than in the graph.

runtime-suspension

There is most likely some room to improve further but this is already good enough for blocks of work of at least 0.1 s.

hkaiser:
@msimberg very nice results! Would you mind adding the benchmark itself to the `tests/performance/local` folder in the repo?

msimberg:
Don't mind at all, will add them soon.

PR #3141, opened 2018-02-05
Fixing small_vector compatibility with older boost version

sithhell:
This fixes problems with swapping boost::container::small_vector on older boost
versions. In addition, on boost < 1.58, it uses std::vector since
boost::container::small_vector was introduced with boost 1.58


hkaiser:
Good thinking!

PR #3143, opened 2018-02-05
Fix stack overhead

msimberg:
## Proposed Changes

- Separate debug definitions into `hpx/config/debug.hpp`
- Include `debug.hpp` in the new `threads_stack.hpp` config file

This is the same as #3097, but makes sure `HPX_DEBUG` is defined for the thread stacksize defines.


hkaiser:
> The original fix also factored out into a separate file the stack size definitions. Should we still do that before merging?

Ohh, I'm blind, please disregard what I said. This LGTM!

Issue #3077, opened 2017-12-15
Remove thread_pool_executors


msimberg:
The `thread_pool_executor`s are currently broken (and have been for a while - the `thread_pool_executors` test hangs occasionally) and serve no real purpose now that one can suspend threads. It does not seem worth the effort to fix them.

hkaiser:
I agree. We should provide alternatives for this when removing them, however. In the end it boils down to providing proper documentation explaining how to replace them.

biddisco:
As per #2997  - if soome can tell me what these executors are for, I will have a go at doing some replacement/cleanup.

PR #3149, opened 2018-02-07
Remove thread_pool_attached_executors as they are not used anywhere

biddisco:
This is partly for #3077
The thread_pool_attached_executor is not used in any code and has no obvious purpose. The pool_executor that works with the thread pools created by the resource manager supercedes it I believe.


biddisco:
Note that I'm expecting some push back for this PR and I partly wanted to test if everything builds ok with this removed.

sithhell:
The only place where it is used in `hpx::compute::block_executor`. One of the features this executor provides is the ability to schedule tasks to a given set of cores, while still allowing for the possibility of stealing. So my suggestion would be, instead of removing it, allow it to attach to a given thread pool configured using the resource partitioner.

biddisco:
All one needs to do is create a pool with the cores you want to use and use a pool executor.

sithhell:
Pools can't be created dynamically, and there is no stealing between pools.

biddisco:
Dynamic pools will be supported in future. Feel free to create an issue requesting this.

The executor does not control stealing, it only tells the tasks which cores to run on - stealing is the job of the scheduler and this belongs to the pool.

msimberg:
@biddisco do you think we can close this? I think the `thread_pool_attached_executor`s could be replaced by something better in the future, but at the moment they don't have a direct replacement and don't hurt anyone (or have they caused headaches for you while working on your executors?). In the end they're similar to a NUMA executor in that they let you schedule tasks on a subset of resources of a thread pool, without actually creating an actual thread pool.

msimberg:
I'm going to close this for now. We can always reopen once we have a proper plan for what to do with all the executors.

PR #3152, opened 2018-02-07
Documentation for runtime suspension

msimberg:
Completes #3003. 

## Proposed Changes

- Add a section to the manual on how to suspend and resume the runtime, and a short motivation on why one might want to do this
- Move `thread_pool_base` out of the detail namespace as it's part of the public API
- Document member functions in `thread_pool_base` related to suspension
- Update docs `CMakeLists.txt` and `hpx.idx` to include new documentation

Note: all non-suspension related functions in `thread_pool_base` are still hidden in the documentation as I'm not sure how many of them should/would be used by a normal user. Most of them will need to be documented eventually though, but that's independent of this PR.


docs/manual/applications.qbk
```diff
@@ -187,4 +187,50 @@ available, such as for instance to provide your own entry point function instead
 of [funcref hpx_main `hpx_main`]. Please refer to the function documentation for
 more details (see: [headerref hpx/hpx_start.hpp `hpx/hpx_start.hpp`]).
 
+[heading:flexible Suspending and resuming the __hpx__ runtime]
+
+In some applications it is required to combine __hpx__ with other runtimes. To
+support this use case __hpx__ provides two functions: [funcref hpx::suspend
+`hpx::suspend`] and [funcref hpx::resume `hpx::resume`]. [funcref hpx::resume
```
sithhell: This looks like a typo. Did you mean to write `[funcref hpx::suspend` at the end of this line?

msimberg: Line 193? Yes, that's intentional. Maybe the linebreak looks odd in the middle of the funcref thing but boostbook handles that well.

sithhell: No, I'm pretty sure you meant suspend and not resume at the end of this line. Suspend blocks, resume doesn't.

sithhell: Essentially, what's written is this:

`[funcref hpx::resume `hpx::suspend`] is a blocking call which will wait for all scheduled tasks to finish executing and then put the thread pool OS threads to sleep.`

PR #3154, opened 2018-02-08
More refactorings based on clang-tidy reports

hkaiser:


sithhell:
Which clang-tidy checks did you run?

hkaiser:
@sithhell I don't have an explicit list. I enabled some of the tests interactively that looked interesting to me ;)

PR #3156, opened 2018-02-08
Fixing serialization of classes with incompatible serialize signature

### LINKED ISSUE #3134: HPX_SERIALIZATION_SPLIT_FREE does not stop compiler from looking for serialize() method
I want to enable HPX serialization of a library type, whose definition I cannot change. For this purpose I define a pair of free functions `save()`/`load()` and make use of `HPX_SERIALIZATION_SPLIT_FREE`. Unfortunately, the type in question has a method called `serialize()`, which has nothing to do with HPX and it apparently gets into the way of name lookup.

```c++
#include <hpx/runtime/serialization/serialize.hpp>

struct has_serialize {
  int i;
  double d;

  void serialize() {
    // Does something unrelated to HPX serialization ...
  }
};

template <typename Archive>
void save(Archive &ar, has_serialize const &hs, const unsigned int) {
  std::cout << "saving" << std::endl;
}

template <typename Archive>
void load(Archive &ar, has_serialize &hs, const unsigned int) {
  std::cout << "loading" << std::endl;
}

HPX_SERIALIZATION_SPLIT_FREE(has_serialize);

int main() {

  has_serialize in, out;

  std::vector<char> buf;

  // Serialize
  hpx::serialization::output_archive oa(buf, hpx::serialization::disable_data_chunking);
  oa << in;

  // Deserialize
  hpx::serialization::input_archive ia(buf);
  ia >> out;

  return 0;
}
```
```
include/hpx/runtime/serialization/access.hpp:88:21: error: no matching function for call to ‘has_serialize::serialize(hpx::serialization::input_archive&, int)’
                     const_cast<typename util::decay<T>::type&>(
                     ^~~~~~~~~~
hpx_bug.cpp:7:8: note: candidate: void has_serialize::serialize()
   void serialize() {
        ^~~~~~~~~
hpx_bug.cpp:7:8: note:   candidate expects 0 arguments, 2 provided
```
This example compiles if I comment out `has_serialize::serialize()`.

`HPX_SERIALIZATION_SPLIT_FREE_TEMPLATE` seems to be affected too.

It's worth noting that Boost.Serialization does not suffer from this issue.

AntonBikineev:
This fixes #3134

### DIFF
--- a/hpx/runtime/serialization/access.hpp
+++ b/hpx/runtime/serialization/access.hpp
@@ -8,19 +8,17 @@
 #define HPX_SERIALIZATION_ACCESS_HPP
 
 #include <hpx/runtime/serialization/serialization_fwd.hpp>
-#include <hpx/traits/has_member_xxx.hpp>
 #include <hpx/traits/polymorphic_traits.hpp>
 #include <hpx/util/decay.hpp>
 
 #include <string>
 #include <type_traits>
+#include <utility>
 
 namespace hpx { namespace serialization
 {
     namespace detail
     {
-        HPX_HAS_MEMBER_XXX_TRAIT_DEF(serialize);
-
         template <class T> HPX_FORCEINLINE
         void serialize_force_adl(output_archive& ar, const T& t, unsigned)
         {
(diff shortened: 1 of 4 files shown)


hpx/runtime/serialization/access.hpp
```diff
@@ -36,6 +34,32 @@ namespace hpx { namespace serialization
 
     class access
     {
+        template <class T>
+        class has_serialize
+        {
+            struct two_chars
+            {
+                char _[2];
+            };
```
sithhell: Can we use `std::true_type` and `std::false_type` here?

AntonBikineev: Not that I can think of.
Do you mean to use std::true/false_type as the return types for the test functions? If so, I don't think it makes a big difference. If you mean to inherit from std::true/false_type, then see the related discussion in #3134 why it didn't work for gcc.

AntonBikineev:
@hkaiser thanks for letting me work on this!

hpx/runtime/serialization/access.hpp
```diff
@@ -36,6 +34,32 @@ namespace hpx { namespace serialization
...
+            // to implement if there hadn't been an issue with gcc:
+            // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=82478
+            template <class T1, class = decltype(
+                    std::declval<typename std::remove_const<T1>::type &>().
+                        serialize(std::declval<output_archive &>(), 0u))>
+            static two_chars test(int);
+
+        public:
+            static constexpr bool value =
+                sizeof(test<T>(0)) == sizeof(two_chars);
```
sithhell: related to the above discussion ... I was thinking of why for example `value = decltype(test<T>(0))::value;` didn't work. It looks like this is unrelated to the discussion in #3134. If this is the only way to get it to work, there's nothing one can do, right?

AntonBikineev:
@sithhell yeah, I agree. It does look better now.

PR #3158, opened 2018-02-08
Fixing a race with timed suspension (second attempt)

sithhell:
When doing a timed suspension on a thread, there were several race conditions
between setting the different participating thread states, which should now be
fixed. The problem was observable with the wait_for_1751 regression test.

This is the second attempt (after #3153), fixing the previous issues.

### DIFF
--- a/hpx/runtime/threads/thread_helpers.hpp
+++ b/hpx/runtime/threads/thread_helpers.hpp
@@ -23,6 +23,7 @@
 #include <hpx/util/steady_clock.hpp>
 #include <hpx/util/thread_description.hpp>
 
+#include <atomic>
 #include <chrono>
 #include <cstddef>
 #include <cstdint>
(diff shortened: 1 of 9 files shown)


hpx/runtime/threads/thread_helpers.hpp
```diff
@@ -104,11 +105,24 @@ namespace hpx { namespace threads
...
         error_code& ec = throws);
 
+    inline thread_id_type set_thread_state(thread_id_type const& id,
+        util::steady_time_point const& abs_time,
+        thread_state_enum state = pending,
+        thread_state_ex_enum stateex = wait_timeout,
+        thread_priority priority = thread_priority_normal,
+        error_code& ec = throws)
+    {
+        std::atomic<bool> timer_started(false);
```
hkaiser: Won't this atomic go out of scope before the timer thread can access it?

sithhell: Good catch!

sithhell:
The inspect failure is not related to this PR. Pycicle tests run through and I verified that everything works locally. This PR is now potentially ready to be merged.

hkaiser:
> The inspect failure is not related to this PR. Pycicle tests run through and I verified that everything works locally. This PR is now potentially ready to be merged.

At least one of the inspect failures seems to be related to your changes (using atomic). Would you mind fixing the other one while you're at it as well?

sithhell:
Right. missed the missing atomic header. The other inspect error is already fixed on master.

PR #3159, opened 2018-02-08
Support Checkpointing Components

aserio:
Extended current implementation of `hpx::util::checkpoint` to properly handle components.

Summary:
 - Fixed `checkpoint.cpp` to make floats constant
   - Makes MSVC happy
 - Adding `checkpoint_component.cpp`
 - Raise abstraction level in checkpoint
   - Create new function arch_data that will be overloaded to properly
      handle components
 - Adding in components to test
 - Working component checkpointing!
   - Checkpoints a server
   - Still needs to create a new client
 - Create new client on restored server
 - Adding functionality to enable checkpointing of clients
   - Checkpoints the server the client points to
   - Still need to add functionality to restore which would
      create a new client with the resurrected server
 - Allowing components to be restored with provided client
   - Users can now use clients to  checkpoint and restore
      servers
 - Updating Documentation
 - Preparing unit test for checkpointing components
 - Fixing 1d_stencil_4_checkpoint example for Windows/Mac
 - Fixing compilation on gcc
 - Adding documentation for checkpointing components
 - Clean up code with Clang-Format
   - Update year on license

### DIFF
--- a/hpx/util/checkpoint.hpp
+++ b/hpx/util/checkpoint.hpp
@@ -26,6 +26,7 @@
 #include <cstddef>
 #include <fstream>
 #include <iosfwd>
+#include <memory>
 #include <sstream>
 #include <string>
 #include <type_traits>
@@ -205,23 +206,28 @@ namespace util {
     namespace detail {
         // Properly handle non clients
         template <typename T,
-            typename U = typename std::enable_if<
-                !hpx::traits::is_client<T>::value>::type>
-        void arch_data(hpx::serialization::output_archive& ar, T&& t)
+            typename U = typename std::enable_if<!hpx::traits::is_client<
+                typename std::decay<T>::type>::value>::type>
+        T&& prep(T&& t)
         {
-            ar << t;
+            return std::forward<T>(t);
         }
         // Properly handle Clients to compoents
-        template <typename Client, typename Server,
-            typename U =
-                typename std::enable_if<traits::is_client<Client>::value>::type>
-        void arch_data(hpx::serialization::output_archive& ar,
-            hpx::components::client_base<Client, Server>&& c)
+        template <typename T,
+            typename U = typename std::enable_if<hpx::traits::is_client<
+                typename std::decay<T>::type>::value>::type>
+        hpx::future<std::shared_ptr<
+            typename std::decay<T>::type::server_component_type>>
+        prep(T&& c)
         {
             // Use shared pointer to serialize server
-            hpx::future<std::shared_ptr<Server>> f_ptr =
-                hpx::get_ptr<Server>(c.get_id());
-            ar << f_ptr.get();
+            using server_type =
+                typename std::decay<T>::type::server_component_type;
+            return c.then(
+                [](typename std::decay<T>::type && c)
+                {
+                    return hpx::get_ptr<server_type>(c.get_id());
+                });
         }
         struct save_funct_obj
         {
@@ -231,8 +237,9 @@ namespace util {
                 // Create serialization archive from checkpoint data member
                 hpx::serialization::output_archive ar(c.data);
                 // Serialize data
-                int const sequencer[] = {// Trick to expand the variable pack
-                    0, (arch_data(ar, std::move(ts)),0)...  // Takes advantage of the comma operator
+                int const sequencer[] = {
+                    // Trick to expand the variable pack
+                    0, (ar << ts, 0)...    // Takes advantage of the comma oper.
                 };
                 (void) sequencer;    // Suppress unused param. warnings
                 return std::move(c);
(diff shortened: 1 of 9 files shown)


hpx/util/checkpoint.hpp
```diff
@@ -236,10 +262,13 @@ namespace util {
     ///
     /// Save_checkpoint takes any number of objects which a user may wish
     /// to store and returns a future to a checkpoint object.
-    /// Additionally the function can take a policy as a first object which
-    /// changes its behavior depending on the policy passed to it. Most
-    /// notably, if a sync policy is used save_checkpoint will simply return a
-    /// checkpoint object.
+    /// This funciton can also store a component either by passing a
```
hkaiser: typo: 'function', same applies several times below

hpx/util/checkpoint.hpp
```diff
@@ -195,8 +201,28 @@ namespace util {
...
+        template <typename Client, typename Server,
+            typename U =
+                typename std::enable_if<traits::is_client<Client>::value>::type>
+        void arch_data(hpx::serialization::output_archive& ar,
+            hpx::components::client_base<Client, Server>&& c)
+        {
+            // Use shared pointer to serialize server
+            hpx::future<std::shared_ptr<Server>> f_ptr =
+                hpx::get_ptr<Server>(c.get_id());
+            ar << f_ptr.get();
```
sithhell: You should directly serialize the future here, otherwise you'll easily deadlock.

hpx/util/checkpoint.hpp
```diff
@@ -461,16 +504,43 @@ namespace util {
...
+        // Properly handle client/server restoration
+        template <typename Client, typename Server>
+        void restore_impl(hpx::serialization::input_archive& ar,
+            hpx::components::client_base<Client, Server>& c)
+        {
+            std::shared_ptr<typename hpx::components::client_base<Client,
+                Server>::server_component_type>
+                server_ptr;
+            // Revive server
+            ar >> server_ptr;
```
sithhell: Same here, load a future directly.

hpx/util/checkpoint.hpp
aserio: @sithhell How would I get a deadlock here?

sithhell: Checkpointing will block until the future is ready. This might be a problem, doesn't have to.
There's special support for futures though which you might want to support anyway.

hkaiser: Could you elaborate what problems might arise from a local only serialization (besides making it more asynchronous)?

sithhell: Depends on where you run the code. It should not lead to a deadlock, probably, when you just checkpoint a given object hierarchy. It all depends on what get_ptr is doing and whether the code that depends on the checkpointing to succeed is responsible for allowing the call to `get_ptr` or `c.get_id()` (which is another blocking call) succeed.

hkaiser: get_ptr will succeed for local (to its invocation) objects only, and check-pointing does not call `get_id()`, iirc.

sithhell: I am not concerned that those calls might fail. I am mostly concerned that a operation that might follow checkpointing will only make those operations to return the result. This was a major concern for sending parcels, as said, not sure the same applies here. This is just something that caught my attention.

aserio: @sithhell I confused about what changes you would like to see me make. Would you like to see the `ar << f_ptr.get()` wrapped in a dataflow?

aserio: @sithhell , @hkaiser and I have come up with a way which should address the issue. Let me push that solution for your review.

aserio: @sithhell  do these changes satisfy your request?

hkaiser:
I looked through the code we have to see what we can do to come to a mutually acceptable solution for this. I think we can create a specialization for `hpx::traits::serialization_access_data<>` to used for check-pointing that would fully handle the specifics. This trait already isolates the handling of (general) futures and of id_types from the actual serialization process. Providing a new specialization for check-pointing should allow to do the dereferencing for id_types we need.

msimberg:
What's the status of this PR?

hkaiser:
@msimberg this still needs some work on the serialization end.

hkaiser:
Let's keep this open for now This is needed functionality.

hkaiser:
@aserio I have rebased and slightly cleaned up this branch for the new serialization module. We agreed with @sithhell that we will not support check-pointing managed id_type's, if a user needs to check-point a component he/she can use clients. Please verify and review.

hkaiser:
This should be good to go now. Please review.

aserio:
LGTM!

hkaiser:
The errors on pycicle are unrelated.

src/runtime/naming/name.cpp
```diff
@@ -261,16 +261,29 @@ namespace hpx { namespace naming
...
+                    "client instead");
+            }
+
             auto& split_gids = ar.get_extra_data<
                 serialization::detail::preprocess_gid_types>();
 
             if (split_gids.has_gid(*this))
             {
+                // archive doesn't support credit splitting or no splitting
+                // required
```
sithhell: This comment is not correct. The gid has been split already and we don't need to do anything further.

hkaiser: Good catch, thanks; it's fixed on master now.

PR #3164, opened 2018-02-11
Fixing component migration

sithhell:
- Simplified migration handling in primary namespace
 - Fixing various race conditions

### DIFF
--- a/hpx/runtime/agas/addressing_service.hpp
+++ b/hpx/runtime/agas/addressing_service.hpp
@@ -1377,9 +1377,9 @@ public:
     /// start/stop migration of an object
     ///
     /// \returns Current locality and address of the object to migrate
-    hpx::future<std::pair<naming::id_type, naming::address> >
-        begin_migration_async(naming::id_type const& id);
-    hpx::future<bool> end_migration_async(naming::id_type const& id);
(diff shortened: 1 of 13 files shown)


hpx/runtime/agas/addressing_service.hpp
```diff
@@ -1377,9 +1377,9 @@ struct HPX_EXPORT addressing_service
     /// start/stop migration of an object
     ///
     /// \returns Current locality and address of the object to migrate
-    hpx::future<std::pair<naming::id_type, naming::address> >
-        begin_migration_async(naming::id_type const& id);
-    hpx::future<bool> end_migration_async(naming::id_type const& id);
+    std::pair<naming::id_type, naming::address>
+        begin_migration(naming::id_type const& id);
+    bool end_migration(naming::id_type const& id);
```
hkaiser: These operations are potentially remote. What was your rationale to change this to run synchronously?

sithhell: The main rationale was that those two functions were always executed on the components service locality. Remote execution never happened in the migration implementation we have.

hkaiser: Yes, you're right - thanks.

PR #3170, opened 2018-02-13
Add cmake flag -DHPX_WITH_FAULT_TOLERANCE=ON (OFF by default)

KADichev:
to ignore connection reset signals of connections, so HPX does not terminate if a node crashes.


hpx/runtime.hpp
```diff
@@ -345,6 +345,12 @@ namespace hpx
         boost::scoped_ptr<components::server::memory> memory_;
         boost::scoped_ptr<components::server::runtime_support> runtime_support_;
     };
+    ///////////////////////////////////////////////////////////////////////////
+    /// Keep the factory object alive which is responsible for the given
+    /// component type. This a purely internal function allowing to work
+    /// around certain library specific problems related to dynamic
+    /// loading of external libraries.
+    HPX_EXPORT bool tolerate_node_faults();
```
hkaiser: This comment does not match the code.

hkaiser:
What is the rationale of making this functionality optional? Wouldn't it be better to be always able to handle this kind of exception?

hpx/runtime.hpp
KADichev: sorry, copy/paste :-( I'll add a proper comment

KADichev:
The functionality is optional, because I'm not sure if ignoring a "connection_reset" signal due to suddenly failing remote endpoints is the right way of handling this -- it is a workaround to avoid HPX raising an exception, and terminating, and allowing to explore recovery (which we are working on).

hkaiser:
> The functionality is optional, because I'm not sure if ignoring a "connection_reset" signal due to suddenly failing remote endpoints is the right way of handling this -- it is a workaround to avoid HPX raising an exception, and terminating, and allowing to explore recovery (which we are working on).

In the end this should be the default, it shouldn't even be possible to disable this. I would have no objections to go ahead with merging this PR if that's the eventual goal.

sithhell:
The goal should be to have a fault tolerance HPX in the end. I am all for having it optional, since there's always a trade off between fault tolerance and performance. This is also the first step in that direction so it makes sense to have it defaulted to being turned off for now.

hkaiser:
> The goal should be to have a fault tolerance HPX in the end. I am all for having it optional, since there's always a trade off between fault tolerance and performance. 

Ok. Fair enough.

> This is also the first step in that direction so it makes sense to have it defaulted to being turned off for now.

This PR proposes to have it set to `ON` by default.

hpx/runtime.hpp
hkaiser: @KADichev ping?

KADichev:
I fixed the comment and pushed the changes to the branch Let me know if you need anything else.

hpx/runtime.hpp
KADichev: I just updated the branch. Not sure if anything else is required of me here. Never did PR before.

hkaiser:
@KADichev now inspect is unhappy: https://9559-4455628-gh.circle-artifacts.com/0/tmp/circle-artifacts.TwEixCv/hpx_inspect_report.html

Issue #3166, opened 2018-02-11
Application (Octotiger) gets stuck on hpx::finalize when only using one thread


G-071:
## Expected Behavior

I am currently trying to move the changes I made in Octotiger to the master branch. Therefore I am moving away from HPX 1.0.0 to a more recent version. Expected behaviour: hpx::finalize still works as before for one or more threads.

## Actual Behavior

Octotiger (current master branch) itself compiles with a recent(ish) commit of HPX and also works as intended - unless I only uses a single thread. In this case Octotiger still produces the the right result but gets stuck on hpx::finalize. This problem did not occur with HPX 1.0.0.

See inital issue: STEllAR-GROUP/octotiger#61

## Steps to Reproduce the Problem

  1. Run octotiger -t1 -Problem=moving_star -Max_level=3 -Odt=0.3 -Stoptime=0.2 -Xscale=20.0 -Omega=0.1 -Stopstep=9 -Disableoutput

## Specifications

  - HPX Version:  STEllAR-GROUP/hpx@29da281b87cfa1171b42f5e2f22caa29b3c4be2d 
  - Octotiger Version: STEllAR-GROUP/octotiger@afef65c6aa09785ff33b01ca8254b25a98f08faf
  - Vc Version: STEllAR-GROUP/Vc@157026d3adf3494922269d11a2e4dede09bf867c 
  - gcc 5.4.0, boost 1.63

msimberg:
@G-071 There was a problem like this earlier which should already be fixed in master. Could you try again with 0867f8aa21796812fbd834188a7687616e555dc3 or newer (i.e. anything after #3110 was merged)?

G-071:
By now, I have a workaround for STEllAR-GROUP/octotiger#60 so I rebuilt everything with the current HPX master ( ef7360bf21cabe9ccc813e9d757746ec44feaa5a ) - The problem still persists!

msimberg:
@G-071 Thanks for testing. Seeing as you already had the problem with c9f2c09d9378d920f7d68ce857ca3729014fc601 it is most certainly not what I suspected, as you've just confirmed as well.

I will try to get octotiger running to reproduce the problem. I assume you're on Linux? Does debug or release make a difference?

msimberg:
[...]

Issue #2705, opened 2017-06-20
Cleanup required for resource partitioner


shoshijak:
- Merge threadmanager and threadmanager_impl (since threadmanager is not templated on the scheduler any more, this type erasure mechanism is not needed any more) 
- Merge runtime and runtime_impl (same reasoning)
- Namespace review: move thread_pool and thread_pool_impl out of the detail namespace, and move threadmanager (and runtime?) into the detail namespace
- Scheduler cleanup: reduce the number of schedulers but make the remaining schedulers more flexible, for example, by turning flags on/off. Flags could be things like "stealing/none", "differentiation between priorities/not", "numa sensitive/not" etc.
- Redundancy: both runtime and resource_partitioner have a pointer to the thread manager

PR #3172, opened 2018-02-13
Fix lifo queue backend

msimberg:
This addresses the issue in #3166.

The issue in this case was that when `shutdown_all` waits for all threads to finish it calls `suspend(pending)` which eventually calls `schedule_last` for the `shutdown_all` thread. For the lifo queue backend `schedule_last` is the same as `schedule` and `shutdown_all` will be scheduled as the next thread again, leaving no chance for the remaining threads to run. I suspect we haven't seen this with more than 1 thread because then other OS threads will be able to steal the threads and/or most tests and examples wait for all futures before calling `hpx::finalize`.

## Proposed Changes

- Use deque for the non-abp lifo and fifo queue backends so that `schedule_last` can actually schedule a thread last in the queue.
- Add a test which reproduces the issue.
- Fix some naming in the documentation related to schedulers.

Note: from my preliminary tests deque is neither slower nor faster than stack or queue, but I need to check with more examples to see that this holds more generally.

### DIFF
--- a/src/runtime/threads/detail/scheduled_thread_pool.cpp
+++ b/src/runtime/threads/detail/scheduled_thread_pool.cpp
@@ -38,6 +38,9 @@ template class HPX_EXPORT hpx::threads::detail::scheduled_thread_pool<
 template class HPX_EXPORT hpx::threads::detail::scheduled_thread_pool<
     hpx::threads::policies::local_priority_queue_scheduler<hpx::compat::mutex,
         hpx::threads::policies::lockfree_abp_fifo>>;
+template class HPX_EXPORT hpx::threads::detail::scheduled_thread_pool<
+    hpx::threads::policies::local_priority_queue_scheduler<hpx::compat::mutex,
+        hpx::threads::policies::lockfree_abp_lifo>>;
 #endif
 
 #if defined(HPX_HAVE_HIERARCHY_SCHEDULER)
--- a/docs/manual/commandline.qbk
+++ b/docs/manual/commandline.qbk
@@ -85,9 +85,9 @@ described in the table below:
     [[`--hpx:print-bind`]       [print to the console the bit masks calculated from the
                                  arguments specified to all `--hpx:bind` options.]]
     [[`--hpx:queuing arg`]      [the queue scheduling policy to use, options are
-                                 'local/l', 'local-priority-fifo/lo', 'local-priority-lifo', 'abp/a',
-                                 'abp-priority', 'hierarchy/h', and 'periodic/pe'
-                                 (default: local-priority-fifo/lo)]]
+                                 'local/l', 'local-priority-fifo/lo', 'local-priority-lifo',
+                                 'abp-priority-fifo/a', 'abp-priority-lifo', 'hierarchy/h', and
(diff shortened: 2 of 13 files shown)


src/runtime/threads/detail/scheduled_thread_pool.cpp
```diff
@@ -38,6 +38,9 @@ template class HPX_EXPORT hpx::threads::detail::scheduled_thread_pool<
 template class HPX_EXPORT hpx::threads::detail::scheduled_thread_pool<
     hpx::threads::policies::local_priority_queue_scheduler<hpx::compat::mutex,
         hpx::threads::policies::lockfree_abp_fifo>>;
+template class HPX_EXPORT hpx::threads::detail::scheduled_thread_pool<
+    hpx::threads::policies::local_priority_queue_scheduler<hpx::compat::mutex,
+        hpx::threads::policies::lockfree_abp_lifo>>;
 #endif
```
hkaiser: If you add this specialization then you probably should add a command line option for ir as well.

hkaiser:
Is the underlying assumption for this PR that `boost::lockfree:stack` is broken?

src/runtime/threads/detail/scheduled_thread_pool.cpp
msimberg: Yes, you're right, will add that.

hkaiser: OTOH, instead of adding more schedulers to HPX core, we should consider creating a better mechanism (using the thread-pools) allowing for all schedulers (except the default one) to live outside of HPX core. What do you think?

msimberg:
No, not at all. It's simply that the scheduling loop relies on `schedule_last` to actually schedule a task last, but with a stack this requirement is broken and is by definition not possible to do.

src/runtime/threads/detail/scheduled_thread_pool.cpp
msimberg: Not thought about it too much, but yes, that would probably be a good thing. How far outside HPX core are you thinking? This probably goes hand in hand with having a nicer way to create thread pools in general.

biddisco:
Note that #2705 also mentions scheduler cleanup. Some schedulers should be removed and we should consider replacing the default scheduler.

msimberg:
To follow up on the performance, from what I can tell there is no practical difference between using a deque, stack or queue (tested with lots of small tasks with `for (...) hpx::async([](){});` and the stencil examples). I think this is safe to merge without causing a performance hit for the default scheduler.

I'm also taking the liberty of quoting @biddisco from irc, saying there is no difference for his particular application:
```
13:15 <jbjnr> at the moment, abp_lifo
13:15 <heller_> deque is slower since the implementation uses dcas, which is not lockfree
13:15 <heller_> IIRC
13:15 <jbjnr> I vary them from time to time to experiemnt, but don't seem to get much differnceeither way
```

hkaiser:
Just a side remark - I don't really trust the ABP schedulers (perhaps without reason), so we may not want to use one of those as the default scheduler.

PR #3174, opened 2018-02-15
Fix format to

### LINKED ISSUE #3013: hello_world occasionally prints multiple lines from a single OS-thread
`hello_world` occasionally outputs something like this:

```
> ./bin/hello_world
hello world from OS-thread 3 on locality 0
hello world from OS-thread 3 on locality 0
hello world from OS-thread 0 on locality 0
hello world from OS-thread 1 on locality 0
hello world from OS-thread 2 on locality 0
```

Not sure if related, but it also sometimes output something like this:
```
> ./bin/hello_world
hoello world from OS-thread 0 on locality MemFree:     00
hoello world from OS-thread 0 on locality MemFree:     00
hello world from OS-thread 3 on locality 0
hello world from OS-thread 2 on locality 0
```

msimberg:
Fixes #3013.

## Proposed Changes

- Adds an overload to `hpx::util::format_to` taking a `hpx::iostreams::ostream` to make sure that the correct operators (`<<`) are called.

When a `hpx::iostreams::ostream` is used as a `std::stream` the correct virtual functions are not called, thus bypassing locks that are supposed to be taken in the `hpx::iostreams::ostream` implementation (base class functions are not virtual?). A proper fix would make sure the correct functions are called, but at the moment I have no fix for that.

### DIFF
--- a/examples/quickstart/hello_world.cpp
+++ b/examples/quickstart/hello_world.cpp
@@ -42,10 +42,10 @@ std::size_t hello_world_worker(std::size_t desired)
     if (current == desired)
     {
         // The HPX-thread has been run on the desired OS-thread.
-        char const* msg = "hello world from OS-thread %1% on locality %2%";
+        char const* msg = "hello world from OS-thread %1% on locality %2%\n";
 
         hpx::util::format_to(hpx::cout, msg, desired, hpx::get_locality_id())
-                  << std::endl << hpx::flush;
+            << hpx::flush;
 
         return desired;
     }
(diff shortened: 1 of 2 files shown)


hkaiser:
For the records, the actual problem is that `hpx::cout` has it's own `operator<<(T)` that does the locking (see https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/components/iostreams/ostream.hpp#L330-L335). This operator will not be invoked if the stream is used through its `std::ostream` base class. This should be fixed by a separate PR.

PR #3179, opened 2018-02-17
Various minor changes to support FLeCSI

hkaiser:
- adding more constructors to resource::partitioner (aligned with hpx::init)
- adding hpx::parallel::execution::pool_executor alias (for consistency)
- adding default constructor for pool-executor
- adding support to request minimal number of cores (threads) (hpx.force_min_os_threads)
- Making sure resume/suspend is not called for schedulers that don't support it

### DIFF
--- a/hpx/include/thread_executors.hpp
+++ b/hpx/include/thread_executors.hpp
@@ -1,4 +1,4 @@
-//  Copyright (c) 2007-2015 Hartmut Kaiser
+//  Copyright (c) 2007-2018 Hartmut Kaiser
 //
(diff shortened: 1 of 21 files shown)


msimberg:
@hkaiser Excellent catches regarding suspension. Sorry for missing this. This looks good to me, but for the long term I should make the checks more fine grained. Suspension on static schedulers is only disabled because of the problem of suspending a thread from the same pool, but suspending the whole pool from another pool should in fact work without problems. I'll try to work out something for that.

PR #3185, opened 2018-02-20
Remove unused scoped_unlock, unlock_guard_try

K-ballo:


K-ballo:
> Should we add a
> 
> ```
> #if defined(HPX_HAVE_SCOPED_LOCK_COMPATIBILITY)
> template <typename Lock>
> using scoped_unlock<Lock> = unlock_guard<Lock>;
> #endif
> ```
> 
> and follow the usual deprecation steps instead of just dropping scoped_unlock without proper phasing-out?

Sounds good, specially if compatibility defaults to OFF.

K-ballo:
Added deprecation compatibility flag.

PR #3188, opened 2018-02-22
Optionally disable printing of diagnostics during terminate.

hkaiser:
This also removes the call to std::abort in the termination handler as the system is aborting anyways.

### DIFF
--- a/src/runtime.cpp
+++ b/src/runtime.cpp
@@ -64,14 +64,15 @@ namespace hpx
             util::attach_debugger();
         }
 
-        std::cerr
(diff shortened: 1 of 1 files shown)


sithhell:
Out of curiosity, what's the original problem this PR is addressing?

hkaiser:
> Out of curiosity, what's the original problem this PR is addressing?

If you work with Python loading modules that depend on HPX, then forcefully existing the python interpreter causes HPX to print the full debug info trace, which is just confusing. This allows to disable printing things, if required.

PR #3189, opened 2018-02-22
Deleting one include of median from BOOST library to arithmetics_counter file

victor-ludorum:
Two times **"#include <boost/accumulators/statistics/median.hpp>"** is included. So, one should be removed .
@hkaiser Would you mind to review !!


victor-ludorum:
@hkaiser Sir, I have updated it . Sorry, I haven't recognized that previously.

PR #3192, opened 2018-02-23
Move type info into hpx::debug namespace and add print helper functions

biddisco:
## Proposed Changes
Adds an extra couple of type info print helpers and moves the demangler into the debug namespace.
Might not compile on all platforms because we had some problems with it in the past, so wanting to see what pycicle gives after a build/test of this PR.

### DIFF
--- a/hpx/util/debug/demangle_helper.hpp
+++ b/hpx/util/debug/demangle_helper.hpp
@@ -10,25 +10,42 @@
 #include <hpx/config.hpp>
 #include <string>
 #include <type_traits>
+#include <typeinfo>
 
-#if defined(__GNUC__)
+// --------------------------------------------------------------------
+// Always present regardless of compiler : used by serialization code
+// --------------------------------------------------------------------
+namespace hpx { namespace util { namespace debug
+{
+    template <typename T>
+    struct demangle_helper
+    {
+        char const* type_id() const
+        {
+            return typeid(T).name();
+        }
+    };
+}}}
+
+#if defined(__GNUG__)
+
 #include <cxxabi.h>
 #include <stdlib.h>
 
+// --------------------------------------------------------------------
+// if available : demangle an arbitrary c++ type using gnu utility
+// --------------------------------------------------------------------
 namespace hpx { namespace util { namespace debug
 {
-    // --------------------------------------------------------------------
-    // demangle an arbitrary c++ type using gnu utility
-    // --------------------------------------------------------------------
     template <typename T>
-    class demangle_helper
+    class cxxabi_demangle_helper
     {
     public:
-        demangle_helper()
+        cxxabi_demangle_helper()
           : demangled_(abi::__cxa_demangle(typeid(T).name(), 0, 0, 0))
         {}
 
-        ~demangle_helper()
+        ~cxxabi_demangle_helper()
         {
             free(demangled_);
         }
@@ -46,18 +63,10 @@ namespace hpx { namespace util { namespace debug
 
 #else
 
-#include <typeinfo>
-
 namespace hpx { namespace util { namespace debug
 {
     template <typename T>
-    struct demangle_helper
-    {
-        char const* type_id() const
-        {
-            return typeid(T).name();
-        }
-    };
+    using cxxabi_demangle_helper = demangle_helper<T>;
 }}}
 
 #endif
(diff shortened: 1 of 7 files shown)


hpx/util/demangle_helper.hpp
```diff
@@ -36,6 +41,7 @@ namespace hpx { namespace util
     private:
         char* demangled_;
     };
+
```
hkaiser: The same structure below needs to be moved into `hpx::debug` as well

hpx/util/demangle_helper.hpp
```diff
@@ -67,6 +73,31 @@ namespace hpx { namespace util
...
+    // --------------------------------------------------------------------
+    // print type information
+    // usage : std::cout << print_type<args...>("separator")
+    // separator is appended if the number of types > 1
+    // --------------------------------------------------------------------
+    template <typename T=void>
+    inline std::string print_type(const char *delim="")
+    {
+        return std::string(debug::type_id<T>::typeid_.type_id());;
```
hkaiser: Superfluous ';'

msimberg:
Should these be moved to an `hpx/debug` folder to match the namespace?

hkaiser:
> Should these be moved to an hpx/debug folder to match the namespace?

Good point!

biddisco:
would `hpx/util/debug` be acceptable, or would people prefer `hpx/debug`

hkaiser:
> would hpx/util/debug be acceptable, or would people prefer hpx/debug

I'd vote for `hpx/util/debug/`.

hpx/util/demangle_helper.hpp
```diff
@@ -1,73 +0,0 @@
...
-//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
-
-#if !defined(HPX_UTIL_DEMANGLE_HELPER_OCT_28_2011_0410PM)
-#define HPX_UTIL_DEMANGLE_HELPER_OCT_28_2011_0410PM
-
-#include <hpx/config.hpp>
-
-// disable the code specific to gcc for now as this causes problems
-// (see #811: simple_central_tuplespace_client run error)
-#if 0 // defined(__GNUC__)
```
biddisco: Am I right in thinking that the reason why the code was #ifdefed out here is because the demangle_helper uses `abi::__cxa_demangle(typeid(T).name(), 0, 0, 0)` which returns a `char *` and this char * is deleted when the helper goes out of scope - but the `char *` is held onto by the (serialization) code that  uses it.

If this is the case, is there anything we can do to make the serialization code keep the memory alive?
For debugging, the demangled version is very useful, in the `debug::print(...)` code, the `char *` is copied into a string, so no memory violations occur. Perhaps we should just use the undemangled version in the serialization, and keep the demangled one for debug stuff?

sithhell: for that, we'd just need to implement proper resource management. `std::string`, `std::shared_ptr<char>` or `std::vector<char>` are available.

biddisco: I played with this some more and something very odd is going on. The code segfaults in function registration of pointers when the debug mangler is used, but it's not caused by the pointer being invalid AFAICT. the test `any_serialization_test_exe` fails if the GNU cxxabi demangler is used and I'm baffled as to what is going wrong.

biddisco: I think I'll leave the serialization code using the standard typeid stuff and enable the demangler for debugging and printing types, it is very useful to me and worth having, but I'll just provide two types and use one for debug and the other for serialization. Unless someone has a clue as to why the cxxabi demangler causes problems

biddisco:
I have made 2 versions of the debug mangler
serialization code always uses plain typeid for safety, until we understand why the code fails when using the cxxabi version. The cxxabi version is enabled for debug printing on systems that support it.

I'd like this to go into 1.2 release if possible.

hpx/util/debug/demangle_helper.hpp
```diff
@@ -0,0 +1,126 @@
...
+    class cxxabi_demangle_helper
+    {
+    public:
+        cxxabi_demangle_helper()
+          : demangled_(abi::__cxa_demangle(typeid(T).name(), 0, 0, 0))
+        {}
+
+        ~cxxabi_demangle_helper()
+        {
+            free(demangled_);
```
sithhell: Due to this, the class can't be copied. I think it would make sense to use proper memory management here. As I suggested previously, using either `std::vector<char>`, `std::unique_ptr<char, deleter>` or even `std::string` might help here.

biddisco: just to clarify - the code in question was copied directly from the old implementation and used only in one place to initialize a static var that is used for the symbol translation (so copying is not planned or supported for this class). I will however use a unique_ptr if that makes you happy.

PR #3202, opened 2018-02-27
Making sure resource partitioner is not accessed if its not valid

hkaiser:


hkaiser:
@msimberg this got merged too early, there are still missing `return` statements.

PR #3205, opened 2018-02-28
Fixing return type calculation for bulk_then_execute.

### LINKED ISSUE #3182: bulk_then_execute has unexpected return type/does not compile
The following snippet of code does not do what I would expect it to do:
```
void fun1(int, shared_future<void>) {...}
int fun2(int, shared_future<void>) {...}
...
{
    shared_future<void> f = ...;
    std::vector<int> v(100);
    // tested with exec as pool_executor, thread_pool_executor or this_thread_executor
    bulk_then_execute(exec, &fun1, v, f); // Returns future<void>
    // bulk_then_execute(exec, &fun2, v, f); // Does not compile
}
```

## Expected Behavior

- with `fun1` I expect either that `.get()`ing the `future<void>` would block until all 100 invocations of `fun1` have finished or that it returns a `vector<future<void>>` like `bulk_async_execute`
- with `fun2` I expect it to compile and return a `vector<future<int>>`

## Actual Behavior

- with `fun1` `get()` does not wait for all invocations to finish
- with `fun2` it fails to compile with the following error:
```
/home/simbergm/src/hpx-worktrees/fix-examples/hpx/parallel/executors/thread_execution.hpp:213:25: error: could not convert ‘hpx::traits::future_access<hpx::lcos::future<R>>::create(boost::intrusive_ptr<U>&&) [with SharedState = hpx::lcos::detail::future_data_base<std::vector<hpx::lcos::future<int>, std::allocator<hpx::lcos::future<int> > > >; R = std::vector<hpx::lcos::future<int>, std::allocator<hpx::lcos::future<int> > >]()’ from ‘hpx::lcos::future<std::vector<hpx::lcos::future<int>, std::allocator<hpx::lcos::future<int> > > >’ to ‘std::enable_if<true, hpx::lcos::future<std::vector<int> > >::type {aka hpx::lcos::future<std::vector<int> >}’
             std::move(p));
                         ^
```

## Specifications

  - HPX Version: 8b0f81c
  - Platform (compiler, OS): gcc 6.3.0, Ubuntu 16.04

hkaiser:
- flyby: added util::functional::unwrap[_n|_all]
- some restructuring required by circular #include dependencies

Fixes #3182

Unfortunately this PR contains a lot of unrelated changes that were necessary in order to disentangle a couple of circular #include dependencies that were introduced by the initial fix to the original problem.

### DIFF
--- a/docs/CMakeLists.txt
+++ b/docs/CMakeLists.txt
@@ -143,7 +143,7 @@ set(doxygen_dependencies
     "${PROJECT_SOURCE_DIR}/hpx/performance_counters/manage_counter_type.hpp"
     "${PROJECT_SOURCE_DIR}/hpx/runtime_fwd.hpp"
     "${PROJECT_SOURCE_DIR}/hpx/runtime/applier_fwd.hpp"
-    "${PROJECT_SOURCE_DIR}/hpx/runtime/basename_registration.hpp"
+    "${PROJECT_SOURCE_DIR}/hpx/runtime/basename_registration_fwd.hpp"
     "${PROJECT_SOURCE_DIR}/hpx/runtime/find_here.hpp"
     "${PROJECT_SOURCE_DIR}/hpx/runtime/find_localities.hpp"
     "${PROJECT_SOURCE_DIR}/hpx/runtime/get_colocation_id.hpp"
--- a/hpx/apply.hpp
+++ b/hpx/apply.hpp
@@ -7,11 +7,11 @@
 #define HPX_APPLY_APR_16_20012_0943AM
 
 #include <hpx/config.hpp>
-#include <hpx/lcos/future.hpp>
 #include <hpx/runtime/applier/apply.hpp>
 #include <hpx/runtime/applier/apply_continue.hpp>
 #include <hpx/runtime/threads/thread_executor.hpp>
 #include <hpx/runtime/threads/thread_helpers.hpp>
+#include <hpx/runtime_fwd.hpp>
 #include <hpx/traits/is_executor.hpp>
 #include <hpx/traits/is_launch_policy.hpp>
 #include <hpx/util/bind_action.hpp>
--- a/hpx/compute/cuda/target_distribution_policy.hpp
+++ b/hpx/compute/cuda/target_distribution_policy.hpp
@@ -12,7 +12,7 @@
 
 #if defined(HPX_HAVE_CUDA)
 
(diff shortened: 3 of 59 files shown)


msimberg:
Thanks @hkaiser, this works as intended (below is likely unrelated).

Out of curiousity, HPX futures don't seem to wait in the destructors if get/wait was not called earlier. This seems to be a contentious topic for the standards, but is there a reason for HPX choosing to go this way? (I'm asking because I wanted to see what the thread pool executors do now if one does not get the future before destroying the executor -- currently it hangs, but I have not investigated further).

hkaiser:
> Out of curiousity, HPX futures don't seem to wait in the destructors if get/wait was not called earlier. This seems to be a contentious topic for the standards, but is there a reason for HPX choosing to go this way? (I'm asking because I wanted to see what the thread pool executors do now if one does not get the future before destroying the executor -- currently it hangs, but I have not investigated further).

Yes, HPX futures don't wait on destruction if the value was not extracted. Please note that even the standard prescribes this _only_ for futures that have been returned from `std::async`. The rationale for the standard prescribing this is to avoid for the user to easily being able to create 'runaway' threads, i.e. threads that the user has no way of controlling as there is no 'handle' to them available anymore once the future instance has gone away.

We have deliberately diverged from this for two reasons: a) in HPX we can't have runways threads as the runtime waits for all threads to terminate eventually before exiting, also there is always a way for us to access still running threads, if needed, and b) (probably more important) we wanted for all futures to be equal in semantics, no matter how they were created (through `async` or any other means).

msimberg:
@hkaiser I added some missing includes which for some reason pycicle (and my local build) caught but circleci did not ([here](http://cdash.cscs.ch/viewBuildError.php?buildid=1857)). Will check if pycicle complains about something else after this.

hkaiser:
@msimberg Thanks a lot!

msimberg:
So this seems to have broken the `sequenced_executor` test ([link](http://rostam.cct.lsu.edu/builders/hpx_gcc_4_9_boost_1_57_centos_x86_64_release/builds/220/steps/run_unit_tests/logs/tests.unit.parallel.executors.sequenced_executor%20%280.28%20sec%29)).

I think the test itself may not be correct though. From the docs: "`hpx::parallel::execution::sequenced_executor`: creates groups of sequential execution agents which execute in the calling thread." It looks to me like the test should check that the thread ids are equal, not different. Am I interpreting the docs correctly?

Issue #2455, opened 2017-01-16
Add more arithmetic performance counters


hkaiser:
Add more arithmetic performance counters exposing statistic properties calculated from the values gathered by querying arbitrary underlying counters. These new counters would work along the lines of the existing arithmetics counters (like `/arithmetics/add`) which can be invoked with a list of other counters to query in order to produce the required result.

For instance:
```
hello_world \
    --hpx:threads=4 \
    --hpx:print-counter=/threads/count/cumulative
    --hpx:print-counter=/arithmetics/average@/threads{locality#0/worker-thread#*}/count/cumulative
```

could produce something like:

```
hello world from OS-thread 1 on locality 0
hello world from OS-thread 0 on locality 0
hello world from OS-thread 3 on locality 0
hello world from OS-thread 2 on locality 0
/threads{locality#0/worker-thread#0}/count/cumulative,1,0.0025214,[s],27
/threads{locality#0/worker-thread#1}/count/cumulative,1,0.0025453,[s],33
/threads{locality#0/worker-thread#2}/count/cumulative,1,0.0025683,[s],29
/threads{locality#0/worker-thread#3}/count/cumulative,1,0.0025904,[s],33
/arithmetics/average@/threads{locality#0/worker-thread#*}/count/cumulative,1,0.0026002,[s],30.5
```

etc.

aserio:
We should probably make this a GSoC project

diehlpk:
@hkaiser  https://github.com/STEllAR-GROUP/hpx/wiki/GSoC-2017-Project-Ideas#add-more-arithmetic-performance-counters

Sumit2318:
Sir, would you please guide me How to start contributing to this project?

hkaiser:
> Sir, would you please guide me How to start contributing to this project?

This is a nice and simple task, just perfect for GSoC. To start off, you should build HPX on your machine and make sure the examples are running properly. The next step would be to familiarize yourself with the concepts of performance counters in HPX. I think the docs should give you a good head start to this (see http://stellar-group.github.io/hpx/docs/html/hpx/manual/performance_counters.html).

Once you understand them and how to use them, I'd suggest trying to find the existing simple arithmetic performance counter implementations (i.e. `/arithmetic/add` and friends) and to try to analyze how things actually work (in the code). This gives you a good platform to start thinking about a) what additional counters might be useful, and b) how to implement them.

[...]

PR #3206, opened 2018-02-28
Addition of new arithmetic performance counter "Count"

victor-ludorum:
References #2455 

## Proposed Changes
This PR has added one new arithmetic performance counter which is ```count``` . This mainly returns the count of values present in the accumulator containing values of the underlying counters.

### DIFF
--- a/src/performance_counters/registry.cpp
+++ b/src/performance_counters/registry.cpp
@@ -987,7 +987,7 @@ namespace hpx { namespace performance_counters
                 gid = components::server::construct<counter_t>(
                     complemented_info, base_counter_names);
             }
-             else if (p.countername_ == "count") {
+            else if (p.countername_ == "count") {
                 typedef hpx::components::component<
                     performance_counters::server::arithmetics_counter_extended<
                         boost::accumulators::tag::count>
--- a/src/performance_counters/registry.cpp
+++ b/src/performance_counters/registry.cpp
@@ -987,6 +987,14 @@ namespace hpx { namespace performance_counters
                 gid = components::server::construct<counter_t>(
                     complemented_info, base_counter_names);
             }
+             else if (p.countername_ == "count") {
+                typedef hpx::components::component<
+                    performance_counters::server::arithmetics_counter_extended<
+                        boost::accumulators::tag::count>
+                > counter_t;
+                gid = components::server::construct<counter_t>(
+                    complemented_info, base_counter_names);
+            }
             else {
                 HPX_THROWS_IF(ec, bad_parameter,
                     "registry::create_arithmetics_counter",
--- a/src/runtime.cpp
+++ b/src/runtime.cpp
@@ -645,6 +645,16 @@ namespace hpx
               &performance_counters::default_counter_discoverer,
               ""
             },
+            // arithmetics count counter
+            { "/arithmetics/count", performance_counters::counter_aggregating,
+              "returns the count value of all values of the specified "
+              "base counters; pass the required base counters as the parameters: "
+              "/arithmetics/count@<base_counter_name1>,<base_counter_name2>",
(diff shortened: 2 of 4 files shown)


src/performance_counters/registry.cpp
```diff
@@ -986,6 +986,14 @@ namespace hpx { namespace performance_counters
                 > counter_t;
                 gid = components::server::construct<counter_t>(
                     complemented_info, base_counter_names);
+            }
+             else if (p.countername_ == "count") {
```
hkaiser: The indentation seems to be off here.

src/runtime.cpp
```diff
@@ -645,6 +645,16 @@ namespace hpx
               &performance_counters::default_counter_discoverer,
               ""
             },
+            // arithmetics count counter
+            { "/arithmetics/count", performance_counters::counter_aggregating,
+              "returns the count value of all values of the specified "
+              "base counters; pass the required base counters as the parameters: "
+              "/arithmetics/count@<base_counter_name1>,<base_counter_name2>",
```
hkaiser: If the counter returns the number of values in the accumulator, wouldn't it always return the number of base-counters specified as its argument?

victor-ludorum: @hkaiser Thanks for the review 😊 . Actually , Yes it will return the number of base-counters . Sorry for being little unclear . As mean, median and variance are done on the base counter values . Similarly ```count``` is also there to return the count values .

hkaiser:
@victor-ludorum while I understand that this is a small project for you to better understand how performance counters work (and kudos to you that you got it in place!), I'm not sure if adding this counter to the HPX core would be useful. What would be a possible use case for it?

victor-ludorum:
@hkaiser Thanks !! Actually , I have planned to work on this project . Therefore, I want to check that whether _the steps which I have thought to implement any new counter is correct or not_ . Therefore , I have created this PR so that you or any experienced one should suggest the steps is accurate or not .
So, Thanks for the review !!
**Usage of Count :** 
As there are several counters having statistical properties like ```mean```, ```median``` and ```variance``` are already implemented . But I have not seen any count which may state the number of values in accumulator . I know this is a basic one . But I have thought this should be implemented and should be present in the HPX performance counter . Because , sometime we may need the number of values which are present on which operations are performed . This will be given only by ```count``` .

>  I understand that this is a small project for you to better understand how performance counters work 

Please suggest , how can I improve my work . On which part I may work that should be good and more informative .

victor-ludorum:
@hkaiser Hello sir 😊 !! Do I close this PR . Before closing I just want one advice that whether this process is appropriate to add any new arithmetic performance counter to HPX . 
Thanks !!

hkaiser:
@victor-ludorum why should we close this PR? We can merge it as soon as all tests pass. Currently, inspect is unhappy (see here: https://9666-4455628-gh.circle-artifacts.com/0/tmp/circle-artifacts.SeTzLMh/hpx_inspect_report.html).

victor-ludorum:
Okay sir. So how can I rectify that so that it May pass all test cases.

victor-ludorum:
Hello @hkaiser In the details given in this [report](https://9666-4455628-gh.circle-artifacts.com/0/tmp/circle-artifacts.SeTzLMh/hpx_inspect_report.html) . It is giving error for **"Endline whitespace"**  . And I have given those white spaces as to segregate them and all other counters are also added in the same way . So, how Can I rectify that error . Do I omit all white spaces and then check for the report or I may till the present state get accepted .

hkaiser:
@victor-ludorum the empty lines are ok, they just contain some (at least one) space characters. Remove those and all will be fine.

victor-ludorum:
Okay @hkaiser  sir!! I will do it and then examine . Thanks for review!!

PR #3209, opened 2018-03-01
Fix locking problems during shutdown

msimberg:
This is to test if this helps for the Windows shutdown problems, but I will still see if we can get by with the `stopped_` and `terminated_` flags being `atomic<bool>`s instead (the mutex has to be locked in any case for the condition variables).

## Proposed Changes

- Unlock mutex while yielding/suspending HPX thread during shutdown
- Includes #3203 

HPX threads can migrate between OS threads. Holding an OS mutex in an HPX thread which gets migrated to another OS thread is undefined behaviour (this is checked on Windows). For #3203 a mutex was being held while being destructed.

### DIFF
--- a/src/runtime/components/server/runtime_support_server.cpp
+++ b/src/runtime/components/server/runtime_support_server.cpp
@@ -827,25 +827,6 @@ namespace hpx { namespace components { namespace server
         }
     }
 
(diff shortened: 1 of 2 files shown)


hkaiser:
This might actually make #3203 obsolete.

PR #3213, opened 2018-03-05
Relax locks for thread suspension to avoid holding locks when yielding

msimberg:
Fixes locks held during suspension.

## Proposed Changes

- Unlock locks earlier so that they are not held while suspending (HPX threads)
- This still keeps the guarantee the new threads are not created on suspended threads (when elasticity is enabled)

This fixes a problem uncovered by #3203 and #3209.


msimberg:
I do hope this is okay, but I'm not convinced yet, so please don't merge yet. I will get back to this on Thursday.

msimberg:
This is good AFAICT, but will revert if there are any problems.

PR #3229, opened 2018-03-12
Fix sync traversals of containers passed as r-value

Naios:
## Proposed Changes

  - This fixes the reference correctness of `traverse_pack` when traversing a container passed as rvalue


hkaiser:
@Naios could you please fix the remaining inspect check errors (https://9740-4455628-gh.circle-artifacts.com/0/tmp/circle-artifacts.pTMGwoc/hpx_inspect_report.html)? We can't move ahead with merging otherwise. Thanks!

Naios:
@hkaiser I corrected the outstanding issue

hkaiser:
@Naios you can ignore the AppVeyor problem (this is unrelated to your work), but could you please look into the errors reported by our pycicle CI system?

Naios:
@hkaiser I took a look into the error, this is a greater bug with which affects the whole GCC (6.x) family.
I could reproduce it on one of my remote machines using GCC 6.3.
Currently I don't have the resources to reproduce it (currently I'm missing time to do it and also I don't have an environment for debugging GCC properly).

Issue #3082, opened 2017-12-18
Simplify custom thread pool creation


msimberg:
Currently creating a new thread pool with a custom scheduling policy is easy:

```
rp.create_thread_pool("default", hpx::resource::scheduling_policy::local_priority_fifo)
```

but any other customization requires using the callback interface. For example to add a single scheduler mode flag:
```
rp.create_thread_pool("default",
    [](hpx::threads::policies::callback_notifier& notifier,
        std::size_t num_threads, std::size_t thread_offset,
        std::size_t pool_index, std::string const& pool_name)
    -> std::unique_ptr<hpx::threads::detail::thread_pool_base>
    {
        typename local_priority_queue_scheduler::init_parameter_type init(num_threads);
        std::unique_ptr<local_priority_queue_scheduler> scheduler(new local_priority_queue_scheduler(init));

        auto mode = hpx::threads::policies::scheduler_mode(
            hpx::threads::policies::do_background_work |
            hpx::threads::policies::reduce_thread_priority |
            hpx::threads::policies::delay_exit |
            hpx::threads::policies::enable_elasticity);

        std::unique_ptr<hpx::threads::detail::thread_pool_base> pool(
            new hpx::threads::detail::scheduled_thread_pool<local_priority_queue_scheduler>(
                std::move(scheduler), notifier, pool_index, pool_name, mode,
                thread_offset));

        return pool;
    });
```

While flexible, the callback interface is unnecessarily complicated for most use cases.

biddisco:
I would like to keep this issue alive. I will (slowly) work on it as part of my scheduler cleanup.

PR #3230, opened 2018-03-12
Add optional scheduler mode parameter to create_thread_pool function

msimberg:
## Proposed Changes

- Add a `default_mode` value to the scheduler mode enum
- Add a an optional scheduler mode parameter to `create_thread_pool` with a default value of `default_mode`

Addresses #3082, but I'm not sure what more we might want. This at least simplifies the use cases in the suspension tests (one does not need to use the callback version of `create_thread_pool` anymore to change scheduler modes).

### DIFF
--- a/hpx/runtime/resource/detail/partitioner.hpp
+++ b/hpx/runtime/resource/detail/partitioner.hpp
@@ -53,7 +53,9 @@ namespace hpx { namespace resource { namespace detail
 
     private:
         init_pool_data(const std::string &name,
-            scheduling_policy = scheduling_policy::unspecified);
+            scheduling_policy = scheduling_policy::unspecified,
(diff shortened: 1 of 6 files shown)


hpx/runtime/resource/detail/partitioner.hpp
```diff
@@ -53,7 +54,8 @@ namespace hpx { namespace resource { namespace detail
 
     private:
         init_pool_data(const std::string &name,
-            scheduling_policy = scheduling_policy::unspecified);
+            scheduling_policy = scheduling_policy::unspecified,
+            hpx::util::optional<hpx::threads::policies::scheduler_mode> = hpx::util::nullopt);
```
hkaiser: Wouldn't it be simpler to define an enumerator value representing a 'default_mode'?

PR #3231, opened 2018-03-12
Adding global_factorial example implemented based on global_init

NK-Nikunj:
## Proposed Changes
This PR shows a very basic implementation of the GSoC project "A C++ Runtime Replacement".

The example is working and after building an executable is created by the name of global_factorial ( to keep it different from factorial, which is already an example in quickstart ) in the bin directory.


hkaiser:
@NK-Nikunj I thought about this a bit more. While a global object is a very elegant solution, it will not provide us with a seamless integration of HPX into arbitrary existing applications. For that we would need to invoke the C-main function on an HPX thread (very much what happens now if `#include <hpx/hpx_main.hpp>` is used - but more elegant. Any ideas you might have would be appreciated.

NK-Nikunj:
@hkaiser I am investigating ways through which we could invoke the C-main function on an HPX thread.

NK-Nikunj:
@hkaiser I have implemented the runtime system with following features:-
  - main now runs as an HPX thread. This means that all hpx components can be called in main
  - Things now run without getting into hpx_main
  - User now has to call hpx::complete() to start the termination of the program
  - After correcting initialization sequence, everything will run seamlessly

hkaiser:
@NK-Nikunj could you address the reported inspect problems, please (see: https://9795-4455628-gh.circle-artifacts.com/0/tmp/circle-artifacts.E1deIzW/hpx_inspect_report.html)?

hkaiser:
Unfortunately there are still some tabs in those files: https://9800-4455628-gh.circle-artifacts.com/0/tmp/circle-artifacts.nkLjCos/hpx_inspect_report.html

examples/global_init/hpx_initialize.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+//////////////////////////////////////////////////////////////////////////////
+// This implemented function calls the destruct object to start the exit
+// sequence
+namespace hpx {
+    void complete() {
+        // Finalize everything
+        destruct destroy;
+    }
+}
```
hkaiser: Why is this necessary? Wouldn't invoking hpx::finalize() from the main function of the application trigger the destruction of the global object?

NK-Nikunj: @hkaiser I was not aware if hpx::finalize() trigger destruction of the global objects as well. So, I created an alternative way of hpx::complete() which would first destruct runtime system then de-initialize the initialization sequences that will be implemented in manage_global_runtime.

hkaiser: `hpx::finalize()` informs the runtime that it is safe to exit. This should eventually cause the runtime to exit and the main application should terminate. At this point the system calls the destructors of all global objects (after the C-main function returned).

NK-Nikunj: ok, so calling hpx::finalize() in main itself should work properly. I'll make amendments to my code.

hkaiser: Also, wasn't the goal of the whole example to show that the user does not need to invoke any special functionality in order to use HPX?

hkaiser: What does this example do differently compared to the existing example `init_globally`?

NK-Nikunj: This example runs main directly as hpx thread without using hpx_main. This gives main function complete functionalities of an hpx thread i.e. futures and async can run without giving null id runtime error. In init_globally you could not get all functionalities of hpx thread in main, I received null id error for futures and async. This is because main is registered as hpx thread, so only a few components can be used.
Furthermore, I felt that this implementation is better since it completely drops the idea of hpx_main, which was the key aspect of my GSoC project (i.e. running things directly from main without going through a special function).

NK-Nikunj:
Closing this pr wrt #3357 which is a better implementation as compared to the given pr.

PR #3234, opened 2018-03-12
Disable background thread when networking is off

msimberg:
## Proposed Changes

- Disable the background thread completely in `scheduling_loop` if `HPX_HAVE_NETWORKING=OFF`
- The flag can still be set but has no effect
- This is unlikely to give a big speedup in any application, but it's nice to disable as the background thread doesn't do anything useful with networking off

@biddisco Might give you a *tiny* speedup for Cholesky, unless you've already disabled the background thread.

WIP as I will still run the tests with pycicle and networking off.

### DIFF
--- a/cmake/HPX_AddTest.cmake
+++ b/cmake/HPX_AddTest.cmake
@@ -44,8 +44,13 @@ macro(add_hpx_test category name)
           "${CMAKE_BINARY_DIR}/bin/hpxrun.py"
           ${_exe}
           "-e" "${expected}"
-          "-l" "${${name}_LOCALITIES}"
           "-t" "${${name}_THREADS_PER_LOCALITY}")
+
+  if(HPX_WITH_NETWORKING)
+      list(APPEND cmd "-l" "${${name}_LOCALITIES}")
+  else()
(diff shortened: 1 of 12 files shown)


msimberg:
FYI, this timed out on pycicle so I'll take this as an opportunity to check what breaks with networking off (or what tests etc. should be disabled, if they're not already disabled). Consider this low-priority exploration.

hkaiser:
@msimberg Thanks! And yes, tests will be a problem - we never really made this consistent.

msimberg:
This ran successfully with `HPX_HAVE_NETWORKING=OFF` now. I will keep a pycicle instance running with networking off for master only.

I've
- forced one locality (in CMake) for tests if networking is off
- ignored networking related options in `hpxrun.py` if localities is 1 (as they're disabled in HPX)
- completely disabled some tests that checked that the test is running with more than one locality
- done some small adjustments to tests that use multiple localities but don't require it (they use e.g. `find_remote_localities` and will do almost nothing since the vector is empty, but I kept them enabled as they should still compile correctly)

PR #3163, opened 2018-02-10
libfabric parcelport w/o HPX_PARCELPORT_LIBFABRIC_ENDPOINT_RDM

ct-clmsn:
HPX_PARCELPORT_LIBFABRIC_ENDPOINT_RDM fixes.

Fixes #2650 

## Any background context you want to provide?

platforms without LIBFABRIC_ENDPOINT_RDM turned on have compile time errors. this patch resolves compile issues. this patch also fixes an issue where the passive endpoint is not turned into an active endpoint for IPC. without this fix active endpoints are never created; active endpoints are created only in blocks of code with the HPX_PARCELPORT_LIBFABRIC_ENDPOINT_RDM #define guard. active endpoints are necessary for IPC.

PR #3235, opened 2018-03-13
Fixing Libfabric Parcelport

sithhell:
This patch makes the libfabric parcelport compile and run again after the recent changes


hkaiser:
@sithhell is this related to #3163?

sithhell:
No, this change makes the non RDM libfabric parcelport work again. There where multiple compiler errors related to the yield_while changes and the PP internal logging. #3163 is fixing a problem we have with active endpoints.

PR #3238, opened 2018-03-13
Readding accidently removed std::abort

sithhell:
This abort was removed with #3188.

Without this change, applications can not be stopped by for example hitting ctrl+c or when an exception occurs.

### DIFF
--- a/src/runtime.cpp
+++ b/src/runtime.cpp
@@ -59,7 +59,7 @@
 namespace hpx
 {
     ///////////////////////////////////////////////////////////////////////////
-    HPX_NORETURN void handle_termination(char const* reason)
+    void handle_termination(char const* reason)
     {
         if (get_config_entry("hpx.attach_debugger", "") == "exception")
         {
@@ -134,6 +134,7 @@ namespace hpx
                 << "{what}: " << (reason ? reason : "Unknown signal") << "\n"
                 << full_build_string();           // add full build information
         }
+        std::abort();
     }
 }
 


sithhell:
Yes, my use case is a application running in a terminal that I want to end
with Ctrl+C. This is currently broken. As well as applications with
segfaults (or other signals). They go on forever.

Am 13.03.2018 5:17 nachm. schrieb "Hartmut Kaiser" <notifications@github.com
>:

*@hkaiser* commented on this pull request.

These std::abort calls were removed on purpose. The rationale is that the
application is going to be terminated at that point anyways. Do you have a
use case where those are needed?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/STEllAR-GROUP/hpx/pull/3238#pullrequestreview-103504784>,
or mute the thread

hkaiser:
> Yes, my use case is a application running in a terminal that I want to end with Ctrl+C. This is currently broken. As well as applications with segfaults (or other signals). They go on forever.

Ok, then this should be re-added for non-windows systems only as there we're sure that the process is being terminated once we reach that point in the execution.

sithhell:
Really? If the termination handler doesn't abort, the process won't continue to run? That sounds odd to me ...
I guess what we rather want to have is to move the abort into the if statement to not cause the python application to abort in that case?

hkaiser:
On Windows, if the termination handler returns TRUE, then no other handler functions are called and the system terminates the process (see: https://docs.microsoft.com/en-us/windows/console/handlerroutine).

sithhell:
Thanks @hkaiser for the explanation. I pushed a new version of the changes removing `std::abort` for windows.

PR #3239, opened 2018-03-13
Changing std::rand() to a better inbuilt PRNG generator.

### LINKED ISSUE #2954: Use of `std::rand` has UB risk in tests

Anushi1998:
Fixes #2954 
This PR is in continuation of PR #3204.

### DIFF
--- a/tests/unit/parallel/algorithms/copyif_random.cpp
+++ b/tests/unit/parallel/algorithms/copyif_random.cpp
@@ -12,12 +12,16 @@
 #include <iostream>
 #include <iterator>
 #include <numeric>
+#include <random>
 #include <string>
 #include <vector>
 
 #include "test_utils.hpp"
 
 ////////////////////////////////////////////////////////////////////////////
+unsigned int seed = std::random_device{}();
+std::mt19937 gen(seed);
+
 template <typename ExPolicy>
 void test_copy_if(ExPolicy policy)
 {
@@ -32,7 +36,7 @@ void test_copy_if(ExPolicy policy)
     std::vector<int> c(10007);
     std::vector<int> d(c.size());
     auto middle = std::begin(c) + c.size()/2;
-    std::iota(std::begin(c), middle, std::rand());
+    std::iota(std::begin(c), middle, gen());
     std::fill(middle, std::end(c), -1);
 
     hpx::parallel::copy_if(policy,
@@ -68,7 +72,7 @@ void test_copy_if_async(ExPolicy p)
     std::vector<int> c(10007);
     std::vector<int> d(c.size());
     auto middle = std::begin(c) + c.size()/2;
-    std::iota(std::begin(c), middle, std::rand());
+    std::iota(std::begin(c), middle, gen());
     std::fill(middle, std::end(c), -1);
 
     auto f =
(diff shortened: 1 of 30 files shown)


msimberg:
@Anushi1998 it looks like some tests are now failing. If you click on pycicle daint-3239... details and then the red 3 under Test/Fail you can see which ones failed. In this case it might be that the range of `std::rand` and your new generator are not the same so you might need to do some adjustments to make the tests happy again.

Anushi1998:
@msimberg 
The test failing on files have all in common
`std::iota(std::begin(c), middle, gen());` i.e `gen()` is used only in `std::iota`
Can you help me with why this is failing here and why not in the previous PR?
There are files(in previous PR) `tests/performance/local/foreach_scaling.cpp` like this which had used `std::iota`

Anushi1998:
@msimberg `gen()` is giving random number in range [0,gen.max()] and `gen.max()` is given by (2^w-1) where w is word length.I thought that it is not exceeding INT_MAX (since we also have `mt19937_64`)(so I tried `uniform_int_distribution<>(0,gen.max()-100007)`) but I just checked that it(`gen.max()`) was actually overflowing Integer!
So I better will make the upper limit to INT_MAX!

msimberg:
Oh, even better. Thanks for correcting me! You seem to have gotten the hang of it. You might also find `std::numeric_limits` useful (instead of `INT_MAX`).

Anushi1998:
Yeah, Sure! I was actually checking for any way to pass w in `mt19937` constructor but unfortunately I haven't till now! I will better use `std::uniform_int_distribution`.

tests/unit/parallel/algorithms/copyif_random.cpp
```diff
@@ -21,7 +23,7 @@
 ////////////////////////////////////////////////////////////////////////////
 unsigned int seed = std::random_device{}();
 std::mt19937 gen(seed);
-std::uniform_int_distribution<> dis(0,std::numeric_limits<int>::max());
+std::uniform_int_distribution<> dis(0,boost::integer_traits<int>::max());
```
hkaiser: Why did you go away from `std::numeric_limits`?

Anushi1998: Because Circle CI was failing `violation of Boost min/max guidelines`

hkaiser: Yah, sure. If you look around you'll find this: http://www.boost.org/doc/libs/1_34_0/more/lib_guide.htm (amongst other things). Essentially the Boost guidelines say that in order to avoid `min` or `max` being expanded as a macro you need to accordingly protect all of their uses. In  case of `std::numeric_limits<T>::max()` you would have to write `(std::numeric_limits<T>::max)()`.

Anushi1998: Thanks for correcting me :)
I will fix it right away.

Anushi1998:
@hkaiser I think the _pycicle daint-3239-gcc-6.2.0 Test_ is unrelated because I haven't changed these files.It can be removed probably by restarting tests.What's your opinion?

Anushi1998:
@msimberg @hkaiser Please review this PR and let me know if I missed something :)

msimberg:
Thanks @Anushi1998!

PR #3241, opened 2018-03-14
WIP Guided pool executor

biddisco:
Fixes #
This adds new executor/scheduler features such as scheduler_hints that can be passed from executors to control how tasks are placed by the scheduler (numa domains in this first version)

## Proposed Changes
New executor types that pass task arguments through a custom user defined function (templated by the executor type) so that the tasks can be queried for information (such as memory placement) before being scheduled. The query only happens once the args for the task a ready so that data locality (memory addresses of the future contents) can be obtained when they are valid.

This PR contains some commits that need to be moved onto other branches, but are included here for testing purposes on my home setup.


msimberg:
Closing this in favor of #3415 and #3420.

PR #3186, opened 2018-02-20
Remove hierarchy, periodic priority and throttling schedulers

msimberg:
## Proposed Changes

- Remove hierarchy, periodic priority and throttling schedulers
- Clean up documentation for command line options related to schedulers
  - all affinity options can be used with all schedulers
  - remove short options for schedulers from documentation (I'm thinking these should be completely disallowed, or be explicitly set, but I'm not sure yet so leaving them for now. Opinions?)
- Remove `ensure_hwloc` function (missed this when making hwloc compulsory)

PR #3146, opened 2018-02-06
Changing the coroutine implementations to do a lazy init

sithhell:
This changes the context implementations to perform a lazy initialization
of their respective contexts. This potentially leads to less resource usage.

PR #3243, opened 2018-03-16
Remove wait or add new

biddisco:
This is a cleanup of the scheduling system to remove the work_items from thread_queue and wait_or_add_new.
It relies on #3146 (lazy init) and also #3186 (remove schedulers) which ought to be merged or closed at the same time (those two PRs are included in this one).

## Proposed Changes
Removes work_items and wait_or_add_new - improves performance of scheduling_loop  -

After this, stack reuse needs to be examined


hkaiser:
If the staged threads are not a thing anymore, then the related performance counters (including their registration, implementation, and documentation) need to be removed as well. 

Same is true for the corresponding thread state.

A note in the release notes ('What's New' in the docs) saying that the thread state and those counters are gone might be appropriate as well.


Some pointers (not a complete list):

Docs:
https://github.com/STEllAR-GROUP/hpx/blob/remove_wait_or_add_new/docs/whats_new.qbk#L76
https://github.com/STEllAR-GROUP/hpx/blob/remove_wait_or_add_new/docs/manual/existing_performance_counters.qbk#L733
etc.

Configuration:
https://github.com/STEllAR-GROUP/hpx/blob/remove_wait_or_add_new/src/util/runtime_configuration.cpp#L246-L247

Thread state:
https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/runtime/threads/thread_enums.hpp#L39-L42
https://github.com/STEllAR-GROUP/hpx/blob/master/src/runtime/threads/threadmanager.cpp#L241
 
Counters:
https://github.com/STEllAR-GROUP/hpx/blob/master/src/runtime/threads/threadmanager.cpp#L1569-L1577
https://github.com/STEllAR-GROUP/hpx/blob/master/src/runtime/threads/threadmanager.cpp#L1745-L1755
https://github.com/STEllAR-GROUP/hpx/blob/master/src/runtime/threads/threadmanager.cpp#L1745-L1755
https://github.com/STEllAR-GROUP/hpx/blob/master/src/runtime/threads/threadmanager.cpp#L1745-L1755
etc.

biddisco:
> Same is true for the corresponding thread state.

The thread state still exists, it just isn't necessary to store staged threads in a separate queue

msimberg:
This looks very nice, but I'll have a closer look once the other PRs are merged and the conflicts are resolved (when the diff is a bit smaller).

biddisco:
Unfortunately, this branch has been rebased so many times in the last year that at some point, something got messed up. I'm looking at it from fresh to try to unpick what's the right code to have/use.

hkaiser:
@biddisco let's get this in asap - it's the only way.

hkaiser:
@biddisco I think we can close this now as we all agree that `wait_or_add_new` should not be completely removed. Additionally, many of the issues that were intended to be solved by this have been addressed by #3768. Please reopen, if needed

PR #3245, opened 2018-03-19
Apex refactoring with guids

khuck:
- Adds apex_new_task() call when HPX threads are constructed.  The resulting object pointer (apex::task_wrapper* cast as a void*) is stored in the thread.
  - Updates the apex start/yield/stop calls to take the apex::task_wrapper object pointer.  The apex::stop call will destroy the object, so HPX doesn't have to manage that memory.
  - Removes the start/stop around the background thread - the measurement of that thread actually is handled when the work is scheduled in the regular scheduler.

This pull request updates the APEX/HPX integration to track task dependencies, generate GUIDs within APEX for tracking tasks, and adds those GUIDs to the OTF2 trace output.  The version of APEX that cmake will pull has been modified from "master" to "v2.0", a new tag in APEX representing these API changes.  v2.0 is currently a "beta" release, if all goes well it will become an official release soon.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1539,7 +1539,7 @@ if(HPX_WITH_APEX)
   include(GitExternal)
   git_external(apex
     https://github.com/khuck/xpress-apex.git
-    generate_guids
+    v2.0
     ${_hpx_apex_no_update}
     VERBOSE)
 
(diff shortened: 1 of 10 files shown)


msimberg:
All right, I'll go ahead and add this to the release. We won't really know if something breaks as rostam and pycicle have Apex off, but this looks fairly harmless. Thanks @khuck!

PR #2911, opened 2017-09-20
Fixing CUDA problems

sithhell:
This fixes various issues related to building HPX applications with either nvcc or cuda-clang

PR #3246, opened 2018-03-20
Assorted fixes for CUDA

gentryx:
Fixes #2911

### DIFF
--- a/hpx/lcos/detail/async_implementations.hpp
+++ b/hpx/lcos/detail/async_implementations.hpp
@@ -368,7 +368,8 @@ namespace hpx { namespace detail
                             id, std::move(addr), std::forward<Ts>(vs)...);
                     } else {
                         f = hpx::async(action_invoker<action_type>(),
-                                       addr.address_, addr.type_, std::forward<Ts>(vs)...);
+                                       addr.address_,
+                                       addr.type_, std::forward<Ts>(vs)...);
 
                         return keep_alive(std::move(f), id, std::move(r.second));
                     }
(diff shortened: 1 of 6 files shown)


hkaiser:
Why is adding an `else` case solving a compiler warning? I don't get that.

gentryx:
NVCC was complaining that the trailing return statements were unreachable. I guess this is because nvcc is a little over zealous for analyzing const expressions? I agree that it doesn't make much sense, it doesn't hurt either, though.

gentryx:
Huh, any idea why CircleCI would fail? This particular issue (docker login failing) doesn't seem to be related to my change.

hkaiser:
@gentryx it decided to run on your circleci account (no idea why), so it can't access our (CI) environment variables (like the docker credentials).

msimberg:
I don't mind merging this, but it does undo some clang-tidy warnings that were fixed in #3154, which will come back next time someone decides to run clang-tidy. Being able to successfully compile is more important than being clang-tidy though.

What version of NVCC were you using? How much of HPX did you try to compile?

gentryx:
Necessary for compilation is only the change to hpx/runtime/components/server/managed_component_base.hpp.

The other changes are just to make nvcc compile without issuing tons of warnings. So, I should be good if someone ran clang-tidy in the future. :-)

gentryx:
I'm using CUDA 9.1.85, I was compiling from HPX only what's required by LibGeoDecomp. Not sure which parts that are exactly.

gentryx:
PTAL

hkaiser:
@gentry: @msimberg is the master of ceremonies for the upcoming release - he will decide whether this should go in or not.

msimberg:
Thanks! Will add this to the release.

gentryx:
Thanks! :-)

PR #3249, opened 2018-03-21
Updating People.qbk

aserio:
- Adding new contributors to the Acknowledgement section


hkaiser:
@msimberg this should be added to the release, I think.

msimberg:
Thank you @aserio!

PR #3250, opened 2018-03-22
Apex refactoring with guids

khuck:
This commit adds parent dependencies for APEX measurement.  The parent dependencies will now be in the OTF2 output as well as the APEX taskgraph output.


hkaiser:
@msimberg I think this is a leftover from the previous patch you have merged. Do you think this could still go into the release?

msimberg:
Thanks! I will add it to the release.

PR #3254, opened 2018-03-24
Adding trait for actions allowing to make runtime decision on whether to execute it directly

hkaiser:


hkaiser:
The reported failures are unrelated to the change at hands.

PR #3261, opened 2018-03-26
Apex refactoring fix rebind

khuck:
This fixes cases where HPX threads go through the rebind_base() process, and the APEX data pointer is no longer valid.  In those cases, we update the APEX data pointer with the new task information.  That code is added in src/runtime/threads/coroutines/detail/context_base.cpp because it requires access to members of thread_id_type and thread_data, which are not available in context_base.hpp.


hkaiser:
@khuck something is not quite right with this patch, it caused (almost) all tests to fail running on daint. Could you have another look, please?

khuck:
@hkaiser I'll take a look when I can.  I am surprised that it failed, since that build would have been without APEX, I think.

msimberg:
Daint has been slightly unreliable the last few days/weeks (slurm timeouts etc.). I have a feeling the failures are unrelated and would maybe wait for a rerun to see if the failures are still there.

PR #3265, opened 2018-03-28
Split parallel tests on CircleCI

msimberg:
Splits up parallel tests into separate steps as they take nearly an hour to run (all others take less than 25 minutes with most taking less than 10 minutes). This obviously does not increase throughput but does give faster feedback for tests that are ready.

We'll also more easily see failures in non-partitioned_vector parallel tests as they're not in the same stage as the partitioned_vector tests.

## Proposed Changes

- Split up parallel tests on CircleCI
- Move the block tests directly in the tests/unit/parallel folder into a subfolder (makes the CircleCI config slightly easier to handle)


msimberg:
I put the subtests under tests.headers.parallel instead of tests.unit.parallel, will fix tomorrow.

msimberg:

PR #3266, opened 2018-03-28
Replace boost::format with custom sprintf-based implementation

K-ballo:
The format replacement specification grammar takes the form: `'{' [arg] [':' spec] '}'`.
- The first field, `arg`, is a numeric value and it specifies the 1-based index of the argument to use in the replacement (the 0-th argument is the format string itself). If not present it is derived from the index of the replacement within the format string.
- The second field, `spec`, is an arbitrary value and it specifies the format to use in the replacement. The meaning depends on the type of the value under replacement:
  - For fundamental types, pointers, and strings, the format specifier follows the format of the `printf` family, except that the type conversion specifier is optional and is derived from the type of the value under replacement if not present.
  - For types that opt into custom formatting via the ADL customization point, the meaning and interpretation of the format specifier is entirely up to user defined overload.
  - For every other type the format specification must be empty.

The ADL customization point signature is:
```
template <typename T>
void format_value(std::ostream& os, boost::string_ref spec, T const& value);
```

### DIFF
--- a/examples/1d_stencil/1d_stencil_4_throttle.cpp
+++ b/examples/1d_stencil/1d_stencil_4_throttle.cpp
@@ -42,7 +42,7 @@ using hpx::performance_counters::stubs::performance_counter;
 using hpx::performance_counters::counter_value;
 using hpx::performance_counters::status_is_valid;
 static bool counters_initialized = false;
-static const char * counter_name = "/threadqueue{locality#%d/total}/length";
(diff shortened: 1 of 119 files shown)


K-ballo:
> I'd assume this is modelled after one of the standardization papers?

P0645, loosely, yes. I decided not to deviate from `printf` style format specification.

Issue #3053, opened 2017-12-06
segmented algorithm tests take too long to compile


biddisco:
Compilation of the segmented algorithm tests takes a very long time, uses too much memory and has caused nodes to fail when I use `make -j` by mistake.

We should make segmented algorithms (or just their tests) optional until such time as they can be made more user friendly.

aserio:
@biddisco How long is a long time? I have been doing some thinking about this and I believe that we intend to develop more APIs on top of the segmented algorithms. Therefore, we should at least continue to make sure they are integrated properly into the library.

sithhell:
Am 07.12.2017 11:54 nachm. schrieb "Adrian Serio" <notifications@github.com
>:

@biddisco <https://github.com/biddisco> How long is a long time? I have
been doing some thinking about this and I believe that we intend to develop
more APIs on top of the segmented algorithms. Therefore, we should at least
continue to make sure they are integrated properly into the library.


I don't have the raw numbers. But they don't only take a long time
(minutes) but also a ridiculous amount of memory (several gigabytes). The
compilation is severely broken in that regard.
Before you think about developing new algorithms on top of it, this needs
to be fixed.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/STEllAR-GROUP/hpx/issues/3053#issuecomment-350119560>,
or mute the thread

sithhell:
As a very rough point of measure, on circleci, the parallel unit tests take 1h8m to build. All the remaining ones take 34m. A significant amount of time is spent with building the segmented algorithms tests.

aserio:
So we want to add a flag so that users can disable the parallel algorithm unit tests?

PR #3273, opened 2018-04-06
Splitting tests to avoid compiler OOM

### LINKED ISSUE #3272: CircleCI does not upload generated docker image any more
The new CircleCI V2 scripts do not upload the generated docker image anymore as some of the parallel algorithm tests fail to compile (OOM problems?). This causes all kind of problems for projects depending on the latest HPX master features.

Would it be possible to upload the docker file whenever the core libraries compile successfully (at least until the issue with the parallel algorithms has been resolved)?

hkaiser:
Fixes #3272

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -309,9 +309,16 @@ jobs:
               make -j1 tests.unit.parallel.segmented_algorithms.partitioned_vector_for_each
               make -j1 tests.unit.parallel.segmented_algorithms.partitioned_vector_transform
               make -j1 tests.unit.parallel.segmented_algorithms.partitioned_vector_transform_binary
+              make -j1 tests.unit.parallel.segmented_algorithms.partitioned_vector_transform_binary1
+              make -j1 tests.unit.parallel.segmented_algorithms.partitioned_vector_transform_binary2
+              make -j1 tests.unit.parallel.segmented_algorithms.partitioned_vector_transform_binary3
               make -j1 tests.unit.parallel.segmented_algorithms.partitioned_vector_exclusive_scan
+              make -j1 tests.unit.parallel.segmented_algorithms.partitioned_vector_inclusive_scan
+              make -j1 tests.unit.parallel.segmented_algorithms.partitioned_vector_inclusive_scan2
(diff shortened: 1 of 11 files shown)


hkaiser:
This also fixes #3053

msimberg:
Very nice! Now the problem is the install step:

```
Concurrent upstream jobs persisted the same file(s) into the workspace:
  - tests/headers/CMakeFiles/runtime_runtime_mode_hpp_lib.dir/flags.make
  - tests/headers/CMakeFiles/runtime_parcelset_fwd_hpp_lib.dir/link.txt
  - tests/unit/parallel/algorithms/CMakeFiles/
...
```

This is because the test steps all attach to the same workspace, persist their workspaces, and the install step then tries to merge them. This should fail officially as well (according to [this](https://circleci.com/blog/deep-diving-into-circleci-workspaces/)). We could try deleting those files in all but one of the parallel steps... @sithhell have you encountered this problem with other projects? How badly do we want parallel builds?

hkaiser:
@msimberg @sithhell could this be resolved with high priority please? We really need an updated docker image for our other project testing environment asap.

msimberg:
@hkaiser in that case I will try to make everything sequential at least as a temporary solution (it *should* work, we'll see). Then we can play around again to make it run in parallel.

msimberg:
But we still need this one so I will go ahead and merge this. Thanks!

hkaiser:
@msimberg Wouldn't it be sufficient to single out one of the steps to run after everything has finished to do the installation?

msimberg:
@hkaiser hmm, yes, it won't be a full install then but I suppose that's enough for you?

hkaiser:
@msimberg I'd need a full install, but your suggestion sound like a possible solution.

msimberg:
I need to think before I write on a Friday afternoon. `make install` depends on examples but not tests. Let's see what happens in #3275.

sithhell:
The parallel workspace merge is certainly a problem. It is a little unclear to me why the failure happens, since only the examples persist files to the workspace (or should, at least)... I think in general, it is a good idea to only upload a new docker image once the tests pass. Keeps us motivated to fix the remaining spurious failures ;)

PR #3277, opened 2018-04-07
Support external plugins

hkaiser:
A couple of minor changes allowing for external libraries to load their own plugins through HPX's plugin infrastructure. 

These are changes required by Phylanx.

### DIFF
--- a/hpx/plugins/plugin_registry.hpp
+++ b/hpx/plugins/plugin_registry.hpp
@@ -38,12 +38,10 @@ namespace hpx { namespace plugins
     /// functions to be exposed by a plugin's registry instance.
     ///
     /// \tparam Plugin   The plugin type this registry should be responsible for.
-    template <typename Plugin>
+    template <typename Plugin, char const* const Name,
+        char const* const Section, char const* const Suffix>
     struct plugin_registry : public plugin_registry_base
     {
-        ///
-        ~plugin_registry() {}
-
         /// \brief Return the ini-information for all contained components
         ///
         /// \param fillini  [in] The module is expected to fill this vector
@@ -54,14 +52,14 @@ namespace hpx { namespace plugins
         /// \return Returns \a true if the parameter \a fillini has been
         ///         successfully initialized with the registry data of all
         ///         implemented in this module.
-        bool get_plugin_info(std::vector<std::string>& fillini)
+        bool get_plugin_info(std::vector<std::string>& fillini) override
         {
             using namespace boost::assign;
-            fillini += std::string("[" HPX_PLUGIN_CONFIG_SECTION ".plugins.") +
+            fillini += std::string("[") + Section + ".plugins." +
                 unique_plugin_name<plugin_registry>::call() + "]";
-            fillini += "name = " HPX_PLUGIN_STRING;
-            fillini += std::string("path = ") + util::find_prefixes(
-                    "/" HPX_PLUGIN_PATH_SUFFIX, HPX_PLUGIN_STRING);
+            fillini += std::string("name = ") + Name;
+            fillini += std::string("path = ") +
+                util::find_prefixes(std::string("/") + Suffix, Name);
             fillini += "enabled = 1";
 
             char const* more = traits::plugin_config_data<Plugin>::call();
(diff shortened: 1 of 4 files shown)


hpx/plugins/plugin_registry.hpp
```diff
@@ -57,11 +57,11 @@ namespace hpx { namespace plugins
...
         {
             using namespace boost::assign;
-            fillini += std::string("[hpx.plugins.") +
+            fillini += std::string("[" HPX_PLUGIN_CONFIG_SECTION ".plugins.") +
                 unique_plugin_name<plugin_registry>::call() + "]";
             fillini += "name = " HPX_PLUGIN_STRING;
-            fillini += std::string("path = ") +
-                util::find_prefixes("/hpx", HPX_PLUGIN_STRING);
+            fillini += std::string("path = ") + util::find_prefixes(
+                    "/" HPX_PLUGIN_PATH_SUFFIX, HPX_PLUGIN_STRING);
```
sithhell: Wouldn't it be better to have `HPX_PLUGIN_CONFIG_SECTION` and `HPX_PLUGIN_PATH_SUFFIX` as a parameter to the plugin registry instead? Making it relying on the global `config.hpp` header looks error prone. The following should work as well: https://wandbox.org/permlink/4emy0DIwSOzxmSjs

And provides a much cleaner interface for the module declarations.

hkaiser: This will break NVCC most likely, but we might be able to keep this code out of view of this compiler.

hkaiser: Done.

hpx/plugins/plugin_registry.hpp
```diff
@@ -89,15 +87,32 @@ namespace hpx { namespace plugins
...
+    typedef hpx::plugins::plugin_registry<PluginType, __##pluginname##_string, \
+        __##pluginname##_section, __##pluginname##_suffix>                     \
+        __##pluginname##_plugin_registry_type;                                 \
+    HPX_REGISTER_PLUGIN_BASE_REGISTRY(                                         \
+        __##pluginname##_plugin_registry_type, pluginname)                     \
+    HPX_DEF_UNIQUE_PLUGIN_NAME(                                                \
+        __##pluginname##_plugin_registry_type, pluginname)                     \
+    template struct hpx::plugins::plugin_registry<PluginType,                  \
+        __##pluginname##_string, __##pluginname##_section,                     \
+        __##pluginname##_suffix>;                                              \
```
sithhell: Should we rather use `HPX_PP_CAT` to concatenate the PP tokens?

hkaiser: This is only necessary if the parameters `pluginname` is a macro itself.. Not sure if it's worth the trouble.

sithhell:
http://cdash.cscs.ch/testDetails.php?test=913265&build=6076

I think this test failure is related.

hkaiser:
> http://cdash.cscs.ch/testDetails.php?test=913265&build=6076
> I think this test failure is related.

Yes, this is related but was fixed by the latest commit.

hpx/plugins/plugin_registry.hpp
```diff
@@ -89,15 +87,32 @@ namespace hpx { namespace plugins
...
-    template struct hpx::plugins::plugin_registry<PluginType >;               \
-/**/
+#define HPX_REGISTER_PLUGIN_REGISTRY_2(PluginType, pluginname)                 \
+    HPX_REGISTER_PLUGIN_REGISTRY_5(                                            \
+        PluginType, pluginname, HPX_PLUGIN_NAME, "hpx", "hpx")                 \
+    /**/
+#define HPX_REGISTER_PLUGIN_REGISTRY_4(                                        \
+    PluginType, pluginname, pluginsection, pluginsuffix)                       \
+    HPX_REGISTER_PLUGIN_REGISTRY_5(PluginType, pluginname, HPX_PLUGIN_NAME,    \
+        pluginsection, "pluginsuffix")                                         \
```
sithhell: why is `pluginsuffix` quoted here?

hpx/plugins/plugin_registry.hpp
sithhell: Ok, we'll see how it goes.

hpx/plugins/plugin_registry.hpp
hkaiser: Yep, caught that yesterday as well.... Should be fine now.

hkaiser:
@sithhell canm we go ahead with this now?

PR #3283, opened 2018-04-13
Do not unwrap ready future

hkaiser:
- if the outer future of a future<Future> is ready we don't need to allocate a new
  shared state while unwrapping


hkaiser:
This breaks one of our applications (the application hangs randomly). I'm going to revert this change for now.

PR #3284, opened 2018-04-13
Background thread can use relaxed atomics to manipulate thread state

hkaiser:


hkaiser:
The reported errors are related to daint and not related to the proposed changes.

sithhell:
The errors do indeed look to be unrelated, but not special to daint. rostam is showing the same problems.

PR #3289, opened 2018-04-15
Fixing Docker image creation

sithhell:
- Using multi-stage Dockerfil
 - Removing docker-compile.pl

Flyby: removing parallel performance tests from 'all' target


hkaiser:
> Flyby: removing parallel performance tests from 'all' target

Why's that?

hkaiser:
> > > Flyby: removing parallel performance tests from 'all' target
> >
> > Why's that?
>
> Because we don't install them, and in the workflow, they were rebuilt during the install step

Are those still built, at least?

sithhell:
Sure, here: https://circleci.com/gh/STEllAR-GROUP/hpx/12827

PR #3292, opened 2018-04-17
Add new cuda kernel synchronization with hpx::future demo

biddisco:
This example shows how to launch a kuda kernel and bind it to
an hpx::future so that continuations may be attached to it
in the same way as a CPU task.

A new helper class is introduced that can be used with the old
cuBlas +future example code. This commit unifies the two to make a
cleaner integration.

The cuda_future_helper might benefit from being moved into the main
hpx::compute tree rather than being in examples.


hkaiser:
Let's go ahead with this. We can always move the code into the core, if needed.

Issue #3297, opened 2018-04-23
Crash with APEX when running Phylanx lra_csv with > 1 thread


khuck:
## Expected Behavior

I don't expect it to crash.

## Actual Behavior

HPX is crashing in the rebind_task method, when APEX is setting a new task wrapper for the re-used thread_data object.  The parent_thread_id_ 's m_apex_data member is corrupted. 

## Steps to Reproduce the Problem

Configure HPX with these additional APEX settings:
```
-DHPX_WITH_APEX=TRUE \
-DAPEX_WITH_ACTIVEHARMONY=TRUE \
-DACTIVEHARMONY_ROOT=/usr/local/activeharmony/4.6 \
-DAPEX_WITH_OTF2=TRUE -DOTF2_ROOT=/usr/local/otf2/2.1 \
-DAPEX_WITH_PAPI=TRUE -DPAPI_ROOT=/usr/local/papi/5.5.0/ \
-DHPX_WITH_APEX_NO_UPDATE=TRUE
```
...then build the Phylanx software and run the lra_csv example with more than 1 HPX worker thread.  Higher thread counts will crash more reliably, and consistently with >4 threads.

## Specifications

HPX master, Apple Clang++ 9.1.0 (Xcode), OSX 10.13.4

biddisco:
@khuck I'm getting problems with the code after the recent merge. Specifically there's an 
`assert(wrapper);` call in update_task that is being triggered. I can comment it out and all seems well until it hangs on shutdown. Didn't get a chance to diagnose it.
Reopening this issue for now.

PR #3298, opened 2018-04-25
Apex task wrapper memory bug

### LINKED ISSUE #3296: Building HPX with APEX configuration gives compiler warnings
## Expected Behavior

I don't expect compiler warnings.

## Actual Behavior

The compiler is warning about unused variables in the annotated_function() constructors.  This might be related to annotated_function issues

Configure HPX with these additional APEX related config options:
```
-DHPX_WITH_APEX=TRUE -DAPEX_WITH_ACTIVEHARMONY=TRUE \
-DACTIVEHARMONY_ROOT=/usr/local/activeharmony/4.6 \
-DAPEX_WITH_OTF2=TRUE -DOTF2_ROOT=/usr/local/otf2/2.1 \
-DAPEX_WITH_PAPI=TRUE -DPAPI_ROOT=/usr/local/papi/5.5.0/ \
-DHPX_WITH_APEX_NO_UPDATE=TRUE
```

## Specifications

This happens with Apple Clang++ 9.1.0 (Xcode) on OSX 10.13.4

khuck:
Fixes #3296 and #3297

## Proposed Changes

  - parent/child dependency is captured when the thread_init_data object is constructed (and the parent is still alive)
  - the apex_data pointer is transferred to the hpx_thread object when the thread is constructed or rebound
  - all dependencies are now correct, and HPX doesn't crash.

This is to support task dependency tracking in HPX and Phylanx.  Currently, this will fix dependencies in OTF2 files completely.  Unfortunately, annotated functions are double-counted when capturing a task graph in APEX, but that is an APEX problem and will soon be fixed.  I also fixed a few compiler warnings.

### DIFF
--- a/hpx/runtime/threads/thread_init_data.hpp
+++ b/hpx/runtime/threads/thread_init_data.hpp
@@ -13,6 +13,7 @@
 #include <hpx/runtime/threads/thread_enums.hpp>
 #include <hpx/runtime/threads_fwd.hpp>
 #include <hpx/util/thread_description.hpp>
+#include <hpx/util/apex.hpp>
 
 #include <cstddef>
 #include <cstdint>
@@ -38,6 +39,11 @@ namespace hpx { namespace threads
 #if defined(HPX_HAVE_THREAD_PARENT_REFERENCE)
             parent_locality_id(0), parent_id(nullptr), parent_phase(0),
 #endif
+#ifdef HPX_HAVE_APEX
+        /* HPX_HAVE_APEX forces the HPX_HAVE_THREAD_DESCRIPTION
+         * and HPX_HAVE_THREAD_PARENT_REFERENCE settings to be on */
+            apex_data(apex_new_task(description,parent_id)),
+#endif
             priority(thread_priority_normal),
             num_os_thread(std::size_t(-1)),
             stacksize(get_default_stack_size()),
@@ -55,6 +61,11 @@ namespace hpx { namespace threads
 #if defined(HPX_HAVE_THREAD_PARENT_REFERENCE)
             parent_locality_id(rhs.parent_locality_id), parent_id(rhs.parent_id),
             parent_phase(rhs.parent_phase),
+#endif
+#ifdef HPX_HAVE_APEX
+        /* HPX_HAVE_APEX forces the HPX_HAVE_THREAD_DESCRIPTION
+         * and HPX_HAVE_THREAD_PARENT_REFERENCE settings to be on */
+            apex_data(apex_new_task(description,parent_id)),
 #endif
             priority(rhs.priority),
             num_os_thread(rhs.num_os_thread),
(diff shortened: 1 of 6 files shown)


hpx/runtime/threads/thread_init_data.hpp
```diff
@@ -81,6 +92,11 @@ namespace hpx { namespace threads
 #endif
 #if defined(HPX_HAVE_THREAD_PARENT_REFERENCE)
             parent_locality_id(0), parent_id(nullptr), parent_phase(0),
```
hkaiser: Shouldn't this rather (deep-)copy the data from the parameter `desc`?

hpx/runtime/threads/thread_init_data.hpp
```diff
@@ -81,6 +92,11 @@ namespace hpx { namespace threads
...
+#endif
+#ifdef HPX_HAVE_APEX
+        /* HPX_HAVE_APEX forces the HPX_HAVE_THREAD_DESCRIPTION
+         * and HPX_HAVE_THREAD_PARENT_REFERENCE settings to be on */
+            apex_data(apex_new_task(description,parent_id)),
```
hkaiser: Shouldn't this rather (deep-)copy the data from the parameter desc?

hpx/runtime/threads/thread_init_data.hpp
```diff
@@ -104,6 +120,11 @@ namespace hpx { namespace threads
         threads::thread_id_type parent_id;
         std::size_t parent_phase;
 #endif
+#ifdef HPX_HAVE_APEX
+        /* HPX_HAVE_APEX forces the HPX_HAVE_THREAD_DESCRIPTION
+         * and HPX_HAVE_THREAD_PARENT_REFERENCE settings to be on */
+        void * apex_data;
```
hkaiser: Where is the memory block returned from `apex_new_task()` deleted?

hpx/runtime/threads/thread_init_data.hpp
khuck: The description is deep-copied in the apex_new_task() call:
https://github.com/STEllAR-GROUP/hpx/blob/c2a3be669721168fbf59c70ef09fc229e285ffa9/src/util/apex.cpp#L22 through line 29

hpx/runtime/threads/thread_init_data.hpp
khuck: It is deleted by APEX in the apex::stop() call when the task_wrapper is no longer needed:
https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/runtime/threads/detail/scheduling_loop.hpp#L586

khuck:
@hkaiser I had to check in a minor change - so the CircleCI tests are re-running.  I thought the over-counting of tasks was related to annotated functions, but it was because the default constructor for thread_init_data() should NOT be capturing a task relationship.  See https://github.com/STEllAR-GROUP/hpx/pull/3298/commits/c2a3be669721168fbf59c70ef09fc229e285ffa9

khuck:
@hkaiser one more comment - don't merge this pull request yet.  I am thinking I might not need the apex_update_task() method at all now.  but I have to test with Phylanx to be sure, and I am waiting on that build.  I may refactor that out with this same pull request.  Stay tuned.

biddisco:
```
/home/biddisco/src/hpx-master/apex/src/apex/apex.cpp: In function ‘void apex::stop(std::shared_ptr<apex::task_wrapper>&)’:
/home/biddisco/src/hpx-master/apex/src/apex/apex.cpp:833:22: error: type ‘class std::shared_ptr<apex::task_wrapper>’ argument given to ‘delete’, expected pointer
         delete(tt_ptr);
                      ^
/home/biddisco/src/hpx-master/apex/src/apex/profiler_listener.cpp: In member function ‘virtual void apex::profiler_listener::on_task_complete(std::shared_ptr<apex::task_wrapper>&)’:
```

biddisco:
This passed all tests - we should merge it quick!

PR #3299, opened 2018-04-26
Performance improvements

hkaiser:
This PR is a collection of performance improvements:

- AGAS: all non-migratable (non-managed) components now store their lva and component type directly in the gid
- AGAS: all migratable component now have a special bit set in their gid
- LCOS: future state is now stored as a `std::atomic<state>`


hkaiser:
More testing is needed. I'll close this PR and create separate ones for each of the things proposed here.

PR #3303, opened 2018-04-27
Changed std::rand to a better inbuilt PRNG Generator

### LINKED ISSUE #2954: Use of `std::rand` has UB risk in tests

Anushi1998:
Fixes #2954
This PR is in continuation of PR #3239 .

### DIFF
--- a/tests/unit/parallel/algorithms/for_loop_induction.cpp
+++ b/tests/unit/parallel/algorithms/for_loop_induction.cpp
@@ -13,6 +13,7 @@
 #include <cstddef>
 #include <iostream>
 #include <numeric>
+#include <random>
 #include <string>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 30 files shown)


hkaiser:
@Anushi1998 what's wrong with `gen`, `generator`, or just `g`?

Anushi1998:
@hkaiser `gen` is used in some [files](https://github.com/Anushi1998/hpx/blob/b44d0d6dfge350f8275d1db9568fe53f2782a7127d/tests/unit/parallel/algorithms/remove_tests.hpp#L100) .
I am comfortable using `generator` and `g`.

Anushi1998:
Apologies for the delay.

Anushi1998:
@hkaiser I couldn't generate artifact for CircleCi error. How can I solve the issue?

hkaiser:
The CircleCI error is unrelated. Please ignore.

Anushi1998:
Sure :)

PR #3315, opened 2018-05-14
WIP: Fix CircleCI config to really push docker image only on master

msimberg:


.circleci/config.yml
```diff
@@ -917,7 +917,8 @@ jobs:
               docker login -u ${DOCKER_USER} -p ${DOCKER_PASS}
               docker push ${TARGET_IMAGE_NAME}
           filters:
-              branches: master
+              branches:
+                  only: master
```
sithhell: Good catch!

msimberg:
This doesn't do what I hoped it would do. I think the filter can't be applied to a single run step, but has to be applied to a whole workflow step. I'll get back to this.

msimberg:
I'll close this and open another when I have time to look at it again.

PR #3316, opened 2018-05-14
Fix includes for 1d_stencil_*_omp examples

msimberg:
Makes sure the `1d_stencil` examples with OpenMP build.

Includes `hpx/config.hpp` before including `boost/program_options.hpp` to avoid this:
```
In file included from /home/simbergm/src/hpx/hpx/util/format.hpp:9:0,
                 from /home/simbergm/src/hpx/examples/1d_stencil/print_time_results.hpp:10,
                 from /home/simbergm/src/hpx/examples/1d_stencil/1d_stencil_1_omp.cpp:28:
/home/simbergm/src/hpx/hpx/config.hpp:14:2: error: #error Boost.Config was included before the hpx config header. This might lead to subtile failures and compile errors. Please include <hpx/config.hpp> before any other boost header
 #error Boost.Config was included before the hpx config header. This might lead to subtile failures and compile errors. Please include <hpx/config.hpp> before any other boost header
```

### DIFF
--- a/examples/1d_stencil/1d_stencil_1_omp.cpp
+++ b/examples/1d_stencil/1d_stencil_1_omp.cpp
@@ -12,7 +12,9 @@
 // The only difference to 1d_stencil_1 is that this example uses OpenMP for
 // parallelizing the inner loop.
 
-#include <hpx/config/defines.hpp>   // avoid issues with Intel14/libstdc++4.4 nullptr
+// Include before any Boost header and avoid issues with Intel14/libstdc++4.4
+// nullptr
+#include <hpx/config.hpp>
 
 #include <boost/program_options.hpp>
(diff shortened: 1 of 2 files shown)


examples/1d_stencil/1d_stencil_1_omp.cpp
```diff
@@ -13,6 +13,7 @@
 // parallelizing the inner loop.
 
 #include <hpx/config/defines.hpp>   // avoid issues with Intel14/libstdc++4.4 nullptr
+#include <hpx/config.hpp> // Include before any boost header
```
hkaiser: The first #include could be removed now. Including `<hpx/config.hpp>` should be sufficient.

PR #3317, opened 2018-05-15
Fix compilation with VS 15.7.1 and /std:c++latest

hkaiser:
MSVC somehow confuses the two templates that have the same name. Renaming one fixes things...


K-ballo:
LGTM

PR #3326, opened 2018-05-21
Make sure logging is up and running while modules are being discovered.

hkaiser:
- add more logging to module discovery
- flyby: don't log hpx::success as error


hkaiser:
@parsa please check again using this branch. This should give you detailed logs for the module discovery in HPX.

parsa:
@hkaiser It fails to compile:
```
c:\repos\hpx\src\exception.cpp(89): error C2678: binary '!=': no operator found which takes a left-hand operand of ty
pe 'const boost::system::system_error' (or there is no acceptable conversion) [C:\Repos\hpx\cmake-build-debug\src\hpx.v
cxproj]
```

PR #3329, opened 2018-05-22
fix dox for creating component instances through 'new_' is

hrasadi:
'new_' will only create server instances, in order to create client objects directly, one should create the component's future using 'new_' and pass it the the copy constructor of the client.

Documentation is updated to reflect this.

### DIFF
--- a/docs/manual/components.qbk
+++ b/docs/manual/components.qbk
@@ -175,22 +175,22 @@ calls of the server side implementation objects corresponding to the
 
     // create one instance on the given locality
     hpx::id_type here = hpx::find_here();
-    client_type c = hpx::new_<client_type>(here, ...);
+    client_type c = hpx::new_<some_component_type>(here, ...);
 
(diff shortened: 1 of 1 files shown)


hkaiser:
I'm not sure if this change reflects the actual behavior. IIRC, `new_<>` can be used in two ways, using the component type or using the client type. If used with the component type it will return a `future<id_type>`, if used with a client type it will return this client type pre-populated with the corresponding lazy `id_type`.

hrasadi:
Problem is I get a compile error when using _new_<client_type>_ pattern. It complains that 'no matching function for call to 'new_''. If this is not an expected behavior, I will try to put together a small demo case and send it here.

msimberg:
@hrasadi were you able to find a small reproducing test case? Is this still a problem for you?

msimberg:
Let's reopen if this comes up again.

PR #3331, opened 2018-05-24
Add quickstart section to documentation

msimberg:
This adds a quick start section to the beginning of the documentation.

- It goes through the basics of building HPX and a single node application using HPX, links to relevant parts of documentation (nothing new)
- Adds another simplest hello world which uses `hpx::init` explicitly
- Adds a small example which uses `async`, `future`s, `when_all`, `then` and parallel algorithms


docs/quickstart.qbk
```diff
@@ -0,0 +1,282 @@
...
+__hpx__ executables end up in the `bin` directory in your build directory. You
+can now run `simplest_hello_world` and should see the following output:
+
+``
+./bin/simplest_hello_world_1
+Hello World!
+``
+
+You've just run an example which prints `Hello World!` from the __hpx__ runtime.
+The source for the example is in `examples/quickstart/simplest_hello_world.cpp`.
```
hkaiser: Shouldn't this be `simplest_hello_world1.cpp`?

docs/quickstart.qbk
```diff
@@ -0,0 +1,282 @@
...
+
+[import ../examples/quickstart/simplest_hello_world_1.cpp]
+
+[simplest_hello_world_1_getting_started]
+
+It looks almost like a regular `C++` hello world with the exception of the two
+includes and `hpx::cout`. When you include `hpx_main.hpp` some things will be
+done behind the scenes to make sure that `main` actually get's launched on the
+__hpx__ runtime. So while it looks almost the same you can now use futures,
+`async`, parallel algorithms and more which make use of the __hpx__ runtime with
```
hkaiser: The only difference to a 'normal' C++ program is that your `main` function _must_ return a value (normally main is allowed not to have an explicit `return` statement).

docs/quickstart.qbk
```diff
@@ -0,0 +1,282 @@
...
+__hpx__ runtime. ]
+
+[heading Writing task-based applications]
+
+So far we haven't done anything that can't be done using the C++ standard
+library. In this section we will give a short overview of what you can do with
+__hpx__ on a single node. The essence is to avoid global synchronization and
+break up your application into small, composable tasks whose dependencies
+control the flow of your application. Remember, however, that __hpx__ allows you
+to write distributed applications in similarly to how you would write
```
hkaiser: superfluous `in`

docs/quickstart.qbk
```diff
@@ -0,0 +1,282 @@
...
+
+Try copying the contents to your `main.cpp` file and look at the output. It can
+be a good idea to go through the program step by step with a debugger. You can
+also try changing the types or adding new arguments to functions to make sure
+you can get the types to match. The type of the `then` method can be especially
+tricky to get right (the continuation needs to take the future as an argument).
+
+[note __hpx__ programs accept command line arguments. The most important one is
+`--hpx:threads=N` to set the number of OS-threads used by __hpx__. __hpx__ uses
+all available threads by default. Play around with the example above and see
```
hkaiser: Should that be `all available cores by default`?

docs/quickstart.qbk
msimberg: Yes, thanks!

docs/quickstart.qbk
msimberg: Very good point. We already got bitten by this at CSCS, so it would be very good to point that out. I'll add a note.

docs/quickstart.qbk
msimberg: Thanks!

docs/quickstart.qbk
msimberg: Yes, or even all available processing units.

Jakub-Golinowski:
Great examples! Thank you very much.

docs/quickstart.qbk
```diff
@@ -0,0 +1,288 @@
...
+unpack it or clone the repository directly using `git`:
+
+``
+    git clone https://github.com/STEllAR-GROUP/hpx.git
+``
+
+It is also recommended that you check out the latest stable tag:
+
+``
+    git checkout 1.1.0
```
hkaiser: We should add a note to the release-procedure document to not to forget changing this version (here: 1.1.0) during release preparations.

msimberg: I'm counting on point 26 (the note about grepping old version numbers) in the [release procedure](https://github.com/STEllAR-GROUP/hpx/blob/master/docs/release_procedure.rst) catching that.

We could also have a `latest` tag or something like that which would always point to the latest release and say `git checkout latest`, but I'm not sure if that's a good idea (tags shouldn't change and so on...).

hkaiser: Ok, thanks!

PR #3334, opened 2018-05-24
Making sure early logging settings don't get mixed with others

hkaiser:
Fixes currently observed initialization errors seen when running with SLURM


hkaiser:
The reported errors on daint seem to be unrelated. Everything else is green. @msimberg ok to merge?

PR #3338, opened 2018-05-28
Replacing hpx_main's #define main with overriding __libc_start_main

NK-Nikunj:
This is my first GSoC'18 commit. Currently, this commit provides code for refining **hpx_main.hpp** for **gcc** and **clang** based compilers. I am working on **msvc** based compilers and will provide code for the same soon.

## Proposed Changes

  - Refining **hpx_main.hpp** by using wrapper for ``__libc_start_main``

I have created a wrapper for ``__libc_start_main`` and created a hook for the function. I changed the signature of ``__libc_Start_main`` by providing my own entry point (``__initializing_main``) instead of ``main``. From here I call the ``hpx::init`` to start the HPX runtime system. From the initialized system, I call the ``main`` function.

For a user, including this header will enable him to use HPX functionality directly from ``int main``.

## Current challenges and refinement required

This implementation has a few drawbacks which I'm currently working on. These are as follows:
 - It will NOT work for any global object using HPX functionality.
 - It has used global scope variables, which in my opinion is not a good practice

Any suggestions and reviews are welcomed.

### DIFF
--- a/hpx/hpx_main.hpp
+++ b/hpx/hpx_main.hpp
@@ -1,4 +1,5 @@
 //  Copyright (c) 2007-2014 Hartmut Kaiser
+//  Copyright (c) 2018 Nikunj Gupta
 //
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
@@ -31,31 +32,31 @@ char** __envp = nullptr;
 ///////////////////////////////////////////////////////////////////////////////
 // Default implementation of __hpx_entry() for initializing the HPX runtime
 // system
-// 
+//
 int __hpx_entry(int argc, char* argv[])
 {
-	// Call to the Compiler's main with HPX runt time
-	// initiated.
-	actual_main(__argc, __argv, __envp);
+    // Call to the Compiler's main with HPX runt time
+    // initiated.
+    actual_main(__argc, __argv, __envp);
 
-	return hpx::finalize();
+    return hpx::finalize();
 }
 
 ///////////////////////////////////////////////////////////////////////////////
 // Default entry point for the runtime system.
-// 
+//
 // Note: This function is now the equivalent of Comipler's main i.e the entry
 // point has been shifted from "main" to "initializing_main"
-// 
+//
 int __initializing_main (int argc, char** argv, char** envp)
 {
-	__argc = argc;
-	__argv = argv;
-	__envp = envp;
+    __argc = argc;
+    __argv = argv;
+    __envp = envp;
 
-	std::vector<std::string> const cfg = {
-		// allow for unknown command line options
-		"hpx.commandline.allow_unknown!=1",
+    std::vector<std::string> const cfg = {
+        // allow for unknown command line options
+        "hpx.commandline.allow_unknown!=1",
 	};
 
     using hpx::util::placeholders::_1;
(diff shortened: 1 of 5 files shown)


hpx/hpx_main.hpp
```diff
@@ -9,13 +10,92 @@
...
+
+#elif defined(HPX_HAVE_STATIC_LINKING)
 #include <hpx/hpx_main_impl.hpp>
-#endif
 
 // We support redefining the plain C-main provided by the user to be executed
 // as the first HPX-thread (equivalent to hpx_main()). This is implemented by
 // a macro redefining main, so we disable it by default.
 #define main hpx_startup::user_main
+#endif
```
hkaiser: This disables the `#define main` on all platforms that do not support your new initialization code. I think the static linking is orthogonal to your new functionality.

tests/unit/runtime/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+    add_hpx_pseudo_dependencies(tests.unit.runtime
+                                tests.unit.runtime.${test})
+
+    # add dependencies to pseudo-target
+    add_hpx_pseudo_dependencies(tests.unit.runtime.${test}
+                                ${test}_test_exe)
+  endforeach()
+
+else()
+  message(FATAL_ERROR "This test is meant only for gcc and clang compilers.")
```
hkaiser: If you unconditionally raise an error on MSVC it will break the whole build system. I think there is no need for this `else()` case altogether.

tests/unit/runtime/runtime_initialization.cpp
```diff
@@ -0,0 +1,32 @@
+//  Copyright (c) 2018 Nikunj Gupta
+//
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// Simple test verifying runtime initialization sequence
+
+#define _GNU_SOURCE
```
hkaiser: Why is this necessary?

tests/unit/runtime/runtime_initialization.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+#include <dlfcn.h>
```
hkaiser: Why is this #include needed?

hpx/hpx_main.hpp
NK-Nikunj: @hkaiser yes, static linking is orthogonal to my implementation. I should probably change it to gnuc, clang and not HPX_HAVE_STATIC_LINKING and keep the rest same. This should solve the issue from what I think.

tests/unit/runtime/CMakeLists.txt
NK-Nikunj: It was added as a precaution for msvc based build which will currently not work. But yes it is redundant and can be removed. Since it will cause the whole system to break, I'll remove it altogether.

tests/unit/runtime/runtime_initialization.cpp
NK-Nikunj: @hkaiser I added it to write portable code. It provides access to various nonstandard GNU/Linux extension functions. I have not made any use of it, but I added keeping in mind that I might need it in the final implementation.

hkaiser: I don't think it to be a good idea to request from the user to define this in order for your system to work. Wouldn't there be a way to just `#include <hpx/hpx_main.hpp>` and be done with it?

tests/unit/runtime/runtime_initialization.cpp
hkaiser: Same comment as above. The user shouldn't see all of this.

NK-Nikunj:
Not necessarily. But I can make them inline if you want me to.

hkaiser:
Using names starting with `'__'` are reserved for the implementation, could you use names without that please (feel free to move the functions into a namespace, though)?

NK-Nikunj:
Closing due to better implementation in the pr #3357

PR #3341, opened 2018-05-28
Allowing for all command line options to be used as configuration settings

### LINKED ISSUE #3339: setting command line options/flags from init cfg is not obvious
Try using 
```
std::vector<std::string> cfg = {
    "hpx.os_threads=2",
    "hpx.bind=balanced",
    "hpx.print-bind"
};
```
hpx.print-bind is not accepted as a valid config entry. I'm not sure why not. Is there a difference between those with values and those without?

Clearly there is a special syntax, because on the command line one says
`--hpx:threads=2` but in the cfg, one says `hpx.os_threads=2`, it's not obvious why they should be different. Why can one not just put command line options in the cfg?

hkaiser:
Fixes #3339

This patch allows for command line options to be used instead of the configuration setting as explained in #3339.

@bcumming, @biddisco please check whether this solves your problem

### DIFF
--- a/hpx/util/command_line_handling.hpp
+++ b/hpx/util/command_line_handling.hpp
@@ -84,6 +84,9 @@ namespace hpx { namespace util
             boost::program_options::options_description const& help);
 
         void handle_attach_debugger();
+
+        std::vector<std::string> preprocess_config_settings(
+            int argc, char** argv);
     };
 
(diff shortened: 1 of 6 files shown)


hkaiser:
Hmmm, looks like many tests fail now. Will fix.

PR #3344, opened 2018-05-31
Reduce amount of information printed for unknown command line options

### LINKED ISSUE #3321: hpx bad_parameter handling is awful
## Expected Behavior
`./bin/hello_world_exe --test` 
should simply print a warning about 'unknown parameter'

## Actual Behavior
An exception is thrown
Pages of rubbish that nobody wants to see with a bad_parameter somewhere at the end

What would be better is simply displaying a warning.
We had a mini hpx workshop at ETHZ and most of the participants found the output from HPX to be excessive and said that the user could ask for the stack trace and extra config info by adding `--hpx:print-errors` or something of that kind ...

hkaiser:
Fixes #3321

### DIFF
--- a/examples/quickstart/simplest_hello_world_2.cpp
+++ b/examples/quickstart/simplest_hello_world_2.cpp
@@ -18,8 +18,8 @@ int hpx_main(int, char**)
     return hpx::finalize();
 }
 
-int main()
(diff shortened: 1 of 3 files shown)


hkaiser:
@biddisco please verify that this fixes the issue you reported.

PR #3355, opened 2018-06-26
Reverting changes introduced by #3283 as those make applications hang

hkaiser:
@justwagle please verify whether this fixes the application hangs you're seeing.


justwagle:
@hkaiser This gets rids of the hangs I was seeing. Thank you very much.

PR #3357, opened 2018-06-27
Static/dynamic executable implementation

NK-Nikunj:
## Proposed Changes

 - This pr adds static/dynamic executable implementation by using linker option `-Wl,-wrap`.

I have used the linker option `-Wl,-wrap` to wrap the `__libc_start_main` (unlike last time where I used `dlfcn` to hook into it). By using the linker option the function can be understood as `__wrap___libc_start_main` and `__real___libc_start_main`. Any call to the actual function will be received by our wrapper function. It is from here that we change the entry point of the C runtime system to our own custom function (`initializing_main` in our case). From here we configure the HPX runtime system and initialize it with the entry point being `hpx::hpx_entry`. From the initialized system we call the C main function with arguments.

## Using this method to generate an executable

Since this implementation requires linking to `libhpx_wrap.a` (build during HPX build itself), I have added a flag by the name of `INCLUDE_HPX_MAIN` (subject to change, if required). Addition of the flag to `add_executable` will take care of the linking part.

## TODO

 - Add flag to all examples and tests
 - Output error if the flag is not provided but `hpx_main.hpp` is included
 - Output error if the header is provided but the flag is not

## Failed attempt

I tried to initialize the HPX runtime system in such a manner that it would work for any global scoped object (utilizing HPX functionality) but it failed. The reason for it to fail was due to mismatch of the initialization sequence of std library, HPX runtime system and C runtime system. I tried to force initialize buffers and libraries but it still did not work as expected (It was able to run a few codes correctly while producing errors for the rest).

I am still trying to figure out ways to incorporate global objects into HPX but I can't figure out a way as of now.


cmake/HPX_SetupTarget.cmake
```diff
@@ -187,9 +187,8 @@ function(hpx_setup_target target)
     set(_USE_CONFIG 0)
   endif()
 
-  # linker instructions
   if(NOT target_NOLIBS)
-    set(hpx_libs hpx)
+    set(hpx_libs hpx ${hpx_libs})
     if(NOT target_STATIC_LINKING)
       set(hpx_libs ${hpx_libs})
```
hkaiser: Can we remove `${hpx_libs}` here now?

cmake/HPX_SetupTarget.cmake
```diff
@@ -204,6 +203,11 @@ function(hpx_setup_target target)
     if(DEFINED HPX_LIBRARIES)
       set(hpx_libs ${hpx_libs} ${HPX_LIBRARIES})
     endif()
+
+    # add linker flags for libhpx_init.a for linux environment
+    if(CMAKE_SYSTEM_NAME STREQUAL Linux)
+      set_target_properties(${target} PROPERTIES LINK_FLAGS "-Wl,-wrap=__libc_start_main")
+    endif()
```
hkaiser: This adds the overloaded entry point unconditionally. Is that what we want? Shouldn't this be done for executables only (or at least those modules that link against hpx_init)?

hpx/hpx_main.hpp
```diff
@@ -9,6 +10,18 @@
 #include <hpx/config.hpp>
 #include <hpx/hpx_init.hpp>
 
+// We support different implementation depending upon the Operating
+// System in use.
+#if defined(__linux__)
```
hkaiser: We usually check for Linux as follows:
```
#if defined(__linux) || defined(__linux__) || defined(linux)
...
#endif
```

Different flavors of Linux define different pp constants.

hpx/hpx_main.hpp
```diff
@@ -9,6 +10,18 @@
...
+
+// include_libhpx_wrap here is an override for the one present in
+// src/hpx_wrap.cpp. The value of this variable defines if we need
+// to change the program's entry point or not.
+extern int include_libhpx_wrap;
+int include_libhpx_wrap = 1;
+
```
hkaiser: Could you place these global variables into a separate namespace, please? In other spots we use `namespace hpx_startup {}` for this.

src/hpx_wrap.cpp
```diff
@@ -0,0 +1,127 @@
+//  Copyright (c) 20018 Nikunj Gupta
+//
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// The following implementation is only possible for Linux systems.
+#if defined(__linux__)
```
hkaiser: Same as above, please use all three flavors of the Linux pp.

src/hpx_wrap.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+namespace hpx{
+    // Main entry point of HPX runtime system
+    extern int hpx_entry(int argc, char* argv[]);
+}
+
+// HPX's implented program's entry point
+extern int initialize_main(int argc, char** argv, char** envp);
+
+// Real libc function __libc_start_main
```
hkaiser: Please add a comment with a link to the source code of the original libc implementation.

src/hpx_wrap.cpp
```diff
@@ -0,0 +1,127 @@
...
+                void (*rtld_fini) (void), void (* stack_end));
+
+
+////////////////////////////////////////////////////////////////////////////////
+// Global pointers
+
+// actual_main is responsible to store the pointer to the main()
+// function. POSIX implementation of main requires pointer to envp
+// so it is stored as well.
+int (*actual_main)(int, char**, char**) = nullptr;
```
hkaiser: Please move these globals into a separate namespace.

src/hpx_wrap.cpp
```diff
@@ -0,0 +1,127 @@
...
+int initialize_main(int argc, char** argv, char** envp)
+{
+    // initializing envp pointer to utilize when
+    // calling the actual main.
+    __envp = envp;
+
+    // Configuring HPX system before runtime
+    std::vector<std::string> const cfg = {
+        "hpx.commandline.allow_unknown!=1",
+    };
```
hkaiser: Please also disable the short command line options, similar to https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/hpx_main_impl.hpp#L31-L34

cmake/HPX_SetupTarget.cmake
NK-Nikunj: oh, I forgot to remove it from there. I'll do it now

cmake/HPX_SetupTarget.cmake
NK-Nikunj: Yes, this should be done for executables only. I will add another parameter to the if statement to add for target type executables.

hpx/hpx_main.hpp
NK-Nikunj: I will update it as you have described.

PR #3365, opened 2018-07-05
Fix order of hpx libs in HPX_CONF_LIBRARIES.

Jakub-Golinowski:
Fixes building issue that surfaced when building OpenCV with HPX backend (https://github.com/opencv/opencv/pull/11897). 

The hpx_init depends on hpx therefore it should be before it in the list of hpx libraries.

## Proposed Changes

  - inserts hpx_init before hpx in the list of hpx libraries in cmake HPX_CONF_LIBRARIES


cmake/HPX_GeneratePackage.cmake
```diff
@@ -60,7 +60,7 @@ if(HPX_WITH_STATIC_LINKING)
   set(HPX_PKG_LIBRARIES "\${libdir}/libhpx.a ${HPX_PKG_LIBRARIES}")
   set(HPX_PKG_DEBUG_LIBRARIES "\${libdir}/libhpxd.a ${HPX_PKG_DEBUG_LIBRARIES}")
 else()
-  set(HPX_CONF_LIBRARIES "general;hpx;general;hpx_init;${HPX_LIBRARIES}")
+  set(HPX_CONF_LIBRARIES "general;hpx_init;hpx;general;${HPX_LIBRARIES}")
```
hkaiser: This will break cmake dependency handling. the 'general;' has to prefix the library name. If you need to change the linking order, please modify this to 
```
set(HPX_CONF_LIBRARIES "general;hpx_init;general;hpx;${HPX_LIBRARIES}") 
```

Jakub-Golinowski: Ok, thank you for pointing that out. I made the changes as suggested.

khuck:
@hkaiser It doesn't fix anything in Phylanx, but it doesn't seem to make things worse.  What does this change do? I see that the library ordering later in the same cmake file stays the same (libhpx then libhpx_init), so I don't know what this flag is used for.  What about HPX_PKG_LIBRARIES_BAZEL, or HPX_PKG_LIBRARIES?

hkaiser:
@khuck well, this change is supposed to modify the link order for `libhpx_init.a` and `libhpx.so` for libraries that depend on HPX.. 

Also, thanks for noticing the same problem in other places - @Jakub-Golinowski would you mind fixing those as well?

Jakub-Golinowski:
I changed the order of libraries in all occurrences as suggested. If this is OK, let me know and I will squash it to single before merge.

hkaiser:
@Jakub-Golinowski could you rebase this branch onto master, please? We can't merge it otherwise.

Jakub-Golinowski:
@hkaiser rebased against master and squashed to 1 commit.

PR #3367, opened 2018-07-07
Adds Mac OS implementation to hpx_main.hpp

NK-Nikunj:
## Proposed Changes

  - Changes current `hpx_main.hpp`'s implementation of `#define` to a more robust solution using linker flags.

In this PR I'm extending my Linux implementation for `hpx_main` to Mac OS X. The implementation here is achieved using another linker flag to change the entry point of the system to `initialize_main` instead of `main`.
Using some linker tricks function stack is determined at runtime depending upon the inclusion of `hpx_main.hpp`.


NK-Nikunj:
@hkaiser yes I can unify the two implementations using PPs wherever necessary. I kept it separated so as to keep code look clearer. Should I unify the two?

hkaiser:
> @hkaiser yes I can unify the two implementations using PPs wherever necessary. I kept it separated so as to keep code look clearer. Should I unify the two?

@NK-Nikunj: yes, please try to unify the two, but within reason ;-) This will probably be a balance between code duplication and readability.

ac-alpha:
@hkaiser I ran a build for it and tests seems to run fine (except for 1 which constantly fails). You can find the gist to the tests here: https://gist.github.com/ac-alpha/e1efd4e078ce8577148596949da7bfc9

hkaiser:
@ac-alpha thanks very much! That is very helpful.

PR #3368, opened 2018-07-08
Updates doc with recent hpx_wrap implementation

NK-Nikunj:
This PR updates the docs to include the recent implementation for `hpx_main.hpp`. It now adds a `caution` for using `hpx_main.hpp` only once and in the executable. It also explains in brief how __HPX__ is able to seamlessly initiate things from main.

@hkaiser , I can add information specific to Linux and Apple implementations as well, if you want to include it in the docs.

### DIFF
--- a/docs/manual/hpx_main_usage.qbk
+++ b/docs/manual/hpx_main_usage.qbk
@@ -1,0 +1,55 @@
+[/=============================================================================
+    Copyright (C) 2018 Nikunj Gupta
+
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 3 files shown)


docs/manual/hpx_main_usage.qbk
```diff
@@ -0,0 +1,55 @@
+[/=============================================================================
```
hkaiser: This file needs to be added to the docs somehow, please see here for examples on how to do this: https://github.com/STEllAR-GROUP/hpx/blob/master/docs/hpx.qbk#L333-L359. Have you been able to build the docs?

PR #3371, opened 2018-07-09
Adds HPX_LINKER_FLAGS to HPX applications without editing their source codes

NK-Nikunj:
Fixes breaking of HPX applications due to recent merged pull request (#3357 )

## Proposed Changes

Creates a separate static library for `hpx_wrap.cpp`. If the option `HPX_WITH_DYNAMIC_HPX_MAIN` is set to `ON` then it will be created. It would also add the specific `HPX_LINKER_FLAGS` to `CMAKE_EXE_LINKER_FLAGS` which will be propagated to HPX specific apps. If it is set to `OFF` then it will not generate the static library. The linker flags will not be added as well. Therefore, the system will default to the previous implementation.

## Use in HPX applications (Ex: `phylanx`)

Since `CMAKE_EXE_LINKER_FLAGS` is updated in `HPXConfig`, it is propagated to the application using HPX. The linker flag contains the value for `HPX_LINKER_FLAGS` which can be empty or the linker flag specific to Linux implementation (depending upon the option `HPX_WITH_DYNAMIC_HPX_MAIN`).

If HPX was built with the option set to `OFF` then the library won't be built and the linker flags won't be updated as well. This would mean that any application utilizing HPX will use the previous implementation.

If HPX was built with the option set to `ON` then the library will be built and linker flags will be updated and will propagate to the HPX applications (using `HPXConfig`). This would mean that any application utilizing HPX will use the current implementation (in case of Linux).


cmake/HPX_SetupTarget.cmake
```diff
@@ -188,7 +188,11 @@ function(hpx_setup_target target)
   endif()
 
   if(NOT target_NOLIBS)
-    set(hpx_libs hpx)
+    if(HPX_WITH_DYNAMIC_HPX_MAIN AND ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux"))
```
hkaiser: Shouldn't this happen for executables only?

src/CMakeLists.txt
```diff
@@ -316,11 +330,64 @@ foreach(_flag ${HPX_TARGET_COMPILE_DEFINITIONS_VAR})
   endif()
 endforeach()
 
+################################################################################
+# libhpx_wrap
+
```
hkaiser: All of this should be disabled for anything but Linux

cmake/HPX_SetupTarget.cmake
NK-Nikunj: @hkaiser if we limit it to Executables only, then some shared objects that link to it in phylanx will throw a linker error for not defined `__real___libc_start_main`. Providing this flag to non-executables will not produce unexpected results.

src/CMakeLists.txt
NK-Nikunj: oh yes! forgot that.

PR #3374, opened 2018-07-10
Adds hpx_wrap to HPX_LINK_LIBRARIES which links only when specified.

NK-Nikunj:
Fixes Failing tests in Phylanx

## Proposed Changes

  - Excludes `libhpx_wrap.a` from `HPX_LIBRARIES` list. It is now linked only when specified.
  - Exports `HPX_LINKER_FLAGS` instead of changing `CMAKE_EXE_LINKER_FLAGS`. It should be added for executables along with the link library (that is exported by the name of `HPX_LINK_LIBRARIES`).


hkaiser:
@NK-Nikunj please rebase your branch onto master, otherwise we can't go ahead with this.

src/CMakeLists.txt
```diff
@@ -393,6 +392,11 @@ foreach(_flag ${HPX_TARGET_COMPILE_OPTIONS_VAR})
   endif()
 endforeach()
 
+set(export_hpx_wrap "")
+if(HPX_WITH_DYNAMIC_HPX_MAIN AND ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux"))
+  set(export_hpx_wrap "hpx_wrap")
+endif()
+
```
hkaiser: Couldn't this be simply added to the `hpx_targets` list?

NK-Nikunj: hpx_targets is later exported with dependencies. So hpx_wrap will be added as a target dependency (please correct me if I'm wrong).

hkaiser: Fair point.

NK-Nikunj: @hkaiser I have tested it including in `hpx_targets`. Turns out `hpx_wrap` was added to Phylanx `.so` files because `hpx_wrap` was added in `HPX_CONF_LIBRARIES` in GeneratePackage.
`hpx_targets` has nothing to do with final `HPX_LIBRARIES` exported (and it's potential linkage with HPX applications). So, I've added `hpx_wrap` back to `hpx_targets`.
I've tested this configuration with Debug/Release and `HPX_WITH_DYNAMIC_MAIN` ON/OFF.

PR #3375, opened 2018-07-11
Replacing wrapper for __libc_start_main with main

NK-Nikunj:
Fixes platform issues with new `hpx_wrap` implementation

## Proposed Changes

The reason for failing `__libc_start_main` on a different platform (Eg. Powerpc) was due to a different function signature all together for `__libc_start_main`. This would mean that we will have to implement it for various platforms.

To overcome the above limitations, I am changing the wrapper function from `__libc_start_main` to `main`. Since we need to change the entry point only, `main` will fulfill our needs.

With these changes, I think, we cover all of linux.


biddisco:
I can confirm that code now runs correctly on PowerPC using HPX_WITH_DYNAMIC_HPX_MAIN=ON

NK-Nikunj:
John sir,

It's great to see it working on powerpc. I'm experiencing a bad
connectivity so I'm unable to converse on IRC.

Could you please run the tests on the build to check if they all pass. If
possible please try to build phylanx as well to check if HPX applications
run well.

Thanks
Regards
Nikunj Gupta

PR #3386, opened 2018-07-18
Adding support for boost::container::small_vector to dataflow

hkaiser:


hkaiser:
@Naios would you mind reviewing this, please?

hpx/util/detail/pack_traversal_impl.hpp
```diff
@@ -363,9 +363,8 @@ namespace util {
                 typename OldType, typename OldAllocator,
                 // Check whether the second argument of the container was
                 // the used allocator.
-                typename std::enable_if<std::uses_allocator<
-                    Base<OldType, OldAllocator>, OldAllocator>::value>::type* =
-                    nullptr,
+                typename Enable = typename std::enable_if<std::uses_allocator<
```
Naios: Is this change neccessary? Overall the old style is preferrable because arguments are not part of the template method signature otherwise. See https://stackoverflow.com/a/32167562/2303378 for details.

hkaiser: Good point, I'll revert this.

hkaiser: Reverted.

PR #3388, opened 2018-07-23
Add several functionalities allowing to optimize synchronous action invocation

hkaiser:
- add hpx::sync, semantically equivalent to hpx::async().get(), except that in the
  case of local execution of a direct action no future needs to be created
- add hpx::unwrap_result 'distribution policy' that can be used with hpx::async and
  direct actions that return a future on the server side. That allows to avoid
  creating the outer future in case of local synchronous execution.
- several minor optimizations for hpx::future and hpx::unwrapping

### DIFF
--- a/hpx/lcos/detail/async_unwrap_result_implementations.hpp
+++ b/hpx/lcos/detail/async_unwrap_result_implementations.hpp
@@ -10,6 +10,7 @@
 #include <hpx/lcos/future.hpp>
 #include <hpx/lcos/detail/async_implementations.hpp>
 #include <hpx/lcos/detail/async_unwrap_result_implementations_fwd.hpp>
+#include <hpx/lcos/detail/sync_implementations.hpp>
 #include <hpx/lcos/packaged_action.hpp>
 #include <hpx/runtime/launch_policy.hpp>
 #include <hpx/runtime/naming/address.hpp>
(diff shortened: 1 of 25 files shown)


hpx/lcos/detail/sync_implementations.hpp
```diff
@@ -0,0 +1,141 @@
...
+            }
+        }
+
+        return async_impl<Action>(policy, id, std::forward<Ts>(vs)...).get();
+    }
+
+    template <typename Action, typename ...Ts>
+    typename hpx::traits::extract_action<Action>::type::local_result_type
+    sync_impl(hpx::detail::sync_policy, hpx::id_type const& id, Ts&&... vs)
+    {
```
sithhell: Do we really need to duplicate the code here?

hpx/lcos/detail/async_unwrap_result_implementations.hpp
```diff
@@ -0,0 +1,239 @@
...
+                return f;
+            }
+        }
+        return f;
+    }
+
+    template <typename Action, typename ...Ts>
+    typename hpx::traits::extract_action<Action>::type::local_result_type
+    async_unwrap_result_impl(hpx::detail::sync_policy, hpx::id_type const& id,
+        Ts&&... vs)
```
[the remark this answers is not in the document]
hkaiser: Yah, we could factor this out.

hkaiser:
@sithhell your requested changes have been applied.

hkaiser:
@sithhell better now?

hpx/lcos/detail/sync_implementations.hpp
sithhell: Those two functions still look almost identical. Do we really need that here?

hkaiser: Ok, ok - you got me there. I was hoping you wouldn't notice ;-) Should be fine now, though.

hkaiser:
> I guess the only remaining issue I have is that the newly introduced unwrap_result functionality isn't really a distribution policy but something else. 

@sithhell I have renamed the policy to `unwrapped_result_policy`, so from the users' standpoint lal is well. All of the distribution policies have to be revisited after the executors dust has settled anyways.

PR #3392, opened 2018-07-24
Add performance counters background overhead and background work duration

justwagle:
## Proposed Changes

  - This pull request adds 2 new counters:
     - threads/time/background-work-duration 
     - threads/background-overhead

These counters measure the total time spent doing background work and the fraction of time spent doing background work with respect to the overall time spent running the scheduler. 
These counters need to be explicitly enabled (off by default) and hence should not add any overheads when not enabled.

### DIFF
--- a/hpx/runtime/threads/detail/scheduled_thread_pool.hpp
+++ b/hpx/runtime/threads/detail/scheduled_thread_pool.hpp
@@ -242,6 +242,11 @@ namespace hpx { namespace threads { namespace detail
 
         std::int64_t get_cumulative_duration(std::size_t, bool);
 
+#ifdef HPX_HAVE_ADAPTIVE_COALESCING_COUNTERS
+        std::int64_t get_background_work_duration(std::size_t, bool);
+        std::int64_t get_background_overhead(std::size_t, bool);
+#endif    //HPX_HAVE_ADAPTIVE_COALESCING_COUNTERS
+
 #if defined(HPX_HAVE_THREAD_IDLE_RATES)
         std::int64_t avg_idle_rate_all(bool reset);
         std::int64_t avg_idle_rate(std::size_t, bool);
@@ -361,6 +366,13 @@ namespace hpx { namespace threads { namespace detail
         // tfunc_impl timers
         std::vector<std::uint64_t> exec_times_, tfunc_times_;
         std::vector<std::uint64_t> reset_tfunc_times_;
+
+#ifdef HPX_HAVE_ADAPTIVE_COALESCING_COUNTERS
+        std::vector<std::uint64_t> background_duration_,
+            reset_background_duration_;
+        std::vector<std::uint64_t> reset_background_tfunc_times_;
+        std::vector<std::uint64_t> reset_background_overhead_;
+#endif    //HPX_HAVE_ADAPTIVE_COALESCING_COUNTERS
 
         std::vector<std::int64_t> idle_loop_counts_, busy_loop_counts_;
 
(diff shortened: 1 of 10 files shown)


hpx/runtime/threads/detail/scheduled_thread_pool.hpp
```diff
@@ -362,6 +367,15 @@ namespace hpx { namespace threads { namespace detail
         std::vector<std::uint64_t> exec_times_, tfunc_times_;
         std::vector<std::uint64_t> reset_tfunc_times_;
 
+        //background duration for adaptive coalescing counters
+        std::vector<std::uint64_t> background_duration_;
```
hkaiser: Shouldn't this be inside the pp guards below as well?

src/runtime/threads/executors/thread_pool_executors.cpp
```diff
@@ -339,14 +339,14 @@ namespace hpx { namespace threads { namespace executors { namespace detail
 
             // FIXME: turn these values into performance counters
             std::int64_t executed_threads = 0, executed_thread_phases = 0;
-            std::uint64_t overall_times = 0, thread_times = 0;
+            std::uint64_t overall_times = 0, thread_times = 0, bg_work=0;
```
hkaiser: Please follow the formatting we use everywhere else in this file (use spaces surrounding `=`)

hpx/runtime/threads/detail/scheduled_thread_pool.hpp
justwagle: I needed to exclude it from the guard as I am using that variable to create an object of type detail::scheduling_counters (https://github.com/justwagle/hpx/blob/9eac86a642d4e43cc5b359f7d8444170f166cda8/hpx/runtime/threads/detail/scheduled_thread_pool_impl.hpp#L576). If I included it in the guard, I got an error stating that background_duration_ was undefined when the counter was disabled.

hpx/runtime/threads/detail/scheduled_thread_pool.hpp
```diff
@@ -242,6 +242,11 @@ namespace hpx { namespace threads { namespace detail
 
         std::int64_t get_cumulative_duration(std::size_t, bool);
 
+#if defined(HPX_HAVE_ADAPTIVE_COALESCING_COUNTERS) && defined(HPX_HAVE_THREAD_IDLE_RATES)
```
sithhell: What's the relation between adaptive coalescing and the background thread timings? Why can't I just enable the performance counters without the adaptive coalescing option?

justwagle: @sithhell  The plan is to use these counters to do adaptive parcel coalescing so I used the name adaptive coalescing, but I am open to changing it to something else like HPX_HAVE_BACKGROUND_THREAD_COUNTERS. Let me know if you want me to do that instead.

sithhell: yes, that sounds better. And I guess they are orthogonal to the idle rate counters as well.

hkaiser: They are orthogonal, but they internally use the same data sources, iirc,

justwagle:
@sithhell I made the changes you suggested. Its now HPX_WITH_BACKGROUND_THREAD_COUNTERS / HPX_HAVE_BACKGROUND_THREAD_COUNTERS. Thanks you very much.

PR #3393, opened 2018-07-24
Explicitly instantiate configured schedulers

K-ballo:
Fixes schedule_last_test linkage failures on msvc


hkaiser:
Yay, finally some code clang-tidy has not seen yet ;-)

PR #3394, opened 2018-07-26
Fix small mistake in primary_namespace_server.cpp

msimberg:


hkaiser:
Thanks for catching this!

PR #3396, opened 2018-07-30
Allowing to register thread event functions (start/stop/error)

hkaiser:


hkaiser:
@tianyi93 this should help you with registering callbacks for various thread events.

tianyi93:
@hkaiser  Yes, it works as what we want. Thanks!

PR #3397, opened 2018-08-01
Fixing code for older gcc versions

hkaiser:


hkaiser:
@msimberg this should fix the issue you were seeing.

msimberg:
Thanks @hkaiser for the quick fix! It's in fact with clang 3.8, but I guess the fix would be the same. Since this is most likely not too critical for anyone would you mind not merging until I set up a pycicle builder with that clang version as well.

hkaiser:
> Since this is most likely not too critical for anyone would you mind not merging until I set up a pycicle builder with that clang version as well.

@msimberg sure. just merge it yourself whenever you think is appropriate.

PR #3398, opened 2018-08-04
Atomic future state

hkaiser:
This converts the state of a future into an atomic variable which greatly reduces the need for a lock in the future's shared state. Measurements in Phylanx showed a performance gain of 8-10% just from this change.

### DIFF
--- a/hpx/lcos/detail/future_data.hpp
+++ b/hpx/lcos/detail/future_data.hpp
@@ -37,6 +37,7 @@
 #include <boost/container/small_vector.hpp>
 #endif
 
+#include <atomic>
 #include <chrono>
 #include <cstddef>
 #include <exception>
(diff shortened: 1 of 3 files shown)


hpx/lcos/detail/future_data.hpp
```diff
@@ -357,7 +348,7 @@ namespace detail
         {
             result_type* value_ptr = reinterpret_cast<result_type*>(&storage_);
             construct(value_ptr);
-            state_ = value;
+            state_.store(value, std::memory_order_relaxed);
```
K-ballo: the next constructor should probably use relaxed too.

src/lcos/detail/future_data.cpp
```diff
@@ -237,31 +239,48 @@ namespace hpx { namespace lcos { namespace detail
...
+        {
+            std::unique_lock<mutex_type> l(mtx_);
+            if (is_ready())
+            {
+#if !(defined(HPX_DISABLE_ASSERTS) || defined(BOOST_DISABLE_ASSERTS) ||        \
+    defined(NDEBUG))
+                // HPX_ASSERT might suspend
+                auto t = std::move(on_completed_);
+                l.unlock();
+                HPX_ASSERT(t.empty());
```
K-ballo: for this condition to hold, setting the state to ready needs to be done under the lock.

hpx/lcos/detail/future_data.hpp
hkaiser: It already does.

src/lcos/detail/future_data.cpp
hkaiser: Good catch... I think we don't need to verify this condition in the first place.

hpx/lcos/detail/future_data.hpp
K-ballo: no, there's another constructor in between. github doesn't let me comment on it since it's not part of the diff

msimberg:
This looks pretty stable on pycicle and I think it could be merged. If there are any rare race conditions introduced by this we won't find out using pycicle anyway.

msimberg:
@biddisco mentioned today that he'd like to add some more (CDash) timing outputs to performance tests (it's currently there for only one or two tests) before this PR goes in so that we could see the effect. Adding timing outputs to all performance tests is unlikely to happen soon, but a few could be done quickly. Suggestions for useful tests are welcome.

hkaiser:
I'd like to have that in master asap as we depend on this change in Phylanx. We're seeing a significant speedup there.

hkaiser:
@biddisco can this be merged now?

biddisco:
too late!

PR #3400, opened 2018-08-07
Fix floating point exception scheduler_base idle backoff

### LINKED ISSUE #3395: Floating point bug in hpx/runtime/threads/policies/scheduler_base.hpp
## Expected Behavior
Octo-tiger to run

## Actual Behavior
Floating point exception causes crash.

When running octo-tiger I get a floating point crash within HPX sometimes. I've traced it to line 138 of hpx/runtime/threads/policies/scheduler_base.hpp. The offending call is std::pow(2.0,double(wait_count_)). When wait_count is large enough, this causes an overflow. In particular, I got the error when wait_count_=1033. Seems like wait_count should never get that high, so I'm not sure what's going on here. 

The only way I have found to consistently reproduce it is to run octo-tiger with a specific start up file (q=0.7) on queenbee. It crashes when reading the startup file. When I set HPX_WITH_THREAD_MANAGER_IDLE_BACKOFF=OFF the problem goes away. I have seen the bug randomly occur in other conditions. 

## Specifications

  - HPX Version: master
  - Platform (compiler, OS): gcc, linux

msimberg:
Fixes #3395.

### DIFF
--- a/hpx/runtime/threads/policies/scheduler_base.hpp
+++ b/hpx/runtime/threads/policies/scheduler_base.hpp
@@ -32,6 +32,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <exception>
+#include <limits>
 #include <memory>
 #include <mutex>
 #include <set>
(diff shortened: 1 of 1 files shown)


msimberg:
Opening this before I forget. This seems to work but waiting for confirmation from @dmarce1.

PR #3402, opened 2018-08-08
Allow debug option to be enabled only for Linux systems with dynamic main on

NK-Nikunj:
This PR fixes the breaking of master on non Linux systems for not finding out about `hpx_start` and `include_libhpx_wrap`.


NK-Nikunj:
@hkaiser since Mac OSX PR has been merged, I have added apple specific macro to debug as well.

PR #3404, opened 2018-08-08
fixing multiple difinition of main() in linux

hapoo:

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5,7 +5,15 @@
 
 #include <hpx/config.hpp>
 
+// We support different implementation depending upon the Operating
+// System in use.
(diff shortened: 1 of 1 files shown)


sithhell:
When exactly do the multiple definitions materialize? Do you have a test case? I haven't encountered this so far.

hkaiser:
@sithhell it happens if libhpx_init.a is on the linker line before the object file that implements main().

NK-Nikunj:
@hkaiser `hpx_wrap` is usually the first linked library for HPX in case of executables. So I do not think it's due to `libhpx_init.a` being first on the linker line. It might be due to libraries linked with `libhpx_init` causing multiple definition error when linked with an executable. I will dig deeper into it. The PR is correct since I forgot to put a #define guard in src/main.cpp thinking that hpx_main.hpp was the only place where it was implemented.

PR #3405, opened 2018-08-09
Adding DHPX_HAVE_THREAD_LOCAL_STORAGE=ON to builds

parsa:
This PR adds `-DHPX_HAVE_THREAD_LOCAL_STORAGE=ON` to HPX builds. This is done since the HPX image created by CircleCI is used by the [HPX project](https://github.com/STEllAR-GROUP/hpxmp).


sithhell:
The main reason we had it turned of was because the corresponding unit tests failed. Is this fixed now or is someone committed to fix it?

hkaiser:
I was not aware of unit tests failing because of this. If that happens again, let's fix them.

PR #3406, opened 2018-08-10
Adding thread local allocator and use it for future shared states

hkaiser:
This helps reducing overheads caused by the allocation of future shared states.

This is WIP, currently. It can however already be used for experimentation.

### DIFF
--- a/hpx/lcos/future.hpp
+++ b/hpx/lcos/future.hpp
@@ -39,6 +39,7 @@
 #include <hpx/util/result_of.hpp>
 #include <hpx/util/serialize_exception.hpp>
 #include <hpx/util/steady_clock.hpp>
+#include <hpx/util/thread_allocator.hpp>
 #include <hpx/util/void_guard.hpp>
 
 #if defined(HPX_HAVE_AWAIT)
@@ -1479,16 +1480,31 @@ namespace hpx { namespace lcos
     >::type
     make_ready_future(T&& init)
     {
-        typedef typename hpx::util::decay_unwrap<T>::type result_type;
-        typedef lcos::detail::future_data<result_type> shared_state;
-        typedef typename shared_state::init_no_addref init_no_addref;
+        using result_type = typename hpx::util::decay_unwrap<T>::type;
 
-        boost::intrusive_ptr<shared_state> p(
-            new shared_state(init_no_addref{}, std::forward<T>(init)),
-            false);
+        using base_allocator =
+            util::thread_allocator<detail::future_data_refcnt_base>;
+        using shared_state = typename traits::detail::shared_state_allocator<
+                lcos::detail::future_data<result_type>, base_allocator
+            >::type;
+
+        using allocator = typename std::allocator_traits<base_allocator>::
+            template rebind_alloc<shared_state>;
+        using traits = std::allocator_traits<allocator>;
+
+        using init_no_addref = typename shared_state::init_no_addref;
+
+        using unique_ptr = std::unique_ptr<shared_state,
+            util::allocator_deleter<allocator>>;
+
+        allocator alloc{};
+        unique_ptr p(traits::allocate(alloc, 1),
+            util::allocator_deleter<allocator>{alloc});
+        traits::construct(
+            alloc, p.get(), init_no_addref{}, alloc, std::forward<T>(init));
 
         return hpx::traits::future_access<future<result_type>>::create(
-            std::move(p));
+            p.release(), false);
     }
 
     // Extension (see wg21.link/P0319)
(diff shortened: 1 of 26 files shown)


hpx/lcos/future.hpp
```diff
@@ -1499,17 +1515,32 @@ namespace hpx { namespace lcos
...
+
+        allocator alloc{};
+        unique_ptr p(traits::allocate(alloc, 1),
+            util::allocator_deleter<allocator>{alloc});
+        traits::construct(
+            alloc, p.get(), init_no_addref{}, default_construct{}, alloc);
 
         return hpx::traits::future_access<future<result_type>>::create(
-            std::move(p));
+            p.release(), false);
```
biddisco: Is the only reason for going to the trouble of creating a unique_ptr with an  `allocator_deleter` and then releasing it right after, to ensure that if an exception is thrown, all is cleared up?

hkaiser: @biddisco yes, that is correct.

hkaiser:
> Since this allocator is doing essentially what jemalloc/tcmalloc do, why not just create an arena in jemalloc for future shared states and use that?

@biddisco I didn't know this was possible. I'll investigate. From what I see however, I think tcmalloc/jemalloc can't match the performance of this allocator. I'll try to collect some hard data for this.

biddisco:
What you gain by using (for example) jemalloc is some memory garbage collection thrown in, so that if you keep allocating new pages of memory and then don't need them at some point, it will do some cleanup.
This might not be a big deal if you are only allocating small shared state instances - but it's a shame to see the code base grow in this way. For task stacks, we also need a thread local allocator, and maybe we should just use the one you have included. I'd be very interested in seeing a comparison of this allocator and a dedicated jemalloc arena of the same size (meaning an arena created for objects of size N bytes).

hkaiser:
> What you gain by using (for example) jemalloc is some memory garbage collection thrown in, so that if you keep allocating new pages of memory and then don't need them at some point, it will do some cleanup.

The allocator proposed here does that as well. Overall I agree, however. I'm all for ditching this PR if we could reuse functionalities in jemalloc (or similar) and still get decent performance.

> I'd be very interested in seeing a comparison of this allocator and a dedicated jemalloc arena of the same size (meaning an arena created for objects of size N bytes).

Good thinking. I'll prepare some benchmarks.

hkaiser:
@biddisco would you be able to point me to the documentation that describes using special jemalloc arena allocators?

sithhell:
Something worth considering as well is this here: https://accu.org/index.php/journals/2533

src/util/thread_allocator.cpp
```diff
@@ -0,0 +1,344 @@
...
+        thread_local_allocator& operator=(thread_local_allocator&&) = delete;
+
+        static thread_local_allocator& get_tls_allocator();
+        void process_pending_requests(bool forceall = false);
+
+    private:
+        void cleanup();
+
+        friend HPX_EXPORT void* thread_alloc(std::size_t size);
+        friend HPX_EXPORT void thread_free(void* addr);
```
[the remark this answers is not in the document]
hkaiser: We need it for MSVC :/

hkaiser:
> I would have expected an allocator that is completely free of synchronization primitives. That is, if data is going to get deleted on a thread that doesn't own the data, we could just add it to our freelist and not give it back to the original one.

I'm not sure if this scheme would allow us to track the number of blocks allocated in a page (in order to know when a page can be freed). At least it would require to make the block counter atomic (again) which turned out to have a significant impact...

src/util/thread_allocator.cpp
```diff
@@ -0,0 +1,344 @@
...
+                return reinterpret_cast<alloc_block*>(&data);
+            }
+
+            alloc_block const* get() const
+            {
+                return reinterpret_cast<alloc_block const*>(&data);
+            }
+
+            // FIXME: should we align this on page boundaries?
+            typename std::aligned_storage<PAGE_SIZE>::type data;
```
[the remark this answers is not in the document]
hkaiser: Aligning the pages a PAGE boundaries does not guarantee the alignment of the blocks in any way. I'm not too sure about this anymore for this reason.

sithhell:
Here is the data running the test on a two socket machine with 8 cores per socket:
https://docs.google.com/spreadsheets/d/18NYIeEuQ46MfsGhZn0smCXLXdGp7nMkmur-Z54r5db8/edit?usp=sharing

The `std::allocator` is backed with jemalloc. While we get an improvement from 1 to 4 cores, we can observe that the current implementation doesn't scale as good as the jemalloc implementation. I think the potential is clear here. In other tests, I couldn't observe performance regression with other benchmarks.

src/runtime/agas/interface.cpp
```diff
@@ -22,10 +22,17 @@
 
 namespace hpx { namespace agas
 {
+// request the reference to the AGAS client only once
+naming::resolver_client& get_agas_client()
+{
+    static naming::resolver_client& client = naming::get_agas_client();
```
msimberg: @hkaiser This doesn't play well with the tests that initialize the runtime multiple times (the ones that are failing on CircleCI). Does this change make a big difference?

hkaiser: @msimberg Good point! I forgot about the multiple-init scenarios. I'll roll this change back. This could also explain the hangs. Thanks!

hkaiser:
> The std::allocator is backed with jemalloc. While we get an improvement from 1 to 4 cores, we can observe that the current implementation doesn't scale as good as the jemalloc implementation. I think the potential is clear here. In other tests, I couldn't observe performance regression with other benchmarks.

@sithhell I'm not sure where our allocator loses. Any idea?

sithhell:
Here is the graph with the updated test:
https://docs.google.com/spreadsheets/d/e/2PACX-1vTiupxfK5Hgb7pGmIEKnCjzaq-fw_GimM-roV9NJOyEcYHGlbEQ6O9tDLsd-uzp3IdVfMm92coaWeMu/pubchart?oid=1063489052&format=interactive

The situation didn't change much. I don't think not excluding large allocations skewed the results. Using this new allocator should always be at least as fast as the system (in that case jemalloc) allocator. I don't know what causes the small regression here. The phylanx application reported good speedups, so I think this PR is a proper basis for future optimizations.

src/util/thread_allocator.cpp
```diff
@@ -0,0 +1,494 @@
...
+        }
+        else
+        {
+            detail::alloc_page* page = blk->get_ptr();
+            thread_local_allocator* allocator = page->allocator;
+            thread_local_allocator* current =
+                &thread_local_allocator::get_tls_allocator();
+
+            if (allocator != current)
+            {
```
sithhell: We need to add a `hpx::util::ignore_all_while_checking il;` here. When destructors are run, we might hold some locks.

hkaiser: Fair point, thanks. Done.

msimberg:
Could this eventually also be used for thread data/stack allocation, or is there something fundamentally incompatible for that purpose (e.g. the maximum block size seems quite small)?

sithhell:
It can't be used for the stack. The main difference is that we lazily allocate the stacks (only mmap without actual physically being backed up). `thread_data` is a different story

hkaiser:
This can be closed now as part of the changes have been incorporated into master (see #3465). The allocator itself is not needed as it does not improve the picture over jemalloc. I will create a separate PR enabling the use of jemalloc on windows

PR #3411, opened 2018-08-14
Test reverting all commits since a working appveyor build

msimberg:
This is just a test to see if appveyor starts building again with a previously working commit or if there has been some change in appveyor itself.


msimberg:
This one is not needed. We ran out of disk space on AppVeyor.

PR #3413, opened 2018-08-14
Fixing multiple defines of APEX_HAVE_HPX

khuck:
APEX_HAVE_HPX has to be defined in CMakeLists.hpx of APEX - otherwise, APEX doesn't know to include hpx/config.h.  If it is defined in both places, it gets defined twice and generates warnings.  If it isn't defined in CMakeLists.hpx, then hpx/config.h doesn't get included, and APEX isn't configured correctly.  Also changing version of APEX to be develop to get the latest APEX changes which includes making sure this define is defined.  Before HPX is released, APEX should be tagged and that tag should be used instead of develop.


CMakeLists.txt
```diff
@@ -1565,7 +1564,7 @@ if(HPX_WITH_APEX)
   include(GitExternal)
   git_external(apex
     https://github.com/khuck/xpress-apex.git
-    v2.0.1
+    develop   # this should be set to a known good tag when HPX is released!
```
hkaiser: Should we turn this into a CMake configuration variable? May I ask you to at least add a note to the release procedure here: https://github.com/STEllAR-GROUP/hpx/blob/master/docs/release_procedure.rst?

khuck:
@hkaiser I don't know why appveyor failed, but all other tests failed.  I added HPX_WITH_APEX_TAG as a CMake variable, defaulting to v2.0.1 (the most current release tag) if not overridden by another tag or branch name.  If my CMakeLists.txt settings are OK, can we merge this request?  Thanks!

PR #3415, opened 2018-08-16
Scheduler NUMA hint and shared priority scheduler

msimberg:
This contains parts of #3241, namely:

- Instead of a simple thread number there is now a `thread_schedule_hint` which tells functions to schedule a thread anywhere, on a specific thread, or on a NUMA domain.
- An updated `shared_priority_scheduler` example (this still contains logging, etc.)
- The same scheduler moved into core with command line options to use it etc. This is yet another scheduler™, but the intention is to get it into master so that it doesn't keep going out of date compared to master, and to later replace the existing schedulers with one based on the `shared_priority_scheduler`. There is currently no documentation about the new scheduler since it's there just for integration and testing purposes. This has been cleaned up and made consistent with existing schedulers.
- In the process I've slightly modified the way a fallback worker thread is chosen to hopefully make it a bit clearer.
- Note: (worker) thread suspension in its current form is slightly problematic with this scheduler as queues are shared between multiple cores. Tests pass but it'll likely need a revision in the future.

~@biddisco will still add some of his pending changes to this, WIP until then.~

Required for #3420.


biddisco:
@msimberg I've done a bit of cleaning, squashed quite a few commits and tidied things up a bit.
I'm not 100% happy because quite a few commits are still out of order and messed up, but I don't have the patience to edit it further at the moment.

biddisco:
I have not tested this branch with any of my newer guided pool executor stuff, so I will now have a go at getting that to work to see if this branch does what it is supposed to do.

msimberg:
@biddisco Thanks! I'll happily reorder/squash/fix inspect errors once you're happy with the functionality. No point doing that yet if you still want to change things.

hpx/runtime/threads/policies/shared_priority_queue_scheduler.hpp
```diff
@@ -802,14 +802,14 @@ namespace policies {
             }
             else if (priority == thread_priority_low)
             {
-                lp_queues_[domain_num].queues_[hp_lookup_[
-                    q_index % hp_queues_[domain_num].num_cores]]->
+                lp_queues_[domain_num].queues_[lp_lookup_[
```
biddisco: Good catch. The evil demons of cut'n paste have been working their magic here.

src/util/command_line_handling.cpp
```diff
@@ -774,7 +774,7 @@ namespace hpx { namespace util
 #endif
 
         // handle setting related to schedulers
-        queuing_ = detail::handle_queuing(cfgmap, vm, "local-priority-fifo");
+        queuing_ = detail::handle_queuing(cfgmap, vm, "shared-priority");
```
biddisco: There's a good reason why this scheduler was in a branch and not being pushed into master.

msimberg: I'm not sure if I'm missing something but I agree with you. To be clear, this is temporary.

biddisco: Has the dashboard shown any interesting effects either good/bad of making this the default?

sithhell: This would also mean that the shared-priority scheduler policy needs to be built always.

examples/resource_partitioner/shared_priority_queue_scheduler.hpp
```diff
@@ -0,0 +1,1550 @@
+//  Copyright (c) 2017-2018 John Biddiscombe
```
sithhell: Do we still need to have this as part of the examples since it is now a "real" scheduler?

src/util/command_line_handling.cpp
msimberg: I will still change it back to the old default before merging.

msimberg:
@sithhell That could make sense. Do you have something in mind already where that could be useful? Also, do you think other hints could be useful in the future? I could see the hint reflecting the whole machine topology at some point, but I'd rather not go there (in this PR at least).

examples/resource_partitioner/shared_priority_queue_scheduler.hpp
msimberg: Not a must, but we left it in to have at least one example for users to look at. Since it has all the logging etc it's also a bit easier to learn from.

sithhell:
Mikael Simberg <notifications@github.com> schrieb am Di., 28. Aug. 2018
11:50:

> @sithhell <https://github.com/sithhell> That could make sense. Do you
> have something in mind already where that could be useful? Also, do you
> think other hints could be useful in the future? I could see the hint
> reflecting the whole machine topology at some point, but I'd rather not go
> there (in this PR at least).
>

The use case I had in mind was that of stack reuse. We probably don't want
to reuse a stack that comes from a different numa domain.
Other hints I can think of are: pinned threads, that is threads that aren't
allowed to be stolen.

biddisco:
The idea behind the schedule_hint was to form a starting point upon which we can build. I would suggest that what we have in place is a working prototype that allows certain tasks to be steered towards certain resources.
We should leave it as it is and open a more feature-full and well planned issue for the next level of developments.
One of the things I'm hoping we get from a closer integration with something like Kokkos, is a better managment of thread affinities and interoperability with the executor/scheduler.
the development of this was driven by the lack of a way for the executors/schedulers to interact nicely and we clearly need to look long and hard at what the 'best' way to o it is.

msimberg:
I changed the default back to the local priority scheduler. This is good to go from my side.

PR #3421, opened 2018-08-17
Add CDash timing output to future overhead test (for graphs)

biddisco:
This adds `DartMeasurement` tags to the output so that timing can be plotted in CDash.

The idea behind this is that if we take some simple test that provides a useful benchmark, we ought to be able to see if atomic futures and scheduling fixes make a noticable difference over time.

### DIFF
--- a/tests/performance/local/future_overhead.cpp
+++ b/tests/performance/local/future_overhead.cpp
@@ -41,6 +41,10 @@ using hpx::util::high_resolution_timer;
 using hpx::cout;
 using hpx::flush;
 
+#define CDASH_TIMING(name,time) \
+    cout << "<DartMeasurement name=\"" << name << "\" " \
+    << "type=\"numeric/double\"> " << time << " </DartMeasurement> \n";
+
 ///////////////////////////////////////////////////////////////////////////////
 // we use globals here to prevent the delay from being optimized away
 double global_scratch = 0;
@@ -93,6 +97,7 @@ void measure_action_futures(std::uint64_t count, bool csv)
             "invoked {1} futures (actions) in {2} seconds\n",
             count,
             duration) << flush;
+    CDASH_TIMING("FutureOverhead_Actions", duration)
 }
 
 void measure_function_futures_wait_each(std::uint64_t count, bool csv)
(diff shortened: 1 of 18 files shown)


tests/performance/local/future_overhead.cpp
```diff
@@ -41,6 +41,10 @@ using hpx::util::high_resolution_timer;
 using hpx::cout;
 using hpx::flush;
 
+#define CDASH_TIMING(name,time) \
+    cout << "<DartMeasurement name=\"" << name << "\" " \
+    << "type=\"numeric/double\"> " << time << " </DartMeasurement> \n";
+
```
hkaiser: Could we turn this into a (inline) function (instead of a MACRO), please? Also, I think this could be moved into the header `hpx/util/lightweight_test.hpp`

biddisco:
I've added a utility and updated all existing tests that print CDash timing to use it. 
Currently only the sort/scan/reduce_by_key/etc tests that I was working on for VTKm are making use of it.

biddisco:
It seems that the benchmarks (like future_overhead) are not run as part of the testing process, so I will add them. see #3422 - then with luck we will at least get some baseline values of timing for tests on daint (or elsewhere) as part of the nightly testing.

biddisco:
I have added CDash compatible timing output to a number of benchmarks and included them in the tests suite for nightly and PR testing.

These tests have been excluded from the ctest rule because they do not run on HPX threads and the python run launcher adds `--hpx:threads=xxx` causing them to fail. They can be run by hand as before and there is no need right now to worry about them.
```
 delay_baseline
 delay_baseline_threaded
 function_object_wrapper_overhead
 nonconcurrent_fifo_overhead
 nonconcurrent_lifo_overhead
 print_heterogeneous_payloads
 ```

These tests either fail/segfault/abort, so they are excluded from testing until they are fixed. A separate issue will be opened for them
```
 print_heterogeneous_payloads
 hpx_homogeneous_timed_task_spawn_executors
 skynet
 timed_task_spawn
 hpx_tls_overhead
 native_tls_overhead
 coroutines_call_overhead
 serialization_performance
```

PR #3425, opened 2018-08-21
Use source cache and workspace between job steps on CircleCI

### LINKED ISSUE #3424: CircleCI builds broken for forked repositories
It seems that CircleCI has changed its policies on caching with forked repositories, leading to build failures on PRs from forked repositories: [https://discuss.circleci.com/t/saving-cache-stopped-working-warning-skipping-this-step-disabled-in-configuration/24423](https://discuss.circleci.com/t/saving-cache-stopped-working-warning-skipping-this-step-disabled-in-configuration/24423). Still unsure if we have to disable caching or if there is a proper fix for this.

msimberg:
Fixes #3424.

Note: this does a shallow local clone to avoid copying the whole 900 MB repository using the workspace between work steps. This is not strictly necessary and can be reverted if we ever want to query the git repo for more information than the current hash. I did this because copying the whole repo using the workspace was significantly slower than doing the same with the cache, but since there doesn't seem to be a way to save to the cache anymore on PRs from forked repos this is a small optimization to get it back to the same performance as before.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,25 +36,42 @@ jobs:
       - image: stellargroup/build_env:debian_clang
     working_directory: /hpx
     steps:
+      # Try to restore source cache. Keys are tried in order. The checkout step
+      # correctly handles the case when there is a cache entry.
+      - restore_cache:
+          keys:
+            - source-v2-{{ .Branch }}-{{ .Revision }}
+            - source-v2-{{ .Branch }}-
+            - source-v2-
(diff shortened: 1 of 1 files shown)


sithhell:
The problem with using workspaces was parallel updates. If that's fine now, this should be good

msimberg:
Yeah, I remember the problems. I don't think anything changed so it might still come back and haunt us, but for a few builds this has at least worked...

PR #3429, opened 2018-08-23
One thread per core

biddisco:
## Proposed Change
This adds a new `--hpx:threads=cores` option which will trigger creation of one thread per core, as opposed to `--hpx:threads=all` that  creates one thread per PU. On systems with hyperthreads, it is usually the case that N threads per core gives worse performance than one.

A second patch in the PR makes `--hpx:threads=cores` the default instead of `all`

When converting one of the solvers to HPX here at CSCS it is better to have a default (i.e. no command line specified at all) that gives one thread per core - this means that users who try it out do not find the hpx version gives much worse performance than the old one due to hyperthreads competing for resources.


biddisco:
I'd be interested in knowing if @hkaiser approves of this PR because it is a potentially breaking change in terms of people's daily use of HPX

msimberg:
If this goes in the documentation should be updated as well ([here](https://github.com/STEllAR-GROUP/hpx/blob/63c3ef8f3caa8c7e6ee426fd6fe818eb76117163/docs/manual/commandline.qbk#L65-L67), and this one is already outdated before this PR; and [here](https://github.com/STEllAR-GROUP/hpx/blob/63c3ef8f3caa8c7e6ee426fd6fe818eb76117163/docs/manual/config_defaults.qbk#L97-L99)).

biddisco:
Docs updated

docs/manual/commandline.qbk
```diff
@@ -63,8 +63,10 @@ described in the table below:
     [[`--hpx:pu-step`]          [the step between used processing unit numbers for this
                                  instance of __hpx__ (default: 1)]]
     [[`--hpx:threads arg`]      [the number of operating system threads to spawn
-                                 for this __hpx__ locality (default: 1, using 'all' will
-                                 spawn one thread for each processing unit)]]
+                                 for this __hpx__ locality (default: 'cores')
+                                 valid entries are numeric values 1,2,3... or 'all' which will
+                                 spawn one thread per processing unit (includes all hyperthread 
+                                 cores), 'cores' will spwan one thread per core (no hyperthreads).
```
msimberg: Tyop: spwan.

msimberg:
@biddisco thanks!

biddisco:
@msimberg I've added docs for numa-balanced - no idea if the image will look any good in the new help. It's a lot bigger than the old one.

biddisco:
Grrr. messed up the image. force pushing again in 2 mins

msimberg:
@biddisco thanks for updating the numa-balanced docs. I always wondered how it was different... There's not technical reason the image wouldn't look good in the new docs even if it's bigger. Could I just ask you to remove the red underlines from "numa" if you still have the original around?

sithhell:
The picture still seems to be messed up. The word numa is underlined.

biddisco:
Fixed the picture. Spell checked underlined it without me noticing.
Please merge as soon as you are happy - I'm fed up with this PR :)

msimberg:
Added a couple of small fixes. Looks good to me at least and works as advertised.

PR #3432, opened 2018-08-24
Don't track task dependencies for distributed runs

khuck:
Because the parent_task pointer isn't reliable (even when the
parent and child tasks have the same locality) don't track task
dependencies when performing distributed runs.

Fixes failing distributed runs in the phylanx unit tests when configured with APEX

## Proposed Changes

  - don't dereference the parent task pointer when num_localities > 1


biddisco:
Does this mean that we can't use APEX at all distributed mode, or only that certain parent/child relationships are not available when running distributed?

hkaiser:
Frankly, I'd rather fix things than finding workarounds (at least in mid-term). How can we reproduce the problems you were trying to circumvent?

khuck:
@biddisco It just means that the parent_task pointer is unreliable in distributed runs, so we can't track the dependencies across localities.  Everything else would still be functional.
@hkaiser There are two issues - when the child locality and parent locality aren't the same, we can't dereference the parent task to get the APEX data stored with the parent - I don't see a way to fix this, other than sending a deep copy of the parent's APEX data with the remote task request. The second issue is that even when the parent and child localities are the same, the parent task pointer has already been freed when the child is instantiated (and/or the parent pointer is bogus).  I don't see these problems with a non-distributed run, but that could be a red herring (there are some actions that don't happen in single-node executions).  My motivation with this fix is just to disable the feature for distributed runs so that tests don't fail and other unrelated work (i.e. parcel coalescing) can proceed without hacking up the code first.

hkaiser:
> @hkaiser There are two issues - when the child locality and parent locality aren't the same, we can't dereference the parent task to get the APEX data stored with the parent - I don't see a way to fix this, other than sending a deep copy of the parent's APEX data with the remote task request. 

Couldn't this be resolved/consolidated at the end? APEX stores all the information about the tasks it has seen anyways, doesn't it?

> The second issue is that even when the parent and child localities are the same, the parent task pointer has already been freed when the child is instantiated (and/or the parent pointer is bogus). 

Yeah, I was wondering all the time how you circumvent this particular problem. This could be rectified however if we hooked APEX into the task registration instead of the task creation. Alternatively, we wait for @sithhell to remove the task registration altogether (he is planning to do that soon) at which point the tasks are created right away, guaranteeing that the parent task is still alive.

> I don't see these problems with a non-distributed run, but that could be a red herring (there are some actions that don't happen in single-node executions). 

This is most probably a red herring, even if I don't understand why you don't see this in non-distributed runs. 

> My motivation with this fix is just to disable the feature for distributed runs so that tests don't fail and other unrelated work (i.e. parcel coalescing) can proceed without hacking up the code first.

I understand. I would at least like to see another ticket though, reminding us that this needs fixing.

khuck:
@hkaiser APEX can't resolve this at the end of execution - here's the full explanation:

in thread_init_data, APEX stores an apex_task_wrapper object, which is a typedef of ```std::shared_ptr<apex::task_wrapper> apex_task_wrapper```.  That object contains an ```apex::task_identifier``` pointer (either its name and/or address), an ```apex::profiler``` pointer (only if the task is currently executing), a GUID (generated by APEX or provided by the runtime) and a smart pointer to the parent ```apex::task_wrapper```.

The GUID identifies the specific instance of that task.  The task_identifier identifies what type of task it is (i.e. the HPX description).

For the OTF2 support of task dependencies, all that is needed is the GUID of the parent.  To generate the task dependency graph of task types, only the task_identifier is needed.

Regardless, the apex_task_wrapper of the parent is not available without dereferencing the parent_id pointer, which is of the type ```hpx::threads::thread_id_type```, which has only one data member, a pointer to ```hpx::threads::thread_data``` (which contained the parent's apex_task_wrapper, but is now either gone or in another process space).

APEX *could, theoretically* maintain a map of parent_id addresses to ```apex::task_wrapper``` pointers.  But that would have to be shared across all threads in a process, and would be a bottleneck - it would have to be locked every time a task was created and/or destroyed.  But maybe this is a solution if/when the user wants to track dependencies.  I'd have to implement it to see how much of a bottleneck it would be.

Regardless, I am more than happy to create a new issue so we don't lose track of the fact that task dependencies can't be tracked across localities.  I think there's a solution here, but I don't know all the details of how tasks are created, let alone across localities...

khuck:
@hkaiser Is there something that I am supposed to do to help get this pull request merged?  The two CI failures are unrelated, as near as I can tell. I don't know why the other 3 didn't report back.

khuck:
@hkaiser all ok now?

PR #3434, opened 2018-08-30
Fix threads cores

biddisco:
Fixes a mistake introduced in #3429, hpx:threads=all was not working as expected and one test was failing due to a problem with hpxrun.py

### DIFF
--- a/src/util/command_line_handling.cpp
+++ b/src/util/command_line_handling.cpp
@@ -355,6 +355,8 @@ namespace hpx { namespace util
                 threads_str = vm["hpx:threads"].as<std::string>();
                 if ("all" == threads_str)
                 {
+                    default_threads = thread::hardware_concurrency();
+                    batch_threads = env.retrieve_number_of_threads();
                     if (batch_threads == std::size_t(-1))
                     {
                         batch_threads = thread::hardware_concurrency();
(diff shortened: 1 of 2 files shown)


src/util/command_line_handling.cpp
```diff
@@ -355,6 +355,8 @@ namespace hpx { namespace util
                 threads_str = vm["hpx:threads"].as<std::string>();
                 if ("all" == threads_str)
                 {
+                    default_threads = thread::hardware_concurrency();
+                    batch_threads = env.retrieve_number_of_threads();
```
hkaiser: This is a change to the original implementation of `--hpx:threads=all`. Could you give a rationale for why it is needed?

sithhell: 'all' always considered the batch environment previously.

hkaiser: @sithhell Sure, but not at this spot. Did we miss to properly handle the environment earlier?

biddisco: The reason is because if you look a couple of lines above, the default and batch threads are set to those values, but since "threads=cores" is the new default, they immediately get reset to something else, so then inside the `if == all` clause I reset them back to what they would have been if I hadn't made `cores` the new default. The code is a bit messy and probably needs a rewrite, but that should go into a separate issue I think. (also the command line handling code is called twice and I'm puzzled as to why it needs to be done twice, so that should be looked at at the same time)

hkaiser: Thanks for the explanations, makes perfect sense.

PR #3435, opened 2018-08-30
Fixing 1205

### LINKED ISSUE #1205: TSS is broken
The TSS related tests currently fail. This means tss is non-functional and disabled by default for now.

sithhell:
Fixes #1205

## Proposed Changes

  - Fixing double delete problem:
    - This adds a move ctor and move assignment for tss_node to avoid double deletes
    in case of temporary objects.
  - Unbreaking unit test:
    - The test had a timing issues. The TSS dtors are called after the task making
    the future ready returns.
  - Properly reseting the coroutine context to avoid an assertion being triggered.

### DIFF
--- a/hpx/runtime/threads/coroutines/detail/tss.hpp
+++ b/hpx/runtime/threads/coroutines/detail/tss.hpp
@@ -51,6 +51,25 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail
             value_(val)
         {}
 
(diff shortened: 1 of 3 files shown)


hkaiser:
@tianyi93 this fixes the HPX thread local storage problem that could be causing the hpxMP issues you're seeing.

tianyi93:
@hkaiser Thanks, I will try to see if it can solve those issues!

Issue #1668, opened 2015-07-18
Adapt all parallel algorithms to Ranges TS


hkaiser:
The proposed ranges library includes slight (mostly non-breaking) changes to the standard algorithms. Those are described in N4569: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/n4569.pdf

Edit: here is an updated version of this document: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4685.pdf

We should adapt our algorithms to conform to this proposal (as much as possible).

Here is the list of algorithms which need to be adapted:
- [x] `adjacent_difference` (#5564)
- [x] `adjacent_find`  (#4996)
- [x] `all_of` `any_of` `none_of` (#2985)
- [x] `copy` `copy_if` `copy_n` (#1911)
- [x] `move` (#3028)
- [x] `count` `count_if` (#3169)
- [x] `equal` (#4883) `mismatch` (#4884)
- [x] `exclusive_scan` (#5436) `inclusive_scan` (#5439)
- [x] `reduce` 
- [x] `transform` (see #1671)
- [x] `fill` `fill_n` (see #3194)
- [x] `find` `find_end` `find_first_of` `find_if` `find_if_not` (#4885)
- [x] `for_each` `for_each_n` (see #1671)
- [x] `generate` `generate_n` (see #1968)
- [x] `is_heap` `is_heap_until` (see #2788)
- [x] `is_partitioned` (#5006)
- [x] `is_sorted` `is_sorted_until` (#5025)
- [x] `lexicographical_compare` (#5350)
- [x] `max_element` `min_element` `minmax_element` (#1968)
- [x] `make_heap` (#4964)
- [x] `sort` (#1911)
- [x] `partial_sort` `partial_sort_copy` (#5630)
- [x] `nth_element` (#5592)
- [x] `stable_sort` (#4817)
- [x] `partition` (#2778, #5540)
- [x] `partition_copy` (see #2716, #5540) 
- [x] `stable_partition` (#5540)
- [x] `remove` `remove_if` (see #3086)
- [x] `remove_copy` `remove_copy_if` (#1922)
- [x] `replace` `replace_copy` `replace_copy_if` `replace_if` (#1922)
- [x] `reverse` `reverse_copy` (#1922)
- [x] `rotate` `rotate_copy` (#1922)
- [x] `search` `search_n` (#5066)
[...]

PR #3436, opened 2018-08-31
Add projection function to find_* (and fix very bad bug)

brjsp:
This started as an attempt to make find_* Ranges TS compliant (see #1668). But during testing, i discovered and fixed a bad bug in find_end making it unusable with custom comparison operator.
Because of this bug, i am making this pull request now, before writing a container version of the above algorithms. The regular versions are already usable as given in this pull request.

### DIFF
--- a/docs/CMakeLists.txt
+++ b/docs/CMakeLists.txt
@@ -118,6 +118,7 @@ set(doxygen_dependencies
     "${PROJECT_SOURCE_DIR}/hpx/parallel/container_algorithms/copy.hpp"
     "${PROJECT_SOURCE_DIR}/hpx/parallel/container_algorithms/count.hpp"
     "${PROJECT_SOURCE_DIR}/hpx/parallel/container_algorithms/fill.hpp"
+    "${PROJECT_SOURCE_DIR}/hpx/parallel/container_algorithms/find.hpp"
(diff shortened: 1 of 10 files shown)


hkaiser:
Thanks a lot!

Please also fix the inspect errors reported here: https://30382-4455628-gh.circle-artifacts.com/0/hpx_inspect_report.html

brjsp:
After a discussion with @K-ballo on chat, i have removed plain `find` from the pull request, because it is really hard to do correctly (and in a way not requiring C++14).

brjsp:
Actually doing a projection version of find_if and find_if_not would mean having to do also do this for the segmented algorithms, so not doing it now.

hkaiser:
@brjsp thanks for working on this. I have not followed all the discussion around this - could you give a short overview about the current state of this PR, please?

brjsp:
@hkaiser find_end and find_first_of have added projection and range version, the bug in find_end is fixed, other functions are untouched, should be ready for merging.

msimberg:
@hkaiser yeah, this looks safe to merge. @brjsp could I just ask you to add `hpx/parallel/container_algorithms/find.hpp` [here](https://github.com/STEllAR-GROUP/hpx/blob/1458fcc0c1f58135a6f3563701b784eb96838cf0/docs/CMakeLists.txt#L121). Thanks a lot for your work!

PR #3439, opened 2018-09-03
Use correct MPI CXX libraries for MPI parcelport

cgarth:
## Proposed Changes

  - Use MPI_CXX_LIB_NAMES instead of MPI_C_LIB_NAMES to identify MPI parcelport link libraries

- This fixes compilation with HPX_WITH_PARCELPORT_MPI=ON on stock Ubuntu 18.04 (OpenMPI 2.1.1 and CMake 3.10.2).

### DIFF
--- a/plugins/parcelport/CMakeLists.txt
+++ b/plugins/parcelport/CMakeLists.txt
@@ -64,6 +64,10 @@ macro(add_parcelport name)
       ${HPX_STATIC_PARCELPORT_PLUGINS_SOURCES}
       CACHE INTERNAL "" FORCE)
 
+    set_source_files_properties(${${name}_SOURCES}
+      PROPERTIES COMPILE_FLAGS
+      "${${name}_COMPILE_FLAGS}")
+
     hpx_libraries(${${name}_DEPENDENCIES})
 
   else()
--- a/plugins/parcelport/mpi/CMakeLists.txt
+++ b/plugins/parcelport/mpi/CMakeLists.txt
@@ -41,6 +41,8 @@ if(HPX_WITH_PARCELPORT_MPI)
   endif()
 
   macro(add_parcelport_mpi_module)
+    # find MPI again to re-populate variables in this context
+    find_package(MPI)
     hpx_debug("add_parcelport_mpi_module")
     if(MPI_CXX_INCLUDE_PATH)
       include_directories(${MPI_CXX_INCLUDE_PATH})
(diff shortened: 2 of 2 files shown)


cgarth:
Upon further investigation it turns out that the Ubuntu 18.04 OpenMPI package is configured with "--enable-mpi-cxx". Including <mpi.h> in C++ code (but not actually using C++ bindings) creates a dependency on libmpi_cxx, which leads to missing symbols during link.

Rebuilding the package without "--enable-mpi-cxx" solves this for me, as does defining "-DOMPI_SKIP_MPICXX" in CFLAGS.

biddisco:
Out of curiosity - if you did
```
extern "C" {
#include <mpi.h>
}
```
would that stop mpi from adding cxx linkage?

cgarth:
No, it does not, or rather, it produces (many) other errors. Relevant part from my mpi.h:
```/*
 * Conditional MPI 2 C++ bindings support.  Include if:
 *   - The user does not explicitly request us to skip it (when a C++ compiler
 *       is used to compile C code).
 *   - We want C++ bindings support
 *   - We are not building OMPI itself
 *   - We are using a C++ compiler
 */
#if !defined(OMPI_SKIP_MPICXX) && OMPI_BUILD_CXX_BINDINGS && !OMPI_BUILDING
#if defined(c_plusplus) || defined(__cplusplus)
#include "openmpi/ompi/mpi/cxx/mpicxx.h"
#endif
#endif
```

Apparently, `__cplusplus` is still defined under `extern "C"`, which leads to "conflicting declaration" errors.

In case it helps, here's a minimal test case wrapped in a Dockerfile:
```
FROM ubuntu:18.04

ENV DEBIAN_FRONTEND=noninteractive

RUN apt-get update      \
 && apt-get install -y  \
        g++		\
        libopenmpi-dev

WORKDIR /root

RUN /bin/echo -e 'extern "C" {\n#include <mpi.h>\n}' > mpitest.cpp
RUN g++ $(mpicc --showme:compile) mpitest.cpp -o mpitest $(mpicc --showme:link)
```

sithhell:
If the standard Ubuntu package leads to a link failure, then that's bad. We should have that fixed. What would be the best way? Add the macro or the libraries?

cgarth:
More info: CMake (>= 3.10.3) allows to handle this situation through an option `MPI_CXX_SKIP_MPICXX`, which in the above setup produces
```
MPI_CXX_COMPILE_FLAGS: -pthread -DMPICH_SKIP_MPICXX -DOMPI_SKIP_MPICXX -D_MPICC_H
```
and thus disables C++ support in OpenMPI's `<mpi.h>`.

However, these flags are not propagated in `plugins/parcelport/mpi/CMakeLists.txt`. They get passed to `hpx_add_compile_flag()`, but are not used in the actual compilation.

biddisco:
Oops. I didn't realize there was a patch applied here too. I submitted another one elsewhere.

biddisco:
The status of this PR is that if #3458 is accepted, then we can close this one. If it is rejected, then we should pursue this one further. They both solve the issue, but via slightly different approaches.

sithhell:
Can we close this now?

hkaiser:
This can be closed now as #3458 has been merged.

PR #3444, opened 2018-09-06
Pass special MPI compile definitions from parcelport to main hpx library

biddisco:
Fixes #3439

## Proposed Changes
This is intended to allow the user to compile hpx with settings like
cmake -DMPI_C_COMPILER=$HOME/apps/mpich/bin/mpicc -DMPI_CXX_SKIP_MPICXX=ON ...
and have the MPI defined MPICH_SKIP_MPICXX OMPI_SKIP_MPICXX definitions
passed through to the hpx library where they are needed

### DIFF
--- a/plugins/parcelport/CMakeLists.txt
+++ b/plugins/parcelport/CMakeLists.txt
@@ -25,7 +25,7 @@ macro(add_parcelport name)
   set(name "parcelport_${name}")
   set(options STATIC)
   set(one_value_args FOLDER)
-  set(multi_value_args SOURCES HEADERS DEPENDENCIES COMPILE_FLAGS LINK_FLAGS)
+  set(multi_value_args SOURCES HEADERS DEPENDENCIES COMPILE_FLAGS COMPILE_DEFINITIONS LINK_FLAGS)
   cmake_parse_arguments(${name} "${options}" "${one_value_args}" "${multi_value_args}" ${ARGN})
 
   if(${${name}_STATIC})
@@ -65,6 +65,7 @@ macro(add_parcelport name)
       CACHE INTERNAL "" FORCE)
 
     hpx_libraries(${${name}_DEPENDENCIES})
+    hpx_add_target_compile_definition("${${name}_COMPILE_DEFINITIONS}")
 
   else()
     hpx_debug("adding parcelport plugin: ${name}")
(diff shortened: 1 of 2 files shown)


biddisco:
I'd be grateful if @cgarth would see if this makes his compilation work the way it should.

biddisco:
There seems to be a problem with this. It's adding link flags as well as compile defs and it should not. I will try to fix again.

cgarth:
Hi John, debugging this further (offline) today I found a very similar solution, with one additional detail: I have to re-`find_package(MPI)` at the top of `add_parcelport_mpi_module()`, otherwise all `MPI_...` variables are unset in the macro. 

Updated #3439.

biddisco:
Turns out this patch was fine, but I had corrupted my cache with failed attempts earlier, hence the dodgy link flags.

I'm not sure what to do, you have fixed it in #3439 - but I don't like the fact that we call find_package twice. It should not be necessary. My fix, fixes it with a slight variation.

However It turns out we already "do the right thing" here 
https://github.com/STEllAR-GROUP/hpx/blob/457670e0581fe03ee4b4b4b12b0a757eff445307/plugins/parcelport/mpi/CMakeLists.txt#L24
but it doesn't seem to work as expected, so I think in actual fact, the bug is simply a dodgy or incorrect use of macro. The comment above that line, indicates that something was known to not be working properly, so I'll have a look at seeing what's wrong there.

cgarth:
Confirmed that this fixes compilation on stock Ubuntu 18.04 if -DMPI_CXX_SKIP_MPICXX=ON is passed. Thanks!

Would it be acceptable to set(MPI_CXX_SKIP_MPICXX ON) by default? This would make master compile without any special flags, i.e. "cmake ; make" would just work on systems with OpenMPI C++ bindings enabled.

biddisco:
I realize on inspection that you identified that the `hpx_add_compile_flag` line was the problem in your original patch. I understand the problem now
hpx_add_compile_flag sets
`set(CMAKE_${_lang}_FLAGS${_conf} "${CMAKE_${_lang}_FLAGS${_conf}} ${FLAG}")`
which really means
`CMAKE_CXX_FLAGS_RELEASE/DEBUG/OTHER` but this is a cache var, and setting it in the macro location, loses it elsewhere as it is not propagated. So we should really add the flags to a GLOBAL PROPERTY variable as is done in most of the other macros, and then set it just on the flags when the target is added.

biddisco:
This has been superceded by #3458

PR #3445, opened 2018-09-07
Fix various warnings

msimberg:
Fixes:

- Warnings about missing `override` in `shared_priority_queue_scheduler`
- Deprecation warnings in `topology.cpp`

Flyby: reset counters in `shared_priority_queue_scheduler` in `reset_thread_distribution`.

### DIFF
--- a/src/runtime/threads/topology.cpp
+++ b/src/runtime/threads/topology.cpp
@@ -516,8 +516,9 @@ namespace hpx { namespace threads
 
         {
             std::unique_lock<hpx::util::spinlock> lk(topo_mtx);
-            int ret = hwloc_get_area_membind_nodeset(topo,
-                reinterpret_cast<void const*>(lva), 1, nodeset, &policy, 0);
+            int ret =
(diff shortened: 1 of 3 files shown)


src/runtime/threads/topology.cpp
```diff
@@ -837,7 +838,7 @@ namespace hpx { namespace threads
     {
         hwloc_bitmap_t cpuset  = mask_to_bitmap(mask, HWLOC_OBJ_PU);
         hwloc_bitmap_t nodeset = hwloc_bitmap_alloc();
-        hwloc_cpuset_to_nodeset_strict(topo, cpuset, nodeset);
+        hwloc_cpuset_to_nodeset(topo, cpuset, nodeset);
```
biddisco: It may not be valid for people using older hwloc. We'll see on the dashboards (though this code probably isn't used much)

hkaiser:
I think it's too early to abandon HWLOC V1.x. Too many systems come with that version by default, still.

PR #3446, opened 2018-09-07
Fix some thread suspension timeouts

msimberg:
Adds an early return to `get_next_thread` in `local_(priority_)queue_scheduler` when
suspending worker thread (i.e. when not running).

The old behavior was to still try to steal pending threads from other worker threads. The current behavior is similar to `wait_or_add_new`, which already had an early return to not steal staged threads
when suspending. This should fix worker thread suspension cases where the HPX thread trying to suspend a worker thread would get scheduled on another thread (correctly) but then be stolen back immediately by the worker thread being suspended before another worker thread has the chance to run it. This should probably have been there all along.

Quick testing showed an immediate improvement but I'll keep this WIP until I've soak tested it a bit longer (and I'm sure it doesn't break other tests again).

Found a few more problems, most of them related to termination on an error:

- `scheduler_base::select_active_pu` could get stuck in an infinite loop if all threads have stopped (e.g. because of an error), now it falls back to less strict checks for the thread state if no thread has a suitable state.
- `scheduled_thread_pool::stop_locked/remove_processing_unit_internal` used to set the thread state to `state_stopping` even if an error occurred. The worker thread would then not stop from `state_stopping` if there are suspended threads left and nothing to schedule the suspended threads.

- (unrelated) `scheduled_thread_pool::get_os_thread_count` calls `threads_.size()` without a lock which can happen at the same time as a `threads_.resize()` (with a lock). I can't confirm that this actually is a problem, but it smells like the problem of very rare hangs during startup. I changed this to return the atomic `thread_count_` instead, which is in any case more appropriate (since `threads_.size()` doesn't have to reflect how many threads exist at the moment, although right now they are the same modulo race conditions since we don't fully remove worker threads dynamically).

### DIFF
--- a/hpx/runtime/threads/detail/scheduled_thread_pool.hpp
+++ b/hpx/runtime/threads/detail/scheduled_thread_pool.hpp
@@ -156,7 +156,7 @@ namespace hpx { namespace threads { namespace detail
 
         std::size_t get_os_thread_count() const
         {
-            return threads_.size();
+            return thread_count_;
         }
 
         std::size_t get_active_os_thread_count() const
--- a/hpx/runtime/threads/detail/scheduled_thread_pool_impl.hpp
+++ b/hpx/runtime/threads/detail/scheduled_thread_pool_impl.hpp
@@ -141,7 +141,7 @@ namespace hpx { namespace threads { namespace detail
     void scheduled_thread_pool<Scheduler>::report_error(
         std::size_t num, std::exception_ptr const& e)
     {
-        sched_->Scheduler::set_all_states(state_terminating);
+        sched_->Scheduler::set_all_states_at_least(state_terminating);
         this->thread_pool_base::report_error(num, e);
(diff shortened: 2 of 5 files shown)


msimberg:
While this does seem to help, it's not the end of the story. There's still something else causing timeouts, so I'm still looking into this.

hkaiser:
@msimberg I'm seeing regular timeouts on circleci for #3406. AFAICS, all hangs happen for tests that are somehow related to the partitioner. Do you think your fixes here are related?

msimberg:
@hkaiser hard to say. My feeling is these changes won't help, but I wouldn't mind having a look at your branch. I'll try it out later today.

msimberg:
Updated this with a few more fixes, but more soak testing to follow.

msimberg:
Any objections to merging this? I could not get the resource tests to fail locally anymore after the latest changes and a few hours of soak testing, both Debug and Release.

PR #3450, opened 2018-09-07
Constexpr-fying bind and other functional facilities

K-ballo:


msimberg:
@K-ballo this time it looks like pycicle is complaining for a reason. It seems like gcc 7.3 (C++ 17) has problems while gcc 6.2 (C++ 17), gcc 4.9 (C++ 11), and clang 3.8 (C++ 11) are okay. Clang 6.0 timed out so could be a problem there as well.

PR #3457, opened 2018-09-12
Fixing more size_t -> int16_t (and similar) warnings

hkaiser:


hkaiser:
@msimberg could we merge this, please? Those warnings I get are just annoying...

PR #3458, opened 2018-09-13
Fix a problem with scope of CMAKE_CXX_FLAGS and hpx_add_compile_flag

biddisco:
CMAKE_CXX_FLAGS_XXX is only valid at directory scope, so flags added
in subdirs are not passed up to the parent. Modify the
add_compile_flag macro to set a global property with flags that can
be read back in the base dir and passed to the executables/libs.

This fixes PR #3439

In the future a target_compile_option would be preferable, but
some more fixes would be required.


biddisco:
@sithhell @cgarth This is the 'correct' fix for the MPI_CXX flags problem (I think). Please feel free to test/comment as you see fit.

biddisco:
The macro `hpx_remove_target_compile_definition` was removed because it was not used anywhere in the codebase and it was no longer applicable with the changes I made to `hpx_add_compile_flag`. 
I have not removed `hpx_target_compile_flag` that's just a bad choice of diff display. I modified the signature of `hpx_add_compile_flag` which was incorrect. A couple of the other macros are also slightly wrong, but I do not want to change them in this PR and might fix them later. Since they are basically working at the moment I don't consider it a high priority.

biddisco:
Anyone want to approve/reject this?

hkaiser:
The failing test is new to this PR. Not sure that I understand it, though... @biddisco could you have another look, please.

biddisco:
Thanks for pointing out the failing test. I have applied a fix that appears to work locally. Hopefully it is fixed on circleci too. I don't usually build the headers tests, I should probably enable them on cdash builds.

PR #3463, opened 2018-09-23
- cmake : create the parent directory before symlinking

### LINKED ISSUE #3462: CMake configuration shows a minor and inconsequential failure to create a symlink
## Expected Behavior

When running cmake, the output should be clean without any errors.

## Actual Behavior

Instead, we get a error saying `failed to create symbolic link '/home/jayesh/build/hpx/share/hpx': No such file or directory` which can be seen in the full output [here](https://gist.github.com/jayeshbadwaik/005cf0c5426a91fe1995e37be9f3dc8b). This happens because the parent directory of the target directory does not yet exist. Ultimately, the error is not fatal and the install continues correctly. However, it is annoying.

The fix is minor as shown [here](https://github.com/jayeshbadwaik/hpx/commit/e1c7ab4e953ab60bcc68dda2110a1d2682e608aa).  

  1. `git clone git@github.com:STEllAR-GROUP/hpx.git`
  2. `mkdir build && cd build`
  3. `cmake ../hpx`

## Specifications
  - HPX Version: master (git commit: 61e7a441e4a)
  - Platform (compiler, OS): ArchLinux, gcc-8.2

jayeshbadwaik:
The share directory is not created when the cmake tries to symlink
    HPX installation to share directory. This creates a minor and
    eventually inconsequential error in the configuration as shown
    below:

	failed to create symbolic link '/home/jayesh/build/hpx/share/hpx': No such file or directory

    This commit fixes it by creating a directory beforehand.

Fixes #3462

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1839,6 +1839,12 @@ endif()
 add_plugin_modules()
 
 ################################################################################
+# Prepare the Build Directory
+################################################################################
+file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/share)
+
+
+################################################################################
 # Configure the header to include all compile definitions
 ################################################################################
 generate_config_defines_header(${CMAKE_BINARY_DIR})


msimberg:
The CircleCI install failure is unrelated. I'll open a separate PR to fix that.

jayeshbadwaik:
Thanks :-)

PR #3465, opened 2018-09-24
This patch adds various optimizations extracted from the thread_local_allocator work

hkaiser:
- make AGAS properly (fast-)resolve local-only ids
- avoid reading atomic in future state  more than necessary
- marking certain hot-spot functions as force-inline
- extend `future_access_customization_point` to support addref-control while creating a future from a shared state
- `unwrapping_result_policy` now holds id by reference
- flyby: fix warnings in `when_each` test
- flyby: split `partitioned_vector_exclusive_scan` test

This also adds overloads for APIs that produce futures taking an allocator (dataflow, make_ready_future, etc.)

This may supersede #3406 if we decide not to add the `thread_local_allocator` proposed there.

### DIFF
--- a/hpx/lcos/dataflow.hpp
+++ b/hpx/lcos/dataflow.hpp
@@ -50,6 +50,7 @@
 #include <exception>
 #include <functional>
 #include <iterator>
+#include <memory>
 #include <type_traits>
 #include <utility>
 
@@ -365,6 +366,30 @@ namespace hpx { namespace lcos { namespace detail
     }
 
     ///////////////////////////////////////////////////////////////////////////
+    template <
+        typename Allocator, typename Policy, typename Func, typename ...Ts,
+        typename Frame = dataflow_frame<
+            typename std::decay<Policy>::type,
+            typename std::decay<Func>::type,
+            util::tuple<typename std::decay<Ts>::type...>>>
+    typename Frame::type create_dataflow_alloc(
+        Allocator const& alloc, Policy && policy, Func && func, Ts &&... ts)
+    {
+        // Create the data which is used to construct the dataflow_frame
+        auto data = Frame::construct_from(
+            std::forward<Policy>(policy), std::forward<Func>(func));
+
+        // Construct the dataflow_frame and traverse
+        // the arguments asynchronously
+        boost::intrusive_ptr<Frame> p = util::traverse_pack_async_allocator(
+            alloc, util::async_traverse_in_place_tag<Frame>{},
+            std::move(data), std::forward<Ts>(ts)...);
+
+        using traits::future_access;
+        return future_access<typename Frame::type>::create(std::move(p));
+    }
+
+    ///////////////////////////////////////////////////////////////////////////
     template <typename FD, typename Enable = void>
     struct dataflow_dispatch;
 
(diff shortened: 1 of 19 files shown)


hpx/lcos/dataflow.hpp
```diff
@@ -477,6 +570,22 @@ namespace hpx { namespace lcos { namespace detail
...
+        template <
+            typename Allocator, typename F, typename ...Ts,
+            typename Enable = typename std::enable_if<
+                !traits::is_action<typename std::decay<F>::type>::value
+            >::type>
+        HPX_FORCEINLINE static auto
+        call_alloc(Allocator const& alloc, F && f, Ts &&... ts)
+//         ->  decltype(dataflow_dispatch<launch>::call_alloc(
+//                 alloc, launch::async, std::forward<F>(f),
+//                 std::forward<Ts>(ts)...))
```
sithhell: any reason this is commented? It seems that this line is problematic on older clangs.

hpx/lcos/detail/future_data.hpp
```diff
@@ -400,9 +400,9 @@ namespace detail
         ///               error description if <code>&ec == &throws</code>.
         virtual result_type* get_result(error_code& ec = throws)
         {
-            if (!get_result_void(ec))
-                return nullptr;
+            if (get_result_void(ec) != nullptr)
             return reinterpret_cast<result_type*>(&storage_);
+            return nullptr;
```
sithhell: This is messed up. Doesn't look right at all.

hpx/runtime/components/unwrapping_result_policy.hpp
```diff
@@ -29,35 +29,17 @@ namespace hpx { namespace components
...
-        /// Create a new \a unwrapping_result_policy representing the
-        /// target locality
-        ///
-        /// \param id     [in] The global address of the target object
-        ///
-        unwrapping_result_policy operator()(id_type const& id) const
-        {
-            return unwrapping_result_policy(id);
-        }
+        unwrapping_result_policy(id_type const& id)
```
sithhell: I don't particularly like this change. This leads to a copy of the id_type. Wouldn't it we better if we move it into the async call instead? What about lifetime?

hpx/lcos/dataflow.hpp
hkaiser: Sorry, it's a left-over. Will fix.

hpx/lcos/detail/future_data.hpp
hkaiser: Merge conflict... WIll fix - thanks for spotting.

hpx/runtime/components/unwrapping_result_policy.hpp
hkaiser: No, unwrapping_policy holds the id by reference. No copy is happening anymore (at least that is the intent...).

sithhell: The copy is happening in the call to async (and friends).

hkaiser: Huh? Could you elaborate?

sithhell: These calls: https://github.com/STEllAR-GROUP/hpx/pull/3465/files#diff-a1d7921ce43cab0e4a7fb54f51558858R85

Do capture the id, don't they? I might be totally off here, but it looks like it could be moved there as well.

hkaiser: @sithhell so do I understand correctly that you propose to convert the `unwrapping_result_policy` into a one-shot facility, allowing to move the `id_type` to be possibly moved through into the function object created by async? If so, that would require that we can move the initial `id_type` instance as well  (the one the `unwrapping_result_policy` is created from), which is not the case most of the time, I belive.

hkaiser: After this discussing this offline, @sithhell agreed for this to go forward as is.

hkaiser:
@msimberg would you mind merging this at your earliest convenience? @sithhell has approved it.

biddisco:
Mikael is on vacation, so I'll take over

hkaiser:
@biddisco thanks!

PR #3467, opened 2018-09-25
Add compiler specific fallthrough attributes if C++17 attribute is not available

msimberg:
These warnings come up when building HPX with C++11 or 14 and a new compiler such as gcc 7.


hkaiser:
The remaining error is known and unrelated.

PR #3469, opened 2018-09-27
Enable (existing) code for extracting stack pointer on Power platform

hkaiser:
@khuck this might solve the stackoverflows we're seeing on your Power9 platform. Please verify.


hkaiser:
This still does not solve the issues we're seeing (https://github.com/STEllAR-GROUP/phylanx/issues/584). We need to keep investigating things...

khuck:
@sithhell @hkaiser since this was merged into master on Feb 1, I am seeing the error return for Power8 when compiled with GCC 7.1 or Clang 7 - for all HPX programs, even simple hello world or fibonacci:

```
{what}: Stack overflow: HPX(out_of_memory)

  what():  Stack overflow: HPX(out_of_memory)
./build-hpx.sh: line 70: 33088 Aborted                 (core dumped) ./bin/fibonacci
```

The HPX commit right before this one 5364e25ddaeff6e0b82420af1acb7b175c50ea2b works fine.

hkaiser:
@khuck either our stack overflow detection is wrong or the stack overflow is real. How do we find out?

khuck:
@hkaiser The default for Linux is off, and I don’t set it.  Should I?

khuck:
@hkaiser enabling HPX_WITH_STACKOVERFLOW_DETECTION gives the same result.

hkaiser:
@khuck: @sithhell knows all of this code better than I. Let's see what he has to say.

sithhell:
This is very odd, I'll have to run it myself and see what's happening then.

PR #3478, opened 2018-10-03
Making sure the application's exit code is properly propagated back to the OS

hkaiser:
This fixes the issue that some tests don't report their errors back to CircleCI.

### DIFF
--- a/src/hpx_wrap.cpp
+++ b/src/hpx_wrap.cpp
@@ -61,16 +61,17 @@ namespace hpx_start
     // main entry point of the HPX runtime system
     int hpx_entry(int argc, char* argv[])
     {
-        #if defined(__linux) || defined(__linux__) || defined(linux)
+#if defined(__linux) || defined(__linux__) || defined(linux)
         // Call to the main() function
         int return_value = __real_main(argc, argv);
-        #else /* APPLE */
+#else /* APPLE */
         // call to the main() function
         int return_value = main(argc, argv);
-        #endif
+#endif
 
-        //Finalizing the HPX runtime
-        return hpx::finalize(return_value);
+        // Finalizing the HPX runtime
+        hpx::finalize();
+        return return_value;
     }
 }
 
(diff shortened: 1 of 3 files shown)


hkaiser:
@NK-Nikunj please verify

src/hpx_wrap.cpp
```diff
@@ -61,16 +61,17 @@ namespace hpx_start
...
         // call to the main() function
         int return_value = main(argc, argv);
-        #endif
+#endif
 
-        //Finalizing the HPX runtime
-        return hpx::finalize(return_value);
+        // Finalizing the HPX runtime
+        hpx::finalize();
+        return return_value;s
```
NK-Nikunj: @hkaiser there's an extra 's'

NK-Nikunj:
@hkaiser Just ran the builds and checked. This seems to resolve the issue. You were right, `hpx::finalize` does not propagate the return value (I cross-validated that as well).

src/hpx_wrap.cpp
hkaiser: That's fixed now, thanks!

src/hpx_wrap.cpp
```diff
@@ -61,16 +61,17 @@ namespace hpx_start
...
         // call to the main() function
         int return_value = main(argc, argv);
-        #endif
+#endif
 
-        //Finalizing the HPX runtime
-        return hpx::finalize(return_value);
+        // Finalizing the HPX runtime
+        hpx::finalize();
+        return return_value;
```
biddisco: OK. I thought that finalize could return it, but if this works properly

hkaiser: Yah, I thought that as well … :/
But it works this way, we verified.

hkaiser:
The fact that it works is confirmed by a couple of tests now failing. Those were failing before but were not reported as such.

src/hpx_wrap.cpp
biddisco: would the "proper" fix to be to make sure that `return hpx::finalize(return_value);` did what we expect it to?

hkaiser: I think such an overload for `hpx::finalize()` would be possible but dangerous. We already have an overload taking a `double`.

PR #3479, opened 2018-10-05
Fix FLAGS - Use correct version of target_compile_options

biddisco:
The previous PR #3458 somehow got pushed by me with the wrong flag setting CMake call.
This one uses `target_compile_options` for the main hpx target and adds the PUBLIC setting so that all targets that depend on hpx also pick up the same flags.

@khuck noted on IRC that warnings had increased and he was correct.


biddisco:
This should be merged as soon as pycicle/circleci show that it hasn't made anything worse.

PR #3480, opened 2018-10-08
Towards a more lightweigh transfer action

sithhell:
This patch removes the extra `transfer_continuation_action` and the `transfer_base_action`
among other things that haven't been used to reduce binary bloat.

Flyby: Fixing header tests


sithhell:
Doesn't work out as intended. The compile time overhead seems to be indeed even higher.

PR #3483, opened 2018-10-08
Fixing type attribute warning for transfer_action

sithhell:
This fixes warnings occuring after recent changes to cmake.

### DIFF
--- a/hpx/runtime/actions/basic_action.hpp
+++ b/hpx/runtime/actions/basic_action.hpp
@@ -861,6 +861,7 @@ namespace hpx { namespace serialization
 #define HPX_REGISTER_ACTION_1(action)                                         \
     HPX_REGISTER_ACTION_2(action, action)                                     \
 /**/
+#if defined(HPX_MSVC) || defined(HPX_MINGW)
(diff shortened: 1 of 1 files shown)


hpx/runtime/actions/basic_action.hpp
```diff
@@ -861,6 +861,7 @@ namespace hpx { namespace serialization
 #define HPX_REGISTER_ACTION_1(action)                                         \
     HPX_REGISTER_ACTION_2(action, action)                                     \
 /**/
+#if defined(HPX_MSVC) || defined(HPX_MINGW)
```
msimberg: Why does this get rid of any warnings? It looks to me like the if and else branches are exactly the same...

sithhell: Good catch ... looks like I didn't commit the latest version...

msimberg:
@sithhell this caused undefined references with GCC. Were the warnings only a problem with clang?

sithhell:
No. The warnings were a problem with gcc, not with clang. Very strange... I
am using gcc here as well. I'll have a look.

PR #3486, opened 2018-10-09
Fix comparison for expect_connecting_localities config entry

### LINKED ISSUE #3453: Two HPX applications can't run at the same time.
## Expected Behavior

If one HPX application is running, another HPX application should be able to run.

## Actual Behavior

The second application will fail with this exception:
```
  what():  <unknown>: HPX(network_error)
```
This is particularly problematic on the Phylanx buildbot server, different configurations can't be built and tested concurrently because one of them will fail.

Run two HPX applications on the same host concurrently.

## Specifications

  - Platform (compiler, OS):  GCC 7.1 on Red Hat Linux

msimberg:
I think this fixes #3453, although I'm not quite sure if there are side effects with multiple localities. This at least works for me with a single locality.

The `expect_connections` variable was set opposite to what the comment above it says, and the comment sounds like the correct thing to do, so I've changed it to expect connections only if the number of localities is greater than one (or if it's set explicitly of course).

@khuck could you check if this works for phylanx?

### DIFF
--- a/src/runtime/components/server/runtime_support_server.cpp
+++ b/src/runtime/components/server/runtime_support_server.cpp
@@ -272,7 +272,7 @@ namespace hpx { namespace components { namespace server
 {
     ///////////////////////////////////////////////////////////////////////////
     runtime_support::runtime_support(hpx::util::runtime_configuration & cfg)
-      : stop_called_(false), stop_done_(false), terminated_(false),
+      : stop_called_(true), stop_done_(true), terminated_(false),
         dijkstra_color_(false), shutdown_all_invoked_(false),
         modules_(cfg.modules())
     {}
--- a/src/util/command_line_handling.cpp
+++ b/src/util/command_line_handling.cpp
@@ -648,7 +648,8 @@ namespace hpx { namespace util
         ini_config += std::string("hpx.expect_connecting_localities=") +
             (expect_connections ? "1" : "0");
 
-        if (num_localities_ != 1 || expect_connections)
+        if (num_localities_ != 1 || expect_connections ||
+            env.found_batch_environment())
         {
             initial_hpx_port =
                 boost::lexical_cast<std::uint16_t>(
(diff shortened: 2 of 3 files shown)


hkaiser:
> The expect_connections variable was set opposite to what the comment above it says, and the comment sounds like the correct thing to do, so I've changed it to expect connections only if the number of localities is greater than one (or if it's set explicitly of course).

@msimberg why shouldn't we expect connections if the application runs on one locality only?

msimberg:
Please correct me if I'm wrong but I thought `expect_connecting_localities` is exactly for localities connecting dynamically after initialization. If I've set one locality I would think that's in most cases because I really only want to use one locality and don't expect any others to connect, but if that's not what I want I can still set `expect_connecting_localities=1`.

That said, this does seem to cause problems on some tests so it may not be the way to go.

hkaiser:
@msimberg yes, makes sense.

msimberg:
Note: this passes all tests now. I had to change the following:

- always enable parcel ports when in batch environments
- don't notify waiting main if errors happen before main is waiting for finalize

The fixes are not the prettiest, but seem to work.

msimberg:
Closing since #3599 was merged.

src/runtime/components/server/runtime_support_server.cpp
```diff
@@ -272,7 +272,7 @@ namespace hpx { namespace components { namespace server
 {
     ///////////////////////////////////////////////////////////////////////////
     runtime_support::runtime_support(hpx::util::runtime_configuration & cfg)
-      : stop_called_(false), stop_done_(false), terminated_(false),
+      : stop_called_(true), stop_done_(true), terminated_(false),
         dijkstra_color_(false), shutdown_all_invoked_(false),
```
hkaiser: @msimberg: Is that change still necessary?

msimberg: Good catch. In principle yes, since the change is independent from the networking changes. It just happened to be triggered once these changes were introduced. I'll check and open a new PR with that change only if needed.

PR #3485, opened 2018-10-09
Use BUILD_INTERFACE generator expression to fix cmake flag exports

biddisco:
Fixes #3484 

## Proposed Changes
Use BUILD_INTERFACE generator expression to make flags only inherited by targets n the build tree

PR #3488, opened 2018-10-12
Revert "Use BUILD_INTERFACE generator expression to fix cmake flag exports"

hkaiser:
Reverts STEllAR-GROUP/hpx#3485

I'm going to revert this change as the original PR breaks the builds on Windows. I'm seeing:
```
2>Building Custom Rule D:/Devel/hpx/src/CMakeLists.txt
2>CMake does not need to re-run because D:/Devel/hpx/build/All.vcpkg/src/CMakeFiles/generate.stamp is up-to-date.
2>PUBLIC
2>PUBLIC
2>error_code.cpp
2>c1xx : fatal error C1083: Cannot open source file: 'PUBLIC': No such file or directory
2>exception.cpp
2>c1xx : fatal error C1083: Cannot open source file: 'PUBLIC': No such file or directory
2>exception_list.cpp
...
```


hkaiser:
The problem was solved by: https://github.com/STEllAR-GROUP/hpx/commit/458e0b647cbdff2c57fa6bc1f14f971361108bfa

PR #3489, opened 2018-10-12
Remove deprecated options for 1.2.0

msimberg:
This removes the CMake options and code for:

- `HPX_WITH_LOCAL_DATAFLOW_COMPATIBILITY`
- `HPX_WITH_GENERIC_EXECUTION_POLICY`

because they have been deprecated since 0.9.99.

In principle `HPX_WITH_ASYNC_FUNCTION_COMPATIBILITY` should also be removed according to the release procedures, but with more frequent releases and semver considerations I'm tempted to leave it. Any opinions?


hkaiser:
> In principle HPX_WITH_ASYNC_FUNCTION_COMPATIBILITY should also be removed according to the release procedures, but with more frequent releases and semver considerations I'm tempted to leave it. Any opinions?

I'd say let's get rid of it.

msimberg:
> > In principle HPX_WITH_ASYNC_FUNCTION_COMPATIBILITY should also be removed according to the release procedures, but with more frequent releases and semver considerations I'm tempted to leave it. Any opinions?
> 
> I'd say let's get rid of it.

All right, I'll remove it. I'll add it still to this PR, so please don't merge until then.

msimberg:
For some reason I missed a whole bunch of other options that also need to be removed. I'll remove them asap.

PR #3491, opened 2018-10-15
Changing Base docker image

sithhell:
This new base docker image providing better support for C++17


hkaiser:
@sithhell would you care to see this through?

sithhell:
@hkaiser done. All but the parallel algorithms (same old problem ...) works.

PR #3493, opened 2018-10-15
Remove deprecated options for 1.2.0 part 2

msimberg:
Removes:

* Boost chrono compatibility
* v1 executor compatibility
* execution policy compatibility
* transform reduce compatibility (`inner_product`)


hkaiser:
Another ~9k source lines down!

PR #3497, opened 2018-10-18
Note that cuda support requires cmake 3.9

biddisco:


hkaiser:
@biddisco is this enforced in the CMake files?

biddisco:
Updated PR with quick check in CMake to raise minimal version if HPX_WITH_CUDA is enabled

biddisco:
Updated to do it using your suggestion. I was being lazy and I'm glad you didn't let me get away with it.

CMakeLists.txt
```diff
@@ -15,6 +15,10 @@
 # We require at least CMake V3.3.2
 cmake_minimum_required(VERSION 3.3.2 FATAL_ERROR)
 
+if (HPX_WITH_CUDA AND CMAKE_VERSION VERSION_LESS 3.9)
```
hkaiser: Isn't `HPX_WITH_CUDA` defined only further down in this file?

biddisco: yes, but the first time someone runs cmake, they will either use `cmake -DHPX_WITH_CUDA` on the command line, or they will run it once with `ccmake` then change the gui setting to enable `HPX_WITH_CUDA` and then when they configure, they'll get the error.

PR #3498, opened 2018-10-19
Remove thread specific ptr

sithhell:
`hpx::util::thread_specific_ptr` introduces an extra level of indirection and dynamic memory allocation where none should be needed. `thread_local` could be used instead, this patch is using `HPX_NATIVE_TLS` to allow for some backwards compatibility.

PR #3499, opened 2018-10-19
Improving spinlock pools

sithhell:
This patch introduces a multiplicative hashing algorithm for the spinlock pools to avoid modulo operations in the critical paths.
This, together with #3498 shows up to 10% performance increase on the Phylanx simple loop benchmark: https://docs.google.com/spreadsheets/d/15aNdbbyrWg3L56kY5Ud56sflX3Iy3tqbOC5dBaDijsc/edit#gid=0

### DIFF
--- a/hpx/lcos/local/spinlock.hpp
+++ b/hpx/lcos/local/spinlock.hpp
@@ -15,7 +15,6 @@
 #include <hpx/config.hpp>
 
 #include <hpx/runtime/threads/thread_helpers.hpp>
-#include <hpx/runtime/threads/topology.hpp>
 #include <hpx/util/detail/yield_k.hpp>
 #include <hpx/util/itt_notify.hpp>
 #include <hpx/util/register_locks.hpp>
--- a/hpx/lcos/local/spinlock.hpp
+++ b/hpx/lcos/local/spinlock.hpp
@@ -15,6 +15,7 @@
 #include <hpx/config.hpp>
 
 #include <hpx/runtime/threads/thread_helpers.hpp>
+#include <hpx/runtime/threads/topology.hpp>
 #include <hpx/util/detail/yield_k.hpp>
 #include <hpx/util/itt_notify.hpp>
 #include <hpx/util/register_locks.hpp>
@@ -127,7 +128,6 @@ namespace hpx { namespace lcos { namespace local
 #endif
         }
     };
-
 }}}
 
 #endif // HPX_B3A83B49_92E0_4150_A551_488F9F5E1113
--- a/hpx/lcos/local/spinlock_pool.hpp
+++ b/hpx/lcos/local/spinlock_pool.hpp
@@ -14,6 +14,7 @@
 #define HPX_LCOS_LOCAL_SPINLOCK_POOL_HPP
 
 #include <hpx/config.hpp>
+#include <hpx/util/fibhash.hpp>
 #include <hpx/lcos/local/spinlock.hpp>
 
 #include <cstddef>
@@ -24,7 +25,7 @@ namespace hpx { namespace lcos { namespace local
 #if HPX_HAVE_ITTNOTIFY != 0
     namespace detail
     {
-        template <typename Tag>
+        template <typename Tag, std::size_t N>
         struct itt_spinlock_init
         {
             itt_spinlock_init();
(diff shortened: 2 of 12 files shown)


msimberg:
Does the size of the spinlock pool make much of a difference?

hkaiser:
> Does the size of the spinlock pool make much of a difference?

Not sure. Using a prime number seems to be a good idea, though. Also, I'd suggest to pad the locks such as each is loaded into a separate cache line. That should help decreasing false sharing.

hpx/util/spinlock_pool.hpp
```diff
@@ -41,20 +42,20 @@ namespace hpx { namespace util
     }
 #endif
 
-    template <typename Tag>
+    template <typename Tag, std::size_t N = 128>
     class spinlock_pool
```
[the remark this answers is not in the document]
sithhell: In theory, yes. In this implementation, no: The reason is the shift operation instead of division. We could manually clamp the value, but that would add another operation. I think requiring a power of two isn't that bad.

sithhell:
> > Does the size of the spinlock pool make much of a difference?
> 
> Not sure. Using a prime number seems to be a good idea, though. Also, I'd suggest to pad the locks such that each is loaded into a separate cache line. That should help decreasing false sharing.

No, using a prime number instead won't change anything. In fact, it would eliminate the purpose of having a fast implementation of the hash. It is all explained nicely here: https://probablydance.com/2018/06/16/fibonacci-hashing-the-optimization-that-the-world-forgot-or-a-better-alternative-to-integer-modulo/ (also part of the fibhash header).

Increasing the size makes a difference for sure: The bigger the array, the less conflicts we have. I haven't played around with it though.

Padding the locks is definitely something we can try.

hkaiser:
@sithhell, @msimberg ping?

msimberg:
This would probably be good to go as is, but would it be worth it to make the default 128 (CMake) configurable? Or at least try a few other values for `N` before we merge this?

sithhell:
I updated the PR to make the size configurable and pad to cache line boundaries.

The number not being a prime isn't a problem at all, since the fibhash itself takes care of the proper distribution. I tested with different sizes but couldn't observe significant performance difference (unless the number is too small) with future_overhead and phylanx simple_loop_test.

hpx/lcos/local/spinlock_pool.hpp
```diff
@@ -14,39 +14,46 @@
...
+        template <typename Tag, std::size_t N>
         struct itt_spinlock_init
         {
             itt_spinlock_init();
             ~itt_spinlock_init();
         };
-    }
 #endif
+        // special struct to ensure cache line alignemnt
+        struct spinlock_holder
```
msimberg: Couldn't `spinlock` have `alignas(64)` instead?

sithhell: Yes, I wasn't sure if that was working on all our platforms... I remember that some had problems with over alignment

hpx/lcos/local/spinlock_pool.hpp
```diff
@@ -14,39 +14,46 @@
...
+        {
+            // pad to 64 bytes
+            char cacheline_pad[64 - sizeof(lcos::local::spinlock)];
```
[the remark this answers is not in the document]
biddisco: I would suggest we add an `hpx::topology::get_cache_line_size` and one for `page_size` as well. then there is a single place where all code can get the value from and it can be adjusted by platform/architecture in a less invasive manner than having 64 and 4096 in code scattered about.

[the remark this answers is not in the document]
hkaiser: Good thinking!

hkaiser:
@sithhell please verify my changes I applied in an attempt to address the review comments

hpx/lcos/local/spinlock.hpp
```diff
@@ -15,6 +15,7 @@
 #include <hpx/config.hpp>
 
 #include <hpx/runtime/threads/thread_helpers.hpp>
+#include <hpx/runtime/threads/topology.hpp>
```
sithhell: why did you add this include?

hkaiser: Nice catch - it's a leftover. Will remove.

hkaiser: Done.

PR #3505, opened 2018-10-25
Fixing type attribute warning for transfer_action

msimberg:
Retry of #3483. I didn't change anything from @sithhell's original, so I'm expecting pycicle gcc builds to fail.

### DIFF
--- a/hpx/components/containers/partitioned_vector/export_definitions.hpp
+++ b/hpx/components/containers/partitioned_vector/export_definitions.hpp
@@ -9,7 +9,11 @@
 #include <hpx/config/export_definitions.hpp>
 
 #if defined(HPX_PARTITIONED_VECTOR_MODULE_EXPORTS)
-# define HPX_PARTITIONED_VECTOR_EXPORT HPX_SYMBOL_EXPORT
+# if defined(HPX_MSVC) || defined(HPX_MINGW)
+#  define HPX_PARTITIONED_VECTOR_EXPORT HPX_SYMBOL_EXPORT
+# else
+#  define HPX_PARTITIONED_VECTOR_EXPORT
+# endif
 #else
 # define HPX_PARTITIONED_VECTOR_EXPORT HPX_SYMBOL_IMPORT
 #endif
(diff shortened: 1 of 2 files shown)


msimberg:
Apparently the builds on daint are ok with gcc... The errors on daint were [these](http://rostam.cct.lsu.edu/builders/hpx_gcc_7_boost_1_65_centos_x86_64_release/builds/130/steps/build_unit_tests/logs/stdio), so something is different on rostam.

hkaiser:
> Apparently the builds on daint are ok with gcc... The errors on daint were these, so something is different on rostam.

Could it be different `--fvisibility=…` settings (or whatever the command line options are)?

msimberg:
> > Apparently the builds on daint are ok with gcc... The errors on daint were these, so something is different on rostam.
> 
> Could it be different `--fvisibility=…` settings (or whatever the command line options are)?

I had hoped that would be the problem, but that doesn't seem to be the case. We can maybe have a look next week.

msimberg:
I'm not sure what is going on. Manually triggered builds with the same configuration that failed before don't fail now, and nothing relevant has changed. I think I'll just have to try merging again, and retry if it fails.

src/components/containers/partitioned_vector/partitioned_vector_component_double.cpp
```diff
@@ -21,7 +21,10 @@ HPX_REGISTER_PARTITIONED_VECTOR(double);
 #pragma warning(disable: 5037)
 #endif
 
-template class HPX_PARTITIONED_VECTOR_EXPORT
+template class
+#if defined(HPX_MSVC) || defined(HPX_MINGW)
+HPX_PARTITIONED_VECTOR_EXPORT
+#endif
```
hkaiser: Could we handle that by defining a separate pp constant (similarly to `HPX_PARTITIONED_VECTOR_SPECIALIZATION_EXPORT`, see here: https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/components/containers/partitioned_vector/export_definitions.hpp#L17-L21)?

msimberg: Ooh, yes, absolutely. Thanks for pointing that out.

hkaiser:
@sithhell, @msimberg what's the status here. Can we go ahead and merge this now?

PR #3506, opened 2018-10-28
WIP Improve CMake Testing  

biddisco:
Currently, the CMake driven tests are run on machines like daint at CSCS by calling srun from the hpx-runwrapper.py script. Unfortunately, calling srun 600+ times during testing places a large burden on slurm and causes false positive test fails due to slurm timeouts etc.

This branch removes the runwrapper invocation for tests that only use a single locality, so they are run directly rather than via runwrapper/srun/mpiexec

This PR  is an experimental branch to try to solve issues with pycicle/daint and not intended to be merged yet.

### DIFF
--- a/cmake/HPX_AddTest.cmake
+++ b/cmake/HPX_AddTest.cmake
@@ -29,105 +29,53 @@ macro(add_hpx_test category name)
     set(_exe "${${name}_EXECUTABLE}")
   endif()
 
-  set(expected "0")
-
-  if(${name}_FAILURE_EXPECTED)
-    set(expected "1")
-  endif()
-
   set(args)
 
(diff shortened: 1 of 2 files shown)


msimberg:
@biddisco is this still a thing?

msimberg:
To add an alternative to this, `hpxrun.py` also doesn't handle interrupts correctly. Right now e.g. with `srun` as the run command, it tries to kill the child process but, `srun` would require two `Ctrl-C`s to actually stop. I don't know how robust that can be made but if that were fixed we'd also have the correct behaviour on multi-locality runs.

Even then, `ctest` might keep us from doing this properly: https://gitlab.kitware.com/cmake/cmake/issues/17288 (`ctest` only sends `SIGKILL` which can't be handled).

Issue #3474, opened 2018-10-03
Modernize CMake used in HPX


biddisco:
## Expected Behavior
Modern CMake thoughout

## Actual Behavior
The CMake commands
```
add_compile_options
include_directories
link_directories
link_libraries
```
are deprecated and should be replaced with  
```
target_compile_options
target_include_directories
target_link_libraries
```
This is mostly inspired by things I didn't like when implementing  #3458. we need a bit of a makeover in places.

chinz07:
It might also be a good idea to use compile features (cxx_std_11, ...) instead of explicit flags to enable newer standards. This should reduce the probability for compatibility problems as CMake can choose a suitable version based on the requirements for all targets.

biddisco:
Yes. The whole CMake setup needs a good cleaning up. The macros that appear throughout are awful as are as things like adding `_exe` extensions to targets which is terribly annoying. 
Are you volunteering?

biddisco:
Regarding the CXX 11/14/17 - we do try to support different compilers and so it is nice to force a build to use C++11 or C++14 depending on what the user asks for (or the testing CI stuff), so we probably wouldn't wat to do away with it altogether. The problem is that we support new language features if they are available, and support older stuff f not, so allowing cmake to decide for us will prevent the old stuff being tested. If we can keep the old functionality, but clean things up, then great ...

chinz07:
> Yes. The whole CMake setup needs a good cleaning up. The macros that appear throughout are awful as are as things like adding `_exe` extensions to targets which is terribly annoying.
> Are you volunteering @chinz07 ?

I would love to help with this issue. For me as a consumer, HPX's arcane build system is one of the major pain points. Incidentally, my project has a similar structure (runtime library + several (optional) backends). It might be possible to adapt some of the code patterns. I will review the current build system and compile a preliminary plan of action for the modernization. Afterwards, we can discuss how to include any corner case or additional constraints I might have missed. I think this is the most reasonable course of action.

> Regarding the CXX 11/14/17 - we do try to support different compilers and so it is nice to force a build to use C++11 or C++14 depending on what the user asks for (or the testing CI stuff), so we probably wouldn't wat to do away with it altogether. The problem is that we support new language features if they are available, and support older stuff f not, so allowing cmake to decide for us will prevent the old stuff being tested. If we can keep the old functionality, but clean things up, then great ...

[...]

Issue #3257, opened 2018-03-26
migrate_component test times out occasionally


msimberg:
The `migrate_component` test occasionally times out. From what I can tell it always happens at the `test_migrate_busy_component2` step.

Typical output:

```
test_migrate_component: ->{0000000200000000, 0000000000000000}
test_migrate_component: <-{0000000200000000, 0000000000000000}
test_migrate_busy_component: ->{0000000200000000, 0000000000000000}
test_migrate_busy_component: <-{0000000200000000, 0000000000000000}
test_migrate_component2: ->{0000000200000000, 0000000000000000}
....................................................................................................
test_migrate_component2: <-{0000000200000000, 0000000000000000}
....................................................................................................
test_migrate_busy_component2: ->{0000000200000000, 0000000000000000}
```

biddisco:
just to bump this. I ran a full set of tests on my laptop a few days ago and noticed this was one of the few that failed.

hkaiser:
Darn, I really thought we did fix this once and for all :/

biddisco:
Not a timeout, just a regular fail. I'll look into it when I have a moment
```
test 880
    Start 880: tests.unit.component.migrate_component

880: Test command: /bin/python "/home/biddisco/build/hpx/bin/hpxrun.py" "/home/biddisco/build/hpx/bin/migrate_component_test" "-e" "0" "-t" "2" "-v" "--"
880: Test timeout computed to be: 1500
880: test_migrate_component: ->{0000000100000000, 0000000000000000}
880: test_migrate_component: <-{0000000100000000, 0000000000000000}
880: test_migrate_lazy_component: ->{0000000100000000, 0000000000000000}
880: test_migrate_lazy_component: <-{0000000100000000, 0000000000000000}
880: test_migrate_lazy_component_client: ->{0000000100000000, 0000000000000000}
880: test_migrate_lazy_component_client: <-{0000000100000000, 0000000000000000}
880: test_migrate_busy_component: ->{0000000100000000, 0000000000000000}
880: test_migrate_busy_component: <-{0000000100000000, 0000000000000000}
880: test_migrate_lazy_busy_component: ->{0000000100000000, 0000000000000000}
880: test_migrate_lazy_busy_component: <-{0000000100000000, 0000000000000000}
880: test_migrate_lazy_busy_component_client: ->{0000000100000000, 0000000000000000}
880: test_migrate_lazy_busy_component_client: <-{0000000100000000, 0000000000000000}
[...]

PR #3514, opened 2018-10-31
Cleaning up the build system

chinz07:
## Proposed Changes

- Remove some dead or obsolete code (hpx_is_target and SubProject.cmake)
- Replace most macros with functions

The remaining macros can not be replaced easily without an extensive rewrite. Since the PR is already rather large, I will submit these changes as a separate PR.

To verify that the behavior of the build system has not changed, I have compared the generated compiler invocations and the hashes of every object file with their old values. The modified build system produces bit-for-bit identical results for all files (except version.cpp.o) if the default settings are used.

This PR is the first step towards resolving issue #3474. As part of a followup PR I will start to replace the deprecated directory-based commands (add_compile_options, ...).


chinz07:
I am unable to reproduce the failures on CircleCI. Even if I use the `stellargroup/build_env:debian_clang` Docker image and work through the steps specified in the config file, all tests succeed on my machine. Judging from other runs of the test suite on CircleCI these tests do timeout occasionally. Are these failures expected?

The failures on daint are related to #3506?

Any help would be appreciated.

msimberg:
Hi @chinz07, you're right that those failures are unrelated. The CircleCI failure is this one: #3257, and the daint one is related to but not caused by #3506. Sorry about that. This PR is looking really good and much appreciated! For now you're mostly safe to assume that errors that are not build-related are not caused by your PR.

msimberg:
> > and the daint one is related to but not caused by #3506.
> 
> Could you please elaborate on that. What's a related fail, I had a wuick look on cdash but couldn't spot anything obvious

@biddisco I might've looked too quickly. It looks like a plain old hang in HPX whereas your PR is about slurm timeouts. So not related.

chinz07:
@msimberg Thanks for the quick reply. Since the produced binaries were bit-for-bit identical, I anticipated that much. But better safe than sorry. ODR violations due to missing flags and preprocessor definitions can cause quite a pain.

If the failures are unrelated, the PR is ready to be merged.

chinz07:
@msimberg Agreed. That's what I intended anyway. Sorry for not mentioning it explicitly. In the meantime, I will continue to prepare the follow-up PRs. Hopefully, we can merge those as fast as possible after the release so everybody can properly test the refactored build system.

PR #3521, opened 2018-11-02
Fixing argument handling in induction and reduction clauses for parallel::for_loop

hkaiser:


hkaiser:
@msimberg I believe the remaining errors are unrelated...

msimberg:
@hkaiser yes, the install failure is unrelated. If you want you can rebase on latest master and it should go away, but it's okay like this as well. I'll wait for the pycicle tests to pass but looks good otherwise. Thanks for fixing the build (and for fixing the problem in the first place).

PR #3523, opened 2018-11-03
Fix logging regression

K-ballo:
Fix logging regression after 2972f39e1944a564aa1013f7786d55f4f6281740


hkaiser:
@msimberg this should go into the release, I think.

msimberg:
@hkaiser yep, definitely want this one. Thanks @K-ballo!

PR #3524, opened 2018-11-05
Do not build guided pool examples with older GCC compilers

biddisco:
Buildbot confirms that this now passes on the gcc 4_9 builder. I do not know how to fix the problem correctly, so the safest thing is to simply disable the tests on gcc4.9 and lower

### DIFF
--- a/examples/resource_partitioner/CMakeLists.txt
+++ b/examples/resource_partitioner/CMakeLists.txt
@@ -9,11 +9,16 @@ set(example_programs
   simplest_resource_partitioner_1
   simplest_resource_partitioner_2
 )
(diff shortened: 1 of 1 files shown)


biddisco:
FYI http://rostam.cct.lsu.edu/builders/hpx_gcc_4_9_boost_1_60_centos_x86_64_release/builds/148
is a buildbot output where the compilation fail can be seen to be suppressed.

biddisco:
I'm not proud of myself, but I wasted too much time trying to fix it for an old compiler and at some point you have to give up and move on - and no - nobody else will ever use this feature on gcc 4.9, so we don't need to worry about it.
We can remove gcc 4.x support after this release yes?

PR #3531, opened 2018-11-06
Warnings, warnings...

msimberg:
I'll try to make this the last one so that I can enable `-Werror`.


hkaiser:
Note, this includes fixes from #3505

msimberg:
> Note, this includes fixes from #3505

Yes, you're right. I'll merge them separately but based this off that branch so that I can see when I get to zero warnings on pycicle and rostam.

hkaiser:
What can we do to make sure those don't creep back in?

msimberg:
> What can we do to make sure those don't creep back in?

`-Werror`? I plan to turn it on for pycicle builds. I know there will be times when that's annoying but long term I think it's the only way to keep these from coming back. There will likely be differences between what we build on daint with pycicle and what gets built on rostam with buildbot (already happened), but the majority of warnings on the compilers we care about most will be caught before anything is merged.

msimberg:
I added a new CMake option `HPX_WITH_COMPILER_WARNINGS_AS_ERRORS` for turning `-Werror` on. The latest clang and gcc are now clean with the configurations on daint and rostam. There are a couple of warnings still with our oldest supported compilers that I take as false positives that have been fixed in newer compilers.

CMakeLists.txt
```diff
@@ -1386,6 +1390,15 @@ if(HPX_WITH_COMPILER_WARNINGS)
...
 
+# Configure compiler warnings as errors
+if(HPX_WITH_COMPILER_WARNINGS_AS_ERRORS)
+  if(MSVC)
+    hpx_add_compile_flag(-WX LANGUAGES C CXX Fortran)
+  else()
+    hpx_add_compile_flag_if_available(-Werror LANGUAGES CXX C Fortran)
+  endif()
+endif()
+
```
hkaiser: This may break builds on MSVC as `-WX` will trigger tons of errors on superficial warnings generated in system headers (Boost). IIRC, `-WX` implies `-W4` which is the actual problem. But I will check again.

hpx/hpx_fwd.hpp
```diff
@@ -1,98 +0,0 @@
...
-#include <hpx/runtime/get_worker_thread_num.hpp>
-#include <hpx/runtime/message_handler_fwd.hpp>
-#include <hpx/runtime/set_parcel_write_handler.hpp>
-#include <hpx/runtime/trigger_lco.hpp>
-
-#include <hpx/lcos/async_callback_fwd.hpp>
-#include <hpx/lcos/async_fwd.hpp>
-
-#endif
-
```
hkaiser: Yay! Finally!

CMakeLists.txt
msimberg: Okay, thanks for checking. I'm happy to not enable this on Windows as well if it's not useful there. It is off by default though.

sithhell:
Can we move on with this?

msimberg:
@sithhell yeah, thanks for bumping this. I wanted rostam to be in better shape before merging. Specifically this: https://github.com/STEllAR-GROUP/hermione-buildbot/pull/3. Does that look like a reasonable thing to do?

PR #3537, opened 2018-11-08
Fix guided exec

biddisco:
These patches appear to fix a couple of problems that sneaked through the guided executor merge.

Committing this to see if all tests pass with the slight tweak to the executor interface.

### DIFF
--- a/hpx/parallel/executors/thread_execution.hpp
+++ b/hpx/parallel/executors/thread_execution.hpp
@@ -129,7 +129,7 @@ namespace hpx { namespace threads
         std::is_same<typename hpx::util::decay<Hint>::type,
             hpx::threads::thread_schedule_hint>::value
     >::type
-    post(Executor && exec, F && f, Ts &&... ts, Hint && hint)
+    post(Executor && exec, F && f, Hint && hint, Ts &&... ts)
     {
         exec.add(
             util::deferred_call(std::forward<F>(f), std::forward<Ts>(ts)...),
--- a/hpx/parallel/executors/thread_execution.hpp
+++ b/hpx/parallel/executors/thread_execution.hpp
@@ -126,7 +126,8 @@ namespace hpx { namespace threads
     HPX_FORCEINLINE
     typename std::enable_if<
         hpx::traits::is_threads_executor<Executor>::value &&
-        std::is_same<Hint, hpx::threads::thread_schedule_hint>::value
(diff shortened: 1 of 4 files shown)


hpx/parallel/executors/thread_execution.hpp
```diff
@@ -126,7 +126,8 @@ namespace hpx { namespace threads
     HPX_FORCEINLINE
     typename std::enable_if<
         hpx::traits::is_threads_executor<Executor>::value &&
-        std::is_same<Hint, hpx::threads::thread_schedule_hint>::value
+        std::is_same<typename hpx::util::decay<Hint>::type,
+            hpx::threads::thread_schedule_hint>::value
     >::type
     post(Executor && exec, F && f, Ts &&... ts, Hint && hint)
```
hkaiser: How can there be an argument after a variadic one? I always assumed that the variadic list of arguments should be the last one.

hpx/runtime/threads/executors/guided_pool_executor.hpp
```diff
@@ -248,8 +252,8 @@ namespace hpx { namespace threads { namespace executors
                       << util::debug::print_type<result_type>() << "\n"
                       << "async_execute : Numa Hint   : "
                       << util::debug::print_type<pool_numa_hint<Tag>>() << "\n"
-                      << "async_execute : Hint   : "
-                      << util::debug::print_type<H>() << "\n";
+/*                      << "async_execute : Hint   : "
+                      << util::debug::print_type<H>() << "\n"*/;
```
hkaiser: Please remove code that is not necessary.

hpx/parallel/executors/thread_execution.hpp
biddisco: Sorry. Turns out I don't know what I'm doing after all.
Good job those variadic args are not actually used here. I've switched the order round now.
(gcc didn't mind about the variadic in the middle it seems)

hkaiser: @biddisco may I suggest that you create a (completely separate) customization point for your executor that is enable_if'ed based on the guided_executor type? I think adding the additional condition on the scheduling hint may cause ambiguities.

biddisco: That's a good idea. Didn't think of that. I'll have a look later.

PR #3541, opened 2018-11-09
Fixing the build system of the MPI parcelport

### LINKED ISSUE #3475: HPX fails to link if the MPI parcelport is enabled
## Expected Behavior

HPX should link successfully.

## Actual Behavior

Linking HPX fails with the following linker errors:

```
/usr/bin/ld: CMakeFiles/hpx.dir/version.cpp.o: in function `MPI::Op::Init(void (*)(void const*, void*, int, MPI::Datatype const&), bool)':
version.cpp:(.text._ZN3MPI2Op4InitEPFvPKvPviRKNS_8DatatypeEEb[_ZN3MPI2Op4InitEPFvPKvPviRKNS_8DatatypeEEb]+0x19): undefined reference to `ompi_mpi_cxx_op_intercept'
/usr/bin/ld: CMakeFiles/hpx.dir/version.cpp.o: in function `MPI::Intracomm::Clone() const':
version.cpp:(.text._ZNK3MPI9Intracomm5CloneEv[_ZNK3MPI9Intracomm5CloneEv]+0x3f): undefined reference to `MPI::Comm::Comm()'
/usr/bin/ld: CMakeFiles/hpx.dir/version.cpp.o: in function `MPI::Graphcomm::Clone() const':
version.cpp:(.text._ZNK3MPI9Graphcomm5CloneEv[_ZNK3MPI9Graphcomm5CloneEv]+0x38): undefined reference to `MPI::Comm::Comm()'
/usr/bin/ld: CMakeFiles/hpx.dir/version.cpp.o: in function `MPI::Cartcomm::Sub(bool const*) const':
version.cpp:(.text._ZNK3MPI8Cartcomm3SubEPKb[_ZNK3MPI8Cartcomm3SubEPKb]+0xa4): undefined reference to `MPI::Comm::Comm()'
/usr/bin/ld: CMakeFiles/hpx.dir/version.cpp.o: in function `MPI::Intracomm::Create_graph(int, int const*, int const*, bool) const':
version.cpp:(.text._ZNK3MPI9Intracomm12Create_graphEiPKiS2_b[_ZNK3MPI9Intracomm12Create_graphEiPKiS2_b]+0x3b): undefined reference to `MPI::Comm::Comm()'
/usr/bin/ld: CMakeFiles/hpx.dir/version.cpp.o: in function `MPI::Cartcomm::Clone() const':
version.cpp:(.text._ZNK3MPI8Cartcomm5CloneEv[_ZNK3MPI8Cartcomm5CloneEv]+0x38): undefined reference to `MPI::Comm::Comm()'
/usr/bin/ld: CMakeFiles/hpx.dir/version.cpp.o:version.cpp:(.text._ZNK3MPI9Intracomm11Create_cartEiPKiPKbb[_ZNK3MPI9Intracomm11Create_cartEiPKiPKbb]+0xae): more undefined references to `MPI::Comm::Comm()' follow
/usr/bin/ld: CMakeFiles/hpx.dir/version.cpp.o:(.data.rel.ro._ZTVN3MPI8DatatypeE[_ZTVN3MPI8DatatypeE]+0x78): undefined reference to `MPI::Datatype::Free()'
/usr/bin/ld: CMakeFiles/hpx.dir/version.cpp.o:(.data.rel.ro._ZTVN3MPI3WinE[_ZTVN3MPI3WinE]+0x48): undefined reference to `MPI::Win::Free()'

```

Compile HPX with CMake >= 3.10 and OpenMPI with HPX_WITH_PARCELPORT_MPI enabled.

## Specifications

  - HPX Version: 1.2.0 (commit: e7d1ae4c5ba91b841fd2fa7fed99b712d0270cfe)
  - Platform (compiler, OS): Arch Linux (GCC 8.2.1, OpenMPI 3.1.2, CMake 3.12.2)

I have a, rather ugly, workaround for the linker errors. If the issue should be fixed immediately I could submit a PR containing the workaround. Alternatively, I am currently working on a proper fix. The root cause seems to be a problem with the scope of several CMake variables, exacerbated by the use of macros. It should be possible to get rid of both the macros and most of variables by using modern CMake targets and user-defined target properties. I have made good experiences using this approach in my current project.

chinz07:
Fixes #3475.

## Proposed Changes

  - Emulate add_compile_definitions. It was only introduced as part of CMake 3.11.
  - Link against the C++-enabled MPI libraries

Unfortunately, our previous fix of disabling MPI's C++ interface via `MPI_CXX_SKIP_MPICXX` is broken. Newer versions of CMake introduce it as a cache variable. Therefore, we can not reliable
disable the C++ interface programmatically and have to link against the proper set of libraries.

If the C++ interfaces causes problems, `MPI_CXX_SKIP_MPICXX` can still be enabled manually.

### DIFF
--- a/plugins/parcelport/mpi/CMakeLists.txt
+++ b/plugins/parcelport/mpi/CMakeLists.txt
@@ -10,7 +10,6 @@ include(HPX_AddLibrary)
 # Decide whether to use the MPI based parcelport
 ################################################################################
 if(HPX_WITH_PARCELPORT_MPI)
-  set(MPI_CXX_SKIP_MPICXX TRUE)
   find_package(MPI)
 
   if(NOT MPI_CXX_FOUND)
@@ -58,11 +57,11 @@ if(HPX_WITH_PARCELPORT_MPI)
     endif()
 
     set(_mpi_libraries)
-    if(MPI_C_LIBRARIES)
-      set(_mpi_libraries ${_mpi_libraries} ${MPI_C_LIBRARIES})
-    elseif(MPI_C_LIB_NAMES)
+    if(MPI_CXX_LIBRARIES)
+      set(_mpi_libraries ${_mpi_libraries} ${MPI_CXX_LIBRARIES})
+    elseif(MPI_CXX_LIB_NAMES)
       # cmake V3.10.2 has changed the way MPI libraries are being discovered
-      foreach(lib ${MPI_C_LIB_NAMES})
+      foreach(lib ${MPI_CXX_LIB_NAMES})
         set(_mpi_libraries ${_mpi_libraries} ${MPI_${lib}_LIBRARY})
       endforeach()
     endif()
--- a/plugins/parcelport/mpi/CMakeLists.txt
+++ b/plugins/parcelport/mpi/CMakeLists.txt
@@ -30,7 +30,9 @@ if(HPX_WITH_PARCELPORT_MPI)
   endif()
 
   if(MPI_CXX_COMPILE_DEFINITIONS)
-    add_compile_definitions(${MPI_CXX_COMPILE_DEFINITIONS})
+    # Emulate add_compile_definitions.
+    # This command was only introduced as part of CMake 3.11, but we already require it for versions >= 3.10.
+    set_property(DIRECTORY APPEND PROPERTY COMPILE_DEFINITIONS ${MPI_CXX_COMPILE_DEFINITIONS})
   endif()
 
   # This list is to detect whether we run inside an mpi environment.
--- a/tests/headers/CMakeLists.txt
+++ b/tests/headers/CMakeLists.txt
@@ -35,7 +35,6 @@ endif()
 # add the MPI include path here, because for the main library, it's only
 # added for the plugin.
 if(HPX_WITH_PARCELPORT_MPI)
-  set(MPI_CXX_SKIP_MPICXX TRUE)
   find_package(MPI)
 
   if(MPI_CXX_INCLUDE_PATH)


justwagle:
Works for me, Thanks!!

plugins/parcelport/mpi/CMakeLists.txt
```diff
@@ -56,11 +57,11 @@ if(HPX_WITH_PARCELPORT_MPI)
...
 
     set(_mpi_libraries)
-    if(MPI_C_LIBRARIES)
-      set(_mpi_libraries ${_mpi_libraries} ${MPI_C_LIBRARIES})
-    elseif(MPI_C_LIB_NAMES)
+    if(MPI_CXX_LIBRARIES)
+      set(_mpi_libraries ${_mpi_libraries} ${MPI_CXX_LIBRARIES})
+    elseif(MPI_CXX_LIB_NAMES)
       # cmake V3.10.2 has changed the way MPI libraries are being discovered
-      foreach(lib ${MPI_C_LIB_NAMES})
```
sithhell: Are the C libraries always included in the CXX library list? What if we encounter a MPI library without CXX bindings at all?

chinz07: Yes, the C libraries are always included. CXX is the C++-compatible C interface. The C++ interface, if it is available, is either always included (CMake < 3.10) or needs to be explicitly requested via the MPICXX component (CMake >= 3.10) as far as CMake is concerned.

The documentation for the MPI find module (https://cmake.org/cmake/help/v3.10/module/FindMPI.html) clearly states this from version 3.10 onward. Unfortunately, the documentation for older versions is not as explicit. But judging by the actual code, the CXX variants of all variables and targets always include the information from the C versions.

Depending on the MPI implementation, the C++ interface might still be included even if only the C interface has been requested. Its existence is just not guaranteed.

PR #3542, opened 2018-11-09
Fix numa lookup from pu when using hwloc 2.x

### LINKED ISSUE #3538: numa handling incorrect for hwloc 2
## Expected Behavior
Ability to lookup numa domain from pu/core 

## Actual Behavior
always returns zero as the numa domain when using hwloc 2

Ref : https://www.open-mpi.org/projects/hwloc/doc/v2.0.1/a00327.php

Related to #3181

biddisco:
Fixes #3538 I think


src/runtime/threads/topology.cpp
```diff
@@ -678,7 +707,7 @@ namespace hpx { namespace threads
 
     std::size_t topology::get_number_of_numa_nodes() const
     {
-        int nobjs =  hwloc_get_nbobjs_by_type(topo, HWLOC_OBJ_NODE);
+        int nobjs =  hwloc_get_nbobjs_by_type(topo, HWLOC_OBJ_NUMANODE);
```
hkaiser: Will `HWLOC_OBJ_NUMANODE` be available for older versions of hwloc?

msimberg: I've just removed that commit. The reason is that when `HWLOC_OBJ_NUMANODE` does exist it's an enum value, and thus we can't check `if defined(HWLOC_OBJ_NUMANODE)`. The version check that we had should be sufficient and what is recommended by hwloc for compatibility.

PR #3543, opened 2018-11-13
added Ubuntu dependency list to readme

PixelOfDeath:


msimberg:
Thanks for your PR. This is much appreciated! We just made a new release with some changes to the README on the release branch. Unfortunately that means that this change doesn't apply anymore. We removed the build instructions from the README and have them only in our documentation to avoid having to keep the two in sync.

Would you mind rebasing onto master (I've just merged the release branch into master) and adding a section over [here](https://github.com/STEllAR-GROUP/hpx/blob/master/docs/sphinx/manual/building_hpx.rst#how-to-install-hpx-on-arch-distributions) instead? You could more or less copy the Arch Linux instructions and replace it with the Ubuntu commands.

sithhell:
Since #3632 got merged, we can close this.

PR #3548, opened 2018-11-14
Replace boost::scoped_ptr with std::unique_ptr

K-ballo:


src/runtime/threads/topology.cpp
```diff
@@ -477,7 +477,7 @@ namespace hpx { namespace threads
                 // Strict binding not supported or failed, try weak binding.
                 if (hwloc_set_cpubind(topo, cpuset, HWLOC_CPUBIND_THREAD))
                 {
-                    boost::scoped_ptr<char> buffer(new char [1024]);
+                    std::unique_ptr<char> buffer(new char [1024]);
```
K-ballo: This one should have been `boost::scoped_array<char>`, now `std::unique_ptr<char[]>`.

hkaiser: Could you create a new PR for this, pls?

PR #3562, opened 2018-11-22
Removing the target suffixes

chinz07:
## Proposed Changes

- Remove the '_lib' and '_exe' suffixes from all targets.
- Fix the Bzip2 find module

Having to use different target names depending on the context is confusing and error-prone.
Above all, it makes refactoring the build system a pain.

Since the suffixes is not necessary to disambiguate the existing targets,
this PR simply drops them to unify the target names.

While testing the changes, I also noted that the BZip2 find module is broken.
In addition to the intended changes, the PR also contains a fix for this issue.

This PR is a part of the build system refactoring (see #3474).


chinz07:
Case in point: I missed an occurrence of '_lib'. This should now be fixed.

chinz07:
@hkaiser I have updated the documentation accordingly.

hkaiser:
> I have updated the documentation accordingly.

@chinz07 Thanks! Could you add a sentence or two to the 'What's New' topic for the V1.3 release ('Breaking Changes' section) as well?

chinz07:
@hkaiser Done.

PR #3563, opened 2018-11-22
Drop util::[unique_]function target_type

K-ballo:
The functionality is not particularly useful, and has a high cost in binary bloat.

### DIFF
--- a/hpx/util/detail/basic_function.hpp
+++ b/hpx/util/detail/basic_function.hpp
@@ -23,7 +23,6 @@
 #include <cstring>
 #include <string>
 #include <type_traits>
-#include <typeinfo>
 #include <utility>
 
 namespace hpx { namespace util { namespace detail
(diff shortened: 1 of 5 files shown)


hkaiser:
I think we rely on `unique_function<>::target()` in Phylanx (https://github.com/STEllAR-GROUP/phylanx/blob/master/src/execution_tree/compiler/compiler.cpp#L409). Would you see a way to have it optionally available, at least for `unique_function`?

K-ballo:
Ideally we should be able to clean phylanx's design to not depend on this kind of functionality, but if not we could at least replace it with a low level unsafe unchecked function returning a void pointer to the stored callable.

hkaiser:
Is the bloat caused by the `target_type()` function or by `target()`? I think that `target()` itself does not depend on `type_info` so it shouldn't add much to our vtables… Phylanx relies on `target()` only, so leaving that while removing just `target_type()` would be an option as well.

K-ballo:
The bloat is caused by `target_type()`. And we can implement `target()` without it, despite its semantics being defined in terms of `target_type()`. I'll measure and report back.

hkaiser:
@K-ballo FWIW, `target()` is already implemented without relying on `type_info`.

K-ballo:
`::target()` now restored.

PR #3582, opened 2018-12-07
Towards Modularized HPX

sithhell:
This commit adds the basic infrastructure to develop modular HPX
libraries. That is, that each library in the libs subdirectory is
supposed to be a self contained CMake project, with clear dependencies

In order to streamline our development process, I think it is paramount, that we improve the structure of our code base with clearer dependencies. This is a proposal on how this might look like.


msimberg:
Thanks @sithhell (I got it right this time!). Does this kind of setup mean that as a user I'd have to do `find_package(HPX.libA)` and then `find_package(HPX.libB)`? Instead of `find_package(HPX)` and then use as link targets `HPX::libA` and `HPX::libB`? On the other hand, I guess the top `CMakeLists.txt` can re-export targets from subprojects so that you get both?

sithhell:
Yes, both should be possible at the end.

sithhell:
The documentation generated will look like this:
https://stellar-group.gitlab.io/-/hpx/-/jobs/147110145/artifacts/build/share/hpx/docs/html/libs/_example/docs/index.html

Each added library, will get an entry below the "Modularized HPX Libraries" entry.

sithhell:
I updated the commit on this PR to reflect the various comments received on #3629 and #3628.

The documentation process has been simplified and automatic generation of API reference documentation has been added. Together with that, the self-containment of headers is now automatically tested/generated as well.

The changes are isolated in this PR and can be seen in action in #3629.

PR #3584, opened 2018-12-07
Untangle function storage from object pointer

K-ballo:


hpx/util/detail/basic_function.hpp
```diff
@@ -27,6 +27,8 @@
 
 namespace hpx { namespace util { namespace detail
 {
+    static const std::size_t function_storage_size = 3*sizeof(void*);
```
sithhell: Why is this not constexpr?

hpx/util/detail/basic_function.hpp
```diff
@@ -119,24 +126,29 @@ namespace hpx { namespace util { namespace detail
...
         void swap(function_base& f) noexcept
         {
             std::swap(vptr, f.vptr);
-            std::swap(object, f.object); // swap
+            std::swap(object, f.object);
+            std::swap(storage, f.storage);
+            if (object == &f.storage)
+                object = &storage;
+            if (f.object == &storage)
+                f.object = &f.storage;
```
sithhell: What's the purpose of these assignments?

hpx/util/detail/basic_function.hpp
K-ballo: It wasn't before and there was no reason to add it now, other than perhaps a cosmetic one.

hpx/util/detail/basic_function.hpp
K-ballo: The object pointer may point to embedded storage, in which case it would be pointing to the wrong embedded storage after the swap. If the object pointer was pointing to the other's embedded storage before, it has to point to its own embedded storage now.

PR #3586, opened 2018-12-08
 Make bind propagate value category

K-ballo:
Make bind propagate value category to the target callable. Make one_shot into a generic wrapper that forces rvalue-qualified invocation.

### DIFF
--- a/hpx/util/bind_front.hpp
+++ b/hpx/util/bind_front.hpp
@@ -24,10 +24,6 @@ namespace hpx { namespace util
 {
     namespace detail
     {
-        template <typename F>
-        class one_shot_wrapper;
-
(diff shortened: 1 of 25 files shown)


hpx/util/bind_front.hpp
```diff
@@ -13,6 +13,7 @@
 #include <hpx/util/decay.hpp>
 #include <hpx/util/detail/pack.hpp>
 #include <hpx/util/invoke.hpp>
+#include <hpx/util/one_shot.hpp>
```
biddisco: Why include this here? It isn't used in the edited version as far as I can see.

K-ballo: Looks like a leftover of the refactoring, same for `bind_back.hpp`

PR #3590, opened 2018-12-13
Fix typo/mistake in thread queue cleanup_terminated

biddisco:
I noticed performance issues due to excessive memory use and believe that this error has crept into the master branch and needs to be fixed.

### DIFF
--- a/hpx/runtime/threads/policies/thread_queue.hpp
+++ b/hpx/runtime/threads/policies/thread_queue.hpp
@@ -509,7 +509,7 @@ namespace hpx { namespace threads { namespace policies
                 while (true)
                 {
                     std::lock_guard<mutex_type> lk(mtx_);
-                    if (cleanup_terminated_locked(false))
+                    if (cleanup_terminated_locked(true))
                     {
                         return true;
                     }


hpx/runtime/threads/policies/thread_queue.hpp
```diff
@@ -509,7 +509,7 @@ namespace hpx { namespace threads { namespace policies
                 while (true)
                 {
                     std::lock_guard<mutex_type> lk(mtx_);
-                    if (cleanup_terminated_locked(false))
+                    if (cleanup_terminated_locked(true))
```
msimberg: The result should be the same with `true` and `false`. The difference is that with `true` you shouldn't actually need the while loop; with `false` you need it to make sure all threads are actually deleted. Is there maybe a bug in `cleanup_terminated_locked`? What's the difference when you change it to `true`?

biddisco: I'm looking at this again. You're right that the two paths are the same, but I must have something different on my branch that made it break. I reverted that patch and my tests are still passing, so I must have fixed the problem elsewhere and made a mistake with this.

msimberg: Looking at this a bit more closely there's at least one potential bug in `cleanup_terminated`: threads are only recycled in `cleanup_terminated_locked` if `delete_all` is `true`. I think the result is still correct to some extent, but considering we've been discussing how to reuse stacks we should first make sure that the current implementation actually does what we want it to do.

biddisco: Yup. When I run without this patch, the memory consumption for the futures_overhead test goes through the roof if work is added to the otherwise empty tasks causing tasks to be added to the queues faster than they are removed by worker threads.

I have just pushed rebased versions of the branches lazy_thread_init and remove_wait_or_add_new as part of my own cleaning process.

msimberg: *Small* typo: I meant to say that threads are only recycled if `delete_all` is `false`. Something is not doing what we want it to do...

hkaiser:
@biddisco what should we do with this? Is this still relevant?

biddisco:
I'm closing this because I have other upcoming PRs that supercede it and I no longer consider this to be valid.

PR #3593, opened 2018-12-18
Add include(GNUInstallDirs) to HPXMacros.cmake

hkaiser:
Fixes problems in dependent projects missing `CMAKE_INSTALL_LIBDIR` et.al.

### DIFF
--- a/cmake/templates/HPXMacros.cmake.in
+++ b/cmake/templates/HPXMacros.cmake.in
@@ -5,6 +5,8 @@
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "@HPX_CMAKE_MODULE_PATH@")
+
+include(GNUInstallDirs)
 include(HPX_Utils)
 
 function(hpx_check_compiler_compatibility)


junghans:
Sorry to be late for the party, would it be better to hard-coded the path in the `.cmake` files, which get installed.

Otherwise a user could set `CMAKE_INSTALL_LIBDIR` to some custom path when building hpx, but then `HPXMacros.cmake` will overwrite that with the default coming from `GNUInstallDirs`.

hkaiser:
@junghans could you elaborate please? I'm not sure I fully understand what you're proposing.

junghans:
@hkaiser, if I e.g. build hpx with `-DCMAKE_INSTALL_LIBDIR=/some/path` the hpx libraries get installed in `/some/path`, however, the `HPXMacros.cmake`, which gets installed alongside, doesn't not remember that `CMAKE_INSTALL_LIBDIR` was `/some/path`, but uses the default from `GNUInstallDirs` instead.

hkaiser:
@junghans ok, thanks for the explanation. Do you see any way to resolve this?

junghans:
@hkaiser we would need to preprocess the files included in `HPX_Utils.cmake` (i.e. HPX_%, with % being  Message
  Option AppendProperty   AddCompileFlag   AddCompileTest  AddLinkFlag  AddComponent  AddLibrary  AddExecutable  AddTest  AddPseudoDependencies  AddPseudoTarget ShortenPseudoTarget  AddLibrarySources  AddLibraryHeaders  AddSourceGroup  ExportTargets HandleComponentDependencies  ForceOutOfTreeBuild  SetCMakePolicy  SetLibName  SetupTarget)
and then replace `${CMAKE_INSTALL_LIBDIR}` etc by `@CMAKE_INSTALL_LIBDIR@`.

junghans:
@sithhell additionally you can just configure the dependent project with the same `CMAKE_INSTALL_*DIR` options as hpx was build.

sithhell:
@junghans So what you are suggesting is that the defaults are to put everything in the same location as HPX? Isn't it a bit saner to have the system default as the default (IIUC, that's what GNUInstallDirs.cmake does)?

junghans:
I am saying if you install hpx in some obscure path `HPX_Marcos.cmake` won't remember as it always uses the paths from `GNUInstallDirs` and not actually the ones specified by `CMAKE_INSTALL_*DIR`. However one can help the dependent package by setting e.g. `CMAKE_INSTALL_LIBDIR`, but you are right this will also lead to dependent package being installed there, which might be not desired.

Background here is that, the openmpi & mpich versions of hpx in Fedora live in `$MPI_INCLUDE`, `$MPI_LIB` and `$MPI_BIN` (see [here](https://fedoraproject.org/wiki/Packaging:MPI) for details), but the `HPX_Marcos.cmake` installed alongside (in `$MPI_LIB`) does reflect these paths, but uses the defaults path from `GNUInstallDirs`.

PR #3599, opened 2018-12-28
Making sure networking is disabled for default one-locality-runs

### LINKED ISSUE #3453: Two HPX applications can't run at the same time.

hkaiser:
Fixes #3453

### DIFF
--- a/hpx/runtime/agas/addressing_service.hpp
+++ b/hpx/runtime/agas/addressing_service.hpp
@@ -126,8 +126,7 @@ public:
     resolved_localities_type resolved_localities_;
 
     addressing_service(
-        parcelset::parcelhandler& ph
-      , util::runtime_configuration const& ini_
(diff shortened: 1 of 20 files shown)


hkaiser:
@biddisco I wouldn't know how to test this... Any ideas?

biddisco:
Not really. I just wondered if there was one because the config logic looked a bit strange and easy to mess up.

Issue #3553, opened 2018-11-16
Need a synchronization method for batches of tasks


biddisco:
It frequently occurs that we have loops of the kind
```
for (int i=0; i<large_number; ++i) {
    hpx::async(task, stuff);
}
wait_here_for_tasks_to_complete();
// access some shared structure that is being modified by the tasks
```
Examples of this can be seen in tests like the future overhead performance test, the network storage tests and various other places in the code.
https://github.com/STEllAR-GROUP/hpx/blob/6df436bf5b260716aad5663c943c2f7f5d76857e/tests/performance/local/future_overhead.cpp

In some ways this is related to fork_join parallelism since we have to do N lots of work and block on the result. In the future_overhead test we have a `wait_each`, `wait_all` and a thread counter implementation. The `wait_all` version is most well suited to the _hpx way of doing things_, but is expensive since we must store futures and wait on them. the thread counter implementation is intrusive since it queries the pool at start and at end to ensure no other tasks are running. Another version in my own local tree uses a sliding semaphore, but this requires the user to decorate their task with a decrement of the counter. In other tests, we have used a simepl atomic counter and this also requires modification of the user's task to update the value.

Ideally, we should be able to add tasks (to a thread pool for example) and supply a `latch` so that we can be signalled when the last one has completed.
```
const in num_tasks = 1000000;
hpx::lco::latch my_latch(num_tasks);
for  (int i=0; i<num_tasks; ++i) {
    hpx::async(my_latch, task, args);
}
my_latch.wait();
```
An alternative that uses the existing executor style API
```
const in num_tasks = 1000000;
hpx::lco::latch my_latch(num_tasks);
for  (int i=0; i<num_tasks; ++i) {
    hpx::async(an_executor.with(my_latch);, task, args);
}
my_latch.wait();
```
When tasks complete, the existing internal plumbing would decrement the latch.

I believe something of this kind would be a very useful mechanism to have in our code. I welcome comments on what the best API might be. It could be used on some of the parallel::algorithms to remove some `wait_all` uses.

biddisco:
This is also connected to the idea of N-ary tasks. In that case, promoting the callable object (as suggested in #3348) to be created earlier and storing things like counters in a task structure. I didn't want to write this in the issue descriptions, but just tag it here as a random thought.

biddisco:
Note that in the examples I pasted in above, it would actually be nice to have this functionality with `hpx::apply` since we don't need to return the futures and wait on them.

[...]

PR #3600, opened 2018-12-31
Parallel executor aggregated

hkaiser:
Working towards #3553, this might also be interesting to use the new executor with regard to #3579 (paging @biddisco).

This PR introduces a new executor: `parallel_executor_aggregated` that does not create a future for each task executed by `bulk_execute`, but only one future representing the whole execution. The synchronization is implemented using a simple counting semaphore.

This also introduces a couple of (mostly minor) optimizations for all algorithms relying on `foreach_partitioner::call()` (e.g. `for_each`) and `partitioner::call_with_index()` (for instance `for_loop`).

This PR also extends the for_each benchmark allowing to compare the results.

### DIFF
--- a/hpx/parallel/algorithms/for_each.hpp
+++ b/hpx/parallel/algorithms/for_each.hpp
@@ -170,6 +170,7 @@ namespace hpx { namespace parallel { inline namespace v1
             }
         };
 
+        ///////////////////////////////////////////////////////////////////////
         template <typename Iter>
         struct for_each_n : public detail::algorithm<for_each_n<Iter>, Iter>
         {
@@ -383,7 +384,22 @@ namespace hpx { namespace parallel { inline namespace v1
 
             template <typename ExPolicy, typename InIter, typename F,
                 typename Proj>
-            static InIter
+            static typename std::enable_if<
+                hpx::traits::is_random_access_iterator<InIter>::value, InIter
+            >::type
+            sequential(ExPolicy && policy, InIter first, InIter last, F && f,
+                Proj && proj)
+            {
+                return util::loop_n<typename std::decay<ExPolicy>::type>(first,
+                    static_cast<std::size_t>(std::distance(first, last)),
+                    invoke_projected<F, Proj>{f, proj});
+            }
+
+            template <typename ExPolicy, typename InIter, typename F,
+                typename Proj>
+            static typename std::enable_if<
+                !hpx::traits::is_random_access_iterator<InIter>::value, InIter
+            >::type
             sequential(ExPolicy && policy, InIter first, InIter last, F && f,
                 Proj && proj)
             {
--- a/hpx/parallel/algorithms/for_each.hpp
+++ b/hpx/parallel/algorithms/for_each.hpp
@@ -81,6 +81,43 @@ namespace hpx { namespace parallel { inline namespace v1
             }
         };
 
+        template <typename F>
+        struct invoke_projected<F, util::projection_identity>
+        {
+            typename hpx::util::decay<F>::type& f_;
+            util::projection_identity& proj_;
+
+            template <typename T>
+            HPX_HOST_DEVICE HPX_FORCEINLINE
+            typename std::enable_if<
+               !hpx::traits::is_value_proxy<T>::value
+            >::type
+            call(T && t)
+            {
+                T && tmp = std::forward<T>(t);
+                hpx::util::invoke(f_, tmp);
+            }
+
+            template <typename T>
+            HPX_HOST_DEVICE HPX_FORCEINLINE
+            typename std::enable_if<
+                hpx::traits::is_value_proxy<T>::value
+            >::type
+            call(T && t)
+            {
+                auto tmp = std::forward<T>(t);
+                hpx::util::invoke_r<void>(f_, tmp);
+            }
+
+            template <typename Iter>
+            HPX_HOST_DEVICE HPX_FORCEINLINE
+            void operator()(Iter curr)
+            {
+                call(*curr);
+            }
+        };
+
+        ///////////////////////////////////////////////////////////////////////
         template <typename ExPolicy, typename F, typename Proj>
         struct for_each_iteration
         {
@@ -203,7 +240,7 @@ namespace hpx { namespace parallel { inline namespace v1
         for_each_(ExPolicy && policy, SegIter first, SegIter last, F && f,
             Proj && proj, std::true_type);
 
-        /// Segmented implementaion using for_each.
+        /// Segmented implementation using for_each.
         template <typename ExPolicy, typename FwdIter, typename Size, typename F,
             typename Proj>
         typename util::detail::algorithm_result<ExPolicy, FwdIter>::type
--- a/hpx/parallel/algorithms/for_each.hpp
+++ b/hpx/parallel/algorithms/for_each.hpp
@@ -84,8 +84,13 @@ namespace hpx { namespace parallel { inline namespace v1
         template <typename F>
         struct invoke_projected<F, util::projection_identity>
         {
+            invoke_projected(typename hpx::util::decay<F>::type& f,
+                    util::projection_identity)
+              : f_(f)
+            {
+            }
+
             typename hpx::util::decay<F>::type& f_;
-            util::projection_identity& proj_;
 
             template <typename T>
             HPX_HOST_DEVICE HPX_FORCEINLINE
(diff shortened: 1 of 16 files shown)


biddisco:
Thanks @hkaiser - pity though because I have implemented a 'signalling_executor' that uses a semaphore to do the same and allows the user to wait when spawning too many tasks. I'll have a look at this and get back to you.

biddisco:
I was planning on making a PR, but haven't finished tweaking the code yet. I wanted to add an 'auto-block' option etc so that I can replace my current implementation
```
{
   hpx::threads::executors::signalling_executor<parallel_executor> signal_exec(exec, 5000);
   for (std::uint64_t i = 0; i < count; ++i) {
       hpx::apply(signal_exec, &null_function);
       if (i % 1000 == 0) signal_exec.wait();
   }
}
``` 
with
```
{
   hpx::threads::executors::signalling_executor<parallel_executor> signal_exec(exec, 5000, auto_block=true);
   for (std::uint64_t i = 0; i < count; ++i) {
       hpx::apply(signal_exec, &null_function);
   }
}
``` 
and it will block on `apply` when you have more than 5000 (in this example) tasks outstanding. This means we don't need sliding semaphores any more for things like cholesky and it also solves the problem in cholesky of when you start with N=100, you want M iterations in flight, but as N decreases, you want M*K iterations in flight, now you can just say, keep 5000 tasks going and wake me when less than this are out there, block me when I spawn too many.

One thing I want to add is the ability for the task that is spawning many many subtasks to prote itself to high priority so that when it is woken, it can spawn more tasks asap without being in the queue anf waiting for the outstanding ones to drain.

Is there an API `set_state` call anywhere that allows task X to be created as a normal priority task, but upgrade itself to HP when it suspends? I didn't look yet if anything is already in there.

hkaiser:
> One thing I want to add is the ability for the task that is spawning many many subtasks to promote itself to high priority so that when it is woken, it can spawn more tasks asap without being in the queue and waiting for the outstanding ones to drain.

We currently do not support this, but I think it can be easily added. Please create a ticket and I'll see what I can do.

hkaiser:
>
> ```
> {
>    hpx::threads::executors::signalling_executor<parallel_executor> signal_exec(exec, 5000);
>    for (std::uint64_t i = 0; i < count; ++i) {
>        hpx::apply(signal_exec, &null_function);
>        if (i % 1000 == 0) signal_exec.wait();
>    }
> }

What is the difference between the `5000` passed to the constructor and the `1000` used to decide whether to wait?
```

**Comment by hkaiser** [comment]
>> One thing I want to add is the ability for the task that is spawning many many subtasks to promote itself to high priority so that when it is woken, it can spawn more tasks asap without being in the queue and waiting for the outstanding ones to drain.
>> Is there an API set_state call anywhere that allows task X to be created as a normal priority task, but upgrade itself to HP when it suspends? I didn't look yet if anything is already in there.
>
> We currently do not support this, but I think it can be easily added. Please create a ticket and I'll see what I can do.

Actually, after looking through the code I think you should be already able to do this. You can set the priority of a thread as:
```
// set the priority to use when this thread is scheduled next
thread_id_type id = threads::get_self_id();
id->set_priority(threads::thread_priority_boost);
…
```
Please note that `thread_priority_boost` will increase the threads priority to 'high' only for the next time is being scheduled. If you need to permanently raise its priority use `thread_priority_high`.

Another way to do this would be to simply call suspend:
```
threads::suspend(threads::pending_boost);   // note: reschedules this thread right away
```
The difference is that the latter raises the priority of the thread immediately while the former will apply the new priority only after it becomes suspended for the next time.

**Comment by biddisco** [comment]
> What is the difference between the 5000 passed to the constructor and the 1000 used to decide whether to wait?
just that if I were to do
```
if (i % 5000 == 0) signal_exec.wait();
```
then after spawning 5000 tasks, a few hundred might have already completed, so the wait call will return as only 4750 are outstanding and so it might not wait until we get to 10,000. By testing every 1000, I would only get to 6000 before it passed. A trivial tweak just to make sure I don't overdo the queuing of tasks.

> Actually, after looking through the code I think you should be already able to do this. You can set the priority of a thread as:
...

This is just what I wanted. If the task suspends, then it can either be upgraded first, or when it is woken, as long as it goes into an HP queue and is executed first, then it probably doesn't matter. I shall experiment. Thanks for the pointers.

**Comment by biddisco** [comment]
NB. I forgot to mention that it's partly because I don't like the 
```
if (i % 5000 == 0) signal_exec.wait();
```
That I thought I'd add a blocking option to the executor constructor so that you could just call apply/async and have the executor block/wait if the number of tasks got too large. Makes the user side code look much nicer and removes the need for those checks. Making it optional of course allows both manual or automatic waiting modes.

**Comment by hkaiser** [comment]
> if (i % 5000 == 0) signal_exec.wait();

Couldn't you do this for any executor (well, if it supported `wait`)?

**Comment by biddisco** [comment]
> Couldn't you do this for any executor (well, if it supported `wait`)?

If executors in general supported `wait` then yes. My own implementation uses a template param so that you can use any executor underneath and add the signalling/wait capability on top. 

```   executors::signalling_executor<parallel_executor> signal_exec(exec, 5000);```

I feel like I should combine my executor with yours? I don't change the way futures are produced, but just use `apply` instead of `async` if I don't need the individual futures.

**Comment by hkaiser** [comment]
Could we go ahead and merge this? Anybody interested in reviewing?

**Comment by biddisco** [comment]
I'll take a good look at it tonight/tomorrow

File: hpx/parallel/executors/parallel_executor_aggregated.hpp [hunk 1]
```diff
@@ -0,0 +1,356 @@
...
+        struct wrapped_func
+        {
+            template <typename Value>
+            void operator()(
+                lcos::local::counting_semaphore& sem, Value const& value)
+            {
+                // invoke user function
+                hpx::util::invoke_fused(f_,
+                    hpx::util::tuple_cat(
+                        hpx::util::forward_as_tuple(value), ts_));
```
sithhell: What happens when f_ throws an exception? The semaphore will never be signaled.

hpx/parallel/executors/parallel_executor_aggregated.hpp
```diff
@@ -0,0 +1,356 @@
...
```
sithhell: same as above

hpx/parallel/algorithms/for_each.hpp
```diff
@@ -81,6 +81,43 @@ namespace hpx { namespace parallel { inline namespace v1
             }
         };
 
+        template <typename F>
+        struct invoke_projected<F, util::projection_identity>
+        {
+            typename hpx::util::decay<F>::type& f_;
+            util::projection_identity& proj_;
```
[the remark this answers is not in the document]
hkaiser: Good catch, it's superfluous now. I'll remove it.

hpx/parallel/executors/parallel_executor_aggregated.hpp
hkaiser: Excellent catch.

msimberg:
FWIW, I've found when implementing the Kokkos backend that even a semaphore adds significant overhead when going to small/many tasks, especially on multiple NUMA domains. There I've for now settled on essentially just using an atomic counter, i.e. essentially something like this:

```
atomic<int> counter = n;
for (int i = 0; i < n; ++i) {
    apply([&counter, f]() { f(); --counter; });
}
yield_while([&counter]() { counter != 0; });
```

The spinlocks are what kills the performance. It's arguably very ugly, but abstracted away in an executor I think it would be worth the benefit.

hkaiser:
@msimberg the only reason I could think of why `counting_semaphore` is slower than your solution is that we have both, the internal spinlock and the actual semaphore sitting on the same cache line (see: https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/lcos/local/counting_semaphore.hpp#L104-L105). Otherwise (assuming no contention on the semaphore) it does not add anything in addition to your code.

hkaiser:
@msimberg, @sithhell, @biddisco: I believe this is good to go now.

hpx/parallel/util/partitioner.hpp
```diff
@@ -46,12 +46,12 @@ namespace hpx { namespace parallel { namespace util
             ExPolicy && policy,
             FwdIter first, std::size_t count, F && f)
         {
-            // estimate a chunk size based on number of cores used
+                    // estimate a chunk size based on number of cores used
```
msimberg: Looks like there is just added (unnecessary) indentation in this file.

hkaiser: Uhh, no idea how that happened... Will fix.

hkaiser: Should be fine now. Thanks!

hkaiser:
Since there don't seem to be any objections I'm going to merge this now.

PR #3606, opened 2019-01-04
Add nvcc to exclude constexpress since is it not supported by nvcc

diehlpk:
Fixes constexpress defines for nvcc compiler 

## Proposed Changes

  - Add nvcc compiler to this check https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/config/constexpr.hpp#L12

Since we have `!defined(HPX_MSVC_NVCC)` we should add NVdida's nvcc compiler too.

### DIFF
--- a/hpx/config/constexpr.hpp
+++ b/hpx/config/constexpr.hpp
@@ -9,7 +9,7 @@
 
 #include <hpx/config/defines.hpp>
 
-#if defined(HPX_HAVE_CXX11_CONSTEXPR) && !defined(HPX_MSVC_NVCC)
+#if defined(HPX_HAVE_CXX11_CONSTEXPR) && !defined(HPX_MSVC_NVCC) && !defined(__NVCC__)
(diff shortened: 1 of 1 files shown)


hpx/config/constexpr.hpp
```diff
@@ -9,7 +9,7 @@
 
 #include <hpx/config/defines.hpp>
 
-#if defined(HPX_HAVE_CXX11_CONSTEXPR) && !defined(HPX_MSVC_NVCC)
+#if defined(HPX_HAVE_CXX11_CONSTEXPR) && !defined(HPX_MSVC_NVCC) && !defined(__CUDACC__)
```
hkaiser: Please use `__NVCC__` instead of `__CUDACC__` as the latter is also defined by clang.

diehlpk: But constexpress work with cuda clang and we might want to use it there or? It seems they just not work on nvidia nvcc.

PR #3610, opened 2019-01-08
Don't wait for stop_condition in main thread

msimberg:
Skip waiting for `stop_condition_` if in main thread since main thread notifies `stop_condition_`.

This came up while doing #3486. Assertions thrown during init in main thread end up calling `notify_waiting_main` which waits for `stop_condition`. However, since the main thread is also expected to notify `stop_condition_` it would deadlock.


hkaiser:
@msimberg: We're seeing occasional hangs during shutdown(?) lately. Could that fix those?

msimberg:
@hkaiser maybe... Anything in particular that is timing out? In phylanx I suppose?

hkaiser:
> Anything in particular that is hanging? In phylanx I suppose?

@msimberg: we see it on CircleCI mostly on Phylanx, but HPX itself is not immune to this: https://circleci.com/gh/STEllAR-GROUP/hpx/53520.

PR #3613, opened 2019-01-09
Adding support for gitlab runners

sithhell:
This adds support for gitlab runners and enables a (semi modular)
setup to add new configurations. This will allow for a basis to add
more PR based tests for more configurations.
    
The whole workflow was slightly changed:
    - The tests are split up between unit, regression, performance and
    examples. Due to the enhanced infrastructure, we don't need to split
    them anymore
    - Checking the code with clang-tidy is enabled by default. A
    .clang-tidy file has been added for the clang-tidy configuration
    - We build both, a debug and release version. Only the debug version
    is currently pushed to dockerhub on the master branch
    
This is to replace the slow circle-ci setup (which has been replaced
with a dummy config)


.clang-tidy
```diff
@@ -0,0 +1,8 @@
+Checks: '-*,modernize-use-nullptr,misc-use-after-move,misc-virtual-near-miss,misc-multiple-statement-macro,misc-move-constructor-init,misc-move-forwarding-reference,misc-assert-side-effect,misc-dangling-handle,misc-non-copyable-objects,misc-forwarding-reference-overload,misc-unused-raii'
```
hkaiser: Boost license?

hpx/runtime/threads/coroutines/detail/context_base.hpp
```diff
@@ -102,22 +102,27 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail
...
+          m_apex_data(nullptr)
+          ,
 #endif
-            m_type_info(),
-            m_thread_id(id),
-            continuation_recursion_count_(0)
+          m_type_info()
+          , m_thread_id(id)
+          , continuation_recursion_count_(0)
         {}
```
hkaiser: This formatting is messed up, I believe.

hkaiser:
FWIW, we could leave the inspect checks to be run on CircleCI.

.clang-tidy
sithhell: good point.

hpx/runtime/threads/coroutines/detail/context_base.hpp
sithhell: hmm, this is what came out with clang-format-7, do we require a even newer version?

sithhell:
Ok, looks like the gitlab runner isn't started when having PRs from a fork... back to the drawing board...

hpx/runtime/threads/coroutines/detail/context_base.hpp
hkaiser: V7 should be fine. might be a bug...

sithhell:
Closing due to wrong reference branch

PR #3619, opened 2019-01-16
Store debug logs and core dumps on CircleCI

msimberg:
## Proposed Changes

- Add CMake option `HPX_WITH_TESTS_DEBUG_LOG` which when enabled will add the command line argument `--hpx:debug-hpx-log` to tests. `HPX_WITH_TESTS_DEBUG_LOG_DESTINATION` sets the output destination.
- Enable `HPX_WITH_TESTS_DEBUG_LOG` on CircleCI and store output as artifacts.
- Enable core dumps on CircleCI and store output as artifacts.

Trying to get more information about timeouts and failures in our tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,10 +13,17 @@ convert_xml: &convert_xml
     name: Converting XML
     when: always
     command: |
-        mkdir ${CIRCLE_JOB}
+        mkdir -p ${CIRCLE_JOB}
         xsltproc \
           /hpx/conv.xsl Testing/`head -n 1 < Testing/TAG`/Test.xml \
           > ${CIRCLE_JOB}/Test.xml
+
+move_debug_log: &move_debug_log
+    name: Moving debug logs
+    when: always
+    command: |
+        mkdir -p ${CIRCLE_JOB}
+        cp debug-log.txt ${CIRCLE_JOB} || true
 
 gh_pages_filter: &gh_pages_filter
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -13,11 +13,26 @@ convert_xml: &convert_xml
...
+        mkdir -p ${CIRCLE_JOB}
+        # Ignore errors if there are no core dumps
+        cp core.* ${CIRCLE_JOB} || true
+
+move_debug_log: &move_debug_log
+    name: Moving debug logs
+    when: on_fail
+    command: |
+        mkdir -p ${CIRCLE_JOB}
+        cp debug-log.txt ${CIRCLE_JOB} || true
```
sithhell: When interpreting this correctly, this will move the log of the last run test, right? This might not have been the test that failed, I guess.

msimberg: The logs get appended to the file. It'll contain logs from all the tests in that step. The downside of this is that the files can end up huge (GBs), but I've set it to run on failures only for that reason.

msimberg: So far just by checking what failed on CircleCI and searching for the logs of that test in the file. It has worked well enough so far. I was considering creating a log with a name corresponding to each test. I'm sure we could parse the ctest xml output to determine which tests failed but I didn't feel particularly motivated to do so. Do you think it would worth it?

PR #3620, opened 2019-01-16
Adding support for gitlab runner

sithhell:
This adds support for gitlab runners and enables a (semi modular) setup to add new configurations. This will allow for a basis to add more PR based tests for more configurations. The whole workflow was slightly changed:
 - The tests are split up between unit, regression, performance and examples. Due to the enhanced infrastructure, we don't need to split them anymore
 - Checking the code with clang-tidy is enabled by default. A .clang-tidy file has been added for the clang-tidy configuration
 - We build both, a debug and release version.

This can be seen as a first step to add more and unified testing using the packet.net infrastructure to our codebase. The tests are run via a gitlab mirror which starts gitlab runners

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,3 +1,8 @@
+# Copyright (c) 2019 Thomas Heller
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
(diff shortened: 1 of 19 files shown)


msimberg:
Hmm, you say push an image to docker hub but I can't see docker push anywhere. Is that implicit? And if it is actually there, we should probably disable the push from CircleCI?

sithhell:
@msimberg good point. I disabled that step since we decided to keep circle CI around for the time being...

Issue #3034, opened 2017-12-01
Use C++17 structured bindings for serialization


hkaiser:
In addition to the HPX serialization allowing to automatically serialize empty structs without explicit serialization functions we should add support for serializing brace-initializable types using structured bindings available in C++17 (see here for some examples: http://playfulprogramming.blogspot.com/2016/12/serializing-structs-with-c17-structured.html).

raymonwhite42:
I'll work on this soon.

hkaiser:
FWIW, serialization of empty types is handled here: https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/runtime/serialization/access.hpp#L101.

shantanudwvd:
hi,
i would like to work on this. Can anyone guide me?

hkaiser:
@shantanudwvd have you read the article linked above?

shantanudwvd:
Yes I have read it. 




Get Outlook for Android







On Thu, Mar 22, 2018 at 10:13 PM +0530, "Hartmut Kaiser" <notifications@github.com> wrote:












@shantanudwvd have you read the article linked above?

hkaiser:
> Yes I have read it.

Good, so what's you question? Also, I think we might be better off discussing this either on IRC or on the mailing list.

shantanudwvd:
i'm quite comfortable with c++ but i'm new to this. So can you please guide me?

hkaiser:
FWIW, here is an example where this technique was used: https://cista.rocks/.

hkaiser:
This can be closed as #3625 was merged

PR #3625, opened 2019-01-20
Added automatic serialization for simple structs (see #3034)

wteiwewte:
- Added automatic serialization for simple, brace-initializable structs. Every struct's field must be serializable and public.
  - Added one unit test for automatic serialization and removed fail_compile_non_serializable_class test (now it compiles, so no need to have it)
-  Arity function doesn't support copyable/movable structs with one field (because then exists constructor with one argument)
- Up to 15 fields (can be extended with simple copy&paste)


hpx/runtime/serialization/access.hpp
```diff
@@ -76,15 +108,34 @@ namespace hpx { namespace serialization
...
-                // finds only serialize-member function and doesn't
-                // perform ADL
+                template<class T1> struct dependent_false : std::false_type {};
+
+
                 template <class Archive>
                 static void call(Archive& ar, T& t, unsigned)
                 {
-                    detail::serialize_force_adl(ar, t, 0);
+                    if constexpr (has_serialize_adl<T>::value)
```
hkaiser: Could you please guard the use of `if constexpr` with a feature test macro. The whole new serialization functionality should be made available when compiling in C++17 mode only anyways, so enabling `if constexpr` at the same time should be ok.

wteiwewte:
- Added `__cpp_if_constexpr` test macro.

- `arity` function checks if there is no public constructor with given number of arguments. `hpx::util::tuple` has variadic constructor, so for every number of arguments it has public constructor, thus with `hpx::util::tuple` it won't work (same for `std::tuple`)

hpx/runtime/serialization/access.hpp
hkaiser: `__cpp_if_constexpr` is not supported when compiling with a C++11 compiler. You might want to add a feature test here: https://github.com/STEllAR-GROUP/hpx/tree/master/cmake/tests, https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_PerformCxxFeatureTests.cmake#L198, and here: https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_AddConfigTest.cmake#L661. You can then use the preprocessor macro to hide the new code from older compilers.

hkaiser:
@AntonBikineev is this something you'd be interested in reviewing?

hkaiser:
@wteiwewte could you please add the missing license/copyright as reported here: https://55221-4455628-gh.circle-artifacts.com/0/hpx_inspect_report.html? The other reported errors seem to be unrelated.

wteiwewte:
Sorry for the delay. I added missing licenses and feature tests (through cmake).

hkaiser:
@wteiwewte inspect is still unhappy with your proposed changes (see: https://circleci.com/gh/STEllAR-GROUP/hpx/57135). Additionally, could you please rebase this branch to resolve the conflicts with master?

hkaiser:
@wteiwewte Thanks a lot for working on this! @msimberg I think this is good to go now.

hkaiser:
@wteiwewte Congrats to your first accepted PR! Please get in contact with @aserio if you would like to receive a STE||AR-Group t-shirt. We give that away to our first-time contributors.

PR #3626, opened 2019-01-21
Fix cancelable_action_client example

msimberg:
Fixes errors like this:

```
terminating with uncaught exception of type hpx::thread_interrupted: std::exception
```

in the `cancelable_action_client` example.

## Proposed Changes

- Use atomic to protect `id_` instead of spinlock mutex

The commit message contains more details on what happens.

Side note: this is triggered because of `spinlock` calling `suspend` after a few iterations (in `yield_k`). `spinlock` to me sounds like it should never suspend the current thread. Should that be changed as well?

### DIFF
--- a/examples/cancelable_action/cancelable_action/server/cancelable_action.hpp
+++ b/examples/cancelable_action/cancelable_action/server/cancelable_action.hpp
@@ -11,7 +11,7 @@
 #include <hpx/include/actions.hpp>
 #include <hpx/include/threads.hpp>
 
-#include <mutex>
+#include <atomic>
 
 ///////////////////////////////////////////////////////////////////////////////
 namespace examples { namespace server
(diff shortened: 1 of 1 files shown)


sithhell:
Regarding the call to suspend: This is actually a misnomer ... what it really does is yielding the current thread. no suspension going on there...

examples/cancelable_action/cancelable_action/server/cancelable_action.hpp
```diff
@@ -67,19 +70,18 @@ namespace examples { namespace server
         // Cancel the lengthy action above
         void cancel_it()
         {
-            std::lock_guard<mutex_type> l(mtx_);
-            if (id_ != hpx::thread::id()) {
-                hpx::thread::interrupt(id_);
```
sithhell: hmm, wouldn't it make more sense to unlock `mtx_` before calling `hpx::thread::interrupt(id_)`?
Something like this:
```

hpx::thread_id::id id;
{
    std::lock_guard<mutex_type> l(mtx_);
    std::swap(id, id_);
}
if (id != hpx::thread::id())
{
    hpx::thread::interrupt(id);
}
````

msimberg: Yeah, that could be changed, but it's actually taking the lock in the `reset_id` destructor that causes the problem. To be honest, the example wouldn't even need an atomic (in this particular example and use case).

msimberg:
> Regarding the call to suspend: This is actually a misnomer ... what it really does is yielding the current thread. no suspension going on there...

True. However, the point is that `suspend` and `yield` both call `interruption_point`. And `yield_k` also eventually does timed suspension.

sithhell:
> > Regarding the call to suspend: This is actually a misnomer ... what it really does is yielding the current thread. no suspension going on there...
> 
> True. However, the point is that `suspend` and `yield` both call `interruption_point`. And `yield_k` also eventually does timed suspension.

fair enough.

examples/cancelable_action/cancelable_action/server/cancelable_action.hpp
sithhell: Ok, I haven't looked too closely. Just noticed the slight change of semantics with your proposed changes...

PR #3628, opened 2019-01-21
Modular Assert

sithhell:
This changeset requires #3582.

Moving on with the modularization by adding HPX.Assert. HPX.Assert has HPX.PP and HPX.Config as dependencies, which are included in this PR.

The assert implementation has been improved to not evaluate the expression to assert on when assertions are turned of (similar to the optional message). Documentation has been added to all three modules.


libs/assert/tests/unit/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+  $<TARGET_FILE:hpx.assert.test.assert_fail>)
+
+set_tests_properties(tests.unit.assert.test.assert_fail
+  PROPERTIES WILL_FAIL $<$<CONFIG:Debug>:true>)
+
+add_executable(hpx.assert.test.assert_succeed assert_succeed.cpp)
+target_link_libraries(hpx.assert.test.assert_succeed hpx.assert)
+add_test(NAME tests.unit.assert.test.assert_succeed
+  COMMAND $<TARGET_FILE:hpx.assert.test.assert_succeed>)
+
```
hkaiser: Why is `assert_fail` run using `hpxrun.py` but `assert_succeed` is run directly?

libs/config/src/asio.cpp
```diff
@@ -0,0 +1,8 @@
+//  Copyright (c) 2019 Thomas Heller
+//
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/config/asio.hpp>
+
+// This file is intentionally empty to ensure self contained headers
```
hkaiser: What are these source files needed for?

src/exception.cpp
```diff
@@ -562,18 +509,20 @@ namespace hpx
...
+        std::string const* env = xi.get<hpx::detail::throw_env>();
+        if (env && !env->empty())
+            strm << "{env}: " << *env;
+
         std::string const* back_trace =
             xi.get<hpx::detail::throw_stacktrace>();
         if (back_trace && !back_trace->empty()) {
             // FIXME: add indentation to stack frame information
             strm << "{stack-trace}: " << *back_trace << "\n";
         }
```
hkaiser: We've had the stack backtrace first as users where complaining that they can't find it in the output easily...

tests/performance/local/CMakeLists.txt
```diff
@@ -40,9 +40,9 @@ if(HPX_WITH_EXAMPLES_OPENMP)
      )
 
   set(openmp_homogeneous_timed_task_spawn_FLAGS NOLIBS
-      DEPENDENCIES ${boost_library_dependencies})
-  set(openmp_homogeneous_timed_task_spawn_FLAGS NOLIBS
-      DEPENDENCIES ${boost_library_dependencies})
+      DEPENDENCIES ${boost_library_dependencies} hpx.config hpx.assert)
+  set(openmp_parallel_region_FLAGS NOLIBS
```
hkaiser: `hpx.assert` already depends on `hpx.config`. Would it be possible to automatically recursively append dependencies? Otherwise every HPX application will have to list many sub-libraries, which is probably a pain in the end. I think having a top-level dependency `hpx` should be sufficient...

libs/assert/tests/unit/CMakeLists.txt
sithhell: `assert_fail` calls std::abort, which will make the `WILL_FAIL` make fail. This is a cmake problem which can be circumvented with a such a driver script. There is no particular reason why `assert_succeed` is run directly, this could be changed for sure.

libs/assert/include/hpx/assert.hpp
```diff
@@ -0,0 +1,82 @@
...
+#define HPX_ASSERT_MSG(...) HPX_ASSERT(__VA_ARGS__)
+#else
+#define HPX_ASSERT(...)                                                        \
+    {                                                                          \
+    }
+#define HPX_ASSERT_MSG(...)                                                    \
+    {                                                                          \
+    }
+#endif
+#endif
```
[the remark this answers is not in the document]
sithhell: Yes. Is there any usecase that requires this functionality?

libs/assert/include/hpx/assert.hpp
```diff
@@ -0,0 +1,82 @@
...
+#include <hpx/config.hpp>
+#include <hpx/assert/current_function.hpp>
+#include <hpx/assert/evaluate_assert.hpp>
+#include <hpx/assert/source_location.hpp>
+#include <hpx/pp/stringize.hpp>
+
+#include <string>
+#include <type_traits>
+
+namespace hpx { namespace assertion {
```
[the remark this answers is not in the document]
sithhell: `namespace assert` is a problem once someone included `cassert`.

sithhell:
The mentioned unrelated changes are mostly due to clang-format being applied on the changes...

libs/create_library_skeleton.py
```diff
@@ -0,0 +1,210 @@
...
+    print('Usage: %s <lib_name>' % sys.argv[0])
+    print('Generates the skeleton for lib_name in the current working directory')
+    sys.exit(1)
+
+lib_name = sys.argv[1]
+
+# CMake minimum version
+cmake_version = '3.3.2'
+
+cmake_header =  '# Copyright (c) 2019 The STE||AR-Group\n'
```
msimberg: Wouldn't python's multiline strings work here? And down below in the calls to `f.write`. Just for readability...

libs/config/src/asio.cpp
hkaiser: Why do we need explicit source files for this? Those could be generated (as before), couldn't they?

libs/assert/include/hpx/assert.hpp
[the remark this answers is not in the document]
hkaiser: Then we should name the module `assertion` as well.

sithhell:
In order to make the review easier, I factored out the first commit to have a preprocessor library into #3629. Once this one is merged, I will do the same for the config library, and once that is done, the assert library as proposed here can be reviewed more easily.

libs/config/src/asio.cpp
sithhell: The biggest issue with the autogenerated header tests is that you easily end up with dangling headers. This completely eliminates this issue at the cost of added verbosity.

sithhell:
> > The main motivation the assertion handling was changed (at least in the implementation) is that our current implementation always executes the expression to assert on. This implementation gets rid of that behavior and is able to have the assertion completely optimized away for release builds.
> 
> @sithhell This is a bug indeed. However the same could have been achieved by simply fixing the pp expressions that guard `HPX_ASSERT`.

it's not the expressions that guard `HPX_ASSERT` it's the implementation itself. See here:
https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/util/assert.hpp#L36-L38

hkaiser:
> it's not the expressions that guard HPX_ASSERT it's the implementation itself. See here:
https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/util/assert.hpp#L36-L38

@sithhell Sure, just removing that broken bit and use this: https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/util/assert.hpp#L40, fixes this bug however (plus fixing the pp guard). And frankly, I'm not sure why those lines (L36-L38) were added in the first place...

Mind you, I'm not opposed to changing the implementation of assert (if this is necessary or beneficial). I'm opposed to combining restructuring efforts with implementation changes for no apparent reason.

hkaiser:
@sithhell … and in case I have not made this clear: thanks for driving this forward! I think this can help cleaning up the structure of HPX significantly.

libs/create_library_skeleton.py
hkaiser: I'd even suggest separating the template text into their own files and letting the script fill in the blanks. Python is good at that kind of work. That will simplify maintenance.

libs/config/src/asio.cpp
hkaiser: Please elaborate, I'm not sure I understand what you mean. The generated files look exactly as the files you created manually (**shiver**), well - perhaps sans the license comment.

libs/assert/include/hpx/assert.hpp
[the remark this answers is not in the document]
hkaiser: Alternatively, let's call the namespace `_assert` or `assert_`.

msimberg:
@sithhell what's the plan for this and #3629?

libs/config/src/asio.cpp
msimberg: @sithhell could you explain how we might end up with dangling headers (unused headers, I presume?) more easily if the tests are auto-generated?

tests/performance/local/CMakeLists.txt
msimberg: I think dependencies should be specified explicitly if they're used directly in the module/application, not transitively (CMake takes care of this case).

msimberg:
Closing in favor of a cleaned up version coming up soon.

PR #3633, opened 2019-01-25
Replace deprecated boost endian macros

K-ballo:


K-ballo:
> What's the minimum boost version having predef? Should we check that in our CMake scripts?
1.55.0, our minimum required one.

Issue #3634, opened 2019-01-27
The build fails if shared_future<>::then is called with a thread executor


chinz07:
## Expected Behavior

The build should succeed.

## Actual Behavior

The build fails with the error
```
/opt/hpx-debug/include/hpx/parallel/executors/thread_execution.hpp:118:18: error: ‘class hpx::threads::executor’ has no member named ‘get_stacksize’
             exec.get_stacksize(),
             ~~~~~^~~~~~~~~~~~~
```
The full error message is:
```
In file included from /opt/hpx-debug/include/hpx/parallel/executors/default_executor.hpp:14,
                 from /opt/hpx-debug/include/hpx/parallel/executors.hpp:11,
                 from /home/christopher/git/work/Qubus/qubus/src/dataflow.cpp:3:
/opt/hpx-debug/include/hpx/parallel/executors/thread_execution.hpp: In instantiation of ‘typename std::enable_if<hpx::traits::is_threads_executor<Policy>::value>::type hpx::threads::post(Executor&&, F&&, Ts&& ...) [with Executor = hpx::threads::executor&; F = std::pair<hpx::threads::thread_state_enum, hpx::threads::thread_id_type> (hpx::lcos::detail::continuation<hpx::lcos::shared_future<void>, hpx::util::detail::one_shot_wrapper<hpx::util::detail::bound_back<qubus::dataflow_graph::schedule_modification(const qubus::object&)::<lambda(const hpx::lcos::shared_future<void>&)> > >, qubus::access_token>::*&)(boost::intrusive_ptr<hpx::lcos::detail::future_data_base<hpx::traits::detail::future_data_void> >&&); Ts = {boost::intrusive_ptr<hpx::lcos::detail::continuation<hpx::lcos::shared_future<void>, hpx::util::detail::one_shot_wrapper<hpx::util::detail::bound_back<qubus::dataflow_graph::schedule_modification(const qubus::object&)::<lambda(const hpx::lcos::shared_future<void>&)> > >, qubus::access_token> >, boost::intrusive_ptr<hpx::lcos::detail::future_data_base<hpx::traits::detail::future_data_void> >}; typename std::enable_if<hpx::traits::is_threads_executor<Policy>::value>::type = void]’:
[...]

PR #3637, opened 2019-01-28
Merging the executor-enabled overloads of shared_future<>::then

chinz07:
The separate overload for thread executors caused problems since
the implementation of `then` requires a `scheduled_executor` and not
just an `executor`.

The merged overload still does not test if the user-provided executor
does actually satisfy the requirements of a `scheduled_executor`.
Since future<>::then does exhibit the same issue, this should be
fixed in a separate PR.

This commit also adds a regression test demonstrating the issue.

This fixes issue #3634.

### DIFF
--- a/tests/regressions/lcos/shared_future_then_with_executor_3634.cpp
+++ b/tests/regressions/lcos/shared_future_then_with_executor_3634.cpp
@@ -1,0 +1,25 @@
+//  Copyright (c) 2019 Christopher Hinz
+//
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// This test case demonstrates the issue described in #3634:
+// The build fails if shared_future<>::then is called with a thread executor
+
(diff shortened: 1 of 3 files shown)


tests/regressions/lcos/shared_future_then_with_executor_3634.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include <hpx/parallel/executors.hpp> // Workaround for a missing header file
+#include <hpx/runtime/threads/executors/pool_executor.hpp>
+
+int main()
+{
+    hpx::threads::executors::pool_executor executor("default");
+
+    auto future = hpx::make_ready_future().share();
+
+    future.then(executor, [](auto future) { future.get(); });
```
msimberg: Could you change this to `hpx::shared_future<void>` instead of `auto`. `auto` is not allowed with C++11 (http://cdash.cscs.ch//viewBuildError.php?buildid=37278).

chinz07: Of course I can. Thanks for the remark.

PR #3644, opened 2019-02-01
Attempt to tackle timeouts during startup

sithhell:
This PR enables the deadlock detection for our spinlocks and unlocks the lock before throwing an exception during loading of components.


hkaiser:
This didn't fix the timeouts :/ (https://circleci.com/gh/STEllAR-GROUP/hpx/58464)

sithhell:
No, migrate_component suffers from a different problem than those at startup... no idea what yet...

PR #3656, opened 2019-02-04
Avoiding lifetime problems with sync_put_parcel

sithhell:


hkaiser:
This change does not make sense to me. The promise is guaranteed to outlive the execution of the callback as it will be destroyed only after the future has become ready (https://github.com/STEllAR-GROUP/hpx/blob/master/src/runtime/parcelset/parcelhandler.cpp#L98). 

If there really is a race between the invocation of the callback and the synchronization with the future (which I miss to see), we should move the promise into the capture instead of wrapping it into a shared_ptr instead.

sithhell:
Not quite. When a future is triggered, we set the state to value/exception after copying over the value/exception. When waiting on a future, we first check if the state is not empty (without taking the lock), thus leading to the lifetime issue this PR fixes.

hkaiser:
The promise is not accessed anymore once we have set the value. The `get` will never return before we have set the value, thus we will not access the promise anymore. Even if the value is set before `get` is called on the future, I still don't see where the race is. Sorry.

sithhell:
Thanks for the discussion. Closing this in favor of #3663.

PR #3657, opened 2019-02-04
Avoid dangling references in wait_all

sithhell:


hkaiser:
@sithhell as said before, this does not make any sense to me. The reference is guaranteed to stay valid as the referenced data (https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/lcos/wait_all.hpp#L394) will be kept alive until after all futures have become ready. If this change silences one of the sanitizers then we have bigger problems than a dangling reference.

sithhell:
As with #3656, I assume that there is a similar race between waiting on the future and the lifetime of the stuff we wait on.

hkaiser:
Well, assumptions are not enough to warrant copying everything instead of just referring to it. I really would like to understand what's wrong here.

hkaiser:
@sithhell Can we close this in favor of #3663 as well?

sithhell:
> @sithhell Can we close this in favor of #3663 as well?

I'm not sure yet. I am still getting lots of failures on the sanitizers branch

sithhell:
After some more tests, I am certain we can close this now as it has been resolved by #3663 and #3666.

PR #3658, opened 2019-02-04
Fixing execution_parameters

sithhell:
Fixed the handling of reference wrappers to be able to safely copyable.

### DIFF
--- a/hpx/parallel/executors/execution_parameters.hpp
+++ b/hpx/parallel/executors/execution_parameters.hpp
@@ -433,7 +433,6 @@ namespace hpx { namespace parallel { namespace execution
         template <typename T, typename Wrapper, typename Enable = void>
         struct maximal_number_of_chunks_call_helper
         {
-            maximal_number_of_chunks_call_helper(Wrapper&) {}
         };
 
         template <typename T, typename Wrapper>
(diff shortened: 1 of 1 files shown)


hpx/parallel/executors/execution_parameters.hpp
```diff
@@ -617,13 +581,7 @@ namespace hpx { namespace parallel { namespace execution
...
 
             unwrapper(wrapper_type wrapped_param)
-              : base_member_helper<wrapper_type>(std::move(wrapped_param))
-              , maximal_number_of_chunks_call_helper<T, wrapper_type>(this->member_)
-              , get_chunk_size_call_helper<T, wrapper_type>(this->member_)
-              , mark_begin_execution_call_helper<T, wrapper_type>(this->member_)
-              , mark_end_execution_call_helper<T, wrapper_type>(this->member_)
-              , processing_units_count_call_helper<T, wrapper_type>(this->member_)
-              , reset_thread_distribution_call_helper<T, wrapper_type>(this->member_)
+              : base_member_helper<wrapper_type>(wrapped_param)
```
hkaiser: This does not look right, `wrapper_param` is now always copied in the constructor.

hpx/parallel/executors/execution_parameters.hpp
```diff
@@ -640,13 +598,7 @@ namespace hpx { namespace parallel { namespace execution
...
```
hkaiser: Same here, this does not look right.

hpx/parallel/executors/execution_parameters.hpp
hkaiser: I understand that the base classes do not need the reference to the derived class anymore (good thinking on that on!). I was referring to the constructor `base_member_helper<wrapper_type>(wrapped_param)` where you have removed the `std::move`s causing the argument to be copied eventually.

hkaiser: Right, I missed that.

PR #3659, opened 2019-02-04
Parallel executor latch

K-ballo:
Change `parallel_executor::spawn` to wait on a `local::latch` instead of hierarchical `when_all`s. 

Flyby: reuse the spawn thread to spawn the last chunk of tasks.


sithhell:
any idea how much performance improvement that gives us?

hpx/parallel/executors/parallel_executor.hpp
```diff
@@ -162,53 +163,44 @@ namespace hpx { namespace parallel { namespace execution
...
 
-                while (size != 0)
+                while (size > chunk_size)
                 {
-                    std::size_t curr_chunk_size = (std::min)(chunk_size, size);
-
-                    hpx::future<void> f = async_execute(
-                        spawn_func, this, std::ref(results), base,
-                        curr_chunk_size, num_tasks, std::ref(func), it,
+                    async_execute(
```
sithhell: Any reason why you didn't use `post` instead of `async_execute` here? This would save another allocation of the shared state.

K-ballo: No reason, will change.

K-ballo:
> any idea how much performance improvement that gives us?

No. Is there a benchmark we could run?

I would not expect the change to be noticeable over the bulk tasks being executed.

sithhell:
> > any idea how much performance improvement that gives us?
> 
> No. Is there a benchmark we could run?
> 
> I would not expect the change to be noticeable over the bulk tasks being executed.

This might be worth a try:
https://github.com/STEllAR-GROUP/hpx/blob/master/tests/performance/local/foreach_scaling.cpp

sithhell:
There's also one test failing related to this change:
https://circleci.com/gh/STEllAR-GROUP/hpx/60041

hkaiser:
> WRT to performance gain, I am seeing an overall improvement with a speedup between 1.0 and 2.23 on my 8 core machine here.

That roughly corresponds to the performance gain we were seeing for the parallel_executor_aggregated PR (see #3600).

K-ballo:
Using `async_execute` and dropping the result leads to trouble with the _deferred_ policy, as deferred execution is never triggered.
Using `post` leads to trouble with the _sync_ policies, as they spawn new threads to schedule the tasks (test checks for inline execution).
Augmenting `post` to acknowledge _sync_ policies causes the _then_ tests to fail (test checks for no inline execution of the continuation, regardless of policy).

hkaiser:
> Augmenting `post` to acknowledge _sync_ policies causes the _then_ tests to fail (test checks for no inline execution of the continuation, regardless of policy).

This sounds like a problem with the test

K-ballo:
> > Augmenting `post` to acknowledge _sync_ policies causes the _then_ tests to fail (test checks for no inline execution of the continuation, regardless of policy).
> 
> This sounds like a problem with the test

I'll adjust the test accordingly.

K-ballo:
Once adjusted, the _then_ tests are now detecting that new threads are being spanwed for the _deferred_ policy (a synchronous policy).

hkaiser:
@K-ballo can I help with this in any way?

K-ballo:
> @K-ballo can I help with this in any way?

@hkaiser you could look at the parallel_policy_executor test and confirm it reflects the expected semantics, or suggest design changes if it doesn't.

hkaiser:
@K-ballo the only things I think are not quite clear on what the expected behavior should be are the `then_execute` and `bulk_then_execute` calls using the `deferred` policy. 

Would that require to launch the operations synchronously only once the returned future is asked for its value? Does it make sense to support that case in the first place?

Currently, the `[bulk_]then_execute` customization points simply wrap the required operation(s) into a continuation that is attached to the returned future. This continuation is _immediately_ launched through the executor using its `post` operation. This is not what I would expect as the execution should be triggered only once the returned future is queried.

The current tests for the `deferred` policy don't check for whether the invocation was `deferred` in the first place. They only check whether the invocation was executed on the calling thread.

A possible solution would be to implement the `[bulk_]then_execute` executor functions for `parallel_policy_executor<deferred>`. This would circumvent the behavior implemented by the customization point allowing us to actually defer the invocation of the operation. I can have a look into this.

K-ballo:
@hkaiser The impression I get is that the executor shouldn't even be attempting to do the hierarchical spawn for sync policies.

hkaiser:
> The impression I get is that the executor shouldn't even be attempting to do the hierarchical spawn for sync policies.

Yes, that sounds about right. Things have to be executed synchronously anyways. There is no need to complicate matters in this case...

K-ballo:
> @hkaiser The impression I get is that the executor shouldn't even be attempting to do the hierarchical spawn for sync policies.

This has simplified code slightly, but has not altered behavior as `[bulk_]then_execute` makes use of the fallback synthesized implementation. The fallback implementation appears to not honor policies, that is not changed in this PR.

hkaiser:
>  The fallback implementation appears to not honor policies

I think they should not. The launch policy is a property of the executor.

sithhell:
I'm not a fan merging a PR which introduces a failing test at this point. Can you think of a way to avoid this?

K-ballo:
> I'm not a fan merging a PR which introduces a failing test at this point.

The test is not supposed to be failing, there's something actually wrong.

sithhell:
> > I'm not a fan merging a PR which introduces a failing test at this point.
> 
> The test is not supposed to be failing, there's something actually wrong.

So what do you propose? I understand that the test is already broken on master and that your PR just makes the problem visible. What I am wondering is whether we can fix the test first, and then merge your PR.

K-ballo:
> > > I'm not a fan merging a PR which introduces a failing test at this point.
> > 
> > 
> > The test is not supposed to be failing, there's something actually wrong.
> 
> So what do you propose? I understand that the test is already broken on master and that your PR just makes the problem visible. What I am wondering is whether we can fix the test first, and then merge your PR.

I propose to fix whatever is wrong with the PR. The test has been adjusted to not be broken given the existing behavior so it should not be failing, it should be checking for known bad behavior.

K-ballo:
This should be fine now. @hkaiser please double check the test before merging.

hkaiser:
@K-ballo I have added the missing launch policy support for `then_execute` and `bulk_then_execute`. I think this is good to go now.

K-ballo:
> @K-ballo I have added the missing launch policy support for `then_execute` and `bulk_then_execute`. I think this is good to go now.

It seems there was a bad rebase, but it was eventually resolved correctly. LGTM once the missing includes are added.

hkaiser:
@msimberg This is good to go now. Please merge at your leasure.

msimberg:
@hkaiser this does look good to go but unless you need this right now I'd like to see this pass through pycicle once more (there's been some delays and some of the latest builds seem to be with older commits). @sithhell just a heads up, this seems to introduce new warnings on the CUDA build: http://cdash.cscs.ch//viewBuildError.php?buildid=42110. Do you think these will make your life harder on the CUDA branch, or are you changing so much that it won't matter?

hkaiser:
@msimberg Sure, no rush. Also, those warnings should be gone.

msimberg:
> Also, those warnings should be gone.

Indeed they are, thanks! The clang 3.9 builder only fails because of the change I made for #3566, but this PR passed before that change so in this goes...

PR #3662, opened 2019-02-05
Attempt to fix exception handling

sithhell:
- Do not pass exception_ptr as a reference when handling completions.
   This avoids dangling references during completion handling
 - Fixing correct detection if RP is running or not (needed for reporting exceptions if the runtime isn't up yet)
 - Removing typeid based lookup for exception_info and cleanup


src/exception.cpp
```diff
@@ -251,21 +252,22 @@ namespace hpx { namespace detail
...
-               hpx::detail::throw_thread_name(thread_name),
-               hpx::detail::throw_function(func),
-               hpx::detail::throw_file(file),
-               hpx::detail::throw_line(line),
-               hpx::detail::throw_env(env),
-               hpx::detail::throw_config(config),
-               hpx::detail::throw_state(state_name),
-               hpx::detail::throw_auxinfo(auxinfo)));
+            throw_with_info(e,
+                std::move(hpx::exception_info().set(
```
hkaiser: This `move` looks wrong (same below, btw).

sithhell: It's not wrong per se. `hpx::exception_info::set` returns a reference to itself, which can be safely moved since it is never used again afterward.

K-ballo:
- `exception_with_info_base` was necessary for implementing `diagnostic_info`, but it seems we left that part of the proposal out (HPX has its own diagnotsic info facilities).
- another functionality of the proposal that we might not be using is the support for attaching new exception information during handling, if so we could remove the entire logic of chaining nodes, etc.
- replacing `typeid` with `dynamic_cast` seems to complicate the code for no apparent benefit, what's the rationale for the change?

sithhell:
> * `exception_with_info_base` was necessary for implementing `diagnostic_info`, but it seems we left that part of the proposal out (HPX has its own diagnotsic info facilities).

can you link to that proposal?

> * another functionality of the proposal that we might not be using is the support for attaching new exception information during handling, if so we could remove the entire logic of chaining nodes, etc.
> * replacing `typeid` with `dynamic_cast` seems to complicate the code for no apparent benefit, what's the rationale for the change?

The rationale was to make it less heavyweight. The proposed changes don't need the header `typeinfo` and `tuple`. It slightly reduces binary sizes.

K-ballo:
> > * `exception_with_info_base` was necessary for implementing `diagnostic_info`, but it seems we left that part of the proposal out (HPX has its own diagnotsic info facilities).
> 
> can you link to that proposal?

[P0640](https://wg21.link/p0640), now withdrawn. The HPX implementation was derived from [this](https://gist.github.com/K-ballo/6b7bebde6db546f25be4adadaf18edc4).



> > * replacing `typeid` with `dynamic_cast` seems to complicate the code for no apparent benefit, what's the rationale for the change?
> 
> The rationale was to make it less heavyweight. The proposed changes don't need the header `typeinfo` and `tuple`. It slightly reduces binary sizes.

`dynamic_cast` needs more runtime type information than `typeid`, it supports partial down-casts as well as side-casts.

hkaiser:
FWIW, I'm seeing segmentation faults during exception handling on this branch. I'll try to construct a test reproducing the problem.

sithhell:
@K-ballo I reverted the changes to exeption_info and only left the fixes regarding move/forward in place.

sithhell:
@hkaiser this probably has something to do with me messing up the copyable requirement... this should be fixed now.

K-ballo:
> @K-ballo I reverted the changes to exeption_info and only left the fixes regarding move/forward in place.

There's a lot of unused cruft in the `exception_info` mechanism, let's look at removing it in a future PR. I'm particularly curious at what the effect would be of just removing the `tuple` dependency (while retaining `typeid` based lookups).

sithhell:
Are there still any objections to this?

msimberg:
@sithhell judging by the number of comments there are likely no more objections. However, [this](http://cdash.cscs.ch//viewBuildError.php?buildid=43426) would still need to be fixed.

sithhell:
@msimberg thanks, this should be fixed now.

hpx/runtime/threads/detail/scheduled_thread_pool_impl.hpp
```diff
@@ -567,6 +567,8 @@ namespace hpx { namespace threads { namespace detail
 
                 // run the work queue
                 hpx::threads::coroutines::prepare_main_thread main_thread;
+                // avoid unused variable error
+                (void)main_thread;
```
msimberg: `HPX_UNUSED`?

PR #3672, opened 2019-02-08
Change new build system to use existing macros related to pseudo dependencies

hkaiser:


hkaiser:
> What's the advantage of using the custom functions over plain add_custom/add_dependency?

This prevent for the targets to be created in Visual Studio project files. There, those are not needed and simply get in the way by slowing down things.

sithhell:
> > What's the advantage of using the custom functions over plain add_custom/add_dependency?
> 
> This prevent for the targets to be created in Visual Studio project files. There, those are not needed and simply get in the way by slowing down things.

Fair enough ... I was hoping to get rid of our HPX specific macros here :(

PR #2404, opened 2016-11-22
Forward the explicitly given result_type in the hpx invoke

atrantan:
This pull request tries to solve the issue when result_of cannot deduce the return type of the callable object, i.e. it happens when the callable object is an HPX_DEVICE lambda.

PR #3674, opened 2019-02-08
Reduce functional utilities call depth

K-ballo:
Reduce the number of function calls necessary to invoke the target callable in the functional utilities.

This PR relays on reverting the workaround introduced in #2404, and instead fixes the underlying bug that is believed to have caused the original issue (for which there is no report and no tests).

Each "flatten" commit removes one frame from the call stack. Most changes result in code simplification, but two commits (titled "flatten ... some more") result in more complex code in order to embed index sequences into the wrappers.

### DIFF
--- a/hpx/util/invoke.hpp
+++ b/hpx/util/invoke.hpp
@@ -173,9 +173,9 @@ namespace hpx { namespace util
             template <typename F, typename... Ts>
             HPX_CONSTEXPR HPX_HOST_DEVICE
             typename util::invoke_result<F, Ts...>::type
-            operator()(F && f, Ts &&... vs)
+            operator()(F && f, Ts &&... vs) const
             {
-                return util::invoke(std::forward<F>(f),
+                return HPX_INVOKE(std::forward<F>(f),
                     std::forward<Ts>(vs)...);
             }
         };
@@ -185,9 +185,9 @@ namespace hpx { namespace util
         {
             template <typename F, typename... Ts>
             HPX_CONSTEXPR HPX_HOST_DEVICE
-            R operator()(F && f, Ts &&... vs)
+            R operator()(F && f, Ts &&... vs) const
             {
-                return util::invoke_r<R>(std::forward<F>(f),
+                return HPX_INVOKE_R(R, std::forward<F>(f),
                     std::forward<Ts>(vs)...);
             }
         };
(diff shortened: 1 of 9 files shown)


hpx/util/invoke.hpp
```diff
@@ -162,24 +162,20 @@ namespace hpx { namespace util
...
             template <typename F, typename... Ts>
             HPX_CONSTEXPR HPX_HOST_DEVICE
             typename util::invoke_result<F, Ts...>::type
-            operator()(F && f, Ts &&... vs) const
+            operator()(F && f, Ts &&... vs)
             {
-                typedef typename util::invoke_result<F, Ts...>::type R;
-
-                return hpx::util::void_guard<R>(), util::invoke(std::forward<F>(f),
+                return util::invoke(std::forward<F>(f),
```
hkaiser: Couldn't this use `HPX_INVOKE` instead (saving another call level)?

hpx/util/invoke.hpp
```diff
@@ -189,14 +185,13 @@ namespace hpx { namespace util
         {
             template <typename F, typename... Ts>
             HPX_CONSTEXPR HPX_HOST_DEVICE
-            R operator()(F && f, Ts &&... vs) const
+            R operator()(F && f, Ts &&... vs)
             {
-                return hpx::util::void_guard<R>(), util::invoke_r<R>(std::forward<F>(f),
+                return util::invoke_r<R>(std::forward<F>(f),
                     std::forward<Ts>(vs)...);
```
hkaiser: Same here, this could use `HPX_INVOKE_R` instead.

hpx/util/invoke.hpp
K-ballo: Yes, they could be. These did not show up anywhere, they seem to be unused.

sithhell:
It seems like const isn't propagated properly:
https://stellar-group.gitlab.io/-/hpx/-/jobs/158592760/artifacts/tests.unit.html#1b048cec-0235-474d-8aae-fbcc0372a85e

hkaiser:
There are also additional test failures caused by the proposed changes: https://stellar-group.gitlab.io/-/hpx/-/jobs/159443858/artifacts/tests.unit.html#775cc3d3-3e99-462e-91fa-e374cdeb7c28

K-ballo:
> It seems like const isn't propagated properly:

Caused by the different semantics between `decltype(e)` and `decltype((e))`.

PR #3679, opened 2019-02-09
Simplify basic action implementations

K-ballo:


hpx/runtime/components/server/console_error_sink.hpp
```diff
@@ -20,9 +20,7 @@ namespace hpx { namespace components { namespace server
     // console logging happens here
     void console_error_sink(std::exception_ptr const&);
 
-    typedef actions::action<
-        void (*)(std::exception_ptr const&), console_error_sink
-    > console_error_sink_action;
+    HPX_DEFINE_PLAIN_ACTION(console_error_sink, console_error_sink_action);
```
K-ballo: Why is this needed? and is it the correct way to declare such an action?

hpx/runtime/actions/basic_action.hpp
```diff
@@ -290,23 +308,18 @@ namespace hpx { namespace actions
...
-                        lva, comptype, std::forward<Ts>(vs)...)));
-            }
+                target = {};
 
+            using thread_function = detail::thread_function<Derived>;
             return traits::action_decorate_function<Derived>::call(lva,
-                util::one_shot(util::bind(
-                    typename Derived::thread_function(target),
-                    lva, comptype, std::forward<Ts>(vs)...)));
+                thread_function(target, lva, comptype,
```
hkaiser: Should `target` be moved here?

hpx/runtime/actions/basic_action.hpp
```diff
@@ -315,55 +328,44 @@ namespace hpx { namespace actions
...
-                    thread_function(std::move(cont), lva, invoker(),
-                        lva, comptype, std::forward<Ts>(vs)...));
-            }
+                target = {};
 
+            using thread_function = detail::continuation_thread_function<Derived>;
             return traits::action_decorate_function<Derived>::call(lva,
-                thread_function(target, std::move(cont), lva, invoker(),
-                    lva, comptype, std::forward<Ts>(vs)...));
+                thread_function(target, std::move(cont), lva, comptype,
```
hkaiser: Same here, should `target` be moved?

hpx/runtime/actions/basic_action.hpp
K-ballo: It's now a local copy so it should at least not hurt moving it.

hkaiser: Moving it would safe an atomic increment. I'd suggest doing it.

hpx/runtime/actions/basic_action.hpp
```diff
@@ -180,108 +263,43 @@ namespace hpx { namespace actions
...
-                    // report this error to the console in any case
-                    hpx::report_error(std::current_exception());
-                }
-
-                // Verify that there are no more registered locks for this
-                // OS-thread. This will throw if there are still any locks
-                // held.
-                util::force_error_on_lock();
-                return threads::thread_result_type(threads::terminated,
-                    threads::invalid_thread_id);
```
sithhell: What happened to the exception handling?

K-ballo: The nested `thread_function` class moved to namespace scope, and now resides at lines 88:149

sithhell: Thanks, missed that!

sithhell:
clang doesn't like all of your changes:
```
/hpx/source/hpx/runtime/actions/component_action.hpp:91:5: error: 'action' defined as a class template here but previously declared as a struct template [-Werror,-Wmismatched-tags]
    class action<R (Component::*)(Ps...), F, Derived>
    ^
/hpx/source/hpx/runtime/actions/basic_action.hpp:528:5: note: did you mean class here?
    struct action;
    ^~~~~~
    class
```

hkaiser:
There are also additional test failures caused by the proposed changes: https://stellar-group.gitlab.io/-/hpx/-/jobs/159443858/artifacts/tests.unit.html#775cc3d3-3e99-462e-91fa-e374cdeb7c28

hkaiser:
@sithhell, @K-ballo: were the test failures referenced above fixed now?

K-ballo:
> @K-ballo: were the test failures referenced above fixed now?

Yes, by that commit after the reference. A const-qualified component type was sometimes used to do the pinning whereas a mutable component is required, and since pinning is SFINAE-based it would simply ignore the request instead of diagnosing the issue.

PR #3682, opened 2019-02-09
Ignore unitialized chunks. Check proper indices.

### LINKED ISSUE #3442: HPX set_difference, set_intersection failure cases
Hello, 

I am an intern on the Visual C++ Libraries Team this summer and have been working on implementing some of their parallel algorithms.  While working on implementing the set operations in parallel, I was comparing the performance of my implementation to hpx’s.  In doing so, I found some cases in which I got an “Access violation writing location” or “RangeChecks implementation code detected an out of range array access” exception (attempting to run the same executable multiple times resulted in almost always throwing this exception).   I don’t believe this is expected behavior, so Billy O’Neal suggested I email you my findings.  

The benchmarking program I was using can be found attached.  I would comment out one of the algorithms’ calls to test the performance of the other.  The specific cases in which I fairly consistently got an exception (in 5-6 tries running the executable, I got the exception every time) were: 
set_difference
n=5: all cases
n=50: Cases 1-6, 9
set_intersection:
n=50: Case 9
n=500000: Case 9
n=5000000: Case 9
n=50000000: Case 9

There were some cases where running the executable 5-6 times would finish and allow me to collect data for that case, but I’m afraid I didn’t record those (I was initially just interested in gathering benchmarking results, and it was later suggested that I should notify you of these issues).  

Please let me know if you need more information.  

Best,

Miya Natsuhara
mailto:t-minat@microsoft.com

pmikolajczyk41:
Fixes #3442 

## Proposed Changes

  - Unitialized chunks are ignored - runtime errors avoided.
  - Proper indices are being checked while determining whether whole chunk should be handled by next.

### DIFF
--- a/hpx/parallel/algorithms/detail/set_operation.hpp
+++ b/hpx/parallel/algorithms/detail/set_operation.hpp
@@ -107,8 +107,8 @@ namespace hpx { namespace parallel { inline namespace v1 { namespace detail
                 std::size_t end1 = (std::min)(start1 + step, std::size_t(len1));
 
                 if (start1 >= end1)
-        			return;
-                
+                    return;
+
                 bool first_partition = (start1 == 0);
                 bool last_partition = (end1 == std::size_t(len1));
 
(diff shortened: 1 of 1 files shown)


hkaiser:
@pmikolajczyk41 if you need to change a PR after you created it, just push to the same branch, the PR will update. Thanks!

PR #3686, opened 2019-02-11
Make sure tests.unit.build are run after install on CircleCI

msimberg:
This makes sure that build unit tests are run after installing HPX and that they would fail if HPX has not been installed.

Note that this contains a redundant `ninja install` because of the way workspaces are handled on CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -283,6 +283,10 @@ jobs:
       - attach_workspace:
           at: /hpx
       - run:
+          name: Installing HPX for build unit tests
+          command: |
+              ninja install
+      - run:
           name: Building Unit Tests
           command: |
               ninja -j2 -k 0 tests.unit.build
@@ -1002,6 +1006,11 @@ jobs:
           command: |
               ./bin/hello_world --hpx:bind=none
               ninja -j2 install
+          working_directory: /hpx/build
+          when: always
+      - run:
+          name: Testing installed HPX
+          command: |
               hello_world --hpx:bind=none
               ldconfig
               hpxcxx --exe=hello_world_test_build ../source/examples/quickstart/hello_world.cpp -g -lhpx_iostreamsd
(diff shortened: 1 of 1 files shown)


examples/hello_world_component/CMakeLists.txt
```diff
@@ -33,5 +33,5 @@ if(EXISTS "${HPX_DIR}")
     COMPONENT_DEPENDENCIES iostreams
   )
 else()
-  message(WARNING "HPX_DIR=${HPX_DIR} does not exist. Did you forget to run the install rule?")
+  message(FATAL_ERROR "HPX_DIR=${HPX_DIR} does not exist. Did you forget to run the install rule?")
```
sithhell: This should not be a fatal error. I still want to run the unit tests without installing

examples/hello_world_component/Makefile
```diff
@@ -3,6 +3,10 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt
 
+ifeq ($(wildcard $(HPX_DIR)/*),)
+  $(error "HPX_DIR=$(HPX_DIR) does not exist. Did you forget to run the install rule?")
```
sithhell: Same as with the cmake based test...

examples/hello_world_component/CMakeLists.txt
msimberg: Fair point, but then we have to be very careful to make sure that we don't end up back to where it's run before installation and we don't notice errors.

msimberg: I've changed this back to a warning (same in the pkgconfig test).

sithhell: Well, it boils down to how we setup our testers. The rostam testers did install before running the unit tests, for example.

PR #3688, opened 2019-02-11
Decouple serialization from function vtables

K-ballo:
Separating the serialization logic from the function vtables allows:
- making function vtable instances constexpr objects,
- reusing function instantiations for serializable/non-serializable versions,
- various function simplifications (2+1 vtables vs 2*2 vtables),

...at the cost of an extra vtable pointer for the serializable functions.

### DIFF
--- a/hpx/util/detail/basic_function.hpp
+++ b/hpx/util/detail/basic_function.hpp
@@ -305,7 +305,7 @@ namespace hpx { namespace util { namespace detail
         template <typename T>
         static serializable_vtable const* get_serializable_vtable() noexcept
         {
-            return detail::get_vtable<serializable_vtable, T>();
+            return detail::get_serializable_vtable<VTable, T>();
         }
 
     protected:
--- a/hpx/util/detail/basic_function.hpp
+++ b/hpx/util/detail/basic_function.hpp
@@ -14,6 +14,7 @@
 #include <hpx/traits/get_function_address.hpp>
 #include <hpx/traits/get_function_annotation.hpp>
 #include <hpx/traits/is_callable.hpp>
+#include <hpx/util/assert.hpp>
 #include <hpx/util/detail/empty_function.hpp>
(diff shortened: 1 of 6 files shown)


hpx/util/detail/basic_function.hpp
```diff
@@ -200,117 +249,120 @@ namespace hpx { namespace util { namespace detail
...
             if (!is_empty)
             {
-                std::string function_name = this->vptr->name;
-                ar << function_name;
+                std::string name;
+                ar >> name;
+                serializable_vptr = detail::get_serializable_vtable<vtable>(name);
 
-                this->vptr->save_object(this->object, ar, version);
+                vptr = serializable_vptr->vptr;
```
sithhell: Is there a way to avoid the extra storage for `vptr` inside `serializable_vptr`? It seems superfluous to me.

K-ballo: Maybe, I get lost around the function registration logic.. How would it work?

sithhell: maybe by adding a `serializable_vtable::get_vptr` virtual function, which gives us back the needed vtable pointer?

K-ballo: No, the serializable vtable type only knows the function vtable type, not the target callable type. That gets erased in the constructor, where the function is registered. I think the function vtable pointer would have to be stored alongside the serializable vtable pointer in the intrusive registration factory, where it could later be retrieved from by function name, but I find that code hard to navigate.

sithhell: Ok, that might be something for a different PR then.

msimberg:
@K-ballo clang 3 and C++11 do not agree with your changes: http://cdash.cscs.ch//viewBuildError.php?buildid=40884.

msimberg:
This seems to be ready to go, but let's get #3710 in first to make sure that everything is ok.

PR #3691, opened 2019-02-12
Link with libsupc++ if needed for thread_local

msimberg:
Attempt to fix [these](http://rostam.cct.lsu.edu/builders/hpx_clang_3_9_boost_1_62_centos_x86_64_release/builds/214/steps/build_core/logs/stdio) errors occuring with clang < 4.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1243,6 +1243,10 @@ if(NOT WIN32)
     endif()
   endif()
 
+  if(HPX_HAVE_LIBSUPCPP)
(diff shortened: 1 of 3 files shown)


msimberg:
Any objections to this? We could remove this if/once we decide to drop clang 3.

msimberg:
It turns out this only helped for 3.9 but not 3.8. Looking into why is very low priority for me though. We should very strongly consider dropping clang 3 directly after 1.3.0, if not already for 1.3.0.

PR #3692, opened 2019-02-12
Only disable constexpr with clang-cuda, not nvcc+gcc

msimberg:
Trying to fix some of the compilation errors occuring with nvcc+gcc. The combo nvcc++gcc *does* support constexpr so only disable it if using clang-cuda.


hkaiser:
The failing test is most likely unrelated.

PR #3700, opened 2019-02-13
Added barrier::increment function to increase total number of thread

tianyi93:
Added barrier:: increment function to increase the total number of thread
Test added

### DIFF
--- a/hpx/lcos/local/barrier.hpp
+++ b/hpx/lcos/local/barrier.hpp
@@ -46,9 +46,10 @@ namespace hpx { namespace lcos { namespace local
         /// releasing all waiting threads as soon as the last \a thread
         /// entered this function.
         void wait();
+        void count_up();
 
     private:
-        std::size_t const number_of_threads_;
+        std::size_t number_of_threads_;
         std::size_t total_;
 
         mutable mutex_type mtx_;
--- a/src/lcos/local/local_barrier.cpp
+++ b/src/lcos/local/local_barrier.cpp
@@ -67,5 +67,11 @@ namespace hpx { namespace lcos { namespace local
             }
         }
     }
+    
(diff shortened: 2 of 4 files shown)


hpx/lcos/local/barrier.hpp
```diff
@@ -46,9 +46,10 @@ namespace hpx { namespace lcos { namespace local
         /// releasing all waiting threads as soon as the last \a thread
         /// entered this function.
         void wait();
+        void increase();
```
hkaiser: Let's align the name of this function with previous art. The closest would probably `std::latch::count_down()`. I'd suggest to rename this to `barrier::count_up()`. Would that work for everybody?

src/lcos/local/local_barrier.cpp
```diff
@@ -67,5 +67,9 @@ namespace hpx { namespace lcos { namespace local
             }
         }
     }
+    void barrier::increase(){
+        std::unique_lock<mutex_type> l(mtx_);
+        number_of_threads_+=1;
+    }
```
hkaiser: Please format your code to match the formatting of the surrounding code.

src/lcos/local/local_barrier.cpp
```diff
@@ -67,5 +67,9 @@ namespace hpx { namespace lcos { namespace local
             }
         }
     }
+    void barrier::increase(){
+        std::unique_lock<mutex_type> l(mtx_);
+        number_of_threads_+=1;
```
hkaiser: `++number_of_threads_;` would probably be more readable.

tianyi93:
@hkaiser I did those changes based on the comments, it seems that still some checks are failing, please let me know if there is anything I should improve, thanks

PR #3701, opened 2019-02-14
MoodyCamel lockfree

biddisco:
This PR replaces some of our usage of boost::lockfree fifo queue with an alternative lockfree queue from MoodyCamel. Please see the commit log for extra info.

Improvements come in two places. 1) we get better performance with our task queues as this queue outperforms the boost one. Particularly on higher thread count numbers. 2) The removal of an alloc/dealloc in the task description queue reduces work in jemalloc/other allocators which can hit posix locks when they are maxxed out (particularly noticable in the future_overhead test for example).

I would like to have approval from @cameron314 that the inclusion of the boost license in the concurrentqueue.h file is in accordance with his terms and conditions.

### DIFF
--- a/hpx/runtime/threads/policies/lockfree_queue_backends.hpp
+++ b/hpx/runtime/threads/policies/lockfree_queue_backends.hpp
@@ -17,8 +17,12 @@
 #include <boost/lockfree/queue.hpp>
 #endif
 
+// Does not rely on CXX11_STD_ATOMIC_128BIT
+#include <hpx/util/lockfree/concurrentqueue.hpp>
+
 #include <cstddef>
 #include <cstdint>
+#include <utility>
 
 namespace hpx { namespace threads { namespace policies
 {
@@ -73,6 +77,57 @@ struct lockfree_fifo
 };
 
 ///////////////////////////////////////////////////////////////////////////////
+// MoodyCamel FIFO
+template <typename T>
+struct moodycamel_fifo_backend
+{
+    typedef moodycamel::ConcurrentQueue<T> container_type;
+    typedef T value_type;
+    typedef T& reference;
+    typedef T const& const_reference;
+    typedef T && rval_reference;
+    typedef std::uint64_t size_type;
+
+    moodycamel_fifo_backend(
+        size_type initial_size = 0
+      , size_type num_thread = size_type(-1)
+        )
+      : queue_(std::size_t(initial_size))
+    {}
+
+    bool push(rval_reference val, bool /*other_end*/ = false)
+    {
+        return queue_.enqueue(std::move(val));
+    }
+
+    bool push(const_reference val, bool /*other_end*/ = false)
+    {
+        return queue_.enqueue(val);
+    }
+
+    bool pop(reference val, bool steal = true)
+    {
+        return queue_.try_dequeue(val);
+    }
+
+    bool empty()
+    {
+        return (queue_.size_approx()==0);
+    }
+
+  private:
+    container_type queue_;
+};
+
+struct concurrentqueue_fifo
+{
+    template <typename T>
+    struct apply
+    {
+        typedef moodycamel_fifo_backend<T> type;
+    };
+};
+
 // LIFO
 #if defined(HPX_HAVE_CXX11_STD_ATOMIC_128BIT)
 struct lockfree_lifo;
--- a/hpx/runtime/threads/policies/thread_queue_mc.hpp
+++ b/hpx/runtime/threads/policies/thread_queue_mc.hpp
@@ -1,0 +1,1153 @@
+//  Copyright (c) 2007-2019 Hartmut Kaiser
+//  Copyright (c) 2011      Bryce Lelbach
+//
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#if !defined(HPX_THREADMANAGER_THREAD_QUEUE_MC)
+#define HPX_THREADMANAGER_THREAD_QUEUE_MC
+
+#include <hpx/config.hpp>
+#include <hpx/assertion.hpp>
+#include <hpx/error_code.hpp>
+#include <hpx/runtime/config_entry.hpp>
+#include <hpx/runtime/threads/policies/lockfree_queue_backends.hpp>
+#include <hpx/runtime/threads/policies/queue_helpers.hpp>
+#include <hpx/runtime/threads/thread_data.hpp>
+#include <hpx/runtime/threads/policies/thread_queue.hpp>
+#include <hpx/throw_exception.hpp>
+#include <hpx/allocator_support/internal_allocator.hpp>
+#include <hpx/util/cache_aligned_data.hpp>
+#include <hpx/util/function.hpp>
+#include <hpx/util/get_and_reset_value.hpp>
+#include <hpx/timing/high_resolution_clock.hpp>
+#include <hpx/thread_support/unlock_guard.hpp>
+#include <hpx/datastructures/tuple.hpp>
+
+#ifdef HPX_HAVE_THREAD_CREATION_AND_CLEANUP_RATES
+#   include <hpx/util/tick_counter.hpp>
+#endif
+
+#include <boost/lexical_cast.hpp>
+
+#include <atomic>
+#include <cstddef>
+#include <cstdint>
+#include <exception>
+#include <functional>
+#include <list>
+#include <map>
+#include <memory>
+#include <mutex>
+#include <string>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
+///////////////////////////////////////////////////////////////////////////////
+namespace hpx { namespace threads { namespace policies
+{
+#ifdef HPX_HAVE_THREAD_QUEUE_WAITTIME
+    ///////////////////////////////////////////////////////////////////////////
+    // We control whether to collect queue wait times using this global bool.
+    // It will be set by any of the related performance counters. Once set it
+    // stays set, thus no race conditions will occur.
+    extern HPX_EXPORT bool maintain_queue_wait_times;
+#endif
+#ifdef HPX_HAVE_THREAD_MINIMAL_DEADLOCK_DETECTION
+    ///////////////////////////////////////////////////////////////////////////
+    // We globally control whether to do minimal deadlock detection using this
+    // global bool variable. It will be set once by the runtime configuration
+    // startup code
+    extern bool minimal_deadlock_detection;
+#endif
+
+    ///////////////////////////////////////////////////////////////////////////
+    // // Queue back-end interface:
+    //
+    // template <typename T>
+    // struct queue_backend
+    // {
+    //     typedef ... container_type;
+    //     typedef ... value_type;
+    //     typedef ... reference;
+    //     typedef ... const_reference;
+    //     typedef ... size_type;
+    //
+    //     queue_backend(
+    //         size_type initial_size = ...
+    //       , size_type num_thread = ...
+    //         );
+    //
+    //     bool push(const_reference val);
+    //
+    //     bool pop(reference val, bool steal = true);
+    //
+    //     bool empty();
+    // };
+    //
+    // struct queue_policy
+    // {
+    //     template <typename T>
+    //     struct apply
+    //     {
+    //         typedef ... type;
+    //     };
+    // };
+    template <typename Mutex = std::mutex,
+        typename PendingQueuing = lockfree_lifo,
+        typename StagedQueuing = lockfree_lifo,
+        typename TerminatedQueuing = lockfree_fifo>
+    class thread_queue_mc
+    {
+    private:
+        // we use a simple mutex to protect the data members for now
+        typedef Mutex mutex_type;
+
+        // don't steal if less than this amount of tasks are left
+        int const min_tasks_to_steal_pending;
+        int const min_tasks_to_steal_staged;
+
+        // create at least this amount of threads from tasks
+        int const min_add_new_count;
+
+        // create not more than this amount of threads from tasks
+        int const max_add_new_count;
+
+        // number of terminated threads to discard
+        int const max_delete_count;
+
+        // number of terminated threads to collect before cleaning them up
+        int const max_terminated_threads;
+
+        // this is the type of a map holding all threads (except depleted ones)
+        using thread_map_type = std::unordered_set<thread_id_type,
+            std::hash<thread_id_type>, std::equal_to<thread_id_type>,
+            util::internal_allocator<thread_id_type>>;
+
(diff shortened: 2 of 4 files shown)


hpx/runtime/threads/policies/lockfree_queue_backends.hpp
```diff
@@ -11,6 +11,7 @@
 #include <hpx/config.hpp>
 
 #include <hpx/util/lockfree/deque.hpp>
+#include <hpx/concurrent/concurrentqueue.h>
```
msimberg: Why not `hpx/util/lockfree/concurrentqueue.h` like the deque?

hpx/runtime/threads/policies/thread_queue.hpp
```diff
@@ -196,8 +196,8 @@ namespace hpx { namespace threads { namespace policies
         typedef typename PendingQueuing::template
             apply<thread_description*>::type work_items_type;
 
-        typedef typename StagedQueuing::template
-            apply<task_description*>::type task_items_type;
+        typedef lockfree_fifo::
```
msimberg: Either `StagedQueueing` should go from the template, or the other queues should still be supported?

hpx/concurrent/concurrentqueue.h
```diff
@@ -0,0 +1,3648 @@
...
+                    continue;
+                }
+
+                // Good, reference count has been incremented (it wasn't at zero), which means we can read the
+                // next and not worry about it changing between now and the time we do the CAS
+                auto next = head->freeListNext.load(std::memory_order_relaxed);
+                if (freeListHead.compare_exchange_strong(head, next, std::memory_order_acquire, std::memory_order_relaxed)) {
+                    // Yay, got the node. This means it was on the list, which means shouldBeOnFreeList must be false no
+                    // matter the refcount (because nobody else knows it's been taken off yet, it can't have been put back on).
+                    assert((head->freeListRefs.
```
msimberg: If we don't want to support the other queue backends for terminated items, this one could be useful for `cleanup_terminated`.

hpx/runtime/threads/policies/lockfree_queue_backends.hpp
biddisco: When we were talking about integrating libcds with hpx, we had the idea of making hpx/concurrent as a new place for concurrent stuff. But I'm quite happy to put it into `util/lockfree/` since its a single header

msimberg: Or move deque into concurrent?

hpx/runtime/threads/policies/thread_queue.hpp
biddisco: StagedQueuing is still used for the thread_alloc_ template.
Not sure if it's actually needed though.

hpx/concurrent/concurrentqueue.h
biddisco: I thought I'd start by changing the drop in replacments from fifo to new queue. The lifo stuff I left for now as I still have benchmarking I'm doing across schedulers that might help us decide on whether we want to remove other stuff.

hpx/runtime/threads/policies/lockfree_queue_backends.hpp
biddisco: Either works for me

hpx/runtime/threads/thread_init_data.hpp
```diff
@@ -50,6 +50,31 @@ namespace hpx { namespace threads
...
+            parent_phase = rhs.parent_phase;
+#endif
+#ifdef HPX_HAVE_APEX
+        // HPX_HAVE_APEX forces the HPX_HAVE_THREAD_DESCRIPTION
+        // and HPX_HAVE_THREAD_PARENT_REFERENCE settings to be on
+            apex_data = apex_new_task(description, parent_locality_id, parent_id );
+#endif
+            return *this;
+        }
+
```
hkaiser: Good catch!

hkaiser:
@biddisco Have you thought about using the advanced stuff in the ModyCamel queue (https://github.com/cameron314/concurrentqueue#advanced-features) to speed up stealing? We know what thread we will steal from, so we could use the producer's token. Not sure how much this would help, though.

biddisco:
> @biddisco Have you thought about using the advanced stuff in the ModyCamel queue (https://github.com/cameron314/concurrentqueue#advanced-features) to speed up stealing? We know what thread we will steal from, so we could use the producer's token. Not sure how much this would help, though.

Yes. I will look at using producer and consumer tokens. Adding them should be straightforward as we can assign them to each thread/queue at startup. I want to get a basic working version in first and once we are happy that nothing is broken, then make more improvements.

biddisco:
What we really need is a dequeue version of the queue so that we can support lifo/fifo operation and steal from the stale end and keep hot items on the same thread.

cameron314:
Queue author here, license and comments inserted at top of concurrentqueue.h are fine :-)

hpx/runtime/threads/policies/thread_queue.hpp
msimberg: It's not actually used, so it could go.

hpx/concurrent/concurrentqueue.h
msimberg: Sure. I'm only specifically talking about `terminated_items_` since other queue backends can't be used there anymore.

biddisco:
> 
> 
> Queue author here, license and comments inserted at top of concurrentqueue.h are fine :-)

Thank you very much. I will also look though our contributors section and add a thanks to you in there somewhere. (assuming this PR gets merged)

Since I have your attention - would it be a lot of work to turn the queue into a dequeue so we can pull from either end? As noted in an earlier comment, when tasks are stolen from one thread queue to another, stealing from the cold end is preferred, but on the owning thread, stealing from the hot end can improve cache reuse. Is this something that might one day be supported?

cameron314:
Sorry, turning the queue into a dequeue is pretty much impossible. It would need to be completely redesigned.

Do you need the fifo-ness of a queue? This queue provides fifo order only per-producer, with no global order. If you don't need order at all, the internal free-list data structure is probably closer to what you want, since it pops out hot items where possible.

biddisco:
> Do you need the fifo-ness of a queue? This queue provides fifo order only per-producer, with no global order. If you don't need order at all, the internal free-list data structure is probably closer to what you want, since it pops out hot items where possible.

Thanks for he information. The fact that the queue isn't linearizable might cause some unexpected behaviour in terms of task ordering, but it should not be a deal breaker for us since each core is generally maintaining its own queue and stealing takes place when the queues drain on one core. It is mostly the ability to steal from the cold end (of another thread's queue) rather than the hot one that is of interest, (but I don't have much in the way of numbers to back up my claim anyway).

hpx/runtime/threads/policies/thread_queue.hpp
biddisco: I'll clean it up some more then.

Do @hkaiser or @sithhell want to comment on whether we should move lockfree stuff into concurrent, or just move the concurrentqueue into util/lockfree for now. (seems like putting it into util/lockfree for the time being is probably the simplest option, though I still like the idea of having a concurrent top level dir for this stuff)

hpx/concurrent/concurrentqueue.h
biddisco: ok. I did change `typename TerminatedQueuing = lockfree_fifo` in the shared priority scheduler to use it but I didn't realize that it was the only one we should use. the other schedulers are using lifo at the moment and I didn't touch them. I'll look again at the code.

hpx/runtime/threads/policies/lockfree_queue_backends.hpp
sithhell: I don't mind. The 'concurrent/concurrentqueue.h' has lots of redundant information. Also, please use the `.hpp` file ending to stay consinstent with the rest.

sithhell:
I think non ordering of tasks in the queues is totally fine. After all, it looks like we have, and only require, weak consistency here, which is IMHO a good thing as it naturally reflects the actual scheduling of "stuff" in the hardware, I actually like that "hot" items are preferred. I think this is also where the speed comes from?

hpx/runtime/threads/policies/lockfree_queue_backends.hpp
biddisco: I've renamed the file to `.hpp` extension and moved it to hpx/util/lockfree/ - force pushed with the commits edited accordingly.

msimberg:
This looks good to me, but there are a lot of timeouts on CircleCI. It looks like you're up to date with master, so you have the latest bugfixes. It would be good to figure out if they're the fault of this PR (I don't spot anything odd), or if the new queue is triggering more race conditions because of changed timings.

I also want to make sure everyone is aware and on-board with hard-coding the staged queues to the new queue backend (if we want to keep support for the others I guess it would be doable with various helpers and wrappers). It looks like in this case the implementation matters more than any theoretical benefits we'd get from LIFO or ABP LIFO, but all we have so far is results from the future overhead test.

biddisco:
I'm fixing inspect failures for the PR and the clang-tidy reported problems, but I have a question.
Inspect does not like the many long lines in the new queue. I would like to make the queue exempt from our inspect checks because there is a strong possibility that we will want to check for new versions reasonably regularly and update the `concurrentqueue.hpp` file without having to make changes every time.
1) does anyone object to making the queue inspect exempt
2) if we are ok to do that, is there a file somewhere I can add the queue to to make it exempt (I think there is but I'm too lazy to look and am hoping someone can just tell me).

biddisco:
Force-pushed this PR again : I've fixed inspect errors, added `// NOLINT` to the offending clang-tidy errors in ConcurrentQueue and opened an issue on the original repo here https://github.com/cameron314/concurrentqueue/issues/149 just to make sure the clang-tidy problems are noted.

msimberg:
>     2. if we are ok to do that, is there a file somewhere I can add the queue to to make it exempt (I think there is but I'm too lazy to look and am hoping someone can just tell me).
I think adding `hpx-no-inspect` to the top of the file should do it: https://github.com/STEllAR-GROUP/hpx/blob/cde5fd7adfaae9570ad6d3139e8f8a2e8ac057fd/tools/inspect/inspect.cpp#L21-L22.

hpx/runtime/threads/policies/thread_queue.hpp
hkaiser: I'm impartial to this. Having it in util/lockfree is fine, I guess.

msimberg:
> Sorry, turning the queue into a dequeue is pretty much impossible. It would need to be completely redesigned.
@cameron314 sorry to bug you more, but what about LIFO (i.e. stack)?

cameron314:
@msimberg Nope, same. Like most lock-free data structures, it's pretty difficult to turn into anything it's not already :-)

msimberg:
@biddisco started a couple of builds on rostam:
- http://rostam.cct.lsu.edu/builders/hpx_clang_6_boost_1_68_centos_x86_64_debug/builds/139
- http://rostam.cct.lsu.edu/builders/hpx_clang_6_boost_1_68_centos_x86_64_release/builds/143

Doesn't tell us what's wrong, but these tests seem to fail regularly:
- tests.unit.parallel.executors.thread_pool_executors
- tests.unit.parallel.executors.timed_thread_pool_executors
- tests.unit.threads.resource_manager
- tests.regressions.components_dir.distributed.tcp.bulk_new_3054

biddisco:
I have modified this PR to include the MoodyCamel concurrentqueue as an additional lockfree backend and not replace the existing lockfree fifo. Then switched the shared_priority scheduler aloone to use it in one place. This allows us to incrementally add support/use of this backend without changing other schedulers and breaking the failing executor tests that caused problems with the previous version of the PR

biddisco:
I've cleaned this up to fix all unexpected header test fails etc and it should be mergeable.
There are occasional dashboard fails, but they are most likely caused by other random problems with builds etc. This PR should not touch any existing tests unless they explicitly use the shared-priority scheduler.

hpx/runtime/threads/policies/lockfree_queue_backends.hpp
```diff
@@ -17,8 +17,12 @@
 #include <boost/lockfree/queue.hpp>
 #endif
 
+// Does not rely on CXX11_STD_ATOMIC_128BIT
```
hkaiser: Is this comment a left-over or is it still true? If yes, please guard all of the code using 128Bit atomics (using `HPX_HAVE_CXX11_STD_ATOMIC_128BIT`).

hkaiser: Sorry, I misread the comment, please disregard.

biddisco: if the moody camel code used 128 bit atomics internally, would there be a simple way of checking (ie compiling with a certain compiler flag and triggering an error?)

cameron314: The `moodycamel::ConcurrentQueue` uses atomics of the `size_t` and `index_t` types from its traits, which are `std::size_t` by default. There's no use case for 128-bit atomics in the queue.

hkaiser: @cameron314 thanks for clarifying!

hpx/runtime/threads/policies/thread_queue_mc.hpp
```diff
@@ -0,0 +1,1153 @@
...
+    // {
+    //     template <typename T>
+    //     struct apply
+    //     {
+    //         typedef ... type;
+    //     };
+    // };
+    template <typename Mutex = std::mutex,
+        typename PendingQueuing = lockfree_lifo,
+        typename StagedQueuing = lockfree_lifo,
```
msimberg: This template parameter needs to go (just like in the template for `shared_priority_queue_scheduler`)

msimberg:
It looks like this is pretty much ready to go in. The only thing I don't particularly like is that most of `thread_queue_mc` just duplicates what's in `thread_queue`. Would inheriting from `thread_queue` be an option (assuming it doesn't add any overhead)?

biddisco:
I have another branch ready for a PR that builds upon this work. In the next PR, the shared_priority_queue is cleaned up (removing template params as above) and the thread_queue_mc is substantially redesigned. I put this thread_queue_mc in because it forms a nice starting point for the next layer of work. It's almost a copy/paste in this version but diverges significantly in the next version.

hpx/runtime/threads/policies/thread_queue_mc.hpp
msimberg: What I mean here is that `shared_priority_queue_scheduler` passes its `TerminatedQueuing` template parameter as `StagedQueuing` to `thread_queue_mc`.

PR #3704, opened 2019-02-17
Reduce use of binders

K-ballo:
Replace several uses of binders with simpler lighter alternatives.

### DIFF
--- a/hpx/lcos/dataflow.hpp
+++ b/hpx/lcos/dataflow.hpp
@@ -193,8 +193,10 @@ namespace hpx { namespace lcos { namespace detail
             // simply schedule new thread
             parallel::execution::parallel_policy_executor<launch::async_policy>
                 exec{policy};
-            parallel::execution::post(exec, &dataflow_frame::done,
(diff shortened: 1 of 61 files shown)


hpx/lcos/dataflow.hpp
```diff
@@ -193,8 +193,10 @@ namespace hpx { namespace lcos { namespace detail
             // simply schedule new thread
             parallel::execution::parallel_policy_executor<launch::async_policy>
                 exec{policy};
-            parallel::execution::post(exec, &dataflow_frame::done,
-                std::move(this_), std::move(futures));
+            parallel::execution::post(exec,
+                [HPX_CAPTURE_MOVE(this_)](Futures&& futures) mutable -> void {
+                    return this_->done(std::move(futures));
+                }, std::move(futures));
```
hkaiser: What's the rationale of this change?

K-ballo: Lifting the mem-ptr value from runtime into compile time, yielding a smaller, simpler, and easier to inline callable.

PR #3705, opened 2019-02-17
Fix and simplify make_ready_future overload sets

K-ballo:


hkaiser:
Should we use the new (global) `init_no_addref` for `id_type` as well (see https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/runtime/naming/name.hpp#L806)?

K-ballo:
> Should we use the new (global) `init_no_addref` for `id_type` as well (see https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/runtime/naming/name.hpp#L806)?

Possibly yes, but then it would have to graduate from `lcos::detail` to `util` or something like that.

hkaiser:
>> Should we use the new (global) `init_no_addref` for `id_type` as well (see https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/runtime/naming/name.hpp#L806)?
>
> Possibly yes, but then it would have to graduate from `lcos::detail` to `util` or something like that.

@K-ballo Let's leave that for a separate PR. 
@msimberg this is good to go.

K-ballo:
Removed `init_no_addref` changes from this PR.

hkaiser:
Let's go ahead and get this in.

PR #3710, opened 2019-02-18
Fix the C++11 build, after #3704

K-ballo:


msimberg:
Thanks @K-ballo for doing this. Looks good, but let's wait for the builds to come in. This should have priority over other PRs, since we're currently blind to failures.

PR #3711, opened 2019-02-19
Ffwd scheduler

margabue:
This scheduler implements the fast, fly-weight delegation algorithm, that was shown to have a much higher performance in certain situations. The idea is derived form the article at https://blog.acolyer.org/2017/12/04/ffwd-delegation-is-much-faster-than-you-think/ that comprehensively sums up the key aspects of the mechanism. The original implementation was adapted to work without a dedicated hardware thread and instead utilizes message passing between the threads, in order to fit with hpx's design.

## Proposed Changes
Adding an additional scheduler, that can be chosen by the user.

Scheduler based on the fast, fly-weight algorithm that was implemented during my bachelor thesis under supervision of @sithhell.


msimberg:
@margabue this (and the link you provided) looks very interesting. Thanks! The new scheduler seems to be using a single shared queue for all worker threads (with the `thread_queue` class unchanged). Is there something missing or is that exactly the point of the new scheduler? Would you mind summarizing the key point of ffwd (delegation, presumably?) and how this implementation differs from the one in the article? Do you happen to have any performance numbers comparing this to our existing schedulers?

It looks like you may have been a bit overeager in deleting files. For example the `apex` folder and `.gitignore` should stay, and some `CMakeLists.txt` are missing.

hkaiser:
@margabue Thanks for working on this! This may add a valuable data-point and functionality to our scheduling system. I'm eagerly looking forward to seeing some performance numbers.

Additionally, may I suggest you merge all of your commits into one single commit in the end? That might help understanding your submission.

msimberg:
@margabue @sithhell what is the status of this? Is this likely to be picked up again? Let us know if there's anything we can do to help.

msimberg:
@margabue (@sithhell) I'm closing this for now as it's been inactive for a while, but please let us know if we can help in any way. It would still be nice to have this in at some point.

PR #3716, opened 2019-02-24
Add barrier reset function, requested by hpxMP for reusing barrier

tianyi93:
Add barrier reset function, requested by hpxMP for reusing barrier

### DIFF
--- a/hpx/lcos/local/barrier.hpp
+++ b/hpx/lcos/local/barrier.hpp
@@ -46,7 +46,12 @@ namespace hpx { namespace lcos { namespace local
         /// releasing all waiting threads as soon as the last \a thread
         /// entered this function.
         void wait();
+        /// The function \a count_up will increase the number of \a threads
+        /// to be waited in \a wait function.
         void count_up();
+        /// The function \a reset will reset the number of \a threads
+        /// as given by the function parameter \a number_of_threads.
+        void reset(std::size_t number_of_threads);
 
     private:
         std::size_t number_of_threads_;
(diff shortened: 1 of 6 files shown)


tianyi93:
@hkaiser 
The ```local_barrier_count_up_reset``` test is failing if I add
```
total_ = barrier_flag;
```
under the ```reset``` function.

It seems that the ```total_```has already been set to ```barrier_flag``` when it exit the ```wait ```function,
I am not sure why this behavior happens when I set ```total_ ```to ```barrier_flag``` again in ```reset ```function.

tianyi93:
@hkaiser 
```
            // get entering threads to wake up
            if (total_ == barrier_flag)
            {
                cond_.notify_all(std::move(l));
            }
```
It seems that this is waking up all the waiting threads if I set ```total_```to ```barrier_flag``` in the reset function. Some previous threads might still haven't reached this line when new thread reached cond_.wait on the reseted barrier.

If I understand correctly, the following lines has already gracefully makes the newer coming threads wait untill total_ is equal to barrier_flag 
```
while (total_ > barrier_flag)
        {
            // wait until everyone exits the barrier
            cond_.wait(l, "barrier::wait");
        }
```
```
if (total_ == barrier_flag)
            {
                cond_.notify_all(std::move(l));
            }
```

hkaiser:
You might want to add a `HPX_ASSERT(total_ == barrier_flag);` to `reset()` (or similar), then. The precondition for `reset()` should be that there are no pending barrier operations in flight.

PR #3727, opened 2019-03-03
Adding example demonstrating how to customize the memory management for a component

hkaiser:
- adding traits::component_heap_type
- flyby: adding `local_new(launch::sync, ...)
- flyby: don't bind gids that hold lva


McKillroy:
It would be great if someone could review and approve this - I really want to use the possibility of having a custom allocator for certain types. Thanks a ton @hkaiser for bringing this in and so quick!

PR #3734, opened 2019-03-06
Limiting executor

biddisco:
This adds an executor that counts tasks and can be used to limit the number that go into the queues at any given time - and also to wait on them so that it is not necessary to wait on a vector of futures. It is similar to the aggregating executor and the recent changes to parallel algorithms to use a latch.

This executor can be used to replace sliding_semaphore and provides a much simpler and cleaner way of doing it.

### DIFF
--- a/hpx/runtime/threads/executors/limiting_executor.hpp
+++ b/hpx/runtime/threads/executors/limiting_executor.hpp
@@ -6,20 +6,17 @@
 #ifndef HPX_RUNTIME_THREADS_LIMITING_EXECUTOR_HPP
 #define HPX_RUNTIME_THREADS_LIMITING_EXECUTOR_HPP
 
+#include <hpx/config.hpp>
+#include <hpx/apply.hpp>
 #include <hpx/parallel/executors/execution_fwd.hpp>
 #include <hpx/runtime/threads/executors/default_executor.hpp>
-#include <hpx/runtime/threads/thread_executor.hpp>
 #include <hpx/runtime/threads/executors/pool_executor.hpp>
+#include <hpx/runtime/threads/thread_executor.hpp>
 #include <hpx/runtime/threads/thread_pool_base.hpp>
-#include <hpx/apply.hpp>
-#include <hpx/util/thread_description.hpp>
-#include <hpx/util/thread_specific_ptr.hpp>
-#include <hpx/util/deferred_call.hpp>
-#include <hpx/lcos/dataflow.hpp>
 #include <hpx/util/invoke.hpp>
-#include <hpx/util/pack_traversal.hpp>
 #include <hpx/util/yield_while.hpp>
 
+#include <atomic>
 #include <cstddef>
 #include <cstdint>
 #include <string>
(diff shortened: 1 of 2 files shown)


biddisco:
just noticed that I committed this with a bunch of commented out bits and debug comments in place. also forgot to document it. I'll add that tomorrow.

tests/performance/local/future_overhead.cpp
```diff
@@ -52,10 +83,17 @@ std::uint64_t num_iterations = 0;
...
+        const int array_size = 4096;
+        std::array<double, array_size> dummy;
+        for (std::uint64_t i = 0; i < num_iterations; ++i) {
+            for (std::uint64_t j = 0; j < array_size; ++j) {
+                dummy[j] = 1.0 / (2.0 * i * j + 1.0);
+            }
+        }
+        return dummy[0];
+    }
+    return 0.0;
```
msimberg: Did this change just to have some more work?

hpx/runtime/threads/executors/limiting_executor.hpp
```diff
@@ -0,0 +1,218 @@
...
+#include <hpx/parallel/executors/execution_fwd.hpp>
+#include <hpx/runtime/threads/executors/default_executor.hpp>
+#include <hpx/runtime/threads/thread_executor.hpp>
+#include <hpx/runtime/threads/executors/pool_executor.hpp>
+#include <hpx/runtime/threads/thread_pool_base.hpp>
+#include <hpx/apply.hpp>
+#include <hpx/util/thread_description.hpp>
+#include <hpx/util/thread_specific_ptr.hpp>
+#include <hpx/util/deferred_call.hpp>
+#include <hpx/lcos/dataflow.hpp>
```
msimberg: This looks like it can be removed, probably others too.

hpx/runtime/threads/executors/limiting_executor.hpp
```diff
@@ -0,0 +1,218 @@
...
+        future<typename util::invoke_result<F, Ts...>::type>
+        async_execute(F && f, Ts &&... ts)
+        {
+            typedef typename util::detail::invoke_deferred_result<
+                    F, Ts...>::type result_type;
+
+            count_up();
+            lcos::local::futures_factory<result_type()> p(
+                executor_,
+                [this, f{std::move(f)},
```
msimberg: You can use `HPX_CAPTURE_MOVE` here.

tests/performance/local/future_overhead.cpp
biddisco: yes. When the num_iterations>0 the work done in the loop is much more substantial. The reason for this is to allow testing of memory consumption. If many tasks have stacks attached on creation (they should not) then when thousands are siting in waiting queues, the memory explodes. This is just to allow me to test not just task overheads, but also memory overheads and catch problems when testing new lazy stack init type PRs

hkaiser:
@msimberg, @biddisco: I addressed all review comments, let's get this merged asap.

biddisco:
Thanks

PR #3736, opened 2019-03-08
Fix guided_pool_executor after dataflow changes caused compilation fail

biddisco:
Fixes current problems on dashboard caused by #3704


hpx/runtime/threads/executors/guided_pool_executor.hpp
```diff
@@ -17,6 +17,7 @@
 #include <hpx/util/pack_traversal.hpp>
 #include <hpx/util/thread_description.hpp>
 #include <hpx/util/thread_specific_ptr.hpp>
+#include <hpx/traits/is_future_tuple.hpp>
 
 #include <cstddef>
 #include <cstdint>
```
hkaiser: Please `#include <iostream>` conditionally

msimberg:
@biddisco https://github.com/STEllAR-GROUP/hpx/blob/71f3d4def724d497eab8f7f26d233d633fa52786/examples/resource_partitioner/CMakeLists.txt#L13-L15 is the check I mentioned. It looks like it should only exclude gcc < 5, instead of only including gcc >= 5.

hpx/runtime/threads/executors/guided_pool_executor.hpp
```diff
@@ -435,10 +423,14 @@ namespace hpx { namespace threads { namespace executors
...
+#ifdef GUIDED_POOL_EXECUTOR_FAKE_NOOP
+                         int>(" | ")
+#else
                          decltype(unwrapped_futures_tuple)>(" | ")
+#endif
                       << "\n"
-                      << "dataflow-frame: Result      : "
-                      << util::debug::print_type<Result>() << "\n";
+                      /*<< "dataflow-frame: Result      : "
+                      << util::debug::print_type<Result>() */<< "\n";
```
msimberg: Can the commented-out code be removed?

hkaiser:
@msimberg, @biddisco I applied the requested changes, I think this should be merged asap as it fixes a couple of problems that annoy us to no end with other PRs.

PR #3739, opened 2019-03-09
More fixes to make sure that std::flush really flushes all output

hkaiser:
This is in addition to #3728 as we have missed some spots there.

### DIFF
--- a/examples/quickstart/potpourri.cpp
+++ b/examples/quickstart/potpourri.cpp
@@ -17,6 +17,13 @@ void final_task(hpx::future<hpx::util::tuple<hpx::future<double>, hpx::future<vo
     hpx::cout << "in final_task" << hpx::endl;
 }
 
+// Avoid ABI incompatibilities between C++11/C++17 as std::rand has exception
+// specification in libstdc++.
+int rand_wrapper()
+{
+    return std::rand();
+}
+
 int main(int, char**)
 {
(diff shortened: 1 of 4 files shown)


examples/quickstart/potpourri.cpp
```diff
@@ -17,6 +17,13 @@ void final_task(hpx::future<hpx::util::tuple<hpx::future<double>, hpx::future<vo
     hpx::cout << "in final_task" << hpx::endl;
 }
 
+// Avoid ABI incompatibilities between C++11/C++17 as std::rand has exception
+// specification in libstdc++.
+int rand()
+{
+    return std::rand();
```
sithhell: This seems to not work out nicely. Clang with libstdc++ gives me this:
```
/builds/stellar-group/hpx/source/examples/quickstart/potpourri.cpp:23:1: warning: all paths through this function will call itself [-Winfinite-recursion]
```

msimberg: There are quite a few other uses of `generate(..., std::rand)` elsewhere, which might be problematic in the same way(?). Would it make sense to fix those in a separate PR?

hkaiser: @sithhell yah, that's clang being over-eager in reporting. I changed the code for this to go away, however.

Issue #3744, opened 2019-03-15
Separate spinlock from same cacheline as internal data for all LCOs


hkaiser:
Most of our LCOs (`local::counting_semaphore`, `local::barrier`, `local::latch`, etc.) place the spinlock and the data it is protecting on the same cache line. In case of higher contention this causes unnecessary false sharing as accessing the spinlock from a different core moves the associated data to that core as well.

We should introduce proper padding for the local LCOs to minimize this effect.

PR #3745, opened 2019-03-17
Introduced cache_aligned_data and cache_line_data helper structure

hkaiser:
- reduce false sharing in local priority scheduler and thread_queue 
- reduce false sharing in latch and condition_variable
- disabled thread stealing counters by default (to reduce overheads)

The changes result in an overall performance improvement by 30% (for_each benchmark).

Working towards #3744

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1439,6 +1439,12 @@ if(HPX_WITH_COMPILER_WARNINGS)
     hpx_add_compile_flag_if_available(-Wformat=2)
     hpx_add_compile_flag_if_available(-Wno-format-nonliteral)
 
+    # enable C++17 style alignment on allocations for pre=C++17
+    hpx_add_compile_flag_if_available(-faligned-new)
+
+    # prevent warnings about over-aligned types for pre-C++17
+    hpx_add_compile_flag_if_available(-Wno-aligned-new)
+
     # Self initialization is dangerous
     hpx_add_compile_flag_if_available(-Winit-self)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -519,8 +519,8 @@ hpx_option(HPX_WITH_THREAD_STACK_MMAP BOOL
   CATEGORY "Thread Manager" ADVANCED)
 
 hpx_option(HPX_WITH_THREAD_MANAGER_IDLE_BACKOFF BOOL
-  "HPX scheduler threads do exponential backoff on idle queues (default: ON)"
-  ON
+  "HPX scheduler threads do exponential backoff on idle queues (default: OFF)"
+  OFF
   CATEGORY "Thread Manager" ADVANCED)
 
 hpx_option(HPX_WITH_STACKTRACES BOOL
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -600,8 +600,8 @@ if(HPX_WITH_THREAD_CUMULATIVE_COUNTS)
 endif()
 
 hpx_option(HPX_WITH_THREAD_STEALING_COUNTS BOOL
-  "Enable keeping track of counts of thread stealing incidents in the schedulers (default: ON)"
-  ON CATEGORY "Thread Manager" ADVANCED)
+  "Enable keeping track of counts of thread stealing incidents in the schedulers (default: OFF)"
+  OFF CATEGORY "Thread Manager" ADVANCED)
 
 if(HPX_WITH_THREAD_STEALING_COUNTS)
   hpx_add_config_define(HPX_HAVE_THREAD_STEALING_COUNTS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1383,13 +1383,17 @@ if(HPX_WITH_COMPILER_WARNINGS)
     hpx_add_compile_flag(-W3 LANGUAGES C CXX)
     # According to the ifort Windows manual, W3 isn't supported
     hpx_add_compile_flag(-W1 LANGUAGES Fortran)
-    # Boost.Lockfree triggers 'warning C4307: '+' : integral constant overflow'
-    # which is benign
-    hpx_add_compile_flag(-wd4307)
 
     # MSVC2012/2013 are overeager to report 'qualifier applied to function type
     # has no meaning; ignored'
     hpx_add_compile_flag(-wd4180)
+
+    # Boost.Lockfree triggers 'warning C4307: '+' : integral constant overflow'
+    # which is benign
+    hpx_add_compile_flag(-wd4307)
+
+    # object allocated on the heap may not be aligned
+    hpx_add_compile_flag(-wd4316)
 
     # max symbol length exceeded
     hpx_add_compile_flag(-wd4503)
--- a/hpx/lcos/local/latch.hpp
+++ b/hpx/lcos/local/latch.hpp
@@ -13,6 +13,7 @@
 #include <hpx/util/assert.hpp>
 #include <hpx/util/cache_aligned_data.hpp>
 
+#include <atomic>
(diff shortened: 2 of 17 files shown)


hpx/runtime/threads/policies/thread_queue.hpp
```diff
@@ -1128,7 +1152,6 @@ namespace hpx { namespace threads { namespace policies
         std::size_t max_count_;     // maximum number of existing HPX-threads
         task_items_type new_tasks_; // list of new tasks to run
 
-        std::atomic<std::int64_t> new_tasks_count_; // count of new tasks to run
 #ifdef HPX_HAVE_THREAD_QUEUE_WAITTIME
```
msimberg: Hmm, what makes `new_tasks_count_` special? Compared to `thread_map_count_`, work_items_count_` and `terminated_items_count_`.

hkaiser: The timings around that variable were sticking out like a sore thumb. The others might need special handling as well - that will require more investigations, though.

hkaiser:
> Just curious, do you know which changes had the biggest impact (LCOs or schedulers)? 

@msimberg neither ;-) It is the disabling of the stealing counters that gives the biggest effect.

hpx/lcos/local/latch.hpp
```diff
@@ -106,11 +107,14 @@ namespace hpx { namespace lcos { namespace local
...
             HPX_ASSERT(n >= 0);
 
-            std::unique_lock<mutex_type> l(mtx_);
-            HPX_ASSERT(counter_ >= n);
+            std::ptrdiff_t new_count = (counter_ -= n);
+            HPX_ASSERT(new_count >= 0);
 
-            if ((counter_ -= n) == 0)
-                cond_.notify_all(std::move(l));    // release the threads
+            if (new_count == 0)
```
sithhell: This change slightly changes the behavior of the latch. Especially when calling `count_up` and `count_down` concurrently, the notification of the condition might happen even though the count was increased in another thread. This might require additional synchronization for the caller. This synchronization might have been needed previously in those use cases anyway.

hpx/lcos/local/latch.hpp
```diff
@@ -131,15 +134,18 @@ namespace hpx { namespace lcos { namespace local
...
         void wait() const
         {
-            std::unique_lock<mutex_type> l(mtx_);
-            if (counter_ > 0)
-                cond_.wait(l, "hpx::local::latch::wait");
+            if (counter_.load(std::memory_order_acquire) > 0)
+            {
+                std::unique_lock<mutex_type> l(mtx_.data_);
+                if (counter_.load(std::memory_order_relaxed) > 0)
+                    cond_.data_.wait(l, "hpx::local::latch::wait");
```
sithhell: A loop might have been better here:
```
while (counter_.load(std::memory_order_acquire) > 0)
{
    std::unique_lock<mutex_type> l(mtx_.data_);
    cond_.data_.wait(l, "hpx::local::latch::wait");
}

File: hpx/lcos/local/latch.hpp (hunk 2 above)
Comment (hkaiser) [reply hpx/lcos/local/latch.hpp]: The pre-condition for `count_up` is that the internal counter_ is not zero, i.e. `count_up` is called only before the final thread arrives. This should take care of the situation you described.

File: hpx/lcos/local/latch.hpp (hunk 3 above)
Comment (hkaiser) [reply hpx/lcos/local/latch.hpp]: Could you elaborate, please? I don't see the use case for this. If the `counter_` reaches zero all threads should be notified once. There does not seem to be a need for the loop.

File: CMakeLists.txt [hunk 4]
```diff
@@ -519,8 +519,8 @@ hpx_option(HPX_WITH_THREAD_STACK_MMAP BOOL
   CATEGORY "Thread Manager" ADVANCED)
 
 hpx_option(HPX_WITH_THREAD_MANAGER_IDLE_BACKOFF BOOL
-  "HPX scheduler threads do exponential backoff on idle queues (default: ON)"
-  ON
+  "HPX scheduler threads do exponential backoff on idle queues (default: OFF)"
```
msimberg: Side note: the `wait_count_` variable would also be a candidate for "unnecessary" sharing (i.e. it's shared across all threads, whereas there could easily be one per thread). I think turning this off is a really good default though.

In the worst case the idle backoff makes things slower because of additional latency from waking threads up, but in the best case it might even help reduce contention on thread queues.

hkaiser:
Are there any objections left to go ahead and merge this?

msimberg:
@hkaiser this is the only concern: http://cdash.cscs.ch//viewBuildError.php?buildid=46532. Not sure if adding `-faligned-new` will work or help. In the worst case it could be disabled for nvcc.

hkaiser:
@msimberg thanks for pointing this out. I guess we should disable alignment for NVCC. I'll see what I can do.

hkaiser:
@msimberg I thought I had disabled the alignment for nvcc (here: https://github.com/STEllAR-GROUP/hpx/pull/3745/files#diff-88c5f8cc3a54823a2cb084376808e72eR17 and here: https://github.com/STEllAR-GROUP/hpx/pull/3745/files#diff-88c5f8cc3a54823a2cb084376808e72eR17). The compilation seems to still fail, however (http://cdash.cscs.ch//buildSummary.php?buildid=46620). Do I need to anything else (I'm *puzzled*)?

msimberg:
> I'm _puzzled_?

Me too. I'll try building this branch myself tomorrow to see what's up.

msimberg:
@hkaiser it's actually unrelated to nvcc. It's newer the newer gcc combined with C++11 that's complaining about it. I'd suggest adding `hpx_add_compile_flag_if_available(-faligned-new)` since it's good to have it enabled outside of just C++17. Alternatively `hpx_add_compile_flag_if_available(-Wno-aligned-new)` (I think that would work), since it's a rather benign warning.

hkaiser:
@msimberg I have added those flags. For some reason the pycicle runs didn't get started or didn't go through. Could you manually trigger the CUDA tests there, please?

msimberg:
Thanks @hkaiser. The build is [here](http://cdash.cscs.ch/viewBuildError.php?buildid=47287), just waiting for the tests to finish. The compiler flags helped and this looks ready to go, although it seems we've regressed on the CUDA build. Not because of this PR though.

msimberg:
Hmm, I take that back. There's a lot of builds that have either timed out or been canceled (by something being merged to master), and I can't tell which is which (#3506 is very relevant...). Let's at least wait for this last round of builds to finish before merging.

msimberg:
There are definitely more timeouts on this branch compared to master. I'm hoping #3769 will help at least a bit in seeing what is causing timeouts, so let's have a look again after that gets merged.

hkaiser:
> There are definitely more timeouts on this branch compared to master.

@msimberg I guess this is a good sign. Timings have changed sufficiently for lurking bugs to surface...

msimberg:
@hkaiser you're right. It's not knowing if it's actually good or bad that bothers me. I'll fire off a few builds on rostam to cross-check.

msimberg:
That was helpful:
- http://rostam.cct.lsu.edu/builders/hpx_gcc_4_9_boost_1_60_centos_x86_64_debug/builds/257
- http://rostam.cct.lsu.edu/builders/hpx_gcc_4_9_boost_1_60_centos_x86_64_debug/builds/245
I think I know what the problem is (I already had these timeouts without realizing I was testing this branch), but I just want to understand what changed first. I'll let you know when I have something.

hkaiser:
I'm closing this in favor of #3768, which includes the same change set and additionally fixes the test issues that have surfaced here.

PR #3748, opened 2019-03-20
Fixed unusable behavior of the clang code analyzer.

m-diers:
Fixed an unusable behaviour of the the clang code analyzer.
The clangbackend of the qtcreator 4.8.1 aborts with the code analysis at runtime internally.
I couldn't find out exactly what it is, but the following fix helped.
is_brace_constructible() 20 lines above served as template for the fix.

### DIFF
--- a/hpx/traits/brace_initializable_traits.hpp
+++ b/hpx/traits/brace_initializable_traits.hpp
@@ -85,7 +85,8 @@ namespace hpx { namespace traits
 
     template <typename T, std::size_t N>
     constexpr auto is_paren_constructible()
-    -> detail::is_paren_constructible<T, std::make_index_sequence<N>>
+    -> decltype(detail::is_paren_constructible(
+            static_cast<T*>(nullptr), std::make_index_sequence<N>{}))
     {
         return {};
     }


hpx/traits/brace_initializable_traits.hpp
```diff
@@ -85,7 +85,8 @@ namespace hpx { namespace traits
 
     template <typename T, std::size_t N>
     constexpr auto is_paren_constructible()
-    -> detail::is_paren_constructible<T, std::make_index_sequence<N>>
+    -> decltype(detail::is_paren_constructible(
+            static_cast<T*>(nullptr), std::make_index_sequence<N>{}))
```
msimberg: Something's not right with this: http://cdash.cscs.ch//viewBuildError.php?buildid=45979 (those failures are coming from a test using this functionality). Unlike `is_brace_constructible`, `is_paren_constructible` doesn't have a two-argument constructor so this isn't going to work. I wouldn't dare to say if one should be added though. @wteiwewte  would you be able to have a look at this?

@m-diers if you have any more details about the failure that would probably help in fixing it.

m-diers: i suspect i did some crap yesterday. i can't reproduce it today and the test fails on me today, too.
i will close this pr first, and as soon as i have brought light into the dark again, i will get back to you.
sorry for the effort. ;-(

msimberg: No problem, thanks for looking into it!

m-diers:
Put this pr back until I can reproduce my problem again.

hpx/traits/brace_initializable_traits.hpp
m-diers: Found origin, it was an old gcc-toolchain which was used for some reason.

PR #3749, opened 2019-03-21
Fix lifetime problem in run_as_hpx_thread

msimberg:
Fixes a problem in `run_as_hpx_thread` where the condition variable could be accessed after going out of scope.

### DIFF
--- a/hpx/runtime/threads/run_as_hpx_thread.hpp
+++ b/hpx/runtime/threads/run_as_hpx_thread.hpp
@@ -19,6 +19,7 @@
 #include <condition_variable>
 #include <cstdlib>
 #include <exception>
+#include <memory>
 #include <mutex>
 #include <thread>
 #include <type_traits>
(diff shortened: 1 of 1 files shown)


hkaiser:
@msimberg Could you please elaborate on what condition the cv could go out of scope prematurely?

hpx/runtime/threads/run_as_hpx_thread.hpp
```diff
@@ -68,12 +71,12 @@ namespace hpx { namespace threads
                         std::lock_guard<hpx::lcos::local::spinlock> lk(mtx);
                         stopping = true;
                     }
```
msimberg: When this scope ends the lock is released and the spawning thread is free to continue past `cond->wait` (line 79 below). That leaves some time for `run_as_hpx_thread` to return before `cond->notify_all` gets called.

hkaiser: Ok, thanks!

PR #3751, opened 2019-03-22
Fixing problems with lcos::barrier and iostreams

sithhell:
- Fixing memory leak with hpx::cout and friends during shutdown
 - Fixing potential memory corruption with lcos barrier

### DIFF
--- a/hpx/components/iostreams/ostream.hpp
+++ b/hpx/components/iostreams/ostream.hpp
@@ -132,14 +132,14 @@ namespace hpx { namespace iostreams
                 detail::get_outstream(tag));
         }
 
-//         ///////////////////////////////////////////////////////////////////////
-//         void release_ostream(char const* name, naming::id_type const& id);
-//
-//         template <typename Tag>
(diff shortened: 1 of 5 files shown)


hpx/components/iostreams/ostream.hpp
```diff
@@ -281,7 +281,7 @@ namespace hpx { namespace iostreams
             }
 
             // FIXME: find a later spot to invoke this
-//             detail::release_ostream(tag, this->get_id());
+            detail::release_ostream(tag, this->get_id());
```
msimberg: Is the `FIXME` still relevant?

sithhell: probably not...

sithhell:
> I however distinctly remember that we commented out the `release_ostream` function as it was causing issues during shutdown. Have those issues been fixed?

Looks like it indeed. There doesn't seem to be any problems related to that... at least not on the test runners. (not sure what's going on with pycicle...)

PR #3760, opened 2019-03-30
Fix callable vtable for GCC4.9

K-ballo:
Workaround for GCC regression when forming function pointers to [[noreturn]] functions in constant expressions


msimberg:
For reference, [this](http://rostam.cct.lsu.edu/builders/hpx_gcc_4_9_boost_1_60_centos_x86_64_debug/builds/250) was a previously failing build with gcc 4.9, and [this](http://rostam.cct.lsu.edu/builders/hpx_gcc_4_9_boost_1_60_centos_x86_64_debug/builds/251) is after the fix.

PR #3761, opened 2019-04-01
Update python scripts to work with python 3

diehlpk:
## Proposed Changes

  - Update the python scripts to work with python 3

### DIFF
--- a/python/hpx/process.py
+++ b/python/hpx/process.py
@@ -14,12 +14,12 @@ from sys import float_info, platform
 from threading import Thread, Lock
 from time import sleep, time
 from subprocess import Popen, STDOUT, PIPE
-from types import StringType
+#from types import StringType
 from shlex import split
 from signal import SIGKILL
 from os import kill
 from platform import system
-from Queue import Queue, Empty
+from queue import Queue, Empty
 from errno import ESRCH
 # TODO: implement for Windows
 
@@ -64,7 +64,7 @@ def kill_process_tree(parent_pid, signal=SIGKILL):
   for pid in find_process_tree(parent_pid):
     try:
       kill(int(pid), signal)
-    except OSError, err:
+    except OSError as err:
       if ESRCH != err.errno:
         raise err
       else:
@@ -96,7 +96,7 @@ class process(object):
     # other Exceptions aren't in python
     try:
       self._proc.wait()
-    except Exception, err:
+    except Exception as err:
       self._error = err
 
   def _finish(self, thread):
--- a/python/hpx/process.py
+++ b/python/hpx/process.py
@@ -14,7 +14,6 @@ from sys import float_info, platform
 from threading import Thread, Lock
 from time import sleep, time
 from subprocess import Popen, STDOUT, PIPE
-#from types import StringType
 from shlex import split
 from signal import SIGKILL
 from os import kill
--- a/python/scripts/hpx_environment.py
+++ b/python/scripts/hpx_environment.py
@@ -1,6 +1,7 @@
 #! /usr/bin/env python 
 #
 # Copyright (c) 2011 Bryce Lelbach
+# Copyright (c) 2019 Patrick Diehl
 #
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
@@ -25,11 +26,11 @@ parser = OptionParser(usage=usage)
 (options, driver) = parser.parse_args()
 
 if 0 == len(driver):
-  print "No compiler driver specified."
+  print ("No compiler driver specified.")
   exit(1) 
 elif 1 != len(driver):
-  print "More than one compiler driver specified."
+  print ("More than one compiler driver specified.")
   exit(1)
 
-print identify(driver[0])
+print (identify(driver[0]))
 
--- a/python/scripts/hpx_invoke.py
+++ b/python/scripts/hpx_invoke.py
@@ -2,6 +2,7 @@
 #
 # Copyright (c) 2009 Maciej Brodowicz
 # Copyright (c) 2011 Bryce Lelbach
+# Copyright (c) 2018 Patrick Diehl
 #
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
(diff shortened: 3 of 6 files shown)


python/scripts/hpx_environment.py
```diff
@@ -1,6 +1,7 @@
 #! /usr/bin/env python 
 #
 # Copyright (c) 2011 Bryce Lelbach
+# Copyright (c) 2019 Patrick Diehl
 #
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
```
hkaiser: You should add a `from __future__ import print_function` at the very first statement to this script to ensure compatibility with Python V2 (we don't know what `/usr/env/python` refers to).

python/hpx/process.py
```diff
@@ -14,12 +14,12 @@
...
 from time import sleep, time
 from subprocess import Popen, STDOUT, PIPE
-from types import StringType
+#from types import StringType
 from shlex import split
 from signal import SIGKILL
 from os import kill
 from platform import system
-from Queue import Queue, Empty
+from queue import Queue, Empty
```
hkaiser: How about writing this as
```
import sys
sys.version_info[0] < 3:
    from Queue import Queue, Empty
else:
    from queue import Queue, Empty
```
instead?

python/scripts/hpx_invoke.py
```diff
@@ -10,16 +11,16 @@
...
 from optparse import OptionParser
 
-if exists(join(path[0], "../hpx")):
-  path.append(join(path[0], ".."))
-if exists(join(path[0], "../share/hpx/python/hpx")):
-  path.append(join(path[0], "../share/hpx/python"))
+import sys
+
+if exists(join(sys.path[0], "../hpx")):
+  sys.path.append(join(sys.path[0], ".."))
```
hkaiser: Isn't `path` defined in the `os` module, i.e. `os.path.append(join(sys.path[0], ".."))`? That should also work on both platforms, Python V2 and V3.

python/scripts/hpx_invoke.py
```diff
@@ -63,16 +64,16 @@ def rstrip_last(s, chars):
 (options, cmd) = parser.parse_args()
 
 if None == options.program:
-  print "No program specified"
+  print ("No program specified")
```
hkaiser: Same here, please import `__future__`.

python/scripts/hpx_optsweep.py
```diff
@@ -2,6 +2,7 @@
 #
 # Copyright (c) 2009 Maciej Brodowicz
 # Copyright (c) 2011 Bryce Lelbach
+# Copyright (c) 2019 Patrick Diehl
 #
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
```
hkaiser: Please import `__future__` as outlined above.

python/scripts/hpx_run_test.py
```diff
@@ -1,6 +1,7 @@
 #! /usr/bin/env python
 #
 # Copyright (c) 2012 Bryce Adelstein-Lelbach
+# Copyright (c) 2019 Patrick Diehl
 #
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
```
hkaiser: Please import `__future__` as outlined above.

python/scripts/hpx_environment.py
diehlpk: Actually, I dropped the python 2 support intentionally since python 2.x reaches the end of life in around 8 months (https://pythonclock.org/). However, if we still want to support python 2.x, I can add these.

hkaiser: Sure, Python V2 officially ends its live in a couple of month. That does not automatically remove V2 from user's machines for a long time, though. I'd like to see it being held compatible for the time being, not sure how others feel.

msimberg: I'm in favor of dropping Python 2 just for the sake of it being very old. But if importing `__future__` (and the related changes with that) isn't a lot of work then why not. I don't think users will revolt if they will have to install Python 3 though.

diehlpk: The import of future has not to be done if we only support python 2.7, since print is defined as a function there to. I tested my changes with python 2.7 on Fedora and all scripts run without any issues. I would prefer to keep them as the are and just say HPX needs ptthon 2.7 or later. I hope none is using python 2.6 any more.

python/scripts/hpx_invoke.py
diehlpk: See reply above.

python/scripts/hpx_optsweep.py
diehlpk: See reply above

python/scripts/hpx_run_test.py
diehlpk: See reply above

python/scripts/hpx_invoke.py
diehlpk: This worked for python 2.7 and python 3. So I think we can keep it as it is.

python/hpx/process.py
diehlpk: I am not sure why this works for python 2.7 and python 3 in the current version.

python/hpx/process.py
```diff
@@ -14,12 +14,12 @@
 from threading import Thread, Lock
 from time import sleep, time
 from subprocess import Popen, STDOUT, PIPE
-from types import StringType
+#from types import StringType
```
msimberg: Can this be deleted?

python/scripts/hpx_run_test.py
```diff
@@ -22,7 +23,7 @@
 import sys, os, string
 import os.path as osp
 
-from types import StringType
+#from types import StringType
```
msimberg: Remove?

python/scripts/hpx_run_test.py
```diff
@@ -217,18 +218,18 @@ def read_callback(fd, job):
...
 
-    print "-", ("Passed" if test_passed else "Failed")
+    print ("-", ("Passed" if test_passed else "Failed"))
 
-    if "always" == options.log or ("fail" == options.log and not test_passed):
+    if ("always" == options.log or ("fail" == options.log and not test_passed)):
       f = None
 
-      if not options.log_stdout:
+      if (not options.log_stdout):
```
msimberg: I don't think the parenthesis are needed. They (arguably) make the code less readable.

python/hpx/process.py
diehlpk: Yes

python/scripts/hpx_run_test.py
diehlpk: Yes

python/scripts/hpx_run_test.py
diehlpk: yes, it is removed

PR #3766, opened 2019-04-03
Recycle threads more aggressively

msimberg:
Testing more aggressive recycling of threads (rather than deletion). This separates the notion of cleaning up all or some terminated threads, and whether they should be recycled or deleted. This will also only delete threads when the worker thread isn't running.

To do:
- We may need a limit on how many threads are in the recycling heaps (instead of only on how many terminated threads there are)
- Test performance and see if it helps with out of memory errors

This may need further changes, but putting it up here already now in case someone has better ideas about this.

### DIFF
--- a/hpx/runtime/threads/policies/thread_queue.hpp
+++ b/hpx/runtime/threads/policies/thread_queue.hpp
@@ -447,7 +447,7 @@ namespace hpx { namespace threads { namespace policies
         ///
         /// This returns 'true' if there are no more terminated threads waiting
         /// to be deleted.
-        bool cleanup_terminated_locked(bool delete_all = false)
+        bool cleanup_terminated_locked(bool delete_all, bool recycle_threads)
         {
 #ifdef HPX_HAVE_THREAD_CREATION_AND_CLEANUP_RATES
             util::tick_counter tc(cleanup_terminated_time_);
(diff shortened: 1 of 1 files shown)


msimberg:
I ran the `future_overhead` test and the results are... interesting. With a small number of threads it doesn't seem to make much of a difference. With a lot of threads it does this:

master:
```
36 threads                                                                                                                                                                                              
invoked 500000 futures (actions) in 2.888646 seconds                                                                                                                                                    
invoked 500000 futures (functions, wait_each) in 0.676633 seconds                                                                                                                                       
invoked 500000 futures (functions, wait_all) in 0.566526 seconds                                                                                                                                        
invoked 500000 futures (functions, thread count) in 0.390299 seconds
```

this branch:
```
36 threads
invoked 500000 futures (actions) in 1.044909 seconds
invoked 500000 futures (functions, wait_each) in 0.881757 seconds
invoked 500000 futures (functions, wait_all) in 0.919252 seconds
invoked 500000 futures (functions, thread count) in 0.762666 seconds
```

I don't know why this is happening but it's probably worth investigating closer.

msimberg:
I'm going to close this now. Testing again shows practically identical results compared to master now.

PR #3768, opened 2019-04-03
Adding flags to scheduler allowing to control thread stealing and idle back-off

### LINKED ISSUE #3744: Separate spinlock from same cacheline as internal data for all LCOs

hkaiser:
This adds a couple of things:

- adding `mark_end_of_scheduling` hook to execution parameters allowing to be  called back after the scheduling phase of the parallel algorithm
- added invocation of mark_begin/mark_end scheduling parameters to sequential  for_each and for_loop
- added flags to scheduling loop that allows to control thread stealing and  idle back-off (effective only for schedulers that support it)
- changed implementation of static_priority_scheduler to full rely on  local_priority_scheduler except that it starts of with stealing disabled
- made fast_idle_mode scheduler flag more useful as it now speeds up idle  back-off in case of unsuccessful thread stealing (no work available through  thread stealing)
- set `HPX_WITH_THREAD_MANAGER_IDLE_BACKOFF=ON` by default (can now be disabled  at runtime)
- added `add_scheduler_mode`/`remove_scheduler_mode`/`add_remove_scheduler_mode`  APIs allowing to control the scheduler mode dynamically
- flyby: removed unused arguments for get_next_thread, added argument  controlling enable_stealing for wait_or_add_new
- modified for_each_scaling test to not measure data initialization  also added command line options `--disable_stealing` and `--fast_idle_mode`
- duplicated scheduler mode across all cores (with cache alignment) to reduce false sharing

~~This relies on #3745 being merged first.~~ This subsumes #3745

Fixes #3744


hpx/runtime/threads/detail/scheduling_loop.hpp
```diff
@@ -822,10 +846,11 @@ namespace hpx { namespace threads { namespace detail
             // if nothing else has to be done either wait or terminate
             else
             {
-                ++idle_loop_count;
+                --idle_loop_count;
```
msimberg: Does this need to be reversed? Just thinking about naming. Right now it's more like `idle_loop_countdown` ;) and it's different from the other counts that actually grow.

hpx/runtime/threads/policies/local_priority_queue_scheduler.hpp
```diff
@@ -247,7 +247,7 @@ namespace hpx { namespace threads { namespace policies
                 num_pending_misses += high_priority_queues_[num_thread].data_->
                     get_num_pending_misses(reset);
             }
-            if (num_thread == 0)
+            if (num_thread == num_queues_-1)
```
msimberg: Does this make a difference?

hkaiser: There was some inconsistency that we were accessing the low-priority queue from different cores, sometimes it was core zero sometimes the last one. I made it consistent to reduce false  sharing on that queue.

hpx/runtime/threads/detail/scheduling_loop.hpp
hkaiser: I reversed this as it helped me not to carry around the max value to the point where it needed to be reset.

hkaiser:
@msimberg, @biddisco: this is ready to go from my end, please review.

msimberg:
What was the reason the `shared_priority_queue_scheduler` was timing out?

hkaiser:
@msimberg the `wait_or_add_new` was not properly setting its return value forcing the scheduler loop to wait forever for it to signal its termination status; see for instance here: https://github.com/STEllAR-GROUP/hpx/pull/3768/files#diff-c730f66e3d702c09217b1bbcc6b570f2R1317.

PR #3769, opened 2019-04-03
Fix SIGSEGV handler

msimberg:
Makes sure program terminates on SIGSEGV. Also replaces `nasty_cast` with `reinterpret_cast`.


hpx/runtime/threads/coroutines/detail/context_linux_x86.hpp
```diff
@@ -82,16 +82,6 @@ namespace hpx { namespace threads { namespace coroutines
...
     {
-        template <typename TO, typename FROM>
-        TO nasty_cast(FROM f)
-        {
-            union {
-                FROM f; TO t;
-            } u;
-            u.f = f;
-            return u.t;
-        }
```
msimberg: @sithhell, @K-ballo, @hkaiser I'm trying to sneak this one in at the same time. AFAICT `reinterpret_cast` will do the right thing where `nasty_cast` was used, but am I missing some weird corner of the standards?

hkaiser:
The filing suspend_resume test will be fixed by merging #3768

PR #3781, opened 2019-04-10
Fix generic context coroutines after lazy stack allocation changes

msimberg:
Build with generic context coroutines running [here](http://rostam.cct.lsu.edu/builders/hpx_gcc_6_boost_1_64_centos_x86_64_debug/builds/265). I'll update one of the pycicle builders to test this as well.


msimberg:
The fact that `fcontext_context_impl` is now a template class. Can I actually define the static member functions in the cpp file for a template class?

hkaiser:
> The fact that fcontext_context_impl is now a template class. 

Ahh, sorry, I missed that.

> Can I actually define the static member functions in the cpp file for a template class?

You could, but only if you know the types the template is going to be instantiated with.

PR #3645, opened 2019-02-01
Enabling sanitizers on gitlab runner

sithhell:
This PR enables the undefined and address sanitizer for the gitlab runners.

Builds on #3784.

PR #3784, opened 2019-04-10
Fix leak and address sanitizer problems

msimberg:
Separates the problems fixed in #3645 from the GitLab runner related changes.

### DIFF
--- a/hpx/config/compiler_specific.hpp
+++ b/hpx/config/compiler_specific.hpp
@@ -115,5 +115,11 @@
 #define HPX_CDECL
 #endif
 
+#if defined(__has_feature)
+#  if __has_feature(address_sanitizer)
+#    define HPX_HAVE_ADDRESS_SANITIZER
+#  endif
+#endif
+
 #endif
 
--- a/hpx/config/compiler_specific.hpp
+++ b/hpx/config/compiler_specific.hpp
@@ -115,7 +115,7 @@
 #define HPX_CDECL
 #endif
 
-#if defined(__has_feature)
+#if defined(HPX_HAVE_SANITIZERS) && defined(__has_feature)
 #  if __has_feature(address_sanitizer)
 #    define HPX_HAVE_ADDRESS_SANITIZER
 #  endif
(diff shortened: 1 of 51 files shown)


hpx/config/compiler_specific.hpp
```diff
@@ -115,5 +115,11 @@
 #define HPX_CDECL
 #endif
 
+#if defined(__has_feature)
```
hkaiser: How does this relate to the `HPX_HAVE_SANITIZERS` pp constant defined above?

hpx/runtime/threads/coroutines/detail/context_impl.hpp
```diff
@@ -117,6 +117,10 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail
 #elif (defined(__linux) || defined(linux) || defined(__linux__)) &&            \
     !defined(__bgq__) && !defined(__powerpc__) && !defined(__s390x_)
 
+#if defined(HPX_HAVE_ADDRESS_SANITIZER)
+#define HPX_CONTEXT_HAVE_ADDRESS_SANITIZER
```
hkaiser: Why do we need this extra pp constant?

hpx/config/compiler_specific.hpp
msimberg: Good question. `HPX_HAVE_SANITIZERS` should presumably be enabled (by the user) if *any* sanitizer is enabled, whereas `HPX_HAVE_ADDRESS_SANITIZER` is only enabled for the address sanitizer. I'm not sure if we need the distinction.

hkaiser: In this case, shouldn't the above be wrapped in a `#ifdef HPX_HAVE_ADDRESS_SANITIZER/#endif`, then?

sithhell: In principle, the user could set the asan flag independently of setting the CMake variable `HPX_WITH_SANITIZERS`. This will probably lead to linker errors though since the sanitizers are linked in completely dynamically. To make everything consistent, this should be wrapped in a `#ifdef HPX_HAVE_SANITIZERS/#endif`

hpx/runtime/threads/coroutines/detail/context_impl.hpp
sithhell: We probably don't. Looks like this is a leftover.

msimberg:
Let's go ahead with this. Thanks @sithhell for the massive number of fixes in here!

PR #3786, opened 2019-04-10
Bump minimum required versions for 1.3.0 release

msimberg:
Bump minimum Boost version to 1.61 and GCC to 5.

### DIFF
--- a/hpx/config.hpp
+++ b/hpx/config.hpp
@@ -33,9 +33,9 @@
 
 #include <boost/version.hpp>
 
-#if BOOST_VERSION < 105800
+#if BOOST_VERSION < 106100
 // Please update your Boost installation (see www.boost.org for details).
-#error HPX cannot be compiled with a Boost version earlier than 1.58.0
+#error HPX cannot be compiled with a Boost version earlier than 1.61.0
 #endif
 
 #include <hpx/util/detail/pp/cat.hpp>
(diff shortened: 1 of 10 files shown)


hpx/config.hpp
```diff
@@ -33,9 +33,9 @@
 
 #include <boost/version.hpp>
 
-#if BOOST_VERSION < 105800
+#if BOOST_VERSION < 106100
```
hkaiser: Can we leave this unchanged? It will most likely work, we just don't support it anymore. Perhaps a #warning instead?

msimberg: Good point. I can do that.

msimberg: Hmm. Thinking about this a bit more, I'd rather not just warn about it this. The only versions that could work are 1.59 and 1.60, anything older definitely won't work. A warning can be too easily overlooked, so I'd rather just be explicit and strict about it. If anything maybe we don't need to bump the minimum version 1.61, but rather 1.59 or something, but sticking to the last 10 releases is much more straightforward.

K-ballo: This should be a hard error, unless perhaps explicitly requested otherwise (will void warranty). Whether there's value in making this a supported option vs letting people try their luck and patch the code, I do not know.

hkaiser: Ok, fair point.

msimberg:
Ok to merge?

PR #3787, opened 2019-04-10
Bump minimal boost supported version to 1.61.0

K-ballo:


hpx/config.hpp
```diff
@@ -33,9 +33,9 @@
 
 #include <boost/version.hpp>
 
-#if BOOST_VERSION < 105800
+#if BOOST_VERSION < 106100
```
hkaiser: Should we bail out or should we just warn, it might work after all.

msimberg:
@K-ballo this is nearly the same as #3786 ;) But I see you've been more proactive in bumping recommended version numbers and some other changes. I'll move those changes over to the other branch. Would you mind checking if there's some difference you don't agree with?

K-ballo:
> @K-ballo this is nearly the same as #3786 ;)

I hadn't noticed that other PR. This was on an old branch, waiting for the release of Boost 1.70.0.

Issue #3792, opened 2019-04-12
Barrier Hangs When Locality Zero not included


folshost:
## Expected Behavior

A barrier should register successfully when being executed by localities 1 and 2 (from {0,1,2})

## Actual Behavior

The barrier hangs, apparently on register_lock.

```
//  Copyright (c) 2019 Weile Wei
//  Copyright (c) 2019 Maxwell Reeser
//  Copyright (c) 2019 Hartmut Kaiser
//
//  Distributed under the Boost Software License, Version 1.0. (See accompanying
//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)

#include <hpx/hpx.hpp>
#include <hpx/hpx_init.hpp>
#include <hpx/include/iostreams.hpp>
#include <hpx/lcos/barrier.hpp>
#include <hpx/runtime/components/component_factory.hpp>

void run_barrier_test(std::vector<size_t> locs)
{
    if (std::find(locs.begin(), locs.end(), hpx::get_locality_id()) == locs.end())
        return;

    std::string barrier_name = 
        "/loc_list/barrier" + std::to_string(locs[0]) + std::to_string(locs[1]);
    hpx::lcos::barrier b(barrier_name, locs.size(), hpx::get_locality_id());
    b.wait();
}

int hpx_main() 
{
    std::cout << "Hello world from locality " << hpx::get_locality_id() << std::endl;
    std::vector<size_t> locs_0{ 0, 1 };
    run_barrier_test(locs_0);
    std::vector<size_t> locs_1{ 0, 2 };
    run_barrier_test(locs_1);
    std::vector<size_t> locs_2{ 1, 2 };
    run_barrier_test(locs_2);
    return hpx::finalize();
}

int main(int argc, char *argv[]) { return hpx::init(argc, argv); }
```

Run the above code in Visual Studio with three localities with command line arguments "--hpx:localities=3 --hpx:node={0-2} --hpx:run-hpx-main --hpx:threads=1" with the appropriate value for --hpx:node specified for each locality. The first two runs of run_barrier_test will complete successfully, while the third will hang.

  - HPX Version: Most recent on github
  - Platform (compiler, OS): running on Windows 64-bit, with vcpkg for libraries, in VS2017

folshost:
@sithhell can you please take a look at this?

sithhell:
Thanks for the testcase, I'll have a look ASAP.

sithhell:
[...]

PR #3794, opened 2019-04-12
Fixing #3792

sithhell:
This fixes the problems reported in #3792. The main issue lies within an
implementation detail of the barrier. The rank of a barrier is
completely decoupled from the locality id. For that reason, the barrier
root is always rank 0. If no barrier with rank 0 is given, we get the
reported hang. The added test shows both a workaround to that issue and
adds an additional barrier constructor that takes a vector of ranks, and
the rank of the given barrier instantiation.

### DIFF
--- a/src/lcos/barrier.cpp
+++ b/src/lcos/barrier.cpp
@@ -21,6 +21,7 @@
 #include <cstddef>
 #include <string>
 #include <utility>
+#include <vector>
 
 ///////////////////////////////////////////////////////////////////////////////
 namespace hpx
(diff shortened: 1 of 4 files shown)


src/lcos/barrier.cpp
```diff
@@ -59,6 +59,21 @@ namespace hpx { namespace lcos {
                 base_name, node_->get_unmanaged_id(), (*node_)->rank_).get();
     }
 
+    barrier::barrier(std::string const& base_name, std::vector<std::size_t> const& ranks, std::size_t rank)
+    {
+        auto rank_it = std::find(ranks.begin(), ranks.end(), hpx::get_locality_id());
```
hkaiser: Shouldn't this use the `rank` parameter instead of `get_locality_id()`? Alternatively, could we get rid of the explicit `rank`?

sithhell: Good point, this should be `rank` indeed. I don't think we should get rid of the explicit `rank` parameter here.

hkaiser: @sithhell Is there a use case where `rank != hpx::get_locality_id()`?

sithhell: Anything where you need a barrier on a topology that is not the same as the one of the localities.

PR #3798, opened 2019-04-15
count and count_if accepts different iterator types

pmikolajczyk41:


tests/regressions/parallel/count_3646.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    HPX_TEST_EQ(result, stdResult);
+}
+
+void test_count_if()
+{
+    using Iter = BitCountingIterator<std::int64_t{33}>;
+    using Sent = Sentinel<std::int64_t>;
+
+    auto predicate = [](auto v) { return v == 1; };
```
msimberg: Would you mind changing the `auto v` to spell out the type explicitly? For now we still try to support compiling with C++11.

pmikolajczyk41: I am very sorry that it took so long - I was completely offline for the Easter time.

msimberg: No problem at all. Thanks for updating it, that should do it! I'll let the CI do a final check, but then we should be able to merge this.

PR #3799, opened 2019-04-16
Remove compat namespace and its contents

msimberg:
Use standard library functionality where applicable (`thread`, `mutex`, `condition_variable`), and move the rest to the util namespace (`barrier`).

Trying my luck with this one. The only reason I can see for not removing this is that we didn't officially deprecate it. There are still some places that use functionality from Boost.Threads explicitly. One of them is https://github.com/STEllAR-GROUP/hpx/blob/bc423e98065c34db0d5ae5455f79f6b15c0ab0e1/plugins/parcelport/unordered_map.hpp#L43-L44 which we could get rid of if we drop C++11. Another is the fallback implementation for `thread_specific_ptr` which is off by default. The rest seem to be in tests using either `thread_specific_ptr` or again `boost::shared_lock` which could be removed if we drop C++11.

Note: I'm not suggesting we drop C++11 right now, just sometime pretty soon.


biddisco:
Only  the verbs parcelport uses the map and we might want to just dump the verbs PP. Unfortunately, we might not be able to run the libfabric PP on top of verbs because the verbs provider for libfabric doesn't support `FI_SOURCE` which we use (see https://github.com/ofiwg/libfabric/wiki/Provider-Feature-Matrix-master for info).
It should be possible to run one of the libfabric RXD utility providers on top of the verbs provider to get `FI_SOURCE` but I have not tried o tested this. If it works, then we do not need the verbs PP any more and we can do a big cleanup/purge. I will investigate this further.

hkaiser:
> Note: I'm not suggesting we drop C++11 right now, just sometime pretty soon.

I hope we will never 'drop' C++11 as such (would I be allowed to use a `for` loop in that case?) 

We should be more careful in formulating this as 'Unconditionally assuming for all of C++14 to be available', instead.

msimberg:
> > Note: I'm not suggesting we drop C++11 right now, just sometime pretty soon.
> 
> I hope we will never 'drop' C++11 as such (would I be allowed to use a `for` loop in that case?)
> 
> We should be more careful in formulating this as 'Unconditionally assuming for all of C++14 to be available', instead.

You're right. "Require C++14 support" is exactly what I mean by this.

msimberg:
What's the opinion on this now?

Note that the comment about requiring C++14 was just a possible next step. This PR does not make C++14 a requirement. It just replaces the `compat` features with equivalent `std` features which are in C++11. What's not clear to me if something somewhere still requires the compatibility layer.

msimberg:
I'm going to interpret silence on this one as acceptance. Will merge in the next few days if no-one objects.

hkaiser:
@khuck IIRC, APEX needed `Boost.Thread` on some platforms at some point. Is that still true?

khuck:
@hkaiser We only use boost/assign.hpp with HPX, and boost/regex.hpp on Intel MIC (i.e. Cori), because the intel compiler was missing std::regex support.  No use of boost threads

PR #3800, opened 2019-04-16
Attempt to fix Posix context switching after lazy init changes

hkaiser:
Blind edit of the POSIX context switching, fixing it after lazy init changes

### DIFF
--- a/hpx/runtime/threads/coroutines/detail/context_posix.hpp
+++ b/hpx/runtime/threads/coroutines/detail/context_posix.hpp
@@ -235,7 +235,6 @@ namespace hpx { namespace threads { namespace coroutines
                   ? static_cast<std::ptrdiff_t>(default_stack_size)
                   : stack_size),
                 m_stack(nullptr),
-                cb_(&cb),
                 funp_(&trampoline<CoroutineImpl>)
             {
             }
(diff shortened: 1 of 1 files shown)


hpx/runtime/threads/coroutines/detail/context_posix.hpp
```diff
@@ -396,7 +408,6 @@ namespace hpx { namespace threads { namespace coroutines
             // declare m_stack_size first so we can use it to initialize m_stack
             std::ptrdiff_t m_stack_size;
             void * m_stack;
-            void * cb_;
```
sithhell: why has this been removed?

khuck:
@hkaiser @sithhell I committed the changes to fix this branch.  It builds with Clang 7 on POWER8 now.

hkaiser:
@khuck thanks a lot!

hpx/runtime/threads/coroutines/detail/context_posix.hpp
hkaiser: @sithhell do we still need it?

sithhell: oh right, missed that we don't need it anymore...

PR #3802, opened 2019-04-17
Numa binding allocator

biddisco:
## Proposed Changes
This introduces a numa_allocator that can provide memory with interleaved/first touch or user defined numa binding per page of memory.
The allocator uses a 'binder' object that defines the way pages are bound and users can supply their own binders. A 1D and 2D example (unit test) are provided that mimic the way we bind memory in our matrix allocation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -739,7 +739,9 @@ jobs:
           when: always
           command: |
               ulimit -c unlimited
-              ctest -T test --no-compress-output --output-on-failure -R tests.unit.topology
+              ctest -T test --no-compress-output --output-on-failure \
+                  -R tests.unit.topology \
+                  -E tests.unit.topology.numa_allocator
       - run:
           <<: *convert_xml
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -725,6 +725,32 @@ jobs:
       - store_artifacts:
           path: tests.unit.threads
 
+  tests.unit.topology:
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: /hpx
+      - run:
+          name: Building Unit Tests
+          command: |
+              ninja -j2 -k 0 tests.unit.topology
+      - run:
+          name: Running Unit Tests
+          when: always
+          command: |
+              ulimit -c unlimited
+              ctest -T test --no-compress-output --output-on-failure -R tests.unit.topology
+      - run:
+          <<: *convert_xml
+      - run:
+          <<: *move_core_dump
+      - run:
+          <<: *move_debug_log
+      - store_test_results:
+          path: tests.unit.topology
+      - store_artifacts:
+          path: tests.unit.topology
+
   tests.unit.traits:
     <<: *defaults
     steps:
@@ -1205,6 +1231,8 @@ workflows:
           <<: *core_dependency
       - tests.unit.threads:
           <<: *core_dependency
+      - tests.unit.topology:
+          <<: *core_dependency
       - tests.unit.traits:
           <<: *core_dependency
       - tests.unit.util:
@@ -1273,6 +1301,7 @@ workflows:
             - tests.unit.resource
             - tests.unit.serialization
             - tests.unit.threads
+            - tests.unit.topology
             - tests.unit.traits
             - tests.unit.util
             - tests.unit.libs
--- a/examples/transpose/transpose_block_numa.cpp
+++ b/examples/transpose/transpose_block_numa.cpp
@@ -575,7 +575,7 @@ int main(int argc, char* argv[])
     std::size_t numa_nodes = get_num_numa_nodes(topo, vm);
     std::pair<std::size_t, std::size_t> pus =
         get_num_numa_pus(topo, numa_nodes, vm);
-    std::size_t num_cores = topo.get_number_of_numa_node_cores(0);
+    //std::size_t num_cores = topo.get_number_of_numa_node_cores(0);
 
     // Initialize and run HPX, this example requires to run hpx_main on all
     // localities
--- a/examples/transpose/transpose_block_numa.cpp
+++ b/examples/transpose/transpose_block_numa.cpp
@@ -583,8 +583,8 @@ int main(int argc, char* argv[])
         "hpx.run_hpx_main!=1",
         "hpx.numa_sensitive=2",  // no-cross NUMA stealing
         // block all cores of requested number of NUMA-domains
-        hpx::util::format("hpx.cores={}", numa_nodes * num_cores),
-        hpx::util::format("hpx.os_threads={}", numa_nodes * pus.second)
+//        hpx::util::format("hpx.cores={}", numa_nodes * num_cores),
+//        hpx::util::format("hpx.os_threads={}", numa_nodes * pus.second)
     };
 
     std::string node_name("numanode");
@@ -605,8 +605,9 @@ int main(int argc, char* argv[])
             pus.second-1                             // core:0-%d
         );
     }
-    cfg.push_back(bind_desc);
+//    cfg.push_back(bind_desc);
 
+//    return hpx::init();
     return hpx::init(desc_commandline, argc, argv, cfg);
 }
 
--- a/hpx/parallel/util/numa_binding_allocator.hpp
+++ b/hpx/parallel/util/numa_binding_allocator.hpp
@@ -164,7 +164,7 @@ namespace hpx { namespace compute { namespace host {
         using numa_binding_helper_ptr = std::shared_ptr<numa_binding_helper<T>>;
 
         // default constructor produces an unusable allocator
-        numa_binding_allocator() {}
+        numa_binding_allocator() = default;
 
         // construct without a memory binder function
         // only first touch or interleave policies are valid
@@ -209,7 +209,7 @@ namespace hpx { namespace compute { namespace host {
             LOG_NUMA_MSG("numa_binding_allocator : Copy allocator rebind");
         }
 
-        // Move assignment
+        // Move constructor
         numa_binding_allocator(numa_binding_allocator&& rhs)
           : binding_helper_(std::move(rhs.binding_helper_))
           , policy_(rhs.policy_)
--- a/hpx/parallel/util/numa_binding_allocator.hpp
+++ b/hpx/parallel/util/numa_binding_allocator.hpp
@@ -9,9 +9,9 @@
 #include <hpx/config.hpp>
 
 #include <hpx/parallel/execution_policy.hpp>
-#include <hpx/runtime/threads/topology.hpp>
-#include <hpx/runtime/threads/executors/pool_executor.hpp>
 #include <hpx/runtime/threads/executors/guided_pool_executor.hpp>
+#include <hpx/runtime/threads/executors/pool_executor.hpp>
+#include <hpx/runtime/threads/topology.hpp>
 #include <hpx/util/assert.hpp>
 
 #include <sstream>
@@ -19,15 +19,15 @@
 #include <vector>
 
 #include <cstddef>
-#include <type_traits>
-#include <utility>
 #include <iostream>
 #include <memory>
+#include <type_traits>
+#include <utility>
 
 #if defined(__linux) || defined(linux) || defined(__linux__)
-# include <linux/unistd.h>
-# include <sys/mman.h>
-# define NUMA_ALLOCATOR_LINUX
+#include <linux/unistd.h>
+#include <sys/mman.h>
+#define NUMA_ALLOCATOR_LINUX
 #endif
 
 // Can be used to enable debugging of the allocator page mapping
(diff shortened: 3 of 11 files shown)


biddisco:
PS. 2 extra flyby commits are included that fix issues related to NUMA that were found when working on this.

biddisco:
PPS. This allocator could/should replace the numa_allocator that is already present in HPX, but for now the two can coexist (even if the original is not used anywhere). This one was named numa_binding_allocator to distinguish it from the former.

examples/transpose/transpose_block_numa.cpp
```diff
@@ -583,8 +583,8 @@ int main(int argc, char* argv[])
         "hpx.run_hpx_main!=1",
         "hpx.numa_sensitive=2",  // no-cross NUMA stealing
         // block all cores of requested number of NUMA-domains
-        hpx::util::format("hpx.cores={}", numa_nodes * num_cores),
-        hpx::util::format("hpx.os_threads={}", numa_nodes * pus.second)
+//        hpx::util::format("hpx.cores={}", numa_nodes * num_cores),
+//        hpx::util::format("hpx.os_threads={}", numa_nodes * pus.second)
```
hkaiser: Can we remove the commented code sections here? Or is this an oversight?

hpx/parallel/util/numa_binding_allocator.hpp
```diff
@@ -0,0 +1,569 @@
...
+{
+    struct numa_binding_allocator_tag {};
+
+    template <>
+    struct HPX_EXPORT pool_numa_hint<numa_binding_allocator_tag> {
+      // The call operator () must return an int type
+      // The arguments must be const ref versions of the equivalent task arguments
+      int operator()(const std::size_t& domain) const {
+          LOG_NUMA_MSG("allocator pool_numa_hint returns domain " << domain);
+          return domain;
```
hkaiser: This will cause `size_t -> int` conversion warnings on some platforms (implicit narrowing conversion).

hpx/parallel/util/numa_binding_allocator.hpp
```diff
@@ -0,0 +1,569 @@
...
+
+        // The allocator uses the pool name to get numa bitmap masks needed by
+        // the allocation function. The "default" pool is assumed
+        // @TODO use an executor to retrieve the pool name
+        virtual const std::string &pool_name() const { return pool_name_; }
+
+        // Return the total memory consumption in bytes
+        virtual std::size_t memory_bytes() const { return 0; }
+
+//#ifdef NUMA_BINDING_ALLOCATOR_DEBUG_PAGE_BINDING
```
hkaiser: Should the debug-support be always on?

hpx/parallel/util/numa_binding_allocator.hpp
```diff
@@ -0,0 +1,569 @@
...
+        template <typename U>
+        struct rebind
+        {
+            typedef numa_binding_allocator<U> other;
+        };
+
+        using numa_binding_helper_ptr = std::shared_ptr<numa_binding_helper<T>>;
+
+        // default constructor produces an unusable allocator
+        numa_binding_allocator() {}
```
hkaiser: Use `= default;` instead?

hpx/parallel/util/numa_binding_allocator.hpp
```diff
@@ -0,0 +1,569 @@
...
+            LOG_NUMA_MSG("numa_binding_allocator : no binder function");
+            HPX_ASSERT(policy_ != threads::hpx_hwloc_membind_policy::membind_user);
+        }
+
+        // construct using a memory binder function for placing pages
+        // according to a user defined pattern using first touch policy
+        numa_binding_allocator(numa_binding_helper_ptr bind_func,
+                       threads::hpx_hwloc_membind_policy policy,
+                       unsigned int flags)
+            : binding_helper_(bind_func)
```
hkaiser: This could use `std::move()` to avoid the (atomic) reference count.

hpx/parallel/util/numa_binding_allocator.hpp
```diff
@@ -0,0 +1,569 @@
...
+        // Move assignment
+        numa_binding_allocator& operator=(numa_binding_allocator && rhs)
+        {
+            binding_helper_ = rhs.binding_helper_;
+            policy_         = rhs.policy_;
+            flags_          = rhs.flags_;
+
+            LOG_NUMA_MSG("numa_binding_allocator : Move assignment");
+            return *this;
+        }
```
hkaiser: You should add a move-constructor if you have a move-assignment operator.

hpx/parallel/util/numa_binding_allocator.hpp
```diff
@@ -0,0 +1,569 @@
...
+                    threads::topology().
+                        set_area_membind_nodeset(page_ptr, pagesize,
+                                                 nodesets[dom]->get_bmp());
+                }
+                page_ptr += pageN;
+            }
+        }
+
+    private:
+        hpx::lcos::local::spinlock display_mutex;
```
hkaiser: This seems to be only needed if `NUMA_BINDING_ALLOCATOR_DEBUG_PAGE_BINDING` is defined

tests/unit/topology/allocator_binder_linear.hpp
```diff
@@ -0,0 +1,83 @@
...
+// This simply binds pages according to a linear sequence
+// with increasing numa domain number modulo the total domains available
+// ------------------------------------------------------------------------
+template <typename T>
+struct linear_numa_binder : hpx::compute::host::numa_binding_helper<T>
+{
+    linear_numa_binder(std::size_t num_pages) :
+        hpx::compute::host::numa_binding_helper<T>()
+    {
+        const std::size_t CACHE_LINE_SIZE = sysconf(_SC_LEVEL1_DCACHE_LINESIZE);
```
hkaiser: `sysconf` is a Linux-specific function call. This should be hidden on other platforms.

tests/unit/topology/allocator_binder_linear.hpp
```diff
@@ -0,0 +1,83 @@
+#ifndef ALLOCATOR_BINDER_LINEAR_HPP
```
hkaiser: Copyright/license is missing.

tests/unit/topology/allocator_binder_linear.hpp
```diff
@@ -0,0 +1,83 @@
...
+    {
+        std::intptr_t offset = page_ptr - base_ptr;
+        std::size_t   index  = (offset / elements_page_);
+        return index % domains;
+    }
+
+    // This is for debug purposes
+    virtual std::string description() const override
+    {
+        std::ostringstream temp;
```
hkaiser: The corresponding `#include` is missing from this file.

examples/transpose/transpose_block_numa.cpp
biddisco: The old example was crashing on startup because it set those params and hey didn't work as expected with the resource partitioner stuff. I commented out stuff to make the example run without error, but it doesn't not really work any more and it needds further work. For now, I think this should be left 'as is' until I get time to patch it up and make it work with the new design of allocators and RP.

hpx/parallel/util/numa_binding_allocator.hpp
biddisco: I'll fix it - thanks.

hpx/parallel/util/numa_binding_allocator.hpp
biddisco: no  it shouldn't.

hpx/parallel/util/numa_binding_allocator.hpp
biddisco: ok

hpx/parallel/util/numa_binding_allocator.hpp
biddisco: you mean the shared_ptr? I would not worry here. this is not a critical piece of code, the allocator will be constructed once before doing math operations and kept alive until termination most likely

tests/unit/topology/allocator_binder_linear.hpp
biddisco: I have added a function to topology to get the page size on startup (mostly platform independent) and store it in a static var that is then returned by calls to hpx::threads::get_memory_page_size() - this should resolve this problem.

biddisco:
I believe I've addressed all the issues raised by the review. Please have another look. Thanks

msimberg:
@biddisco mind having a look at this: https://circleci.com/gh/STEllAR-GROUP/hpx/82304? There are a few more unresolved comments as well. Would you mind either fixing them or saying why not? Thanks!

biddisco:
> @biddisco mind having a look at this: https://circleci.com/gh/STEllAR-GROUP/hpx/82304? There are a few more unresolved comments as well. Would you mind either fixing them or saying why not? 

I'm slightly confused by the circle CI message, it says `('Executable /hpx/build/bin/numa_allocator_test does not exist'` - but there is no error in the build step that might indicate why it doesn't exit.
I'll keep experimenting - however - "There are a few more unresolved comments as well" - I went through them one by one - which ones did I not address?

msimberg:
Agree. It's a bit odd. I've restarted the build but I don't think that's going to help. Maybe it's just a matter of a typo/spelling difference somewhere? As for unresolved comments I was referring to the ones about constructors.

msimberg:
@biddisco you might also want to enable some parts conditionally: http://cdash.cscs.ch/viewBuildError.php?buildid=51348 (C++11 Clang 4 build). (Note: the hwloc version is very old there, I'll bump that up to something that has the needed enums/macros as long as it's not too high.)

tests/unit/topology/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+
+  # add example executable
+  add_hpx_executable(${test}_test
+                     SOURCES ${sources}
+                     ${${test}_FLAGS}
+                     EXCLUDE_FROM_ALL
+                     HPX_PREFIX ${HPX_BUILD_PREFIX}
+                     FOLDER "Tests/Unit/Topology/")
+
+  add_hpx_unit_test("util" ${test} ${${test}_PARAMETERS})
```
msimberg: `"util"` should be `topology` and then you'll need to add an entry to the CircleCI config for `tests.unit.topology`.

biddisco: Done and pushed a new commit. If it doesn't pass all tests now, then I want a refund of all the time I've spent on it.

.circleci/config.yml
```diff
@@ -720,7 +720,33 @@ jobs:
...
+      - run:
+          <<: *move_core_dump
+      - run:
+          <<: *move_debug_log
+      - store_test_results:
+          path: tests.unit.topology
+      - store_artifacts:
+          path: tests.unit.topology
+
+tests.unit.traits:
```
msimberg: This probably needs to be indented.

biddisco:
I hope that all the review comments have been addressed. I fixed the constructor, have upgraded some places to use threads::get_memory_page_size(), reduced the test footprint and if the CircleCi fix to the test runner stops it running docker, then I believe that this is good to go. Pleas let me know if I've forgotten to address any issues.

msimberg:
@biddisco I know you'll thank us for this, but could you have a final look at the Codacy report. [This one](https://app.codacy.com/app/STEllAR-GROUP/hpx/file/34612186970/issues/source?bid=12589126&fileBranchId=12589177#l293) in particular looks like a real bug.

biddisco:
> @biddisco I know you'll thank us for this, but could you have a final look at the Codacy report. [This one](https://app.codacy.com/app/STEllAR-GROUP/hpx/file/34612186970/issues/source?bid=12589126&fileBranchId=12589177#l293) in particular looks like a real bug.

Not bad. I am impressed that this was found by an automated tool. I had obviously bodged a merge - or pasted a snippet in by mistake (or during testing and forgotten to remove it) - the offending piece has been removed.

I do not know what to do about 
_Redundant assignment of '*vaddr' to itself._
```
*vaddr = *vaddr;
```
This is a deliberate assignment using a volatile pointer to force a first touch without changing the memory. For me this is a false positive - is there any way of telling codacity to ignore it? I can add extra instructions to make it go away, but I like it this way and would prefer not to change it.

msimberg:
@biddisco I *think* the checks are done with `cppcheck` and there might be a way to suppress those warnings (special comments like with `clang-tidy`). I'd say don't worry right now since as you said it's deliberate, and the checks are only done on changed files (so we won't have that one pop on every PR).

Looks like you lost a license header while doing the latest changes. And I added a commit to exclude the `numa_allocator` test on CircleCI. Please don't force push over it!

biddisco:
> Looks like you lost a license header while doing the latest changes. And I added a commit to exclude the `numa_allocator` test on CircleCI. Please don't force push over it!

oops. I forgot about your circleci fix and I accidentally pushed a commit with another binder for executors that wasn't supposed to be part of that PR which is why the license header was missing.
I have force pushed again with your patch this time. Hope things pass after this apart from the codacity self assignment error.

hpx/parallel/util/numa_binding_allocator.hpp
```diff
@@ -0,0 +1,639 @@
...
+            }
+            else if (policy_ == threads::hpx_hwloc_membind_policy::membind_user)
+            {
+                result = reinterpret_cast<pointer>(
+                    threads::topology().allocate(n * sizeof(T)));
+                threads::hwloc_bitmap_ptr bitmap =
+                    threads::get_thread_manager().get_pool_numa_bitmap(
+                        binding_helper_->pool_name());
+                //
+                result = reinterpret_cast<pointer>(
```
msimberg: @biddisco not sure if you're aware, but `result` is still being assigned to twice here.

Issue #3718, opened 2019-02-24
Please make the python executable configurable through cmake


yurivict:
On FreeBSD several versions of python can be installed at the same time, with different executable names.

hkaiser:
@yurivict In general, HPX does not require any specific python executable to be built. We only install a python script `hpx_run.py` that can be helpful for running HPX applications. This script should be compatible with any Python version, so I don't think we need to make the Python executable configurable as this script will be used by the user in the end.

yurivict:
But you embed a specific executable name ```python```, and this fails.

hkaiser:
@yurivict Fair enough.

msimberg:
I think we have some old Python 2-style print statements hanging around, but otherwise nothing that would still require Python 2 (AFAIK). This should be a straightforward change if someone feels like having a go at this. And I don't think we need to support Python 2 anymore. It's time to let it go...

hkaiser:
@msimberg it shouldn't be a problem to change the code such it is compatible with any Python version.

diehlpk:
@hkaiser @msimberg pr #3761 provides the changes such that our scripts run with python3 on Fedora 29.

msimberg:
The question is: is there a need for the user to actually change the python executable, or is it enough that it works with both Python 2 and 3 and the user doesn't care which? @yurivict I'm guessing in your case it was actually the latter, and not that you necessarily need to be able to set the Python executable?

yurivict:
> is there a need for the user to actually change the python executable

The package should always work with a specific version of the python executable. Reasons: 1. predictability 2. you might not be able to guess what the executable is.

This is why there should be a configuration variable for it.

msimberg:
@yurivict does #3803 do what you had in mind?

msimberg:
@yurivict I'm closing this now that #3803 has been merged. Please reopen if needed.

PR #3803, opened 2019-04-17
Update hpxcxx to look at lib64 and use python3

stevenrbrandt:
Fixes issues using hpxcxx on the singularity image

## Proposed Changes

  - check explicitly for the hpx_application.pc file
  - use python3
  - check lib64 in addition to lib


cmake/HPX_GeneratePackage.cmake
```diff
@@ -53,6 +53,12 @@ foreach(dir ${HPX_LIBRARY_DIR})
   set(HPX_PKG_LIBRARY_DIR "${HPX_PKG_LIBRARY_DIR} -L${dir}")
 endforeach()
 
+find_program(FULL_PYTHON_CMD python3)
```
hkaiser: We already have `PYTHON_EXECUTABLE` set to the python interpreter found by the build system.

hkaiser: Also, we shouldn't explicitly limit things to Python V3.

stevenrbrandt:
@hkaiser it never limited things to Python v3, but it no longer matters because it now uses @PYTHON_EXECUTABLE@.

msimberg:
Is it a must to use the exact path of the Python executable? This means that the path will be system specific, which can be annoying for cross-compilation. Not a huge deal, but if it's not necessary I'd rather not hardcode it like that.

hkaiser:
#3718 was asking for exactly this. I'm not sure what's the best way out...

PR #3804, opened 2019-04-18
Fix latch

msimberg:
Add `notified_` member variable to `latch` to avoid `wait` returning before `count_down` has notified other threads. Details on the race in the commit message.

This seems to be the cause of the segfaults we've seen with the parallel algorithms after #3745.

This was the best I could do for now for avoiding taking the lock. It's still only taken twice: once for the notify and once when waiting. Ideas for avoiding that while keeping the behavior correct are welcome.

### DIFF
--- a/hpx/lcos/local/latch.hpp
+++ b/hpx/lcos/local/latch.hpp
@@ -55,7 +55,7 @@ namespace hpx { namespace lcos { namespace local
         /// Postconditions: counter_ == count.
         ///
         explicit latch(std::ptrdiff_t count)
-          : mtx_(), cond_(), counter_(count)
+          : mtx_(), cond_(), counter_(count), notified_(count == 0)
         {
         }
 
@@ -89,7 +89,10 @@ namespace hpx { namespace lcos { namespace local
             std::unique_lock<mutex_type> l(mtx_.data_);
             HPX_ASSERT(counter_ > 0);
             if (--counter_ == 0)
+            {
+                notified_ = true;
                 cond_.data_.notify_all(std::move(l));    // release the threads
+            }
             else
                 cond_.data_.wait(l, "hpx::local::latch::count_down_and_wait");
         }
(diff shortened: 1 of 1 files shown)


hpx/lcos/local/latch.hpp
```diff
@@ -134,12 +138,9 @@ namespace hpx { namespace lcos { namespace local
         ///
         void wait() const
         {
-            if (counter_.load(std::memory_order_acquire) > 0)
-            {
-                std::unique_lock<mutex_type> l(mtx_.data_);
-                if (counter_.load(std::memory_order_relaxed) > 0)
-                    cond_.data_.wait(l, "hpx::local::latch::wait");
-            }
+            std::unique_lock<mutex_type> l(mtx_.data_);
```
hkaiser: Can we do a double-test still? I.e. once before grabbing the lock, just testing the counter and then, if necessary grab the lock and perform a repeated test under the lock?

msimberg: Hmm, don't think so. Just testing `counter_ > 0` outside the lock would get us to the same situation as before. Testing `counter_ > 0 || !notified` is significantly less racey, but still racey. Do you mean something else?

hkaiser: Makes sense, thanks!

msimberg:
Hmm... there's more to this: http://cdash.cscs.ch/testDetails.php?test=7767107&build=50026. I also need to fix the assert in `reset`, so no merging, please.

msimberg:
@hkaiser after these fixes I can't reproduce the segfaults anymore, even though I've seen two happen on pycicle. I think this should still be a strict improvement (at least in correctness, hopefully not a big regression in performance) and try to figure out if more segfaults still happen on rostam. Anything against merging this?

hkaiser:
> Anything against merging this?

@msimberg please go ahead.

PR #3806, opened 2019-04-18
Moving include path configuration to before APEX

### LINKED ISSUE #3805: HPX won't compile with  -DHPX_WITH_APEX=TRUE
## Expected Behavior

HPX should build

## Actual Behavior

```
[  0%] Building CXX object apex/src/apex/CMakeFiles/apex.dir/apex.cpp.o
cd /home/users/khuck/src/phylanx/tools/buildbot/build-delphi-x86_64-Linux-gcc/hpx-Debug/apex/src/apex && /packages/gcc/7.1/bin/g++  -DAPEX_HAVE_ACTIVEHARMONY -DAPEX_HAVE_HPX_CONFIG -DAPEX_HAVE_OTF2 -DAPEX_HAVE_PAPI -DAPEX_HAVE_POWERCAP_POWER -DAPEX_HAVE_PROC -DBOOST_ALL_NO_LIB -DDEBUG -DHPX_COROUTINE_EXPORTS -DHPX_EXPORTS -DHPX_LIBRARY_EXPORTS -D_GNU_SOURCE -I/home/users/khuck/src/phylanx/tools/buildbot/src/hpx -isystem /usr/local/packages/boost/1.65.0-gcc7/include -I/home/users/khuck/src/phylanx/tools/buildbot/src/hpx/apex/src/apex -I/home/users/khuck/src/phylanx/tools/buildbot/build-delphi-x86_64-Linux-gcc/hpx-Debug -I/home/users/khuck/src/phylanx/tools/buildbot/build-delphi-x86_64-Linux-gcc/hpx-Debug/apex/src/apex -I/home/users/khuck/src/phylanx/tools/buildbot/src/hpx/apex/src/contrib -I/include -I/usr/local/packages/papi/papi-knl/5.5.0/include -I/usr/local/packages/activeharmony/4.6.0-gcc/include -I/usr/local/packages/otf2-2.1/include  -fPIC -g   -fPIC -std=c++17 -o CMakeFiles/apex.dir/apex.cpp.o -c /home/users/khuck/src/phylanx/tools/buildbot/src/hpx/apex/src/apex/apex.cpp
In file included from /home/users/khuck/src/phylanx/tools/buildbot/src/hpx/hpx/config.hpp:31:0,
                 from /home/users/khuck/src/phylanx/tools/buildbot/src/hpx/apex/src/apex/apex.cpp:12:
/home/users/khuck/src/phylanx/tools/buildbot/build-delphi-x86_64-Linux-gcc/hpx-Debug/hpx/config/version.hpp:15:10: fatal error: hpx/pp/cat.hpp: No such file or directory
 #include <hpx/pp/cat.hpp>
          ^~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [apex/src/apex/CMakeFiles/apex.dir/apex.cpp.o] Error 1
```

configure hpx with -DHPX_WITH_APEX=TRUE 

## Specifications

any platform, any compiler

khuck:
Fixes #3805

## Proposed Changes

Move include path configuration to earlier in the file so that APEX has the include paths set correctly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1619,6 +1619,12 @@ if((NOT HPX_WITH_APEX) AND HPX_WITH_ITTNOTIFY)
 endif()
 
 ################################################################################
+# Set basic search paths for the generated HPX headers
+################################################################################
+include_directories("${CMAKE_BINARY_DIR}/include")
+include_directories("${CMAKE_BINARY_DIR}")
+
+################################################################################
 # Enable integration with Apex event counters
 ################################################################################
 
@@ -1810,11 +1816,6 @@ function(generate_config_defines_header TARGET_DIRECTORY)
     NAMESPACE default
     FILENAME "${TARGET_DIRECTORY}/hpx/config/defines.hpp")
 endfunction()
-
-################################################################################
-# Set basic search paths for the generated HPX headers
-################################################################################
-include_directories("${CMAKE_BINARY_DIR}")
 
 ################################################################################
 # Add custom targets for tests


hkaiser:
Thanks @khuck!

@msimberg, @sithhell: looks like we should change the new module directories shouldn't be linked to `<build_dir>/include/hpx` but to `<build_dir>/hpx` instead.

msimberg:
Thanks @hkaiser and @khuck!

> @msimberg, @sithhell: looks like we should change the new module directories shouldn't be linked to `<build_dir>/include/hpx` but to `<build_dir>/hpx` instead.

Yeah, that would probably be a good idea. @khuck I guess this does the job for now?

sithhell:
By adding the `hpx.pp` module the include paths get set correctly. Since APEX depends on those (indirectly through config), I think the correct way of fixing this is to include the modules as dependencies instead of globally setting include directories (see https://github.com/khuck/xpress-apex/pull/99)

msimberg:
IMO indirect dependencies should get taken care of implicitly by direct dependencies. I.e. in this case since apex only indirectly depends on `pp`, the `hpx` target should be exposing that dependency correctly which it might not be doing right now. I think this case is trickier as well because of the circular dependency between apex and hpx.

sithhell:
IIUC, the only dependency here is for the config module (which doesn't exist yet), once we have that, there shouldn't be an indirect dependency anymore. All in all, I agree. The transition period is hard...

msimberg:
I think this is a special case because of the circular dependency. Setting `hpx` as a dependency *should* really be enough, and if it's not we need to make sure it is. In this case though we might not be able to avoid setting global `include_directories`.

sithhell:
I don't think it's a special case. There's no circular dependency in this case. All that's missing is the dependency that's there so far.

hkaiser:
We can close this as #3811 will help fixing this properly.

PR #3809, opened 2019-04-19
Adding HPX Library "Config"

sithhell:
Further step to modularize HPX. This puts the main config macros into
their own module. The cmake scripts have been streamlined to further
depend on target specific include directories.

Flyby: Fixing library linkage mode


libs/config/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+add_subdirectory(examples)
+add_subdirectory(src)
+add_subdirectory(tests)
+
+install(
+  DIRECTORY include/hpx
+  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/hpx
+  COMPONENT config)
+
+message(STATUS "config: Configuring done")
```
hkaiser: We should collect all header files for this module such that those are being added to the generated VS solution files. This also means we have to properly create a source-group for those.

The main problem with source groups in cmake is, that the `source_group` has to be created in the same (cmake) directory as the corresponding `add_library`/`add_executable` is located (see also #3810).

libs/config/include/hpx/config/version.hpp
```diff
@@ -89,4 +91,6 @@ namespace hpx
     }
 #endif
 
+/// \endcond
+
 #endif
```
hkaiser: What's the rationale to not generate the version.hpp file anymore?

libs/config/src/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+target_include_directories(hpx.config PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../include>
+  $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}>)
+
+include(HPX_AddSourceGroup)
+add_hpx_source_group(NAME hpx.config ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../include/hpx
+  CLASS "Header Files" TARGETS ${headers})
+add_hpx_source_group(NAME hpx.config ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../include/hpx
+  CLASS "Source Files" TARGETS ${sources})
+
```
hkaiser: Ok, I see the source_group definitions now, please (partially) disregard my previous comment. Unfortunately, this will not properly define the source groups as 

a) the `source_group` invocation has to happen in the same directory as the corresponding `add_library` is located
b) all files listed in a source group must be explicitly added to that `add_library`

msimberg: @hkaiser could you clarify the problem. To my eyes both of those conditions are fulfilled above, so I must be missing something.

hkaiser: @msimberg no, unfortunately none of those conditions are met. 

a) The `source_group` call needs to happen in the same cmake scope as the corresponding `add_library` the files of the source group are part of (see Brad's answer here: https://gitlab.kitware.com/cmake/cmake/issues/18856)

b) The files added to the source group have to be passed to the corresponding `add_library` call. Currently, the files are not directly passed to the `add_library` invocation, see: https://github.com/STEllAR-GROUP/hpx/blob/master/src/CMakeLists.txt#L229-L244.

The best is to have a look at how we add our Natvis files (the VS debugger visualizers) to the generated VS solution:
image
That happens here: https://github.com/STEllAR-GROUP/hpx/blob/master/src/CMakeLists.txt#L88-L91.

sithhell:
Let's wait for this until #3811 has been merged.

msimberg:
Closing in favor of a cleaned up version coming up soon.

PR #3811, opened 2019-04-19
Clean up PP module and library skeleton

### LINKED ISSUE #3810: Source groups are not created for files in modules
The new module-oriented cmake scripts do not properly place the files into source groups. This prevents the header files from showing up in the generated VS solution files. The source files are all placed into the same root solution folder. That will greatly reduce usability as eventually most of the files will be placed into the same solution folder...

msimberg:
Attempt to clean up and address some of the issues in our modules for the upcoming release. Please comment here if you find other problems.

- Require Python 3 in `create_library_skeleton.py`. This is because it uses format strings. Since this is only for internal use, I think it's fine to require 3 here.
- Rename `hpx.pp` to `hpx_preprocessor` for consistency (the underscore) and understanding (full name). I haven't changed the names of the header directories, but will do so if we agree on that.
- Use rst for all READMEs and clean up some of the markup (to be consistent with the main docs).
- Put generated headers and symlinks to libs headers under `include/generated` and `include/libs` to have a common place for them.
- Attempts to fix source groups. (Fixes #3810.)

This hopefully includes all the fixes from #3809 without the `config` modules. Also probably replaces #3806.


CMakeLists.txt
```diff
@@ -1604,6 +1604,12 @@ endif()
 hpx_libraries(${HWLOC_LIBRARIES})
 include_directories(${HWLOC_INCLUDE_DIR})
 
+################################################################################
+# Set search paths for generated and libs headers
+################################################################################
+include_directories("${CMAKE_BINARY_DIR}/include/generated")
+include_directories("${CMAKE_BINARY_DIR}/include/libs")
```
sithhell: The main reason I am reluctant to have those here is that "hidden" dependencies are masked that way. We won't notice if we forgot to add some dependencies while developing our new modules.
This is exactly the hack which renders the whole effort to be ignored. Without those lines, we have to invest something now, rather than invest heavily again once we decide to remove them, because we think everything is done.

msimberg: You're absolutely right. I'm not even sure the libs ones are needed right now. I'll try removing them. The generated ones I'm not sure about. I'd like to see them gone as well from the global includes but not sure how. APEX finds e.g. the version header only because of this right now. Suggestions on how to avoid that?

msimberg:
@hkaiser I tried to do something for the source groups. Could you give it a try when you have time?

CMakeLists.txt
```diff
@@ -1604,6 +1604,11 @@ endif()
 hpx_libraries(${HWLOC_LIBRARIES})
 include_directories(${HWLOC_INCLUDE_DIR})
 
+################################################################################
+# Set search paths for generated headers.
+################################################################################
+include_directories("${CMAKE_BINARY_DIR}/include/generated")
```
hkaiser: Befor @khuck's change for APEX we had just:
```
include_directories("${CMAKE_BINARY_DIR}")
```
so that HPX could find its own generated files (`hpx/config/defines.hpp`, etc.). How are those found now?

msimberg: Good observation. I moved the generated files to `include/generated` to have them sit alongside `include/libs` which held the symlinks to headers from our modules. But I've removed the symlinks so I could move the generated files back to where they were. @sithhell did you have some use case in mind with the symlinks that I've now broken? I've added the module include directories to the build-tree CMake and pkgconfig files which as far as I can see would've been the only use case for the symlinks.

hkaiser:
@msimberg also, should we add #3813 to this as well?

hkaiser:
> I tried to do something for the source groups. Could you give it a try when you have time?

@msimberg A quick try shows that whatever you did has not had any positive effect on the source group issue. I will have a closer look later.

msimberg:
I moved the generated headers back to where they were. The only thing that now still doesn't work is APEX integration. We a few options, none of which are good, but we'll just have to pick one.
1. Add back the collected symlinked headers for modules and add a global `include_directories` to point to there (as in #3806).
2. Add global `include_directories` only for the parts that APEX needs as a special case. This is least disruptive in both HPX and APEX.
3. Add `hpx_preprocessor` as a dependency to APEX. This sounds bad at first glance, but considering APEX depends on a small subset of HPX (I think) the list of modules it'd have to depend on shouldn't grow very large.

hkaiser:
@msimberg I think we should go with option 3 (add dependencies to APEX). That is the least intrusive and most cmake-like variant.

hkaiser:
@msimberg do you still plan to rename the files (pp -> preprocessor)?

msimberg:
> @msimberg I think we should go with option 3 (add dependencies to APEX). That is the least intrusive and most cmake-like variant.

Yeah, that's fair enough. It's not 100% right because APEX should depend on the `hpx` target which would give `hpx_preprocessor` for free, but it doesn't because of the circular dependency. Let's go with this though.

> @msimberg do you still plan to rename the files (pp -> preprocessor)?

Yeah, I'll do it. We've already broken compatibility by moving them so I think it makes sense. The last thing I'm _not_ going to change though is the macro names.

hkaiser:
>  It's not 100% right because APEX should depend on the hpx target which would give hpx_preprocessor for free, but it doesn't because of the circular dependency. Let's go with this though.

I rather see at as APEX being one of the modules we will create in the future.

> The last thing I'm not going to change though is the macro names.

Fair enough!

msimberg:
> > It's not 100% right because APEX should depend on the hpx target which would give hpx_preprocessor for free, but it doesn't because of the circular dependency. Let's go with this though.
> 
> I rather see at as APEX being one of the modules we will create in the future.

I think we agree! I meant "should" more in the sense that it should right now but, yes, ideally in the future it wouldn't.

libs/preprocessor/README.rst
```diff
@@ -0,0 +1,20 @@
...
+This library contains useful preprocessor macros:
+
+* ``HPX_PP_CAT``: Concatenate two tokens
+* ``HPX_PP_EXPANDS``: Expands a preprocessor token
+* ``HPX_PP_NARGS``: Determines the number of arguments passed to a variadic macro
+* ``HPX_PP_STRINGIZE``: Turns a token into a string
+* ``HPX_PP_STRIP_PARENS``: Strips parenthesis from a token
+
+Extensive documentation can be found `here
+<https://stellar-group.github.io/hpx/docs/sphinx/latest/html/libs/pp/docs/index.html>`__.
```
hkaiser: This is not true ;-)

msimberg: Damn it, I changed it to say just `Documentation can be found...`. Even that's a bit of an overstatement. I promise that's the last thing I'll change and then I think we can merge this.

PR #3813, opened 2019-04-19
Make the deprecation warnings issued by the old pp headers optional

hkaiser:
This is probably something that should be added to all modules, possibly generically.


msimberg:
Wouldn't it make more sense to have the warnings be controlled by `HPX_WITH_DEPRECATION_WARNINGS` and have `HPX_WITH_PREPROCESSOR_COMPATIBILITY_HEADERS` control whether the headers are included at all?

hkaiser:
> Wouldn't it make more sense to have the warnings be controlled by `HPX_WITH_DEPRECATION_WARNINGS`

I thought about this and was not sure whether this would be too coarse grain in the end. However, I think I'm impartial to this.

> and have HPX_WITH_PREPROCESSOR_COMPATIBILITY_HEADERS control whether the headers are included at all?

What do you have in mind? Wrapping all of the contents of the compatibility headers into such a guard? I.e. disabling pulling in the correct header files? Hmmm. not sure if this wouldn't be too radical.

msimberg:
> > Wouldn't it make more sense to have the warnings be controlled by `HPX_WITH_DEPRECATION_WARNINGS`
> 
> I thought about this and was not sure whether this would be too coarse grain in the end. However, I think I'm impartial to this.

Perhaps. But in the end an end user should deal with all deprecations as soon as possible, no matter where they come from.

> > and have HPX_WITH_PREPROCESSOR_COMPATIBILITY_HEADERS control whether the headers are included at all?
> 
> What do you have in mind? Wrapping all of the contents of the compatibility headers into such a guard? I.e. disabling pulling in the correct header files? Hmmm. not sure if this wouldn't be too radical.

Something like that, yes. In the end it's like any other compatibility layers we've had.

In this particular case I'm not even sure we would need the compatibility headers. The preprocessor headers were in a `detail` directory after all so if someone depended on them they've asked for trouble. There'll be other cases where we do need them though, so it's good to sort this out now.

msimberg:
> What do you have in mind? Wrapping all of the contents of the compatibility headers into such a guard? I.e. disabling pulling in the correct header files? Hmmm. not sure if this wouldn't be too radical.

By the way, why do you think this would be too radical? I might not be thinking this through properly.

For `pp` though I would honestly just backtrack and remove the compatibility headers.

hkaiser:
> For pp though I would honestly just backtrack and remove the compatibility headers.

Remove them now? I think we should leave them in place for at least a release.

msimberg:
> > For pp though I would honestly just backtrack and remove the compatibility headers.
> 
> Remove them now? I think we should leave them in place for at least a release.

Yep :) for the reason I mentioned earlier (they were in a detail directory). But I'm also impartial here. We just need to decide on something before the release. If we leave them I'd go with having the deprecation warnings controlled by `HPX_WITH_DEPRECATION_WARNINGS` and have the headers be conditionally included so that we can turn it off and then remove them two releases from now.

hkaiser:
> If we leave them I'd go with having the deprecation warnings controlled by HPX_WITH_DEPRECATION_WARNINGS and have the headers be conditionally included so that we can turn it off and then remove them two releases from now.

Fine with me.

msimberg:
@hkaiser I've tried to incorporate all of the above (there's also an option to turn off the deprecation warning just for `preprocessor`). I couldn't find a cleaner way to exclude the compatibility headers within the project, so ideas are welcome for that.

PR #3814, opened 2019-04-21
Allow setting default scheduler mode via the configuration database

hkaiser:


src/runtime/threads/policies/scheduler_base.cpp
```diff
@@ -51,6 +52,18 @@ namespace hpx { namespace threads { namespace policies
...
+        // if there is a cfg setting for the default_scheduler_mode, use that
+        // instead of the built-in default
+        std::string default_scheduler_mode = hpx::get_config_entry(
+            "hpx.default_scheduler_mode", std::string());
+
+        if (!default_scheduler_mode.empty())
+        {
+            mode = scheduler_mode(hpx::util::safe_lexical_cast<std::size_t>(
+                default_scheduler_mode));
+            HPX_ASSERT((mode & ~scheduler_mode::all_flags) == 0);
```
msimberg: A message would probably be good here saying that the config entry contains illegal scheduler options.

hkaiser: Yah, I was thinking of throwing an exception, I might just do that instead.

msimberg: Yeah, that works. I was thinking of `HPX_ASSERT_MSG` but either is fine with me.

msimberg:
@hkaiser just added a short message to the assertion. I suppose you're done with this otherwise? We can merge if the build succeeds.

hkaiser:
> @hkaiser just added a short message to the assertion. I suppose you're done with this otherwise? We can merge if the build succeeds.

Yah looks good, sorry - didn't have time to look into this.

hpx/parallel/util/numa_allocator.hpp
```diff
@@ -124,6 +124,11 @@ namespace hpx { namespace parallel { namespace util
             }
             hpx::wait_all(first_touch);
 
+            for (auto && f : first_touch)
+            {
+                f.get();        // rethrow exceptions
+            }
+
```
msimberg: I think this officially exposes the brokenness of our NUMA handling(?) on CircleCI (https://circleci.com/gh/STEllAR-GROUP/hpx/86317). #3802 also had problems with this. We might just have to disable or remove this example and leave it to @biddisco to replace this `numa_allocator` with his new and improved one ;)

PR #3820, opened 2019-04-23
Assorted CMake fixes

### LINKED ISSUE #3817: HPX_INCLUDE_DIRS contains non-existent directory
Working on Kokkos+HPX. In `HPXConfig.cmake`, the following appears in the list of HPX_INCLUDE_DIRS, but it doesn't actually get installed or created.
````
/Users/jjwilke/Programs/install/hpx/include/hpx/external
````
from
````
set(HPX_INCLUDE_DIRS "/Users/jjwilke/Programs/install/hpx/include;/Users/jjwilke/Programs/install/hpx/include/hpx/external;/opt/local/include;/Users/jjwilke/Programs/spack/opt/spack/darwin-highsierra-x86_64/clang-10.0.0-apple/hwloc-2.0.2-jwtizf4qjcn7gh2p2eho5vbrldqg6mbj/include")
````
This actually doesn't cause errors (well, warnings) until I build a kokkos-dependent project. 
````
  Imported target "kokkos" includes non-existent path

    "/Users/jjwilke/Programs/install/hpx/include/hpx/external"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
````
Configure still succeeds, but it would be better not to have the warning

### PR DESCRIPTION
**Pull request by msimberg**
Attempts to fix #3817 and #3819.

### CODE REVIEW

File: libs/preprocessor/CMakeLists.txt [hunk 1]
```diff
@@ -29,7 +29,7 @@ set(headers
   ${CMAKE_CURRENT_SOURCE_DIR}/include/hpx/util/detail/pp/strip_parens.hpp
 )
 
-add_library(hpx_preprocessor STATIC ${headers})
+add_library(hpx_preprocessor ${hpx_libs_link_mode} ${headers})
```
hkaiser: Shouldn't modules always be linked statically?

msimberg: Need not after this change... ;) We could keep it as it was, but not sure if there's any reason to restrict it to always be linked statically.

I'll revert this for now so we don't end up with any funny surprises from this for the release. Statically linking is pretty safe...

hkaiser: @msimberg My only concern would be that we end up with dozens of very small shared libraries that will (possibly) significantly slow down startup on large distributed runs.

msimberg: Ok, that could actually end up being a problem. So definitely staying with static for now. Thanks for noticing!

PR #3821, opened 2019-04-24
Disable barrier_3792 test when networking is disabled

msimberg:
The test doesn't make much sense with only one locality (and it hangs) so we disable it.

### DIFF
--- a/tests/regressions/lcos/CMakeLists.txt
+++ b/tests/regressions/lcos/CMakeLists.txt
@@ -11,7 +11,6 @@ set(tests
     async_callback_non_deduced_context
     async_unwrap_1037
     barrier_hang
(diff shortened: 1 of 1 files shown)


msimberg:
@hkaiser I don't think so, although they're a bit inconsistent. Some tests just "work" (usually do nothing) with networking disabled, and some are explicitly disabled. That could of course be cleaned up as well, but I'll leave that for another time. This one (`barrier_3792`) is causing pycicle builder with networking disabled to time out, so if it still times out there's probably more (that also means please don't merge before we see that that one is back).

msimberg:
All right, pycicle looks good for this (now I just need to figure out what's wrong with the clang 4 build...).

PR #3822, opened 2019-04-24
Change includes in hpx_wrap.cpp to more specific includes

msimberg:
Including `hpx/hpx.hpp` pulls in unnecessary dependencies.

### DIFF
--- a/src/hpx_wrap.cpp
+++ b/src/hpx_wrap.cpp
@@ -1,4 +1,4 @@
-//  Copyright (c) 20018 Nikunj Gupta
+//  Copyright (c) 2018 Nikunj Gupta
 //
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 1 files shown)


src/hpx_wrap.cpp
```diff
@@ -23,10 +23,13 @@ namespace hpx_start
     HPX_SYMBOL_EXPORT bool include_libhpx_wrap __attribute__((weak)) = false;
 }
 
+#include <hpx/hpx_finalize.hpp>
 #include <hpx/hpx_init.hpp>
-#include <hpx/hpx.hpp>
+#include <hpx/runtime/runtime_mode.hpp>
+#include <hpx/util/function.hpp>
 
 #include <iostream>
```
hkaiser: Do we need `<iostream>` for this file? Looks like a left-over.

msimberg: Good catch! Probably not.

msimberg: Removed.

Issue #3646, opened 2019-02-01
Parallel algorithms should accept iterator/sentinel pairs


hkaiser:
Our parallel algorithms currently expect for the begin/end iterator pairs to be of the same type. This is an unnecessary limitation. We should relax that. 

Here is the list of algorithms that need to be adapted:

- [x] `adjacent_difference` (#5564)
- [x] `adjacent_find`  (#4996)
- [x] `all_of` `any_of` `none_of`
- [x] `copy` `copy_if` (#4821)
- [x] `move`
- [x] `count` `count_if`
- [x] `equal` (#4883) `mismatch` (#4884)
- [x] `exclusive_scan` (#5436)
- [x] `inclusive_scan` (#5439)
- [x] `reduce` (see #3650, #4836)
- [x] `transform` (#4855, #4888, #5051)
- [x] `fill`  (#4860)
- [x] `find` `find_end` `find_first_of` `find_if` `find_if_not`  (#4885)
- [x] `for_each`  `for_each_n` (#4867)
- [x] `for_loop` `for_loop_strided` `for_loop_n` `for_loop_strided_n` (#4980)
- [x] `generate` `generate_n` (#4896)
- [x] `is_heap` `is_heap_until` (#4965)
- [x] `is_partitioned` (#5006)
- [x] `is_sorted` `is_sorted_until` (#5025)
- [x] `lexicographical_compare` (#5350)
- [x] `max_element` `min_element` `minmax_element` (#5241)
- [x] `make_heap` (#4964)
- [x] `sort` (#5460)
- [x] `partial_sort` `partial_sort_copy` (#5630)
- [x] `nth_element` (#5592)
- [x] `stable_sort` (see #4803)
- [x] `partition` `partition_copy` `stable_partition` (#5540)
- [x] `remove` `remove_if` (#5125)
- [x] `remove_copy` `remove_copy_if` (#5150)
- [x] `replace` `replace_copy` `replace_copy_if` `replace_if` (#5192)
- [x] `reverse` `reverse_copy` (#5225)
- [x] `rotate` `rotate_copy` (#5459)
- [x] `search` (#5066)
- [x] `set_difference` `set_intersection` `set_symmetric_difference` `set_union` `includes` (#4970)
- [x] `merge` `inplace_merge` (#4970)
- [x] `swap_ranges` (#5449)
- [x] `uninitialized_copy` `uninitialized_copy_n` (#5371)
[...]

PR #3829, opened 2019-04-27
lexicographic_compare accepts different iterator types (#3646)

wteiwewte:


msimberg:
@wteiwewte you might want to check either the implementation or the test: https://circleci.com/gh/STEllAR-GROUP/hpx/85740 ;)

wteiwewte:
I checked tests and didn't find any bugs. On my fork those tests are passing. What factors can contribute to different execution on ci ?

hkaiser:
@weilewei the clang compiler on CircleCI has vectorization enabled by default (even for Debug builds). I'm not saying that this is the reason for the issues we're seeing, we just ran into this in different contexts before...

msimberg:
@wteiwewte what compiler are you using? I see from our CI that the test fails with clang 4, clang 8, and gcc 9, but not with gcc 5. Is it possible that you're using some version of gcc older than 9? You could also use the docker image we use on CircleCI to get the exact same environment (`docker run -it stellargroup/build_env:ubuntu`).

PR #3831, opened 2019-05-02
Fix CircleCI config for modules

msimberg:
Add libs tests to CircleCI config.

Flyby: use `hpx_option` for all options in modules.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -772,6 +772,31 @@ jobs:
       - store_artifacts:
           path: tests.unit.util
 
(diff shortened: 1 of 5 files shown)


msimberg:
> I assume `hpx_libs_link_mode` is not used anywhere anymore, LGTM.

Yep. It used to let one create dynamically linked modules, but since we're sticking to `STATIC` I'm removing it. We can always reintroduce it if needed. Sorry about the sneaky force-pushes, by the way.

PR #3833, opened 2019-05-02
make latch counter_ protected for deriving latch in hpxmp

tianyi93:


hkaiser:
@tianyi93 After thinking about this some more, we should have made all members `protected`, if for nothing else but consistency...

tianyi93:
@hkaiser Yes, I will do that.

PR #3834, opened 2019-05-03
Disable transpose_block_numa example on CircleCI

msimberg:

This call to hwloc: https://github.com/STEllAR-GROUP/hpx/blob/8a8cec8dee6f0dc12200c124b6cb3bf43264bc27/src/runtime/threads/topology.cpp#L522-L528 returns `-1` and sets `errno = EPERM` (i.e. permission denied) on CircleCI. My guess is that this hasn't come up before because of recent changes to exception handling (#3662, #3814), or CircleCI launching containers differently. Since we have pycicle running as well without containers I think we can safely disable the example on CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -232,7 +232,12 @@ jobs:
           when: always
           command: |
               ulimit -c unlimited
-              ctest -T test --no-compress-output --output-on-failure -R tests.examples
+              # NOTE: transpose_block_numa is disabled because
+              # hwloc_get_area_membind_nodeset (which is used by the
+              # numa_allocator) fails with EPERM.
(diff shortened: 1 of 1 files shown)


biddisco:
Would it be a good idea to add some kind of cmake check that detects if we are inside a docker container and then we can disable certain tests in a more formal way
https://stackoverflow.com/questions/20010199/how-to-determine-if-a-process-runs-inside-lxc-docker

hkaiser:
> Would it be a good idea to add some kind of cmake check that detects if we are inside a docker container and then we can disable certain tests in a more formal way

Let's do this. I'd suggest to leave it for a later PR, though as it needs more investigation. I think we should go ahead and merge this PR now as a stopgap measure.

PR #3835, opened 2019-05-03
make all latch members protected for consistency

tianyi93:
make all latch members protected for consistency


msimberg:
@hkaiser can we ignore the codacity check? It doesn't show any details for me. Not sure if I need to be logged in for me to see something.

hkaiser:
@msimberg yes, I wanted to try Codacy to see what this offers. It looks somewhat useful to me, but we can always remove the checking if it gets in the way.

msimberg:
Sure, I'm all for more checks. Let's try it and see what it gives as. And now the check green anyway, so I guess it just wasn't set up properly yet or something like that.

PR #3843, opened 2019-05-06
This fixes the launch_process/launched_process pair of tests

hkaiser:
- we have to make sure networking stays enabled for both parts of the test


msimberg:
Build on rostam: http://rostam.cct.lsu.edu/builders/hpx_gcc_6_boost_1_64_centos_x86_64_debug/builds/306. Thanks @hkaiser!

PR #3845, opened 2019-05-07
Assertion module

msimberg:
Based on #3628. Depends on #3832.

Open questions from the previous PR:
- Changes to assertion functionality: I've kept the changes as they were in the previous PR, but I'll change it back to the way it is on master if this is still too big to review.

Resolved(?) from previous PR:
- The module is now named `assertion` (including headers and namespaces).
- I removed the changes to the order in which information is printed for exceptions and assertions. I think the problem here is the sometimes massive lists of environment variables. We could make that optional output? In a separate PR in any case.
- Changed the tests to use the usual `add_hpx_executable` (with `NOLIBS` and `NO_HPXINIT`) and `add_hpx_unit_test`.
- The header tests are generated.

Other changes:
- Turn on PIC for the static libraries linked against the main `hpx` library. This is a requirement at least on Linux/x86_64. Perhaps not on Windows? It's unconditionally on at the moment.


msimberg:
By the way, inspect is complaining about `hpx/assertion.hpp` missing in `hpx/assertion.hpp` because it defines `HPX_ASSERT`. I don't want to ignore the whole file, and I'm not sure how to best deal with these situations yet, so I'll ignore it completely. I'm not convinced it's maintainable to keep adding include checks for our own headers, since it'll never be complete anyway and a pain to update. I won't remove it either because it's already been added.

hkaiser:
> Turn on PIC for the static libraries linked against the main hpx library. This is a requirement at least on Linux/x86_64. Perhaps not on Windows? It's unconditionally on at the moment.

Windows does not know anything about PIC.

hkaiser:
> By the way, inspect is complaining about hpx/assertion.hpp missing in hpx/assertion.hpp because it defines HPX_ASSERT.

Use `// hpxinspect:noinclude:HPX_ASSERT` to suppress only the check for `HPX_ASSERT`

msimberg:
> > Turn on PIC for the static libraries linked against the main hpx library. This is a requirement at least on Linux/x86_64. Perhaps not on Windows? It's unconditionally on at the moment.
> 
> Windows does not know anything about PIC.

Gotcha.

> > By the way, inspect is complaining about hpx/assertion.hpp missing in hpx/assertion.hpp because it defines HPX_ASSERT.
> 
> Use `// hpxinspect:noinclude:HPX_ASSERT` to suppress only the check for `HPX_ASSERT`

Argh, thank you!

aurianer:
This is good to go for me :) I applied your suggested changes (also for HPX_ASSERT_MSG) and I made some forgotten changes too.

libs/assertion/CMakeLists.txt
```diff
@@ -0,0 +1,130 @@
...
+target_include_directories(hpx_assertion PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+  $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}>)
+
+if(HPX_ASSERTION_WITH_COMPATIBILITY_HEADERS)
+  target_include_directories(hpx_assertion PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include_compatibility>)
+endif()
+
+include(HPX_AddSourceGroup)
```
hkaiser: We need to add common COMPILE_DEFINITIONS here (see #3878 for an example)

hkaiser:
@aurianer please rebase this on master as it might remove some of the changes that have been merged by now from this PR.

hkaiser:
@aurianer there are (new) linker failures on pycicle. Could you have a look, please?

aurianer:
Back in the game ;)

libs/assertion/include/hpx/assertion.hpp
```diff
@@ -0,0 +1,86 @@
+//  Copyright (c) 2019 Thomas Heller
+//
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+//  Make HPX inspect tool happy: hpxinspect:noassert_macro
+//                               hpxinspect:noinclude:HPX_ASSERT
```
hkaiser: This is duplicated (see lines below)

libs/assertion/include/hpx/assertion.hpp
```diff
@@ -0,0 +1,86 @@
...
+#define HPX_ASSERT_MSG_(expr, msg, ...) msg
+/// \encond
+
+#if defined(HPX_DEBUG)
+#define HPX_ASSERT(...)                                                        \
+    HPX_ASSERT_(HPX_ASSERT_EXPR_(__VA_ARGS__, ""),                             \
+        HPX_ASSERT_MSG_(__VA_ARGS__, std::string(), ""))                       \
+    /**/
+
+#define HPX_ASSERT_MSG(...) HPX_ASSERT(__VA_ARGS__)
```
hkaiser: Should we really make `HPX_ASSERT` and `HPX_ASSERT_MSG` equivalent?

libs/assertion/include/hpx/assertion/current_function.hpp
```diff
@@ -0,0 +1,59 @@
...
+//  See accompanying file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt
+//
+//  http://www.boost.org/libs/assert
+//
+//  Adapted to HPX naming scheme
+
+namespace hpx { namespace assertion { namespace detail {
+
+    inline void current_function_helper()
```
hkaiser: Could this function be `constexpr`?

libs/assertion/include_compatibility/hpx/util/assert.hpp
```diff
@@ -0,0 +1,26 @@
...
+//  Copyright (c) 2001, 2002 Peter Dimov and Multi Media Ltd.
+//  Copyright (c) 2007 Peter Dimov
+//  Copyright (c) Beman Dawes 2011
+//
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+//  Make HPX inspect tool happy: hpxinspect:noassert_macro
+//                               hpxinspect:noinclude:HPX_ASSERT
+//                               hpxinspect:nodeprecatedname:BOOST_ASSERT
```
hkaiser: Are those comments still needed?

libs/assertion/src/assertion.cpp
```diff
@@ -0,0 +1,48 @@
...
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/assertion.hpp>
+
+#include <iostream>
+#include <string>
+
+namespace hpx { namespace assertion {
+    namespace {
```
hkaiser: What's the rationale of having this in an unnamed namespace (and not in `namespace detail` for instance)?

libs/assertion/src/assertion.cpp
```diff
@@ -0,0 +1,48 @@
...
+        assertion_handler& get_handler()
+        {
+            static assertion_handler handler = nullptr;
+            return handler;
+        }
+    }
+
+    void set_assertion_handler(assertion_handler handler)
```
hkaiser: Should we export the functions in this file?

libs/assertion/src/source_location.cpp
```diff
@@ -0,0 +1,17 @@
...
+//
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/assertion/source_location.hpp>
+
+#include <iostream>
+
+namespace hpx { namespace assertion {
+    std::ostream& operator<<(std::ostream& os, source_location const& loc)
```
hkaiser: Do we have to export this function?

libs/assertion/tests/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+# Copyright (c) 2019 The STE||AR-Group
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+if (NOT HPX_WITH_TESTS)
```
hkaiser: We do more nowadays... (see for instance here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/preprocessor/tests/CMakeLists.txt#L6)

libs/assertion/include_compatibility/hpx/util/assert.hpp
aurianer: In this specific file, we don't really mind because it will be removed soon (just for compatibility of headers), but I will check if they are needed in `libs/assertion/include/hpx/assertion.hpp`

libs/assertion/include/hpx/assertion.hpp
msimberg: Very good catch, but they are actually still different because of the way `HPX_ASSERT` and the helpers `HPX_ASSERT_EXPR_` and `HPX_ASSERT_MSG_` work. I.e. the message is printed correctly. It's another question if it's actually readable like this.

@sithhell do you remember why `HPX_ASSERT` or `HPX_ASSERT_MSG` would need to be variadic?

libs/assertion/src/assertion.cpp
aurianer: Already exported in `libs/assertion/include/hpx/assertion.hpp`

libs/assertion/src/source_location.cpp
aurianer: Apparently only used internally so no need to export it for now

libs/assertion/tests/CMakeLists.txt
aurianer: Sorry I missed this change when I rebased

libs/assertion/include/hpx/assertion.hpp
sithhell: In this implementation, they are equivalent indeed. In fact, `HPX_ASSERT_MSG` isn't necessarily needed anymore.
However, there is nothing wrong with having it implemented like this:
```
#define HPX_ASSERT(expr) HPX_ASSERT_(expr, std::string())
#define HPX_ASSERT_MSG(expr, msg) HPX_ASSERT_(expr, msg)
```
with the proposed implementation, the only thing you need to remember however is HPX_ASSERT which can take an optional message argument.

libs/assertion/src/source_location.cpp
sithhell: It's probably a good idea. I think this class can be nicely reused by a future exception module.

libs/assertion/include/hpx/assertion.hpp
hkaiser: Yes, that's what I'd like to see happening.

aurianer:
Done ^^

msimberg:
Nice, let's get this in. The failing tests on pycicle are not new... Thanks @aurianer!

PR #3846, opened 2019-05-08
Add tests for libs header tests

msimberg:
Previously only build targets were created, but no tests. This adds the tests.


msimberg:
The CircleCI build is missing because I had testing enabled through my account at that time, which seems to exclude it from being tested through the organization account. Here's the build on my account though: https://circleci.com/gh/msimberg/hpx/2142. I'll go ahead and merge this.

Issue #3808, opened 2019-04-19
local_dataflow_boost_small_vector fails to build with Boost 1.70


msimberg:
Build logs from [rostam](http://rostam.cct.lsu.edu/builders/hpx_gcc_8_boost_1_70_centos_x86_64_release/builds/0/steps/build_unit_tests/logs/stdio). I'm guessing this has to do with the default `Allocator` for `small_vector` being `void` in 1.70 (see [here](https://www.boost.org/doc/libs/1_70_0/doc/html/boost/container/small_vector.html)). I'm hoping those more knowledgeable will find the fix trivial...

hkaiser:
How much more incompatible can one make a container? Grrr.

msimberg:
I opened another issue (#3871) to keep track of the actual issue that caused this.

PR #3862, opened 2019-05-16
Temporary fix to local_dataflow_boost_small_vector test

msimberg:
"Fixes" #3808. This sets the allocator explicitly in the test to avoid problems with dataflow not knowing what to do with a `void` allocator.

@hkaiser I'd be happy with this as a solution for the release to not have to rush in a proper fix.

I'm also tempted to say that this isn't even our problem since standard containers don't support `void` as an allocator either, but you're a better judge than I am of how bad this is and what we can reasonably do to support it.

### DIFF
--- a/tests/unit/lcos/local_dataflow_boost_small_vector.cpp
+++ b/tests/unit/lcos/local_dataflow_boost_small_vector.cpp
@@ -25,8 +25,8 @@
 // #include <boost/range/begin.hpp>
 // #include <boost/range/end.hpp>
 
-template <typename T> //, typename Allocator = std::allocator<T>>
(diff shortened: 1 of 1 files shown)


msimberg:
Just for the record, [before](http://rostam.cct.lsu.edu/builders/hpx_gcc_8_boost_1_70_centos_x86_64_debug/builds/27) and [after](http://rostam.cct.lsu.edu/builders/hpx_gcc_8_boost_1_70_centos_x86_64_debug/builds/28) (since I haven't updated pycicle to have a Boost 1.70 builder).

All right, then let's go ahead with this for now.

PR #3864, opened 2019-05-16
Helper functions that are useful in numa binding and testing of allocator

biddisco:
This is a commit that was left off the original branch but is useful for other tests
Adds methods to query number of threads user requested on start (useful when creating custom pools and deciding how many to put in each), a utility function to access the topology from the RP.
Also adds a cast function to the allocator to convert the binder to a user type.
If this could go into the next release, it would be very useful.

### DIFF
--- a/hpx/parallel/util/numa_binding_allocator.hpp
+++ b/hpx/parallel/util/numa_binding_allocator.hpp
@@ -33,7 +33,7 @@
 
 // Can be used to enable debugging of the allocator page mapping
 //#define NUMA_BINDING_ALLOCATOR_INIT_MEMORY
-// #define NUMA_BINDING_ALLOCATOR_DEBUG_PAGE_BINDING
+//#define NUMA_BINDING_ALLOCATOR_DEBUG_PAGE_BINDING
 
 #if defined(NUMA_BINDING_ALLOCATOR_DEBUG_PAGE_BINDING) && !defined(HPX_MSVC)
 #include <plugins/parcelport/parcelport_logging.hpp>
@@ -630,9 +630,16 @@ namespace hpx { namespace compute { namespace host {
         }
 
     public:
+        // return the binding helper cast to a specific type
+        template <typename Binder>
+        std::shared_ptr<Binder> get_binding_helper_cast() const
+        {
+            return std::dynamic_pointer_cast<Binder>(binding_helper_);
+        }
+
         std::shared_ptr<numa_binding_helper<T>> binding_helper_;
-        threads::hpx_hwloc_membind_policy policy_;
-        unsigned int flags_;
+        threads::hpx_hwloc_membind_policy       policy_;
+        unsigned int                            flags_;
 
     private:
         mutable std::mutex init_mutex;
--- a/hpx/runtime/resource/detail/partitioner.hpp
+++ b/hpx/runtime/resource/detail/partitioner.hpp
@@ -15,7 +15,7 @@
 #include <hpx/runtime/threads/topology.hpp>
 #include <hpx/util/command_line_handling.hpp>
 #include <hpx/util/tuple.hpp>
-
+#include <hpx/util/assert.hpp>
 #include <boost/program_options.hpp>
 
 #include <atomic>
@@ -127,6 +127,15 @@ namespace hpx { namespace resource { namespace detail
         // resource partitioner called in hpx_init
         void configure_pools();
 
+        // returns the number of threads(pus) requested
+        // by the user at startup.
+        // This should not be called before the RP has parsed the config and
+        // assigned affinity data
+        std::size_t threads_needed() {
+            HPX_ASSERT(pus_needed_ != std::size_t(-1));
+            return pus_needed_;
+        }
+
         ////////////////////////////////////////////////////////////////////////
         scheduling_policy which_scheduler(std::string const& pool_name);
         threads::topology &get_topology() const;
@@ -224,7 +233,7 @@ namespace hpx { namespace resource { namespace detail
         // holds all of the command line switches
         util::command_line_handling cfg_;
         std::size_t first_core_;
-        std::size_t cores_needed_;
+        std::size_t pus_needed_;
 
         // contains the basic characteristics of the thread pool partitioning ...
         // that will be passed to the runtime
--- a/hpx/runtime/resource/partitioner.hpp
+++ b/hpx/runtime/resource/partitioner.hpp
@@ -259,6 +259,13 @@ namespace hpx { namespace resource
         // Access all available NUMA domains
         HPX_EXPORT std::vector<numa_domain> const& numa_domains() const;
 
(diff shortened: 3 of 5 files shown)


hpx/parallel/util/numa_binding_allocator.hpp
```diff
@@ -630,9 +630,16 @@ namespace hpx { namespace compute { namespace host {
         }
 
     public:
+        // return the binding helper cast to a specific type
+        template <typename Binder>
+        std::shared_ptr<Binder> get_binding_helper_cast() const
+        {
+            return std::dynamic_pointer_cast<Binder>(binding_helper_);
+        }
```
msimberg: Where is this easier to use than just calling `std::dynamic_pointer_cast` directly?

hpx/runtime/resource/detail/partitioner.hpp
```diff
@@ -127,6 +127,14 @@ namespace hpx { namespace resource { namespace detail
         // resource partitioner called in hpx_init
         void configure_pools();
 
+        // returns the number of threads requested by the user at start
+        // should not be called before the RP has parsed the config and
+        // assigned affinity data
+        std::size_t threads_required() {
+            HPX_ASSERT(cores_needed_ != std::size_t(-1));
+            return cores_needed_;
```
msimberg: Requested? Required? Needed? Threads? Cores? Maybe the member variable could be renamed if it's actually talking about threads and not cores.

hpx/runtime/resource/partitioner.hpp
```diff
@@ -259,6 +259,9 @@ namespace hpx { namespace resource
         // Access all available NUMA domains
         HPX_EXPORT std::vector<numa_domain> const& numa_domains() const;
 
+        HPX_EXPORT std::size_t get_number_requested_threads();
+        HPX_EXPORT const hpx::threads::topology &get_topology() const;
```
msimberg: What's wrong with https://github.com/STEllAR-GROUP/hpx/blob/062f097c88fbe5de8c8a8e88dd42ee7cc18b203d/hpx/runtime/threads/topology.hpp#L426 or https://github.com/STEllAR-GROUP/hpx/blob/062f097c88fbe5de8c8a8e88dd42ee7cc18b203d/hpx/runtime.hpp#L152-L155?

hkaiser: Also, I don't want to start a religious war about 'east const' and 'west const', but I would like to emphasize that all of HPX uses 'east const', thus it might be appropriate if you adjusted your style to this.

hpx/parallel/util/numa_binding_allocator.hpp
biddisco: The fact that it's a one line function shouldn't be a reason for not having it in the place where it belongs. I could point to thousands of functions in HPX and say "Where is this easier to use than just writing the code yourself inline?". I use it in several places and it makes user code more readable (IMHO) than having std::dynamic_pointer_cast<long_name_in_here>. It's a convenience wrapper,nothing more.

hpx/runtime/resource/detail/partitioner.hpp
biddisco: This is a more problem with HPX itself. HPX uses threads in many places where cores is implied and vice versa. (For example I changed all the variable names in my scheduler to refer to queue_ids, but you changed them back to thread id's when that was not what I intended). This is just more of the same. I'll give it another name.

hpx/runtime/resource/partitioner.hpp
biddisco: At startup the user may have code that get's the RP, and also needs to get the topology to find out details about the node. The user shiouldn't need to know that the topology is available from the runtime (that hasn't been created yet). IMHO the topology should be available from the resource_partitioner as that's the place where the main management of the node takes place and development of our newer thread pool creation etc is taking place.
Getting it from the runtime ought to be deprecated in favour of the resource partitioner.

msimberg: By the way, there's one way to get it: https://github.com/STEllAR-GROUP/hpx/blob/062f097c88fbe5de8c8a8e88dd42ee7cc18b203d/examples/resource_partitioner/shared_priority_queue_scheduler.hpp#L1371-L1372, which as far as I can tell is equivalent to the one you've added? The difference seems to be that yours is in `partitioner` whereas the other one is in `detail::partitioner` which `hpx::resource::get_partitioner` returns.

hpx/runtime/resource/detail/partitioner.hpp
msimberg: Agreed, so let's try to make it more consistent step by step? Required and needed are basically synonyms so I'd prefer not to have both uses. Requested is slightly different (the comment above the function uses requested) than required.

And by the way, I'm just suggesting a change here. If you had good reasons for choosing a name I'm just asking you to clarify why. The code doesn't tell me why.

hpx/parallel/util/numa_binding_allocator.hpp
msimberg: Fair enough. I was just asking since there's no code in here (e.g. in the NUMA allocator test) that uses it so I can't judge its usefulness myself.

hpx/runtime/resource/partitioner.hpp
biddisco: ok, then I will remove mine. I must have not found the other one when I was looking for it.

hpx/parallel/util/numa_binding_allocator.hpp
biddisco: I remember why I did it now. There is no `get_binding_helper() ` function accessor in the allocator currently, so when a user needs to get it (it may be in a different scope from the original setup of the allocator), we'd need to add
```
 std::shared_ptr<Binder> get_binding_helper_cast() const
        {
            return binding_helper_;
        }
```
and the user would  need to then do
```
  std::shared_ptr<binder_type_2D> binder2 =         
      std::dynamic_pointer_cast<binder_type_2D>(a.localMatrix().get_allocator().get_binding_helper());
```
so rather than add a get_binder function all alone and then require the user to cast it, I just added the single one with the cast already in place. 

I think I'll keep it that way as it's a bit painful if the user has to cast it directly

hpx/runtime/resource/partitioner.hpp
biddisco: OK. I've fixed the `var const &` east/west thingy and left the `get_topology` call in there - as discussed with @msimberg - if the user does
```
hpx::resource::partitioner rp(desc_cmdline, argc, argv);
```
where `rp` is a `resource::partitioner` - then it's a bit odd that they then have to call
```
auto &rp = resource::get_partitioner(); 
auto const& topo = rp.get_topology(); 
```
to get a detail partitioner and retrieve the topology from it. We probably should not be exposing `detail` level objects to the user.

PR #3865, opened 2019-05-17
adds hpx_target_compile_option_if_available

### LINKED ISSUE #3863: shouldn't "-faligned-new" be a usage requirement?
Including HPX libraries from my project moves the -faligned-new problem to my build.

I think that it is an HPX-related thing so it may be better to set this flag as usage requirement.

> @hkaiser it's actually unrelated to nvcc. It's the newer gcc combined with C++11 that's complaining about it. I'd suggest adding `hpx_add_compile_flag_if_available(-faligned-new)` since it's good to have it enabled outside of just C++17. Alternatively `hpx_add_compile_flag_if_available(-Wno-aligned-new)` (I think that would work), since it's a rather benign warning.
> 
> _Originally posted by @msimberg in https://github.com/STEllAR-GROUP/hpx/pull/3745#issuecomment-477168298_

albestro:
Fixes #3863

## Proposed Changes
- add hpx_target_compile_option_if_available CMake function
- now -faligned-new is added to target compile options (if available) and gets exported as usage requirement
- minor-change: adds Visual Studio Code project files to .gitignore


.gitignore
```diff
@@ -13,3 +13,4 @@
 .ycm_extra_conf.py
 apex
 packages
+.vscode
```
msimberg: This is likely accidental. It can go in if you want, but add back the newline in that case.

msimberg:
@hkaiser I'll add the feature test.

msimberg:
Updated. I'm slightly worried that this is a futile attempt because older GCCs and Clangs don't even warn about this, which means that the feature test for those versions passes and they might do whatever they feel like. I considered checking if an allocation is aligned correctly, but one might get lucky with that.

This is still certainly better than before (most importantly that the `-faligned-new` flag is correctly forwarded). If we want to avoid all of this we could of course enable it only with C++17 the official way.

hkaiser:
> This is still certainly better than before (most importantly that the -faligned-new flag is correctly forwarded). If we want to avoid all of this we could of course enable it only with C++17 the official way.

@msimberg: Older compilers - even if silently accepting the code - will not align properly in any case. Their allocators usually don't support this to begin with. OTOH, based on the way we use `alignas`, we don't care if the whole cache-aligned block is actually aligned at a cache line boundary. All we care about is that the neighboring data item is at least one cache line length away. Thus disabling the use of `alignas` for our `cache_line_block<>` for anything but in C++17 mode wouldn't hurt.

msimberg:
@hkaiser thanks for pointing that out. I completely forgot that you added a nice fallback for when alignas isn't available (it was right in front of my nose). I kept worrying that we'll lose all the benefits you introduced with pre-C++17. Now it's enabled only for C++17 (and I've un-disabled the warnings).

msimberg:
@albestro sorry for removing the essence of this PR, but it was really important that you noticed the problem, so thanks for that!

PR #3870, opened 2019-05-24
Module cache

aurianer:
Add `cache` module. Depends on #3832 config module.


hkaiser:
@aurianer is this ready now?

aurianer:
As far as I concerned yes, but I'm not sure as some picycle tests failed due to socket timeout, what do you think ?

hkaiser:
@aurianer the pycicle problems seem to be unrelated, however @msimberg might know better.

aurianer:
Thanks !

PR #3874, opened 2019-05-27
Allow to suppress target keywords in hpx_setup_target for backwards compatibility

hkaiser:
In one of our projects HPX is used as a backend for a library that does not use target keywords (PUBLIC/PRIVATE) in their build system. This change allows us to disable their use for compatibility.


hkaiser:
@msimberg could you have a look at this, please? I'd very much appreciate to have that in asap.

hkaiser:
@msimberg no worries. It was FLeCSI that broke, they rely on non-keyword target dependency specifications in their build system, so this wasn't our fault...

PR #3875, opened 2019-05-27
Converted typedefs in actions submodule to using directives

hkaiser:
- flyby: fixed MSVC warning in internal allocator


hkaiser:
This has been subsumed by #3894 and #3892. Closing.

PR #3876, opened 2019-05-28
Module hardware

aurianer:
Add `hardware` module. Depends on #3845 assertion module.


aurianer:
Need to parse `EXCLUDE` arguments for `add_hpx_lib_header_tests()`, arguments currently ignored => redefinition build error.

PR #3888, opened 2019-06-05
Fix async_customization dataflow example and Clarify what's being tested

biddisco:
This is a minor cleanup of the async_customization example that was incorrect when the guided_executor was fixed.

### DIFF
--- a/examples/resource_partitioner/async_customization.cpp
+++ b/examples/resource_partitioner/async_customization.cpp
@@ -478,19 +478,18 @@ namespace hpx { namespace threads { namespace executors
 
 int hpx_main()
 {
-    int val = 0;
 
     test_async_executor exec;
-    val = test("Testing async custom executor", exec);
+    test("Testing async custom executor", exec);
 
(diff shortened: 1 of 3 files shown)


examples/resource_partitioner/async_customization.cpp
```diff
@@ -6,6 +6,7 @@
 #include <hpx/include/parallel_executors.hpp>
 #include <hpx/runtime/threads/executors/default_executor.hpp>
 #include <hpx/runtime/threads/executors/pool_executor.hpp>
+#define GUIDED_EXECUTOR_DEBUG 1
```
hkaiser: Should this be left in? Or is it an oversight?

biddisco: That one is deliberate, the example/test overrides the debug setting to dump out useful information.

biddisco:
I've fixed the codacy complaint and rebased onto latest master. Added one compilation fix for libfabric parcelport caused by new counters.

PR #3891, opened 2019-06-07
Removal of the default intel config include

m-diers:
That was overlooked during introducing the config module.
Fixes compilations with intel 2019 update 4


hkaiser:
@m-diers Please get in contact with @aserio if you are interested in receiving one of our STE||AR-Group t-shirts. We give those away to our first time contributors.

PR #3894, opened 2019-06-08
Working around VS2019 problem with make_action

hkaiser:
Fixes recent compilation problems using VS2019.


hkaiser:
[[15:47]	K-ballo: hkaiser: I can reproduce the issue, the fix works, and the bases are no longer needed](http://irclog.cct.lsu.edu/ste~b~~b~ar/2019-06-08#232680;)

PR #3899, opened 2019-06-09
Remove old compatibility bases from make_action

K-ballo:


K-ballo:
> We should now be able to remove the specialization of the extract_action trait as well.

Was that specialization there only for handling cases of `extract_action<make_action<...>>`?

hkaiser:
@K-ballo looks like I was wrong when I said we can remove the extract_action trait (https://circleci.com/gh/STEllAR-GROUP/hpx/94660)

K-ballo:
> @K-ballo looks like I was wrong when I said we can remove the extract_action trait (https://circleci.com/gh/STEllAR-GROUP/hpx/94660)

Those actually break with just the base change too.

We'll have to start with only deprecating those bases, fixing the examples, and tightening `extract_action` to specialize only for the `make_action` traits.

hkaiser:
> Those actually break with just the base change too.

Not sure about this...

> We'll have to start with only deprecating those bases, fixing the examples, and tightening extract_action to specialize only for the make_action traits.

I fully agree. Let's do it in stages.

hkaiser:
@msimberg I think this is good to go. Please review and/or merge at your leasure.

PR #3900, opened 2019-06-09
Making sure errors during command line processing are properly reported and will not cause assertions

### LINKED ISSUE #3890: hpx.ini not working
## Expected Behavior
Loading my setting from hpx.ini

## Actual Behavior

Ignoring the settings or - if forced to load hpx.ini with 
--hpx:ini=hpx.ini
crashing in the MSVC runtime with a call to abort()

1. Debug build of my app
2. hpx.ini like this:

```
[hpx.logging]
# level = ${HPX_LOGLEVEL:0} ## uses HPX_LOGLEVEL env var and defaults to 0
# level = ${HPX_LOGLEVEL:4} ## uses HPX_LOGLEVEL env var and defaults to 0
level = 4
#destination = ${HPX_LOGDESTINATION:console}
destination = file(hpx.log)
#destination = file(hpx.log)
format = ${HPX_LOGFORMAT:(T%locality%/%hpxthread%.%hpxphase%/%hpxcomponent%) P%parentloc%/%hpxparent%.%hpxparentphase% %time%($hh:$mm.$ss.$mili) [%idx%]|\\n}

```

3a. starting the app = no logging
3b. starting the app with --hpx:ini=hpx.ini causing a call to abort() in the runtime
"Debug Error!
Program: ....
abort() has been called
etc. yadda yadda

Please describe your environment
Windows 10, MSVC 2017
- HPX Version: 1.3.0

hkaiser:
Fixes #3890

### DIFF
--- a/hpx/runtime/resource/detail/create_partitioner.hpp
+++ b/hpx/runtime/resource/detail/create_partitioner.hpp
@@ -47,13 +47,14 @@ namespace hpx { namespace resource { namespace detail
         int argc, char** argv, std::vector<std::string> ini_config,
         resource::partitioner_mode rpmode = resource::mode_default,
         runtime_mode mode = runtime_mode_default,
-        bool check = true);
+        bool check = true, int* result = nullptr);
 
 #if !defined(HPX_EXPORTS)
     inline partitioner& create_partitioner(
         int argc, char** argv,
         resource::partitioner_mode rpmode = resource::mode_default,
-        runtime_mode mode = runtime_mode_default, bool check = true)
+        runtime_mode mode = runtime_mode_default, bool check = true,
+        int* result = nullptr)
     {
         boost::program_options::options_description desc_cmdline(
             std::string("Usage: ") + HPX_APPLICATION_STRING + " [options]");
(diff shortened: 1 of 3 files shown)


hpx/runtime/resource/detail/create_partitioner.hpp
```diff
@@ -47,13 +47,14 @@ namespace hpx { namespace resource { namespace detail
         int argc, char** argv, std::vector<std::string> ini_config,
         resource::partitioner_mode rpmode = resource::mode_default,
         runtime_mode mode = runtime_mode_default,
-        bool check = true);
+        bool check = true, int* result = nullptr);
```
msimberg: Am I missing something, or is `int *` basically used as an `optional<int>`?

hkaiser: I used `int*` with a default value to avoid having to change any of the existing code relying on this function. It's not really (conceptually) a `optional<int>`

msimberg: Ok, fair enough.

PR #3901, opened 2019-06-10
Factoring out container remapping into a separate trait

### LINKED ISSUE #3871: dataflow does not support void allocators
Details in #3808. While #3862 fixed the immediate issue of building the test, this issue is to keep track of the underlying issue of `dataflow` not handling a `void` allocator in `small_vector`.

hkaiser:
This removes the dependency of the unwrapping code on boost::small_vector

Fixes #3871


msimberg:
Thanks @hkaiser for doing this. Since the CDash upgrade we're not getting any logs from builds, so can't tell what those failing pycicle builds are doing at the moment. Hopefully we'll get it back up next week.

PR #3902, opened 2019-06-11
Add add_hpx_module function

aurianer:
To factorize module's CMakeLists.txt because redundant code in each module.


cmake/HPX_AddModule.cmake
```diff
@@ -0,0 +1,165 @@
...
+  foreach(header_file ${tmp_headers})
+      FILE(APPEND ${${name}_HEADER_ROOT}/${name}_module.hpp
+      "#include <${header_file}>\n"
+    )
+    list(APPEND headers "${${name}_HEADER_ROOT}/${header_file}")
+  endforeach(header_file)
+
+  foreach(header_file ${headers})
+      message(${header_file})
+  endforeach(header_file)
```
hkaiser: This is probably an oversight. Alternatively you may want to use `hpx_debug(…)` instead of `message(…)` allowing to control the amount of output generated.

libs/cache/CMakeLists.txt
```diff
@@ -5,150 +5,64 @@
...
+    FOLDER ${CMAKE_CURRENT_SOURCE_DIR}
+    SOURCE_ROOT ${source_root}
+    HEADER_ROOT ${header_root}
+    COMPAT_HEADER_ROOT ${compat_header_root}
+    SOURCES ${${module_name}_sources}
+    HEADERS ${${module_name}_headers}
+    COMPAT_HEADERS ${${module_name}_compat_headers}
+    DEPENDENCIES hpx_config
+    CMAKE_SUBDIRS examples tests
 )
```
hkaiser: Should we add AUTOGLOB capabilities to this function (see add_hpx_library() and friends)? I know that generally people discourage the use of file system globbing for cmake projects as it might break cmake's own dependency tracking. It might however simplify creating modules significantly as the user wouldn't have to explicitly list all header and source files...

libs/cache/CMakeLists.txt
```diff
@@ -5,150 +5,64 @@
...
-  ${CMAKE_CURRENT_SOURCE_DIR}/include/hpx/cache/entries/lfu_entry.hpp
-  ${CMAKE_CURRENT_SOURCE_DIR}/include/hpx/cache/entries/lru_entry.hpp
-  ${CMAKE_CURRENT_SOURCE_DIR}/include/hpx/cache/entries/size_entry.hpp
-  ${CMAKE_CURRENT_SOURCE_DIR}/include/hpx/cache/policies/always.hpp
-  ${CMAKE_CURRENT_SOURCE_DIR}/include/hpx/cache/statistics/local_full_statistics.hpp
-  ${CMAKE_CURRENT_SOURCE_DIR}/include/hpx/cache/statistics/local_statistics.hpp
-  ${CMAKE_CURRENT_SOURCE_DIR}/include/hpx/cache/statistics/no_statistics.hpp
+set(source_root ${CMAKE_CURRENT_SOURCE_DIR}/src)
+set(header_root ${CMAKE_CURRENT_SOURCE_DIR}/include)
+set(compat_header_root ${CMAKE_CURRENT_SOURCE_DIR}/include_compatibility)
```
hkaiser: These three variables are always defined the same (for all modules). Should we use those as defaults inside add_hpx_module if they are not specified explicitly?

libs/cache/CMakeLists.txt
aurianer: I don't know cause keeping the headers like this makes it safer, and we don't create modules that often, what do you think ?

cmake/HPX_AddModule.cmake
```diff
@@ -0,0 +1,166 @@
...
+  hpx_debug("Add module ${name}: COMPAT_HEADER_ROOT: ${COMPAT_HEADER_ROOT}")
+
+  string(TOUPPER ${name} name_upper)
+
+  # HPX options
+
+  hpx_option(HPX_${name_upper}_WITH_TESTS
+    BOOL
+    "Build HPX ${name} module tests. (default: ON)"
+    ON ADVANCED
```
hkaiser: Shouldn't this use `${HPX_WITH_TESTS}` for the default value (see for instance: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/config/CMakeLists.txt#L21)?

cmake/HPX_AddModule.cmake
```diff
@@ -0,0 +1,166 @@
...
+      hpx_add_config_define_namespace(
+          DEFINE HPX_${name_upper}_HAVE_COMPATIBILITY_HEADERS
+          NAMESPACE ${name_upper})
+    endif()
+  endif()
+
+  # Write full path for the sources files
+  include(HPX_CMakeUtils)
+  prepend(sources ${SOURCE_ROOT} ${${name}_SOURCES})
+  prepend(headers ${HEADER_ROOT} ${${name}_HEADERS})
```
hkaiser: Will this do the right thing even if the files are already fully specified?

cmake/HPX_AddModule.cmake
```diff
@@ -0,0 +1,166 @@
...
+      "#include <${header_file}>\n"
+    )
+    list(APPEND headers "${HEADER_ROOT}/${header_file}")
+  endforeach(header_file)
+
+  foreach(header_file ${headers})
+      hpx_debug(${header_file})
+  endforeach(header_file)
+
+  message(STATUS "${name}: Configuring")
```
hkaiser: We recently switched to using `hpx_info()` here (see: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/config/CMakeLists.txt#L16).

cmake/HPX_AddModule.cmake
```diff
@@ -0,0 +1,166 @@
...
+  
+  write_config_defines_file(
+    NAMESPACE ${name_upper}
+    FILENAME "${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/hpx/${name}/config/defines.hpp")
+  
+  foreach(dir ${${name}_CMAKE_SUBDIRS})
+    add_subdirectory(${dir})
+  endforeach(dir)
+  
+  message(STATUS "${name}: Configuring done")
```
hkaiser: This is not used by the other modules anymore

cmake/HPX_AddModule.cmake
aurianer: No, but should we specify the full path here ? cause the structure in each module is the same : `src/`, `include/`, `include_compatibility`

libs/cache/CMakeLists.txt
```diff
@@ -5,150 +5,55 @@
...
-  "Enable warnings for deprecated facilities. (default: ${HPX_WITH_DEPRECATION_WARNINGS})"
-  ${HPX_WITH_DEPRECATION_WARNINGS} ADVANCED
-  CATEGORY "Modules")
-
-if(HPX_CACHE_WITH_DEPRECATION_WARNINGS)
-  hpx_add_config_define_namespace(
-      DEFINE HPX_CACHE_HAVE_DEPRECATION_WARNINGS
-      NAMESPACE CACHE)
-endif()
+project(HPX.${module_name} CXX)
```
hkaiser: May I suggest _not_ to use the `${module_name}` cmake variable throughout the file for concrete modules to improve readability?

cmake/HPX_AddModule.cmake
msimberg: I think we should be able to go with relative paths for now. It forces us to use the same structure in all the modules. If we need absolute paths at some point in the future I think we can easily add an option for that when needed.

libs/cache/CMakeLists.txt
hkaiser: Yah, it's probably not a big issue - definitely not a showstopper for this PR.

cmake/HPX_AddModule.cmake
hkaiser: Fair point.

cmake/HPX_AddModule.cmake
aurianer: sorry :/

libs/cache/CMakeLists.txt
msimberg: I kind of agree. The question is if we can put the `project` call inside `hpx_add_module as well. Then we would barely need to use either `${module_name}` or `cache` in the `CMakeLists.txt`. @aurianer could you check if that works?

libs/cache/CMakeLists.txt
```diff
@@ -5,8 +5,6 @@
 
 cmake_minimum_required(VERSION 3.3.2 FATAL_ERROR)
 
-project(HPX.cache CXX)
-
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 include(HPX_AddDefinitions)
```
hkaiser: One last nitpick: this is not needed anymore. Instead we should add a `include(HPX_Message)` now. After that this PR is good to go, I think. Thanks a lot!

aurianer: Thank you !

libs/cache/CMakeLists.txt
aurianer: Actually it has been removed

hkaiser:
Part of the idea of introducing modules was to enable people to use those independently. Wouldn't that imply that the CMakeLists.txt for the modules have to have their own `project()` statement?

aurianer:
Sorry I misunderstood your previous comment, I'll try to put it inside the function

aurianer:
Sorry apparently there is come problems, I will check those :/

aurianer:
Ok this should be good to go, sorry for the wait ^^

cmake/HPX_AddModule.cmake
```diff
@@ -0,0 +1,181 @@
...
+
+  set_target_properties(hpx_${name} PROPERTIES
+    FOLDER "Core/Modules")
+
+  install(TARGETS hpx_${name} EXPORT HPXTargets
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+    COMPONENT ${name}
+  )
```
hkaiser: Should we really install the generated binaries? I thought we need those while building the HPX core library only/

cmake/HPX_AddModule.cmake
```diff
@@ -0,0 +1,181 @@
...
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+    COMPONENT ${name})
+
+  if(HPX_${name_upper}_WITH_COMPATIBILITY_HEADERS)
+    install(
+      DIRECTORY include_compatibility/hpx
+      DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+      COMPONENT ${name})
+  endif()
+
```
hkaiser: We need to additionally install the generated header file, if needed:
```
  if (${name}_GLOBAL_HEADER_GEN)
    install(
      FILES ${CMAKE_BINARY_DIR}/hpx/${name}.hpp
      DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/hpx
      COMPONENT ${name})
  endif()
```

cmake/HPX_AddModule.cmake
```diff
@@ -0,0 +1,181 @@
...
+  include(HPX_Message)
+  include(HPX_Option)
+
+  hpx_info("  ${name}")
+
+  # Compatibility header off is not specified
+  if (NOT ${name}_COMPATIBILITY_HEADERS)
+      set(${name}_COMPATIBILITY_HEADERS OFF)
+  endif()
```
hkaiser: This `if()`/`endif()` is superfluous (but not harmful).

cmake/HPX_AddModule.cmake
aurianer: Let's see with @msimberg when he returns, but installing the module libraries could simplify the compilation process if we consider the modules as standalone in the future, just linking them from the binaries, but as I said, I'm not sure of what mickael wants to do with it ^^

msimberg: I'm not 100% sure where I want to go with this but having them installed for now won't hurt. Basically we can keep it or remove it now and it won't make a difference because users shouldn't be linking manually to these anyway. Eventually I would like at least a few high-level shared libraries that can be used explicitly. Exposing all the small static libraries would be nice as well, but we'd need to make sure that things work correctly when linking with a static library and a shared library that includes that static library (probably won't work).

hkaiser: What if we made this optional using a `INSTALL_BINARIES` tag for the `add_hpx_module()` function (default could be `OFF` for now)? I'm a bit hesitant to install those now as users might think using them this is a supported scenario... - I know this is more wishful thinking than anything else - but anyways.

PR #3903, opened 2019-06-11
Adding new performance counters

hkaiser:
- `/threads/time/background-send-duration`, `/threads/time/background-receive-duration`
- `/threads/time/background-send-overhead`, `/threads/time/background-receive-overhead`

These counters currently return meaningful values for the MPI parcelport only

@justwagle please verify whether the new counters provide you with the information you need

### DIFF
--- a/hpx/runtime/threads/detail/scheduled_thread_pool_impl.hpp
+++ b/hpx/runtime/threads/detail/scheduled_thread_pool_impl.hpp
@@ -45,6 +45,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <exception>
+#include <functional>
 #include <iosfwd>
 #include <memory>
 #include <string>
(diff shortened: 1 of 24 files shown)


hpx/runtime/threads/detail/scheduled_thread_pool_impl.hpp
```diff
@@ -1387,6 +1398,224 @@ namespace hpx { namespace threads { namespace detail
...
+        std::int64_t tfunc_total = 0;
+        std::int64_t reset_tfunc_total = 0;
+
+        if (num != std::size_t(-1))
+        {
+            tfunc_total = counter_data_[num].tfunc_times_;
+            reset_tfunc_total =
+                counter_data_[num].reset_background_receive_tfunc_times_;
+
+            bg_total = counter_data_[num].background_duration_;
```
justwagle: I think this line should be `counter_data_[num].background_receive_duration_`      instead of `counter_data_[num].background_duration_ ` ?

hkaiser: Excellent catch! Should be fixed now.

PR #3921, opened 2019-06-17
Add ccls cache directory to gitignore

msimberg:
Ignore cache directory produced by [ccls](https://github.com/MaskRay/ccls).


msimberg:
Yeah, I know I could, but I don't want to encourage that. This keeps even small changes like this visible to someone who might not otherwise have noticed.

PR #3922, opened 2019-06-17
Add Windows builds with Travis CI

msimberg:
Removes the old Appveyor config and adds a Travis CI config for Windows.

Windows people (@hkaiser, @K-ballo, @parsa) could you please have a look?

At the moment it seems like pseudo dependencies are broken again with examples and tests on Windows, so I've disabled that. Also, the Boost version is [outdated](https://travis-ci.com/msimberg/hpx/jobs/208569529)? I used the same vcpkg dependencies that were used on Appveyor before.

I think this still needs some polishing, but whatever we end up with is probably better than what we have right now (nothing).


hkaiser:
@msimberg I have updated the HPX dependencies to the latest Boost (the file name has not changed). The old file is still available, if you need it (http://stellar-group.org/files/vcpkg-export-hpx-dependencies-boost1.66-1.7z).

msimberg:
I think we can go ahead with this. I don't have a solution for the pseudo-dependencies right now and building `core` will be better than nothing for now.

msimberg:
I'll go ahead with this now. Let's see if Travis picks it up once it's merged.

PR #3923, opened 2019-06-17
Allow aligning objects bigger than the cache line size

msimberg:


hpx/util/cache_aligned_data.hpp
```diff
@@ -30,7 +30,8 @@ namespace hpx { namespace util
 
         // pad to cache line size bytes
         Data data_;
-        char cacheline_pad[threads::get_cache_line_size() - sizeof(Data)];
+        char cacheline_pad[threads::get_cache_line_size() -
+            (sizeof(Data) % threads::get_cache_line_size())];
     };
```
hkaiser: We'll need to adapt the `static_assert` above in this case.

hkaiser: We'll need to adapt the `static_assert` above in this case.

msimberg: Indeed, good catch. Thanks!

msimberg: Indeed, good catch. Thanks!

msimberg:
The codacy warning is on point but a must in this case.

hkaiser:
> The codacy warning is on point but something we can't really change in this case...

I have flagged the issue to be ignored. It will not show up for future changes to this PR anymore.

msimberg:

PR #3924, opened 2019-06-17
Add schedule hint executor parameters

msimberg:
Adds task placement executor parameters to complement chunk size parameters.

Implements
- `chunked_placement`: places contiguous tasks close to each other
- `round_robin_placement`: places tasks round-robin...
- `scheduler_placement`: leaves the decision to the scheduler

I've also added a `default_schedule` which combines `static_chunk_size` and `chunked_placement`. It's also the default for `parallel_executor`. This is almost like OpenMP's static schedule except work stealing may make it not completely static. To do this properly we'd need to be able to create tasks that can't be stolen.

Most importantly this makes the "static schedule" that we had before with `static_chunk_size` as static as we can without unstealable tasks. Earlier hierarchical spawning would interfere with round robin scheduling.

I'll post some benchmark numbers soon, but on multiple NUMA domains the improvement can be 10's of %. On single NUMA domains the difference is negligible. I assumed that `chunked_placement` would be faster than `round_robin` but there's practically no difference. The biggest improvement seems  to come from deterministic scheduling.

### DIFF
--- a/hpx/async_launch_policy_dispatch.hpp
+++ b/hpx/async_launch_policy_dispatch.hpp
@@ -66,15 +66,13 @@ namespace hpx { namespace detail
             !traits::is_action<Action>::value
         >::type>
     {
-        template <typename F, typename ...Ts>
-        HPX_FORCEINLINE static
-        typename std::enable_if<
+        template <typename F, typename... Ts>
+        HPX_FORCEINLINE static typename std::enable_if<
             traits::detail::is_deferred_invocable<F, Ts...>::value,
-            hpx::future<
-                typename util::detail::invoke_deferred_result<F, Ts...>::type
-            >
-        >::type
-        call(launch policy, F && f, Ts&&... ts)
+            hpx::future<typename util::detail::invoke_deferred_result<F,
+                Ts...>::type>>::type
+        call(launch policy, threads::thread_schedule_hint hint, F&& f,
+            Ts&&... ts)
         {
             typedef typename util::detail::invoke_deferred_result<F, Ts...>::type
                 result_type;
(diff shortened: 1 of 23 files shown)


msimberg:
This works for my test cases but not in the general case. I had made the assumption that I can pass the scheduling hint in the `shape` parameter to `bulk_async_execute` but that means one can't just pass anything iterable as the shape, like a `vector<int>` for example.

Another approach I considered was querying the scheduling hints in the executor, but the parameters are no longer available once you're inside `bulk_async_execute`. Passing the hints on to the partitioner tasks is too late because by then the task has already been spawned. We could do all of this instead by extending the block executor(s) to handle different task placements, but it feels wrong that chunk sizes get to be executor parameters whereas placement would have to be an executor. Could we actually require that the shape has to contain elements which are tuples with the first element a scheduling hint? That seems overly restrictive. Am I overthinking this?

Any ideas @hkaiser?

hkaiser:
@msimberg I think you could use a similar technique as I described here: https://github.com/executors/executors/issues/9#issuecomment-270212206. IOW create a slim executor-shim that holds the scheduler hint and passes it along to the wrapped executor's `bulk_execute` through its variadic interface.

msimberg:
| num elements | chunked | round robin | no hint (scheduler round robin) |
| --- | --- | --- | --- |
| 1000000 |  |  |  |
|  |  |  |  |
|  |  |  |  |
|  |  |  |  |
|  |  |  |  |
|  |  |  |  |

msimberg:
Thanks @hkaiser for the link. We might need to discuss this in real time as I'm not quite sure how that technique helps here. Are you suggesting that I could keep a task placement executor parameter with that technique?

Some selected timings from the benchmark in this PR. For smaller inputs there's no difference anymore and I think our task overheads start dominating anyway. This is on Daint multicore (36 threads, 2 * 18 core Xeons).

| number of elements | no hint (scheduler round robin) | chunked | round robin |
| - | - | - | - |
| 1000000 | 0.00015662 | 0.00014406 | 0.000157818 |
| 10000000 | 0.000999082 | 0.000818076 | 0.000801452 |
|1000000000  | 0.219809 | 0.153083 | 0.153302 |

msimberg:
Closing this as the approach won't work without major changes or unpleasant restrictions. @hkaiser pointed out at it might be worth making the executor parameters explicitly visible inside executors to enable this kind of functionality. The properties system used in newer executors proposals should also allow this kind of customization.

PR #3925, opened 2019-06-18
Move version check to config module

msimberg:
Move version check definitions to config module. They should've been defined there in the first place (instead of duplicated in the core library and config). Moving this to the config library means that using modules also have the version check. Should also fix the linker errors in the assertion PR.

Flyby: use non-compatibility headers from the cache module.

### DIFF
--- a/cmake/templates/modules.cpp.in
+++ b/cmake/templates/modules.cpp.in
@@ -3,10 +3,14 @@
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-#include <hpx/cache/force_linking.hpp>
+#include <hpx/config/export_definitions.hpp>
 
+@MODULE_FORCE_LINKING_INCLUDES@
 ///////////////////////////////////////////////////////////////////////////////
-namespace hpx { namespace cache
+namespace hpx
 {
-    void force_linking() {}
-}}
+    // Enforce for this function to be linked into the main HPX library
+    HPX_EXPORT void force_linking()
+    {@MODULE_FORCE_LINKING_CALLS@
+    }
+}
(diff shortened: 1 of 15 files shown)


hkaiser:
@msimberg this change will cause linker errors on platforms that have to explicitly export symbols in order for them to be visible to the outside. 

This also highlights a potential issue with our module setup I was hoping wouldn't turn into a real problem. If a symbol needs to be exported from the main HPX library but is defined in one of the modules we need to somehow enforce for this symbol to be linked into the main HPX library and (re-)exported from there. I'm not quite sure how this could be done.

hkaiser:
@msimberg I think I have a solution for this. I will push to this PR once I'm confident things work.

msimberg:
@hkaiser ok, thanks for looking into this.

> If a symbol needs to be exported from the main HPX library but is defined in one of the modules

This to me feels pretty much inevitable. I see most of HPX being defined in modules and there being one, or hopefully more than one, meta-library.

msimberg:
So why does this work? We're creating a dummy `HPX_EXPORT force_linking() {}` in each module, which we then "use" in the main library in `modules.cpp`. And if we use at least one exported symbol from a module in the main library, *all* exported symbols will be exported by the main library as well?

hkaiser:
@msimberg this works for now as (at least on Windows) the linker picks symbols based on object file granularity. IOW, once one symbol is needed from a file, all symbols from this file are linked in. This is not foolproof obviously (you can instruct the linker to apply function-level granularity, but we don't do that in our case anyways). This also requires to make sure that at least one top-level (i.e. API) symbol from each translation unit is referenced. But this could be done from the module-based `force_linking()` function.

hkaiser:
> We're creating a dummy HPX_EXPORT force_linking() {} in each module

@msimberg please note, that the module based `force_linking()` functions are not exported (there is no need to) but just referenced from the one in the main module. That one however _should_ be exported to force the linker not to discard the symbol (as its not used anywhere).

msimberg:
@hkaiser rebased and fixed the conflicts. Ready to go from your side? I'll just wait and see that at least CircleCI approves of this.

PR #3928, opened 2019-06-21
fixed number of tasks and number of threads for heterogeneous slurm environments

m-diers:
Fixed for heterogeneous slurm environments
- number of tasks determination
- number of threads determination

Added test to show the slurm environment with its variables.

Changed the behavior with hyperthreading. Before it was determined by hpx::topology, now it depends on the slurm argument --threads-per-core, which is intended for this purpose, according to the approach `2019-03-18 14:23 <hkaiser> mdiers_: all depends on the environment`.

Some more informations about it: http://irclog.cct.lsu.edu/ste~b~~b~ar/2019-03-18


hkaiser:
@m-diers could you please add the missing #include's to your PR (see https://98428-4455628-gh.circle-artifacts.com/0/hpx_inspect_report.html). We would be able to merge it then.

PR #3931, opened 2019-06-23
Adding all_to_all

hkaiser:
Adding `all_to_all` facility

- flyby: create new module for collectives, `all_to_all` is the first (and only)
         facility in this module
- flyby: adding `promise::get_shared_future` to allow extracting a shared future
         from a promise without access to any (shared-)future that was extracted
         from this promise before
- flyby: minor fixes to `gather_here`/`gather_there`

This relies on #3902 being merged first


hkaiser:
@msimberg I'd need some help integrating the doxygen documentation for the `all_to_all` to the new documentation system (see: https://github.com/STEllAR-GROUP/hpx/pull/3931/files#diff-43aea3f47587805148adec9676d6b19dR14). Any pointers/guidelines would be much appreciated.

biddisco:
What is the underlying implementation based on? I didn't look at the code yet but will review it later in the week if you wish. Does it use any tree based algorithms or anything fancy to minimize latency/bandwidth or total messages sent etc?

hkaiser:
@biddisco the implementation is straightforward, nothing special. It creates a central point of collecting the information where all participants extract the data from. My goal was not to have something very high performant, I needed the abstraction. We can always get fancy implementation-wise later.

libs/collectives/include/hpx/collectives/all_to_all.hpp
```diff
@@ -0,0 +1,483 @@
...
+    ///             values send by all participating sites. It will become
+    ///             ready once the all_to_all operation has been completed.
+    ///
+    template <typename T>
+    hpx::future<std::vector<T> >
+    all_to_all(char const* basename, hpx::future<T> result,
+        std::size_t num_sites = std::size_t(-1),
+        std::size_t generation = std::size_t(-1),
+        std::size_t this_site = std::size_t(-1),
+        std::size_t root_sire = 0);
```
msimberg: Should probably be `root_site`.

libs/collectives/include/hpx/collectives/all_to_all.hpp
```diff
@@ -0,0 +1,483 @@
...
+    ///             values send by all participating sites. It will become
+    ///             ready once the all_to_all operation has been completed.
+    ///
+    template <typename T>
+    hpx::future<std::vector<typename std::decay<T>::type> >
+    all_to_all(char const* basename, T && result,
+        std::size_t num_sites = std::size_t(-1),
+        std::size_t generation = std::size_t(-1),
+        std::size_t this_site = std::size_t(-1),
+        std::size_t root_sire = 0);
```
msimberg: `root_sire`.

msimberg:
> @msimberg I'd need some help integrating the doxygen documentation for the `all_to_all` to the new documentation system (see: https://github.com/STEllAR-GROUP/hpx/pull/3931/files#diff-43aea3f47587805148adec9676d6b19dR14). Any pointers/guidelines would be much appreciated.

It should "just work", i.e. anything not in detail is added as doxygen docs now. I'll check that it does the right thing though.

Another thing this does is create a cyclic dependency between this and the core library. I don't think we can avoid this at the moment without having all the prerequisites in place as modules, but I just want to point out that we should be very careful about not forgetting cyclic dependencies like these. I'm really happy to see this as a module though!

hkaiser:
> Another thing this does is create a cyclic dependency between this and the core library. 

I thought about this. In this case it doesn't create an issue as `all_to_all` is header only...

libs/collectives/include/hpx/collectives/all_to_all.hpp
hkaiser: Doh! Will fix.

hkaiser:
@msimberg this is good to go from my end, please review.

cmake/HPX_AddCompileTest.cmake
```diff
@@ -51,10 +51,13 @@ function(add_hpx_headers_compile_test category name)
...
-  file(GLOB_RECURSE headers ${DO_CONFIGURE_DEPENDS} "${PROJECT_SOURCE_DIR}/include/hpx/*hpp")
+  file(GLOB_RECURSE headers ${DO_CONFIGURE_DEPENDS}
+      "${PROJECT_SOURCE_DIR}/include/hpx/*hpp")
+
   set(all_headers)
-  add_custom_target(tests.headers.${lib})
-  add_dependencies(tests.headers tests.headers.${lib})
+  add_custom_target(tests.headers.modules.${lib})
+  add_dependencies(tests.headers.modules tests.headers.modules.${lib})
+
```
msimberg: We should probably still have `tests.headers.modules` as a dependency of `tests.headers`.

hpx/lcos/local/promise.hpp
```diff
@@ -253,16 +287,24 @@ namespace hpx { namespace lcos { namespace local
 
         // Returns: A future<R> object with the same shared state as *this.
         // Throws: future_error if *this has no shared state or if get_future
-        //         has already been called on a promise with the same shared
-        //         state as *this.
+        //         or get_shared_future has already been called on a promise
```
msimberg: These don't of course show up in docs because it's not in doxygen format. Probably better to leave it for another PR though because there are plenty of other places where we have this kind of almost-doxygen docs.

libs/CMakeLists.txt
```diff
@@ -9,11 +9,28 @@ include(HPX_Message)
 
 set(HPX_LIBS
   cache
+  collectives
   config
   preprocessor
   CACHE INTERNAL "" FORCE
 )
 
+if (HPX_WITH_TESTS_UNIT)
```
msimberg: `create_library_skeleton.py` should also be updated with this.

libs/_example/tests/CMakeLists.txt
```diff
@@ -4,28 +4,31 @@
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 if (NOT HPX_WITH_TESTS)
+  hpx_set_option(HPX__EXAMPLE_WITH_TESTS VALUE OFF FORCE)
```
msimberg: I'm considering removing the example module altogether now that we have a few actual modules in place. Documentation about creating modules should probably anyway go into the developer documentation. Do you think there's value in keeping it?

libs/CMakeLists.txt
msimberg: And `tests.unit.modules` should probably be added as a dependency to `tests.unit` just like the header tests.

libs/collectives/docs/index.rst
```diff
@@ -0,0 +1,12 @@
...
+
+   Distributed under the Boost Software License, Version 1.0. (See accompanying
+   file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+.. _libs_collectives:
+
+=====
+collectives
+=====
+
```
msimberg: Would you mind adding at least one sentence describing what the module provides? And this is absolutely not important, but I've tried to keep the over- and underlines of the heading the same length as the heading itself.

libs/collectives/examples/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2019 The STE||AR-Group
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+if (HPX_WITH_TESTS AND HPX_WITH_TESTS_EXAMPLES)
+  add_hpx_pseudo_target(tests.examples.collectives)
```
msimberg: Should these also go under `tests.examples.modules`? Don't really care either way.

msimberg:
> > @msimberg I'd need some help integrating the doxygen documentation for the `all_to_all` to the new documentation system (see: https://github.com/STEllAR-GROUP/hpx/pull/3931/files#diff-43aea3f47587805148adec9676d6b19dR14). Any pointers/guidelines would be much appreciated.
> 
> It should "just work", i.e. anything not in detail is added as doxygen docs now. I'll check that it does the right thing though.

@hkaiser you don't need to do anything additional for this to work.

hpx/lcos/local/promise.hpp
hkaiser: I agree. I simply went with what we have already in the same file.

libs/_example/tests/CMakeLists.txt
hkaiser: Yah, let's remove it. Should be done in a separate PR, though.

hkaiser:
@msimberg I think I have addressed all of your comments, thanks!

msimberg:
@hkaiser the `force_linking.hpp` header was missing. I extended `add_hpx_module` to generate it instead, but could you verify that it still works correctly for you on Windows?

hkaiser:
@msimberg thanks for looking into this. I applied a couple of minor changes to streamline the force-linking even further. Should be good to go now.

PR #3933, opened 2019-06-26
Remove cudadevrt from compile/link flags as it breaks downstream projects

biddisco:
I'm not sure if cudadevrt is still required, but removing this doesn't cause any problems on my builds, and fixes link problems with projects that use HPX with cuda enabled


msimberg:
I have nothing against this change, but what exactly breaks with it? What about Windows...?

biddisco:
Any project that does a `find_package(HPX)` and uses HPX when the HPX has cuda enabled gets link errors about not finding cudadevrt. This was the easiest fix for me.

PR #3934, opened 2019-06-26
Add set function for sliding_semaphore max_difference 

biddisco:
This allows the semaphore distance to be adjusted after initial use.

Largely obsolete now, due to the limiting_executor, but I have some code still using this.

### DIFF
--- a/hpx/lcos/local/detail/sliding_semaphore.hpp
+++ b/hpx/lcos/local/detail/sliding_semaphore.hpp
@@ -35,6 +35,13 @@ namespace hpx { namespace lcos { namespace local { namespace detail
           : max_difference_(max_difference), lower_limit_(lower_limit), cond_()
         {}
 
+        void set_max_difference(std::int64_t max_difference,
+                                std::int64_t lower_limit)
+        {
+            max_difference_ = max_difference;
+            lower_limit_    = lower_limit;
+        }
+
         void wait(std::unique_lock<mutex_type>& l, std::int64_t upper_limit)
         {
             HPX_ASSERT_OWNS_LOCK(l);
--- a/hpx/lcos/local/detail/sliding_semaphore.hpp
+++ b/hpx/lcos/local/detail/sliding_semaphore.hpp
@@ -35,9 +35,12 @@ namespace hpx { namespace lcos { namespace local { namespace detail
           : max_difference_(max_difference), lower_limit_(lower_limit), cond_()
         {}
 
-        void set_max_difference(std::int64_t max_difference,
+        void set_max_difference(std::unique_lock<mutex_type> &l,
+                                std::int64_t max_difference,
                                 std::int64_t lower_limit)
         {
+            HPX_ASSERT_OWNS_LOCK(l);
+
             max_difference_ = max_difference;
             lower_limit_    = lower_limit;
         }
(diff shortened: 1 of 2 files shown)


hpx/lcos/local/detail/sliding_semaphore.hpp
```diff
@@ -35,6 +35,13 @@ namespace hpx { namespace lcos { namespace local { namespace detail
           : max_difference_(max_difference), lower_limit_(lower_limit), cond_()
         {}
 
+        void set_max_difference(std::int64_t max_difference,
```
msimberg: Is it safe to call this at any time? And is just `set` a better name? Just because it sets both `max_difference` and `lower_limit`. Looks useful though.

biddisco: The max diff and lower limit don't change constantly (unlike the internal counter), so changing them will not be a big deal, it might happen that you change it, whilst a task is completing and one extra or less is alowed through. for the kind of stuff we're using it for, this isn't worth putting locks on. but I'll bear it in mind next time I tweak it.
I can change the name to ust `set`, but then I'd have to go through all my code and change that too, so I'd rather just leave it. Once I've transitioned the cholesky code to use limiting executor, I' can tweak this then if it makes you happy.

Note that it turns out that the limiting execuitor isn't a simple drop-in replacement in some code, because the guided executor has template issues when I try to use it with the limiting one, so I'm working on that.

msimberg: Of course it makes me happy. That's what I live for.

hpx/lcos/local/detail/sliding_semaphore.hpp
```diff
@@ -35,6 +35,13 @@ namespace hpx { namespace lcos { namespace local { namespace detail
...
+                                std::int64_t lower_limit)
+        {
+            max_difference_ = max_difference;
+            lower_limit_    = lower_limit;
+        }
```
hkaiser: We will have to acquire a lock to set those values in any case. Otherwise, platforms with a weaker memory model might end up writing half of the values in racey situations.

biddisco:
Added a mutex. Hope this can be merged now.

hpx/lcos/local/sliding_semaphore.hpp
```diff
@@ -62,6 +62,21 @@ namespace hpx { namespace lcos { namespace local
...
+        ///                 [in] The max difference between the upper limit
+        ///                 (as set by wait()) and the lower limit (as set by
+        ///                 signal()) which is allowed without suspending any
+        ///                 thread calling wait().
+        /// \param lower_limit  [in] The initial lower limit.
+        void set_max_difference(std::int64_t max_difference,
+                std::int64_t lower_limit = 0)
+        {
+            std::unique_lock<mutex_type> l(mtx_);
+            sem_.set_max_difference(l, max_difference, lower_limit);
```
hkaiser: This will require to move the lock. Alternatively you can turn the first argument of `set_max_difference` into a reference.

biddisco: sorry. bad cut'n'paste. there are several mistakes in there actually (in the other functions too). good catch.

biddisco:
My bad. just that one mistake. fixed now. please check.

PR #3936, opened 2019-06-26
Fixing installation of binaries on windows

hkaiser:
- on windows, all executable code must go into CMAKE_INSTALL_BINDIR

### DIFF
--- a/cmake/HPX_AddComponent.cmake
+++ b/cmake/HPX_AddComponent.cmake
@@ -102,18 +102,34 @@ function(add_hpx_component name)
     set(exclude_from_all EXCLUDE_FROM_ALL)
   else()
     if(${name}_PLUGIN AND NOT HPX_WITH_STATIC_LINKING)
-      set(install_destination ${CMAKE_INSTALL_LIBDIR}/hpx)
+      if(MSVC)
+        set(library_install_destination ${CMAKE_INSTALL_BINDIR}/hpx)
+      else()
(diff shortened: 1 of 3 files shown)


cmake/HPX_AddComponent.cmake
```diff
@@ -102,18 +102,34 @@ function(add_hpx_component name)
...
+      set(archive_install_destination ${${name}_INSTALL_SUFFIX})
+      set(runtime_install_destination ${${name}_INSTALL_SUFFIX})
     endif()
     set(_target_flags
       INSTALL
       INSTALL_FLAGS
-        DESTINATION ${install_destination}
+        LIBRARY DESTINATION ${library_install_destination}
+        ARCHIVE DESTINATION ${archive_install_destination}
+        RUNTIME DESTINATION ${runtime_install_destination}
```
msimberg: Do we actually need `RUNTIME DESTINATION` here? Not that it hurts, but a component should never be an executable. Same in `HPX_AddLibrary.cmake`.

hkaiser: Indeed. That can be removed.

hkaiser: Correction: on Windows the `RUNTIME` path is required as its used to install the generated dll's.

msimberg: Ah, ok. Then it stays.

PR #3943, opened 2019-06-28
Remove example module

msimberg:
Move module structure overview to contribution documentation.


docs/sphinx/contributing/modules.rst
```diff
@@ -4,13 +4,13 @@
...
-==============
+================
+Module structure
+================
 
-This is an example module used to explain the structure of an |hpx| module.
+This section explains the structure of an |hpx| module.
 
 The tool `create_library_skeleton.py
 <https://github.com/STEllAR-GROUP/hpx/blob/master/libs/create_library_skeleton.py>`_
```
hkaiser: Shouldn't we at least add a sentence or two here? Now that we remove the example there is no 'go by example' anymore...

msimberg: What do you have in mind? I agree in principle, but `create_library_skeleton.py` essentially gives one a minimal example module.

hkaiser: Well, perhaps just a sentence explaining that running this script (how - any command line arguments?) generates a minimal set of files resembling an HPX module...

msimberg: Updated it to explain that and a bit of `add_hpx_module`.

msimberg:
I also updated `create_library_skeleton.py` with the `add_hpx_module` and force-linking changes.

docs/sphinx/contributing/modules.rst
hkaiser: Thanks!

PR #3946, opened 2019-06-28
[Modules] Move the location of the generated headers

aurianer:
- Move the generated headers into the `CMAKE_CURRENT_BIN_DIR`.
- Install those headers (including config/defines.hpp).
- Remove the duplicated `write_config_defines_file`.

### DIFF
--- a/cmake/HPX_AddModule.cmake
+++ b/cmake/HPX_AddModule.cmake
@@ -147,7 +147,6 @@ function(add_hpx_module name)
   target_link_libraries(hpx_${name} ${${name}_DEPENDENCIES})
   target_include_directories(hpx_${name} PUBLIC
     $<BUILD_INTERFACE:${HEADER_ROOT}>
-    $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}>
     $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>)
 
   if(HPX_${name_upper}_WITH_COMPATIBILITY_HEADERS)
--- a/cmake/HPX_AddModule.cmake
+++ b/cmake/HPX_AddModule.cmake
@@ -88,7 +88,7 @@ function(add_hpx_module name)
   # This header generation is disabled for config module specific generated
   # headers are included
   if (${name}_GLOBAL_HEADER_GEN)
-    set(global_header "${CMAKE_BINARY_DIR}/hpx/${name}.hpp")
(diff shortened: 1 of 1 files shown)


cmake/HPX_AddModule.cmake
```diff
@@ -152,34 +153,30 @@ function(add_hpx_module name)
...
+  )
   
   write_config_defines_file(
     NAMESPACE ${name_upper}
-    FILENAME "${CMAKE_BINARY_DIR}/hpx/${name}/config/defines.hpp")
-
-  write_config_defines_file(
-    NAMESPACE ${name_upper}
-    FILENAME "${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/hpx/${name}/config/defines.hpp")
+    FILENAME "${CMAKE_CURRENT_BINARY_DIR}/include/hpx/${name}/config/defines.hpp")
```
hkaiser: Does this has to come before the `install()` above in order to be properly picked up? I'm not sure...

cmake/HPX_AddModule.cmake
```diff
@@ -108,7 +108,7 @@ function(add_hpx_module name)
   target_link_libraries(hpx_${name} ${${name}_DEPENDENCIES})
   target_include_directories(hpx_${name} PUBLIC
     $<BUILD_INTERFACE:${HEADER_ROOT}>
-    $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}>)
+    $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>)
```
msimberg: I think we still need `${CMAKE_BINARY_DIR}` here, *and* `${CMAKE_CURRENT_BINARY_DIR}/include` for the generated headers from this module.

cmake/HPX_AddModule.cmake
aurianer: Yep sorry!

cmake/HPX_AddModule.cmake
aurianer: Yep !

aurianer: Could you just explain to me why `$<BUILD_INTERFACE:${CMAKE_BINARY_DIR}>` is still necessary ?

aurianer:
Ready to go, but one test is failing because of a timeout.

cmake/HPX_AddModule.cmake
msimberg: For the record, we concluded that it isn't necessary.

PR #3949, opened 2019-06-30
This adds a configuration file that will cause for stale issues to be automatically closed

hkaiser:
- The time frames currently set will consider a PR as stale after 180 and will
  close it after another month.

Marking issues with 'tag: pinned' keeps them open. If we decided to go for this we should first mark all issues that shouldn't be closed.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,27 @@
+# Copyright (c) 2019 Hartmut Kaiser
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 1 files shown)


hkaiser:
FWIW, I went through older tickets and have marked some of them as 'pinned' (so they won't get closed). @heller,  @biddisco, @msimberg Please mark tickets dear to you as well.

msimberg:
I'll merge this already now. @sithhell, @biddisco if you haven't had time look through the issues you'll still have 6 months to do so and a 1 month warning before an issue is closed.

PR #3954, opened 2019-07-02
Move components to separate subdirectory in root of repository

msimberg:
This moves the components out from the main `hpx` and `src` directories into a `components` subdirectory with a structure similar to that of the modules. Eventually they could even all go in the same subdirectory but keeping them separate for now since they are treated somewhat differently.


hkaiser:
Good thinking. All components build cleanly on top of HPX, as such they are add-ons. I don't think they should be integrated with the modules as we have them today as those are the parts the main library is composed of.

msimberg:
Yeah, there are definitely differences in use. But the line becomes a bit more blurry especially if parts of HPX become optional. Then the optional libraries are also just libraries that sit cleanly on top of core HPX. Some optional libraries would provide components and some would not. `add_hpx_component` creates a shared (component) library which provides one or more (HPX) components. Using the word component in two different contexts could be confusing to some. Anyway, not changing any of this now! Just providing some food for thought...

msimberg:
Not quite ready to go yet... The include directories need fixing.

msimberg:
All right, local testing looks promising. This would be ready for re-review. I've for now duplicated the `create_library_skeleton.py` script to do roughly the equivalent for components. I expect them to converge over time. I also did some refactoring of the CMake functions for header tests. Hopefully a bit less duplication now.

components/component_storage/tests/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) 2019 The STE||AR-Group
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+include(HPX_Option)
```
hkaiser: Is this needed here? Shouldn't be the target-specific modules be included instead?

components/containers/partitioned_vector/tests/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) 2019 The STE||AR-Group
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+include(HPX_Option)
```
hkaiser: Same here, is this the correct module being included?

components/component_storage/tests/CMakeLists.txt
msimberg: Not strictly needed. Do you mean by target-specific modules CMake modules that are in `components/X/cmake`? Including this here would make sense if we wanted to build components standalone, but I'm not sure that makes sense (at least right now) so I'll remove it for now. The modules do the same thing here to be able build them standalone, but that still needs testing. Also, the modules are subprojects while the components are not. And the modules set `CMAKE_MODULE_PATH` so that it could actually work standalone. All in all, components are not ready for this yet, so I'll remove it.

components/containers/partitioned_vector/tests/CMakeLists.txt
msimberg: Same as above.

PR #3957, opened 2019-07-03
Algorithms module

msimberg:
Adds an algorithms module (headers from `hpx/parallel/algorithms` and `container_algorithms`). Segmented algorithms will go in a separate module. No compatibility headers this time since I didn't move the headers. Not sure yet whether to move them.

Note: this is another slightly naughty module like `collectives` but works out since it's a header only module.

Flyby: remove global include and link directories (might break things, testing to see what...).


libs/algorithms/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+  hpx/parallel/container_algorithms/minmax.hpp
+  hpx/parallel/container_algorithms/merge.hpp
+  hpx/parallel/container_algorithms/is_heap.hpp
+  hpx/parallel/container_algorithms/generate.hpp
+  hpx/parallel/container_algorithms/for_each.hpp
+  hpx/parallel/container_algorithms/find.hpp
+  hpx/parallel/container_algorithms/fill.hpp
+  hpx/parallel/container_algorithms/count.hpp
+  hpx/parallel/container_algorithms/copy.hpp
+  hpx/parallel/container_algorithms/all_any_none.hpp
```
hkaiser: Should we separate the container algorithms?

libs/algorithms/tests/unit/algorithms/CMakeLists.txt
```diff
@@ -140,7 +140,7 @@ foreach(test ${tests})
 
   source_group("Source Files" FILES ${sources})
 
-  set(folder_name "Tests/Unit/Parallel/Algorithms")
+  set(folder_name "Tests/Unit/Modules/Algorithms/Algorithms")
```
hkaiser: Probably a typo...

msimberg:
> The general question is whether this is a 'module' (i.e. is a part of the core library) or a 'component' (i.e. builds on top of and independently of the main library). And yes 'module' and 'component' are arbitrary and confusing names - but I hope you see what I mean...
> 
> Also, should distinguish between the two categories at all?

That's kind of what I was getting at in #3954. With the arbitrary and confusing terminology this is a "component", but I'd call them either "modules" or "libraries" in the end. Then it's just a matter of which ones are included by default and which ones are optional. The main reason I've made this a "module" now is that it has been part of the main HPX library so far and it would be a breaking change to make it a "component". At some point I do want to make that breaking change of making it optional, but not now.

libs/algorithms/tests/unit/algorithms/CMakeLists.txt
msimberg: Intentional, but maybe a bad choice. The module is now called algorithms, and it has two sub-somethings called algorithms and container algorithms, so the last `Algorithms` just distinguishes between the two. I should probably just collapse them into just one namespace.

libs/algorithms/CMakeLists.txt
msimberg: I considered this, but since the container algorithms are more or less just overloads of the algorithms (correct me if I'm wrong) I thought they fit together. The other reason is that it's all headers so lumping them together doesn't really cost anything in compile time (this is not a good reason in general though...). I'm open to having them separate as well.

msimberg:
I rebased this on top of #3954 to avoid later conflicts. That should be merged first.

hkaiser:
@msimberg shouldn't we actually move the headers if we turn this into a module?

hkaiser:
@msimberg The header "hpx/parallel/numeric.hpp" belongs to this module as well (i.e. the equivalent header for <numeric>).

libs/algorithms/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+include(HPX_AddModule)
+add_hpx_module(algorithms
+    DEPRECATION_WARNINGS
+    INSTALL_BINARIES OFF
+    FORCE_LINKING_GEN ON
+    HEADERS ${algorithms_headers}
+    # TODO: Really depends on main HPX library, but cannot add that as a
+    # dependency yet. Need to have the parts that algorithms depends on in
+    # separate modules first.
+    DEPENDENCIES
```
hkaiser: Shouldn't this module depend on hpx_config?

hkaiser: Most likely it depends on hpx_assert as well.

libs/algorithms/tests/unit/algorithms/CMakeLists.txt
hkaiser: @msimberg this should either be changed to "Tests/Modules/Unit/Algorithms/Algorithms" or we should change the folder names for the modules hpx_assertion, hpx_cache, and hpx_collectives. The same is probably true for the regression and perf tests.

hkaiser:
I think the regression tests have not been moved to the module.

libs/algorithms/CMakeLists.txt
msimberg: Yes, you're right. Thanks!

libs/algorithms/tests/unit/algorithms/CMakeLists.txt
msimberg: Good catch. The test names are `tests.unit.modules....` so I'd rather change the folder names in the other modules because I find it slightly more useful to do `make tests.unit` than `tests.modules`. We could of course have both but that could get messy... Any preference? If not I'll stick with `Tests/Unit/Modules` and `tests.unit.modules`.

msimberg:
@hkaiser I think I've addressed all your comments (folder names changed, but can be changed again if you prefer something else). I moved the numeric and memory headers as well, but not `hpx/include/numeric.hpp`, `hpx/include/memory.hpp` and all the `hpx/include/<algorithm_name>` headers since I think they can be just convenience headers provided by the core HPX library (and some of them include headers from segmented algorithms, which I'd like to keep separate).

msimberg:
Note that I also added the performance tests for the algorithms module as unit tests. This means they'll run when you do `ctest`. I wouldn't mind that but I can imagine some people would mind. Are there strong opinions against this?

libs/algorithms/tests/unit/algorithms/CMakeLists.txt
hkaiser: I have no preference, just would like to have it consistent.

PR #3960, opened 2019-07-04
fix compilation with CUDA 10 and GCC 6

gentryx:
Avoid clashing of public classes (e.g. formatter) and template parameters (e.g. formatter) by renaming the template parameters.

This was a problem when compiling host code with NVCC (from NVIDIA CUDA 10) with GCC 6 as the host compiler.


gentryx:
Renamed template parameters as requested.

msimberg:
The test failure is unrelated. Thanks @gentryx!

gentryx:
Thanks Mikael!

PR #3961, opened 2019-07-05
Making hpx::util::tuple<Ts...> and std::tuple<Ts...> convertible

### LINKED ISSUE #1403: Make std::tuple tuple-like for HPX
This would enable seamless use of std::tuple with all HPX API functions which expect or return a hpx::util::tuple.

sithhell:
Fixes #1403 by allowing the use of hpx::util::get with std::tuple and
std::get with hpx::util::tuple::get. Same goes for the utilities
tuple_size and tuple_element.
An additional conversion operator from hpx::util::tuple<Ts..> to
std::tuple<Ts..> has been added to make the two types interchangable.


hpx/util/tuple.hpp
```diff
@@ -35,7 +36,39 @@ namespace hpx { namespace util
...
+        return hpx::util::tuple_element<I, hpx::util::tuple<Ts...>>::get(tuple);
+    }
+
+    template <std::size_t I, typename ...Ts>
+    typename hpx::util::tuple_element<I, hpx::util::tuple<Ts...>>::type const&
+    get(hpx::util::tuple<Ts...> const& tuple)
+    {
+        return hpx::util::tuple_element<I, hpx::util::tuple<Ts...>>::get(tuple);
+    }
+}
```
hkaiser: Should we add overloads for rvalue tuples as well?

hpx/util/tuple.hpp
```diff
@@ -303,6 +336,18 @@ namespace hpx { namespace util
...
+            operator std::tuple<Ts...>() const &
+            {
+                return std::make_tuple(get<Is>()...);
+            }
+
+            HPX_HOST_DEVICE
+            operator std::tuple<Ts...>() &&
+            {
+                return std::make_tuple(std::move(get<Is>())...);
+            }
```
hkaiser: Do we need overloads for lvalue references and const rvalue references as well?

hpx/util/tuple.hpp
```diff
@@ -500,6 +550,16 @@ namespace hpx { namespace util
...
+
+        operator std::tuple<Ts...>() const&
+        {
+            return _impl;
+        }
+
+        operator std::tuple<Ts...>() &&
+        {
+            return _impl;
+        }
```
hkaiser: Same here - do we need the overloads for lvalue and const rvalue references?

hpx/util/tuple.hpp
```diff
@@ -656,6 +721,24 @@ namespace hpx { namespace util
...
+        {
+            return std::get<I>(tuple);
+        }
+
+        static HPX_CONSTEXPR HPX_HOST_DEVICE HPX_FORCEINLINE type const&
+        get(std::tuple<Ts...> const& tuple) noexcept
+        {
+            return std::get<I>(tuple);
+        }
+    };
```
hkaiser: Overloads for rvalue refs?

hpx/util/tuple.hpp
K-ballo: Note that adding these overloads to namespace `std` is UB. If they are still wanted, then there should be 4 of them to match `std::` expected interface.

hpx/util/tuple.hpp
```diff
@@ -303,6 +336,18 @@ namespace hpx { namespace util
                         I, typename detail::at_index<I, Ts...>::type
                     > const&>(*this).value();
             }
+
+            HPX_HOST_DEVICE
+            operator std::tuple<Ts...>() const &
+            {
+                return std::make_tuple(get<Is>()...);
```
K-ballo: `std::make_tuple` decays and unwraps, it will yield the wrong `std::tuple` specialization.

hpx/util/tuple.hpp
K-ballo: Not here, the two rvalue ref overloads are synthesized by `util::get`.

hpx/util/tuple.hpp
sithhell: The fact that it is UB can be neglected here. There should be no consequences for existing implementations or any implementation in the future.

sithhell: The overloads have been added.

hpx/util/tuple.hpp
sithhell: So returning `std::tuple<Ts...>(get<Is>()...)` it is.

msimberg:
@sithhell would you have time to look at the build failures?

sithhell:
This should be ready to go now.

libs/datastructures/include/hpx/datastructures/tuple.hpp
```diff
@@ -1017,4 +1092,10 @@ namespace hpx { namespace util {
 #pragma warning(pop)
 #endif
 
+#if defined(__clang__)
+#  pragma clang diagnostic pop
+#elif defined (__GNUC__)
+#  pragma GCC diagnostic pop
+#endif
```
hkaiser: This `GCC diagnostic pop` has no corresponding `push`

libs/datastructures/tests/unit/any.cpp
```diff
@@ -6,10 +6,9 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 #include <hpx/datastructures/any.hpp>
-#include <hpx/hpx_init.hpp>
 #include <hpx/testing.hpp>
 
-#include <hpx/util/storage/tuple.hpp>
+#include <hpx/datastructures/tuple.hpp>
```
hkaiser: This is not correct. `util/storage/tuple.hpp` is distinct from `datastructures/tuple.hpp`.

sithhell: Indeed, `util/storage/tuple.hpp` is a strange thing, more a any_vector and not used in this test...

libs/datastructures/include/hpx/datastructures/tuple.hpp
sithhell: Good catch...

libs/datastructures/tests/unit/any.cpp
sithhell: The include has been removed...

libs/datastructures/include/hpx/datastructures/tuple.hpp
```diff
@@ -54,7 +60,26 @@ namespace hpx { namespace util {
...
+    struct tuple_size<hpx::util::tuple<Ts...>>
+      : public hpx::util::tuple_size<hpx::util::tuple<Ts...>>
+    {};
+
+    template <std::size_t I, typename... Ts>
+    struct tuple_element<I, hpx::util::tuple<Ts...>>
+      : public hpx::util::tuple_element<I, hpx::util::tuple<Ts...>>
+    {};
+
+    using hpx::util::get;
```
hkaiser: Isn't this a bit dangerous? This relies on the std library itself using `get<>()` in an unqualified way. Is that true for all std libraries we care about? Besides, doesn't this associate the HPX namespace with `namespace std`; could this have other implications?

At the very least I would suggest to only conditionally enable this integration, such that users would have to explicitly buy-in. 

I'm not sure if there are other ways to achieve this, but if there are, let's consider less dangerous techniques instead.

libs/datastructures/include/hpx/datastructures/tuple.hpp
```diff
@@ -306,6 +331,18 @@ namespace hpx { namespace util {
...
+            operator std::tuple<Ts...>() const &
+            {
+                return std::make_tuple(get<Is>()...);
+            }
+
+            HPX_HOST_DEVICE
+            operator std::tuple<Ts...>() &&
+            {
+                return std::make_tuple(std::move(get<Is>())...);
+            }
```
hkaiser: Didn't we at some point have also overloads for `&` and `const &&` ? Don't we need those?

libs/datastructures/include/hpx/datastructures/tuple.hpp
```diff
@@ -502,6 +544,16 @@ namespace hpx { namespace util {
...
+
+        operator std::tuple<Ts...>() const&
+        {
+            return _impl;
+        }
+
+        operator std::tuple<Ts...>() &&
+        {
+            return _impl;
+        }
```
hkaiser: Same here: don't we also need overloads for `&` and `const&&`?

libs/datastructures/include/hpx/datastructures/tuple.hpp
```diff
@@ -667,6 +724,24 @@ namespace hpx { namespace util {
...
+        {
+            return std::get<I>(tuple);
+        }
+
+        static HPX_CONSTEXPR HPX_HOST_DEVICE HPX_FORCEINLINE type const&
+        get(std::tuple<Ts...> const& tuple) noexcept
+        {
+            return std::get<I>(tuple);
+        }
+    };
```
hkaiser: What about overloads for rvalue references here?

libs/datastructures/include/hpx/datastructures/tuple.hpp
sithhell: The `using` was my interpretation of: http://irclog.cct.lsu.edu/ste~b~~b~ar/2019-07-06#239285;

libs/datastructures/include/hpx/datastructures/tuple.hpp
sithhell: Good point... don't know why I lost them ... was probably due to resolving the merge conflicts :/

PR #3963, opened 2019-07-05
Fix problem export modules when HPX_WITH_APEX ON

aurianer:
Needs the addition of hpx_assertion as a dependency in apex project (open PR).
- Export modules when HPX_WITH_APEX = ON
- Add the missing EXPORT option in `add_hpx_module` function


aurianer:
Ok actually I was pointing on a old version of apex, so I had some old errors that have been fixed (by not exporting apex) whereas I just exported the modules, but not exporting apex is more smart ^^ I'll close this !

PR #3964, opened 2019-07-05
Fixing merge problems

hkaiser:
The headers in the collectives module where added concurrently to the assertion module, thus where missed while fixing the related #include changes.


hkaiser:
I'll go ahead and merge this to make master green on CircleCI.

PR #3965, opened 2019-07-08
Change current_function_helper back to non-constexpr

msimberg:
void constexpr functions not supported in C++11 (https://cdash.cscs.ch/viewBuildError.php?buildid=59491).

### DIFF
--- a/libs/assertion/include/hpx/assertion/current_function.hpp
+++ b/libs/assertion/include/hpx/assertion/current_function.hpp
@@ -12,9 +12,11 @@
 //
 //  Adapted to HPX naming scheme
 
+#include <hpx/config.hpp>
+
 namespace hpx { namespace assertion { namespace detail {
 
-    constexpr void current_function_helper()
+    HPX_CXX14_CONSTEXPR inline void current_function_helper()
     {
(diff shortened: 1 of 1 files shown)


libs/assertion/include/hpx/assertion/current_function.hpp
```diff
@@ -14,7 +14,7 @@
 
 namespace hpx { namespace assertion { namespace detail {
 
-    constexpr void current_function_helper()
+    void current_function_helper()
     {
```
hkaiser: We have `HPX_CXX14_CONSTEXPR` for these cases.

msimberg: Doh, you're right. Will change.

msimberg: Done.

msimberg:
This should be happier now. Do we always want to fall back to `inline` if constexpr is not available?

hkaiser:
> This should be happier now. Do we always want to fall back to inline if constexpr is not available?

We probably have to in order to avoid linker errors.

PR #3966, opened 2019-07-08
Add missing export option in add_hpx_module

aurianer:
The EXPORT option was used but not parsed.


aurianer:
I tried to stay coherent with the `COMPATIBILTY_HEADERS` and `INSTALL_BINARIES` with are ON/OFF args, but it's up to you, I really don't know what's better

hkaiser:
> I tried to stay coherent with the COMPATIBILTY_HEADERS and INSTALL_BINARIES with are ON/OFF args, but it's up to you, I really don't know what's better

Yah, those are `options` as well, I think... Specifying `ON`/`OFF` for those seems to be redundant. But I'm happy to be convinced otherwise. This is not a real issue.

aurianer:
> Yah, those are `options` as well, I think... Specifying `ON`/`OFF` for those seems to be redundant. But I'm happy to be convinced otherwise. This is not a real issue.

Ok I'll change that !

PR #3967, opened 2019-07-09
Remove leftover PIC settings from main CMakeLists.txt

msimberg:
Forgot to remove this from the assertion PR after it was added as a per-target option (for modules).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -304,10 +304,6 @@ else()
   set(hpx_library_link_mode SHARED)
 endif()
 
-if (NOT MSVC)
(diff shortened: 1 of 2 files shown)


msimberg:
> Would this still be ok if the core libraries are built statically?

Good question. I think yes, since PIC is only needed when we're creating a shared library. We shouldn't need PIC at all when everything is a static library. But I'll have to try it out to be sure.

msimberg:
1. We had another `set(CMAKE_POSITION_INDEPENDENT_CODE ON)` in `src/CMakeLists.txt`. I removed that and made it target specific instead (`hpx_init`).

2. Static linking doesn't even work on master. See #3970.

PR #3971, opened 2019-07-10
Add testing module

aurianer:


libs/testing/docs/index.rst
```diff
@@ -0,0 +1,13 @@
...
+   Distributed under the Boost Software License, Version 1.0. (See accompanying
+   file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+.. _libs_testing:
+
+=======
+testing
+=======
+
+The testing module contains useful macros for testing like `HPX_TEST` for example.
```
msimberg: `HPX_TEST` should actually be surrounded by double backticks. My bad. @aurianer would you mind updating that?

hkaiser:
@msimberg I'll have a look. Please give me a day or so.

hkaiser:
@aurianer, @msimberg I took the liberty to push a couple of changes to this branch. Please feel free to roll back whatever you feel is not the way you'd like it:

- flyby: reformat files using clang-format (sorry)
- flyby: replace `BOOST_CURRENT_FUNCTION` with `HPX_ASSERT_CURRENT_FUNCTION`, add missing #include
- flyby: move assertion tests into module folder

Exporting the symbols is required however, except if we decided to not change the way the testing infrastructure code is linked. Before, it was linked directly with the application (if necessary, it was part of hpx_init.lib). Now, the testing code is part of the core library.

hkaiser:
@aurianer, @msimberg As said before, this PR changes the way the testing infrastructure is linked. It is now part of the core HPX library while before it was linked statically to tests only. While this is a not a significant change I would like to understand the rationale behind this decision.

msimberg:
@hkaiser ignorance on my part. Thanks for pointing it out! Was it not implicitly linked to all users of HPX? In any case, I think it should be separate from `hpx_init` at least. As a module we'll eventually be able to optionally link it to executables that need it (tests...), but having it in the core library wouldn't hurt now, would it? I don't have any strong preference, I just didn't realize that it used to be part of `hpx_init` before. Sorry for the confusion.

Also, thanks for the fixes!

PR #3972, opened 2019-07-11
Add condition to clone apex + use of new cmake var APEX_ROOT

aurianer:
Allow compilation with an existing apex directory outside of the hpx source directory. Need to specify a cmake APEX_ROOT option to point at this folder, if not specified, the normal clone is done.

Easier to work on several modules, I'll have one apex worktree per module, so that I don't have to change the headers and dependencies every time I switch between them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1691,15 +1691,22 @@ if(HPX_WITH_HPXMP)
     set(_hpxmp_no_update NO_UPDATE)
   endif()
 
-  # handle hpxMP library
-  include(GitExternal)
-  git_external(hpxmp
-    https://github.com/STEllAR-GROUP/hpxMP.git
-    ${HPX_WITH_HPXMP_TAG}
-    ${_hpxmp_no_update}
-    VERBOSE)
+  if (NOT HPXMP_ROOT)
+    # handle hpxMP library
+    include(GitExternal)
+    git_external(hpxmp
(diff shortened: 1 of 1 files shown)


msimberg:
Looks good, but you might want to do the same with hpxMP. It has exactly the same structure.

CMakeLists.txt
```diff
@@ -1628,24 +1628,33 @@ if(HPX_WITH_APEX)
...
+      hpx_error("Apex could not be found and HPX_WITH_APEX=On")
+    endif()
+  endif(NOT APEX_ROOT)
+
+  if (NOT APEX_SOURCE_DIR)
+    set(APEX_SOURCE_DIR ${APEX_ROOT}/src/apex)
   endif()
+  LIST(APPEND CMAKE_MODULE_PATH "${APEX_ROOT}/cmake/Modules")
+  include_directories(${APEX_SOURCE_DIR})
+  add_subdirectory(${APEX_SOURCE_DIR} ${CMAKE_BINARY_DIR}/apex/src/apex)
```
msimberg: I know this will work, but I feel a bit uncomfortable defining `APEX_SOURCE_DIR` here. It's actually defined by APEX itself as well and could in principle change once you do `add_subdirectory(${APEX_SOURCE_DIR} ...)`. Maybe you can just do
```
add_subdirectory(${APEX_ROOT}/src/apex ...)
include_directories(${APEX_SOURCE_DIR})
```
?

PR #3975, opened 2019-07-12
Enable deprecation warnings on CircleCI

msimberg:
Testing how much fails. Some included headers still use the old compatibility headers from modules, will fix at least those.


msimberg:
This fixes the unused variable warnings that show up now on the `daint-clang-newest` pycicle builder (I changed it to release mode).

tests/unit/threads/lockfree_fifo.cpp
```diff
@@ -52,7 +52,7 @@ void worker_thread(std::uint64_t num_thread)
     for (std::uint64_t i = 0; i < items; ++i)
     {
         bool result = get_next_thread(num_thread);
-        BOOST_TEST(result);
+        HPX_TEST(result);
```
hkaiser: Excellent catch!

PR #3978, opened 2019-07-15
Fix clang-tidy step on CircleCI

msimberg:

The clang-tidy checks have been disabled (unintentionally) on CircleCI since I changed it to do use a shallow clone. We still need the shallow clone to avoid copying the whole repo between steps, but I've changed it to do a full clone just for the clang-tidy step.


sithhell:
TBH, I think it would be best to use the `CMAKE_CXX_CLANG_TIDY` functionality and remove the script entirely. This catches more errors and doesn't need to be maintained...
Just make sure that the `.clang-tidy` file is being used.

msimberg:
Good idea. I didn't realize there was such an option to CMake. I'll update it to use that instead.

hkaiser:
Does `CMAKE_CXX_CLANG_TIDY` work if we only have a shallow copy of our repo? Or will this always run clang-tidy on all targets?

msimberg:
This will run it on all targets, so no need for a full clone. It's potentially a bit slower to do it like this (haven't timed it) but gets rid of the additional script.

hpx/runtime/parcelset/parcelport_impl.hpp
```diff
@@ -913,6 +913,7 @@ namespace hpx { namespace parcelset
             else
             {
                 ++operations_in_flight_;
+                // NOLINTNEXTLINE(bugprone-use-after-move)
                 HPX_ASSERT(num_parcels < parcels.size());
```
msimberg: This is a false positive, but I'm wondering if we couldn't avoid it. @sithhell it seems like clang-tidy is complaining because it can't see that the lambda which evaluates the expression is evaluated right away. I'm wondering what the lambda gives us? I don't see immediately why we couldn't just evaluate the expression right away. (I'm talking about this: https://github.com/STEllAR-GROUP/hpx/blob/2d1581bc6a01c47778c4380193b6b6b613d78fb0/libs/assertion/include/hpx/assertion.hpp#L54-L59.)

sithhell: The lambda eliminates unused variable warnings, iirc. I'm a little clueless why this warning appears in this case though...

msimberg: Maybe it did in some earlier version, but in release mode it definitely won't stop unused variable warnings. I think it's a fair, but somewhat paranoid warning. I'll keep it as it is for now.

libs/collectives/include/hpx/collectives/all_reduce.hpp
```diff
@@ -444,7 +444,7 @@ namespace hpx { namespace lcos
 
         return all_reduce(
             hpx::find_from_basename(std::move(name), root_site),
-            std::move(local_result), std::forward<F>(op), this_site);
+            std::forward(local_result), std::forward<F>(op), this_site);
```
msimberg: @hkaiser could you please check that the move here and in `all_to_all.hpp` weren't intentional?

hkaiser: The `forward` looks like the right thing to use. Thanks for fixing this!

PR #3980, opened 2019-07-15
Module logging

aurianer:


msimberg:
A bit of explanation: The logging functionality is now split up into the basic logging module which defines logs, but not their behaviour, and the initialization of logging (`init_logging.cpp`) which actually sets up the logs (based on configuration options) and sets the formatters (which use runtime information; thread ids, localities, etc.).

aurianer:
I set clang-format `ReflowComments` option to true otherwise, inspect was failing on some comment.
And I set `FixNamespaceComments` to true so that we have the name of the namespace as a comment at the closing bracket

hkaiser:
> I set clang-format ReflowComments option to true otherwise, inspect was failing on some comment.

I'd suggest not to do that. This completely destroys doxygen comments, IIRC. I'd rather fix inspect to be silent about the problems it reports otherwise.

hkaiser:
> And I set FixNamespaceComments to true so that we have the name of the namespace as a comment at the closing bracket

How do people feel about this? What's the benefit of this? (note: I'm impartial here)

msimberg:
> > And I set FixNamespaceComments to true so that we have the name of the namespace as a comment at the closing bracket
> 
> How do people feel about this? What's the benefit of this? (note: I'm impartial here)

I slightly prefer it over not having the comment. It makes knowing what scope you're looking at slightly easier. But I don't really mind either way.

aurianer:
I temporarily removed `ExperimentalAutoDetectBinPacking` cause it was messing with some lines of code, what do you think of that ?

hkaiser:
> I temporarily removed ExperimentalAutoDetectBinPacking cause it was messing with some lines of code, what do you think of that ?

Would you have an example of what happened?

aurianer:
> > I temporarily removed ExperimentalAutoDetectBinPacking cause it was messing with some lines of code, what do you think of that ?
> 
> Would you have an example of what happened?

This for example (on `copy.hpp` the algorithms module):
```
-                              hpx::shared_future<std::size_t> curr,
-                              hpx::shared_future<std::size_t> next) mutable {
+                             hpx::shared_future<std::size_t>
+                                  curr,
+                             hpx::shared_future<std::size_t>
+                                  next) mutable {
```

hkaiser:
@aurianer Yeah, this change makes perfect sense!

aurianer:
Cool thanks !

msimberg:
I just wanted to add that clang-format was not just wrapping it unnecessarily early, but it was in fact not idempotent, i.e. it would wrap the lines as in the example the second time clang-format is applied.

aurianer:
> I just wanted to add that clang-format was not just wrapping it unnecessarily early, but it was in fact not idempotent, i.e. it would wrap the lines as in the example the second time clang-format is applied.

aurianer:
sorry ^^

PR #3981, opened 2019-07-15
Segmented algorithms module

msimberg:
Same deal as #3957 (also depends on it), but for segmented algorithms.


hkaiser:
The following file should be added to this module:
```
#include <hpx/traits/segmented_iterator_traits.hpp> 
```

msimberg:
> The following file should be added to this module:
> 
> ```
> #include <hpx/traits/segmented_iterator_traits.hpp> 
> ```

Thanks, good catch! I'll update this first thing on Monday.

libs/algorithms/CMakeLists.txt
```diff
@@ -124,6 +124,6 @@ add_hpx_module(algorithms
     # TODO: Really depends on main HPX library, but cannot add that as a
     # dependency yet. Need to have the parts that algorithms depends on in
     # separate modules first.
-    DEPENDENCIES hpx_assertion hpx_config hpx_timing
+    DEPENDENCIES hpx_assertion hpx_config hpx_iterator_support hpx_timing
```
hkaiser: Wouldn't this create a circular dependency between the `algorithms` and `segmented_algorithm` modules?

msimberg: Should not as far as I can tell, but I may have missed something. What is the cycle you're thinking about? FYI, @aurianer is working on getting the checks for cyclic dependencies. I'll have a look in the morning to make sure there's no cycle.

hkaiser: Hmmm, not sure what I was thinking. There is no cycle for sure - my brain was trying to convince me that there was one, however ;-) Sorry for the noise.

PR #3982, opened 2019-07-15
Parallel executors module

msimberg:
Same deal as #3957 and #3981 (depends on both), but for the executors in `hpx::parallel::execution`. The `algorithms` and `segmented_algorithms` modules depend on this module.


hkaiser:
The following files should be added to this module:
```
#include <hpx/traits/executor_traits.hpp>
#include <hpx/traits/is_execution_policy.hpp>
#include <hpx/traits/is_executor.hpp>
#include <hpx/traits/is_timed_executor.hpp>
#include <hpx/traits/is_executor_parameters.hpp>
```

PR #3985, opened 2019-07-15
Module timing

aurianer:
Timing utilities.


aurianer:
Good to go ! :)

libs/algorithms/tests/performance/CMakeLists.txt
```diff
@@ -17,6 +17,16 @@ set(benchmarks
...
+ set(benchmark_inplace_merge_FLAGS DEPENDENCIES hpx_timing)
+ set(benchmark_is_heap_FLAGS DEPENDENCIES hpx_timing)
+ set(benchmark_is_heap_until_FLAGS DEPENDENCIES hpx_timing)
+ set(benchmark_merge_FLAGS DEPENDENCIES hpx_timing)
+ set(benchmark_partition_copy_FLAGS DEPENDENCIES hpx_timing)
+ set(benchmark_partition_FLAGS DEPENDENCIES hpx_timing)
+ set(benchmark_remove_FLAGS DEPENDENCIES hpx_timing)
+ set(benchmark_remove_if_FLAGS DEPENDENCIES hpx_timing)
+ set(benchmark_unique_FLAGS DEPENDENCIES hpx_timing)
+ set(benchmark_unique_copy_FLAGS DEPENDENCIES hpx_timing)
```
msimberg: Hmm, this *should* not be needed just like `config`, `assertion`, `format`, etc. are not needed. Why is this a special case?

libs/timing/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+  hpx/util/scoped_timer.hpp
+  hpx/util/steady_clock.hpp
+)
+
+# cmake will not create a separate VS project without any source files, thus
+# this adds a dummy (empty) source file to the target
+# Default location is $HPX_ROOT/libs/timing/src
+set(timing_sources
+  timing.cpp
+)
```
msimberg: This one can actually go since the generated `force_linking.cpp` file is there.

libs/algorithms/tests/performance/CMakeLists.txt
aurianer: Right !

msimberg:
@aurianer did you already apply clang-format to this? If not, please do so (once we've decided on the format) and we can merge this :)

aurianer:
> @aurianer did you already apply clang-format to this? If not, please do so (once we've decided on the format) and we can merge this :)

Yep :)

Issue #3940, opened 2019-06-27
Convert pseudo-doxygen function documentation into actual doxygen documentation


msimberg:
We have several comments which are already almost in the style of doxygen, but don't show up because of the wrong type of comment (`//` instead of `///`) and the comment itself doesn't follow doxygen syntax. Example: https://github.com/STEllAR-GROUP/hpx/blob/c1317e3c664dc14c00d320568f29080ffda5aac0/hpx/lcos/promise.hpp#L70-L145. Related to #985.

PR #3990, opened 2019-07-19
Documentation tags fixes

tapaswenipathak:
Fixes https://github.com/STEllAR-GROUP/hpx/issues/3940

## Proposed Changes
Documentation tags fixes


hpx/lcos/promise.hpp
```diff
@@ -167,97 +167,97 @@ namespace lcos {
...
+        /// \brief Abandons any shared state
         ~promise()
         {}
 
-        // Effects: Abandons any shared state (30.6.4) and then as if
-        //          promise(std::move(other)).swap(*this).
-        // Returns: *this.
+        /// \brief Abandons any shared state (30.6.4) and then as if
+        ///          promise(std::move(other)).swap(*this).
+        /// \return *this.
```
msimberg: Please be consistent with `\return` and `\returns`. I think I prefer `\returns` as it reads more naturally (and it's what we use in most places in the codebase).

hpx/lcos/promise.hpp
```diff
@@ -67,90 +67,90 @@ namespace lcos {
...
-        // Effects: constructs a promise object and a shared state.
+        /// \brief constructs a promise object and a shared state.
         promise()
           : base_type()
         {}
 
-        // Effects: constructs a promise object and a shared state. The
-        // constructor uses the allocator a to allocate the memory for the
-        // shared state.
+        /// \breif constructs a promise object and a shared state. The
```
msimberg: Typo, please fix.

hpx/lcos/promise.hpp
```diff
@@ -67,90 +67,90 @@ namespace lcos {
...
-        // Throws: future_error if *this has no shared state or if
-        //         get_shared_future has already been called on a promise
-        //         with the same shared state as *this.
-        // Error conditions:
-        //   - future_already_retrieved if get_shared_future has already been
-        //     called on a promise with the same shared state as *this.
-        //   - no_state if *this has no shared state.
+        /// \return A shared_future<Result> object with the same shared state
+        ///          as *this.
+        /// \throw future_error if *this has no shared state or if
```
msimberg: I think `\throws` reads better. Could you change all `\throw` to `\throws`?

hpx/lcos/promise.hpp
```diff
@@ -67,90 +67,90 @@ namespace lcos {
...
-        //     exception.
-        // Error conditions:
-        //   - promise_already_satisfied if its shared state already has a
-        //     stored value or exception.
-        //   - no_state if *this has no shared state.
+        /// \brief Effects atomically stores the value r in the shared state and makes
+        ///          that state ready (30.6.4).
+        /// \throw
+        ///   - future_error if its shared state already has a stored value or
+        /// \exception    exception.
```
msimberg: I think `\exception` doesn't fit here, or am I missing something?

hpx/lcos/promise.hpp
```diff
@@ -67,90 +67,90 @@ namespace lcos {
...
         promise(std::allocator_arg_t, Allocator const& a)
           : base_type(std::allocator_arg, a)
         {}
 
-        // Effects: constructs a new promise object and transfers ownership of
-        //          the shared state of other (if any) to the newly-
-        //          constructed object.
-        // Postcondition: other has no shared state.
+        /// \brief constructs a new promise object and transfers ownership of
+        ///          the shared state of other (if any) to the newly-
```
msimberg: Could you please check the indentation for multi-line blocks like this? It's a small thing but I think it would be nice to have it like so:
```
        /// \brief constructs a new promise object and transfers ownership of
        //         the shared state of other (if any) to the newly-
```
where `constructs` and `the` are aligned.

tapaswenipathak:
@msimberg do you suggest any more enhancements?

hpx/lcos/promise.hpp
```diff
@@ -167,97 +164,91 @@ namespace lcos {
...
-        //     object of R.
-        // Error conditions:
-        //   - promise_already_satisfied if its shared state already has a
-        //     stored value or exception.
-        //   - no_state if *this has no shared state.
+        /// \brief     atomically stores the value r in the shared state and makes
+        ///            that state ready (30.6.4).
+        /// \throws    future_error if its shared state already has a stored value or
+        ///            exception, or any exception thrown by the constructor
+				///            selected to copy an object of R.
```
msimberg: Stray tabs:
```suggestion
        ///            selected to copy an object of R.
```

hpx/lcos/promise.hpp
```diff
@@ -167,97 +164,91 @@ namespace lcos {
...
-        //   - promise_already_satisfied if its shared state already has a
-        //     stored value or exception.
-        //   - no_state if *this has no shared state.
+        /// \brief     atomically stores the exception pointer p in the shared
+        ///            state and makes that state ready (30.6.4).
+        /// \throws    future_error if its shared state already has a stored value
+        ///            or exception.
+        ///            promise_already_satisfied if its shared state already has a
+        ///            stored value or exception.
+        ///            no_state if *this has no shared state.
```
msimberg: This is the only part that still needs some work. The parts that were under "Error conditions" need better formatting (see [here](https://en.cppreference.com/w/cpp/thread/promise/set_value) under "Exceptions" for an example). `future_error` is the type of the exception, whereas `promise_already_satisfied` and `no_state` are error codes associated with the exception. So maybe change it to something like (ignore the ugly formatting):
```
/// \throws future_error if [... as before ...] exception. The error code
/// is set to promise_already_satisfied if its shared state...
/// The error code is set to no_state if...
```
The same goes for all the other "Error conditions" sections in this file. I think after that we should be able to merge this. Thanks for working on this!

hpx/lcos/promise.hpp
```diff
@@ -231,12 +231,11 @@ namespace lcos {
...
-        /// \throws    future_error if its shared state already has a stored value or
-        ///            exception, or any exception thrown by the constructor
-        ///            selected to copy an object of R.
-        ///            promise_already_satisfied if its shared state already has a
-        ///            stored value or exception.
-        ///            no_state if *this has no shared state.
+				///
+				/// \throws    future_error if [... as before ...] exception. The error code
+				///            is set to promise_already_satisfied if its shared state already
+				///            has a stored value or exception.
```
hkaiser: This does look like you used tabs instead of spaces...

hpx/lcos/promise.hpp
```diff
@@ -167,97 +164,90 @@ namespace lcos {
...
-        //   - promise_already_satisfied if its shared state already has a
-        //     stored value or exception.
-        //   - no_state if *this has no shared state.
+        /// \brief     atomically stores the value r in the shared state and makes
+        ///            that state ready (30.6.4).
+				///
+				/// \throws    future_error if [... as before ...] exception. The error code
+				///            is set to promise_already_satisfied if its shared state already
+				///            has a stored value or exception.
+        ///            The error code is set to no_state if *this has no shared state.
```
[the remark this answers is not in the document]
msimberg: Ignore this. You've already taken care of it.

msimberg:
@tapaswenipathak sorry, I had started a review, but forgot to finish it. Please have a look at my comments.

hpx/lcos/promise.hpp
msimberg: What I meant was to update them to look like this: https://github.com/STEllAR-GROUP/hpx/blob/8b6c014fe5aa2f6e507ac48065006bca20835e84/hpx/lcos/promise.hpp#L232-L238. I liked that change (except that it shouldn't say `[... as before ...]`).

tapaswenipathak: Cool, will do.

tapaswenipathak: Sure.

hpx/lcos/promise.hpp
```diff
@@ -59,223 +58,216 @@ namespace lcos {
...
-        {}
+        {
+        }
 
-        // Effects: constructs a promise object and a shared state. The
-        // constructor uses the allocator a to allocate the memory for the
-        // shared state.
+        /// \brief    constructs a promise object and a shared state. The
+        ///           constructor uses the allocator a to allocate the memory for the
+        ///           shared state.
```
hkaiser: IIRC, Doxygen ignores/skips `\brief` sections spanning more than one line. I think just leaving off the `\brief` directive should be fine, though.

msimberg:
@tapaswenipathak do you plan to work on this?

tapaswenipathak:
@msimberg I will implement the remaining things; sorry for the delay.

hpx/lcos/promise.hpp
tapaswenipathak: @msimberg I still don't get you, wdym?

hpx/lcos/promise.hpp
tapaswenipathak: No I think, you can find http://www.doxygen.nl/manual/docblocks.html.

tapaswenipathak:
folks: let me know if you have any more comments.

hpx/lcos/promise.hpp
msimberg: Sorry if I was unclear. Having both `\throws` and `\exception` doesn't make much sense. I'd make it one `\throws` section which explains the two different error codes that can be set.

Easiest might be to look at https://en.cppreference.com/w/cpp/thread/promise/set_value and use the documentation from there.

The same change should be made for all the other member functions. Does that make it clearer?

tapaswenipathak: I can see what you are saying. I can remove that, np.

tapaswenipathak: I have another tiny input: You can have a `\error` tag as well, should I add that or would you just prefer adding `\throws` and `\exception` wherever strictly required?

msimberg: As far as I can tell `\error` is not a built-in Doxygen command: http://www.doxygen.nl/manual/commands.html#cmdxrefitem. Do you have another source for this? I think we'll be okay without `\error`.

tapaswenipathak: Ok.

tapaswenipathak:
@msimberg: any issues in merging?

PR #3991, opened 2019-07-19
Fix dependencies and flags for for header tests

aurianer:


hkaiser:
@aurianer something is still wrong with the header tests. Some of those look for `#include <hpx/dataflow.hpp>`, a header file that does not exist.

msimberg:
The file does actually exist. I think the `add_hpx_header_tests` call for collectives just needs to have `NOHPX_INIT` and `NOLIBS` removed since it depends on the core HPX library.

PR #3996, opened 2019-07-21
Attempt to solve issue where -latomic does not support 128bit atomics

hkaiser:
Hopefully fixes https://github.com/laristra/flecsi-third-party/pull/88

### DIFF
--- a/hpx/runtime/threads/policies/lockfree_queue_backends.hpp
+++ b/hpx/runtime/threads/policies/lockfree_queue_backends.hpp
@@ -24,18 +24,13 @@ namespace hpx { namespace threads { namespace policies
 {
 
 struct lockfree_fifo;
-struct lockfree_lifo;
 
 ///////////////////////////////////////////////////////////////////////////////
 // FIFO
 template <typename T>
 struct lockfree_fifo_backend
 {
-#if defined(HPX_HAVE_CXX11_STD_ATOMIC_128BIT)
-    typedef boost::lockfree::deque<T> container_type;
-#else
     typedef boost::lockfree::queue<T> container_type;
-#endif
 
     typedef T value_type;
     typedef T& reference;
@@ -51,20 +46,12 @@ struct lockfree_fifo_backend
 
     bool push(const_reference val, bool /*other_end*/ = false)
     {
-#if defined(HPX_HAVE_CXX11_STD_ATOMIC_128BIT)
-        return queue_.push_left(val);
-#else
         return queue_.push(val);
-#endif
     }
 
     bool pop(reference val, bool steal = true)
     {
-#if defined(HPX_HAVE_CXX11_STD_ATOMIC_128BIT)
-        return queue_.pop_right(val);
-#else
         return queue_.pop(val);
-#endif
     }
 
     bool empty()
@@ -87,14 +74,13 @@ struct lockfree_fifo
 
 ///////////////////////////////////////////////////////////////////////////////
 // LIFO
+#if defined(HPX_HAVE_CXX11_STD_ATOMIC_128BIT)
+struct lockfree_lifo;
+
 template <typename T>
 struct lockfree_lifo_backend
 {
-#if defined(HPX_HAVE_CXX11_STD_ATOMIC_128BIT)
     typedef boost::lockfree::deque<T> container_type;
-#else
-    typedef boost::lockfree::queue<T> container_type;
-#endif
 
     typedef T value_type;
     typedef T& reference;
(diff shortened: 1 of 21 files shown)


hpx/runtime/threads/policies/lockfree_queue_backends.hpp
```diff
@@ -66,11 +85,17 @@ struct lockfree_fifo
...
 
+///////////////////////////////////////////////////////////////////////////////
 // LIFO
 template <typename T>
 struct lockfree_lifo_backend
 {
+#if defined(HPX_HAVE_CXX11_STD_ATOMIC_128BIT)
     typedef boost::lockfree::deque<T> container_type;
+#else
+    typedef boost::lockfree::queue<T> container_type;
```
msimberg: Shouldn't this be `boost::lockfree::stack`?

hpx/runtime/threads/policies/lockfree_queue_backends.hpp
```diff
@@ -85,14 +110,24 @@ struct lockfree_lifo_backend
...
     bool push(const_reference val, bool other_end = false)
     {
+#if defined(HPX_HAVE_CXX11_STD_ATOMIC_128BIT)
         if (other_end)
             return queue_.push_right(val);
         return queue_.push_left(val);
+#else
+        // without 128 atomics this turns into a fifo queue
+        return queue_.push(val);
+#endif
```
msimberg: This is just a reminder that this is subtly broken if items are not pushed to the other end when `other_end = true` (#3166 and #3172). I have nothing against this as long as the flecsi people are aware of this. Otherwise it might be better to disable the lifo queue altogether like the abp queues.

hpx/runtime/threads/policies/lockfree_queue_backends.hpp
hkaiser: Good point, will change.

hpx/runtime/threads/policies/lockfree_queue_backends.hpp
hkaiser: Yes, I know. However I think you're right. Disabling might be the best option as leaving it as it is right now does not provide any benefits - it just duplicates functionalities in some obscure way,

hpx/runtime/threads/policies/lockfree_queue_backends.hpp
```diff
@@ -21,11 +26,17 @@ namespace hpx { namespace threads { namespace policies
...
 
+///////////////////////////////////////////////////////////////////////////////
 // FIFO
 template <typename T>
 struct lockfree_fifo_backend
 {
+#if defined(HPX_HAVE_CXX11_STD_ATOMIC_128BIT)
     typedef boost::lockfree::deque<T> container_type;
+#else
+    typedef boost::lockfree::queue<T> container_type;
```
msimberg: By the way, this could actually always be `queue` instead of `deque`. I simply changed all of them to `deque` in #3172 because I thought it doesn't hurt to have them all use the same underlying datastructure. The other queues actually need `deque`.

hkaiser:
@msimberg I have addressed your comments, please have a look.

tests/unit/resource/shutdown_suspended_pus.cpp
```diff
@@ -75,7 +75,7 @@ int main(int argc, char* argv[])
                 hpx::resource::scheduling_policy::local_priority_fifo,
                 hpx::resource::scheduling_policy::local_priority_lifo,
```
msimberg: I think `local_priority_lifo` needs to be `#ifdef`ed away like the ABP scheduler, no? Same in the other tests.

hkaiser: Uhh, I missed those. Thanks!

hkaiser:
@msimberg ok now?

tests/unit/resource/suspend_runtime.cpp
```diff
@@ -68,7 +68,7 @@ int main(int argc, char* argv[])
             hpx::resource::scheduling_policy::local_priority_fifo,
             hpx::resource::scheduling_policy::local_priority_lifo,
```
msimberg: Sorry @hkaiser, one more...

msimberg:
@hkaiser looks good except for that one remaining `lifo` that needs to be ifdefed out. Thanks! Of course, our CI isn't going to test any of these changes... Do you want me to give it a run with 128 bit atomics off? Or can you get the flecsi people to try it out?

tests/unit/resource/suspend_runtime.cpp
hkaiser: Grrr...

hkaiser:
@msimberg we asked the flecsi people to give it a go.

tests/unit/resource/suspend_runtime.cpp
hkaiser: Fixed now.

hpx/runtime/threads/policies/lockfree_queue_backends.hpp
hkaiser: As is turns out, this change or the change to `lockfree::stack` above has caused a performance regression. I will prepare a PR reverting this part of the patch.

hkaiser: See #4111 for the patch that reverts this particular change.

PR #3997, opened 2019-07-22
Export handle_assert function

msimberg:
Fixes GCC release builds. The function was not declared exported inside the assertion module, only for consumers of the module.

[Before](http://rostam.cct.lsu.edu/builders/hpx_gcc_8_boost_1_70_centos_x86_64_release/builds/85) and [after](http://rostam.cct.lsu.edu/builders/hpx_gcc_8_boost_1_70_centos_x86_64_release/builds/86).

### DIFF
--- a/libs/assertion/include/hpx/assertion/evaluate_assert.hpp
+++ b/libs/assertion/include/hpx/assertion/evaluate_assert.hpp
@@ -8,7 +8,6 @@
 
 #include <hpx/config.hpp>
 
-#if defined(HPX_DEBUG)
 #include <hpx/assertion/source_location.hpp>
 
 #include <string>
@@ -32,5 +31,4 @@ namespace hpx { namespace assertion { namespace detail
     /// \endcond
 }}}
 
-#endif
 #endif
--- a/libs/assertion/include/hpx/assertion/force_linking.hpp
+++ b/libs/assertion/include/hpx/assertion/force_linking.hpp
@@ -16,10 +16,8 @@ namespace hpx { namespace assertion
 {
     struct force_linking_helper
     {
-#if defined(HPX_DEBUG)
         void (*handle_assert)(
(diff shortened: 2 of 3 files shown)


msimberg:
Yeah, you're right. In this case `handle_assert` is only being forward-declared elsewhere, but not declared before the definition I changed (no header file). The assertion module would probably need some rearranging, but this is the minimal change to fix this.

msimberg:
@hkaiser mind if I merge this?

hkaiser:
> mind if I merge this?

@msimberg frankly, I still don't understand why this is necessary. The function is declared as `HPX_EXPORT` [here](https://github.com/STEllAR-GROUP/hpx/blob/master/libs/assertion/include/hpx/assertion/evaluate_assert.hpp#L20) and this header file is (recursively) included by the file  you propose to change. Is this a GCC bug?

hkaiser:
Ahh, now I see it. The declaration is guarded by an `#if defined(HPX_DEBUG)`. Should we guard the definition as well?

msimberg:
@hkaiser mind if I merge this?

> Ahh, now I see it. The declaration is guarded by an `#if defined(HPX_DEBUG)`. Should we guard the definition as well?

Very good point. Thanks for pointing that out! Yes, that or remove the guard. I'd prefer removing the guard because then the only thing that's guarded is the `HPX_ASSERT` macro and the rest stays the same. Less confusing or all the same?

hkaiser:
> Yes, that or remove the guard. I'd prefer removing the guard because then the only thing that's guarded is the HPX_ASSERT macro and the rest stays the same (in debug and release). 

Fine with me. Thanks!

msimberg:
@hkaiser how about this? (build here: http://rostam.cct.lsu.edu/builders/hpx_gcc_8_boost_1_70_centos_x86_64_release/builds/96)

PR #3998, opened 2019-07-23
Add process thread binding option

msimberg:
Adds a new thread binding option `--hpx:bind=process` which looks at the CPU mask of the process to choose which PUs to use for worker threads. If threads is `"cores"` (default) or `"all"` it overrides the number of threads to the number of PUs in the CPU mask, and if an explicit number has been set it uses that.

This is useful for making thread bindings consistent with the behaviour of other applications in batch environments. If this works well it could even replace the batch environment support we have (assuming all of them use mpi).

### DIFF
--- a/docs/sphinx/manual/launching_and_configuring_hpx_applications.rst
+++ b/docs/sphinx/manual/launching_and_configuring_hpx_applications.rst
@@ -1823,7 +1823,7 @@ Here is a full grammar describing the possible format of mappings:
 
 .. productionlist::
    mappings: `distribution` | `mapping` (";" `mapping`)*
-   distribution: "compact" | "scatter" | "balanced" | "numa-balanced"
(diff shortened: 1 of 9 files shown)


msimberg:
Updated based on feedback from @rasolca at CSCS. Instead of `--hpx:bind=process` there's now a flag `--hpx:use-process-mask` which applies to *all* the predefined distributions (it used to do a compact distribution on the available PUs). This means that one can enable hyperthreading for the process mask but have HPX use a smaller number of threads e.g. with the balanced distribution. @rasolca might be able to elaborate on use cases for this. `--hpx:use-process-mask` also implies `--hpx:ignore-batch-env`.

msimberg:
@hkaiser please have a look when you have the time. This should be ready to go.

PR #4000, opened 2019-07-26
Module errors

aurianer:


hpx/custom_exception_info.hpp
```diff
@@ -378,9 +100,52 @@ namespace hpx
...
+            }
+            catch (...)
+            {
+                return std::current_exception();
+            }
+
+            // need this return to silence a warning with icc
+            HPX_ASSERT(false);
+            return std::exception_ptr();
+        }
```
hkaiser: Is it necessary to _define_ this in a header file?

libs/errors/include/hpx/errors/exception_list.hpp
```diff
@@ -35,7 +34,10 @@ namespace hpx
     class HPX_EXCEPTION_EXPORT exception_list : public hpx::exception
     {
     private:
-        typedef hpx::lcos::local::spinlock mutex_type;
+        // TODO: Does this need to be hpx::lcos::local::spinlock?
+        // typedef hpx::lcos::local::spinlock mutex_type;
+        // TODO: Add correct initialization of boost spinlock.
+        typedef boost::detail::spinlock mutex_type;
```
hkaiser: What's the rationale of this change?

hpx/custom_exception_info.hpp
msimberg: Well, It *is* a templated function, but we might be able to get by with explicit instantiations. Not sure.

libs/errors/include/hpx/errors/exception_list.hpp
msimberg: Not having to depend on `lcos` in this module (with the assumption that we don't actually need HPX thread yielding in the spinlock). But saying that I realize that `exception_list` might live more happily in the `algorithms` module in which case we should be able to keep it as `hpx::lcos::local::spinlock`. What do you think?

hpx/custom_exception_info.hpp
hkaiser: We do explicitly instantiate this function (see: https://github.com/STEllAR-GROUP/hpx/blob/master/src/exception.cpp#L412-L461).

libs/errors/include/hpx/errors/exception_list.hpp
hkaiser: Is it used only by the algorithms? I'm not sure. Also, assuming we don't need HPX yielding sounds like a bit of a stretch... However, I see your point...

msimberg: Hrm, almost only by algorithms (and executors)... two other files include it.

I know no HPX thread yielding does sound a bit like a stretch, but I don't see it as a performance problem since we're dealing with exceptions anyway in which case something has already gone wrong. That leaves the question whether this can deadlock. I can't prove this won't happen but I'm hoping our tests are extensive enough to tell us.

Having `lcos` as a dependency would make modularization very difficult so I'd rather try to make it work with the boost spinlock before resorting to that.

hpx/custom_exception_info.hpp
msimberg: Sorry. Then this can definitely just be declared. Thanks!

libs/errors/include/hpx/errors/exception_list.hpp
hkaiser: Ok, fair enough. BTW, there is also `hpx::util::spinlock` that relies on Boost's version but additionally does lock-tracking for HPX. So using this instead might actually be the better option.

msimberg: Agreed, but `hpx::util::spinlock` depends on exceptions making that problematic as well. This should one of, if not the, only module that would have to do this (depend directly on `boost::spinlock`). `hpx::util::spinlock` is coming up right after this module.

hpx/custom_exception_info.hpp
msimberg: Looking a bit more carefully I see that it's `throw_exception` that's explicitly instantiated, not `construct_exception`. We would probably manage with the same explicit instantiations for `construct_exception` as well. Would you prefer that?

hkaiser: Yes, you are right. However, `construct_exception` is instantiated by `get_exception`, which is implicitly instantiated by `throw_exception` so there is no need to explicitly instantiate that as well.

msimberg: Sorry for the confusion (I'm slowly remembering the changes I made). That part has changed on this branch. `throw_exception` does not call `construct_exception` anymore. We can still instantiate `construct_exception` explicitly though.

hkaiser: Right, I should have looked more carefully. Let's go with explicitly instantiating the functions, then.

msimberg:
The failure on CircleCI looks unrelated.

PR #4003, opened 2019-07-29
Refactoring to remove (internal) dependencies

msimberg:
Refactoring to make it easier to create threadmanager, scheduler, resource partitioner modules:
- Moved network-related callbacks out of `scheduler_base` depended on parcelport and AGAS
- Separated threadmanager performance counters out of the class and into separate file
- Move futurized thread pool suspend/resume functions out of the thread pool classes into free functions
- Move around command line handling functionality to remove unnecessary dependencies
- Pass affinity data and callbacks into threadmanager/schedulers/thread pools instead of getting them through the global resource partitioner object


examples/resource_partitioner/async_customization.cpp
```diff
@@ -502,27 +502,33 @@ int main(int argc, char** argv)
...
-          {
+    rp.create_thread_pool("custom",
+        [](hpx::threads::policies::callback_notifier& notifier,
+            std::size_t num_threads, std::size_t thread_offset,
+            std::size_t pool_index, std::string const& pool_name,
+            hpx::threads::detail::network_background_callback_type const&
+                network_background_callback,
+            hpx::threads::policies::detail::affinity_data const& affinity_data,
+            std::size_t max_background_threads, std::size_t max_idle_loop_count,
+            std::size_t max_busy_loop_count)
```
hkaiser: May I suggest to combine all of the `size_t`'s in a special struct to reduce the number of arguments?

msimberg: Yep, will do, it's getting a bit out of hand.

hkaiser:
This also fixes #3082, I think.

msimberg:
> This also fixes #3082, I think.

Yeah, I think you're right. It's not what I imagined when I first opened that issue, but I think not having to explicitly forward all unchanged arguments removes the worst part of customizing thread pools. I don't know if @biddisco had something else in mind though.

I think this is ready for review from my side, modulo any test/build failures I haven't noticed yet. There's more rearranging to do at least for `topology`, but that's coming in a separate PR with a topology module.

msimberg:
All right, that should be the last problem fixed for this PR.

hpx/runtime/threads/detail/scheduling_loop.hpp
```diff
@@ -559,23 +547,18 @@ namespace hpx { namespace threads { namespace detail
         bool may_exit = false;
 
 #if defined(HPX_HAVE_NETWORKING)
-        bool networking_is_enabled = hpx::is_networking_enabled();
```
hkaiser: What is the rationale of enabling background threads if networking is disabled?

msimberg: It's not enabled, the logic has just moved up a bit (to avoid the call to `hpx::is_networking_enabled` and thus a dependency on the runtime configuration). I instead set the number of background threads to 0 in the thread manager if networking is disabled with the same effect.

hpx/util/command_line_handling.hpp
```diff
@@ -24,6 +24,21 @@
...
+        util::command_line_handling const& cfg, std::size_t num_threads);
+
+    HPX_EXPORT std::string get_affinity_domain(util::command_line_handling const& cfg);
+
+    HPX_EXPORT std::size_t get_affinity_description(
+        util::command_line_handling const& cfg, std::string& affinity_desc);
+
+    HPX_EXPORT std::size_t get_pu_offset(util::command_line_handling const& cfg);
+
+    HPX_EXPORT std::size_t get_pu_step(util::command_line_handling const& cfg);
```
hkaiser: Do we have to export these functions?

src/runtime/resource/detail/detail_partitioner.cpp
```diff
@@ -916,7 +918,15 @@ namespace hpx { namespace resource { namespace detail
         cfg_.parse_result_ = cfg_.call(desc_cmdline, argc, argv);
 
         // set all parameters related to affinity data
-        pus_needed_ = affinity_data_.init(cfg_);
+        std::string affinity_description;
+        get_affinity_description(cfg_, affinity_description);
+
+        pus_needed_ = affinity_data_.init(cfg_.num_threads_,
+            hpx::util::safe_lexical_cast<std::size_t>(
+                get_config_entry("hpx.cores", 0), 0),
```
hkaiser: Shouldn't this query the number of cores in a more general fashion (i.e. taking into account command line options, batch environment, etc.)?

msimberg: If I follow the logic correctly this does in fact include those things that you mention. The command line option (`--hpx:cores`) and batch environments are all handled together and the config entry `hpx.config` is updated to contain the value resulting from those. The behaviour is also unchanged from before.

hpx/util/command_line_handling.hpp
msimberg: No, I don't think so. Will remove. Thanks!

src/runtime/resource/detail/detail_partitioner.cpp
hkaiser: Ok, fair enough.

PR #4004, opened 2019-07-29
Shorten CTest timeout on CircleCI

msimberg:
Explicitly sets the timeout to 60 seconds (I *think* it's enough, let's see...) on CircleCI instead of waiting for the 10 minute timeout that CircleCI has by default.


msimberg:
It turns out that `tests.unit.resource.suspend_runtime` was running for 8-9 minutes in a typical run on CircleCI (when it would take around a minute on Piz Daint). Not surprising now that I know about it but, but that's most likely the reason it's been timing out so often (and not that it actually hangs). I've changed it to run for a certain amount of time instead of a certain number of iterations.

PR #4007, opened 2019-07-30
Fixing migration for components that have actions returning a client

hkaiser:
- flyby: properly unmark as migrated component that do not need to be migrated as target locality is the same as current locality
- flyby: adding missing functions to `gather_here`/`gather_there`

@parsa this should fix the hang in your application, please re-check

### DIFF
--- a/hpx/lcos/gather.hpp
+++ b/hpx/lcos/gather.hpp
@@ -335,6 +335,25 @@ namespace hpx { namespace lcos
 
         ///////////////////////////////////////////////////////////////////////
         template <typename T>
+        hpx::future<std::vector<typename std::decay<T>::type> >
+        gather_data_direct(hpx::future<hpx::id_type> f, T&& result,
+            std::size_t site)
+        {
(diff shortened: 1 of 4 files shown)


msimberg:
I have no real review comments on this. Just curious, does this happen to fix any of the failures we see on the `migrate_component` test, or does this fix something different?

hkaiser:
> I have no real review comments on this. Just curious, does this happen to fix any of the failures we see on the migrate_component test, or does this fix something different?

I don't think this touches on the migration failures we're seeing for the polymorphic migration test. This PR fixes two separate issues (and a flyby for gather): 

- components with actions that return client instances were unpinning the component instance too early, causing for objects to be potentially migrated while still being accessed
- migration of a component instance to the same locality it already lives on (which should be a no-op) didn't unpin the instance at all causing subsequent deadlocks
- `gather` was simply broken (it didn't even compile) for certain use cases, this change fixes that

PR #4008, opened 2019-07-31
Extend cache aligned data to handle tuple-like data

biddisco:
This allows the user to create a structure such as
```
  cache_line_multidata<
    int,
    int
  > rollover_counters_;
```
and access the elements individually. The elements will occupy the same cache-line and are padded to the next cache-line boundary.

### DIFF
--- a/hpx/util/cache_aligned_data.hpp
+++ b/hpx/util/cache_aligned_data.hpp
@@ -63,63 +63,15 @@ namespace hpx { namespace util
     // The tuple is cache aligned and fully occupies 1 or more cache lines
     // use get<0>, get<1>, ... get<N> to access the internal elements
 #if defined(HPX_HAVE_CXX11_ALIGNAS) && defined(HPX_HAVE_CXX17_ALIGNED_NEW) &&  \
-    defined(HPX_HAVE_CXX17_FOLD_EXPRESSIONS) && !defined(__NVCC__)
+    !defined(__NVCC__)
     template <typename ... Data>
     struct alignas(threads::get_cache_line_size()) cache_line_multidata
     {
-        template <typename ... T>
-        cache_line_multidata(T...ts) : data_(ts...) {};
-
-        template <std::size_t I>
-        auto &get() {
-            return std::get<I>(data_);
-        }
-
-        template <std::size_t I>
-        auto const &get() const {
-            return std::get<I>(data_);
-        }
-
-        // the storage of the data in tuple form
-        std::tuple<Data...> data_;
-
-        // pad to multiple of cache line size bytes
-        // use zero padding if multiple of line size already
-        static constexpr unsigned int padding1 =
-            (0 + ... + sizeof(Data)) % threads::get_cache_line_size();
-        static constexpr unsigned int padding2 =
-            (padding1==0) ? 0 : threads::get_cache_line_size() - padding1;
-        char cacheline_pad[padding2];
-    };
-
-#elif defined(HPX_HAVE_CXX14_STD_INTEGER_SEQUENCE)
-    namespace detail
-    {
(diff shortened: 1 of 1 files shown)


hpx/util/cache_aligned_data.hpp
```diff
@@ -55,6 +55,84 @@ namespace hpx { namespace util
...
+        }
+
+        template <std::size_t I>
+        auto const &get() const {
+            return std::get<I>(data_);
+        }
+
+        // pad to cache line size bytes
+        std::tuple<Data...> data_;
+        char cacheline_pad[(0 + ... + sizeof(Data)) % threads::get_cache_line_size()];
```
msimberg: I think `sizeof...(Data)` should work here. Then you should be able to get rid of the other implementation.

biddisco: `sizeof()...` on its own will give you the pack size, but not the bytes used by the elements, that's the reason for the long-winded template stuff. the fold operator does it very nicely for c++17

NB. force pushed because I had
`(0 + ... + sizeof(Data)) % threads::get_cache_line_size()`
instead of 
`threads::get_cache_line_size() - ((0 + ... + sizeof(Data)) % threads::get_cache_line_size())`
in both cases

msimberg: You're right, sorry for the noise.

hkaiser: @biddisco: wouldn't be `sizeof(std::tuple<Data...>)` be sufficient here?

hkaiser:
@biddisco I'd go even further to ask whether we would need your facility at all. One could always write `cache_aligned_data<std::tuple<int, int>>` for the same effect... 

I do appreciate that your facility might provide a minor convenience over explicitly spelling out the `std::tuple`, in the end I however would prefer the explicit option as this also opens the door to using `hpx::util::tuple` or anything similar.

biddisco:
doh! `sizeof(tuple<>)` would be much simpler. I didn't want `cache_aligned_data<std::tuple<>>` just because it seemed a bit messy at the time. I could be persuaded to just use it though and bin the PR.

msimberg:
@biddisco so how badly do you want this in? Your padding calculations made me realize that the existing padding that I changed to handle sizes bigger than a cache line adds extra unnecessary padding. Yours blows up when the tuple is bigger than a cache line.

If we close this I'll open a separate PR to fix the calculation. If we keep this open those could both be fixed here.

hkaiser:
@msimberg I think we should ditch this PR and add the fixes you mentioned to a separate one.

biddisco:
Sorry. I had my stuff in a separate class and shoved it into the cache_line_data at the last minute because I thought I could sneak it in more easily. I did have a version that handled tuple size >cache line properly, but I guess I broke it when I copied the code into here. I'll close this PR for now.

biddisco:
The version I have here locally in my other branch has the correct padding for size>cache line size. I messed up when I pasted it into the PR branch. It is fixed for the original struct and the tuple version.
Resubmit? or are you all opposed to the multidata tuples version?

msimberg:
I'm only slightly in favor of `cache_aligned_data<tuple<...>>`. In terms of access your version is only slightly shorter:
```
padded_tuple_data.get<i>();
```
over
```
hpx::util::get<i>(padded_tuple_data.data_);
```
but I think the latter is more familiar and I don't have to wonder if `.get<N>` does the same as `hpx::util::get<N>`.

As I said, only very slightly in favor of what's there though.

hkaiser:
> The version I have here locally in my other branch has the correct padding for size>cache line size. I messed up when I pasted it into the PR branch. 

@biddisco please create a new PR for these fixes.

PR #4009, opened 2019-07-31
Check circular dependencies in a circle ci step

aurianer:
Check for circular `#include` dependencies for the modules, the `cpp-dependencies` being installed in the Dockerfile from STE||AR-GROUP/docker_build_env

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,6 +107,7 @@ jobs:
           name: Check circular dependencies with cpp-dependencies
           command: |
               cpp-dependencies --dir /hpx/source/libs --graph-cycles /hpx/tools/cpp-dependencies/circular_deps.dot
+              dot /hpx/tools/cpp-dependencies/circular_deps.dot -Tsvg -o /hpx/tools/cpp-dependencies/circular_deps.svg
               if [[ $(wc -l /hpx/tools/cpp-dependencies/circular_deps.dot | awk '{print $1}') -gt 2 ]]; then exit 1; fi
 
   configure:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,6 +97,18 @@ jobs:
             - ./source
             - ./conv.xsl
 
+  # Check circular dependencies with cpp-dependencies tool
+  check_circular_deps:
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: /hpx
+      - run:
+          name: Check circular dependencies with cpp-dependencies
+          command: |
+              cpp-dependencies --dir /hpx/source/libs --graph-cycles /hpx/tools/cpp-dependencies/circular_deps.dot
+              if [[ $(wc -l /hpx/tools/cpp-dependencies/circular_deps.dot | awk '{print $1}') -gt 2 ]]; then exit 1; fi
+
   configure:
     <<: *defaults
     steps:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -181,6 +181,20 @@ jobs:
...
+      - attach_workspace:
+          at: /hpx
+      - run:
+          name: Check circular dependencies with cpp-dependencies
+          command: |
+              ./cpp-dependencies --dir /hpx/source/libs /hpx/tools/cpp-dependencies/circular_deps.dot
+              if [[ $(wc -l /hpx/tools/cpp-dependencies/circular_deps.dot | awk '{print $1}') -gt 2 ]]; then exit 1; fi
+      - store_artifacts:
+          path: /hpx/tools/cpp-dependencies/circular_deps.dot
+
```
msimberg: The step needs to be added to the list of steps at the bottom of the config as well. I think it should be enough if this depends on the checkout step, and install depends on this step.

.circleci/config.yml
```diff
@@ -96,6 +96,20 @@ jobs:
...
+  # Check circular dependencies with cpp-dependencies tool
+  check_circular_deps:
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: /hpx
+      - run:
+          name: Check circular dependencies with cpp-dependencies
+          command: |
+              ./cpp-dependencies --dir /hpx/source/libs /hpx/tools/cpp-dependencies/circular_deps.dot
```
hkaiser: This probably misses a `--graph-cyclic` command line option.

.circleci/config.yml
```diff
@@ -96,6 +96,20 @@ jobs:
...
+              if [[ $(wc -l /hpx/tools/cpp-dependencies/circular_deps.dot | awk '{print $1}') -gt 2 ]]; then exit 1; fi
+      - store_artifacts:
+          path: /hpx/tools/cpp-dependencies/circular_deps.dot
```
hkaiser: Should we store a rendered dot file instead?

aurianer: I kept the dot file to check if there is some graph cycles by file length and added a svg generation to visualize the graph :)

.circleci/config.yml
```diff
@@ -96,6 +96,23 @@ jobs:
...
+      - run:
+          name: Check circular dependencies with cpp-dependencies
+          command: |
+              cpp-dependencies --dir /hpx/source/libs --graph-cycles /hpx/tools/cpp-dependencies/circular_deps.dot
+              dot /hpx/tools/cpp-dependencies/circular_deps.dot -Tsvg -o /hpx/tools/cpp-dependencies/circular_deps.svg
+              if [[ $(wc -l /hpx/tools/cpp-dependencies/circular_deps.dot | awk '{print $1}') -gt 2 ]]; then exit 1; fi
+      - store_artifacts:
+          path:
+            - /hpx/tools/cpp-dependencies/circular_deps.dot
+            - /hpx/tools/cpp-dependencies/circular_deps.svg
```
msimberg: I think these have to go in two separate `- store_artifacts` entries (see https://circleci.com/gh/STEllAR-GROUP/hpx/116132 and https://circleci.com/docs/2.0/configuration-reference/#store_artifacts).

PR #4012, opened 2019-08-05
Fix CUDA compilation

msimberg:
Fixes CUDA compilation.

1. `function_base` constructor was `constexpr` which `nvcc` didn't like. This was causing the `there was an error in verifying the lgenfe output!` errors. I couldn't figure out if this is a known bug or not. @K-ballo was the `constexpr` critical for performance?
2. `construct`/`bulk_construct`/`bulk_destroy` didn't instantiate `detail::launch` in device code which meant that kernel launches would fail.

### DIFF
--- a/hpx/compute/cuda/allocator.hpp
+++ b/hpx/compute/cuda/allocator.hpp
@@ -172,7 +172,7 @@ namespace hpx { namespace compute { namespace cuda
         template <typename ... Args>
         HPX_HOST_DEVICE void bulk_construct(pointer p, std::size_t count, Args &&... args)
         {
-#if defined(HPX_COMPUTE_DEVICE_CODE)
+#if defined(HPX_COMPUTE_CODE)
             int threads_per_block = (hpx::util::min)(1024, int(count));
             int num_blocks =
                 int((count + threads_per_block - 1) / threads_per_block);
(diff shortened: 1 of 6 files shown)


msimberg:
Rostam build here: http://rostam.cct.lsu.edu/builders/hpx_cuda_10_1_gcc_8_boost_1_70_centos_x86_64_release/builds/106. I've enabled CUDA builds on pycicle again as well.

K-ballo:
>     1. @K-ballo was the `constexpr` critical for performance?

The `constexpr` is there for correctness, to guarantee constant initialization and thus avoid static initialization order issues. That said, we have a workaround for non-conforming compilers in https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/util/detail/empty_function.hpp#L27

CUDA might need to enable the workaround as well.

msimberg:
> > ```
> > 1. @K-ballo was the `constexpr` critical for performance?
> > ```
> 
> The `constexpr` is there for correctness, to guarantee constant initialization and thus avoid static initialization order issues. That said, we have a workaround for non-conforming compilers in https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/util/detail/empty_function.hpp#L27
> 
> CUDA might need to enable the workaround as well.

The workaround seems to work nicely for CUDA, thanks!

hkaiser:
@msimberg could you please rebase on master to remove the unrelated changes?

msimberg:
@hkaiser which unrelated changes are you talking about? This is already rebased on latest master.

hkaiser:
> which unrelated changes are you talking about? This is already rebased on latest master.

Hmmm, this PR has all the recent changes to master, much more than just your CUDA fixes. Not sure what happened...

PR #4017, opened 2019-08-07
Require C++14

### LINKED ISSUE #3944: Require C++14
Officially all our supported compilers have full C++14 support. The open question has always been `nvcc`, particularly on Windows. Could those who use Windows and/or `nvcc` let me know if there have been any corner cases where you've been forced to go down to C++11 because C++14 wasn't fully supported.

HPX also doesn't currently compile in C++11 mode (https://cdash.cscs.ch/viewBuildError.php?buildid=56536). I take the blame for that because the CI wasn't running properly. Nevertheless, I think it's time to require C++14 because it's clearly something people already want to use (see the errors above, and @biddisco's eagerness to use newer features in the guided executor). We'd also be able to drop a big bunch of feature checks from the CMake configuration.

Related: #3696.

msimberg:
Fixes #3944.

I've been slighly overeager in removing checks with the assumption that if we require C++14 then we don't need to check if C++14 features are available. This has the benefit of speeding up CMake configuration, but I realize that this is naive and that I'll most likely need to add back some of the checks. The goal is we would only add them back in very special cases where we know that the compiler doesn't fully support C++14 but for some reason we still want to make a special case for that.


hpx/exception_info.hpp
```diff
@@ -227,15 +223,13 @@ namespace hpx
     ///////////////////////////////////////////////////////////////////////////
     template <typename E, typename F>
     auto invoke_with_exception_info(E const& e, F&& f)
-      -> decltype(std::forward<F>(f)(std::declval<exception_info const*>()))
```
K-ballo: These should be `decltype(auto)` to avoid decay.

hpx/util/spinlock.hpp
```diff
@@ -24,17 +24,11 @@ struct spinlock
...
     boost::detail::spinlock m = BOOST_DETAIL_SPINLOCK_INIT;
-#else
-    boost::detail::spinlock m;
-#endif
 
   public:
     spinlock(char const* /*desc*/ = nullptr)
-#if !defined(HPX_HAVE_CXX11_NSDMI)
       : m(BOOST_DETAIL_SPINLOCK_INIT)
-#endif
```
K-ballo: This should be removed, the condition is negated.

cmake/HPX_DetectCppDialect.cmake
```diff
@@ -6,77 +6,46 @@
...
-            check_cxx_compiler_flag(-std=c++1y HPX_WITH_CXX1Y)
-          endif()
-
-          if(HPX_WITH_CXX1Y)
-            set(CXX_FLAG -std=c++1y PARENT_SCOPE)
-            hpx_info("C++ mode used: C++1y")
-          else()
-            # ... otherwise try -std=c++11
-            check_cxx_compiler_flag(-std=c++11 HPX_WITH_CXX11)
-            if(HPX_WITH_CXX11)
```
hkaiser: I think we should warn if somebody uses `HPX_WITH_CXX11` (or similar), for instance in build scripts, so that people are made aware of this (breaking) change.

cmake/HPX_PerformCxxFeatureTests.cmake
```diff
@@ -10,187 +10,6 @@
...
-    REQUIRED "HPX needs support for C++11 alias templates")
-
-  hpx_check_for_cxx11_alignas(
-    DEFINITIONS HPX_HAVE_CXX11_ALIGNAS)
-
-  hpx_check_for_cxx11_auto(
-    REQUIRED "HPX needs support for C++11 auto")
-
-  hpx_check_for_cxx11_constexpr(
-    DEFINITIONS HPX_HAVE_CXX11_CONSTEXPR)
```
hkaiser: What about `constexpr` support in NVCC? Does it support this sufficiently by now?

msimberg: CUDA + gcc on linux is fine. CUDA clang and MSVC are another question, but constexpr is ifdefed out for those as before. I wonder if those builds work anyway...

cmake/HPX_DetectCppDialect.cmake
msimberg: Yep, that's a good point. We may have to hold off with this PR and just give a warning in the next release... and then a hard error after that.

msimberg: Added a hard error for `HPX_WITH_CXX11` and co. We should definitely wait with this PR until after the release and add a warning for this release.

I'd argue that we should actually deprecate the `HPX_WITH_CXX` options and set `CXX_STANDARD` instead (or have the user set that).

hkaiser: Ok, I'd be fine with this change. Those would have to be phased out in the usual way, however. People have build scripts they use every day that possibly have the `HPX_WITH_CXX` switches, removing support for those needs to be communicated prominently.

msimberg: Absolutely. It's a separate issue though and I'll have a look at it in a separate PR.

CMakeLists.txt
```diff
@@ -409,9 +409,6 @@ endif()
 # Native TLS configuration
 ################################################################################
 set(HPX_WITH_NATIVE_TLS_DEFAULT ON)
-if(APPLE)
-  set(HPX_WITH_NATIVE_TLS_DEFAULT ${HPX_WITH_CXX11_THREAD_LOCAL})
-endif()
```
msimberg: This means that mac users need Xcode >= 8, but I think that's a reasonable requirement as it was released in 2016. One less thing to worry about.

msimberg:
I'm starting to finalize this. I've made some additional replacements for keywords that used to be hidden behind macros: `thread_local`, `constexpr` and the move/forward lambda captures. I don't know yet if all of these changes will go through with `nvcc` but we'll see. Is there someone who could try `nvcc` on Windows? Or even better, do we have a possibility to set up a test machine for it? I assume it's broken at the moment as it's not tested.

msimberg:
This should now be ready to go (the one pycicle failure is due to a pycicle configuration problem).

msimberg:
`inspect` is complaining about the SPDX tag that was removed from the license file in #4292, that can be ignored.

msimberg:
I'd like to move on with this before it gets outdated again. Reviews would be appreciated.

libs/algorithms/include/hpx/parallel/algorithms/detail/distance.hpp
```diff
@@ -14,9 +14,8 @@
 namespace hpx { namespace parallel { inline namespace v1 { namespace detail {
     // provide implementation of std::distance supporting iterators/sentinels
     template <typename InIterB, typename InIterE>
-    HPX_CXX14_CONSTEXPR inline
-        typename std::iterator_traits<InIterB>::difference_type
-        distance(InIterB first, InIterE last, std::input_iterator_tag)
+    typename std::iterator_traits<InIterB>::difference_type distance(
```
K-ballo: dropped `constexpr inline`, intentional?

libs/execution/include/hpx/parallel/execution_policy.hpp
```diff
@@ -33,7 +33,7 @@
 namespace hpx { namespace parallel { namespace execution {
     ///////////////////////////////////////////////////////////////////////////
     /// Default sequential execution policy object.
-    static task_policy_tag HPX_CONSTEXPR_OR_CONST task;
+    static task_policy_tag constexpr task;
```
K-ballo: odd placing (east constexpr?)

libs/resource_partitioner/examples/CMakeLists.txt
```diff
@@ -25,19 +26,14 @@ if(HPX_WITH_MAX_CPU_COUNT AND HPX_WITH_SHARED_PRIORITY_SCHEDULER)
     oversubscribing_resource_partitioner
     simple_resource_partitioner
   )
+endif()
 
-  if (HPX_WITH_CXX14_RETURN_TYPE_DEDUCTION)
```
K-ballo: Should this no longer be guarded by the above `if()`?

tests/performance/local/native_tls_overhead.cpp
```diff
@@ -20,21 +20,17 @@
 #include <thread>
 #include <vector>
 
-#if defined(__has_feature)
```
K-ballo: bad lose `if`
also when `__has_feature` is not defined then `__has_feature(cxx_thread_local)` is ill-formed as it evaluates to `0(0)`

tests/performance/local/transform_reduce_binary_scaling.cpp
```diff
@@ -21,7 +21,7 @@
 struct plus
 {
     template <typename T1, typename T2>
-    auto operator()(T1 && t1, T2 && t2) const -> decltype(t1 + t2)
```
K-ballo: this drops SFINAE

tests/unit/util/pack_traversal.cpp
```diff
@@ -752,7 +752,7 @@ static void test_strategic_tuple_like_traverse()
 struct duplicate_mapper
 {
     template <typename T>
-    auto operator()(T arg) -> decltype(hpx::util::spread_this(arg, arg))
```
K-ballo: this drops SFINAE

docs/CMakeLists.txt
```diff
@@ -80,7 +80,6 @@ set(doxygen_definition_list
     "HPX_ALWAYS_EXPORT="
     "extern="
     "HPX_NOEXCEPT="
-    "HPX_CONSTEXPR="
```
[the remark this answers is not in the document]
msimberg: Good catch. They are actually already gone from the code, but apparently got left behind here.

libs/algorithms/include/hpx/parallel/algorithms/detail/distance.hpp
msimberg: Nope, very good catch!

libs/resource_partitioner/examples/CMakeLists.txt
msimberg: No, with the rationale that if we require C++14 support we don't check for it unnecessarily. I'd only add checks for C++11 and 14 features that are not supported by compilers that we really care about (e.g. atomics need some special care).

tests/performance/local/native_tls_overhead.cpp
msimberg: Good catch. This whole block should go. I now assume `thread_local` is a keyword, and if not compilation will fail.

msimberg: Deleted this whole block.

msimberg:
Thanks @K-ballo for the review. I think I've fixed the problems you pointed out.

libs/resource_partitioner/examples/CMakeLists.txt
K-ballo: Further above, sorry. These commands were previously guarded by `if(HPX_WITH_MAX_CPU_COUNT AND HPX_WITH_SHARED_PRIORITY_SCHEDULER)`.

msimberg: Grr, yep. Fixed now.

cmake/HPX_AddConfigTest.cmake
```diff
@@ -424,285 +242,6 @@ function(hpx_check_for_cxx11_std_atomic_128bit)
...
-    SOURCE cmake/tests/cxx11_std_shared_ptr.cpp
-    FILE ${ARGN})
-endfunction()
-
-###############################################################################
-function(hpx_check_for_cxx11_std_shared_ptr_lwg3018)
-  add_hpx_config_test(HPX_WITH_CXX11_SHARED_PTR_LWG3018
-    SOURCE cmake/tests/cxx11_std_shared_ptr_lwg3018.cpp
-    FILE ${ARGN})
-endfunction()
```
hkaiser: Are we sure that this is really implemented by all compilers by now?

cmake/HPX_AddConfigTest.cmake
```diff
@@ -424,285 +242,6 @@ function(hpx_check_for_cxx11_std_atomic_128bit)
...
-
-###############################################################################
-function(hpx_check_for_cxx14_return_type_deduction)
-  add_hpx_config_test(HPX_WITH_CXX14_RETURN_TYPE_DEDUCTION
-    SOURCE cmake/tests/cxx14_return_type_deduction.cpp
-    FILE ${ARGN}
-    CMAKECXXFEATURE cxx_return_type_deduction)
-endfunction()
```
hkaiser: I'm not convinced we should remove the C++14 checks at this point. While we required C++11 for a while, we still had all the checks enabled to allow for the compilers to catch up. But please convince me!

hpx/lcos/channel.hpp
```diff
@@ -457,8 +455,7 @@ namespace hpx { namespace lcos
                 send_channel<T>, lcos::server::channel<T>
             > base_type;
 
-        HPX_CONSTEXPR_OR_CONST
-        static std::size_t default_generation = std::size_t(-1);
+        constexpr static std::size_t default_generation = std::size_t(-1);
```
hkaiser: Should we write `constexpr static` or `static constexpr`? I don't care, but it should be consistent.

hpx/util/detail/pack_traversal_async_impl.hpp
```diff
@@ -304,7 +304,7 @@ namespace util {
         {
             explicit static_async_range(Target*) {}
```
hkaiser: This can be `constexpr` as well.

libs/config/include/hpx/config/constexpr.hpp
```diff
@@ -5,42 +5,31 @@
...
-#else
-
-// clang-format off
-#if defined(HPX_HAVE_CXX11_CONSTEXPR) && !defined(HPX_MSVC_NVCC) &&            \
-    !(defined(__NVCC__) && defined(__clang__))
-#   define HPX_CONSTEXPR constexpr
-#   define HPX_CONSTEXPR_OR_CONST constexpr
-#else
-#   define HPX_CONSTEXPR
-#   define HPX_CONSTEXPR_OR_CONST const
```
hkaiser: Should we leave the definition of `#define HPX_CONSTEXPR_OR_CONST constexpr` in place as well?

libs/coroutines/src/detail/coroutine_self.cpp
```diff
@@ -14,7 +14,7 @@
 namespace hpx { namespace threads { namespace coroutines { namespace detail {
     coroutine_self*& coroutine_self::local_self()
     {
-        HPX_NATIVE_TLS coroutine_self* local_self_ = nullptr;
```
hkaiser: Hah, this was a bug to begin with! I wonder how it ever worked?

tests/unit/resource/CMakeLists.txt
```diff
@@ -16,7 +16,6 @@ set(tests
     suspend_thread_timed
 )
 
-# TODO: Enable unconditionally when C++14 is required.
 if (HPX_CXX_STANDARD GREATER 11)
```
hkaiser: Shouldn't this be unconditional now?

cmake/HPX_AddConfigTest.cmake
msimberg: No, I can add this one back. I can't find any reliable information on when it's been fixed so better keep it.

cmake/HPX_AddConfigTest.cmake
msimberg: Fair enough, but it comes down to what compilers we're planning on supporting. The pycicle builds with the oldest gcc and clang, and nvcc/clang-cuda are happy (they are forced to C++14). Do you know if MSVC has problems with any of these? Intel should be fine with anything that's in the standard library (rather than compiler) since it uses libstc++. Some of the checks are completely unused. I promise to add back whichever checks are necessary if it turns out a compiler that we care about doesn't support them. I won't fight back if you'd rather have them around (it does give a bit of peace of mind...).

Lastly (I know this isn't a very strong argument) if a compiler by now does not support C++14 fully it's probably not worth the effort to support it (I realize that in this case we already had the effort to support such cases, and I'm removing it).

tests/unit/resource/CMakeLists.txt
msimberg: Yes! Good catch.

libs/coroutines/src/detail/coroutine_self.cpp
hkaiser: According to the standard it is:
>When thread_local is applied to a variable of block scope the storage-class-specifier static is implied if it does not appear explicitly.

See, and now I have learned some thing new. I think we should have the `static` explicitly, however.

libs/config/include/hpx/config/constexpr.hpp
hkaiser: Doh! Didn't see that.

cmake/HPX_AddConfigTest.cmake
msimberg: I added this one and the `quick_exit` check back.

msimberg:
The last few commits fix a few more inconsistencies (`constexpr` placement, `static thread_local`, adds back missing feature checks).

msimberg:
I've rebased this on master again. Are there any final comments or can we go ahead with this if builds pass?

msimberg:
Just waiting for pycicle to finish the last builds. They should finish tonight.

msimberg:
Rebased once again... This time I'll only wait for one of the builders to finish. If all looks good there I'll merge this.

PR #4019, opened 2019-08-08
WIP: Module topology

aurianer:


libs/topology/include/hpx/topology/topology.hpp
```diff
@@ -382,7 +378,12 @@ namespace hpx { namespace threads
 
         std::size_t num_of_pus_;
 
-        mutable hpx::util::spinlock topo_mtx;
+        using mutex_type = boost::detail::spinlock;
```
msimberg: Sorry @hkaiser, I misremembered how many places need to use boost spinlock directly. There shouldn't be many more though...

libs/topology/include/hpx/topology/topology.hpp
```diff
@@ -392,7 +393,7 @@ namespace hpx { namespace threads
         std::vector<std::size_t> socket_numbers_;
         std::vector<std::size_t> numa_node_numbers_;
         std::vector<std::size_t> core_numbers_;
-//        std::vector<std::size_t> pu_numbers_; // (empty vector)
+        //        std::vector<std::size_t> pu_numbers_; // (empty vector)
```
msimberg: I think we can just delete this line.

tests/unit/topology/CMakeLists.txt
```diff
@@ -14,7 +14,9 @@ if (HPX_WITH_SHARED_PRIORITY_SCHEDULER AND HPX_WITH_CXX14_RETURN_TYPE_DEDUCTION)
     THREADS_PER_LOCALITY -2
     ARGS --size=128 --nb=512 --tiles-per-domain=2 --col-proc=1 --row-proc=1)
 endif()
-set(numa_allocator_FLAGS COMPONENT_DEPENDENCIES iostreams)
+set(numa_allocator_FLAGS
+  COMPONENT_DEPENDENCIES iostreams
+  DEPENDENCIES hpx_topology)
```
msimberg: This is probably a leftover from my changes but I don't think we need to have `hpx_topology` as dependency here. We do `add_hpx_executable` without `NOLIBS` so we should get it automatically through `hpx` as a dependency.

libs/topology/include/hpx/topology/topology.hpp
msimberg: In fact, the concurrency module will be next and we can make this depend on the concurrency module allowing us to keep `hpx::util::spinlock` here. @aurianer will take care of updating this.

hkaiser:
Just a thought: we may want to merge this with the already existing hardware module. 

Also, just an observation: this module is the only one depending on hwloc, should that be reflected in the build system for this module?

aurianer:
- @msimberg told me that the hardware module is pretty different but probably badly named, whereas this one is a wrapper of hwloc
- This will be reflected in the build system once my target directives will be ready, soon I promise ^^
I'm still doing some tests for all the options and it's pretty long haha

msimberg:
@aurianer please check the build with GCC. We might need to export the `topology` struct: https://cdash.cscs.ch//viewBuildError.php?buildid=69260.

aurianer:
> @aurianer please check the build with GCC. We might need to export the `topology` struct: https://cdash.cscs.ch//viewBuildError.php?buildid=69260.

Checked and for now I just inlined the `get_machine_affinity_mask` function, seems to work locally :)

msimberg:
> What's the rationale of removing `HPX_HAVE_THREAD_TARGET_ADDRESS`?

Two reasons:
1. It was unused, and
2. It created a dependency on naming.

I think we discussed this sometime earlier, and concluded that that functionality never really went anywhere and if we'd like it back we might want to start over. I assumed the intention of that feature was something a bit more automatic than manually passing NUMA hints, but if we'd like it back I think we could build the feature more cleanly on top of the scheduler hints.

libs/topology/include/hpx/topology/topology.hpp
msimberg: @aurianer let's change this back to `hpx::util::spinlock` now. (I think the timeout on the `numa_allocator` test comes from the Boost spinlock requiring a special initial value.)

PR #4020, opened 2019-08-08
Module concurrency

aurianer:


libs/concurrency/include/hpx/concurrency/register_locks.hpp
```diff
@@ -26,7 +26,7 @@ namespace hpx { namespace util {
     template <typename Lock, typename Enable = void>
     struct ignore_while_checking;
 
-#if defined(HPX_HAVE_VERIFY_LOCKS) || defined(HPX_EXPORTS)
+#if defined(HPX_HAVE_VERIFY_LOCKS)
```
hkaiser: What is the rationale for this change?

msimberg: Me being clueless... We had problems getting this to link correctly and tried a few things. I think @auraner will change it to be `|| defined HPX_MODULE_EXPORTS`.

Part of the problems came from the `HPX_DEBUG` here: https://github.com/STEllAR-GROUP/hpx/blob/58bc53642b4e13601e071f62bcbc1d3716140b7c/examples/hello_world_component/CMakeLists.txt#L30. Would you happen to know what the idea behind it is? Is it explicitly there to check that we are (more or less) ABI compatible between debug and release?

hkaiser: Yes, this was an attempt to keep things ABI compatible between Release/Debug - and yes I know we don't do this consistently...

msimberg: Thanks, makes sense. We'll try not to make things at least worse then, in case we still decide to do this consistenly across the codebase at some point.

msimberg:
This builds and passes all tests now (the remaining failures reported by pycicle are unrelated). @hkaiser if you have time a review would be appreciated. The only potentially surprising change that was made is that we moved `cache_line_size` to the `concurrency` module to allow `topology` to use the spinlocks in `concurrency` without a circular dependency. Not perfect, but I figured it fits well enough in `concurrency` to be worth it.

libs/concurrency/src/itt_notify.cpp
```diff
@@ -286,36 +281,28 @@ namespace hpx { namespace util { namespace itt
...
-            reinterpret_cast<std::size_t>(sh_.handle_));
+        id_ = HPX_ITT_MAKE_ID(
+            domain_.domain_, reinterpret_cast<std::size_t>(sh_.handle_));
 
         HPX_ITT_TASK_BEGIN_ID(domain_.domain_, id_, sh_.handle_);
-
-        if (name.kind() == util::thread_description::data_type_address)
-        {
-            add_metadata(sh_.handle_, name.get_address());
-        }
```
hkaiser: Has this been moved somewhere else?

src/hpx_init.cpp
```diff
@@ -340,6 +340,53 @@ namespace hpx
...
+                            back_trace);
+                }
+            }
+        }
+
+        bool register_locks_predicate()
+        {
+            return threads::get_self_ptr() != nullptr;
+        }
+#endif
```
hkaiser: Does this code needs to live in hpx_init.cpp (i.e. does it need to be linked to the executable)? Wouldn't it be ok to have it exposed from the main core library?

msimberg: No, and yes, I think. All of the handlers above could go in a separate file (just like `termination_handler` in fact).

hkaiser: I agree. This file should be part of HPX core, however, not HPX init.

msimberg: Yep, that's what I understood from your first comment, although my reply might have been a bit unclear.

aurianer: I'll do it in a separate PR :)

PR #4021, opened 2019-08-12
Warn about compiling in C++11 mode

msimberg:
Add a deprecation warning when compiling in C++11 mode (and hard error for C++0X). In preparation for #4017.


msimberg:
How does

cmake/HPX_DetectCppDialect.cmake
```diff
@@ -67,12 +67,13 @@ function(hpx_detect_cpp_dialect_non_msvc)
             if(HPX_WITH_CXX11)
               set(CXX_FLAG -std=c++11 PARENT_SCOPE)
               hpx_info("C++ mode used: C++11")
+              hpx_warn("Compiling in C++11 mode is deprecated. HPX will require C++14 support in future releases. Set HPX_WITH_CXX14 (or newer) to ON during CMake configuration to enable C++14 support.")
```
msimberg: How does this sound @hkaiser?

hkaiser: Sounds good!

PR #4024, opened 2019-08-16
Fix padding calculation in cache_aligned_data.hpp

msimberg:
Correctly sets the padding to 0 when the size of the data is a multiple of the cache line size. If the calculation looks too cryptic I can change it to something a bit more explicit.

I've also made `get_cache_line_padding_size` and `get_cache_line_size` unconditionally `constexpr` since they have to be `constexpr`.

### DIFF
--- a/hpx/util/cache_aligned_data.hpp
+++ b/hpx/util/cache_aligned_data.hpp
@@ -10,8 +10,21 @@
 #include <hpx/config.hpp>
 #include <hpx/runtime/threads/topology.hpp>
 
-namespace hpx { namespace util
-{
+#include <cstddef>
+
+namespace hpx { namespace util {
+    namespace detail {
+        // Computes the padding required to fill up a full cache line after
+        // data_size bytes.
+        constexpr std::size_t get_cache_line_padding_size(
+            std::size_t data_size)
+        {
+            return (threads::get_cache_line_size() -
+                       (data_size % threads::get_cache_line_size())) %
+                threads::get_cache_line_size();
+        }
+    }    // namespace detail
+
     ///////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


hpx/util/cache_aligned_data.hpp
```diff
@@ -10,8 +10,19 @@
...
-namespace hpx { namespace util
-{
+namespace hpx { namespace util {
+    namespace detail {
+        // Computes the padding required to fill up a full cache line after
+        // data_size bytes.
+        constexpr std::size_t get_cache_line_padding_size(
+            std::size_t data_size)
+        {
+            return (threads::get_cache_line_size() -
```
biddisco:
```
    template <typename Data>
    struct cache_aligned_data
    {
        static constexpr unsigned int padding1 = sizeof(Data) % threads::get_cache_line_size();
        static constexpr unsigned int padding2 = (padding1==0) ? 0 : threads::get_cache_line_size()- padding1;
        
        constexpr std::size_t padding() const {
            return padding2;
        }
        
        // pad to cache line size bytes
        Data data_;
        char cacheline_pad[padding2];        
    };
```

biddisco: I use an extra check to pad with zero if the cache is aligned already (modulus returns zero). then if you have a 64 byte struct or something, it doesn't use 128 bytes.

msimberg: This calculation has the same effect (that's the whole point of this PR). But as I said I'll change it to something more explicit like your version if it's not very clear from the way it's written now. Is that what you're suggesting?

biddisco: sorry. I misread the code. Looks fine.

PR #4025, opened 2019-08-16
Refactoring thread queue configuration options

msimberg:
Passes various thread queue related configuration options (miscellaneous min/max counts, stack sizes, max idle backoff time) explicitly into the thread queues instead of getting them from the configuration. The configuration values are extracted in the thread manager instead.

Also adds a test which creates a standalone `thread_pool_os_executor` without the runtime.

Flyby:
- Deprecate `HPX_SCHEDULER_MAX_TERMINATED_THREADS` CMake option in favor of setting it through configuration values.
- Remove hardcoded and duplicated max thread count values from thread queues and schedulers in favor of setting it through configuration values.
- Fix thread queue constructor arguments (`max_queue_thread_count_` passed as `queue_num`).
- clang format most of the touched files.

@biddisco this probably conflicts with your changes to `thread_queue_mc` and `shared_priority_queue_scheduler`. I'll revert my changes to those files if the conflicts are bad and rebase once you've got your PR up.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -624,16 +624,16 @@ hpx_option(HPX_WITH_SPINLOCK_POOL_NUM STRING
 
 hpx_add_config_define(HPX_HAVE_SPINLOCK_POOL_NUM ${HPX_WITH_SPINLOCK_POOL_NUM})
 
-# Count number of terminated threads before forcefully cleaning up all of
-# them. Note: terminated threads are cleaned up either when this number is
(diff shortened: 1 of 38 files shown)


msimberg:
@biddisco thanks for having a look.

1. Yes, I'm sorry about the whitespace changes. I decided in the end to apply them since we're doing it with the modules anyway already. If I hadn't applied clang-format now it would've happened in the next few modules anyway. The goal is to enforce the formatting on PRs once we've applied the changes everywhere.
2. Good idea, will add a note there. (Short story: it tests that one can create a `thread_pool_os_executor` without starting the runtime.)

PR #4035, opened 2019-08-22
Fixing support for mpirun

### LINKED ISSUE #4032: Simple hello world fails to launch correctly
## Expected Behavior

From this example code:

```
#include <hpx/hpx_init.hpp>
int hpx_main(int argc, char* argv[]) {
	printf( "Hello World!\n");
	return hpx::finalize();
}
int main(int argc, char* argv[]) {
	hpx::init(argc, argv);
}
```

I expect there to be one instance of hpx_main called and the output to be:
```
Hello World!
```
## Actual Behavior

When run with mpirun -np 2, the output is 
```
Hello World!
Hello World!
```

I compiled the above code using cmake file:
```
project(hello CXX)
find_package(HPX REQUIRED NO_CMAKE_PACKAGE_REGISTRY)

add_hpx_executable(
  hello
  DEPENDENCIES
  SOURCES
    main.cpp
)
```

and executed it with mpirun -np 2 ./hello

## Specifications

  - HPX Version: Top of master
  - Platform (compiler, OS): GNU 8.2 Ubuntu 18.04

hkaiser:
- flyby: remove dependency on Boost.Assign from command line handling

Fixes #4032

### DIFF
--- a/plugins/parcelport/mpi/mpi_environment.cpp
+++ b/plugins/parcelport/mpi/mpi_environment.cpp
@@ -15,7 +15,6 @@
 #include <hpx/util/command_line_handling.hpp>
 #include <hpx/plugins/parcelport/mpi/mpi_environment.hpp>
 
-#include <boost/assign/std/vector.hpp>
 #include <boost/lexical_cast.hpp>
 #include <boost/tokenizer.hpp>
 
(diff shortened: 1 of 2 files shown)


dmarce1:
This fixes the problem.

Issue #3440, opened 2018-09-03
Explore feasibility of removing all dependencies on Boost


biddisco:
A discussion at CSCS took place where the anti-boost argument went as follows

Q: The idea is that eventually the features in HPX will be available via the c++ standard?
A: Ideally, yes most of the HPX features will be in c++ 
Q: Will the c++ standard have a dependency on boost then?
A: Of course not.
Q: Well then, you need to get rid of the boost dependency in HPX before it will be adopted by users X,Y,Z

Currently we make use of `program options` which could be replaced (given a lot of effort), by [cxxopt](https://github.com/jarro2783/cxxopts) or an alternative.

other boost libraries that we use are

- atomic - do we need it other than for our compat layer?
- filesystem - this should be in c++17?
- program_options - could be replaced by alternative in principle
- regex - why not use std::regex? (some work converting expressions)
- system - no idea if we really need this
- chrono - should be in std::
- context ?
- thread ?
- spirit ?

It would seem on casual inspection that we rely on context/thread and others could be removed apart from spirit that is used in parsing thread binding options (yes? elsewhere?).

Are there other parts of our code that rely on boost that are not listed here?
How much effort would it take to absorb context/thread into HPX? (and do it in such a way that we could still benefit from upstream fixes)

Does anyone feel strongly about this? (either in favour or against using boost).

K-ballo:
- `boost::atomic` is mostly gone, only one use case remains. Some deprecated executor is making use of an invalid `atomic<>` specialization which standard implementations diagnose, we are waiting for that executor to simply be removed.

- `boost::regex` is a tricky one, `std::regex` lacks multiline support before 17, and standard implementations aren't capable of dealing with big inputs as boost one is (matches that take a couple seconds to boost take up to 20mins for std, when it doesn't bail out early due to complexity/limits). Many inspect checks rely on matching regexes with lot of backtracking on entire source files, switching to `std::regex` ruins those checks completely. There's been an attempt to replace inspect checks regexes with token-based matchers, it was eventually abandoned in favor of clang's ast matcher but it could be revived.

[...]

PR #4036, opened 2019-08-23
Add C++17 filesystem compatibility header

msimberg:
Adds a basic compatibility layer for C++17 filesystem. Mostly just forwards `std::filesystem` to `hpx::filesystem` if it's available and otherwise forwards `boost::filesystem` instead. The appropriate `error_code` is introduced into the `hpx::filesystem` namespace.

~It's not 100% transparent because of `error_code`. I've so far put explicit ifdefs where needed. The other main difference is that `canonical` doesn't take a base path for relative paths. In some cases we set this to `initial_path` but I'm not sure we care about the difference between that and `current_path`. I could probably put together something that does the same with `std` features.~

Where no naming changes were required I've just assumed that the functions behave the same, but let's see first if the tests even pass.

If someone wants to try `std::filesystem` it requires GCC 9, or an older GCC and explicit linking to `stdc++fs`. I think it's not yet in any released `libc++` but it can be linked to explicitly with `c++fs` as well.

Part of #3440.

### DIFF
--- a/libs/filesystem/include/hpx/filesystem.hpp
+++ b/libs/filesystem/include/hpx/filesystem.hpp
@@ -1,0 +1,76 @@
+//  Copyright (c) 2019 Mikael Simberg
+//
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+/// \file
+/// This file provides a compatibility layer using Boost.Filesystem for the
+/// C++17 filesystem library. It is *not* intended to be a complete
+/// compatibility layer. It only contains functions required by the HPX
+/// codebase. It also provides some functions only available in Boost.Filesystem
+/// when using C++17 filesystem.
+
+#if !defined(HPX_FILESYSTEM_HPP)
+#define HPX_FILESYSTEM_HPP
+
+#include <hpx/config.hpp>
+
+#if defined(HPX_HAVE_CXX17_FILESYSTEM)
+#include <filesystem>
+#include <system_error>
+
+namespace hpx { namespace filesystem {
+    using namespace std::filesystem;
+    using std::error_code;
+    using std::filesystem::canonical;
+
+    inline path initial_path()
+    {
+        static path ip = current_path();
+        return ip;
+    }
+
+    inline path basename(path const& p)
+    {
+        return p.stem();
+    }
+
+    inline path canonical(path const& p, path const& base)
+    {
+        if (p.is_relative())
+        {
+            return canonical(base / p);
+        }
+        else
+        {
+            return canonical(p);
+        }
+    }
+
+    inline path canonical(path const& p, path const& base, error_code& ec)
+    {
+        if (p.is_relative())
+        {
+            return canonical(base / p, ec);
+        }
+        else
+        {
+            return canonical(p, ec);
+        }
+    }
+
+}}    // namespace hpx::filesystem
+#else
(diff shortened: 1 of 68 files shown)


src/runtime/components/server/runtime_support_server.cpp
```diff
@@ -1581,13 +1580,12 @@ namespace hpx { namespace components { namespace server
...
                 boost::char_separator<char> sep(HPX_INI_PATH_DELIMITER);
                 tokenizer tokens(component_path, sep);
-                boost::system::error_code fsec;
                 for(tokenizer::iterator it = tokens.begin(); it != tokens.end(); ++it)
                 {
                     lib = fs::path(*it);
                     fs::path lib_path =
                         lib / std::string(HPX_MAKE_DLL_STRING(component));
-                    if(fs::exists(lib_path, fsec))
+                    if(fs::exists(lib_path))
```
hkaiser: This will throw an exception, no? I think for the purpose of the filesystem API we can introduce a `filesystem::error_code` that maps onto the appropriate type (`std` or `boost`) and use that for the filesystem API calls we have.

src/runtime/components/server/runtime_support_server.cpp
```diff
@@ -2015,13 +2013,12 @@ namespace hpx { namespace components { namespace server
...
```
hkaiser: same as above

src/util/init_ini_data.cpp
```diff
@@ -39,8 +36,12 @@ namespace hpx { namespace util
     bool handle_ini_file (section& ini, std::string const& loc)
     {
         try {
-            namespace fs = boost::filesystem;
+            namespace fs = filesystem;
+#if defined(HPX_HAVE_CXX17_FILESYSTEM)
+            std::error_code ec;
+#else
             boost::system::error_code ec;
+#endif
```
hkaiser: Yah, this confirms it. Let's introduce `hpx::filesystem::error_code` instead.

msimberg: Good idea. Thanks!

msimberg:
Updated. The error code now comes from either `std` or `boost` depending on which `filesystem` is used. I also added the wrappers for `canonical`, `initial_path` and `basename` that exist in `boost::filesystem` but not in `std::filesystem`. Boost.Filesystem is also not linked to if `filesystem` exists in the standard library.

libs/filesystem/include/hpx/filesystem.hpp
```diff
@@ -0,0 +1,76 @@
...
+        static path ip = current_path();
+        return ip;
+    }
+
+    inline path basename(path p)
+    {
+        return p.stem();
+    }
+
+    inline path canonical(path p, path base)
```
hkaiser: Couldn't these be passed by `const&` instead?

libs/filesystem/include/hpx/filesystem.hpp
```diff
@@ -0,0 +1,76 @@
...
+    using std::error_code;
+    using std::filesystem::canonical;
+
+    inline path initial_path()
+    {
+        static path ip = current_path();
+        return ip;
+    }
+
+    inline path basename(path p)
```
hkaiser: Couldn't this be passed by `const&` instead?

libs/filesystem/include/hpx/filesystem.hpp
msimberg: Yep, and should probably be passed like that as well. That's what the `std::filesystem::canonical` takes anyway. Good catch, thanks!

PR #4038, opened 2019-08-25
Adding MPark's variant (V1.4.0) to HPX

hkaiser:
This PR adds MPark's implementation of `std::variant` (https://github.com/mpark/variant) to HPX' datastructures module.

- adding serialization support for hpx::util::variant
- flyby: add support for skipping whole files to inspect (without modifying them)

This is in preparation of moving `boost::program_options` to HPX (see #3440)


hkaiser:
Actually, we don't need variant for program_options, just optional and any. It's probably fine to close this without merging...

msimberg:
> Actually, we don't need variant for program_options, just optional and any. It's probably fine to close this without merging...

We do use variant, even though it's in only one place (`parse_affinity_options`) so it probably wouldn't hurt to have this in for that. In any case, the serialization test and inspect changes are useful as well.

hkaiser:
@msimberg: we will have to wait for @mpark to disable the warning/error generated by the variant header (see https://github.com/mpark/variant/issues/68). Alternatively, we could disable this diagnostic in our build system (for now).

hkaiser:
> We do use variant, even though it's in only one place (parse_affinity_options) so it probably wouldn't hurt to have this in for that. In any case, the serialization test and inspect changes are useful as well.

@msimberg: isn't `variant` used only in the context of the Spirit parser? In this case replacing it would be feasible but not absolutely necessary.

msimberg:
Yes, only for the parser. And if spirit is optional, so is variant... Good point! And do we actually need the variant serialization? Maybe you're right that we should just close this without merging.

PR #4040, opened 2019-08-26
Program options module

hkaiser:
Working towards #3440, relies on #4039.


hkaiser:
@msimberg I think this is good to go now. Please review.

tests/unit/threads/CMakeLists.txt
```diff
@@ -79,10 +80,10 @@ foreach(test ${tests})
 
 endforeach()
 
-set_property(TARGET lockfree_fifo_test APPEND
-    PROPERTY COMPILE_DEFINITIONS "HPX_NO_VERSION_CHECK")
-
-target_include_directories(lockfree_fifo_test PRIVATE ${HPX_SOURCE_DIR})
+target_compile_definitions(lockfree_fifo_test
+  PRIVATE HPX_MODULE_STATIC_LINKING HPX_NO_VERSION_CHECK)
```
msimberg: Why do we need this definition?

msimberg:
We'll need to disable deprecation warnings in the fallback test for `boost::program_options`. I think the relevant switch for that is `-Wdeprecated-declarations"` for both clang and gcc. You'll probably know the switch for MSVC.

It looks like both the examples and tests for HPX.Compute are still using `boost::program_options`.

hkaiser:
> We'll need to disable deprecation warnings in the fallback test for boost::program_options. I think the relevant switch for that is -Wdeprecated-declarations" for both clang and gcc. 

How do I do that nowadays for a single target?

msimberg:
Something like [this](https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/plugins/parcelport/mpi/mpi.hpp), most likely around the include, not the usage (not sure). Adding `-Wno-deprecated-declarations` to the target might also work (not sure if the order will be correct). I can give it a try tomorrow if you'd prefer not to deal with waiting for CircleCI results.

hkaiser:
> It looks like both the examples and tests for HPX.Compute are still using boost::program_options.

@msimberg Could you point me to those files, please. I can't see any uses of `boost::program_options` in my source tree anymore :/

msimberg:
> @msimberg Could you point me to those files, please. I can't see any uses of `boost::program_options` in my source tree anymore :/

It's the `.cu` files:
https://github.com/STEllAR-GROUP/hpx/blob/98a89c8e7ea75203d497da43e66354dac0e4d363/examples/compute/cuda/data_copy.cu#L17

hkaiser:
Uhh, right :/ Will fix.

msimberg:
Thanks @hkaiser! This looks good now. CircleCI failed on the build unit tests step, but I think because of CircleCI, not because of this PR. I've restarted it to see if it finishes this time.

PR #4050, opened 2019-08-28
Creating plugin module

hkaiser:
This also gets rid of any last uses of `boost::any` and `boost::shared_ptr`.


hpx/plugins/binary_filter_factory_base.hpp
```diff
@@ -10,8 +10,8 @@
 #include <hpx/plugins/plugin_factory_base.hpp>
 #include <hpx/runtime/serialization/binary_filter.hpp>
 
-#include <hpx/util/plugin.hpp>
-#include <hpx/util/plugin/export_plugin.hpp>
+#include <hpx/plugin.hpp>
```
msimberg: Should we try to do the same as with the other modules where `hpx/modulename.hpp` just includes all (public) headers from the module (`GLOBAL_HEADER_GEN ON`)? And maybe rename this one `hpx/plugin/plugin.hpp`?

libs/plugin/include/hpx/plugin/detail/dll_dlopen.hpp
```diff
@@ -37,44 +32,47 @@
...
+            union
+            {
+                FROM f;
+                TO t;
             } u;
             u.f = f;
             return u.t;
         }
-    }
+    }    // namespace very_detail
```
msimberg: Unrelated but this one should probably go to the same place the `nasty_cast` from the coroutines went (unless we really need it of course...).

hkaiser:
As it turns out, older `std::smart_ptr` implementations don't support deleters (https://cdash.cscs.ch//viewBuildError.php?buildid=70255). I'm not sure how to proceed. We could use `std::shared_ptr` conditionally, falling back to `boost::shared_ptr` if needed. Alternatively we could stop supporting those platforms, however I'm not sure what the new minimal compiler versions would have to be.

hkaiser:
> As it turns out, older std::smart_ptr implementations don't support deleters (https://cdash.cscs.ch//viewBuildError.php?buildid=70255). I'm not sure how to proceed. We could use std::shared_ptr conditionally, falling back to boost::shared_ptr if needed. Alternatively we could stop supporting those platforms, however I'm not sure what the new minimal compiler versions would have to be.

As @K-ballo has clarified (see http://irclog.cct.lsu.edu/ste~b~~b~ar/2019-08-29#1567079474-1567079570;), that the issue is not the missing support for deleters, but the missing support for binding function pointers (see: https://cplusplus.github.io/LWG/issue3018). This will allow to create a feature test to detect support for our use case.

libs/plugin/include/hpx/plugin/detail/dll_dlopen.hpp
msimberg: Ah, good question. I replaced it with a `reinterpret_cast`. Do you know if we really can't get away with one of the `std` casts here?

hpx/plugins/binary_filter_factory_base.hpp
msimberg: Sorry, no reason. I thought the current `hpx/plugin.hpp` contained something in addition to just including other headers, but I see that's not the case.

hkaiser:
@msimberg I don't understand where the errors on pycicle come from. Those look very unrelated to me.

msimberg:
> @msimberg I don't understand where the errors on pycicle come from. Those look very unrelated to me.

Yep, I think they are if you're talking about [this](https://cdash.cscs.ch/viewTest.php?onlyfailed&buildid=70436) and [this](https://cdash.cscs.ch/viewTest.php?onlyfailed&buildid=70420). I haven't had the time or motivation to look into them yet.

hkaiser:
All the recent builds on Pycicle are green. I have no idea why it shows errors above.

hkaiser:
@msimerg This is good to go now, please have a look.

msimberg:
@hkaiser just a heads up: I merged master into this branch (there was a conflict).

PR #4054, opened 2019-08-30
Add troubleshooting section to manual

msimberg:
Adds a short troubleshooting section to the manual. Generally I think we should try to keep tips like these in their appropriate sections, but some things pop up frequently enough that I think it's worth having a separate section just for common problems. Doesn't contain much yet, but it'll hopefully expand over time.

### DIFF
--- a/docs/sphinx/manual/troubleshooting.rst
+++ b/docs/sphinx/manual/troubleshooting.rst
@@ -1,0 +1,51 @@
+..
+    Copyright (C) 2019 Mikael Simberg
+
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+.. _troubleshooting:
+
+===============
+Troubleshooting
+===============
+
+This section contains commonly encountered problems when compiling or using HPX.
+
+.. _troubleshooting_program_options:
+
+``Undefined reference to boost::program_options``
+=================================================
+
+Boost.ProgramOptions is not ABI compatible between all C++ versions and
+compilers. Because of this you may see linker errors similar to this:
+
+.. code-block:: text
+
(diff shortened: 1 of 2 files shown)


hkaiser:
FWIW, @McKillroy has started an HPX FAQ documentation (https://github.com/McKillroy/hpx_snippets), where this kind of information could go.

docs/sphinx/manual/troubleshooting.rst
```diff
@@ -0,0 +1,48 @@
...
+Boost.ProgramOptions is not ABI compatible between all C++ versions and compilers.
+Because of this you may see linker errors similar to this:
+
+.. code-block:: text
+
+   ...: undefined reference to `boost::program_options::operator<<(std::ostream&, boost::program_options::options_description const&)'
+
+if you are not linking to a compatible version of Boost.ProgramOptions. We
+recommend that you always build |boost|_ libraries and |hpx| with the same
+compiler and C++ standard.
```
hkaiser: Update this with regards to the new `hpx::program_options` module?

msimberg: Yes, my memory is indeed short...

msimberg: Done.

msimberg:
> FWIW, @McKillroy has started an HPX FAQ documentation (https://github.com/McKillroy/hpx_snippets), where this kind of information could go.

Yeah, that's really nice indeed. But long term we want to have these kind of things in our documentation, no? We shouldn't need to rely on @McKillroy maintaining this.

McKillroy:
Just to clarify - my idea was just to have something I could quickly and easily work on, with a positive side effect you could use it and integrate what looks useful to you. 
I totally don't want to create some extra "grey" HPX documentation. 
The best effect could be that some of my stuff simply become superfluous, because it's all covered in the official documentation.

PR #4057, opened 2019-08-31
Printing configuration summary for modules as well

hkaiser:
- generate configuration options as strings for each module and print this with `--hpx:info`
~~- introduce global configuration option `HPX_WITH_COMPATIBILITY_HEADERS`~~


cmake/HPX_AddModule.cmake
```diff
@@ -33,30 +33,34 @@ function(add_hpx_module name)
...
-      "Enable compatibility headers for old headers"
-      ON ADVANCED
-      CATEGORY "Modules")
-    if(HPX_${name_upper}_WITH_COMPATIBILITY_HEADERS)
-      hpx_add_config_define_namespace(
-          DEFINE HPX_${name_upper}_HAVE_COMPATIBILITY_HEADERS
-          NAMESPACE ${name_upper})
-    endif()
+  set(_compatibility_headers_default OFF)
+  if(${HPX_WITH_COMPATIBILITY_HEADERS} AND ${name}_COMPATIBILITY_HEADERS)
```
msimberg: I see the appeal of the global option but we may end up with compatibility headers that have been introduced in different versions for different modules. This puts all the modules on the same "cycle". We also end up with `HPX_WITH_MODULENAME_COMPATIBILITY_HEADERS` variables for modules that don't have any. I'd prefer not doing this, but don't mind if there's a strong case for this.

hkaiser: This is a fair point. I wanted to have a global option I could use for a sensible default. I'll revert this particular change.

hkaiser: That has been taken care of now.

PR #4058, opened 2019-08-31
Do not build networking tests when networking disabled

biddisco:


tests/unit/CMakeLists.txt
```diff
@@ -22,6 +22,10 @@ set(subdirs
     util
    )
 
+if(HPX_WITH_NETWORKING)
+  set(subdirs ${subdirs} parcelset)
+endif()
```
msimberg: Please remove `parcelset` from the list above.

msimberg: Ping @biddisco?

biddisco: doh! fixed.

PR #4059, opened 2019-08-31
Renaming module hpx_parallel_executors into hpx_execution

hkaiser:
Just giving it a shorter name. Also, not all implemented executors are 'parallel'...


hkaiser:
… or should we name it `hpx_execution` - might be even more appropriate ...

msimberg:
> … or should we name it `hpx_execution` - might be even more appropriate ...

Good idea. `execution` would match the standard library header. With that name we would also end up with a global header called `hpx/execution.hpp` which I think is nicely consistent. The related question to this is if we should move the executors into `hpx::execution` or even just `hpx`. The documentation doesn't help here, but we've noticed people having trouble figuring out e.g. that `par` is in `hpx::parallel::execution` (plus it's quite a mouthful). I'd be in favour of going all the way up to `hpx` since that would make the `std::par` to `hpx::par` replacement more natural, but `hpx::execution` would be more consistent with what we've been doing with some of the other modules.

hkaiser:
@msimberg The standard has `par` and friends in `std::execution::par` (https://en.cppreference.com/w/cpp/algorithm/execution_policy_tag). If we are going to change the namespaces we should move them into `hpx::execution`. Otherwise I agree 100%.

msimberg:
> @msimberg The standard has `par` and friends in `std::execution::par` (https://en.cppreference.com/w/cpp/algorithm/execution_policy_tag). If we are going to change the namespaces we should move them into `hpx::execution`. Otherwise I agree 100%.

Thanks for correcting me! That makes the decision easier I guess. Let's do `hpx_execution` for the module name then and try to move the functionality to `hpx::execution` eventually as well.

libs/execution/CMakeLists.txt
```diff
@@ -5,7 +5,9 @@
 
 cmake_minimum_required(VERSION 3.3.2 FATAL_ERROR)
 
-set(parallel_executors_headers
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
```
msimberg: What is this one for? I don't see any `libs/execution/cmake` directory...

hkaiser: This is more for consistency with other modules than anything else.

msimberg: Ah, indeed. Some modules have that line, some don't. I've been trying to remove them since the modules just had empty `cmake` directories, but now it looks like none of them have the directory anymore. I'd rather remove the line in all modules and include it only if needed. What do you think?

hkaiser: >  I'd rather remove the line in all modules and include it only if needed. What do you think?

Fair enough!

PR #4061, opened 2019-09-02
Add clang-format checking step for modules

### LINKED ISSUE #4052: Enforce clang-format style for modules
Since modules are being reformatted before going in we could already enforce that their formatting stays clang-formatted on CircleCI.

aurianer:
Fixes #4052

Add a circleci step to check the clang-formatting of the modules


aurianer:
I'm formatting some modules as they are not all completely formatted, and I found out about this change, is  this wanted ?
```
-          : base_type(val), ref_count_(0)
+          : base_type(val)
+          , ref_count_(0)
```

hkaiser:
> I'm formatting some modules as they are not all completely formatted, and I found out about this change, is this wanted ?

We agreed to go with clang-formatted modules. So let's accept whatever it produces.

.circleci/config.yml
```diff
@@ -116,6 +116,24 @@ jobs:
...
+    steps:
+      - attach_workspace:
+          at: /hpx
+      - run:
+          name: Check that the modules are well clang-formatted
+          command: |
+              cd /hpx/source/libs && shopt -s globstar # to activate the ** globbing
+              clang-format-8 -i **/*.{cpp,hpp}
+              git diff --exit-code --name-status > /tmp/modified_format_files.txt
+              if [[ $(git diff --exit-code --name-status) ]]; then exit 1; fi
```
hkaiser: Do we really have to execute git twice?

msimberg: @aurianer I *think* you can leave out both the `if` and the `cd /` below. The working directory should get reset for each step. And if `git diff ...` is the last line in the step the exit code of that will be used for the whole step (the redirection doesn't affect it).

aurianer: Yeah I thought that the exit code would be the one of the redirection but it's not !

msimberg:
This needs rebasing again unfortunately :/ We should be able to merge this next though.

aurianer:
I think the test failures are unrelated to the formatting, if you can merge this as soon as possible that'd be great cause it's pretty time consuming to keep it up-to-date with master, lots of conflicts ^^

PR #4062, opened 2019-09-02
The cpu_features tool fails to compile on anything but x86 architectures

hkaiser:
- flyby: removed obsolete explicit linking of the cpu_features tool with the Boost.ProgramOptions library
- flyby: fixed x86 architecture tag on windows


msimberg:
While I like the changes, I'm wondering what the purpose of that tool is? Is it still used for something?

hkaiser:
@msimberg: this tool was never used for anything except for informational purposes. I think it was planned to be used for detecting certain hardware capabilities that them could be used for configuration time decisions. @brycelelbach might remember more.

hkaiser:
@msimberg: should we leave this tool in place or would you suggest I remove it with this PR?

msimberg:
Well, I'd get rid of it, but that's just me. If someone needs it, it could easily live in a separate repo. Have you ever used it?

hkaiser:
Ok, I have removed it. Thanks!

PR #4064, opened 2019-09-03
Fix program_options test with older boost versions

msimberg:
This just disables some more tests when using older versions of Boost that don't support the tested functionality. Flyby: disable building `program_options` tests as part of the `all` target.

### DIFF
--- a/libs/program_options/examples/config_file_types.cpp
+++ b/libs/program_options/examples/config_file_types.cpp
@@ -27,7 +27,8 @@ bool check_float(Double1 test, Double2 expected)
     double expected_d = double(expected);
 
     double seperation = expected_d * (1. + FLOAT_SEPERATION) / expected_d;
-    if ((test_d < expected_d + seperation) && (test_d > expected_d - seperation))
+    if ((test_d < expected_d + seperation) &&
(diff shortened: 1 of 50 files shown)


msimberg:
I've started a couple of builds on rostam as well (no links since they're still in the queue).

msimberg:
@hkaiser did you apply `clang-format` on the `program_options` files? If not, should we? I guess not applying `clang-format` would make it easier to port changes from Boost.ProgramOptions if we feel like keeping them in sync. But in that case I'll add some `// clang-format off` comments for that...

hkaiser:
> @hkaiser did you apply clang-format on the program_options files? If not, should we? I guess not applying clang-format would make it easier to port changes from Boost.ProgramOptions if we feel like keeping them in sync. But in that case I'll add some // clang-format off comments for that...

I did apply clang-format to all files. My reasoning was that this library has not changed much over the years and I don't expect heavy changes down the road.

msimberg:
> I did apply clang-format to all files. My reasoning was that this library has not changed much over the years and I don't expect heavy changes down the road.

Good, sounds reasonable. Let me just push some formatting changes then before merging. It looks like not all files were formatted.

PR #4069, opened 2019-09-03
Moving more facilities to the collectives module

hkaiser:
This depends on #4068


libs/collectives/CMakeLists.txt
```diff
@@ -11,27 +11,59 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
...
+      hpx_cache
       hpx_concurrency
       hpx_config
       hpx_errors
       hpx_format
+      hpx_iterator_support
       hpx_logging
+      hpx_parallel_executors
       hpx_preprocessor
+      hpx_program_options
```
msimberg: Does the module itself need `program_options` or just the tests?

msimberg:
Good idea! Should `spmd_block` be in `collectives` as well?

libs/collectives/CMakeLists.txt
hkaiser: The module depends on things in the big runtime-blob which in turn depend on program_options. I had to add all those dependencies here in order to be able to compile things. I did not try to disentangle this at this point.

hkaiser:
> Good idea! Should spmd_block be in collectives as well?

Yes, probably so. I can add it.

libs/collectives/tests/performance/osu/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+foreach(benchmark ${benchmarks})
+
+  set(sources
+      osu_base.cpp
+      ${benchmark}.cpp)
+
+  source_group("Source Files" FILES ${sources})
+
+  # add example executable
+  add_hpx_executable(${benchmark}
```
msimberg:
```suggestion
  add_hpx_executable(${benchmark}_test
```
The `add_hpx_performance_test` function expects the `_test` suffix... Same below for `coll_benchmarks`.

hkaiser:
This is good to go now

libs/collectives/tests/performance/osu/osu_bibw.cpp
```diff
@@ -141,6 +137,6 @@ void run_benchmark(hpx::program_options::variables_map & vm)
 
         double bw = (size / 1e6 * loop * window_size * 2) / t.elapsed();
 
-        hpx::cout << std::left << std::setw(10) << size << bw << std::endl;
+        std::cout << std::left << std::setw(10) << size << bw << std::endl;
```
msimberg: Was this change intentional?

msimberg:
The only thing I'm a bit worried about is that the `osu_bibw` test now times out on all the debug builds on pycicle. I can't see anything that stands out though so it may just be temporary... I've started a couple of builds on rostam to see if it happens there as well.

biddisco:
Just a clarification question:
In the vtk library (+paraview etc), the modularization was done in such a way that for each module there is usually a set of submodules, with one being the parallel (MPI) version of each algorithm, so a serial build uses module graphics, and a parallel one can also pull in graphics/parallel etc etc

Are we going to have collectives (on node) and collectives/distributed (multinode) as distinct modules - and is this pattern intended for other modules too?

msimberg:
@biddisco yes. Collectives are pretty much by definition distributed so there won't be a local version of this module, but there is already an `algorithms` module with local parallel algorithms and `segmented_algorithms` for the distributed equivalents. The rest of the lcos are going to be split up the same way. They won't be "submodules" in the way you describe them, but there will be local and distributed modules wherever it makes sense.

libs/collectives/tests/performance/osu/osu_bibw.cpp
hkaiser: Yes, I wanted to remove the dependency on iostreams, the benchmark will print on locality zero only anyways.

hkaiser:
> The only thing I'm a bit worried about is that the osu_bibw test now times out on all the debug builds on pycicle. 

@msimberg I think those were real timeouts. I have reduced the windows size used for this test in debug mode (number of messages send/received concurrently), that should take care of the problem.

PR #4071, opened 2019-09-04
Fixing MPI detection for PMIx

sithhell:
Adding PMIx environment variable. This allows to automatically detect the MPI environment on summit.


weilewei:
It works now. Thanks.

PR #4075, opened 2019-09-05
WIP: Module threading

aurianer:


msimberg:
> Should we extract the coroutines code into a separate module?

This is essentially already that. We might want to call this `coroutines` instead?

hkaiser:
> > Should we extract the coroutines code into a separate module?
>
> This is essentially already that. We might want to call this `coroutines` instead?

Hmmm. Then I'd suggest not to include the following files:

- `hpx/threading/apex.hpp`
- `hpx/threading/thread_enums.hpp` 
- `hpx/threading/thread_id_type.hpp`
- `hpx/threading/detail/combined_tagged_state.hpp`

msimberg:
Yep, good idea. I think `thread_id_type` is very tied into the coroutines though, and I feel that's fine as well (I don't think it needs to be separated). Some parts from `thread_enums.hpp` are needed for coroutines but those could be separated out. `combined_tagged_state.hpp` seems very much separate. `apex` is still a work in progress... it depends on `thread_description` and that still needs fixing even if it were to stay in this module.

hkaiser:
Let's move those files into the coroutines directory, then. Also, I think the module itself should be called coroutines (as you suggested), that will also slightly flatten the directory structure.

libs/threading/include/hpx/threading/coroutines/detail/context_generic_context.hpp
```diff
@@ -41,49 +41,54 @@
...
+    void* __splitstack_makecontext(
+        std::size_t, void * [HPX_COROUTINES_SEGMENTS], std::size_t*);
+    void __splitstack_releasecontext(void * [HPX_COROUTINES_SEGMENTS]);
+    void __splitstack_resetcontext(void * [HPX_COROUTINES_SEGMENTS]);
+    void __splitstack_block_signals_context(
+        void * [HPX_COROUTINES_SEGMENTS], int* new_value, int* old_value);
     void __splitstack_getcontext(void * [HPX_COROUTINES_SEGMENTS]);
     void __splitstack_setcontext(void * [HPX_COROUTINES_SEGMENTS]);
 }
+// clang-format on
```
msimberg: What happens with this block if we let `clang-format` do its job?

PR #4078, opened 2019-09-05
Disable parcel-layer if networking is disabled

hkaiser:
This PR removes a lot of code that is unnecessary for the case when networking is disabled (all of the parcel-port and related support code). This reduces the compile time of the main  core module by about 20% (on my machine).


msimberg:
The pycicle builder `daint-gcc-newest` is the one to look out for. That one has networking disabled. Thanks for doing this @hkaiser.

hkaiser:
This is good to go now, please review.

sithhell:
> The pycicle builder daint-gcc-newest is the one to look out for. That one has networking disabled. 

Why does it still run the migrate_component test?

hpx/runtime/actions/base_action.hpp
```diff
@@ -112,6 +112,7 @@ namespace hpx { namespace actions
             was_object_migrated(hpx::naming::gid_type const&,
                 naming::address_type) = 0;
 
+#if defined(HPX_HAVE_NETWORKING)
         /// Return a pointer to the filter to be used while serializing an
```
sithhell: Hmmm, I was sure that we can get rid of the entire `base_action` and `transfer_action` business.

hpx/runtime/actions/basic_action.hpp
```diff
@@ -696,8 +704,15 @@ namespace hpx { namespace serialization
     }}                                                                        \
 /**/
 #endif
+#else // HPX_HAVE_NETWORKING
+#define HPX_REGISTER_ACTION_2(action, actionname)                             \
+    HPX_DEFINE_GET_ACTION_NAME_(action, actionname)                           \
```
sithhell: Do we need an action name when networking is disabled?

hpx/runtime/actions/detail/action_factory.hpp
```diff
@@ -17,26 +17,26 @@
 #include <unordered_map>
 #include <vector>
 
-namespace hpx { namespace actions { namespace detail
-{
+namespace hpx { namespace actions { namespace detail {
+
     struct action_registry
```
sithhell: This shouldn't be needed when networking is disabled.

hpx/runtime/parcelset/put_parcel.hpp
```diff
@@ -33,6 +34,7 @@
```
sithhell: is it not possible to `#ifdef` out the entire file here as well?

src/runtime/actions/detail/action_factory.cpp
```diff
@@ -21,7 +21,9 @@ namespace hpx { namespace actions { namespace detail
 
     void action_registry::register_factory(std::string const& type_name, ctor_t ctor)
```
sithhell: Can't we remove that entirely?

src/runtime/agas/addressing_service.cpp
```diff
@@ -280,6 +295,75 @@ void addressing_service::launch_bootstrap(
...
+    gva runtime_support_address(here
+      , components::get_component_type<components::server::runtime_support>()
+      , 1U, rt.get_runtime_support_lva());
+
+    naming::gid_type lower, upper;
+    get_id_range(HPX_INITIAL_GID_RANGE, lower, upper);
+    rt.get_id_pool().set_range(lower, upper);
+
+    register_name("/0/agas/locality#0", here);
+    if (is_console())
```
sithhell: We are alwas the console when networking is dsiabled, aren't we?

src/runtime/parcelset/locality.cpp
```diff
@@ -21,21 +21,25 @@ namespace hpx { namespace parcelset
 {
     void locality::save(serialization::output_archive& ar, const unsigned int) const
     {
+#if defined(HPX_HAVE_NETWORKING)
         std::string t = type();
```
sithhell: Hmm, can we at least have an assert or something here for networking disabled?

hpx/runtime/actions/base_action.hpp
hkaiser: Didn't we?

hkaiser:
> Why does it still run the migrate_component test?

Why not? Attempting to migrate a component to the same locality is a valid operation (doesn't do anything) but we've had bugs for this case before, so I left it in.

hkaiser:
> All distributed tests are still compiled an run

Ok, I didn't consistently go over the tests, granted.

> a lot of stuff concerning action registration hasn't been touched yet.

Action registration is completely gone, I believe - at least the part needed for distributed operation.

hpx/runtime/actions/detail/action_factory.hpp
hkaiser: Right... I wasn't sure if this is still needed or not.

hpx/runtime/actions/base_action.hpp
sithhell: no, it is still completely there, it is not `#ifdef`-ed out completely

hkaiser: That's removed now.

hkaiser:
@sithhell I think all of your comments have been addressed now.

sithhell:
There's a build failure here: https://cdash.cscs.ch//viewBuildError.php?buildid=74625

components/containers/unordered/tests/unit/CMakeLists.txt
```diff
@@ -10,8 +10,13 @@ set(tests
 
 set(unordered_map_FLAGS
     COMPONENT_DEPENDENCIES unordered)
+
+if(HPX_WITH_NETWORKING)
+  set(__localities LOCALITIES 2)
+endif()
```
sithhell: Wouldn't it be better to perform this check here
https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_AddTest.cmake#L15

maybe even not add the test at all if networking was disabled?

components/iostreams/tests/regressions/CMakeLists.txt
```diff
@@ -8,9 +8,12 @@ set(tests
     no_output_1173
    )
 
+
 set(lost_output_2236_FLAGS COMPONENT_DEPENDENCIES iostreams)
 
-set(no_output_1173_PARAMETERS LOCALITIES 2)
+if(HPX_WITH_NETWORKING)
+  set(no_output_1173_PARAMETERS LOCALITIES 2)
+endif()
```
sithhell: Same as above.

components/containers/unordered/tests/unit/CMakeLists.txt
hkaiser: > Wouldn't it be better to perform this check here
https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_AddTest.cmake#L15

Good thinking!

> maybe even not add the test at all if networking was disabled?

I think that all of these tests run perfectly well on one locality, just with reduced test coverage. We might have to remove some of them, but I would like to be selective about this.

hkaiser: > Wouldn't it be better to perform this check here
https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_AddTest.cmake#L15

Actually, this is already in place: https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_AddTest.cmake#L55-L59

PR #4080, opened 2019-09-06
Updating module dependencies

hkaiser:
Updating module dependencies from output generated by cpp-dependencies


msimberg:
Thanks @hkaiser! Did you do this manually or could we maybe automate this?

hkaiser:
> Did you do this manually or could we maybe automate this?

This was a manual process. I was too lazy to think about automating it ;-)

hkaiser:
This looks like it's good to go now.

PR #4081, opened 2019-09-07
Adding option allowing to force using Boost.FileSystem

hkaiser:
- this is useful in situations where other parts of an application rely on Boost.FileSystem

### DIFF
--- a/libs/filesystem/CMakeLists.txt
+++ b/libs/filesystem/CMakeLists.txt
@@ -5,17 +5,40 @@
 
 cmake_minimum_required(VERSION 3.3.2 FATAL_ERROR)
 
+set(__filesystem_compatibility_default ON)
+if(HPX_WITH_CXX17_FILESYSTEM)
+  set(__filesystem_compatibility_default OFF)
+endif()
+
+# Compatibility with using Boost.FileSystem, introduced in V1.4.0
+hpx_option(HPX_FILESYSTEM_WITH_BOOST_FILESYSTEM_COMPATIBILITY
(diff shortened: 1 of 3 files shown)


msimberg:
@hkaiser would you mind explaining the motivation behind this? Shouldn't a user set up their Boost filesystem dependency themselves if they depend on it directly? Or do we actually expose std/boost filesystem somewhere in our ("public") API.

hkaiser:
@msimberg We do not expose the filesystem in our documented API, AFAIK. However, sneaky as I am, I'm using HPX internals in other projects (like Phylanx) and there we have seen compatibility problems.

msimberg:
All right... Do you have an example of the problems you've had? Is it something you could eventually resolve in Phylanx itself?

hkaiser:
> All right... Do you have an example of the problems you've had? Is it something you could eventually resolve in Phylanx itself?

I'll see what I can do.

hkaiser:
@msimberg: the issue we were running into (now I do remember) was that if HPX (silently) found the `std::filesystem` (while compiling in C++17, which is the default nowadays) then this forced our dependent libraries into using C++17 as well,  but in a non-obvious way (mostly again because of the defaults we have there, which in that case was using C++14). In the end the users were suddenly facing compilation problems where there were no problems before. This is while we do not use the filesystem types in our APIs we still (indirectly) #include those files from our exposed headers.

Having a way to explicitly control which filesystem library to use would be helpful.

msimberg:
Fair point! Explicit is better. Thanks!

PR #4083, opened 2019-09-10
Adapting thread_queue_mc to behave even if no 128bit atomics are available

hkaiser:
@biddisco please verify that this does not break anything on your end.

### DIFF
--- a/hpx/runtime/threads/policies/shared_priority_queue_scheduler.hpp
+++ b/hpx/runtime/threads/policies/shared_priority_queue_scheduler.hpp
@@ -68,10 +68,10 @@ namespace hpx { namespace threads { namespace policies {
     class shared_priority_queue_scheduler : public scheduler_base
     {
     public:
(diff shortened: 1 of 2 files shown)


biddisco:
remove the template defaults will not break anything

hkaiser:
> removing the template defaults will not break anything

Thanks @biddisco! I was more worried about (arbitrarily) choosing `PendingQueuing` as the third template argument for the `thread_queue_mc` here: https://github.com/STEllAR-GROUP/hpx/pull/4083/files#diff-6cd28becc08171d26d3f09c07c085302R73. Is that what you had in mind?

biddisco:
I wouldn't worry too much. the latest changes (incoming) to thread_queue_mc have substantially hanged the class and I'll absorb any of your changes into it when I do my PR

hkaiser:
@rtohid: this should fix your issues you're having with building FLeCSI.

PR #4086, opened 2019-09-10
Add coloured output support for building with ninja

biddisco:
gcc and clang do not output coloured text when running in
a non interactive terminal. Ninja intercepts output and appears
as one of these terminals. This patch adds a force colored flag
to the build when ninja is used (or enabled manually)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1175,6 +1175,25 @@ if(HPX_WITH_CXX14_RETURN_TYPE_DEDUCTION)
 endif()
 
 ################################################################################
+# Colored output (when using ninja)
+################################################################################
+if (CMAKE_MAKE_PROGRAM MATCHES ninja)
+  hpx_option(CMAKE_FORCE_COLORED_OUTPUT BOOL
+      "Always produce ANSI-colored output (GNU/Clang only)." TRUE ADVANCED)
+else ()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1174,6 +1174,25 @@ if(HPX_WITH_CXX14_RETURN_TYPE_DEDUCTION)
...
+  hpx_option(CMAKE_FORCE_COLORED_OUTPUT BOOL
+      "Always produce ANSI-colored output (GNU/Clang only)." TRUE ADVANCED)
+else ()
+  hpx_option(CMAKE_FORCE_COLORED_OUTPUT BOOL
+    "Always produce ANSI-colored output (GNU/Clang only)." FALSE ADVANCED)
+endif ()
+
+if (${CMAKE_FORCE_COLORED_OUTPUT})
+  if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    add_compile_options (-fdiagnostics-color=always)
```
msimberg: Can you change this to `hpx_add_compile_option` (I think)? We wouldn't want this to leak this to someone who uses the `hpx` target (once that's in place).

msimberg:
@biddisco do you still want this merged?

msimberg:
@biddisco please reopen if you're still interested in having this in. I'm of the opinion that this is something best handled by the user setting those flags themselves.

PR #4087, opened 2019-09-10
Making hpx::util::any compatible with C++17

hkaiser:
Fixes #4082


hkaiser:
@msimberg this is good to go now.

PR #4088, opened 2019-09-11
Construct lock object when it is needed in select_active_pu

biddisco:
minor optimization

### DIFF
--- a/examples/resource_partitioner/shared_priority_queue_scheduler.hpp
+++ b/examples/resource_partitioner/shared_priority_queue_scheduler.hpp
@@ -758,8 +758,6 @@ namespace hpx { namespace threads { namespace policies { namespace example {
                 {"HINT_NONE" COMMA "HINT....." COMMA "ERROR...." COMMA  "NORMAL..."});
             LOG_CUSTOM_VAR(const char *msg = nullptr);
 
-            std::unique_lock<pu_mutex_type> l;
-
             using threads::thread_schedule_hint_mode;
 
             switch (data.schedulehint.mode) {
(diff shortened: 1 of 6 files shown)


examples/resource_partitioner/shared_priority_queue_scheduler.hpp
```diff
@@ -758,8 +758,6 @@ namespace hpx { namespace threads { namespace policies { namespace example {
                 {"HINT_NONE" COMMA "HINT....." COMMA "ERROR...." COMMA  "NORMAL..."});
             LOG_CUSTOM_VAR(const char *msg = nullptr);
 
-            std::unique_lock<pu_mutex_type> l;
```
msimberg: I did put the lock here on purpose to make sure that whatever PU is chosen is not suspended before the work actually gets scheduled on it. Note that the lock is default constructed. No mutex is involved yet at this point. Did you measure a speedup from doing this?

biddisco: The lock is default constructed. All that ahppens is some internal vars are filled and nothing happens. It serves no purpose here until the lock is actually taken - so there's no need to construct it here. Since the lock is never acquired outside of the `select_active_pu` function, I just moved the lock construction inside the function. Just a tidying p operation.

msimberg: *When* it's actually locked the lock will be held to the end of this function, past the `create_thread` call at the end. The scope of the lock is different when it's constructed here as opposed to inside `select_active_pu`.

biddisco: ok. good point. I'll rethink this.

PR #4089, opened 2019-09-11
Make counters in coroutines optional

msimberg:
Makes the counters optional and turns them off by default. Saves us some cache misses when creating and rebinding threads. It makes a small but noticeable difference in the `future_overheads` benchmark.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -340,7 +340,7 @@ if("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
 endif()
 
 hpx_option(HPX_WITH_IO_COUNTERS BOOL
-  "Build HPX runtime (default: ${HPX_WITH_IO_COUNTERS_DEFAULT})"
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -602,6 +602,13 @@ if(HPX_WITH_THREAD_STEALING_COUNTS)
   hpx_add_config_define(HPX_HAVE_THREAD_STEALING_COUNTS)
 endif()
 
+hpx_option(HPX_WITH_COROUTINE_COUNTS BOOL
```
hkaiser: nitpick: should we name this `..._COUNTERS` instead?

biddisco: yes

msimberg: Counters was my first choice but there was prior art with counts... There are actually examples of both. `HPX_WITH_IO_COUNTERS`, `HPX_WITH_THREAD_STEALING_COUNTS`. I'm absolutely okay with changing it to counters.

msimberg: I changed the name to `HPX_WITH_COROUTINE_COUNTERS`.

PR #4090, opened 2019-09-11
Execution context

sithhell:
## Introducing execution_context for thread_data
    
    This patch replaces the use of the coroutine_self TLS pointer
    by using the newly introduced execution_context for yield, suspend
    and resume operations. What's missing is yield_to.
    
    This has the side effect of not needing to check whether we are
    running in an HPX thread when using synchronization primitives.

##  Implementing basic execution_context
    
    This patch implements basic functionality for a execution_context to be
    used for abstracting different threading implementations. The default is
    to rely on std facilities, that is OS threads.
    
    Flyby: Moving lock registration to execution in order to enable lock detection
    when yielding/suspending from generic contexts


libs/CMakeLists.txt
```diff
@@ -88,7 +89,7 @@ foreach(lib ${HPX_LIBS})
   add_subdirectory(${lib})
 
   set(MODULE_FORCE_LINKING_INCLUDES
-    "${MODULE_FORCE_LINKING_INCLUDES}#include <hpx/${lib}/force_linking.hpp>\n")
+    "${MODULE_FORCE_LINKING_INCLUDES}\n#include <hpx/${lib}/force_linking.hpp>\n")
```
hkaiser: This is most likely unintentional

libs/execution/include/hpx/execution/this_thread.hpp
```diff
@@ -0,0 +1,53 @@
...
+        detail::execution_context_storage* storage_;
+        execution_context_base* old_;
+    };
+
+    hpx::execution::execution_context execution_context();
+
+    void yield(const char* desc = "hpx::execution::this_thread::yield");
+    void yield_k(std::size_t k,
+        const char* desc = "hpx::execution::this_thread::yield_k");
+    void suspend(const char* desc = "hpx::execution::this_thread::suspend");
```
hkaiser: Will we have to `HPX_EXPORT` these functions?

src/lcos/local/detail/condition_variable.cpp
```diff
@@ -237,45 +229,33 @@ namespace hpx { namespace lcos { namespace local { namespace detail
...
-                                threads::get_thread_state(id))
-                        << "(" << id << "): "
-                        << threads::get_thread_description(id);
+                //                 LERR_(fatal)
+                //                         << "condition_variable::abort_all:"
+                //                         << " pending thread: "
+                //                         << get_thread_state_name(
+                //                                 threads::get_thread_state(id))
+                //                         << "(" << id << "): "
+                //                         << threads::get_thread_description(id);
```
hkaiser: Is this intentional?

hkaiser:
> We decided to move the parallel executors to a module named execution. Is this planned to be integrated there?

To answer my own question: I don't think these should be merged. The parallel executors stuff belongs into `execution` as the related namespace is named the same way. 

Should we call this new module `context` instead? Any better ideas?

sithhell:
> We decided to move the parallel executors to a module named execution. Is
> this planned to be integrated there?
>
> To answer my own question: I don't think these should be merged. The
> parallel executors stuff belongs into execution as the related namespace
> is named the same way.
>
> Should we call this new module context instead? Any better ideas?
>
I personally would merge the two. In fact, with the help of this
execution_context, the execution module could contain:
 - synchronization primitives (mutex, spinlock, condition_variable and
friends): those need the context to yield, suspend and resume
 - future: those need the sync primitives
 - executors, at least defining the least minimal concepts:
execution_context should expose an executor member function, to be able to
spawn new execution contexts. Executors need futures.

That's why I believe those three functionalities belong in the same module.

The namespace execution should not be limited to this concept however, as
the HPX thread specific executors need to depend on our coroutine
implementation, and this should, IMHO, be built on top of the execution
module. This creates a logical structure to reuse the HPX concepts with
ease with other thread pool implementations. I hope this makes sense

>

sithhell:
FWIW, the execution context is a concept that has been originally introduced by earlier versions of p0443. This is a slight variation of it, merging ece

sithhell:
While thinking about it once more, I think it would be split and the changed to context and agent.

hkaiser:
@sithhell Why have you closed this?

sithhell:
> @sithhell Why have you closed this?

@hkaiser this was by accident...

libs/execution/include/hpx/execution/this_thread.hpp
sithhell: Right. All symbols have been exported properly now.

src/lcos/local/detail/condition_variable.cpp
sithhell: No, i re-enabled logging again...
This indentation was a result of clang-format.

msimberg:
I haven't looked through this in detail yet, but shouldn't this make it *easier* for us to have lcos, futures, executors, etc. in different modules? In any case, why not just call this `execution_context` as it's the main concept defined in this module?

hkaiser:
We've had some discussion about this on IRC yesterday. The thing introduced by this PR should be called `agent_proxy` (`agent_view`, `agent_ref` - I like the first best) as it does not own the execution agent it is referring to. It exposes a helper API that most likely will not be seen by any user but that is needed to abstract the suspension mechnisms for various execution agents (`std::thread`, `hpx::thread`) .

hpx/lcos/local/recursive_mutex.hpp
```diff
@@ -35,7 +35,7 @@ namespace hpx { namespace lcos { namespace local
 
         private:
             std::atomic<std::uint64_t> recursion_count;
-            std::atomic<hpx::execution::agent> locking_context;
+            std::atomic<hpx::execution::agent_ref> locking_context;
```
hkaiser: Is it ok to use a non-POD type with `std::atomic`?

sithhell:
> I haven't looked through this in detail yet, but shouldn't this make it
> *easier* for us to have lcos, futures, executors, etc. in different
> modules?
>
Depends...

> In any case, why not just call this execution_context as it's the main
> concept defined in this module?
>
... mainly on where we want to go in the long run. If we want to go some
route along P0443, it doesn't make sense to rename it. It should provide
and define the basic interface needed to implement the above mentioned.
However, some basic infrastructure classes need to be included in this
module to avoid circular dependencies between modules.
Even if we go a different route. Lcos, futures and executors will always
depend on each other in some way. If not by including specific
implementations, at the very least, by defining the concepts, generic APIs
to use the concepts and the traits.

However, as soon as the context returns an executor, there needs to be an
implementation of a future inside this module as the future needs to know
about executors (at least by accessing the traits).

A specific thread pool implementation can be used by just depending on the
module, as demonstrated by this PR.

I don't see a problem that other modules reuse the execution namespace, for
example.

>

sithhell:
> We've had some discussion about this on IRC yesterday. The thing
> introduced by this PR should be called agent_proxy (agent_view, agent_ref
> - I like the first best) as it does not own the execution agent it is
> referring to. It exposes a helper API that most likely will not be seen by
> any user but that is needed to abstract the suspension mechnisms for
> various execution agents (std::thread, hpx::thread) .
>
I went for `agent_ref`, goes along with things like `function_ref`.

>

hkaiser:
@msimberg this is good to go, AFAICS.

libs/basic_execution/include/hpx/basic_execution/agent_ref.hpp
```diff
@@ -0,0 +1,104 @@
...
+            std::chrono::time_point<Clock, Duration> const& sleep_time,
+            char const* desc = "hpx::basic_execution::agent_ref::sleep_until")
+        {
+            sleep_until(hpx::util::steady_time_point{sleep_time}, desc);
+        }
+
+        // TODO:
+        // affinity
+        // thread_num
+        // executor
```
msimberg: Do we leave this for another PR?

libs/basic_execution/include/hpx/basic_execution/resource_base.hpp
```diff
@@ -0,0 +1,21 @@
...
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+////////////////////////////////////////////////////////////////////////////////
+
+#ifndef HPX_EXECUTION_RESOURCE_BASE_HPP
+#define HPX_EXECUTION_RESOURCE_BASE_HPP
+
+namespace hpx { namespace basic_execution {
+
+    /// TODO: implement, this is currently just a dummy
+    struct resource_base
```
msimberg: For another PR?

libs/config/include/hpx/config.hpp
```diff
@@ -209,7 +209,7 @@
 #  endif
 #endif
 #if !defined(HPX_SPINLOCK_DEADLOCK_DETECTION_LIMIT)
-#  define HPX_SPINLOCK_DEADLOCK_DETECTION_LIMIT 1000000
+#  define HPX_SPINLOCK_DEADLOCK_DETECTION_LIMIT 4294967294
```
msimberg: That's around an hour of 1000 ns sleeps. The old value is a second. Maybe we should have something in between?

msimberg:
I pushed a couple of small changes (docs formatting, use generated global header, remove thousands separator for C++11). Were the TODOs intended to be left for another PR?

libs/config/include/hpx/config.hpp
hkaiser: Yah, this doesn't make sense. @sithhell is this a leftover from testing?

libs/basic_execution/include/hpx/basic_execution/agent_ref.hpp
hkaiser: Yes, let's leave this. Nobody knows how this should look like.

libs/basic_execution/include/hpx/basic_execution/resource_base.hpp
hkaiser: Same here. Let's leave this for now.

sithhell:
TODOs were left for future PRs, yes

libs/config/include/hpx/config.hpp
sithhell: the value could be smaller, probably. The old value lead to the detection kicking in though. Note that we don't sleep for 1000 nsecs for every k (only every second iteration when k is large). In any case, this probably needs some adaption.

hkaiser:
@sithhell I rebased this one more time to account for the stackless changes on master. Could you please have a look?

libs/basic_execution/include/hpx/basic_execution/register_locks.hpp
```diff
@@ -17,6 +17,7 @@
 
 ///////////////////////////////////////////////////////////////////////////////
 namespace hpx { namespace util {
+
```
msimberg: Last question, what was the rationale for moving `register_locks` into this module? Fewer dependencies?

sithhell: The rationale is that the lock detection is a property of the execution agent, IMHO, as such it makes sense to have that here.

sithhell:
@hkaiser looks good. I also decreased the deadlock detection limit by a factor of 4.

hpx/lcos/local/recursive_mutex.hpp
sithhell: Yes, that's fine, `agent_ref` should provide the necessary properties as described here: https://en.cppreference.com/w/cpp/atomic/atomic (under Primary template)

hkaiser:
>  I also decreased the deadlock detection limit by a factor of 4.

Thanks @sithhell, what's the point of deadlock detection if it doesn't run, after all...

msimberg:
This looks new: https://circleci.com/gh/STEllAR-GROUP/hpx/137937.

hkaiser:
> This looks new: https://circleci.com/gh/STEllAR-GROUP/hpx/137937.

Yes, I've seen that - working on it...

hkaiser:

PR #4091, opened 2019-09-11
Split runtime into local and distributed parts

msimberg:
Splits the runtime into a local class and a distributed class which inherits the local one. Current non-features of the local runtime that will be dealt with when needed:

- logging
- performance counters

This PR merges `runtime_impl` into `runtime` and then moves the distributed features into `runtime_distributed` inheriting from `runtime`. One can then access `runtime` or `runtime_distributed` depending on what features one needs, but `runtime_distributed` is not guaranteed to be initialized. Naming: is it clearer to have `runtime_local` and `runtime_distributed`? Or `runtime_local` and `runtime`?

Outside of the runtime implementation there is now a CMake option `HPX_WITH_DISTRIBUTED_RUNTIME` which is on by default. When this is off a number of source files are not compiled, a number of tests and examples are disabled, and many includes are excluded. Components, parcelports, logging, performance counters, plugins are all disabled when the distributed runtime is disabled.

To do:
- [x] format runtime files
- [x] remove unnecessary includes
- [x] change pycicle to test local runtime
- [x] squash/clean up history

To do in a separate PR:
- move the local and distributed runtime into separate modules


msimberg:
This is starting to take shape (meaning I'm back to a working state where tests pass). The usual distributed runtime is still the default and only option. I think having it as a runtime option is less valuable since the difference in startup times is negligible. I will instead try a compile-time option which completely turns off components and actions and see how much of a difference it makes.

msimberg:
I've added an experimental option to only build the local runtime (no plugins, components, actions, performance counters etc.). I get the following timings when building `hello_world_` with a single job:

- local runtime: 8 minutes
- networking off: 17 minutes
- networking on: 21 minutes

Max memory usage is only slightly improved (around 1 GB with the local runtime, 1.3 GB with networking on and off).

CMakeLists.txt
```diff
@@ -753,54 +753,66 @@ if(HPX_WITH_AGAS_DUMP_REFCNT_ENTRIES)
...
+hpx_option(HPX_WITH_EXPERIMENTAL_LOCAL_RUNTIME BOOL
+  "Enable the experimental local runtime (this disables all distributed features of HPX, including components and actions) (default: OFF)"
+  OFF CATEGORY "Build Targets" ADVANCED)
+
+if(HPX_WITH_LOGGING AND HPX_WITH_EXPERIMENTAL_LOCAL_RUNTIME)
+  hpx_error("HPX_WITH_EXPERIMENTAL_LOCAL_RUNTIME=ON requires HPX_WITH_LOGGING=OFF")
+endif()
+
+if(NOT HPX_WITH_EXPERIMENTAL_LOCAL_RUNTIME)
+  hpx_add_config_define(HPX_HAVE_DISTRIBUTED_RUNTIME)
```
hkaiser: May I suggest to use consistent names here? Either use `HPX_WITH_DISTRIBUTED_RUNTIME` or `HPX_HAVE_EXPERIMENTAL_LOCAL_RUNTIME`. I like the first one better, FWIW.

hpx/runtime.hpp
```diff
@@ -314,22 +412,50 @@ namespace hpx
         notification_policy_type::on_error_type on_error_func(
             notification_policy_type::on_error_type&&);
 
+        virtual std::uint32_t get_locality_id(error_code& ec)
```
hkaiser: Could this member function be `const`?

hpx/runtime.hpp
```diff
@@ -314,22 +412,50 @@ namespace hpx
...
+        {
+            return 0;
+        }
+
+        virtual std::size_t get_num_worker_threads()
```
hkaiser: Same here (and below): could this be `const`?

hpx/runtime/runtime_fwd.hpp
```diff
@@ -15,15 +15,16 @@
 namespace hpx
 {
     class HPX_API_EXPORT runtime;
-
-    ///////////////////////////////////////////////////////////////////////////
-    class HPX_API_EXPORT runtime_impl;
+    class HPX_API_EXPORT runtime_distributed;
```
hkaiser: This probably needs to be declared only if `HPX_HAVE_DISTRIBUTED_RUNTIME` is set.

hpx/runtime/runtime_fwd.hpp
```diff
@@ -15,15 +15,16 @@
...
 
     /// The function \a get_runtime returns a reference to the (thread
     /// specific) runtime instance.
     HPX_API_EXPORT runtime& get_runtime();
     HPX_API_EXPORT runtime*& get_runtime_ptr();
 
+    HPX_API_EXPORT runtime_distributed& get_runtime_distributed();
+    HPX_API_EXPORT runtime_distributed*& get_runtime_distributed_ptr();
```
hkaiser: Same here, do we always need this functions to be defined?

libs/CMakeLists.txt
```diff
@@ -49,6 +46,15 @@ set(HPX_LIBS
...
 )
 
+if(NOT HPX_WITH_EXPERIMENTAL_LOCAL_RUNTIME)
+  set(HPX_LIBS ${HPX_LIBS}
+    collectives
+    resiliency
+    segmented_algorithms
+    CACHE INTERNAL "list of HPX modules" FORCE
+  )
+endif()
```
hkaiser: I think resiliency does not rely on distributed capabilities. Everything is purely local.

src/CMakeLists.txt
```diff
@@ -93,57 +227,609 @@ if(MSVC)
...
-add_hpx_library_headers(hpx
-  GLOB_RECURSE GLOBS "${PROJECT_SOURCE_DIR}/hpx/compat/*.hpp"
-  APPEND)
-
-# libhpx_init headers
-add_hpx_library_headers(hpx_init
-  GLOB GLOBS "${PROJECT_SOURCE_DIR}/hpx/hpx_main*.hpp")
-
-add_hpx_library_headers(hpx_external GLOB APPEND)
+if(MSVC)
```
hkaiser: Why `MSVC` only?

src/hpx_init.cpp
```diff
@@ -337,143 +348,182 @@ namespace hpx
...
-                else {
-                    std::string msg ("Invalid command line option value"
-                        "for --hpx:list-counters: ");
-                    msg += option;
-                    msg += ", allowed values are 'minimal' and 'full'";
-                    throw detail::command_line_error(msg.c_str());
+#if defined(HPX_HAVE_DISTRIBUTED_RUNTIME)
+            runtime_distributed* rtd =
+                dynamic_cast<hpx::runtime_distributed*>(&rt);
+            if (rtd != nullptr)
```
hkaiser: I think an `assert` would be fine here. If the distributed runtime is compiled in, this has to succeed.

src/hpx_init.cpp
```diff
@@ -819,16 +886,20 @@ namespace hpx
             std::terminate();
         }
 
+#if defined(HPX_HAVE_DISTRIBUTED_RUNTIME)
         components::server::runtime_support* p =
             reinterpret_cast<components::server::runtime_support*>(
-                  get_runtime().get_runtime_support_lva());
+                  get_runtime_distributed().get_runtime_support_lva());
```
hkaiser: Couldn't we turn `get_runtime_support_lva` into a `virtual` function instead?

hkaiser: No... I see it now - please disregard this comment.

CMakeLists.txt
msimberg: Yeah, probably a good idea. I wanted to have `EXPERIMENTAL` in the CMake option, but I didn't want to have `HPX_WITH_EXPERIMENTAL_DISTRIBUTED_RUNTIME` because that's a lie. Perhaps I'll just make it clear in the description of the option that turning it off is experimental.

hpx/runtime/runtime_fwd.hpp
msimberg: No. I think this was a result of me planning to make this a runtime option, in which case the distributed runtime might not be available. As a compile time option only this doesn't make sense anymore.

libs/CMakeLists.txt
msimberg: Makes sense. Nice!

hpx/runtime.hpp
msimberg: Yep!

hpx/runtime.hpp
msimberg: I'm pretty sure it can. Thanks!

src/CMakeLists.txt
msimberg: Because it's only relevant for `MSVC`. But that's ignored in `add_hpx_source_group` so it doesn't make sense to duplicate that check here. Thanks for noticing! `hpx_SOURCES` below should also be `hpx_HEADERS`.

msimberg:
Thanks @hkaiser for looking through this. I know this is probably a bit of a paint to look through. If there's anything I can do to make it more manageable to review for you, let me know.

I'll take next week to rebase this on current master, but I definitely want to wait with this until after the release. Too many corner cases that could go wrong...

CMakeLists.txt
msimberg: I updated it to be `HPX_WITH_DISTRIBUTED_RUNTIME`.

msimberg:
I just fixed what I hope was the last bug on this branch, and discovered another one that I don't know how to fix.

The `init_globally` example was failing with `pure virtual method called` during shutdown, from calling `default_agent::yield`. This was 1) due to TSS initialization happening later on this branch leading to networking being enabled and parcelports being initialized, and 2) due to this meaning that during shutdown parcelports get cleaned up and they call `yield`. The problem is that since the runtime was started in a global object, and the thread local agent storage is destructed before the global runtime management object is destructed, one can't call `yield` anymore.

I've "fixed" it by restoring the TSS initialization to an earlier stage, but the actual bug still remains, and would likely always manifest itself in distributed runs with the global object. This is the line that calls `yield`: https://github.com/STEllAR-GROUP/hpx/blob/b819125e44949554de09755e40c66d512a6187bd/hpx/runtime/parcelset/parcelport_impl.hpp#L176.

I'd welcome reviews on this now. For convenience of rebasing I've squashed everything into one commit. I know this can make reviewing a bit more challenging, so I will happily untangle the commits if wished. Please let me know something is unclear about the changes. The main files to look at are `{hpx/src}/runtime*`.

examples/quickstart/file_serialization.cpp
```diff
@@ -10,6 +10,7 @@
 #include <hpx/hpx.hpp>
 #include <hpx/hpx_main.hpp>
 #include <hpx/serialization/traits/serialization_access_data.hpp>
```
aurianer: I think as we include the global generated header we can remove the `hpx/serialization/traits/serialization_access_data.hpp` one

CMakeLists.txt
```diff
@@ -1792,7 +1815,9 @@ add_subdirectory(init)
 ################################################################################
 # Add components
 ################################################################################
-add_subdirectory(components)
+if(HPX_WITH_DISTRIBUTED_RUNTIME)
+  add_subdirectory(components)
```
sithhell, line 1819: Why can't we have components in a non distributed world?

CMakeLists.txt
```diff
@@ -1759,9 +1780,11 @@ add_hpx_pseudo_target(core)
 ################################################################################
 # Configure compression and other plugins
 ################################################################################
-set(HPX_STATIC_PARCELPORT_PLUGINS "" CACHE INTERNAL "" FORCE)
-add_hpx_pseudo_target(plugins)
-add_subdirectory(plugins)
+if(HPX_WITH_DISTRIBUTED_RUNTIME)
+  set(HPX_STATIC_PARCELPORT_PLUGINS "" CACHE INTERNAL "" FORCE)
+  add_hpx_pseudo_target(plugins)
+  add_subdirectory(plugins)
```
sithhell, line 1786: Why can't we have plugins in a non distributed world?

CMakeLists.txt
msimberg: Because that's what classes are for. `HPX_WITH_NETWORKING=OFF` gives you components without turning networking if that's what you're looking for.

CMakeLists.txt
msimberg: That's a fair point. I haven't actually disabled the plugin functionality. Disabling the plugins at this level was more for simplicity as none of the plugins we have at the moment make much sense with a local-only runtime. I can go and disable them individually instead.

CMakeLists.txt
msimberg: And to clarify even further, there are actually components/plugins which would be useful also locally: the counters components. But since they still at the moment require components and actions they will stay disabled.

hpx/apply.hpp
```diff
@@ -33,6 +36,10 @@ namespace hpx { namespace detail
     // Define apply() overloads for plain local functions and function objects.
     // dispatching trait for hpx::apply
 
+#if !defined(HPX_HAVE_DISTRIBUTED_RUNTIME)
+    template <typename Func, typename Enable = void>
+    struct apply_dispatch;
+#endif
```
hkaiser, line 42: Where is this template declared if `HPX_HAVE_DISTRIBUTED_RUNTIME` is actually defined? Should we really separate that?

hpx/async_launch_policy_dispatch.hpp
```diff
@@ -66,7 +70,12 @@ namespace hpx { namespace detail
     ///////////////////////////////////////////////////////////////////////////
     template <typename Action>
     struct async_launch_policy_dispatch<Action,
-        typename std::enable_if<!traits::is_action<Action>::value>::type>
+#if defined(HPX_HAVE_DISTRIBUTED_RUNTIME)
+        typename std::enable_if<!traits::is_action<Action>::value>::type
+#else
+        void
+#endif
+        >
```
hkaiser, line 78: Why is this specialization (for Actions) enabled if `HPX_HAVE_DISTRIBUTED_RUNTIME` is off?

hpx/lcos_fwd.hpp
```diff
@@ -85,7 +87,11 @@ namespace hpx
 
     using lcos::future;
     using lcos::shared_future;
+#if defined(HPX_HAVE_DISTRIBUTED_RUNTIME)
     using lcos::promise;
+#else
+    using lcos::local::promise;
+#endif
```
hkaiser, line 94: This is very dangerous as it (silently) changes semantics of user code depending on whether `HPX_HAVE_DISTRIBUTED_RUNTIME` is defined or not.

hpx/runtime_distributed.hpp
```diff
@@ -123,6 +113,18 @@ namespace hpx
...
+#if !defined(HPX_COMPUTE_DEVICE_CODE)
+            //   tell main locality to start application exit, duplicated requests
+            // will be ignored
+            apply<components::server::runtime_support::shutdown_all_action>(
+                hpx::find_root_locality(), shutdown_timeout);
+#endif
+
+            return 0;
+        }
+
```
hkaiser, line 127: I think this is safe to be defined in a .cpp file.

hpx/runtime_distributed.hpp
```diff
@@ -346,45 +393,98 @@ namespace hpx
...
+            components::component_type type, error_code& ec) const
+        {
+            return agas_client_.get_num_localities(type, ec);
+        }
+
+        lcos::future<std::uint32_t> get_num_localities(
+            components::component_type type) const
+        {
+            return agas_client_.get_num_localities_async(type);
+        }
```
hkaiser, line 434: Moving all of those implementations into a .cpp file might reduce compile times and direct dependencies.

hpx/apply.hpp
msimberg: No, it doesn't make sense. Especially after #4478 the base template will be in `async_base` and always declared. I'll change this once that's in master and this is rebased.

hpx/async_launch_policy_dispatch.hpp
msimberg: Same thing applies as in https://github.com/STEllAR-GROUP/hpx/pull/4091#discussion_r403913899. Thanks!

hpx/lcos_fwd.hpp
msimberg: Good point, will change.

msimberg:
> Should we disable all command line options that are related to performance counters (i.e. remove them from view, here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/command_line_handling/src/parse_command_line.cpp#L584-L641)?

Yes, it might be a good idea. I'm not sure yet if silently ignoring them or giving an error would be better. I'll look into keeping the options, but giving an error if they're specified on the command line.

sithhell:
@msimberg thanks for your answers. I understand that currently it is mostly about reducing compile times to get more and more people use the local only parts of HPX. I understand that this is a workaround to that problem, which eventually will be solved in a much cleaner way once modularization is completed

hkaiser:
>> Should we disable all command line options that are related to performance counters (i.e. remove them from view, here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/command_line_handling/src/parse_command_line.cpp#L584-L641)?
> 
> Yes, it might be a good idea. I'm not sure yet if silently ignoring them or giving an error would be better. I'll look into keeping the options, but giving an error if they're specified on the command line.

Simply removing them in the first place would be one option. The command line handling would then throw an error if somebody tries to use those in a purely local scenario. Touching up on every of those options individually would allow for a better error message however. It's a QoI issue and I'd leave it to you to decide which way to go.

biddisco:
I did not read all the comments in this thread, but ...

It's not just about compile times. Its also about the sheer scale of the baggage that must be waded through when working with HPX code. When a new user asks questions about component this and component that and then you discover they're not even working in distributed mode, you realize they've completely misunderstood how it all fits together. Then when they have a bug and you dig into the code to try to fix it, you realize you can't - because you have to modify not just the 3 files that matter - but another 20 that depend on it/them through some strange connection.

We have at multiple projects that have created their own thread pools, made their own simple task frameworks and their developers run away screaming when you show them hpx - not because it's badly written or terrible performance, but because of intellectual overhead and baggage - they don't want a Rolls-Royce to get them from A to B, they just want something that works and they can understand without investing half a lifetime into looking at the code. 

HPX is currently big and bloated with tons of code that isn't really ever used. It had it's 12th? birthday recently and only 35 people out there (according to the survey) are actually using it. At least 20 (for example) at CSCS won't touch it with a ten foot pole because of the bloat, no doubt there are countless others around the world.

Separating the local from the distributed is a necessary step in making the library palatable to more users. Additionally, I've worked through the parcelport code extensively - and I don't find it performs very well - so separating the distributed and local is a useful step towards cleaning up some of the API that interfaces the two layers.

msimberg:
I think I've addressed all the review comments now, and that the builds pass (let's see what I've missed...).

msimberg:
The tests seem to be as clean as they get with our current master. `gcc-newest` is the builder on daint which tests the local runtime (used to do `HPX_WITH_NETWORKING=OFF`, not sure if we need both). For the record, there are about 500 tests with the local runtime, and 800 with the distributed runtime. That's still quite a few tests with the local runtime, and it could be a bit higher if there were a local `iostreams`.

hkaiser:
> The tests seem to be as clean as they get with our current master. gcc-newest is the builder on daint which tests the local runtime (used to do HPX_WITH_NETWORKING=OFF, not sure if we need both). For the record, there are about 500 tests with the local runtime, and 800 with the distributed runtime. That's still quite a few tests with the local runtime, and it could be a bit higher if there were a local iostreams.

A simple `namespace hpx { using std::cout; }` should give us a local iostreams.

msimberg:
> > The tests seem to be as clean as they get with our current master. gcc-newest is the builder on daint which tests the local runtime (used to do HPX_WITH_NETWORKING=OFF, not sure if we need both). For the record, there are about 500 tests with the local runtime, and 800 with the distributed runtime. That's still quite a few tests with the local runtime, and it could be a bit higher if there were a local iostreams.
> 
> A simple `namespace hpx { using std::cout; }` should give us a local iostreams.

Agreed. I'd still like to give it a go to use the buffering and IO pools of HPX's current `iostreams` without components, but your solution practically glows with simplicity. It'd definitely do as a temporary solution. On the other hand most uses of `hpx::cout` in our tests and examples don't benefit much from the buffering when run locally, and one can still use the IO pools explicitly if there's heavier IO.

hkaiser:
>>> The tests seem to be as clean as they get with our current master. gcc-newest is the builder on daint which tests the local runtime (used to do HPX_WITH_NETWORKING=OFF, not sure if we need both). For the record, there are about 500 tests with the local runtime, and 800 with the distributed runtime. That's still quite a few tests with the local runtime, and it could be a bit higher if there were a local iostreams.
>> 
>> A simple namespace hpx { using std::cout; } should give us a local iostreams.
> 
> Agreed. I'd still like to give it a go to use the buffering and IO pools of HPX's current iostreams without components, but your solution practically glows with simplicity. It'd definitely do as a temporary solution. On the other hand most uses of hpx::cout in our tests and examples don't benefit much from the buffering when run locally, and one can still use the IO pools explicitly if there's heavier IO.

I understand. Dispatching the use of `std::cout` to one of the IO threads requires a bit more work. This wouldn't be necessary for the tests (no performance requirements), but for users of `hpx::cout` the solution I suggested above would possibly have surprising side effects.

msimberg:
> I understand. Dispatching the use of `std::cout` to one of the IO threads requires a bit more work. This wouldn't be necessary for the tests (no performance requirements), but for users of `hpx::cout` the solution I suggested above would possibly have surprising side effects.

In any case I would like to leave that out of this PR (especially since `namespace hpx { using std::cout; }` does change blocking semantics). Once this PR is in I'll collect next steps into one or more issues.

CMakeLists.txt
```diff
@@ -745,11 +750,31 @@ if(HPX_WITH_AGAS_DUMP_REFCNT_ENTRIES)
   hpx_add_config_define(HPX_HAVE_AGAS_DUMP_REFCNT_ENTRIES)
 endif()
 
+hpx_option(HPX_WITH_DISTRIBUTED_RUNTIME BOOL
+  "Enable the distributed runtime (default: ON). Turning off the distributed runtime completely disallows the creation and use of components and actions. Turning this option off is experimental!"
+  ON CATEGORY "Build Targets" ADVANCED)
```
hkaiser, line 755: Would the checks above properly pick up this option?

CMakeLists.txt
```diff
@@ -745,11 +750,31 @@ if(HPX_WITH_AGAS_DUMP_REFCNT_ENTRIES)
...
+  hpx_error("HPX_WITH_NETWORKING=ON requires HPX_WITH_DISTRIBUTED_RUNTIME=ON")
+endif()
+
+if(HPX_WITH_LOGGING AND NOT HPX_WITH_DISTRIBUTED_RUNTIME)
+  hpx_error("HPX_WITH_DISTRIBUTED_RUNTIME=OFF requires HPX_WITH_LOGGING=OFF")
+endif()
+
+if(HPX_WITH_APEX AND NOT HPX_WITH_DISTRIBUTED_RUNTIME)
+  hpx_error("HPX_WITH_DISTRIBUTED_RUNTIME=OFF requires HPX_WITH_APEX=OFF")
+endif()
```
hkaiser, line 776: Do we really have to disable APEX here? I think disabling the performance counter related API in the APEX/HPX interface should be sufficient.

msimberg: APEX does depend on actions: https://github.com/khuck/xpress-apex/blob/6bc59aeb18ae5badce1c42fe35c9a15dcc382b06/src/apex/profiler_listener.cpp#L1728-L1756. That might be one of the only places though and making it work without the distributed runtime is probably doable.

CMakeLists.txt
msimberg: Excellent catch! I've fixed that now.

CMakeLists.txt
hkaiser: That doesn't seem to actually need an action as it launches things on the current locality only. I think this could be changed easily - @khuck?

khuck: I'm confused...why does distributed require APEX to be disabled?  What's the context here?  Are actions deprecated?  What's the change required to just schedule the task locally?

hkaiser: Kevin, actions are not deprecated. We're trying to enable a minimal configuration of HPX that essentially is reduced to purely local operation. In this case, actions and components would be not available. The code in APEX does not actually require actions as everything is limited to local execution. Instead of using async/apply to run the function that is associated with the action you could simply use async/apply to execute the function itself:
```
void apex_schedule_process_profiles() {
    if(get_hpx_runtime_ptr() == nullptr) return;
    if(!thread_instance::is_worker()) return;
    if(hpx_shutdown) {
        APEX_TOP_LEVEL_PACKAGE::profiler_listener::process_profiles_wrapper();
    } else {
        if(!consumer_task_running.test_and_set(memory_order_acq_rel)) {
            try {
                hpx::apply(&apex_internal_process_profiles);
            } catch(...) {
                // During shutdown, we can't schedule a new task,
                // so we process profiles ourselves.
                profiler_listener::process_profiles_wrapper();
            }
        }
    }
}
```

msimberg: And just to be explicit, the corresponding header: https://github.com/khuck/xpress-apex/blob/6bc59aeb18ae5badce1c42fe35c9a15dcc382b06/src/apex/profiler_listener.cpp#L70, would need to be `#include <hpx/local_async.hpp>` after that. The performance counters in APEX would at the moment have to be fully disabled though.

msimberg: I think this also doesn't have to resolved before this is merged.

msimberg:
Just to be clear, I'm not planning to add anything more to this PR (except for cosmetic and bug fixes, if needed). All the extensions to bring more functionality to the local runtime will be done separately.

msimberg:
Are there any further comments on this?

msimberg:
I've started https://github.com/STEllAR-GROUP/hpx/projects/5 to keep track of issues related to the local runtime.

msimberg:
I'll merge this once CI passes if there are no more comments.

msimberg:
I only fixed an inspect failure for a missing include in the last force push. Merging now.

PR #4093, opened 2019-09-16
Fix initialization of global_num_tasks in parallel_executor

msimberg:
Fixes deadlocks typically seen in the `1d_stencil` example caused by yielding while initializing a static variable. Details in commit message.

Also initializes `global_num_tasks` in the `parallel_executor` based on the number of threads in the current pool (or default pool if not on an HPX thread).

### DIFF
--- a/libs/execution/include/hpx/parallel/executors/parallel_executor.hpp
+++ b/libs/execution/include/hpx/parallel/executors/parallel_executor.hpp
@@ -25,10 +25,14 @@
 #include <hpx/parallel/executors/fused_bulk_execute.hpp>
 #include <hpx/parallel/executors/post_policy_dispatch.hpp>
 #include <hpx/parallel/executors/static_chunk_size.hpp>
-#include <hpx/runtime/get_worker_thread_num.hpp>
 #include <hpx/runtime/launch_policy.hpp>
 #include <hpx/runtime/serialization/serialize.hpp>
+#include <hpx/runtime/threads/policies/scheduler_base.hpp>
+#include <hpx/runtime/threads/thread_data.hpp>
+#include <hpx/runtime/threads/thread_data_fwd.hpp>
 #include <hpx/runtime/threads/thread_helpers.hpp>
+#include <hpx/runtime/threads/thread_pool_base.hpp>
+#include <hpx/runtime/threads/threadmanager.hpp>
 #include <hpx/traits/future_traits.hpp>
 #include <hpx/traits/is_executor.hpp>
 #include <hpx/util/unwrap.hpp>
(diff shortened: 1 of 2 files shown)


libs/execution/include/hpx/parallel/executors/parallel_executor.hpp
```diff
@@ -172,12 +176,19 @@ namespace hpx { namespace parallel { namespace execution
...
+                auto pool = threads::get_self_ptr() ?
+                    threads::get_self_id()->get_scheduler_base()->get_parent_pool() :
+                    &hpx::threads::get_thread_manager().default_pool();
+                num_tasks =
+                    (std::min)(std::size_t(128), pool->get_os_thread_count());
+            }
+            else
+            {
+                num_tasks = num_tasks_;
+            }
```
hkaiser: I'd suggest to use a `static thread_local` instead. Otherwise this looks very racy (might be benign, though)!

msimberg: Not sure I follow. I wanted to completely avoid `static` variables, so these are local. A `static thread_local` in principle has the same problem as just `static`, although it should be *much* less likely to fail. Which part looks racy to you? `num_tasks_` could potentially be changed since it's public (it's a struct) but I would think that's usually only set in the constructor. We can also make that private.

hkaiser: Well, `static thread_local` wouldn't have the same problem. But I agree that getting rid of that static altogether is a Good Thing (tm). We could initialize things in the constructor now, however. 

>  We can also make that private.

Yes, that's a good idea.

msimberg: Updated to intialize the `num_tasks_` in the constructor (it was already private). Also clang-formatted the file, so the diff looks a bit bigger...

libs/execution/include/hpx/parallel/executors/parallel_executor.hpp
```diff
@@ -93,10 +95,23 @@ namespace hpx { namespace parallel { namespace execution
...
+            Policy l = detail::get_default_policy<Policy>::call(),
+            std::size_t spread = 4, std::size_t tasks = std::size_t(-1))
+          : policy_(l)
+          , num_spread_(spread)
+          , num_tasks_(tasks)
+        {
+            if (num_tasks_ == std::size_t(-1))
+            {
+                auto pool = threads::get_self_ptr() ?
+                    threads::get_self_id()
```
hkaiser: Isn't `get_self_ptr()` already returning the pointer you need? I think there is no need to call `get_self_id()` afterwards (it accesses a thread_local, so might be a bit more costly than expected). Also, older gcc will complain about a constexpr constructor not having an empty body. I think you could wrap the `if()` into a static function to call in the member initializer:
```
    , num_tasks_(init_num_tasks(tasks))
```
or similar to work around this.

msimberg: Yes, on the `get_self_ptr`. What I have is unnecessary. And I'll try removing the constexpr from the constructor.

PR #4099, opened 2019-09-19
Module coroutines

aurianer:
Coroutines module as suggested by #4075


libs/CMakeLists.txt
```diff
@@ -93,7 +94,7 @@ foreach(lib ${HPX_LIBS})
   add_subdirectory(${lib})
 
   set(MODULE_FORCE_LINKING_INCLUDES
-    "${MODULE_FORCE_LINKING_INCLUDES}#include <hpx/${lib}/force_linking.hpp>\n")
+    "${MODULE_FORCE_LINKING_INCLUDES}\n#include <hpx/${lib}/force_linking.hpp>\n")
```
hkaiser: This is most likely unintentional, is it?

libs/coroutines/include/hpx/coroutines/detail/context_base.hpp
```diff
@@ -298,54 +299,54 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail
...
+        //
+        //    if (reset)
+        //        m_allocation_counters.get(heap_num).store(0);
+        //    return result;
+        //}
+        //
+        //static std::uint64_t increment_allocation_count(std::size_t heap_num)
+        //{
+        //    return ++m_allocation_counters.get(heap_num);
+        //}
```
hkaiser: We can probably remove this.

libs/coroutines/include/hpx/coroutines/detail/context_base.hpp
```diff
@@ -387,15 +388,14 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail
...
-
         // Nothrow.
         void do_yield() noexcept
         {
             swap_context(*this, m_caller, detail::yield_hint());
         }
 
         // Nothrow.
-        void do_invoke() throw ()
+        void do_invoke() throw()
```
hkaiser: This should be `noexcept` instead of `throw()`. It's an oversight we forgot to change earlier.

aurianer:
Ok because we were wondering if we were integrating `apex.{hpp,cpp}` in this module too ?

hkaiser:
> Ok because we were wondering if we were integrating apex.{hpp,cpp} in this module too ?

I think the APEX integration will have to be changed to rely on callbacks APEX will register itself with. This will be the only way to avoid circular dependencies, I believe.

msimberg:
This should be ready now modulo any build failures that I've introduced. `invalid_thread_id` stayed with type `invalid_thread_id_tag` and specializations for the comparisons in `thread_id` were added. All APEX and `thread_data`/`thread_description` functionality is outside of this module and will go in a separate module (probably the next one...).

hpx/runtime/threads/thread_data_fwd.hpp
```diff
@@ -167,5 +171,18 @@ namespace hpx { namespace threads
...
 }}
 
+namespace std {
+    template <>
+    struct hash<::hpx::threads::thread_id_type>
+    {
+        std::size_t operator()(::hpx::threads::thread_id_type const& v) const
+            noexcept
+        {
+            std::hash<const ::hpx::threads::thread_data*> hasher_;
```
hkaiser: Not that I want to kick off a religious discussion here... But doesn't all of HPX use `east const` notations?

hpx/runtime/threads/thread_init_data.hpp
```diff
@@ -63,9 +63,15 @@ namespace hpx { namespace threads
...
-            apex_data = apex_new_task(description, parent_locality_id, parent_id );
+            // HPX_HAVE_APEX forces the HPX_HAVE_THREAD_DESCRIPTION
+            // and HPX_HAVE_THREAD_PARENT_REFERENCE settings to be on
+            if (description.kind() == description.data_type_description) {
+                apex_data = util::apex_new_task(description.get_description(),
+                        parent_id);
+            } else {
+                apex_data = util::apex_new_task(description.get_address(),
+                        parent_id);
+            }
```
hkaiser: Could we wrap this initialization into a helper function? It's used all over the place now...

libs/coroutines/include/hpx/coroutines/coroutine.hpp
```diff
@@ -34,49 +34,46 @@
...
 #include <cstddef>
 #include <cstdint>
 #include <limits>
 #include <utility>
 
-namespace hpx { namespace threads { namespace coroutines
-{
+namespace hpx { namespace threads { namespace coroutines {
     /////////////////////////////////////////////////////////////////////////////
+    template <typename ThreadData>
```
hkaiser: Would it be more appropriate to rely on an abstract base class instead of making the coroutines be templatized on the `thread_data` type? We will need to type-erase the `thread_data` anyways in order to add stack-less HPX threads later on.

src/runtime_handlers.cpp
```diff
@@ -44,6 +46,11 @@ namespace hpx { namespace detail {
         std::abort();
     }
 
+    bool enable_parent_task_handler()
+    {
+        return (hpx::get_initial_num_localities() == 1);
+    }
+
```
hkaiser: Should this be `ifdef`'ed for `HPX_HAVE_APEX`?

hpx/runtime/threads/thread_data_fwd.hpp
msimberg: No need for religion :) The const is where it always was and I'll happily move it.

hpx/runtime/threads/thread_init_data.hpp
msimberg: Yeah, this changed because of other reasons which don't make sense anymore. Will change.

libs/coroutines/include/hpx/coroutines/coroutine.hpp
msimberg: The reason I prefer this is that this way the coroutines module doesn't need to know *anything* about thread data (which is appropriate as the only use in the module is to store a pointer to thread data, and only return it to the outside world). However, if getting stackless threads in later will be easier with an abstract base class, then it's an option of course. How much of the current `thread_data` interface do you think would be required in `thread_data_base`? `thread_id` could of course be templated *and* there could be an abstract base class in the same module as `thread_data`?

src/runtime_handlers.cpp
msimberg: Yep, good catch.

aurianer:
Thanks a lot @msimberg !

hpx/runtime/threads/thread_data_fwd.hpp
hkaiser: We've had a `west const` in our code base? Unheard of! ;-)

libs/coroutines/include/hpx/coroutines/coroutine.hpp
hkaiser: The fact that `thread_id` actually holds the pointer to the thread data is an implementation detail that shouldn't ever have leaked into the code base. But you know how these things happen... For this reason, I think `thread_id` does not need to be templated on the thread data. Conceptually it's just an opaque id that somehow identifies the thread instance. 

It might be sufficient to store a `void*` or even a `std::size_t` in the `thread_id` and add a (friend) helper function `thread_data* get_thread_data(thread_id id);` that knows how to wield the magic.

For the abstract base class, I think we would need it in any case if we decided to add stack-less threads. I don't see another way for the scheduler to be agnostic of the thread type to handle. We will need some type of type erasure there for sure.

hpx/runtime/applier/apply.hpp
```diff
@@ -444,7 +444,6 @@ namespace hpx
 #ifdef HPX_HAVE_APEX
             data.apex_data = hpx::util::apex_new_task(
                 data.description,
-                data.parent_locality_id,
                 data.parent_id);
```
hkaiser: The idea of passing the `parent_locality_id` to APEX was to enable cross-locality tracking of things. Could you elaborate on the rationale of removing this, please?

hpx/runtime/threads/detail/create_thread.hpp
```diff
@@ -77,7 +77,7 @@ namespace hpx { namespace threads { namespace detail
         {
             if (data.priority == thread_priority_default &&
                 thread_priority_high_recursive ==
-                    threads::get_self_id()->get_priority())
+                    get_thread_id_data(threads::get_self_id())->get_priority())
```
hkaiser: We could even wrap this into a single function, something like `threads::get_self_data()` or similar. That would simplify writing the above to: `threads::get_self_data()->get_priority()`.

hpx/runtime/threads/thread_data_fwd.hpp
```diff
@@ -165,7 +171,19 @@ namespace hpx { namespace threads
...
+}}    // namespace hpx::threads
+
+namespace std {
+    template <>
+    struct hash<::hpx::threads::thread_id>
+    {
+        std::size_t operator()(::hpx::threads::thread_id const& v) const
+            noexcept
+        {
+            std::hash<::hpx::threads::thread_data const*> hasher_;
```
hkaiser: Thanks! ;-)

libs/coroutines/include/hpx/coroutines/thread_id_type.hpp
```diff
@@ -59,93 +46,76 @@ namespace threads {
...
-            std::basic_ostream<Char, Traits>& os, thread_id_type const& id)
+            std::basic_ostream<Char, Traits>& os, thread_id const& id)
         {
             os << id.get();
             return os;
         }
 
     private:
-        thread_data* thrd_;
+        void* thrd_;
```
hkaiser: May I suggest to use a `private: using thread_id_repr = void*;` (or similar) to remove the `void*` from the API?

libs/coroutines/CMakeLists.txt
```diff
@@ -0,0 +1,83 @@
...
+
+set(coroutines_sources
+  detail/context_base.cpp
+  detail/coroutine_impl.cpp
+  detail/coroutine_self.cpp
+  detail/tss.cpp
+  swapcontext.cpp
+  swapcontext32.ipp
+  swapcontext64.ipp
+  )
```
hkaiser: The last two files (if not three) are not needed on Windows.

hpx/runtime/threads/detail/create_thread.hpp
msimberg: Doh. Good idea!

hpx/runtime/applier/apply.hpp
msimberg: Well, we might've been too eager with this, but my reasoning was that the argument is unused. If it is to be made to work (again?) it can be put back in.

libs/coroutines/CMakeLists.txt
msimberg: I'd leave them in not to duplicate logic that's already in `swapcontext.cpp`. It'll be empty on anything that's not linux. And the platform detection variables in CMake are not fine-grained enough to have the same logic as in `swapcontext.cpp` (CMake only gives us `UNIX` whereas the file is non-empty only for linux). What do you think?

hpx/runtime/threads/detail/create_thread.hpp
msimberg: I've added a `threads::get_self_id_data` (I left `id` in there to indicate that it comes from the thread id). `get_thread_id_data` is still there as well for the cases when one gets a thread id from elsewhere.

libs/coroutines/CMakeLists.txt
hkaiser: I agree wrt including `swapcontext.cpp`. The `.ipp` files are not directly compiled but are included by that file. Do we need to have them here?

hpx/runtime/applier/apply.hpp
hkaiser: This might have been in the planning all the time. I'd suggest to leave it in for the time being, for this reason.

libs/coroutines/CMakeLists.txt
msimberg: Good point. I think they can be completely removed from the list. Thanks!

msimberg: They're out now.

hpx/runtime/applier/apply.hpp
msimberg: They're back.

PR #4103, opened 2019-09-24
Add this to lambda capture list

ShmuelLevine:
The implicit capture of 'this' by default by-value capture is deprecated in C++20.  This revision is intended to address this change (and to eliminate numerous warnings when building code).

### DIFF
--- a/hpx/runtime/parcelset/parcelhandler.hpp
+++ b/hpx/runtime/parcelset/parcelhandler.hpp
@@ -194,11 +194,10 @@ namespace hpx { namespace parcelset
         ///                 id (if not already set).
         HPX_FORCEINLINE void put_parcel(parcel p)
         {
-            auto f =
-                [=, this](boost::system::error_code const& ec, parcel const& p) -> void
-                {
-                    invoke_write_handler(ec, p);
-                };
+            auto f = [this](boost::system::error_code const& ec,
+                         parcel const& p) -> void {
+                invoke_write_handler(ec, p);
+            };
 
             put_parcel(std::move(p), std::move(f));
         }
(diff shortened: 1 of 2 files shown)


hpx/runtime/parcelset/parcelhandler.hpp
```diff
@@ -195,7 +195,7 @@ namespace hpx { namespace parcelset
         HPX_FORCEINLINE void put_parcel(parcel p)
         {
             auto f =
-                [=](boost::system::error_code const& ec, parcel const& p) -> void
+                [=, this](boost::system::error_code const& ec, parcel const& p) -> void
```
hkaiser: Do we really need the '=' in the capture list?

hpx/runtime/parcelset/parcelhandler.hpp
```diff
@@ -235,7 +235,7 @@ namespace hpx { namespace parcelset
         ///                 id (if not already set).
         void put_parcels(std::vector<parcel> parcels)
         {
-            std::vector<write_handler_type> handlers(parcels.size(), [=](
+            std::vector<write_handler_type> handlers(parcels.size(), [=, this](
```
hkaiser: Same here, do we still need the '=' in the capture list?

msimberg:
@ShmuelLevine unfortunately adding this leads to warnings for older standards (i.e. older than 20). @hkaiser's suggestion might help with some of them. If that's not an option we might need something like [this](https://github.com/STEllAR-GROUP/hpx/blob/1000bb00e66cb490cb9f3bad84bf5924f75f964f/libs/functional/tests/unit/function/function_test.cpp#L17-L29) to disable warnings with older standards. For reference the warnings are: [clang](https://cdash.cscs.ch//viewBuildError.php?buildid=75318) and [gcc](https://cdash.cscs.ch//viewBuildError.php?buildid=75365).

hkaiser:
@msimberg We already have `HPX_CAPTURE_MOVE` and `HPX_CAPTURE_FORWARD`. Why not add `HPX_CAPTURE_THIS`?

hkaiser:
This looks good now. The APEX builder refers to an outdated commit.

PR #4104, opened 2019-09-24
Extend scheduler_mode with new work_stealing and task assignment modes

biddisco:
Replace numa_sensitive variable and make core stealing and numa stealing
more consistent. Allow schedulers to have their stealing modes changed
by executors/users or algorithms

Note - This will probably not compile as is, so it cannot be merged until it has run a few times on the dashboard and I've fixed the diffs that will cause trouble between master and my work branch that this came from.


hpx/runtime/threads/policies/static_queue_scheduler.hpp
```diff
@@ -72,7 +72,12 @@ namespace hpx { namespace threads { namespace policies
           : base_type(init, deferred_initialization)
         {}
 
-        virtual bool has_thread_stealing(std::size_t num_thread) const override
+        virtual bool has_work_stealing_core() const override
+        {
+            return false;
+        }
+
+        virtual bool has_work_stealing_numa() const override
```
msimberg: Shouldn't these just come from `scheduler_base` which checks if the flags are set? The static schedulers disable stealing initially but there's nothing stopping someone from enabling stealing at runtime. Or, if these two functions should always return false we should also throw if someone tries to enable stealing.

src/runtime/threads/threadmanager.cpp
```diff
@@ -582,17 +597,24 @@ namespace hpx { namespace threads {
...
+                std::string affinity_desc;
+                std::size_t numa_sensitive =
+                    hpx::util::get_affinity_description(cfg_, affinity_desc);
                 // instantiate the scheduler
                 typedef hpx::threads::policies::
                     shared_priority_queue_scheduler<>
                         local_sched_type;
                 local_sched_type::init_parameter_type init(
-                    thread_pool_init.num_threads_, {4, 4, 64},
+                    thread_pool_init.num_threads_, {1, 1, 64},
```
msimberg: Did you mean to change these?

hpx/runtime/threads/policies/local_priority_queue_scheduler.hpp
```diff
@@ -1256,7 +1246,7 @@ namespace hpx { namespace threads { namespace policies {
             });
 
             // check for the rest and if we are NUMA aware
-            if (numa_sensitive_ != 2 && any(first_mask & pu_mask))
+            if (has_work_stealing_numa() && any(first_mask & pu_mask))
```
msimberg: Is the old distinction of `numa_sensitive = 0/1/2` not necessary/used anymore? I don't remember what the distinction between the three states was to be honest.

hpx/runtime/threads/policies/scheduler_mode.hpp
```diff
@@ -32,15 +32,18 @@ namespace hpx { namespace threads { namespace policies
             ///< scheduler to dynamically increase and reduce the number of
             ///< processing units it runs on. Setting this value not succeed for
             ///< schedulers that do not support this functionality.
-        enable_stealing = 0x20,          ///< This option allows for certain
+        enable_stealing_core = 0x20,          ///< This option allows for certain
             ///< schedulers to explicitly disable thread stealing
-        enable_idle_backoff = 0x40,     ///< This option allows for certain
+        enable_stealing_numa = 0x40,          ///< This option allows for certain
+            ///< schedulers to explicitly disable thread stealing
```
msimberg: These could do with more thorough descriptions. What's the distinction between the two modes? Should they be `enable_work_stealing` and `enable_work_stealing_numa`? Is `enable_stealing_core == false` and `enable_stealing_numa == true` an allowed configuration? And what would it do if it is?

hpx/runtime/threads/policies/static_queue_scheduler.hpp
biddisco: The current design has a virtual call to check the flags, the schedulers that don't allow stealing, just reset the flags back to off when returning the flags, even if someone has manually set them on. I don't like it, but I just added a numa version to allow the set and get of the flags. 
Ideally we would just remove the schedulers that don't support these features, remove the virtual calls, not need to throw if someone tries to enable unsupported features. For now I wanted to simply get the new flag added.

src/runtime/threads/threadmanager.cpp
biddisco: yes. It was intentional, but I can remove it to a later PR as it is an unrelated flyby.

hpx/runtime/threads/policies/scheduler_mode.hpp
biddisco: `enable_stealing_core = OFF` and `enable_stealing_NUMA=ON` is not really meaningful, but I added `_core` simply a a way of making it clear what was going on in the code to distinguish the two flags.

biddisco:
I have redone this PR and substantially improved it as follows

1. Remove unnecessary duplication of cache-aligned scheduling mode variable
There is no need to have one copy per thread, so clean this up and change all places that call the get_scheduler_mode() function

2. Replace numa_sensitive variable and make (core) stealing and numa stealing
more consistent
Add scheduler mode flags for a number of new scheduling choices, make work_stealing and work_stealing_numa more consistent in their use

3. Allow schedulers to have their stealing modes changed by executors/users or algorithms
Code can switch stealing on/off programmatically

4. Make get_scheduler_mode (called very frequently) non virtual and instead make set_scheduler_mode virtual and ensure it is always called when setting mode and overridden by schedulers that do not support certain modes.
This is an optimization - get() is called on every pass of the scheduling loop and used to be virtual. Instead make set() (called rarely) virtual and switch off flags there.

biddisco:
NB. I have also addressed one complaint in the original PR which was that the docs for scheduler modes was incorrect/incomplete. I have added new modes and added a description for each.

libs/threadmanager/src/threadmanager.cpp
```diff
@@ -363,21 +367,25 @@ namespace hpx { namespace threads {
...
+
                 std::unique_ptr<local_sched_type> sched(
                     new local_sched_type(init));
 
+                // set the default scheduler flags
+                sched->add_scheduler_mode(thread_pool_init.mode_);
+                // conditionally set/unset this flag
+                sched->update_scheduler_mode(
+                    policies::enable_stealing_numa, !numa_sensitive);
+
```
msimberg: Could we pass the default mode directly to the constructor?

hpx/runtime/threads/policies/static_queue_scheduler.hpp
```diff
@@ -29,6 +29,8 @@
 
 #include <hpx/config/warnings_prefix.hpp>
 
+// JB set work stealing false in mode set
+
```
msimberg: Leftover comment?

libs/threadmanager/src/threadmanager.cpp
biddisco: Yes. The PR is still slightly WIP and there are a few missing things that I'm cleaning up and fixing now. The default mode was missing in the last commit, but fixed in the next one.
The numa sensitive mode is not default because the command line can override it, so by default be allow numa-stealing, but it can be turned off, the update_mode(blah, false) does this.

hpx/runtime/threads/policies/static_queue_scheduler.hpp
biddisco: yes. Fixed in upcoming commit.
Existing schedulers override get_scheduler_mode to set the stealing flags to false if they do not support it. I have changed to to override the set_xxx instead and left that comment to remind me to check all schedulers are doing the right thing with flag they do not supoort

biddisco:
Once I'm happy that the PR is ready I will remove the WIP from the title and ask for a final set of reviews.
NB. The embedded scheduler fractal_executor test is failing. I'm looking at that now, but would not want that test to hold up this PR if it turns out to be a fundamental problem. Views?

biddisco:
I think the tests are now passing with my latest comit. I will check the clang-format and any other tidyups now.

hkaiser:
@biddisco please fix the clang-format issues, otherwise this PR is good to go, I think.

biddisco:
Fingers crossed that this is ok now.

hpx/runtime/threads/policies/scheduler_base.hpp
```diff
@@ -261,7 +279,7 @@ namespace hpx { namespace threads { namespace policies {
         // the scheduler mode is simply replicated across the cores to
         // avoid false sharing, we ignore benign data races related to this
         // variable
-        std::vector<util::cache_line_data<std::atomic<scheduler_mode>>> modes_;
+        std::atomic<scheduler_mode> mode_;
```
msimberg: It might still be good to leave this on its own cache line to keep it separate from the mutexes below.

msimberg: And the comment above needs to be updated...

hpx/runtime/threads/policies/static_queue_scheduler.hpp
msimberg: Remove?

hpx/runtime/threads/policies/scheduler_base.hpp
biddisco: good catch. I meant to have it in a cache line thingy, but completely forgot.

PR #4108, opened 2019-09-25
Making sure C++ standard is properly detected and propagated

hkaiser:
- enable C++17 features only if available and compilation mode is C++17 (__cplusplus >= 201500), same for C++14  (__cplusplus >= 201402)
- making sure MSVC actually generates proper __cplusplus preprocessor constant
- flyby: fix CMakeLists.txt to properly detect MSVC versions


cmake/HPX_AddDefinitions.cmake
```diff
@@ -80,7 +80,21 @@ function(write_config_defines_file)
 
   set(hpx_config_defines "\n")
   foreach(def ${DEFINITIONS_VAR})
-    set(hpx_config_defines "${hpx_config_defines}#define ${def}\n")#"
+    # C++17 specific variable
+    string(FIND ${def} "HAVE_CXX17" _pos)
+    if(NOT ${_pos} EQUAL -1)
+      set(hpx_config_defines
+         "${hpx_config_defines}#if __cplusplus >= 201500\n#define ${def}\n#endif\n")
```
sithhell: This is MSVC specific, right?

hkaiser: No it isn't. It's gcc specific - normally `__cplusplus == 201703` for C++17 (iirc). I used this to decide on a constant: https://renenyffenegger.ch/notes/development/languages/C-C-plus-plus/preprocessor/macros/predefined/__cplusplus.

PR #4109, opened 2019-09-26
Overload for start without entry point/argv.

### LINKED ISSUE #4107: Initializing runtime without entry point function and command line arguments
While trying to integrate HPX into an existing simulation framework, I ran into the issue of initializing the runtime while providing neither an entry point function nor command line arguments. Due to the specifics of our framework at the point where HPX should be initialized I cannot access argc/argv, as the corresponding code is located in a module loaded dynamically at runtime. Additionally, it is necessary to suspend/resume the runtime in our application.

Following the example in the [corresponding section of the manual](https://stellar-group.github.io/hpx/docs/sphinx/latest/html/manual/starting_the_hpx_runtime.html#suspending-and-resuming-the-hpx-runtime), I could not find an overload for `hpx::start()` or `hpx::init()` which allows me to initialize the runtime as described above while ideally passing the configuration to HPX via a vector of strings.

severinstrobl:
Fixes #4107.

This overload allows starting the runtime while providing no entry point
and no command line arguments to parse.

### DIFF
--- a/hpx/hpx_start.hpp
+++ b/hpx/hpx_start.hpp
@@ -770,6 +770,9 @@ namespace hpx
     inline bool start(std::nullptr_t f,
         int argc, char** argv, std::vector<std::string> const& cfg,
         hpx::runtime_mode mode = hpx::runtime_mode_default);
+
+    inline bool start(std::nullptr_t f, std::vector<std::string> const& cfg,
+        hpx::runtime_mode mode = hpx::runtime_mode_default);
 /// \endcond
 }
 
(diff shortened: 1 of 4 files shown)


hpx/hpx_start.hpp
```diff
@@ -770,6 +770,9 @@ namespace hpx
     inline bool start(std::nullptr_t f,
         int argc, char** argv, std::vector<std::string> const& cfg,
         hpx::runtime_mode mode = hpx::runtime_mode_default);
+
+    inline bool start(std::nullptr_t f, std::vector<std::string> const& cfg,
```
hkaiser: Could you add the doxygen comments for this (similarly to the other functions, please?

severinstrobl: Sorry, didn't see this in time. Will do.

hkaiser:

hkaiser:
@severinstrobl Thanks again for contributing. We are giving away a STE||AR-Group t-shirt for our first-time contributors. If you would like to own one please get in contact with @aserio, he will be able to get it to you.

PR #4111, opened 2019-09-28
Fixing performance regression introduced earlier

hkaiser:
@tianyi93 Could you please verify whether this fixes your performance issue?

### DIFF
--- a/hpx/runtime/threads/policies/lockfree_queue_backends.hpp
+++ b/hpx/runtime/threads/policies/lockfree_queue_backends.hpp
@@ -25,269 +25,265 @@
 #include <cstdint>
 #include <utility>
 
-namespace hpx { namespace threads { namespace policies
-{
+namespace hpx { namespace threads { namespace policies {
 
-struct lockfree_fifo;
+    struct lockfree_fifo;
 
-///////////////////////////////////////////////////////////////////////////////
-// FIFO
-template <typename T>
-struct lockfree_fifo_backend
-{
-    typedef boost::lockfree::queue<T> container_type;
-
-    typedef T value_type;
-    typedef T& reference;
-    typedef T const& const_reference;
-    typedef std::uint64_t size_type;
-
-    lockfree_fifo_backend(
-        size_type initial_size = 0
-      , size_type num_thread = size_type(-1)
-        )
-      : queue_(std::size_t(initial_size))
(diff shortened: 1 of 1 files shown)


biddisco:
The diff here is quite large and mostly stuff moved around, with some changes for 128 bit atomic presence - can you summarize please what is the difference that causes a speed improvement - thanks.

hkaiser:
> The diff here is quite large and mostly stuff moved around, with some changes for 128 bit atomic presence - can you summarize please what is the difference that causes a speed improvement - thanks.

This PR is very small. The gist is we changed everything to `lockfree::queue` (from `lockfree::deque`), which has caused the issue.

hkaiser:
> > The diff here is quite large and mostly stuff moved around, with some changes for 128 bit atomic presence - can you summarize please what is the difference that causes a speed improvement - thanks.
>>
> This actual change is very small. The gist is we changed everything to `lockfree::queue` (from `lockfree::deque`), which has caused the issue. See here for the important patch: https://github.com/STEllAR-GROUP/hpx/pull/4111/files#diff-6dece8d84d34dc5dc3d0df6e67f26f7dR37-R69.
>
> Everything else are changes caused by clang-format.

Sorry for this. I went back and redid the patch separating the actual fix from the reformatting.

tianyi93:
@hkaiser I did a small test showing that the performance issue has been fixed. I will do the whole benchmark run. Thanks!

hkaiser:
>  I did a small test showing that the performance issue has been fixed.

@tianyi93 Thanks! I'll merge this as soon as the tests have cycled.

msimberg:
Have I understood correctly that `boost::lockfree::deque` is actually faster than `boost::lockfree::queue` as a FIFO queue? I admit I did just assume `queue` would be better than `deque`.

hkaiser:
> Have I understood correctly that boost::lockfree::deque is actually faster than boost::lockfree::queue as a FIFO queue? I admit I did just assume queue would be better than deque.

@msimberg yes, that seems to be the case. I was not expecting this either. We ran into the performance regression while testing hpxMP. There is mostly use `hpx::apply` to schedule threads which are then joined using `hpx::lcos::local::latch`. Both facilities a very lightweight, thus the effect of the used scheduler was very pronounced.

Issue #3816, opened 2019-04-23
Add assertion macros that print their values


msimberg:
Having alternative assertion macros in the style of 
```
HPX_ASSERT_EQ(x, 3);
```
which print the value of `x` when the assertion fails would be nice to have to make debugging slightly easier.

hkaiser:
Sounds like a feature for the assert module sitting in the pipeline ;-)

msimberg:
> Sounds like a feature for the assert module sitting in the pipeline ;-)

Indeed!

tapaswenipathak:
Hi folks: Can I PR or is the ticket for internal folks?

msimberg:
@tapaswenipathak feel free to work on this. You might want to have a look at what's in the testing module: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/testing/include/hpx/testing.hpp. We should try to avoid duplicating functionality in both modules.

hkaiser:
> feel free to work on this. You might want to have a look at what's in the testing module: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/testing/include/hpx/testing.hpp. We should try to avoid duplicating functionality in both modules.

We should keep in mind that testing is different from using assertions. If those can share functionalities - fine. I think this is not a requirement, though.

tapaswenipathak:
Thanks for the brain dump folks, I didn't read much, I think I can PR that macro out. I will ask for any questions, thanks.

PR #4112, opened 2019-09-28
Assertions: Add assertion macros that print their values

tapaswenipathak:
Fixes https://github.com/STEllAR-GROUP/hpx/issues/3816

## Proposed Changes

using `evaluate_assert` https://github.com/STEllAR-GROUP/hpx/blob/master/libs/assertion/include/hpx/assertion/evaluate_assert.hpp.

I think ticket has sufficient information.


libs/assertion/include/hpx/assertion.hpp
```diff
@@ -58,16 +58,27 @@ namespace hpx { namespace assertion {
             __FILE__, __LINE__, HPX_ASSERT_CURRENT_FUNCTION},                  \
         HPX_PP_STRINGIZE(expr), [&]() { return msg; });                        \
 /**/
+#define HPX_ASSERT_EQ(x, count)
```
msimberg: This definition is not necessary (and leads to errors due to `HPX_ASSERT_EQ` begin redefined). I'd also suggest naming the two arguments e.g. `expr1` and `expr2` to be consistent with the `HPX_TEST_*` macros and since the second argument isn't necessarily a number.

libs/assertion/include/hpx/assertion.hpp
```diff
@@ -58,16 +58,27 @@ namespace hpx { namespace assertion {
...
+#define HPX_ASSERT_EQ_(x, count)
+    ::hpx:assertion::detail::evaluate_count_assert(
```
msimberg: `evaluate_count_assert` isn't defined anywhere. Did you maybe forget to add it?

msimberg:
Thanks for starting this! A natural next step (in this PR) would be to then actually check for uses of `HPX_ASSERT` where this could be used instead, and do the replacement.

libs/assertion/include/hpx/assertion.hpp
tapaswenipathak: What?! How do you visualise the `HPX_ASSERT_EQ(x, 3);` item with `HPX_ASSERT_EQ(expr1, expr2);`? 

- Do you require a macro equating two expressions? and returning a boolean value?
- Do you require a macro equating x and a digit and return digit if equating fails?

libs/assertion/include/hpx/assertion.hpp
tapaswenipathak: Should be `evaluate_assert`.

tapaswenipathak:
ack, ok.

libs/assertion/include/hpx/assertion.hpp
msimberg: Please have a look at https://github.com/STEllAR-GROUP/hpx/blob/master/libs/testing/include/hpx/testing.hpp to get some ideas on how you can implement it.

tapaswenipathak: Can you please answer:

- Do you require a macro equating two expressions? and returning a boolean value?
- Do you require a macro equating expression and a value and return digit if equating fails?

Ref: https://github.com/STEllAR-GROUP/hpx/issues/3816#issuecomment-527876871.

msimberg: The `HPX_ASSERT_EQ` macro should take two *expressions*, with the only requirement that the types of the two expressions are equatable (have operator `==`) and that they can be written to a stream (have operator `<<`). A macro in general can not return a value, but the macro should also not evaluate to a value. The macro should check for equality between the two values and if they are not equal call `handle_assert` with a message which prints both the expression (`expr1 == expr2`) and the evaluated values of the two expressions (e.g. `1 == 2`, or `"abc" == "def"`). This is where the current `HPX_ASSERT` macro eventually ends up: https://github.com/STEllAR-GROUP/hpx/blob/414380e50e55ed1f4ebfde57f3bda7018d6d1cf0/libs/assertion/include/hpx/assertion/evaluate_assert.hpp#L22-L30, and then: https://github.com/STEllAR-GROUP/hpx/blob/414380e50e55ed1f4ebfde57f3bda7018d6d1cf0/libs/assertion/src/assertion.cpp#L31-L48. It may be easiest to add functions `evaluate_assert_eq`, `evaluate_assert_neq` etc. functions to handle the different comparisons. Does that make it clearer?

tapaswenipathak: Cool, ok.

msimberg:
@tapaswenipathak can we help in any way here?

msimberg:
@tapaswenipathak I'll close this one as well for now to avoid CI rebuilds. Please reopen when you get back to it.

PR #4113, opened 2019-09-29
Module datastructures

hkaiser:
This PR fixes the include directory for traits headers related to the datastructures module. It also adds missing compatibility headers for those traits files.

Most importantly, this patch proposes a `moveonly_any` that is capable of storing move-only data types.


hkaiser:
@msimberg, @sithhell I think this is good to go now. Please review. I have no idea why the clang-oldest builder is unhappy...

msimberg:
This doesn't answer your question but the error message is definitely related to this PR: https://cdash.cscs.ch//viewBuildError.php?buildid=75704. @sithhell any ideas?

hkaiser:
> This doesn't answer your question but the error message is definitely related to this PR

I know :/ I think this version of clang does not properly implement `std::is_void<>` (unlikely, but a possible reason, see: https://github.com/STEllAR-GROUP/hpx/pull/4113/files#diff-9a4d5912ff0cdb713bd9aa5930718dc3R599-R603). Possibly it could be the `enable_if` applied to a forward declaration... Not sure.

hkaiser:
@msimberg I have modified the implementation to get away without the SFINAE'd forwarding declaration. Let's see if this changes things for this compiler.

hkaiser:
@msimberg things are better now: https://cdash.cscs.ch//buildSummary.php?buildid=75954

sithhell:
Hmm, too bad. I don't like the name `moveonly_any`. I would have proposed `unique_any` to go along with `unique_function`.

msimberg:
@sithhell sorry, I didn't realize there was still an ongoing discussion about naming. I have no opinion on this so feel free to rename it if you come to a concensus on it.

hkaiser:
I'm fine with any name, so `unique_any` sounds good to me.

PR #4114, opened 2019-09-30
Module serialization

hkaiser:
This PR separates the basic serialization functionality into a new module. It removes the dependency of our `output_archive` from the special handling of id_types (credit splitting) and futures (await). The actual logic for those is kept in place for now, it is just separated into their own independent types.

This depends on #4113.

This will also enable us to move forward on #3159.


src/runtime/naming/name.cpp
```diff
@@ -239,6 +241,35 @@ namespace hpx { namespace naming
 
         ///////////////////////////////////////////////////////////////////////
         // prepare the given id, note: this function modifies the passed id
+        void handle_credit_splitting(
+            serialization::output_archive& ar, id_type_impl& gid,
+            serialization::detail::preprocess_gid_types* split_gids)
+        {
+            auto* handle_futures = ar.try_get_extra_data<
```
sithhell: I am not very confident with those changes here. I'd rather like to have the original semantics restored, as in, assume we always have a future handler and always have a split gid handler. Let's work on other use cases as soon as we need them.

src/runtime/naming/name.cpp
```diff
@@ -239,6 +241,35 @@ namespace hpx { namespace naming
...
+                    });
+
+                handle_futures->await_future(
+                    *traits::future_access<decltype(f)>::get_shared_state(f));
+            }
+            else
+            {
+                auto f = split_gid_if_needed(gid);
+                handle_futures->await_future(
+                    *traits::future_access<decltype(f)>::get_shared_state(f));
```
sithhell: This is definitely not correct and will lead to memory leaks if the gid was split.

src/runtime/naming/name.cpp
hkaiser: Under normal circumstances (i.e with `output_archives` constructed without specifying extra data) should we always rely on credit-splitting and/or future handling to be available? Or should we remove the ability to construct `output_archives` without extra data in the first place?

My reasoning was that users normally don't need credit-splitting or handling of (HPX) futures. They simply want to construct an archive and have their serialization needs covered.

src/runtime/naming/name.cpp
hkaiser: Right. I think we can just throw (assert?) if we get to this point as it shouldn't even happen in the first place.

src/runtime/naming/name.cpp
sithhell: That's correct, but in our parcel layer, we know that those might need to be serialized. We absolutely need to handle them. So in our case, we can just add them when constructing the archive (we need them anyway), let's try to move on with this with as little extra changes, and deal with additional usage requirements along the lines.

src/runtime/naming/name.cpp
sithhell: For now, I suggest to not change the current behavior (an assert to catch potential mistakes might do it for now just as well).

src/runtime/naming/name.cpp
hkaiser: That would either require to a) touch on all examples and tests that use serialization directly or b) have any `output_archive` that is constructed without explicit extra data support all of the internal HPX extra data items. Both options don't look appealing to me.

sithhell:
There seems to be a problem with gcc in C++17 mode:
http://rostam.cct.lsu.edu/builders/hpx_gcc_8_boost_1_70_centos_x86_64_debug/builds/156/steps/build_core/logs/stdio

hkaiser:
> There seems to be a problem with gcc in C++17 mode:

@sithhell this has been fixed. I'm currently working on the remaining test failure, which is unrelated to the external archive data.

src/runtime/naming/name.cpp
hkaiser: That's fixed now.

hkaiser:
@msimberg, @sithhell from my end, this is good to go. All tests pass. I'd be fine with merging this now even if @sithhell has some ideas on how to better encapsulate the extra archive data. Those could be implemented separately.

PR #4116, opened 2019-10-01
Fix automatically setting C++ standard

msimberg:
Attempting to fix C++ standard detection when `HPX_WITH_CXXAB` is not set explicitly. The problems occur when the C++ standard is not set explicitly and the compiler defaults to a standard older than C++11 (like gcc <= 5 and clang <= 5).

@hkaiser I've attempted to fix this, but please have a look to see that I haven't undone your changes in #4108. In particular, I didn't quite understand the removal of some `set(... PARENT_SCOPE)` calls and the purpose of setting the default standard to 14. We should either set the standard to 14 when `HPX_WITH_CXXAB` isn't use and remove the automatic detection of the standard or stick to what we've had which is to detect the highest standard supported by the compiler. I've gone back to the latter, but I think we can go either way with this. We could also require `HPX_WITH_CXXAB` to be set and error out otherwise.


hkaiser:
@msimberg I'm not sure I didn't break anything on the non-windows-side. I trust you have it under control.

msimberg:
Your trust is not warranted but I tried my best. Things are happier on rostam (http://rostam.cct.lsu.edu/builders/hpx_clang_4_boost_1_63_centos_x86_64_debug/builds/403) at least so let's go with this for now.

Issue #4118, opened 2019-10-03
static_chunk_size is not obeyed in scan algorithms


jefftrull:
## Expected Behavior

I assumed that supplying an execution policy with a fixed chunk size would result in each individual task working on a "chunk" of that size from the input.

## Actual Behavior

Each individual task has a size that is a function of the overall input size and the number of threads. The static chunk size does not appear to affect it.

## Steps to Reproduce the Problem

  1. Build the attached code with HPX master
  1. Run under gdb with a breakpoint at [the size extraction](https://github.com/STEllAR-GROUP/hpx/blob/162c0bd3c728b8fe562babd02af5f158240014ac/libs/algorithms/include/hpx/parallel/util/scan_partitioner.hpp#L133)
  1. Print `size` and observe it is not the static_chunk_size (50000)

## Specifications

  - HPX Version: master from git
  - Platform (compiler, OS): gcc 8.3.0, Ubuntu 19.04

jefftrull:
```c++
// reproducing chunk size issue
// Jeff Trull
#include <random>

#include <hpx/hpx_init.hpp>
#include <hpx/hpx.hpp>
#include <hpx/include/parallel_numeric.hpp>
#include <hpx/include/parallel_algorithm.hpp>

int main(int argc, char* argv[])
{
    // By default this should run on all available cores
    std::vector<std::string> const cfg = {
        "hpx.os_threads=2"
    };

    // Initialize and run HPX
    return hpx::init(argc, argv, cfg);
}

int hpx_main(int argc, char **argv)
{
    using namespace hpx::parallel;
    using namespace hpx::util;

    // create a fixed chunk size to be used in the algorithm
    execution::static_chunk_size fixed(50000);
    // this does not seem to be obeyed!
    auto ex = execution::par.with(fixed);

    // run the below with a breakpoint at scan_partitioner.hpp:133 where each chunk is extracted
    // I find that the size of each chunk is a function of the data size and the number of threads
    // but not the "fixed" chunk size I supplied

    // create and fill random vector of desired size
    std::random_device rnd_device;
    std::mt19937 mersenne_engine{rnd_device()};
    std::uniform_int_distribution<int> dist{1, 20};

    std::size_t sz = 16777216;

    std::vector<int> data(sz);
    std::generate(data.begin(), data.end(),
                  [&]() { return dist(mersenne_engine); });

    std::vector<int> result(sz + 1);

[...]

PR #4119, opened 2019-10-03
Making sure chunk_size and max_chunk are actually applied to parallel algorithms if specified

hkaiser:
Fixies #4118

@jefftrull Could you please verify whether this fixes your issue?

Note this also moves three tests into the execution module that were left behind earlier. Those are unrelated to the issue at hand.


jefftrull:
I confirm that this allows me to specify the chunk size without using the `maximal_number_of_chunks` customization point

hkaiser:
This looks good, I'll go ahead and merge it.

PR #4121, opened 2019-10-04
Fixing migration test if networking is disabled.

hkaiser:


hkaiser:
The test is fixed, no other problems occurred - merging,,,

PR #4122, opened 2019-10-06
Add kernel launch helper utility (+saxpy demo) and merge in octotiger …

biddisco:
…changes

Add a demo that computes saxpy on the GPU and launches it using
the cudaLaunchKernel function via the stream helper.

Tested for compatibility with latest Octotiger at Hackathon.


biddisco:
Don't merge yet - I think I forgot to rename one function call in a test that uses this.

biddisco:
This should be ok to merge now. I see a couple of fails on the dashboard, but they do not appear to be related to the PR

msimberg:
Thanks @biddisco. Yep, failures are unrelated... merging.

PR #4124, opened 2019-10-08
Don't link to Boost.System unless required

msimberg:
Boost.System is header only from 1.69 onwards. It still produces a stub library which one can link to, but this disables linking when it's not necessary. This means that in the default configuration with GCC 9 and Boost >= 1.69 HPX doesn't require any compiled libraries from Boost anymore.

Flyby: Fix a couple of other Boost-related things.


msimberg:
This worked fine locally but not on pycicle. It turned out that `Boost_VERSION_STRING` is only defined in CMake 3.14 onwards. Should be good to go now.

PR #4125, opened 2019-10-08
Changing extra_archive_data implementation

sithhell:
This patch does not rely on any magic indexing of the archive extra data.
If an extra data for an archive is required, it is being constructed
lazily, with the addition of enforcing post conditions with asserts
when the extra data is being destructed.

This is the follow up from #4114


msimberg:
I *think* the `daint-gcc-newest` failure is a problem with the builder, but let's just wait a day and see what it says the next round.

PR #4131, opened 2019-10-10
Applying fixes based on reports from PVS Studio

hkaiser:


hpx/lcos/local/recursive_mutex.hpp
```diff
@@ -44,14 +44,15 @@ namespace hpx { namespace lcos { namespace local
         template <>
         struct thread_id_from_mutex<lcos::local::spinlock>
         {
-            typedef std::size_t thread_id_type;
+            using thread_id_type = std::size_t;
 
-            static thread_id_type invalid_id() { return ~0u; }
+            static thread_id_type invalid_id() { return ~thread_id_type(0); }
```
msimberg: By the way, is this meant to be something different that `invalid_thread_id`?

hkaiser: Good catch, this is most likely the same.

msimberg: https://docs.microsoft.com/en-us/windows/win32/procthread/thread-handles-and-identifiers claims that `GetCurrentThreadId` won't use 0 for a valid id. For `pthread_self` it's not as clear, but `0` is likely not worse than `~0` for an invalid thread id. So we should be okay with just using `invalid_thread_id`.

hkaiser:
The remaining failures are unrelated, this can go ahead. Please review.

PR #4133, opened 2019-10-11
Fix block_executor

### LINKED ISSUE #4132: parallel fill leaves elements unfilled
## Expected Behavior
The parallel fill should fill all elements of the vector passed to it

## Actual Behavior
Seems like a power of 2 smaller than the array is filled

Compile and run the following code:
```
#include <hpx/hpx.hpp>
#include <hpx/include/parallel_fill.hpp>
#include <hpx/include/compute.hpp>
#include <hpx/include/parallel_executors.hpp>
#include <hpx/hpx_main.hpp>

void prvec(std::vector<float> v){
    for(auto i : v)
      std::cout << i << " ";
    std::cout << std::endl;

};

hpx::compute::host::block_executor<> getexec(){
    auto host_targets = hpx::compute::host::get_local_targets();
    typedef hpx::compute::host::block_executor<> executor_type;
    decltype(host_targets) v;
    for(auto host : host_targets) {
        if(v.size() == 4)
            break;
        v.push_back(host);
    }
    executor_type exec(v);
    return exec;

};

void fill_example(){
    auto exec = getexec();
    std::vector<float> vd;
    for(int i=0;i<5;i++) vd.push_back(1.f);
    prvec(vd);
    hpx::parallel::fill(hpx::parallel::execution::par.on(exec),vd.begin(),vd.end(),2.0f);
    prvec(vd);
};

int main(int argc, char **argv) {
    fill_example();
    return 0;
}
```

  - HPX Version: top of master
  - Platform (compiler, OS): gcc 9.2.1

msimberg:
Fixes #4132. In general fixes the `block_executor` to make sure it forwards all chunks to the underlying executors.

@hkaiser this fixes the problem, but I'm unsure about my `std::advance`s. A forward iterator does not guarantee O(1) complexity for it, does it? I found the current version easier to read, but I'll obviously change it to do fewer `std::advance`s if needed.

### DIFF
--- a/hpx/compute/host/block_executor.hpp
+++ b/hpx/compute/host/block_executor.hpp
@@ -150,27 +150,29 @@ namespace hpx { namespace compute { namespace host
                 >::type
             > > results;
             std::size_t cnt = util::size(shape);
-            std::size_t part_size = cnt / executors_.size();
+            std::size_t num_executors = executors_.size();
 
             results.reserve(cnt);
 
-            try {
+            try
+            {
                 auto begin = util::begin(shape);
                 for (std::size_t i = 0; i != executors_.size(); ++i)
                 {
+                    std::size_t part_begin_offset = (i * cnt) / num_executors;
+                    std::size_t part_end_offset =
+                        ((i + 1) * cnt) / num_executors;
+                    auto part_begin = begin;
                     auto part_end = begin;
-                    std::advance(part_end, part_size);
-                    auto futures =
-                        parallel::execution::bulk_async_execute(
-                            executors_[i],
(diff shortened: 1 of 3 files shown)


hkaiser:
> this fixes the problem, but I'm unsure about my std::advances. A forward iterator does not guarantee O(1) complexity for it, does it? I found the current version easier to read, but I'll obviously change it to do fewer std::advances if needed.

The parallel algorithms generally refuse to compile for forward (and output) iterators; thus you should be fine. Besides, the whole purpose of `std::advance` is to abstract away the different iterator categories as those would require different code for advancing them otherwise.

Issue #4135, opened 2019-10-14
Use object libraries for modules


msimberg:
We should consider using [object libraries](https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries) for creating `libhpx` out of the modules. This should be significantly cleaner and may let us avoid the `force_linking` mechanisms. However, we would need CMake 3.12 for this (object libraries exist before that, but are somewhat crippled). I'm tentatively adding this to next year's release, when I think we could reasonably bump the minimum CMake version to 3.12 or higher.

hkaiser:
Sounds good to me! Good thinking.

hkaiser:
Keeping this open.... (adding `pinned` label).

PR #4136, opened 2019-10-14
HPX.Compute modules

msimberg:
Adds two modules, `hpx_compute` and `hpx_compute_cuda`. I thought separating the two makes sense in case the HPX.Compute world is expanded some time in the future. `hpx_compute` could be called `hpx_compute_host` as well, but I consider it a kind of "base" module for HPX.Compute.

On Linux, since nothing from `libhpx_compute{_cuda}` is used in `libhpx`, no symbols actually get added to `libhpx` (I thought our `force_linking` contraptions would do this, but it seems like that's not the case). To get around this I've added a `--whole-archive` option when on `UNIX`. I think eventually the proper way of doing this is #4135.


libs/compute/include/hpx/compute/host/target.hpp
```diff
@@ -25,7 +25,7 @@
 
 namespace hpx { namespace compute { namespace host
 {
-    struct target
+    struct HPX_EXPORT target
```
hkaiser: If we remove the `HPX_EXPORT` here we will have to add it at least for the serialization functions below.

msimberg: Note: I've *added* the `HPX_EXPORT`. I could remove it from `native_handle_type` below though...

hkaiser: Uhh, sorry - I'm still sleeping, apparently… And yes, we don't need it for the `native_handle_type`

libs/compute/tests/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+if (HPX_WITH_TESTS_BENCHMARKS)
+  add_hpx_pseudo_target(tests.performance.modules.compute)
+  add_hpx_pseudo_dependencies(tests.performance.modules tests.performance.modules.compute)
+  add_subdirectory(performance)
+endif()
+
+if (HPX_WITH_TESTS_HEADERS)
+  # TODO: Why is this needed?
+  add_library(hpx_includes INTERFACE)
+  target_include_directories(hpx_includes INTERFACE ${CMAKE_SOURCE_DIR})
```
msimberg: @hkaiser note that I've cheated here! I don't understand why I need this (all the other modules are happy without it). It may be that this problem goes away by itself, but let me at least try once more to get rid of it before we merge this.

msimberg: I still don't get this. Let's go with this for now. It will definitely disappear once all the modules are in place.

msimberg: The reason is the `NOLIBS` option. I'm cleaning this up in #4147.

PR #4137, opened 2019-10-15
Fixing TSS destruction

sithhell:
Deleting TSS before binding the final result of the task function is being set and the coroutine state has been reset.

The thread specific destructors might need to synchronize 'as if' they still run on a valid HPX thread.


libs/coroutines/src/detail/coroutine_impl.cpp
```diff
@@ -96,6 +96,8 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
                         result_last.first == thread_state_enum::terminated);
                 }
 
+                this->reset_tss();
+
```
hkaiser: Shouldn't the rest of `this` be reset somewhere now as well?

hkaiser: Ahh, it is - please ignore me.

hkaiser: But here is a real question: shouldn't the TSS be reset only if the `state == terminated`?

sithhell: https://github.com/STEllAR-GROUP/hpx/pull/4137/files/db357236eaa7cc7b97c682ac9cb0181ceb4cfa15#diff-ae32f64c0aa2f9db307e2e779cf067e3R95

What about this?

msimberg: Did you come to a consensus on whether TSS should be reset if there's an exception?

sithhell: Yes, it should be reset if there is an exception. I just (force) pushed a fix to remedy this.

PR #4138, opened 2019-10-16
Forking boost::intrusive_ptr and adding it as hpx::intrusive_ptr 

hkaiser:
Also creates a module `hpx_memory`

Incidentally, this removes the dependency on `boost::intrusive_ptr`.


libs/memory/include/hpx/memory/serialization/intrusive_ptr.hpp
```diff
@@ -0,0 +1,33 @@
+//  Copyright (c) 2014 Thomas Heller
```
msimberg: This looks identical to `libs/serialization/include/hpx/serialization/intrusive_ptr.hpp`. Shouldn't we keep only the one in the serialization module? This way we can use `intrusive_ptr` without depending on `serialization`.

libs/memory/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  EXCLUDE_FROM_GLOBAL_HEADER
+    "hpx/memory/detail/sp_convertible.hpp"
+    "hpx/memory/serialization/intrusive_ptr.hpp"
+  SOURCES ${memory_sources}
+  HEADERS ${memory_headers}
+  COMPAT_HEADERS ${memory_compat_headers}
+  DEPENDENCIES
+    hpx_config
+    hpx_assertion
+    hpx_serialization
```
msimberg: See my other comment. We should try to not depend on serialization here (although it's a much lighter dependency now!).

libs/memory/include/hpx/memory/serialization/intrusive_ptr.hpp
hkaiser: @msimber The one in the serialization module is for `boost::intrusive_ptr`. Leaving the serialization of `hpx::intrusive_ptr` there as well would make serialization depend on this module (and in the end on any other modules that requires serialization in any way). 

My rationale was to have the serialization module support only 'ambient' data types and to move all serialization of specific HPX types into their respective module. That makes those modules depend on `hpx_serialization`, but I thought that this is not a problem as serialization is a very fundamental functionality anyways. However I could see it being a problem in your case...

I have however excluded the header `memory/serialization/intrusive_ptr.hpp` from the generated module header. We could let this not depend on `hpx_serialization` and the user has to `#include <hpx/memory/serialization/intrusive_ptr.hpp>` explicitly anyways. So we could require for adding a dependency to `hpx_serialization` wherever this header is used as well.

sithhell: I'd see it the other way around. Serialization is an integral part and should be always a dependency. As you mentioned, it should be pretty lightweight now and shouldn't come with much bloat by itself, anything else leads to too much complexity for users.

hkaiser: Well, that's what I said, didn't I? Maybe I mis-phrased things...

msimberg: Thanks @hkaiser. I was simply hoping we could avoid the dependency if possible, but since serialization is lightweight now let's do it like this!

msimberg:
GCC 9 seems to want braces for the `HPX_ASSERT` in case it's empty over here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/logging/include/hpx/logging/writer/named_write.hpp#L337. How it's not complaining about that on master is odd since you haven't changed that line...

libs/memory/tests/unit/intrusive_ptr.cpp
```diff
@@ -0,0 +1,1060 @@
...
+namespace n_assignment {
+
+    void copy_assignment()
+    {
+        HPX_TEST(N::base::instances == 0);
+
+        {
+            hpx::intrusive_ptr<X> p1;
+
+            p1 = p1;
```
msimberg: Clang warns about the self-assignment. We can add the same as here: https://github.com/STEllAR-GROUP/hpx/blob/8e77499f2f243342f13fc546f6e65803d8b2ac94/libs/functional/tests/unit/function/function_ref.cpp#L128-L135.

hkaiser: Will do, thanks!

PR #4139, opened 2019-10-16
Remove runtime includes from coroutines module

msimberg:
These were not supposed to be there anymore after #4099. Nothing from the headers was used though so removing them is enough.


msimberg:
I was wrong... TSS complicates things. I'll try to get this fixed asap.

msimberg:
This is hopefully good now. `tss.cpp` was only using `hpx::threads::get_self_ptr` from outside the module instead of `coroutine_self::get_self`.

PR #4142, opened 2019-10-17
Add a new thread_pool_executor

msimberg:
- Adds overloads to `register_thread/work` that take a `thread_pool_base*` as an argument
- Moves the `register_thread/work` functions to a separate file, deprecates them from the `hpx::applier` namespace with a CMake option
- Adds overloads to `async_launch/post_policy_dispatch` that take a `thread_pool_base*` and a `thread_schedule_hint` as arguments
- Adds a new `thread_pool_executor` that takes a `thread_pool_base` as an argument and dispatches work onto that thread pool. It also does hierarchical scheduling like in `create_thread_hierarchical` in the `future_overhead` test.
- Adds another standalone thread pool example using the `thread_pool_executor` which explicitly sets up a new thread pool and executor, and spawns work using the executor
- Cleans up some additional places that try to access the global runtime (`this_thread_executor`), and some unneeded headers

This will probably need some massaging before it can go in.
- Right now the `post/async` dispatch functions take a subset of the options that `register_thread/work` and `futures_factory::apply` take. Should we go all in and add the remaining options (stack size, priority). We could also just pass a `thread_init_data` much earlier in the process.
- The new `thread_pool_base` argument could eventually be an `execution_resource` (from #4090)? All calls eventually lead to `pool->create_work/thread` anyway.
- `thread_pool_executor` is already taken as a name, but describes the new executor quite well. The other `thread_pool_executor` is the one with embedded schedulers.
- `thread_pool_executor` is very similar to `parallel_policy_executor` and we should avoid duplicating them too much. Should we change the `parallel_policy_executor` to have an optional constructor which takes a `thread_pool_base*`? If it's `nullptr` we fall back to getting the current or default pool?

Finally, the hierarchical spawning in `thread_pool_executor` is horrible on the first iteration, but otherwise consistently 10-20% faster than the current hierarchical spawning:

thread_pool_executor


biddisco:
I don't see the new executor added to the `future_overhead` test? is that an oversight?

msimberg:
> I don't see the new executor added to the `future_overhead` test? is that an oversight?

Maybe... if I move the new hierarchical scheduling into the existing parallel executor we don't really need to add it to the test.

msimberg:
I've renamed the old `thread_pool_executor` to `embedded_thread_pool_executor` without any compatibility options as it was in the detail namespace. The normal usage is through `local_queue_executor` and friends anyway.

msimberg:
I think I've resolved all the conflicts resulting from #4145. Any remaining comments besides the inevitable inspect warnings that I'll get? I've left the new executor as a separate executor for now until I come up with a good way of merging it with `parallel_executor` (and `parallel_executor_aggregated`).

msimberg:
The `daint-gcc-oldest` build problem is unrelated to this PR. I'll take care of that in a separate PR.

hkaiser:
Let's get this merged, LGTM!

PR #4145, opened 2019-10-17
Adding basic support for stackless threads

### LINKED ISSUE #4030: Allow threads to skip context switching
Add an option to invoke HPX threads directly on the worker thread stack instead of switching to a new stack. This would be a useful option for tasks that don't do much, e.g. the tasks in simple data-parallel for loops. This may lead to nice performance improvements in our thread scheduling.

A simple proof-of-concept is probably a good first step to see if we actually gain anything by this.

hkaiser:
Fixes #4030

### DIFF
--- a/libs/coroutines/include/hpx/coroutines/stackless_coroutine.hpp
+++ b/libs/coroutines/include/hpx/coroutines/stackless_coroutine.hpp
@@ -1,0 +1,245 @@
+//  Copyright (c) 2013 Hartmut Kaiser
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0.
+//  (See accompanying file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef HPX_COROUTINE_STACKLESS_COROUTINE_HPP_20130724
+#define HPX_COROUTINE_STACKLESS_COROUTINE_HPP_20130724
+
+#include <hpx/config.hpp>
+#include <hpx/assertion.hpp>
+#include <hpx/coroutines/coroutine.hpp>
+#include <hpx/coroutines/detail/tss.hpp>
+#include <hpx/coroutines/thread_enums.hpp>
+#include <hpx/coroutines/thread_id_type.hpp>
+#include <hpx/functional/detail/reset_function.hpp>
+#include <hpx/functional/unique_function.hpp>
+#include <hpx/type_support/unused.hpp>
+
+#include <cstddef>
+#include <utility>
+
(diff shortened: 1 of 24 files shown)


hkaiser:
@msimberg I think that this requires a bit more experimentation/time. Let's wait with this one until we are sure that we see a gain and do not introduce undue overheads to the common case. The code as is decreases the performance of running 500000 tasks by about 3% compared to master (for the stack-full case), but improves things for the new stack-less threads by 7% (compared to master). I would like to do a better analysis and avoid any decrease in performance compared to what we have.

hkaiser:
FWIW, the latest changes ensure that stack-full threads perform no worse than what we have on master. Stack-less threads give a performance win of about 7%. I think this can go ahead once all tests have been cleaned up.

hkaiser:
@msimberg, @sithhell: this is good to be merged, I think.

libs/coroutines/include/hpx/coroutines/stackless_coroutine.hpp
```diff
@@ -0,0 +1,239 @@
...
+        {
+            HPX_ASSERT(exited());
+
+            util::detail::reset_function(f_);
+
+#if defined(HPX_HAVE_THREAD_PHASE_INFORMATION)
+            phase_ = 0;
+#endif
+#if defined(HPX_HAVE_THREAD_LOCAL_STORAGE)
+            detail::delete_tss_storage(thread_data_);
```
sithhell: Shouldn't we apply the tss fixes to the stackless coroutines as well?

msimberg:
This is extreme bikeshedding, but I think `thread_data_stackfull` should be spelled `thread_data_stackful`. If you don't mind the change I volunteer to do the change.

libs/coroutines/include/hpx/coroutines/stackless_coroutine.hpp
hkaiser: Well, we could, certainly - even if just for consistency. There is no way the TSS problem can happen for stackless threads, I believe.

hkaiser:
> This is extreme bikeshedding, but I think thread_data_stackfull should be spelled thread_data_stackful. If you don't mind the change I volunteer to do the change.

Good point. I'll do that change as I need to touch up on the TSS stuff anyways.

hkaiser:
> What got rid of the added overhead with stackful coroutines that you had initially?

Replacing the `virtual operator()()` with this removed the overhead:
https://github.com/STEllAR-GROUP/hpx/pull/4145/files#diff-11f383c8f29752a630098984712b40bbR586-R593

hkaiser:
@msimberg I have renamed the stackful thread_data as requested.

PR #4147, opened 2019-10-22
Resource partitioner and threadmanager modules

msimberg:
Adds a `resource_partitioner` and a `threadmanager` module.

I've removed the runtime checks from the partitioner completely without adding global functions that do the check (I found having wrappers for some of the partitioner member functions ugly, and doing it for all of them adds tedious boilerplate). Best would probably still be to add overloads to `init`/`start` taking a partitioner, but I may leave that for another PR.

Builds on #4142 since it takes care of certain runtime dependencies for these modules (the diff is very large because of that).


libs/resource_partitioner/src/detail_partitioner.cpp
```diff
@@ -33,27 +32,13 @@ namespace hpx { namespace resource { namespace detail {
...
     {
-        if (get_runtime_ptr() != nullptr)
-        {
-            HPX_THROW_EXCEPTION(bad_parameter, func, message);
-        }
-        else
-        {
-            throw std::invalid_argument(func + ": " + message);
-        }
+        HPX_THROW_EXCEPTION(bad_parameter, func, message);
```
hkaiser: Does this assume to run on an HPX thread? Throwing an `hpx::exception` on a non-HPX thread is generally a bad idea... (it tries to collect a stack-backtrace by spawning a new HPX Thread).

msimberg: You're right, I didn't check this properly. I think this is okay though. Pre-runtime start we only construct a lightweight exception because the handler isn't installed at that point. During runtime, and outside of HPX threads, we construct a full/custom exception but leave out some information that only applies to HPX threads. We do get the stacktrace though, but that is guarded by this: https://github.com/STEllAR-GROUP/hpx/blob/458e59a071c7abd2330f33c10443387088566fec/src/util/backtrace/backtrace.cpp#L410-L411. If anything, the check here (partitioner) should've been with `get_self_ptr`, not `get_runtime_ptr`.

I see we don't print any of the information stored in the exception if it's thrown on a non-HPX thread, but that's orthogonal to this issue. It would certainly be convenient if we were allowed to `HPX_THROW_EXCEPTION`s anywhere, not just on HPX threads, and it seems that is the case.

msimberg:
The reason for my confusion about failing header tests in #4136 was the `NOLIBS` option. Since these modules still depend on the core HPX library I'm leaving out the `NOLIBS` option here (and removing it and the hack for `hpx_compute`  header tests).

Issue #2396, opened 2016-11-16
BOOST_ROOT paths are not used at link time


BryantLam:
This issue was observed with HPX release 0.9.99 and git commit 88af985.

On my system, there are some older Boost libraries installed in /usr/lib64, so I built Boost 1.61.0 in my home directory and used `BOOST_ROOT` at cmake-time to reference it.

At compile time, the `BOOST_ROOT` path is correctly used as `-isystem $BOOST_ROOT/include`. However, at link time, the Boost libraries are referenced (e.g., `-lboost_system`) but the path is not used at all.

The end result creates cryptic undefined-reference linking errors because the older Boost libraries are being used from /usr/lib64. Setting `LIBRARY_PATH=$BOOST_PATH/lib` does not get around this issue because /usr/lib64 is searched before `LIBRARY_PATH` is:
```
gcc --print-search-dirs | tr ':' '\n'
```

I would have expected something like ~~`-L$BOOST_ROOT/lib`~~ an absolute path to the `$BOOST_PATH/lib/<lib.so>` at link time.

BryantLam:
This issue does not occur when `HPX_WITH_STATIC_LINKING=ON`.

hkaiser:
Adding `-DBOOST_LIBRARYDIR=<your_boost_library_dir>` to the cmake command line should fix the issue.

BryantLam:
Unfortunately, setting `BOOST_LIBRARYDIR` did not work. Using `-DBoost_DEBUG:BOOL=ON`, I was able to verify Boost paths were detected correctly.

I noticed that during link time, the last argument is an `-rpath` list that includes jemalloc (so at least that one was detected correctly), but Boost is not in the list. Setting `-DBoost_NO_SYSTEM_PATHS:BOOL=ON` at cmake-time did not help.

This issue may be a problem with cmake/HPX_SetupBoost.cmake. Even if it is not, the general guidance is to not use `LINK_DIRECTORIES()` anymore [[1]](http://cmake.3232098.n2.nabble.com/LINK-DIRECTORIES-td3389697.html)[[2]](http://cmake.3232098.n2.nabble.com/Problem-with-LINK-DIRECTORIES-td6992023.html).

sithhell:
We use target link libraries, which should do the job just fine.
Do you have the environment variables CPATH or CPLUS_INCLUDE_PATH set? If
yes, for some reason, the FindBoost.cmake script doesn't set the path to
[...]

PR #4154, opened 2019-10-24
Added CMake policy CMP0060 for HPX applications.

severinstrobl:
While the policy CMP0060 is activated for building HPX itself, for HPX
applications #2396 can lead to errors during linking if the wrong Boost
installation is picked up by the CMake module.


msimberg:
The pycicle error is most likely due to #4028, which was fixed by #4157, so I'm merging this. Thanks @severinstrobl!

PR #4164, opened 2019-10-29
Affinity and batch environment modules

msimberg:
Builds on #4147.


hkaiser:
@msimberg it is _very_ difficult (close to impossible) to understand the changes in this PR as the formatting changes are intermixed with real edits. Could you try (in the future) to separate those into single commits (i.e. aggregate all commits into s set of dedicated minimal ones)?

msimberg:
@hkaiser absolutely, sorry about that. I'll separate them better in the future. Note that most of the changes come from #4142 and #4147 and the changes in this PR are actually quite minimal. Starting with #4142 should make the changes much more apparent.

msimberg:
@hkaiser this should be good to go now. The main change besides moving files around was moving the detection of MPI to the command line handling module where it can access the runtime configuration. The batch environments are also now passed a flag indicating if MPI is available instead of checking it themselves.

msimberg:
@hkaiser I've reverted the spirit-related formatting changes.

PR #4166, opened 2019-10-29
Add an INTERNAL_FLAGS option to link to hpx_internal_flags

aurianer:
This prevents external projects from inheriting hpx internal compile flags when using `hpx_setup_target`, `add_hpx_component`, `add_hpx_library` or `add_hpx_executable`.


aurianer:
I used a `sed` command to add the `INTERNAL_FLAGS` option, I fixed the indentation problems but I may have missed some

PR #4174, opened 2019-10-31
Reestablish global HPX_LIBRARIES and HPX_INCLUDE_DIRS

### LINKED ISSUE #4163: Add back HPX_LIBRARIES and HPX_INCLUDE_DIRS
We should still try to emulate these in the next release but deprecate the use of them.

aurianer:
- Fixes #4163 

- Fix some examples pseudo dependencies

Note : The `HPX_LIBRARIES` and `HPX_INCLUDE_DIRS` will be visible once #4165 is merged


cmake/HPX_SetupMPI.cmake
```diff
@@ -18,38 +18,30 @@ if(HPX_WITH_NETWORKING AND HPX_WITH_PARCELPORT_MPI)
...
-  if (MPI_CXX_LIBRARIES)
-    if(${CMAKE_VERSION} VERSION_LESS "3.12.0")
-      set_property(TARGET hpx::mpi APPEND PROPERTY
-        INTERFACE_LINK_LIBRARIES ${MPI_CXX_LIBRARIES})
-    else()
-      target_link_libraries(hpx::mpi INTERFACE ${MPI_CXX_LIBRARIES})
-    endif()
-  endif()
-  # Ensure compatibility with older versions
-  if (MPI_LIBRARY)
```
msimberg: Good catch! For others, `MPI_LIBRARY` and `MPI_EXTRA_LIBRARY` have been deprecated since at least CMake 3.5 so we can safely remove these.

msimberg:
I think we'll be better off reusing the functionality that generates the pkgconfig files. The problem with sprinkling `hpx_libraries` into the `HPX_SetupX.cmake` files is that 1) we're bound to miss some, and 2) it doesn't work so nicely when those files get included in `HPXConfig.cmake` (we don't want to be adding to `HPX_LIBRARIES` etc. from that file). If we use `hpx_collect_usage_requirements` the targets will be the truth and everything else is generated from those.

msimberg:
Thinking about this a bit more, since `HPX_LIBRARIES` and `HPX_INCLUDE_DIRS` are only used within CMake (not pkgconfig) we don't need to resolve targets recursively. It's enough if we get the interface properties of the `hpx` target and stick them in those variables.

aurianer:
Replaced by #4176

PR #4179, opened 2019-11-01
Moving zip_iterator and transform_iterator to iterator_support module

hkaiser:


msimberg:
So far these were left out of the `iterator_support` module because they cause a circular dependency (https://140433-4455628-gh.circle-artifacts.com/0/hpx/artifacts/circular_deps.svg). I'm not sure if creating a separate module or trying to resolve the dependency will be better. Ideas?

hkaiser:
> So far these were left out of the iterator_support module because they cause a circular dependency (https://140433-4455628-gh.circle-artifacts.com/0/hpx/artifacts/circular_deps.svg). I'm not sure if creating a separate module or trying to resolve the dependency will be better. Ideas?

The file `hpx/util/detail/reserve.hpp` has two functions: `reserve_if_reservable` and `reserve_if_random_access_by_range`. Only the former is needed by the serialization module and only the latter requires to include the range-related things from util. Thus duplicating (and possibly renaming) the implementation of `reserve_if_reservable` (which is trivial) in the serialization module should do the trick. Thanks for noticing!

libs/iterator_support/CMakeLists.txt
```diff
@@ -33,13 +36,16 @@ add_hpx_module(iterator_support
...
+    "hpx/iterator_support/traits/is_iterator.hpp"
+    "hpx/iterator_support/traits/is_range.hpp"
   SOURCES ${iterator_support_sources}
   HEADERS ${iterator_support_headers}
   COMPAT_HEADERS ${iterator_support_compat_headers}
   DEPENDENCIES
     hpx_config
+    hpx_datastructures
+    hpx_functional
+    hpx_serialization
```
msimberg: There's still what *seems* like a cycle (`iterator_support` - `serialization` - `util` - `iterator_support`) according to the dependencies, but the dependency checker isn't complaining. Either the checker is wrong, or there are unnecessary dependencies in this list.

hkaiser: Thanks for noticing. The dependency of `serialization` on `util` was a leftover which should be gone now.

PR #4181, opened 2019-11-04
Rename base libraries var

aurianer:
- Transform the current `HPX_LIBRARIES` cache variable which was linking only base libraries like `rt`, `dl`, etc. to the `hpx_base_libraries` target leaving the basic `HPX_LIBRARIES` to be used in other projects
- Remove duplicated link to external asio in `HPX_SetupBoost`
- Fix typo that resulted in a bug in `hpx_add_example_target_dependencies`
- Fix some examples cmake dependencies
- Remove the use of `hpx_library_dir` and instead `target_link_directories`
- Refactor CUDA setup


hkaiser:
Wouldn't this break dependent projects that were relying on `HPX_LIBRARIES`?

msimberg:
> Wouldn't this break dependent projects that were relying on `HPX_LIBRARIES`?

Depends what "relying on" means. I always imagined the use case was `(target_)link_libraries(... ${HPX_LIBRARIES})` and that still works perfectly with this. Same for `HPX_INCLUDE_DIRS` etc. (since they will be propagated from the targets in `HPX_LIRBARIES`). If someone was relying on `HPX_LIBRARIES` containing full paths and library names (e.g. for pkgconfig file generation) it won't work. Are you doing that in phylanx? Have we made any promises about what `HPX_LIBRARIES` can be used for?

hkaiser:
@msimberg all is well, I just wanted to make sure this wouldn't break anything down the road. Thanks!

msimberg:
> @msimberg all is well, I just wanted to make sure this wouldn't break anything down the road. Thanks!

Your suspicion is warranted ;) Everything *should* be okay, but I may of course have missed something. We will find out...

aurianer:
Just changed the `HPX_BASE_LIBRARIES` into the target `hpx_base_libraries`

msimberg:
@hkaiser by the way, I missed initially the context for your question. I thought your comment was on #4176. But I'm sure you got by now that renaming or removing `HPX_LIBRARIES` is ok because of #4176...

msimberg:
@aurianer APEX still relies on the `hpx_libraries` function: https://circleci.com/gh/STEllAR-GROUP/hpx/140775?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link. Since APEX is essentially an HPX module now we can probably go and do the necessary changes to APEX itself.

aurianer:
Actually it doesn't on `origin/develop` but I'm not sure which tag the tests are fetching

cmake/HPX_SetupCUDA.cmake
```diff
@@ -0,0 +1,67 @@
...
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+
+if(HPX_WITH_CUDA)
+
+  find_package(CUDA REQUIRED)
+  set(HPX_WITH_COMPUTE ON)
```
msimberg: I wonder if this one has to be `PARENT_SCOPE`... let's give it a try and fix it afterwards if needed.

aurianer: When calling `include(<cmake script>)` the scope stays identical and as `HPX_SetupCuda.cmake` is called in the top `CMakeLists.txt`, I would say there is no `PARENT_SCOPE` ^^

msimberg: Indeed, very good point. Thanks!

PR #4184, opened 2019-11-05
Resetting some cached variables to make sure those are re-filled

hkaiser:
- this helps if the user reuses an existing cmake cache after switching branches


aurianer:
Thanks !

PR #4196, opened 2019-11-08
minor fixes for building on OSX Darwin

khuck:
## Proposed Changes

  - minor change to CMakeLists.txt for link flag
  - std::on_quick_exit() not implemented on OSX, so don't call it.


hkaiser:
> std::on_quick_exit() not implemented on OSX, so don't call it.

@khuck, this is solved in #4189 in a more portable way.

khuck:
@msimberg  OK, but the APPLE - specific CMake change is still required.  Apple doesn't understand --whole-archive...

hkaiser:
> @msimberg OK, but the APPLE - specific CMake change is still required. Apple doesn't understand --whole-archive...

@khuck: yes, certainly. Could you modify this PR such that the linker flag change remains?

khuck:
@hkaiser @msimberg done

src/CMakeLists.txt
```diff
@@ -328,7 +328,11 @@ foreach(_module ${HPX_LIBS})
   if(UNIX OR APPLE)
     # add the full archive (some are not used directly in libhpx and need to be
     # forced into libhpx with this option)
+  if(APPLE)
+    set(_module_target "-Wl,-all_load" "${_module_target}")
+  else(APPLE)
```
msimberg: Sorry for the nitpick. Could we do:
```
if(UNIX)
  set(... whole-archive ...)
elseif(APPLE)
  set(... all_load ...)
endif()
```
instead and skip the nested ifs?

msimberg: Sorry for the nitpick. Could we do:
```
if(UNIX)
  set(... whole-archive ...)
elseif(APPLE)
  set(... all_load ...)
endif()
```
instead and skip the nested ifs?

msimberg:
Thanks @khuck, I didn't realize that option would be different on apple.

src/CMakeLists.txt
khuck: @msimberg done

PR #4201, opened 2019-11-11
Fix APEX build

msimberg:
Adds APEX temporarily as a dependency to the `threadmanager` module.

### DIFF
--- a/libs/compute_cuda/CMakeLists.txt
+++ b/libs/compute_cuda/CMakeLists.txt
@@ -62,3 +62,7 @@ add_hpx_module(compute_cuda
     hpx_serialization
   CMAKE_SUBDIRS examples tests
 )
+
+if (TARGET hpx::apex)
+  target_link_libraries(hpx_compute_cuda PRIVATE hpx::apex)
+endif()
--- a/libs/threadmanager/CMakeLists.txt
+++ b/libs/threadmanager/CMakeLists.txt
@@ -45,3 +45,7 @@ add_hpx_module(threadmanager
     hpx_topology
   CMAKE_SUBDIRS examples tests
 )
+
+if (TARGET hpx::apex)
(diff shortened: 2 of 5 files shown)


libs/threadmanager/CMakeLists.txt
```diff
@@ -45,3 +45,9 @@ add_hpx_module(threadmanager
     hpx_topology
   CMAKE_SUBDIRS examples tests
 )
+
+if (TARGET hpx::apex)
+  target_link_libraries(hpx_threadmanager PRIVATE hpx::apex)
+  get_target_property(_apex_inc_dir hpx::apex INTERFACE_INCLUDE_DIRECTORIES)
+  target_include_directories(hpx_compute INTERFACE "${_apex_inc_dir}")
```
hkaiser: Why not just add `hpx::apex` as a dependency to `hpx_compute` as well?

msimberg: `hpx_compute` already depends on APEX. I'll add it to `hpx_compute_cuda` though.

libs/compute_cuda/CMakeLists.txt
```diff
@@ -62,3 +62,9 @@ add_hpx_module(compute_cuda
     hpx_serialization
   CMAKE_SUBDIRS examples tests
 )
+
+if (TARGET hpx::apex)
+  target_link_libraries(hpx_threadmanager PRIVATE hpx::apex)
```
hkaiser: Is this what you intended? Shouldn't this be `hpx_compute_cuda` instead?

msimberg: Uh, no... Thanks!

msimberg:
@hkaiser I removed the include directories. Things seem to be happy without them. They were probably a leftover from somewhere else.

msimberg:
Fixed in #4214.

PR #4192, opened 2019-11-08
Set up CUDA in HPXConfig.cmake

msimberg:
Makes sure the `hpx::cuda` target is available for dependent projects.

PR #4202, opened 2019-11-11
Fix CUDA configuration

msimberg:
Continues #4192.

@weilewei would you mind trying this out to see if things are better?


weilewei:
@msimberg I tried and it does work! Thanks for the work.

PR #4206, opened 2019-11-12
Runtime configuration module

msimberg:
Adds a `runtime_configuration` module, and moves dependencies around to make it independent:

- `attach_debugger` is now in the debugging module
- Moves setup of some global options (like lock detection) to `hpx_init.cpp`
- Make `trace_on_new_stack` a helper function outside of the `backtrace` class to make it independent of `async`/`futures_factory`

Includes commits from #4164.


msimberg:
@hkaiser please consider this and #4207 as drafts. I'll let you know when I have them actually working.

hkaiser:
@msimberg Couldn't we resolve the dependencies by encapsulating the necessary functionalities in (partially or fully specialized) traits? Things are pulled together in hpx_init, where we could rely on the compiler to pull in and integrate all the pieces without any need for the pieces to depend on each other. If a compile-time is not viable, we could rely on some self-registration mechanism implemented by the parcel-ports (some global object instance that registers callbacks in its constructor).

msimberg:
> @msimberg Couldn't we resolve the dependencies by encapsulating the necessary functionalities in (partially or fully specialized) traits? Things are pulled together in hpx_init, where we could rely on the compiler to pull in and integrate all the pieces without any need for the pieces to depend on each other. If a compile-time solution is not viable, we could rely on some self-registration mechanism implemented by the parcel-ports (some global object instance that registers callbacks in its constructor).

I'm trying to understand if we're actually talking about the same solution without me quite seeing it. To be clear, I think the second solution is actually not so bad. I'm not at all against `hpx_init.cpp` pulling everything together (and effectively depending on all the parts of the runtime). I'm against the different parts of the runtime depending on each other (as are you probably too). For it to be possible for `hpx_init.cpp` to pull everything together it needs to be the *only* place where it can be initialized, not like now where the resource partitioner can initialize parts before `hpx::init` is called.

I wanted to avoid global objects registering callbacks if possible because they're difficult to reason about (and may break things like initializing the runtime with a global object... I'm sure we can make it work, but I'd rather not).

biddisco:
Does this affect the way we setup custom pools? these are done before hpx_init is called, so some initialization of things currently happens out of order...

msimberg:
> Does this affect the way we setup custom pools? these are done before hpx_init is called, so some initialization of things currently happens out of order...

Number 2 would affect the way we set up custom pools. But it would just be a matter of delaying the assignment of hardware to pools by wrapping it in a lambda.

msimberg:
Based on our discussion yesterday with @hkaiser we'll try to go with passing resource partitioner init parameters through `hpx::init/start`. This'll make it easier to collect and handle all runtime dependencies in `hpx_init.cpp`. @hkaiser since this doesn't introduce any circular dependencies among the modules, I'd like to leave refactoring the RP initalization for another PR. Would that be okay with you?

msimberg:
The last formatting commit was quite rough on some files, so it may be easier to look at the previous commits individually.

msimberg:
This would be ready to review now. I'm leaving out the points discussed earlier regarding parcelport configuration options from this PR, but once #4388 is in we can take care of that as well.

PR #4207, opened 2019-11-12
Command line handling module

msimberg:
Adds a command line handling module. To do this I've added a few smaller modules:

- `prefix` handles finding the HPX prefix. This could've lived in `util` but it requires enough other modules and the use case is rather clear that I thought it's worth it to have a separate module for this.
- `asio` contains helper functions built on top of Boost.Asio.
- `version` contains functions to return version information. This required moving all the generated `config_defines.hpp` for each module to the `config` module. Without this the `version` module would depend on *all* other modules. Ideas for other ways to solve this are welcome.

Also changed:
- The `sed_transform` files moved into the `util` module.
- `init_logging`, and MPI and parcelset init moved to `hpx_init.cpp`

Includes commits from #4164 and #4206.


hpx/runtime/threads/policies/thread_queue_mc.hpp
```diff
@@ -16,6 +16,7 @@
 #include <hpx/errors.hpp>
 #include <hpx/functional/function.hpp>
 #include <hpx/runtime/threads/policies/lockfree_queue_backends.hpp>
+#include <hpx/runtime/threads/policies/maintain_queue_wait_times.hpp>
```
hkaiser: Should this also `#include <hpx/runtime/threads/policies/deadlock_detection.hpp>`?

libs/affinity/src/parse_affinity_options.cpp
```diff
@@ -112,56 +105,42 @@ namespace hpx { namespace threads { namespace detail
...
+            spec = qi::uint_ >> -(qi::int_) |
+                partlit("all") >> qi::attr(spec_type::all_entities());
 
             BOOST_SPIRIT_DEBUG_NODES(
-                (start)(mapping)(distribution)(thread_spec)(pu_spec)(specs)
-                (spec)(socket_spec)(core_spec)(processing_unit_spec)
-            );
+                (start)(mapping)(distribution)(thread_spec)(pu_spec)(specs)(
+                    spec)(socket_spec)(core_spec)(processing_unit_spec));
         }
```
hkaiser: Could we please not reformat the spirit parsers?

libs/batch_environments/src/slurm_environment.cpp
```diff
@@ -279,92 +272,90 @@ namespace hpx { namespace util { namespace batch_environments
...
 
-            hostlist =
-                (+ranges)[
-                    phoenix::bind(construct_nodelist(),
-                        phoenix::ref(nodes), qi::_1)
-                ];
+            hostlist = (+ranges)[phoenix::bind(
+                construct_nodelist(), phoenix::ref(nodes), qi::_1)];
 
             nodelist = hostlist % ',';
```
hkaiser: Please do not reformat the Spirit parsers.

libs/command_line_handling/src/command_line_handling.cpp
```diff
@@ -1451,22 +1516,6 @@ namespace hpx { namespace util
...
-            num_localities_ = static_cast<std::size_t>(mpi_environment::size());
-        }
-#endif
-
-        if (num_localities_ != 1 || node_ != 0 || rtcfg_.enable_networking())
-        {
-            parcelset::parcelhandler::init(&argc, &argv, *this);
-        }
-#endif
-
```
hkaiser: Where does this code live now?

libs/affinity/src/parse_affinity_options.cpp
msimberg: My bad. Yes, definitely don't want to reformat them.

libs/batch_environments/src/slurm_environment.cpp
msimberg: Yep. My mistake.

hpx/runtime/threads/policies/thread_queue_mc.hpp
msimberg: Yes, thanks!

libs/command_line_handling/src/command_line_handling.cpp
msimberg: In `hpx_init.cpp`: https://github.com/STEllAR-GROUP/hpx/blob/c6192fff99f1f71dfc37dea8518bd668c46b02d0/src/hpx_init.cpp#L703-L719. It may not be the best final place, but I'd like to separate setting up configuration options and actually acting on them as much as possible.

Clearly this (or #4206) isn't working right though with tests failing so you might want to hold off looking at this too deeply (I meant to make this a draft pull request).

hkaiser: Ok, thanks!

hpx/runtime.hpp
```diff
@@ -355,6 +356,19 @@ namespace hpx
...
+        // retrieve the command line arguments for the current locality
+        HPX_API_EXPORT bool retrieve_commandline_arguments(
+            hpx::program_options::options_description const& app_options,
+            hpx::program_options::variables_map& vm);
+
+        ///////////////////////////////////////////////////////////////////////////
+        // retrieve the command line arguments for the current locality
+        HPX_API_EXPORT bool retrieve_commandline_arguments(
+            std::string const& appname, hpx::program_options::variables_map& vm);
+    }    // namespace util
```
hkaiser: Wouldn't these function declarations belong into the new module command_line handling?

hpx/runtime/threads/policies/maintain_queue_wait_times.hpp
```diff
@@ -0,0 +1,19 @@
...
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef HPX_RUNTIME_THREADS_POLICIES_MAINTAIN_QUEUE_WAIT_TIMES_HPP
+#define HPX_RUNTIME_THREADS_POLICIES_MAINTAIN_QUEUE_WAIT_TIMES_HPP
+
+#include <hpx/config.hpp>
+
+namespace hpx { namespace threads { namespace policies {
+#ifdef HPX_HAVE_THREAD_MINIMAL_DEADLOCK_DETECTION
+    extern bool minimal_deadlock_detection;
```
hkaiser: This seems to declare the wrong variable.

libs/debugging/include/hpx/debugging/attach_debugger.hpp
```diff
@@ -0,0 +1,20 @@
...
+#ifndef HPX_DEBUGGING_ATTACH_DEBUGGER_HPP
+#define HPX_DEBUGGING_ATTACH_DEBUGGER_HPP
+
+#include <hpx/config.hpp>
+#include <string>
+
+namespace hpx { namespace util {
+    /// Tries to break an attached debugger, if not supported a loop is
+    /// invoked which gives enough time to attach a debugger manually.
+    void HPX_EXPORT attach_debugger();
```
hkaiser: Nitpick: we normally write `HPX_EXPORT void attach_debugger();` Let's try to be consistent.

src/hpx_init.cpp
```diff
@@ -657,6 +660,54 @@ namespace hpx
...
+                            static_cast<std::size_t>(mpi_environment::size());
+                    }
+#endif
+
+                    if (cms.num_localities_ != 1 || cms.node_ != 0 ||
+                        cms.rtcfg_.enable_networking())
+                    {
+                        parcelset::parcelhandler::init(&argc, &argv, cms);
+                    }
+#endif
```
hkaiser: Could this be moved into a separate function, please?

src/runtime/threads/policies/maintain_queue_wait_times.cpp
```diff
@@ -0,0 +1,18 @@
+//  Copyright (c) 2005-2017 Hartmut Kaiser
+//  Copyright (c)      2011 Bryce Adelstein-Lelbach
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/config.hpp>
+
```
hkaiser: This should #include the corresponding header file.

src/runtime/threads/policies/deadlock_detection.cpp
```diff
@@ -0,0 +1,18 @@
+//  Copyright (c) 2005-2017 Hartmut Kaiser
+//  Copyright (c)      2011 Bryce Adelstein-Lelbach
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/config.hpp>
+
```
hkaiser: This should #include the corresponding header file.

cmake/HPX_PrintSummary.cmake
```diff
@@ -62,12 +62,16 @@ function(create_configuration_summary message module_name)
...
+  if("${module_name}" STREQUAL "hpx")
     set(_base_dir_local "hpx/config")
     set(_base_dir "hpx/config")
     set(_template "config_defines_strings.hpp.in")
   else()
-    set(_base_dir_local "libs/${module_name}/include/hpx/${module_name}/config/")
+    # NOTE: We generate the config_strings.hpp files for each module into the
+    # config module directory. This ensures that modules like version can access
+    # the config strings of all modules, instead of making the version module
+    # depend on all other modules.
```
hkaiser: Wouldn't it be conceptually correct for the version module to depend on all other modules it pulls configuration information from?

hpx/runtime.hpp
msimberg: They've intentionally been moved out of the module because they are free functions that access the runtime and command line handling through the global runtime object. There could be a version of this function that is a member function of `runtime_configuration`, but this specific function belongs close to the runtime.

cmake/HPX_PrintSummary.cmake
msimberg: It would be nice conceptually, but the version module (really, the configuration module) has to depend on *all* other modules to get their options. We can't do this without creating cycles (e.g. `runtime_configuration -> version -> config -> runtime_configuration`), at least with the current setup. We could probably do it with some sort of more dynamic setup.

msimberg:
This is also ready for another round of reviews.

Issue #3818, opened 2019-04-23
Transitive HPX target


jjwilke:
Instead of exporting HPX_LIBRARIES and HPX_INCLUDE_DIRS, consider having target hpx transitively include these. This would more closely follow "modern cmake" where I just:

````
find_package(HPX)
target_link_libraries(target hpx)
````
and I'm done.

msimberg:
This is long overdue and would be very nice to have. It would allow us to get rid of the special `add_hpx_executable` as well.

msimberg:
By the way, if anyone feels like doing this: we actually already do provide targets, they just don't work correctly. In particular, they seem to be missing include directories (probably something else as well). But this would probably not be as tough to do as I imagined at first.

msimberg:
#4028 will get us most of the way there. Some more fixing and testing will likely be required but most of the work will be there to make this work.

aurianer:
On it !

msimberg:
This is effectively possible right now, but I'd ask anyone itching for this to not rely on the current targets just yet. For libraries one can link to the `hpx` target and for applications one can link to `hpx_init hpx_wrap` and things should just work. For components one needs to manually set a few compile definitions which would normally be taken care of by the setup macros (`hpx_setup_target`).

While we'll probably keep the `hpx`, `hpx_init` and `hpx_wrap` targets around we should still do some bikeshedding on the targets. Presumably we'd want to namespace all public targets with `HPX::` (note, all caps as in the package name). We already have two helper targets `hpx::application` and `hpx::component` for dealing with pkgconfig but we'd likely want something similar for CMake users.

* `HPX::hpx` for libraries: no magic, just the base library
* `HPX::executable`: links with `hpx`, `hpx_wrap`, and `hpx_init` (executable, and not application, because of `add_executable`). Alternatively, merge `hpx_init` into `hpx` and require explicit linking with `hpx_wrap`/`HPX::wrap_main` when using the `hpx_main/main` wrapping magic. This way executables can also just link to `HPX::hpx` unless they want the extra magic.
* `HPX::component`: is anyone aware of a way to do e.g. `target_compile_definitions(mytarget HPX_APPLICATION_NAME=mytarget)` but with a target, i.e. `target_link_libraries(mytarget magic_target_that_can_introspect_mytarget)`?
* `HPX::plugin`: same as above but for plugins

[...]

PR #4208, opened 2019-11-12
Add release notes for 1.4.0

msimberg:
Adds release notes for the 1.4.0 release.

Note: they are currently just in the form of bullet points. Everyone, please have a look and let me know if you'd like to have something added (or removed) to the list.

### DIFF
--- a/docs/sphinx/releases/whats_new_1_4_0.rst
+++ b/docs/sphinx/releases/whats_new_1_4_0.rst
@@ -14,11 +14,492 @@
 General changes
 ===============
 
+* We have added the collectives ``all_to_all`` and ``all_reduce``.
+* We have added APIs for resiliency, which allows replication and replay for
+  failed tasks. See the :ref:`documentation <libs_resiliency>` for more details.
+* Components can now be checkpointed.
+* Performance improvements to schedulers and coroutines. A significant change is
+  the addition of stackless coroutines. These are to be used for tasks that do
+  not need to be suspended and can reduce overheads noticeably in applications
+  with short tasks. A stackless coroutine can be created with the new stack size
+  ``thread_stacksize_nostack``.
+* We have added an implementation of ``unique_any``, which is a non-copyable
+  version of ``any``.
+* The ``shared_priority_queue_scheduler`` has been improved. It now has lower
+  overheads than the default scheduler in many situations. Unlike the default
+  scheduler it fully supports NUMA scheduling hints. Enable it with the command
+  line option :option:`--hpx:queuing`\ ``=shared-priority``.
+* We have added the performance counters ``background-receive-duration`` and
+  ``background-receive-overhead`` for inspecting the time and overhead spent on
+  receiving parcels in the background.
+* Compilation time has been further improved when ``HPX_WITH_NETWORKING=OFF``.
+* We no longer require compiled Boost dependencies in certain configurations.
+  This requires at least Boost 1.70, compiling on x86 with GCC 9, clang (libc++)
+  9, or VS2019 in C++17 mode. The dependency on Boost.Filesystem can explicitly
+  be turned on with ``HPX_FILESYSTEM_WITH_BOOST_FILESYSTEM_COMPATIBILITY=ON``
+  (it is off by default if the standard library supports ``std::filesystem``).
+  Boost.ProgramOptions has been copied into the HPX repository. We have a
+  compatibility layer for users who must explicitly use Boost.ProgramOptions
+  instead of the ProgramOptions provided by HPX. To remove the dependency
+  ``HPX_PROGRAM_OPTIONS_WITH_BOOST_PROGRAM_OPTIONS_COMPATIBILITY`` must be
+  explicitly set to ``OFF``. This option will be removed in a future release. We
+  have also removed several other header-only dependencies on Boost.
+* It is now possible to use the process affinity mask set by tools like
+  ``numactl`` and various batch environments with the command line option
+  :option:`--hpx:use-process-mask`. Enabling this option implies
+  :option:`--hpx:ignore-batch-env`.
+* It is now possible to create standalone thread pools without starting the
+  runtime. See the ``standalone_thread_pool_executor.cpp`` test in the
+  ``execution`` module for an example.
+* Tasks annotated with :cpp:class:`hpx::util::annotated_function` now have their
+  correct name when using APEX to generate OTF2 files.
+* Cloning of APEX was defective in previous releases (it required manual
+  intervention to check out the correct tag or branch). This has been fixed.
+* The option ``HPX_WITH_MORE_THAN_64_THREADS`` is now ignored and will be
+  removed in a future release. The value is instead derived directly from
+  ``HPX_WITH_MAX_CPU_COUNT`` option.
+* We have deprecated compiling in C++11 mode. The next release will require a
+  C++14 capable compiler.
+* We have deprecated support for the Vc library. This option will be replaced
+  with SIMD support from the standard library in a future release.
+* We have significantly refactored our CMake setup. This is intended to be a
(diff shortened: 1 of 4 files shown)


msimberg:
Expanded a bit on the points. @hkaiser, @aserio could I ask you to specifically have a look at the collectives and component checkpointing points. I feel like I'm slightly underselling them... Are they even accurate?

@aserio could I ask you to update the grant information in the README (if there's anything to update)? You can push it here or do it in a separate PR, either is fine.

msimberg:
This also includes new contributors in `people.rst` now.

docs/sphinx/releases/whats_new_1_4_0.rst
```diff
@@ -14,11 +14,59 @@
...
 ===============
 
+* We have added the collectives ``all_to_all`` and ``all_reduce``.
+* Components can now be checkpointed.
+* We have added the performance counters ``background-receive-duration`` and
+  ``background-receive-overhead`` for inspecting the time and overhead spent on
+  receiving parcels in the backgroun d.
+* Compilation time has been further improved when ``HPX_HAVE_NETWORKING=OFF``.
+* We no longer require compiled Boost dependencies in certain configurations.
+  This requires at least Boost 1.70, compiling on x86 with GCC 9 or newer in
```
hkaiser: VS2019 in C++17 mode supports filesystem as well.

docs/sphinx/releases/whats_new_1_4_0.rst
```diff
@@ -14,11 +14,59 @@
 General changes
 ===============
 
+* We have added the collectives ``all_to_all`` and ``all_reduce``.
+* Components can now be checkpointed.
+* We have added the performance counters ``background-receive-duration`` and
+  ``background-receive-overhead`` for inspecting the time and overhead spent on
+  receiving parcels in the backgroun d.
```
hkaiser: There is a stray space in the last word.

docs/sphinx/releases/whats_new_1_4_0.rst
```diff
@@ -14,11 +14,59 @@
...
+  with SIMD support from the standard library in a future release.
+* We have significantly refactored our CMake setup. This is intended to be a
+  non-breaking change and will allow for using HPX through CMake targets in the
+  future.
+* We have continued modularizing the HPX library. In the process we have
+  rearranged many header files into module-specific directories. All moved
+  headers have compatibility headers which forward from the old location to the
+  new location, together with a deprecation warning. The compatibility headers
+  will eventually be removed.
+
```
hkaiser: Here is a list of more changes (in no particular order): stack-less coroutines, unique_any and friends, SPDX licensing tags, reworking cmake (targets!), clang-formatting modules, resiliency APIs, performance improvements (mostly scheduler, etc.), there might be more...

Also, what compilers are deprecated with this release? gcc v5, clang v???

docs/sphinx/releases/whats_new_1_4_0.rst
msimberg: Good catch!

docs/sphinx/releases/whats_new_1_4_0.rst
msimberg: I've missed so many things (big release!). Will definitely add the stackless coroutines, unique_any etc., resiliency, the performance improvements, and SPDX license tag. I've mentioned the CMake targets but don't want to advertize it too much before #3818. Is clang-formatting relevant to users?

hkaiser: > Is clang-formatting relevant to users?

Probably not so much if taken in isolation. It's more of a QOI issue that underlines our efforts towards a homogeneous user experience.

msimberg: Fair point. I'll add it. Thanks!

hkaiser: @msimberg Should we add a (small) paragraph elaborating on the modularization efforts? I mean talking about non-cyclicity, maintainability, the possibility to use separate modules without pulling in all of HPX (serialization!), etc.. We might want to outline the overall goal of this refactoring, the possible (and already real) benefits of going through this. The modularization is a major part of this release, it will also be _the_ motivation for releasing HPX V2.0 (once done).

msimberg: Yes, but no!? It's important but I don't think it belongs in the release notes in that detail. There is a note on modularization and the practical implications of headers moving around. Again I don't want to advertise the targets too much before we finalize the names, namespacing, static/shared libraries, granularity etc. unless we make it clear that it's "experimental" in the sense that things might still move around. We should definitely advertise the whole effort as much as possible though in the release announcement and everywhere else! It's probably even worthy of a blog post on its own. What do you think?

aserio:
> 
> 
> Expanded a bit on the points. @hkaiser, @aserio could I ask you to specifically have a look at the collectives and component checkpointing points. I feel like I'm slightly underselling them... Are they even accurate?
> 
> @aserio could I ask you to update the grant information in the README (if there's anything to update)? You can push it here or do it in a separate PR, either is fine.

I have looked over the grants listed and it looks good to me. @msimberg Do you want to add CSCS in the first paragraph of the Acknowledgement section (both in the ReadMe and the People section)?

docs/sphinx/releases/whats_new_1_4_0.rst
hkaiser: > It's probably even worthy of a blog post on its own. 

Absolutely! Otherwise, I'm fine with not expanding it here too much.

msimberg:
> > Expanded a bit on the points. @hkaiser, @aserio could I ask you to specifically have a look at the collectives and component checkpointing points. I feel like I'm slightly underselling them... Are they even accurate?
> > @aserio could I ask you to update the grant information in the README (if there's anything to update)? You can push it here or do it in a separate PR, either is fine.
> 
> I have looked over the grants listed and it looks good to me. @msimberg Do you want to add CSCS in the first paragraph of the Acknowledgement section (both in the ReadMe and the People section)?

Thanks @adrian. I think we're mentioned in both sections already and that's fine by us. Feel free to move things around if you feel it's appropriate though.

msimberg:
@hkaiser (@biddisco, @sithhell) regarding dropping support for compilers in this version:

For the previous release I went with the policy of dropping compilers whose *last* release (i.e. counting all the patch releases) was more than three years ago. We may want to become a bit stricter and make that three years since the first release. This would drop GCC 5 and 6 for this release. On one hand I like to be conservative with this if there's no need to drop support for compilers. On the other hand GCC 7 is the default on e.g. Ubuntu LTS, and more importantly GCC 4-6 in particular has not stopped surprising us with odd behaviour.

Opinions? I'm leaning towards dropping GCC 5 and 6.

docs/sphinx/releases/whats_new_1_4_0.rst
```diff
@@ -14,11 +14,73 @@
...
+  the addition of stackless coroutines. These are to be used for tasks that do
+  not need to be suspended and can reduce overheads noticeably in applications
+  with short tasks. A stackless coroutine can be created with the new stack size
+  ``thread_stacksize_nostack``.
+* We have added an implementation of ``unique_any``, which is a non-copyable
+  version of ``any``.
+* We have added the performance counters ``background-receive-duration`` and
+  ``background-receive-overhead`` for inspecting the time and overhead spent on
+  receiving parcels in the background.
+* Compilation time has been further improved when ``HPX_HAVE_NETWORKING=OFF``.
```
hkaiser: I think this should be `HPX_WITH_NETWORKING=OFF` instead.

msimberg:
Added a note on static linking not working.

PR #4209, opened 2019-11-12
Fix CUDA 10 build

msimberg:
Disables `handle_local_exception` code from being compiled in device mode. Flybys:
- Use CUDA's `assert` for handling assertions on device (no custom message, but prints file and location)
- Change a couple of `register_*` calls from the `hpx::applier` namespace to `hpx::threads`

### DIFF
--- a/libs/algorithms/include/hpx/parallel/util/detail/handle_local_exceptions.hpp
+++ b/libs/algorithms/include/hpx/parallel/util/detail/handle_local_exceptions.hpp
@@ -10,6 +10,7 @@
 #define HPX_PARALLEL_UTIL_DETAIL_HANDLE_LOCAL_EXCEPTIONS_OCT_03_2014_0142PM
 
 #include <hpx/config.hpp>
+#include <hpx/assertion.hpp>
 #include <hpx/async.hpp>
 #include <hpx/errors.hpp>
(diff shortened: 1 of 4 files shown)


libs/algorithms/include/hpx/parallel/util/detail/handle_local_exceptions.hpp
```diff
@@ -144,46 +164,66 @@ namespace hpx { namespace parallel { namespace util { namespace detail {
...
+#else
             hpx::terminate();
+#endif
         }
 
         HPX_NORETURN static void call(
             std::exception_ptr const&, std::list<std::exception_ptr>&)
         {
+#if defined(HPX_COMPUTE_DEVICE_CODE)
+            HPX_ASSERT(false);
```
hkaiser: We might want to call `std::terminate()` here (is there such thing in device code?).

msimberg: As far as I know, no. I think an assertion is the closest we can get that kind of behaviour (with the downside that it's only enabled in debug mode). This fails fairly gracefully with the messages being printed and the next device synchronization failing (which we turn into an exception).

PR #4219, opened 2019-11-15
Assorted small changes

msimberg:
- Changes as many `HPX_TEST`s to more specific `HPX_TEST_EQ/NEQ/LT/LTE/RANGE` as possible
- Removes the `u` suffix from the version components (we end up with `1u.4u.0u` strings in some places)
- Increases the default backtrace depth to 20 to capture more useful frames (the last 5 are usually just error handling)

Not tested so will likely still need some fixing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -550,8 +550,8 @@ if(HPX_WITH_STACKTRACES OR HPX_WITH_THREAD_BACKTRACE_ON_SUSPENSION)
   endif()
 
   hpx_option(HPX_WITH_THREAD_BACKTRACE_DEPTH STRING
-    "Thread stack back trace depth being captured (default: 5)"
-    "5"
+    "Thread stack back trace depth being captured (default: 20)"
+    "20"
     CATEGORY "Thread Manager" ADVANCED)
   hpx_add_config_define(HPX_HAVE_THREAD_BACKTRACE_DEPTH
     ${HPX_WITH_THREAD_BACKTRACE_DEPTH})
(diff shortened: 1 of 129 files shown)


CMakeLists.txt
```diff
@@ -528,8 +528,8 @@ if(HPX_WITH_STACKTRACES OR HPX_WITH_THREAD_BACKTRACE_ON_SUSPENSION)
   endif()
 
   hpx_option(HPX_WITH_THREAD_BACKTRACE_DEPTH STRING
-    "Thread stack back trace depth being captured (default: 5)"
-    "5"
+    "Thread stack back trace depth being captured (default: 20)"
+    "20"
```
hkaiser: I wanted to apply this change for quite some time, thanks!

msimberg:
This should be ready to go now. The only thing I'm unsure about is the removal of the `u` suffix from the version numbers. I have a feeling they were added to suppress warnings about implicit conversions, but most likely only on Windows. Travis is happy, but then I don't know if you Windows people run with more warnings enabled usually. @hkaiser would you (or someone else using Windows) know or have time to check if the removal of the `u` suffixes produces warnings for you?

hkaiser:
> would you (or someone else using Windows) know or have time to check if the removal of the u suffixes produces warnings for you?

@msimberg will do, thanks!

msimberg:
@hkaiser do you think we can go ahead with this or would you still like to check what the removal of the `u` suffix does on Windows?

msimberg:
I'm going to go ahead with this one now. @hkaiser please let me know if this breaks things for you and I'll fix them (hopefully the GH actions builders will tell us soon).

PR #4224, opened 2019-11-18
Deprecate verbs parcelport

msimberg:
Flyby: fixes verbs and other parcelport related options.

### DIFF
--- a/cmake/toolchains/BGION-gcc.cmake
+++ b/cmake/toolchains/BGION-gcc.cmake
@@ -41,8 +41,8 @@ set(HPX_PLATFORM "native")
 # Disable generic coroutines (and use posix version)
 set(HPX_WITH_GENERIC_CONTEXT_COROUTINES OFF CACHE BOOL "disable generic coroutines")
 
-# BGAS nodes support ibverbs
-set(HPX_WITH_PARCELPORT_IBVERBS ON CACHE BOOL "")
+# BGAS nodes support ibverbs, but it is deprecated
+set(HPX_WITH_PARCELPORT_VERBS OFF CACHE BOOL "")
 
 # Always disable the tcp parcelport as it is non-functional on the BGQ.
 set(HPX_WITH_PARCELPORT_TCP ON CACHE BOOL "")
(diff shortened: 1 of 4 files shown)


cmake/toolchains/BGION-gcc.cmake
```diff
@@ -41,8 +41,8 @@ set(HPX_PLATFORM "native")
 # Disable generic coroutines (and use posix version)
 set(HPX_WITH_GENERIC_CONTEXT_COROUTINES OFF CACHE BOOL "disable generic coroutines")
 
-# BGAS nodes support ibverbs
-set(HPX_WITH_PARCELPORT_IBVERBS ON CACHE BOOL "")
+# BGAS nodes support ibverbs, but it is deprecated
+set(HPX_WITH_PARCELPORT_VERBS OFF CACHE BOOL "")
```
hkaiser: Should we enable the MPI parcelport by default instead?

cmake/toolchains/XeonPhi.cmake
```diff
@@ -37,7 +37,7 @@ set(CMAKE_CROSSCOMPILING ON)
 set(HPX_PLATFORM "XeonPhi")
 
 # Always disable the ibverbs parcelport as it is non-functional on the BGQ.
-set(HPX_WITH_PARCELPORT_IBVERBS OFF CACHE BOOL "Enable the ibverbs based parcelport. This is currently an experimental feature")
+set(HPX_WITH_PARCELPORT_VERBS OFF CACHE BOOL "Enable the ibverbs based parcelport. This is currently an experimental feature")
```
hkaiser: Same here, should we enable MPI instead?

biddisco:
just a comment. at some point we will remove the parcelport completely, however, it would be nice if we created an `hpx/obsolete` folder/module that contained useful, but no longer functional or tested code so that if we ever need it again, it was somewhat easier to find than going through old branches etc. (comments?)

sithhell:
Why not just use tags? Or create a wiki page pointing to the commit before it got removed.

biddisco:
compromise...?
in `hpx/obsolete/obsolete.txt`
Parcelport Verbs was removed in git SHA  f47b3814bf5db1ec483611397708f40f4cb57db3
(or maybe - the last commit before it was removed is .....)

PR #4230, opened 2019-11-19
Add OS X builds to Travis CI

msimberg:
Travis CI has a timeout of 50 minutes per build which is not enough to build everything, but this will at least build the core libraries and perhaps examples.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,9 @@
 
 language: cpp
 
-os: windows
+os:
+  - windows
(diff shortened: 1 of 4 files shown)


msimberg:
> LGTM, except perhaps for the names of the new scripts introduced. Could those be a tad more descriptive, please?

Yeah, sorry about that. I matched the names with the stages of the travis config. I'll change them to describe what they actually do.

PR #4231, opened 2019-11-19
Module local lcos

aurianer:
Add `static_reinit`, `synchronization` and `local_lcos` module, and move some dependencies to make it independent:
- move `static_reinit`, `reinitializable` into a `static_reinit` module
- add a `synchronization` module with all the synchronization primitves which do not return futures
- add a `local_lcos` with lcos returning futures
- move `first_argument` into the `functional` module
- remove some unused headers


libs/functional/CMakeLists.txt
```diff
@@ -21,14 +21,17 @@ set(functional_headers
...
   hpx/functional/function.hpp
   hpx/functional/function_ref.hpp
   hpx/functional/invoke.hpp
   hpx/functional/invoke_fused.hpp
   hpx/functional/mem_fn.hpp
   hpx/functional/one_shot.hpp
   hpx/functional/protect.hpp
+  hpx/functional/reinitializable_static.hpp
   hpx/functional/result_of.hpp
+  hpx/functional/static_reinit.hpp
```
msimberg: `first_argument` fits here nicely but `reinitializable_static` and `static_reinit` not so much. Maybe add them to a module `hpx_static_reinit`? `hpx_reinitialiazable_static` is a bit of a mouthful but also works. `hpx_static` could have too many other meanings...

libs/local_lcos/include/hpx/local_lcos.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+#include <hpx/lcos/future.hpp>
+#include <hpx/local_lcos/and_gate.hpp>
+#include <hpx/local_lcos/packaged_continuation.hpp>
+#include <hpx/local_lcos/packaged_task.hpp>
+#include <hpx/local_lcos/promise.hpp>
+#include <hpx/local_lcos/receive_buffer.hpp>
+#include <hpx/local_lcos/trigger.hpp>
+
+#endif
```
msimberg: I think we can generate this file now, but keep the original `hpx/include/local_lcos.hpp` in place. So far we've kept all the `hpx/include/*` headers in place in the main include directory. I think once we've moved everything we can go and deprecate that directory separately.

libs/local_lcos/include/hpx/local_lcos/packaged_continuation.hpp
```diff
@@ -1051,55 +1000,54 @@ namespace hpx { namespace lcos { namespace detail
...
+    //            typename traits::detail::shared_state_ptr_for<inner_type>::type
+    //                inner_shared_ptr_type;
+    //        return inner_shared_ptr_type(
+    //            traits::future_access<inner_type>::detach_shared_state(
+    //                std::move(f)),
+    //            false);
+    //    }
+    //
+    //    return unwrap_impl(std::move(fut), ec);
+    //}
```
msimberg: @hkaiser do you know if we can get rid of this?

libs/local_lcos/include/hpx/local_lcos/recursive_mutex.hpp
```diff
@@ -164,21 +164,21 @@ namespace hpx { namespace lcos { namespace local
...
+            //    ::boost::system_time const& target)
+            //{
+            //    if (mtx.timed_lock(target))
+            //    {
+            //        locking_thread_id.exchange(current_thread_id);
+            //        recursion_count.store(1);
+            //        return true;
+            //    }
+            //    return false;
+            //}
```
msimberg: @hkaiser same question with these: could we remove them? They seem to rely on boost functionality that we use nowhere else (and is perhaps even deprecated).

libs/local_lcos/tests/unit/CMakeLists.txt
```diff
@@ -0,0 +1,113 @@
...
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+set(tests
+  apply_local
+  apply_local_executor
+  async_local
+  async_local_executor
```
msimberg: The `apply` and `async` tests can stay I think. They'll go in a separate module.

libs/local_lcos/CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
...
+  hpx/local_lcos/channel.hpp
+  hpx/local_lcos/composable_guard.hpp
+  hpx/local_lcos/conditional_trigger.hpp
+  hpx/local_lcos/condition_variable.hpp
+  hpx/local_lcos/counting_semaphore.hpp
+  hpx/local_lcos/detail/condition_variable.hpp
+  hpx/local_lcos/detail/counting_semaphore.hpp
+  hpx/local_lcos/detail/sliding_semaphore.hpp
+  hpx/local_lcos/event.hpp
+  hpx/local_lcos/futures_factory.hpp
```
msimberg: Can or should `hpx/future.hpp` also be in this module?

aurianer: I'll do a `future` module, I removed `futures_factory.hpp` from this module, that was a mistake ^^

hkaiser:
What's the rationale of deprecating `receive_buffer`?

msimberg:
> What's the rationale of deprecating `receive_buffer`?

That's on me. It's used in exactly one example, and nowhere in the main codebase. Is it useful to keep around? I'm easy to convince to keep it; I just don't want to keep it around for the sake of keeping it around.

hkaiser:
> > What's the rationale of deprecating `receive_buffer`?
>
> That's on me. It's used in exactly one example, and nowhere in the main codebase. Is it useful to keep around? I'm easy to convince to keep it; I just don't want to keep it around for the sake of keeping it around.

We could make it part of the example. That would keep it around, but remove the need for us to document or maintain it...

msimberg:
> > > What's the rationale of deprecating `receive_buffer`?
> > 
> > 
> > That's on me. It's used in exactly one example, and nowhere in the main codebase. Is it useful to keep around? I'm easy to convince to keep it; I just don't want to keep it around for the sake of keeping it around.
> 
> We could make it part of the example. That would keep it around, but remove the need for us to document or maintain it...

Fair enough! That sounds like a good compromise.

aurianer:
> > > > What's the rationale of deprecating `receive_buffer`?
> > > 
> > > 
> > > That's on me. It's used in exactly one example, and nowhere in the main codebase. Is it useful to keep around? I'm easy to convince to keep it; I just don't want to keep it around for the sake of keeping it around.
> > 
> > 
> > We could make it part of the example. That would keep it around, but remove the need for us to document or maintain it...
> 
> Fair enough! That sounds like a good compromise.

It's actually used in `channel.hpp` so we'll keep it around sorry ^^

aurianer:
To remove the circular dependency between the `local_lcos` and the `execution` module, we need to split `local_lcos` into "synchronization lcos" the others (the ones which returns futures).
@hkaiser would you have a good name for those ?

hkaiser:
> would you have a good name for those ?

@aurianer how about `sync` or `synchronization` for the low-lever stuff and `local_lcos` for the ones that depend on `hpx::future`

libs/local_lcos/include/hpx/local_lcos/recursive_mutex.hpp
hkaiser: I think those were meant to provide a compatibility layer with Boost timing facilities. We're past that stage, so let's remove them.

hpx/include/traits.hpp
```diff
@@ -54,13 +54,13 @@
...
-#include <hpx/traits/is_future.hpp>
+#include <hpx/is_instance/traits/is_executor.hpp>
+#include <hpx/is_instance/traits/is_executor_parameters.hpp>
+#include <hpx/is_instance/traits/is_future.hpp>
 #include <hpx/traits/is_future_range.hpp>
 #include <hpx/traits/is_future_tuple.hpp>
-#include <hpx/traits/is_launch_policy.hpp>
-#include <hpx/traits/is_timed_executor.hpp>
+#include <hpx/is_instance/traits/is_launch_policy.hpp>
+#include <hpx/is_instance/traits/is_timed_executor.hpp>
```
msimberg: I think all five of these traits should stay where they are. They belong with the executor/future (`execution`)  module.

aurianer: Yep I revert these

aurianer:
There is still some dependencies to some main headers but they will move in the execution module or in the threading module :)

change_includes.sh
```diff
@@ -15,10 +15,15 @@
 # specified in the deprecation message in include_compatibility (so it is better
```
hkaiser: I'd like for this file to be moved to the `tools/` subdirectory, if possible.

hpx/util/annotated_function.hpp
```diff
@@ -18,10 +18,7 @@
 #include <hpx/type_support/decay.hpp>
 #include <hpx/util/thread_description.hpp>
 
-#if HPX_HAVE_ITTNOTIFY != 0
-#include <hpx/runtime/get_thread_name.hpp>
-#include <hpx/concurrency/itt_notify.hpp>
```
msimberg: I think the `itt_notify.hpp` include still has to stay.

libs/synchronization/docs/index.rst
```diff
@@ -0,0 +1,18 @@
...
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+.. _libs_synchronization:
+
+===============
+synchronization
+===============
+
+TODO: High-level description of the library.
```
msimberg: Let's add a sentence here about what this module does.

libs/local_lcos/docs/index.rst
```diff
@@ -0,0 +1,18 @@
...
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+.. _libs_local_lcos:
+
+==========
+local_lcos
+==========
+
+TODO: High-level description of the library.
```
msimberg: Same here, let's add at least a high-level description of what the module does.

libs/synchronization/tests/unit/CMakeLists.txt
```diff
@@ -3,3 +3,96 @@
...
+  condition_variable
+  counting_semaphore
+  local_latch
+  local_barrier
+  local_barrier_count_up
+  local_barrier_reset
+  local_dataflow
+  local_dataflow_boost_small_vector
+  local_dataflow_executor
+  local_dataflow_std_array
```
msimberg: The dataflow tests should maybe not be in this module?

msimberg:
Just a couple of things from me as well, but looks good otherwise!

libs/synchronization/tests/unit/CMakeLists.txt
aurianer: Yeah I wasn't sure about those I'll leave them where they were !

aurianer:
Ok all should be fine now :)

msimberg:
I've just pushed a commit to turn on compatibility headers in the `local_lcos` module that was causing the APEX build to fail. Merging now.

hkaiser:
@aurianer, @msimberg Thanks everybody for this! Very nice!

PR #4233, opened 2019-11-20
Remove project() from modules + fix CMAKE_SOURCE_DIR issue

### LINKED ISSUE #4232: Configure of HPX project using CMake FetchContent fails
## Expected Behavior

My project uses CMake's FetchContent module to fetch, configure and build HPX. Configuring and building my project should trigger a successful configure and build of HPX.

## Actual Behavior

Configuring of the HPX project fails with messages like this:

```
CMake Error: File /home/kor/development/project/hello_hpx/cmake/templates/global_module_header.hpp.in does not exist.
CMake Error at /tmp/hello_hpx/Debug-default/build/_deps/hpx-src/cmake/HPX_AddModule.cmake:102 (configure_file):
  configure_file Problem configuring file
Call Stack (most recent call first):
  /tmp/hello_hpx/Debug-default/build/_deps/hpx-src/libs/algorithms/CMakeLists.txt:126 (add_hpx_module)
```

  1. git clone https://github.com/kordejong/hello_hpx.git
  1. ./debug_build.sh

## Specifications

I will attach a file with details and error messages.

  - HPX Version: 1.4.0-rc1
  - Platform (compiler, OS): Linux, gcc

aurianer:
Fixes #4232

- Remove `project()` directive from modules to allow reference to hpx root through `PROJECT_SOURCE_DIR`
- Make corresponding change to the (old) use of `PROJECT_SOURCE_DIR`


aurianer:
> Wouldn't this PR imply to remove the project() statement from all existing modules as well? Also, IIRC, we wanted to end up with self-contained modules that could be imported into external projects independently from HPX itself. Wouldn't that require having a project() statement in each module (I'm honestly not sure...)?

- Yes I already removed it cause the `project()` statement was in the `HPX_AddModule` macro if I'm not mistaken.
- Yes that would be the final goal, but I'm not quite sure of how to do that for now without putting a `${CMAKE_PROJECT_DIR}/../..` to go back to the root of the hpx project, so I figured it was the cleanest solution for the next release as the module are not yet standalone, what do you think ?

Issue #4148, opened 2019-10-22
cacheline_data delivers poor perfomance with C++17 compared to C++14


sithhell:
## Expected Behavior

When running programs compiled with either C++17 or C++14 should have roughly the same performance.

## Actual Behavior

Programs compiled with C++17 are significantly slower. While investigating, it turns out that this is the culprit:
https://github.com/STEllAR-GROUP/hpx/blob/master/libs/concurrency/include/hpx/concurrency/cache_line_data.hpp#L46-L64

This has been verified by performing the following isolated benchmark:
clang: http://quick-bench.com/tn0AzNrQpe3t5gczE8pAJu1N5PI
gcc: http://quick-bench.com/SvM8Otk9FvU96uElajGEMAfw7mI

sithhell:
While fixing this issue, this needs to be addressed as well:
http://quick-bench.com/PkunktMrNXhZ5pdOusj9s9MRQ7c

as in, the generated constructor also initializes the padding data

biddisco:
Nice test. Disturbing result. Bug in the compilers? Seems odd that both gcc and clang are showing it in c++17 mode.

sithhell:
It looks like malloc is just bad when requesting aligned memory... at least that's what the profiler is telling me. So more a malloc bug. Do we know which allocator was in use when the problem was reported first?

hkaiser:
@sithhell we were using tcmalloc for this.

msimberg:
Do we know enough about the behaviour here? Should we maybe revert back to using manual padding instead of alignas for the release until we understand the problem better?

hkaiser:
@msimberg yes, I was planning to suggest doing exactly that - roll things back to manual alignment.

hkaiser:
This has been fixed by merging #4235

PR #4235, opened 2019-11-21
Don't use alignas in cache_{aligned,line}_data

msimberg:
"Fixes" #4148 by reverting to the non-alignas versions with all C++ standards.

### DIFF
--- a/libs/concurrency/include/hpx/concurrency/cache_line_data.hpp
+++ b/libs/concurrency/include/hpx/concurrency/cache_line_data.hpp
@@ -11,6 +11,7 @@
 #include <hpx/config.hpp>
 
 #include <cstddef>
+#include <utility>
 
 namespace hpx {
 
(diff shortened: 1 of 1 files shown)


libs/concurrency/include/hpx/concurrency/cache_line_data.hpp
```diff
@@ -41,55 +41,41 @@ namespace hpx {
...
-        };
-#else
         template <typename Data>
         struct cache_aligned_data
         {
+            cache_aligned_data()
+              : data_{}
+            {
+            }
+            cache_aligned_data(Data data)
```
hkaiser: You might want to consider adding two constructors instead, one taking a (const) lvalue ref and one taking an rvalue reference.

msimberg: Added the constructors.

hkaiser:
@msimberg Please go ahead and merge this. LGTM.

PR #4236, opened 2019-11-21
Move module print summary to modules

aurianer:
Prepare the independence of the modules by moving the configuration summary for modules


libs/algorithms/CMakeLists.txt
```diff
@@ -144,3 +144,6 @@ add_hpx_module(algorithms
   CMAKE_SUBDIRS examples tests
 )
 
+include(HPX_PrintSummary)
+create_configuration_summary(
+  "  Module configuration summary (algorithms):" "algorithms")
```
msimberg: Can we move these into the `add_hpx_module` function?

hkaiser: Sure, good thinking!

PR #4237, opened 2019-11-21
fixing annotated task to use the name, not the desc

khuck:
Fixes annotating HPX threads with APEX

## Proposed Changes

  - minor change to use the passed in name


hpx/util/annotated_function.hpp
```diff
@@ -83,7 +83,7 @@ namespace hpx { namespace util
             /* update the task wrapper in APEX to use the specified name */
             threads::set_self_timer_data(
                 external_timer::update_task(threads::get_self_timer_data(),
-                desc_));
+                std::string(name)));
```
msimberg: This needs a `#include <string>` (https://144987-4455628-gh.circle-artifacts.com/0/hpx_inspect_report.html).

khuck:
@msimberg I added the header

PR #4240, opened 2019-11-22
Mostly fix clang CUDA compilation

msimberg:
Fixes most of the compilation problems with clang CUDA. I'm adding a pycicle builder which uses the new Cray compiler based on clang 9. The clang 8/CUDA 10 builder on rostam still doesn't like these changes but the error messages look like it could be a problem with the clang setup, not clang itself.

I'm completely disabling the `partitioned_vector.cu` example for now. Our `#if defined(HPX_COMPUTE_DEVICE_CODE)` guards in components/actions end up excluding too much so that the required device functions don't get instantiated at all.

I'm also disabling the `hello_compute` example with clang. It causes an ICE with the Cray clang compiler. I don't know yet if it's an upstream clang problem or a Cray problem.


msimberg:
This time this should be ready for real. I've messed up the pycicle builder so results might not be correct. Merging this can wait until I have that sorted out though.

msimberg:
The pycicle builder is doing its job now. I'm going to go ahead and merge this.

PR #4241, opened 2019-11-22
Fix start_stop_callbacks test

msimberg:
The main thread startup callbacks were called after `hpx_main` was scheduled. This meant that the test failed occasionally if `hpx_main` was scheduled quickly enough. This moves the main thread tss init (which calls the startup functions) to before `hpx_main` is scheduled.

### DIFF
--- a/src/runtime_impl.cpp
+++ b/src/runtime_impl.cpp
@@ -415,6 +415,11 @@ namespace hpx
 
         lbt_ << "(1st stage) runtime_impl::start: booting locality " << here();
 
+        // Register this thread with the runtime system to allow calling
+        // certain HPX functionality from the main thread. Also calls
+        // registered startup callbacks.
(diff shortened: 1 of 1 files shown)


src/runtime_impl.cpp
```diff
@@ -415,6 +415,11 @@ namespace hpx
 
         lbt_ << "(1st stage) runtime_impl::start: booting locality " << here();
 
+	// Register this thread with the runtime system to allow calling
+	// certain HPX functionality from the main thread. Also calls
+	// registered startup callbacks.
+	init_tss("main-thread", 0, 0, "", "", false);
```
hkaiser: Looks like the indentation got messed up here.

msimberg: Grr, good catch! Should be fixed now.

PR #4245, opened 2019-11-25
Adding high performance channels

hkaiser:
This is the first step towards a new work-stealing scheduler. The PR adds two variation of high-performance channel types:

- a MPMC (multiple producer/multiple consumer) channel
- a MPSC (multiple producer/single consumer) channel
- a SPSC (single producer/single consumer) channel

~~Those are either unbuffered (`size == 0`) or bounded buffer channels.~~ All channels are buffered only now.


hkaiser:
@msimberg would you think this could still go into V1.4? I have marked this PR for the next release right now, but it looks to be ready to go.

msimberg:
@hkaiser, yeah, I don't see why not! Sorry, this slipped from my todo list. Just a couple of questions:
1. What exactly is the difference between buffered and unbuffered? It looks to me like `size == 0` is just a specialization of `size == 1`. What am I missing?
2. Channel = bounded queue that can be closed?

hkaiser:
> Just a couple of questions:
> What exactly is the difference between buffered and unbuffered? It looks to me like size == 0 is just a specialization of size == 1. What am I missing?

Those are very similar. Unbuffered is slightly more efficient compared to `size == `. Also, the `size == 1` is a corner case rarely anybody would ever want to use.

> Channel = bounded queue that can be closed?

Yah, in essence - these channels are thread-safe bounded queues.

hkaiser:
@msimberg Let's hold off on this one for a couple of days, I think I would like to have a MPSC variation added before merging. I'm not sure yet which of those I will actually need for the scheduler I'm working on, let's have them all ;-)

hkaiser:
@msimberg I was thinking to remove the unbuffered option for the channels now. Those put more strain on the caches as both ends (set and get) access the same variable in the channel (`head_`). This would also simplify the implementation...

msimberg:
> @msimberg Let's hold off on this one for a couple of days, I think I would like to have a MPSC variation added before merging. I'm not sure yet which of those I will actually need for the scheduler I'm working on, let's have them all ;-)

Sure, no problem. Not that you need it, but you have my blessing to merge this once you're happy with it (assuming tests pass!).

> @msimberg I was thinking to remove the unbuffered option for the channels now. Those put more strain on the caches as both ends (set and get) access the same variable in the channel (`head_`). This would also simplify the implementation...

Simple is good. I'm very curious now what you have planned for a new scheduler...

msimberg:
By the way, this should probably go into the `synchronization` module (#4231).

hkaiser:
> By the way, this should probably go into the synchronization module (#4231).

Indeed. Let's merge that one first, I will adapt this PR afterwards.

hkaiser:
The channels have been moved to the synchronization module now.

hkaiser:
The failures on CircleCI and Daint are unrelated

PR #4246, opened 2019-11-26
Expose CMake targets officially

### LINKED ISSUE #3818: Transitive HPX target

msimberg:
Fixes #3818.

- Exports HPX targets into the `HPX::` namespace.
- Adds `hpx_wrap` and `hpx_init` to the `hpx` target `INTERFACE` dependencies conditional on the target being an executable.
- Adds targets `HPX::component` and `HPX::plugin` which set required compile definitions.
- Slightly simplifies `hpx_setup_target`

Note that this does not remove the macros as they still do things that can't be done with targets (AFAIK):

- creating source groups
- setting the output directories (especially for plugins)
- setting library versions

I consider these convenience features though, except for the plugin output directory (can the search path be customized?).

This requires CMake 3.12 because of the generator expression `TARGET_NAME_IF_EXISTS`. I don't currently see a way to implement that part differently but maybe there is a way to make it work with older CMakes.

To do:
- [x] describe the requirements for a plugin.


msimberg:
Note: this is still a draft.

docs/sphinx/manual/creating_hpx_projects.rst
```diff
@@ -230,8 +230,66 @@ Additionally, if you wish to require |hpx| for your project, replace the
...
+example, the plugins created by |hpx| itself are installed into the ``hpx``
+subdirectory in the library install directory (typically ``lib`` or ``lib64``).
+When using the ``HPX::plugin`` target you need to install your plugins into an
+appropriate directory. You may also want to set the location of your plugin in
+the build directory with the ``*_OUTPUT_DIRECTORY*`` CMake target properties to
+be able to load the plugins in the build directory. Once you've set the install
+or output directory of your plugin you need to tell your executable where to
+find it at runtime. You can do this either by setting the environment variable
+``HPX_COMPONENT_PATHS`` or the ini setting ``hpx.component_paths`` (see
+:option:`--hpx:ini`) to the directory containing your plugin.
```
msimberg: @hkaiser your eyes on this paragraph would be greatly appreciated. I didn't try this myself, but this is what I understood from the source. The last part on `HPX_COMPONENT_PATHS`/`hpx.component_paths` is especially important.

msimberg:
Before I continue with this I'd like a decision on if we want to bump the minimum required version of CMake. Requiring 3.12 (or 3.13) would allow us to get rid of most, if not all, of the temporary workarounds we've had to put in place to correctly handler targets. Is there a good reason *not* to bump the minimum version?

hkaiser:
I wouldn't object to bumping the minimal cmake version as it's normally no issue to update cmake (even locally) if needed.

msimberg:
@biddisco (do I even need to ask...), @sithhell, opinions? Users!? Any reason not to go up to CMake 3.13?

docs/sphinx/manual/creating_hpx_projects.rst
hkaiser: I will go back to the sources for that. I only remember that we did change things there about a year ago...

biddisco:
I'm happy with updating cmake versions (I'm using 3.13.4) locally.

What's the status of this branch. Is it still a draft, or should we start testing it?

msimberg:
> I'm happy with updating cmake versions (I'm using 3.13.4) locally.
> 
> What's the status of this branch. Is it still a draft, or should we start testing it?

It used to work, and should still more or less work. I just didn't want to continue working on it until I know what version I can work with. It seems like we're getting to a consensus on bumping the minimum version so I can pick this up again.

Using it in external projects is where it needs testing, so if you have one for checking corner cases and other weirdness testing would be appreciated. https://github.com/STEllAR-GROUP/hpx/pull/4246/files#diff-96041d0f76292ea6455f6463c0f71aa8R233-R292 is meant to summarize how targets should be used in dependent projects. Comments on that section would also be helpful.

msimberg:
I think this is mostly functional now with the new targets. Is there someone who would like to volunteer to try this out (both that their code builds unchanged, and that it works with the new targets)?

For all normal use cases linking to `HPX::hpx` is meant to be enough (both libraries and applications). For components and plugins one needs to link to `HPX::component` or `HPX::plugin` and for plugins one may optionally set the install path to something useful.

hkaiser:
Will the 'old way' of doing things still work? I think we should support that for at least a release cycle to allow people to update their build systems on their own time.

msimberg:
The old way is at least meant to work. I may of course have broken things which is why I'm asking for others to test that their projects also work unchanged with these changes.

aurianer:
Thanks a lot !

msimberg:
I'd like to go ahead with this to get some real-world testing, assuming there are no more comments, and that phylanx is ready for the change (https://github.com/STEllAR-GROUP/phylanx/pull/1112 is as far as I know enough, but I did not test everything in phylanx).

msimberg:
Just for the record, I tried adding `SYSTEM` to our external interface targets, but decided against it in the end. It seems like CMake deals with it just fine, but our pkgconfig file generation does not take this into account and can produce things like `-isystem/usr/local` which completely breaks compilers. I'd like to stick with non-`SYSTEM` interface targets in this PR, as long as I don't have a simple way to filter out the compiler's include paths.

docs/sphinx/manual/creating_hpx_projects.rst
hkaiser: The description looks about right. Sorry it took so long...

hkaiser:
> 
> 
> I'd like to go ahead with this to get some real-world testing, assuming there are no more comments, and that phylanx is ready for the change ([STEllAR-GROUP/phylanx#1112](https://github.com/STEllAR-GROUP/phylanx/pull/1112) is as far as I know enough, but I did not test everything in phylanx).

I still need to get back to the Phylanx PR. However, let's go ahead with this one for now regardless.

msimberg:
> I still need to get back to the Phylanx PR. However, let's go ahead with this one for now regardless.

Sounds good. I'll be away Friday and the weekend, so let's wait until Monday with merging so that I have the full week to fix things when they inevitably break.

PR #4247, opened 2019-11-27
Add execution in module headers

aurianer:
Add execution in the name of the execution module headers, that way we know directly that it is coming from the `execution` module

 Remove the `execution.hpp` dependency to `dataflow` to avoid a circular dependency using `unwrap`


msimberg:
I think if we're doing this we should go for `hpx/execution` (instead of `hpx/execution/parallel`) right away. I see us potentially moving the actual executor implementations into an `executors` module at some point. The question is if those should live in `hpx/execution/executors` or `hpx/executors`. These are quite commonly included files so I'd like avoid changing our minds later and moving them multiple times. Opinons?

hkaiser:
> I think if we're doing this we should go for hpx/execution (instead of hpx/execution/parallel) right away. 

I weakly support this suggestion.

msimberg:
I feel like if we go with `hpx/execution/parallel` we can't stick with `hpx::parallel::execution` as the namespace. That'll confuse users to no end. And changing it to `hpx::execution::parallel` would break code (even if there's a compatibility layer) for no good reason (IMO).

PR #4249, opened 2019-11-30
Scheduler improvement

biddisco:
This PR contains a number of improvments to the shared-priority scheduler.

The thread holder is responsible for creating and deleting tasks, this means that each thread has only one copy of terminated items etc (instead of 1 per high/low/normal priority queue)

stealing has been redesigned to allow high priority stealing in preference to others in a more controllable way.

create_thread/add_new have been rewritten to behave according to stealing rules

new steal_high_priority_first or steal_local_first rules are implemented

new flags for round robin or thread_parent task placement are added

The scheduler stealing mode and work placement can be changed on the fly during operation, so that once an algorithm completes, the stealing or placement mode can be changed

A new priority and queue is introduced, this is bound_priority - tasks launched with the bound priority are bound to a particular core/queue and cannot be stolen. This is useful for openmp style task launching as well as numa touching of memory etc

Many small tweaks and optimizations are in place to improve efficiency


biddisco:
Note that this PR might still need some cleaning up and is mostly done now to see how tests pass/fail. It is probably not ready for merging just yet.

libs/coroutines/include/hpx/coroutines/thread_enums.hpp
```diff
@@ -93,6 +93,10 @@ namespace hpx { namespace threads {
             priority queue and will be executed before normal/low priority
             tasks are taken (some schedulers modify the behavior slightly and
             the documentation for those should be consulted). */
+        thread_priority_bound = 6,          /*!< Task goes onto a special
+            high priority queue and will never be stolen by another thread
+            after initial assignment. This should be used for thread placement
+            tasks such as OpenMP type for loops. */
```
msimberg: Should this be a hint instead?

libs/debugging/CMakeLists.txt
```diff
@@ -4,17 +4,20 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-cmake_minimum_required(VERSION 3.6.3 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.3.2 FATAL_ERROR)
 
 set(debugging_headers
   hpx/debugging/demangle_helper.hpp
+  # do not add hpx/debugging/print.hpp as it includes threads
+  # instead users should add this manually on demand
```
msimberg: It would be nice if the generic parts of `print.hpp` stayed in `print.hpp` and the threading related specializations moved to a separate file outside of this module. Independent of that, the header should still go into this list. Leaving it out doesn't get rid of any dependencies.

src/runtime/threads/executors/pool_executor.cpp
```diff
@@ -16,177 +16,166 @@
...
-            HPX_ASSERT(invalid_thread_id != id || !run_now);
+        threads::thread_id_type id = threads::invalid_thread_id;
+        pool_.create_thread(data, id, initial_state, run_now, ec);
+        if (ec)
+            return;
 
-            if (&ec != &throws)
-                ec = make_success_code();
-        }
+        //HPX_ASSERT(invalid_thread_id != id || !run_now);
```
msimberg: Intentional?

tests/performance/local/future_overhead.cpp
```diff
@@ -1,4 +1,5 @@
-//  Copyright (c) 2018 Mikael Simberg
+//  Copyright (c) 2018=2019 Mikael Simberg
```
msimberg:
```suggestion
//  Copyright (c) 2018-2019 Mikael Simberg
```

tests/performance/local/future_overhead.cpp
```diff
@@ -389,6 +411,20 @@ void measure_function_futures_create_thread_hierarchical_placement(
...
+        sched->add_remove_scheduler_mode(
+            hpx::threads::policies::scheduler_mode(
+                hpx::threads::policies::assign_work_thread_parent),
+            hpx::threads::policies::scheduler_mode(
+                hpx::threads::policies::enable_stealing |
+                hpx::threads::policies::enable_stealing_numa |
+                hpx::threads::policies::assign_work_round_robin |
+                hpx::threads::policies::steal_after_local |
+                hpx::threads::policies::steal_high_priority_first));
+    }
```
msimberg: Can we enable these flags unconditionally or will the local/static schedulers complain about unsupported flags?

tests/unit/resource/CMakeLists.txt
```diff
@@ -15,6 +17,8 @@ set(tests
     suspend_thread_timed
 )
 
+set(cross_pool_injectionr_PARAMETERS THREADS_PER_LOCALITY all)
```
msimberg:
```suggestion
set(cross_pool_injection_PARAMETERS THREADS_PER_LOCALITY all)
```

tests/unit/resource/CMakeLists.txt
```diff
@@ -15,6 +17,8 @@ set(tests
...
+set(scheduler_priority_check_PARAMETERS THREADS_PER_LOCALITY all)
```
msimberg: Should we keep it at `4` instead of `all`?

tests/unit/resource/CMakeLists.txt
```diff
@@ -37,6 +41,7 @@ foreach(test ${tests})
     ${${test}_FLAGS}
     EXCLUDE_FROM_ALL
     HPX_PREFIX ${HPX_BUILD_PREFIX}
+    DEPENDENCIES hpx_debugging hpx_testing
```
msimberg: No need for these. `add_hpx_executable` links to libhpx (if it doesn't we fix that instead).

tests/unit/resource/cross_pool_injection.cpp
```diff
@@ -0,0 +1,258 @@
...
+    } while (counter > 0);
+
+    return hpx::finalize();
+}
+
+int main(int argc, char* argv[])
+{
+    std::size_t max_threads = std::thread::hardware_concurrency();
+    std::vector<std::string> cfg = {
+        "hpx.os_threads=" + std::to_string(max_threads)};
```
msimberg: This should not be needed if we set the number of threads for the test. Or this should be `hpx.os_threads=all`.

tests/unit/resource/cross_pool_injection.cpp
```diff
@@ -0,0 +1,258 @@
...
+    // attach continuations to them that run on different
+    // random pools
+    const int loops = 10000;
+    //
+    std::cout << "1: Starting HP " << loops << std::endl;
+    std::atomic<int> counter(loops);
+    for (int i = 0; i < loops; ++i)
+    {
+        // high priority
+        std::size_t random_pool_1 = st_rand() % num_pools;
```
msimberg: This could maybe use a uniform distribution from `0` to `num_pools` instead.

hpx/runtime/threads/policies/shared_priority_queue_scheduler.hpp
```diff
@@ -27,1327 +33,1333 @@
...
-                return num_pending_misses;
-            }
-
-            std::size_t domain_num = d_lookup_[num_thread];
+                // ------------------------------------------------------------
+                // just function aliases
+                static constexpr auto& get_thread_number =
+                    threads::detail::get_thread_num_tss;
+                static constexpr auto& get_thread_numbers =
+                    threads::detail::get_thread_numbers_tss;
```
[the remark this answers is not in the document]
biddisco: not sure I understand what you are trying to say

src/runtime/threads/detail/thread_num_tss.cpp
```diff
@@ -7,17 +7,24 @@
...
         }
-    }
+
+        thread_tuple& thread_numbers_tss()
+        {
+            HPX_NATIVE_TLS thread_tuple thread_numbers_tss_ = {
+                std::uint16_t(-1), std::uint16_t(-1)};
+            return thread_numbers_tss_;
+        }
+    }    // namespace
```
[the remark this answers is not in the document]
biddisco: actually, the thread number and the pool id. it turns out the thread indexing was changed/broken at some point during recent changes and so I need the pool index to correctly get the local/global thread numbering right. 
I thought about removing the old thread_num_tss but decided that it'd be better to leave the existing stuff as-is and after this has been used for a while and is considered safe/stable to thn do a cleanup in the next version

hpx/runtime/threads/policies/shared_priority_queue_scheduler.hpp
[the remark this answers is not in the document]
msimberg: I'm just suggesting you use `using ...` instead of the function references as the only thing you gain with the reference is that you can remove `_tss`. Just my preference though.

libs/coroutines/include/hpx/coroutines/thread_enums.hpp
biddisco: Ideally we would have a better mechanism for passing this info, but for now I decided to go with a thread priority because the bound queues are considered high-priority as well as steal-free. As per the previous comment, I believe we need a root-and-branch cleanup of the design of some of this stuff and would prefer to get a working redesign of the scheduler in now and then make other changes later. 
The problem with a 'hint' is that it is not mandatory - ideally we should have `hints` and `orders` - which must be obeyed. Also the other schedulers cannot currently handle this and I didn't want to start redesigning all of them to accommodate a new  feature.
option 1) get this PR in - option 2) keep adding stuff to it for another year and never get it done. I went with option 1) in the hope that we can get the scheduler improvements in now and do some further design later.

src/runtime/threads/detail/thread_num_tss.cpp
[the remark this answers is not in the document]
msimberg: Then we can maybe call them that? And I guess there's no real need to pack them into a tuple instead of having them separate like the global thread index. The tuple has no names for the elements so it's hard to know what's what.

libs/debugging/CMakeLists.txt
biddisco: I agree. Unfortunately, the most useful info sometimes is the thread id stuff, so I always use it. where should we put this debug printer so that it doesn't cause unwanted dependencies. if I add print.hpp to this, then debugging.hpp pulls in threads and messes up otherwise clean dependencies. (there is some other stuff that uses debugging.hpp and that shouldn't need threads)

tests/unit/resource/CMakeLists.txt
biddisco: good spot. you must really read the code thoroughly to notice stuff like that. I'm impressed.

tests/unit/resource/CMakeLists.txt
biddisco: the cross pool test needs as many threads as possible to be useful, so I'd prefer all, the scheduler check doesn't matter much but all gives a more interesting output

tests/unit/resource/cross_pool_injection.cpp
biddisco: ok. I fix that/. probably just a left over cut'n'paste from another test

tests/unit/resource/cross_pool_injection.cpp
biddisco: that would be more 'modern c++' - I'll do it

libs/debugging/CMakeLists.txt
msimberg: Keep the non-thread related functionality in `print.hpp`, but move the customizations into `thread_print.hpp` or something like that in one of the scheduler directories (most likely `hpx/runtime/threads/detail`). A lot of this customization looks like it's also duplicating functionality that the format module does, but let's discuss that live tomorrow.

libs/coroutines/include/hpx/coroutines/thread_enums.hpp
msimberg: It is just a hint now because the other schedulers don't support it. I think it should be independent of priorities because you might want normal and low priority bound tasks. Whether it should be a hint or something entirely else though I don't know...

src/runtime/threads/detail/thread_num_tss.cpp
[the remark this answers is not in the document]
biddisco: I didn't want to use extra bytes so I packed 2x16bit nums into a tuple. The existing use of the thread_num stuff was totally undocumented, so I thought I could sneak the pool Id in there too without any objection. I'd like to leave the tuple as-is,  but I'll add some comments in the code to explain what's going on

libs/coroutines/include/hpx/coroutines/thread_enums.hpp
biddisco: yes. we would like low priority bound tasks, and high, but adding support for those will require more work and will delay the PR considerably ...

biddisco:
just noticed some other commits that I thought were removed are still present at the end of this PR. I will check I don't have other branches merged into this.

src/runtime/threads/detail/thread_num_tss.cpp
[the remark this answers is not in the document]
biddisco: I added some comments. it's local thread id and pool id, the other one is global thread id, so they actually complement each other.

libs/coroutines/include/hpx/coroutines/thread_enums.hpp
biddisco: I'll see how much I need to change to use a hint instead

hpx/runtime/threads/detail/scheduled_thread_pool_impl.hpp
```diff
@@ -134,8 +133,11 @@ namespace hpx { namespace threads { namespace detail
            << "\n"
            << "is running on PUs : \n";
         os << std::hex << HPX_CPU_MASK_PREFIX << get_used_processing_units()
+           << " "
+           << std::bitset<HPX_HAVE_MAX_CPU_COUNT>(get_used_processing_units())
```
hkaiser: What's the rationale not to use the existing `mask_type` for this?

hpx/runtime/threads/detail/thread_num_tss.hpp
```diff
@@ -10,20 +10,30 @@
...
-namespace hpx { namespace threads { namespace detail
-{
+namespace hpx { namespace threads { namespace detail {
+    // set/get the global thread Id to/from thread local storage
     HPX_EXPORT std::size_t set_thread_num_tss(std::size_t num);
     HPX_EXPORT std::size_t get_thread_num_tss();
 
+    // this tuple holds the local thread Id and the pool index
+    // associated with the thread
+    using thread_tuple = std::tuple<std::uint16_t, std::uint16_t>;
```
hkaiser: Any reason not to use `hpx::util::tuple`? Also, using tuple's in contexts like this usually suggests some kind of 'laziness' on behalf of the programmer. An explicit struct would probably be more descriptive.

hpx/runtime/threads/executors/guided_pool_executor.hpp
```diff
@@ -8,47 +8,53 @@
...
+
+#if !defined(GUIDED_POOL_EXECUTOR_DEBUG)
+#define GUIDED_POOL_EXECUTOR_DEBUG false
 #endif
 
-#include <hpx/config/warnings_prefix.hpp>
+namespace hpx {
+    // cppcheck-suppress ConfigurationNotChecked
+    static hpx::debug::enable_print<GUIDED_POOL_EXECUTOR_DEBUG> gpx_deb(
+        "GP_EXEC");
```
hkaiser: Are we sure that the use of the `enable_print` like this does not cause any ODR violations?

hpx/runtime/threads/executors/limiting_executor.hpp
```diff
@@ -44,34 +43,40 @@ namespace hpx { namespace threads { namespace executors
...
         {
-            if (++count_ > upper_threshold_) {
-                hpx::util::yield_while([&](){
-                    return (count_ > lower_threshold_);
-                });
+            if (++count_ > upper_threshold_)
+            {
+                hpx::util::yield_while(
+                    [&]() { return (count_ > lower_threshold_); }, nullptr,
+                    hpx::threads::pending, false);
```
hkaiser: Unrelated to this PR: why isn't `limiting_executor` using a sliding semaphore (or similar) to actually suspend the 'waiting' threads instead of constantly rescheduling them?

hpx/runtime/threads/policies/queue_holder_numa.hpp
```diff
@@ -0,0 +1,303 @@
...
+        //        {
+        //            return work_items_count_;
+        //        }
+
+        //        // This returns the current length of the staged queue
+        //        std::int64_t get_staged_queue_length(
+        //            std::memory_order order = std::memory_order_seq_cst) const
+        //        {
+        //            return new_tasks_count_.load(order);
+        //        }
```
hkaiser: Can this be removed?

hpx/runtime/threads/policies/queue_holder_numa.hpp
```diff
@@ -0,0 +1,303 @@
...
+
+        void increment_num_pending_misses(std::size_t num = 1) {}
+        void increment_num_pending_accesses(std::size_t num = 1) {}
+        void increment_num_stolen_from_pending(std::size_t num = 1) {}
+        void increment_num_stolen_from_staged(std::size_t num = 1) {}
+        void increment_num_stolen_to_pending(std::size_t num = 1) {}
+        void increment_num_stolen_to_staged(std::size_t num = 1) {}
```
hkaiser: Will those counters be available for the new scheduler?

hpx/runtime/threads/detail/thread_num_tss.hpp
biddisco: I didn;t realize tuples were the lazy man's struct! I will use a struct and give the members proper names to make things explicit. If it doesn't add overheads when fetching the data (I think it is returned by ref), I'll combine all 3 into one struct and unify them.

hpx/runtime/threads/executors/guided_pool_executor.hpp
biddisco: I have not had any link problems though I can see that if 2 TUs had different values of the macro, it would be a problem. I think it's up to individuals to not do that if they use it

hpx/runtime/threads/policies/queue_holder_numa.hpp
biddisco: ok

hpx/runtime/threads/policies/queue_holder_numa.hpp
biddisco: I've not yet needed them but I am conscious of the fact that they're missing. I actually find the total number of tasks executed by each thread to be a useful indicator. Since the scheduler supports `thread_parent` or `round_robin` injection of tasks we expect the number stolen to be large when many tasks are created by a single thread using `thread_parent` (example - the futures_overhad test) - when thread parent mode is used, it might be the case that all task are created by one thread, so stolen numbers will always be huge. with round robin creation, one expects less. Ultimately, the idle rate and number of tasks executed by each thread gives a good estimate of how well the tasks are being executed.
I didn't provide a counter for the total executed, though I did have an internal counter on one branch that I removed during cleaning. Do we have an existing counter for tasks executed that I could use for it?

biddisco: NB. Would a histogram counter be possible for the tasks executed per core?

hpx/runtime/threads/detail/scheduled_thread_pool_impl.hpp
biddisco: on machines where `HPX_HAVE_MAX_CPU_COUNT` is less than 64 bits, the output is just an uint, I wanted to see a binary bit mask explicitly - hence the cast to std::bitset on cout. (debugging).

biddisco:
As a sanity check to ensure this branch is not broken after the tweaks and rebases I have run tests using commit e1937fadb2b3c75e on 4 machines.
There appears to be a slight but not serious regression in performance, but the numbers are consistent with what I've seen in previous tests.
[pics here, was not able to upload image directly](https://pasteboard.co/IK0yCbX.png)
The machine ault - which is an AMD Naples architecture shows slightly worse performance than the local scheduler, but others are ok. The main benefit of this scheduler is that it give as good performance as the local on when workloads are placed on threads (launch=hierarchical = less stealing), but much better performance than the other schedulers when work is not placed strategically (launch=apply)

libs/debugging/CMakeLists.txt
biddisco: One thing I am not sure I agree with or understand - is the rationale for adding `print.hpp` to the CMakeLists file in the first place. Adding it to the list of headers causes it to be added to the 'module header' which means that `#include <hpp/debugging.hpp>` pulls in `print.hpp`. Is it intended that if you include a module, you have to include every single header declared in that module - this seems to be a very expensive way of exposing modules.
Is it intended that users `#include` the whole module, or are they encouraged to only include the headers they are using?

hkaiser: I don't think we have a clearly defined policy at this point. Both options have their advantages and disadvantages, Conceptually, all headers that are part of the module will be considered when doing the cyclic dependency analysis, regardless whether a particular header is included by the main module header file or not.

I personally weakly prefer including the main module header in modules that use a particular other module, but refer to just the files needed from inside the same module.

src/runtime/threads/detail/thread_num_tss.cpp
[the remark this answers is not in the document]
msimberg: The comments in the `.cpp` file say global thread num and pool id, while the `.hpp` file says local thread num and pool id. I guess it's actually the latter? (Note: please don't push any changes now so that we see what the CI says about the latest changes.)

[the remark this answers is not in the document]
biddisco: It should be local thread num.

Issue #4227, opened 2019-11-18
Fibonacci: segfaults and scalability issues


mratsim:
I've been measuring the overhead of various multithreading runtimes. HPX seems to have a couple of issues with a naive fibonacci.

## Expected Behavior

Fibonacci examples complete in a reasonable time

## Actual Behavior

HPX either segfaults or takes a very long time for no reason

### Problem 1: Sudden spike in time required to compute fibonacci.

1. Compile all the quickstart fibonacci examples
2. Run `bin/fibonacci_one --n-value 22`
3. Run `bin/fibonacci_one --n-value 23`
4. The time required for `fibonacci_direct` changes from 0.218 seconds to 8.263 s

![DeepinScreenshot_select-area_20191118184749](https://user-images.githubusercontent.com/22738317/69076516-f21fc680-0a33-11ea-831c-b62424f0d2b7.png)

### Problem 2: Segfaults

1. Run fibonacci_await with any n and it segfaults

![DeepinScreenshot_select-area_20191118185011](https://user-images.githubusercontent.com/22738317/69077381-c9003580-0a35-11ea-8b34-97c081ca3604.png)

### Problem 3: a  rare segfault will make all the subsequent call take very long-time
I manage to segfault the `fibonacci` binary with a value of 30 and I couldn't reclaim control of my console even with Ctrl+C, following that the performance of everything was very degraded.

Before a segfault on fibonacci 30

![DeepinScreenshot_select-area_20191118185907](https://user-images.githubusercontent.com/22738317/69077310-a3732c00-0a35-11ea-8dbc-2a291cd00c98.png)

After a segfault on fibonacci 30
1. Run `bin/fibonacci --n-value 23`, it takes 12s

![DeepinScreenshot_select-area_20191118185630](https://user-images.githubusercontent.com/22738317/69077104-2942a780-0a35-11ea-85dc-d05ad9bef38a.png)

## Specifications

... Please describe your environment

  - HPX Version: master - git: fd67ac0f86
  - Platform (compiler, OS): GCC, Linux

```
Core library:
  HPX_WITH_ACTION_BASE_COMPATIBILITY=ON
  HPX_WITH_AGAS_DUMP_REFCNT_ENTRIES=OFF
  HPX_WITH_APEX=OFF
  HPX_WITH_ATTACH_DEBUGGER_ON_TEST_FAILURE=OFF
  HPX_WITH_AUTOMATIC_SERIALIZATION_REGISTRATION=ON
[...]

PR #4250, opened 2019-12-02
Inline execution of scoped tasks, if possible

hkaiser:
This adds a new executor `scoped_parallel_executor` that causes for launched tasks to be executed inline (in the context of invoking thread) if possible. This optimization is applicable for tasks that do not exit the scope of their parent tasks.

- flyby: adding stack size detection on Windows
- flyby: fixing scheduler hint pass through to various functions
- flyby: specialize `cache_line_data` for items that don't require padding
- flyby: fixing HPX_WITH_SWAP_CONTEXT_EMULATION on Windows

This goes towards resolving #4227


hkaiser:
@msimberg I'll close this as well for now as I can't really make it work.

msimberg:
Ok, shame :( Would you mind writing a sentence or two here on what the difficulty was, in case someone other than you picks this up in the future?

hkaiser:
The main problem I faced was to control the lifetime of the `thread_data` instances. In order to manage threads that might be executed directly (inline), those are referred to from different places in the code concurrently. It is not clear how to decide when to let instance go out of scope without reintroducing reference counting.

Issue #3348, opened 2018-06-14
Add N-ary tasks to improve scheduling overheads


biddisco:
## New feature
During recent discussions with the kokkos people and also as a result of our failed attempt to have executors made variadic, we speculated that task management could be simplified in hpx by making some large changes to the way tasks are managed/created.

1. currently hpx::async forwards arguments to executors, tasks are created and eventually passed to the scheduler. The thread_data struct holds our low level task https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/runtime/threads/thread_data.hpp and is created somewhere between the executor and scheduler using a function like register_thread_nullary (or similar, I don't recall exactly)
2. when a parallel::for_loop is created, we might create 32/64/128/more individual tasks that go onto queues and require a considerable amount of management. In kokkos they just create a single task with an atomic counter with value 32/64/128/more in it that each thread decrements until it reaches zero. Each thread takes a 'chunk' of work corresponding to the index of the counter and the original ranges of the loop. We would need to be careful about iterator requirements (random access/forward etc). but if a loop over data that was iterable required 128 tasks, we could use the same strategy.
3. Variadic executors help us introspect arguments and take action based on values. We could promote the creation of thread_data to the async/executor interface and move the arguments into the callable in such a way that we can still access them. Our lowest level deferred callable type would become a function object with extra information. This would help us pass information to the scheduler and we could create n-ary tasks from parallell::for_loop and other parallel::algorithms that were simple iterations over ranges.
4. @sithhell hinted that we might move the shared state of the future into this struct to save extra mallocs between task create/destroy.
5. Schedulers could use lists instead of queues and by using intrusive pointers we could store prev/next in the thread_data struct and elide more mallocs for list nodes.
6. Having non-stealable tasks that use affinity to bind to cores would be simplified because we could add non-steal flags to the thread_data and the scheduler could steal or take work from the middle of a queue and no just the front/back.

This idea needs more thought, but I believe that a reworking of the task creation process in HPX could produce significant gains in performance. Please review this list and add more thoughts or comment on the above.

biddisco:
Discussion on IRC (@sithhell @msimberg) and elsewhere have raised interesting points that need to be itemized and acted on.
Our principle performance related problem is fork-join parallelism and being competitive with other libraries such as OpenMP etc.
Strategies that could help are:

[...]

PR #4251, opened 2019-12-02
Local workrequesting scheduler

hkaiser:
This adds a new experimental scheduler that works on a different basic principle compared to the others we have. Instead of actively stealing tasks from neighboring cores, with this scheduler a core that ran out of work posts a 'steal request' to the other cores which in turn respond with sending tasks that can be 'spared' back to the requesting core. This drastically reduces the contention in the scheduling loop itself. Essentially, the thread queues are not shared between cores anymore.

Currently this scheduler produces slightly worse results (about 10%) compared to the local priority scheduler we use by default. I'm sure there are ways to improve things, however.  

This relies on #4245

Please note that this PR also sets the new scheduler as the default for all applications. This is done to ensure all tests pass when run with this PR.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,5 @@
 # Copyright (c) 2016 Thomas Heller
-# Copyright (c) 2016-2018 Hartmut Kaiser
+# Copyright (c) 2016-2019 Hartmut Kaiser
 #
 # SPDX-License-Identifier: BSL-1.0
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
@@ -42,9 +42,11 @@ AlwaysBreakTemplateDeclarations: true
 BinPackArguments: true
 BinPackParameters: true
 BraceWrapping:
+#    AfterCaseLabel: true       # added in V10
     AfterClass: true
     AfterControlStatement: true
     AfterEnum: true
+#    AfterExternBlock: true     # added in V10
     AfterFunction: true
     AfterNamespace: false
     AfterStruct: true
@@ -52,6 +54,9 @@ BraceWrapping:
     BeforeCatch: true
     BeforeElse: true
     IndentBraces: false
+#    SplitEmptyFunction: true   # added in V10
+#    SplitEmptyNamespace: true  # added in V10
+#    SplitEmptyRecord: true     # added in V10
 BreakBeforeBinaryOperators: None
 BreakBeforeBraces: Custom
 BreakBeforeInheritanceComma: true
@@ -102,7 +107,7 @@ PenaltyExcessCharacter: 1000000
 PenaltyReturnTypeOnItsOwnLine: 200
 PointerAlignment: Left
 ReflowComments: false
-SortIncludes:    true
+SortIncludes: true
 SpaceAfterCStyleCast: true
 SpaceAfterTemplateKeyword: true
 SpaceBeforeAssignmentOperators: true
--- a/.clang-format
+++ b/.clang-format
@@ -42,11 +42,9 @@ AlwaysBreakTemplateDeclarations: true
 BinPackArguments: true
 BinPackParameters: true
 BraceWrapping:
-#    AfterCaseLabel: true       # added in V10
     AfterClass: true
     AfterControlStatement: true
     AfterEnum: true
-#    AfterExternBlock: true     # added in V10
     AfterFunction: true
     AfterNamespace: false
     AfterStruct: true
@@ -54,9 +52,6 @@ BraceWrapping:
     BeforeCatch: true
     BeforeElse: true
     IndentBraces: false
-#    SplitEmptyFunction: true   # added in V10
-#    SplitEmptyNamespace: true  # added in V10
-#    SplitEmptyRecord: true     # added in V10
 BreakBeforeBinaryOperators: None
 BreakBeforeBraces: Custom
 BreakBeforeInheritanceComma: true
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -694,7 +694,7 @@ endif()
 # Scheduler configuration
 ################################################################################
 hpx_option(HPX_WITH_THREAD_SCHEDULERS STRING
-  "Which thread schedulers are built. Options are: all, abp-priority, local, static-priority, static, shared-priority. For multiple enabled schedulers, separate with a semicolon (default: all)"
+  "Which thread schedulers are built. Options are: all, abp-priority, local, workstealing, static-priority, static, shared-priority. For multiple enabled schedulers, separate with a semicolon (default: all)"
   "all"
   CATEGORY "Thread Manager" ADVANCED)
 
@@ -711,6 +711,10 @@ foreach(_scheduler ${HPX_WITH_THREAD_SCHEDULERS_UC})
   if(_scheduler STREQUAL "LOCAL" OR _all)
     hpx_add_config_define(HPX_HAVE_LOCAL_SCHEDULER)
     set(HPX_WITH_LOCAL_SCHEDULER ON CACHE INTERNAL "")
+  endif()
+  if(_scheduler STREQUAL "WORKSTEALING" OR _all)
+    hpx_add_config_define(HPX_HAVE_LOCAL_WORKSTEALING_SCHEDULER)
+    set(HPX_WITH_LOCAL_WORKSTEALING_SCHEDULER ON CACHE INTERNAL "")
   endif()
   if(_scheduler STREQUAL "STATIC-PRIORITY" OR _all)
     hpx_add_config_define(HPX_HAVE_STATIC_PRIORITY_SCHEDULER)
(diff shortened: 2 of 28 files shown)


biddisco:
Interesting idea. I might steal that and try it out too!

hpx/runtime/threads/policies/lockfree_queue_backends.hpp
```diff
@@ -63,6 +63,8 @@ namespace hpx { namespace threads { namespace policies {
         bool pop(reference val, bool steal = true)
         {
 #if defined(HPX_HAVE_CXX11_STD_ATOMIC_128BIT)
+            if (steal)
+                return queue_.pop_left(val);
```
msimberg: Note that this makes `lockfree_fifo_backend` *exactly* the same as `lockfree_abp_fifo_backend` below (modulo the handling of 128 bit atomics). Same thing with `lockfree_lifo_backend`.

hkaiser: Fair point. We might be able to get rid of the abp backends... I think it is important to have the queue stealing from the opposite end in any case.

hkaiser:
This is ready for review now. @biddisco Would you mind running your scripts to generate perf graphs including this scheduler?

src/util/command_line_handling.cpp
```diff
@@ -1011,7 +1011,7 @@ namespace hpx { namespace util
 #endif
 
         // handle setting related to schedulers
-        queuing_ = detail::handle_queuing(cfgmap, vm, "local-priority-fifo");
+        queuing_ = detail::handle_queuing(cfgmap, vm, "local-workstealing-fifo");
```
msimberg: Should we already change the default? I suppose the answer depends on the performance results... And should we add the new scheduler to the suspension tests?

biddisco:
I don't know why pasting of images doesn't work any more. Here is a graph
https://pasteboard.co/IKqwXS8.png
seems to stop working on the new schedulers once the thread count exceeds some value. The test is still running, but I don't think we'll get any more data for the new schedulers.

src/util/command_line_handling.cpp
hkaiser: This is done on this branch only to run all tests using the new scheduler. Before merging I was planning to revert this change.

hkaiser:
@biddisco thanks for running the tests. As it looks like, there are still issues with this. I'll have a look. Also, I have some ideas how to further improve things. I'd say let's leave this unmerged for the time being until all problems have been fixed.

biddisco:
I like the idea of cores requesting tasks rather than explicitly stealing them, one thing I was planning on adding next to my scheduler is simply a bitmask that says when each core is empty of work - then new tasks being spawned (when using round robin as opposed to thread parent) can be added to cores with a '1' in the mask - and cores that want to steal can only bother stealing from cores with a '0' in the mask. This is a 'halfway' approach to what you're doing. 

(My feeling at the moment is that we are limited by the rate at which we can inject new work, rather than the rate at which we can dispatch it. The future_overhead test is a rather extreme case as it tens or hundreds of threads all vying for a limited amount of work). When we are injecting one task at a time with no actual work to be done, the cores trying to steal from each other are largely irrelevant - until they prevent the new work being added.

hkaiser:
> I like the idea of cores requesting tasks rather than explicitly stealing them, one thing I was planning on adding next to my scheduler is simply a bitmask that says when each core is empty of work - then new tasks being spawned (when using round robin as opposed to thread parent) can be added to cores with a '1' in the mask - and cores that want to steal can only bother stealing from cores with a '0' in the mask. This is a 'halfway' approach to what you're doing.

I think this would rather increase contention and not reduce it. But might be worth a try...

> (My feeling at the moment is that we are limited by the rate at which we can inject new work, rather than the rate at which we can dispatch it. The future_overhead test is a rather extreme case as it tens or hundreds of threads all vying for a limited amount of work). When we are injecting one task at a time with no actual work to be done, the cores trying to steal from each other are largely irrelevant - until they prevent the new work being added.

I agree. I'm also working on the concept of 'splittable tasks' that will reduce the overheads in bulk-async scenarios. Also, #4250 should help reducing overheads in general fork-join (and recursive parallelism) contexts (once its actually done). 

In general I think we are close to exhausting the optimization opportunities for our schedulers as they are. We should try to work on alternative approaches that fundamentally change the way we schedule work in clearly defined scenarios.

biddisco:
> In general I think we are close to exhausting the optimization opportunities for our schedulers as they are. We should try to work on alternative approaches that fundamentally change the way we schedule work in clearly defined scenarios.

Exactly this. I also refer back to #3345 for previous discussions on this topic.

hkaiser:
> I also refer back to #3345 for previous discussions on this topic.

I this really what you meant? That ticket looks unrelated.

biddisco:
> > I also refer back to #3345 for previous discussions on this topic.
> 
> I this really what you meant? That ticket looks unrelated.

sorry I meant #3348 (updated comment accordingly for future reference)

.clang-format
```diff
@@ -42,16 +42,21 @@ AlwaysBreakTemplateDeclarations: true
...
     AfterFunction: true
     AfterNamespace: false
     AfterStruct: true
     AfterUnion: true
     BeforeCatch: true
     BeforeElse: true
     IndentBraces: false
+#    SplitEmptyFunction: true   # added in V10
+#    SplitEmptyNamespace: true  # added in V10
+#    SplitEmptyRecord: true     # added in V10
```
msimberg: Do we want to add these? If they're useful we can update the docker image (once clang 10 is out), but otherwise these can maybe be left out of this PR?

CMakeLists.txt
```diff
@@ -694,7 +694,7 @@ endif()
 # Scheduler configuration
 ################################################################################
 hpx_option(HPX_WITH_THREAD_SCHEDULERS STRING
-  "Which thread schedulers are built. Options are: all, abp-priority, local, static-priority, static, shared-priority. For multiple enabled schedulers, separate with a semicolon (default: all)"
+  "Which thread schedulers are built. Options are: all, abp-priority, local, workstealing, static-priority, static, shared-priority. For multiple enabled schedulers, separate with a semicolon (default: all)"
```
msimberg: Bikeshedding that should be done before we merge this: `workstealing` isn't a very descriptive name, considering most of the other schedulers also steal work. `workrequesting` could be more appropriate.

hpx/runtime/threads/policies/local_priority_queue_scheduler.hpp
```diff
@@ -128,7 +128,7 @@ namespace hpx { namespace threads { namespace policies {
           , affinity_data_(init.affinity_data_)
           , num_queues_(init.num_queues_)
           , num_high_priority_queues_(init.num_high_priority_queues_)
-          , low_priority_queue_(0, thread_queue_init_)
+          , low_priority_queue_(init.num_queues_ - 1, thread_queue_init_)
```
msimberg: Is this just for consistency with `wait_or_add_new` only taking low priority tasks on the last thread? AFAICT we don't actually use the queue index in `thread_queue` or the backends, so we might want to consider removing that argument unless there's a use case for it.

hpx/runtime/threads/policies/local_workstealing_scheduler.hpp
```diff
@@ -0,0 +1,1608 @@
...
+            // victims
+            std::size_t const num_victims = num_queues_ - count(req.victims_);
+
+            // generate one more random number
+            std::size_t selected_victim = 0;
+            bucket_size =
+                (std::numeric_limits<std::int16_t>::max)() / num_victims;
+            do
+            {
+                selected_victim = uniform_int_(gen_) / bucket_size;
```
msimberg: Could the random numbers be drawn directly from the correct range with `uniform_int_distribution(0, num_victims)`? Same in the previous while-loop.

src/util/command_line_handling.cpp
msimberg: Leaving this unresolved just so we don't forget to change this back.

.clang-format
hkaiser: We should leave those out, definitely.

CMakeLists.txt
hkaiser: I like `workrequesting`!

hpx/runtime/threads/policies/local_priority_queue_scheduler.hpp
hkaiser: Done. I have removed the index from the thread queue.

msimberg:
@hkaiser is this in need of more reviewing or still in progress?

hkaiser:
> @hkaiser is this in need of more reviewing or still in progress?

@msimberg I put this on ice for now because of time constraints. It might be better to close it now and revive it later.

msimberg:
@hkaiser I'll close this for now to save us some build cycles. Just reopen when you pick it up again.

PR #4257, opened 2019-12-03
Fix apex annotation for async dispatch

biddisco:
When an annotated task is passed though the asyn dispatch, we can strip
off the task 'name' or annotation and pass it through to the thread
creation so that it is used when the task is initialized and apex
get the right name. We make use of
traits::get_function_annotation<typename hpx::util::decay<F>::type>::call(f)
to achieve this.

### DIFF
--- a/libs/execution/include/hpx/parallel/executors/thread_execution.hpp
+++ b/libs/execution/include/hpx/parallel/executors/thread_execution.hpp
@@ -119,13 +119,13 @@ namespace hpx { namespace threads {
         hpx::traits::is_threads_executor<Executor>::value &&
         std::is_same<typename hpx::util::decay<Hint>::type,
             hpx::threads::thread_schedule_hint>::value>::type
-    post(Executor&& exec, F&& f, Hint&& hint, const char *annotation, Ts&&... ts)
+    post(
+        Executor&& exec, F&& f, Hint&& hint, const char* annotation, Ts&&... ts)
     {
         exec.add(
             util::deferred_call(std::forward<F>(f), std::forward<Ts>(ts)...),
-            annotation,
-            threads::pending, true,
-            exec.get_stacksize(), std::forward<Hint>(hint), throws);
+            annotation, threads::pending, true, exec.get_stacksize(),
+            std::forward<Hint>(hint), throws);
     }
 
     ///////////////////////////////////////////////////////////////////////////
--- a/libs/execution/include/hpx/parallel/executors/thread_execution.hpp
+++ b/libs/execution/include/hpx/parallel/executors/thread_execution.hpp
@@ -54,7 +54,7 @@ namespace hpx { namespace threads {
         lcos::local::futures_factory<result_type()> p(
             std::forward<Executor>(exec),
             util::deferred_call(std::forward<F>(f), std::forward<Ts>(ts)...));
-        p.apply();
+        p.apply(hpx::traits::get_function_annotation<typename hpx::util::decay<F>::type>::call(f));
         return p.get_future();
     }
 
@@ -105,7 +105,8 @@ namespace hpx { namespace threads {
     {
         exec.add(
             util::deferred_call(std::forward<F>(f), std::forward<Ts>(ts)...),
-            "hpx::parallel::execution::post", threads::pending, true,
+            hpx::traits::get_function_annotation<typename hpx::util::decay<F>::type>::call(f),
+                    threads::pending, true,
             exec.get_stacksize(), threads::thread_schedule_hint(), throws);
     }
     ///////////////////////////////////////////////////////////////////////////
@@ -115,11 +116,12 @@ namespace hpx { namespace threads {
         hpx::traits::is_threads_executor<Executor>::value &&
         std::is_same<typename hpx::util::decay<Hint>::type,
             hpx::threads::thread_schedule_hint>::value>::type
-    post(Executor&& exec, F&& f, Hint&& hint, Ts&&... ts)
+    post(Executor&& exec, F&& f, Hint&& hint, const char *annotation, Ts&&... ts)
     {
         exec.add(
             util::deferred_call(std::forward<F>(f), std::forward<Ts>(ts)...),
-            "hpx::parallel::execution::post", threads::pending, true,
+            annotation,
+            threads::pending, true,
             exec.get_stacksize(), std::forward<Hint>(hint), throws);
     }
 
--- a/libs/execution/include/hpx/parallel/executors/thread_execution.hpp
+++ b/libs/execution/include/hpx/parallel/executors/thread_execution.hpp
@@ -51,10 +51,12 @@ namespace hpx { namespace threads {
         typedef typename util::detail::invoke_deferred_result<F, Ts...>::type
             result_type;
 
+        char const* annotation = hpx::traits::get_function_annotation<
+            typename hpx::util::decay<F>::type>::call(f);
         lcos::local::futures_factory<result_type()> p(
             std::forward<Executor>(exec),
             util::deferred_call(std::forward<F>(f), std::forward<Ts>(ts)...));
-        p.apply(hpx::traits::get_function_annotation<typename hpx::util::decay<F>::type>::call(f));
+        p.apply(annotation);
         return p.get_future();
     }
 
@@ -103,11 +105,12 @@ namespace hpx { namespace threads {
         hpx::traits::is_threads_executor<Executor>::value>::type
     post(Executor&& exec, F&& f, Ts&&... ts)
     {
+        char const* annotation = hpx::traits::get_function_annotation<
+            typename hpx::util::decay<F>::type>::call(f);
         exec.add(
             util::deferred_call(std::forward<F>(f), std::forward<Ts>(ts)...),
-            hpx::traits::get_function_annotation<typename hpx::util::decay<F>::type>::call(f),
-                    threads::pending, true,
-            exec.get_stacksize(), threads::thread_schedule_hint(), throws);
+            annotation, threads::pending, true, exec.get_stacksize(),
+            threads::thread_schedule_hint(), throws);
     }
     ///////////////////////////////////////////////////////////////////////////
     // post()
--- a/hpx/async_launch_policy_dispatch.hpp
+++ b/hpx/async_launch_policy_dispatch.hpp
@@ -90,7 +90,8 @@ namespace hpx { namespace detail
                 std::forward<F>(f), std::forward<Ts>(ts)...));
             if (hpx::detail::has_async_policy(policy))
             {
-                threads::thread_id_type tid = p.apply(pool, policy,
+                threads::thread_id_type tid = p.apply(pool,
+                    "async_launch_policy_dispatch", policy,
                     policy.priority(), threads::thread_stacksize_default, hint);
                 if (tid && policy == launch::fork)
                 {
--- a/hpx/async_launch_policy_dispatch.hpp
+++ b/hpx/async_launch_policy_dispatch.hpp
@@ -148,8 +148,8 @@ namespace hpx { namespace detail
             lcos::local::futures_factory<result_type()> p(util::deferred_call(
                 std::forward<F>(f), std::forward<Ts>(ts)...));
 
-            p.apply(pool, policy, policy.priority(),
-                threads::thread_stacksize_default, hint);
+            p.apply(pool, "async_launch_policy_dispatch::call", policy,
+                policy.priority(), threads::thread_stacksize_default, hint);
             return p.get_future();
         }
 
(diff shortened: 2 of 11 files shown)


biddisco:
This is an experimental patch that @khuck should try. It intercepts the annotated task function name and passes it through the async dispatch so that when the task is created, the task annotation is passed in. This solves the primary problem of tasks being renamed after the OTF2 trace info is created.

There may be other places where special handling is needed.

The code that handles stacksize, priority, policy, annotation &etc should be consolidated into a task_description type object. @hkaiser I welcome your feedback on that point. We could simplify a lot of the internal API by packing those items into a struct earlier on. Executors tat forward information through to other executors all need special treatment currently to handle these parameters. simplification would be a big help.

khuck:
@biddisco @hkaiser I am testing it now...

khuck:
@biddisco So very close!  The only thing that didn't work is the "then" clause, for example:

```
auto f1 = hpx::async(HP_executor,
    hpx::util::annotated_function(&dummy_task, "HP task"), "HP task", hp_m).
    then(SYNCHRONIZATION, hpx::util::annotated_function([&](auto &&){
        counter--;
    }, "then after HP"));
```

My example is here: https://github.com/khuck/jb_error
The "correct" task graph is here: https://github.com/khuck/jb_error/blob/master/taskgraph_pre_sync.png
All trace events are correct except the "then" clauses.

khuck:
@msimberg @biddisco @hkaiser I fixed the "then" clause (in hpx/lcos/local/packaged_continuation.hpp) and now it's annotated correctly.  I'll push that change in a few...

khuck:
@msimberg I just found the same thing.  I'm making that change (it's cleaner) and testing it now...

khuck:
@msimberg Interesting... it did NOT do the same thing.  It grabbed the annotation from the hpx::async() function, not the "then" function.  I'm going to leave my change as it is, and not revert it.

msimberg:
> @msimberg Interesting... it did NOT do the same thing. It grabbed the annotation from the hpx::async() function, not the "then" function. I'm going to leave my change as it is, and not revert it.

@khuck note that https://github.com/STEllAR-GROUP/hpx/blob/1ee979c7160f1c387790176df6bb7a12bf4802d6/src/util/thread_description.cpp#L56-L82
has to be changed to give priority to `altname`. With that change I'm 99% sure
```
hpx::util::thread_description desc(f_, "hpx::parallel::execution::parallel_executor::post");
```
will do the same thing.

khuck:
@msimberg I just confirmed that in the debugger.  I'll try that fix, too...

biddisco:
Apologies. I intentionally didn't fix the `.then()` case. The `async` case is used by most of the code paths and one would need to make another change around the `parallel::post` area to fix `.then()` - I was really just asking if this was the right approach from @hkaiser 's point of view. Since I've modified the async dispatch signatures in a few places I thought it would be rejected anyway.

khuck:
@msimberg @biddisco OK, I pushed my changes... looks like annotations are correct now.

biddisco:
Nice teamwork everyone. Question. Since we have now fixed the 'real' problem of tasks being created with one name and then renamed later, does this mean that some of the 'fixes' and checks that have been added to apex can now be removed or simplified?

khuck:
> Nice teamwork everyone. Question. Since we have now fixed the 'real' problem of tasks being created with one name and then renamed later, does this mean that some of the 'fixes' and checks that have been added to apex can now be removed or simplified?

I was thinking the same thing.  If the functions are annotated before starting, we don't need the update_task() calls, other than when annotate_function() is used (and should be refactored to be called "rename_function" to make it less ambiguous).  Also, the annotated_function() shouldn't call annotate_function() because it isn't necessary now.

msimberg:
Just to put the idea out there: John was at one point considering adding overloads to `async`/`apply` that would take a function annotation as the first argument to make annotating functions a bit less verbose. In other words something like this:

```
hpx::async("my important task", f, arg1, arg2);
```

It would purely be a simple wrapper making `f` an `annotated_function` but it's quite a nice shorthand. It's obviously not standard, but I think it could be a worthy addition. I'll open a separate issue if someone doesn't object badly.

khuck:
> Just to put the idea out there: John was at one point considering adding overloads to `async`/`apply` that would take a function annotation as the first argument to make annotating functions a bit less verbose. In other words something like this:
> 
> ```
> hpx::async("my important task", f, arg1, arg2);
> ```
> 
> It would purely be a simple wrapper making `f` an `annotated_function` but it's quite a nice shorthand. It's obviously not standard, but I think it could be a worthy addition. I'll open a separate issue if someone doesn't object badly.

No objections!

Also, John and I talked about swapping the order of the arguments to annotated_function() to be name, function rather than function, name.  Thoughts?  It would make it easier to annotate lambdas, for example.

khuck:
@biddisco The pycicle failures are due to some changes in HPX that exposed the fact that both APEX and HPX use concurrentqueue from https://github.com/cameron314/concurrentqueue, but unfortunately different versions. I patched APEX to use the concurrentqueue.hpp that is part of HPX instead of downloading the other one from GitHub.  The latest APEX develop branch has this fix merged.

biddisco:
just for clarification, I'm not in favour of
```hpx::async("function name", &task_rddr, args...)```
but I am in favour of
```hpx::async(task_descriptor("function name", ...), &task_rddr, args...)```
```
where a `task_descriptor` could be some new type that is analagous to `annotated_function`

hkaiser:
> just for clarification, I'm not in favour of
> ```hpx::async("function name", &task_rddr, args...)```
> but I am in favour of
>```hpx::async(task_descriptor("function name", ...), &task_rddr, args...)```
> where a `task_descriptor` could be some new type that is analogous to `annotated_function`

What would be the difference between `task_descriptor` and the existing `annotated_function` (besides the name)?

libs/execution/include/hpx/parallel/executors/thread_execution.hpp
```diff
@@ -54,7 +54,7 @@ namespace hpx { namespace threads {
         lcos::local::futures_factory<result_type()> p(
             std::forward<Executor>(exec),
             util::deferred_call(std::forward<F>(f), std::forward<Ts>(ts)...));
-        p.apply();
+        p.apply(hpx::traits::get_function_annotation<typename hpx::util::decay<F>::type>::call(f));
```
hkaiser: This will potentially access a moved-from object (see the `std::forward<F>(f)` above).

libs/execution/include/hpx/parallel/executors/thread_execution.hpp
```diff
@@ -105,7 +105,8 @@ namespace hpx { namespace threads {
     {
         exec.add(
             util::deferred_call(std::forward<F>(f), std::forward<Ts>(ts)...),
-            "hpx::parallel::execution::post", threads::pending, true,
+            hpx::traits::get_function_annotation<typename hpx::util::decay<F>::type>::call(f),
+                    threads::pending, true,
```
hkaiser: Same here, see above.

khuck:
@biddisco @msimberg @hkaiser I would like to push some more updates to this branch, please let me know if that's OK:
1) renamed "annotate_function" to "rename_function"
2) changed order of annotated_function(f, name) to annotated_function(name, f)
3) added a directory of unit tests in tests/unit/apex with 8 new tests
4) other minor fixes
I'll push these when you give me the OK.

hkaiser:
@khuck the changes sound sensible to me. I'd suggest to keep those on a separate PR, however.

msimberg:
@khuck also in favour of those changes (including them being on a separate branch). My only wish is that you keep the old overloads around and mark them deprecated with `HPX_DEPRECATED(HPX_DEPRECATED_MSG)`.

libs/execution/include/hpx/parallel/executors/thread_execution.hpp
biddisco: > >     hpx::async(task_descriptor("function name", ...), &task_rddr, args...)
> What would be the difference between task_descriptor and the existing annotated_function (besides the name)?

2 principal differences. 
The first (which doesn't work, but I'll mention it anyway) is that currently we have
`hpx::async(annotated_function(name, func, args));`
and I would prefer
`hpx::async(annotated_function(name, func), args);`
but that's no good for lambdas unfortunately.

The second is that we currently have 
execution policies
executor parameters
function annotations
executors that add params without the user really knowing it is going on
Internally the stacksize, priority, hints, annotations are not handled very consistently and it seems like an execution_parameter that wrapped all of those up (plus new ones we have not yet added) 

A real difficulty is n executors like guided_pool and limiting (others?) that wrap other executors, can't easily forward this information onwards (partly caused by the fact that thread_executors have stacksize, but non thread executors don't) - there isn't a single place to get the info from and cleanly pass it around. 

So all I realy meant was that a 'task_descriptor' would be some kind of execution paramaters object tht could old lots of this kind of thing in one place and then the info could be managed more easily (similar to the thread_init struct that mikael added to the pools/schedulers etc).

teonnik:
From a Slack conversation with John:

I am running a HPX + APEX (master with the recent #4261, networking is **off**) / MPI application on Piz Daint. When running on multiple processes, I only get a trace for a single process. Even with a clean directory, I get:

```
[OTF2] src/otf2_archive_int.c:3945: error: File does already exist: Could not create archive trace directory!
[OTF2] src/otf2_archive_int.c:1108: error: File does already exist: Couldn't create directories on root.
Closing OTF2 event files...
Writing OTF2 definition files...
Writing OTF2 Global definition file...
Writing OTF2 Node information...
Writing OTF2 Communicators...
Closing the archive...
done.
OTF2 Error: 17, File does already exist
....
```
from most processes. Only 1 process is able to write the OTF2 file.

biddisco:
@khuck So if we use N MPI ranks, but `HPX_WITH_NETWORKING=OFF` then I guess apex doesn't see each of the ranks as part of the same job, and it overwrites the trace files (or something similar). Is there an easy way to tell apex that it should create one trace file per rank/process and not clobber the other ones? (also the time sync for each rank might need some adjustment)

msimberg:
I pushed a few commits to take care of the remaining build failures and the potential use after move. In a small test the task names seem to still work as before. I also snuck in a small commit fixing the use of a deprecated affinity header. I can remove this if you prefer.

libs/execution/include/hpx/parallel/executors/thread_execution.hpp
hkaiser: @biddisco Ok, understood. I think all of this falls into the domain of the future executor work we will have to do in any case. I'd suggest leaving things as they are for now.

msimberg:
I had to make one final formatting commit. Besides the formatting the pycicle apex build and CircleCI were happy. Merging.

khuck:
> @khuck So if we use N MPI ranks, but `HPX_WITH_NETWORKING=OFF` then I guess apex doesn't see each of the ranks as part of the same job, and it overwrites the trace files (or something similar). Is there an easy way to tell apex that it should create one trace file per rank/process and not clobber the other ones? (also the time sync for each rank might need some adjustment)

Then HPX is lying to APEX about how many ranks there are, and who is rank zero.  see: https://github.com/STEllAR-GROUP/hpx/blob/d0011113dd4f299fc17adc76b3d5afe5255823e6/src/runtime_impl.cpp#L408-L411

biddisco:
@khuck if I were to add command line support to tell apex that we had N localities and replace the `hpx::get_initial_num_localities()` and `hpx::get_locality_id()` with something else - would the rest of apex function as expected, or are there other places where apex would assume that communication with other localities was required. 
In other words - I can spoof a value that tells apex we have 2,4,8 etc ranks and so it will initialize itself correctly and enable writing out of files per rank - but does apex use hpx actions or other types that involve communication anywhere at all in the code?

khuck:
> @khuck if I were to add command line support to tell apex that we had N localities and replace the `hpx::get_initial_num_localities()` and `hpx::get_locality_id()` with something else - would the rest of apex function as expected, or are there other places where apex would assume that communication with other localities was required.
> In other words - I can spoof a value that tells apex we have 2,4,8 etc ranks and so it will initialize itself correctly and enable writing out of files per rank - but does apex use hpx actions or other types that involve communication anywhere at all in the code?

That should work.  As of right now, APEX doesn't use any distributed features of HPX.  When the OTF library is initialized each process has to know its "rank", otherwise they all assume they are rank 0.

biddisco:
@khuck 
> That should work. As of right now, APEX doesn't use any distributed features of HPX. When the OTF library is initialized each process has to know its "rank", otherwise they all assume they are rank 0.
I have implemented the patch to https://github.com/STEllAR-GROUP/hpx/blob/d0011113dd4f299fc17adc76b3d5afe5255823e6/src/runtime_impl.cpp#L408-L411 so that we pass mpi rank and mpi size in instead of hpx localities etc when NETWORKING is OFF. Superficially this seems to be working and I am indeed getting trace files generated for N ranks and they can be loaded in vampir - however, they are incomplete/corrupted.
```
Initializing instrumentation with rank,size 0 2
Initializing instrumentation with rank,size 1 2
Rank 0 of 2.
Rank 1 of 2.
[OTF2] src/otf2_archive_int.c:1108: error: Unknown error code: Couldn't create directories on root.
OTF2 Error: INVALID, Unknown error code
```
I get messages of this kind when I do some debugging (I added the cout of rank/size) and the message appears to be emitted from https://github.com/khuck/xpress-apex/blob/d89100ba49511a16ab8c64f760f7d1bd2548e77b/src/apex/otf2_listener.cpp#L402-L405 the call to `OTF2_Archive_SetCollectiveCallbacks`

any idea what might be wrong?

biddisco:
I recompiled the OTF2 library in debug mode and traced the problem. The call to otf2 broadcast error was returning an unset pointer which said in effect "error" as rank 0 wrote the file/dir archive successfully. I've squashed that and now everything appears to proceed normally, but the trace files are not complete.
I do not know where to look next. Hints welcome please. since this issue is closed. I will copy these last two comments to a new one.

PR #4258, opened 2019-12-03
Couple of minor changes

hkaiser:
- adding stack size detection on Windows
- specialize cache_line_data for items that don't require padding
- fixing HPX_WITH_SWAP_CONTEXT_EMULATION on Windows

### DIFF
--- a/libs/concurrency/include/hpx/concurrency/cache_line_data.hpp
+++ b/libs/concurrency/include/hpx/concurrency/cache_line_data.hpp
@@ -11,6 +11,7 @@
 #include <hpx/config.hpp>
 
 #include <cstddef>
+#include <type_traits>
 #include <utility>
 
 namespace hpx {
@@ -40,6 +41,13 @@ namespace hpx {
                            (data_size % threads::get_cache_line_size())) %
                     threads::get_cache_line_size();
             }
+
+            template <typename Data>
+            struct needs_padding
+              : std::integral_constant<bool,
+                    detail::get_cache_line_padding_size(sizeof(Data)) != 0>
+            {
+            };
         }    // namespace detail
 
         // NOTE: We do not use alignas here because asking for overaligned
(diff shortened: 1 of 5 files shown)


libs/concurrency/include/hpx/concurrency/cache_line_data.hpp
```diff
@@ -40,6 +41,13 @@ namespace hpx {
                            (data_size % threads::get_cache_line_size())) %
                     threads::get_cache_line_size();
             }
+
+            template <typename Data>
+            struct needs_padding
+              : std::integral_constant<bool,
+                    detail::get_cache_line_padding_size(sizeof(Data)) != 0>
+            {
+            };
```
msimberg: Just curious, where do we use this with 0-sized data?

libs/coroutines/include/hpx/coroutines/detail/get_stack_pointer.hpp
```diff
@@ -45,6 +48,8 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
     }
 }}}}    // namespace hpx::threads::coroutines::detail
 
+#else
+#define HPX_HAVE_THREADS_GET_STACK_POINTER
```
msimberg: Would you mind inverting the `#if defined(HPX_WINDOWS)` and moving this `#define` up to the top? I think it makes the logic easier to follow.

msimberg:
@hkaiser are #4254 and this the last changes you wanted to have in for 1.4.0?

libs/concurrency/include/hpx/concurrency/cache_line_data.hpp
hkaiser: It's any data that has a size evenly divisible by the cache line size. I ran into this while working on the new scheduler. MSVC emits a warning for zero-sized arrays.

hkaiser:
> are #4254 and this the last changes you wanted to have in for 1.4.0?

Yes, indeed. Thanks!

libs/concurrency/include/hpx/concurrency/cache_line_data.hpp
msimberg: Ah right, makes sense. Thanks!

libs/coroutines/include/hpx/coroutines/detail/get_stack_pointer.hpp
hkaiser: Done.

PR #4260, opened 2019-12-04
Fix docs compilation problems

aurianer:
Fixes error in docs


aurianer:
Ok another error appeared, working on it

PR #4261, opened 2019-12-04
Fixing HPX/APEX linking and dependencies for external projects like Phylanx

khuck:
Fixes linking problems with external executables. Fixes flatbed problems from apex dependencies.

### DIFF
--- a/cmake/HPX_SetupApex.cmake
+++ b/cmake/HPX_SetupApex.cmake
@@ -15,66 +15,47 @@
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 if(HPX_WITH_APEX)
-  set(_hpx_apex_no_update)
-  if(HPX_WITH_APEX_NO_UPDATE)
-    set(_hpx_apex_no_update NO_UPDATE)
-  endif()
-
-  # We want to track parent dependencies
-  hpx_add_config_define(HPX_HAVE_THREAD_PARENT_REFERENCE)
-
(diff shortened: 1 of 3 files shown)


khuck:
The build-and-test CI failure has to do with documentation, I have no idea why that failed or whether it's relevant

hkaiser:
> The build-and-test CI failure has to do with documentation, I have no idea why that failed or whether it's relevant

Yes, Auriane is currently working on fixing this (see also #4260).

msimberg:
The pycicle builder seems to agree with this as well, which is good. It does not test that APEX actually works though (and we should try to think of a more thorough test for this; perhaps grepping stdout and otf2 files for keywords like correct task names etc.). @khuck I take it you checked that this works with e.g. Phylanx? Also, did you end up not needing to change anything related to rpath handling?

khuck:
> The pycicle builder seems to agree with this as well, which is good. It does not test that APEX actually works though (and we should try to think of a more thorough test for this; perhaps grepping stdout and otf2 files for keywords like correct task names etc.). @khuck I take it you checked that this works with e.g. Phylanx? Also, did you end up not needing to change anything related to rpath handling?

Correct, the rpath handling is now fixed.  For testing, there is a tests/unit/apex directory, so running ctest -R tests.unit.apex should check whether the build was valid.  Although the only way to really test this change is to use HPX with another project, which is the Phylanx test.  I have a buildbot server that does just that, http://omega.nic.uoregon.edu:8020

PR #4259, opened 2019-12-04
Module naming base

aurianer:
Move the `address_type`, `component_type` and `invalid_locality_id` to a `naming_base` module to eliminates some dependencies.

PR #4265, opened 2019-12-06
Threading module

msimberg:
Adds a threading module. It grew bigger than I hoped but I think its current form is an acceptable tradeoff. It currently includes:

- `thread_data` and friends
- `execution_agent`
- `thread_data`-related helper functions (`get_self_ptr`, `get_pool` etc.)
- `scheduler_base`
- `thread_pool_base`

It does not include the actual schedulers, thread pool implementations, the `hpx::thread` class, or the IO service pools (these will come in follow-up PRs). The critical dependency chain is `thread_data_stackful` -> `execution_agent` -> `scheduler_base` -> `thread_data_stackful`. After a discussion with @sithhell it seems like breaking this up is not worth the effort (at least now), but ideas are welcome.

I initially named the module `thread_data` in the hope that it would only contain `thread_data`/`thread_data_stackful`/`thread_data_stackless`, but it should be renamed now. Options: `threading`, `threading_base`, `thread_base`, `basic_threading`, `basic_thread`. Others? I vote for `threading_base`.

Refactoring that had to happen for this:
- Move things around in the `thread_data_fwd.hpp`, `thread_helpers.hpp`, `thread_pool_helpers.hpp` headers to make sure this module is self-contained.
- Add a callback to get a default `boost::asio::io_service` instead of relying on the runtime
- Add a callback to get the locality id.
- Move the `state` enum to this module to have it defined for the scheduler state. I did not rename it.
- Move the `yield_k`/`yield_while` functionality to `basic_execution`. They originally depended on `thread_state_enum` but did not use that information. I think we can do without that.
- Move the `get_action_name` declaration to the `functional` module.

To do:
- [x] rebase on naming module PR (#4259)
- [x] rename the module (`threading_base`)
- [x] update deprecated headers to new ones
- [x] format module


msimberg:
This is currently failing until the deprecated headers are changed and until #4264 is merged.

libs/thread_data/src/execution_agent.cpp
```diff
@@ -128,7 +125,9 @@ namespace hpx { namespace threads {
         thread_id_type id = self_.get_thread_id();
 
         // handle interruption, if needed
-        threads::interruption_point(id);
+        thread_data* thrd_data = get_thread_id_data(id);
```
aurianer: Should we remove the include `thread_data.hpp` with the use of `get_thread_id_data` ?

aurianer:
Thanks a lot ! :)

libs/thread_data/src/execution_agent.cpp
msimberg: Most likely not. Good catch!

libs/synchronization/include/hpx/synchronization/spinlock.hpp
```diff
@@ -71,8 +71,7 @@ namespace hpx { namespace lcos { namespace local {
 
             for (std::size_t k = 0; !acquire_lock(); ++k)
             {
-                util::detail::yield_k(k, "hpx::lcos::local::spinlock::lock",
-                    hpx::threads::pending_boost);
+                util::detail::yield_k(k, "hpx::lcos::local::spinlock::lock");
```
hkaiser: What motivated this change?

libs/threading_base/include/hpx/threading_base/annotated_function.hpp
```diff
@@ -114,36 +113,40 @@ namespace hpx { namespace util
...
+            operator()(Ts&&... ts)
             {
                 return util::invoke(f_, std::forward<Ts>(ts)...);
             }
 
             template <typename Archive>
             void serialize(Archive& ar, unsigned int const /*version*/)
             {
-                ar & f_;
+                ar& f_;
```
hkaiser: This is where clang-format always gets things wrong... Nothing we can do about this, I guess *sigh*.

libs/threading_base/include/hpx/threading_base/create_thread.hpp
```diff
@@ -8,64 +8,63 @@
...
+    inline void create_thread(policies::scheduler_base* scheduler,
+        thread_init_data& data, threads::thread_id_type& id,
+        thread_state_enum initial_state = pending, bool run_now = true,
+        error_code& ec = throws)
     {
         // verify parameters
-        switch (initial_state) {
+        switch (initial_state)
+        {
         case pending:
```
hkaiser: Let's add `HPX_FALLTHROUGH` while we're at it...

libs/threading_base/include/hpx/threading_base/external_timer.hpp
```diff
@@ -0,0 +1,332 @@
...
+
+    HPX_EXPORT void set_enable_parent_task_handler(
+        enable_parent_task_handler_type f);
+
+    namespace external_timer {
+
+        /* HPX provides a smart pointer to a data object that maintains
+     * information about an hpx_thread.  Any library (i.e. APEX)
+     * that wants to use this callback API needs to extend this class.
+     */
```
hkaiser: This formatting looks wrong as well. Should we switch to using `//` style comments?

libs/threading_base/include/hpx/threading_base/scheduler_state.hpp
```diff
@@ -0,0 +1,37 @@
...
+        state_pre_sleep = 7,
+        state_sleeping = 8,
+        state_pre_shutdown = 9,
+        state_shutdown = 10,
+        state_stopping = 11,
+        state_terminating = 12,
+        state_stopped = 13,
+        last_valid_runtime_state = state_stopped
+    };
+}
```
hkaiser: Have you merged the scheduler state and the runtime state?

libs/threading_base/include/hpx/threading_base/thread_init_data.hpp
```diff
@@ -32,28 +30,36 @@ namespace hpx { namespace threads
...
     public:
         thread_init_data()
-          : func(),
+          : func()
+          ,
 #if defined(HPX_HAVE_THREAD_DESCRIPTION)
-            description(),
+          description()
+          ,
 #endif
```
hkaiser: This formatting looks wrong as well :/

libs/threading_base/include/hpx/threading_base/thread_init_data.hpp
```diff
@@ -63,59 +69,71 @@ namespace hpx { namespace threads
...
+          /* HPX_HAVE_APEX forces the HPX_HAVE_THREAD_DESCRIPTION
          * and HPX_HAVE_THREAD_PARENT_REFERENCE settings to be on */
-            timer_data(util::external_timer::new_task(description,
-                parent_locality_id, parent_id )),
-#endif
-            priority(rhs.priority),
-            schedulehint(rhs.schedulehint),
-            stacksize(rhs.stacksize),
-            scheduler_base(rhs.scheduler_base)
+          timer_data(util::external_timer::new_task(
```
hkaiser: C-style comments...

libs/threading_base/include/hpx/threading_base/thread_init_data.hpp
```diff
@@ -63,59 +69,71 @@ namespace hpx { namespace threads
...
-        /* HPX_HAVE_APEX forces the HPX_HAVE_THREAD_DESCRIPTION
+          /* HPX_HAVE_APEX forces the HPX_HAVE_THREAD_DESCRIPTION
          * and HPX_HAVE_THREAD_PARENT_REFERENCE settings to be on */
-            timer_data(util::external_timer::new_task(description,
-                parent_locality_id,parent_id)),
-#endif
-            priority(priority_), schedulehint(os_thread),
-            stacksize(stacksize_),
-            scheduler_base(scheduler_base_)
+          timer_data(util::external_timer::new_task(
```
hkaiser: C-style comments...

libs/threading_base/include/hpx/threading_base/annotated_function.hpp
msimberg: Yeah, this one might be worth a `// clang-format off`.

libs/threading_base/include/hpx/threading_base/external_timer.hpp
msimberg: Yikes. I think that's a good idea. I'll take care of the other formatting problems as well.

libs/threading_base/include/hpx/threading_base/scheduler_state.hpp
msimberg: No. They are merged on master as well. I've only moved the file (or the enum definition). I wanted to separate them initially but decided against it because we e.g. use the scheduler states to compute the thread manager state (before writing this comment I thought the runtime state was computed from the scheduler states), and it's nice to have them more or less synchronized. There are some states which are used by the schedulers but not by the runtime (the ones related to suspension). I'm kind of indifferent on whether they should be the same or not.

libs/synchronization/include/hpx/synchronization/spinlock.hpp
msimberg: Cleanup. The thread state parameter is there on master for the free functions `yield_k` and `yield_while`, but not passed on to the `agent_base::yield_k` member function. This is for example what `execution_agent` does: https://github.com/STEllAR-GROUP/hpx/blob/d639fba90dfb76ca9711b68fba6284368dccc5c7/src/runtime/threads/execution_agent.cpp#L56-L75. I'll happily go the other way and add back the the parameter to `agent_base::yield_k` (doesn't introduce any problems with dependencies).

hkaiser: Good catch! The important piece is however, that when spinning on an HPX spinlock we yield the HPX thread with `priority_boost`. This way the thread will not sit indefinitely long in the queues.

msimberg: Indeed, `priority_boost` is the important part. I think what's in `execution_agent` seems like a fair solution. No need to think about which status to give, but still does something reasonable by default. The thresholds for when to yield in what way may of course need to be tweaked.

msimberg:
@hkaiser just FYI, I'll get back to #4206 and #4207 once we get this merged (as it's bound to lead to conflicts again).

msimberg:
The CircleCI failure is spelling only (fixed here: #4362). Unfortunately we can't really trust the pycicle builders at the moment (potentially due to the update; I will clean and restart them).

Are there remaining comments on this one?

msimberg:
Only the spell checker is complaining now on CircleCI.

msimberg:
I'd like to go ahead with this one. Are there more comments?

PR #4270, opened 2019-12-09
Refactor future implementation

aurianer:
Add the futures files inside the execution module

- Remove some dependencies
- Move `thread_timed_execution.hpp` back in `hpx/parallel/executors` to avoid a circular dependency between the `execution` and `local_lcos` module

Goes on top of #4247


hkaiser:
I'm not sure if this is the right granularity. In comparison with the other modules this one seems to be very large. I think the future type itself belongs into the local_lcos module (with the exception for the `future_data<id_type>` specialization). This would allow to step by step move things like (local) async into their own modules. For the component related files - I think we first should extract the file related to actions and AGAS, and only afterwards anything related to components. Actions and AGAS are unfortunately inseparable, but components can be built on top of those (I think).

aurianer:
> I'm not sure if this is the right granularity. In comparison with the other modules this one seems to be very large. I think the future type itself belongs into the local_lcos module (with the exception for the `future_data<id_type>` specialization). This would allow to step by step move things like (local) async into their own modules. For the component related files - I think we first should extract the file related to actions and AGAS, and only afterwards anything related to components. Actions and AGAS are unfortunately inseparable, but components can be built on top of those (I think).

- The first thing is that the futures and the execution traits are very tight together so they kind of belong together
- Concerning the fact that this module is too big, I totally agree for now but we planned to put the executors in their own module and deprecate the `execution/parallel/traits/detail/vc/` files. So at the end I would think the size is fine, what do you think ?
- Last thing, I'm not sure I understand your reference to the component related files cause I'm not doing anything related to that in this PR ^^

hkaiser:
> - The first thing is that the futures and the execution traits are very tight together so they kind of belong together, and putting futures in `local_lcos` would create a circular dependency between `execution` and `local_lcos`

I believe we can break this circular endency by moving only the basic future to local_lcos, leaving all things related to executors separated in the `executors` module. This involves separating the template specialization for executors into separate file, however.

> - Concerning the fact that this module is too big, I totally agree for now but we planned to put the executors in their own module and deprecate the `execution/parallel/traits/detail/vc/` files. So at the end I would think the size is fine, what do you think ?

ok

> - Last thing, I'm not sure I understand your reference to the component related files cause I'm not doing anything related to that in this PR ^^ 

Please ignore this comment, I did misunderstand what you were doing here.

msimberg:
> I believe we can break this circular dependency by moving only the basic future to `local_lcos`, leaving all things related to executors separated in the `executors` module. This involves separating the template specialization for executors into a separate file, however.

Good point. How much of `future` do you think we can move to `local_lcos` though? Do you specifically mean `future_base` (the `is_future` etc. traits are ok too)? The implementations of `future` and `shared_future` eventually need the executor `then` customization point and launch policies, which are both undoubtedly execution-y. `packaged_continuation` definitely needs the `execution` module. Do you see a way to fully define `hpx::future` in `local_lcos` without depending on the the `execution` module? Templating `future` and `shared_future` on those implementation details is definitely one possibility but that will show up in the types (don't know if that's a bad thing).

msimberg:
@hkaiser this is still slightly in progress, but I think the general changes required for the future/execution are in place and your feedback on it would be appreciated.

hpx/lcos/future.hpp
```diff
@@ -413,126 +408,48 @@ namespace hpx { namespace lcos { namespace detail
...
-            typename hpx::traits::detail::shared_state_ptr<result_type>::type p =
-                detail::make_continuation_alloc<continuation_result_type>(
-                    alloc, std::move(fut), std::forward<Policy_>(policy),
-                    std::forward<F>(f));
-            return hpx::traits::future_access<future<result_type> >::create(
-                std::move(p));
+            // dummy impl to fail compilation if this function is called
+            static_assert(dummy_false_type<Future>::value, "Cannot use the \
+                    dummy implementation of future_then_dispatch, please use \
+                    one of the template specialization.");
```
hkaiser: Instead of the `dummy_false_type<Future>` you could simply say `sizeof(Future) == 0` (but that's not a necessary change). The error message may be improved however by referencing possibly missing #includes/module dependencies.

msimberg:
Is it clear to you why this happens on Windows: https://travis-ci.org/STEllAR-GROUP/hpx/jobs/643968739#L2384-L2404? Have we lost some sort of laziness property with the rearranged templates?

hkaiser:
> Is it clear to you why this happens on Windows: https://travis-ci.org/STEllAR-GROUP/hpx/jobs/643968739#L2384-L2404? Have we lost some sort of laziness property with the rearranged templates?

I'll try to have a look.

aurianer:
> > Is it clear to you why this happens on Windows: https://travis-ci.org/STEllAR-GROUP/hpx/jobs/643968739#L2384-L2404? Have we lost some sort of laziness property with the rearranged templates?
> 
> I'll try to have a look.

Did you have time to look at this ? Otherwise I can try to have a look

msimberg:
The error is triggered by `.then` called with `launch::sync`. For some reason MSVC tries to use (or at least evaluates the return type for) this overload: https://github.com/aurianer/hpx/blob/d8a2a1052997b48b3be79175727fb92dbeccdb47/hpx/lcos/future.hpp#L952-L954, even though the second argument clearly isn't an `error_code`: https://github.com/aurianer/hpx/blob/d8a2a1052997b48b3be79175727fb92dbeccdb47/hpx/runtime/parcelset/put_parcel.hpp#L165-L171. Is this expected behaviour on MSVC (or by the standard)? Is there something we can do to work around this? Do we need to introduce a trait for the `T0` type here: https://github.com/aurianer/hpx/blob/d8a2a1052997b48b3be79175727fb92dbeccdb47/hpx/lcos/future.hpp#L961-L962, and SFINAE with the inverse in the first overload?

@hkaiser (perhaps @K-ballo)?

K-ballo:
> Is this expected behaviour on MSVC (or by the standard)?

Yes, this is expected standard behavior.

> Is there something we can do to work around this?

The `future_then_result` instantiation needs to be lazy.

hpx/lcos/future.hpp
```diff
@@ -1041,53 +950,20 @@ namespace hpx { namespace lcos
...
-                typename std::decay<F>::type>::value &&
-            !hpx::traits::is_one_way_executor<
-                typename std::decay<F>::type>::value &&
-            !hpx::traits::is_two_way_executor<
-                typename std::decay<F>::type>::value &&
-            !hpx::traits::is_threads_executor<
-                typename std::decay<F>::type>::value
-          , hpx::traits::future_then_result<future, F>
-        >::type
+        typename hpx::traits::future_then_result<future, F>::type
```
msimberg: Something like this perhaps:
```suggestion
        typename util::lazy_enable_if<hpx::traits::is_callable<F(future&&)>::value, typename hpx::traits::future_then_result<future, F>::type>::type
```?

Comment (msimberg) [reply hpx/lcos/future.hpp]: And the inverse condition below on `T0`.

Comment (hkaiser) [reply hpx/lcos/future.hpp]: I'd say rather this
```suggestion
        typename util::lazy_enable_if<hpx::traits::is_callable<F(future&&)>::value, hpx::traits::future_then_result<future, F>>::type
```
You want to delay the instatiation of `future_then_result<>:;type` after all

msimberg: @aurianer mind giving this a try?

hpx/lcos/future.hpp
```diff
@@ -1042,15 +951,8 @@ namespace hpx { namespace lcos
...
-            !hpx::traits::is_launch_policy<
-                typename std::decay<F>::type>::value &&
-            !hpx::traits::is_one_way_executor<
-                typename std::decay<F>::type>::value &&
-            !hpx::traits::is_two_way_executor<
-                typename std::decay<F>::type>::value &&
-            !hpx::traits::is_threads_executor<
-                typename std::decay<F>::type>::value
-          , hpx::traits::future_then_result<future, F>
+            hpx::traits::is_invocable<F(future&&)>::value,
```
msimberg:
```suggestion
            hpx::traits::is_invocable<F, future&&>::value,
```
(as pointed out by @K-ballo)

hpx/lcos/future.hpp
```diff
@@ -1367,49 +1250,24 @@ namespace hpx { namespace lcos
...
-            !hpx::traits::is_launch_policy<
-                typename std::decay<F>::type>::value &&
-            !hpx::traits::is_threads_executor<
-                typename std::decay<F>::type>::value &&
-            !hpx::traits::is_one_way_executor<
-                typename std::decay<F>::type>::value &&
-            !hpx::traits::is_two_way_executor<
-                typename std::decay<F>::type>::value
-          , hpx::traits::future_then_result<shared_future, F>
+            hpx::traits::is_invocable<F(shared_future&&)>::value,
```
msimberg:
```suggestion
            hpx::traits::is_invocable<F, shared_future&&>::value,
```
(as pointed out by @K-ballo)

hpx/lcos/future.hpp
```diff
@@ -1059,35 +961,16 @@ namespace hpx { namespace lcos
                 std::forward<F>(f), ec);
         }
 
-        template <typename Policy, typename F>
+        template <typename T0, typename F>
         typename util::lazy_enable_if<
-            hpx::traits::is_launch_policy<
-                typename std::decay<Policy>::type
-            >::value,
+            !hpx::traits::is_invocable<F(future&&)>::value,
```
msimberg:
```suggestion
            !hpx::traits::is_invocable<T0, future&&>::value &&
            hpx::traits::is_invocable<F, future&&>::value,
```

hpx/lcos/future.hpp
```diff
@@ -1367,49 +1250,24 @@ namespace hpx { namespace lcos
...
                 std::forward<F>(f), ec);
         }
 
-        template <typename Policy, typename F>
+        template <typename T0, typename F>
         typename util::lazy_enable_if<
-            hpx::traits::is_launch_policy<
-                typename std::decay<Policy>::type
-            >::value,
+            !hpx::traits::is_invocable<F(shared_future&&)>::value,
```
msimberg:
```suggestion
            !hpx::traits::is_invocable<T0, shared_future&&>::value &&
            hpx::traits::is_invocable<F, shared_future&&>::value,
```

hpx/lcos/future.hpp
```diff
@@ -1059,35 +961,16 @@ namespace hpx { namespace lcos
...
-
-        template <typename Executor, typename F>
-        typename util::lazy_enable_if<
-            hpx::traits::is_one_way_executor<
-                typename std::decay<Executor>::type>::value ||
-            hpx::traits::is_two_way_executor<
-                typename std::decay<Executor>::type>::value ||
-            hpx::traits::is_threads_executor<
-                typename std::decay<Executor>::type>::value
-          , hpx::traits::future_then_executor_result<Executor, future, F>
```
msimberg: @hkaiser `future_then_executor_result` remains as an unwanted dependency. Would you suggest we:
1. Use `future_then_result` instead and get rid of `future_then_executor_result`? Are there actually cases where `async_execute` would return something other than `future_then_result`, and is this something we want to keep allowing?
2. We introduce a `future_then_T0_result` where `T0` is some useful name describing something policy- or executor-like, which we then specialize in the executor module again like for `future_then_dispatch`?

The second sounds like it's definitely doable, but I'd rather simplify *if* we don't actually to specialize the return type per executor.

hpx/lcos/future.hpp
aurianer: I did not manage to make it compile, it fails on this :
```
In file included from /users/aurianer/projects/hpx_module_futures/src/runtime_impl.cpp:12:
In file included from /apps/daint/UES/aurianer/build/hpx/build_module_futures/libs/collectives/include/hpx/collectives.hpp:15:
/users/aurianer/projects/hpx_module_futures/libs/collectives/include/hpx/collectives/broadcast.hpp:426:46: error: no matching member
      function for call to 'then'
            hpx::when_all(broadcast_futures).then(&return_void).get();
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
```
Any hints for me ?

aurianer: Ok actually it was my bad sorry ^^

aurianer: Still does not compile for :
```
/users/aurianer/projects/hpx_module_futures/libs/collectives/include/hpx/collectives/gather.hpp:325:27: error: no matching member
      function for call to 'then'
            return result.then(
                   ~~~~~~~^~~~
/users/aurianer/projects/hpx_module_futures/hpx/lcos/future.hpp:957:9: note: candidate template ignored: substitution failure [with F =
      (lambda at /users/aurianer/projects/hpx_module_futures/libs/collectives/include/hpx/collectives/gather.hpp:326:17)]: implicit
      instantiation of undefined template 'hpx::traits::is_callable<(lambda at
      /users/aurianer/projects/hpx_module_futures/libs/collectives/include/hpx/collectives/gather.hpp:326:17), hpx::lcos::future<bool>
      &&>'
        then(F && f, error_code& ec = throws)
        ^
/users/aurianer/projects/hpx_module_futures/hpx/lcos/future.hpp:969:9: note: candidate function template not viable: requires at least 2
      arguments, but 1 was provided
        then(T0 && t0, F && f, error_code& ec = throws)
        ^

```

hpx/lcos/future.hpp
msimberg: It looks like using `future_then_result` for the launch policy/executor overload does indeed work in our codebase. The question is then if that'll always be correct and if we still want to keep `future_then_T0_result` in some form.

aurianer: @hkaiser any comments on this ?

msimberg: Ping @hkaiser. As this works I'd like to move on with this PR. We can decide on what trait to use here later as well.

hpx/lcos/future.hpp
```diff
@@ -14,43 +14,36 @@
...
 #include <hpx/traits/future_traits.hpp>
-#include <hpx/execution/traits/is_executor.hpp>
 #include <hpx/traits/is_future.hpp>
-#include <hpx/traits/is_launch_policy.hpp>
-#include <hpx/type_support/always_void.hpp>
 #include <hpx/type_support/decay.hpp>
-#include <hpx/type_support/identity.hpp>
-#include <hpx/type_support/lazy_enable_if.hpp>
-#include <hpx/type_support/void_guard.hpp>
-#include <hpx/util/serialize_exception.hpp>
```
msimberg, line 41: I think you'll need to add back this line (`serialize_exception.hpp`). The pycicle builder that's failing is running with networking off. The builders with networking on are likely getting the header in the right place by mistake, but the header is needed in here.

msimberg:
It looks like this is working quite well with `decltype(auto)` now (the GitHub Actions builders are over at @aurianer's fork, but they seem to be building this just fine also on Windows). The pycicle builder just needs to have that one header fixed. @hkaiser are you happy with this solution (assuming the test results end up greenish)?

hpx/lcos/future.hpp
aurianer: Ok thanks !

msimberg:
I suggest we go ahead with this as it is. It might still need some cleanup but exactly what to clean up will be more apparent once files get moved to their modules, which would be the next step.

aurianer:
I just rebased this :)

PR #4272, opened 2019-12-12
Fix pushing of documentation

msimberg:
PROJECT_BINARY_DIR is not set in scripts that are called indirectly through `cmake -P`.


aurianer:
Thanks !

PR #4275, opened 2019-12-13
Make sure APEX is linked to even when not explicitly referenced

### LINKED ISSUE #4273: Clients of HPX 1.4.0-rc2 with APEX ar not linked to libhpx-apex
## Expected Behavior

When using a build of HPX with support for APEX I expect it to be possible to generate a trace without having to instrument my code.

```
# Expected: a trace is printed
APEX_SCREEN_OUTPUT=1 ./fibonacci
```

## Actual Behavior

Without calling an APEX function from my code my program is not linked to libhpx-apex and no trace is generated. When I do call an APEX function, my program is linked to libhpx-apex and a trace can be generated. The same behavior can be seen with the HPX examples, like with fibonacci. No trace can be generated from it. apex_fibonacci can generate a trace.

```
# Actual: no trace is printed
APEX_SCREEN_OUTPUT=1 ./fibonacci
```

I have  a project for testing building different HPX variants (build type, features, ...). With this project I verify whether building HPX works as advertised on my platforms, without being distracted by client code.

  1. git clone https://github.com/kordejong/hello_hpx.git
  1. Edit `rel_with_deb_info.sh`: comment all rebuild commands except for the one for the apex variant. This script shows the arguments passed to CMake.
  1. Run the script. It will build HPX in a temp location (see `build_common.sh`)
  1. Run `APEX_SCREEN_OUTPUT=1 ./fibonacci` in the build directory (`/tmp/hello_hpx/RelWithDebInfo-apex/build/_deps/hpx-build/bin`). Also, `ldd fibonacci` shows that the program is not linked to libhpx_apex.

## Specifications

  - HPX Version: 1.4.0-rc2
  - Platform (compiler, OS): Kubuntu 18.04, gcc-7
  - Subet of the output of CMake configure of HPX: [hpx-configure.txt](https://github.com/STEllAR-GROUP/hpx/files/3954926/hpx-configure.txt)
  - Output of `fibonacci --hpx:info`: [hpx-info.txt](https://github.com/STEllAR-GROUP/hpx/files/3954912/hpx-info.txt)

msimberg:
Fixes #4273.


msimberg:
@kordejong would you mind giving this a try? I couldn't make CMake apply the flag only to apex. This adds `-Wl,-no-as-needed` to all dependent libraries which may result in other non-references libraries being linked as well, but at least apex gets linked.

kordejong:
I think you lost a dash in front of `no-as-needed`. After I put it in, the build continued. It is still busy, but I noticed that fibonacci and some other examples I checked are linked to apex.

msimberg:
Very good catch, thanks! I did indeed lose one. I've added it now.

kordejong:
The build finished fine, so I think your PR is good.

PR #4277, opened 2019-12-17
MPI_futures

biddisco:
Add new hpx::mpi library for futures with async MPI calls
    
This adds support for getting a future from an asynchronous
MPI call and having the future ready when the call completes.
Background progress is made in MPI by simply attaching a callback
into the scheduling loop to handle MPI Requests.

Only one test exists so far. Others will be added in the coming days. Submitting this PR early so that users at CSCS can test it prior to merging.


hkaiser:
@biddisco Could you add a simple executor that uses your new `hpx::mpi::async` to schedule tasks? I think that would greatly improve the usability of your proposed facilities.

biddisco:
> 
> 
> @biddisco Could you add a simple executor that uses your new `hpx::mpi::async` to schedule tasks? I think that would greatly improve the usability of your proposed facilities.

Absolutely Yes. I presume you are thinking along the lines of an executor representing a communicator?
The current version (lets call it 1) is intended to allow a drop in replacement of existing code that goes from
```
MPI_Isend(&tokens[i], 1, MPI_INT, rank_to, i, MPI_COMM_WORLD, &request);
```
to
```
auto f_send = hpx::mpi::async(
  MPI_Isend, &tokens[i], 1, MPI_INT, rank_to, i, MPI_COMM_WORLD);
```
so upgrading existing code is super-trivial.

Executors for communicators could be considered a level 2 approach as it requires the user to make more significant structural changes to their code - but yes. I will be working on that next as we have to use copies of communicators if we have multiple collective operations in flight simultaneously (two on the same communicator can lead to disaster). At that point, having a communicator as an executor makes perfect sense, though our executors will have to maintain 'state' such as a communicator handle and so they deviate somewhat from the standard's view of an executor.

I plan on upgrading the `cuda_future_helper` code to mimic the same style so that we can have `hpx::cuda::async(....)` and `hpx::cuda::apply(....)` (I will add `hpx::mpi::apply` thought I'm not sure anyone will want to use a fire and forget mpi call... maybe). When I do that, I can see cuda streams being an analogue to the executor/communicator pattern.

hpx/runtime/threads/policies/scheduler_base.hpp
```diff
@@ -275,6 +275,14 @@ namespace hpx { namespace threads { namespace policies {
             return thread_queue_init_.small_stacksize_;
         }
 
+        void set_user_polling_function(const std::function<void(void)> &f) {
+            user_polling_function_ = f;
+        }
+
+        void user_polling_function() {
+            user_polling_function_();
```
msimberg: I think this needs a check to see if `user_polling_function_` is empty.

libs/mpi/docs/index.rst
```diff
@@ -0,0 +1,18 @@
...
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+.. _libs_mpi:
+
+===
+mpi
+===
+
+TODO: High-level description of the library.
```
msimberg: @biddisco this is a good place to write documentation on using MPI explicitly in HPX.

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    // This is not completely safe as it will return when the request list is
+    // empty, but cannot guarantee that other communications are not about
+    // to be launched in outstanding continuations etc.
+    void wait() {
+        // Yield until there is only this and background threads left.
+        hpx::util::yield_while([&]() {
+            return (active_requests_.size()>0);
+        });
```
msimberg: Would a `counting_semaphore` work here?

libs/mpi/tests/unit/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
...
+set(subdirs
+   )
+
+# add sub directories
+foreach(subdir ${subdirs})
+  add_hpx_pseudo_target(tests.unit.modules.mpi.${subdir})
+  add_subdirectory(${subdir})
+  add_hpx_pseudo_dependencies(tests.unit.modules.mpi
+    tests.unit.modules.mpi.${subdir})
+endforeach()
```
msimberg: Can this and the compile tests part be removed or are you planning on adding something there?

libs/mpi/tests/unit/mpi_ring_async.cpp
```diff
@@ -0,0 +1,157 @@
...
+                msg_send(rank, size, rank_to, rank_from, tokens[i], i);
+            });
+        }
+    }
+
+    // Our simple counter should reach zero when all send/recv pairs are done
+    while (counter != 0)
+    {
+        hpx::this_thread::yield();
+    }
```
msimberg: Would a `latch` work here?

hkaiser:
> The current version (lets call it 1) is intended to allow a drop in replacement of existing code that goes from
> ```
>MPI_Isend(&tokens[i], 1, MPI_INT, rank_to, i, MPI_COMM_WORLD, &request);
> ``` 
> to
> ```
> auto f_send = hpx::mpi::async(
>   MPI_Isend, &tokens[i], 1, MPI_INT, rank_to, i, MPI_COMM_WORLD);
> ```
> so upgrading existing code is super-trivial.

What I had in mind was to directly go from 
```
MPI_Isend(&tokens[i], 1, MPI_INT, rank_to, i, MPI_COMM_WORLD, &request);
``` 
to
```
auto f_send = hpx::async(hpx::mpi::executor{},
  MPI_Isend, &tokens[i], 1, MPI_INT, rank_to, i, MPI_COMM_WORLD);
```
instead. I'd like to avoid introducing yet another API level `async`. Even more as such an executor could be directly used with `future::then`, `dataflow` etc.

I would consider `hpx::mpi::async` and implementation detail (which doesn't even have to be called `async`, but that's not important) and the executor would be the level 1 API.

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,186 @@
...
+        // return a new future with the mpi::future_data shared state
+        using traits::future_access;
+        return future_access<hpx::future<void>>::create(std::move(data));
+    }
+
+    // -----------------------------------------------------------------
+    // Background progress function for MPI async operations
+    // Checks for completed MPI_Requests and sets mpi::future ready
+    // when found
+    void poll()
```
hkaiser: Can this go into `namespace detail`?

libs/mpi/include/hpx/mpi/mpi_future.hpp
hkaiser: Also, can this go into `namespace detail`?

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,186 @@
...
+            : hpx::lcos::detail::future_data<void>(no_addref)
+        {}
+
+        // The native MPI request handle owned by this future data
+        MPI_Request request_;
+    };
+
+    // -----------------------------------------------------------------
+    // utility function to add a new request to the list to be tracked
+    void add_to_request_list(future_data_ptr data)
```
hkaiser: Can this go into `namespace detail`?

hkaiser:
> I plan on upgrading the cuda_future_helper code to mimic the same style so that we can have hpx::cuda::async(....) and hpx::cuda::apply(....) (I will add hpx::mpi::apply thought I'm not sure anyone will want to use a fire and forget mpi call... maybe). When I do that, I can see cuda streams being an analogue to the executor/communicator pattern.

May I suggest to go the executor route directly? The required effort is not much larger than what you plan to create. It would however avoid cluttering the API and better integrate your work with the existing infrastructure.

libs/mpi/include/hpx/mpi/mpi_future.hpp
biddisco: yes, this should be in detail, but I'd argue that the `poll` function is intended to be referenced by the end user so it should be in the top level namespace

biddisco:
```
    sched->set_user_polling_function(&hpx::mpi::poll);
    sched->add_scheduler_mode(hpx::threads::policies::enable_user_polling);
```

biddisco:
Yes you're right. I could just do away with the `hpx::mpi::async` and `hpx::cuda:async` and use an executor for that.

Regarding using the current thread. For posting async MPI requests such as `MPI_Isend`, `MPI_Irecv`, `MPI_Ibcast` and all the rest, yes, I'm happy that they can go out on the calling thread. `MPI_THREAD_MULTIPLE` must of course be used in the `init` function, but if you want low latency, you simply cannot start creating tasks to do these operations, they should be immediate.

That's not to say that we will not permit users to create separate pool and launch MPI related work on that pool (via a pool executor). This is absolutely intended to be supported. for now however, we do want to be able to launch work on the current thread and have any threads doing the polling. the purpose of this 
```
    sched->set_user_polling_function(&hpx::mpi::poll);
    sched->add_scheduler_mode(hpx::threads::policies::enable_user_polling);
```
is to tell the pool to execute the callback on any thread in the pool. the user can however disable polling on the main pool and only enable it in an mpi pool (example/test for this will be provided)

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,186 @@
...
+                               , "R", rank
+                               , "request", hpx::debug::hex<8>(original_req)
+                               , "list size", active_requests_.size());
+            }
+            mpi_debug.timed(timer
+                           , "R", rank
+                           , "request", hpx::debug::hex<8>(original_req)
+                           , "list size", active_requests_.size());
+
+            ++i;
```
teonnik: This is a bug: ++i should go in the else clause. Note that erase returns the next iterator.

Also, I am not sure if the debug statement is needed here. The other debug statements seem sufficient.

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,186 @@
...
+    hpx::future<void> async(F f, Ts &&...ts)
+    {
+        future_data_ptr data =
+            new future_data(future_data::init_no_addref{});
+
+        // invoke the call to MPI_Ixxx
+        f(std::forward<Ts>(ts)..., &data->request_);
+
+        // add the new shared state to the list for tracking
+        add_to_request_list(data);
```
teonnik: The PR can easily support `MPI_THREAD_SERIALIZED` by rewriting this to sth like:

```
{
        std::lock_guard<std::mutex> lk(list_mtx_);
        // invoke the call to MPI_Ixxx
        f(std::forward<Ts>(ts)..., &data->request_);

        // add the new shared state to the list for tracking
        add_to_request_list(data);
}
```

This ensures that non-blocking calls are serialized even if made from different threads.

hpx/runtime/threads/policies/scheduler_base.hpp
```diff
@@ -307,6 +315,8 @@ namespace hpx { namespace threads { namespace policies {
 
         std::atomic<std::int64_t> background_thread_count_;
 
+        std::function<void(void)> user_polling_function_;
```
hkaiser: Please use `hpx::util::unique_function_nonser` for this. It's more lightweight than any other `std::function` implementations we know.

hpx/runtime/threads/detail/scheduling_loop.hpp
```diff
@@ -933,6 +933,12 @@ namespace hpx { namespace threads { namespace detail
                     params.inner_();
             }
 
+            if (scheduler.SchedulingPolicy::get_scheduler_mode() &
+                    policies::enable_user_polling)
+            {
+                scheduler.user_polling_function();
+            }
```
hkaiser: Shouldn't this better go to the background thread instead?

libs/mpi/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+  if (MPI_LIBRARY)
+    target_link_libraries(hpx::mpi INTERFACE ${MPI_LIBRARY})
+  endif()
+  if (MPI_EXTRA_LIBRARY)
+    target_link_libraries(hpx::mpi INTERFACE ${MPI_EXTRA_LIBRARY})
+  endif()
+
+  set_property(TARGET hpx::mpi PROPERTY INTERFACE_COMPILE_OPTIONS ${MPI_CXX_COMPILE_FLAGS})
+    set_property(TARGET hpx::mpi PROPERTY INTERFACE_COMPILE_DEFINITIONS ${MPI_CXX_COMPILE_DEFINITIONS})
+endif()
```
hkaiser: I would suggest to not re-discover MPI here but rely on the MPI logic that is already in place. Simply do not add this module to the mix if `hpx::mpi` is not a target.

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,186 @@
+//  Copyright (c) 2019 John Biddiscombe
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef HPX_LIBS_MPI_FUTURE_HPP
+#define HPX_LIBS_MPI_FUTURE_HPP
+
+#include <mpi.h>
```
hkaiser: Please #include HPX headers first (at least `#include <hpx/config.hpp>`) to avoid issues with Windows based compilations.

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,186 @@
...
+#include <array>
+#include <utility>
+#include <tuple>
+#include <list>
+#include <mutex>
+//
+#include <hpx/local_lcos/promise.hpp>
+#include <hpx/util/yield_while.hpp>
+#include <hpx/memory/intrusive_ptr.hpp>
+#include <hpx/debugging/print.hpp>
```
hkaiser: This needs clang-formatting.

libs/mpi/include/hpx/mpi/mpi_future.hpp
hkaiser: Should this lock be acquired unconditionally only to support `MPI_THREAD_SERIALIZED`?

hkaiser: Also, I don't like the fact that we are using a `std::mutex` on a HPX thread. This could possibly stall the whole scheduler progress.

biddisco: We can perform a check at init time to see what level of thread support is available and then conditionally make use of a lock when invoking mpi functions.

Note that the polling function DOES happen on a std thread and not on an hpx thread and we use std::mutexes there (with a try_lock to avoid uneccessary waits). The sending happens on an HPX thread and we could use an hpx mutex - however, we wish to avoid high latencies on sends, so it might still be worth using a std::mutex. We don't want a thread suspension to kick in when we have an 'urgent' send, even an asynchronous one. I will look into options here. If the user has selected serialized threading in MPI, then they probably aren't too bothered by performance and we can use a lock...

hpx/runtime/threads/detail/scheduling_loop.hpp
biddisco: The background thread runs as an hpx task and can be suspended. Since we're only calling MPI on this and it pretty much expects to be on a std::thread type thread, I'm not sure that it really mattters where we call it. For this initial implementation I put it outside the background threading work to ensure I know what's going on, however it could be moved into the background work once I'm happy that all is behaving sensibly. (does it really make any differnce where we do it? what are the pros and cons?)

libs/mpi/CMakeLists.txt
biddisco: I will move all the logic to a common location. currently if networking is off, the find mpi stuff doesn't get called, so I duplicated it here. I want to fix it as it's implemented incorrectly anyway, it uses MPI_CXX_LIBRARY and includes instead of C_LIBRARY and includes etc etc

libs/mpi/include/hpx/mpi/mpi_future.hpp
hkaiser: Fine with me. Note however that a `std::mutex` might incur much higher overheads than an hpx mutex.

hpx/runtime/threads/detail/scheduling_loop.hpp
hkaiser: Well, cons are that the current implementation adds overheads to the inner part of the scheduling loop for everything, not just for your MPI executor.

biddisco:
Comment: I do not want to introduce mpi executors into this PR yet for three reasons
1) We have code using this at CSCS that works well and will not benefit from adding an executor. I would like our users to be able to move to hpx master rather than working on a branch.
2) Using an executor for mpi async calls and using `hpx::async` dispatch requires wrapping each mpi call with a forwarding layer to handle the extra parameter (MPI_Request) in order to make the `return_type` handling work in the hpx framework. 
3) Once a user has an mpi_executor, they will attempt to use it on non mpi functions and I would like to spend more time on idiot-proofing this.
Items 2,3 require a little work that I would prefer to leave until a later date.

biddisco:
Note that point 2 above omits one important detail. 
different releases of MPI have different versions/numbers of async calls and we will have to make sure that all are protected via blocks of `#if MPI_VERSION_MAJOR/MINOR` that will need to be maintained. All distros such as mvapch, mpich, openmpi, craympi, intel mpi will need to be checked for which calls are present and supported. In principle they should all be the same for a given mpi version. 

...But not wrapping them places the burden on the user and makes our implementation much simpler.

hkaiser:
> Comment: I do not want to introduce mpi executors into this PR yet for three reasons
> 
> We have code using this at CSCS that works well and will not benefit from adding an executor. I would like our users to be able to move to hpx master rather than working on a branch.

This would make an API 'official' that we know will change. I wouldn't like for this to happen, if we can avoid it.

> Using an executor for mpi async calls and using hpx::async dispatch requires wrapping each mpi call with a forwarding layer to handle the extra parameter (MPI_Request) in order to make the return_type handling work in the hpx framework.

Sure, so what? 

OTOH, I think that using decltype(auto) instead of some explicit return type calculation could mitigate this issue.

> Once a user has an mpi_executor, they will attempt to use it on non mpi functions and I would like to spend more time on idiot-proofing this.

The explicit wrappers you mentioned above could be used to circumvent that.

All in all, I'm all in favor of moving this forward, but not if it manifests APIs we know we want to change in the future. This will essentially force us to support the non-executor interface forever. 

How about a compromise: I will try to find some time to do the necessary work to wrap your `mpi::async` into an executor (I would like to rename that while I'm at it, btw). Your main reason for pushing this right now as is seems to be lack of time to do additional work, so I hope this would allow for a mutually agreeable solution.

hkaiser:
> Note that point 2 above omits one important detail.
> different releases of MPI have different versions/numbers of async calls and we will have to make sure that all are protected via blocks of #if MPI_VERSION_MAJOR/MINOR that will need to be maintained. All distros such as mvapch, mpich, openmpi, craympi, intel mpi will need to be checked for which calls are present and supported. In principle they should all be the same for a given mpi version.
> ...But not wrapping them places the burden on the user and makes our implementation much simpler.

I would assume that your users currently employ only a handful of MPI functions in their applications anyways. So let's start with those functions and see how far we can push things.

hkaiser:
@biddisco many of the comments made on this PR have not been answered/addressed yet. Could you respond, please?

libs/mpi/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+  hpx_option(HPX_WITH_MPI_FUTURES BOOL
+    "Enable support for returning futures from MPI asynchronous calls"
+    " (default: ON if networking+MPI already enabled)"
+    ON CATEGORY "MPI")
+else()
+  hpx_option(HPX_WITH_MPI_FUTURES BOOL
+    "Enable support for returning futures from MPI asynchronous calls"
+    " (default: ON if networking+MPI already enabled)"
+    OFF CATEGORY "MPI")
+endif()
```
hkaiser: What's the difference between the two options (except for the default value)?

hkaiser: We usually name module-specific configuration constants as `HPX_<modulename>_WITH_<option>`. Could you please adapt this?

libs/mpi/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+find_package(MPI REQUIRED)
+
+if(${CMAKE_VERSION} VERSION_LESS "3.12.0")
+    message(ERROR "cmake 3.12 or newer required")
+endif()
+
+# All cmake version don't have the same found variable set
+if(NOT MPI_FOUND AND NOT MPI_C_FOUND)
+  hpx_error("MPI could not be found and hpx::mpi library was requested\n
+  set MPI_MPI_CXX_COMPILER to point to your MPI installation's mpicxx")
```
hkaiser: This error message is misleading. Finding the MPI libraries is not related to finding mpicxx.

libs/mpi/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+    target_link_libraries(hpx::mpi INTERFACE ${MPI_EXTRA_LIBRARY})
+  endif()
+
+  set_property(TARGET hpx::mpi PROPERTY INTERFACE_COMPILE_OPTIONS ${MPI_CXX_COMPILE_FLAGS})
+    set_property(TARGET hpx::mpi PROPERTY INTERFACE_COMPILE_DEFINITIONS ${MPI_CXX_COMPILE_DEFINITIONS})
+endif()
+
+# Default location is $HPX_ROOT/libs/mpi/include
+set(mpi_headers
+    hpx/mpi/mpi_future.hpp
```
hkaiser: This misses the header `comm_wrapper.hpp`

libs/mpi/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+  # Ensure compatibility with older versions
+  if (MPI_LIBRARY)
+    target_link_libraries(hpx::mpi INTERFACE ${MPI_LIBRARY})
+  endif()
+  if (MPI_EXTRA_LIBRARY)
+    target_link_libraries(hpx::mpi INTERFACE ${MPI_EXTRA_LIBRARY})
+  endif()
+
+  set_property(TARGET hpx::mpi PROPERTY INTERFACE_COMPILE_OPTIONS ${MPI_CXX_COMPILE_FLAGS})
+    set_property(TARGET hpx::mpi PROPERTY INTERFACE_COMPILE_DEFINITIONS ${MPI_CXX_COMPILE_DEFINITIONS})
```
hkaiser: Indentation?

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,447 @@
...
+                : hpx::lcos::detail::future_data<int>(no_addref), request_(request)
+            {}
+
+            // constructor used for creation directly by invoke
+            future_data(init_no_addref no_addref)
+                : hpx::lcos::detail::future_data<int>(no_addref)
+            {}
+
+            // The native MPI request handle owned by this future data
+            MPI_Request request_;
```
hkaiser: Do you need `public` access to this member?

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,447 @@
...
+            }
+        }
+
+        // -----------------------------------------------------------------
+        // used internally to add a request to the main polling vector/list
+        // that is passed to MPI_Testany
+        void add_to_request_list(future_data_ptr data)
+        {
+            // this will make a copy and increment the ref count
+            active_futures_.push_back(data);
```
hkaiser: How does this ensure thread-safety?

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,447 @@
...
+#include <hpx/local_lcos/promise.hpp>
+#include <hpx/memory/intrusive_ptr.hpp>
+#include <hpx/resource_partitioner/partitioner.hpp>
+#include <hpx/runtime/thread_pool_helpers.hpp>
+#include <hpx/threading_base/print.hpp>
+
+// The preferred mode of operation uses a std::vector and MPI_Testany
+// but the use of a list and individual tests is supported for legacy
+// debug reasons, uncomment one of these #defines to choose which is used
+//#define HPX_MPI_LIST_STORAGE
```
hkaiser: Please remove this commented `#define` and the corresponding code below.

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,447 @@
...
+                    , debug::str<>("Poll <ERR>")
+                    , "MPI_ERROR", err_buff
+                    , "status", debug::dec<>(status.MPI_ERROR)
+                    , "index", debug::dec<>(index)
+                    , "flag", debug::dec<>(flag));
+                // clang-format on
+            }
+        };
+        // if there are more than 25% NULL request handles in our lists, compact them
+        if (detail::active_futures_.size()>0) {
```
hkaiser: Please use `!detail::active_futures_.empty()` instead.

libs/mpi/include/hpx/mpi/mpi_future.hpp
hkaiser: Do you need `public` access to this member?

libs/mpi/include/hpx/mpi/mpi_future.hpp
hkaiser: Please remove this `#define` and the code below that is wrapped by it.

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,447 @@
...
+    // return a future from an async call to MPI_Ixxx function
+    template <typename F, typename... Ts>
+    hpx::future<int> async(F f, Ts &&...ts)
+    {
+        // create a future data shared state
+        detail::future_data_ptr data =
+            new detail::future_data(detail::future_data::init_no_addref{});
+        // invoke the call to MPI_Ixxx, ignore the returned result for now
+        int result = f(std::forward<Ts>(ts)..., &data->request_);
+        (void)result; // silence unused var warning
```
hkaiser: Why is the return value ignored?

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,447 @@
...
+                    detail::active_futures_[unsigned(index)]->set_data(MPI_SUCCESS);
+                    // remove the request from our list to prevent retesting
+                    detail::active_requests_[unsigned(index)] = MPI_REQUEST_NULL;
+                    detail::active_futures_[unsigned(index)]  = nullptr;
+                }
+            }
+            else {
+                keep_trying = false;
+                int N = 1023;
+                char err_buff[1024];
```
hkaiser: Please use `err_buff[N+1]` instead.

libs/CMakeLists.txt
```diff
@@ -108,21 +112,37 @@ set(MODULE_FORCE_LINKING_CALLS)
...
+      "${CONFIG_STRINGS_MODULES_INCLUDES}#include <hpx/${lib}/config/config_strings.hpp>\n")
+    set(CONFIG_STRINGS_MODULES_ENTRIES
+      "${CONFIG_STRINGS_MODULES_ENTRIES}\n        { \"${lib}\", ${lib}::config_strings },")
+  endif()
 endforeach()
 
+configure_file(
+    "${PROJECT_SOURCE_DIR}/cmake/templates/libs_enabled.hpp.in"
+    "${PROJECT_BINARY_DIR}/hpx/config/libs_enabled.hpp"
+    @ONLY)
```
hkaiser: Where is the generated file used?

libs/mpi/tests/unit/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
...
+    add_hpx_unit_compile_test(
+      "modules.mpi"
+      ${compile_test}
+      SOURCES ${sources}
+      ${${compile_test}_FLAGS}
+      FOLDER "Tests/Unit/Modules/MPI/CompileOnly")
+
+  endforeach()
+
+endif()
```
hkaiser: This is not needed, there are no compile-only tests.

libs/mpi/tests/unit/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
...
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+set(tests
+  mpi_ring_async
+#  mpi_simple_async
+  )
+
+set(mpi_ring_async_PARAMETERS THREADS_PER_LOCALITY 4)
```
hkaiser: Should this test run on more than one locality?

biddisco:
> How about a compromise: I will try to find some time to do the necessary work to wrap your `mpi::async` into an executor (I would like to rename that while I'm at it, btw). 

Why? What is the point? MPI is a C library, with a C API that mpi users are accustomed to. The API implemented is very simple and straightforward and easy to maintain. Adding an executor makes the code harder to use, harder to maintain, forwards a simple call to a C API through template specializations and customization points, requires wrapping every MPI call and gains the user nothing at all. It is simply a waste of time. 
Who actually cares if they call
```
auto mpi_exec = hpx::mpi::executor();
auto f = hpx::async(mpi_exec, hpx::mpi::wrapper::MPI_Isend, a,b,c,d);
```
or
```
auto f = hpx::mpi::async(MPI_Isend, a,b,c,d);
```

>All in all, I'm all in favor of moving this forward, but not if it manifests APIs we know we want to change in the future. This will essentially force us to support the non-executor interface forever.

I disagree, there is no need to have a more complex API - and we can change the API if we want at any time (we can add a deprecation message if we want).

hkaiser:
> > How about a compromise: I will try to find some time to do the necessary work to wrap your mpi::async into an executor (I would like to rename that while I'm at it, btw).
> 
> Why? What is the point? MPI is a C library, with a C API that mpi users are accustomed to. The API implemented is very simple and straightforward and easy to maintain. Adding an executor makes the code harder to use, harder to maintain, forwards a simple call to a C API through template specializations and customization points, requires wrapping every MPI call and gains the user nothing at all. It is simply a waste of time.

None of the above (see my PR related to this - #4407). The overall API of HPX stays consistent while adding a `hpx::mpi::async` clutters the API with yet another specialized function.

> Who actually cares if they call
> auto mpi_exec = hpx::mpi::executor();
> auto f = hpx::async(mpi_exec, hpx::mpi::wrapper::MPI_Isend, a,b,c,d);
> 
> or
> auto f = hpx::mpi::async(MPI_Isend, a,b,c,d);

I do care, otherwise we wouldn't have tis discussion.

> > All in all, I'm all in favor of moving this forward, but not if it manifests APIs we know we want to change in the future. This will essentially force us to support the non-executor interface forever.
> 
> I disagree, there is no need to have a more complex API - and we can change the API if we want at any time (we can add a deprecation message if we want).

See #4407. `hpx::async(mpi::executor{}, ...)` is nicely consistent with everything we have so far. And it's not more complex than calling a specialized `mpi::async(...)`. The additional complexity is minimal if at all noticable.

biddisco:
OK. what you have done is fine. Thanks. I will clean up some of the other issues. There are some references to comm wrappers that should not be in this PR (they are part of the next one). I will merge your commit into my PR and force push once I've removed the other crud.

biddisco:
I have merged @hkaiser 's mpi_executor branch into this one with a small fix and removed the unwanted comms stuff that was left in by mistake.

I believe that most of the issues with this are resolved. One outstanding question was 

> "Well, cons are that the current implementation adds overheads to the inner part of the scheduling loop for everything, not just for your MPI executor."

The overhead is really just a single `if (mode_set)` clause and on consideration, I believe it is better to put the polling outside of the background thread checks. Firstly, MPI may take locks using pthread style mutexes and the existing hpx background work checks happen on an hpx worker task that should not ideally be suspended (it doesn't really make any difference since it's in a background task, but the logic in the main scheduling loop is not expecting this behaviour). Secondly, the overheads are lower using a non background task and using the main worker thread because the hpx background checks are placed on a hpx task that is generated 'on the fly' and context switched into, whereas the mpi::poll is just a function call. 

In principle we can put this in a background task, but I would advise against it until this code has been field tested and there are more tests/users in place. There is probably more chance of a deadlock rather than less if we move this inside the background task, and it is more expensive to do it, so there needs to be a good reason. If we could instead hide the cost of the extra `if` in the main scheduling loop, then there would be no reason to worry. I'll look into a better integration of this in the main loop as the current implementation was really just a first attempt to get stuff going - though it has now proved to be a very useful thing.

libs/mpi/CMakeLists.txt
```diff
@@ -0,0 +1,103 @@
...
+    hpx/mpi/mpi_future.hpp
+)
+
+# Default location is $HPX_ROOT/libs/mpi/include_compatibility
+set(mpi_compat_headers)
+
+set(mpi_sources)
+
+include(HPX_AddModule)
+add_hpx_module(mpi
```
msimberg: Was the plan to call this `mpi_support`? Or should we stick with `mpi`?

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,515 @@
...
+#include <hpx/threading_base/print.hpp>
+#include <hpx/lcos/detail/future_data.hpp>
+
+// The preferred mode of operation uses a std::vector and MPI_Testany
+// but the use of a list and individual tests is supported for legacy
+// debug reasons, uncomment one of these #defines to choose which is used
+//#define HPX_MPI_LIST_STORAGE
+#define HPX_MPI_VECTOR_STORAGE
+
+namespace hpx { namespace mpi {
```
msimberg: My preference would be `hpx::mpi::experimental` (i.e. `experimental` as the leaf namespace, just like `detail`). (Note: if the module name is `mpi_support` the namespace should probably be `hpx::mpi_support`).

libs/mpi/CMakeLists.txt
biddisco: Originally I thought of MPI_Support, but I think mpi is fine, though I would like the CMake variable to be `HPX_WITH_MPI_SUPPORT`. I need to write some docs to explain a few things - like what happens if you enable `HPX_WITH_MPI_SUPPORT` and at the same time have `HPX_WITH_NETWORKING=ON` and MPI PARCELPORT enabled - this has not been tested - it should work, but users would need to be super careful about which communicators they use. I suspect we need to isolate the communicator we use in the `MPI_Testany` and suchlike. Making some notes on this would be sensible. A blog articles with the same info in too (as discussed yesterday).

libs/mpi/include/hpx/mpi/mpi_future.hpp
biddisco: `hpx::mpi::experimental` is fine

biddisco: I will do some more tidying up of this PR today

libs/mpi/include/hpx/mpi/mpi_future.hpp
```diff
@@ -0,0 +1,515 @@
...
+        ~enable_user_polling()
+        {
+            mpi::finalize(pool_name_);
+        }
+
+        template <typename F>
+        void wait(F&& f)
+        {
+            return mpi::wait(std::forward<F>(f));
+        }
```
hkaiser: I know, I added this member function `wait` here. I think we should not have that, however. Let's not advertise the existence of a global wait facility.

libs/mpi/tests/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+      add_hpx_pseudo_dependencies(tests.performance.modules tests.performance.modules.mpi)
+      add_subdirectory(performance)
+    endif()
+
+    if (HPX_WITH_TESTS_HEADERS)
+      add_hpx_header_tests(
+        modules.mpi
+        HEADERS ${mpi_headers}
+        HEADER_ROOT ${PROJECT_SOURCE_DIR}/include
+        NOLIBS
```
msimberg: I think you'll need to remove `NOLIBS` here. If header tests fail on CircleCI you can try removing it here (not having `NOLIBS` lets the header tests use includes from the main include directory).

msimberg:
By the way, could you please add yourself (your GitHub username) as the code owner for this module: https://github.com/STEllAR-GROUP/hpx/blob/master/.github/CODEOWNERS.

libs/mpi/CMakeLists.txt
```diff
@@ -0,0 +1,103 @@
+# Copyright (c) 2019 The STE||AR-Group
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+cmake_minimum_required(VERSION 3.3.2 FATAL_ERROR)
```
teonnik: This line should probably go. It is overriding the policies set in the top CMakeLists.txt file. In particular, policy CMP0074 which is useful when MPI is installed in a custom directory: the MPI_ROOT variable is ignored. Moreover, the version specified is less than the version of the top CMake file.

cmake/HPX_SetupApex.cmake
```diff
@@ -53,6 +53,10 @@ if(HPX_WITH_APEX AND NOT TARGET APEX::apex)
     target_link_libraries(APEX::apex INTERFACE APEX::apex)
   else()
     target_link_libraries(APEX::apex INTERFACE apex)
+    if(TARGET Mpi::mpi)
+      target_compile_definitions(apex PRIVATE APEX_HAVE_MPI)
+      target_link_libraries(APEX::apex PRIVATE Mpi::mpi)
+    endif()
```
msimberg, line 59: Does APEX not set up this dependency by itself? We might still need this here because APEX being special... It looks like APEX has a CMake variable called `USE_MPI` which we could set instead but I don't know if that's intended for public use. @khuck?

libs/CMakeLists.txt
```diff
@@ -112,6 +113,10 @@ endif()
 hpx_info("")
 hpx_info("Configuring modules:")
 
+# generate a file thatliosts all enabled modules for checks that might be
```
msimberg, line 116:
```suggestion
# Generate a file that lists all enabled modules for checks that might be
```

libs/local_lcos/CMakeLists.txt
```diff
@@ -45,6 +45,8 @@ add_hpx_module(local_lcos
   SOURCES ${local_lcos_sources}
   HEADERS ${local_lcos_headers}
   COMPAT_HEADERS ${local_lcos_compat_headers}
+  EXCLUDE_FROM_GLOBAL_HEADER
+    "hpx/local_lcos/force_linking.hpp"
```
msimberg, line 49: Good catch on this one. I wonder if we can just use the generated force linking functionality instead though...

libs/mpi/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+
+# setup MPI, if necessary
+if(NOT TARGET Mpi::mpi)
+  include(HPX_SetupMPI)
+  setup_mpi()
+endif()
+
+# Default location is $HPX_ROOT/libs/mpi/include
+set(mpi_headers
+    hpx/mpi/force_linking.hpp
```
msimberg, line 36: Same here, I wonder if we can't just use the usual force linking functionality.

libs/threading_base/include/hpx/threading_base/scheduler_base.hpp
```diff
@@ -270,8 +271,18 @@ namespace hpx { namespace threads { namespace policies {
...
         }
 
+        void set_user_polling_function(util::function_nonser<void()> const& f)
+        {
+            user_polling_function_ = f;
+        }
+
+        void user_polling_function()
+        {
+            user_polling_function_();
```
msimberg, line 281:
```suggestion
            if (user_polling_function_) user_polling_function_();
```
or even

```suggestion
            HPX_ASSERT(user_polling_function_);
            user_polling_function_();
```
?

libs/mpi/src/mpi_future.cpp
```diff
@@ -0,0 +1,434 @@
...
+
+        void register_polling(hpx::threads::thread_pool_base& pool)
+        {
+            auto* sched = pool.get_scheduler();
+
+            mpi_debug.debug(debug::str<>("Setting mode"),
+                detail::get_mpi_info(), "enable_user_polling");
+
+            // always set polling function before enabling polling
+            sched->set_user_polling_function(&hpx::mpi::poll);
```
msimberg, line 332: It would be nice if this weren't tied to the scheduler and could be done with just plain HPX threads. If another executor will also silently set the user polling function, it should at least support setting multiple such functions. That's for outside this PR though.

cmake/HPX_SetupApex.cmake
hkaiser: I left this in as the original PR was handling this in a special way. I have no means of verifying it, however.

libs/local_lcos/CMakeLists.txt
hkaiser: Yes, I thought about it, but then decided that this would be outside of the scope of this PR.

libs/mpi/CMakeLists.txt
hkaiser: I'll create a separate ticket for this.

libs/mpi/src/mpi_future.cpp
hkaiser: Yes, I agree. Let's leave this for another PR though. I'll create a ticket for this.

libs/mpi/CMakeLists.txt
```diff
@@ -0,0 +1,103 @@
...
+)
+
+if (HPX_WITH_APEX)
+  cmake_policy(SET CMP0079 NEW)
+  # Reload the Apex setup but this time, with MPI
+  set(HPX_HAVE_LIB_MPI ON)
+  target_compile_definitions(apex_flags INTERFACE APEX_HAVE_MPI)
+  target_link_libraries(apex PRIVATE ${MPI_CXX_LIBRARY} ${MPI_C_LIBRARY} ${MPI_LIBRARIES})
+  target_include_directories(apex PRIVATE ${MPI_INCLUDE_PATH} ${MPI_C_INCLUDE_DIRS})
+endif()
```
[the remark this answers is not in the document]
hkaiser: I moved this to the SetupAPEX, see below.

weilewei:
This PR works for my ring G algorithm test on Summit. Thanks!

Source code: https://github.com/weilewei/Ring_example_MPI_CUDA/blob/hpx_mpi_async/G2_ring_hpx.cpp

libs/mpi/include/hpx/mpi/mpi_future.hpp
hkaiser: Yes, you do...

hpx/runtime/threads/detail/scheduling_loop.hpp
hkaiser: See also #4507

hkaiser:
@msimberg, @biddisco I think I have addressed all review comments now. Please review one last time.

hkaiser:
> This PR works for my ring G algorithm test on Summit. Thanks!
> Source code: https://github.com/weilewei/Ring_example_MPI_CUDA/blob/hpx_mpi_async/G2_ring_hpx.cpp

Thanks @weilewei. Please note however, that once merged, the facilities introduced by this PR will move to the `namespace hpx::mpi::experimental`. You will have to adapt your code accordingly.

weilewei:
> > This PR works for my ring G algorithm test on Summit. Thanks!
> > Source code: https://github.com/weilewei/Ring_example_MPI_CUDA/blob/hpx_mpi_async/G2_ring_hpx.cpp
> 
> Thanks @weilewei. Please note however, that once merged, the facilities introduced by this PR will move to the `namespace hpx::mpi::experimental`. You will have to adapt your code accordingly.

noted. Thanks.

cmake/HPX_SetupApex.cmake
khuck: @msimberg APEX doesn't depend on MPI support when used with HPX.  In fact, in my head they are mutually exclusive... why does APEX require MPI for this PR?

hkaiser: @khuck no idea. @biddisco had that in his initial version for this PR. I'll go ahead and remove this, then.

msimberg: I don't know the original reason either, but does APEX require MPI to be on to correctly handle multiple localities? I think the idea is that you can have the module introduced in this PR enabled with MPI on, but HPX as would have networking off (and thus no MPI parcelport). In that case HPX has no intrinsic concept of "localities" but would instead verbatim use MPI ranks to identify different processes. If APEX handles that no matter if MPI is enabled or not I think this can be removed, but only @biddisco knows for sure.

hkaiser: I have removed it for now. If things break for @biddisco after that we can always re-introduce this later. I'd rather have this PR in master asap than to wait.

hkaiser:
@msimberg I think the APEX failure is caused by a missing dependency on Apex during pkgconfig handling. I don't understand the CUDA problems, however. Some seem to be caused by some build system issue, the other looks like NVCC being braindead (no news there). Can we merge this now?

msimberg:
Yes, the APEX error is most likely caused by me enabling build tests. The CUDA error is strange, but apparently already on master. They are most likely not caused by this PR. I'll have to look at these next week, but let's go ahead with this to avoid further rebasing.

PR #4280, opened 2019-12-18
Fix scheduler related configuration values

msimberg:
Fixes some very bad configuration values related to scheduler stealing etc. This was bad copy-pasting by me.

### DIFF
--- a/hpx/runtime/threads/detail/scheduling_loop.hpp
+++ b/hpx/runtime/threads/detail/scheduling_loop.hpp
@@ -930,7 +930,11 @@ namespace hpx { namespace threads { namespace detail
 #endif
                 // call back into invoking context
                 if (!params.inner_.empty())
+                {
                     params.inner_();
+                    context_storage = hpx::basic_execution::this_thread::
+                        detail::get_agent_storage();
+                }
             }
 
             // something went badly wrong, give up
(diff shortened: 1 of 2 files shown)


msimberg:
@kordejong these are probably causing at least part your weird traces. These are also values you can play around with as I don't think we've put too much effort into finding optimal values (and they'll likely vary by application). You can change them with `--hpx:ini=hpx.thread_queue.<option_name>=X`.

msimberg:
@hkaiser this is a clear regression since the last release, so this would be good to have in 1.4.0.

msimberg:
This also revealed a problem in how agent storage was handled (i.e. not handled) in embedded schedulers. I think it's as simple as the fix I've proposed (reset the `context_storage` variable in the scheduling loop after the embedded scheduler has yielded to the parent scheduler). The previously failing tests now pass again, but I'll wait to see if this introduces other problems.

msimberg:
> Interesting... I was not aware that the `context_storage` may change during execution of the callbacks...

It's the thread stealing. The faulty config values essentially prohibited thread stealing, and with that fixed the embedded scheduler worker threads can get stolen while suspended. When they wake up they use the storage of the wrong OS thread.

hkaiser:
Nice catch!

PR #4281, opened 2019-12-18
Async customization

biddisco:
This fixes a problem found by @msimberg and improves the test so that in future such problems would be detected.

Force pushed because I pushed the wrong branch ref first time.


hkaiser:
@biddisco What problem does this PR solve?

biddisco:
@msimberg noticed that the implementation I had for the `then_execute` was incorrect and I found by testing that it indeed was wrong. The continuation was fired before the future that should trigger it was ready (I had left an `apply` in there and not used the correct continuation structure)

msimberg:
Thanks @biddisco!

PR #4284, opened 2019-12-19
Remove warnings about missing pkgconfig on Windows

msimberg:


msimberg:
Adding this to 1.4.0 as well since there's time and it's a small change.

PR #4285, opened 2019-12-19
Delete apex.hpp

msimberg:
The file is unused and has been replaced by external_timer.hpp


msimberg:
It is indeed unrelated. It looks like it's a pretty reproducible failure, but I haven't had the time to look into it.

PR #4287, opened 2019-12-25
Forking boost::lexical_cast as a new module

K-ballo:


libs/lexical_cast/include/hpx/lexical_cast/detail/basic_pointerbuf.hpp
```diff
@@ -0,0 +1,149 @@
...
+          : base_type()
+        {
+            this_type::setbuf(0, 0);
+        }
+        const charT* getnext()
+        {
+            return this->gptr();
+        }
+
+#ifndef BOOST_NO_USING_TEMPLATE
```
hkaiser: Can this be removed?

libs/lexical_cast/include/hpx/lexical_cast/detail/converter_lexical.hpp
```diff
@@ -0,0 +1,170 @@
...
+//        Cheng Yang, Matthew Bradbury, David W. Birdsall, Pavel Korzh and other Boosters
+// when:  November 2000, March 2003, June 2005, June 2006, March 2011 - 2014
+
+#ifndef HPX_LEXICAL_CAST_DETAIL_CONVERTER_LEXICAL_HPP
+#define HPX_LEXICAL_CAST_DETAIL_CONVERTER_LEXICAL_HPP
+
+#include <hpx/config.hpp>
+
+#include <boost/type_traits/has_left_shift.hpp>
+#include <boost/type_traits/has_right_shift.hpp>
```
hkaiser: Should those be recreated here as well? Otherwise we replace one boost dependency with another...

libs/lexical_cast/include/hpx/lexical_cast/detail/converter_numeric.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+#ifndef HPX_LEXICAL_CAST_DETAIL_CONVERTER_NUMERIC_HPP
+#define HPX_LEXICAL_CAST_DETAIL_CONVERTER_NUMERIC_HPP
+
+#include <hpx/config.hpp>
+
+#include <limits>
+#include <type_traits>
+
+#include <boost/numeric/conversion/cast.hpp>
```
hkaiser: Can this be forked as well?

libs/lexical_cast/include/hpx/lexical_cast/detail/inf_nan.hpp
```diff
@@ -0,0 +1,158 @@
...
+//        Cheng Yang, Matthew Bradbury, David W. Birdsall, Pavel Korzh and other Boosters
+// when:  November 2000, March 2003, June 2005, June 2006, March 2011 - 2014
+
+#ifndef HPX_LEXICAL_CAST_DETAIL_INF_NAN_HPP
+#define HPX_LEXICAL_CAST_DETAIL_INF_NAN_HPP
+
+#include <hpx/config.hpp>
+
+#include <boost/math/special_functions/fpclassify.hpp>
+#include <boost/math/special_functions/sign.hpp>
```
hkaiser: Same here: should we recreate those functionalities in HPX?

libs/lexical_cast/tests/unit/lexical_cast_float_types_test.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+#if defined(__INTEL_COMPILER)
+#pragma warning(disable: 193 383 488 981 1418 1419)
+#elif defined(HPX_MSVC_WARNING_PRAGMA)
+#pragma warning(disable: 4097 4100 4121 4127 4146 4244 4245 4511 4512 4701 4800)
+#endif
+
+#include <hpx/lexical_cast.hpp>
+
+#include <boost/cstdint.hpp>
```
hkaiser: Could we use `# include <cstdint>` instead?

libs/lexical_cast/tests/unit/lexical_cast_inf_nan_test.cpp
```diff
@@ -0,0 +1,143 @@
...
+#if defined(__INTEL_COMPILER)
+#pragma warning(disable: 193 383 488 981 1418 1419)
+#elif defined(HPX_MSVC_WARNING_PRAGMA)
+#pragma warning(disable: 4097 4100 4121 4127 4146 4244 4245 4511 4512 4701 4800)
+#endif
+
+#include <hpx/lexical_cast.hpp>
+
+#include <boost/math/special_functions/sign.hpp>
+#include <boost/math/special_functions/fpclassify.hpp>
```
hkaiser: Should these be replaced with our own?

libs/lexical_cast/tests/unit/lexical_cast_integral_types_test.cpp
```diff
@@ -0,0 +1,412 @@
...
+#pragma warning(disable: 193 383 488 981 1418 1419)
+#elif defined(HPX_MSVC_WARNING_PRAGMA)
+#pragma warning(disable: 4097 4100 4121 4127 4146 4244 4245 4511 4512 4701 4800)
+#endif
+
+#include <hpx/lexical_cast.hpp>
+
+#include <hpx/testing.hpp>
+
+#include <boost/type_traits/integral_promotion.hpp>
```
hkaiser: Should this be forked as well?

libs/logging/src/logging.cpp
```diff
@@ -78,7 +78,7 @@ namespace hpx { namespace util {
                 }
                 return hpx::util::logging::level::debug;
             }
-            catch (boost::bad_lexical_cast const&)
+            catch (hpx::util::bad_lexical_cast const&)
```
hkaiser: The `try`/`catch` is not necessary anymore, I think.

K-ballo: It is still necessary, but it should catch `std::invalid_argument` instead.

libs/lexical_cast/include/hpx/lexical_cast/detail/converter_numeric.hpp
K-ballo: This is an entire module, and should be handled separately.

libs/lexical_cast/examples/variant_to_long_double.cpp
```diff
@@ -0,0 +1,51 @@
+// Copyright Antony Polukhin, 2013-2019.
```
hkaiser: I think we should remove this particular test as it introduces a dependency on Boost variant.

libs/lexical_cast/include/hpx/lexical_cast/bad_lexical_cast.hpp
```diff
@@ -0,0 +1,81 @@
...
+//        Alexander Nasonov, Antony Polukhin, Justin Viiret, Michael Hofmann,
+//        Cheng Yang, Matthew Bradbury, David W. Birdsall, Pavel Korzh and other Boosters
+// when:  November 2000, March 2003, June 2005, June 2006, March 2011 - 2014
+
+#ifndef HPX_LEXICAL_CAST_BAD_LEXICAL_CAST_HPP
+#define HPX_LEXICAL_CAST_BAD_LEXICAL_CAST_HPP
+
+#include <hpx/config.hpp>
+
+#include <boost/throw_exception.hpp>
```
hkaiser: Could we use our exception handling, instead?

libs/lexical_cast/include/hpx/lexical_cast/detail/converter_numeric.hpp
hkaiser: This is needed for this test only... Should we remove the test?

K-ballo: Sounds good to me

libs/lexical_cast/include/hpx/lexical_cast/bad_lexical_cast.hpp
K-ballo: No, I don't think we can. The `errors` module should be a foundational module, but it has quite a few dependencies, some of which in turn depend on this module.

We could always replace it with a naked `throw`.

libs/lexical_cast/tests/unit/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+
+  # add example executable
+  add_hpx_executable(${test}_test
+    INTERNAL_FLAGS
+    SOURCES ${sources}
+    ${${test}_FLAGS}
+    EXCLUDE_FROM_ALL
+    NOLIBS
+    DEPENDENCIES hpx_lexical_cast hpx_testing
+      hpx_filesystem
```
hkaiser: Why do we need filesystem as a dependent module? Is that for tests only?

K-ballo: Looks like it, yes. I'll move the dependency to tests.

libs/lexical_cast/include/hpx/lexical_cast/detail/converter_numeric.hpp
K-ballo: ...which test is that?

hkaiser: Uhh, my bad - it's not a test.

K-ballo:
It occurs to me that there's too much complexity in `lexical_cast` that we do not need (even after dropping wide char support, locale support, etc). All we need is a pair of `to_string/from_string` overloads.

hkaiser:
> It occurs to me that there's too much complexity in lexical_cast that we do not need (even after dropping wide char support, locale support, etc). All we need is a pair of to_string/from_string overloads.

Right, you have replace most uses of `boost::lexical_cast` with standard facilities. There is not much left that depends on lexical_cast.

hkaiser:
This is superseded by #4288.

PR #4288, opened 2019-12-31
A lightweight take on boost::lexical_cast

K-ballo:


src/util/asio_util.cpp
```diff
@@ -10,13 +10,12 @@
...
 #include <string>
 
-#include <boost/lexical_cast.hpp>
-
 #if defined(HPX_HAVE_NETWORKING)
 #include <hpx/assertion.hpp>
 #include <hpx/config/asio.hpp>
 #include <hpx/errors.hpp>
 #include <hpx/util/asio_util.hpp>
+#include <hpx/util/from_string.hpp>
```
msimberg: The include has to be outside the `HPX_HAVE_NETWORKING` guard.

PR #4289, opened 2019-12-31
Fixing problem if 128 bit atomics are not available

hkaiser:
- flyby: fixing a ton of signed/unsigned and narrowing warnings
- flyby: moving a MSVC specific change from release to master

@msimberg I think this should be added to the release, still.

### DIFF
--- a/hpx/runtime/threads/policies/queue_holder_thread.hpp
+++ b/hpx/runtime/threads/policies/queue_holder_thread.hpp
@@ -210,9 +210,10 @@ namespace hpx { namespace threads { namespace policies {
           , queue_index_(queue)
           , thread_num_(thread_num)
           , owner_mask_(owner)
-          , min_delete_count_(init.min_delete_count_)
-          , max_delete_count_(init.max_delete_count_)
-          , max_terminated_threads_(init.max_terminated_threads_)
+          , min_delete_count_(static_cast<int>(init.min_delete_count_))
+          , max_delete_count_(static_cast<int>(init.max_delete_count_))
+          , max_terminated_threads_(
+                static_cast<int>(init.max_terminated_threads_))
           , terminated_items_(max_thread_count)
         {
             rollover_counters_.data_ =
--- a/tests/unit/resource/CMakeLists.txt
+++ b/tests/unit/resource/CMakeLists.txt
@@ -16,7 +16,7 @@ set(tests
     suspend_thread_timed
 )
 
-if (${HPX_CXX_STANDARD} GREATER 11)
+if (HPX_CXX_STANDARD GREATER 11)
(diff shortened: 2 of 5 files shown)


NK-Nikunj:
@hkaiser , @biddisco HPX is now building fine.

hpx/runtime/threads/policies/queue_holder_thread.hpp
```diff
@@ -210,15 +210,17 @@ namespace hpx { namespace threads { namespace policies {
...
+          , max_delete_count_(static_cast<int>(init.max_delete_count_))
+          , max_terminated_threads_(
+                static_cast<int>(init.max_terminated_threads_))
           , terminated_items_(max_thread_count)
         {
             rollover_counters_.data_ =
                 std::make_tuple(queue_index_, round_robin_rollover);
             tq_deb.debug(debug::str<>("construct"), "D",
-                debug::dec<2>(domain_index_), "Q", debug::dec<3>(queue_index_),
+                debug::dec<2>(static_cast<int>(domain_index_)), "Q",
```
biddisco: The better way to do this would be 
`debug::dec<2,unsigned>(domain_index_)`
instead of
`debug::dec<2>(static_cast<int>(domain_index_))`
if the compiler is generating tons of warnings. (it would be nice if the default `int` could be removed and let the compiler deduce the right type)

hkaiser: @biddisco Sure, please feel free to adjust the PR the way you feel is right.

tests/unit/resource/CMakeLists.txt
```diff
@@ -16,7 +16,9 @@ set(tests
     suspend_thread_timed
 )
 
-if (${HPX_CXX_STANDARD} GREATER 11)
+# TODO: Enable on all platforms and standards when C++14 is required and test is
+# fully fixed.
+if (NOT MSVC AND "${HPX_CXX_STANDARD}" GREATER 11)
```
msimberg: This one has a few fixes now. I think we can go with the one that was in #4293 (i.e. remove this change). I'm not sure why the diff still shows master with the old comparison though...

hpx/runtime/threads/policies/queue_holder_thread.hpp
hkaiser: @biddisco I have changed the implementation of `dec` so it deduces the correct type now. In the future I'd like to map your 'logging' facilities onto the existing logging infrastructure, though.

biddisco:
I'l take a look at this today. I will fix the debug formatters.

biddisco:
Oh. Didn't see the previous comments. Ignore my previous one.

PR #4293, opened 2020-01-03
Applying two small changes fixing various MSVC/Windows problems

hkaiser:

### DIFF
--- a/tests/performance/local/future_overhead.cpp
+++ b/tests/performance/local/future_overhead.cpp
@@ -234,7 +234,7 @@ void measure_function_futures_thread_count(
 
     if (sanity_check != 0)
     {
-        int count = this_pool->get_thread_count_unknown(std::size_t(-1), false);
+        auto count = this_pool->get_thread_count_unknown(std::size_t(-1), false);
         throw std::runtime_error(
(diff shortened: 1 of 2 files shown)


tests/performance/local/future_overhead.cpp
```diff
@@ -442,9 +442,10 @@ void measure_function_futures_create_thread_hierarchical_placement(
     high_resolution_timer walltime;
     for (std::size_t t = 0; t < num_threads; ++t)
     {
-        auto const hint = hpx::threads::thread_schedule_hint(t);
+        auto const hint =
+            hpx::threads::thread_schedule_hint(static_cast<std::int16_t>(t));
         auto spawn_func = [&thread_func, sched, hint, t, count, num_threads,
-                              desc, stack_size]() {
+                              desc, stack_size, prio]() {
```
msimberg: Is this actual non-compliance in MSVC or allowed behaviour according to the standard?

hkaiser: Non-compliance in MSVC requires that even `const` variables have to be captured. OTH, clang/gcc emits a warning if a `const` variable is captured. So this PR makes it non-const to begin with.

PR #4294, opened 2020-01-05
Logging revamp

K-ballo:
Consolidate logging functionality, clean up and modernize. Add `format()`-like logger interface.


src/util/init_logging.cpp
```diff
@@ -43,239 +42,223 @@
...
                 std::size_t phase = self->get_thread_phase();
-                if (0 != phase) {
-                    std::stringstream out;
-                    out << std::hex << std::setw(sizeof(std::uint32_t))
-                        << std::setfill('0') << self->get_thread_phase();
-                    str.prepend_string(out.str());
+                if (0 != phase)
+                {
+                    std::string out =
+                        format("{:04x}", self->get_thread_phase());
```
msimberg: This should probably be output somewhere?

msimberg:
This is the only thing keeping this from going in: https://cdash.cscs.ch//viewBuildError.php?buildid=92906 (it's the CUDA build). Presumably some includes have changed to pull in things that shouldn't be pulled in device code. @K-ballo do you have any suspects right away? If not I'll try building this manually with CUDA as the compiler errors aren't very helpful.

K-ballo:
> This is the only thing keeping this from going in: https://cdash.cscs.ch//viewBuildError.php?buildid=92906 (it's the CUDA build). Presumably some includes have changed to pull in things that shouldn't be pulled in device code. @K-ballo do you have any suspects right away? If not I'll try building this manually with CUDA as the compiler errors aren't very helpful.

Those errors are being triggered from https://github.com/STEllAR-GROUP/hpx/pull/4294/files#diff-e5da93dcb588211b367bdf85a1dc02cdR402 . The code in question creates a message by moving an `std::stringstream`, which is new behavior introduced in this PR.

If you get to build this with CUDA, check whether the following inefficient variation is accepted:
```
message formatted;
formatted << out.str();
```

msimberg:
Sorry @K-ballo, this slipped off my todo list. I'll look at it this week.

msimberg:
@K-ballo that did not change things. We could consider disabling or limiting logging (basic printf?) in CUDA code since whatever is there right now can't be used in CUDA code even if we can get it to compile. Do we only use the logging library through the logging macros?

K-ballo:
> @K-ballo that did not change things. We could consider disabling or limiting logging (basic printf?) in CUDA code since whatever is there right now can't be used in CUDA code even if we can get it to compile.

This was working before (for whatever definition of "working" that was, compiling at least) and isn't working with these changes, with no clear indication of what may be triggering these issues. This should be properly explored by whoever is in charge of CUDA builds, to determine what the rejected patterns are.

msimberg:
> > @K-ballo that did not change things. We could consider disabling or limiting logging (basic printf?) in CUDA code since whatever is there right now can't be used in CUDA code even if we can get it to compile.
> 
> This was working before (for whatever definition of "working" that was, compiling at least) and isn't working with these changes, with no clear indication of what may be triggering these issues. This should be properly explored by whoever is in charge of CUDA builds, to determine what the rejected patterns are.

Indeed. I'll look closer into this.

src/util/init_logging.cpp
msimberg: I have a vague memory of you already fixing this. Did this maybe get reverted at some point?

K-ballo: I remember that too.. maybe it was a different instance? or perhaps something get lost during a rebase

msimberg:
@K-ballo I've rebased this on master and added a commit. It compiles now with CUDA but doesn't have to stay this way if you have suggestions for better ways to do it. I've essentially just ifdefed out the minimal problematic parts. You were spot on with your first suspect. There were just a few more places where strings were moved.

libs/logging/include/hpx/logging/format/named_write.hpp
```diff
@@ -399,8 +399,10 @@ This will just configure "file" twice, ending up with writing only to "two.txt"
             std::stringstream out;
             m_format(out, msg);
 
+#if !defined(HPX_COMPUTE_HOST_CODE)
             message formatted(std::move(out));
             m_destination(formatted);
+#endif
```
K-ballo: I'd expect there would be an `#else` here in some form. This effectively disables formatting in logging.

msimberg: Well, yes and no. Unless we provide a way to redirect messages to `printf` in device code the logging capabilities can't be used in device code anyway. This just makes it compile but does not work exactly as you say. Hence my earlier suggestion to just not build logging at all when compiling device code.

hkaiser: FWIW, I wouldn't expect for logging to be invoked in device code. So I think the current solution is good enough. We could add a (non-HPX)  `assert()` however to make sure.

msimberg: Indeed. `HPX_ASSERT` uses an `assert` which works in device code btw, so it's safe to use `HPX_ASSERT` directly (there's currently no error message, though).

hkaiser: > HPX_ASSERT uses an assert

Hmm, does it? I thought we used our own (https://github.com/STEllAR-GROUP/hpx/blob/master/libs/assertion/include/hpx/assertion.hpp#L59-L65).

msimberg: Note the trailing underscore on the definition you linked ;) `HPX_ASSERT` itself is defined differently if compiling device code: https://github.com/STEllAR-GROUP/hpx/blob/aa59c1bc6e97410a5e7db009feec5693fe0cf4dc/libs/assertion/include/hpx/assertion.hpp#L68-L71.

hkaiser: Yes, you're certainly right!

hkaiser:
@K-ballo I rebased this onto current master. I think we can merge it if the tests look ok.

msimberg:
`gcc-oldest` is now only complaining about name mangling changing from 14 to 17. Let's fix that in a separate PR. If `gcc-cuda` still comes back clean (except for `transform_compute`...) we can go ahead with this.

msimberg:
I built this manually and things look ok. I'll merge this now.

PR #4296, opened 2020-01-07
Fixing collectives locking problem

folshost:
Fixes #

## Proposed Changes

  - Adds call to ignore_while_checking in all_to_all and all_reduce

I forgot to test if it fixed the problem in Phylanx which provided the motivation for the fix, before I committed it last time. We used the wrong ignore function.


folshost:
Fixes problem #4255 was supposed to fix

msimberg:
Thanks @folshost!

PR #4298, opened 2020-01-08
Remove extra call to annotate function that messed up traces

### LINKED ISSUE #3560: Apex integration with HPX not working properly
## Expected Behavior
Nice plots of tasks

## Actual Behavior
All the tasks have the same `parallel::execution::post` name.
I am using the RAII style apex wrapper with `profiler::TaskProfiler tp(taskname, "Trailing Matrix");` that wraps an `hpx::util::annotate_function apex_profiler;` and is instantiated at the start of the task.
Note that is differs from the lambda style `annotated_function` by setting the name of the task _inside_ the task, whereas the lambda version does it on creation on the task.

I attempted to attach a screen shot of vampir, but it doesn't seem to work.

hoping @khuck knows what's wrong

biddisco:
Fixes #3560

### DIFF
--- a/hpx/util/annotated_function.hpp
+++ b/hpx/util/annotated_function.hpp
@@ -137,7 +137,6 @@ namespace hpx { namespace util
                 typename util::decay_unwrap<F>::type, Ts...>::type
             operator()(Ts && ... ts)
             {
-                annotate_function func(name_);
                 return util::invoke(f_, std::forward<Ts>(ts)...);
             }
 
--- a/hpx/lcos/local/futures_factory.hpp
+++ b/hpx/lcos/local/futures_factory.hpp
@@ -131,7 +131,7 @@ namespace hpx { namespace lcos { namespace local {
 
                 threads::register_thread_nullary(pool,
                     util::deferred_call(&base_type::run_impl, std::move(this_)),
-                    util::thread_description(f_, "task_object::apply"),
+                    util::thread_description(f_, annotation),
                     threads::pending, false, priority, schedulehint, stacksize,
                     ec);
                 return threads::invalid_thread_id;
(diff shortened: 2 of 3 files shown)


hpx/util/annotated_function.hpp
```diff
@@ -137,7 +137,6 @@ namespace hpx { namespace util
                 typename util::decay_unwrap<F>::type, Ts...>::type
             operator()(Ts && ... ts)
             {
-                annotate_function func(name_);
```
hkaiser: This call is the whole point of having the `annotated_function` type. Removing this call renders it useless.

khuck: @hkaiser What if we put `#if !defined(HPX_HAVE_APEX)` / `#endif` around it?  There is other code in HPX that will do the annotation correctly when the function happens to be of the type `annotated_function`.

biddisco: @hkaiser I'm not sure I understand. The `annotated_function` wrapper sets the annotation name in the creation of the task and does the 'right' thing. The call to annotate_function inside the `annotated_function` is doing what?

hkaiser: @biddisco right, I misunderstood - I think all is well now.

biddisco:
I guess once the release is out, we should consider a point release in a month or so with some fixes like this in. A lot of people ask me about apex, and having a functional infrastructure would be a good thing.

msimberg:
Thanks @biddisco for looking into this! Let's get this in the release as well. The annotation changes are very small. The ICE (https://cdash.cscs.ch//viewBuildError.php?buildid=91235) is a bit suspicious though as that test was actually changed... I've restarted that build so let's see if it fails again.

msimberg:
This is definitely unrelated: https://cdash.cscs.ch//viewBuildError.php?buildid=91350.

msimberg:
GCC 5 now builds cleanly, thanks @biddisco! https://cdash.cscs.ch/buildSummary.php?buildid=91377

Merging this now.

PR #4299, opened 2020-01-09
Small CMake fixes

msimberg:
Fixes CMake configuration with APEX and CUDA enabled at the same time. Also adds more guards for targets that may be created multiple times when HPX is found multiple times.


msimberg:
I think this will also have to go in 1.4.0.

msimberg:
Merging this now for the release. Please comment still after merging if you notice something odd in the diff.

PR #4301, opened 2020-01-10
Clean up thread executors

### LINKED ISSUE #2997: Executor consolidation and clarification/documentation
Documentation for the following is incomplete/scarce and some executors may no longer be required.
As part of the ongoing cleaning up caused by the introduction of the resource_partitioner, and with a potential 1.1 release planned, the following executors need to be documented and/or deprecated.

The following exist in the runtime/threads or parallel namespaces
- pool_executor
creates tasks that are scheduled on the named thread pool. The thread pool is created at startup by the resource manager and may be customized by the user.
- thread_pool_executor
- thread_pool_attached_executor
- thread_pool_os_executor
- this_thread_executor
- thread_executor
- current_executor
- default_executor
creates tasks by default on the same pool as the parent task. I believe that this executor should be a subclass of pool_executor and should have an additional default-policy constructor option that allows the user to select between
    - running the task on the same pool as the previous/parent task
    - running the task on the "default pool" 
- distribution_policy_executor
creates executors using any of the existing distribution policies (like hpx::components::colocating_distribution_policy et.al.).
- rebind_executor
- sequenced_executor
creates groups of sequential execution agents which execute in the calling thread. The sequential order is given by the lexicographical order of indices in the index space.
- parallel_executor
creates groups of parallel execution agents which execute in threads implicitly created by the executor. This executor uses a given launch policy.
- service_executor
creates groups of parallel execution agents which execute in one of the kernel threads associated with a given pool category (I/O, parcel, or timer pool, or on the main thread of the application).
- timed_executor

Additional executors exist in the compute:: namespace 
- block_executor
Spawns tasks on a vector of target resources (such as numa domains). Needs work to be integrated into the numa domain definitions provided by the resource_manager. 
- concurrent_executor
- default_executor

Additional executors exist for specific scheduler types, these create executors on top of the corresponding HPX schedulers.
- local_priority_queue_executor, 
- local_queue_executor 
- static_priority_queue_executor 
And with an os_ prefix
- local_queue_os_executor
- static_queue_os_executor
- local_priority_queue_os_executor
- static_priority_queue_os_executor
- local_queue_os_executor

Please add a short description to this issue for each of the executor names describing what they do and if they can be removed, please state so. if duplicates or errors exist, please say so. Once all executors have been described or obsoleted, I will produce some documentation and start cleaning up where applicable.

msimberg:
Fixes #2997? Cleans up thread executors. Some are rewritten, some deprecated, and some are simply typedefed to other executors. The changes are not 100% backwards compatible (i.e. some APIs that existed using the old interface have been completely removed). I think this should be okay, but let's gather some feedback until the next release if some of them should be reintroduced.

Summary:

- Moved all executors to `hpx/parallel/executors` and left only forwarding headers and typedefs for the `hpx::threads::executors` namespace in `hpx/runtime/threads/executors`.
- `service_executor` rewritten using the the newer executor interface. The old `service_executor` claimed to block until all outstanding work was done, but this wasn't the case. I've made it non-blocking now since if someone relied on it they were very lucky, or else noone actually relied on that functionality. I'd add it back only if someone requests the functionality.
- `current_executor` is typedefed to `thread_pool_executor` (after some constructors were added to it). The nullary constructor just grabs the current thread pool and submits tasks to that.
- `pool_executor` inherits `thread_pool_executor`.
- `default_executor` is typedefed to `parallel_executor`.
- `embedded_thread_pool_executor` are deprecated behind a CMake option.
- `this_thread_executor` deprecated.
- `thread_pool_attached_executor` deprecated, with a replacement based on `thread_pool_executor` (`compute::host::block_executor` is now based on this replacement).
- `thread_pool_os_executors` deprecated.


msimberg:
I forgot the most important part: more `future_overhead` benchmarks. I only compared the `for_loop` version for the three main parallel executors we have now: `parallel_executor`, `parallel_executor_aggregated`, and `thread_pool_executor` (this one with and without stackless threads, although the stackless one might be buggy because of spinlocks). After 200 tasks (it says futures on the y-axis, but it's really plain tasks) the plots look all the same. Here the stackless threads actually seem to make a difference. `thread_pool_executor` is slightly slower for small numbers of tasks because it actually creates more tasks than `parallel_executor` but after that it's similar or slightly faster.

thread_pool_executor
thread_pool_executor_200

msimberg:
Reviews welcome on this one.

msimberg:
Even codespell should be happy on this one now.

msimberg:
@biddisco would you have time to review this?

hpx/runtime/threads/executors/default_executor.hpp
```diff
@@ -1,115 +1,18 @@
...
-            // violate bounds on the executor's queue size.
-            inline void add_after(
-                util::steady_clock::duration const& rel_time,
-                closure_type&& f, util::thread_description const& description,
-                threads::thread_stacksize stacksize, error_code& ec) override
-            {
-                return add_at(util::steady_clock::now() + rel_time,
-                    std::move(f), description, stacksize, ec);
-            }
+#if !defined(HPX_RUNTIME_THREADS_EXECUTORS_DEFAULT_EXECUTOR)
```
aurianer: Is the removal of the `_HPP` intentional ?

hpx/runtime/threads/executors/limiting_executor.hpp
```diff
@@ -1,207 +1,16 @@
...
-    struct is_two_way_executor<threads::executors::limiting_executor<Executor>>
-      : std::true_type
-    {
-    };
-}}}    // namespace hpx::parallel::execution
-
-#include <hpx/config/warnings_suffix.hpp>
-
-#endif /*HPX_RUNTIME_THREADS_LIMITING_EXECUTOR_HPP*/
+#endif /*HPX_RUNTIME_THREADS_GUIDED_POOL_EXECUTOR*/
```
aurianer: `THREADS_LIMITING_EXECUTOR` instead of `THREADS_GUIDED_POOL_EXECUTOR`?

hpx/runtime/threads/executors/limiting_executor.hpp
```diff
@@ -1,207 +1,16 @@
-//  Copyright (c) 2017-2018 John Biddiscombe
+//  Copyright (c) 2017-2019 John Biddiscombe
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-#ifndef HPX_RUNTIME_THREADS_LIMITING_EXECUTOR_HPP
-#define HPX_RUNTIME_THREADS_LIMITING_EXECUTOR_HPP
+#ifndef HPX_RUNTIME_THREADS_LIMITING_EXECUTOR
```
aurianer: Missing `_HPP`?

msimberg: Good catch on all of these, thanks! Most likely caused by my sloppy find-and-replacing.

msimberg:
This goes for one more CI round because of the rebase.

PR #4304, opened 2020-01-13
Improve pack support utilities

K-ballo:


msimberg:
I guess the builtins may provide a small compilation time decrease?

K-ballo:
> I guess the builtins may provide a small compilation time decrease?

They do, but for our use cases (N<20) I do not expect it to be noticeable. It is still a few less template instantiations.

PR #4306, opened 2020-01-14
Fix schedule hint not being taken from executor

biddisco:
A user on IRC posted a request to bind tasks to threads, but when using the default executor, the schedulehint was not being taken from the executor, instead the default was being used.

### DIFF
--- a/hpx/lcos/local/futures_factory.hpp
+++ b/hpx/lcos/local/futures_factory.hpp
@@ -264,7 +264,7 @@ namespace hpx { namespace lcos { namespace local {
                     parallel::execution::post(*exec_,
                         util::deferred_call(
                             &base_type::run_impl, std::move(this_)),
-                        schedulehint, annotation);
+                        exec_->get_schedulehint(), annotation);
                     return threads::invalid_thread_id;
                 }
 
--- a/libs/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
+++ b/libs/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
@@ -70,6 +70,8 @@ static_assert(false,
 #define SHARED_PRIORITY_SCHEDULER_LINUX
 #endif
 
+// #define SHARED_PRIORITY_SCHEDULER_DEBUG_NUMA
+
 namespace hpx {
     using print_onoff =
         hpx::debug::enable_print<SHARED_PRIORITY_SCHEDULER_DEBUG>;
@@ -200,7 +202,7 @@ namespace hpx { namespace threads { namespace policies {
                     steal_hp_first_ = mode & policies::steal_high_priority_first;
                     core_stealing_ = mode & policies::enable_stealing;
                     numa_stealing_ = mode & policies::enable_stealing_numa;
-                    spq_arr.debug(debug::str<>("scheduler_mode")
+                    spq_deb.debug(debug::str<>("scheduler_mode")
                         , round_robin_ ? "round_robin" : "thread parent"
                         , ','
                         , steal_hp_first_ ? "steal_hp_first" : "steal after local"
@@ -223,7 +225,11 @@ namespace hpx { namespace threads { namespace policies {
 
                 // ------------------------------------------------------------
                 // access thread local storage to determine correct thread
-                // and pool identification
+                // and pool identification. This is used internally by the scheduler
+                // to compute correct queue indexes and offsets relative to a numa
+                // node. It should not be used without care as the thread numbering
+                // internal to the scheduler is not a simple linear indexing
+                // returns -1 to indicate an invalid thread/value/state
                 inline std::size_t local_thread_number()
                 {
                     using namespace hpx::threads::detail;
@@ -232,7 +238,7 @@ namespace hpx { namespace threads { namespace policies {
                     // if the thread belongs to this pool return local Id
                     if (pool_index_ == thread_pool_num)
                         return get_local_thread_num_tss();
-                    return -1;
+                    return std::size_t(-1);
                 }
 
                 // ------------------------------------------------------------
@@ -432,23 +438,35 @@ namespace hpx { namespace threads { namespace policies {
                     if (local_num != thread_num)
                     {
                         run_now = false;
-                        spq_deb.debug(debug::str<>("create_thread"), "pool",
-                            parent_pool_->get_pool_name(), "hint", msg, "dest",
-                            "D", debug::dec<2>(domain_num), "Q",
-                            debug::dec<3>(q_index), "this", "D",
-                            debug::dec<2>(d_lookup_[local_num]), "Q",
-                            debug::dec<3>(local_num), "run_now OVERRIDE ",
-                            run_now, debug::threadinfo<thread_init_data>(data));
+                        // clang-format off
+                        spq_deb.debug(debug::str<>("create_thread")
+                            , "pool", parent_pool_->get_pool_name()
+                            , "hint", msg
+                            , "dest"
+                            , "D", debug::dec<2>(domain_num)
+                            , "Q", debug::dec<3>(q_index)
+                            , "this"
+                            , "D", debug::dec<2>(d_lookup_[local_num])
+                            , "Q", debug::dec<3>(local_num)
+                            , "run_now OVERRIDE ", run_now
+                            , debug::threadinfo<thread_init_data>(data));
+                        // clang-format on
                     }
                     else
                     {
-                        spq_deb.debug(debug::str<>("create_thread"), "pool",
-                            parent_pool_->get_pool_name(), "hint", msg, "dest",
-                            "D", debug::dec<2>(domain_num), "Q",
-                            debug::dec<3>(q_index), "this", "D",
-                            debug::dec<2>(d_lookup_[local_num]), "Q",
-                            debug::dec<3>(local_num), "run_now", run_now,
-                            debug::threadinfo<thread_init_data>(data));
+                        // clang-format off
+                        spq_deb.debug(debug::str<>("create_thread")
+                            , "pool", parent_pool_->get_pool_name()
+                            , "hint", msg
+                            , "dest"
+                            , "D", debug::dec<2>(domain_num)
+                            , "Q", debug::dec<3>(q_index)
+                            , "this"
+                            , "D", debug::dec<2>(d_lookup_[local_num])
+                            , "Q", debug::dec<3>(local_num)
+                            , "run_now", run_now
+                            , debug::threadinfo<thread_init_data>(data));
+                        // clang-format on
                     }
                     numa_holder_[domain_num]
                         .thread_queue(static_cast<std::size_t>(q_index))
@@ -801,6 +819,9 @@ namespace hpx { namespace threads { namespace policies {
                         // @TODO. We should check that the thread num is valid
                         // Create thread on requested worker thread
                         debug::set(msg, "HINT_THREAD");
+                        spq_deb.debug(debug::str<>("schedule_thread"),
+                            "received HINT_THREAD",
+                            debug::dec<3>(schedulehint.hint));
                         thread_num = select_active_pu(
                             l, schedulehint.hint, allow_fallback);
                         domain_num = d_lookup_[thread_num];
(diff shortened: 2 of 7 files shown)


msimberg:
Thanks @biddisco. Note that https://github.com/STEllAR-GROUP/hpx/pull/4301 changes the `default_executor` to be the `parallel_executor` and the path from specifying a hint to having it in the scheduler is a bit more direct (and hopefully less prone to bugs like this).

tests/unit/resource/scheduler_binding_check.cpp
```diff
@@ -0,0 +1,86 @@
+#include <hpx/hpx_init.hpp>
```
hkaiser: This file misses License, copyright, and the SPDX tag.

hpx/lcos/local/futures_factory.hpp
```diff
@@ -265,7 +265,7 @@ namespace hpx { namespace lcos { namespace local {
                     parallel::execution::post(*exec_,
                         util::deferred_call(
                             &base_type::run_impl, std::move(this_)),
-                        schedulehint, annotation);
+                        exec_->get_schedulehint(), annotation);
```
hkaiser: Why does the `schedulehint` provided by the executor takes precedence over the one passed in from the caller? Do we have this behavior everywhere?

biddisco: The one from the caller is always just a default param. The whole thing is a mess.

hkaiser: Why making it an even bigger mess, then? Isn't there a better way?

biddisco: I don't believe this makes it a bigger mess. This fixes a bug where scheduling hints were not being used when set via the executor. In my opinion the mess is caused by a very complex system of overloads on executors for different methods of invocations and default parameters used in many places which hides invocations that might be clumsy or wasteful. Ideally the default params should be removed and the functions invoked with the right parameters - or an invocation signature which does not use the parameters should exist - unfortunately, invocations with hints, without hints, with annotation, without them, with policies and without them etc make the whole thing messy. I'd like to see a more consistent way of passing policies, parameters, scheduling information/hints and any other required parsm to be designed. I hope that as we move over to the new executor proposal we can clean up some of the existing code - but that's too much for this simple fix and deserves a bigger PR at some later date.

hkaiser: Could you please be a bit more specific about what functions should not have default arguments for their parameters?

biddisco: I cannot provide such a list. I am simply trying to fix bugs that prevent me from optimizing and debugging my code. Many of the default params for `schedulehint` were introduced by me, and I take full blame for them - they were introduced because there is a lot of code to understand at the executor dispatch level that is not commented or easily understandable by anyone other than the person who wrote it - it was easier to get the functionality I need by adding these defaults than to edit the many places that would need to be changed.

I hope that the executor cleanups and new proposals make this simpler to maintain.

tests/unit/resource/scheduler_binding_check.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include <hpx/async.hpp>
+#include <hpx/runtime/threads/executors/default_executor.hpp>
+#include <hpx/testing.hpp>
+#include <hpx/parallel/executors/thread_pool_executors.hpp>
```
msimberg: @biddisco I think you only need to rebase this on latest master to have the build errors go away (`hpx/parallel/executors/thread_pool_executors.hpp` not found). That said, you should also change the include to `hpx/execution/executors/thread_pool_executors.hpp`.

biddisco:
Added copyright etc and rebased onto master.

biddisco:
The new test introduced by this PR has been failing on machines like daint. Sorry for that. what I didn't realize is that the `get_worker_thread_num` call actually delves into the core and returns the `thread_num_tss` thread local storage ID. The shared_priority scheduler actually uses internal a differnt numbering scheme when the threads go onto the second numa domain.

What I'd like to do is replace the calls to `thread_num_tss` with calls that go into the current scheduler and return the thread id from there. This allows schedulers to use any mapping of thread Ids to cores.

Does anyone object to that?

hpx/lcos/local/futures_factory.hpp
msimberg: I think long term futures factory should not have the option to take an executor (this functionality seems to be there mostly for the old thread executors). If one wants to create a task with an executor there is one way to do so: through `post` or `async_execute`. The handling of the hints will be more direct after #4301 (I'm also working on unifying the various launch parameters into a struct).

msimberg:
I think we should go ahead with this. This does seem to provide an intermediate fix and the change is minimal. It also adds a very useful test to check that hints are taken into account.

msimberg:
@biddisco you'll unfortunately need to rebase again to have this in...

> The new test introduced by this PR has been failing on machines like daint. Sorry for that. what I didn't realize is that the `get_worker_thread_num` call actually delves into the core and returns the `thread_num_tss` thread local storage ID. The shared_priority scheduler actually uses internally a different numbering scheme when the threads go onto the second numa domain.
> 
> What I'd like to do is replace the calls to `thread_num_tss` with calls that go into the current scheduler and return the thread id from there. This allows schedulers to use any mapping of thread Ids to cores.
> 
> Does anyone object to that?

Why don't we extend the thread_num_tss to return a single struct with local thread id, global thread id, and thread pool id? Internally pools would use the local id, and if one wants to create a schedulehint for the current thread we use the local id (we can create a new executor for this as well to avoid mistakes). The struct already exists for the local thread id and pool id, but they should just be combined into one struct. The `get_worker_thread_num` function should just grab it from the struct.

@biddisco did you check if the test really fails on daint or if we have other problems there (there have been many failures lately not related to PRs... I'm restarting the builders now after the maintenance)? CircleCI is happy with the test and it doesn't look like you've added any special handling for that.

biddisco:
yes @msimberg the test really does fail. I am looking into a fix as we discussed the other day on the call. Didn't spend any time on it since then.

I would be in favour of the `thread_num tss` storing a struct of global thread id, local pool thread id and pool number. This would make things consistent throughout and although we might need to change quite a few places, I think it's the right way to go. I will add that to this PR for review once I'm done with the fix I'm working on.

hkaiser:
>  would be in favour of the thread_num tss storing a struct of global thread id, local pool thread id and pool number. This would make things consistent throughout and although we might need to change quite a few places, I think it's the right way to go. I will add that to this PR for review once I'm done with the fix I'm working on.

I'm not sure that this would be a good move (but I'd go with it if you are convinced that it is...). My rationale would be that having all this information stored with the kernel thread will require to modify it whenever pools are dynamically added, removed, activated, or deactivated. Currently the thread number is a constant value that does not require protected from races. Leaving the additional mappings (from and to logocal thread numbers etc.) in the pools/schedulers would allow us to avoid having any additional protection from concurrent access.

msimberg:
I'd say the local thread ids and pool ids should be as static as the global id, i.e. not change (do the indices have to be contiguous?). Also, it's `thread_local` so I don't see how this is going to be accessed from multiple threads. Are there use cases I'm not thinking about?

msimberg:
If this is going in 1.4.1 it should only contain the small change to get the schedule hint from the executor (I'll make an rc today). Any other changes will have to wait. I'd prefer to have this properly sorted out for 1.5.0 instead though.

biddisco:
After more debugging I finally discovered that my initial diagnosis was incorrect and in fact I do not need to route the get_worker thread number through the scheduler (though I actually think it's a good idea for future use). The problem was that there was a bug in one of my stealing functions that allowed a thread to be stolen when it shouldn't have.

The final fix, is in fact only a few tweaks to the scheduler itself and other thread_num_tss functions can be left alone.

Would it be too late to get this into the patch release if I force push the correct fix?

msimberg:
> Would it be too late to get this into the patch release if I force push the correct fix?

If you have the fix ready please push and then we'll see how small the change is.

hpx/runtime/threads/policies/shared_priority_queue_scheduler.hpp
```diff
@@ -69,6 +69,8 @@ static_assert(false,
 #define SHARED_PRIORITY_SCHEDULER_LINUX
 #endif
 
+// #define SHARED_PRIORITY_SCHEDULER_DEBUG_NUMA
```
msimberg: Remove?

biddisco: it's like a flag that can be swtiched on/off, if you remove it, then anyone debugging might not know that it is useful...
I like to leave these in because this time next week, something else will break and I'll be back debugging it - it gets hard to maintain all the branches when you have to keep removing and putting back small snippets.

biddisco:
I cleaned up the commits for this branch to cover only the fixes necessary to solve the problem + a few extra debug things that are useful in the scheduler.

The stuff that was leftover is a small extension to the base_scheduler to return the thread index and thread numa node index - these I used when I was trying to fix it first time, but turned out not to be needed. However they are useful extensions that I will use elsewhere and I would like to add this last commit to the PR as it belongs in this PR rather than anywhere else. 

Considering that this PR might get into a point release I'm happy to wait with the extr commit and add it to another PR at a later date, but if I can squeeze it in, please say and I'll push it.

tests/unit/resource/CMakeLists.txt
```diff
@@ -22,6 +23,7 @@ set(tests
 
 set(cross_pool_injection_PARAMETERS THREADS_PER_LOCALITY -1)
 set(scheduler_priority_check_PARAMETERS THREADS_PER_LOCALITY -1)
+set(scheduler_binding_check_PARAMETERS THREADS_PER_LOCALITY -1)
```
msimberg: This needs to be enabled only if the shared priority queue scheduler is enabled.

msimberg:
I'll cherry pick the fixes themselves to the release but the new test still needs to be appropriately disabled.

tests/unit/resource/scheduler_binding_check.cpp
```diff
@@ -40,16 +48,21 @@ struct dec_counter
...
-        HPX_TEST_EQ(thread_num, threadnum);
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        std::size_t thread_actual = hpx::get_worker_thread_num();
+        hpx::deb_schbin.debug(hpx::debug::str<>("Running on thread"),
+            thread_actual, "Expected", thread_expected);
+        if (thread_actual != thread_expected)
+        {
+            hpx::deb_schbin.error(hpx::debug::str<>("actual!=expected"), "Got",
+                thread_actual, "Expected", thread_expected);
+        }
```
msimberg: Why do this when `HPX_TEST_EQ` prints essentially the same thing when it fails?

msimberg:
@biddisco would you mind addressing the last comments on this one (and rebasing) so that we can get this in?

hkaiser:
What's the status of this? @biddisco, @msimberg?

msimberg:
AFAIK it's just missing a rebase and that one test needs to be conditionally enabled (one of our pycicle builders doesn't have the shared priority queue scheduler enabled). We already have the main changes in master (because they went into the release).

msimberg:
I've just resolved the conflicts and addressed some of the comments. I did not change anything related to how hints are passed as the thread executors are deprecated anyway.

libs/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
```diff
@@ -427,27 +433,39 @@ namespace hpx { namespace threads { namespace policies {
...
-                            run_now, debug::threadinfo<thread_init_data>(data));
+                        // clang-format off
+                        spq_deb.debug(debug::str<>("create_thread")
+                            , "pool", parent_pool_->get_pool_name()
+                            , "hint", msg
+                            , "dest"
+                            , "D", debug::dec<2>(domain_num)
+                            , "Q", debug::dec<3>(q_index)
+                            , "this"
+                            , "D", debug::dec<2>(d_lookup_[local_num])
```
hkaiser, line 448: This will horribly fail if `local_num == -1` (that's a valid possibility, as it is initialized by `local_thread_number()` which may return `-1`). 

I'm aware that this is not a problem introduced by this PR, I just thought to mention it anyways.

msimberg:
I think #4516 will take care of the failures relating to `default_executor`.

msimberg:
I've rebased this again. Most of the build problems seem to be gone after #4516, but I did not build everything. Let's see what CI says.

msimberg:
This is looking pretty clean on CI now. I think we can leave the thread numbering issues to #4311 or another PR.

PR #4311, opened 2020-01-17
Dataflow annotations for APEX

biddisco:
Fixes #4257 (again) since the previous fixes were not complete.

### DIFF
--- a/hpx/lcos/local/packaged_continuation.hpp
+++ b/hpx/lcos/local/packaged_continuation.hpp
@@ -45,7 +45,8 @@ namespace hpx { namespace traits {
         ///////////////////////////////////////////////////////////////////////////
         // traits specialization to get annotation from dataflow_finalization
         template <typename Continuation>
-        struct get_function_annotation<lcos::detail::continuation_finalization<Continuation>>
+        struct get_function_annotation<lcos::detail::continuation_finalization
+            <Continuation>>
         {
             using function_type = typename Continuation::function_type;
             static char const* call(
@@ -229,7 +230,8 @@ namespace hpx { namespace lcos { namespace detail
         typedef continuation<Future, F, ContResult> continuation_type;
 
         friend struct continuation_finalization<continuation_type>;
-        friend struct traits::get_function_annotation<continuation_finalization<continuation_type>>;
+        friend struct traits::get_function_annotation<continuation_finalization
+            <continuation_type>>;
 
     protected:
         threads::thread_id_type get_id() const
--- a/hpx/lcos/local/packaged_continuation.hpp
+++ b/hpx/lcos/local/packaged_continuation.hpp
@@ -64,7 +64,10 @@ namespace hpx { namespace lcos { namespace detail
     template <typename Continuation>
     struct continuation_finalization {
         //
-        continuation_finalization(Continuation *c) : this_(c) {}
+        explicit continuation_finalization(Continuation* c)
+          : this_(c)
+        {
+        }
         //
         template <typename Future>
         void operator()(Future&& f) const
--- a/hpx/lcos/local/packaged_continuation.hpp
+++ b/hpx/lcos/local/packaged_continuation.hpp
@@ -20,7 +20,6 @@
 #include <hpx/traits/future_traits.hpp>
 #include <hpx/threading_base/annotated_function.hpp>
 #include <hpx/threading_base/thread_description.hpp>
-#include <hpx/functional/traits/get_function_annotation.hpp>
 
 #include <hpx/execution/executors/execution.hpp>
 #include <hpx/execution/detail/post_policy_dispatch.hpp>
@@ -33,53 +32,8 @@
 #include <utility>
 
 ///////////////////////////////////////////////////////////////////////////////
-// forward declare the type we will get function annotations from
 namespace hpx { namespace lcos { namespace detail
 {
-    template <typename Continuation>
-    struct continuation_finalization;
-}}}
-
-namespace hpx { namespace traits {
-    #if defined(HPX_HAVE_THREAD_DESCRIPTION)
-        ///////////////////////////////////////////////////////////////////////////
-        // traits specialization to get annotation from dataflow_finalization
-        template <typename Continuation>
-        struct get_function_annotation<lcos::detail::continuation_finalization
-            <Continuation>>
-        {
-            using function_type = typename Continuation::function_type;
-            static char const* call(
-                lcos::detail::continuation_finalization<Continuation> const& f) noexcept
-            {
-                char const* annotation = hpx::traits::get_function_annotation<
-                    typename hpx::util::decay<function_type>::type>::call(f.this_->f_);
-                return annotation;
-            }
-        };
-    #endif
-}}
-
-namespace hpx { namespace lcos { namespace detail
-{
-    template <typename Continuation>
-    struct continuation_finalization {
-        //
-        explicit continuation_finalization(Continuation* c)
-          : this_(c)
-        {
-        }
-        //
-        template <typename Future>
-        void operator()(Future&& f) const
-        {
-            return this_->async_exec_impl(std::forward<Future>(f));
-        }
-
-        // keep the dataflow frame alive with this pointer reference
-        hpx::intrusive_ptr<Continuation> this_;
-    };
-
     template <typename Source, typename Destination>
     HPX_FORCEINLINE void transfer_result_impl(std::false_type,
         Source&& src, Destination& dest)
@@ -226,12 +180,6 @@ namespace hpx { namespace lcos { namespace detail
 
         typedef typename base_type::mutex_type mutex_type;
         typedef typename base_type::result_type result_type;
-        typedef typename std::decay<F>::type function_type;
-        typedef continuation<Future, F, ContResult> continuation_type;
-
-        friend struct continuation_finalization<continuation_type>;
-        friend struct traits::get_function_annotation<continuation_finalization
-            <continuation_type>>;
 
     protected:
         threads::thread_id_type get_id() const
@@ -429,9 +377,8 @@ namespace hpx { namespace lcos { namespace detail
             }
 
             hpx::intrusive_ptr<continuation> this_(this);
-            char const* annotation = hpx::traits::get_function_annotation<
-                function_type>::call(f_);
-            hpx::util::thread_description desc(f_, annotation);
+            hpx::util::thread_description desc(f_,
+                "hpx::parallel::execution::parallel_executor::post");
 
             parallel::execution::detail::post_policy_dispatch<
                     hpx::launch::async_policy
@@ -476,9 +423,8 @@ namespace hpx { namespace lcos { namespace detail
             }
 
             hpx::intrusive_ptr<continuation> this_(this);
-            char const* annotation = hpx::traits::get_function_annotation<
-                function_type>::call(f_);
-            hpx::util::thread_description desc(f_, annotation);
+            hpx::util::thread_description desc(f_,
+                "hpx::parallel::execution::parallel_executor::post");
 
             parallel::execution::detail::post_policy_dispatch<
                     hpx::launch::async_policy
@@ -522,12 +468,13 @@ namespace hpx { namespace lcos { namespace detail
                 started_ = true;
             }
 
-            continuation_finalization<continuation_type> this_f_(this);
-
-            parallel::execution::post(
-                std::forward<Executor>(exec),
-                std::move(this_f_),
-                std::move(f));
+            hpx::intrusive_ptr<continuation> this_(this);
+            parallel::execution::post(std::forward<Executor>(exec),
+                [this_ = std::move(this_),
+                    f = std::move(f)
+                ]() mutable -> void {
+                    this_->async_exec_impl(std::move(f));
+                });
 
             if (&ec != &throws)
                 ec = make_success_code();
--- a/hpx/lcos/local/packaged_continuation.hpp
+++ b/hpx/lcos/local/packaged_continuation.hpp
@@ -20,6 +20,7 @@
 #include <hpx/traits/future_traits.hpp>
 #include <hpx/threading_base/annotated_function.hpp>
 #include <hpx/threading_base/thread_description.hpp>
+#include <hpx/functional/traits/get_function_annotation.hpp>
 
 #include <hpx/execution/executors/execution.hpp>
 #include <hpx/execution/detail/post_policy_dispatch.hpp>
@@ -32,8 +33,49 @@
 #include <utility>
 
 ///////////////////////////////////////////////////////////////////////////////
+// forward declare the type we will get function annotations from
 namespace hpx { namespace lcos { namespace detail
 {
+    template <typename Continuation>
+    struct continuation_finalization;
+}}}
+
+namespace hpx { namespace traits {
+    #if defined(HPX_HAVE_THREAD_DESCRIPTION)
+        ///////////////////////////////////////////////////////////////////////////
+        // traits specialization to get annotation from dataflow_finalization
+        template <typename Continuation>
+        struct get_function_annotation<lcos::detail::continuation_finalization<Continuation>>
+        {
+            using function_type = typename Continuation::function_type;
+            static char const* call(
+                lcos::detail::continuation_finalization<Continuation> const& f) noexcept
+            {
+                char const* annotation = hpx::traits::get_function_annotation<
+                    typename hpx::util::decay<function_type>::type>::call(f.this_->f_);
+                return annotation;
+            }
+        };
+    #endif
+}}
+
+namespace hpx { namespace lcos { namespace detail
+{
+    template <typename Continuation>
+    struct continuation_finalization {
+        //
+        continuation_finalization(Continuation *c) : this_(c) {}
+        //
+        template <typename Future>
+        void operator()(Future&& f) const
+        {
+            return this_->async_exec_impl(std::forward<Future>(f));
+        }
+
+        // keep the dataflow frame alive with this pointer reference
+        hpx::intrusive_ptr<Continuation> this_;
+    };
+
     template <typename Source, typename Destination>
     HPX_FORCEINLINE void transfer_result_impl(std::false_type,
         Source&& src, Destination& dest)
@@ -180,6 +222,11 @@ namespace hpx { namespace lcos { namespace detail
 
         typedef typename base_type::mutex_type mutex_type;
         typedef typename base_type::result_type result_type;
+        typedef typename std::decay<F>::type function_type;
+        typedef continuation<Future, F, ContResult> continuation_type;
+
+        friend struct continuation_finalization<continuation_type>;
+        friend struct traits::get_function_annotation<continuation_finalization<continuation_type>>;
 
     protected:
         threads::thread_id_type get_id() const
@@ -377,8 +424,9 @@ namespace hpx { namespace lcos { namespace detail
             }
 
             hpx::intrusive_ptr<continuation> this_(this);
-            hpx::util::thread_description desc(f_,
-                "hpx::parallel::execution::parallel_executor::post");
+            char const* annotation = hpx::traits::get_function_annotation<
+                function_type>::call(f_);
+            hpx::util::thread_description desc(f_, annotation);
 
             parallel::execution::detail::post_policy_dispatch<
                     hpx::launch::async_policy
@@ -423,8 +471,9 @@ namespace hpx { namespace lcos { namespace detail
             }
 
             hpx::intrusive_ptr<continuation> this_(this);
-            hpx::util::thread_description desc(f_,
-                "hpx::parallel::execution::parallel_executor::post");
+            char const* annotation = hpx::traits::get_function_annotation<
+                function_type>::call(f_);
+            hpx::util::thread_description desc(f_, annotation);
 
             parallel::execution::detail::post_policy_dispatch<
                     hpx::launch::async_policy
@@ -468,13 +517,12 @@ namespace hpx { namespace lcos { namespace detail
                 started_ = true;
             }
 
-            hpx::intrusive_ptr<continuation> this_(this);
-            parallel::execution::post(std::forward<Executor>(exec),
-                [this_ = std::move(this_),
-                    f = std::move(f)
-                ]() mutable -> void {
-                    this_->async_exec_impl(std::move(f));
-                });
+            continuation_finalization<continuation_type> this_f_(this);
+
+            parallel::execution::post(
+                std::forward<Executor>(exec),
+                std::move(this_f_),
+                std::move(f));
 
             if (&ec != &throws)
                 ec = make_success_code();
--- a/libs/execution/include/hpx/execution/executors/thread_pool_executor.hpp
+++ b/libs/execution/include/hpx/execution/executors/thread_pool_executor.hpp
@@ -90,8 +90,9 @@ namespace hpx { namespace parallel { namespace execution {
             threads::thread_schedule_hint schedulehint, launch policy, F&& f,
             Ts&&... ts)
         {
-            hpx::util::thread_description desc(
-                f, "hpx::parallel::execution::parallel_executor::post");
+            char const* annotation =
+                hpx::traits::get_function_annotation<F>::call(f);
+            hpx::util::thread_description desc(f, annotation);
 
             detail::post_policy_dispatch<decltype(policy)>::call(policy, desc,
                 pool, priority, stacksize, schedulehint, std::forward<F>(f),
--- a/tests/unit/apex/annotation_check.cpp
+++ b/tests/unit/apex/annotation_check.cpp
@@ -16,7 +16,7 @@
 #include <hpx/include/threads.hpp>
 #include <hpx/lcos/when_all.hpp>
 #include <hpx/program_options.hpp>
-#include <hpx/runtime/threads/executors/pool_executor.hpp>
+#include <hpx/execution/executors/pool_executor.hpp>
 #include <hpx/util/annotated_function.hpp>
 #include <hpx/lcos/future.hpp>
 #include <hpx/runtime/launch_policy.hpp>
@@ -203,8 +203,8 @@ void test_none()
 int hpx_main()
 {
     // setup executors
-    hpx::threads::scheduled_executor NP_executor =
-        hpx::threads::executors::pool_executor(
+    hpx::parallel::execution::pool_executor NP_executor =
+        hpx::parallel::execution::pool_executor(
             "default", hpx::threads::thread_priority_default);
     hpx::parallel::execution::parallel_executor par_exec{};
 
--- a/tests/unit/apex/annotation_check.cpp
+++ b/tests/unit/apex/annotation_check.cpp
@@ -29,6 +29,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <type_traits>
 
 //
 // This test generates a set of tasks with certain names, then checks
@@ -99,7 +100,8 @@ std::string exec_string(const Executor &)
 
 // --------------------------------------------------------------------------
 template <typename Executor>
-typename std::enable_if<hpx::traits::is_threads_executor<Executor>::value, std::string>::type
+typename std::enable_if<hpx::traits::is_threads_executor<Executor>::value,
+         std::string>::type
 execution_string(const Executor &exec) { return exec_string(exec); }
 
 template <typename Executor>
--- a/tests/unit/apex/annotation_check.cpp
+++ b/tests/unit/apex/annotation_check.cpp
@@ -1,0 +1,198 @@
+//  Copyright (c) 2019 John Biddiscombe
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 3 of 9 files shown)


hpx/lcos/dataflow.hpp
```diff
@@ -59,8 +59,52 @@
...
+    template <typename Frame>
+    struct dataflow_finalization {
+        //
+        dataflow_finalization(Frame *df) : this_(df) {}
+        using is_void = typename Frame::is_void;
+        //
+        template <typename Futures>
+        void operator()(Futures&& futures)
+        {
+            return this_->execute(is_void{}, std::move(futures));
```
hkaiser: This should be using `std::forward` instead of `std::move`.

hpx/lcos/dataflow.hpp
```diff
@@ -59,8 +59,52 @@
...
+namespace hpx { namespace lcos { namespace detail
+{
+    template <typename Frame>
+    struct dataflow_finalization {
+        //
+        dataflow_finalization(Frame *df) : this_(df) {}
+        using is_void = typename Frame::is_void;
+        //
+        template <typename Futures>
+        void operator()(Futures&& futures)
```
hkaiser: Can the operator be `const`?

hpx/lcos/dataflow.hpp
```diff
@@ -59,8 +59,52 @@
...
+    template <typename Frame>
+    struct dataflow_finalization;
+}}}
+
+namespace hpx { namespace traits {
+#if defined(HPX_HAVE_THREAD_DESCRIPTION)
+    ///////////////////////////////////////////////////////////////////////////
+    // traits specialization to get annotation from dataflow_finalization
+    template <typename Frame>
+    struct get_function_annotation<lcos::detail::dataflow_finalization<Frame>>
```
hkaiser: Please add `#include <hpx/functional/traits/get_function_annotation.hpp>`

hpx/lcos/local/packaged_continuation.hpp
```diff
@@ -32,8 +32,49 @@
...
+    template <typename Continuation>
+    struct continuation_finalization;
+}}}
+
+namespace hpx { namespace traits {
+    #if defined(HPX_HAVE_THREAD_DESCRIPTION)
+        ///////////////////////////////////////////////////////////////////////////
+        // traits specialization to get annotation from dataflow_finalization
+        template <typename Continuation>
+        struct get_function_annotation<lcos::detail::continuation_finalization<Continuation>>
```
hkaiser: Please add `#include <hpx/functional/traits/get_function_annotation.hpp>`.

hpx/lcos/local/packaged_continuation.hpp
```diff
@@ -376,8 +422,9 @@ namespace hpx { namespace lcos { namespace detail
             }
 
             hpx::intrusive_ptr<continuation> this_(this);
-            hpx::util::thread_description desc(f_,
-                "hpx::parallel::execution::parallel_executor::post");
+            char const* annotation = hpx::traits::get_function_annotation<
+                function_type>::call(f_);
```
hkaiser: What will be the fallback in this case (i.e. no specialization for `get_function_annotation` is found?

hpx/lcos/local/packaged_continuation.hpp
```diff
@@ -32,8 +32,49 @@
...
+            static char const* call(
+                lcos::detail::continuation_finalization<Continuation> const& f) noexcept
+            {
+                char const* annotation = hpx::traits::get_function_annotation<
+                    typename hpx::util::decay<function_type>::type>::call(f.this_->f_);
+                return annotation;
+            }
+        };
+    #endif
+}}
```
hkaiser: Is this specialization really needed? The code below already calls `get_function_annotation` on the embedded function `f_`. Please explain.

hpx/lcos/dataflow.hpp
biddisco: yup. one of the builders failed and I found it too. I'll force push these commits shortly

hpx/lcos/local/packaged_continuation.hpp
biddisco: The invocations for version with and without executors and different policies go through different places. I'm not sure I know which bits you are referring to as unneeded.

hpx/lcos/local/packaged_continuation.hpp
biddisco: The default is to return "<unknown>" or something of that kind. I suppose I'll need to check and see what appears in the traces.

hpx/lcos/local/packaged_continuation.hpp
hkaiser: Well, I think the specialization I attached this comment onto is not really needed... What is the use case (i.e. test) that is fixed by _this_ particular change?

biddisco:
```
    {
        hpx::future<int> f1 = hpx::make_ready_future(5);
        results.emplace_back(f1.then(
            NP_executor,
            hpx::util::annotated_function(
                [](auto &&f1) {
                    dummy_task(std::size_t(1000));
                },
                "Continuation with executor")));
    }
```

hkaiser: That code seems to work completely without your patch. I.e.
```
#include <hpx/hpx.hpp>
#include <hpx/hpx_main.hpp>

int main()
{
    hpx::future<int> f1 = hpx::make_ready_future(5);
     hpx::future<void> f2 = f1.then(hpx::parallel::execution::parallel_executor{},
        hpx::util::annotated_function(
            [](auto&& f1)
            {
                std::cout << hpx::threads::get_thread_description(
                    hpx::threads::get_self_id()) << "\n";
            }, "Continuation with executor"));
    f2.get();
    return 0;
}
```
Will print `Continuation with executor` without your patch. What do I miss?

BTW: could you add some tests for your changes, please?

biddisco: try it with 
```
    hpx::threads::scheduled_executor NP_executor =
           hpx::threads::executors::pool_executor(
               "default", hpx::threads::thread_priority_default);
  
```

biddisco: There isn't really a way of testing this. you have to use tools like vampir to load the generated otf files to see if the functions are visible. The file `tests/unit/resource/scheduler_priority_check.cpp` on my current branch has been extended with annotations for some of these and I will commit it to the branch once I'm happy that apex works. The last time I was able to use apex was back in early 2018, so there's no great rush with this.

biddisco: The tests I'm running are ...

```
    hpx::threads::scheduled_executor NP_executor =
        hpx::threads::executors::pool_executor(
            "default", hpx::threads::thread_priority_default);

    std::vector<hpx::future<void>> results;

    {
        hpx::future<int> f1 = hpx::make_ready_future(5);
        hpx::future<int> f2 = hpx::make_ready_future(5);
        results.emplace_back(hpx::dataflow(
            NP_executor,
            hpx::util::annotated_function(
                [](auto &&f1, auto &&f2) {
                    dummy_task(std::size_t(1000));
                },
                "Dataflow with executor"), f1, f2));
    }

    {
        hpx::future<int> f1 = hpx::make_ready_future(5);
        hpx::future<int> f2 = hpx::make_ready_future(5);
        results.emplace_back(hpx::dataflow(
            hpx::util::annotated_function(
                [](auto &&f1, auto &&f2) {
                    dummy_task(std::size_t(1000));
                },
                "Dataflow without executor"), f1, f2));
    }

    {
        hpx::future<int> f1 = hpx::make_ready_future(5);
        hpx::future<int> f2 = hpx::make_ready_future(5);
        results.emplace_back(hpx::dataflow(
            hpx::launch::sync,
            hpx::util::annotated_function(
                [](auto &&f1, auto &&f2) {
                    dummy_task(std::size_t(1000));
                },
                "Dataflow with sync policy"), f1, f2));
    }

    {
        hpx::future<int> f1 = hpx::make_ready_future(5);
        results.emplace_back(f1.then(
            NP_executor,
            hpx::util::annotated_function(
                [](auto &&f1) {
                    dummy_task(std::size_t(1000));
                },
                "Continuation with executor")));
    }

    {
        hpx::future<int> f1 = hpx::make_ready_future(5);
        results.emplace_back(f1.then(
            hpx::util::annotated_function(
                [](auto &&f1) {
                    dummy_task(std::size_t(1000));
                },
                "Continuation without executor")));
    }

    {
        hpx::future<int> f1 = hpx::make_ready_future(5);
        results.emplace_back(f1.then(
            hpx::launch::sync,
            hpx::util::annotated_function(
                [](auto &&f1) {
                    dummy_task(std::size_t(1000));
                },
                "Continuation with sync policy")));
    }

    // wait for completion
    hpx::when_all(results).get();
```

biddisco: I should add some other policies too ...

biddisco: I forgot to add the combinations that use `hpx:unwrapping` - I will commit a test shortly to the branch

biddisco:
I added a check/test that can be run and visually inspected for errors

Without the fixes in this PR
```
Timer                                                : #calls  |    mean  |   total  |  % total  
------------------------------------------------------------------------------------------------
                                      - Continuation :        1      0.001      0.001      0.839
                - Executor non-threaded Continuation :        1      0.001      0.001      0.805
                                - async Continuation :        2      0.001      0.002      1.617
                                 - fork Continuation :        1      0.001      0.001      0.853
                                           <unknown> :        2      0.001      0.002      1.603
                        async_launch_policy_dispatch :        3      0.000      0.000      0.078
            broadcast_call_shutdown_functions_action :        2      0.000      0.000      0.163
                      call_shutdown_functions_action :        2      0.001      0.003      2.243
   hpx::parallel::execution::parallel_executor::post :        9      0.001      0.008      5.716
                              load_components_action :        1      0.006      0.006      4.746
                   primary_namespace_colocate_action :        2      0.000      0.000      0.105
                                          run_helper :        1      0.012      0.012      9.336
                                 shutdown_all_action :        1      0.001      0.001      0.512
                                           APEX Idle :                          0.094     71.386
------------------------------------------------------------------------------------------------
```
This PR
```
Timer                                                : #calls  |    mean  |   total  |  % total                                                                                               
------------------------------------------------------------------------------------------------                                                                                              
                                      - Continuation :        1      0.001      0.001      0.534                                                                                              
                                          - Dataflow :        1      0.001      0.001      0.531                                                                                              
                - Executor non-threaded     Dataflow :        1      0.001      0.001      0.553                                                                                              
                - Executor non-threaded Continuation :        1      0.001      0.001      0.521
                    - Executor threaded     Dataflow :        1      0.001      0.001      0.526
                    - Executor threaded Continuation :        1      0.001      0.001      0.524
                                - apply     Dataflow :        1      0.001      0.001      0.528
                                - async     Dataflow :        2      0.001      0.002      1.055
                                - async Continuation :        2      0.001      0.002      1.068
                             - deferred     Dataflow :        1      0.001      0.001      0.531
                                 - fork     Dataflow :        1      0.001      0.001      0.514
                                 - fork Continuation :        1      0.001      0.001      0.528
hpx::components::server::runtime_support::shutdow... :        1      0.001      0.001      0.294
hpx::components::server::runtime_support::load_co... :        1      0.016      0.016      7.879
hpx::components::server::runtime_support::call_sh... :        4      0.001      0.003      1.447
hpx::agas::server::primary_namespace::colocate_ac... :        2      0.000      0.000      0.072
hpx::actions::action<void (*)(hpx::components::se... :        2      0.000      0.000      0.109
                        async_launch_policy_dispatch :        3      0.000      0.000      0.089
                                          run_helper :        1      0.017      0.017      8.487
                                           APEX Idle :                          0.150     74.211
------------------------------------------------------------------------------------------------
```

biddisco:
oops. had `async` in there twice. removed one and force pushed, counts in output will be off by one.

biddisco:
note that I added a minus sin to the task names solely to make them appear at the start of the list alphabetically. I'll tweak that to make it more sensible next time.

khuck:
@biddisco With `APEX_CSV_OUTPUT=1` you get the output above (or you can call `apex::apex_options::use_screen_output(true);` at the start of the test program).  Having done either of those two, you can add these types of test properties to the ctest definition:

```
set_tests_properties(dataflow_apex_test PROPERTIES PASS_REGULAR_EXPRESSION "- Dataflow :        1")
set_tests_properties(dataflow_apex_test PROPERTIES PASS_REGULAR_EXPRESSION "- Executor non-threaded     Dataflow :        1")
...
```

...or call `apex::get_profile` like happens in hpx/tests/unit/apex/apex_action_count.cpp and do the test programmatically.

biddisco:
Thanks @khuck - I have added a call to `apex::apex_options::use_screen_output(true);` and prefixed the task names in such a way that they appear alphabetically in a checkable way. Then with ctest we can scan for the right string and pass/fail the test.

The git log includes this comment - even with the patches on my branch, not all tasks are correctly detected. There is still work to be done, but this makes the majority of tasks we are using in our cholesky code visible, (though not yet on N>1 mpi ranks).
```
    several continuation tasks are not correctly picked up by apex.
    hpx::launch::sync is not handled by apex
    hpx::unwrapping hides annotations - this should be addressed
```
I'm currently testing for the following task names (I'm not checking the count, what we really care about for now is if they stop appearing in the output). 
```
719: Timer                                                : #calls  |    mean  |   total  |  % total  
719: ------------------------------------------------------------------------------------------------
719:                                           1-Dataflow :        1      0.001      0.001      0.626
719:                     1-Executor non-threaded Dataflow :        1      0.001      0.001      0.601
719:                         1-Executor threaded Dataflow :        1      0.001      0.001      0.621
719:                                     1-apply Dataflow :        1      0.006      0.006      3.353
719:                                     1-async Dataflow :        1      0.001      0.001      0.612
719:                                  1-deferred Dataflow :        1      0.001      0.001      0.640
719:                                      1-fork Dataflow :        1      0.001      0.001      0.613
719:                                       2-Continuation :        1      0.001      0.001      0.619
719:                 2-Executor non-threaded Continuation :        1      0.001      0.001      0.608
719:                     2-Executor threaded Continuation :        1      0.001      0.001      0.621
719:                                 2-async Continuation :        1      0.001      0.001      0.625
719:                                  2-fork Continuation :        1      0.001      0.001      0.622
```
(continuations are still missing `apply`, `deferred` launch policies)

biddisco:
Question : since `hpx::annotate_function();` the RAII alternative is no longer functioning (and hence things like `launch::sync` don't work) should we remove it altogether and all places that call it? There are still quite a lot of places in the code where it is sprinkled around.

hkaiser:
> e code where it is sprinkled around.

Did it ever properly 'work' (for APEX)? Is there a way to make it work for APEX? @khuck?

Personally, I would prefer having a RAII version of the functionality. IIUC, today `hpx::annotate_function()` does not really annotate anything but forces a typetrait specialization to be generated that will cause anybody that asks for the name of a function of this function-type will get the annotated string returned. Is my understanding correct?

khuck:
> Thanks @khuck - I have added a call to `apex::apex_options::use_screen_output(true);` and prefixed the task names in such a way that they appear alphabetically in a checkable way. Then with ctest we can scan for the right string and pass/fail the test.
> 
> The git log includes this comment - even with the patches on my branch, not all tasks are correctly detected. There is still work to be done, but this makes the majority of tasks we are using in our cholesky code visible, (though not yet on N>1 mpi ranks).
> 
> ```
>     several continuation tasks are not correctly picked up by apex.
>     hpx::launch::sync is not handled by apex
>     hpx::unwrapping hides annotations - this should be addressed
> ```
> 
> I'm currently testing for the following task names (I'm not checking the count, what we really care about for now is if they stop appearing in the output).
> 
> ```
> 719: Timer                                                : #calls  |    mean  |   total  |  % total  
> 719: ------------------------------------------------------------------------------------------------
> 719:                                           1-Dataflow :        1      0.001      0.001      0.626
> 719:                     1-Executor non-threaded Dataflow :        1      0.001      0.001      0.601
> 719:                         1-Executor threaded Dataflow :        1      0.001      0.001      0.621
> 719:                                     1-apply Dataflow :        1      0.006      0.006      3.353
> 719:                                     1-async Dataflow :        1      0.001      0.001      0.612
> 719:                                  1-deferred Dataflow :        1      0.001      0.001      0.640
> 719:                                      1-fork Dataflow :        1      0.001      0.001      0.613
> 719:                                       2-Continuation :        1      0.001      0.001      0.619
> 719:                 2-Executor non-threaded Continuation :        1      0.001      0.001      0.608
> 719:                     2-Executor threaded Continuation :        1      0.001      0.001      0.621
> 719:                                 2-async Continuation :        1      0.001      0.001      0.625
> 719:                                  2-fork Continuation :        1      0.001      0.001      0.622
> ```
> 
> (continuations are still missing `apply`, `deferred` launch policies)

The `sync` version doesn't create a task, so there is no APEX instrumentation around it.  Which is interesting, because direct actions are instrumented, if I remember correctly.

msimberg:
I think the CircleCI failure is unrelated. I've restarted that build.

hpx/lcos/local/packaged_continuation.hpp
hkaiser: I still believe, that this specialization is not really needed. Please verify. The code below already calls `get_function_annotation` on the embedded function `f_`. That should already be sufficient.

tests/unit/apex/annotation_check.cpp
```diff
@@ -0,0 +1,226 @@
...
+        auto now = std::chrono::steady_clock::now();
+        auto elapsed =
+            std::chrono::duration_cast<std::chrono::microseconds>(now - start);
+        sleep = (elapsed < std::chrono::microseconds(n));
+    } while (sleep);
+}
+
+// --------------------------------------------------------------------------
+// string for a policy
+std::string policy_string(const hpx::launch& policy)
```
hkaiser: We normally use the east-const notation in our code base (not to start a religious war...). I think we should be consistent.

biddisco:
> Why did you clang-format the whole file? This makes it close to impossible to review things.

Our checks run clang format on the whole file. If there is a way of just formatting the part you modified (and have it still pass our tests), please tell me.

hkaiser:
> Why did you clang-format the whole file? This makes it close to impossible to review things.
> 
> Our checks run clang format on the whole file. If there is a way of just formatting the part you modified (and have it still pass our tests), please tell me.

Our checks run clang-format on all files in the modules (`libs`) directory only. There we do the whole-file-formatting as a step on its own (we're trying not to change things while moving a file into a module as much as possible). We have not formatted any of the other files (outside of `libs`) yet and I believe there is no reason to do so. 

So generally speaking, if we do format a whole file (for whatever reason), we do try to separate the formatting from the rest of the changes to enable reviewing.

msimberg:
@biddisco `clang-format` also has an option to format specific lines only: `clang-format -lines=start:end`. It's quite horrible to use manually but many editors have an option to `clang-format` only a selection.

aurianer:
There is also a `git-clang-format`used to format the diff of the commits, but I haven't tried it for now

biddisco:
I removed most of the clang format stuff and squashed the tiny codacy and clang format changes into one commit.

biddisco:
I removed the clan-formatted files and cleaned up inspect, but now clang-format test fails again
```
#!/bin/bash -eo pipefail
cd /hpx/source/libs && shopt -s globstar # to activate the ** globbing
clang-format-8 -i **/*.{cpp,hpp}
git diff --exit-code > /tmp/modified_format_files.txt

Exited with code exit status 1
```
It really looks to me as though it just clang formats all files in the commit and fails you if there is any non zero output. I waste far too much time on inspect and format errors.

biddisco:
ok. my bad. it checks the files in lib and I made some changes there. Cleaned up again and force pushed. fingers crossed.

msimberg:
@biddisco note that the end goal is that one can enable auto-format on save and you never have to think about formatting again. As soon as we get everything formatted once things should become much easier.

kordejong:
I have tested commit 8707534 and can confirm that I now get useful information in my Vampir plots. When annotating lambdas passed to dataflow this information ends up in the trace. Great!

msimberg:
What's currently blocking this?

hkaiser:
> What's currently blocking this?

I'm still not convinced that the specializations added are actually needed. See my comments above.

biddisco:
`hpx::parallel::execution::parallel_executor` isn't a thread_pool (or thread based?) executor and it goes through a different customization point I expect. This PR fixed my problems and the user's ones, but if it isn't of value then just delete it and I'll maintain my own branch for using apex.

hkaiser:
@biddisco I have rebased this branch onto master. Let's see how the tests fare.

libs/execution/include/hpx/execution/executors/thread_pool_executor.hpp
```diff
@@ -90,8 +90,9 @@ namespace hpx { namespace parallel { namespace execution {
             threads::thread_schedule_hint schedulehint, launch policy, F&& f,
             Ts&&... ts)
         {
-            hpx::util::thread_description desc(
-                f, "hpx::parallel::execution::parallel_executor::post");
+            char const* annotation =
+                hpx::traits::get_function_annotation<F>::call(f);
+            hpx::util::thread_description desc(f, annotation);
```
msimberg, line 95: @biddisco was this change actually needed? None of the other places doing the same have changed... If I'm reading the constructor for `thread_description` correctly it will first try to use `hpx::traits::get_function_annotation<F>::call(f)`, and only if it's null will it use the fallback (the second argument). If we don't want the fallback that can be left out.

libs/execution/include/hpx/execution/executors/parallel_executor.hpp
```diff
@@ -206,8 +206,9 @@ namespace hpx { namespace parallel { namespace execution {
         template <typename F, typename... Ts>
         void post(F&& f, Ts&&... ts) const
         {
-            hpx::util::thread_description desc(
-                f, "hpx::parallel::execution::parallel_executor::post");
+            char const* annotation =
+                hpx::traits::get_function_annotation<F>::call(f);
+            hpx::util::thread_description desc(f, annotation);
```
msimberg, line 211: Same question as in `thread_pool_executor.hpp`.

hkaiser:
@msimberg I have added more assertions here that make sure `local_thread_num` is not being used whenever it happens to be `-1`. Those asserts fire, unfortunately - looks like your fixes have not solved all the related problems.

@biddisco The guided executor is still failing to compile since this branch was rebased onto master. I don't know what is going on, care to have a look (https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/2899/workflows/c2f4ac28-1268-4c02-a04b-19f0fd4fa8f6/jobs/177995)?

msimberg:
> @msimberg I have added more assertions here that make sure `local_thread_num` is not being used whenever it happens to be `-1`. Those asserts fire, unfortunately (https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/2899/workflows/c2f4ac28-1268-4c02-a04b-19f0fd4fa8f6/jobs/177989) - looks like your fixes have not solved all the related problems.

`local_num` is still allowed to be `-1`. I had intended for only `thread_num` to be used later which is not allowed to be `-1`. I've added asserts for `thread_num` instead.

I've merged master into this branch now (I've not fixed building the `async_customization` example).

libs/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
```diff
@@ -275,10 +275,11 @@ namespace hpx { namespace threads { namespace policies {
 
                 // ------------------------------------------------------------
                 bool cleanup_terminated(
-                    std::size_t thread_num, bool delete_all) override
+                    std::size_t /*thread_num*/, bool delete_all) override
                 {
                     std::size_t local_num = local_thread_number();
-                    HPX_ASSERT(local_num < num_workers_);
+                    HPX_ASSERT(local_num != std::size_t(-1) &&
+                        local_num < num_workers_);
```
hkaiser, line 282: Didn't you say `local_num` is _allowed_ to be `-1`?

libs/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
```diff
@@ -406,12 +411,14 @@ namespace hpx { namespace threads { namespace policies {
                             d_lookup_[local_num] == domain_num)
```
hkaiser, line 411: This operation will still fail if `local_num == -1`

biddisco:
Changes to thread numbering have nothing to do with this dataflow-annotations branch. They should not be included here. If other changes have been made and merged into this, they should be fixed in their own PR's and branches.

libs/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
```diff
@@ -618,7 +626,8 @@ namespace hpx { namespace threads { namespace policies {
                     bool enable_stealing) override
                 {
                     std::size_t this_thread = local_thread_number();
-                    HPX_ASSERT(this_thread < num_workers_);
+                    HPX_ASSERT(this_thread != std::size_t(-1) &&
+                        this_thread < num_workers_);
```
hkaiser, line 630: Same here, this will fire if `thread_num == -1`.

hkaiser:
> Changes to thread numbering have nothing to do with this dataflow-annotations branch. They should not be included here. If other changes have been made and merged into this, they should be fixed in their own PR's and branches.

Right... The only problem is that the issues manifest themselves on this branch only, I think.

libs/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
msimberg: Ok, almost always... I didn't want to change the behaviour here, just make it clear that `thread_num` is not used. And my changed assert doesn't make much sense with `std::size_t`. In any case, @biddisco will drop these changes and redo the checks if necessary.

biddisco:
I will fix this branch and the thread numbering.

msimberg:
@biddisco are you able to reproduce the segfaults with the apex builder? They might be the failures we were trying to fix earlier.

biddisco:
I do not know what segfaults you are referring to. The dataflow PR looks pretty clean on cdash to me - could you please elaborate.

msimberg:
https://cdash.cscs.ch//viewTest.php?onlyfailed&buildid=105728

biddisco:
Those tests appear to be failing on multiple dashboards 
https://cdash.cscs.ch//testSummary.php?project=2&name=tests.performance.local.resume_suspend&date=2020-04-27
it does not appear to be connected to this PR

msimberg:
> Those tests appear to be failing on multiple dashboards
> https://cdash.cscs.ch//testSummary.php?project=2&name=tests.performance.local.resume_suspend&date=2020-04-27
> it does not appear to be connected to this PR

The tests failed on those two PRs (#4570 and #4551) because the build failed (no tests to run). The tests are not failing on master.

biddisco:
OK. I'll delete this PR and come back to it some other time.

biddisco:
Reopening this after debugging with a clang build. No idea why gcc builds pass and clang ones fail, because they were both dereferencing a nullptr that was taken from the default implementation of `get_function_annotation` - this is fixed by returning "<anonymous>" instead of nullptr so that apex always has something to store. This might fix other apex related problems that I was having when reductions of task names took place in distributed runs. I will check this. 
Thanks @msimberg for being annoying enough that I took some time to investigate this.

biddisco:
FYI: rebased onto latest master.

msimberg:
> Thanks @msimberg for being annoying enough that I took some time to investigate this.

You're welcome, I guess... how come the nullptrs aren't being dereferenced on master? Are there maybe some checks missing for nullptrs (the fallback strings are fine by me though, I'm just wondering what changed)?

hkaiser:
> Reopening this after debugging with a clang build. No idea why gcc builds pass and clang ones fail, because they were both dereferencing a nullptr that was taken from the default implementation of get_function_annotation - this is fixed by returning "" instead of nullptr so that apex always has something to store. This might fix other apex related problems that I was having when reductions of task names took place in distributed runs. I will check this.

May I suggest using something more descriptive as the fallback - perhaps `"<unknown>"`? That might help identifying issues down the road.

Also, @biddisco thanks for not dropping the ball on this!

biddisco:
> May I suggest using something more descriptive as the fallback - perhaps `"<unknown>"`? That might help identifying issues down the road.

It actually returns `"<anonymous>"` but I'm not sure why the markup cha nged my text to an empty string.

biddisco:
> how come the nullptrs aren't being dereferenced on master? Are there maybe some checks missing for nullptrs (the fallback strings are fine by me though, I'm just wondering what changed)?

As far as I can tell, the nullptrs get derefernced to some random stack memory on gcc and cause a segfault in clang. I added printfs and gcc was dumping out rubbish, consistent with some of the stuff I saw being generated in OTF files that woulnd't load properly in vampir.

I should investigate further, but when I saw the nullptr stuff being accessed I changed the unspecialized version of `get_annotated_function` and everything started working, so I stopped then.

msimberg:
@biddisco this just needs formatting now. I think we can merge this after that.

msimberg:
@biddisco I just rebased and clang-formatted this branch. If CI comes back happy I'll merge it.

PR #4316, opened 2020-01-21
Add clang format indentppdirectives

aurianer:
Add indentation for preprocessor directives

- Necessity to change the header guard from `#if !defined` to `#ifndef`in order for the clang-format option to work properly


hkaiser:
Will there be a way to tell clang-format to indent the pp directives by less than 4 characters? I think in this case even one character indentation would be sufficient, possibly 2 characters.

aurianer:
> Will there be a way to tell clang-format to indent the pp directives by less than 4 characters? I think in this case even one character indentation would be sufficient, possibly 2 characters.

I've search for it but didn't find any additional option to do that, it may be possible though cause the example of `IndentPPDirectives` in the clang documentation is indented with only 2 spaces :
https://clang.llvm.org/docs/ClangFormatStyleOptions.html

hkaiser:
@aurianer may I ask you to try separating the #ifndef change from the reformatting? 1300 files are impossible to review otherwise.

aurianer:
> @aurianer may I ask you to try separating the #ifndef change from the reformatting? 1300 files are impossible to review otherwise.

Actually it is already the case :
- first commit : change of the #ifndef
- second commit : add the clang format option in the .clang-format file
- third commit : clang-format all the modules
- last : fix some inspect errors
So if you isolate the first commit from the commit tab, that should be fine ;)

hkaiser:
@aurianer thanks!

libs/algorithms/include/hpx/parallel/datapar/loop.hpp
```diff
@@ -10,22 +10,22 @@
...
+#    include <hpx/parallel/datapar/iterator_helpers.hpp>
+#    include <hpx/parallel/util/loop.hpp>
+#    include <hpx/type_support/decay.hpp>
+
+#    include <algorithm>
+#    include <cstddef>
+#    include <cstdint>
+#    include <iterator>
+#    include <type_traits>
+#    include <utility>
```
msimberg: What do people think of this formatting? My initial thought was that it's ugly, but probably just because it's unusual. I think we can keep it like this because it's consistent with all the other macro formatting.

libs/local_lcos/include/hpx/local_lcos/force_linking.hpp
```diff
@@ -5,10 +5,10 @@
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 #if !defined(HPX_LOCAL_LCOS_FORCE_LINKING_HPP)
```
msimberg:
```suggestion
#ifndef HPX_LOCAL_LCOS_FORCE_LINKING_HPP
```
?

libs/algorithms/include/hpx/parallel/datapar/loop.hpp
hkaiser: I'm weakly against this.

libs/local_lcos/include/hpx/local_lcos/force_linking.hpp
hkaiser: clang-format recognizes the `#ifndef` as #include guards, however it does not recognize the `#if !defined()` version as such.

aurianer:
I think the failure is not related to this PR

hkaiser:
After thinking about this some more: I believe -- if we decided to go with enabling `IndentPPDirectives` -- we should go with `PPDIS_BeforeHash`. The rationale is that `PPDIS_AfterHash` complicates using grep or similar to find things. 

However, I'm still not convinced that indenting the pp directives is a good thing to begin with (considering that the related clang-format options are very limited).

PR #4322, opened 2020-01-23
Undefine VT1 and VT2 after boost includes.

khuck:
The boost asio support includes termios.h.
The termios.h file on ppc64le defines these macros, which
are also used by blaze, blaze_tensor as Template names.
Make sure we undefine them before continuing.

Fixes phylanx bug https://github.com/STEllAR-GROUP/phylanx/issues/1088

## Proposed Changes

  - After boost/asio includes, undefine VT1 and VT2.

As requested by @hkaiser, this is a better solution than using `sed` on the blaze library, and it works.

### DIFF
--- a/hpx/plugins/parcelport/tcp/receiver.hpp
+++ b/hpx/plugins/parcelport/tcp/receiver.hpp
@@ -34,6 +34,16 @@
 #include <boost/asio/placeholders.hpp>
 #include <boost/asio/read.hpp>
 #include <boost/asio/write.hpp>
+/* The boost asio support includes termios.h.
+ * The termios.h file on ppc64le defines these macros, which
+ * are also used by blaze, blaze_tensor as Template names.
+ * Make sure we undefine them before continuing. */
+#ifdef VT1
+#undef VT1
+#endif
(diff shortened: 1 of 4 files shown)


hpx/plugins/parcelport/tcp/receiver.hpp
```diff
@@ -34,6 +34,16 @@
 #include <boost/asio/placeholders.hpp>
 #include <boost/asio/read.hpp>
 #include <boost/asio/write.hpp>
+/* The boost asio support includes termios.h.
+ * The termios.h file on ppc64le defines these macros, which
+ * are also used by blaze, blaze_tensor as Template names.
+ * Make sure we undefine them before continuing. */
+#ifdef VT1
+#undef VT1
+#endif
```
hkaiser: There is no need to check whether the symbols are defined. A simple `#undef` does the trick. It will be ignored if the symbol wasn't defined.

khuck:
@hkaiser yes, this fixes the blaze issue.  I can remove the `#ifdef` checks.

PR #4327, opened 2020-01-24
Use INTERNAL_FLAGS option for all examples and components

msimberg:


biddisco:
What does INTERNAL_FLAGS actually do/hold?

msimberg:
It holds things like warning flags that should not be exposed to users of HPX.

msimberg:
Can we go ahead with this?

aurianer:
LGTM thanks!

PR #4328, opened 2020-01-24
Generate global header for basic_execution module

msimberg:
Remove manually created global header.


msimberg:
The failures are problems on Piz Daint.

PR #4331, opened 2020-01-24
This adds the missing tests for async_colocated and async_continue_colocated

hkaiser:
- fixing a problem revealed by the new tests


hkaiser:
> Thanks! I'm not quite sure exactly what issues codacy is complaining about (it does show some new duplication, but it seems to think there are some issues which I don't see listed...)

It reported those at first but I suppressed them (as they were false positives). No idea how to convince it not to complain anymore, however.

PR #4332, opened 2020-01-24
This is a collection of mostly minor (cleanup) fixes

hkaiser:
This is a collection of mostly minor (cleanup) fixes to various places of the threading subsystem

### DIFF
--- a/hpx/runtime/threads/policies/thread_queue.hpp
+++ b/hpx/runtime/threads/policies/thread_queue.hpp
@@ -312,12 +312,12 @@ namespace hpx { namespace threads { namespace policies {
                 }
                 else if (work_items_.empty())
                 {
-                    add_count =
-                        parameters_
-                            .min_add_new_count_;    // add this number of threads
+                    // add this number of threads
+                    add_count = parameters_.min_add_new_count_;
+
+                    // increase max_thread_count
                     parameters_.max_thread_count_ +=
-                        parameters_
-                            .min_add_new_count_;    // increase max_thread_count //-V101
+                        parameters_.min_add_new_count_;    //-V101
                 }
                 else
                 {
@@ -374,7 +374,7 @@ namespace hpx { namespace threads { namespace policies {
             util::tick_counter tc(cleanup_terminated_time_);
 #endif
 
-            if (terminated_items_count_ == 0)
+            if (terminated_items_count_.load(std::memory_order_acquire) == 0)
                 return true;
 
             if (delete_all)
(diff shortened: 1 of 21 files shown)


hpx/runtime/threads/policies/thread_queue.hpp
```diff
@@ -416,27 +416,28 @@ namespace hpx { namespace threads { namespace policies {
...
                     // this thread has to be in this map
-                    HPX_ASSERT(it != thread_map_.end());
-
-                    recycle_thread(*it);
+                    HPX_ASSERT(thread_map_.find(tid) != thread_map_.end());
 
-                    thread_map_.erase(it);
+                    bool deleted = thread_map_.erase(tid) != 0;
+                    HPX_ASSERT(deleted);
+                    (void)deleted;
```
msimberg:
```suggestion
                    (HPX_UNUSED(deleted);
```

The changes in this function aren't intended to change behaviour its behaviour, right? It's not obvious to me if there's some functional change from recycling the thread last (other than allowing you to check that the `erase` is successful).

libs/coroutines/include/hpx/coroutines/coroutine.hpp
```diff
@@ -47,6 +47,7 @@
 #include <utility>
 
 namespace hpx { namespace threads { namespace coroutines {
+
```
msimberg: Just so we don't go back and forth on formatting changes like this: I tend to remove empty lines like this :) Happy to leave them in the future if you'd prefer?

hpx/runtime/threads/policies/thread_queue.hpp
hkaiser: This change was done to make it consistent with another spot in the code where we do similar things. No functional change was intended.

libs/coroutines/include/hpx/coroutines/coroutine.hpp
hkaiser: Uhh. I personally prefer to have them. But I can live with it if we decided to remove them.

msimberg: No worries, let's keep it. I won't be deleting the ones that are there, but I may not always remember to add them either to new code. This is really no big deal.

msimberg:
Just formatting left on one file (https://159071-4455628-gh.circle-artifacts.com/0/hpx/artifacts/modified_format_files.txt), then this looks pretty much good to go (at least according to CircleCI and Travis).

PR #4341, opened 2020-01-27
Fix launch process test with APEX enabled

msimberg:
The test segfaulted trying to access parent task information from another
locality. This happened because parent task handling was enabled when the
initial number of localities was 1. However, even when the initial number of
localities is 1 additional localities can connect later (as in the launch
process test) leading to APEX trying to acccess task information from another
locality. This changes the behaviour to only enable parent task handling when
networking is disabled.


msimberg:
Going to cheat and merge this even though the apex builder hasn't finished yet. I'll come back to this if it turns out the issue remains.

Issue #4336, opened 2020-01-26
Building HPX 1.4.0 with IO Counters breaks (Windows)


McKillroy:
## Expected Behavior

to build

## Actual Behavior

breaks

## Steps to Reproduce the Problem

build with -DHPX_WITH_IO_COUNTERS=ON

## Specifications

... Please describe your environment

  - HPX Version: 1.4.0
  - Platform (compiler, OS): MSVC 2019, Windows 10 (boith latest)
Error: 

```
[...]

PR #4343, opened 2020-01-27
Give a hard error if IO counters are enabled on non-Linux systems

msimberg:
"Fixes" #4336.

@McKillroy how does this look to you?


McKillroy:
I didn't test it, but I like the idea and the thinking behind it.

PR #4351, opened 2020-01-29
Move generated documentation to hpx-docs repo

msimberg:
This is still WIP until I see that everything gets pushed correctly to the new repo. Currently it's https://github.com/STEllAR-GROUP/hpx-docs, and the docs would be at https://stellar-group.github.io/hpx-docs/latest/html/index.html, but feel free to suggest another name for it. Once this is finalized I'll clean up the `gh-pages` branch here.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -8,9 +8,9 @@
 |circleci_status| |zenodo_doi| |codacy|
 
 Documentation: `latest
-<https://stellar-group.github.io/hpx/docs/sphinx/latest/html/index.html>`_,
+<https://stellar-group.github.io/hpx-docs/latest/html/index.html>`_,
 `development (master)
-<https://stellar-group.github.io/hpx/docs/sphinx/branches/master/html/index.html>`_
+<https://stellar-group.github.io/hpx-docs/branches/master/html/index.html>`_
 
 ===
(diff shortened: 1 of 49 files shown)


hkaiser:
The downside of this is that it separates the docs from the code repository. :/

msimberg:
We can still go with the squash-and-force-push if that seems better. I would like to do something about this though because a 2 GB repo isn't very nice.

I'd argue though that browsing the `gh-pages` branch manually in the repo isn't really a use case for anyone. Either one looks at the source rst files in the `hpx` repos, or the generated docs at `stellar-group.github.io`. Which repo the HTML files are in is secondary *except* for the fact that the URL changes. Is the URL change a no-go in your opinion?

hkaiser:
Ok, looks like I misunderstood things. Please disregard my comment - all is good to go.

msimberg:
@hkaiser I left out the crucial "*generated* documentation" from the title :)

The docs are now pushed to the `hpx-docs` repo: https://stellar-group.github.io/hpx-docs. I'll do all the cleanup on this side once this is merged.

aurianer:
Thanks !

PR #4354, opened 2020-01-29
Lightweight utility to hold a pack as members

K-ballo:


msimberg:
For those of us who don't understand this from a glance, what exactly is the relationship or difference between `tuple` and `member_pack`?

K-ballo:
> For those of us who don't understand this from a glance, what exactly is the relationship or difference between `tuple` and `member_pack`?

This `member_pack` is what was previously the core of `tuple` (which is now implemented on top of `member_pack`). It is the minimum machinery necessary to hold and access a pack as members. In addition, `tuple` has extra convenience machinery like value initialization, conditional implicit conversions, customized assign-through assignment operators, relational operators, interaction with other tuple-like types, etc.

msimberg:
@K-ballo I think you can ignore most of the pycicle build errors (they seem to be due to Piz Daint, rather than this PR). This one does look potentially real though: https://cdash.cscs.ch//viewBuildError.php?buildid=93989?

PR #4355, opened 2020-01-30
Fix spellcheck step in CircleCI config

msimberg:
Only run the step on changed files in pull requests, not on e.g. master.
Also skip HDF5 and PNG files.


msimberg:
Merging this now to quiet down the CircleCI builds.

PR #4358, opened 2020-02-01
Doc updates; generating documentation. Will likely need heavy editing.

rstobaugh:
First attempt generating documentation for a module. Let me know if it's too long/short, if it needs more/less detail, etc.

### DIFF
--- a/libs/serialization/docs/index.rst
+++ b/libs/serialization/docs/index.rst
@@ -11,8 +11,30 @@
 serialization
 =============
 
-This module provides serialization primitives and support for many C++ standard
-library and Boost types.
+This module provides serialization primitives and support for all built-in
+types as well as all C++ Standard Library collection and utility types. This
+list is extended by |hpx| vocabulary types with proper support for global
+reference counting. |hpx|'s mode of serialization is dervied from `Boost's
+serialization model
+<https://www.boost.org/doc/libs/1_72_0/libs/serialization/doc/index.html>`_
+and, as such, is mostly interface compatible with
+its Boost counterpart.
+
+The purest form of serializing data is to copy the content of the payload bit
+by bit; however, this method is impractical for generic C++ types, which might
+be composed of more than just regular built-in types. Instead, |hpx|'s approach
+to serialization is derived from the Boost Serialization library, and is geared
+towards allowing the programmer of a given class explicit control and syntax of
+what to serialize. It is based on operator overloading of two special archive
+types that hold a buffer or stream to store the serialized data and is
+responsible for dispatching the serialization mechanism to the intrusive or
+non-intrusive version. The serialization process is recursive. Each member that
+needs to be serialized must be specified explicitly. The advantage of this
+approach is that the serialization code is written in C++ and leverages all
+necessary programming techniques. The generic, user-facing interface allows
+for effective application of the serialization process without obstructing the
+algorithms that need special code for packing and unpacking. It also allows for
+optimizations in the implementation of the archives.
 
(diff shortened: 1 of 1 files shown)


msimberg:
This looks good to me. Thanks! In this case I think adding examples would be the best documentation, but this does not need to happen in this PR.

libs/serialization/docs/index.rst
```diff
@@ -11,8 +11,18 @@
...
+
+The purest form of serializing data is to copy the content of the payload bit by bit; however, this method is impractical
+for generic C++ types, which might be composed of more than just regular built-in types. Instead, |hpx|'s approach to serialization
+is derived from the Boost Serialization library, and is geared towards allowing the programmer of a given class explicit control
+and syntax of what to serialize. It is based on operator overloading of two special archive types that hold a buffer or stream
+to store the serialized data and is responsible for dispatching the serialization mechanism to the intrusive or non-intrusive version.
+The serialization process is recursive. Each member that needs to be serialized must be specified explicitly. The advantage of this approach
+is that the serialization code is written in C++ and leverages all necessary programming techniques. The generic, user-facing interface allows
+for effective application of the serialization process without obstructing the algorithms that need special code for packing and unpacking.
+It also allows for optimizations in the implementation of the archives.
```
msimberg: The only practical request I have is if you would mind reformatting the paragraphs to be at most 80 characters wide. Some editors can do this automatically as well.

hkaiser:
Since this PR is all about adding documentation for serialization, may I suggest to look at https://www.boost.org/doc/libs/1_72_0/libs/serialization/doc/index.html to get started? All of Boost's documentation is under the BSL, so there is no need to reinvent the wheel. At least a note that the HPX serialization module was designed to be (mostly) interface compatible with the Boost counterpart and a link to their documentation should be added.

rstobaugh:
> This looks good to me. Thanks! In this case I think adding examples would be the best documentation, but this does not need to happen in this PR.

libs/serialization/docs/index.rst
rstobaugh: @msimberg  Will do. You also mentioned that I should include examples of serialization. Do you have any references you would recommend to help me find or build examples?

msimberg:
@rstobaugh did you close this on purpose? As for the examples I don't think we have anything specific in place at the moment but we could reuse the actual implementations we have. For example https://github.com/STEllAR-GROUP/hpx/blob/master/libs/serialization/include/hpx/serialization/complex.hpp just to show how to write the `serialize`/`load`/`save` functions.

rstobaugh:
@msimberg Sorry, I must have hit the "close and comment" button instead of "comment"; I've reopened it

msimberg:
> I can't get my hyperlinks to API references to work; can anyone tell me what I'm doing wrong with this one?

@rstobaugh do you mean this one: https://github.com/STEllAR-GROUP/hpx/pull/4358/files#diff-c8874fed7487477687d7f84b646e05beR39? It should already be correct (it works in our generated documentation: https://stellar-group.github.io/hpx-docs/branches/master/html/libs/serialization/docs/index.html). Could I ask you to try a clean build of the documentation (delete the build directory and configure with CMake again)?

rstobaugh:
@msimberg Yes, that's the one I'm referring to. I'll have the others show me how to do a clean build of the documentation.

msimberg:
@rstobaugh I think we can go ahead with the content of this PR. Examples can come in a separate PR. Would you mind trying to rebasing on master and squashing the commits? Let me know if you need any help with that.

PR #4360, opened 2020-02-03
Move plugins

aurianer:
Move some of the plugins headers inside the top level `plugins/` directory


msimberg:
I suggest we deprecate the `hpx/include/compression*.hpp` headers together with this, or we create a meta-target `hpx_compression` which adds the common headers and links to whichever compression plugins are available.

msimberg:
I think this looks pretty good now.

I have a general question about the compression and coalescing plugins: If using them requires changes in user code they're not quite plugins, are they? If the header says HPX was built with compression but the shared library is not there the application won't work (unlike just gracefully not use the plugin like I would expect a plugin to work)? Am I thinking about them the wrong way?

aurianer:
> I think this looks pretty good now.
> 
> I have a general question about the compression and coalescing plugins: If using them requires changes in user code they're not quite plugins, are they? If the header says HPX was built with compression but the shared library is not there the application won't work (unlike just gracefully not use the plugin like I would expect a plugin to work)? Am I thinking about them the wrong way? @hkaiser, @sithhell?

Any updates on this ?

PR #4363, opened 2020-02-03
Update clang-format config file for version 9

msimberg:
Goes together with https://github.com/STEllAR-GROUP/docker_build_env/pull/19. Brace wrapping after case labels can be kept as it is with a new option but I don't see any obvious options to avoid the other formatting changes. They look like something that should've been applied with clang-format-8 anyway.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,7 +127,7 @@ jobs:
           name: Check that the modules are well clang-formatted
           command: |
               cd /hpx/source/libs && shopt -s globstar # to activate the ** globbing
-              clang-format-8 -i **/*.{cpp,hpp}
+              clang-format -i **/*.{cpp,hpp}
               git diff --exit-code > /tmp/modified_format_files.txt
       - store_artifacts:
           path: /tmp/modified_format_files.txt
--- a/.clang-format
+++ b/.clang-format
@@ -42,6 +42,7 @@ AlwaysBreakTemplateDeclarations: true
 BinPackArguments: true
 BinPackParameters: true
 BraceWrapping:
+    AfterCaseLabel: true
     AfterClass: true
     AfterControlStatement: true
     AfterEnum: true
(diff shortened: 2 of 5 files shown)


aurianer:
Thanks!

biddisco:
Question : Other projects have had trouble with clang-format when rules change from one version to another (and are no longer supported or incompatible or have different effects) - do we have confidence that clang format rules and version support are now stable enough that we can upgrade at will, or should we stick with clang-format-x for everyone (until we know better?).

msimberg:
> Question : Other projects have had trouble with clang-format when rules change from one version to another (and are no longer supported or incompatible or have different effects) - do we have confidence that clang format rules and version support are now stable enough that we can upgrade at will, or should we stick with clang-format-x for everyone (until we know better?).

I'd say we stick with a version until we explicitly update the version, as in this case. The update from 8 to 9 lead to one type of formatting change that could be covered by updating the config file, and a couple of other minor changes that I considered small enough to warrant an update. For the time being we require clang-format-9 even after clang-format-10 is out.

biddisco:
> For the time being (i.e. with this PR) we require clang-format-9 even after clang-format-10 is out.
Correct me if I'm wrong, but doesn't the CI use clang-format-8 currently? (I installed that on my system because the errors I saw on CircleCI had clang-format-8 mentioned somewhere).

msimberg:
> > For the time being (i.e. with this PR) we require clang-format-9 even after clang-format-10 is out.
> 
> Correct me if I'm wrong, but doesn't the CI use clang-format-8 currently? (I installed that on my system because the errors I saw on CircleCI had clang-format-8 mentioned somewhere).

It uses 9 now (I updated the docker image to Ubuntu 19.10 which comes with clang 9 and associated tools, and most importantly a newer CMake). A stricter clang-tidy-9 is why the CircleCI builds are failing at the moment.

msimberg:
I forgot to change `clang-format-8` in the CircleCI config file... I've just pushed that. If CircleCI agrees with the formatting changes I'll merge this.

PR #4365, opened 2020-02-03
Removing more Boost dependencies

hkaiser:
- make use of `boost::exception` (and related facilities) conditional
- replace `boost::counting_iterator`
- replace `boost::io::ios_flags_saver`
- flyby: fully remove the use of `boost::exception` where not needed
- flyby: HPX_WITH_DYNAMIC_HPX_MAIN is not supported on windows

Working towards #3440


hpx/runtime/parcelset/locality.hpp
```diff
@@ -1,6 +1,7 @@
 //  Copyright (c)      2014 Thomas Heller
 //  Copyright (c) 2007-2017 Hartmut Kaiser
 //  Copyright (c) 2007 Richard D. Guidry Jr.
+//  Copyright 2002, 2005 Daryle Walker
```
K-ballo: This looks out of place.

hkaiser: Yah, that's not right. Thanks for spotting.

hkaiser:
I believe the errors on gcc-oldest are unrelated to this PR. It is ready to be reviewed now.

CMakeLists.txt
```diff
@@ -288,7 +288,7 @@ endif()
 hpx_option(HPX_WITH_DYNAMIC_HPX_MAIN BOOL
   "Enable dynamic overload of system ``main()`` (Linux only, default: ON)"
   ON ADVANCED)
-if(HPX_WITH_DYNAMIC_HPX_MAIN)
+if(HPX_WITH_DYNAMIC_HPX_MAIN AND (NOT MSVC))
```
msimberg: Should we also make it a hard configuration error if `HPX_WITH_DYNAMIC_HPX_MAIN=ON` on Windows? There might be quite a few others as well in which case we can do this separately?

hkaiser: Let's do this separately. I agree we should turn non-applicable options into hard configuration errors.

PR #4366, opened 2020-02-04
Suppress or fix clang-tidy-9 warnings

msimberg:
Attempting to fix new warnings after updating from 8 to 9.

### DIFF
--- a/hpx/runtime/threads/policies/shared_priority_queue_scheduler.hpp
+++ b/hpx/runtime/threads/policies/shared_priority_queue_scheduler.hpp
@@ -339,7 +339,7 @@ namespace hpx { namespace threads { namespace policies {
                             // pool - we can schedule on any thread available
                             thread_num =
                                 numa_holder_[0].thread_queue(0)->worker_next(
-                                    static_cast<unsigned int>(num_workers_));
+                                    static_cast<std::size_t>(num_workers_));
                         }
                         else if (!round_robin_) /* thread parent */
                         {
@@ -368,7 +368,7 @@ namespace hpx { namespace threads { namespace policies {
                                 numa_holder_[domain_num]
                                     .thread_queue(
                                         static_cast<std::uint16_t>(q_index))
-                                    ->worker_next(static_cast<unsigned int>(
+                                    ->worker_next(static_cast<std::size_t>(
                                         num_workers_));
                         }
                         thread_num = select_active_pu(l, thread_num);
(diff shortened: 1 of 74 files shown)


hpx/runtime/threads/policies/shared_priority_queue_scheduler.hpp
```diff
@@ -487,9 +487,9 @@ namespace hpx { namespace threads { namespace policies {
                     // High priority tasks first
                     else if (steal_hp_first_)
                     {
-                        for (std::uint16_t d = 0; d < num_domains_; ++d)
+                        for (std::size_t d = 0; d < num_domains_; ++d)
                         {
-                            std::uint16_t dom =
+                            std::size_t dom =
```
hkaiser: I think with this change all of the `static_cast`'s can go away now as well.

hpx/util/thread_aware_timer.hpp
```diff
@@ -50,7 +50,7 @@ namespace hpx { namespace util
 
         std::int64_t elapsed_microseconds() const
         {
-            return std::int64_t((take_time_stamp() - start_time_) * 1e-3);
+            return std::int64_t((take_time_stamp() - start_time_) * 1000);
```
hkaiser: This change is not correct. The code should divide by `1000`, not multiply.

msimberg: Excellent catch, thanks!

msimberg:
Btw, I've suppressed quite a few warnings that to me look like false positives. If you see something that actually seems like a bug please let me know and I'll try to fix it.

msimberg:
Related to the docker image upgrade, the `python` command isn't available anymore (only `python3`). We have a few Python scripts which explicitly use `/usr/bin/env python` which now fail on CircleCI. We can:

- change `python` to `python3`,
- symlink `python` to `python3` in the docker image, or
- generate the shebang line in those scripts like we do for scripts that get installed.

My preference is in the order they're listed. Opinions?

hkaiser:
Since python V2 is being phased out everywhere, I think we are safe to rename the command to python3. I don't know however if this would be correct on all platforms. Option 3 might be the safest way forward.

msimberg:
> Since python V2 is being phased out everywhere, I think we are safe to rename the command to python3. I don't know however if this would be correct on all platforms. Option 3 might be the safest way forward.

That's my thinking as well. I see that we actually install many of the Python scripts. I'll generate them with the Python that we find during configuration. Scripts that aren't installed I'll leave with `python3`.

This is finally passing `clang-tidy` now. The spell checker is complaining about some changes that I missed earlier. I'll leave those out of this PR and look at them separately (or try to ignore them if possible).

msimberg:
I unfortunately don't have the time to update the scripts now until Monday. If someone wants to push to this branch feel free to do so so that we can get builds working for everyone, or we just merge it like this and I change the python shebangs separately next week.

PR #4368, opened 2020-02-06
Add a whitelist of words for weird spelling suggestions

aurianer:
Add a codespell whitelist to ignore some spelling suggestions

Hint : The words to ignore should be separated with a comma


aurianer:
I rebased onto #4366 so that the core build is not stopped

aurianer:
I think this is good to go

msimberg:
Thanks @aurianer for taking care of this!

PR #4370, opened 2020-02-07
Doc updates quick start

rstobaugh:
Grammar edits to documentation.

### DIFF
--- a/docs/sphinx/quickstart.rst
+++ b/docs/sphinx/quickstart.rst
@@ -41,14 +41,14 @@ It is also recommended that you check out the latest stable tag:
 The minimum dependencies needed to use |hpx| are |boost|_ and |hwloc|_. If these
 are not available through your system package manager, see
 :ref:`boost_installation` and :ref:`hwloc_installation` for instructions on how
-to build them yourself. In addition to |boost|_ and |hwloc|_, it is recommended
+to build them yourself. In addition to |boost| and |hwloc|, it is recommended
 that you don't use the system allocator, but instead use either ``tcmalloc``
 from |google_perftools|_ (default) or |jemalloc|_ for better performance. If you
-would like to try |hpx| without a custom allocator at this point you can
+would like to try |hpx| without a custom allocator at this point, you can
 configure |hpx| to use the system allocator in the next step.
 
 A full list of required and optional dependencies, including recommended
-versions is available at :ref:`prerequisites`.
+versions, is available at :ref:`prerequisites`.
 
 Building |hpx|
 ==============
@@ -65,9 +65,9 @@ known to |cmake|, the following gets you started:
     make install
 
 This will build the core |hpx| libraries and examples, and install them to your
-chosen location. If you want to install |hpx| to system folders simply leave out
+chosen location. If you want to install |hpx| to system folders, simply leave out
 the ``CMAKE_INSTALL_PREFIX`` option. This may take a while. To speed up the
-process launch more jobs by passing the ``-jN`` option to ``make``.
+process, launch more jobs by passing the ``-jN`` option to ``make``.
 
 .. tip::
 
@@ -76,13 +76,13 @@ process launch more jobs by passing the ``-jN`` option to ``make``.
 
 .. tip::
 
-   If you want to change |cmake|_ variables for your build it is usually a good
+   If you want to change |cmake| variables for your build, it is usually a good
    idea to start with a clean build directory to avoid configuration problems.
    It is especially important that you use a clean build directory when changing
    between ``Release`` and ``Debug`` modes.
 
-If your dependencies are in custom locations you may need to tell |cmake|_ where
-to find them by passing one or more of the following options to |cmake|_:
+If your dependencies are in custom locations, you may need to tell |cmake| where
+to find them by passing one or more of the following options to |cmake|:
 
 .. code-block:: sh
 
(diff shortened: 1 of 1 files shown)


docs/sphinx/quickstart.rst
```diff
@@ -41,21 +41,21 @@ It is also recommended that you check out the latest stable tag:
 The minimum dependencies needed to use |hpx| are |boost|_ and |hwloc|_. If these
 are not available through your system package manager, see
 :ref:`boost_installation` and :ref:`hwloc_installation` for instructions on how
-to build them yourself. In addition to |boost|_ and |hwloc|_, it is recommended
+to build them yourself. In addition to Boost and HWLOC, it is recommended
```
sithhell: This change needs to be reverted. `|boost|_` and `|hwloc|_` are macros that get expanded.

docs/sphinx/quickstart.rst
```diff
@@ -41,21 +41,21 @@ It is also recommended that you check out the latest stable tag:
...
-versions is available at :ref:`prerequisites`.
+versions, is available at :ref:`prerequisites`.
 
 Building |hpx|
 ==============
 
 Once you have the source code and the dependencies, set up a separate build
 directory and configure the project. Assuming all your dependencies are in paths
-known to |cmake|, the following gets you started:
+known to CMake, the following gets you started:
```
sithhell: same here.

docs/sphinx/quickstart.rst
```diff
@@ -76,13 +76,13 @@ process launch more jobs by passing the ``-jN`` option to ``make``.
 
 .. tip::
 
-   If you want to change |cmake|_ variables for your build it is usually a good
+   If you want to change CMake variables for your build, it is usually a good
```
sithhell: and here.

docs/sphinx/quickstart.rst
```diff
@@ -76,13 +76,13 @@ process launch more jobs by passing the ``-jN`` option to ``make``.
...
    idea to start with a clean build directory to avoid configuration problems.
    It is especially important that you use a clean build directory when changing
    between ``Release`` and ``Debug`` modes.
 
-If your dependencies are in custom locations you may need to tell |cmake|_ where
-to find them by passing one or more of the following options to |cmake|_:
+If your dependencies are in custom locations, you may need to tell CMake where
+to find them by passing one or more of the following options to CMake:
```
sithhell: and the two lines above.

docs/sphinx/quickstart.rst
```diff
@@ -92,22 +92,22 @@ to find them by passing one or more of the following options to |cmake|_:
     -DJEMALLOC_ROOT=/path/to/jemalloc
 
 If you want to try |hpx| without using a custom allocator pass
-``-DHPX_WITH_MALLOC=system`` to |cmake|_.
+``-DHPX_WITH_MALLOC=system`` to CMake.
```
sithhell: and here

docs/sphinx/quickstart.rst
```diff
@@ -92,22 +92,22 @@ to find them by passing one or more of the following options to |cmake|_:
...
 
 .. important::
 
-   If you are building |hpx| for a system with more than 64 processing units you
-   must change the |cmake| variables ``HPX_WITH_MORE_THAN_64_THREADS`` (to
+   If you are building |hpx| for a system with more than 64 processing units, you
+   must change the CMake variables ``HPX_WITH_MORE_THAN_64_THREADS`` (to
```
sithhell: and here.

docs/sphinx/quickstart.rst
```diff
@@ -146,14 +146,14 @@ the distributed aspects of |hpx|.
       ... examples.quickstart.hello_world_distributed
       ... hello_world_distributed
 
-   It is also possible to build e.g. all quickstart examples using ``make
+   It is also possible to build all quickstart examples using ``make
```
sithhell: I think the `e.g.` serves a purpose here. The intention is, to show that one can build only a subset, not just everything (aka the `all` target). There are many more targets which can be displayed with `make help`, or `ninja -t targets`

docs/sphinx/quickstart.rst
```diff
@@ -173,7 +173,7 @@ build an executable using |cmake|_ and |hpx|:
 .. note::
 
    ``COMPONENT_DEPENDENCIES iostreams`` is optional for a minimal project but
-   lets us use the |hpx| equivalent of ``std::cout``, i.e. the |hpx|
+   lets users use the |hpx| equivalent of ``std::cout``, i.e., the |hpx|
```
sithhell: maybe `lets you` instead? Throughout the rest of text, we address the reader with "you"

docs/sphinx/quickstart.rst
msimberg: I think one of the style changes @rstobaugh converged on during GSoD is to only have a link at the first occurrence of something like `Boost` in a section or page (I'm in favor of this!). In that case we can use the form `|boost|` (i.e. no underscore at the end) which does the replacement to the correct capitalization (`Boost`) but does not produce a link.

rstobaugh: @sithhell as msimberg mentioned, prior to GSoD, there was no consistency to when hyperlinks were used. The guideline we decided on was that external links would only be used once per page (unless directing the reader to a specific download) and internal links would always be used for relevant terms.

docs/sphinx/quickstart.rst
rstobaugh: @sithhell The e.g. doesn't flow well grammatically speaking, so I'll look into rewording the sentence to sound better but preserve the meaning. Thanks!

docs/sphinx/quickstart.rst
msimberg: @rstobaugh I may not have been very clear about it, but I think we should still keep `|boost|` and `|hwloc|` (without the underscore) so that we get consistent capitalization more or less automatically.

docs/sphinx/quickstart.rst
```diff
@@ -2,7 +2,7 @@
     Copyright (C) 2018 Mikael Simberg
 
     SPDX-License-Identifier: BSL-1.0
-    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    Distributed under the |boost| Software License, Version 1.0. (See accompanying
```
msimberg, line 5: This one should stay as `Boost` as it's not actually rendered in the documentation.
```suggestion
    Distributed under the Boost Software License, Version 1.0. (See accompanying
```

PR #4374, opened 2020-02-10
Adding overload for `hpx::init`/`hpx::start` for use with resource partitioner

hkaiser:
- this helps resolving linker errors on Windows if the main entry point to HPX is in a shared library
- this also cleans up the API as it makes the use of the resource partitioner explicit

Note that this is a breaking API as the use of the no-argument versions of `init` and `start` is now deprecated when using an explicit resource partitioner.

### DIFF
--- a/hpx/hpx_init.hpp
+++ b/hpx/hpx_init.hpp
@@ -780,6 +780,40 @@ namespace hpx
         hpx::runtime_mode mode = hpx::runtime_mode_default);
 
 /// \cond NOINTERNAL
+    namespace resource {
+
+        // forward declaration only
+        class partitioner;
+    }
+/// \endcond
+
+    /// \brief Main entry point for launching the HPX runtime system.
+    ///
+    /// This is a simplified main entry point, which can be used to set up the
+    /// runtime for an HPX application. All initialization parameters for the
+    /// runtime are taken from the resource partitioner object provided.
+    ///
+    /// \param rp           [in] The resource partitioner object to use for
+    ///                     initializing the runtime.
+    /// \param startup      [in] A function to be executed inside a HPX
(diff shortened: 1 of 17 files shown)


msimberg:
Thanks @hkaiser for starting this. @aurianer has another PR almost ready to go which actually passes a callback to `init`/`start` to create thread pools etc. with the resource partitioner. This PR still allows initializing the rp explicitly outside of `init`/`start` and I was actually hoping to disallow that, but we could go this way as well. The critical change that I wanted to have is to not have the rp grab things like parcelport configs directly but have them passed to it by `init`/`start`. Is this for a specific use case of yours or related to the discussions we had earlier on cleaning up use of the rp with `init`/`start`?

hkaiser:
Thanks @msimberg. This change is to avoid a linker error on Windows under certain circumstances. The problem I'm trying to solve is that in cases were the resource-manager is initialized explicitly (with an explicit HPX-main function) the no-argument overload of `hpx::init` still passes `hpx_main` as a symbol to the runtime, causing linker errors in applications. This new overload exposes a separate entry point function that works around the issue.

This PR also makes passing the resource-partitioner to the runtime explicit. I never liked the implicit 'pseudo-global' resource partitioner that was snug in behind the scenes.

I have no issue to achieve the two goals in any different way, however.

msimberg:
Thanks @hkaiser. I suggest we go with this first (and if you can add a small test demonstrating the linking problems that would be great). We'll adapt @aurianer's changes to fit this.

aurianer:
Thanks !

msimberg:
I already forgot about the test when I merged this. Let's figure out a test on IRC.

PR #4378, opened 2020-02-12
Change C++ standard handling

msimberg:
Playing around with using the standard CMake ways of handling C++ standards flags.

The way I've set it up at the moment is to discourage setting any standard explicitly which means that HPX would by default be built in C++14 mode. This is to allow dependent projects to use any standard from 14 upwards. If HPX is built with 17 dependent applications can't be built with 14.

The standard can still be set explicitly with `CMAKE_CXX_STANDARD` and `HPX_USE_CMAKE_CXX_STANDARD=ON` (`OFF` by default, leads to error to discourage setting `CMAKE_CXX_STANDARD`). The required standard is propagated to dependent projects using the `cxx_std_ab` compile feature.

The `HPX_WITH_CXXAB` options are also still there but produce warnings when used.

To do:
- [x] Make sure C++ standard flag is used in feature tests.


msimberg:
Opening this up to hear suggestions and possible problems with this approach. The goal is to use as much vanilla CMake as possible and avoid handling things by ourselves if not necessary.

aurianer:
LGTM thanks!

msimberg:
In the process I've attempted to (somewhat) clean up the handling of compiler flags. Let me know if you'd prefer to have these in a separate PR (the changes are in separate commits). I've kept the `hpx_add*flags/definitions` macros but set the flags and definitions to the interface targets `hpx_private_flags` (previously `hpx_internal_flags`) and `hpx_public_flags`. HPX targets should link privately and publicly, respectively, to those targets.

msimberg:
The `cxx_std_ab` features require CMake 3.8 by the way, but I've bumped it up in #4246 so I'll leave the minimum unchanged in this PR.

msimberg:
This should be ready to go together with #4246. Any remaining comments?

PR #4380, opened 2020-02-12
Add a helper function to get a future from a cuda stream

msimberg:
Adds a free function to get a future from a stream (instead of requiring a `hpx::compute::cuda::target`).

@G-071

@biddisco I didn't get to creating a module yet, but can add that as well.

### DIFF
--- a/libs/compute_cuda/include/hpx/compute/cuda/target.hpp
+++ b/libs/compute_cuda/include/hpx/compute/cuda/target.hpp
@@ -33,6 +33,11 @@
 
 namespace hpx { namespace compute { namespace cuda {
     ///////////////////////////////////////////////////////////////////////////
+    namespace detail {
+        HPX_API_EXPORT hpx::future<void> get_future(cudaStream_t);
+    }
+
+    ///////////////////////////////////////////////////////////////////////////
     struct target
     {
     public:
@@ -204,6 +209,7 @@ namespace hpx { namespace compute { namespace cuda {
         hpx::id_type locality_;
     };
 
+    using detail::get_future;
     HPX_API_EXPORT target& get_default_target();
 }}}    // namespace hpx::compute::cuda
 
--- a/libs/compute_cuda/include/hpx/compute/cuda/target.hpp
+++ b/libs/compute_cuda/include/hpx/compute/cuda/target.hpp
@@ -13,17 +13,20 @@
 #include <hpx/config.hpp>
 
 #if defined(HPX_HAVE_CUDA)
+#include <hpx/allocator_support/allocator_deleter.hpp>
 #include <hpx/compute/cuda/get_targets.hpp>
 #include <hpx/lcos/future.hpp>
 #include <hpx/runtime/find_here.hpp>
 #include <hpx/runtime/runtime_fwd.hpp>
 #include <hpx/synchronization/spinlock.hpp>
+#include <hpx/traits/future_access.hpp>
 
 #include <hpx/serialization/serialization_fwd.hpp>
 
 #include <cuda_runtime.h>
 
 #include <cstddef>
+#include <memory>
 #include <mutex>
 #include <string>
 #include <utility>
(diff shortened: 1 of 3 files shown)


msimberg:
Please wait with merging this btw. It seems like I messed things up with an untested change. I should probably add a test as well.

msimberg:
> I'll give it a shot and you can clean up my mess!

@hkaiser if you have the time please have a go at it. I'm not productively making sense of what needs to be added...

hkaiser:
@msimberg please recheck, I have not actually compiled my commit (I don't have CUDA readily set up here right now).

msimberg:
> @msimberg please recheck, I have not actually compiled my commit (I don't have CUDA readily set up here right now). And yes, you were right. You had to derive your shared state from `future_data_alloc`, sorry for saying otherwise.

Thanks for the commit! I made it compile (some things needed to be moved to the header file) but I'm getting "data has already been set for this future". Does it look like I butchered your changes? I don't quite understand what code determines if the allocated shared state will be ready or not. The code looks pretty much the same as for `make_ready_future_alloc`.

libs/compute_cuda/include/hpx/compute/cuda/target.hpp
```diff
@@ -32,6 +35,149 @@
...
+        {
+            using init_no_addref =
+                typename lcos::detail::future_data_allocator<void,
+                    Allocator>::init_no_addref;
+            using other_allocator = typename std::allocator_traits<
+                Allocator>::template rebind_alloc<future_data>;
+
+            static void CUDART_CB stream_callback(
+                cudaStream_t stream, cudaError_t error, void* user_data)
+            {
```
hkaiser: This function could be HPX_EXPORT'ed and moved to the .cpp file

libs/compute_cuda/include/hpx/compute/cuda/target.hpp
```diff
@@ -32,6 +35,149 @@
...
+            future_data(init_no_addref no_addref, other_allocator const& alloc,
+                cudaStream_t stream)
+              : lcos::detail::future_data_allocator<void, Allocator>(
+                    no_addref, alloc)
+              , rt_(hpx::get_runtime_ptr())
+            {
+                init(stream);
+            }
+
+            void init(cudaStream_t stream)
```
hkaiser: This function could be HPX_EXPORT'ed and moved to the .cpp file

libs/compute_cuda/include/hpx/compute/cuda/target.hpp
```diff
@@ -32,6 +35,149 @@
...
+            }
+
+            ~release_on_exit()
+            {
+                // release the shared state
+                lcos::detail::intrusive_ptr_release(data_);
+            }
+
+            future_data<Allocator>* data_;
+        };
```
hkaiser: If we moved the functions below to the implementation file, then the two classes above could go there as well.

libs/compute_cuda/include/hpx/compute/cuda/target.hpp
msimberg: Wait, can it really? It accesses the `rt_` member from `future_data` which means that it's effectively also templated on `Allocator`. What am I missing?

msimberg:
This seems to work well after your fixes @hkaiser. Thanks! I've rebased so let's check that pycicle also agrees that things are good.

biddisco:
If this is merged, I will rebase my cuda branch onto it afterwards as this code will move into the cuda_support module and there will be conflicts.

libs/compute_cuda/include/hpx/compute/cuda/target.hpp
hkaiser: It accesses the member through the pointer it receives. The whole function does not depend on the template argument and could be moved outside of the class.

libs/compute_cuda/include/hpx/compute/cuda/target.hpp
hkaiser: This function can't be HPX_EXPORT'ed and moved to the implementation file after all. It's a non-static member function of a template.

libs/compute_cuda/include/hpx/compute/cuda/target.hpp
hkaiser: The first class can be moved unchanged (after the function `stream_callback` was moved), the second class requires some changes in order to be moved however. Instead of referring to `future_data<Allocator>` it could be referring to the base class `future_data_refcnt_base` (https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/lcos/detail/future_data.hpp#L69).

libs/compute_cuda/include/hpx/compute/cuda/target.hpp
msimberg: Right, it can be moved outside the class (it's static after all) but not into the cpp file. It still needs to know to what it should cast the `void*` so it would still need to be templated on `Allocator` (or on the full `future_data` type). Am I still missing something?

hkaiser: Yes, you're certainly right. Sorry for the confusion.

msimberg:
https://cdash.cscs.ch//testDetails.php?test=18391475&build=97142 is new so please don't merge this. I'll try to look into it.

msimberg:
> https://cdash.cscs.ch//testDetails.php?test=18391475&build=97142 is new so please don't merge this. I'll try to look into it.

Apparently it's not only on this branch. I'll have to investigate if we missed it in some PR along the way or if something changed on Piz Daint but we can go ahead with this PR if you're happy with it @hkaiser (I am).

PR #4381, opened 2020-02-13
Fix configuration with CUDA and older CMake versions

msimberg:
Newer CMake (3.15 tested) accepted the old code, but older ones (3.10 tested) complain rightfully that `hpx_base_libraries` is not built by the dependent project when calling `find_package(HPX)`.

I've already added this change to the 1.4.1 release.


aurianer:
Thanks!

PR #4382, opened 2020-02-13
Update spellchecking whitelist and fix more typos

msimberg:
This fixes more typos and expands the whitelist. I've also changed the CircleCI config to not require `spellcheck`, `inspect`, and `docs` for the `install` step (as they don't affect functionality).


examples/transpose/transpose_serial.cpp
```diff
@@ -152,8 +152,8 @@ int main(int argc, char* argv[])
         ( "verbose", "Verbose output")
     ;
 
-    // Initialize and run HPX, this example is serial and therefor only needs on thread,
-    // We just use hpx::init to parse our command line arguments
+    // Initialize and run HPX, this example is serial and therefore only needs
+    // on thread, We just use hpx::init to parse our command line arguments
```
hkaiser: Should it be `one thread` instead?

examples/transpose/transpose_serial_vector.cpp
```diff
@@ -157,8 +157,8 @@ int main(int argc, char* argv[])
         ( "verbose", "Verbose output")
     ;
 
-    // Initialize and run HPX, this example is serial and therefor only needs on thread,
-    // We just use hpx::init to parse our command line arguments
+    // Initialize and run HPX, this example is serial and therefore only needs
+    // on thread, We just use hpx::init to parse our command line arguments
```
hkaiser: Same here: `one thread`.

examples/transpose/transpose_serial.cpp
msimberg: It helps to read around the typo as well... thanks!

aurianer:
Thanks!

PR #4383, opened 2020-02-13
Use configure_file for HPXCacheVariables.cmake

msimberg:
Mainly just to get rid of the message `Installing: .../HPXCacheVariables.cmake` when configuring HPX.


msimberg:
The failures look Piz Daint related, but this can wait for another round of builds.

aurianer:
Thanks!

PR #4385, opened 2020-02-13
Cuda futures

biddisco:
This moves the most basic support for cuda stream handling and setting of futures into a small library called `cuda_support`.

To enable this library, one must use `HPX_WITH_CUDA_SUPPORT=ON` and it will work even when `HPX_WITH_CUDA=OFF`

The motivation for this is to allow applications that use cuda, but do not need any `hpx::` algorithms making use of cuda, and the user has provided their own kernels for all device code. 

This permits the connecting of user kernels via futures obtained from `hpx::cuda::`to arbitrary `hpx` code. 

I would like to go further and rename the main `HPX_WITH_CUDA` to `HPX_WITH_CUDA_COMPUTE` to force the user to choose between basic support and full hpx algorithms in cuda.

This PR includes #4380


cmake/HPX_AddExecutable.cmake
```diff
@@ -140,7 +140,7 @@ function(add_hpx_executable name)
     endforeach()
   endif()
 
-  if(HPX_WITH_CUDA AND NOT HPX_WITH_CUDA_CLANG)
+  if((HPX_WITH_CUDA OR HPX_WITH_CUDA_SUPPORT) AND NOT HPX_WITH_CUDA_CLANG)
```
hkaiser: What's the difference between `HPX_WITH_CUDA` and `HPX_WITH_CUDA_SUPPORT` ?

libs/CMakeLists.txt
```diff
@@ -90,7 +91,7 @@ if(HPX_WITH_TESTS)
   endif()
 
   if (HPX_WITH_TESTS_HEADERS)
-    add_hpx_pseudo_target(tests.headers.modules)
+    add_custom_target(tests.headers.modules)
```
hkaiser: This will break things on Windows. Please stick to using `add_hpx_pseudo_target`.

libs/compute_cuda/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 cmake_minimum_required(VERSION 3.3.2 FATAL_ERROR)
+message("Inside compute_cuda")
```
hkaiser: Can we remove this message, please?

libs/compute_cuda/include/hpx/compute/cuda.hpp
```diff
@@ -9,12 +9,12 @@
 #ifndef HPX_COMPUTE_CUDA_HPP
 #define HPX_COMPUTE_CUDA_HPP
 
+#include <hpx/cuda_support/get_targets.hpp>
+#include <hpx/cuda_support/target.hpp>
```
hkaiser: This will cause the clang-format-check to complain. Please clang-format the module files.

hkaiser: Ahh, I take this back - this file is not part of a module (yet), so things should be good.

libs/CMakeLists.txt
biddisco: I thought I had only moved code around. If I introduced something new and bad in the cmake, I'd better fix it ...

libs/compute_cuda/CMakeLists.txt
biddisco: I thought I already had. Sorry.

libs/compute_cuda/include/hpx/compute/cuda/detail/launch.hpp
```diff
@@ -85,8 +85,8 @@ namespace hpx { namespace compute { namespace cuda { namespace detail {
         }
 
         template <typename DimType>
-        HPX_HOST_DEVICE static void call(target const& tgt, DimType gridDim,
-            DimType blockDim, fun_type f, args_type args)
+        HPX_HOST_DEVICE static void call(hpx::cuda::target const& tgt,
+            DimType gridDim, DimType blockDim, fun_type f, args_type args)
```
hkaiser: While this is not introduced by this PR: could we please change the names of the arguments to conform to our naming conventions (all `lower_case_names`) while we touch this anyways?

libs/cuda_support/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) 2019 The STE||AR-Group
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+cmake_minimum_required(VERSION 3.3.2 FATAL_ERROR)
+
+if (NOT TARGET hpx::cuda)
+  message(FATAL_ERROR "Need cuda")
```
hkaiser: Please use `hpx_error()` instead of `message()`, also please make the error message less terse. Perhaps something along the lines of 
> In order to use the CUDA support module, please enable CUDA support while configuring HPX (add `-DHPX_WITH_CUDA=ON` to the cmake command line).

libs/cuda_support/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+endif()
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+
+# Default location is $HPX_ROOT/libs/cuda_support/include
+set(cuda_support_headers
+  hpx/cuda_support/cuda_future_helper.hpp
```
hkaiser: Please add all header files that are part of this module here.

libs/cuda_support/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+
+# Default location is $HPX_ROOT/libs/cuda_support/include
+set(cuda_support_headers
+  hpx/cuda_support/cuda_future_helper.hpp  
+)
+
+# Default location is $HPX_ROOT/libs/cuda_support/include_compatibility
+set(cuda_support_compat_headers)
```
hkaiser: Please generate forwarding headers to retain compatibility with existing code and add those headers here.

libs/cuda_support/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+
+set(cuda_support_sources
+  cuda_target.cpp
+  get_cuda_targets.cpp
+)
+
+include(HPX_AddModule)
+add_hpx_module(cuda_support
+  COMPATIBILITY_HEADERS OFF
```
hkaiser: This should be set to `ON`

libs/cuda_support/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+    hpx_config
+    hpx_errors
+    hpx_execution
+    hpx_functional
+    hpx_memory
+    hpx_program_options
+    hpx_synchronization
+    hpx_static_reinit
+    hpx_type_support
+    ${CUDA_LIBRARIES} ${CUDA_CUBLAS_LIBRARIES}
```
hkaiser: Shouldn't this depend on `hpx::cuda` instead?

libs/cuda_support/include/hpx/cuda_support/cuda_future_helper.hpp
```diff
@@ -51,7 +50,7 @@
 //        auto fut = helper.async(&call_trivial_kernel, 3.1415);
 //
 // -------------------------------------------------------------------------
-namespace hpx { namespace compute { namespace util
+namespace hpx { namespace cuda
```
hkaiser: Please add definitions to the forwarding headers that ensure backwards compatibility of the symbols defined in here.

libs/cuda_support/include/hpx/cuda_support/get_targets.hpp
```diff
@@ -11,19 +11,19 @@
 
 #include <hpx/config.hpp>
 
-#if defined(HPX_HAVE_CUDA)
+//#if defined(HPX_HAVE_CUDA)
```
hkaiser: Is this not needed anymore?

libs/cuda_support/tests/unit/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+if (NOT HPX_WITH_CUDA_SUPPORT)
+  return()
+endif()
+
+#enable_language(CUDA)
```
hkaiser: Is this not needed anymore?

libs/cuda_support/tests/unit/cuda_future.cpp
```diff
@@ -14,8 +14,9 @@
 #include <hpx/include/parallel_for_each.hpp>
 #include <hpx/include/parallel_for_loop.hpp>
 //
-#include <hpx/compute/cuda/target.hpp>
-#include <hpx/include/compute.hpp>
+#include <hpx/cuda_support/target.hpp>
+#include <hpx/cuda_support/cuda_future_helper.hpp>
+//#include <hpx/include/compute.hpp>
```
hkaiser: Is this header not needed anymore?

hkaiser:
> To enable this library, one must use HPX_WITH_CUDA_SUPPORT=ON and it will work even when HPX_WITH_CUDA=OFF

Could you please elaborate on your rationale for this? Could we just enable this module with the usual `HPX_WITH_CUDA` instead? Eventually, the user will decide what to use from HPX by just #including the modules that are needed,

msimberg:
> > To enable this library, one must use HPX_WITH_CUDA_SUPPORT=ON and it will work even when HPX_WITH_CUDA=OFF
> 
> Could you please elaborate on your rationale for this? Could we just enable this module with the usual `HPX_WITH_CUDA` instead? Eventually, the user will decide what to use from HPX by just #including the modules that are needed,

Just my 2 cents: While I like the idea of being able to turn on and off individual modules, I'm a bit wary to do it too early in an ad-hoc way before we have all modules in place. As I've stated before I'd prefer to have only coarser grained options exposed to the user *or* it would have to be possible to turn on and off any module with the same mechanism (and this comes with the additional complexities of having to check if a disabled module should be enabled because another enabled module depends on it). I think the difference between building only `cuda_support` or `compute_cuda` on top of that is close to zero. The parallel algorithms are all header only and don't make a difference when compiling the main library.

libs/cuda_support/CMakeLists.txt
biddisco: That was left in by mistake. I'll improve that

biddisco:
The reason for creating this module is simply to allow cuda futures, without having to build HPX with cuda enabled. Currently, cuda requires c++14 (and not 17 or 20), and is hard to compile on some machines and architectures. This module breaks the current cuda support into a minimal `cuda_support` that does not need to compile any hpx code at all using nvcc, and allows you to use c++17 or any other model and separately link your kernels to hpx and get futures from their launches - and the full `hpx::compute` that is not well maintained and has had issues with compilation on diverse machines over a period of many years. I would like the beauty of cuda+futures - without the pain of compiling all of hpx with cuda.
This is a precursor for moving cuda offloading to another library such as kokkos, sycl, other.

The old `HPX_WITH_CUDA` can(should) in principle be replaced with `HPX_WITH_CUDA_COMPUTE` - but I didn't go that far.

cmake/HPX_AddExecutable.cmake
biddisco: > What's the difference between HPX_WITH_CUDA and HPX_WITH_CUDA_SUPPORT ?

The` difference is that `HPX_WITH_CUDA` implies full `hpx::compute::cuda` currently, whereas the new CUDA_SUPPORT is a much smaller subset that does not pull in `hpx::compute::cuda` and instead only supplies user facing converters to get a future from a stream or stream event. `HPX_WITH_CUDA_SUPPORT`. `HPX_WITH_CUDA` requires that hpx be compiled with the same compiler restrictions that apply to cuda code (such as c++!4 currently), whereas basic support for futures does not.

libs/CMakeLists.txt
biddisco: this appear to have been introduced by a change in the library generator code in `components/create_component_skeleton.py` - I have removed it from CMakeLists in this PR and will push shortly.

hkaiser: Could you please fix the generator script in this case as well?

hkaiser:
> The reason for creating this module is simply to allow cuda futures, without having to build HPX with cuda enabled. Currently, cuda requires c++14 (and not 17 or 20), and is hard to compile on some machines and architectures. This module breaks the current cuda support into a minimal cuda_support that does not need to compile any hpx code at all using nvcc, and allows you to use c++17 or any other model and separately link your kernels to hpx and get futures from their launches - and the full hpx::compute that is not well maintained and has had issues with compilation on diverse machines over a period of many years. I would like the beauty of cuda+futures - without the pain of compiling all of hpx with cuda.

Ok, understood. Wouldn't it be better in this case to change the build system such that we don't build all of HPX using NVCC. There is no real reason to do that. 

I agree that this PR might be a step in the right direction of deprecating the exiting compute parts. However, I'd like to see the `HPX_WITH_CUDA` setting to be renamed to `HPX_WITH_CUDA_COMPUTE` and a new `HPX_WITH_CUDA` being introduced that simply enables both, `HPX_WITH_CUDA_COMPUTE` and `HPX_WITH_CUDA_SUPPORT`. Oppinions?

msimberg:
Does `libhpx` library actually get compiled with `nvcc` when it's enabled? Or do we just link `libhpx` to the CUDA libraries, and the examples and tests get compiled with `nvcc`? Wouldn't it be a bad idea to build HPX with C++17 and then build kernels with C++14, unless the only thing you include in `cu` files is the kernel itself and don't include anything from HPX? No inline lambdas as kernels?

In any caseI support of cleaning up CUDA support. We should consider using the CUDA language feature (see top of https://cmake.org/cmake/help/latest/module/FindCUDA.html), and https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#module:FindCUDAToolkit. I'm not sure how that works with clang CUDA though.

biddisco:
> ...

Sorry. I didn't mean that nvcc compiles all of hpx, but if a user does a `cmake -DHPX_WITH_CUDA=ON` with testing enabled, followed by a `make install`, then all the tests and  examples that use cuda and require `hpx::compute::cuda` will be built. This step has been difficult for many years and here have been a ton of problems getting all of hpx 'safe' for use with compute::cuda as quite a lot of the core of HPX gets pulled into the nvcc compilation with cuda::compute. This is the part that goes away with this new CMake setting.

In this PR, If `HPX_WITH_CUDA=ON` then nothing changes, the new cuda_support module is built as well as the old compute_cuda module. All that changes is that you can now say `HPX_WITH_CUDA=OFF` but separately enable `HPX_WITH_CUDA_SUPPORT=ON` to get basic cuda support.

> However, I'd like to see the HPX_WITH_CUDA setting to be renamed to HPX_WITH_CUDA_COMPUTE and a new HPX_WITH_CUDA being introduced that simply enables both, HPX_WITH_CUDA_COMPUTE and HPX_WITH_CUDA_SUPPORT. Opinions?

That's basically what I wanted to do, but I stopped slightly short. I will make the change and add some docs and to explain the difference between full cuda support and basic.

biddisco:
@msimberg can you confirm if this is a mistake. All the if/else clauses use pseudo target except the last headers block (and this is what caused the stray entry in this library creation commit).
https://github.com/STEllAR-GROUP/hpx/blob/master/components/create_component_skeleton.py#L191-L199

If that's a mistake and not intentional, I will fix it in this PR

msimberg:
> @msimberg can you confirm if this is a mistake. All the if/else clauses use pseudo target except the last headers block (and this is what caused the stray entry in this library creation commit).
> https://github.com/STEllAR-GROUP/hpx/blob/master/components/create_component_skeleton.py#L191-L199
> 
> If that's a mistake and not intentional, I will fix it in this PR

It definitely looks like a mistake (and seems to be confirmed by @hkaiser's comment). The same goes for `libs/create_library_skeleton.py`.

biddisco:
Accidentally closed this by deleting the wrong branch on my clone. Reopening.

biddisco:
I've fixed all the cmake issues, added compatibility headers, sorted out the python lib generator wscript and made cleanups all around. I think this addresses all the comments made in the PR.

biddisco:
PS. I tried to cherry pick the allocator patch from the other cuda branch, but it didn't compile and I didn't want to spend time on it now. I will try again later if nobody else does it.

msimberg:
> PS. I tried to cherry pick the allocator patch from the other cuda branch, but it didn't compile and I didn't want to spend time on it now. I will try again later if nobody else does it.

You can just un-rebase on that PR (i.e. drop those commits). Merging those changes once this is in master is not a big deal (there'll most likely be no conflicts).

biddisco:
I will rebase this onto master now that #4380 has been merged as it will require some changes.

biddisco:
I've rebased this onto master and verified that all cuda related tests pass - (except for the transform_compute_test that is failing on dashboards already).

libs/cuda_support/docs/index.rst
```diff
@@ -0,0 +1,60 @@
...
+Kernels and CPU work may be freely intermixed.
+
+It is important to note that multiple kernels may be launched
+without fetching a future, and multiple futures may be obtained
+from the helper. Please referr to the unit tests and examples
+for more information.
+
+============
+CMake variables
+============
```
msimberg:
```suggestion
=================
|cmake| variables
=================

```

libs/cuda_support/docs/index.rst
```diff
@@ -0,0 +1,60 @@
...
+HPX_WITH_CUDA_SUPPORT=ON will enable the building of this module which requires
+only the presence of CUDA on the system and only exposes cuda+fuures support.
+
+The CMake setting for HPX_WITH_CUDA=ON (or HPX_WITH_CUDA_COMPUTE=ON) enable more
+sophisticated CUDA based integration using hpx::compute::cuda and allow the user
+to execute parallel algorithms using hpx:: syntax on cuda enabled resources.
+
+
```
msimberg:
```suggestion
```

libs/cuda_support/docs/index.rst
```diff
@@ -0,0 +1,60 @@
...
+It is important to note that multiple kernels may be launched
+without fetching a future, and multiple futures may be obtained
+from the helper. Please referr to the unit tests and examples
+for more information.
+
+============
+CMake variables
+============
+HPX_WITH_CUDA_SUPPORT=ON will enable the building of this module which requires
```
msimberg:
```suggestion
``HPX_WITH_CUDA_SUPPORT=ON`` will enable the building of this module which requires
```
and the same for other CMake variables.

libs/cuda_support/docs/index.rst
```diff
@@ -0,0 +1,60 @@
...
+only the presence of CUDA on the system and only exposes cuda+fuures support.
+
+The CMake setting for HPX_WITH_CUDA=ON (or HPX_WITH_CUDA_COMPUTE=ON) enable more
```
msimberg:
```suggestion
The |cmake| setting for HPX_WITH_CUDA=ON (or HPX_WITH_CUDA_COMPUTE=ON) enable more
```

libs/cuda_support/include/hpx/cuda_support/cuda_future_helper.hpp
```diff
@@ -4,14 +4,13 @@
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-#ifndef CUDA_HELPER_H
-#define CUDA_HELPER_H
+#ifndef HPX_CUDA_SUPPORT_CUDA_FUTURE_HELPER_HPP
+#define HPX_CUDA_SUPPORT_CUDA_FUTURE_HELPER_HPP
 
-#define BOOST_NO_CXX11_ALLOCATOR
+// #define BOOST_NO_CXX11_ALLOCATOR
```
msimberg:
```suggestion
```
?

libs/cuda_support/include_compatibility/hpx/compute/cuda/target.hpp
```diff
@@ -0,0 +1,26 @@
...
+#else
+#warning                                                   \
+    "The header hpx/compute/cuda/target.hpp is deprecated, \
+    please include hpx/cuda_support/target.hpp instead"
+#endif
+#endif
+
+namespace hpx { namespace compute { namespace cuda {
+  using target = hpx::cuda::target;
+}}}
```
msimberg: Please move this to `hpx/cuda_support/target.hpp`. The namespace change should in principle also get a compatibility option. How many people's codes do we think we'll break with this (probably everyone using CUDA support)?

libs/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ include(HPX_Message)
 include(HPX_AddPseudoDependencies)
 include(HPX_AddPseudoTarget)
 
-set(HPX_LIBS
+set(HPX_CANDIDATE_LIBS
```
hkaiser, line 13: Are we sure that `HPX_LIBS` is not used anywhere outside of HPX to refer to the list of modules (APEX)?

libs/cuda_support/include_compatibility/hpx/compute/cuda/target.hpp
hkaiser: I think we're fine with having this here. If user code will be broken after we disable the header compatibility we can deal with it separately.

biddisco:
Just a warning that I am going to rebase this onto latest master and cleanup the unfixed items, so beware if anyone else is working on it.

hkaiser:
Orthogonal suggestion: I would like to rename the existing `mpi` (futures) module into `async_mpi` and the new module proposed here into `async_cuda` for consistency purposes. That could be done in a separate PR, however.

libs/cuda_support/CMakeLists.txt
biddisco: fixed. Note that it was changed in b84e13c789b2a7ed0f7e9c79f1776917f9c2c132 to Cuda::cuda

libs/cuda_support/tests/unit/CMakeLists.txt
biddisco: It wasn't there originally. I added it then commented it out. Gone now.

libs/cuda_support/tests/unit/cuda_future.cpp
biddisco: removed

CMakeLists.txt
```diff
@@ -393,12 +393,26 @@ hpx_option(HPX_WITH_FULL_RPATH BOOL
...
+hpx_option(HPX_WITH_CUDA_SUPPORT BOOL
+  "Enable basic support for CUDA futures (default: OFF)" OFF ADVANCED)
+if(HPX_WITH_CUDA)
+  set(HPX_WITH_CUDA_COMPUTE ON CACHE BOOL "Enable full hpx compute capability with cuda backend" FORCE)
+  set(HPX_WITH_CUDA_SUPPORT ON CACHE BOOL "Enable generation of hpx futures from cuda invocations" FORCE)
+endif()
+if (HPX_WITH_CUDA_COMPUTE)
+  set(HPX_WITH_CUDA ON CACHE BOOL "Enable cuda compute backend as well as cuda/hpx futures" FORCE)
+else()
+  set(HPX_WITH_CUDA OFF CACHE BOOL "Disable hpc compute cuda backend" FORCE)
```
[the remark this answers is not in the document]
biddisco: fixed

CMakeLists.txt
```diff
@@ -393,12 +393,26 @@ hpx_option(HPX_WITH_FULL_RPATH BOOL
...
 hpx_option(HPX_WITH_CUDA_CLANG BOOL
   "Use clang to compile CUDA code (default: OFF)" OFF ADVANCED)
+hpx_option(HPX_WITH_CUDA_COMPUTE BOOL
+  "Enable CUDA compute module (default: OFF) - "
+  "note: enabling this also enables HPX_WITH_CUDA_SUPPORT" OFF ADVANCED)
+hpx_option(HPX_WITH_CUDA_SUPPORT BOOL
+  "Enable basic support for CUDA futures (default: OFF)" OFF ADVANCED)
+if(HPX_WITH_CUDA)
+  set(HPX_WITH_CUDA_COMPUTE ON CACHE BOOL "Enable full hpx compute capability with cuda backend" FORCE)
+  set(HPX_WITH_CUDA_SUPPORT ON CACHE BOOL "Enable generation of hpx futures from cuda invocations" FORCE)
```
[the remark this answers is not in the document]
biddisco: Not sure I understand the question

biddisco:
Passed through all the comments and rebased/force pushed quite a lot of tweaks. Almost all comments addressed I think.

Agree that `async_mpi` and `async_cuda` would be good names and should go into new PR. 
Also `cuda_future_helper` should be upgraded to be executor based like the mpi version.

biddisco:
I was pushing to stellar repo - but this PR was from my clone. Pushed again - Should update now.

CMakeLists.txt
```diff
@@ -375,12 +375,26 @@ hpx_option(HPX_WITH_FULL_RPATH BOOL
...
+hpx_option(HPX_WITH_CUDA_SUPPORT BOOL
+  "Enable basic support for CUDA futures (default: OFF)" OFF ADVANCED)
+if(HPX_WITH_CUDA)
+  set(HPX_WITH_CUDA_COMPUTE ON CACHE BOOL "Enable full HPX compute capability with CUDA backend" FORCE)
+  set(HPX_WITH_CUDA_SUPPORT ON CACHE BOOL "Enable generation of HPX futures from CUDA invocations" FORCE)
+endif()
+if (HPX_WITH_CUDA_COMPUTE)
+  set(HPX_WITH_CUDA ON CACHE BOOL "Enable CUDA compute backend as well as CUDA/HPX futures" FORCE)
+else()
+  set(HPX_WITH_CUDA OFF CACHE BOOL "Disable HPX compute CUDA backend and support" FORCE)
```
msimberg, line 396: The description should say what happens when you turn the option on, and the description should be the on line 381, 394, and 396.

The same goes for the two other options.

CMakeLists.txt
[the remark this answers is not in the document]
msimberg: Sorry, badly worded. It was about the description of the option, but I like what you have there now (except for the inconsistencies...). Please ignore.

CMakeLists.txt
msimberg: Thanks!

msimberg:
@biddisco would you mind marking the comments as resolved if you've fixed them (it's easier for you as you know what you've changed)? Also, could you check why the builds fail on pycicle now?

libs/cuda_support/docs/index.rst
```diff
@@ -0,0 +1,64 @@
...
+    f.then([](hpx::future<void>&& f) {
+        std::cout << "trivial kernel completed \n";
+    }).get();
+
+Kernels and CPU work may be freely intermixed/overlapped
+and synchronized with futures.
+
+It is important to note that multiple kernels may be launched
+without fetching a future, and multiple futures may be obtained
+from the helper. Please referr to the unit tests and examples
```
msimberg, line 46:
```suggestion
from the helper. Please refer to the unit tests and examples
```

biddisco:
I've fixed the conflicts and outstanding items locally, but I'm waiting till the cmake-format branch is merged because I tested it locally and reformatted my cmake stuff. After that's merged, I'll push my latest changes, to avoid pointless conflicts here ...

msimberg:
@biddisco the CMake formatting branch has been merged now. You can rebase on master again.

biddisco:
I have rebased onto current master and updated this PR with new improvements as follows

- cuda helper is now replaced by cuda_executor and cublas_executor
- exceptions are now handled during launch and if a failed cuda/cublas call occurs, the future will be set with an exception state
- cudaEvent_t is now supported. Instead of triggering cuda futures using a callback on a stream (on an NVidia thread), a cudaEvent can be recorded and a new polling method that uses cudaEventQuery checks for completion of cuda functions. The machinery is modelled after the `mpi::future` support recently added, but is somewhat simpler, see message on commit for more info.

This is being pushed for sharing and testing purposes, but some work on the unit tests might still be necessary before a merge is considered. Internally, we are using a mixture of event based and callback based futures in the tests and they should be cleaned up and a small benchmark comparing the two approaches would be useful. A flag passed to the executor constructor to switch between the two types is necessary and will be added in the next patch.

hkaiser:
> A flag passed to the executor constructor to switch between the two types is necessary and will be added in the next patch.

Could those be different executors? I would assume that the implementation for those would be quite different.

biddisco:
> > A flag passed to the executor constructor to switch between the two types is necessary and will be added in the next patch.
> 
> Could those be different executors? I would assume that the implementation for those would be quite different.

Option 1 is have `cuda_executor_callback` and `cuda_executor_polling` - but in actual fact, internally, I've modified the `future_data` object to have two constructors and they are invoked using a tag from the call to `get_future` that the executor calls. So there is really no difference except a simple toggle, one constructor doesn't set the callback, and instead passes an event that can be queried to the internal polling code, the other constructor, does install the callback, but doesn't register an event with the polling code
Option 2 is just have `cuda_executor(target/stream, flag=polling_by_default, callback_if_you_want_it)`

option 2 seems to me to be easier from the point of view of the programmer as we can easily toggle between the two modes in examples and benchmarks with a simple flag/tag. the code needs a bit of cleaning up as I realized I could have simplified a couple of the calls by forwarding the tag I'm using in a few places instead of having multiple `get_future_xxx` methods. I would like to celan these up and put in place option 2 as I do it.

Ultimately, two executors, or one executor with an optional parameter makes little difference to me as they will both end up in the same place internally.

biddisco:
results from new benchmark two runs - 1 thread, and 4 threads on 4 core laptop
```
bin/cuda_executor_throughput_test --iterations=500000 --hpx:threads=1

[HPX CUBLAS executor benchmark] - Starting...
GPU Device 0: "GeForce GTX 1060" with compute capability 6
MatrixA(4,4), MatrixB(4,4), MatrixC(4,4)
Async host->device copy operation completed
Small matrix multiply tests using CUBLAS...

us per iteration 28.94 : Warmup
us per iteration 35.5693 : Callback based executor
us per iteration 7.90474 : Event polling based executor
```
```
bin/cuda_executor_throughput_test --iterations=500000 --hpx:threads=4
[HPX CUBLAS executor benchmark] - Starting...
GPU Device 0: "GeForce GTX 1060" with compute capability 6
MatrixA(4,4), MatrixB(4,4), MatrixC(4,4)
Async host->device copy operation completed
Small matrix multiply tests using CUBLAS...

us per iteration 26.57 : Warmup
us per iteration 30.9399 : Callback based executor
us per iteration 9.25825 : Event polling based executor
```

biddisco:
I believe that the functionality in the new polling method for cuda futures is now quite good and we can keep the callback method for as long as it is not removed by nvidia (note that the reason for adding the polling is partly because on this page [nvidia cudaAddCallback docs](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM_1g74aa9f4b1c2f12d994bf13876a5a2498) it states that "This function is slated for eventual deprecation and removal." so we need the polling support eventually.

Just a heads up: the branch is now essentially good to go, but I will add one more commit in the next day or two - which is to rename the module to "async_cuda" so that the merge of this branch can make a full and clean transition from compute::cuda to the new module.

Issues around the handling of streams between hpx threads can be saved for after this is merged and we can test it on some full application prototypes. 
I have begun work on  #4507 to allow us to install polling hooks for both mpi and cuda in a clean and extensible way, but I would like to see cuda stuff merged, mpi stuff merged and the polling hooks as a new PR.

libs/cuda_support/include/hpx/cuda_support/cublas_executor.hpp
```diff
@@ -0,0 +1,252 @@
...
+    }    // namespace detail
+
+    // -------------------------------------------------------------------------
+    // a simple cublas wrapper helper object that can be used to synchronize
+    // cublas calls with an hpx future.
+    // -------------------------------------------------------------------------
+    struct cublas_executor : cuda_executor
+    {
+        // construct a cublas stream
+        cublas_executor(std::size_t device, bool event_mode = false)
```
teonnik, line 117: A parameter for the pointer mode should also be passed to the executor. https://docs.nvidia.com/cuda/cublas/index.html#scalar-parameters 

The default for most CUBLAS calls is `CUBLAS_POINTER_MODE_HOST`, if the CUBLAS function has an OUT parameter, the mode will actually synchronize with the stream on the thread the task is running, circumventing the `cudaEventQuery()` mechanism and blocking the thread on which the task is running. For such routines, it is good to have the option to change the mode to `CUBLAS_POINTER_MODE_DEVICE`.

libs/cuda_support/include/hpx/cuda_support/cublas_executor.hpp
```diff
@@ -0,0 +1,252 @@
...
+          : hpx::cuda::cuda_executor(device, event_mode)
+        {
+            detail::cub_debug.debug(
+                debug::str<>("cublas_executor"), "event mode", event_mode);
+
+            handle_ = 0;
+            check_cublas_error(cublasCreate(&handle_));
```
teonnik, line 124: CUBLAS handles are an expensive resource, their destruction may cause implicit device synchronization. Have you perhaps considered making a pool of CUBLAS handles from which cublas_executor can get it's CUBLAS handles without creating and destroying them?

https://docs.nvidia.com/cuda/cublas/index.html#cublas-context

libs/cuda_support/include/hpx/cuda_support/cuda_executor.hpp
```diff
@@ -0,0 +1,206 @@
...
+        cuda_executor_base operator=(const cuda_executor_base& other) = delete;
+
+        // -------------------------------------------------------------------------
+        // constructors - create a cuda stream that all tasks invoked by
+        // this helper will use
+        // assume event mode is the default
+        cuda_executor_base(std::size_t device, bool event_mode)
+          : device_(device)
+          , event_mode_(event_mode)
+          , target_(device)
```
teonnik, line 87: `target_` seems to be an abstraction over a `cudaStream_t`. This is similar to my question for the `cublas_executor`, there is some overhead with creating and destroying cudaStream_t as discussed at PyTorch : https://github.com/pytorch/pytorch/issues/9646 . Have you considered making a pool of cudaStream_t which the executor can get a stream from instead?

libs/cuda_support/include/hpx/cuda_support/cuda_future.hpp
```diff
@@ -0,0 +1,343 @@
...
+            void init_callback(cudaStream_t stream)
+            {
+                // Hold on to the shared state on behalf of the cuda runtime
+                // right away as the callback could be called immediately
+                // and it happens on another thread which might cause a race
+                lcos::detail::intrusive_ptr_add_ref(this);
+
+                cudaError_t error =
+                    cudaStreamAddCallback(stream, stream_callback, this, 0);
+                if (error != cudaSuccess)
```
teonnik, line 159: Since `cudaStreamAddCallback()` is deprecated, have you considered using `cudaLaunchHostFunc` instead?

libs/cuda_support/include/hpx/cuda_support/cublas_executor.hpp
biddisco: Thanks. I will add the pointer mode flag to the executor.

libs/cuda_support/include/hpx/cuda_support/cublas_executor.hpp
biddisco: I ad assumed that the cublas executor would be reasonably long lived - ideally created at startup, but the issue of managing many streams via many executors (and many cublas handles) has not yet been addressed.
I suggest we merge this PR without a handle pool and add support for that later. I've got a pool for the events in place and it ought to be straightforward to add an other pool, but it's worth it's own PR I think.

libs/cuda_support/include/hpx/cuda_support/cublas_executor.hpp
hkaiser: I would suggest not to use a `bool` but the explicit CUDA constants. I also think there should be a default for that parameter (whatever is more commonly used).

libs/cuda_support/include/hpx/cuda_support/cuda_executor.hpp
biddisco: Same response really. In octotiger, a collection of (say) 128 streams are created during an initialization step and then when needed, they are handed out in sequentially so that work can be pushed into the GPU and it can be oversubscribed - when all streams are 'busy' - further work can be blocked or the cpu can be given tasks. The streams can be on a single GPU, or divided over multiple ones and synchronization between streams can either use futures or cuda stream semantics - depending on the workflow.

Something like this can be implemented using the targets, or executors as handles onto a preallocated set of streams/handles. `targets` are a hold over from `hpx::compute` and they don't really buy us anything useful, but for this implementation I though it simpler to reuse them rather than getting rid of them and reimplementing a new abstraction over streams. The concept of a `target` is quite nice though. We can/should work on efficient management of streams/handles for the next PR.

libs/cuda_support/include/hpx/cuda_support/cuda_future.hpp
biddisco: Very good idea. I will try that.

libs/cuda_support/include/hpx/cuda_support/cublas_executor.hpp
biddisco: oops. I had a default in there originally, but I removed it when debugging and forgot to put it back. The bool param was an enum originally, but since there were only two, I made them `true_type` and `false_type` - would you prefer an enum?

biddisco: I misunderstood the comments. I had tag type for the event_mode/callback_mode - then switched it explicity to true_type/false_type - not for the pointer mode - please ignore my comment above.

The pointer mode in DLAF is being set on every cublas call - do you think that is actually necessary? Are you not in fact using the same mode for the whole application and could just set it once? do you think it is likely that you'll have it device mode in one executor and host mode in another?

biddisco:
Added pointer_mode to the cublas executor, cleaned up a few minor string uses, made sure clang-format and inspect are passing locally and then renamed the whole library from cuda_support to `async_cuda` 
I would suggest that any improvements to this such as pools, mutexes or changes to the callback function are made in one or more separate PRs otherwise this will never get merged. 

if there's any new feedback on this PR then I'll make the appropriate fixes.

libs/cuda_support/include/hpx/cuda_support/cublas_executor.hpp
teonnik: Since your handle is created and destroyed within the executor it can be set once when the executor is created. The mode depends on the function you are calling, a function like cublas<t>dot() with CUBLAS_POINTER_MODE_HOST will do a `cudaStreamSynchronize()` because with that pointer mode, the OUT parameter (`float *result`) is expected to be available on the CPU when the call completes. With CUBLAS_POINTER_MODE_DEVICE, the OUT parameter is instead on the GPU and no synchronization happens. This section of the docs discusses it : https://docs.nvidia.com/cuda/cublas/index.html#scalar-parameters

> do you think it is likely that you'll have it device mode in one executor and host mode in another?

Yes because I might be schedule a `cublas<t>gemm()` with host mode on one (there are no scalar OUT parameters) and `cublas<t>dot()` with device mode (where there is a scalar OUT parameter, i.e. the final value of the dot product).

libs/cuda_support/include/hpx/cuda_support/cublas_executor.hpp
teonnik: My only concern with managing the lifetime of the CUBLAS handle within the executor is when passing it by value within a closure of a lambda:

```
some_fut.then([cublas_ex, ...] {
    return async(cublas_ex, ....).get() # some cuda call
});
```

this might incur some unwanted synchronizations and may be quite expensive. If the executor is long-lived, perhaps then it would be better to just pass a reference within the closure. What do you think?

biddisco: I thought I  had made the executor non-copyable to prevent that. maybe I accidentally removed that code. The idea is to make it move only and use it by reference and keep it alive until no longer needed.

biddisco: The executors are non copyable, so your lambda would be disallowed. I hope that is satisfactory (for now)

libs/cuda_support/include/hpx/cuda_support/cuda_future.hpp
biddisco: Marking this as resolved as I will look into it in the next PR ...

libs/async_cuda/include/hpx/async_cuda/cuda_future.hpp
```diff
@@ -0,0 +1,345 @@
...
+#include <hpx/config.hpp>
+#include <hpx/async_cuda/cuda_event.hpp>
+#include <hpx/async_cuda/cuda_exception.hpp>
+#include <hpx/basic_execution.hpp>
+#include <hpx/concurrency.hpp>
+#include <hpx/debugging/print.hpp>
+#include <hpx/futures/future.hpp>
+#include <hpx/memory.hpp>
+#include <hpx/runtime_local/thread_pool_helpers.hpp>
+#include <hpx/threading_base.hpp>
```
msimberg, line 20: These need to be updated (https://cdash.cscs.ch//viewBuildError.php?buildid=110625)
```suggestion
#include <hpx/config.hpp>
#include <hpx/async_cuda/cuda_event.hpp>
#include <hpx/async_cuda/cuda_exception.hpp>
#include <hpx/modules/basic_execution.hpp>
#include <hpx/modules/concurrency.hpp>
#include <hpx/debugging/print.hpp>
#include <hpx/futures/future.hpp>
#include <hpx/modules/memory.hpp>
#include <hpx/runtime_local/thread_pool_helpers.hpp>
#include <hpx/modules/threading_base.hpp>
```

libs/async_cuda/tests/unit/cublas_matmul.cpp.orig
```diff
@@ -27,6 +27,7 @@
```
msimberg, line 27: This file shouldn't be here.

msimberg:
The `hpx/include/compute.file` needs to be updated: https://cdash.cscs.ch//viewBuildError.php?buildid=110584. Most likely we should have a compatibility header for `hpx/compute/cuda.hpp` as well (possibly others).

libs/async_cuda/tests/unit/cublas_matmul.cpp.orig
biddisco: aaargggghh.

msimberg:
@hkaiser do you have any final *blocking* comments on this one? There are clearly things that still need to be cleaned up, but it looks like we can leave those for a follow-up PR. It would be good to get this into master asap so that we can figure out what, if anything, this breaks for users before the release.

hkaiser:
> do you have any final blocking comments on this one? 

Let's go ahead as is and solve out problems as we go. Let's have everything in `namespace experimental`, however.

msimberg:
> > do you have any final blocking comments on this one?
> 
> Let's go ahead as is and sort out problems as we go. Let's have everything in `namespace experimental`, however.

In this PR? So far we haven't done any renamings of existing things into `hpx::experimental` (the MPI stuff is new). I'm thinking it may be easiest to do these renamings as the last thing before 2.0.0 simply because we don't need to worry about compatibility at that point anymore. On the other hand I'm all for getting these renamings out of the way as well.

biddisco:
I'm happy to put this stuff into `experimental` namespace (I actually did that originally, but then decided that since the functionality was just the cuda futures helper code moved around, it didn't need to be), however the whole polling for events infrastructure is new and putting the whole lot into experimental makes sense (symmetry with the `mpi::future` stuff)
The only project really using these features is octotiger and they made a copy of the cude-future-helper anyway, so moving the namespace affects nobody (and there's really no point in having compatibility headers anyway as noone is using it)

biddisco:
I am getting fed up with having to fix conflicts caused by all the other PRs being merged. Every time I fix the conflicts, I have to do two new builds, make sure no namespace errors creep in,  run all the cuda and compute tests to make sure they're ok and it takes several hours of time. This has been going on for months now.

msimberg:
> I am getting fed up with having to fix conflicts caused by all the other PRs being merged. Every time I fix the conflicts, I have to do two new builds, make sure no namespace errors creep in, run all the cuda and compute tests to make sure they're ok and it takes several hours of time. This has been going on for months now.

I understand. This time it was a mistake on my part. I actually wanted this in first but accidentally merged another PR with conflicts with this. I'll happily take care of the conflicts on this PR if you'd like.

msimberg:
I'll merge this now to avoid further conflicts. I've opened https://github.com/STEllAR-GROUP/hpx/issues/4766 for the namespace change. I'll leave it up to you @biddisco to decide when to do the change.

PR #4386, opened 2020-02-13
Add warning messages for using thread binding options on macOS

### LINKED ISSUE #4349: `--hpx:use-process-mask` option throw an exception on OS X
## Expected Behavior

I'm aware of the impossibility to do CPU binding on OS X, but usually requesting a binding on this kind of platform does not result in an error.

## Actual Behavior

The following exception is thrown
`hpx::init: hpx::exception caught: hwloc_get_cpubind failed: HPX(kernel_error)`

Specifying the `--hpx:use-process-mask` option results in the exception.

No special code is needed to test it, even just the following one is enough:
```cpp
#include <hpx/hpx_init.hpp>

int hpx_main(int argc, char **argv) {
  return hpx::finalize();
}

int main(int argc, char **argv) {
  auto ret_code = hpx::init(hpx_main, argc, argv);

  return ret_code;
}
```

## Specifications

  - HPX Version: 1.4.0
  - Platform (compiler, OS): clang on OSX

msimberg:
Attempt to fix #4349. @albestro would you mind testing the changes?

### DIFF
--- a/src/util/command_line_handling.cpp
+++ b/src/util/command_line_handling.cpp
@@ -816,6 +816,17 @@ namespace hpx { namespace util
                  vm.count("hpx:ignore-batch-env")) == 0) &&
             !use_process_mask_;
 
+#if defined(__APPLE__)
+        if (use_process_mask_)
+        {
+            std::cerr
+                << "Warning: enabled process mask for thread binding, but "
+                   "thread binding is not supported on macOS. Ignoring option."
+                << std::endl;
+            use_process_mask_ = false;
+        }
+#endif
+
         bool have_mpi = detail::check_mpi_environment(rtcfg_);
 
         util::batch_environment env(nodelist, have_mpi, debug_clp,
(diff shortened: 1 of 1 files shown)


src/util/command_line_handling.cpp
```diff
@@ -816,6 +816,14 @@ namespace hpx { namespace util
                  vm.count("hpx:ignore-batch-env")) == 0) &&
             !use_process_mask_;
 
+#if defined(__APPLE__)
+        std::cerr
+            << "Warning: enabled process mask for thread binding, but "
+               "thread binding is not supported on macOS. Ignoring option."
+            << std::endl;
+        use_process_mask_ = false;
+#end
```
albestro: Do you mean `#endif`?

albestro:
```suggestion
#endif
```

src/util/command_line_handling.cpp
```diff
@@ -816,6 +816,14 @@ namespace hpx { namespace util
                  vm.count("hpx:ignore-batch-env")) == 0) &&
             !use_process_mask_;
 
+#if defined(__APPLE__)
+        std::cerr
```
albestro: Are you printing the warning message even if the `use_process_mask` option has not been used?

src/util/command_line_handling.cpp
msimberg: Excellent catch, thanks!

src/util/command_line_handling.cpp
msimberg: Also not intentional. Thanks!

msimberg:
Thanks @albestro for testing! I've fixed the issues you found. Do the messages themselves seem clear enough for you?

albestro:
Yeah, it looks perfect!
Thank you again!

PR #4388, opened 2020-02-14
Change the resource partitioner initialization

aurianer:
Move the resource partitioner initialization inside `hpx_init`

- Add a `init_params` struct and its `hpx_init` overload to simplify initialization and redundant parameters
- Add a resource partitioner callback (in the init struct) to allow the user to specify the thread pools initialization
- Add base overload `init(argc, argv, init_params=default)`, `init(<hpx_entry_point>, argc, argv, init_params=default)`
- Change the signature of `run_or_start` to take `init_params` as an argument


hpx/hpx_init.hpp
```diff
@@ -779,39 +789,20 @@ namespace hpx
...
-    ///                     returned from the user supplied function \p f.
-    ///
-    /// \note               The created runtime system instance will be
-    ///                     executed in console or worker mode depending on the
-    ///                     configuration passed in `cfg`.
-    inline int init(resource::partitioner& rp,
-        startup_function_type startup = startup_function_type(),
-        shutdown_function_type shutdown = shutdown_function_type());
+        // Utilities to init the thread_pools of the resource partitioner
+        using rp_callback_type = std::function<void(
```
msimberg: Let's do `hpx::util::function_nonser` here.

hpx/hpx_init.hpp
```diff
@@ -779,39 +789,20 @@ namespace hpx
...
+                hpx::resource::partitioner&)>;
+        HPX_EXPORT void set_rp_callback(rp_callback_type f);
```
msimberg: This one we don't need, right?

hpx/hpx_init_impl.hpp
```diff
@@ -297,56 +387,73 @@ namespace hpx
...
     {
-        char *dummy_argv[2] = { const_cast<char*>(HPX_APPLICATION_STRING), nullptr };
+        util::function_nonser<void()> const empty;
 
-        return init(f, 1, dummy_argv, cfg, mode);
+        hpx::init_params iparams;
+        iparams.f = util::bind_back(detail::init_helper, f);
+        iparams.cfg = cfg;
+        iparams.startup = empty;
+        iparams.shutdown = empty;
```
msimberg: Aren't these empty by default? Otherwise you can also do
```suggestion
        iparams.shutdown = {};
```

hpx/hpx_init_impl.hpp
```diff
@@ -355,78 +462,78 @@ namespace hpx
...
          hpx::runtime_mode mode)
     {
-        char* dummy_argv[2] = {
-            const_cast<char*>(HPX_APPLICATION_STRING), nullptr};
+        util::function_nonser<int(hpx::program_options::variables_map& vm)>
+            main_f;
 
-        return init(nullptr, 1, dummy_argv, cfg, mode);
+        hpx::init_params iparams;
+        iparams.f = main_f;
```
msimberg: Same here, I think you can just do 
```suggestion
        iparams.f = {};
```

libs/resource_partitioner/examples/guided_pool_test.cpp
```diff
@@ -302,7 +276,41 @@ int main(int argc, char* argv[])
...
+    // Setup the init parameters
+    hpx::init_params init_args;
+    init_args.argc = argc;
+    init_args.argv = argv;
+    init_args.desc_cmdline_ptr = std::make_shared<
+        hpx::program_options::options_description>(desc_cmdline);
+    init_args.f = static_cast<hpx_main_type>(::hpx_main);
+
+    // Set the callback to init the thread_pools
+    hpx::resource::set_rp_callback(&init_resource_partitioner_handler);
```
msimberg: Let's make this a member of the struct instead.

hpx/hpx_init_params.hpp
```diff
@@ -0,0 +1,76 @@
...
+    /// \brief A struct to contain the hpx::init() parameters
+    struct init_params {
+        // Default args
+        char *dummy_argv[2] = { const_cast<char*>(HPX_APPLICATION_STRING), nullptr };
+        using options_description = hpx::program_options::options_description;
+        options_description default_desc = options_description("Usage: " HPX_APPLICATION_STRING " [options]");
+        // Parameters
+        util::function_nonser<int(hpx::program_options::variables_map& vm)> f;
+        std::shared_ptr<options_description const>
+            desc_cmdline_ptr = std::make_shared<options_description>(default_desc);
```
msimberg: Hrm, if you can do this I would expect `options_description` to actually have a copy constructor... if you have that error message again post it here :)

msimberg: Btw, if we need to keep the `shared_ptr`, `make_shared` passes its arguments to the constructor of `options_description` and you can just do
```suggestion
            desc_cmdline_ptr = std::make_shared<options_description>("Usage" + ...);
```

hpx/hpx_init.hpp
aurianer: What exactly we don't need ? ^^

hpx/hpx_init_impl.hpp
aurianer: For now, the default value is `startup_function_type()`, I'll do this change !

hpx/hpx_init.hpp
msimberg: This was connected to the other comment about passing the callback in the struct instead. We don't need `set_rp_callback` if the callback is passed in the struct.

msimberg:
>     * TODO: find a way to assign the default value of the hpx entry point (usually `hpx_main`) even in the case of `hpx_wrap` usage

I think in the case of `#include <hpx/hpx_main.hpp>` we are ok. The only difficulty with the new struct is distinguishing between users wanting to use `hpx_main` as the default entry point, users forgetting to set the entry point but wanting set it explicitly, and users setting it explicitly to `nullptr`. We can't set `hpx_main` as the default value in the struct itself or we end up with linker errors. I see few options: 

- Adding some custom `init_params` constructors to handle this. This is analogous to the `init`/`start` overloads, but the set of constructors should stay really small for there to be a benefit to this whole PR (at most overloads that take an entry point and argc/argv).
- Set the params via setters instead. We can add some custom logic to these, e.g. for `nullptr` overloads.
- Store the entrypoint as a `optional<hpx_main_type>`. `nullopt` means the user doesn't want an entrypoint. A non-`nullopt` but empty function is the default value and signals that we should try to use `hpx_main`. We can also achieve this with a separate boolean (`use_entrypoint` or something like that).
- Do something intermediate, i.e. have the following overloads (with the entrypoint and argc/argv removed from `init_params`):
```cpp
init(int, char**, init_params = {});
init(function<int(int, char**)>, int, char**, init_params = {});
init(function<int(variables_map&)>, int, char**, init_params = {});
```

Comments from a user perspective would be appreciated. Since we'd still like to keep the most commonly used overloads I quite like the last option. It's a small set of overloads but lets us handle the entrypoint as before. *Or* we get rid of the automatic handling of `hpx_main`(!). I suspect this is not a popular opinion however.

@hkaiser, @biddisco, @sithhell?

hpx/hpx_init.hpp
aurianer: Oh the using, you're right !

hpx/hpx_init_params.hpp
aurianer: yep it's on my todo list, I'll do that asap

hpx/hpx_init.hpp
msimberg: Ah, not sure if I'm misunderstanding but the `using` can stay and is good to have around. I meant that we don't need to store a static `rp_callback_type` in `hpx_init.cpp` and we don't need the `set_rp_callback` to go with it if the callback is in `init_params` instead.

aurianer: Sorry I didn't see your last comment before, yep sure !

hpx/hpx_init_impl.hpp
hkaiser: Why not have appropriate constructors defined for `hpx::init_params`?

hkaiser:
> Adding some custom init_params constructors to handle this. This is analogous to the init/start overloads, but the set of constructors should stay really small for there to be a benefit to this whole PR (at most overloads that take an entry point and argc/argv).

I'd be in favor of defining constructors. Those can define reasonable defaults for all members. For special things the members of `init_params` can always be left `public` for direct access, if needed (at least for the members that are independent of each other). The other options listed (especially the suggestion to use `optional<>`) are not too appealing to me.

aurianer:
> > Adding some custom init_params constructors to handle this. This is analogous to the init/start overloads, but the set of constructors should stay really small for there to be a benefit to this whole PR (at most overloads that take an entry point and argc/argv).
> 
> I'd be in favor of defining constructors. Those can define reasonable defaults for all members. For special things the members of `init_params` can always be left `public` for direct access, if needed (at least for the members that are independent of each other). The other options listed (especially the suggestion to use `optional<>`) are not too appealing to me.

@hkaiser, just to be I understood well, you are talking about the intermediate overloads (last point in @msimberg comment) ?

msimberg:
Just to explicitly rank the options, I prefer:

1. `hpx::init(function, argc, argv, init_params = {})` etc. This keeps the common case simple and does not require doing e.g. `hpx::init(init_params(f, argc, argv))` instead of `hpx::init(f, argc, argv)` (I'm thinking long term for when we remove deprecated overloads). The members of `init_params` already have reasonable defaults set without custom constructors (except for the entrypoint).
2. Constructors for `init_params` but only if they stay at `init_params(function, argc, argv)` and nothing more.

Don't particularly like the others.

hkaiser:
> Just to explicitly rank the options, I prefer:
> 
> hpx::init(function, argc, argv, init_params = {}) etc. This keeps the common case simple and does not require doing e.g. hpx::init(init_params(f, argc, argv)) instead of hpx::init(f, argc, argv) (I'm thinking long term for when we remove deprecated overloads). The members of init_params already have reasonable defaults set without custom constructors (except for the entrypoint).
> Constructors for init_params but only if they stay at init_params(function, argc, argv) and nothing more.
> 
> Don't particularly like the others.

I like that. I would like to have a `hpx::init(argc, argv)` overload, however.

msimberg:
> > Just to explicitly rank the options, I prefer:
> > hpx::init(function, argc, argv, init_params = {}) etc. This keeps the common case simple and does not require doing e.g. hpx::init(init_params(f, argc, argv)) instead of hpx::init(f, argc, argv) (I'm thinking long term for when we remove deprecated overloads). The members of init_params already have reasonable defaults set without custom constructors (except for the entrypoint).
> > Constructors for init_params but only if they stay at init_params(function, argc, argv) and nothing more.
> > Don't particularly like the others.
> 
> I like that. I would like to have a `hpx::init(argc, argv)` overload, however.

Yep, sorry, that was implied. I meant the overloads that I've listed in https://github.com/STEllAR-GROUP/hpx/pull/4388#issuecomment-586333742. I count the `nullptr_t` overloads to that set as well. I think that's a reasonable minimal set of overloads that wouldn't change need to change every time we want to expand `init_params`.

hpx/hpx_init_impl.hpp
```diff
@@ -84,10 +75,44 @@ namespace hpx
...
+        hpx::program_options::options_description const& desc_cmdline,
+        int argc, char** argv, std::vector<std::string> const& cfg,
+        startup_function_type startup, shutdown_function_type shutdown,
+        hpx::runtime_mode mode)
+    {
+        using hpx::program_options::options_description;
+        hpx::init_params iparams;
+        iparams.f = f;
+        iparams.desc_cmdline_ptr = std::make_shared<options_description>(
+            desc_cmdline);
```
hkaiser: Can we assume that the user-supplied `options_description` object stays alive long enough such that we can just hold on to it by reference?

hpx/hpx_init_params.hpp
```diff
@@ -0,0 +1,78 @@
...
+#ifndef DOXYGEN
+    typedef int (*hpx_main_type)(hpx::program_options::variables_map&);
+    typedef int (*hpx_user_main_type)(int argc, char** argv);
+#endif
+
+    /// \struct init_params
+    /// \brief A struct to contain the hpx::init() parameters
+    struct init_params {
+        // Default args
+        char *dummy_argv[2] = { const_cast<char*>(HPX_APPLICATION_STRING), nullptr };
```
hkaiser: I never liked this `const_cast` (it essentially invokes UB). I think we could do something like the following instead:
```
struct init_params {
    static char app_name[] = HPX_APPLICATION_STRING;
    char **argv = { app_name, nullptr };
    // ...
};
```

libs/resource_partitioner/examples/guided_pool_test.cpp
```diff
@@ -302,7 +276,40 @@ int main(int argc, char* argv[])
...
+        vm);
+
+    pool_threads = vm["pool-threads"].as<int>();
+
+    // Setup the init parameters
+    hpx::init_params init_args;
+    init_args.argc = argc;
+    init_args.argv = argv;
+    init_args.desc_cmdline_ptr = std::make_shared<
+        hpx::program_options::options_description>(desc_cmdline);
```
hkaiser: I really would like to investigate options that would avoid using a `shared_ptr` here. Also, we should use constructors on `init_params` as much as possible.

tests/unit/resource/cross_pool_injection.cpp
```diff
@@ -253,8 +250,18 @@ int main(int argc, char* argv[])
...
+    // Setup the init parameters
+    hpx::init_params init_args;
+    init_args.argc = argc;
+    init_args.argv = argv;
+    init_args.f = static_cast<hpx_main_type>(::hpx_main);
+    // set the resource partitioner callback
+    hpx::resource::set_rp_callback(&init_resource_partitioner_handler);
     // now run the test
-    HPX_TEST_EQ(hpx::init(rp), 0);
+    HPX_TEST_EQ(hpx::init(), 0);
```
hkaiser: This should probably receive the `init_params` as its argument.

libs/resource_partitioner/examples/guided_pool_test.cpp
msimberg: Yeah, the `shared_ptr` is just a temporary solution. It's not intended to stay. As for the constructors, I actually prefer leaving them out. Which ones would you define? And should they always essentially have the same signatures as the corresponding `init/start` functions? Assigning them directly like this is IMO much easier to maintain.

hkaiser: Well, I would supply argc/argv directly to `hpx::init`/`hpx::start` _only_ (i.e. not make them part of the `init_params`). Those are too common to go in there. I would prefer:
```
hpx::init_params params = {…};
hpx::init(argc, argv, params);
```
over
```
hpx::init_params params = {argc, argv, ...};
hpx::init(params);
```
any day. Probably the same for `hpx_main`. But I think we agreed on that above already.

For the other parameters, I think I could live with initializing them one by one or through compound initialization.

msimberg: Yes, I think we agree! The entrypoint and argc/argv should not be duplicated in `init_params` and in the `init` arguments. The PR doesn't 100% reflect the last discussions we've had here. Sorry for any potential confusion.

hpx/hpx_init_impl.hpp
aurianer: I wanted to use a reference a first but the options_description is not furnishing a copy constructor nor a assignment operator (but for MSVC), so I still try to figure out what to do to remove it ^^

aurianer: Any ideas?

hkaiser: Couldn't we capture the `options_descriptions` object by reference (potentially storing it as a `std::reference_wrapper<>`)?

hpx/hpx_init_impl.hpp
aurianer: Actually we don't initialize the same parameters in each overload so there would be a lot of different constructors no ?
I agree that would simplify the code inside each overload but as a user I think I would prefer initialize it like that cause it is really fast to copy paste an `iparam.<sth>` assignment and change the 2 words (pretty fast in vim)
But if you want some I can totally make some ! :)

hpx/hpx_init_impl.hpp
aurianer: Thanks a lot ! :)

msimberg:
The last step would be to introduce a compatibility option (`HPX_WITH_INIT_OVERLOADS_COMPATIBILITY`?) and hide all but the ones listed earlier behind that option.

hpx/hpx_init.hpp
```diff
@@ -14,55 +14,222 @@
...
+    ///                     `nullptr` the HPX runtime environment will be started
+    ///                     without invoking \p f.
+    /// \param argc         [in] The number of command line arguments passed
+    ///                     in \p argv. This is usually the unchanged value as
+    ///                     passed by the operating system (to `main()`).
+    /// \param argv         [in] The command line arguments for this
+    ///                     application, usually that is the value as passed
+    ///                     by the operating system (to `main()`).
+    /// \param params       [in] The parameters to the init function of type
+    ///                     `hpx::init_params`. Usually the
```
msimberg: I think we can keep this a bit shorter and just refer to `init_params` documentation.

hpx/hpx_init_impl.hpp
```diff
@@ -84,10 +81,88 @@ namespace hpx
...
+        util::function_nonser<int(hpx::program_options::variables_map&)> main_f;
+        if (argc == 0 || argv == nullptr)
+        {
+            return init(main_f, params.dummy_argc, params.dummy_argv, params);
+        }
+        return init(main_f, argc, argv, params);
+    }
+
+    /* We keep the other overloads to preserve compatibility but they will be
+    removed in the future */
```
msimberg: I think we can remove this comment if these are wrapped in `#if defined(HPX_WITH_X_COMPATIBILITY)`. Also, prefer C++ style comments (`//` instead of `/* */`).

hpx/hpx_init_impl.hpp
```diff
@@ -255,35 +343,34 @@ namespace hpx
...
+
+        hpx::init_params iparams;
+        iparams.desc_cmdline = desc_cmdline;
+        iparams.mode = mode;
 
         if (argc == 0 || argv == nullptr)
         {
-            char *dummy_argv[2] = { const_cast<char*>(app_name.c_str()), nullptr };
-            return init(desc_commandline, 1, dummy_argv, mode);
+            return init(1, iparams.dummy_argv, iparams);
```
msimberg: Does `dummy_argv` need to be in `iparams`?

hpx/hpx_init_params.hpp
```diff
@@ -0,0 +1,83 @@
+//  Copyright (c)      2020 CSCS - ETH Zurich
```
msimberg: Thanks for doing this! :D I think it should be just `ETH Zurich` and we should consider going back to change all our contributions to be `ETH Zurich` as well. Let's start with this one though...

hpx/hpx_init_impl.hpp
aurianer: Not really but I didn't know where to put it as a better place, cause if it static, it will live for the all program no ?

hpx/hpx_init_impl.hpp
aurianer: For now I just removed the comment cause a lot of examples / tests are still using those overloads, I'll create a macro in another PR I think

CMakeLists.txt
```diff
@@ -1065,20 +1065,28 @@ endif()
...
-    "Enable deprecated register_thread/work functions in the hpx::applier namespace (default: ON)"
+  "Enable deprecated register_thread/work functions in the hpx::applier namespace (default: ON)"
     ON ADVANCED)
 if(HPX_WITH_REGISTER_THREAD_COMPATIBILITY)
   hpx_add_config_define(HPX_HAVE_REGISTER_THREAD_COMPATIBILITY)
 endif()
 
+# HPX_WITH_INIT_OVERLOADS_COMPATIBILITY : introduced in V1.5.0?
+hpx_option(HPX_WITH_INIT_OVERLOADS_COMPATIBILITY BOOL
+  "Enable deprecated init() overloads functions (default: ON)"
```
msimberg:
```suggestion
  "Enable deprecated init() and start() overloads functions (default: ON)"
```
?
I wasn't sure if it would be better to make the option `HPX_WITH_INIT_START_OVERLOADS_COMPATIBILITY`. It sounded a bit like a mouthful but let's at least write in the description that it's both `init` and `start`.

hpx/hpx_init_impl.hpp
msimberg: Iether together with the other default statics or you can keep it the way it was (i.e. defined just above). If they're going to be in the struct they should at least be private and static, but I'd prefer keeping them outside.

hpx/hpx_init_impl.hpp
```diff
@@ -296,137 +381,97 @@ namespace hpx
...
-        return init(main_f, desc_commandline, argc, argv, cfg,
-            empty, empty, mode);
+        hpx::init_params iparams;
+        iparams.desc_cmdline = desc_cmdline;
+        iparams.mode = mode;
+        return init(main_f, argc, argv, iparams);
     }
 
     inline int
     init(std::nullptr_t const& f,
```
msimberg:
```suggestion
    init(std::nullptr_t f,
```
(my fault!). Could you please remove the `const&`?

hpx/hpx_init_params.hpp
```diff
@@ -0,0 +1,85 @@
...
+    /// \struct init_params
+    /// \brief A struct to contain the hpx::init() parameters
+    struct init_params {
+        char *default_argv[2] = { detail::app_name , nullptr };
+        // Parameters
+        std::reference_wrapper<hpx::program_options::options_description const>
+            desc_cmdline = detail::default_desc;
+        int dummy_argc = 1;
+        char **dummy_argv = default_argv;
+        std::vector<std::string> cfg;
```
msimberg:
```suggestion
        /// You can add member variable documentation like this.
        std::vector<std::string> cfg;
```

hpx/hpx_init_params.hpp
```diff
@@ -0,0 +1,85 @@
...
+        static shutdown_function_type default_shutdown = shutdown_function_type();
+    }
+
+#ifndef DOXYGEN
+    typedef int (*hpx_main_type)(hpx::program_options::variables_map&);
+    typedef int (*hpx_user_main_type)(int argc, char** argv);
+#endif
+
+    /// \struct init_params
+    /// \brief A struct to contain the hpx::init() parameters
```
msimberg:
```suggestion
    /// \brief Parameters used to initialize the HPX runtime through hpx::init and hpx::start.
```

libs/resource_partitioner/tests/unit/resource_partitioner_info.cpp
```diff
@@ -47,10 +47,10 @@ int main(int argc, char* argv[])
...
     std::vector<std::string> cfg = {"hpx.os_threads=4"};
 
-    // set up the resource partitioner
-    hpx::resource::partitioner rp(argc, argv, std::move(cfg));
+    hpx::init_params init_args;
+    init_args.cfg = std::move(cfg);
 
     // now run the test
-    HPX_TEST_EQ(hpx::init(rp), 0);
+    HPX_TEST_EQ(hpx::init(1, init_args.dummy_argv, init_args), 0);
```
msimberg:
```suggestion
    HPX_TEST_EQ(hpx::init(init_args), 0);
```
If we don't have this overload we should have it. It can be `hpx::init(init_params = {} /*or some reasonable default*/)` in which case it also covers `hpx::init()`.

hpx/hpx_start.hpp
```diff
@@ -22,40 +23,250 @@
...
+    ///                     executed in console or worker mode depending on the
+    ///                     command line arguments passed in `argc`/`argv`. If not
+    ///                     command line arguments are passed, console mode is
+    ///                     assumed.
+    ///
+    /// \note               If no command line arguments are passed the HPX
+    ///                     runtime system will not support any of the default
+    ///                     command line options as described in the section
+    ///                     'HPX Command Line Options'.
+    inline bool start();
```
msimberg:
```suggestion
    inline bool start(init_params const& params = {});
```

libs/resource_partitioner/examples/simplest_resource_partitioner_2.cpp
```diff
@@ -43,7 +41,14 @@ int main(int argc, char* argv[])
...
+}
+
+int main(int argc, char* argv[])
+{
+    // Set the callback to init the thread_pools
+    hpx::init_params init_args;
+    init_args.rp_callback = &init_resource_partitioner_handler;
 
-    hpx::init(rp);
+    hpx::init(argc, argv);
```
msimberg:
```suggestion
    hpx::init(argc, argv, init_args);
```

libs/resource_partitioner/tests/unit/named_pool_executor.cpp
```diff
@@ -142,8 +135,20 @@ int main(int argc, char* argv[])
...
+}
+
+// this test must be run with 4 threads
+int main(int argc, char* argv[])
+{
+    std::vector<std::string> cfg = {
+        "hpx.os_threads=" + std::to_string(max_threads)};
+
+    hpx::init_params init_args;
+    init_args.cfg = std::move(cfg);
```
msimberg:
```suggestion
    init_args.cfg = {
        "hpx.os_threads=" + std::to_string(max_threads)};
```

libs/resource_partitioner/tests/unit/used_pus.cpp
```diff
@@ -38,10 +38,10 @@ int main(int argc, char* argv[])
...
     std::vector<std::string> cfg = {"hpx.os_threads=4"};
 
-    // set up the resource partitioner
-    hpx::resource::partitioner rp(argc, argv, std::move(cfg));
+    hpx::init_params init_args;
+    init_args.cfg = std::move(cfg);
 
     // now run the test
-    HPX_TEST_EQ(hpx::init(rp), 0);
+    HPX_TEST_EQ(hpx::init(1, init_args.dummy_argv, init_args), 0);
```
msimberg: Same here:
```suggestion
    HPX_TEST_EQ(hpx::init(init_args), 0);
```

hpx/hpx_init_impl.hpp
aurianer: Actually done :)

CMakeLists.txt
aurianer: Thanks !

hpx/hpx_init.hpp
aurianer: Yep !

libs/resource_partitioner/tests/unit/resource_partitioner_info.cpp
aurianer: Yeah we have it I just forgot the parameter in there !

hpx/hpx_init_params.hpp
aurianer: Can I also put it like `\var <var_name>  <explanations>` on the very top of the struct ? Cause I feel like otherwise the code is more difficult to read

libs/resource_partitioner/examples/simplest_resource_partitioner_2.cpp
aurianer: Thanks ! I think my rebased actions messed things up sorry..

aurianer:
If it passes all tests, this should be good to go :)

aurianer:
Just changed this because of a build warning on travis-ci MacOS build:
https://github.com/STEllAR-GROUP/hpx/pull/4388/commits/a3db1e360f32abfd6580d6fc5003792ab34c3a9d#diff-6cd28becc08171d26d3f09c07c085302R326
Is that ok ?

hpx/runtime/threads/policies/shared_priority_queue_scheduler.hpp
```diff
@@ -323,7 +323,7 @@ namespace hpx { namespace threads { namespace policies {
                     {
                         debug::set(msg, "HINT_NONE  ");
                         // Create thread on this worker thread if possible
-                        if (local_num < 0)
+                        if (local_num == UINT64_MAX)
```
msimberg: Good catch! We usually do something like this:
```suggestion
                        if (local_num == std::size_t(-1))
```
especially since `size_t` isn't guaranteed to be 64 bits long everywhere.

aurianer: Thanks!

biddisco:
Quite a few problems on the dashboard that need to be fixed before this can proceed

aurianer:
I used `[[maybe_unused]]` for the unused warning to go away, any comments on this ?
https://github.com/STEllAR-GROUP/hpx/pull/4388/files#diff-271309c08d45275ceb7b63a1d45a44bfR68

hkaiser:
> I used [[maybe_unused]] for the unused warning to go away, any comments on this ?
> https://github.com/STEllAR-GROUP/hpx/pull/4388/files#diff-271309c08d45275ceb7b63a1d45a44bfR68

@aurianer `[[maybe_unused]]` is a C++17 feature. Feel free to add a feature test macro and some `HPX_MAYBE_USED` to resolve to the attribute if available. This might not be necessary, however - see below. 

Those (non-const) static variables you define in the header file (https://github.com/STEllAR-GROUP/hpx/pull/4388/files#diff-271309c08d45275ceb7b63a1d45a44bfR62 and below) will create an instance per translation unit this header file is included by. I think this will need some restructuring (i.e. hiding them in a .cpp file or similar), you don't want to make those visible to users anyways.

aurianer:
> > I used [[maybe_unused]] for the unused warning to go away, any comments on this ?
> > https://github.com/STEllAR-GROUP/hpx/pull/4388/files#diff-271309c08d45275ceb7b63a1d45a44bfR68
> 
> @aurianer `[[maybe_unused]]` is a C++17 feature. Feel free to add a feature test macro and some `HPX_MAYBE_USED` to resolve to the attribute if available. This might not be necessary, however - see below.
> 
> Those (non-const) static variables you define in the header file (https://github.com/STEllAR-GROUP/hpx/pull/4388/files#diff-271309c08d45275ceb7b63a1d45a44bfR62 and below) will create an instance per translation unit this header file is included by. I think this will need some restructuring (i.e. hiding them in a .cpp file or similar), you don't want to make those visible to users anyways.

I added the cxx17 test related to maybe unused, thanks !
Could you develop the "hiding them in a cpp file" ? Because I need those in `hpx_init_impl.hpp` and `hpx_start_impl.hpp` which are also header files, so I'm not sure how to do the trick

hkaiser:
> I added the cxx17 test related to maybe unused, thanks !

Thanks!

> Could you develop the "hiding them in a cpp file" ? Because I need those in hpx_init_impl.hpp and hpx_start_impl.hpp which are also header files, so I'm not sure how to do the trick

Well, you could declare them to be `HPX_EXPORT extern` in the header files and define them in a .cpp file that is linked into the core library (hpx_init.cpp?). See for instance [here](https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/hpx_init_impl.hpp#L31) and [here](https://github.com/STEllAR-GROUP/hpx/blob/master/src/custom_exception_info.cpp#L51)

msimberg:
I think you have to exclude `hpx_init_params.cpp` from the `hpx_init` library (it's added to both `hpx` and `hpx_init` now). See https://github.com/STEllAR-GROUP/hpx/blob/4ba7e2e6bb9da77d30a46b0e8f2adee657959d36/src/CMakeLists.txt#L60-L65.

aurianer:
> I think you have to exclude `hpx_init_params.cpp` from the `hpx_init` library (it's added to both `hpx` and `hpx_init` now). See
> 
> https://github.com/STEllAR-GROUP/hpx/blob/4ba7e2e6bb9da77d30a46b0e8f2adee657959d36/src/CMakeLists.txt#L60-L65
> .

Absolutely thanks !

src/hpx_init.cpp
```diff
@@ -781,7 +783,7 @@ namespace hpx
                 std::unique_ptr<hpx::runtime> rt(new runtime_type(cms.rtcfg_));
 
                 result = run_or_start(blocking, std::move(rt),
-                    cms, std::move(startup), std::move(shutdown));
+                    cms, std::move(params.startup), std::move(params.shutdown));
```
hkaiser: Sorry for commenting only after this was merged... You can't move anything that has been marked `const`. I'm surprised clang-tidy didn't report this.

src/hpx_init_params.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include <hpx/hpx_init_params.hpp>
+#include <hpx/runtime/shutdown_function.hpp>
+#include <hpx/runtime/startup_function.hpp>
+#include <hpx/type_support/unused.hpp>
+
+
+namespace hpx { namespace detail {
+
+    const hpx::program_options::options_description default_desc(
+                "Usage: " HPX_APPLICATION_STRING " [options]");
```
hkaiser: This can't be in a `.cpp` file as the macro `HPX_APPLICATION_STRING` is define specifically for each executable.

aurianer: oh yeah sorry !

PR #4391, opened 2020-02-17
Fixing use of allocator for C++20

hkaiser:
This removes the deprecated use of Allocator::template rebind<>


aurianer:
LGTM thanks!

PR #4393, opened 2020-02-17
Use appropriate cache-line size defaults for different platforms

hkaiser:


libs/concurrency/include/hpx/concurrency/cache_line_data.hpp
```diff
@@ -23,8 +23,14 @@ namespace hpx {
         {
 #if defined(HPX_HAVE_CXX17_HARDWARE_DESTRUCTIVE_INTERFERENCE_SIZE)
             return std::hardware_destructive_interference_size;
+#else
```
msimberg: I suppose this could be `#elif defined(__s390__) ...` etc. as well (i.e. without the nesting). Either one works for me but just thought I'd point it out...

hkaiser: I was considering this but decided to leave the whole platform specific branch in a separate `#else` to improve readability.

PR #4395, opened 2020-02-17
Adding a parameter to auto_chunk_size allowing to control the amount of iterations to measure

hkaiser:
- note, this is a breaking change (with minimal potential impact) that adds
  an argument to the callback function that is used by get_chunk_size()
- this also removes unnecessary limitations from `auto_chunk_size` such that 
  the measurement is now always performed


msimberg:
I think this is unused enough that we can make this change without worrying too much. Thanks!

PR #4397, opened 2020-02-19
Doc updates batch environment

rstobaugh:
Creating documentation for batch environments module.

### DIFF
--- a/libs/batch_environments/docs/index.rst
+++ b/libs/batch_environments/docs/index.rst
@@ -11,3 +11,14 @@
 batch_environments
 ==================
 
+This module allows for the detection of execution as batch jobs, a series of
+programs executed without user intervention. All data is preselected and will
+be executed according to preset parameters, such as date or completion of
+another task. Batch environments are especially useful for executing repetitive
+tasks.
+
+|hpx| supports the creation of batch jobs through the Portable Batch System
+(PBS) and SLURM.
+
+For more information on batch environments, see :ref:`running_on_batch_systems`
+and the :ref:`API reference <libs_batch_environments_api>` for the module.


libs/batch_environments/docs/index.rst
```diff
@@ -11,3 +11,16 @@
 batch_environments
 ==================
 
+This module allows for the execution of batch jobs, a series of programs
```
msimberg: I would rather word this as "detection of execution as batch jobs". When an HPX program starts this module is essentially responsible for checking if it thinks the program runs as part of a batch job, and if it is to grab settings from the batch job environment (such as the number of nodes).

biddisco:
The text looks ok, if a little brief, but I think this can be squashed into a single commit instead of 7.

msimberg:
@rstobaugh did someone already show you how to rebase?

msimberg:
@rstobaugh I've rebased this to make it just one commit. Here's a mini-tutorial for how to do it yourself next time on the command line (it might be easier if you're using a GUI):

```
git fetch origin
git checkout your_branch
git rebase origin/master
```

If there were no conflicts this should return successfully. Now, because this rewrites git history you need to force push the changes to the remote:

```
git push --force-with-lease
```

If, after `git rebase` there are messages about conflicts you can do `git status` to see which files have conflicts. When you open the files they will have the conflicts marked like this:

```
<<<<<<<
conflicting changes from master branch
=======
conflicting changes from your branch
>>>>>>>
```

You'll have to choose which changes to keep. Once that's done you have to `git add files/that/had/conflicts`, and do `git rebase --continue`. The rebase process continues like this until there are no more conflicts. Once done `git push --force-with-lease`.

PR #4399, opened 2020-02-19
Add more threading modules

msimberg:
Adds remaining threading modules not part of #4265, i.e.:

- `schedulers`: contains scheduler implementations (including `thread_queue`, queue holders and other helpers)
- `thread_pools`: contains thread pool implementations (i.e. `scheduled_thread_pool` and `scheduling_loop`
- `io_service`: contains the IO service pools built on top of Boost.ASIO
- `thread`: contains `hpx::thread`

The only non-obvious changes I've made are:

- remove `scheduler_specific_ptr` (unused)
- remove `thread_data_fwd.hpp` and replace it with direct includes where necessary
- move `get_thread_count` and `enumerate_threads` out from `threadmanager` to `thread_helpers.hpp/cpp` since they get the thread manager from the global runtime object


biddisco:
Looks like there are still a few problems on the dashboard that need to be cleared up with this PR.

Bueshadow:
Hope

msimberg:
> Looks like there are still a few problems on the dashboard that need to be cleared up with this PR.

@biddisco this is ready for review now.

hpx/async.hpp
```diff
@@ -13,7 +13,7 @@
 #include <hpx/lcos/async_continue.hpp>
 #include <hpx/lcos/future.hpp>
 #include <hpx/runtime/launch_policy.hpp>
-#include <hpx/runtime/threads/thread.hpp>
+#include <hpx/thread.hpp>
```
hkaiser: Can we use `threads` as the name of this module, please? This would avoid naming problems compared to `hpx::thread`.

hkaiser: Possible `threading` would work as well.

libs/thread/include/hpx/thread/thread.hpp
```diff
@@ -23,20 +23,19 @@
...
-namespace hpx
-{
+namespace hpx {
     ///////////////////////////////////////////////////////////////////////////
     using thread_termination_handler_type =
         std::function<void(std::exception_ptr const& e)>;
     void set_thread_termination_handler(thread_termination_handler_type f);
 
-    class thread
+    class HPX_EXPORT thread
```
hkaiser: What's the rationale of `HPX_EXPORT`ing this class?

hpx/async.hpp
msimberg: Yeah, let's do that. I guess `threading` is more consistent with `threading_base`.

libs/thread/include/hpx/thread/thread.hpp
msimberg: As far as I can remember it has to be exported (because of the member functions defined in the cpp file). I can have a go at removing it though.

hpx/async.hpp
msimberg: I've renamed it `threading`.

libs/thread/include/hpx/thread/thread.hpp
msimberg: Yep, it's required.

msimberg:
And this one should be passing everything now as well. Had to fix one of the new header tests.

libs/threading/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) 2019 The STE||AR-Group
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+cmake_minimum_required(VERSION 3.3.2 FATAL_ERROR)
```
aurianer: version 3.13 ?

libs/io_service/tests/performance/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+# Copyright (c) 2019 The STE||AR-Group
+#
```
aurianer: Empty tests CMakeLists can probably be removed

libs/schedulers/tests/performance/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+# Copyright (c) 2019 The STE||AR-Group
```
aurianer: Yep we can probably remove those too ?

libs/thread_pools/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) 2019 The STE||AR-Group
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+cmake_minimum_required(VERSION 3.3.2 FATAL_ERROR)
```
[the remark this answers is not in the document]
msimberg: Thanks, very good catch!

libs/io_service/tests/performance/CMakeLists.txt
msimberg: Yes, we should probably do that... I won't do it here as we've kept them for other modules and that keeps the modules' `tests/CMakeLists.txt` pretty much the same for now.

libs/thread_pools/CMakeLists.txt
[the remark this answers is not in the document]
msimberg: Fixed this and a bunch of others as well. Thanks again!

PR #4405, opened 2020-02-25
Fix cross pool injection test, use default scheduler as falback

biddisco:
The cross pool scheduler test fails for some builds because the shared priority scheduler is not enabled.

This fix should drop down to the default scheduler if the other is not available.

In future I'd like to extend the test to test all schedulers in turn.

### DIFF
--- a/tests/unit/resource/cross_pool_injection.cpp
+++ b/tests/unit/resource/cross_pool_injection.cpp
@@ -238,8 +238,13 @@ int main(int argc, char* argv[])
                         st_rand(
                             0, ((std::max)(std::size_t(1), max_threads / 2)));
                     pool_name = "pool-" + std::to_string(num_pools);
+#if defined(HPX_HAVE_SHARED_PRIORITY_SCHEDULER)
                     rp.create_thread_pool(pool_name,
                         hpx::resource::scheduling_policy::shared_priority);
+#else
+                    rp.create_thread_pool(pool_name,
+                        hpx::resource::scheduling_policy::local_priority_fifo);
+#endif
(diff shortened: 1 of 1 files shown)


biddisco:
This does not appear to break anything and the tests have passed on a couple of dashboard runs, so I'm merging this without review. I'd like to see a bit more green on the dashboard to make PR reviewing easier.

hkaiser:
This essentially stops testing the default scheduler replacing the test with the `shared_priority_scheduler` (which is enabled for our tests). IOW, this does not fix the issue, just hides it. Could we please add a test for the default `local_priority_scheduler`?

biddisco:
This test was designed specifically for the shared_priority scheduler. I will extend the test in the future to check the others, but currently it is not high on my todo list.

hkaiser:
> This test was designed specifically for the shared_priority scheduler. I will extend the test in the future to check the others, but currently it is not high on my todo list.

In this case I'd suggest removing the `local_priority_scheduler` completely and rather disabling the test if the `shared_priority_scheduler` is not available. Otherwise this test would fail as soon as somebody disables that scheduler.

biddisco:
I'll create an issue to extend the test. The shared priority scheduler uses a slightly more complex method of placing tasks on queues/cores and I noticed that when I spawned jobs from one pool using a continuation with an executor on another pool, it caused problems (thread indexing on different pools). The other schedulers use a simpler approach to enqueuing tasks and the test doesn't really stress them - however as more people start using multiple pools and the schedulers are extended - it might be that problems are uncovered and this test will be useful then.

PR #4413, opened 2020-02-27
Merging from V1.4.1 release

hkaiser:
This also updates the version information for V1.5.0


docs/sphinx/quickstart.rst
```diff
@@ -33,7 +33,7 @@ It is also recommended that you check out the latest stable tag:
 
 .. code-block:: sh
 
-    git checkout 1.4.0
+    git checkout 1.5.0
```
msimberg: I'd like to keep this at 1.4.1 since that's the latest actual release.

hkaiser:
> Eventually I'd like to not bump the version number like this because in principle master will now be 1.4.1 plus some commits rather than an unreleased 1.5.0 (the next one could be 2.0.0 at some point but we won't know for sure until when we actually start making the next release). I want to do some research on how other project manage this before changing the process though.

The rationale for me to change the version number on master is to allow for people that follow HPX development (by relying on top of the repository) to have a means in their code to identify new features or breaking changes.

msimberg:
> The rationale for me to change the version number on master is to allow for people that follow HPX development (by relying on top of the repository) to have a means in their code to identify new features or breaking changes.

That's a good point. I'm not sure if bumping the version number like that is a sufficient signal for that (plus we should have the usual compatibility options in place anyway for breaking changes), but this clearly still needs some thought. Let's stick with what we know works then.

msimberg:
On the other hand one could argue that what you're doing with phylanx isn't really recommended practice, and that you know about breaking changes anyway (plus going from 1.5.0 on master to another breaking commit with 1.5.0 on master tells you nothing).

hkaiser:
@msimberg FWFW, other projects solve this by assigning an intermediate version number, something like `V1.4.1-9` that will be used between releases on the master branch.

PR #4426, opened 2020-03-09
Add currently failing tests on GitHub actions to blacklist

msimberg:


msimberg:
By the way, I've now added tests that failed on previous runs, but I suspect that we'll still have some occasional timeouts on tests that don't always fail. In this case feel free to add them to the list of excluded tests at some point in the future.

aurianer:
Thanks !

PR #4430, opened 2020-03-09
Add performance_counters module

msimberg:


msimberg:
This looks like it's ready (CI that has finished so far is happy).

msimberg:
> LGTM, except for the spelling errors.

In this case I'm not sure I would like to make spellcheck quieter: https://169208-4455628-gh.circle-artifacts.com/0/hpx/artifacts/spelling_suggestions.txt. I can easily be convinced to add them to the whitelist though...

hkaiser:
> LGTM, except for the spelling errors.
> 
> In this case I'm not sure I would like to make spellcheck quieter: https://169208-4455628-gh.circle-artifacts.com/0/hpx/artifacts/spelling_suggestions.txt. I can easily be convinced to add them to the whitelist though...

If we leave those spelling diagnostics un-silenced, we should at least make sure the spell-checking step is not flagged as failing.

msimberg:
I silenced them in another PR now.

PR #4432, opened 2020-03-10
Fix missing hwloc dependencies

K-ballo:


K-ballo:
> I wonder how this worked before and why it stopped working now...

The version module didn't use to exist before, and as part of `hpx` the hwloc dependency was getting implicitly satisfied via `topology`.

msimberg:
It seems like our Travis Windows builder has decided to change itself in to a Linux box... Has anyone had problems with Travis recently on other projects?

PR #4434, opened 2020-03-11
Clean up Boost dependencies and copy string algorithms to new module

msimberg:
This cleans up some unused Boost headers, and overloads for Boost types that I deem unnecessary. Please speak up if you think it's essential to keep any of these around.

I've also copied the string algorithms from Boost.Algorithm into a new module. In the first pass this still uses Boost.Range and other utilities from Boost. I attempted to replace `iterator_range` with ours from `iterator_support` but failed. Our ranges don't seem to quite match Boost's. However, I think it's definitely feasible to do the replacement, given a bit more knowledge about how things work in Boost (which I don't have at the moment...).

### DIFF
--- a/libs/CMakeLists.txt
+++ b/libs/CMakeLists.txt
@@ -54,10 +54,11 @@ set(HPX_LIBS
   serialization
   static_reinit
   statistics
+  string_util
   synchronization
   testing
-  threading_base
   thread_support
+  threading_base
   threadmanager
   timing
   topology
--- a/libs/affinity/src/parse_affinity_options.cpp
+++ b/libs/affinity/src/parse_affinity_options.cpp
@@ -13,8 +13,6 @@
 
 #include <hwloc.h>
 
-#include <boost/variant.hpp>
-
 // #define BOOST_SPIRIT_DEBUG
 #define BOOST_SPIRIT_USE_PHOENIX_V3
 #include <boost/fusion/include/adapt_struct.hpp>
--- a/libs/serialization/include/hpx/serialization/input_archive.hpp
+++ b/libs/serialization/include/hpx/serialization/input_archive.hpp
@@ -5,8 +5,6 @@
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-// hpxinspect:nodeprecatedinclude:boost/cstdint.hpp
-
 #ifndef HPX_SERIALIZATION_BASIC_INPUT_ARCHIVE_HPP
 #define HPX_SERIALIZATION_BASIC_INPUT_ARCHIVE_HPP
 
@@ -18,7 +16,6 @@
 #include <hpx/serialization/input_container.hpp>
 #include <hpx/serialization/traits/is_bitwise_serializable.hpp>
 
-#include <boost/cstdint.hpp>
 #include <boost/predef/other/endian.h>
 
 #include <cstddef>
@@ -180,41 +177,6 @@ namespace hpx { namespace serialization {
             load_integral_impl(ul);
             val = static_cast<T>(ul);
         }
-
-#if defined(BOOST_HAS_INT128) && !defined(__NVCC__) && !defined(__CUDACC__)
-        void load_integral(boost::int128_type& t, std::false_type)
-        {
-            load_integral_impl(t);
-        }
-
-        void load_integral(boost::uint128_type& t, std::true_type)
-        {
-            load_integral_impl(t);
-        }
-
-        // On some platforms (gcc) std::is_integral<int128>::value
-        // evaluates to false. Thus, these functions re-route the
-        // serialization for those types to the proper implementation
-        void load_bitwise(boost::int128_type& t, std::false_type)
-        {
-            load_integral_impl(t);
-        }
-
-        void load_bitwise(boost::int128_type& t, std::true_type)
-        {
-            load_integral_impl(t);
-        }
-
-        void load_bitwise(boost::uint128_type& t, std::false_type)
-        {
-            load_integral_impl(t);
-        }
-
-        void load_bitwise(boost::uint128_type& t, std::true_type)
-        {
-            load_integral_impl(t);
-        }
-#endif
         template <class Promoted>
         void load_integral_impl(Promoted& l)
         {
--- a/libs/string_util/include/hpx/string_util/classification.hpp
+++ b/libs/string_util/include/hpx/string_util/classification.hpp
@@ -1,0 +1,49 @@
+//  Copyright (c) 2020 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef HPX_STRING_UTIL_CLASSIFICATION_HPP
+#define HPX_STRING_UTIL_CLASSIFICATION_HPP
+
+#include <cctype>
+#include <string>
+
+namespace hpx { namespace string_util {
+    namespace detail {
+        template <typename CharT, typename Traits, typename Allocator>
+        struct is_any_of_pred
+        {
+            bool operator()(int c) const noexcept
+            {
+                return chars.find(c) != std::string::npos;
+            }
+
+            std::basic_string<CharT, Traits, Allocator> chars;
+        };
+    }    // namespace detail
+
+    template <typename CharT, typename Traits, typename Allocator>
+    detail::is_any_of_pred<CharT, Traits, Allocator> is_any_of(
+        std::basic_string<CharT, Traits, Allocator> const& chars)
+    {
+        return detail::is_any_of_pred<CharT, Traits, Allocator>{chars};
+    }
+
+    inline auto is_any_of(char const* chars)
+    {
+        return detail::is_any_of_pred<char, std::char_traits<char>,
+            std::allocator<char>>{std::string{chars}};
+    }
+
+    struct is_space
+    {
+        bool operator()(int c) const noexcept
+        {
+            return std::isspace(c);
+        }
+    };
+}}    // namespace hpx::string_util
+
+#endif
--- a/libs/string_util/include/hpx/string_util/split.hpp
+++ b/libs/string_util/include/hpx/string_util/split.hpp
@@ -1,0 +1,80 @@
+//  Copyright (c) 2020 ETH Zurich
+//  Copyright (c) 2002-2006 Pavol Droba
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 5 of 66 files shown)


libs/serialization/include/hpx/serialization/input_archive.hpp
```diff
@@ -180,41 +177,6 @@ namespace hpx { namespace serialization {
...
-        void load_bitwise(boost::uint128_type& t, std::false_type)
-        {
-            load_integral_impl(t);
-        }
-
-        void load_bitwise(boost::uint128_type& t, std::true_type)
-        {
-            load_integral_impl(t);
-        }
-#endif
```
hkaiser: Not sure about removing the 128 bit integral types here. OTOH, we might have a replacement already, actually: `util::uint128_t`/`util::int128_t`?

msimberg:
> Should we move all of boost::algorithm to HPX? It might be simpler to just move what we actually need...

I started off with this, but I would absolutely not be opposed to trimming it down to just what we actually use. Note that the string algorithms are already a subset of the Boost.Algorithms library.

> Additionally, I'm not too fond of the module name but have nothing to propose as an alternative at this point.

Same. My other options would be `string_algorithms` or `string_utilities` (or `string_util`).

libs/serialization/include/hpx/serialization/input_archive.hpp
msimberg: As far as I understand those are indeed replacements. I couldn't find any direct uses of Boost's 128 bit types, except for the overloads.

hkaiser: It's probably safe to remove those. If somebody needs support for serializing 128bit integral types it can be added easily.

hkaiser:
> 
> 
> > Should we move all of boost::algorithm to HPX? It might be simpler to just move what we actually need...
> 
> I started off with this, but I would absolutely not be opposed to trimming it down to just what we actually use. Note that the string algorithms are already a subset of the Boost.Algorithms library.

I think that HPX uses only a handful of those string algorithms (like `trim`) most of which can be implemented as two liners. Could you create alist of what's actually needed? I'd be happy

> > Additionally, I'm not too fond of the module name but have nothing to propose as an alternative at this point.
> 
> Same. My other options would be `string_algorithms` or `string_utilities` (or `string_util`).

I like `string_util`

msimberg:
Yeah, you're definitely right. Let's try to do that then. The list of `hpx::string` (or `boost::algorithm::string`) algorithms used at the moment (assuming my search is exhaustive) is:

```
to_lower
trim
trim_copy
split
```

and related classes are:
```
token_compress_off
token_compress_on
is_any_of
is_space
```

`to_lower` should be a trivial application of `std::tolower`. `trim` and `trim_copy` should be easy enough with `std::isspace`. `split` is not trivial but should not be too bad either.

msimberg:
There's now a first iteration of the stripped-down `string_util` module in place. Please have a look. I'm not quite sure the way I've added the overloads (for `basic_string` and `char const*`) is the shortest/most general way. They cover the use cases we have at least. Other comments of course also welcome.

The copy of the Boost string algorithms library still lives in this branch for archiving purposes: https://github.com/msimberg/hpx/tree/boost-string-library.

libs/CMakeLists.txt
```diff
@@ -96,7 +97,7 @@ if(HPX_WITH_TESTS)
   endif()
 
   if (HPX_WITH_TESTS_HEADERS)
-    add_hpx_pseudo_target(tests.headers.modules)
+    add_custom_target(tests.headers.modules)
```
hkaiser: Please don't use `add_custom_target` directly, use `add_hpx_pseudo_target` instead. Otherwise this will break Windows builds.

msimberg: Grr, thanks. I'll make a separate PR to have `create_library_skeleton.py` fixed in master finally.

libs/affinity/src/parse_affinity_options.cpp
```diff
@@ -13,8 +13,6 @@
 
 #include <hwloc.h>
 
-#include <boost/variant.hpp>
-
```
hkaiser: Boost Spirit relies internally on `boost::variant` (for parsing alternatives). We can remove this `#include` here as we don't directly refer to the type. This does not mean however that this code doesn't depend on it anymore...

libs/CMakeLists.txt
msimberg: https://github.com/STEllAR-GROUP/hpx/commit/b3503fc5a4aac7c1bf618993d9e0f726194439d3. I pushed directly to master, and will fix this PR shortly as well.

libs/string_util/include/hpx/string_util/classification.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+            std::basic_string<CharT> chars;
+        };
+    }    // namespace detail
+
+    template <typename CharT>
+    detail::is_any_of_pred<CharT> is_any_of(
+        std::basic_string<CharT> const& chars)
+    {
+        return detail::is_any_of_pred{chars};
```
hkaiser: This relies on CTAD (C++17), could you explicitly provide the `<CharT>` argument, please?

libs/string_util/include/hpx/string_util/classification.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include <cctype>
+#include <string>
+
+namespace hpx { namespace string_util {
+    namespace detail {
+        template <typename CharT>
+        struct is_any_of_pred
+        {
+            bool operator()(int c)
```
hkaiser: Can this be `const` and `noexcept`?

libs/string_util/include/hpx/string_util/classification.hpp
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    inline auto is_any_of(char const* chars)
+    {
+        return detail::is_any_of_pred<char>{std::string{chars}};
+    }
+
+    struct is_space
+    {
+        bool operator()(int c)
```
hkaiser: This can be `const` and `noexcept`

libs/string_util/include/hpx/string_util/classification.hpp
```diff
@@ -0,0 +1,48 @@
...
+    namespace detail {
+        template <typename CharT>
+        struct is_any_of_pred
+        {
+            bool operator()(int c)
+            {
+                return chars.find(c) != std::string::npos;
+            }
+
+            std::basic_string<CharT> chars;
```
hkaiser: `std::basic_string<CharT, Traits, Alloc>`?

libs/string_util/include/hpx/string_util/split.hpp
```diff
@@ -0,0 +1,73 @@
...
+            std::basic_string<CharT, Traits, Allocator> const& s,
+            It const& first, It const& last)
+        {
+            std::size_t const pos = std::distance(std::begin(s), first);
+            std::size_t const count = std::distance(first, last);
+            return s.substr(pos, count);
+        }
+    }    // namespace detail
+
+    enum token_compress_mode_type
```
hkaiser: Let's make this an `enum class`.

hkaiser: Like:
```
    enum class token_compress_mode
    {
        on, off
    };
```

libs/string_util/include/hpx/string_util/split.hpp
```diff
@@ -0,0 +1,73 @@
...
+        auto token_begin = std::begin(str);
+        auto token_end = std::end(str);
+
+        do
+        {
+            token_end = std::find_if(token_begin, std::end(str), pred);
+
+            container.push_back(detail::substr(str, token_begin, token_end));
+
+            token_begin = token_end + 1;
```
hkaiser: This might be dangerous if `std::find_if` above does not find any match.

libs/string_util/include/hpx/string_util/split.hpp
```diff
@@ -0,0 +1,73 @@
...
+                }
+            }
+        } while (token_end != std::end(str));
+    }
+
+    template <typename Container, typename Predicate>
+    void split(Container& container, char const* str, Predicate&& pred,
+        token_compress_mode_type compress_mode = token_compress_off)
+    {
+        split(container, std::string{str}, pred, compress_mode);
```
hkaiser: Should this `std::forward` the `Predicate`?

libs/string_util/include/hpx/string_util/trim.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include <algorithm>
+#include <cstddef>
+#include <string>
+
+namespace hpx { namespace string_util {
+    template <typename CharT, class Traits, class Alloc>
+    void trim(std::basic_string<CharT, Traits, Alloc>& s)
+    {
+        auto first = std::find_if_not(std::cbegin(s), std::cend(s),
+            [](unsigned char c) { return std::isspace(c); });
```
hkaiser: `unsigned char` does not look right here, did you mean `char`. Also, in other spots you use `int` as the argument for the lambda, should this be unified?

msimberg: Thanks! It should also be `int` (`isspace` takes an int).

libs/string_util/include/hpx/string_util/split.hpp
msimberg: Absolutely.

libs/CMakeLists.txt
hkaiser: Thanks! I was planning to change that a while back but forgot as well.

libs/string_util/include/hpx/string_util/split.hpp
msimberg: It is in fact safe because it doesn't get dereferenced in that case, but I'll try to rearrange it a bit to make it clear (or avoid updating `token_begin` in that case).

libs/string_util/include/hpx/string_util/classification.hpp
msimberg: Good catch, will fix.

libs/string_util/include/hpx/string_util/split.hpp
hkaiser: Incrementing an end iterator is UB in any case, even if you don't dereference the result.

msimberg:
> Thanks a lot @msimberg! Very nice.
> 
> For the serialization module, we use `HPX_SERIALIZATION_WITH_BOOST_TYPES` to enable support for Boost types. Should we do something similar here (for `boost::ref`) instead of unconditionally removing the specializations for `boost::reference_wrapper`?

I think it makes sense for things like `variant`, `dynamic_bitset`, or `intrusive_ptr` that we don't have in the standard library at all or only in later standards. I think there's very little reason to keep support for `boost::ref` (except for compatibility, but I think the risk of us breaking someone's code by removing these is low). What do you think?

libs/string_util/include/hpx/string_util/split.hpp
msimberg: I did not know that, very good to know! Thanks!

msimberg:
I think I've addressed the review comments now.

msimberg:
I've only rebased and hopefully fixed the inspect errors. Let's see if CircleCI agrees...

PR #4439, opened 2020-03-16
Link HPX::plugin and component privately in hpx_setup_target

msimberg:
Fix to follow up on #4246. The `component` and `plugin` targets were linked publicly in `hpx_setup_target` leading to dependent targets to also get definitions like `HPX_COMPONENT_EXPORTS` which should stay private.


hkaiser:
Do we need to link against `plugin` targets at all? I'm not sure I have an answer...

msimberg:
> Do we need to link against `plugin` targets at all? I'm not sure I have an answer...

No, you're right. Plugins should never get linked to so they could have either private or public definitions and noone would (should) notice. I thought it's best to have the definitions private in both cases though just to be consistent.

PR #4442, opened 2020-03-17
Adding a tag to all auto-generated files allowing for tools to visually distinguish those

hkaiser:
This is to enable visual identification for generated files in editors.


hkaiser:
Let's not merge this yet. I'm in contact with the author of the tool I was planning to use to make the tag configurable. If that gets implemented I can make it work without this PR.

hkaiser:
We can close this as the tool I had in mind added support for configurable tags (https://github.com/mrlacey/ClearlyEditable/issues/8).

PR #4443, opened 2020-03-18
Add missing tests to CircleCI config

msimberg:


hkaiser:
@msimberg We might have to split the tests.headers target into smaller pieces now as it looks like CTest runs out of memory after a while.

msimberg:
> @msimberg We might have to split the tests.headers target into smaller pieces now as it looks like CTest runs out of memory after a while.

Interesting... I didn't even change the way the header tests are handled. I split them up now into modules header tests and all other header tests. I made the non-modules header tests only run one test at a time. We'll need to split them differently again when more modules are merged but this should do for now.

PR #4444, opened 2020-03-18
Minor CUDA fixes

msimberg:
- Add a dummy `hpx::throws` declaration to suppress errors about `hpx::throws` being undefined in device code (suggestions for better/safer ways to fix this welcome!)
- Remove `HPX_HOST_DEVICE` attribute from defaulted `tuple` constructors; this is apparently not needed (and causes annoying warnings)

### DIFF
--- a/libs/datastructures/include/hpx/datastructures/tuple.hpp
+++ b/libs/datastructures/include/hpx/datastructures/tuple.hpp
@@ -149,12 +149,12 @@ namespace hpx { namespace util {
         // tuple(const tuple& u) = default;
         // Initializes each element of *this with the corresponding element
         // of u.
-        constexpr HPX_HOST_DEVICE tuple(tuple const& /*other*/) = default;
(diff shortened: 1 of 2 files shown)


libs/datastructures/include/hpx/datastructures/tuple.hpp
```diff
@@ -149,12 +149,12 @@ namespace hpx { namespace util {
         // tuple(const tuple& u) = default;
         // Initializes each element of *this with the corresponding element
         // of u.
-        constexpr HPX_HOST_DEVICE tuple(tuple const& /*other*/) = default;
+        constexpr tuple(tuple const& /*other*/) = default;
```
hkaiser: Just out of curiosity: is it that this is not needed because of the `= default;`?

msimberg: Yep, exactly. It will automatically make default constructors `__host__` and `__device__`.

PR #4446, opened 2020-03-19
Resume tasks where they were suspended

msimberg:
Resumes tasks on the worker thread that they were suspended on. This is something I *think* will help in some cases (based on some task plots I've seen), but I don't currently have a benchmark to verify if this makes things better or worse. It does make sense to avoid invalidating caches though.

I've at the same time tried to clean up the thread number implementation. Please comment on naming, exposed functions etc.

### DIFF
--- a/libs/threading_base/src/thread_data.cpp
+++ b/libs/threading_base/src/thread_data.cpp
@@ -67,6 +67,7 @@ namespace hpx { namespace threads {
         , enabled_interrupt_(true)
         , ran_exit_funcs_(false)
         , scheduler_base_(init_data.scheduler_base)
+        , last_worker_thread_num_(std::size_t(-1))
         , stacksize_(init_data.stacksize)
         , queue_(queue)
         , is_stackless_(is_stackless)
@@ -195,6 +196,7 @@ namespace hpx { namespace threads {
             ran_exit_funcs_ = false;
             exit_funcs_.clear();
             scheduler_base_ = init_data.scheduler_base;
+            last_worker_thread_num_ = std::size_t(-1);
 
             HPX_ASSERT(init_data.stacksize == get_stack_size());
 
--- a/libs/threading_base/src/thread_num_tss.cpp
+++ b/libs/threading_base/src/thread_num_tss.cpp
@@ -15,45 +15,55 @@
 
 namespace hpx { namespace threads { namespace detail {
     namespace {
-        std::size_t& thread_num_tss()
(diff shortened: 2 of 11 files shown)


libs/threading_base/src/thread_data.cpp
```diff
@@ -67,9 +67,11 @@ namespace hpx { namespace threads {
         , enabled_interrupt_(true)
         , ran_exit_funcs_(false)
         , scheduler_base_(init_data.scheduler_base)
+        , last_worker_thread_num_(0)
         , stacksize_(init_data.stacksize)
         , queue_(queue)
         , is_stackless_(is_stackless)
+                     // TODO: Add member for storing the last used worker thread.
```
hkaiser: This comment is probably obsolete

libs/threading_base/src/thread_data.cpp
```diff
@@ -67,9 +67,11 @@ namespace hpx { namespace threads {
         , enabled_interrupt_(true)
         , ran_exit_funcs_(false)
         , scheduler_base_(init_data.scheduler_base)
+        , last_worker_thread_num_(0)
```
hkaiser: Is zero a good default value? You might want to use `std::size_t(-1)` instead.

libs/threading_base/src/thread_num_tss.cpp
```diff
@@ -15,57 +15,87 @@
...
-            return thread_num_tss_;
-        }
-
-        thread_pool& thread_pool_tss()
-        {
-            static thread_local thread_pool thread_pool_tss_ = {
-                std::uint16_t(-1), std::uint16_t(-1)};
-            return thread_pool_tss_;
+            static thread_local thread_nums thread_nums_tss_ = {
+                std::uint16_t(-1), std::uint16_t(-1), std::uint16_t(-1)};
```
hkaiser: The members of `thread_nums` are `std::size_t`s (not `std::uint16_t`)

msimberg:
Very good catches @hkaiser, thanks a lot for looking through this! Will update soon.

msimberg:
Pycicle does not seem to want to report the results on this, even though it has run the builds. Things look good so I'm going to merge this (https://cdash.cscs.ch//index.php?project=HPX&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=4446&field2=buildstarttime&compare2=83&value2=0).

PR #4447, opened 2020-03-20
This fixes an obsolete #include

hkaiser:
This also fixes an assertion that fires when MPI is enabled in the build but is not actually used to a distributed application.


hkaiser:
I'll go ahead and merge this as it prevent me from building things

PR #4449, opened 2020-03-20
Fixing build system to actually build variant test

### LINKED ISSUE #4448: New std::variant serialization broken
## Expected Behavior
 to compile
## Actual Behavior
not compiling with error
See gist: https://gist.github.com/McKillroy/cb04b97975db409e7f0c8657e3f85744
## Specifications
  - HPX Version: Commit 8269dd13183efa3bc0fa5346e7a26cc069ee0414
  - Platform (compiler, OS): MSVC 19, latest, Windows 10

hkaiser:
- adding compile only test verifying serialization of std::variant

Fixes #4448

### DIFF
--- a/libs/datastructures/include/hpx/datastructures/tuple.hpp
+++ b/libs/datastructures/include/hpx/datastructures/tuple.hpp
@@ -21,6 +21,9 @@
 #include <tuple>
 #include <type_traits>
 #include <utility>
+#if defined(HPX_HAVE_CXX17_STD_VARIANT)
+#include <variant>
+#endif
 
 #if defined(HPX_MSVC_WARNING_PRAGMA)
 #pragma warning(push)
@@ -468,6 +471,14 @@ namespace hpx { namespace util {
     };
 #endif
 
+#if defined(HPX_HAVE_CXX17_STD_VARIANT)
+    template <typename... Ts>
+    struct tuple_size<std::variant<Ts...>>
+      : std::integral_constant<std::size_t, sizeof...(Ts)>
+    {
+    };
+#endif
+
     // template <size_t I, class Tuple>
     // class tuple_element
     template <std::size_t I, typename T>
@@ -593,6 +604,38 @@ namespace hpx { namespace util {
             std::tuple<Ts...> const&& tuple) noexcept
         {
             return std::get<I>(std::move(tuple));
+        }
+    };
+#endif
+
+#if defined(HPX_HAVE_CXX17_STD_VARIANT)
+    template <std::size_t I, typename... Ts>
+    struct tuple_element<I, std::variant<Ts...>>
+    {
+        using type = typename util::at_index<I, Ts...>::type;
+
+        static constexpr HPX_HOST_DEVICE HPX_FORCEINLINE type& get(
+            std::variant<Ts...>& var) noexcept
+        {
+            return std::get<I>(var);
+        }
+
+        static constexpr HPX_HOST_DEVICE HPX_FORCEINLINE type const& get(
+            std::variant<Ts...> const& var) noexcept
+        {
+            return std::get<I>(var);
+        }
+
+        static constexpr HPX_HOST_DEVICE HPX_FORCEINLINE type&& get(
+            std::variant<Ts...>&& var) noexcept
+        {
+            return std::get<I>(std::move(var));
+        }
+
+        static constexpr HPX_HOST_DEVICE HPX_FORCEINLINE type const&& get(
+            std::variant<Ts...> const&& var) noexcept
+        {
+            return std::get<I>(std::move(var));
         }
     };
 #endif
--- a/libs/datastructures/include/hpx/datastructures/tuple.hpp
+++ b/libs/datastructures/include/hpx/datastructures/tuple.hpp
@@ -21,9 +21,6 @@
 #include <tuple>
 #include <type_traits>
 #include <utility>
-#if defined(HPX_HAVE_CXX17_STD_VARIANT)
-#include <variant>
-#endif
 
 #if defined(HPX_MSVC_WARNING_PRAGMA)
 #pragma warning(push)
@@ -471,14 +468,6 @@ namespace hpx { namespace util {
     };
 #endif
 
-#if defined(HPX_HAVE_CXX17_STD_VARIANT)
-    template <typename... Ts>
-    struct tuple_size<std::variant<Ts...>>
-      : std::integral_constant<std::size_t, sizeof...(Ts)>
-    {
-    };
-#endif
-
     // template <size_t I, class Tuple>
     // class tuple_element
     template <std::size_t I, typename T>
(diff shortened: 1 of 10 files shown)


hkaiser:
@msimberg this can be merged now, I think.

libs/datastructures/include/hpx/datastructures/tuple.hpp
```diff
@@ -468,6 +471,14 @@ namespace hpx { namespace util {
     };
 #endif
 
+#if defined(HPX_HAVE_CXX17_STD_VARIANT)
+    template <typename... Ts>
+    struct tuple_size<std::variant<Ts...>>
+      : std::integral_constant<std::size_t, sizeof...(Ts)>
+    {
+    };
+#endif
```
sithhell: A variant is not a tuple like thing. Further, the active member of the variant is not a compile time property.

libs/datastructures/include/hpx/datastructures/tuple.hpp
```diff
@@ -597,6 +608,38 @@ namespace hpx { namespace util {
...
+            return std::get<I>(std::move(var));
+        }
+
+        static constexpr HPX_HOST_DEVICE HPX_FORCEINLINE type const&& get(
+            std::variant<Ts...> const&& var) noexcept
+        {
+            return std::get<I>(std::move(var));
+        }
+    };
+#endif
```
sithhell: A variant is not a tuple like thing.

hkaiser:
> The changes to tuple_size and tuple_element are conceptually wrong. This needs to be implemented differently.

Could you be more specific, please? I added specializations for `std::variant` in the same way as the existing ones for `std::pair` etc.

libs/datastructures/include/hpx/datastructures/tuple.hpp
hkaiser: And still `std::get` is defined for it... (https://en.cppreference.com/w/cpp/utility/variant/get)

libs/datastructures/include/hpx/datastructures/tuple.hpp
hkaiser: ...and still `std::get` has to work with it. The standard says:

> Index-based value accessor: If `v.index() == I`, returns a reference to the value stored in `v`. Otherwise, throws `std::bad_variant_access`. The call is ill-formed if `I` is not a valid index in the variant.

hkaiser:
> Looks good to me. The clang-newest failure is due to Piz Daint. I can't comment on the validity of the get overloads but, it does make sense to have them there.

FWIW, I did add our specializations for `tuple_size` and `tuple_element` as otherwise gcc V9 throws a tantrum caused by its library relying on using `std::get` on variants (see https://cdash.cscs.ch//viewBuildError.php?buildid=100078).

sithhell:
> Could you be more specific, please? I added specializations for `std::variant` in the same way as the existing ones for `std::pair` etc.

`std::pair` and `std::variant` are fundamentally different data types. One could say that a std::pair is an implementation of a tuple with two elements. A variant however is something different. It can only  ever hold one (or none) of the types in the template.

In other words, a variant is a sum type and a tuple is a product type. Those expose different properties.

sithhell:
`std::get` should be supported, of course. This support should not be implemented in the way that a variant follows the same protocol a a tuple, which in other words means that conceptually, a variant is a tuple, which is obviously not true.

libs/datastructures/include/hpx/datastructures/tuple.hpp
sithhell: Well, `std::get` is defined for it, true. But `std::get` is not (generically) implemented in terms of `std::tuple_size`/`std::tuple_element`.

sithhell: What this additionally enables is structured binding for variants:
```
std::variant<int, float, double> v = ...;
auto [i, f, d] = v;
```
Which you have to agree, does not make a lot of sense.

libs/datastructures/include/hpx/datastructures/tuple.hpp
sithhell: As noted above, those are two orthogonal things.

libs/datastructures/include/hpx/datastructures/tuple.hpp
hkaiser: I agree that we don't need the specialization for `tuple_size`. We do however need it for `tuple_element` (for instance here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/datastructures/include/hpx/datastructures/tuple.hpp#L53). Alternatively we would have to redo all of our implementation for `util::get`.

hkaiser: Separate implementations of `util::adl_barrier::get` for `std::variant` would do the trick as well.

libs/datastructures/include/hpx/datastructures/variant_helper.hpp
```diff
@@ -0,0 +1,63 @@
...
+            return std::get<I>(std::move(var));
+        }
+
+        template <std::size_t I, typename... Ts>
+        constexpr HPX_HOST_DEVICE HPX_FORCEINLINE typename
+            typename util::at_index<I, Ts...>::type const&&
+            get(std::variant<Ts...> const&& var) noexcept
+        {
+            return std::get<I>(std::move(var));
+        }
```
sithhell: While we're at it, the variant `std::get` overloads also support type based access.

sithhell: This can also be implement once a need for that arises.

libs/datastructures/include/hpx/datastructures/variant_helper.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <variant>
+
+namespace hpx { namespace util {
+
+    // This is put into the same embedded namespace as the implementations in
+    // tuple.hpp
+    namespace adl_barrier {
+
+        template <std::size_t I, typename... Ts>
+        constexpr HPX_HOST_DEVICE HPX_FORCEINLINE typename
```
msimberg:
```suggestion
        constexpr HPX_HOST_DEVICE HPX_FORCEINLINE
```
?

https://cdash.cscs.ch//viewBuildError.php?buildid=100368 and https://cdash.cscs.ch//viewBuildError.php?buildid=100419.

msimberg:
Please ignore the `gcc-cuda` results. I enabled MPI on it but it's either badly configured or broken. I've reverted it for now.

libs/datastructures/include/hpx/datastructures/variant_helper.hpp
hkaiser: MSVC is just a piece of work! Two `typename`'s in a row and not even a blink of an eye... Thanks!

msimberg: Well, in defense of MSVC, the older GCCs and clangs also didn't care ;)

msimberg:
It looks like all is good on pycicle, but I'd like to wait to get all the results just be sure. Thanks @hkaiser!

McKillroy:
I'm using it by sending vectors of variants as action parameters in a messaging system and it works like a charm now.

PR #4450, opened 2020-03-21
Fixing compilation with std::filesystem enabled

hkaiser:


McKillroy:
Just for documentation: 
These are the three offending lines when I tried to compile on Windows with C++17 enabled:
https://github.com/STEllAR-GROUP/hpx/blob/master/libs/runtime_configuration/src/init_ini_data.cpp#L450
https://github.com/STEllAR-GROUP/hpx/blob/master/libs/command_line_handling/src/parse_command_line.cpp#L295
https://github.com/STEllAR-GROUP/hpx/blob/master/libs/plugin/include/hpx/plugin/detail/dll_windows.hpp#L78

hkaiser:
I'll go ahead and merge this now. The fix is very small but important.

PR #4454, opened 2020-03-24
Fix hello_word_component.cpp :lines

gonidelis:
Documentation on 'How to build HPX components with pkg-config' hello_world_component.cpp was quoted incompletely 
`:lines7-29.` The correct is 
`:lines 7-30`

### DIFF
--- a/docs/sphinx/manual/creating_hpx_projects.rst
+++ b/docs/sphinx/manual/creating_hpx_projects.rst
@@ -425,8 +425,8 @@ run it on.
 
 .. literalinclude:: ../../examples/quickstart/hello_world_1.cpp
    :language: c++
-   :start-after://[hello_world_1_getting_started
-   :end-before://]
+   :start-after: //[hello_world_1_getting_started
+   :end-before: //]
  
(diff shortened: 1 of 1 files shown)


gonidelis:
@msimberg Oh ok great! I shall fix all the `literalinclude`s then if that's not a problem as copyright and license commented lines will be removed from the manual html page.

docs/sphinx/manual/creating_hpx_projects.rst
```diff
@@ -419,7 +425,9 @@ run it on.
 
 .. literalinclude:: ../../examples/quickstart/hello_world_1.cpp
    :language: c++
-
+   :start-after://[hello_world_1_getting_started
+   :end-before://]
```
msimberg:
```suggestion
   :start-after: //[hello_world_1_getting_started
   :end-before: //]
```

msimberg: And the same below, just for consistency with how we have it elsewhere.

gonidelis: Got it!

PR #4458, opened 2020-03-25
Added example demonstrating how to create and use a wrapping executor

hkaiser:
- this executor wraps any other executor and makes sure that given start
  and stop functions are being called for each HPX thread created by the
  executor

### DIFF
--- a/examples/quickstart/executor_with_thread_hooks.cpp
+++ b/examples/quickstart/executor_with_thread_hooks.cpp
@@ -1,0 +1,230 @@
+//  Copyright (c) 2020 Hartmut Kaiser
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+///////////////////////////////////////////////////////////////////////////////
+// The purpose of this example is to show how to rite an executor that wraps
+// any other executor and adds a hook into thread start and thread exit allowing
+// to associate custom thread data with the tasks that are created by the
+// underlying executor.
+
+#include <hpx/hpx_main.hpp>
+
+#include <hpx/algorithms.hpp>
+#include <hpx/assertion.hpp>
+#include <hpx/execution.hpp>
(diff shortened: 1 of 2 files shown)


examples/quickstart/executor_with_thread_hooks.cpp
```diff
@@ -0,0 +1,224 @@
...
+    auto on_stop = [&]() { ++stops; };
+
+    auto exec = executor_example::make_executor_with_thread_hooks(
+        hpx::parallel::execution::par.executor(), on_start, on_stop);
+
+    hpx::parallel::for_loop(hpx::parallel::execution::par.on(exec), 0, v.size(),
+        [](std::size_t i) {});
+
+    std::cout << "Executed " << starts.load() << " starts and " << stops.load()
+              << " stops\n";
```
msimberg:
```suggestion
              << " stops\n";
              HPX_ASSERT(starts > 0);
              HPX_ASSERT(stops > 0);
```
just as a sanity check?

hkaiser: Ok, I'll add those tests, but will do so a `HPX_TEST_NEQ(starts, std::size_t(0))`.

hkaiser: Hold on... This is an example, not a test. So, `HPX_ASSERT` it is.

examples/quickstart/executor_with_thread_hooks.cpp
```diff
@@ -0,0 +1,228 @@
...
+// The purpose of this example is to show how to rite an executor that wraps
+// any other executor and adds a hook into thread start and thread exit allowing
+// to associate custom thread data with the tasks that are created by the
+// underlying executor.
+
+#include <hpx/hpx_main.hpp>
+
+#include <hpx/algorithms.hpp>
+#include <hpx/execution.hpp>
+#include <hpx/functional.hpp>
```
msimberg: Just inspect left on this one: https://173022-4455628-gh.circle-artifacts.com/0/hpx_inspect_report.html. It needs `<cstddef>` and `hpx/assertion.hpp`.

PR #4460, opened 2020-03-25
Suppress MPI warnings in version.cpp

msimberg:
We have https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/plugins/parcelport/mpi/mpi.hpp, but to avoid a dependency on the plugins in the version module, I've just included the warning suppression pragmas directly in the `version.cpp` file. With this I'll enable the MPI parcelport on pycicle again (on the `gcc-cuda` builder).


msimberg:
Well, at least the build passes without warnings (errors), but the pycicle setup doesn't use MPI correctly. That's independent of this PR though so merging now.

PR #4462, opened 2020-03-25
Merge hpx_init and hpx_wrap into one static library

### LINKED ISSUE #4466: Compile error on Windows, current stable:
## Expected Behavior

To compile

## Actual Behavior

Breaks with: 
\hpx\stable\libs\topology\include\hpx\topology\topology.hpp(27): fatal error C1083: Cannot open include file: 'hwloc.h': No such file or directory

I double-checked cmake vars HWLOC_ROOT, HWLOC_LIBRARY, HWLOC_INCLUDE_DIR for correct values, and it was correct, file was there inside %{HWLOC_INCLUDE_DIR }.

Compile on Windows with cmake

## Specifications

- HPX Version: stable ( 91fa57cf1cccfa4c3d98466a3ba96e44eca85d22 )
- Platform (compiler, OS): MSVC, latest (Version 16.5.1), cmake open folder project, Windows 10

msimberg:
Merges `hpx_init` and `hpx_wrap`, and moves them out of the main source directories. This seems to work on Linux at least, will wait and see that Windows and macOS also like this change. In theory there should be no problems with them being merged.

Fixes #4466 (maybe).


msimberg:
> Merges `hpx_init` and `hpx_wrap`, and moves them out of the main source directories. This seems to work on Linux at least, will wait and see that Windows and macOS also like this change. In theory there should be no problems with them being merged.

Flyby: I've attempted to simplify some things in the `CMakeLists.txt`, e.g. not using `cuda_add_library` for `hpx` and `hpx_init`. It should not be necessary if `hpx_compute_cuda` propagates the CUDA dependencies correctly.

init/CMakeLists.txt
```diff
@@ -0,0 +1,103 @@
...
+
+if(MSVC)
+  set_target_properties(hpx_init PROPERTIES
+    COMPILE_PDB_NAME_DEBUG hpx_initd
+    COMPILE_PDB_NAME_RELWITHDEBINFO hpx_init
+    COMPILE_PDB_OUTPUT_DIRECTORY_DEBUG
+      ${CMAKE_CURRENT_BINARY_DIR}/Debug
+    COMPILE_PDB_OUTPUT_DIRECTORY_RELWITHDEBINFO
+      ${CMAKE_CURRENT_BINARY_DIR}/RelWithDebInfo)
+endif()
```
hkaiser: What about the `Release` and `RelMinSize` build types? Should we add those as well?

init/CMakeLists.txt
```diff
@@ -0,0 +1,103 @@
...
+list(APPEND hpx_init_SOURCES
+  hpx_main.cpp
+  hpx_main_argc_argv.cpp
+  hpx_main_variables_map.cpp
+  hpx_main_winsocket.cpp
+  hpx_user_main.cpp
+  hpx_user_main_argc_argv.cpp
+  main.cpp)
+
+if(HPX_WITH_DYNAMIC_HPX_MAIN AND (("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux") OR (APPLE)))
```
hkaiser: Should the `HPX_WITH_DYNAMIC_HPX_MAIN` option be defined in this CMakeLists file?

init/CMakeLists.txt
hkaiser: Please disregard what I said. There will not be any .pdb files to install for those build types.

init/CMakeLists.txt
msimberg: Yeah, why not! I thought I'd leave it with all the other options, but I see no reason not to move it.

msimberg:
I've updated this again. I had to add a separate config file for the init library, otherwise `HPX_HAVE_DYNAMIC_HPX_MAIN` would not be included in the config definitions. Things are working again on Linux, hoping for the same on the other platforms.

msimberg:
The latest commit changed compile tests to be executables rather than static libraries. This is due to to all of them being written with a main function and a few of them even including `hpx_main.hpp` (which caused failures, since `hpx_main.hpp` is now in a separate include directory). I think this is a convention we can stick to.

cmake/HPX_AddCompileTest.cmake
```diff
@@ -18,14 +18,13 @@ function(add_hpx_compile_test category name)
 
   string(REGEX REPLACE "\\." "_" test_name "${category}.${name}")
 
-  add_hpx_library(
+  add_hpx_executable(
```
hkaiser: Isn't this a compile-only test? Wouldn't `add_hpx_executable` attempt to link as well?

hkaiser: Ok, saw your comment just now. I think this is fine as long as this does not cause linker errors.

msimberg: Good point, I hadn't thought of that. I think though that if we do have linker errors they will likely appear elsewhere as well.

msimberg:
I don't know how it was triggered, but an ugly hack that I put in place for the header tests in the compute modules was causing problems. I've removed the hack as it's not needed anymore.

McKillroy:
Built it on windows and except the broken "quickstart/search_by_key_demo" example it worked.

PR #4464, opened 2020-03-26
Include module

msimberg:
A bit of an absurd module, but here goes. Moves all the `hpx/include` headers into the `hpx_include` module. I want to deprecate these headers as well, but will do so only when we have modules for all the headers in this module (things like runtime, parcelport, etc. are still missing).


tests/headers/CMakeLists.txt
```diff
@@ -13,13 +13,7 @@ set(exclude_from_all_headers)
 # These files are either known to expose incomplete types or are known to
 # require special handling to avoid circular #include dependencies.
 set(exclude_from_headers
-  "${PROJECT_SOURCE_DIR}/hpx/include/compression.hpp"
```
hkaiser: So this PR will stop testing the headers from the former `hpx/include` directory. Is that intentional?

msimberg: They're tested as part of the module (hence no need to exclude these two here; they're excluded in the module instead). See https://github.com/STEllAR-GROUP/hpx/blob/264de211ae34a3a784324b52599c5f0b525b4dba/libs/include/tests/CMakeLists.txt#L34-L43.

hkaiser: Ahh, I did miss that! Thanks!

msimberg:
> You are on a module spree ;-) LGTM, thanks!

Well, this one was *particularly* demanding...

msimberg:
The last commit hopefully satisifies the cyclic dependency checker. The collectives module was using one of the headers from this module.

PR #4465, opened 2020-03-27
Add thread_executors module

aurianer:
Move thread executors into a module


msimberg:
For these kinds of failures https://cdash.cscs.ch//viewBuildError.php?buildid=101308, you can either directly include the correct executor header, but you can also add the collective thread executors headers to `hpx/include/parallel_executors.hpp` (and vice versa) for compatibility.

libs/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2019-2020 The STE||AR-Group
+# Copyright (c) 2020 The STE||AR-Group
```
msimberg, line 1: Not a big deal, but we can probably revert this change and have `2019-2020` in `create_library_skeleton.py`.

hpx/runtime/threads/executors.hpp
```diff
@@ -10,11 +10,11 @@
...
-#include <hpx/runtime/threads/executors/current_executor.hpp>
-#include <hpx/runtime/threads/executors/default_executor.hpp>
-#include <hpx/runtime/threads/executors/service_executors.hpp>
-#include <hpx/runtime/threads/executors/this_thread_executors.hpp>
-#include <hpx/runtime/threads/executors/embedded_thread_pool_executors.hpp>
+#include <hpx/thread_executors/current_executor.hpp>
+#include <hpx/thread_executors/default_executor.hpp>
+#include <hpx/thread_executors/service_executors.hpp>
+#include <hpx/thread_executors/this_thread_executors.hpp>
+#include <hpx/thread_executors/embedded_thread_pool_executors.hpp>
```
msimberg, line 17: I think this whole file (`hpx/runtime/threads/executors.hpp`) can move as well, but it can be replaced by the generated module header. You can then point the compatibility header to `hpx/thread_executors.hpp` instead.

hpx/runtime/threads/executors.hpp
```diff
@@ -10,11 +10,11 @@
...
 #include <hpx/runtime/threads/thread_executor.hpp>
```
msimberg, line 18: Can this file (`hpx/runtime/threads/thread_executor.hpp`) be moved as well or does it cause problems? `limiting_executor` and `sequenced_executor` don't actually need it. `thread_execution_information.hpp` and `thread_execution.hpp` probably need it. If they need we can probably move the file to the `execution` module at least. It already contains some of the base functionality of `thread_executors` and will be removed soon anyway.

hpx/runtime/threads/executors.hpp
aurianer: Yep!

hpx/runtime/threads/executors.hpp
aurianer: Ok!

msimberg: Let's keep `hpx/runtime/threads/thread_executor.hpp` where it was: https://176179-4455628-gh.circle-artifacts.com/0/hpx/artifacts/circular_deps.svg. This is not a dependency worth spending time on.

aurianer:
solved :)

PR #4467, opened 2020-03-28
Add mimalloc find cmake script

aurianer:
Add mimalloc cmake support for hpx


msimberg:
mimalloc already comes with CMake support. That is what the old `find_package(mimalloc ...)` was meant to use (https://github.com/microsoft/mimalloc/blob/master/cmake/mimalloc-config.cmake). Does that not work?

aurianer:
It didn't work when I tried it, but I may have missed something cause I did it quickly
Closing this

msimberg:
You might need to set `-Dmimalloc_DIR` to point at the directory of the config file.

PR #4474, opened 2020-03-30
Use #pragma once in headers

### LINKED ISSUE #4194: Consider using #pragma once
This is a small convenience issue. Every now and then it happens that you want to create a new header file but start off by copying an existing header. Then you forget to update the include guards and spend far too long wondering why your definitions in the new header file aren't actually being defined. `#pragma once` would avoid this problem.

It's not standard but practically supported on every platform that we try to support. Is anyone wildly against this? Would this make someone's life a lot better? Don't care?

msimberg:
Fixes #4194. I was bitten once more by not renaming header guards, so I'm changing our headers to use `#pragma once` since there was no huge argument against it. I've also added a CircleCI check to make sure all headers have `#pragma once` in them. There were quite a few headers which didn't have any guards at all (and a couple that already had `#pragma once`).


hkaiser:
@msimberg I was under the impression that we couldn't do that universally because some of our headers were creating circular `#include` dependencies (i.e. `future.hpp`/`dataflow.hpp`). I tip my hat if you found a way around those!

msimberg:
> @msimberg I was under the impression that we couldn't do that universally because some of our headers were creating circular `#include` dependencies (i.e. `future.hpp`/`dataflow.hpp`). I tip my hat if you found a way around those!

I think we've already dealt with those (by not using `dataflow` in awkward places), and as far as I could tell dataflow was the only one. If it still turns out to be a problem we can always leave that as an exception (the CircleCI check just greps for `#pragma once` which means it can be in a comment as well; I've left it out of e.g. the Windows warnings headers).

It seems I wasn't quite successful with my last replacements yet though, but I've had full builds of this working without problems at an earlier stage.

hkaiser:
> the CircleCI check just greps for #pragma once which means it can be in a comment as well

Wouldn't it be better to add it as a inspect check, in this case?

msimberg:
> > the CircleCI check just greps for #pragma once which means it can be in a comment as well
> 
> Wouldn't it be better to add it as a inspect check, in this case?

Yeah, sure! Grep is just so nice ;)

NK-Nikunj:
We call ourselves C++ standards conforming. Why are we using `#pragma once` then?

hkaiser:
> We call ourselves C++ standards conforming. Why are we using #pragma once then?

Well, that's a valid comment and possible rationale _against_ `#pragma once` after all... ;-)

msimberg:
I agree, that's probably the strongest argument against this. However, the way I see our standards conformance is that it's about our implementations (interfaces, behaviour), not that we cannot use nonstandard features. After all, why do we use for example GNU or clang attributes? That's a bit unfair since those can be enabled depending on the compiler. pragma once is supported on all the compilers we currently support. If there is a reasonable need to support a compiler in the future that doesn't support pragma once, I will change those back to normal header guards. With our current set of supported compilers I don't see a reason to stick with what we have.

msimberg:
Let's discuss this at the next meeting then.

msimberg:
For anyone following this issue but not on the mailing list, we decided to have a vote on this. Please respond on the hpx-devel@stellar-group.org mailing list if you'd like to vote.

msimberg:
> Should we actually edit this file at all? I think the initial idea was to be able to drop in updates of this file in the future without too much work. If we decided to start editing this file, we should add a comment at the top outlining what changes need to be applied for future versions in order to make it compatible.

I'm guessing you're referring to the moodycamel header. I think we should, but keep the changes minimal (it already has `clang-format off`). We essentially own this file now, and if someone else also has the same header integrated in their project things will not build correctly if we keep the moodycamel namespace (this is what happened with APEX).

hkaiser:
>> Should we actually edit this file at all? I think the initial idea was to be able to drop in updates of this file in the future without too much work. If we decided to start editing this file, we should add a comment at the top outlining what changes need to be applied for future versions in order to make it compatible.
> 
> I'm guessing you're referring to the moodycamel header. I think we should, but keep the changes minimal (it already has clang-format off). We essentially own this file now, and if someone else also has the same header integrated in their project things will not build correctly if we keep the moodycamel namespace (this is what happened with APEX).

I'm fine with that, and I agree to the rationale. Let's add that comment then and go on with our lives.

msimberg:
@khuck this PR currently breaks the APEX build because I've moved the moodycamel queue to the `hpx::concurrency` namespace. I can go ahead and make a PR to change the namespace for you in APEX, or revert your commit(s) which got rid of your version of the moodycamel queue. Up to you :)

khuck:
> @khuck this PR currently breaks the APEX build because I've moved the moodycamel queue to the `hpx::concurrency` namespace. I can go ahead and make a PR to change the namespace for you in APEX, or revert your commit(s) which got rid of your version of the moodycamel queue. Up to you :)

I'll patch APEX.  What do I need to do?  If you want to make a PR for APEX, that would be excellent, I don't imagine it's many changes.

msimberg:
> I'll patch APEX. What do I need to do? If you want to make a PR for APEX, that would be excellent, I don't imagine it's many changes.

Yep, the change is small: https://github.com/khuck/xpress-apex/pull/121. Will you tag a new version after that change?

khuck:
> > I'll patch APEX. What do I need to do? If you want to make a PR for APEX, that would be excellent, I don't imagine it's many changes.
> 
> Yep, the change is small: [khuck/xpress-apex#121](https://github.com/khuck/xpress-apex/pull/121). Will you tag a new version after that change?

Done, new tag is v2.1.9

msimberg:
> Done, new tag is v2.1.9

Thanks!

I've updated the tag in this PR. Assuming this builds correctly with APEX now this PR is ready to go. I'm ignoring the one typo reported by codespell (referer) since it's a URL, but I'm also not adding it to the whitelist since I consider it a real typo.

msimberg:
Any further comments on this?

PR #4475, opened 2020-03-30
Fix rp init changes

aurianer:
- Fixes #4388 
- Fixes some the handling of `HPX_APPLICATION_STRING` in `hpx_wrap` (the name of the application using `hpx_main.hpp` were set to "Unknown HPX application" default value)
- Makes `hpx_wrap` use the the `hpx::init` with the new init struct


aurianer:
I've broken some things.. working on it

hpx/hpx_init.hpp
```diff
@@ -193,7 +193,7 @@ namespace hpx
 #if !defined(HPX_HAVE_STATIC_LINKING)
     inline
 #endif
-    int init(init_params const& params = init_params());
+    int init(init_params& params);
```
msimberg: General comment on all these overloads. The last argument was optional because it gives us essentially double the overloads for free. Whatever the solution ends up being, we still want the following to work: `hpx::init()`, `hpx::init(argc, argv)`. The last argument can remain optional, but you won't be able to use `init_params()` as the default value like that anymore. It has to be a static again.

hpx/hpx_init_params.hpp
```diff
@@ -59,13 +59,15 @@ namespace hpx
...
         HPX_EXPORT void on_abort(int signal) noexcept;
         // Default params to initialize the init_params struct
-        extern HPX_EXPORT const hpx::program_options::options_description
-            default_desc;
         extern HPX_EXPORT startup_function_type default_startup;
         extern HPX_EXPORT shutdown_function_type default_shutdown;
         HPX_MAYBE_UNUSED extern HPX_EXPORT int dummy_argc;
         // TODO: make it only one parameter, probably add a cast
         HPX_MAYBE_UNUSED extern HPX_EXPORT char **dummy_argv;
+        static char app_name[] = HPX_APPLICATION_STRING;
```
msimberg: Let's maybe add a comment here to say why this one isn't in the cpp file.

init/src/hpx_wrap.cpp
```diff
@@ -23,14 +25,14 @@ namespace hpx_start
     // by this code.
     HPX_SYMBOL_EXPORT extern bool include_libhpx_wrap;
     HPX_SYMBOL_EXPORT bool include_libhpx_wrap __attribute__((weak)) = false;
+    HPX_SYMBOL_EXPORT extern std::string app_name_libhpx_wrap;
```
msimberg: This most likely has to be weak as well. If `hpx_main.hpp` isn't included by the application the name will never be defined, but that's ok since it also won't be used.

hpx/hpx_init.hpp
hkaiser: Hmm, why do you want to pass the argument as a non-const reference? That would imply that the user supplied data will be changed by the function. Is that what we desire?

hpx/hpx_init_params.hpp
hkaiser: I think this could even be `constexpr` nowadays.

init/CMakeLists.txt
```diff
@@ -77,6 +77,9 @@ target_link_libraries(hpx_init PUBLIC hpx)
 target_link_libraries(hpx_init PRIVATE hpx_private_flags)
 target_link_libraries(hpx_init PRIVATE hpx::boost)
 target_compile_definitions(hpx_init PRIVATE HPX_APPLICATION_EXPORTS)
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+  target_compile_options(hpx_init PRIVATE -U_ZN9hpx_start20app_name_libhpx_wrapE)
+endif()
```
hkaiser: Uhh, really? A mangled name in the build system? This doesn't look right to me.

init/include/hpx/hpx_main.hpp
```diff
@@ -18,13 +18,17 @@
...
+
 namespace hpx_start
 {
     // include_libhpx_wrap here is an override for the one present in
     // src/hpx_wrap.cpp. The value of this variable defines if we need
     // to change the program's entry point or not.
     HPX_SYMBOL_EXPORT extern bool include_libhpx_wrap;
     HPX_SYMBOL_EXPORT bool include_libhpx_wrap = true;
+    HPX_SYMBOL_EXPORT extern std::string app_name_libhpx_wrap;
+    HPX_SYMBOL_EXPORT std::string app_name_libhpx_wrap = HPX_APPLICATION_STRING;
```
hkaiser: Let's make this `constexpr` instead:
```
constexpr char const app_name_libhpx_wrap[] = HPX_APPLICATION_STRING;
```

init/CMakeLists.txt
aurianer: Any suggestion ? I'm out of ideas..

init/include/hpx/hpx_main.hpp
aurianer: I'm not fully understanding what's going on with the hpx_wrap thing but when I just put the line you suggested, `app_name_libhpx_wrap` is then undefined in `src/hpx_wrap.cpp`

The problem is that the for certain tests which use hpx_wrap, the `HPX_APPLICATION_STRING` is defined in `hpx/hpx_main.hpp` and set to "Unknown HPX application", my goal was to use the same method with `__attribute__((weak))` symbols to make it overwrite with the right application name but the `__attribute__((weak))` trick seems to work only with gcc, I really need help on that one..

hpx/hpx_init.hpp
aurianer: Agreed, I've changed that as soon as I added the mutable for `startup` and `shutdown`, thanks !

hpx/hpx_init_params.hpp
aurianer: Actually I get this when I try to make it a constexpr:
```
/src/hpx_init_params.cpp:18:31:error: cannot initialize an array element of type 'char *' with an lvalue of type 'const char [24]'
    char *default_argv[2] = { app_name, nullptr };
                              ^~~~~~~~
```

src/hpx_init_params.cpp
```diff
@@ -11,14 +11,10 @@
...
-
-    const hpx::program_options::options_description default_desc(
-                "Usage: " HPX_APPLICATION_STRING " [options]");
     startup_function_type default_startup = startup_function_type();
     shutdown_function_type default_shutdown = shutdown_function_type();
     HPX_MAYBE_UNUSED int dummy_argc = 1;
     // TODO: make it only one parameter, probably add a cast
-    char app_name[] = HPX_APPLICATION_STRING;
     char *default_argv[2] = { app_name, nullptr };
     HPX_MAYBE_UNUSED char **dummy_argv = default_argv;
```
msimberg, line 19: Btw, `app_name` is also wrong in `default_argv`. Not many people would probably notice in their applications. And the `default_startup/shutdown` functions aren't needed anymore. So maybe we just move these back into the overloads that need them. The duplication will be mostly removed when the deprecated overloads are gone.

aurianer: On it !

msimberg:
The `gcc-newest` failures are my fault. I disabled logging on that builder which results in unused variable warnings (errors with `-Werror`).

PR #4478, opened 2020-03-30
Async modules

msimberg:
Adds three new modules:
- `async_combinators`: contains all the `when_*` and `wait_*` functions and `split_future`
- `local_async`: `async`, `apply`, and `dataflow` for local tasks
- `async`: all the distributed overloads of the above.

I'm not 100% sure about the names, but I think I could live with them. We end up with funny headers like `hpx/async/async.hpp`, but I chose `asyncy` as the module name because it's the main functionality exposed by the modules (I consider `apply` and `dataflow` special cases).

@hkaiser I've moved all the applier files into the `async` module. Does that make sense to you or make life more difficult for later refactorings?

In `dataflow.hpp` I managed in some way to split out the distributed overloads and specializations into a separate file but ended up with an ugly two-level dispatch (`async_dispatch` and `async_dispatch_impl`). Can anyone think of a cleaner way to implement the dispatch over there?

- [x] use new headers (CircleCI will fail currently)


libs/async/include/hpx/async/async_callback.hpp
```diff
@@ -0,0 +1,276 @@
...
+namespace hpx {
+    template <typename F, typename... Ts>
+    HPX_FORCEINLINE auto async_cb(F&& f, Ts&&... ts) -> decltype(
+        detail::async_cb_dispatch<typename util::decay<F>::type>::call(
+            std::forward<F>(f), std::forward<Ts>(ts)...))
+    {
+        return detail::async_cb_dispatch<typename util::decay<F>::type>::call(
+            std::forward<F>(f), std::forward<Ts>(ts)...);
+    }
+}    // namespace hpx
```
hkaiser: Does this API function belong into the `local_async` module?

msimberg: I'm not sure. `async_cb_dispatch` only has `call` overloads for actions (AFAICT). Is it meant to have local dispatch points as well? We can of course still move `async_cb` itself into the `local_async` module in case `async_cb` for local tasks ever gets implemented.

hkaiser: Fair points. I didn't look at the dispatch points, so I think we're fine with what you propose.

msimberg:
This should be close to ready now. The only thorn is the `guided_pool_executor` which still uses `dataflow`. I've for now moved it to the `local_async` module (include path unchanged). @biddisco do you see a way to not use `dataflow` in your executor? Considering it's using `sync`, you might be able to get by with not the full `dataflow` but only part of its functionality?

I also added a bit of documentation on getting more information on circular dependencies that I found useful while making this.

hkaiser:
> This should be close to ready now. The only thorn is the guided_pool_executor which still uses dataflow. I've for now moved it to the local_async module (include path unchanged). @biddisco do you see a way to not use dataflow in your executor? Considering it's using sync, you might be able to get by with not the full dataflow but only part of its functionality?

The circular dependencies are created by dataflow depending on executors, I presume (I have not checked). If this is so, I think we will have to separate the executor specializations from dataflow in the same way as we have done for `async` and `future`. 

> I also added a bit of documentation on getting more information on circular dependencies that I found useful while making this.

👍 thanks! You might want to add a link to where to actually find the `cpp-depenencies` tool, however.

msimberg:
> > This should be close to ready now. The only thorn is the guided_pool_executor which still uses dataflow. I've for now moved it to the local_async module (include path unchanged). @biddisco do you see a way to not use dataflow in your executor? Considering it's using sync, you might be able to get by with not the full dataflow but only part of its functionality?
> 
> The circular dependencies are created by dataflow depending on executors, I presume (I have not checked). If this is so, I think we will have to separate the executor specializations from dataflow in the same way as we have done for `async` and `future`.

Yes, the local dataflow specializations depend on executors, as do async, apply, and sync. For some reason I'd prefer to not depend on dataflow in executors if possible, but I can't really say why. What you're suggesting definitely sounds doable though. That would effectively move the whole current `local_async` module into the execution (or executors, once that's there) module, except for the dataflow/async/etc. declarations along with the dispatch declarations (async_base module?). Does that sound like what you had in mind?

> > I also added a bit of documentation on getting more information on circular dependencies that I found useful while making this.
> 
> +1 thanks! You might want to add a link to where to actually find the `cpp-depenencies` tool, however.

Good point, added!

msimberg:
@hkaiser I've updated this now. All of the meat of local async etc. is in `hpx/execution/{apply,async,dataflow,sync}.hpp`. I had actually left the launch policy specializations in the (distributed) async module first by mistake. I've kept the `local_async` module with forwarding headers to `execution` just for symmetry with `async`, but it is of course redundant. I like it but I don't have to keep it around.

msimberg:
I'll merge this one after the two executors modules (#4494 and #4465).

libs/async_base/include/hpx/async_base/sync.hpp
```diff
@@ -0,0 +1,32 @@
...
+#if !defined(HPX_ASYNC_BASE_SYNC_JUL_21_2018_0937PM)
+#define HPX_ASYNC_BASE_SYNC_JUL_21_2018_0937PM
+
+#include <hpx/config.hpp>
+#include <hpx/type_support/decay.hpp>
+
+#include <utility>
+
+namespace hpx { namespace detail {
+    // dispatch point used for async implementations
```
hkaiser, line 16: Did you mean `sync implementations`?

libs/execution/include/hpx/execution/apply.hpp
```diff
@@ -0,0 +1,65 @@
...
+    struct apply_dispatch
+    {
+        template <typename F, typename... Ts>
+        HPX_FORCEINLINE static typename std::enable_if<
+            traits::detail::is_deferred_invocable<F, Ts...>::value, bool>::type
+        call(F&& f, Ts&&... ts)
+        {
+            parallel::execution::parallel_executor exec;
+            parallel::execution::post(
+                exec, std::forward<F>(f), std::forward<Ts>(ts)...);
```
hkaiser, line 37: I think we know that `parallel_executor` implements `post`, thus we can replace this with:
```suggestion
            exec.post(std::forward<F>(f), std::forward<Ts>(ts)...);
```
There is no reason to go through the customization point again.

libs/execution/include/hpx/execution/dataflow.hpp
```diff
@@ -192,21 +174,23 @@ namespace hpx { namespace lcos { namespace detail
...
             parallel::execution::parallel_policy_executor<launch::async_policy>
                 exec{policy};
-            parallel::execution::post(exec,
+            parallel::execution::post(
+                exec,
                 [this_ = std::move(this_)](Futures&& futures) -> void {
                     return this_->done(std::move(futures));
-                }, std::move(futures));
+                },
+                std::move(futures));
```
hkaiser, line 182: Same here, since we know that `parallel_policy_executor<launch::async_policy>` directly implements a `post` method, we can save us going through the customization point and call it directly on the executor instead.

libs/execution/include/hpx/execution/dataflow.hpp
```diff
@@ -242,10 +226,12 @@ namespace hpx { namespace lcos { namespace detail
...
             parallel::execution::parallel_policy_executor<launch::fork_policy>
                 exec{policy};
-            parallel::execution::post(exec,
+            parallel::execution::post(
+                exec,
                 [this_ = std::move(this_)](Futures&& futures) -> void {
                     return this_->done(std::move(futures));
-                }, std::move(futures));
+                },
+                std::move(futures));
```
hkaiser, line 234: Ditto here, let's call `post` directly.

msimberg:
Grr, sorry @hkaiser! I missed your most recent comments. I'll take care of them in another PR.

hkaiser:
> Grr, sorry @hkaiser! I missed your most recent comments. I'll take care of them in another PR.

No worries, those were mostly unrelated anyways. Thanks!

PR #4479, opened 2020-03-30
Add hints to allow to specify MIMALLOC_ROOT

aurianer:
Add hints for consistency so that we can specify `MIMALLOC_ROOT` variable as for other libraries


aurianer:
Closing this cause cmake is case sensitive and `mimalloc_ROOT` works perfectly fine !

Issue #4482, opened 2020-03-31
Use CMake CUDA language support


msimberg:
Since we now require CMake 3.13 we could potentially significantly clean up our CUDA support (https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html, `enable_language(CUDA)`). This would be a good task for someone familiar (or willing to get familiar) with CMake's CUDA support. The only caveat is that we'd have to make sure that clang CUDA compilation still works normally.

msimberg:
For libraries that only need the CUDA runtime libraries we should be using https://cmake.org/cmake/help/v3.17/module/FindCUDAToolkit.html#module:FindCUDAToolkit and the `CUDA::cudart` target (CMake 3.17 onwards).

msimberg:
In progress in #5162.

PR #4483, opened 2020-03-31
Clean up imported CMake target naming

### LINKED ISSUE #4476: Use CMake targets defined by FindBoost
There's no need for us (anymore) to define our own interface targets for Boost. Since we've bumped the minimum CMake version we can now rely on those targets being there and it saves us quite a bit of boilerplate.

msimberg:
For most imported targets I've followed the `Dependency::dependency` naming convention. I've also moved the target creation to the `Find*.cmake` files. For the allocators I've named the target `hpx_dependencies_allocator`. It has no namespace just like our other internal targets (e.g. `hpx_private_flags`). The old `hpx::boost` target is now called `hpx_dependencies_boost` as it's also an internal target and depends on multiple Boost targets. In other cases I've simply replaced the `hpx::boost::*` targets by `Boost::*`.

~The only bigger change I did was to change the MPI handling. It now links to the target `MPI::MPI_CXX` (provided by CMake's `FindMPI.cmake`). I haven't actually tried this though.~ I dropped the MPI changes (except for renaming). I've opened #4484 to track that instead.

For cleaning up CUDA handling I've opened #4482.

All of this will be slightly simpler as well once we can propagate dependencies correctly from modules (#4135).

Fixes #4476. Fixes #4483.


msimberg:
Let's see if this passes tests now. It turns out our pkgconfig file generation did not handle imported targets (not the same as interface targets, apparently). I've added code to handle it at least for Boost.

msimberg:
This looks ready to go in.

PR #4486, opened 2020-04-01
Fixing an initialization order problem

hkaiser:


hkaiser:
> Looks sane enough! How did the old code manifest itself as a problem? Anything we can do to test it?

The old code worked (for me) by chance. Initialization order of globals is generally undefined. The old initialization sequence was most probably changed because of the modularization related reshuffling. Could be platform specific as well... I ran into issues now manifesting themselves by strange segfaults during startup (initialization) pointing to the vector I wrapped into a function local static now.

msimberg:
Sigh... filesystem problems again on Piz Daint. This looks very safe to merge though based on the other results (and the code), so I'll do so.

PR #4487, opened 2020-04-01
Properly dispatch exceptions thrown from hpx_main to be rethrown from hpx::init/hpx::stop

hkaiser:


msimberg:
Hmm, this is throwing more/in a different place than some of our tests expect. We do have some tests that catch an exception thrown in `hpx_main` (`error_callback` is the simplest). What sort of exceptions were not propagated up through init/main?

msimberg:
`report_error` is mean to store exceptions thrown, which the runtime then rethrows after everything has been shut down. Could it be that `report_error` is not getting called, or encounters an error itself (did the application hang?)?

hkaiser:
> Hmm, this is throwing more/in a different place than some of our tests expect. We do have some tests that catch an exception thrown in hpx_main (error_callback is the simplest). What sort of exceptions were not propagated up through init/main?

I'll investigate.

msimberg:
@hkaiser I've attempted to clean up the thread number handling. `local_num` is allowed to be -1, for example when creating a thread from a non-worker thread (like spawning `hpx_main`/`run_helper` as was the case here). @biddisco could you have a look to see if the changes to the `shared_priority_queue_scheduler` make sense?

hkaiser:
> @hkaiser I've attempted to clean up the thread number handling. local_num is allowed to be -1, for example when creating a thread from a non-worker thread (like spawning hpx_main/run_helper as was the case here).

Thanks @msimberg. I did suspect that much, but was not sure what the correct handling in the scheduler should be. Let's see how the tests fare.

PR #4490, opened 2020-04-02
Update GitHub actions configs

msimberg:
- Try to trigger builds over here for pull requests.
- Try to use our container image for linux builds.


msimberg:
Pff... `No space left on device` on the Linux builder. I'll reduce it to also only build examples. We still have CircleCI and pycicle for building everything. At least GitHub actions gives us very quick feedback on basic problems in pull requests.

PR #4491, opened 2020-04-02
Add missing current_executor header

aurianer:
Fixes master
3 headers added in the CMakeLists.txt: `limiting_executor`, `guided_pool_executor`, `current_executor`


msimberg:
We could by the way consider adding some sort of inspect check to make sure those file lists are up to date. CMake will complain if there are superfluous files, and a simple grep + find would probably be enough to find missing files. That's outside of this PR though.

PR #4493, opened 2020-04-03
Pack traversal module

msimberg:
Includes `pack_traversal` and `unwrap`. Technically `pack_traversal` would not need to be together with `unwrap`. The latter depends on the former. Since they're practically only used together I put them in the same module. If someone thinks they should be separate I can do that as well.


libs/pack_traversal/src/pack_traversal.cpp
```diff
@@ -5,30 +5,30 @@
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
```
hkaiser, line 5: Shouldn't this file go into `.../tests/unit`?

libs/performance_counters/include/hpx/performance_counters/performance_counter_set.hpp
```diff
@@ -12,10 +12,10 @@
 #include <hpx/errors.hpp>
 #include <hpx/lcos/dataflow.hpp>
 #include <hpx/lcos/future.hpp>
+#include <hpx/pack_traversal/unwrap.hpp>
```
hkaiser, line 15: This module now depends on pack_traversal

libs/resource_partitioner/examples/async_customization.cpp
```diff
@@ -26,8 +26,8 @@
 #include <hpx/functional/result_of.hpp>
 #include <hpx/lcos/dataflow.hpp>
 #include <hpx/lcos/when_all.hpp>
+#include <hpx/pack_traversal/pack_traversal.hpp>
```
hkaiser, line 29: This module now depends on pack_traversal

libs/segmented_algorithms/include/hpx/parallel/segmented_algorithms/detail/scan.hpp
```diff
@@ -15,10 +15,10 @@
 #include <hpx/iterator_support/traits/is_iterator.hpp>
 #include <hpx/lcos/dataflow.hpp>
 #include <hpx/lcos/future.hpp>
+#include <hpx/pack_traversal/unwrap.hpp>
```
hkaiser, line 18: This module now depends on pack_traversal

libs/pack_traversal/src/pack_traversal.cpp
msimberg: ...

Well done me. Thanks @hkaiser!

libs/performance_counters/include/hpx/performance_counters/performance_counter_set.hpp
msimberg: This is another thing that we could check in CI (#4497).

aurianer:
LGTM thanks!

hkaiser:
@aurianer Looks like the documentation refers to one of the headers that were moved here.

msimberg:
> @aurianer Looks like the documentation refers to one of the headers that were moved here.

@aurianer is innocent! It is the docs file list that's wrong, and I thought I fixed it... Will check again.

msimberg:
All the CircleCI checks looked clean now. I've just rebased because of conflicts. Let's see if everything is still ok.

PR #4494, opened 2020-04-03
Module executors

aurianer:
Move the executors from the execution module to the executors module


libs/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2019-2020 The STE||AR-Group
+# Copyright (c) 2020 The STE||AR-Group
```
msimberg, line 1: Let's keep this as it is as well.

libs/executors/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+  hpx/execution/executors.hpp
+  hpx/execution/executors/limiting_executor.hpp
+  hpx/execution/executors/parallel_executor_aggregated.hpp
+  hpx/execution/executors/parallel_executor.hpp
+  hpx/execution/executors/pool_executor.hpp
+  hpx/execution/executors/restricted_thread_pool_executor.hpp
+  hpx/execution/executors/sequenced_executor.hpp
+  hpx/execution/executors/service_executors.hpp
+  hpx/execution/executors/thread_pool_attached_executors.hpp
+  hpx/execution/executors/timed_executors.hpp
```
msimberg, line 25: Should we go with just `hpx/executors/*_executor.hpp` now? This seems like it'll be the final stop for the executors.

hkaiser: I'd support that.

libs/CMakeLists.txt
aurianer: Yes I revert it on my other PR but I guess the script is actually changed on master, thanks!

libs/executors/CMakeLists.txt
aurianer: Yep I'll do that, I just have to fix the current circular dependency :)

aurianer:
> LGTM, thanks!
> 
> Do other modules now depend on this new module instead of the already existing `execution` 

I changed some of them but I'll wait for the thread_executors module PR to be merged so I can be sure all is in place

libs/execution/include/hpx/execution/execution_policy_fwd.hpp
```diff
@@ -15,22 +15,26 @@
...
-    struct parallel_policy;
+    template <typename Executor>
+    struct parallel_policy_impl;
 
     template <typename Executor, typename Parameters>
     struct parallel_policy_shim;
 
-    struct parallel_task_policy;
+    template <typename Executor>
+    struct parallel_task_policy_impl;
```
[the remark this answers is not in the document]
hkaiser: The alias would work perfectly well. Thanks!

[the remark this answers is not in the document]
aurianer: Update: dropped the template changes as in some cases the `parallel_policy_impl` template was instanciated  with `block_executor` which didn't have a default `constexpr` constructor
I just moved the execution_policy files in the executors module for now, we may create an `execution_policies` in the future

libs/execution/include/hpx/execution/executors/timed_executors.hpp
```diff
@@ -90,7 +84,7 @@ namespace hpx { namespace parallel { namespace execution {
                         .get())
             {
                 auto predecessor = make_ready_future_at(abs_time);
-                return execution::then_execute(sequenced_executor(),
+                return execution::then_execute(exec,
```
msimberg, line 87: The old code was as far as I can tell an optimization (reduces the number of tasks spawned). The new behaviour works, but leads to changes in the counts here: https://github.com/STEllAR-GROUP/hpx/blob/9600265fab389b32a8966fe3a7c476f832ec6fe1/libs/execution/tests/unit/minimal_timed_async_executor.cpp#L276-L281. That's in principle harmless. Does someone think this change is going badly harm performance? An alternative would be to move all of the timed execution functionality yet another separate module, which could depend on the new `executors` module. Thoughts?

msimberg: There was one other test that broke as a result of this change: `suspend_thread_timed`. The fix is explained in the [commit message](https://github.com/STEllAR-GROUP/hpx/pull/4494/commits/627ce6107fccbdf9f8acdff2ce0f1113ab2e44e7). I'm not convinced there isn't a better way to fix this though. @hkaiser any bright ideas?

hkaiser: > There was one other test that broke as a result of this change: suspend_thread_timed. The fix is explained in the commit message. I'm not convinced there isn't a better way to fix this though. @hkaiser any bright ideas?

@msimberg this commit does not exist anymore. Could you copy the message here for posterity, please?

libs/execution/include/hpx/execution/exception_list.hpp
```diff
@@ -15,11 +15,30 @@
...
+    template <typename Executor, typename Parameters>
+    struct parallel_task_policy_shim;
+
+    struct parallel_unsequenced_policy;
+
+    inline namespace v1 {
+        struct dataseq_task_policy;
+    }   // namespace v1
+
+}}}  // namespace hpx::parallel::execution
```
hkaiser, line 40: Does this belong into this header?

libs/execution/include/hpx/execution/executors/thread_execution.hpp
```diff
@@ -205,23 +204,16 @@ namespace hpx { namespace threads {
...
                 std::forward<Future>(predecessor), std::forward<Executor>(exec),
-                [func = std::move(func),
-                    exec_current = std::move(exec_current)](
-                    future_type&& predecessor) mutable -> result_future_type {
-                    return hpx::parallel::execution::async_execute(exec_current,
-                        hpx::util::functional::unwrap{},
-                        func(std::move(predecessor)));
+                [func = std::move(func)](
+                    future_type&& predecessor) mutable -> vector_result_type {
+                    return hpx::util::unwrap(func(std::move(predecessor)));
```
hkaiser, line 216: This possibly changes the thread that executes the lambda. Are we sure we want that?

libs/timed_execution/include/hpx/timed_execution/timed_executors.hpp
```diff
@@ -407,55 +405,6 @@ namespace hpx { namespace parallel { namespace execution {
...
-      : std::true_type
-    {
-    };
-
-    template <>
-    struct is_two_way_executor<parallel::execution::parallel_timed_executor>
-      : std::true_type
-    {
-    };
-    /// \endcond
```
hkaiser, line 456: Has this been moved some where else?

libs/execution/include/hpx/execution/executors/timed_executors.hpp
msimberg: Yeah, sorry about that. I went back to move the timed functionality into a separate module. I've archived the old state of the branch [here](https://github.com/msimberg/hpx/commits/module_executors_timed_changes), [this](https://github.com/msimberg/hpx/commit/ae699e435811b91e7bb1dc69f9b8337f88f03a7e) is the commit, and just in case I wipe that branch the commit message was:
```
Changes to timed_executor to eagerly create timed threads lead to the
suspend_thread_timed segfaulting for schedulers that don't support suspension.
This happened because when an exception is thrown trying to suspend a worker
thread, the thread pool gets shut down immediately, but the timed threads are
still waiting to spawn. When they wake up and try to schedule a thread on the
thread pool the thread pool is gone and the test segfaults.
```

msimberg: I'll mark this as resolved now since the change isn't relevant anymore.

libs/timed_execution/include/hpx/timed_execution/timed_executors.hpp
msimberg: This part is in progress, sorry... This functionality will end up in `timed_executors`.

libs/execution/include/hpx/execution/executors/thread_execution.hpp
msimberg: I'm aware. I'm going by the tests and they seem to accept this. Since this is deprecated functionality I did not investigate this further. Again, this functionality could move to the `thread_executors` module, allowing us to keep the old behaviour

libs/execution/include/hpx/execution/exception_list.hpp
msimberg: Most likely not. I'll see if it can be removed.

libs/timed_execution/include/hpx/timed_execution/timed_executors.hpp
msimberg: This has been updated.

libs/execution/include/hpx/execution/exception_list.hpp
msimberg: Fixed.

msimberg:
This should now have all the pieces roughtly in place (thanks @aurianer for those module checking scripts, they're amazing!). I suspect not everything will build yet though, so still going to do a round of checking that.

libs/execution/include/hpx/execution/executors/thread_execution.hpp
msimberg: I take that back. The `thread_pool_os_executors` test occasionally fails after this change. I've reverted the change and moved all remaining thread executor related files to `thread_executors`.

Note that the changes in this PR modules do slightly break compatibility (so much for that...). In 1.4.1 one could include `hpx/parallel/execution.hpp` and get all the execution related headers (executors, execution, thread execution, and timed execution). This is not possible anymore (well, we could keep that true, but have chosen not to). The catch-all header `hpx/include/parallel_execution.hpp` is immune against these changes. I think these changes are low risk though as most users should be on the newer executors in `hpx::parallel::execution` and not `hpx::threads::executors`. Timed execution I would guess is rare.

msimberg:
This should be ready to go now. The CircleCI failures are because of the documentation...

PR #3427, opened 2018-08-22
Fixes #3426

NK-Nikunj:
Fixes #3426 

Changes the `extern bool` to actual bool variable in `hpx_init.cpp`. Turns out Mac OSX works differently for `weak` variables as compared to Linux. In Linux, if a variable is declared `extern weak` then it allows to compile and link without producing any errors. For Mac OSX, however, it produces errors while linking (it does compile without errors though).

I do not understand as to why we didn't get such an error before while merging the Mac OSX implementation. As of now, this PR should fix things.

**PS.** Tested with Mac OSX Sierra and Ubuntu Budgie 18.04.

PR #4499, opened 2020-04-06
Attempting to fix hpx_init linking on macOS

msimberg:
There were still some linking errors on macOS that we didn't notice with the previous PR (#4475). This seems to fix them, but I don't know yet if I've broken Linux in return.

### DIFF
--- a/init/src/hpx_wrap.cpp
+++ b/init/src/hpx_wrap.cpp
@@ -25,8 +25,8 @@ namespace hpx_start
     // by this code.
     HPX_SYMBOL_EXPORT extern bool include_libhpx_wrap;
     HPX_SYMBOL_EXPORT bool include_libhpx_wrap __attribute__((weak)) = false;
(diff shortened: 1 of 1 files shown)


init/src/hpx_wrap.cpp
```diff
@@ -25,8 +25,8 @@ namespace hpx_start
     // by this code.
     HPX_SYMBOL_EXPORT extern bool include_libhpx_wrap;
     HPX_SYMBOL_EXPORT bool include_libhpx_wrap __attribute__((weak)) = false;
-    HPX_SYMBOL_EXPORT extern std::string app_name_libhpx_wrap
-        __attribute__((weak));
```
NK-Nikunj, line 29: extern and weak usually don't go together in macOS (see #3427). That's a good catch.

aurianer:
Thanks for fixing this !

Issue #4497, opened 2020-04-06
Add modules CI checks


msimberg:
We can add the following checks:

- All header files need to be listed in `CMakeLists.txt`.
- All direct dependencies should be listed in `CMakeLists.txt` (and conversely no transitive dependencies need to be listed), this should be doable when module headers follow the `hpx/<modulename>` naming convention.

hkaiser:
@aurianer I think we should list all direct dependencies (i.e. modules for which a header is included) even if those would be pulled in transitively.

aurianer:
I agree but I prefer to wait until the modules are pretty much in place, because some of the modules have complicated dependencies just because other modules are not yet formed, I'll update the script as soon as they are all approximately in place :)

msimberg:
@aurianer this is all implemented in the current checks, right? (Except for the "no transitive dependencies should be listed"-check, but that is trickier anyway...)

PR #4503, opened 2020-04-08
Add CMakelists file check

aurianer:
Fixes partially #4497

Check if the files in the module are listed as sources and headers in the  `CMakeLists.txt` of the respective module


msimberg:
Thanks for doing this! Let's see how many are missing...

.circleci/config.yml
```diff
@@ -117,6 +117,21 @@ jobs:
...
+  # Check if the files of the module are listed in the CMakeLists.txt
+  check_cmakelists_files:
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: /hpx
+      - run:
+          name: Check if the headers of the modules are listed in their CMakeLists.txt
+          command: |
+              ./tools/check_cmakelists_files.sh > /tmp/missing_files.txt
```
hkaiser, line 129: This should be `/hpx/source/tools/...` instead, I believe:
```suggestion
              /hpx/source/tools/check_cmakelists_files.sh > /tmp/missing_files.txt
```

aurianer: Oh yeah sorry I was testing it locallly, put a source_dir variable to change it and forgot to use it ^^ thanks !

aurianer:
> Thanks for doing this! Let's see how many are missing...

I saw around 10 when doing it locally but I haven't fix them to see if the circleci step is working correctly ^^

aurianer:
I've added the script to check the module's dependencies, for now it only checks if the global module header is included (i.e. `hpx/<module_name>.hpp`)
However the code is similar for both scripts so I still have to refactor these and do only one circleci step outputing in 2 different files whether it is a missing file or a missing dependency

hkaiser:
@aurianer This is great! Thanks!

hkaiser:
> However the code is similar for both scripts so I still have to refactor these and do only one circleci step outputing in 2 different files whether it is a missing file or a missing dependency

I think it might be better/easier to keep the scripts separate. They are simple enough.

aurianer:
> > However the code is similar for both scripts so I still have to refactor these and do only one circleci step outputing in 2 different files whether it is a missing file or a missing dependency
> 
> I think it might be better/easier to keep the scripts separate. They are simple enough.

The thing is that we are iterating over the modules directory twice with 2 separate scripts and in addition factorizing using function produces a less global variables, but I can understand it is maybe easier to maintain so you choose (see the factorization in the last commit compared to the one before)

hkaiser:
> The thing is that we are iterating over the modules directory twice with 2 separate scripts and in addition factorizing using function produces a less global variables, but I can understand it is maybe easier to maintain so you choose (see the factorization in the last commit compared to the one before)

Sure, no problem. Let's go with the refactored script.

libs/compute_cuda/CMakeLists.txt
```diff
@@ -26,6 +26,9 @@ if(HPX_WITH_CUDA)
     hpx/compute/cuda/traits/access_target.hpp
     hpx/compute/cuda/transfer.hpp
     hpx/compute/cuda/value_proxy.hpp
+    hpx/compute/detail/get_proxy_type.hpp
+    hpx/compute/detail/iterator.hpp
+    hpx/compute/detail/target_distribution_policy.hpp
```
hkaiser, line 31: I think we can (generally) exclude headers in `/detail` from the generated global module header.

libs/plugin/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ set(plugin_headers
   hpx/plugin/virtual_constructor.hpp
   hpx/plugin/detail/dll_dlopen.hpp
   hpx/plugin/detail/dll_dlopen.hpp
```
hkaiser, line 23: This header is listed twice. Should we add a check for that?

aurianer: I feel like it is visually easily detectable but I can add a check for that if you wish, what are the consequences in visual studio if one of the file is listed twice ?

libs/compute_cuda/CMakeLists.txt
aurianer: Should I remove all the `detail/` files I added ? In some module they are already there, should I remove them too?
I'm not entirely sure what are those headers for in visual studio but when a header is just forward declaring some objects and including the detail file for the main implementation, don't you want to be able to access the detail file in visual studio too ?

libs/plugin/CMakeLists.txt
hkaiser: No consequences, so no need to add a check. I was merely asking whether this is worth doing.

libs/compute_cuda/CMakeLists.txt
hkaiser: The `/detail` headers should be listed in the headers list in any case. I think we should exclude them from the generated module header, though.

aurianer: yes sorry for misunderstanding your comment!

libs/plugin/CMakeLists.txt
aurianer: Yep I'll do that but probably later in another PR

libs/compute_cuda/CMakeLists.txt
```diff
@@ -40,12 +43,19 @@ include(HPX_AddModule)
 add_hpx_module(compute_cuda
   FORCE_LINKING_GEN
   GLOBAL_HEADER_GEN OFF
+  EXCLUDE_FROM_GLOBAL_HEADER
+    hpx/compute/cuda/detail/launch.hpp
+    hpx/compute/cuda/detail/scoped_active_target.hpp
+    hpx/compute/detail/get_proxy_type.hpp
+    hpx/compute/detail/iterator.hpp
+    hpx/compute/detail/target_distribution_policy.hpp
```
msimberg, line 51: I think instead of doing this manually we can just add a check to exclude any headers with `/detail/` in their path over here: https://github.com/STEllAR-GROUP/hpx/blob/502eeb4f8c03fab421756f3a63f69760cdd969d4/cmake/HPX_AddModule.cmake#L98-L100.

aurianer: I thought about that but some of the detail headers were needed for example `execution_information` using `rp.get_pu_mask()` resulted in an `access to incomplete type 'hpx::resource::detail::partitioner' error`
I guessed I make some little changes in the exposed `partitioner.hpp` and exclude detail/ in the `HPX_AddModule.cmake`
Will do

msimberg: We can explicitly include a detail header in cases like that.

hkaiser: Yes, that only tells us that we didn't #include the `/detail` header in places where it's needed. Let's fix that instead.

cmake/HPX_AddModule.cmake
```diff
@@ -95,7 +95,8 @@ function(add_hpx_module name)
     set(module_headers)
     foreach(header_file ${${name}_HEADERS})
       # Exclude the files specified
-      if (NOT (${header_file} IN_LIST ${name}_EXCLUDE_FROM_GLOBAL_HEADER))
+      if ((NOT (${header_file} IN_LIST ${name}_EXCLUDE_FROM_GLOBAL_HEADER)) AND
+        (NOT("${header_file}" MATCHES "detail")))
```
msimberg, line 99: Let's exclude "force_linking.hpp" here as well so we get #4506 done as well.

aurianer:
I just added some new missing files and dependencies which appeared after rebasing, let's wait for the tests and this should be good to go :)

msimberg:
Thanks @aurianer! The remaining CircleCI failures are still due to #4521 which has been fixed in master. Merging. Just to check: does this check only that all direct module dependencies are set as dependencies, or does it also check that no non-direct transitive dependencies are set as dependencies (I know the latter might not be that straightforward)?

aurianer:
For now it checks only the direct dependencies to `global_headers` (simple grep for `hpx/<module>.hpp`), I will have to improve that in the future but I left it like that for now as the modules are not complete and some dependencies are just temporary

PR #4504, opened 2020-04-08
Fix C++ standard handling

msimberg:


msimberg:
I'll keep this open until I'm sure the issue on the mailing list is fixed.

msimberg:
I've pushed another standard related fix as well: the pkgconfig files weren't getting the standard flag. I've also enabled the build unit tests on pycicle which would've caught this (some of them build explicitly with C++17).

msimberg:
Merging this now in any case. Any remaining problems relating to standard handling will come in a later PR.

PR #4508, opened 2020-04-08
Only test pull requests (not all branches) with GitHub actions

msimberg:
It seems like GitHub actions was not smart enough to only trigger one build per push when the triggers were `[push, pull_request]`. Please don't merge yet as I want to make sure that `pull_request` doesn't mean builds will be triggered only once when a pull request is opened.

If someone wants the builders to run on their branch without opening a PR, they can temporarily add `push` to the triggers.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -6,7 +6,7 @@
 
 name: Linux CI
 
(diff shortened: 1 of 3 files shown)


msimberg:
Yep, seems to work... merging.

PR #4514, opened 2020-04-12
Adding polymorphic_executor

### LINKED ISSUE #4489: Polymorphic executor
I'm using two different executors `hpx::mpi::executor` (from John's branch) and `hpx::threads::executors::pool_executor` in my code. It would be great if I could toggle between the two dynamically. In other words, I'm looking for a `std::function` equivalent for executors.

Another use of a polymorphic executor for me is to reduce compile times. A function accepting a single polymorphic executor can be (potentially) instantiated only once instead of multiple times with different executor types.

hkaiser:
- flyby: fixing various flaws in executor implementation

Fixes #4489

### DIFF
--- a/libs/execution/include/hpx/execution/executors/parallel_executor.hpp
+++ b/libs/execution/include/hpx/execution/executors/parallel_executor.hpp
@@ -183,7 +183,7 @@ namespace hpx { namespace parallel { namespace execution {
         HPX_FORCEINLINE
             hpx::future<typename hpx::util::detail::invoke_deferred_result<F,
                 Future, Ts...>::type>
-            then_execute(F&& f, Future&& predecessor, Ts&&... ts)
+            then_execute(F&& f, Future&& predecessor, Ts&&... ts) const
         {
             using result_type =
                 typename hpx::util::detail::invoke_deferred_result<F, Future,
@@ -305,8 +305,8 @@ namespace hpx { namespace parallel { namespace execution {
         /// \cond NOINTERNAL
         template <typename Result, typename F, typename Iter, typename... Ts>
         void spawn_sequential(std::vector<hpx::future<Result>>& results,
-            lcos::local::latch& l, std::size_t base, std::size_t size,
-            F const& func, Iter it, Ts const&... ts) const
+            lcos::local::latch& l, std::size_t base, std::size_t size, F&& func,
+            Iter it, Ts&&... ts) const
         {
             // spawn tasks sequentially
             HPX_ASSERT(base + size <= results.size());
@@ -322,8 +322,7 @@ namespace hpx { namespace parallel { namespace execution {
         template <typename Result, typename F, typename Iter, typename... Ts>
         void spawn_hierarchical(std::vector<hpx::future<Result>>& results,
             lcos::local::latch& l, std::size_t base, std::size_t size,
-            std::size_t num_tasks, F const& func, Iter it,
-            Ts const&... ts) const
+            std::size_t num_tasks, F&& func, Iter it, Ts&&... ts) const
         {
             if (size > num_tasks)
             {
--- a/libs/execution/include/hpx/execution/executors/parallel_executor_aggregated.hpp
+++ b/libs/execution/include/hpx/execution/executors/parallel_executor_aggregated.hpp
@@ -63,18 +63,21 @@ namespace hpx { namespace parallel { namespace execution {
         }
 
         /// \cond NOINTERNAL
-        bool operator==(parallel_policy_executor_aggregated const& rhs) const
-            noexcept
+        // clang-format produces inconsistent results between versions
+        // clang-format off
+        bool operator==(
+            parallel_policy_executor_aggregated const& rhs) const noexcept
         {
             return num_spread_ == rhs.num_spread_ &&
                 num_tasks_ == rhs.num_tasks_;
         }
 
-        bool operator!=(parallel_policy_executor_aggregated const& rhs) const
-            noexcept
+        bool operator!=(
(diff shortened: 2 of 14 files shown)


hkaiser:
@teonik I think this is ready for experimentation now. Here is a short documentation:

The type `hpx::parallel::executors::polymorphic_executor<>` is a template the expects the function signature for the functions it can schedule as its template argument. It needs to be constructed from any other executor.

For instance:
```
double f(int arg) 
{ 
    return double(arg); 
}

using executor = hpx::parallel::executors::polymorphic_executor<double(int)>;

void polymorphic_scheduling(executor const& exec)
{
    // use it to schedule the function f
    hpx::future<double> fut = hpx::parallel::async_execute(exec, f, 42);
    HPX_TEST_EQ(fut.get(), 42.0);
}

void schedule_things()
{
    // create polymorphic executor from parallel_executor
    hpx::parallel::executors::parallel_executor par_exec;
    polymorphic_scheduling(executor(par_exec));

    // create polymorphic executor from sequential_executor
    hpx::parallel::executors::sequential_executor seq_exec;
    polymorphic_scheduling(executor(seq_exec));
}
```

Please note, that the function signatures that can be scheduled using `then_execute`, `bulk_async_execute` and `bulk_then_execute` are slightly different. 

Assuming the function signature is `R(Ts...)`, then the corresponding signatures for the scheduling functions above are:
-  `R(hpx::shared_future<void> const&, Ts...)` (for `then_execute`), 
- `R(std::size_t, Ts...)` (for `bulk_async_execute`), and 
- `R(std::size_t, hpx::shared_future<void> const&, Ts...)` (for `bulk_then_execute`). 

This also means that the `bulk_` functions can only be used with a `Shape` (range-)object that exposes something convertable to `std::size_t` as its `value_type` and that the `then_` variations can only pass on `shared_future<void>` to the embedded executors.

The other specific property of the `polymorphic_executor<>` is that it implements all 7 customization points directly, all of which simply forward their arguments to the embedded executor (the one it was constructed from).

For a full set of examples, see for instance here: https://github.com/STEllAR-GROUP/hpx/blob/5ed7b01e473848c5d0f21d6d6c8fb7f3b1b1b3a9/libs/execution/tests/unit/polymorphic_executor.cpp.

libs/execution/include/hpx/execution/executors/parallel_executor.hpp
```diff
@@ -322,8 +322,7 @@ namespace hpx { namespace parallel { namespace execution {
         template <typename Result, typename F, typename Iter, typename... Ts>
         void spawn_hierarchical(std::vector<hpx::future<Result>>& results,
             lcos::local::latch& l, std::size_t base, std::size_t size,
-            std::size_t num_tasks, F const& func, Iter it,
-            Ts const&... ts) const
+            std::size_t num_tasks, F&& func, Iter it, Ts&&... ts) const
         {
```
sithhell, line 326: Changing the arguments to forwarding references would require to forward them inside the function body, no?

libs/execution/include/hpx/execution/executors/parallel_executor_aggregated.hpp
```diff
@@ -125,13 +128,13 @@ namespace hpx { namespace parallel { namespace execution {
                 lcos::local::latch l(size);
                 if (hpx::detail::has_async_policy(Policy{}))
                 {
-                    spawn_hierarchical(l, size, num_tasks, std::forward<F>(f),
+                    spawn_hierarchical(l, size, num_tasks, f,
```
sithhell, line 131: why do you not forward `f` here?

libs/execution/include/hpx/execution/executors/polymorphic_executor.hpp
```diff
@@ -0,0 +1,930 @@
...
+                return lhs.impl_->equal_to(*rhs.impl_);
+            }
+
+            friend bool operator!=(shape_iter const& lhs, shape_iter const& rhs)
+            {
+                return !(lhs == rhs);
+            }
+
+        protected:
+            hpx::memory::intrusive_ptr<shape_iter_impl_base> impl_;
```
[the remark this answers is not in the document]
hkaiser: Good point. Unique ownership might be sufficient. We need to reallocate the implementation object on assignment anyways.

libs/execution/include/hpx/execution/executors/parallel_executor.hpp
hkaiser: The arguments are used more than once inside the body.

libs/execution/include/hpx/execution/executors/parallel_executor_aggregated.hpp
hkaiser: I ran into dangling references. We need to revisit argument handling inside asynchronous algorithms thoroughly. But that's for a separate PR, I think. For now, this change fixed my issues.

hkaiser:
@msimberg this is good to go (once the CI cycled).

libs/execution/tests/unit/polymorphic_executor.cpp
```diff
@@ -0,0 +1,203 @@
...
+        .get();
+    HPX_TEST(count == v.size());
+}
+
+void static_check_executor()
+{
+    using namespace hpx::traits;
+
+    static_assert(has_sync_execute_member<executor>::value,
+        "!has_sync_execute_member<executor>::value");
```
msimberg, line 153: *Really* important:
```suggestion
        "has_sync_execute_member<executor>::value");
```
Feel free to wait for CI and make this change just before merging.

hkaiser: How did that happen? :/

PR #4515, opened 2020-04-12
Replace backslashs by slashs in windows paths

### LINKED ISSUE #4513: HPXConfig.cmake contains ill-formed paths when library paths use backslashes
## Expected Behavior

`find_package(HPX REQUIRED)` should work.

## Actual Behavior

`HPXConfig.cmake` contains statements like `set(HPX_HWLOC_ROOT "C:\local\hwloc-win64-build-2.1.0")` resulting in the following when used:
```
1> [CMake] CMake Error at C:/stellar/inst-hpx-stable-release/lib/cmake/HPX/HPXConfig.cmake:79 (set):
1> [CMake]   Syntax error in cmake code at
1> [CMake] 
1> [CMake]     C:/stellar/inst-hpx-stable-release/lib/cmake/HPX/HPXConfig.cmake:79
1> [CMake] 
1> [CMake]   when parsing string
1> [CMake] 
1> [CMake]     C:\local\hwloc-win64-build-2.1.0
1> [CMake] 
1> [CMake]   Invalid character escape '\l'.
1> [CMake] Call Stack (most recent call first):
1> [CMake]   CMakeLists.txt:8 (find_package)
```

  1. Build HPX with a command line like `cmake ..\hpx -DCMAKE_BUILD_TYPE=Release -G "Ninja" -DCMAKE_INSTALL_PREFIX=C:\stellar\inst-hpx-stable-release -DHWLOC_ROOT=C:\local\hwloc-win64-build-2.1.0 -DHPX_WITH_TESTS=OFF -DHPX_WITH_EXAMPLES=OFF`
  1. Use `find_package(HPX REQUIRED)` in a project.
  1. Run `cmake ..\demo-hpx -G "Visual Studio 16 2019" -A x64 -DHPX_ROOT=C:\stellar\inst-hpx-stable-release`

## Specifications

  - HPX Version: stable tag (e7090becdafd73ebcea5bc010976676188d1ffac)
  - Platform (compiler, OS): Windows 10, VS 2019 x64 command prompt, CMake 3.17.0

aurianer:
Fixes #4513


cmake/HPX_AdaptWindowsPaths.cmake
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2020 ETH Zurich
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+function(adapt_path_if_win var)
+  if(WIN32)
+    string(REPLACE "\\" "/" new_var "${${var}}")
```
hkaiser, line 9: I think this needs to be crafted a bit more carefully. We shouldn't replace `\\` (i.e. an already escaped backslash with '//'). The following might be a starting point (from the top of my head, not verified):
```suggestion
    # replace any backslash/forward-slash that is not followed by another
    # backslash/forward-slash
    string(REGEX REPLACE "[\\\\/][^\\\\/]*$" "/" newvar "${${var}}")
```
Optionally you could now replace all remaining (now escaped) backslashes with a single forward-slash as well.

zao:
This PR seems to be based off a rather old master commit and will as such not build correctly on my machine, so I cannot test it as it currently is.

zao:
I rebased my checkout and this does not solve the problem.

The paths in `HPXConfig.cmake` are invalid CMake syntax as the backslashes in them are interpreted as escapes, CMake complains as soon as it tries parsing the string literal.

This assumedly needs to be done somewhere before synthesising the template.

msimberg:
Can we make use of https://cmake.org/cmake/help/latest/command/file.html#path-conversion? Just having everything as CMake paths internally should be enough (unless someone is offended that their backslashes were converted to forward slashes, but I think they'll have to live with it...).

aurianer:
> This PR seems to be based off a rather old master commit and will as such not build correctly on my machine, so I cannot test it as it currently is.

Sorry I thought you were on the stable tag, and I probably didn't update it before committing :/

aurianer:
> Can we make use of https://cmake.org/cmake/help/latest/command/file.html#path-conversion? Just having everything as CMake paths internally should be enough (unless someone is offended that their backslashes were converted to forward slashes, but I think they'll have to live with it...).

@msimberg I tried the `file(TO_CMAKE_PATH)` but it would replace the `C:` as a `C;` and I don't know why..

aurianer:
> I rebased my checkout and this does not solve the problem.
> 
> The paths in `HPXConfig.cmake` are invalid CMake syntax as the backslashes in them are interpreted as escapes, CMake complains as soon as it tries parsing the string literal.
> 
> This assumedly needs to be done somewhere before synthesising the template.

@zao yes I understand the problem, what I did was just trying to replace the backslashes with slashes, would you mind trying to remove the `if(WIN32)` from the `cmake/HPX_AdaptWindowsPaths.cmake` to make it replace it always, if it works I can try to find out the right platform after

aurianer:
I rebased on the latest stable tag now

aurianer:
> > Can we make use of https://cmake.org/cmake/help/latest/command/file.html#path-conversion? Just having everything as CMake paths internally should be enough (unless someone is offended that their backslashes were converted to forward slashes, but I think they'll have to live with it...).
> 
> @msimberg I tried the `file(TO_CMAKE_PATH)` but it would replace the `C:` as a `C;` and I don't know why..

Oh yeah I see native path now !

zao:
> @zao yes I understand the problem, what I did was just trying to replace the backslashes with slashes, would you mind trying to remove the `if(WIN32)` from the `cmake/HPX_AdaptWindowsPaths.cmake` to make it replace it always, if it works I can try to find out the right platform after

This changed nothing. The file is still ill-formed and bails out on the `set` line.

hkaiser:
> Can we make use of https://cmake.org/cmake/help/latest/command/file.html#path-conversion? Just having everything as CMake paths internally should be enough (unless someone is offended that their backslashes were converted to forward slashes, but I think they'll have to live with it...).
> 
> @msimberg I tried the file(TO_CMAKE_PATH) but it would replace the C: as a C; and I don't know why..

This looks like a cmake problem to me. A workaround would be to use `file(TO_CMAKE_PATH)` and  (on windows) replace all `;` to `:`.

aurianer:
> > Can we make use of https://cmake.org/cmake/help/latest/command/file.html#path-conversion? Just having everything as CMake paths internally should be enough (unless someone is offended that their backslashes were converted to forward slashes, but I think they'll have to live with it...).
> > @msimberg I tried the file(TO_CMAKE_PATH) but it would replace the C: as a C; and I don't know why..
> 
> This looks like a cmake problem to me. A workaround would be to use `file(TO_CMAKE_PATH)` and (on windows) replace all `;` to `:`.

Ok I'll try that !

cmake/HPX_AdaptWindowsPaths.cmake
aurianer: When I did that I based it on the fact that do a `cd <name1>//<name2>` is equivalent to do a `cd <name1>/<name2>` on Linux, but I actually haven't check on windows..
I think I'll go with your suggestion of the `TO_CMAKE_PATH` and then replace semi column thanks!

zao:
b569140 seems to make the build handle paths for hwloc with arbitrary mixes of slash types. 👍

PR #4516, opened 2020-04-14
Fix various builds

msimberg:
Fixes build unit tests with APEX and clang CUDA. Also fixes nvcc build failures, and the `default_executor` headers.

The first was failing due to APEX depending on HPX and pulling in the generator expressions `$<TARGET_NAME_IF_EXISTS:{HPX::,}hpx>` into the pkgconfig file generation. I've simply ignored those generator expressions as HPX gets linked correctly anyway. The second was failing because of trying to add a flag to `hpx_public_flags` in `HPXConfig.cmake` that didn't need to be added anymore. The nvcc failures were triggered by `guided_pool_executor.hpp` being included in the the stream test when it wasn't included before. The failure was caused by trying to use `decltype(*this)` inside a lambda which I guess nvcc just doesn't know how to deal with (plain `*this` is fine...). I changed it to use `std::decay` instead, which seems to be equivalent in that use case.


msimberg:
> I changed it to use `std::decay` instead, which seems to be equivalent in that use case.

Okay, not quite... still looking into this.

PR #4518, opened 2020-04-14
Module distributed lcos

aurianer:
Add the `lcos_distributed` module


components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_decl.hpp
```diff
@@ -12,8 +12,8 @@
 
 #include <hpx/config.hpp>
 #include <hpx/assertion.hpp>
-#include <hpx/lcos/wait_all.hpp>
-#include <hpx/lcos/when_all.hpp>
+#include <hpx/distributed_lcos/wait_all.hpp>
+#include <hpx/distributed_lcos/when_all.hpp>
```
hkaiser, line 16: The `wait_all` and `when_all` facilities are purely local (so are `wait_any`, `wait_n`, `when_any`. and `when_n`)

examples/qt/qt.cpp
```diff
@@ -7,7 +7,7 @@
 #include <hpx/hpx.hpp>
 #include <hpx/hpx_init.hpp>
 #include <hpx/include/thread_executors.hpp>
-#include <hpx/lcos/future_wait.hpp>
+#include <hpx/distributed_lcos/future_wait.hpp>
```
hkaiser, line 10: `future_wait` is purely local.

hpx/lcos/dataflow.hpp
```diff
@@ -25,7 +25,7 @@
 #include <hpx/functional/deferred_call.hpp>
 #include <hpx/functional/invoke_fused.hpp>
 #include <hpx/functional/traits/is_action.hpp>
-#include <hpx/lcos/detail/future_transforms.hpp>
+#include <hpx/distributed_lcos/detail/future_transforms.hpp>
```
hkaiser, line 28: These are local facilities as well.

components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_decl.hpp
msimberg: The are also in àsync_combinators` in #4478 (which I will merge today if CI looks good).

msimberg:
> The sync_fwd.hpp file needs to be split into a file declaring the local `sync()` and one declaring the remote `sync()` overloads.

Also in #4478.

components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_decl.hpp
aurianer: ok thanks !

examples/qt/qt.cpp
aurianer: sorry

libs/distributed_lcos/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+  base_lco_with_value.cpp
+  base_lco_with_value_1.cpp
+  base_lco_with_value_2.cpp
+  base_lco_with_value_3.cpp
+  promise.cpp
+  )
+
+include(HPX_AddModule)
+add_hpx_module(distributed_lcos
+  COMPATIBILITY_HEADERS OFF
```
msimberg, line 52:
```suggestion
  COMPATIBILITY_HEADERS ON
```

libs/distributed_lcos/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+  DEPRECATION_WARNINGS
+  FORCE_LINKING_GEN
+  GLOBAL_HEADER_GEN OFF
```
msimberg, line 55:
```suggestion
  GLOBAL_HEADER_GEN ON
```

libs/distributed_lcos/CMakeLists.txt
aurianer: May I change the default in the python script ? I always forget about those..

aurianer:
> Looks good to me except for the compatibility options. Thanks!

Doesn't look that good ^^'
https://183927-4455628-gh.circle-artifacts.com/0/hpx/artifacts/circular_deps.svg

aurianer:
> Looks good to me except for the compatibility options. Thanks!

Doesn't look that good ^^'
https://183927-4455628-gh.circle-artifacts.com/0/hpx/artifacts/circular_deps.svg

libs/distributed_lcos/CMakeLists.txt
msimberg: Of course. It's supposed to help us do the right thing after all.

msimberg:
@hkaiser I've stripped this down to just move `channel` to the new `lcos` module. The other files in the `hpx/lcos` directory (ignoring future-related ones) have circular dependencies with the `async` module that I don't dare try to resolve here. They can be added to this module later if it seems like they fit.

hkaiser:
I don't want to be picky, but we kind of adopted the naming convention to append the attributes to the module name. Could we name this one `lcos_distributed` instead of `distributed_lcos`? This is not a requirement, however - just a suggestion.

msimberg:
> I don't want to be picky, but we kind of adopted the naming convention to append the attributes to the module name. Could we name this one `lcos_distributed` instead of `distributed_lcos`? This is not a requirement, however - just a suggestion.

I'm glad you're picky. This one is called just `lcos` now, but `lcos_distributed` is more consistent with other module names. I'll rename this as well.

msimberg:
> I'm glad you're picky. This one is called just `lcos` now, but `lcos_distributed` is more consistent with other module names. I'll rename this as well.

Done. `local_lcos` and the rest will be renamed separately though.

aurianer:
Thanks @msimberg!

PR #4521, opened 2020-04-15
Moves more miscellaneous files to modules

msimberg:
Feel free to disagree with where I've put things:

- `hpx/lcos/local/spmd_block.hpp` moved to `local_lcos`
- `hpx/parallel/spmd_block.hpp` and `hpx/parallel/task_block.hpp` moved to `algorithms`
- `numa_allocator.hpp` and `numa_binding_allocator.hpp` moved to `compute`
- `hpx/util/zip_iterator.hpp` moved to `segmented_algorithms` (unlike its name, it actually only contained some specializations for segmented iterators, `algorithms` has similar `zip_iterator.hpp` files)
- the `tagged_x` base functionality moved to `datastructures`, the future specializations to `algorithms` where they are used
- Vc serialization moved to `serialization` (external types)
- `serializable_any` moved to `serialization`. This one could perhaps do with some refactoring, but `tuple` serialization is also in `serialization` so it fits to that extent. `serialization` depends for now on `datastructures` which means that `datastructures` can't depend on `serialization`. I'll look into this separately (it might be that `serialization` only depends on `datastructures` because of `tuple`).
- some `thread_executor` related files moved to `thread_executors`. I think this is ok in terms of dependencies (`cpp-dependencies` locally claimed it was ok). It does mean that `thread_executors` depends on things like `threadmanager`. It is a bit unnecessarily heavy, but since we're `thread_executors` I won't spend time on that unless required by other dependencies.
- `thread_aware_timer` is deprecated. Was this dear to someone?


hkaiser:
> hpx/parallel/spmd_block.hpp and hpx/parallel/task_block.hpp moved to algorithms

If `hpx/lcos/local/spmd_block.hpp` was moved to `local_lcos`, then `hpx/parallel/spmd_block.hpp` should go into the `distributed_lcos` module, I think. `task_block` belongs into algorithms, however.

hkaiser:
> thread_aware_timer is deprecated. Was this dear to someone?

It's not used anywhere, I believe. Let's get rid of it (eventually).

msimberg:
> > hpx/parallel/spmd_block.hpp and hpx/parallel/task_block.hpp moved to algorithms
> 
> If `hpx/lcos/local/spmd_block.hpp` was moved to `local_lcos`, then `hpx/parallel/spmd_block.hpp` should go into the `distributed_lcos` module, I think. `task_block` belongs into algorithms, however.

This mostly a naming issue... the way it currently stands is:

- `hpx/parallel/spmd_block.hpp` basically just forwards `hpx/lcos/local/spmd_block.hpp` to the `hpx::parallel` namespace. The latter actually contains the (local) implementation of `spmd_block`. `collectives` has a distributed version of `spmd_block`. We'll have to think about the `collectives`/`(distributed_)lcos` split.
- `hpx/parallel/task_block.hpp` is all there is for `task_block`. It contains a (local) `task_block`.

libs/serialization/include/hpx/serialization/serializable_any.hpp
```diff
@@ -97,17 +97,17 @@ namespace hpx { namespace util { namespace detail { namespace any {
 
         void save_object(void* const* object, OArch& ar, unsigned)
         {
-            ar & Vtable::get(object);
+            ar& Vtable::get(object);
```
hkaiser, line 100: `//clang-format off` ?

msimberg: Yeah... thanks.

hkaiser:
Here are a couple more files that could be moved now:
- `hpx/traits/pointer_category.hpp` --> `iterator_support`
- `hpx/traits/is_launch_policy.hpp` --> `execution`
- `hpx/runtime/launch_policy.hpp` --> `execution`
- `hpx/util/functional/segemented_iterator_helpers.hpp` --> `segmented_algorithms`
- `hpx/util/custom_exception_info.hpp` --> `errors`

(there might be more)

msimberg:
> Here are a couple more files that could be moved now:
> 
>     * `hpx/traits/pointer_category.hpp` --> `iterator_support`
> 
>     * `hpx/traits/is_launch_policy.hpp` --> `execution`
> 
>     * `hpx/runtime/launch_policy.hpp` --> `execution`
> 
>     * `hpx/util/functional/segemented_iterator_helpers.hpp` --> `segmented_algorithms`
> 
>     * `hpx/util/custom_exception_info.hpp` --> `errors`
> 
> 
> (there might be more)

Thanks! There's also a whole bunch of other traits that will be moved soon with the future module(s). `custom_exception_info.hpp` very much depends on the runtime at the moment (which is why it's not in the errors module).

PR #4525, opened 2020-04-15
Changing sub-name sequencing for experimental namespace

hkaiser:
This change will simplify pulling things from the experimental namespace into the main namespace down the road as it makes different things in the experimental namespace more granular. I.e. if we at some point want to start the process of moving the facilities that are now in `namespace hpx::experimental::mpi` into the main namespace, this can be easiliy done with:
```
#if defined(HPX_EXPERIMENTAL_MPI_COMPATIBILITY
namespace hpx { namespace mpi {
    using namespace hpx::experimental::mpi;
}}
#endif
```
This wouldn't be possible with the current naming scheme.

### DIFF
--- a/libs/mpi/include/hpx/mpi/mpi_executor.hpp
+++ b/libs/mpi/include/hpx/mpi/mpi_executor.hpp
@@ -17,7 +17,7 @@
 
 #include <mpi.h>
 
-namespace hpx { namespace mpi { namespace experimental {
+namespace hpx { namespace experimental { namespace mpi {
 
     struct executor
     {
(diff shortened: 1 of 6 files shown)


hkaiser:
After walking around the block I'm not sure anymore what I was thinking. The current scheme is perfectly fine and can be used in a similar way as outlined above. Closing...

msimberg:
You had me thinking real hard with this one. That said, if someone thinks that `hpx::experimental::mpi` actually feels better than `hpx::mpi::experimental` I think we can go with that as well. The argument for the latter is mostly for consistency with `detail`, but that doesn't need to stay that way either...

PR #4305, opened 2020-01-14
Implementing `hpx::functional::tag_invoke`

sithhell:
This PR introduces `hpx::functional::tag_invoke` which is an implementation
of https://wg21.link/p1895. This is needed to implement https://wg21.link/p0443.

PR #4526, opened 2020-04-15
Shared state for cv

hkaiser:
This fixes a potential race wrt `condition_variable[_any]` destructor.

This relies on #4512 being merged first (only the last commit is unique to this branch)


cmake/tests/cxx17_inline_variable.cpp
```diff
@@ -0,0 +1,12 @@
...
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+inline constexpr int v = 42;
+
+int main()
+{
+    return 0;
+}
```
msimberg, line 12: By the way, @sithhell has the same functionality in #4305 (https://github.com/STEllAR-GROUP/hpx/pull/4305/files#diff-b6fbaa7a023d645956d24534dff6909e). The other one looks a bit more comprehensive, although I guess this one would do the job as well.

libs/concurrency/include/hpx/concurrency/cache_line_data.hpp
```diff
@@ -71,10 +71,7 @@ namespace hpx {
             typename NeedsPadding = typename detail::needs_padding<Data>::type>
         struct cache_aligned_data
         {
-            cache_aligned_data()
-              : data_{}
-            {
-            }
```
msimberg, line 77: I think we explicitly kept it like this, since some compilers go and initialize the padding otherwise: http://quick-bench.com/T-Z9iZ5Wn7E4WE2DoOqGEwVQPOE (i.e. GCC does it, clang doesn't seem to care either way). I suppose a comment would've been good.

libs/threading/tests/unit/stop_tokenrace.cpp
```diff
@@ -0,0 +1,238 @@
...
+int main()
+{
+    test_callback_register();
+    test_callback_unregister();
+
+    test_callback_concurrent_unregister();
+    test_callback_concurrent_unregister_other_thread();
+
+    // this test terminates execution
+    // test_callback_throw();
```
msimberg, line 235: Is this intentionally disabled?

cmake/tests/cxx17_inline_variable.cpp
hkaiser: Yes, I saw that. I was planning to consolidate this once one of the PRs is merged.

libs/concurrency/include/hpx/concurrency/cache_line_data.hpp
hkaiser: Excellent catch! I forgot about that. I'll revert this change.

libs/threading/tests/unit/stop_tokenrace.cpp
hkaiser: Yes, we have no way of testing whether `std::terminate` was actually called. Any ideas?

msimberg: Hmm, is it just that last `request_stop` (line 221) that would terminate? I don't know if this is a good idea, but does one have to call `exit` in the termination handler? If not that could be one option (not sure how messy the synchronization would be). Otherwise maybe separating out that test and printing something unique in the termination handler, and making CMake check for that output?

hkaiser: If the callbacks throw an exception, `std::terminate` will be called (see [C++20](http://wg21.link/n4849), 32.3.5.1.4). 

Using CMake to detect some unique output would work.

hkaiser:
Once the tests have cycled this is good to go.

msimberg:
This is still a problem: https://cdash.cscs.ch//viewBuildError.php?buildid=104881. Although I'm not sure it's a real problem. We may just have to ignore that warning.

hkaiser:
> This is still a problem: https://cdash.cscs.ch//viewBuildError.php?buildid=104881. Although I'm not sure it's a real problem. We may just have to ignore that warning.

I thought, the `NOLINTNEXTLINE` would take care of this. I'll add a `(void*)` to cast away the type as well. As the comment says, the code is UB on purpose, so we can safely work around the error.

msimberg:
> I thought, the `NOLINTNEXTLINE` would take care of this. I'll add a `(void*)` to cast away the type as well. As the comment says, the code is UB on purpose, so we can safely work around the error.

`NOLINTNEXTLINE` is only for clang-tidy warnings. This is a plain compiler warning (oddly enough only with nvcc...).

hkaiser:
Darn, the very test that is supposed to verify the proposed fix is segfaulting :/ Can anybody reproduce this (as it works for me)?

libs/threading/tests/unit/CMakeLists.txt
```diff
@@ -9,19 +9,23 @@ set(threading_tests
     condition_variable2
     condition_variable3
     condition_variable4
+    condition_variable_race
```
aurianer, line 12: Just for consistency, let's maybe remove the "_" to `condition_variable_race` or add it to the `stop_tokenrace*`

hkaiser:
> > Darn, the very test that is supposed to verify the proposed fix is segfaulting :/ Can anybody reproduce this (as it works for me)?
> 
> The cmake option of the CI which triggers the seg fault for me is -DHPX_WITH_THREAD_LOCAL_STORAGE=On

Unfortunately, that does not reproduce it for me. :/

aurianer:
> > > Darn, the very test that is supposed to verify the proposed fix is segfaulting :/ Can anybody reproduce this (as it works for me)?
> > 
> > 
> > The cmake option of the CI which triggers the seg fault for me is -DHPX_WITH_THREAD_LOCAL_STORAGE=On
> 
> Unfortunately, that does not reproduce it for me. :/

Yeah I can't reproduce it anymore either..

msimberg:
I can try to reproduce this but I need some time for it.

hkaiser:
@sithhell, @msimberg this looks clean now. Do you have any objections to go ahead and merge this?

msimberg:
Thanks for looking into this @hkaiser!

PR #4527, opened 2020-04-16
Creating actions_base and actions module

hkaiser:
This PR adds the beginnings for the following modules
- actions_base
- actions
- components_base

This by no means complete as many headers will have to be moved, still.


libs/CMakeLists.txt
```diff
@@ -140,11 +137,6 @@ foreach(lib ${HPX_CANDIDATE_LIBS})
   if (HPX_${lib}_LIBRARY_ENABLED)
     set(HPX_LIBS ${HPX_LIBS} ${lib} CACHE INTERNAL "list of Enabled HPX modules" FORCE)
 
-    string(TOUPPER ${lib} uppercase_lib)
-
-    set(MODULE_ENABLED_LIB_DEFINES
-      "${MODULE_ENABLED_LIB_DEFINES}#define HPX_HAVE_LIB_${uppercase_lib}\n")
-
```
msimberg, line 147: I guess this was removed just because it's unused?

hkaiser: Hmmm. This is missing from the module generator script. I'll fix that independently.

hkaiser: See #4529

hkaiser:
This has been updated to current master now.

hkaiser:
I'd like to leave this PR at that. There are many more headers that need moving, but I would like to do that one at a time.

In general, the next steps would be to create the `agas` and `components` modules. I don't expect too many difficulties with that, however.

msimberg:
Thanks @hkaiser for updating this! Pycicle has again not been running... There was a build error with clang 5 earlier and I'd like to see if it's gone. I've merged master again to get the jenkins configs onto this branch.

hkaiser:
\o/

PR #4531, opened 2020-04-16
Update scripts

aurianer:
- Fix the error message of the compatibility header
- Update the `change_includes.sh` to allow manual specification of headers (see inside the file)
- [Here](https://gist.github.com/aurianer/cf3dce01676ae0c645c6c60a463d16eb) is the little guide to use those module scripts


libs/cache/tests/unit/local_mru_cache.cpp
```diff
@@ -4,8 +4,8 @@
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-#include <hpx/cache/entries/lru_entry.hpp>
-#include <hpx/cache/local_cache.hpp>
+#include <hpx/util/cache/entries/lru_entry.hpp>
+#include <hpx/util/cache/local_cache.hpp>
```
hkaiser, line 8: This is most probably not an intended change.

aurianer: indeed !

aurianer:
> Should we move all scripts into the tools folder?

Yes apart from the `create_library_skeleton.py` one which is based on relative paths I believe

PR #4532, opened 2020-04-16
Fixing broken url in README.rst

hkaiser:
- adding more current and past funding sources


README.rst
```diff
@@ -134,25 +134,35 @@ Acknowledgements
 ================
 
 We would like to acknowledge the NSF, DoE, DTIC, DARPA, the Center for
-Computation and Technology (CCT) at Louisiana State University, and the
+Computation and Technology (CCT) at Louisiana State University, The Swiss
```
aurianer, line 137: Sorry I'm late but wouldn't it be better to have `the` in lower case as it is for CCT ?

PR #4536, opened 2020-04-17
Add generic allocator for execution policies

msimberg:
Adds `hpx::compute::host::detail::policy_allocator` which allows creating an allocator for use with `hpx::compute::vector` with any execution policy, not just with the `block_executor`. The `block_allocator` has been rewritten to use `policy_allocator`, but the implementation itself is unchanged from `block_allocator` on master.

I have also changed the `static_chunk_size` executor parameters to only return powers of two chunk sizes (benchmarks below the fold).

I was unsure if it would be enough to make it generic over executors, but doing it over policies means that one can also switch allocation to be serial by changing the policy. `policy_allocator` is now in `detail`, and I don't know yet if we want to expose it directly.

My main motivation with this is to have be able to use the default `par` policy and get good performance on e.g. the stream benchmark. I.e. the following should be enough:

```
auto policy = par;
allocator<double> alloc(policy);
vector<double> vec(n, alloc);
foo(policy, vec.begin(), vec.end());
```

The lack of template argument deduction on e.g. the allocator (only works if *no* template arguments are specified, but we need at least the element type) means that real code wouldn't actually look this clean. Are there ways to achieve roughly the above level of (non-)verbosity?

---

The follow-up for this is that I'd like to change the default executor for `par` to be based on the `thread_pool_executor` implementation. This is based on the benchmarks in #4301 and the following stream benchmarks. I benchmarked the following combinations (on Piz Daint multicore I should add, two-socket Xeon with 18 cores each, 36 threads total, no hyperthreading; 100 iterations per test):

- `par.on(thread_pool_executor)` and `policy_allocator(thread_pool_executor)`
- `par.on(block_executor)` and `block_allocator`
- `par` and `policy_allocator(par)`
- `par` and `std::allocator`

On big arrays (100000000 elements, ~700 MB) the first two achieve roughly the same peak bandwidth, but the one based on `thread_pool_executor` has much lower variance (and average). The default `par` combination gives roughly 25% lower bandwidth. `std::allocator` gives more than 50% lower bandwidth.

The chunk size change again doesn't change the peak bandwidth (especially not for big arrays), but significantly reduces variance with small arrays. For example, with the `thread_pool_executor` and 100000 elements we get this with the old chunk size:
```
Function    Best Rate MB/s  Avg time     Min time     Max time             
Copy:           24044.9     0.000336     0.000067     0.007587             
Scale:          22141.6     0.000290     0.000072     0.007240             
Add:            37952.4     0.000286     0.000063     0.006938             
Triad:          36632.3     0.000295     0.000066     0.007082
```
and this with the new chunk size:
```
Function    Best Rate MB/s  Avg time     Min time     Max time             
Copy:           28912.1     0.000241     0.000055     0.003331             
Scale:          29232.8     0.000209     0.000055     0.003072             
Add:            46112.2     0.000208     0.000052     0.003155             
Triad:          40975.8     0.000214     0.000059     0.003560
```

The follow-up on the follow-up is to improve performance with small grain sizes.

### DIFF
--- a/libs/algorithms/tests/regressions/count_3646.cpp
+++ b/libs/algorithms/tests/regressions/count_3646.cpp
@@ -39,6 +39,32 @@ struct BitCountingIterator : public Iterator<std::int64_t, stopValue>
         return countBits(this->state + n);
     }
 
+    BitCountingIterator& operator++()
+    {
+        ++(this->state);
+        return *this;
+    }
+
+    BitCountingIterator operator++(int)
+    {
+        auto copy = *this;
+        ++(*this);
+        return copy;
+    }
+
+    BitCountingIterator& operator--()
+    {
+        --(this->state);
+        return *this;
+    }
+
+    BitCountingIterator operator--(int)
+    {
+        auto copy = *this;
+        --(*this);
+        return copy;
+    }
+
 private:
     std::int64_t countBits(std::int64_t v) const
     {
--- a/libs/compute/include/hpx/compute/host/block_allocator.hpp
+++ b/libs/compute/include/hpx/compute/host/block_allocator.hpp
@@ -1,4 +1,5 @@
 ///////////////////////////////////////////////////////////////////////////////
+//  Copyright (c) 2020 ETH Zurich
 //  Copyright (c) 2016 Thomas Heller
 //  Copyright (c) 2016 Hartmut Kaiser
 //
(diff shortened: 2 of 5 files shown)


libs/compute/include/hpx/compute/host/block_allocator.hpp
```diff
@@ -35,241 +36,242 @@
...
-        // Constructs an object of type T in allocated uninitialized storage
-        // pointed to by p, using placement-new
-        template <typename U, typename... Args>
-        void construct(U* p, Args&&... args)
+        block_allocator()
+          : base_type(policy_type(
+                executor_type(target_type(1)), executor_parameters_type()))
         {
-            executor_.execute(hpx::util::functional::placement_new<U>(), p,
-                std::forward<Args>(args)...);
```
msimberg, line 237: I think this function never got instantiated before. Is `execute` some *really* old executor member function? I've changed it to just do placement new directly, but did `execute` have some other behaviour?

libs/execution/include/hpx/execution/executors/static_chunk_size.hpp
```diff
@@ -58,16 +58,28 @@ namespace hpx { namespace parallel { namespace execution {
...
             // Make sure the internal round robin counter of the executor is
             // reset
             execution::reset_thread_distribution(*this, exec);
 
-            // by default use static work distribution over number of
-            // available compute resources, create four times the number of
-            // chunks than we have cores
-            return (num_tasks + 4 * cores - 1) / (4 * cores);    // -V112
+            // Return a chunk size that is a power of two; and that leads to at
+            // least 2 chunks per core, and at most 4 chunks per core.
```
msimberg, line 75: I haven't compared e.g. chunk sizes that lead to 4-8 (or even more) chunks per core. I thought 2-4 per core would be a conservative number for good performance even though in the worst case 2 might lead to considerable load imbalance.

libs/compute/include/hpx/compute/host/block_allocator.hpp
hkaiser: Yes, this should be `executor_.sync_execute()` instead. That name was changed a long time ago and we missed fixing it here.

libs/execution/include/hpx/execution/executors/static_chunk_size.hpp
hkaiser: @scheherzade can you comment on this?

hkaiser:
> The lack of template argument deduction on e.g. the allocator (only works if no template arguments are specified, but we need at least the element type) means that real code wouldn't actually look this clean. Are there ways to achieve roughly the above level of (non-)verbosity?

C++17 CTAD should help with that.

msimberg:
> > The lack of template argument deduction on e.g. the allocator (only works if no template arguments are specified, but we need at least the element type) means that real code wouldn't actually look this clean. Are there ways to achieve roughly the above level of (non-)verbosity?
> 
> C++17 CTAD should help with that.

That's what I was hoping as well, but from my reading (and testing) so far [it may not be possible](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction#Notes): "If a template argument list is specified, deduction does not take place." I may just be using it wrong.

libs/compute/include/hpx/compute/host/block_allocator.hpp
hkaiser: Well, actually, I think we should use the customization points instead (all over this file). So this would have to use `parallel::execution::sync_execute(executor_, …)`.

libs/compute/include/hpx/compute/host/block_allocator.hpp
```diff
@@ -35,241 +36,242 @@
...
 
-        typedef std::false_type is_always_equal;
-        typedef std::true_type propagate_on_container_move_assignment;
+            using value_type = T;
+            using pointer = T*;
+            using const_pointer = const T*;
+            using reference = T&;
+            using const_reference = T const&;
+            using size_type = std::size_t;
+            using difference_type = std::ptrdiff_t;
```
hkaiser, line 55: Some of those are obsolete (as of C++20, see https://en.cppreference.com/w/cpp/memory/allocator). It's better to provide a specialization for `std::allocator_traits` instead (if needed at all - I don't think it's needed).

msimberg: Thanks, I learned something new! I'll remove the ones that aren't needed.

msimberg: I looked into this a bit more and I think we can keep (most of) these. What's deprecated are the typedefs for `std::allocator`, and they should presumably be accessed through `std::allocator_traits` instead. What doesn't seem to be deprecated is using typedefs for `value_type` etc. in a custom allocator. `std::allocator_traits` can be specialized, or it gets a particular typedef from the allocator itself (https://en.cppreference.com/w/cpp/memory/allocator_traits). Not having those typedefs in the allocator also means they're not available for use in the member function signatures. Using `std::allocator_traits` in the member function signatures requires the specialization again (otherwise it uses an incomplete type). On top of this the `hpx::compute::traits::allocator_traits` depends on the typedefs. That could be rewritten to use `std::allocator_traits` but I think that belongs in a separate PR.

All of this sounds complicated enough that I must be doing something wrong. Examples of specializing `std::allocator_traits` are scarce...

libs/compute/include/hpx/compute/host/block_allocator.hpp
msimberg: This was the only that could use a customization point. I've updated it. `bulk_construct` uses a partitioner.

libs/execution/include/hpx/execution/executors/static_chunk_size.hpp
msimberg: I should note that the idea for powers of two comes from Kokkos. Kokkos gets by with much higher number of chunks per core (up to 100), because they don't end up with a task per chunk.

We could make this 4-8 which would mean *at least* as many chunks per core as we have on master right now.

libs/compute/include/hpx/compute/host/block_allocator.hpp
hkaiser: Leaving the typedefs here is not a problem at all. The only reason I mentioned it was that we should not rely on those to be available in code referring to the allocator's properties. The best might be to make the deprecated typedefs `private`.

msimberg: Thanks, that makes sense. If it's ok with you I'd like to leave it like this here and open a separate PR to make `hpx::compute::traits::allocator_traits` not depend on the typedefs.

libs/algorithms/tests/regressions/count_3646.cpp
```diff
@@ -39,6 +39,32 @@ struct BitCountingIterator : public Iterator<std::int64_t, stopValue>
...
+        return *this;
+    }
+
+    BitCountingIterator operator--(int)
+    {
+        auto copy = *this;
+        --(*this);
+        return copy;
+    }
+
```
msimberg, line 67: This test curiously started failing. It was just lucky with the old chunk size though. For some reason the parallel policy lead to the superclass (`Iterator`) being dereferenced ([here](https://github.com/STEllAR-GROUP/hpx/blob/3df4e4489ff5a360a5838f88b3782489c614827f/libs/algorithms/include/hpx/parallel/util/loop.hpp#L172-L175), the first one dereferences correctly and the others don't). Can someone see why it wouldn't use the dereference operator from `BitCountingIterator`?

hkaiser: I'll have a look

msimberg: Thank you! Note that I've force pushed in the meantime.

hkaiser: @msimberg I'm not able to reproduce the issue locally :/ No idea what's wrong...

msimberg: I should've said it explicitly, but these changes fix the problem. I just don't understand why it doesn't work correctly without. Did you run the latest commit on this branch?

msimberg: The number of threads also affects the result... because of its influence on the chunk size.

hkaiser: Yes, I ran the latest commit on this branch.

hkaiser: Ok, you added `operator++()` and friends. Without them the code was using the corresponding operators implemented by the base class. Those however return `Iterator` which caused all subsequent operations to directly work with the base class only...

msimberg: Uhh, right. For some reason I expected them to return a `BitCountingIterator` cast to `Iterator`, but looking at it again I'm not sure how that would ever happen. Then this makes more sense, thanks!

PR #4538, opened 2020-04-17
Move channel documentation examples to examples directory

### LINKED ISSUE #4519: Documentation problem
From IRC: 
```
[13:23]	tiagofg[m]: I noticed that in HPX documentation, in the 
section "2.5.6 Writing single-node HPX applications", in "channels", 
the example that is there has a mistake. When I do "cout << c.get();" 
will fail, because get() return a future.
[13:25]	tiagofg[m]: to obtain the value 42, I think the solution 
may be "cout << c.get().get();" or "cout << c.get(hpx::launch::sync);"
```

msimberg:
Fixes #4519. @tiagofgon would you mind looking through this (I know you can't see the final result without building it but the examples are marked with `//[start` and `//]` comments in the two new examples)?

I do wish we had a better way of doing this. It may be possible to go the other way around (i.e. examples in docs and extract them into source files).

### DIFF
--- a/examples/quickstart/local_channel_docs.cpp
+++ b/examples/quickstart/local_channel_docs.cpp
@@ -1,0 +1,63 @@
+//  Copyright (c) 2016 Hartmut Kaiser
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// This example is meant for inclusion in the documentation.
+
+#include <hpx/hpx_main.hpp>
+#include <hpx/include/apply.hpp>
+#include <hpx/include/iostreams.hpp>
+#include <hpx/include/lcos.hpp>
+
+#include <numeric>
+#include <string>
+#include <vector>
+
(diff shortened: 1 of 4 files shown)


examples/quickstart/local_channel_docs.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <string>
+#include <vector>
+
+///////////////////////////////////////////////////////////////////////////////
+void minimal_channel()
+{
+    //[local_channel_minimal
+    hpx::lcos::local::channel<int> c;
+    hpx::future<int> f = c.get();
+    HPX_ASSERT(f.is_ready() == false);
```
msimberg, line 24: @tiagofgon do the assertions confuse more than they help?

hkaiser: If anything, I'd write this as `HPX_ASSERT(!f.is_ready());`, but a comment might help.

tiagofgon: I think it help, it's understandable. With the next examples become clear what's pretends.

msimberg: Autopilot... I update the assertions.

PR #4540, opened 2020-04-18
Making sure MPI support is enabled if MPI futures are used but networking is disabled

### LINKED ISSUE #4539: Compilation Error when HPX_MPI_WITH_FUTURES=ON
## Expected Behavior

Compilation succeeds.

## Actual Behavior

Compilation fails with:

```
/home/teonnik/code/hpx/src/hpx_init.cpp:391:23: error: no member named 'mpi_environment' in namespace 'hpx::util'
                util::mpi_environment::init(&argc, &argv, cms);
                ~~~~~~^
/home/teonnik/code/hpx/src/hpx_init.cpp:393:52: error: no member named 'mpi_environment' in namespace 'hpx::util'
                    static_cast<std::size_t>(util::mpi_environment::size());
```

Some of the flags are probably irrelevant:

```
cmake -D CMAKE_BUILD_TYPE=Debug \
       -D BUILD_SHARED_LIBS=ON \
       -D CMAKE_INSTALL_PREFIX=<install_dir> \
       -D HPX_WITH_CXX14=ON \
       -D HPX_WITH_CUDA=ON \
       -D HPX_WITH_MALLOC=tcmalloc \
       -D TCMALLOC_ROOT=<tcmalloc_dir> \
       -D HWLOC_ROOT=<hwloc_dir> \
       -D BOOST_ROOT=<boost_dir> \
       -D MPI_ROOT=<mpi_dir> \
       -D HPX_MPI_WITH_FUTURES=ON \
       -D HPX_WITH_MAX_CPU_COUNT=128 \
       -D HPX_WITH_NETWORKING=OFF \
       -D HPX_WITH_EXAMPLES=OFF \
       -D HPX_WITH_TESTS=OFF \
       -D HPX_WITH_APEX=ON \
       -D APEX_WITH_OTF2=ON \
       -D OTF2_ROOT=<otf_dir> \
       <src_dir>
```

## Specifications

  - HPX Version: master branch, commit f829319ed1ce76e433fb6c6c5242df762566106d
  - compiler: clang 9.0.1
  - cmake: 3.17.1
  - mpich: 3.3.2
  - hwloc: 2.2.0
  - boost: 1.72.0
  - gperftools: 2.7-1
  - otf2: 2.2
  - OS: Arch Linux
  - Kernel: x86_64 Linux 5.6.4-arch1-1
  - CPU: Intel Core i7-5600U @ 4x 3.2GHz
  - GPU: Intel Corporation HD Graphics 5500

hkaiser:
Fixes #4539


hkaiser:
@msimberg looks like we need to adapt some of the pycicle builders to explicitly set `HPX_MPI_WITH_FUTURES=OFF`.

hkaiser:
> @msimberg looks like we need to adapt some of the pycicle builders to explicitly set HPX_MPI_WITH_FUTURES=OFF.

Alternatively, we could load the MPI module such that it can be found by the failing configurations.

CMakeLists.txt
```diff
@@ -788,13 +788,14 @@ if(HPX_WITH_NETWORKING)
...
-      # below does not add the extra options to the top level directory
-      hpx_add_config_define(OMPI_IMPORTS)
-    endif()
   endif()
 
+  # we have to decide here whether we want to have MPI futures
+  hpx_option(HPX_MPI_WITH_FUTURES BOOL
+    "Enable support for returning futures from MPI asynchronous calls"
+    " (default: ON if networking+MPI already enabled)"
+    ON CATEGORY "MPI")
```
msimberg, line 797: I don't think the description matches the behaviour. Now MPI futures are on whenever networking is enabled. To keep things simple we can have the MPI parcelport and MPI futures options be independent. If one is using MPI futures one is unlikely to use the parcelport as well.

plugins/parcelport/mpi/CMakeLists.txt
```diff
@@ -53,4 +55,25 @@ if(HPX_WITH_PARCELPORT_MPI)
...
+      hpx_command_line_handling
+      hpx_runtime_configuration
+      hpx_synchronization
+      hpx_util
+      hpx_dependencies_boost
+      Mpi::mpi
+    INCLUDE_DIRS
+      ${PROJECT_SOURCE_DIR}
+    FOLDER
+      "Core/Plugins/Parcelport/MPI")
```
msimberg, line 78: We should probably move `mpi_environment.cpp` to the MPI module. Having that file here also makes the MPI parcelport not a plugin (even though it's already statically in libhpx).

tests/headers/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ set(exclude_from_headers
   "${PROJECT_SOURCE_DIR}/hpx/lcos/local/futures_factory.hpp"
 )
 
-if(HPX_WITH_PARCELPORT_MPI)
+if(HPX_WITH_NETWORKING AND HPX_WITH_PARCELPORT_MPI)
```
msimberg, line 19:
```suggestion
if(HPX_WITH_NETWORKING OR HPX_WITH_PARCELPORT_MPI)
```

but we can also do `if(TARGET Mpi::mpi)` (elsewhere as well) assuming the target is set up early enough.

plugins/parcelport/mpi/CMakeLists.txt
hkaiser: That makes the parcel-port depend on the MPI module, is that what we want?

tests/headers/CMakeLists.txt
hkaiser: No, NETWORKING should always be enabled if the parcel-port is.

CMakeLists.txt
hkaiser: If NETWORKING is enabled, the default for the MPI module is to be enabled as well. If NETWORKING is disabled, the default for the MPI module is to be disabled (see: https://github.com/STEllAR-GROUP/hpx/pull/4540/files/8fcea143f8d9291d2176472e06ad1b816b5d8fe2#diff-af3b638bc2a3e6c650974192a53c7291R814-R817)

plugins/parcelport/mpi/CMakeLists.txt
msimberg: I'd rather have that than the other way around (ok, it's command line handling depending on the MPI parcelport). If necessary we can have `mpi_base` with the MPI environment handling, and have the MPI futures module and the MPI parcelport depend on that.

CMakeLists.txt
msimberg: The MPI part is missing. Now `HPX_MPI_WITH_FUTURES` is `ON` if networking is on even if the MPI parcelport is off. That's why most of the pycicle configurations fail (no MPI, except for implicitly on the CUDA builds because they use the Cray compilers). I'm not sure why the configuration without networking still wants MPI though (https://cdash.cscs.ch//viewConfigure.php?buildid=104086).

msimberg: I interpret `ON if networking+MPI already enabled` as networking and MPI parcelport by the way.

tests/headers/CMakeLists.txt
hkaiser: FWIW, I think I added the check everywhere.

plugins/parcelport/mpi/CMakeLists.txt
hkaiser: Ok. I'll see what I can do.

CMakeLists.txt
hkaiser: > The MPI part is missing. Now HPX_MPI_WITH_FUTURES is ON if networking is on even if the MPI parcelport is off. That's why most of the pycicle configurations fail (no MPI, except for implicitly on the CUDA builds because they use the Cray compilers). I'm not sure why the configuration without networking still wants MPI though (https://cdash.cscs.ch//viewConfigure.php?buildid=104086).

Makes sense now! I was wondering what's going on.... I'll try to fix this.

hkaiser:
@msimberg I have addressed your comments, please review one more time.

plugins/parcelport/CMakeLists.txt
```diff
@@ -15,6 +15,9 @@ if(HPX_WITH_NETWORKING)
     verbs
     mpi
     tcp)
+elseif(HPX_MPI_WITH_FUTURES)
+  set(parcelport_plugins ${parcelport_plugins}
+    mpi)
```
msimberg, line 20: Is this still needed? The MPI parcelport is anyway guarded by `HPX_WITH_NETWORKING`.

libs/version/CMakeLists.txt
```diff
@@ -12,7 +12,8 @@ set(version_headers hpx/version.hpp)
 
 set(version_sources version.cpp)
 
-if(HPX_WITH_PARCELPORT_MPI)
+if((HPX_WITH_NETWORKING AND HPX_WITH_PARCELPORT_MPI) OR HPX_MPI_WITH_FUTURES)
+  include(HPX_SetupMPI)
```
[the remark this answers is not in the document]
aurianer: Apparently to fix some compilation issues (see [here](https://github.com/STEllAR-GROUP/hpx/commit/ed56387187e)) but we may add it directly to the `hpx` target as the mpi include is in `src/hpx_init.cpp`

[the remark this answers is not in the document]
msimberg: Ah, I meant if you know why we include `HPX_SetupMPI` and then use the macro, instead of just including `HPX_SetupMPI` (and removing the macro of course)?

libs/thread_executors/CMakeLists.txt
```diff
@@ -50,6 +51,7 @@ set(thread_executors_compat_headers
 
 set(thread_executors_sources
   embedded_thread_pool_executors.cpp
+  force_linking.cpp
```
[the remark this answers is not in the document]
hkaiser: One of the examples (`quick_start/fractals_executor`) didn't link on Windows.

.circleci/config.yml
```diff
@@ -208,6 +211,7 @@ jobs:
             cmake \
                 /hpx/source \
                 -G "Ninja" \
+                -DHPX_MPI_WITH_FUTURES=Off \
```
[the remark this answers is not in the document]
hkaiser: @msimberg I tried this but ran into issues. Could you help with this?

libs/threading_base/include/hpx/threading_base/set_thread_state.hpp
```diff
@@ -246,6 +253,7 @@ namespace hpx { namespace threads { namespace detail {
             }
 
             // state has changed since we fetched it from the thread, retry
+            // NOLINTNEXTLINE(bugprone-branch-clone)
```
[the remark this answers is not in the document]
hkaiser: Yes, I was surprised. Even more as the 'errors' reported are false positives. Did the clang version we use on CircleCI change recently?

plugins/parcelport/CMakeLists.txt
hkaiser: You are right, this is not needed anymore.

libs/version/CMakeLists.txt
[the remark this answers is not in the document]
hkaiser: How do you suggest to proceed? Should we leave this to another PR?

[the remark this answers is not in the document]
msimberg: Let's leave it as it is in this PR. It's not critical to change it now.

libs/threading_base/include/hpx/threading_base/set_thread_state.hpp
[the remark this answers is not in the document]
msimberg: Ah, wait, that's clang-tidy. It hasn't changed after the change from 8 to 9 a month or two ago... curious.

.circleci/config.yml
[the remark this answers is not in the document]
msimberg: Sure, I'll give it go.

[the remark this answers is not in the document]
msimberg: @hkaiser `" (default: ON if networking+MPI already enabled)"` is not a good default value for boolean CMake option ;) We can do multiline strings like this in CMake:

```
"Enable support for returning futures from MPI asynchronous calls \
(default: ON if networking+MPI already enabled)"
```

but the second line has can't be indented unless we want a bunch of extra spaces in there. I don't have a preference.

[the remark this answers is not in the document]
hkaiser: Thanks for applying this change. What should we do with the `-DHPX_MPI_WITH_FUTURES=Off` in the CIs?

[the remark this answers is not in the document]
msimberg: Ah, sorry. I meant to remove those as well. I tested it locally and didn't need to explicitly turn it off there. I'll remove them as well to make sure things are ok on CircleCI and friends.

msimberg:
@hkaiser I just pushed a commit to use the wrapped mpi header instead of `mpi.h` directly to get rid of warnings. Otherwise this looks good to go now.

hkaiser:
@msimberg: The pycicle errors are unrelated. I think this is finally good to go.

biddisco:
Why are the commits
- Breaking indirect circular dependency between command_line_handling and runtime_configuration
- Making sure thread_executors are actually linked into the core library
part of this PR? Is there any reason why they can't go into their own PR. I would like to make some tweaks to this PR, but these commits look out of place to me

hkaiser:
> Why are the commits
> 
> Breaking indirect circular dependency between command_line_handling and runtime_configuration
> Making sure thread_executors are actually linked into the core library
> 
> part of this PR? Is there any reason why they can't go into their own PR. I would like to make some tweaks to this PR, but these commits look out of place to me

You are right. The second one is unrelated. The first one however is something that needs to be fixed for this PR to pass the circular dependency tests. 

In any case, I'm going to merge this PR now as everything seems to be under control.

biddisco:
I wanted investigate this crash and to make sure the tests passed with this PR before it was merged.
```
~/apps/openmpi/4.0.2/bin/mpiexec -n 2 --oversubscribe bin/mpi_ring_async_test
Rank   0 of   2 scheduler is core-local_priority_queue_scheduler


Rank   0 of   2 Sent token   1 to   rank   1 tag   2
Rank   0 of   2 Sent token   1 to   rank   1 tag   5
Rank   0 of   2 Sent token   1 to   rank   1 tag   3
Rank   0 of   2 Sent token   1 to   rank   1 tag   6
Rank   0 of   2 Sent token   1 to   rank   1 tag   8
Rank   0 of   2 Sent token   1 to   rank   1 tag  11
Rank   0 of   2 Sent token   1 to   rank   1 tag   9
Rank   0 of   2 Sent token   1 to   rank   1 tag  10
Rank   0 of   2 Sent token   1 to   rank   1 tag  12
Rank   0 of   2 Sent token   1 to   rank   1 tag  16
Rank   0 of   2 Sent token   1 to   rank   1 tag  17
Rank   0 of   2 Sent token   1 to   rank   1 tag  14
Rank   0 of   2 Sent token   1 to   rank   1 tag  13
Rank   0 of   2 Sent token   1 to   rank   1 tag   1
Rank   0 of   2 Sent token   1 to   rank   1 tag   4
Rank   0 of   2 Sent token   1 to   rank   1 tag  15
Rank   1 of   2 Recv token   1 from rank   0 tag   3
Rank   0 of   2 Sent token   1 to   rank   1 tag  18
Rank   0 of   2 Sent token   1 to   rank   1 tag   7
Rank   1 of   2 Recv token   1 from rank   0 tag   4
Rank   1 of   2 Sent token   2 to   rank   0 tag   3
Rank   1 of   2 Recv token   1 from rank   0 tag   7
Rank   1 of   2 Recv token   1 from rank   0 tag   2
Rank   1 of   2 Recv token   1 from rank   0 tag  10
Rank   1 of   2 Sent token   2 to   rank   0 tag   4
Rank   1 of   2 Recv token   1 from rank   0 tag   0
Rank   1 of   2 Recv token   1 from rank   0 tag  12
Rank   1 of   2 Recv token   1 from rank   0 tag   6
Rank   1 of   2 Recv token   1 from rank   0 tag  16
Rank   1 of   2 Recv token   1 from rank   0 tag   9
Rank   1 of   2 Recv token   1 from rank   0 tag  19
Rank   1 of   2 Sent token   2 to   rank   0 tag  12
Rank   0 of   2 Recv token   2 from rank   1 tag   4
Rank   0 of   2 Recv token   2 from rank   1 tag  12
Rank   1 of   2 Sent token   2 to   rank   0 tag   7
Rank   1 of   2 Sent token   2 to   rank   0 tag  10
Rank   1 of   2 Sent token   2 to   rank   0 tag   9
Rank   1 of   2 Recv token   1 from rank   0 tag  15
Rank   1 of   2 Sent token   2 to   rank   0 tag   2
Rank   1 of   2 Sent token   2 to   rank   0 tag  16
Rank   1 of   2 Sent token   2 to   rank   0 tag  15
Rank   1 of   2 Recv token   1 from rank   0 tag  14
Rank   1 of   2 Recv token   1 from rank   0 tag  17
Rank   1 of   2 Sent token   2 to   rank   0 tag   0
Rank   0 of   2 Sent token   1 to   rank   1 tag   0
Rank   1 of   2 Recv token   1 from rank   0 tag   5
Rank   1 of   2 Recv token   1 from rank   0 tag   1Rank   1 of   2 Recv token   1 from rank   0 tag  11

Rank   1 of   2 Sent token   2 to   rank   0 tag  19
Rank   1 of   2 Recv token   1 from rank   0 tag  13
Rank   1 of   2 Recv token   1 from rank   0 tag  18
Rank   1 of   2 Sent token   2 to   rank   0 tag   6
Rank   1 of   2 Recv token   1 from rank   0 tag   8
Rank   1 of   2 Sent token   2 to   rank   0 tag  17
Rank   1 of   2 Sent token   2 to   rank   0 tag   5
Rank   1 of   2 Sent token   2 to   rank   0 tag  13
Rank   1 of   2 Sent token   2 to   rank   0 tag  11
Rank   1 of   2 Sent token   2 to   rank   0 tag   8
Rank   1 of   2 Sent token   2 to   rank   0 tag  18
Rank   1 of   2 Sent token   2 to   rank   0 tag   1
Rank   1 of   2 Sent token   2 to   rank   0 tag  14Rank   0 of   2 Recv token   2 from rank   1 tag   3
Rank   0 of   2 Recv token   2 from rank   1 tag   2
Rank   0 of   2 Recv token   2 from rank   1 tag  10
Rank   0 of   2 Recv token   2 from rank   1 tag   7
Rank   0 of   2 Recv token   2 from rank   1 tag  17
Rank   0 of   2 Recv token   2 from rank   1 tag   1
Rank   0 of   2 Recv token   2 from rank   1 tag   9
Rank   0 of   2 Recv token   2 from rank   1 tag   0
Rank   0 of   2 Sent token   2 to   rank   1 tag  19
Rank   0 of   2 Recv token   2 from rank   1 tag  11
Rank   0 of   2 Recv token   2 from rank   1 tag  16
Rank   0 of   2 Recv token   2 from rank   1 tag   6
Rank   0 of   2 Recv token   2 from rank   1 tag  18

Rank   0 of   2 Recv token   2 from rank   1 tag   5
Rank   0 of   2 Recv token   2 from rank   1 tag  15
Rank   0 of   2 Recv token   2 from rank   1 tag  19
Rank   0 of   2 Recv token   2 from rank   1 tag   8
Rank   0 of   2 Recv token   2 from rank   1 tag  13
Rank   0 of   2 Recv token   2 from rank   1 tag  14
--------------------------------------------------------------------------
mpiexec has exited due to process rank 0 with PID 0 on
node pop-os exiting improperly. There are three reasons this could occur:

1. this process did not call "init" before exiting, but others in
the job did. This can cause a job to hang indefinitely while it waits
for all processes to call "init". By rule, if one process calls "init",
then ALL processes must call "init" prior to termination.

2. this process called "init", but exited without calling "finalize".
By rule, all processes that call "init" MUST call "finalize" prior to
exiting or it will be considered an "abnormal termination"

3. this process called "MPI_Abort" or "orte_abort" and the mca parameter
orte_create_session_dirs is set to false. In this case, the run-time cannot
detect that the abort call was an abnormal termination. Hence, the only
error message you will receive is this one.

This may have caused other processes in the application to be
terminated by signals sent by mpiexec (as reported here).

You can avoid this message by specifying -quiet on the mpiexec command line.
--------------------------------------------------------------------------
```

hkaiser:
@biddisco ok, I'd appreciate if you still looked into this. Everything worked for me locally.

PR #4545, opened 2020-04-21
Fix header tests

msimberg:
Fixes header test failures that I missed when merging #4521.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -500,32 +500,6 @@ jobs:
       - store_artifacts:
           path: tests.unit.lcos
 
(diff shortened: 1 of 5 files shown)


msimberg:
I've fixed the inspect failure now. Unfortunately there's now a test failure that wasn't there on the original PR and I don't understand why it would show up now. It may have been there all along, but we have to wait for CircleCI to cycle once to see.

msimberg:
Looks like the one failure from the previous run is not showing up again. I'll take it was a spurious failure (or a circleci environment problem). Once (if) the CircleCI builder finishes anyone can feel free to merge this.

PR #4547, opened 2020-04-22
Fix a warning caused by a misplaced HPX_EXPORT

biddisco:
Put the HPX_EXPORT on the original declaration

### DIFF
--- a/hpx/runtime.hpp
+++ b/hpx/runtime.hpp
@@ -57,7 +57,7 @@ namespace hpx
         namespace server
         {
             class runtime_support;
-            class HPX_EXPORT memory;
+            class memory;
         }
     }
 
--- a/hpx/runtime/components/server/memory.hpp
+++ b/hpx/runtime/components/server/memory.hpp
@@ -23,7 +23,7 @@
 namespace hpx { namespace components { namespace server
 {
     ///////////////////////////////////////////////////////////////////////////
-    class memory
+    class HPX_EXPORT memory
     {
     public:
         typedef memory type_holder;


hkaiser:
> Thanks! We need to do a round of cleaning up those warnings before the next release. Or maybe this was the only one? I wonder if we can just remove the forward declaration in runtime.hpp. Would you mind trying that? The warnings would depend on in which order those two files are included, but it may be that we always end up including them in a certain order.

As long as we have the `HPX_EXPORT` in the forwarding header and #include it from the header that has the actual definition the sequence of inclusion shouldn't matter.

msimberg:
> > Thanks! We need to do a round of cleaning up those warnings before the next release. Or maybe this was the only one? I wonder if we can just remove the forward declaration in runtime.hpp. Would you mind trying that? The warnings would depend on in which order those two files are included, but it may be that we always end up including them in a certain order.
> 
> As long as we have the `HPX_EXPORT` in the forwarding header and #include it from the header that has the actual definition the sequence of inclusion shouldn't matter.

Exactly, and this PR changes it to the opposite (`HPX_EXPORT` on the definition, not in the forwarding header), which makes me worried.

hkaiser:
> Exactly, and this PR changes it to the opposite (HPX_EXPORT on the definition, not in the forwarding header), which makes me worried.

Uhh, indeed. I understood it the other way around :/

biddisco:
FYI removing the declaration from hpx/runtime.hpp
```
In file included from /home/biddisco/src/hpx-branches/dataflow-annotations/libs/compute/src/host_target.cpp:11:
/home/biddisco/src/hpx-branches/dataflow-annotations/libs/compute/../../hpx/runtime.hpp:351:45: error: ‘memory’ is not a member of ‘hpx::components::server’; did you mean ‘hpx::memory’?
  351 |         std::unique_ptr<components::server::memory> memory_;
      |                                             ^~~~~~
```
+ more like it

msimberg:
Thanks! This may be a rabbit hole of circular dependencies, but you could also try including `hpx/runtime/components/server/memory.hpp` directly where it's needed (in this case `runtime.hpp`). The forward declaration is probably there to avoid having to pull in that whole header. Otherwise `hpx/runtime/components/server/memory.hpp` should directly include `runtime.hpp`. Or the forward declarations should potentially even be moved to `hpx/runtime_fwd.hpp`.

biddisco:
Don't think we need this PR any more.

Issue #4542, opened 2020-04-21
Move thread creation parameters into a struct


msimberg:
This is about all the overloads of `register_thread`, `futures_factory`, `async_launch_policy_dispatch` etc. that take among other things a scheduling hint, priority, and stack size. This new struct is not quite `thread_init_data` because we do the conversion from any function to the thread function type, as well as converting an enum stack size to the real stack size. We could perhaps merge `thread_init_data` and this new struct.

I've already started some of this cleanup but wanted to open this issue to see if others have objections or better ideas. I would like to use this new struct everywhere and remove the overloads taking individual arguments. In my opinion the only overloads that might deserve a compatibility option are the `register_thread` overloads.

hkaiser:
@msimberg yes, let's merge the structures.

PR #4548, opened 2020-04-22
Fix stack size

### LINKED ISSUE #4543: Segfaults in Release builds using `sleep_for`
Testing our code on different platforms, we noticed that on some hardware for some tests we need a slightly larger stack size (`hpx.stacks.small_size`). However when increasing the small stack size one of the test started to systematically fail in *release* mode.

From our tests we extracted the failing case and reduced to the following minimal reproducible example, which also work only when the default value (0x10000) is used.

The problem was reproduced on different platforms and compilers (linux + gcc, mac + clang)

## Expected Behavior
We expect it should work.

## Actual Behavior

```
➜  build git:(bug) ✗ ./example
➜  build git:(bug) ✗ ./example -Ihpx.stacks.small_size=0x80000
{stack-trace}: 5 frames:
{what}: Segmentation fault: 11
...
```

See [full_error.txt](https://github.com/STEllAR-GROUP/hpx/files/4510069/full_error.txt)

```cpp
#include <chrono>
#include <hpx/hpx.hpp>

int main_hpx(int argc, char** argv) {
  hpx::shared_future<void> shared_future = hpx::make_ready_future();

  hpx::future<void> last_task =
      shared_future.then(hpx::launch::async, [](auto&&) { hpx::this_thread::sleep_for(std::chrono::milliseconds(10)); });

  last_task.get();

  return hpx::finalize();
}

int main(int argc, char** argv) {
  return hpx::init(main_hpx, argc, argv);
}
```

## Specifications

  - HPX Version: 1.4.0
  - Platform (compiler, OS): `Apple LLVM version 10.0.1` on `macOS 10.14.6`

msimberg:
Fixes #4543. The changes are a bit awkard, so as I said in #4543 I'll change `thread_init_data` to use the enum stacksize along with #4542. I also changed the `thread_schedule_hint` one-argument constructor to be explicit to avoid having me pass a stack size for the hint...

### DIFF
--- a/examples/quickstart/component_with_executor.cpp
+++ b/examples/quickstart/component_with_executor.cpp
@@ -24,9 +24,8 @@ struct hello_world_server
             executor_type, hpx::components::component_base<hello_world_server>
         > base_type;
 
-    // run on all available cores
     hello_world_server()
-      : base_type(executor_type(hpx::get_num_worker_threads()))
+      : base_type(executor_type{})
     {}
 
     void print() const
--- a/libs/coroutines/include/hpx/coroutines/thread_enums.hpp
+++ b/libs/coroutines/include/hpx/coroutines/thread_enums.hpp
@@ -199,7 +199,7 @@ namespace hpx { namespace threads {
         {
         }
 
-        constexpr thread_schedule_hint(std::int16_t thread_hint)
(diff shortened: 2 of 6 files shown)


msimberg:
Updated this. One of the examples was passing a number of threads as a scheduling hint. That stopped working after the schedule hint constructor turned explicit.

PR #4551, opened 2020-04-23
Clean up thread creation parameters

msimberg:
This is WIP, or I can split this into multiple PRs. Also, I've only tested this building `core`, not the examples and tests yet. I'll update those if people are happy with this.

What I've done so far:
- Add `initial_state` and `run_now` to `thread_init_data`
- Remove `initial_state` and `run_now` from all the `create_thread/work` functions (thread pool, schedulers, thread queues, wrappers in `create_thread/work.hpp`)
- Disable all `register_*` overloads except for `register_thread/work` which take a pool and `thread_init_data` (I'll hide these behind a real compatibility option, I've just disabled them now to keep things simple)
- Use the reduced set of `register_thread/work` in the core library
- Add `make_thread_function{,_nullary}` for creating appropriate thread functions (the caller of `register_*` is responsible for conversion

To do for this PR:
- [x] Add a real CMake compatibility option
- [x] I won't do this in this PR: Reduce set of overloads even further? For this I would be tempted to go with only one overload: `thread_id_type register_thread(thread_init_data, thread_pool_base = /* default */, error_code = /* default */)`. The reason for this is that I'm not sure the `register_thread/work` distinction has been followed very well. There were cases of `register_thread` with `run_now = false` and vice versa. I think it may be clearer to just have one version, with the returned thread id being the invalid thread id if `run_now = false`. What do other people think about this? I'm happy to go either way as long as people find the behaviour clear.

Edit: I'll clean this PR up and leave the following for a follow-up PR:
- Extend the use of `thread_init_data` to `future_data`, `async/post_launch_policy_dispatch`, and whatever else is passing priorities etc. as separate function arguments.


msimberg:
There is actually a use for not quite going all in with `thread_init_data` above the `register_*` functionality: the function itself might be known quite late. E.g. `futures_factory` wraps the actual function to call. So we may want the function or the launch parameters separate from `thread_init_data`. That way we can pass launch parameters without a function attached to it.

msimberg:
I think I'm done with this PR for this round, modulo build and inspect errors that may still come up. I'll leave cleaning up the launch parameters higher up than `register_thread` to a second PR. I now left `register_thread` and `register_work` as non-deprecated functions (but they only have two overloads each). `register_thread` and `register_work` set `run_now` to the correct value. If one wants different behaviour one can still directly call `create_thread/work`.

libs/async/include/hpx/async/applier/apply_helper.hpp
```diff
@@ -77,9 +77,8 @@ namespace hpx { namespace applier { namespace detail {
 #endif
 #endif
         data.priority = fix_priority<Action>(priority);
-        data.stacksize =
-            threads::get_stack_size(static_cast<threads::thread_stacksize>(
-                traits::action_stacksize<Action>::value));
+        data.stacksize = static_cast<threads::thread_stacksize>(
+            traits::action_stacksize<Action>::value);
```
hkaiser, line 81: Should we make `traits::action_stacksize<Action>::value` to be of type `enum thread_stacksize` instead of the `static_cast` (I know this is old code, but I thought I'd mention it anyways)?

hpx/traits/action_schedule_thread.hpp
```diff
@@ -46,6 +47,7 @@ namespace hpx { namespace traits
             {
                 // by default we forward this to the component type
                 typedef typename Action::component_type component_type;
+                data.initial_state = initial_state;
                 component_type::schedule_thread(lva, comptype, data, initial_state);
```
hkaiser, line 51: While removing the `initial_state` argument here might be a breaking change, I think we should go for it as most probably there will not be a lot of code out there relying on this customization point.

libs/schedulers/include/hpx/schedulers/queue_holder_thread.hpp
```diff
@@ -134,8 +134,7 @@ namespace hpx { namespace threads { namespace policies {
 
         static util::internal_allocator<threads::thread_data> thread_alloc_;
 
-        typedef util::tuple<thread_init_data, thread_state_enum>
-            task_description;
+        typedef util::tuple<thread_init_data> task_description;
```
hkaiser, line 137: Does this still have to be a `tuple` now?

libs/threading_base/include/hpx/threading_base/register_thread.hpp
```diff
@@ -93,6 +93,22 @@ namespace hpx { namespace threads {
         };
     }    // namespace detail
 
+    template <typename F>
+    auto make_thread_function(F&& f)
```
hkaiser, line 97: Could you write this as `decltype(auto)` please?

libs/threading_base/include/hpx/threading_base/register_thread.hpp
```diff
@@ -93,6 +93,22 @@ namespace hpx { namespace threads {
...
+    {
+        return thread_function_type(
+            detail::thread_function<typename std::decay<F>::type>{
+                std::forward<F>(f)});
+    }
+
+    template <typename F>
+    auto make_thread_function_nullary(F&& f)
```
hkaiser, line 105: Same here, please use `decltype(auto)` instead.

libs/threading_base/include/hpx/threading_base/register_thread.hpp
hkaiser: Mostly for consistency/cosmetics, `auto` yields the same result in this case.

hkaiser: It might be evenbetter to write this as:
```
template <typename F>
thread_function_type make_thread_function(F&& f)
{
    return {detail::thread_function<typename std::decay<F>::type>{
        std::forward<F>(f)}};
}
```

tests/performance/local/htts_v2/htts2_hpx.cpp
```diff
@@ -120,13 +120,22 @@ struct hpx_driver : htts2::driver
...
+                      , nullptr, hpx::threads::thread_priority_low);
+                register_work(data1);
+                hpx::threads::thread_init_data data2(
+                        hpx::threads::make_thread_function_nullary(
+                            hpx::util::bind(&hpx_driver::wait_for_tasks
+                                      , std::ref(*this)
+                                      , std::ref(finished)
+                                       ))
+                      , nullptr, hpx::threads::thread_priority_low);
+                register_work(data2);
```
hkaiser, line 138: What's the rationale of duplicating this code?

libs/async/include/hpx/async/applier/apply_helper.hpp
msimberg: Good point. It seems the action stacksize is already using the thread stacksize enum, but inside another enum... Will change this.

libs/schedulers/include/hpx/schedulers/queue_holder_thread.hpp
msimberg: No. I've changed other instances, but missed this one. Thanks!

libs/threading_base/include/hpx/threading_base/register_thread.hpp
msimberg: The second option looks good to me. I'll change that.

tests/performance/local/htts_v2/htts2_hpx.cpp
msimberg: Very good question... there's none. Good catch!

msimberg:
> Just an observation: currently almost all thread functions are explicitly wrapped by a `make_thread_function_nullary()`. Wouldn't it be possible to make that the default to reduce the amount of code users have to write?

Yes, that's a fair point. `thread_init_data` could easily get a constructor for functions with the signature `void()`. I quite like that the conversion is explicit on the calling side though, and in this case users = us mostly. That said, I can definitely add an overload if you'd like.

PR #4552, opened 2020-04-23
Add configuration option for making exceptions less noisy

msimberg:
Adds the option `hpx.exception_verbosity`. It currently takes three values:

- `0`: prints filename, line, function name, and the exception string
- `1`: prints everything except the build configuration and environment variables
- `2`: prints everything

I'm not sure if they need to be shifted to have `0` mean no output, but I don't see that being useful for anyone. We could also use names (like `minimal`, `reduced`, `full`), but using numbers for the level is similar to what we do with logging.

Flyby: fix reading of ini files from `HOME` and `PWD`.

To do:
- [x] documentation


msimberg:
I applied the verbosity setting also in the termination handler now.

PR #4557, opened 2020-04-26
Improve spinlock implementation to perform better in high-contention situations

hkaiser:
This reduces the cache-pressure while acquiring highly contented spinlock instances by mostly reading from it instead of constantly trying to write to it.

As a flyby, this removes the dependency on `boost/smart_ptr/detail/spinlock.hpp` and related Boost facilities.

### DIFF
--- a/hpx/runtime/naming/name.hpp
+++ b/hpx/runtime/naming/name.hpp
@@ -326,17 +326,10 @@ namespace hpx { namespace naming
         {
             HPX_ITT_SYNC_PREPARE(this);
 
-            for (;;)
+            while (!acquire_lock())
             {
-                if (acquire_lock())
-                {
-                    break;
-                }
-
-                for (std::size_t k = 0; is_locked(); ++k)
-                {
-                    util::detail::yield_k(k, "hpx::naming::gid_type::lock");
-                }
+                util::yield_while([this] { return is_locked(); },
+                    "hpx::naming::gid_type::lock");
             }
 
             util::register_lock(this);
--- a/hpx/runtime/naming/name.hpp
+++ b/hpx/runtime/naming/name.hpp
@@ -326,9 +326,17 @@ namespace hpx { namespace naming
         {
             HPX_ITT_SYNC_PREPARE(this);
 
-            for (std::size_t k = 0; !acquire_lock(); ++k)
+            for (;;)
             {
-                util::detail::yield_k(k, "hpx::naming::gid_type::lock");
+                if (acquire_lock())
+                {
+                    break;
+                }
+
+                for (std::size_t k = 0; is_locked(); ++k)
+                {
+                    util::detail::yield_k(k, "hpx::naming::gid_type::lock");
+                }
             }
 
             util::register_lock(this);
--- a/libs/synchronization/include/hpx/synchronization/spinlock.hpp
+++ b/libs/synchronization/include/hpx/synchronization/spinlock.hpp
@@ -50,18 +50,10 @@ namespace hpx { namespace lcos { namespace local {
         {
             HPX_ITT_SYNC_PREPARE(this);
 
-            for (;;)
+            while (!acquire_lock())
             {
-                if (acquire_lock())
-                {
-                    break;
-                }
-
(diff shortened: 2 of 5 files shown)


hpx/runtime/naming/name.hpp
```diff
@@ -326,9 +326,17 @@ namespace hpx { namespace naming
...
-                util::detail::yield_k(k, "hpx::naming::gid_type::lock");
+                if (acquire_lock())
+                {
+                    break;
+                }
+
+                for (std::size_t k = 0; is_locked(); ++k)
+                {
+                    util::detail::yield_k(k, "hpx::naming::gid_type::lock");
+                }
```
msimberg, line 339: `yield_while`?

hpx/runtime/naming/name.hpp
```diff
@@ -326,9 +326,17 @@ namespace hpx { namespace naming
         {
             HPX_ITT_SYNC_PREPARE(this);
 
-            for (std::size_t k = 0; !acquire_lock(); ++k)
+            for (;;)
```
msimberg, line 329: `while (!acquire_lock())`?

libs/synchronization/include/hpx/synchronization/spinlock.hpp
```diff
@@ -69,9 +50,18 @@ namespace hpx { namespace lcos { namespace local {
         {
             HPX_ITT_SYNC_PREPARE(this);
 
-            for (std::size_t k = 0; !acquire_lock(); ++k)
+            for (;;)
```
msimberg, line 53: Same comments as above.

biddisco:
I ran a benchmark to see how much this helps as I wanted to test it before it is merged. 

Something must be wrong with my testing because I see a significant drop in performance
image
Unfortunately, the previous runs of the benchmark were in December, and so many things have changed that  it is hard to be certain what might be responsible, but I'm pasting this image just for reference.

This may be a false reading, so do not be alarmed yet, I will double check my scripts and report back, but I thought I'd mention it now just in case anything else shows problems.

hkaiser:
Uhh, not good. I wouldn't think that this particular change is causing this. So I guess the only thing we can do is to bisect...

biddisco:
I did another build this morning, but this time I turned off everything that might cause overheads
```
  cmake \
      -DCMAKE_BUILD_TYPE=Release \
      -DBOOST_ROOT=$BOOST_ROOT \
      -DHWLOC_ROOT=$INSTALL_ROOT/hwloc/$HWLOC_VER \
      -DHPX_WITH_MALLOC=JEMALLOC \
      -DJEMALLOC_ROOT=$INSTALL_ROOT/jemalloc/$JEMALLOC_VER \
      -DHPX_WITH_NETWORKING:BOOL=OFF \
      -DHPX_WITH_LOGGING:BOOL=OFF \
      -DHPX_WITH_IO_COUNTERS:BOOL=OFF \
      -DHPX_WITH_PARCELPORT_ACTION_COUNTERS:BOOL=OFF \
      -DHPX_WITH_PARCEL_PROFILING:BOOL=OFF \
      -DHPX_WITH_STACKTRACES:BOOL=OFF \
      -DHPX_WITH_THREAD_CUMULATIVE_COUNTS:BOOL=OFF \
      -DHPX_WITH_THREAD_IDLE_RATES:BOOL=OFF \
      -DHPX_WITH_THREAD_MANAGER_IDLE_BACKOFF:BOOL=OFF \
      -DHPX_WITH_THREAD_STEALING_COUNTS:BOOL=OFF \
      -DHPX_WITH_VERIFY_LOCKS:BOOL=OFF \
      -DHPX_WITH_TIMER_POOL:BOOL=OFF \
      -DHPX_WITH_IO_POOL:BOOL=OFF \
      ~/src/hpx
```
The build seems to perform much more in line with expectations - and looks a little better than december

image

Please disregard my previous graph as there must have been something enabled that shouldn't have been.

hkaiser:
@biddisco thanks! This looks more like what I would have expected. Brilliant!

biddisco:
image
For completeness - included is a daint test

PR #4561, opened 2020-04-26
Adding test macros supporting custom output stream

### LINKED ISSUE #4554: Make HPX test infrastructure configurable to use arbitrary stream to print
If two localities are run on the same node, and both print error messages using the `HPX_TEST()` macros and friends, the output streams of the localities will be intermingled as both localities concurrently output to `std::cerr` by default (see: https://github.com/STEllAR-GROUP/phylanx/issues/1150). 

This is to remind ourselves to create facilities similar to `HPX_TEST()` that allow to explicitly specify the stream that will be used for generating the test output.

hkaiser:
Fixes #4554

This adds macro overloads to all `HPX_TEST()` macros allowing to pass any output stream as the respective first arguments. For instance:
```
    HPX_TEST(true);             // uses std::cerr as before
    HPX_TEST(hpx::cout, true);  // uses hpx::cout for generated output
```

### DIFF
--- a/libs/testing/tests/unit/test_macros.cpp
+++ b/libs/testing/tests/unit/test_macros.cpp
@@ -8,17 +8,18 @@
 #include <hpx/testing.hpp>
 
 #include <sstream>
+#include <string>
 
 int main(int argc, char* argv[])
 {
     std::stringstream strm;
 
     // testing macros
-    HPX_TEST(0 == 0);
-    HPX_TEST(strm, 0 == 0);
+    HPX_TEST(true);
+    HPX_TEST(strm, true);
 
-    HPX_TEST_MSG(0 == 0, "should be true");
-    HPX_TEST_MSG(strm, 0 == 0, "should be true");
+    HPX_TEST_MSG(true, "should be true");
+    HPX_TEST_MSG(strm, true, "should be true");
 
     HPX_TEST_EQ(0, 0);
     HPX_TEST_EQ(strm, 0, 0);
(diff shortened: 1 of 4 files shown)


libs/testing/tests/unit/test_macros.cpp
```diff
@@ -0,0 +1,82 @@
...
+    HPX_SANITY_LT(strm, 0, 1);
+
+    HPX_SANITY_LTE(1, 1);
+    HPX_SANITY_LTE(strm, 1, 1);
+
+    HPX_SANITY_RANGE(1, 1, 1);
+    HPX_SANITY_RANGE(strm, 1, 1, 1);
+
+    // there shouldn't be any output being generated
+    HPX_TEST(strm.str().empty());
```
msimberg, line 79: Should there be the opposite test to check that the error messages actually get written to the stream? Perhaps not character-by-character exactly what we expect, but `strm` should at least not be empty if a test fails.

biddisco:
Test change to check stream looks good, but failed on daint. Any idea why?

hkaiser:
> Test change to check stream looks good, but failed on daint. Any idea why?

Those builds are still busy with previous commits. The latest commit should solve the issues reported.

libs/testing/tests/unit/test_macros.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    // now test that something gets written to the stream if an error occurs
+    HPX_TEST(strm, false);
+    HPX_TEST(strm.str().find("test 'false'") != std::string::npos);
+
+    // we have intentionally generated one error
+    return (hpx::util::report_errors() == 1) ? 0 : -1;
```
msimberg, line 87: This is still failing because the failure counters are fixture specific, and `report_errors` reports errors for the global fixture. I suspect all fixtures should actually share the counters, but there may be use cases for not sharing the counters?

hkaiser: Yes, I saw that. I think the counter should be global (at least for now). I don't have any use-case for separating them.

hkaiser: That should be fine now. I have also made the counters atomic, as we possibly access the value from multiple threads concurrently.

hkaiser:
@msimberg this is good to go now, I believe.

PR #4562, opened 2020-04-27
Unregister lock earlier to avoid holding it while suspending

msimberg:
This one came up in the `shared_mutex1` test. There's more, but at least this fixes one failure. The comment above the fix should describe the change.

### DIFF
--- a/libs/synchronization/src/mutex.cpp
+++ b/libs/synchronization/src/mutex.cpp
@@ -45,6 +45,7 @@ namespace hpx { namespace lcos { namespace local {
         if (owner_id_ == self_id)
         {
             HPX_ITT_SYNC_CANCEL(this);
+            l.unlock();
             HPX_THROWS_IF(ec, deadlock, description,
                 "The calling thread already owns the mutex");
             return;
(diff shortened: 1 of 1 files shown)


libs/synchronization/src/mutex.cpp
```diff
@@ -90,18 +90,18 @@ namespace hpx { namespace lcos { namespace local {
...
         HPX_ITT_SYNC_RELEASING(this);
+        // Unregister lock early as the lock guard below may suspend.
+        util::unregister_lock(this);
         std::unique_lock<mutex_type> l(mtx_);
 
         threads::thread_id_type self_id = threads::get_self_id();
         if (HPX_UNLIKELY(owner_id_ != self_id))
         {
-            util::unregister_lock(this);
             HPX_THROWS_IF(ec, lock_error, "mutex::unlock",
```
hkaiser, line 100: While you're looking at this. Could you please add a `l.unlock();` before the exception is thrown?

biddisco: Why are we taking a lock during a call to mutex.unlock?

hkaiser: > Why are we taking a lock during a call to mutex.unlock?

To protect the internal members of the mutex itself.

biddisco: How many levels of mutex-ception are there? Do the mutexes that protect the mutexes also have mutexes protecting their internals? Could you remind me please of what internal members of the mutex need protecting?

hkaiser: @biddisco `hpx::lcos::local::mutex` is implemented using our `detail::condition_variable` which requires protection.

msimberg: Yep, good catch.

Issue #4537, opened 2020-04-17
Linking issue with libhpx_initd.a


teonnik:
I stumbled on an issue with HPX on master (c140bf0ebafe8c1caadc0c122fea8e583988ab2f). We have a small static library (`DLAF_gtest_hpx_main`) providing `main()` to our unit tests (GoogleTest) :

```
GTEST_API_ int test_main(int, char**) {
  std::printf("Running main() from gtest_hpx_main.cpp\n");
  auto ret = RUN_ALL_TESTS();
  hpx::finalize();
  return ret;
}

GTEST_API_ int main(int argc, char** argv) {
  testing::InitGoogleTest(&argc, argv);

  return hpx::init(test_main, argc, argv);
}
```

Linking the library (`DLAF_gtest_hpx_main`) to our tests results in a linker error
```
ld: <install_tree>/lib/libhpx_initd.a(hpx_wrap.cpp.o): in function `hpx_start::hpx_entry(int, char**)':
<build_tree>/init/src/hpx_wrap.cpp:73: undefined reference to `main'
ld: <install_tree>/lib/libhpx_initd.a(hpx_wrap.cpp.o): in function `__wrap_main':
<build_tree>/init/src/hpx_wrap.cpp:137: undefined reference to `main'
```

The issue is resolved if `DLAF_gtest_hpx_main` is linked after `libhpx_initd.a` (thanks Mikael!). 

Our link line for reference:
```
#!/bin/bash

/usr/bin/g++-8  -g \
  -Wl,-rpath -Wl,/home/teonnik/software/spack/opt/spack/linux-arch-broadwell/gcc-8.4.0/mpich-3.3.2-7mdalbfq7raethryd4niu6auxdhroyvw/lib \
  -Wl,-no-as-needed CMakeFiles/test_lapack_tile.dir/test_lapack_tile.cpp.o  -o test_lapack_tile   \
  -L/opt/cuda/lib64  \
  -L/home/teonnik/software/spack/opt/spack/linux-arch-broadwell/gcc-8.4.0/intel-mkl-2020.0.166-pinkbbeche3f7eupbcugs7zoqlgp2asn/mkl/lib/intel64_lin  \
[...]

PR #4564, opened 2020-04-27
Reshuffle internal targets and add HPX::hpx_no_wrap_main target

msimberg:
"Fixes"  #4537 and makes it a bit clearer which targets are internal and which are not.

Adds a layer of indirection to the `HPX::hpx` target so that we can have a version which does not link to `hpx_wrap`. All public targets are now in the `HPX::` namespace and all others are either in `HPXInternal::` or no namespace (the ones that are created in `HPXConfig.cmake`.

~The `hpx` library target is temporarily renamed `hpx_some_name` in search of a better name, because `hpx` is taken by the new interface target.~ The old `hpx` target is called `hpx_core` now. The actual library name is still `libhpx.so` (or the equivalent on Windows).

The `hpx_init` and `hpx_wrap` merge from #4462 has also been undone here.

I do have a nagging feeling that this is backwards and that we should make `hpx_wrap` opt-in, rather than opt-out, but that's something we shouldn't change now and can consider if and when the time comes for HPX 2.


src/CMakeLists.txt
```diff
@@ -103,15 +103,15 @@ endif()
 
 ################################################################################
 # libhpx
-add_library(hpx ${hpx_library_link_mode_core}
+add_library(hpx_some_name ${hpx_library_link_mode_core}
```
aurianer, line 106: `hpx_intermediate` ? `hpx_inter` ?

src/CMakeLists.txt
```diff
@@ -218,24 +218,37 @@ if(MSVC AND HPX_WITH_SWAP_CONTEXT_EMULATION)
...
-target_link_libraries(hpx_interface INTERFACE $<${_is_executable}:$<TARGET_NAME_IF_EXISTS:HPXInternal::hpx_init>>)
-target_compile_definitions(hpx_interface INTERFACE "$<${_is_executable}:HPX_APPLICATION_NAME_DEFAULT=$<TARGET_PROPERTY:NAME>>")
-target_compile_definitions(hpx_interface INTERFACE "$<${_is_executable}:HPX_PREFIX_DEFAULT=\"${HPX_PREFIX}\">")
-target_compile_definitions(hpx_interface INTERFACE "$<${_is_executable}:HPX_APPLICATION_EXPORTS>")
-target_compile_definitions(hpx_interface INTERFACE "$<${_is_library}:HPX_LIBRARY_EXPORTS>")
+target_link_libraries(hpx_interface INTERFACE hpx_interface_no_wrap_main)
+target_link_libraries(hpx_interface INTERFACE $<${_is_executable}:$<TARGET_NAME_IF_EXISTS:hpx_wrap>>)
+target_link_libraries(hpx_interface INTERFACE $<${_is_executable}:$<TARGET_NAME_IF_EXISTS:HPXInternal::hpx_wrap>>)
+
+target_link_libraries(hpx_no_wrap_main INTERFACE hpx_interface_no_wrap_main)
```
aurianer, line 251: Add `hpx_no_wrap_main` to `hpx_targets` otherwise it doesn't get exported

src/CMakeLists.txt
msimberg: I may just call it `hpx_core` for now. The name will be hidden from users so we can change it easily in the future.

src/CMakeLists.txt
msimberg: Good catch!

hkaiser:
@msimberg Please note that in order to solve #4580 I have added special linker flags to executables by associating it with `hpx_init` (https://github.com/STEllAR-GROUP/hpx/blob/master/init/CMakeLists.txt#L103-L110). So far I don't see this in this PR. Should it go somewhere else now?

msimberg:
> @msimberg Please note that in order to solve #4580 I have added special linker flags to executables by associating it with `hpx_init` (https://github.com/STEllAR-GROUP/hpx/blob/master/init/CMakeLists.txt#L103-L110). So far I don't see this in this PR. Should it go somewhere else now?

I've left that unchanged: https://github.com/STEllAR-GROUP/hpx/pull/4564/files#diff-534e69f2776a2ab1d04f5bc130b2a4caL103. I think it can stay there unless I come up with a good reason to move it.

PR #4574, opened 2020-04-27
Changing return type of hardware_concurrency() to unsigned int

### LINKED ISSUE #4569: Annoying warning when compiling - pls suppress or fix it.
## Expected Behavior

no warning

## Actual Behavior

getting the following warning over and over again:
....  include\hpx\threading\jthread.hpp(258): warning C4267: 'return': conversion from 'size_t' to 'unsigned int', possible loss of data 

compile with MSVC, enabling jemalloc

## Specifications

  - HPX Version: stable : c5b1812e0d414c33a9068d101075e2aab9afeb6d
  - Windows 10, MSVC latest 2019

hkaiser:
Fixes #4569

- flyby: making the return value `[[nodiscard]]`

### DIFF
--- a/libs/topology/src/topology.cpp
+++ b/libs/topology/src/topology.cpp
@@ -1521,9 +1521,9 @@ namespace hpx { namespace threads {
         std::size_t num_of_cores_;
     };
 
-    std::size_t hardware_concurrency()
+    unsigned int hardware_concurrency()
     {
         util::static_<hw_concurrency, hardware_concurrency_tag> hwc;
-        return hwc.get().num_of_cores_;
+        return static_cast<std::size_t>(hwc.get().num_of_cores_);
     }
 }}    // namespace hpx::threads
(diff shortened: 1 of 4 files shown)


McKillroy:
Works for me - the warnings are gone.

libs/topology/src/topology.cpp
```diff
@@ -1521,9 +1521,9 @@ namespace hpx { namespace threads {
         std::size_t num_of_cores_;
     };
 
-    std::size_t hardware_concurrency()
+    unsigned int hardware_concurrency()
     {
         util::static_<hw_concurrency, hardware_concurrency_tag> hwc;
-        return hwc.get().num_of_cores_;
+        return static_cast<std::size_t>(hwc.get().num_of_cores_);
```
aurianer, line 1527: Is this supposed to be a cast into `std::size_t` ?

hkaiser: Most likely not (even if it does no harm)! Thanks for noticing!

hkaiser: See #4583 for a fix.

PR #4559, opened 2020-04-26
Fixing compilation if HPX_WITH_THREAD_BACKTRACE_ON_SUSPENSION=On

hkaiser:
Fixes #4555

PR #4575, opened 2020-04-27
Excluding unconditional definition of class backtrace from global header

hkaiser:
Fixes #4559

This prevents `hpx::util::backtrace` from being defined twice if `HPX_WITH_STACKTRACES=Off`.


biddisco:
OK. This fixes the problem. I will approve this and merge it right away.

Issue #4053, opened 2019-08-30
Unify module structure


hkaiser:
Some of the modules we created have associated traits that until now have been placed in different directories. Some modules have them in their base directory (i.e. `hpx/modulename/`), some have them in `hpx/traits`. This should be unified. The consensus on irc was that associated traits should end up in the directory `hpx/modulename/traits/` (http://irclog.cct.lsu.edu/ste~b~~b~ar/2019-08-30#1567164898-1567165343;). 

Here is a list of modules that should be adopted to this new policy:

- [x] algorithms (#4587)
- [x] datastructures (#4113)
- [x] iterator_support (#4179)
- [x] parallel_executors
- [x] compute_cuda

There are possibly more.

hkaiser:
I think we can close this now. Please reopen if necessary.

PR #4587, opened 2020-04-30
Move traits headers to hpx/modulename/traits directory

msimberg:
Fixes (mostly, this should not close) #4053. At least `compute_cuda` still needs fixing but that whole module might need `compute/cuda` moved to `compute_cuda` to match the module name. That will come separately.


msimberg:
Just fixing headers that I forgot to change to the new locations...

aurianer:
Thanks for this !

PR #4588, opened 2020-05-02
cmake-formatting all cmake files

hkaiser:
This PR formats all CMake files using cmake-format (https://github.com/cheshirekow/cmake_format). It also adds a first draft for a format configuration file.

This is just a result of some experimentation and we should discuss whether we want to do this at all. If we want to go ahead with this we should probably also add a build step on CircleCI verifying the formatting.


hkaiser:
There is one open question: `cmake-format` itself is licensed under GPL-V3. Would that be an obstacle to using it on our code base?

cmake/FindHPX_BLAS.cmake
```diff
@@ -4,9 +4,16 @@
...
+  LIBRARIES
+  blas
+  libblas
+  LIBRARY_PATHS
+  lib64
+  lib
+  HEADERS
+  cblas.h
+  HEADER_PATHS
+  include
```
msimberg, line 18: Questionable if this is better...

hkaiser: Yah, I might have missed adding `hpx_find_package` to the list of custom commands...

hkaiser: Actually, `hpx_find_package` does not exist anymore and those files are all obsolete. I'll remove them.

hkaiser:
@msimberg I have added more custom commands and a couple of cmake-format: off regions now. Things look much better. 

@aurianer, @msimberg In general I don't care too much about the actual style, if you have some preference, please feel free to apply changes. So far I have left almost all configurations settings at their default.

CMakeLists.txt
```diff
@@ -18,63 +18,71 @@ cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
 
 # Overrides must go before the project() statement, otherwise they are ignored.
 
-################################################################################
+# ##############################################################################
```
msimberg, line 21: Is that space intentional? Looks like the hashruler options affect this. Don't mind either way, just curious if it was on purpose.

hkaiser: Yes, cmake-format inserts a space after the first `'#'`, no idea why. I'd disable this, but then the tools stops normalizing the length of hash rulers. In the end, I'm completely impartial to this. Let me know how you like it most.

hkaiser:
@msimberg I could use some help with adding the CircleCI step that verifies the formatting. Would you have some time to lend an hand, please?

CMakeLists.txt
msimberg: If the space means the lengths are consistent let's go with that, I don't mind.

msimberg:
> @msimberg I could use some help with adding the CircleCI step that verifies the formatting. Would you have some time to lend an hand, please?

Absolutely! We also need cmake-format in the docker image. I'll take care of it. Do I get it correctly that there's `cmake-format` *and* `cmake-lint`? Did you try the latter or should we leave that out for now (I just saw a `lint` section in the config, hence the question)?

hkaiser:
> @msimberg I could use some help with adding the CircleCI step that verifies the formatting. Would you have some time to lend an hand, please?
> 
> Absolutely! We also need cmake-format in the docker image. I'll take care of it. Do I get it correctly that there's cmake-format and cmake-lint? Did you try the latter or should we leave that out for now (I just saw a lint section in the config, hence the question)?

I have not tried `cmake-lint`. The changes here are the result of applying `cmake-format -i cmake/*.cmake **/CMakeLists.txt`

msimberg:
@hkaiser I've opened https://github.com/STEllAR-GROUP/docker_build_env/pull/22 and pushed changes to the CircleCI config here. The new step will fail at the moment. I added a dummy commit which violates the formatting just to check that the CircleCI step actually works. That should obviously be dropped before merging...

msimberg:
It seems like the docs don't build anymore most likely because sphinx was updated... Let's see if other PRs fail on that step now that the new image is there. If it needs fixing let's not tie that into this PR though as it's not critical to have that running now. I'll look at that separately.

https://cdash.cscs.ch//viewBuildError.php?buildid=107053 is a bit baffling to me though. As far as I can tell there are no changes to the CUDA options or files. I'll see if I can reproduce this.

msimberg:
> https://cdash.cscs.ch//viewBuildError.php?buildid=107053 is a bit baffling to me though. As far as I can tell there are no changes to the CUDA options or files. I'll see if I can reproduce this.

It was just a little typo in `HPX_AddModule.cmake` which meant the source files for the `compute_cuda` module weren't built.

aurianer:
Thanks for this!

PR #4592, opened 2020-05-04
Futures module

### LINKED ISSUE #4576: Future module
The last local module outside of the runtime is a future (base) module.

msimberg:
Fixes #4576.


libs/futures/include/hpx/lcos/local/futures_factory.hpp
```diff
@@ -124,24 +118,39 @@ namespace hpx { namespace lcos { namespace local {
...
+                Spawner&& spawner, error_code& ec)
+            {
+                this->check_started();
+
+                typedef typename Base::future_base_type future_base_type;
+                future_base_type this_(this);
+
+                spawner(
+                    util::deferred_call(&base_type::run_impl, std::move(this_)));
+
```
msimberg, line 144: @hkaiser I've made `futures_factory` take a "spawner" instead of relying directly on any particular method of spawning tasks (`register_thread_nullary`, `post` etc.). This works nicely to make this part of futures independent of execution (the `post` customization point) and `register_thread` (in principle not a problem, but does not need to be a dependency). It also removes the special case of handling thread executors differently.

`packaged_continuation` has similar dependencies on `post` and `post_policy_dispatch` for `async_exec` and `async`. Do you see a problem with giving `packaged_continuation` the same treatment as `task_object`? `packaged_continuation` could then live with the rest of the futures files and the `make_continuation*` wrappers would then live in the execution module.

As an alternative the `post` etc. customization points and `post_policy_dispatch`/`async_launch_policy_dispatch` declarations could move to the `async_base` or `basic_execution` module. This would mean that all the different ways of spawning tasks are available for the futures module. In that case I can also revert this particular change to `futures_factory`.

Do you have a preference or other suggestion?

hkaiser: @msimberg I like the change to `futures_factory`, that makes it much more reusable. I have no objections to adapt `packaged_continuation` either. Also I like the move of the launching facilities to `async_base`. Thanks!

msimberg: I should test my changes properly before saying they work nicely. The change to make `futures_factory::apply` take a templated spawner does not work because it would have to go through a templated member function in `task_base` (which is not allowed). Because of performance concerns I would not like to use `hpx::function` (or `unique_function_nonser`), but this concern may be unfounded?

I've reverted that change now as we can go ahead without that change. I've made `packaged_continuation` take a generic spawner though, and I've moved the execution customization points to `basic_execution`.

msimberg:
Note that I haven't yet changed any file paths or formatted any files. Doing that now.

libs/futures/include/hpx/lcos/local/futures_factory.hpp
hkaiser: > The change to make futures_factory::apply take a templated spawner does not work because it would have to go through a templated member function in task_base (which is not allowed).

Why wouldn't that be allowed?

msimberg: > > The change to make futures_factory::apply take a templated spawner does not work because it would have to go through a templated member function in task_base (which is not allowed).
> 
> Why wouldn't that be allowed?

I left out a word: *virtual* template member function. There may be another way around it though...

msimberg: Since this doesn't cause circular dependencies and removing thread executors will get rid of that dependency anyway I'm going to leave things as they are in `futures_factory`.

msimberg:
This is now nearing a point where it looks usable. Summary:

- As usual it's mostly headers that have changed locations.
- Git considers some of the traits files to have been deleted and added, instead of moved (because the formatting changes them enough to make them look like new files). They are unchanged though besides the formatting.
- `packaged_continuation` gained a spawner argument for its `attach`, `async_impl` etc. functions. The spawner takes care of launching. The helper functions `make_continuation*` which supply the spawner moved to the execution module.
- `future_data_base` gained a handler that can be set for reporting exceptions in `run_on_completed`. This removes the dependency on the runtime (`custom_exception_info`). It calls `std::terminate` if the handler isn't set.

msimberg:
I should note that `serialize_exception` is still a dependency from the main library. I'm not sure what to do with it yet and will deal with that in a separate PR.

hkaiser:
> future_data_base gained a handler that can be set for reporting exceptions in run_on_completed. This removes the dependency on the runtime (custom_exception_info). It calls std::terminate if the handler isn't set.

@msimber: may I suggest to just throw a `std::runtime_exception` instead of calling `std::terminate` if the handler is not set? I could envision situations in the future where somebody wanted to use `hpx::future` outside of the runtime and would be very surprised needing to set that handler explicitly.

hkaiser:
> The execution customization points moved to basic_execution as they're lower level than async, apply and friends. (Side note: maybe basic_execution should be execution_base...)

I'd agree to the renaming. We should probably do a global sweep through the names at some point...

msimberg:
> > future_data_base gained a handler that can be set for reporting exceptions in run_on_completed. This removes the dependency on the runtime (custom_exception_info). It calls std::terminate if the handler isn't set.
> 
> @msimber: may I suggest to just throw a `std::runtime_exception` instead of calling `std::terminate` if the handler is not set? I could envision situations in the future where somebody wanted to use `hpx::future` outside of the runtime and would be very surprised needing to set that handler explicitly.

Good idea. Perhaps even just rethrow the exception that was just caught?

msimberg:
I've just fixed the inspect errors on this one. This should be good to go now.

msimberg:
@hkaiser if you'd like to prioritize reviewing, this one should go before all the newer half-baked PRs that I've opened.

libs/async_base/src/launch_policy.cpp
```diff
@@ -32,26 +31,24 @@ namespace hpx
...
     ///////////////////////////////////////////////////////////////////////////
-    namespace detail
-    {
+    namespace detail {
         void policy_holder_base::load(
             serialization::input_archive& ar, unsigned)
         {
             int value = 0;
-            ar & value;
+            ar& value;
```
hkaiser, line 39: `// clang-format off`?

libs/execution/include/hpx/execution/traits/executor_traits.hpp
```diff
@@ -277,3 +283,41 @@ namespace hpx { namespace traits {
     using executor_parameters_type_t =
         typename executor_parameters_type<Executor>::type;
 }}    // namespace hpx::traits
+
+#if defined(HPX_HAVE_THREAD_EXECUTORS_COMPATIBILITY)
+namespace hpx { namespace threads {
+    class executor;
+}}    // namespace hpx::threads
```
hkaiser, line 290: This is already declared further up (L19-21)

libs/futures/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+
+set(futures_headers
+    hpx/futures/future.hpp
```
hkaiser, line 12: `hpx/future.hpp` ?

libs/async_base/src/launch_policy.cpp
msimberg: Yes, thanks!

libs/futures/CMakeLists.txt
msimberg: Unforunately no. `async` and friends should be included through `hpx/future.hpp` as well. It has to go in another module.

Issue #4074, opened 2019-09-05
HPX module "configurations"


msimberg:
This issue is for discussing if and how to expose modules to users:

- size of modules; small *and* large?
- static/dynamic?
- build configurations?

msimberg:
There are a few high-level "natural" levels of features that build on top of each other:
- utilities (testing, assertions, configuration, etc.)
- thread pools only
- synchronization primitives/futures (manual thread pool management, explicitly pass executors to apply/async)
- parallel algorithms
- local runtime (configuration management, default pool for apply/async)
- distributed runtime (including support for components/actions)
- distributed synchronization primitives, collectives
- distributed parallel algorithms
- optional components (already linked separately)

In addition to this users may want to pick smaller modules individually (hpx_concurrency). The question is: can we support both use cases? Can we support both use cases simultaneously (can I link to `libhpx_threading` ("high-level" module) and `libhpx_serialization` ("low-level" module)? Is the distinction necessary?

How do we handle build configurations? Should it be possible to individually disable modules? Is it up to the user to ensure that a module's dependencies are all enabled?

Given how complex things could get I'm leaning towards only exposing these "high-level" configurations/libraries to the user.

That's more questions than answers and should get us started...

PR #4593, opened 2020-05-04
HEP 1: build configurations

msimberg:
Meta: This is a PR and not an issue because I thought it would be easier to discuss, review, and revise what we want using the reviewing features. If we like the format of a "HEP" we can also take that into use, in which case merging this would indicate that we agree on what to do. Deciding on that is independent of the "configuration" topic here though.

This PR proposes a set of targets for us to expose, attempting to come to a conclusion for #4074.

### DIFF
--- a/docs/sphinx/heps/hep1.rst
+++ b/docs/sphinx/heps/hep1.rst
@@ -1,0 +1,55 @@
+HEP 1: |hpx| build configurations and exported libraries
+========================================================
+
+Motivation
+----------
+
+The motivation for this change is to expose smaller parts of |hpx| for users to
+consume. This is meant to address both the case where a user builds |hpx|
+manually and and wishes to build a subset of |hpx|, and the case where a user
+uses a preinstalled |hpx| (distribution packages, cluster modules) which may
+have most features enabled but still only wishes to pull in a subset of |hpx|.
+For the latter the user may choose to include headers and link to only a subset
+of |hpx| or they may choose to start only the local runtime even though |hpx|
+was built with the distributed runtime.
+
+Implementation
+--------------
+
+|hpx| will provide the following targets and corresponding libraries:
+
+- ``HPX::hpx_core``: Thread pools, schedulers, coroutines, and utilities;
+  contains functionality that is essential for building |hpx| but not the main
+  user-facing functionality itself (e.g. ``hpx::thread``, ``hpx::future``).
+- ``HPX::hpx_local``: Depends on ``HPX::hpx_core``; contains functionality
+  corresponding to the standard libraries for concurrency and parallelism, and
+  local extensions (e.g. ``hpx::thread``, ``hpx::async``, parallel algorithms,
+  synchronization primitives meant for use on ``hpx::thread``\ s).
+- ``HPX::hpx``: Depends on everything above; contains the local and distributed
+  runtimes, distributed extensions to the local library for concurrency and
+  parallelism, and utilities and functionality that depend on the runtime.
+
+Including headers from ``HPX::hpx_local`` will not include headers from
+``HPX::hpx``. The granularity at which features can be disabled at compile time
+in |hpx| are the targets in this HEP because it is a manageably small set of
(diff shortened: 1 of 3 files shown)


msimberg:
This might be more productive with a concrete list of which modules go where. I'll put together a dependency graph and list of what we have so far once the futures and local runtime modules are in.

msimberg:
I've just updated the text. Feedback would be appreciated from both people who want to use the local runtime and traditional distributed HPX users. I don't want to make the experience better for one type of users at the expense of the other type.

As for modules, here's what we currently have (including PRs) and how I'd split them up. I've simplified the high level configurations to only three main libraries.

 `HPX::core`: thread pools, schedulers, coroutines and utilities

- `affinity`
- `allocator_support`
- `asio`
- `assertion`
- `basic_execution`
- `cache`
- `concepts`
- `concurrency`
- `config`
- `coroutines`
- `datastructures`
- `debugging`
- `errors`
- `filesystem`
- `format`
- `functional`
- `hardware`
- `hashing`
- `io_service`
- `iterator_support`
- `logging`
- `memory`
- `naming_base`
- `plugin`
- `prefix`
- `preprocessor`
- `schedulers`
- `serialization`
- `static_reinit`
- `statistics`
- `string_util`
- `synchronization`
- `testing`
- `thread_pools`
- `thread_support`
- `threading_base`
- `timing`
- `topology`
- `type_support`
- `util`
- `version`

`HPX:local`: local standard parallelism and concurrency

- `algorithms`
- `async_base`
- `async_combinators`
- `execution`
- `executors`
- `futures`
- `local_async`
- `local_lcos`
- `pack_traversal`
- `threading`
- `timed_execution`

`HPX::hpx`: all the rest
- `async`
- `batch_environments`
- `checkpoint`
- `collectives`
- `command_line_handling`
- `distributed_executors`
- `include`
- `lcos_distributed`
- `mpi`
- `mpi_base`
- `performance_counters`
- `program_options`
- `resiliency`
- `resource_partitioner`
- `runtime_configuration`
- `runtime_local`
- `segmented_algorithms`
- `threadmanager`
- everything that is not yet modularized

Undecided: would fit in `HPX::local` but depend on runtime. i would put the in ``HPX::hpx`` right now.
- `compute`
- `compute_cuda`
- `thread_executors`

For users worried that ``HPX::core`` still contains too much we can introduce additional CMake options, orthogonal to these configurations, to disable e.g. serialization or statistics.

See the updated file for details on how I'd handle the local and distributed runtime entry points. It would allow for both a compile time and run time choice between the two runtimes.

docs/sphinx/heps/hep1.rst
```diff
@@ -0,0 +1,54 @@
...
+unlike the modules we have now. That means the smallest part of HPX one can
+build is ``HPX::core``. If a user needs an even smaller part of HPX they can
+link to the static module libraries. They will have to build at least
+``HPX::core`` though.
+
+``HPX::hpx`` includes:
+
+- local runtime
+- distributed runtime
+- distributed LCOs
```
msimberg, line 39: For distributed HPX users: I'd be happy to split this further if there's a need for it, so please comment if you think that's the case. I'd start with this simply because this is simpler and I would guess few distributed HPX users would need only the LCOs but not the runtime (not even possible at the moment, but could be done in a similar way to how we get the default thread pool). I'm not a distributed HPX user so I rely on your feedback here.

hkaiser: I don't think it makes much sense to split this even further for now.

msimberg:
I opened #4625 to add the main page for HEPs. This is now on top of that PR.

msimberg:
I'll close this as well. #4649 will implement part of the changes suggested here. What we should still consider is if even smaller configurations than `HPX_WITH_DISTRIBUTED_RUNTIME=OFF` should be available. My feeling right now is that it won't be necessary, given that it will be possible to at least link to smaller parts of `libhpx` (also to avoid adding even more configuration options).

PR #4595, opened 2020-05-05
Making sure main.cpp is built as a library if HPX_WITH_DYNAMIC_MAIN=OFF

hkaiser:
Fixes linking of HPX applications that rely on hpx_main.hpp if `HPX_WITH_DYNAMIC_MAIN=OFF`. This is the main mode on Windows but will work on other platforms as well.


hkaiser:
> Grr, it was right there in the CI logs... Sorry and thank you!

No worries!

PR #4596, opened 2020-05-06
Fix performance regression in block_executor

msimberg:
*How to make your PRs look good: Introduce a performance regression without benchmarking (#4301), followed by another PR which improves performance a bit (#4536).*

The improvements in #4536 were real, but ignore that #4301 introduced a regression. The regression was caused by the `block_executor` previously relying on thread executors which eagerly created threads for bulk execution (`run_now = true`), whereas the new underlying executor `restricted_thread_pool_executor` only creates task descriptions. In addition the `restricted_thread_pool_executor` did hierarchical spawning. Both of these seem to have caused the performance regression (fixing just one of them is not enough). This is based on the stream benchmark and @m-diers observations on his application.

In this PR I've extended the `block_executor` constructor to take priority, stacksize, and hint to be passed on to the underlying executor. The default priority is high, which forces immediate thread creation, and thus restores the old behaviour (mostly, the old threads were normal priority). I've also removed the hierarchical spawning from the `restricted_thread_pool_executor`.

The question is if this (controlling thread creation with the priority) nicely models what we want. I think it is a reasonable design as I wouldn't want to expose something like the `run_now` parameter to users. I also wouldn't want to force bulk execution to always create threads. The notion of a tight numerical parallel for loop like those in the stream benchmark being high priority feels more or less right (we want all of the for loop tasks to be executed asap without being interrupted by otherwise spawned regular tasks).

This is a draft as I want to redo the benchmarks in #4301 and #4536.

### DIFF
--- a/libs/compute/include/hpx/compute/host/block_executor.hpp
+++ b/libs/compute/include/hpx/compute/host/block_executor.hpp
@@ -43,9 +43,16 @@ namespace hpx { namespace compute { namespace host {
         typedef hpx::parallel::execution::static_chunk_size
             executor_parameters_type;
 
-        block_executor(std::vector<host::target> const& targets)
+        block_executor(std::vector<host::target> const& targets,
+            threads::thread_priority priority = threads::thread_priority_high,
+            threads::thread_stacksize stacksize =
+                threads::thread_stacksize_default,
+            threads::thread_schedule_hint schedulehint = {})
           : targets_(targets)
           , current_(0)
+          , priority_(priority)
+          , stacksize_(stacksize)
+          , schedulehint_(schedulehint)
         {
             init_executors();
         }
@@ -238,12 +245,17 @@ namespace hpx { namespace compute { namespace host {
             for (auto const& tgt : targets_)
             {
                 auto num_pus = tgt.num_pus();
-                executors_.emplace_back(num_pus.first, num_pus.second);
+                executors_.emplace_back(num_pus.first, num_pus.second,
+                    threads::thread_priority_high);
             }
         }
         std::vector<host::target> targets_;
         std::atomic<std::size_t> current_;
         std::vector<Executor> executors_;
+        threads::thread_priority priority_ = threads::thread_priority_high;
+        threads::thread_stacksize stacksize_ =
+            threads::thread_stacksize_default;
+        threads::thread_schedule_hint schedulehint_ = {};
     };
 }}}    // namespace hpx::compute::host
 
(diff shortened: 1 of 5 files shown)


msimberg:
Finally some benchmarks (and with this I think this is ready to be merged modulo linting; I've rebased on the latest master):

# Stream

For stream we were doing well with block executor in 1.4.1 until I made it worse on master. This branch basically restores the old performance for stream.

## Large arrays

### 1.4.1 `block_executor`
```
Memory per array = 781.25 MiB (= 0.762939 GiB).
Function    Best Rate MB/s  Avg time     Min time     Max time
Copy:           76450.7     0.021574     0.021431     0.023405
Scale:          75844.3     0.021736     0.021602     0.022370
Add:            86259.7     0.028615     0.028491     0.029447
Triad:          87665.2     0.028129     0.028034     0.029529
```

### This branch `block_executor`
```
Function    Best Rate MB/s  Avg time     Min time     Max time
Copy:           76818.0     0.021468     0.021328     0.022025
Scale:          75794.6     0.021769     0.021616     0.022544
Add:            86359.0     0.028640     0.028458     0.029616
Triad:          87852.2     0.028096     0.027974     0.028775
```

All of these numbers are mostly within noise thresholds. No progress, but no regression either.

## Small arrays

### 1.4.1 `block_executor`
```
Memory per array = 0.078125 MiB (= 7.62939e-05 GiB).
Function    Best Rate MB/s  Avg time     Min time     Max time
Copy:             439.3     0.000460     0.000373     0.002166
Scale:            431.8     0.000462     0.000379     0.001871
Add:              665.4     0.000466     0.000369     0.002544
Triad:            670.9     0.000447     0.000366     0.002006
```

### This branch `block_executor`
```
Memory per array = 0.078125 MiB (= 7.62939e-05 GiB).
Function    Best Rate MB/s  Avg time     Min time     Max time
Copy:             698.9     0.000331     0.000234     0.001073
Scale:            695.0     0.000334     0.000236     0.001046
Add:             1024.0     0.000333     0.000240     0.001153
Triad:           1011.8     0.000327     0.000243     0.001027
```

Here there's a small improvement (big in the worst case times; average/min improvement is not as dramatic, but still better), but in general HPX does badly with small arrays.

Sorry I don't have actual graphs for this. That would be something nice to have in the future to compare to other stream benchmarks.

# Future overhead

This is a bit more interesting. I added hierarchical spawning to the `thread_pool_executor` without thinking much about the threshold at which it actually does hierarchical spawning. I made it do hierarchical spawning if there are at least two chunks for a worker thread. I've updated the default threshold to be 6 based measurements on daint, but it's also customizable now as the best value is likely to change from system to system. With 6 it'll do hierarchical spawning for half of the tasks on average (the default static chunk size creates 4-8 chunks per worker thread). With this most of the funny behaviour in https://github.com/STEllAR-GROUP/hpx/pull/4301#issuecomment-575627988 is gone. 
thread_pool_executor_upto_100_tasks
thread_pool_executor_over_100_tasks

PR #4598, opened 2020-05-06
Making sure current_executor and service_executor functions are linked into the core library

hkaiser:


hkaiser:
> Thanks! Besides checking the CI logs properly, what can we do to avoid this? These are header only. I don't understand why this is required...

Hmm, not sure. If all is header only then the `HPX_EXPORT` is probably the culprit. I should have tried removing that first.

hkaiser:
@msimberg I have redone this by simply removing `HPX_EXPORT` from header only types. That will do the trick with much less changes. Please re-review.

PR #4603, opened 2020-05-07
Deal with distributed modules in their respective CMakeLists.txt

msimberg:
@hkaiser I've based this on the CMake formatting branch.


hkaiser:
@msimberg sorry, I force pushed the cmake-format branch (before seeing this), you'll likely have to redo this :/

PR #4605, opened 2020-05-08
Enable static symbols for stacktraces in CircleCI config

msimberg:
@hkaiser you mentioned that this lead to segfaults in some tests. Did you notice if that was reproducible? If this causes problems let's not merge it, but if things work well this is really helpful for seeing where failures happen.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -204,6 +204,7 @@ jobs:
                 -DCMAKE_CXX_CLANG_TIDY=clang-tidy \
                 -DHPX_WITH_THREAD_LOCAL_STORAGE=On \
                 -DHPX_WITH_STACKTRACES_STATIC_SYMBOLS=On \
+                -DHPX_WITH_STACKTRACES_DEMANGLE_SYMBOLS=Off \
                 -DHPX_WITH_TESTS_DEBUG_LOG=On \
                 -DHPX_WITH_TESTS_DEBUG_LOG_DESTINATION=/hpx/build/debug-log.txt \
                 -DHPX_WITH_SPINLOCK_DEADLOCK_DETECTION=On \
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -203,6 +203,7 @@ jobs:
                 -DHPX_WITH_DEPRECATION_WARNINGS=On \
                 -DCMAKE_CXX_CLANG_TIDY=clang-tidy \
                 -DHPX_WITH_THREAD_LOCAL_STORAGE=On \
+                -DHPX_WITH_STACKTRACES_STATIC_SYMBOLS=On \
                 -DHPX_WITH_TESTS_DEBUG_LOG=On \
                 -DHPX_WITH_TESTS_DEBUG_LOG_DESTINATION=/hpx/build/debug-log.txt \
                 -DHPX_WITH_SPINLOCK_DEADLOCK_DETECTION=On \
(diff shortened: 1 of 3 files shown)


hkaiser:
> @hkaiser you mentioned that this lead to segfaults in some tests. Did you notice if that was reproducible? If this causes problems let's not merge it, but if things work well this is really helpful for seeing where failures happen.

Not sure if this reproducible. I saw several tests failing with this option enabled - time will tell... In the end it could be a bad version of `libdl` playing games with us here.

msimberg:
@hkaiser I suppose this is roughly what you saw earlier: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/3401/workflows/52077540-bbe2-4983-a1a1-2db3c2a690da/jobs/188668/steps? I might have to give up on this one for now...

hkaiser:
> @hkaiser I suppose this is roughly what you saw earlier: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/3401/workflows/52077540-bbe2-4983-a1a1-2db3c2a690da/jobs/188668/steps? I might have to give up on this one for now...

Yes, that's exactly what I was seeing. Are you able to reproduce these effects locally for yourself?

msimberg:
> > @hkaiser I suppose this is roughly what you saw earlier: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/3401/workflows/52077540-bbe2-4983-a1a1-2db3c2a690da/jobs/188668/steps? I might have to give up on this one for now...
> 
> Yes, that's exactly what I was seeing. Are you able to reproduce these effects locally for yourself?

I don't know yet, but I'll find out soon enough.

msimberg:
> Yes, that's exactly what I was seeing. Are you able to reproduce these effects locally for yourself?

No, not with my local clang, but I'll see if I can reproduce it with the docker image.

msimberg:
Re-running the job on CircleCI does lead to the same segfaults reliably, but my local clang or the docker image locally do not lead to the segfaults... It segfaults inside `abi::__cxa_demangle`. Just trying to demangle a string that causes a segfault in a standalone program does not lead to a segfault either, so it's not the particular mangled name that is the problem (although it is quite long). I don't know if there's anything we can do at the moment. We could try this again next time we upgrade the docker image. I'll close this for now.

hkaiser:
@msimberg How about if we added an option that disabled demangling (for the CI)? That would circumvent the issue on the CI and still give us the information we need. One can always demangle things after the fact, after all.

msimberg:
> @msimberg How about if we added an option that disabled demangling (for the CI)? That would circumvent the issue on the CI and still give us the information we need. One can always demangle things after the fact, after all.

That's a very good idea! I'll get back to this and reopen it once I have that option in place.

hkaiser:
> > @msimberg How about if we added an option that disabled demangling (for the CI)? That would circumvent the issue on the CI and still give us the information we need. One can always demangle things after the fact, after all.
> 
> That's a very good idea! I'll get back to this and reopen it once I have that option in place.

This should be essentially just a wrapper around these lines: 
https://github.com/STEllAR-GROUP/hpx/blob/master/libs/debugging/src/backtrace.cpp#L207-L208

msimberg:
I'm not sure what the better default here is: demangled names with potentially missing symbols, or all symbols mangled... If this really is rare we could turn them both on by default and just turn them off for CI, but I think I'd rather have incomplete stacktraces over segfaults as the default.

msimberg:
This seems to work around the segfaults.

PR #4606, opened 2020-05-08
Local runtime module etc.

msimberg:
Adds `runtime_local` and `init_runtime` modules and cleans up some runtime-related dependencies.

The `runtime` class is still named `runtime` for compatibility reasons, but introducing an alias from `runtime` to `runtime_local` may be enough.

The `init_runtime` (`init` was taken) module just holds the `hpx_init/start/suspend(_impl)` files. The paths are unchanged.

`pre_main.cpp` was already merged into `runtime_distributed.cpp` but I forgot to delete it in the local runtime PR (#4091). I've deleted that. Some of the leftover timers depend on the runtime and moved to live in the `runtime_local` module. Some compatibility headers were using the wrong ifdefs which meant the warnings were not triggered from them. I've updated the ifdefs and the warnings that resulted from them.

The `compute` modules depend quite heavily on the resource partitioner (through the runtime) and so depend on `runtime_local` now. The `compute` modules could move out to separate shared libraries/components since we could leave the base `cuda_support` somewhere in the main library. Alternatively the `compute` modules would have to be rewritten to get their topology/affinity information directly from the thread pools or schedulers. The distributed parts of `compute` depend on e.g. `find_here`.

There's a whole big refactoring just waiting to be done on the resource partitioner now that it's not supposed to be initialized before `init/start` (there are actually some tests that still initialize it before `init/start` and those need to be dealt with before this PR makes sense; I'll open a separate PR for that). Essentially all of this: https://github.com/STEllAR-GROUP/hpx/blob/a04b8c1753679f83936791a694a428f1cbf29358/libs/resource_partitioner/src/detail_partitioner.cpp#L886-L907 could be lifted out into `hpx_init.cpp` and the resource partitioner can focus on just partitioning. That needs quite a lot of work though (even if straightforward) so it will definitely stay out of this PR.

I believe the changes I've made to pass component registries up to `hpx_init.cpp` would be much cleaner with the above refactoring, but something like this will have to do for now.


libs/executors/include/hpx/executors/guided_pool_executor.hpp
```diff
@@ -231,29 +224,30 @@ namespace hpx { namespace parallel { namespace execution {
         friend struct guided_pool_executor_shim;
 
     public:
-        guided_pool_executor(const std::string& pool_name, bool hp_sync = false)
-          : pool_(&resource::get_thread_pool(pool_name))
+        guided_pool_executor(
+            threads::thread_pool_base* pool, bool hp_sync = false)
+          : pool_(pool)
```
msimberg, line 229: @biddisco are you ok with this change? The constructor takes an actual thread pool instead of the name of the thread pool, which allows this not to depend on the runtime (threadmanager, partitioner) to ask for a pool with a specific name. The equivalent functionality after this change is `guided_pool_executor(&hpx::resource::get_thread_pool("pool-name"))`.

I would also deprecate `pool_executor` in favor of `thread_pool_executor(get_thread_pool("pool-name"))`. I will for the next release move `pool_executor` to the runtime module (as it's a wrapper enabled by having the runtime), and add it to the `hpx/include/*executors.hpp` headers.

Do you think the `guided_pool_executor` needs a compatibility option before I make this change (I know the linear algebra people use `pool_executor` which is why I'd like to have the compatibility option.)

msimberg: I will also move, but not deprecate, the service executors (`main_pool_executor` etc.) to the runtime module since they cannot exist without the runtime.

msimberg:
This is starting to work.

I had to add an uncomfortably large number of callbacks/handlers to satisfy the thread executors and execution module (they want to fall back to getting pu masks and number of os threads from the runtime if the executor doesn't provide it). I think this is something that we could eventually remove or provide another default.

As always this is mostly files moving around and formatting changes. I think the other changes are described above in the main PR description and the comment on pool and service executors.

libs/execution/src/execution_parameter_callbacks.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include <hpx/config.hpp>
+#include <hpx/execution/detail/execution_parameter_callbacks.hpp>
+
+#include <cstddef>
+#include <type_traits>
+#include <utility>
+
+///////////////////////////////////////////////////////////////////////////////
+namespace hpx { namespace parallel { namespace execution { namespace detail {
+    static get_os_thread_count_type get_os_thread_count_f;
```
hkaiser, line 18: May I suggest to hide these static variables as function local statics to avoid initialization sequencing problems?

libs/executors/src/exception_list_callbacks.cpp
```diff
@@ -0,0 +1,31 @@
...
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/config.hpp>
+#include <hpx/executors/exception_list.hpp>
+
+namespace hpx { namespace parallel { inline namespace v1 { namespace detail {
+    static exception_list_termination_handler_type
+        exception_list_termination_handler_f;
```
hkaiser, line 14: Same here, function local statics might be safer.

msimberg: Yeah, I just updated both files to use function statics.

libs/init_runtime/src/hpx_init.cpp
```diff
@@ -900,7 +900,7 @@ namespace hpx {
     }    // namespace detail
 
     ///////////////////////////////////////////////////////////////////////////
-    int finalize(double shutdown_timeout, double localwait, error_code& ec)
+    HPX_EXPORT int finalize(double shutdown_timeout, double localwait, error_code& ec)
```
hkaiser, line 903: Shouldn't those API functions be marked `HPX_EXPORT` at their declaration instead?

msimberg: Yes... maybe. I need to dig a bit more. We only include the declarations for applications. I think I'll put some of the declarations in a separate header file (some are currently duplicated, some are not included in hpx_init.cpp).

msimberg: I just pushed something that works on Linux. It remains to be seen if it's correct on Windows.

msimberg: I'm starting to suspect that the linker errors weren't at all due to missing or misplaced `HPX_EXPORTS`, but due to the force linking stuff not working. @hkaiser I attempted to add something to fix it based on the other force linking files, but I don't know if it's enough. If this comes back red would you mind telling me exactly what I should add to make it link? :) Does everything that's exported from a module need a force linking helper?

hkaiser: The native MSVC linker adds functions based on whether at least one other symbol from the same object file is needed (that's the default - however, it can do function level linking as well, but we don't enable this). That means that all symbols from an object file will be forced into the shared library if at least one of those symbols is needed. We use this to create an artificial reference to one of the symbols from files that would otherwise be left out (as the core library itself doesn't require linking to any of those symbols). That's what the `force_linking structures are all about` - creating that artificial reference.

hkaiser: Please note also, that this is orthogonal to the export issue. Symbols will be included through the force-linking regardless whether they are to exported or not.

I general, I think this is a bug in the linker. I would expect for it to link in every exported symbol from static libraries, even if they are not directly referenced by the linked module itself..

msimberg:
Since I'm 99% sure #4649 is going to remove the need for the force linking functionality I'm going to make I'm going to ignore the windows build problems on this branch and focus on getting #4649 ready instead.

libs/init_runtime/src/hpx_init.cpp
msimberg: None of the additional `HPX_EXPORT`s were needed in the end but I reshuffled things a bit (added another header file with declarations, so that `hpx_init.cpp` and `hpx_init_impl.hpp` etc. can share the declarations). Since object libraries went on hold indefinitely in #4649, I fixed the force linking for `init_runtime` and `runtime_local` here. Thanks @hkaiser for the explanation! It was the part about one symbol per translation unit that I missed. I thought it was enough to force one symbol per library to be linked.

msimberg:
This has not really changed since last time, except linking now works on Windows, so this is content-wise ready to go. The only things missing are potentially all the linting steps. Since rebasing on master (with #4615) both this and #4649 now suffer from not having `get_exception` instantiated in various (different) situations. I really don't understand why this is happening...

msimberg:
I've incorporated #4670 into this PR and this should be ready to go now (CI came back clean last time).

libs/execution/include/hpx/execution/detail/execution_parameter_callbacks.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include <hpx/config.hpp>
+#include <hpx/topology/topology.hpp>
+
+#include <cstddef>
+
+///////////////////////////////////////////////////////////////////////////////
+namespace hpx { namespace parallel { namespace execution { namespace detail {
+    /// \cond NOINTERNAL
+    using get_os_thread_count_type = std::function<std::size_t()>;
```
hkaiser, line 17: Have you switched to used `std::function` now? Should we generally move to using it instead of our own types?

msimberg: Not really. Out of habit I've been using `std::function` whenever I've added callbacks. I'm happy to use `hpx::function_*` as well, but I'd like to do that in a separate PR so that I can clean up other instances of `std::function` as well.

msimberg: As a reminder: https://github.com/STEllAR-GROUP/hpx/issues/4672.

msimberg:
Just rebased because of #4670...

libs/executors/include/hpx/executors/guided_pool_executor.hpp
biddisco: I'm fine with changing the type from a name to a pool, this is not a critical path and the changes are easy. I don't thinl a compatibility layer is worth the effort as this code is barely used outside of our group.

msimberg:
Are there remaining comments on this one? I'm going to ignore the suggestions by the spell checker.

hkaiser:
@msimberg let's get this merged before touching #4693 again.

PR #4607, opened 2020-05-08
Moving winsocket initialization to core library

hkaiser:
Fixes linker errors if `hpx::init` or `hpx::start` are used from a shared library.


biddisco:
Shouldn't winsocket only be needed in distributed applications (and hence not in core?)

hkaiser:
> Shouldn't winsocket only be needed in distributed applications (and hence not in core?)

Yes, that is something we can sort out eventually. For now this is mostly a stop-gap measure to solve a Phylanx problem.

PR #4609, opened 2020-05-10
Adding an API function that exposes the number of idle cores

### LINKED ISSUE #4599: Accessing the number of idle cores at runtime
## Expected Behavior

Could we have a function to determine the number of idle cores at runtime? I would like to make decisions within a custom parallel executor based on the number of idle cores.

hkaiser:
Fixes #4599

The new API function is `hpx::get_idle_core_count()`, it will return the number of cores that have no threads (staged or pending) waiting in their queues.

### DIFF
--- a/libs/schedulers/include/hpx/schedulers/local_priority_queue_scheduler.hpp
+++ b/libs/schedulers/include/hpx/schedulers/local_priority_queue_scheduler.hpp
@@ -882,6 +882,23 @@ namespace hpx { namespace threads { namespace policies {
             return count;
         }
 
+        // Queries whether a given core is idle
+        bool is_core_idle(std::size_t num_thread) const override
+        {
+            if (num_thread < num_queues_ &&
+                queues_[num_thread].data_->get_thread_count(unknown) != 0)
+            {
+                return false;
+            }
+            if (num_thread < num_high_priority_queues_ &&
+                high_priority_queues_[num_thread].data_->get_thread_count(
+                    unknown) != 0)
+            {
+                return false;
+            }
+            return true;
+        }
+
         ///////////////////////////////////////////////////////////////////////
         // Enumerate matching threads from all queues
         bool enumerate_threads(
--- a/libs/thread_pools/include/hpx/thread_pools/scheduled_thread_pool_impl.hpp
+++ b/libs/thread_pools/include/hpx/thread_pools/scheduled_thread_pool_impl.hpp
@@ -1776,6 +1776,21 @@ namespace hpx { namespace threads { namespace detail {
             thread_count_.load();
     }
 
+    template <typename Scheduler>
+    std::int64_t scheduled_thread_pool<Scheduler>::get_idle_core_count() const
+    {
+        std::int64_t count = 0;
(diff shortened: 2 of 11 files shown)


libs/schedulers/include/hpx/schedulers/local_priority_queue_scheduler.hpp
```diff
@@ -882,6 +882,28 @@ namespace hpx { namespace threads { namespace policies {
...
+            }
+            count += low_priority_queue_.get_thread_count(unknown) == 0;
+
+            for (std::size_t i = 0; i != num_queues_; ++i)
+            {
+                count += queues_[i].data_->get_thread_count(unknown) == 0;
+            }
+
+            return count;
+        }
```
msimberg, line 905: Is this a good idea? It'll return `num_threads * 2 + 1` when all the cores are idle, and at least `2` if one of the cores are idle. Would it be reasonable to to combine the normal and high priority queues (add one if both are empty), and ignore the low priority queue completely?

hkaiser: Fair points!

msimberg: Now it returns 1 only if all the normal and high priority queus are empty, and 0 otherwise... ;)

Something like this?
```
            std::int64_t count = 0;

            for (std::size_t i = 0; i != (std::max)(num_queues_, num_high_priority_queues_); ++i)
            {
                bool idle = true;
                if (i < num_queues_)
                    idle &&= queues_[i].data_->get_thread_count(unknown) == 0;
                if (i < num_high_priority_queues_)
                    idle &&= high_priority_queues_[i].data_->get_thread_count(unknown) == 0;

                if (idle) ++count;
            }

            return count;
```

hkaiser:
> Thanks, looks good to me!

This code does not take into account whether a core is currently running a thread or not. This still needs fixing.

hkaiser:
@msimberg I went back and re-implemented the logic. Could you have another look, please?

libs/thread_pools/include/hpx/thread_pools/scheduled_thread_pool_impl.hpp
```diff
@@ -1776,6 +1776,21 @@ namespace hpx { namespace threads { namespace detail {
...
     }
 
+    template <typename Scheduler>
+    std::int64_t scheduled_thread_pool<Scheduler>::get_idle_core_count() const
+    {
+        std::int64_t count = 0;
+        std::size_t i = 0;
+        for (auto const& data : counter_data_)
+        {
+            if (!data.tasks_active_ && sched_->Scheduler::is_core_idle(i++))
```
hkaiser, line 1786: Uhh, I see now that this is not correct. This code will increment `i` only if the current core is not active.

hkaiser: See #4643 for a fix.

PR #4610, opened 2020-05-10
Making sure the thread-local lock registration data is moving to the core the suspended HPX thread is resumed on

hkaiser:
Currently, if a lock is actively ignored while the executing thread is suspended, then the lock is not properly de-registered if the thread is resumed on a different core. This PR fixes the issue by moving the lock registration data to a different core with the executing thread.

### DIFF
--- a/libs/basic_execution/include/hpx/basic_execution/register_locks.hpp
+++ b/libs/basic_execution/include/hpx/basic_execution/register_locks.hpp
@@ -12,6 +12,11 @@
 
 #include <cstddef>
 #include <functional>
+#include <map>
+#include <memory>
+#ifdef HPX_HAVE_VERIFY_LOCKS_BACKTRACE
+#include <string>
+#endif
 #include <type_traits>
 #include <utility>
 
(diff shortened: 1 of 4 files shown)


msimberg:
I was about to write a bunch of questions since I didn't see how this could work. Now I think I see how it works, but I'll ask the questions anyway to make sure it's what's actually happening. By putting `on_exit_reset_held_locks_data` around thread yielding you're effectively making the ignored and held locks HPX thread-local rather than worker thread-local? And therefore you don't get into trouble with multiple HPX threads ignoring different sets of locks on the same worker thread, or other locks being ignored on the worker thread to which an HPX thread with ignored locks gets moved?

hkaiser:
> I was about to write a bunch of questions since I didn't see how this could work. Now I think I see how it works, but I'll ask the questions anyway to make sure it's what's actually happening. By putting on_exit_reset_held_locks_data around thread yielding you're effectively making the ignored and held locks HPX thread-local rather than worker thread-local? And therefore you don't get into trouble with multiple HPX threads ignoring different sets of locks on the same worker thread, or other locks being ignored on the worker thread to which an HPX thread with ignored locks gets moved?

That is the idea, yes.

PR #4611, opened 2020-05-11
Making auto_chunk_size execute the given function using the given executor

hkaiser:
- flyby: adding sync_execute to parallel_executor and parallel_aggregated_executor

@McKillroy this should fix your issue with the wrapping executor not being invoked when the auto_chunker is used. This however requires that you implement the `sync_execute` method on your executor.

### DIFF
--- a/libs/executors/include/hpx/executors/parallel_executor.hpp
+++ b/libs/executors/include/hpx/executors/parallel_executor.hpp
@@ -15,6 +15,7 @@
 #include <hpx/execution/algorithms/detail/predicates.hpp>
 #include <hpx/execution/detail/async_launch_policy_dispatch.hpp>
 #include <hpx/execution/detail/post_policy_dispatch.hpp>
+#include <hpx/execution/detail/sync_launch_policy_dispatch.hpp>
 #include <hpx/execution/executors/fused_bulk_execute.hpp>
 #include <hpx/execution/executors/static_chunk_size.hpp>
 #include <hpx/execution/traits/is_executor.hpp>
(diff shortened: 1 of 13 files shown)


libs/executors/include/hpx/executors/parallel_executor.hpp
```diff
@@ -167,6 +168,17 @@ namespace hpx { namespace parallel { namespace execution {
...
+        template <typename F, typename... Ts>
+        static
+            typename hpx::util::detail::invoke_deferred_result<F, Ts...>::type
+            sync_execute(F&& f, Ts&&... ts)
+        {
+            return hpx::detail::sync_launch_policy_dispatch<
+                launch::sync_policy>::call(launch::sync, std::forward<F>(f),
+                std::forward<Ts>(ts)...);
+        }
+
```
msimberg, line 181: Should we fall back on `async_execute(...).get()` for executors that don't have `sync_execute` (as far as I could tell we don't do that now, but I might have mislooked)?

I like this change since explicitly doing `sync_launch_policy_dispatch` avoids all the overhead of launching tasks. I don't mind if we do actually require having `sync_execute` instead of using `async_execute` as a fallback since it avoids nasty performance surprises if one actually expects inline execution with sync.

On the other hand `bulk_sync_execute` does fall back to `bulk_async_execute` and `wait_all`, but that's slightly different since at least parallel execution requires spawning some tasks in any case.

hkaiser: > Should we fall back on async_execute(...).get() for executors that don't have sync_execute (as far as I could tell we don't do that now, but I might have mislooked)?

That happens in the customization points. I added this here to ensure that `sync_execute` is really synchronous (and inline) on this executor (which it was not before).

hkaiser: > On the other hand bulk_sync_execute does fall back to bulk_async_execute and wait_all, but that's slightly different since at least parallel execution requires spawning some tasks in any case.

I'm not sure if we should implement `bulk_sync_execute` here as (now that we have `sync_execute`). I wouldn't hurt, I suppose.

msimberg: I don't mind either way, and resolving this does not block this PR at all.

McKillroy:
It seems to work for me now - I could compile the PR, 
and switching back to the autochunker worked again as expected 
with the example thread hook enabled executor.
The lambdas were executed correctly before the chunks and sync_execute was called.
Nice work and thanks a ton ! :D

msimberg:
Is this ready to go?

hkaiser:
> Is this ready to go?

It is ready to go. Thanks!

PR #4615, opened 2020-05-11
Add handler for serializing custom exceptions

msimberg:
This takes care of the circular dependency between the runtime and futures/exceptions, but there *must* be a better way of handling this.

### DIFF
--- a/libs/serialization/src/exception_ptr.cpp
+++ b/libs/serialization/src/exception_ptr.cpp
@@ -1,0 +1,311 @@
+//  Copyright (c)      2020 ETH Zurich
+//  Copyright (c) 2007-2020 Hartmut Kaiser
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/assertion.hpp>
+#include <hpx/errors.hpp>
+#include <hpx/serialization/exception_ptr.hpp>
+#include <hpx/serialization/serialize.hpp>
+
+#if BOOST_ASIO_HAS_BOOST_THROW_EXCEPTION != 0
+#include <boost/exception/diagnostic_information.hpp>
+#include <boost/exception/exception.hpp>
+#endif
+
+#include <cstddef>
+#include <cstdint>
+#include <exception>
+#include <stdexcept>
+#include <string>
+#include <typeinfo>
+
+///////////////////////////////////////////////////////////////////////////////
+namespace hpx { namespace serialization {
+    namespace detail {
+        static save_custom_exception_handler_type save_custom_exception_handler;
+
+        HPX_EXPORT void set_save_custom_exception_handler(
+            save_custom_exception_handler_type f)
+        {
+            save_custom_exception_handler = f;
+        }
+
+        static load_custom_exception_handler_type load_custom_exception_handler;
+
+        HPX_EXPORT void set_load_custom_exception_handler(
+            load_custom_exception_handler_type f)
+        {
+            load_custom_exception_handler = f;
+        }
+    }    // namespace detail
+
+    ///////////////////////////////////////////////////////////////////////////
+    // TODO: This is not scalable, and painful to update.
+    template <typename Archive>
+    void save(Archive& ar, std::exception_ptr const& ep, unsigned int version)
+    {
+        if (detail::save_custom_exception_handler)
+        {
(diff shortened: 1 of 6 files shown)


hkaiser:
We've had a request a while back asking for enabling user defined exceptions to be integrated with our exception handling. Would this be something we should take into account for this PR?

msimberg:
> We've had a request a while back asking for enabling user defined exceptions to be integrated with our exception handling. Would this be something we should take into account for this PR?

Would you have an issue or some other reference on what exactly they wanted? In any case I'd rather not mix it in this PR, but could perhaps look at it separately.

msimberg:
Are there comments on this one?

libs/errors/src/exception.cpp
```diff
@@ -276,6 +276,47 @@ namespace hpx { namespace detail {
...
+        std::string const&);
+    template HPX_EXPORT std::exception_ptr get_exception(
+        std::runtime_error const&, std::string const&, std::string const&, long,
+        std::string const&);
+    template HPX_EXPORT std::exception_ptr get_exception(
+        std::out_of_range const&, std::string const&, std::string const&, long,
+        std::string const&);
+    template HPX_EXPORT std::exception_ptr get_exception(
+        std::invalid_argument const&, std::string const&, std::string const&,
+        long, std::string const&);
```
hkaiser, line 318: Why do we need these explicit instantiations? I think these will be instantiated implicitly by the explicit instantiations of the `throw_exception` functions below.

libs/serialization/src/exception_ptr.cpp
```diff
@@ -0,0 +1,311 @@
...
+            ar & err_value;
+            // clang-format on
+        }
+        else if (hpx::util::boost_system_error == type)
+        {
+            // clang-format off
+            ar & err_value & err_message;
+            // clang-format on
+        }
+    }
```
hkaiser, line 182: I think the body of this function should be encapsulated and installed as the default `custom_exception_handler`. If no custom handler is installed we should call `std::terminate` (as this would be a bug).

libs/serialization/src/exception_ptr.cpp
```diff
@@ -0,0 +1,311 @@
...
+                throw_function_, throw_file_, throw_line_);
+            break;
+
+        // hpx::thread_interrupted
+        case hpx::util::hpx_thread_interrupted_exception:
+            e = hpx::detail::construct_lightweight_exception(
+                hpx::thread_interrupted());
+            break;
+        }
+    }
```
hkaiser, line 304: Same here - let's install this as the default custom handler.

libs/errors/src/exception.cpp
msimberg: They are not instantiated otherwise. I did not dig into why though.

hkaiser: Interesting - hmmm. They were instantiated before, did you change something?

msimberg: I'll try removing them again. Maybe they were only needed in an earlier iteration of this PR. There are not so many changes here so I should be able to figure out why they're not instantiated (if they indeed aren't).

libs/serialization/src/exception_ptr.cpp
hkaiser: Also, may I suggest to return the previously installed custom handler from the `set_...` functions. This would allow for the user to chain handlers if necessary.

msimberg: I can certainly add that, but I'd like to emphasize that I have not intended for this to be user-facing. This is mainly to resolve the internal dependencies that we have.

hkaiser: Ok, fair point.

libs/errors/src/exception.cpp
msimberg: It was indeed not needed anymore. Thanks for pointing this out!

libs/serialization/src/exception_ptr.cpp
```diff
@@ -25,281 +25,316 @@
...
+#endif
+            catch (...)
+            {
+                type = hpx::util::unknown_exception;
+                what = "unknown exception";
+            }
 
-        static load_custom_exception_handler_type load_custom_exception_handler;
+            // clang-format off
+        ar & type & what & throw_function_ & throw_file_ & throw_line_;
```
hkaiser, line 146: Is the indentation off or is it just GitHub playing games here?

msimberg: Uh, no, that's my editor. Thanks!

PR #4616, opened 2020-05-11
Fix cmake-format warning

msimberg:
`OUTPUT` keyword is expected to come first in `add_custom_command` (by `cmake-format`).


msimberg:
> How did you find out that swapping the lines removes the warning?

CMake docs and the cmake-format source code (the warning message was quite distinct). I don't think CMake actually cares about where the `OUTPUT` keyword is but it does choose between two different modes for `add_custom_command` depending on if `OUTPUT` or `TARGET` is given, so it makes sense to have it first.

PR #4617, opened 2020-05-12
Ignore user-set compatibility header options if the module does not have compatibility headers

msimberg:


cmake/HPX_AddModule.cmake
```diff
@@ -85,7 +91,9 @@ function(add_hpx_module name)
   hpx_debug("Add module ${name}: SOURCE_ROOT: ${SOURCE_ROOT}")
   hpx_debug("Add module ${name}: HEADER_ROOT: ${HEADER_ROOT}")
 
-  if(HPX_${name_upper}_WITH_COMPATIBILITY_HEADERS)
+  if(${_have_compatibility_headers_option}
+     AND HPX_${name_upper}_WITH_COMPATIBILITY_HEADERS
+  )
     set(COMPAT_HEADER_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/include_compatibility")
```
aurianer, line 97: I would add the `if(EXISTS)` for the `include_compatibility` dir otherwise @biddisco will have the same pb

biddisco: the `set` is protected by `if(${_have_compatibility_headers_option} AND ...` so it should be ok

msimberg: That's the intention. I admit I didn't actually test this so if it doesn't do that I'll fix it (@biddisco feel free to push as well if you have to fix something).

PR #4618, opened 2020-05-13
Officially export HPX::wrap_main target for explicitly linking to hpx_wrap

msimberg:
I hoped we wouldn't need this but it's clearly a thing, so now it's a public target in the `HPX::` namespace. I'll accompany this with documentation before the next release.

@weilewei would you mind giving this a try? You'll need to link to `HPX::hpx_no_wrap_main` and `HPX::wrap_main` (that's going to be fun to explain in the documentation). `HPX::wrap_main` has to be linked before any static library containing `main`.

### DIFF
--- a/wrap/CMakeLists.txt
+++ b/wrap/CMakeLists.txt
@@ -135,3 +135,19 @@ install(
 hpx_export_internal_targets(hpx_wrap)
 
 add_hpx_pseudo_dependencies(core hpx_wrap)
+
+# Add public target for unconditionally linking to hpx_wrap
+add_library(wrap_main INTERFACE)
+
+target_link_libraries(wrap_main INTERFACE hpx_wrap)
+
+install(
+  TARGETS wrap_main
+  EXPORT HPXTargets
+  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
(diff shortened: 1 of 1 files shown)


hkaiser:
@msimberg could you explain the use case for this, please?

msimberg:
There are various ways you can start the runtime:

- `main` in executable, start with `init/start`: `hpx_wrap` is always linked but the linker does not mind because `main` is in an object file
- `main` in exectable, include `hpx_main.hpp`: `hpx_wrap` is linked and does the wrapping
- `main` in static lib, start with `init/start`: can use `HPX::hpx_no_wrap_main` to avoid having `hpx_wrap` be linked after the static lib with `main` and `main` not being found
- `main` in static lib, include `hpx_main.hpp`: no way to express this currently without the user explicitly linking to `hpx_wrap/HPX::wrap_main` in the right place
- `main` in one static lib (e.g. gtest), include `hpx_main.hpp` in another static library: no way to express this either without explicitly linking to `hpx_wrap/HPX::wrap_main`, and impossible to do on Windows (unless similar linker tricks can be done there)

As far as I can tell there is no way for us to enforce with CMake targets in the general case that `hpx_wrap` is always linked before a static lib containing `main` (or just first in the linker line, full stop). If there is a way to do that we can unconditionally link executables to `hpx_wrap` without worrying about where `main` is defined.

msimberg:
#4797 replaces this, closing.

PR #4622, opened 2020-05-14
Sanitizer fixes

sithhell:
This PR fixes various problems that either ubsan or asan identified. In addition, a sanitizer github is added and some warnings happening with clang-11 have been fixed.


hkaiser:
As discussed, this will need to be changed to introduce a new allocator that returns aligned memory (instead of modifying the existing one).

hkaiser:
I have rebased the PR onto current master. However, we know that aligned allocation may cause some performance degradation, so we should make sure that this does not introduce any performance regressions before merging.

hkaiser:
@msimberg, @aurianer this should be good to go now (once the tests cycle).

msimberg:
@hkaiser thanks for getting this up-to-date! The testing and/or bugs will still need some fixing. We could go with the changes as they are but disable the github action. It looks like there's a common leak on almost all examples from `lt_malloc`. Hard to say if it's our fault or from `libltdl`. It might be best to add a sanitizer build on daint instead of github actions (it's only running the examples, and if I remember correctly it didn't quite handle running all the tests).

That means we need to:
- [ ] Test that performance is unchanged. I suppose `future_overhead` is appropriate here as usual. @biddisco how easy is it for you to launch benchmarks with this PR and master to compare? If you have enough other things to do I'll run some tests.
- [ ] Add sanitizer jenkins configuration.
- [ ] Figure out where the remaining leaks are coming from.

It'd be a shame to block this because of the two last points, but the first is a must before it goes into master. I'll try to find some time this week for the remaining tasks.

hkaiser:
Thanks @msimberg, and yes, this needs more work now that things are running again. Let's shelf this PR for after the release.

msimberg:
@hkaiser the noexcept cast won't work very well with C++14. Did you see compilation errors without the cast?

I'm struggling to reproduce the errors related to runtime pointers, but I can roughly guess what's causing it. I've pushed one change, although I don't expect it to fix it. The notifiers most likely just need to be created later in the constructor.

hkaiser:
> @hkaiser the noexcept cast won't work very well with C++14. Did you see compilation errors without the cast?
> 
> I'm struggling to reproduce the errors related to runtime pointers, but I can roughly guess what's causing it. I've pushed one change, although I don't expect it to fix it. The notifiers most likely just need to be created later in the constructor.

I thought I removed the noexcept stuff again (because of the compilation problems). I'll go back to this and verify.

msimberg:
> > @hkaiser the noexcept cast won't work very well with C++14. Did you see compilation errors without the cast?
> > I'm struggling to reproduce the errors related to runtime pointers, but I can roughly guess what's causing it. I've pushed one change, although I don't expect it to fix it. The notifiers most likely just need to be created later in the constructor.
> 
> I thought I removed the noexcept stuff again (because of the compilation problems). I'll go back to this and verify.

Hmm, I hope I didn't at some point force push over your changes... I dropped the noexcept cast commit from here just to see what the builds say about the runtime pointer problems. Just in case you want the noexcept cast stuff and don't have it locally anymore, I pushed the previous state to `sanitizer_fixes_noexcept` (on this repo, not my fork).

msimberg:
Still struggling with the runtime pointer error. I don't think my initial suspicion was right. It looks like it must be a racy failure, since it does not happen every time. It also looks like it happens during shutdown. However, I don't see what could be going wrong, especially since I can't reproduce the failure locally. I've attempted to add a suppression for it now, in the hope that we could try to get at least the fixes that are already there into master.

hkaiser:
> Still struggling with the runtime pointer error.

FWIW, I still think that this could be a virtual function call happening in the destructor.

msimberg:
> > Still struggling with the runtime pointer error.
> 
> FWIW, I still think that this could be a virtual function call happening in the destructor.

It's the call to the `on_stop_thread` handler in one of the service pools that's triggering this. But your comment made me go back and have another look and I've found another suspect. One of the service pools isn't actually joined before stopping the runtime which could lead to the runtime being (partially) destroyed before the handler is called. I'll add a fix for that and see if it helps.

msimberg:
That seems to have quieted the runtime pointer errors. It was nondeterministic so I can't be 100% sure, but at least this looks promising.

@hkaiser there's one more leak reported from the performance counters. Is that something you could have a look at?

hkaiser:
> That seems to have quieted the runtime pointer errors. It was nondeterministic so I can't be 100% sure, but at least this looks promising.

If it was one of the pools not being stopped in time, then I'd expect it to be non-deterministic. Good catch!

> @hkaiser there's one more leak reported from the performance counters. Is that something you could have a look at?

We could try making sure that the object created [here](https://github.com/STEllAR-GROUP/hpx/blob/master/examples/performance_counters/access_counter_set.cpp#L36) goes out of scope before `hpx::finalize()` is being invoked.

msimberg:
> We could try making sure that the object created [here](https://github.com/STEllAR-GROUP/hpx/blob/master/examples/performance_counters/access_counter_set.cpp#L36) goes out of scope before `hpx::finalize()` is being invoked.

Unfortunately it looks like that didn't make a difference (unless my change doesn't actually do what you meant).

msimberg:
@hkaiser I've rebased, and it's just like you said. The (apparently) last error is the one related to performance counters, already mentioned earlier. Much appreciated if you have the time to have a look!

msimberg:
retest

msimberg:
Thanks @hkaiser for pushing the suppressions. That looks like it worked. However, things don't look very good on the other builders. I've triggered the tests again to see if we really just got unlucky, but judging by the number of failures it doesn't look like it.

hkaiser:
> Thanks @hkaiser for pushing the suppressions. That looks like it worked. However, things don't look very good on the other builders. I've triggered the tests again to see if we really just got unlucky, but judging by the number of failures it doesn't look like it.

Yes, something is off :/

msimberg:
> > Thanks @hkaiser for pushing the suppressions. That looks like it worked. However, things don't look very good on the other builders. I've triggered the tests again to see if we really just got unlucky, but judging by the number of failures it doesn't look like it.
> 
> Yes, something is off :/

To be fair it looks like something's been off for a while. I just never paid attention to it because there were sanitizer errors being reported. I suspect that this will need some bisecting.

msimberg:
@hkaiser this looks ok now after reverting the `reinitializable_static` changes. The only thing that doesn't seem to be working is the sanitizers build on daint. I will have a quick look at it but let's not let that stop this from going in. The sanitizers build using GH actions is at least there and working.

hkaiser:
> this looks ok now after reverting the `reinitializable_static` changes. The only thing that doesn't seem to be working is the sanitizers build on daint. I will have a quick look at it but let's not let that stop this from going in. The sanitizers build using GH actions is at least there and working.

Let's get this in! Thanks for looking into it.

msimberg:
I just disabled the sanitizers build on Daint but will still look into it separately from this PR.

PR #4625, opened 2020-05-15
Add main documentation page for HEPs

msimberg:
Not sure we need the note, but thought it might be good to have in this early stage when we're still figuring out how to use these HEP-things.

### DIFF
--- a/docs/sphinx/heps.rst
+++ b/docs/sphinx/heps.rst
@@ -1,0 +1,22 @@
+..
+    Copyright (C) 2020 ETH Zurich
+
+    SPDX-License-Identifier: BSL-1.0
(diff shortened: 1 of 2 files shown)


msimberg:
Side question: is it "HPX Enhancement Proposal" or "HPX Evolution Proposal"? I always imagined it's the first, but I'm happy to go with the second as well.

msimberg:
Closing this since #4593 and #4626 are closed.

hkaiser:
> Side question: is it "HPX Enhancement Proposal" or "HPX Evolution Proposal"? I always imagined it's the first, but I'm happy to go with the second as well.

Since we don't even know how to spell out `HPX`, why should we disclose what's a `HEP`? ;-)

PR #4626, opened 2020-05-15
HEP 2: Public API for HPX 2.0.0

msimberg:
Here's a very rough first draft of the public API for HPX 2.0.0 (although we can already start putting things into place for however many intermediate releases we have before that).

The header and namespace organization is 99% solved by following the standard library conventions. The rest seems to work out quite nicely with `hpx::distributed` for distributed LCOs, and just `hpx` for runtime functionality. I realized the standard libaries do ``std::experimental::abc`` for namespaces so we might actually want to go with ``hpx::experimental::abc`` as well instead of ``hpx::abc::experimental`` as I suggested in #4277. 

Things I've definitely missed:
- Various traits that go with the proposed public functionality. I need to do another pass to see which traits actually need to be public.
- Functionality required for components and actions.
- Everything related to performance counters.
- Components.
- CMake targets.

Things I've intentionally left out:
- Lots of internal functionality. The list is too long to mention.
- Many executors. Some of them could be experimental (`limiting_executor`?). Many of them could or will be removed (thread executors).
- The runtime, thread manager, partitioner classes as they can be accessed through functions like ``hpx::get_locality_id``. The partitioner could be experimental since it is required for runtime initialization. Needs cleanup though (like #4613).
- Thread pools, schedulers. These could also be experimental as I think they're close to being stable, but they would require some cleanup.
- Compute. Do we consider this stable enough?

What have I missed? Please have a look at the list and see if anything you use in your application is missing from the list and should be considered be part of the public API, or at least experimental. Users please comment about missing functionality. @aurianer, @biddisco, @hkaiser, @K-ballo, @sithhell please just push missing things directly to the branch (no force-pushing now! we can clean up the commits at the end).

### DIFF
--- a/docs/sphinx/heps/hep2.rst
+++ b/docs/sphinx/heps/hep2.rst
@@ -1,0 +1,310 @@
+HEP 2: The public API for |hpx| 2.0.0
+=====================================
+
+Motivation
+----------
+
+|hpx| has evolved to have a rather ad-hoc organization of functionality into
+namespaces, headers, and libraries. This HEP specifies the public API, and
+through which namespaces and headers the public API is accessible, for |hpx|
+2.0.0.
+
+Implementation
+--------------
+
+The public API is presented as
+
+- ``header name``
+  - ``hpx::x``
+  - ``hpx::y``
+
+where including ``header name`` gives you access to at least ``hpx::x`` and
+``hpx::y``. Functionality may be organized differently internally (both headers
+and namespaces).
+
+Functionality that corresponds to, or extends, standard library functionality,
+will be in the same namespace as in the standard library with ``std`` replaced
+by ``hpx``, and a header ``abc`` replaced by ``hpx/abc.hpp``. Functionality that
+is proposed for the standard library, and may thus be in ``std::experimental``
+and a header ``experimental/abc`` does not need to be in the
+``hpx::experimental`` namespace or a ``hpx/experimental`` header if the
+functionality is considered stable in |hpx|. Distributed extensions to local
+parallelism and concurrency will be in the ``hpx::distributed`` namespace and
+``hpx/distributed`` headers.
+
+Utilities
+.........
+
+- ``hpx/functional.hpp``
+  - ``hpx::function``
+  - ``hpx::unique_function``
+  - ``hpx::function_nonser``
+  - ``hpx::unique_function_nonser``
+  - ``hpx::bind``
+  - ``hpx::bind_front``
+  - ``hpx::bind_back``
+  - ``hpx::invoke``
+  - ``hpx::invoke_fused``
+  - ``hpx::result_of``
+  - more?
+
+- ``hpx/tuple.hpp``
+  - ``hpx::tuple``
+
+- ``hpx/any.hpp``
+  - ``hpx::any``
+
+- ``hpx/optional.hpp``
+  - ``hpx::optional``
+
+- ``hpx/format.hpp``
+  - ``hpx::format``?
+  - other utilities that go with ``format``
+
+- ``hpx/chrono.hpp``
+  - ``hpx::high_resolution_clock``?
+  - ``hpx::high_resolution_timer``?
+  - ``hpx::scoped_timer``?
+
+- ``hpx/exception.hpp``
+  - ``hpx::exception``
+  - ``HPX_THROW_EXCEPTION``
+
+- ``hpx/system_error.hpp``
+  - ``hpx::error_code``
+
+- ``hpx/assert.hpp``
+  - ``HPX_ASSERT``
+
+- ``hpx/serialization.hpp``
+  - what functionality actually needs to be public?
+  - Some classes have intrusive serialization in which case there's no need to
+    include ``serialization.hpp`` to get the serialization support; others have
+    it implemented in the serialization module. Should we just state that using
+    serialization for any class requires including ``hpx/serialization.hpp``
+    even though it may already be included transitively?
+
+Local concurrency and parallelism
+.................................
+
+- ``hpx/future.hpp``
+  - ``hpx::future``
+  - ``hpx::promise``
+  - ``hpx::shared_future``
+  - ``hpx::async``
+  - ``hpx::sync``
+  - ``hpx::apply``
+  - ``hpx::dataflow``
+  - ``hpx::launch_policy`` (or ``hpx::execution::launch_policy``?)
+  - ``hpx::is_future`` and friends
+
+- ``hpx/future.hpp``?
+  - ``hpx::when_all``
+  - ``hpx::wait_all``
+  - ``hpx::when_any``
+  - ``hpx::wait_any``
+  - ``hpx::when_some``
+  - ``hpx::wait_some``
+  - ``hpx::when_each``
+  - ``hpx::wait_each``
+  - ``hpx::split_future``
+
+- ``hpx/thread.hpp``
+  - ``hpx::thread``
+  - ``hpx::this_thread::yield``
+  - ``hpx::this_thread::get_id``
+  - ``hpx::this_thread::sleep_for``
+  - ``hpx::this_thread::sleep_until``
+
+- ``hpx/algorithm.hpp``
+  - ``hpx::for_loop`` and all the other parallel algorithms
+
+- ``hpx/algorithm/for_loop.hpp``
+  - ``hpx::for_loop`` and the same for all other individual algorithms (no need to separate range-based and iterator-based, I believe)
+
+- ``hpx/execution.hpp``
+  - ``hpx::execution::sequenced_policy``
+  - ``hpx::execution::unsequenced_policy``
+  - ``hpx::execution::parallel_sequenced_policy``
+  - ``hpx::execution::parallel_unsequenced_policy``
+  - ``hpx::execution::sequenced_task_policy``
+  - ``hpx::execution::parallel_task_policy``
+  - ``hpx::execution::seq``
+  - ``hpx::execution::unseq`` (or is it ``datapar``...?)
+  - ``hpx::execution::par``
+  - ``hpx::execution::par_unseq``
+  - ``hpx::execution::task``
+
+- ``hpx/execution.hpp``? these should probably be public, despite upcoming changes to executors?
+  - ``hpx::execution::post``?
+  - ``hpx::execution::sync_execute``?
+  - ``hpx::execution::async_execute``?
+  - ``hpx::execution::then_execute``?
+  - ``hpx::execution::bulk_sync_execute``?
+  - ``hpx::execution::bulk_async_execute``?
+  - ``hpx::execution::bulk_then_execute``?
+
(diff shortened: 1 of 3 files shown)


docs/sphinx/heps/hep2.rst
```diff
@@ -0,0 +1,310 @@
...
+
+- ``hpx/system_error.hpp``
+  - ``hpx::error_code``
+
+- ``hpx/assert.hpp``
+  - ``HPX_ASSERT``
+
+- ``hpx/serialization.hpp``
+  - what functionality actually needs to be public?
+  - Some classes have intrusive serialization in which case there's no need to
```
biddisco, line 81: "Some classes have intrusive serialization" Are you sure you don't mean "Some classes have implicit serialization..."

docs/sphinx/heps/hep2.rst
```diff
@@ -0,0 +1,310 @@
...
+  - ``hpx::promise``
+  - ``hpx::shared_future``
+  - ``hpx::async``
+  - ``hpx::sync``
+  - ``hpx::apply``
+  - ``hpx::dataflow``
+  - ``hpx::launch_policy`` (or ``hpx::execution::launch_policy``?)
+  - ``hpx::is_future`` and friends
+
+- ``hpx/future.hpp``?
```
biddisco, line 101: yes

docs/sphinx/heps/hep2.rst
msimberg: No, intrusive, as in the serialization is implemented by member functions ("Some classes implement serialization intrusively" might be better, but I'm not planning on leaving that comment there). Using such a class gives you access to its serialization functionality whether you ask for it or not. Other classes have the serialization defined outside of the class (e.g. all the standard types) and including something which gives you the class does not necessarily give you the serialization functionality for it.

msimberg:
How do we want to proceed with this one? I could reduce it to just cover the standard library-equivalent headers and we create another one for the other API headers, or we just keep this open until we're done and happy (i.e. when we release 2.0.0). I'm happy either way.

docs/sphinx/heps/hep2.rst
```diff
@@ -0,0 +1,312 @@
...
+
+- ``hpx/future.hpp``
+  - ``hpx::future``
+  - ``hpx::promise``
+  - ``hpx::shared_future``
+  - ``hpx::async``
+  - ``hpx::sync``
+  - ``hpx::apply``
+  - ``hpx::dataflow``
+  - ``hpx::launch_policy`` (or ``hpx::execution::launch_policy``?)
```
hkaiser, line 98: FWIW, the equivalent would be `std::launch` (https://en.cppreference.com/w/cpp/thread/launch).

docs/sphinx/heps/hep2.rst
```diff
@@ -0,0 +1,312 @@
...
+  - ``hpx::is_future`` and friends
```
hkaiser, line 99: I think we should keep our traits in `namespace traits`.

docs/sphinx/heps/hep2.rst
```diff
@@ -0,0 +1,312 @@
...
+
+- ``hpx/execution.hpp``
+  - ``hpx::execution::sequenced_policy``
+  - ``hpx::execution::unsequenced_policy``
+  - ``hpx::execution::parallel_sequenced_policy``
+  - ``hpx::execution::parallel_unsequenced_policy``
+  - ``hpx::execution::sequenced_task_policy``
+  - ``hpx::execution::parallel_task_policy``
+  - ``hpx::execution::seq``
+  - ``hpx::execution::unseq`` (or is it ``datapar``...?)
```
hkaiser, line 131: It's `unseq`, `datapar` is a different thing.

docs/sphinx/heps/hep2.rst
```diff
@@ -0,0 +1,312 @@
...
+  - ``hpx::distributed::all_reduce``
+  - ``hpx::distributed::broadcast``
+  - ``hpx::distributed::fold``
+  - ``hpx::distributed::gather``
+  - ``hpx::distributed::reduce``
+  - ``hpx::distributed::spmd_block``
+
+- ``hpx/distributed/checkpoint.hpp``?
+  - ``hpx::distributed::checkpoint`` stable?
+  - this is not necessarily distributed only...
```
hkaiser, line 223: This is not distributed. Well, checkpointing `id_type` is distributed, but everything else is not. We might want to split this into local and distributed parts.

docs/sphinx/heps/hep2.rst
K-ballo: A `traits` namespace never made sense to me.

docs/sphinx/heps/hep2.rst
```diff
@@ -0,0 +1,312 @@
...
+
+- ``hpx/experimental/distributed/resiliency.hpp``?
+  - ``hpx::experimental::distributed::async_replay`` and other resiliency?
+  - ready for public use or still experimental?
+  - the standard libraries use ``std::experimental::x``, perhaps we should just
+    follow that instead? the argument for ``hpx::x::experimental`` wasn't that
+    strong anyway...
+  - this is not necessarily distributed only...
```
hkaiser, line 231: This is not distributed at all. All functionality is purely built on local facilities (no actions).

docs/sphinx/heps/hep2.rst
```diff
@@ -0,0 +1,312 @@
...
+Runtime
+.......
+
+- ``hpx/runtime.hpp``?
+  - ``hpx::get_os_thread_count``
+  - ``hpx::get_worker_thread_num``
+  - ``hpx::get_thread_name``
+  - other ``hpx::get_*`` functions
+  - ``hpx::register_thread``
+  - ``hpx::unregister_thread``
```
hkaiser, line 245: I think we don't need to expose these two as API functions.

docs/sphinx/heps/hep2.rst
```diff
@@ -0,0 +1,312 @@
...
+- ``hpx/distributed/runtime.hpp``?
+  - all of the above and (should they be in ``hpx::distributed`` or not?)
+  - ``hpx::find_localities``
+  - ``hpx::find_here``
+
+- ``hpx/init.hpp``?
+  - ``hpx::init`` (rename to ``hpx::start_stop`` or something similar? only
+    leave ``hpx::start`` and ``hpx::stop``? I found the relationship between
+    ``start``, ``stop``, ``init``, and ``finalize`` confusing at least in the
+    beginning.)
```
hkaiser, line 267: `start`/`stop` and `init` are serving different use cases. I think we should keep them. Let's improve the docs if necessary instead.

docs/sphinx/heps/hep2.rst
```diff
@@ -0,0 +1,312 @@
...
+  - ``hpx::finalize``
+  - ``hpx::resume``?
+  - ``hpx::suspend``?
+  - ``hpx::init_parameters``
+
+- performance counters?
+
+- ``hpx/wrap_main.hpp`` (I propose to rename ``hpx_main.hpp`` to
+  ``wrap_main.hpp`` to avoid confusion with the implicit entry point used by
+  ``hpx::init`` and ``hpx::start`` being ``hpx_main``)
```
hkaiser, line 279: +1

docs/sphinx/heps/hep2.rst
msimberg: Something like this should at least not be in `hpx::traits`: https://en.cppreference.com/w/cpp/algorithm/is_execution_policy. I'm mostly indifferent about where non-std traits go, but I'd probably slightly prefer just `hpx` (or `hpx::distributed`).

hkaiser: +0

docs/sphinx/heps/hep2.rst
```diff
@@ -0,0 +1,312 @@
...
+  - ``hpx::is_future`` and friends
+  - ``hpx::when_all``
+  - ``hpx::wait_all``
+  - ``hpx::when_any``
+  - ``hpx::wait_any``
+  - ``hpx::when_some``
+  - ``hpx::wait_some``
+  - ``hpx::when_each``
+  - ``hpx::wait_each``
+  - ``hpx::split_future``
```
sithhell, line 108: I would like to see the algorithms related to P0443 in there as well (see http://wg21.link/p1897)

docs/sphinx/heps/hep2.rst
msimberg: I see the use case, it's more about the naming. I don't have better suggestions though, so we can leave them as they are.

docs/sphinx/heps/hep2.rst
msimberg: I think we're actually good here. We have the enum `hpx::detail::launch_policy`, but also `hpx::launch::async` etc which is "correct" according to our standard-equivalent naming. I just need to update the document with `hpx::launch`.

docs/sphinx/heps/hep2.rst
msimberg: Then the question is how stable this and checkpoint are. `hpx::experimental` or `hpx`? I.e. is there at least one active user for either of them, and is the interface likely to still need significant changes?

docs/sphinx/heps/hep2.rst
msimberg: Fine by me.

docs/sphinx/heps/hep2.rst
msimberg: I'm absolutely for that, but just wanted to emphasize that HPX isn't going to be "done" with whatever API we specify here, so these things can also come later. I guess what you're pushing is that you'd like to see them in 2.0.0?

docs/sphinx/heps/hep2.rst
hkaiser: Let's leave it in experimental. The new executor APIs will allow for this in a much more integrated way.

docs/sphinx/heps/hep2.rst
```diff
@@ -0,0 +1,312 @@
...
+  - ``hpx::barrier``
+
+- ``hpx/latch.hpp``
+  - ``hpx::latch``
+
+- ``hpx/channel.hpp``
+  - ``hpx::channel``
+
+- ``hpx/spmd_block.hpp``
+  - ``hpx::spmd_block``
```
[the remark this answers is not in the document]
hkaiser: This was added by Antoine a while back. It's not in the TS and not used anywhere, IIRC. We should be able to move this to `experimental`.

msimberg:
I'm going to close this now as we've made enough progress on adding things to the public api. I'd say the process for proposing additions to the api is simply to add whatever is to be added to this page: https://hpx-docs.stellar-group.org/branches/master/html/api/public_api.html. Before 2.0.0 we'll have to still do a recap to see that everything we feel should be in the api by then is actually listed and moved to the correct namespace etc.

PR #4633, opened 2020-05-16
Unify execution parameters customization points

hkaiser:
- This unifies the processing of all execution parameters customization
  points such that all invocations are dispatched through the executor
  first before handled by the parameters object (if the executor does not
  implement the functionality)
- This also consolidates the customization points `count_processing_units`
  and `processing_units_count` by removing the former as it was never used
  anywhere. This is a slightly breaking change, which shouldn't have any
  impact on existing code.

### DIFF
--- a/libs/execution/tests/unit/executor_parameters_dispatching.cpp
+++ b/libs/execution/tests/unit/executor_parameters_dispatching.cpp
@@ -1,0 +1,477 @@
+//  Copyright (c) 2020 Hartmut Kaiser
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/hpx.hpp>
+#include <hpx/hpx_init.hpp>
+#include <hpx/include/parallel_executor_parameters.hpp>
+#include <hpx/include/parallel_executors.hpp>
+#include <hpx/testing.hpp>
+
+#include <atomic>
+#include <string>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+// This test verifies that all parameters customization points dispatch
+// through the executor before potentially being handled by the parameters
+// object.
+
+std::atomic<std::size_t> params_count(0);
+std::atomic<std::size_t> exec_count(0);
+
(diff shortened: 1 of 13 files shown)


libs/thread_executors/tests/unit/resource_manager.cpp
```diff
@@ -193,7 +193,8 @@ void test_executors_shrink(std::size_t num_pus)
     using namespace hpx::parallel::execution;
     std::size_t processing_units = hpx::get_os_thread_count();
 
-    processing_units = (processing_units / num_pus) * num_pus;
+    processing_units = (std::min)(
+        ((processing_units + 1) / num_pus) * num_pus - 1, std::size_t(1));
```
msimberg, line 197: Why the `min`? Looks like that makes `processing_units` either 0 or 1, always. I don't quite understand what the test is checking so that may be correct...

hkaiser: Right, this is wrong. I'll go back and fix this.

libs/execution/tests/unit/executor_parameters_dispatching.cpp
```diff
@@ -0,0 +1,477 @@
...
+    test_mark_end_of_scheduling();
+    test_mark_end_execution();
+
+    return hpx::finalize();
+}
+
+int main(int argc, char* argv[])
+{
+    // By default this test should run on all available cores
+    std::vector<std::string> const cfg = {"hpx.os_threads=all"};
```
msimberg, line 470: Does this really need `all`? That means hyperthreads as well. `cores` (default) or `4` would possibly be enough?

msimberg:
Is this ready to go?

libs/execution/tests/unit/executor_parameters_dispatching.cpp
hkaiser: We've been using this in many tests for a long time. This is just a copy&paste of some related test, I believe. I think nowadays using all cores is the default anyways, so we can go ahead and remove such configurations from all places.

msimberg: Then let's keep it like this. We can go and change it everywhere afterwards if we feel like it.

hkaiser:
> Is this ready to go?

Yes, I think this is ready to go. The CI error is unrelated.

Issue #4329, opened 2020-01-24
HPX 2


msimberg:
I'm opening this as a placeholder for discussion on long term plans for HPX (which we may call 2). See this as a wishlist of things that should be finally deprecated, cleaned up, or implemented.

I'll start (please extend this list):
- Follow semver
- Complete modularization
- CMake modernization
- Namespace and header location cleanup (clear policy on what's part of the API, what's experimental, what's private)
- Executors; there should be a clear story about what executor to use in which situations

Many of these things are already in progress but moving to 2 would allow us to "officially" clean up and break compatibility without feeling too bad about ourselves. If needed we can keep 1.X in lifesupport mode for users who can't take all the breaking changes yet.

The namespace cleanup is partly motivated by having e.g. an `experimental` namespace (this has been discussed before, but should be formalized) so that we can introduce new features and if needed break or remove them without deprecation warnings as long as they are experimental.

The alternative to this is of course keep incrementally adding and removing features as we've been doing now, but I think going for a new major release would be good for both visibility and for allowing us to clean up things properly.

McKillroy:
I want to suggest to add the fill Doxygen documentation to the build, along with other documentation improvements. 
E.g. the dependency graphs created by Doxygen are incredibly useful, also when I need to find out which construct requires which #include I refer to the Doxygen docs.

msimberg:
@McKillroy just to be clear, what you mean is to use the "normal" HTML documentation generated by Doxygen, as opposed to the one we include using Breathe in Sphinx (because the latter is not particularly good...)?

McKillroy:
@msimberg Yes. Breathe is an attempt to feed Doxygen into Sphinx, which in theory is a nice idea, but practically doesn't work that well. 

My idea of documentation would be to have the high level docs in Sphinx, because that's what it's good for, and low level stuff and API details in the code, so a programmer can directly supply it while coding.

Doxygen - despite it's shortcomings still imho is the best ad hoc documentation for C++ programmers available, and the graphs are incredibly useful.

[...]

PR #4634, opened 2020-05-17
Adding C++20 counting semaphore API

hkaiser:
- adding `hpx::binary_semaphore`
- adding `hpx/counting_semaphore.hpp` that imports it into namespace hpx
- flyby: adding `hpx/condition_variable.hpp` and `hpx/stop_token.hpp` headers for
         conformance. This also imports `lcos::local::conditional_variable`
         into namespace hpx

### DIFF
--- a/libs/synchronization/include/hpx/semaphore.hpp
+++ b/libs/synchronization/include/hpx/semaphore.hpp
@@ -1,0 +1,42 @@
+//  Copyright (c) 2020 Hartmut Kaiser
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/config.hpp>
+#include <hpx/synchronization/counting_semaphore.hpp>
+
+#include <cstddef>
+
+///////////////////////////////////////////////////////////////////////////////
+// C++20 counting semaphores
+
+namespace hpx {
+
+    // Semaphores are lightweight synchronization primitives used to constrain
+    // concurrent access to a shared resource. They are widely used to
+    // implement other synchronization primitives and, whenever both are
+    // applicable, can be more efficient than condition variables.
+
+    // A counting semaphore is a semaphore object that models a non - negative
+    // resource count. A binary semaphore is a semaphore object that has only
+    // two states.
+    //
+    // Class template counting_semaphore maintains an internal counter that is
+    // initialized when the semaphore is created. The counter is decremented
+    // when a thread acquires the semaphore, and is incremented when a thread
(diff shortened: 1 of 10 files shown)


msimberg:
@hkaiser I summarized our API discussion from yesterday in #4329. Should we already go ahead and collect the "API headers" into the `include` module in this PR (and #4636, #4635)? I can also collect them afterwards while doing all the other headers.

libs/synchronization/include/hpx/semaphore.hpp
```diff
@@ -0,0 +1,42 @@
...
+    /// A counting semaphore is a semaphore object that models a non - negative
+    /// resource count. A binary semaphore is a semaphore object that has only
+    /// two states.
+    ///
+    /// Class template counting_semaphore maintains an internal counter that is
+    /// initialized when the semaphore is created. The counter is decremented
+    /// when a thread acquires the semaphore, and is incremented when a thread
+    /// releases the semaphore. If a thread tries to acquire the semaphore when
+    /// the counter is zero, the thread will block until another thread
+    /// increments the counter by releasing the semaphore.
```
msimberg, line 33: No doxygen?
```suggestion
    // increments the counter by releasing the semaphore.
```

libs/synchronization/include/hpx/semaphore.hpp
```diff
@@ -0,0 +1,42 @@
...
+    template <std::ptrdiff_t LeastMaxValue = PTRDIFF_MAX>
+    using counting_semaphore =
+        hpx::lcos::local::cpp20_counting_semaphore<LeastMaxValue>;
+
+    /// A binary semaphore should be more efficient than the default
+    /// implementation of a counting semaphore with a unit resource count.
```
msimberg, line 39: No doxygen?
```suggestion
    // implementation of a counting semaphore with a unit resource count.
```

hkaiser:
> Should we already go ahead and collect the "API headers" into the include module in this PR (and #4636, #4635)?

I thought we will create at least 3 `include` modules, one for each of the planned configurations. Should we go ahead and make that split now?

msimberg:
> > Should we already go ahead and collect the "API headers" into the include module in this PR (and #4636, #4635)?
> 
> I thought we will create at least 3 `include` modules, one for each of the planned configurations. Should we go ahead and make that split now?

Yes, I'd say that's the plan in the end. But if we decide to skip splitting up the main library for 1.5.0 then we could still already have them in the include module. Either way. All I would like is that the headers from #4626 that we can put in place we do put in place for 1.5.0. It doesn't matter too much which module they are in before the main library is split into three.

hkaiser:
This should be good to go now as well.

msimberg:
Thanks @hkaiser. I just rebased this because of a conflict with #4634 in the CMake files. I'll merge this as soon as some of the GH actions builders come back clean.

PR #4635, opened 2020-05-17
Adding C++20 latch API

hkaiser:
- adding `hpx/latch.hpp` that imports `hpx::lcos::local::latch` into namespace hpx
- flyby: adding `hpx/thread.hpp` for compatibility

### DIFF
--- a/libs/synchronization/include/hpx/synchronization/latch.hpp
+++ b/libs/synchronization/include/hpx/synchronization/latch.hpp
@@ -16,6 +16,7 @@
 
 #include <atomic>
 #include <cstddef>
+#include <cstdint>
 #include <mutex>
 #include <utility>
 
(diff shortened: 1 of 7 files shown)


libs/synchronization/include/hpx/synchronization/latch.hpp
```diff
@@ -72,78 +59,165 @@ namespace hpx { namespace lcos { namespace local {
...
+        ///                 supports.
+        static constexpr std::ptrdiff_t(max)() noexcept
+        {
+            return PTRDIFF_MAX;
+        }
+
+        /// Decrements counter_ by n. Does not block.
         ///
-        /// Requires: counter_ > 0.
+        /// Requires: counter_ >= n and n >= 0.
```
aurianer, line 80: `update` instead of `n`?

PR #4636, opened 2020-05-17
Adding C++20 barrier

hkaiser:
- adding hpx/barrier.hpp that imports lcos::local::cpp20_barrier into namespace hpx

### DIFF
--- a/libs/synchronization/include/hpx/synchronization/barrier.hpp
+++ b/libs/synchronization/include/hpx/synchronization/barrier.hpp
@@ -10,16 +10,219 @@
 #pragma once
 
 #include <hpx/config.hpp>
+#include <hpx/assertion.hpp>
 #include <hpx/synchronization/condition_variable.hpp>
 #include <hpx/synchronization/spinlock.hpp>
 
 #include <climits>
 #include <cstddef>
+#include <cstdint>
+#include <utility>
 
 #include <hpx/config/warnings_prefix.hpp>
 
 ///////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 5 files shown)


libs/synchronization/include/hpx/synchronization/barrier.hpp
```diff
@@ -15,11 +15,213 @@
...
+    ///
+    /// The default value of the CompletionFunction template parameter is an
+    /// unspecified type, such that, in addition to satisfying the requirements
+    /// of CompletionFunction, it meets the Cpp17DefaultConstructible
+    /// requirements (Table 27) and completion() has no effects.
+    ///
+    /// barrier::arrival_token is an unspecified type, such that it meets the
+    /// Cpp17MoveConstructible (Table 28), Cpp17MoveAssignable (Table 30), and
+    /// Cpp17Destructible (Table 32) requirements.
+
```
msimberg, line 88: Do you know if doxygen will take the docs above into account with the empty line (same with the other empty lines further up)?

hkaiser:
> You've been busy, thanks! Do you know if we have any internal uses of barrier that should be replaced by the new one?

I wanted to go through our code in a second sweep to replace thing where possible.

libs/synchronization/include/hpx/synchronization/barrier.hpp
hkaiser: I shouldn't have marked all the standardeese as doxygen comments in the first place.

hkaiser:

PR #4640, opened 2020-05-19
Implementing receiver concept in terms of P0443r13

sithhell:
One deviation of P0443r13 is the usage of tag_invoke as customization
point

This depends on #4305.


libs/basic_execution/include/hpx/basic_execution/receiver.hpp
```diff
@@ -0,0 +1,365 @@
...
+    hpx::functional::tag_invoke(                                               \
+        *this, std::forward<Receiver>(rcv), std::forward<Values>(values)...)
+        template <typename Receiver, typename... Values>
+        constexpr HPX_FORCEINLINE auto tag_invoke_impl(
+            std::integral_constant<bool, true> /* is tag invocable */,
+            Receiver&& rcv, Values&&... values) const
+            noexcept(
+                noexcept(HPX_BASIC_EXECUTION_RECEIVER_SET_VALUE_EXPRESSION))
+                -> decltype(HPX_BASIC_EXECUTION_RECEIVER_SET_VALUE_EXPRESSION)
+        {
```
hkaiser, line 115: Wouldn't it be (slightly) simpler to use `decltype(auto)` as the return type instead of spelling out the fulle `-> decltype(…)`?

libs/basic_execution/include/hpx/basic_execution/receiver.hpp
```diff
@@ -0,0 +1,365 @@
...
+    struct is_receiver_of;
+
+    HPX_INLINE_CONSTEXPR_VARIABLE struct set_value_t
+    {
+#define HPX_BASIC_EXECUTION_RECEIVER_SET_VALUE_EXPRESSION                      \
+    hpx::functional::tag_invoke(                                               \
+        *this, std::forward<Receiver>(rcv), std::forward<Values>(values)...)
+        template <typename Receiver, typename... Values>
+        constexpr HPX_FORCEINLINE auto tag_invoke_impl(
+            std::integral_constant<bool, true> /* is tag invocable */,
```
hkaiser, line 110: use `std::true_type` instead?

libs/basic_execution/include/hpx/basic_execution/receiver.hpp
hkaiser: For some of the customization points (the ones without the `noexcept()` clause) this would also remove the need for defining the macro.

libs/basic_execution/include/hpx/basic_execution/receiver.hpp
```diff
@@ -0,0 +1,365 @@
...
+    }    // namespace detail
+
+    template <typename T, typename... As>
+    struct is_receiver_of
+      : detail::is_receiver_of_impl<is_receiver_v<T>, T, As...>
+    {
+    };
+
+    template <typename T, typename... As>
+    constexpr bool is_receiver_of_v = is_receiver_of<T, As...>::value;
```
hkaiser, line 336: Should these trait types go into `namespace hpx::traits`?

libs/basic_execution/include/hpx/basic_execution/receiver.hpp
sithhell: unfortunately, that's not going to work... it needs to be spelled out fully for SFINAE to work so that the dispatching works properly.
Unless we require C++17 and up, IIRC.

libs/basic_execution/include/hpx/basic_execution/receiver.hpp
sithhell: yup, good point.

libs/basic_execution/include/hpx/basic_execution/receiver.hpp
sithhell: I guess they should, to stay consistent with other parts of the library.

sithhell:
Since this PR is creating a cycle between functional, concurrency and basic_execution caused by itt_notify support, the question is, where this should go. A generic `profiling` module or `itt_notify`?

msimberg:
> Since this PR is creating a cycle between functional, concurrency and basic_execution caused by itt_notify support, the question is, where this should go. A generic `profiling` module or `itt_notify`?

What exactly is the cycle that you see? `functional` should definitely not depend on either of the two other modules. The headers also don't seem to introduce any new dependencies between `concurrency` and `basic_execution`. If there is a cycle I'd call it `itt_notify` since we may want to use the `profiling` name for other things.

msimberg:
Commented on the wrong PR first...

For the record, there's a graph up now and it says `functional` -> `concurrency` -> `basic_execution` -> `functional`. The `functional` -> `concurrency` dependency is worrying, and strange considering you haven't changed any files in `functional`...

hkaiser:
> Commented on the wrong PR first...
> For the record, there's a graph up now and it says functional -> concurrency -> basic_execution -> functional. The functional -> concurrency dependency is worrying, and strange considering you haven't changed any files in functional...

FWIW, `functional` already depends on `concurrency`, which is the worrying part:
```
  "functional" [shape=oval];                           
  "functional" -> "serialization" [color=lightblue];   
  "functional" -> "debugging" [color=blue];            
  "functional" -> "config" [color=blue];               
  "functional" -> "concurrency" [color=blue];          
  "functional" -> "preprocessor" [color=blue];         
  "functional" -> "type_support" [color=blue];         
  "functional" -> "assertion" [color=blue];            
  "functional" -> "datastructures" [color=blue];       
```
Also:
```
functional -> concurrency                                                                                                                
  .\functional\src\basic_function.cpp includes .\concurrency\include\hpx\concurrency\itt_notify.hpp                                      
  .\functional\include\hpx\functional\traits\get_action_name.hpp includes .\concurrency\include\hpx\concurrency\itt_notify.hpp           
  .\functional\include\hpx\functional\traits\get_function_annotation.hpp includes .\concurrency\include\hpx\concurrency\itt_notify.hpp   
```

I think extracting itt_notify from the concurrency module should help resolving the issue.

msimberg:
@sithhell looks like GCC isn't quite happy with this: https://cdash.cscs.ch//viewBuildError.php?buildid=110431 and https://cdash.cscs.ch//viewBuildError.php?buildid=110520. You'll most likely need the same change as here for the first error: https://github.com/STEllAR-GROUP/hpx/pull/4684/files#diff-85a76557329fe869d2511ee895e3649d.

hkaiser:
@sithhell Please review the changes

msimberg:
Thanks @hkaiser for pushing this forward. My only remaining bikeshed-comment is that we maybe put this in `hpx::execution::experimental`? I think in terms of implementation we'll figure any possible kinks out once we actually start using the concepts.

hkaiser:
This is subsumed by #4665, closing.

PR #4641, opened 2020-05-19
Add a pre-commit hook script to check file formatting

biddisco:
image

The use of clang-format and cmake-format in HPX makes it
tedious to check that formatting of files is correct prior to commits;
this hook can be installed to run checks on files that are about to be
committed.

To install the hook, copy the file to your git hooks as follows
```
cp tools/pre-commit .git/hooks/pre-commit
chmod +x .git/hooks/pre-commit
```
If all files are ok, then the commit will happen as normal,
but if any file fails the clang/cmake format checks, then the commit
will be aborted and a message such as the following will be displayed
(see image above)

```
clang-format -style=file -i libs/compute_cuda/include/hpx/compute/cuda.hpp
clang-format -style=file -i libs/cuda_support/include/hpx/cuda_support/target.hpp
cmake-format -i libs/compute_cuda/CMakeLists.txt
cmake-format -i libs/cuda_support/test.cmake
```
The format of the message is intended to allow you to copy and paste it
into your terminal and execute the fixes immediately. Automatic
formatting could be done, but has not been implemented as most
developers will probably prefer to have direct control of the process.

### DIFF
--- a/tools/pre-commit
+++ b/tools/pre-commit
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+# simple pre commit hook script to check that *.cpp and *.hpp files 
+# are correctly clang-formatted and that CMakeLists.txt and *.cmake
+# files are cmake-formatted
+
+# to install this hook, copy this file to your git hooks as follows
+# cp tools/pre-commit .git/hooks/pre-commit
+# chmod +x .git/hooks/pre-commit
+
+red=$(tput setaf 1)
+green=$(tput setaf 2)
+yellow=$(tput setaf 3)
+blue=$(tput setaf 4)
+normal=$(tput sgr0)
+
+cxxfiles=()
+for file in `git diff --cached --name-only --diff-filter=ACMRT | grep -E "\.(cpp|hpp)$"`; do
+    if ! cmp -s <(git show :${file}) <(git show :${file}|clang-format -style=file); then
+        cxxfiles+=("${file}")
+    fi
+done
+
+cmakefiles=()
+for file in `git diff --cached --name-only --diff-filter=ACMRT | grep -E "(CMakeLists\.txt|\.cmake)$"`; do
+    if ! cmp -s <(git show :${file}) <(git show :${file}|cmake-format); then
+        cmakefiles+=("${file}")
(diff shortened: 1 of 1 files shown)


tools/pre-commit
```diff
@@ -0,0 +1,49 @@
...
+        cmakefiles+=("${file}")
+    fi
+done
+
+returncode=0
+
+if [ -n "${cxxfiles}" ]; then
+    printf "# ${blue}clang-format ${red}error pre-commit${normal} : To fix run the following (use git commit ${yellow}--no-verify${normal} to bypass)\n"
+    for f in "${cxxfiles[@]}" ; do
+        printf "clang-format -style=file -i %s\n" "$f"
```
msimberg, line 36: Maybe:
```suggestion
        printf "clang-format -i %s\n" "$f"
```
? Since it'll use the file at the repository root by default.

biddisco: `-style=file` uses the file loaded from the parent directories- I put this in to make it explicit that we will use the file from the repo root - even if the clang-format default were to change (which it won't)
```
  --style=<string>    ...
                               Use -style=file to load style configuration from
                               .clang-format file located in one of the parent
                               directories of the source file (or current
                               directory for stdin).
```

msimberg: Oh, very nice! Never mind then. Thanks!

tools/pre-commit
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
hkaiser, line 1: Please add copyright and license information

msimberg:
I realized just after merging that this will also check source files not in modules. However, I think we're far enough in the modularization that if we format non-modularized files as we edit them it won't be the end of the world. Let's just try to separate the formatting commits from actual changes.

biddisco:
Understood. Fortunately, this is opt-in only - for my own sanity checking. 

However, I will add an extra path check to enforce it only for files in the `libs/` path. I may also add an inspect check as well, because that also annoys me. However, the user would need to set an ENV var to point to the inspect binary since that most likely won't be present on the majority of machines.

PR #4644, opened 2020-05-19
Fixing support for mpirun (and similar)

hkaiser:
- Checking MPI environment before initializing the runtime
- flyby: handling stack size problem for newly created threads

### DIFF
--- a/libs/threading_base/include/hpx/threading_base/scheduler_base.hpp
+++ b/libs/threading_base/include/hpx/threading_base/scheduler_base.hpp
@@ -251,8 +251,9 @@ namespace hpx { namespace threads { namespace policies {
             switch (stacksize)
             {
             case thread_stacksize_small:
+                HPX_FALLTHROUGH;
+            case thread_stacksize_current:
                 return thread_queue_init_.small_stacksize_;
-                break;
 
             case thread_stacksize_medium:
                 return thread_queue_init_.medium_stacksize_;
--- a/libs/resource_partitioner/CMakeLists.txt
+++ b/libs/resource_partitioner/CMakeLists.txt
@@ -48,6 +48,7 @@ add_hpx_module(
     hpx_functional
     hpx_hardware
     hpx_memory
+    hpx_mpi_base
     hpx_pack_traversal
     hpx_plugin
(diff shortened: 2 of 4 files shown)


libs/threading_base/include/hpx/threading_base/scheduler_base.hpp
```diff
@@ -251,8 +251,9 @@ namespace hpx { namespace threads { namespace policies {
             switch (stacksize)
             {
             case thread_stacksize_small:
+                HPX_FALLTHROUGH;
+            case thread_stacksize_current:
```
msimberg, line 255: Did you have a case where `stacksize` still was `thread_stacksize_current` at this point? I would rather put an assert in the first if after `stacksize` has been updated as this shouldn't happen. A created HPX thread should always have one of the actual sizes (small, medium, ...), and a non-HPX thread should return the default which is the small stacksize at the moment.

hkaiser: Yes, that's what I was seeing. I did add an assert inside the `if()`, which fired for me. I tried to investigate but was not able to figure out why this happens.

msimberg: Hrm, I'm clearly wrong in that case. I think what we should do is reset a `thread_stacksize_current` to the a more concrete size when creating the thread. This is most likely a bug from my earlier changes. We can go with this and I'll try to fix it afterwards.

hkaiser:
@khuck wrote in an email (Tue 5/19/2020 5:55 PM): 
> I can confirm that the branch works for me…thanks!  I am able to run MPI programs again.

libs/threading_base/include/hpx/threading_base/scheduler_base.hpp
hkaiser: Yes, I don't like this 'fix' either. See #4648 for a separate ticket.

msimberg:
Thanks @hkaiser for opening the new issue. I'll take care of that one. Let's go with this since having the MPI parcelport broken is suboptimal...

PR #4645, opened 2020-05-20
Add a few more API headers

msimberg:
Follows @hkaiser's lead and adds headers for exceptions, a few utilities, and algorithms.

It may be that we have to not just forward names to the `hpx` namespace, but put them there in the first place to get the documentation to look reasonable, but that's a separate issue.


libs/algorithms/include/hpx/algorithm.hpp
```diff
@@ -0,0 +1,72 @@
...
+    using hpx::parallel::set_difference;
+    using hpx::parallel::set_intersection;
+    using hpx::parallel::set_symmetric_difference;
+    using hpx::parallel::set_union;
+    using hpx::parallel::sort;
+    using hpx::parallel::stable_partition;
+    using hpx::parallel::swap_ranges;
+    using hpx::parallel::unique;
+    using hpx::parallel::unique_copy;
+}    // namespace hpx
```
hkaiser, line 72: This misses some of the algorithms (there might be more): 
- adjacent_difference 
- inner_product
- exclusive_scan 
- inclusive_scan
- find_end
- reduce
- max_element 
- min_element
- includes
- transform
- all uninitialized_xxx algorithms

However some of those belong into different headers (`hpx/numeric.hpp` and `hpx/memory.hpp`)

libs/datastructures/include/hpx/tuple.hpp
```diff
@@ -0,0 +1,21 @@
...
+    using hpx::util::forward_as_tuple;
+    using hpx::util::get;
+    using hpx::util::ignore;
+    using hpx::util::make_tuple;
+    using hpx::util::tie;
+    using hpx::util::tuple;
+    using hpx::util::tuple_cat;
+    using hpx::util::tuple_element;
+    using hpx::util::tuple_size;
+}    // namespace hpx
```
hkaiser, line 21: This reminds me of (C++17) `std::apply`, our equivalent is `invoke_fused`. However we would create a clash with our `apply` if we imported it into namespace hpx.

libs/synchronization/include/hpx/mutex.hpp
```diff
@@ -0,0 +1,21 @@
...
+namespace hpx {
+    using hpx::lcos::local::call_once;
+    using hpx::lcos::local::mutex;
+    using hpx::lcos::local::no_mutex;
+    using hpx::lcos::local::once_flag;
+    using hpx::lcos::local::recursive_mutex;
+    using hpx::lcos::local::shared_mutex;
+    using hpx::lcos::local::spinlock;
+    using hpx::util::unlock_guard;
+}    // namespace hpx
```
hkaiser, line 21: Another note to self: we need to implement `timed_mutex` and `recursive_timed_mutex`.

libs/timing/include/hpx/chrono.hpp
```diff
@@ -0,0 +1,17 @@
...
+#pragma once
+
+#include <hpx/timing/high_resolution_clock.hpp>
+#include <hpx/timing/high_resolution_timer.hpp>
+#include <hpx/timing/steady_clock.hpp>
+
+namespace hpx {
+    using hpx::util::high_resolution_clock;
+    using hpx::util::high_resolution_timer;
+    using hpx::util::steady_time_point;
```
hkaiser, line 16: Does this belong here?

libs/algorithms/include/hpx/algorithm.hpp
msimberg: Good point. I can add those headers as well.

libs/datastructures/include/hpx/tuple.hpp
msimberg: Yes, I noticed that while looking at what's in `<tuple>`! There's not much we can do about that without deprecating the current `hpx::apply` though...

libs/timing/include/hpx/chrono.hpp
msimberg: Not necessarily. I added it since `<chrono>` has `time_point`, and this looks like a simple extension to that. It doesn't have to be here though.

msimberg:
The `hpx/future.hpp` header is going to be a bit problematic because of `hpx::promise`. We might have to deprecate the `hpx::promise` alias first before we change it to `hpx::lcos::local::promise`. I'll have to check if deprecation attributes work on just an alias or if it will apply it to the aliased name as well.

libs/algorithms/include/hpx/algorithm.hpp
msimberg: I'm leaving `hpx/memory.hpp` out from this PR because of #4647. I notice we have an `accumulate` in the `detail` namespace. Should we make that public or just use `std::accumulate` instead since our `accumulate` isn't (and shouldn't) be parallel anyway?

hkaiser: There is no `std::accumulate(par, …)`, it's called `std::reduce`. The code in the detail namespace is just an implementation detail for our algorithms.

libs/synchronization/include/hpx/mutex.hpp
msimberg: And a note to myself before merging this: we already have `timed_mutex` so add it (we don't have `recursive_timed_mutex`).

libs/algorithms/include/hpx/algorithm.hpp
msimberg: I guess the question is if we can use `std::accumulate` as an implementation detail instead of the `accumulate` we have?

hkaiser: Unfortunately, we cannot do that. While our accumulate is functionally equivalent to `std::accumulate` it allows for the begin and end iterators to be different, while the std version assumes that both iterators are of the same type.

msimberg:
I think is ready to go now. The API headers are not complete by any means, but this is step in that direction.

Issue #4127, opened 2019-10-08
Local runtime entry-points


msimberg:
Opening this to discuss how to expose a "local runtime" to users.

I have two main criteria:
1. The current `hpx::init/start` etc. overloads should work unchanged with the headers they require now.
2. The local and distributed runtimes can be enabled (not running) in the same build (to avoid increasing the number of different configurations there are).

For this to work I'd propose to add new headers (`hpx/local/hpx_init.hpp`? `hpx/hpx_init_local.hpp`? not sure how `hpx/hpx_main.hpp` will work out) and overloads of `hpx::init` perhaps taking an additional tag type as a first argument. The goal would be reuse as much of the init infrastructure as possible, but without having to pull in any distributed headers when using the local runtime.

I'd like to hear ideas for other approaches for this, or if this is overkill in the first place.

hkaiser:
@msimberg Could we extend (and reuse) `hpx::runtime_mode` (https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/runtime/runtime_mode.hpp#L21-L31) for this?

msimberg:
I think semantically it would work very well (`runtime_mode_local`?). Thinking about this a bit more I realize point 2 above is most likely unnecessary (and unnecessarily complicated). Point 1 I'd still like to keep and using the runtime mode fits quite well with that (users would have to explicitly ask for the local runtime).

hkaiser:
`runtime_mode_local` sounds fine to me. We might however want to add more overloads for `hpx::init`/`hpx::start` that make it more convenient to specify this. But this might be another issue - reworking those overloads....

msimberg:
Yep, the overloads are relevant but can quite easily be handled separately.

hkaiser:
Another question: how would this functionality be different from what we already have today, namely that networking is not enabled silently if we can detect that the runtime is used locally only? Would `runtime_mode_local` enforce this behavior or do you plan to do additional things in this case?

msimberg:
My intention with this option is to make the choice explicit, but you're right that there isn't a big functional difference. The main difference is that `runtime_mode_local` would pick the local runtime in #4091 which would not even support using actions and components locally. There may be a small speedup in startup even compared to when networking is automatically disabled due to not having to set any of AGAS etc. up. It could be that keeping this choice implicit through compile-time options will be easier in the end though. Thanks @hkaiser for the good questions!

hkaiser:
[...]

PR #4646, opened 2020-05-20
Runtime type

msimberg:
Note: this builds on #4606. Only the last commit belongs to this PR.

This adds a `runtime_type` enum that can be passed in the `init_params` struct, to choose which runtime to start. If the distributed runtime isn't enabled and it's requested an exception will be thrown during init. The default is to use the distributed runtime if available and otherwise the local one.

### DIFF
--- a/libs/init_runtime/include/hpx/hpx_init_params.hpp
+++ b/libs/init_runtime/include/hpx/hpx_init_params.hpp
@@ -94,9 +94,9 @@ namespace hpx {
     /// \var mode           The mode the created runtime environment
     ///                     should be initialized in. There has to be exactly
     ///                     one locality in each HPX application which is
-    ///                     executed in console mode (\a hpx::runtime_mode_console),
+    ///                     executed in console mode (\a hpx::runtime_mode::console),
     ///                     all other localities have to be run in worker mode
-    ///                     (\a hpx::runtime_mode_worker). Normally this is
+    ///                     (\a hpx::runtime_mode::worker). Normally this is
     ///                     set up automatically, but sometimes it is necessary
     ///                     to explicitly specify the mode.
     /// \var rp_mode        The mode the resource partitioner should be created
@@ -114,7 +114,7 @@ namespace hpx {
         std::vector<std::string> cfg;
         mutable startup_function_type startup;
         mutable shutdown_function_type shutdown;
-        hpx::runtime_mode mode = ::hpx::runtime_mode_default;
+        hpx::runtime_mode mode = ::hpx::runtime_mode::default_;
         hpx::resource::partitioner_mode rp_mode = ::hpx::resource::mode_default;
         hpx::resource::rp_callback_type rp_callback;
     };
--- a/libs/runtime_configuration/include/hpx/runtime_configuration/runtime_mode.hpp
+++ b/libs/runtime_configuration/include/hpx/runtime_configuration/runtime_mode.hpp
@@ -23,7 +23,8 @@ namespace hpx {
         runtime_mode_worker = 1,     ///< The runtime is a worker locality
         runtime_mode_connect = 2,    ///< The runtime is a worker locality
                                      ///< connecting late
-        runtime_mode_default = 3,    ///< The runtime mode will be determined
+        runtime_mode_local = 3,      ///< The runtime is fully local
+        runtime_mode_default = 4,    ///< The runtime mode will be determined
                                      ///< based on the command line arguments
         runtime_mode_last
     };
--- a/libs/runtime_configuration/include/hpx/runtime_configuration/runtime_mode.hpp
+++ b/libs/runtime_configuration/include/hpx/runtime_configuration/runtime_mode.hpp
@@ -16,18 +16,40 @@
 namespace hpx {
     /// A HPX runtime can be executed in two different modes: console mode
     /// and worker mode.
-    enum runtime_mode
+    enum class runtime_mode
     {
-        runtime_mode_invalid = -1,
-        runtime_mode_console = 0,    ///< The runtime is the console locality
-        runtime_mode_worker = 1,     ///< The runtime is a worker locality
-        runtime_mode_connect = 2,    ///< The runtime is a worker locality
-                                     ///< connecting late
-        runtime_mode_local = 3,      ///< The runtime is fully local
-        runtime_mode_default = 4,    ///< The runtime mode will be determined
-                                     ///< based on the command line arguments
-        runtime_mode_last
+        invalid = -1,
+        console = 0,     ///< The runtime is the console locality
+        worker = 1,      ///< The runtime is a worker locality
+        connect = 2,     ///< The runtime is a worker locality
+                         ///< connecting late
+        local = 3,       ///< The runtime is fully local
+        default_ = 4,    ///< The runtime mode will be determined
+                         ///< based on the command line arguments
+        last
     };
+
+#if defined(HPX_HAVE_UNSCOPED_ENUM_COMPATIBILITY)
+#define HPX_RUNTIME_MODE_UNSCOPED_ENUM_DEPRECATION_MSG                         \
+    "The unscoped runtime_mode names are deprecated. Please use "              \
+    "runtime_mode::mode instead."
+
+    HPX_DEPRECATED(HPX_RUNTIME_MODE_UNSCOPED_ENUM_DEPRECATION_MSG)
+    static constexpr runtime_mode runtime_mode_invalid = runtime_mode::invalid;
+    HPX_DEPRECATED(HPX_RUNTIME_MODE_UNSCOPED_ENUM_DEPRECATION_MSG)
+    static constexpr runtime_mode runtime_mode_console = runtime_mode::console;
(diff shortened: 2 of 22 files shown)


libs/init_runtime/include/hpx/hpx_init_params.hpp
```diff
@@ -51,6 +51,17 @@ namespace hpx {
...
+    {
+        runtime_type_local,
+        runtime_type_distributed,
+#if defined(HPX_HAVE_DISTRIBUTED_RUNTIME)
+        runtime_type_best = distributed
+#else
+        runtime_type_best = local
+#endif
+    };
+
```
hkaiser, line 64: For the other enums exposed through the API we have an additional API that can return the string representation of a given enum value. Should we add that here as well? This could be used for generating error messages.

libs/init_runtime/include/hpx/hpx_init_params.hpp
```diff
@@ -51,6 +51,17 @@ namespace hpx {
     }    // namespace resource
     /// \endcond
 
+    enum runtime_type
```
hkaiser, line 54: I would like to see our enums in APIs to be slowly migrated to `enum class` types. This would also homogenize names to something like `runtime_type::local`

libs/init_runtime/include/hpx/hpx_init_params.hpp
```diff
@@ -51,6 +51,17 @@ namespace hpx {
...
+    {
+        runtime_type_local,
+        runtime_type_distributed,
+#if defined(HPX_HAVE_DISTRIBUTED_RUNTIME)
+        runtime_type_best = distributed
+#else
+        runtime_type_best = local
+#endif
```
hkaiser, line 62: Should name this `..._default` instead?

msimberg: Caught me. I had it as a scoped enum and used `best` since `default` won't work there. Then I changed it to an unscoped one to be consistent with the other enums, and so that I could use default. I missed this spot though. I'm currently lacking a good replacement for `default`.

libs/init_runtime/include/hpx/hpx_init_params.hpp
msimberg: Agreed.

libs/init_runtime/include/hpx/hpx_init_params.hpp
msimberg: That's a good idea. Will add that.

msimberg:
> Previously we discussed to extend the runtime_mode enum instead of introducing yet another enum type. Would this make sense or is it preferrable to have those separated?

I don't remember the context, but if I made that statement it may have been because I didn't understand `runtime_mode` properly. I think the two enums are mostly orthogonal, except for the fact that with `runtime_type_local` the `runtime_mode` doesn't make a difference.

libs/init_runtime/include/hpx/hpx_init_params.hpp
hkaiser: `default_`?

hkaiser: In any case, whatever we do here we should do for the other enums as well.

msimberg: Absolutely. I don't like adding underscores just to avoid name conflicts, but it may be better than  any alternative name I'll come up with. I'll sleep on it and see if I can come up with something else but otherwise it can be `default_`.

hkaiser: I wouldn't expect anybody using this value explicitly. This would be mostly for use as a default argument value on our API declarations.

hkaiser:
> > Previously we discussed to extend the runtime_mode enum instead of introducing yet another enum type. Would this make sense or is it preferrable to have those separated?
> 
> I don't remember the context, but if I made that statement it may have been because I didn't understand runtime_mode properly. I think the two enums are mostly orthogonal, except for the fact that with runtime_type_local the runtime_mode doesn't make a difference.

Playing devils advocate here... Wouldn't that mean that adding `runtime_mode::local` to the `runtime_mode` enumeration is sufficient? The existing enum values simply imply distributed operation.

libs/init_runtime/include/hpx/hpx_init_params.hpp
hkaiser: For the record, we discussed this here: #4127

hkaiser:
> Previously we discussed to extend the runtime_mode enum instead of introducing yet another enum type. Would this make sense or is it preferrable to have those separated?

For the record, we discussed this here: #4127

msimberg:
My memory... thanks @hkaiser. You're absolutely right about extending `runtime_mode`. It's much simpler to do that instead.

msimberg:
I just pushed some changes but have not tested them properly yet. I'll be back at this one...

msimberg:
If we go with class enums, should we again gradually phase out the unscoped enums? Is there a good convention for doing this kind of compatibility? I can think of two ways:

```
enum class runtime_mode {
    console = 0,
    // ...
    #if compatibility
    runtime_mode_console = console,
    // ...
    #endif
}

#if compatibility
using enum runtime_mode;
#endif
```
but this is likely to introduce name conflicts.

The other would be:
```
enum class runtime_mode {
    console = 0,
    // ...
}

#if compatibility
static constexpr runtime_mode runtime_mode_console = runtime_mode::console;
// ...
#endif
```

I guess we wouldn't break too many people's code with the runtime mode since that's almost always default or specified on the command line. The partitioner mode is another one which probably has few users. If we're going all in with this there are definitely other enums that would break user code by becoming class enums.

hkaiser:
> ```
> enum class runtime_mode {
>     console = 0,
>     // ...
> }
> 
> #if compatibility
> static constexpr runtime_mode runtime_mode_console = runtime_mode::console;
> // ...
> #endif
> ```

This is the way to go, I think.

libs/runtime_configuration/include/hpx/runtime_configuration/runtime_mode.hpp
```diff
@@ -22,8 +22,8 @@ namespace hpx {
         runtime_mode_console = 0,    ///< The runtime is the console locality
         runtime_mode_worker = 1,     ///< The runtime is a worker locality
         runtime_mode_connect = 2,    ///< The runtime is a worker locality
-                                     ///< connecting late
```
hkaiser, line 25: This comment got lost.

wrap/src/hpx_wrap.cpp
```diff
@@ -107,7 +107,6 @@ extern "C" int initialize_main(int argc, char** argv)
         hpx::init_params iparams;
         iparams.desc_cmdline = desc;
         iparams.cfg = cfg;
-        iparams.mode = hpx::runtime_mode_console;
```
hkaiser, line 110: What is the rationale of this change?

msimberg: To have the wrapping use the default runtime mode. `runtime_mode_console` tries to start the distributed runtime, but if it's disabled that will fail. I'd rather have it fall back to the local runtime just like `init` or `start` do. If the distributed runtime is enabled and the runtime mode is `runtime_mode_default` the command line handling will anyway update it to `runtime_mode_console`. There's probably a corner case I'm not aware of though. @NK-Nikunj would you happen to know why the runtime mode was explicitly set to `runtime_mode_console` here?

msimberg: To have the wrapping use the default runtime mode. `runtime_mode_console` tries to start the distributed runtime, but if it's disabled that will fail. I'd rather have it fall back to the local runtime just like `init` or `start` do. If the distributed runtime is enabled and the runtime mode is `runtime_mode_default` the command line handling will anyway update it to `runtime_mode_console`. There's probably a corner case I'm not aware of though. @NK-Nikunj would you happen to know why the runtime mode was explicitly set to `runtime_mode_console` here?

NK-Nikunj: @msimberg I explicitly set it to `runtime_mode_console` because @hkaiser asked me to take the parameters from the `init_globally` example. There was not much thought put into the config. But it may get tricky to have user defined runtime modes in the config rather than a single explicit mode.

msimberg: Thanks @NK-Nikunj! I don't think we need user defined configs in `hpx_wrap.cpp` in the sense that user code would set the runtime mode. `runtime_mode_default` will do the same without any changes from the user, but will let users pass `--hpx:connect` etc. on the command line to change the mode if desired.

msimberg:
@hkaiser I changed `runtime_mode` to be an enum class, but I added a generic compatibility option for unscoped enum compatibility. I suggest we use that one for all enums that need to be changed, and remove the option in 2.0.0. I'll try to change some other enums before 1.5.0 as well.

msimberg:
https://cdash.cscs.ch//testDetails.php?test=21992139&build=109075 seems to be unrelated (I've seen it on other PRs now as well). Seems like either the stacksize test or stacksize handling is still not quite fixed. I'm reopening #4648 as a reminder to look at it again.

PR #4649, opened 2020-05-20
Split off libhpx_core from libhpx

msimberg:
This splits the main library into two parts. This PR will only add libhpx_core (the library that will contain local concurrency and parallelism still needs #4592 and #4615).

To do:
- [x] Move libhpx_core modules to libs/core subdirectory
- [x] Update `HPX_EXPORT` macros in core modules
- [x] Check if we can avoid empty dummy source files
- [x] Update module generation scripts
- [x] Check CMake variable naming consistency
- [x] Rename libs directory to modules
- [x] Remove manually created force linking files
- [x] Check that docs still work correctly

For someone coming back to this later and doesn't want to read the whole history below: This was meant to use CMake's object libraries, but CMake configuration time exploded on Windows with that change. We'll revisit that later.


CMakeLists.txt
```diff
@@ -125,7 +125,8 @@ hpx_info("HPX version: ${HPX_VERSION}")
 unset(HPX_COMPONENTS CACHE)
 unset(HPX_EXPORT_TARGETS CACHE)
 unset(HPX_EXPORT_INTERNAL_TARGETS CACHE)
-unset(HPX_LIBS CACHE)
+unset(HPX_ENABLED_LIBS CACHE)
+unset(HPX_ENABLED_MAIN_LIBS CACHE)
```
hkaiser, line 129: `HPX_ENABLED_CORE_LIBS`?

msimberg: This falls under "check consistency". It's not actually for `hpx_core`, it's for the non-`hpx_core` part, and on top of that the variable may not be needed anymore (the `add_hpx_module` at least in the current iteration adds the modules to the correct library).

msimberg:
I have to mark this as ready to review just to get the GitHub actions builders running. It's not ready to review otherwise.

msimberg:
This is getting close to usable now.

Apologies for the massive list of changed files. For some reason git did not track me moving the modules from one directory to another. (If someone knows how to tell git that I've just moved files, let me know. Squashing might help?). 

The modules are now added with an extra argument: the library that they are part of, and they need to specify which dependencies are other modules. So adding a module might look like `add_hpx_module(core serialization ... MODULE_DEPENDENCIES hpx_config)`.

Modules in the `full` directory belong to `libhpx`. I'm open to suggestions on the name, but it is just an internal name, i.e. the library does not actually have `full` in its name. The modules in `full` also now depend only on `hpx_core` instead of all the individual `hpx_config`, `hpx_preprocessor` etc. modules.

The `MODULE_DEPENDENCIES` option is needed to be able to distinguish "normal" dependencies and module dependencies, since the latter need some special handling wrt to static libraries.

A target `hpx_modulename` is now an object library, whereas `hpx_modulename_static` is a static library (tests that only want certain modules link to the `_static` modules).

There is also a `hpx_core_static` library now. There is no `hpx_full_static` library, but that could easily be added (this worried me a bit: https://github.com/STEllAR-GROUP/hpx/blob/ff74f285ccc0d2bcd98e727ba46d3b3c24a50b19/src/CMakeLists.txt#L488-L492). Might we get around that restriction now that the library is split up?

hkaiser:
The first thing I noticed when trying this out was that cmake now takes an awful long time to generate the build files (on Windows, have not tested any other platform). It actually takes longer to generate the build system (several minutes) than it takes to actually build the core libraries...

msimberg:
> The first thing I noticed when trying this out was that cmake now takes an awful long time to generate the build files (on Windows, have not tested any other platform). It actually takes longer to generate the build system (several minutes) than it takes to actually build the core libraries...

Hmm, thanks for trying it out. That would be quite a horrible regression. Do you mean that only the final generate step takes longer than usual or the whole cmake invocation? How long does generating the build system files take on master? I haven't noticed a regression like that but I have not timed it. I would actually have hoped that this would slightly (maybe not noticeably) even reduce generation times because we don't need to do the force linking files. I'll check how long generation takes on Linux.

hkaiser:
> Hmm, thanks for trying it out. That would be quite a horrible regression. Do you mean that only the final generate step takes longer than usual or the whole cmake invocation? How long does generating the build system files take on master? I haven't noticed a regression like that but I have not timed it. I would actually have hoped that this would slightly (maybe not noticeably) even reduce generation times because we don't need to do the force linking files. I'll check how long generation takes on Linux.

The generation step is much slower (almost 20 minutes for me). It's most probably a cmake problem, but I'm not sure we can explain that to users. Generating build files on master takes less than a minute for me.

hkaiser:
Another observation (probably a left-over): `libs/module.cpp` is still being generated, even if not used anymore.

cmake/HPX_AddModule.cmake
```diff
@@ -204,123 +201,112 @@ function(add_hpx_module name)
...
+    NAME hpx_{modulename}
     ROOT ${CMAKE_CURRENT_BINARY_DIR}/include/hpx
     CLASS "Generated Files"
     TARGETS ${config_header}
   )
 
   set_target_properties(
-    hpx_${name} PROPERTIES FOLDER "Core/Modules" POSITION_INDEPENDENT_CODE ON
+    hpx_${modulename} PROPERTIES FOLDER "Core/Modules" POSITION_INDEPENDENT_CODE
+                                                       ON
```
hkaiser, line 292: Please add a `FOLDER` property for the static libraries, possibly combining the shared and static libraries. For instance:
```
  set_target_properties(
    hpx_${modulename} PROPERTIES FOLDER "Core/Modules/${modulename}"
  )
  set_target_properties(
    hpx_${modulename}_static PROPERTIES FOLDER "Core/Modules/${modulename}"
  )
```

hkaiser: Possibly even `"Core/Modules/Core/${modulename}"` and `"Core/Modules/Full/${modulename}"` as appropriate.

hkaiser:
Collecting observations still:
- Some of the module projects (the but only the `OBJECT` libraries) refer to more than one `config_registry.obj` file (they refer to their own `config_registry.obj` as well as to the corresponding files from their direct dependencies). I'm not sure why that's happening and it could very well be a cmake issue. I'm just mentioning it here to make sure it's not caused by our scripts.
- I'm seeing linker errors that `switch_to_fiber` is undefined (for the tagret `hpx_core`). The corresponding object file is part of the target `hpx_full`, however. Should it be added to both? This could also be caused by the fact that the file `switch_to_fiber.asm` was moved but the CMakeLists.txt referring to it was not correctly updated (see: https://github.com/STEllAR-GROUP/hpx/blob/master/src/CMakeLists.txt#L117)

hkaiser:
FWIW, I'm attaching my patch here for simplicity:
```
diff --git a/cmake/HPX_AddModule.cmake b/cmake/HPX_AddModule.cmake
index 93ea4a954e3..5f105b4baa7 100644
--- a/cmake/HPX_AddModule.cmake
+++ b/cmake/HPX_AddModule.cmake
@@ -287,15 +287,26 @@ function(add_hpx_module libname modulename)
     TARGETS ${config_header}
   )
 
+  if(${libname} STREQUAL "full")
+    set(subfolder "Full")
+  else()
+    set(subfolder "Core")
+  endif()
   set_target_properties(
-    hpx_${modulename} PROPERTIES FOLDER "Core/Modules" POSITION_INDEPENDENT_CODE
-                                                       ON
+    hpx_${modulename}
+    PROPERTIES FOLDER "Core/Modules/${subfolder}/${modulename}"
+               POSITION_INDEPENDENT_CODE ON
   )
 
   # Add a static library that only contains the object library.
   add_library(hpx_${modulename}_static $<TARGET_OBJECTS:hpx_${modulename}>)
   target_link_libraries(hpx_${modulename}_static PUBLIC hpx_${modulename})
 
+  set_target_properties(
+    hpx_${modulename}_static
+    PROPERTIES FOLDER "Core/Modules/${subfolder}/${modulename}"
+  )
+
   foreach(lib ${${modulename}_MODULE_DEPENDENCIES})
     target_link_libraries(hpx_${modulename}_static PUBLIC ${lib}_static)
   endforeach()
diff --git a/modules/CMakeLists.txt b/modules/CMakeLists.txt
index db993eda901..fce2288436c 100644
--- a/modules/CMakeLists.txt
+++ b/modules/CMakeLists.txt
@@ -63,6 +63,9 @@ foreach(lib ${_hpx_libs})
     add_library(hpx_${lib} SHARED src/dummy.cpp)
     add_library(hpx_${lib}_static STATIC src/dummy.cpp)
 
+    set_target_properties(hpx_${lib} PROPERTIES FOLDER "Core")
+    set_target_properties(hpx_${lib}_static PROPERTIES FOLDER "Core")
+
     install(
       TARGETS hpx_${lib}
       EXPORT HPXInternalTargets
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 2cd10140496..c7f77f39aa2 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -112,7 +112,7 @@ if(MSVC)
   if(HPX_WITH_SWAP_CONTEXT_EMULATION)
     enable_language(ASM)
     set(switch_to_fiber_source
-        "${PROJECT_SOURCE_DIR}/modules/coroutines/src/switch_to_fiber.asm"
+        "${PROJECT_SOURCE_DIR}/modules/core/coroutines/src/switch_to_fiber.asm"
     )
     list(APPEND hpx_external_objects_SOURCES "${switch_to_fiber_source}")
     add_hpx_library_sources(
```

msimberg:
> Collecting observations still:
> 
>     * Some of the module projects (the but only the `OBJECT` libraries) refer to more than one `config_entry.obj` file (they refer to their own `config_entry.obj` as well as to the corresponding files from their direct dependencies). I'm not sure why that's happening and it could very well be a cmake issue. I'm just mentioning it here to make sure it's not caused by our scripts.

Excellent catch. Modules in `hpx_core` should depend on `hpx_config_registry` whereas other modules should just depend on `hpx_core` and get the dependency that way. Will fix.

>     * I'm seeing linker errors that `switch_to_fiber` is undefined (for the tagret `hpx_core`). The corresponding object file is part of the target `hpx_full`, however. Should it be added to both? This could also be caused by the fact that the file `switch_to_fiber.asm` was moved but the CMakeLists.txt referring to it was not correctly updated (see: https://github.com/STEllAR-GROUP/hpx/blob/master/src/CMakeLists.txt#L117)
>       Update: After fixing the path to `switch_to_fiber.asm` everything builds fine.

The handling of that file should probably just move to the coroutines module.

>     * Please add a `FOLDER` property to the targets `hpx_core` and `hpx_core_static`:
>       ```
>       set_target_properties(hpx_core PROPERTIES FOLDER "Core")
>       set_target_properties(hpx_core_static PROPERTIES FOLDER "Core")
>       ```

Will do!

Thanks @hkaiser for going through this!

msimberg:
> The generation step is much slower (more than 20 minutes for me). It's most probably a cmake problem, but I'm not sure we can explain that to users. Generating build files on master takes less than a minute for me.

I went through a small sample of GitHub actions builds and there's definitely a regression here (none as spectacular as yours, but enough to be able to look into it). From my small absolutely random sample the CMake configuration step takes about 1.5-2 minutes on Windows on master and 2-2.5 minutes on this branch. On Linux iti's 20-30 seconds on master and 30-40 seconds on this branch.

It's curious that you would have a 10x slowdown though. I think the GitHub actions builders use CMake 3.17.2. Which version are you using?

hkaiser:
> It's curious that you would have a 10x slowdown though. I think the GitHub actions builders use CMake 3.17.2. Which version are you using?

I need to check with the newest cmake still (I'm using 3.15.x ATM). But I think the difference comes from having all tests enabled. I have the impression that those are contributing quite a bit to the overall time.

msimberg:
Quick update: pkgconfig needed some special handling again... It looks like CMake treats object library dependencies differently in different versions. In at least 3.14 and older it puts something like `hpx_algorithms::@<0x012345678>` in `INTERFACE_LINK_LIBRARIES` which CMake knows how to deal with itself, but is slightly annoying for generating the pkgconfig files. I have a hack that seems to work on the CMake versions that I've tested but no guarantees that it'll work everywhere.

Since the change to object libraries is in the end mostly orthogonal to the main changes here I may still drop that change and stick to the old static libraries. Especially if we can't figure out why configuration is taking longer (I don't have clues yet).

hkaiser:
> Since the change to object libraries is in the end mostly orthogonal to the main changes here I may still drop that change and stick to the old static libraries. Especially if we can't figure out why configuration is taking longer (I don't have clues yet).

I agree, it might be a good idea to separate these changes.

msimberg:
> I agree, it might be a good idea to separate these changes.

It's what I should've done in the first place... at least we know what we're getting.

msimberg:
@hkaiser this is back to the cleanup phase, i.e. things are mostly working but still in progress. Do you think you could check just that the configuration time has actually gone back to normal for you on Windows? The GitHub actions builds seem to have gone back down to the 1.5-2 minutes range which is the same as on master, so I think it's "fixed".

msimberg:
Besides linking on Windows (and possible linting) this is again ready. If `/WHOLEARCHIVE` ends up working I'll remove the force linking facilities again. The diff is unfortunately still massive. Redoing the commits made git realize that most of the files have been renamed, but it still thinks some have been deleted and removed.

hkaiser:
> Besides linking on Windows (and possible linting) this is again ready. If /WHOLEARCHIVE ends up working I'll remove the force linking facilities again. The diff is unfortunately still massive. Redoing the commits made git realize that most of the files have been renamed, but it still thinks some have been deleted and removed.

Here is a small patch that makes `/WHOLEARCHIVE` work (with an additional cosmetic change to the generated `FOLDER`):
```diff
diff --git a/cmake/HPX_AddModule.cmake b/cmake/HPX_AddModule.cmake
index f3f8a49808d..540d1db4050 100644
--- a/cmake/HPX_AddModule.cmake
+++ b/cmake/HPX_AddModule.cmake
@@ -341,8 +341,13 @@ function(add_hpx_module libname modulename)
     TARGETS ${config_header}
   )
 
+  # capitalize string
+  string(SUBSTRING ${libname} 0 1 first_letter)
+  string(TOUPPER ${first_letter} first_letter)
+  string(REGEX REPLACE "^.(.*)" "${first_letter}\\1" libname_cap "${libname}")
+
   set_target_properties(
-    hpx_${modulename} PROPERTIES FOLDER "Core/${libname}/${modulename}"
+    hpx_${modulename} PROPERTIES FOLDER "Core/Modules/${libname_cap}"
                                  POSITION_INDEPENDENT_CODE ON
   )
 
@@ -423,7 +428,7 @@ function(add_hpx_module libname modulename)
   elseif(MSVC)
     target_link_options(
       hpx_${libname} PRIVATE
-      "/WHOLEARCHIVE:\"$<TARGET_FILE:hpx_${modulename}>\""
+      /WHOLEARCHIVE:$<TARGET_FILE:hpx_${modulename}>
     )
   endif()
 
```

msimberg:
I broke the CI check for modules in an earlier PR. @aurianer fixed that and updated the script to the new module structure, which showed a couple of missing dependencies in the CMake files. There are two dependencies that this PR will *not* solve, but I think I've sorted them out in #4606, so I'd like this one to go in first. The spellcheck also found some things which I will not fix (nor whitelist).

Ignoring those I think this is now ready. @hkaiser I did in the end have to change `/WHOLEARCHIVE` to `-WHOLEARCHIVE` as you suspected to get it work on the GitHub actions builders. I guess that will work on your setup as well?

msimberg:
Note: It should be possible to have a library with the local concurrency and parallelism now, but I'll do that in a separate PR. We'll do some renaming in between (the conflicts will be spectacular otherwise).

hkaiser:
> I did in the end have to change /WHOLEARCHIVE to -WHOLEARCHIVE as you suspected to get it work on the GitHub actions builders. I guess that will work on your setup as well?

That should be fine, thanks!

msimberg:
Uhm, is this real: https://github.com/STEllAR-GROUP/hpx/pull/4649/checks?check_run_id=712935935 or GitHub's MSVC compiler being funny? It refers to this: https://github.com/STEllAR-GROUP/hpx/blob/7a07598fbfcb4a871d16d568e31dbd211dc74d4f/libs/pack_traversal/include/hpx/pack_traversal/traits/pack_traversal_rebind_container.hpp#L65-L69 saying `a class cannot be its own base class`. Without knowing anything about the code, should it inherit `detail::pack_traversal_rebind_container` instead?

hkaiser:
> Uhm, is this real: https://github.com/STEllAR-GROUP/hpx/pull/4649/checks?check_run_id=712935935 or GitHub's MSVC compiler being funny? It refers to this: hpx/libs/pack_traversal/include/hpx/pack_traversal/traits/pack_traversal_rebind_container.hpp Lines 65 to 69 in 7a07598
>   ```
>            template <typename NewType, typename OldType, typename Enable = void>           
>            struct pack_traversal_rebind_container 
>              : pack_traversal_rebind_container<NewType, OldType> 
>            { 
>            }; 
> ```
>  saying a class cannot be its own base class. Without knowing anything about the code, should it inherit detail::pack_traversal_rebind_container instead?

Yes, no idea how it ever worked... Thanks!

msimberg:
> Yes, no idea how it ever worked... Thanks!

I'm glad something complained about it! I wonder what triggered it now though... I pushed the change here. If this PR won't go in soon I'll open a separate PR to fix it.

msimberg:
All right, the infrastructure is ready in this one now, but since this will cause conflicts with everything I'll wait with this until we're a bit closer to the release and there's not so much renaming, moving, etc. going on. Recreating the modules directory is not a lot of work.

msimberg:
https://github.com/STEllAR-GROUP/hpx/pull/4682 is now up to fix the pack traversal compilation error.

msimberg:
All right, this is now compiling locally and seems to pass CI as well. A short summary again:

- `libhpx` is split into `libhpx`, `libhpx_parallelism`, and `libhpx_core`
- There's currently no difference for users, they still link to `libhpx`/`HPX::hpx` and get the two others linked as a dependency; user-facing changes will be handled separately
- Modules are added with `add_hpx_module(libname modulename ...)`; the library name can be `full` (for `libhpx`), `parallelism`, or `core`, and the module name as before
- Module dependencies are handled specially: dependencies within the library go in `MODULE_DEPENDENCIES` and the rest go in `DEPENDENCIES` (the split is not strictly necessary right now, but make using object libraries in the future a bit easier)
- The rest of the changes are mostly related to visibility

For anyone daring to look at the diff, I recommend you do it locally with this in `gitconfig`:
```
[diff]
renameLimit = 10000
```

@biddisco, @weilewei I volunteer to bring #4909 up to date if this goes in soon. Offer extends to anyone else wanting help with moving files around as well.

hkaiser (approved):
I admit, I have not looked closely. I would suggest to merge this asap to avoid further conflicts and rebases and deal with any possible problems as we go. Thanks for looking into this @msimberg!

msimberg:
> I admit, I have not looked closely. I would suggest to merge this asap to avoid further conflicts and rebases and deal with any possible problems as we go. Thanks for looking into this @msimberg!

Thanks, I'll go ahead and merge this now in that case. Please let me know if you notice anything weird related to this. I'm fairly confident about the changes on Linux, but not quite sure about everything on Windows obviously (except what CI checks). The changes you suggested earlier are there though, and CMake configuration times look unchanged compared to master, so hopefully this won't be too disruptive for everyone.

msimberg:
A little help for updating PRs that have added files: I've used `rsync` for moving files left in the top-level `libs` directory one step down into the hierarchy:

```
rsync -avr --remove-source-files ${HPX_ROOT}/libs/modulename/* ${HPX_ROOT}/libs/libname/modulename/
```

PR #4651, opened 2020-05-20
Update get_function_annotation.hpp

### LINKED ISSUE #4650: thread descriptions are broken...again
## Expected Behavior

The `util::thread_description` constructor uses the `traits::get_function_annotation<F>::call(f);` value to get an annotation, if it exists.  The default behavior used to be to return a `nullptr`, which would then be overridden by the `altname` value, if there was one.

## Actual Behavior

Unfortunately, the new behavior is to return the string `"<anonymous>"`.  Which then overrides whatever annotation is passed in to the `util::thread_description` constructor as `altname`.  See these lines:
https://github.com/STEllAR-GROUP/hpx/blob/9b9392a24c3406237fae4f04c68a2081c1e04ae8/libs/threading_base/include/hpx/threading_base/thread_description.hpp#L85-L89
and
https://github.com/STEllAR-GROUP/hpx/blob/9b9392a24c3406237fae4f04c68a2081c1e04ae8/libs/functional/include/hpx/functional/traits/get_function_annotation.hpp#L20-L24

I tried changing `traits::get_function_annotation<F>::call(f);` to return `nullptr`, but other code now depends on a valid string value. ☹️ 

I can hack a check for `"<anonymous>"`, but that doesn't "make the common case fast". I am open to other ideas...

  1. Configure HPX with APEX support
  1. `export APEX_SCREEN_OUTPUT=1` to see the APEX measurements after execution
  1. Run any example, but examples.quickstart.fibonacci_futures_distributed will reproduce it.

## Specifications

  - Platform (compiler, OS):  gcc 7.3, RHEL linux

khuck:
Fixes #4650

## Proposed Changes

  - return nullptr for anonymous tasks so that annotation can annotate them

### DIFF
--- a/libs/functional/include/hpx/functional/traits/get_function_annotation.hpp
+++ b/libs/functional/include/hpx/functional/traits/get_function_annotation.hpp
@@ -19,7 +19,7 @@ namespace hpx { namespace traits {
     {
         static char const* call(F const& /*f*/) noexcept
         {
-            return "<anonymous>";
+            return nullptr;
         }
     };
 


khuck:
Are the test failures related?

hkaiser:
In the end I have to take responsibility for the original change that caused this problem. I asked @biddisco to ensure some sensible fallback name is used. Thanks for figuring out what the problem was, @khuck!

hkaiser:
> Are the test failures related?

The test failures are unrelated. I'll merge this now.

PR #4653, opened 2020-05-21
Making sure APEX is linked into every application, if needed

hkaiser:
- this is a Windows specific change

This makes APEX usable on windows (to some minimal extend), relies on https://github.com/khuck/xpress-apex/pull/122 being merged.

### DIFF
--- a/hpx/hpx_init_impl.hpp
+++ b/hpx/hpx_init_impl.hpp
@@ -34,6 +34,14 @@ extern HPX_EXPORT char** freebsd_environ;
 extern char** environ;
 #endif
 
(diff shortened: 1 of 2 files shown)


hkaiser:
> Does hpx_main.hpp (or hpx_main_impl.hpp) also need to reference APEX?

I think we don't need to modify those as they relate to the wrapping magic which is not used on Windows anyways.

msimberg:
> > Does hpx_main.hpp (or hpx_main_impl.hpp) also need to reference APEX?
> 
> I think we don't need to modify those as they relate to the wrapping magic which is not used on Windows anyways.

Good point! Thanks!

PR #4655, opened 2020-05-21
Fixing task annotations for actions

### LINKED ISSUE #4652: How to fix distributed action annotation
## Expected Behavior

In the past, distributed actions were annotated correctly.  Now, all localities other than 0 get just `<anonymous task>` tasks.

## Actual Behavior

I think the fix (or at least the problem) is in here:
https://github.com/STEllAR-GROUP/hpx/blob/9b9392a24c3406237fae4f04c68a2081c1e04ae8/libs/async/include/hpx/async/applier/apply.hpp#L388-L409

I am not sure how using a default `threads::thread_init_data`is supposed to do anything useful - it just creates a thread_description with the name `<anonymous task>`, from the default constructor.  The actual action is in the Action and/or Continuation objects, from what I can see in gdb - but I have no idea how to get a valid thread_description out of those objects.

  1.  Configure HPX with APEX
  1.  `export APEX_CSV_OUTPUT=1` before running
  1.  Run a distributed test like:
```
mpirun -np 2 ./bin/fibonacci_futures_distributed --n-value 10 -t 1 -Ihpx.stacks.use_guard_pages=0 --distribute-at 9
```
  1. Then see the apex.1.csv file and the `<anonymous task>` timers should have been `fibonacci_future_action` instead.

## Specifications

  - Platform (compiler, OS):  gcc 7.3 on RHEL

hkaiser:
- flyby: simplify hello_world_distributed

Fixes #4652

### DIFF
--- a/examples/quickstart/hello_world_distributed.cpp
+++ b/examples/quickstart/hello_world_distributed.cpp
@@ -95,8 +95,6 @@ void hello_world_foreman()
             // thread, but no guarantees are given by the scheduler that the
             // task will actually run on that worker thread.
             hpx::parallel::execution::default_executor exec(
-                hpx::threads::thread_priority_default,
-                hpx::threads::thread_stacksize_default,
                 hpx::threads::thread_schedule_hint(
                     hpx::threads::thread_schedule_hint_mode_thread, worker));
             futures.push_back(hpx::async(exec, hello_world_worker, worker));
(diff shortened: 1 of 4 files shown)


hkaiser:
@khuck could you please check whether this fixes the issue for you?

khuck:
testing it now...

khuck:
> @khuck could you please check whether this fixes the issue for you?

Yes, works great!  Thanks!

khuck:
Looks like I have some labeling issues in APEX, but the profiling is working...

PR #4663, opened 2020-05-25
Mpi fixes

### LINKED ISSUE #4662: MPI initialization broken when networking off
Unfortunately, changes were made in PR #4540 to initialization, that broke MPI when networking was enabled, then they were broken again in #4644 for applications that do not use networking. Attempting to run MPI enabled code that uses HPX results in
```
hpx::init: exception caught: Invalid runtime mode specified
```
on all ranks != 0

Unfortunately #4540 was merged before it was ready (and broke local changes I was working on for initialization). #4644 was not tested properly before being merged. I still have a number of changes I would like to make to the MPI support and I would appreciate if others would synchronize with me before merging any more MPI related fixes.

biddisco:
Fixes #4662 and the error mentioned in the comments of #4540 

## Proposed Changes
These are assorted patches that fix some small inconsistencies in the MPI futures layer.

This is being pushed to prevent duplicated effort on the MPI backend, but should not be merged yet as there may be some more patches applied (subject to further testing).


libs/mpi/tests/unit/mpi_ring_async.cpp
```diff
@@ -155,9 +155,10 @@ int main(int argc, char* argv[])
...
     }
-    return hpx::init(argc, argv);
-    //
+    int result = hpx::init(argc, argv);
+
     // Finalize MPI
-    //
     MPI_Finalize();
+
+    return result;
```
hkaiser, line 163: Good catch, thanks!

libs/mpi/src/mpi_future.cpp
```diff
@@ -97,13 +97,9 @@ namespace hpx { namespace mpi { namespace experimental {
...
-            // for debugging only
-            if (mpi_debug.is_enabled())
-            {
-                mpi_debug.debug(debug::str<>("request queued"), get_mpi_info(),
-                    "request", debug::hex<8>(data->request_), "active futures",
-                    debug::dec<3>(get_active_futures().size()));
-            }
+            mpi_debug.debug(debug::str<>("request queued"), get_mpi_info(),
+                "request", debug::hex<8>(data->request_), "active futures",
+                debug::dec<3>(get_active_futures().size()));
```
hkaiser, line 102: I initially added the conditional as otherwise the arguments to the `.debug()` function would be evaluated even if the output was disabled. I assume you rely on the compiler to optimize away everything. Is that a 100% safe assumption?

libs/mpi/tests/unit/mpi_ring_async_executor.cpp
```diff
@@ -94,7 +94,7 @@ int hpx_main(hpx::program_options::variables_map& vm)
                 {
                     // send the incremented token to the next rank
                     ++tokens[i];
-                    hpx::future<void> f_send = hpx::async(
+                    hpx::future<int> f_send = hpx::async(
```
hkaiser, line 97: The return value is never used, why should it be explicit?

libs/mpi_base/src/mpi_environment.cpp
```diff
@@ -193,7 +193,12 @@ namespace hpx { namespace util {
         }
         else
         {
+#if defined(HPX_HAVE_NETWORKING)
             rtcfg.mode_ = hpx::runtime_mode_worker;
+#else
+            // when networking is off, all ranks are equivalent
+            rtcfg.mode_ = hpx::runtime_mode_console;
+#endif
```
hkaiser, line 201: Good catch!

libs/mpi/tests/unit/mpi_ring_async_executor.cpp
hkaiser: As a side note: any exceptions possibly thrown from the `hpx::async` are not propagated by this code and are silently swallowed. I have missed this when working on the code as well. Would you mind adding the proper exception propagation while you are at it?

biddisco: I added the return value for completeness - though we don't use it, but I could see someone complaining that if the call to mpi_isend failed (for whatever reason) then they should be able to retrieve the return value. I have not done exception handling yet (I presume you mean for mpi_calls that fail) and it's not high on my list. I will add it at some point, but I want to concentrate on fixing app related issues first.

hkaiser: I was referring to any exception that might be thrown by `hpx::async`. All of those should end up in the returned future, but the code never calls `.get()` on it, so the exception will not rethrown.

libs/mpi/tests/unit/mpi_ring_async_executor.cpp
```diff
@@ -74,63 +89,84 @@ int hpx_main(hpx::program_options::variables_map& vm)
...
+                tokens[i] = (rank == 0) ? 1 : -1;
+                int rank_from = (size + rank - 1) % size;
+                int rank_to = (rank + 1) % size;
+
+                // all ranks pre-post a receive
+                hpx::future<int> f_recv = hpx::async(
+                    exec, MPI_Irecv, &tokens[i], 1, MPI_INT, rank_from, i);
+
+                // when the recv completes,
+                f_recv.then([=, &tokens, &counter](auto&&) {
```
hkaiser, line 117: Discarding the returned future counters all of our paradigms. Should we adapt this test to follow the usual 'protocol'?

libs/mpi/tests/unit/mpi_ring_async_executor.cpp
```diff
@@ -74,63 +89,84 @@ int hpx_main(hpx::program_options::variables_map& vm)
...
         }
 
-        // Our simple counter should reach zero when all send/recv pairs are done
-        hpx::mpi::experimental::wait([&]() { return counter != 0; });
-
-        // let the user polling go out of scope
-    }
+        // This is needed to make sure that one rank does not shut down
+        // before others have completed. MPI does not handle that well.
+        MPI_Barrier(MPI_COMM_WORLD);
```
hkaiser, line 161: This could probably go away if we handled the futures returned from the `.then()` above explicitly.

biddisco: This is really there for 2 reasons :
Users will look at our unit tests as a base to develop their own codes - adding a barrier before shutting down the runtime is good practice and I put it here for that reason - but also because, if rank 0 has received all its messages, sent all its sends and is "finished" all its futures will be ready and it can complete and exit - but if it tries to exit whilst other ranks are still doing exchanges, then 99% of the time there will be a problem. Adding this barrier is there to guarantee this does not happen and show users the best practice.

libs/mpi/tests/unit/mpi_ring_async_executor.cpp
biddisco: what do you mean by the 'usual' protocol?

libs/mpi/src/mpi_future.cpp
biddisco: I already answered this comment, but github seems to have lost it - the conditional should not be necessary because the debug statement is a no-op when disabled. My experiments with godbolt indicate that all the code is removed, however I will conduct more tests just to reassure myself that no side effects are possible. A small unit test that flexes the debug::print code might prove useful to ensure that functions used in disabled call parameters are never invoked...

libs/mpi/tests/unit/mpi_ring_async_executor.cpp
biddisco: I want to make the code exception safe, but I'd rather do this in another PR as I am in a hurry to get more functionality working on other branches that are more important than this. I'll create an issue for it once this PR is merged (if ever).

libs/mpi/tests/unit/mpi_ring_async_executor.cpp
hkaiser: I would consider not discarding any futures returned from our APIs being good practice.

libs/mpi/tests/unit/mpi_ring_async_executor.cpp
hkaiser: Would having an MPI barrier in the HPX code right before MPI_finalize is called be sufficient? I would not consider using MPI_Barrier in user code as 'good practice'. It makes me cringe when I see it.

biddisco: Adding a barrier before finalize would certainly help, but only if the user allowed hpx to do the init. If they handle init/finalize themselves then it won't help. I do agree that barriers are bad, but a barrier at shutdown is not as evil as one after every iteration - and lets nioot forget that inside hps there are things like `hpx::big_boot_barrier` that serve exactly the same purpose as this except at startup instead of shutdown. I'd prefer to leave it in.

hkaiser: If the users do MPI_init/MPI_finalize themselves we can expect them to handle the barrier in the end anyways as they will do other things requiring to do that anyways. I'd vote for adding an MPI_Barrier before we call MPI_finalize and remove it here.

hkaiser: > and lets not forget that inside hpx there are things like hpx::big_boot_barrier that serve exactly the same purpose

Yes. However those are inside HPX and not in application code. The same would be true for adding an MPI_barrier before calling MPI_finalize. I have no problem if the user decides to add a barrier in their code wherever they want. Mostly I would like to avoid us requiring for the user to do that on our behalf.

biddisco: This is an application example and not part of hpx internals, the example calls mpi_init explicitly on start (as most applications will and does not use any hpx mpi initialization). The MPI_Barrier is not part of any hpx requirement. This test does work without the barrier (does not crash), because I've made a point of adding a wait until all requests are complete - but if the user does not wait on all requests or futures as we do with our counter, then it will probably crash. I am happy to comment out the barrier, but with notes to say that not waiting for all futures to complete is very bad (especially if they mix normal mpi calls with futurized ones). This is supposed to be an example to show users how to safely use the mpi integration and if we don't mention it, they will have crashes and we'll just get people filing issues and spamming IRC. I will also add a note to the docs about it.

hkaiser: Ok, then let's move the barrier into `main()` (after `hpx::init` returned) to disentangle the HPX part from the MPI code.

However, I still think that dropping the futures on the floor is a bad thing, but this might be for a separate PR.

biddisco:
@weilewei please try this latest version of the mpi branch (rebased onto master today) and tell let me know if this fixes #4752 #4662 etcetc. I still have more work to do on this branch, but if this works and passes tests, then I can continue work on a new branch and let this one me merged.

libs/mpi_base/src/mpi_environment.cpp
```diff
@@ -187,6 +197,10 @@ namespace hpx { namespace util {
 
         this_rank = rank();
 
+#if !defined(HPX_HAVE_NETWORKING)
+        // all ranks are just local nodes when networking is off
+        rtcfg.mode_ = hpx::runtime_mode::local;
+#else
```
msimberg, line 203: I don't think this will do what one would expect. This will start only the local runtime. I think actions and components will still need the AGAS etc. stubs even if they don't do anything. Not sure however... I would leave this part unchanged.

libs/command_line_handling/src/command_line_handling.cpp
```diff
@@ -790,7 +792,11 @@ namespace hpx { namespace util {
         }
 #endif
 
+#ifdef HPX_HAVE_LIB_MPI_BASE
         bool have_mpi = util::mpi_environment::check_mpi_environment(rtcfg_);
+#else
+        bool have_mpi = false;
+#endif
```
msimberg, line 799: Is `mpi_base` disabled but the MPI parcelport on a valid configuration which should also call `check_mpi_environment`?

libs/collectives/tests/unit/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ set(tests
     global_spmd_block
 )
 
-if(HPX_HAVE_NETWORKING)
+if(HPX_WITH_NETWORKING)
```
msimberg, line 22: Very good catch!

libs/mpi_base/src/mpi_environment.cpp
msimberg: I rearranged this a bit so that `local` is only used when the distributed runtime is off, `console` when networking is off but distributed runtime is on, and the old code when networking is on.

libs/command_line_handling/src/command_line_handling.cpp
msimberg: I changed the `#ifdef`. @biddisco please check if this makes sense to you.

hkaiser: This should also check whether the MPI parcelport is enabled. I even think that this should always be checked as it will return `false` if MPI is not available anyways.

msimberg: Good point. I removed the ifdefs completely from here now, as they just duplicated ifdefs that were inside `check_mpi_environment`.

biddisco:
The problem is that if you turn off networking, distributed_runtime and don't use async_mpi then mpi_base is not loaded
```
[ 75%] Building CXX object libs/command_line_handling/CMakeFiles/hpx_command_line_handling.dir/src/command_line_handling.cpp.o
/apps/daint/UES/biddisco/src/hpx/libs/command_line_handling/src/command_line_handling.cpp: In member function 'bool hpx::util::command_line_handling::handle_arguments(hpx::util::manage_config&, boost::program_options::variables_map&, std::vector<std::__cxx11::basic_string<char> >&, std::size_t&, bool)':
/apps/daint/UES/biddisco/src/hpx/libs/command_line_handling/src/command_line_handling.cpp:795:31: error: 'hpx::util::mpi_environment' has not been declared
         bool have_mpi = util::mpi_environment::check_mpi_environment(rtcfg_);
                               ^~~~~~~~~~~~~~~
make[3]: *** [libs/command_line_handling/CMakeFiles/hpx_command_line_handling.dir/build.make:63: libs/command_line_handling/CMakeFiles/hpx_command_line_handling.dir/src/command_line_handling.cpp.o] Error 1
make[2]: *** [CMakeFiles/Makefile2:21331: libs/command_line_handling/CMakeFiles/hpx_command_line_handling.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:61919: tests/performance/local/CMakeFiles/future_overhead_test.dir/rule] Error 2
```
you need to protect the call to mpi_environment

msimberg:
> The problem is that if you turn off networking, distributed_runtime and don't use async_mpi then mpi_base is not loaded
> 
> ```
> [ 75%] Building CXX object libs/command_line_handling/CMakeFiles/hpx_command_line_handling.dir/src/command_line_handling.cpp.o
> /apps/daint/UES/biddisco/src/hpx/libs/command_line_handling/src/command_line_handling.cpp: In member function 'bool hpx::util::command_line_handling::handle_arguments(hpx::util::manage_config&, boost::program_options::variables_map&, std::vector<std::__cxx11::basic_string<char> >&, std::size_t&, bool)':
> /apps/daint/UES/biddisco/src/hpx/libs/command_line_handling/src/command_line_handling.cpp:795:31: error: 'hpx::util::mpi_environment' has not been declared
>          bool have_mpi = util::mpi_environment::check_mpi_environment(rtcfg_);
>                                ^~~~~~~~~~~~~~~
> make[3]: *** [libs/command_line_handling/CMakeFiles/hpx_command_line_handling.dir/build.make:63: libs/command_line_handling/CMakeFiles/hpx_command_line_handling.dir/src/command_line_handling.cpp.o] Error 1
> make[2]: *** [CMakeFiles/Makefile2:21331: libs/command_line_handling/CMakeFiles/hpx_command_line_handling.dir/all] Error 2
> make[1]: *** [CMakeFiles/Makefile2:61919: tests/performance/local/CMakeFiles/future_overhead_test.dir/rule] Error 2
> ```
> 
> you need to protect the call to mpi_environment

Thanks @biddisco! Excellent point. I had somehow looked that `check_mpi_environment` was defined in `command_line_handling` but this makes much more sense. I'll change that back.

msimberg:
retest

libs/mpi/tests/unit/mpi_ring_async_executor.cpp
msimberg: @biddisco what's the plan here now? Do we keep things as they are or are you planning to change this still?

biddisco: I tried an experiment where I collected all the futures and waited for them properly instead of using the atomic counter, but it ran about 5% more slowly than the existing version. I would prefer to leave this as is for now and get the PR merged, then if someone wants to clean up the future handling at some point they can. 
If we had an `apply` version of `.then`, we would not need to drop any but could instead decrement counters in continuations that don't return a future.
```
f1 = hpx::async(some_work);
f2 = f1.then(other_work); // ok
// we don't have
f2.then_apply(no_future_returned_from_this);
```

PR #4642, opened 2020-05-19
Implementing operation_state concept in terms of P0443r13

sithhell:
Implementing operation_state concept in terms of P0443r13
    
One deviation of P0443r13 is the usage of tag_invoke as customization
point

This PR depends on #4640

PR #4665, opened 2020-05-25
Implementing sender, receiver, and operation_state concepts in terms of P0443r13

sithhell:
One deviation of P0443r13 is the usage of tag_invoke as customization
point

Depends on #4642


msimberg:
@sithhell could you update the formatting on this one before we merge it? Since the itt module is merged a rebase/merge from master would be nice as well so that we can see that the circular dependency is really gone.

libs/basic_execution/tests/unit/basic_receiver.cpp
```diff
@@ -0,0 +1,288 @@
...
+        mylib::receiver_1 rcv;
+        hpx::basic_execution::set_done(std::move(rcv));
+        HPX_TEST(done_called);
+        done_called = false;
+        hpx::basic_execution::set_error(std::move(rcv), std::exception_ptr{});
+        HPX_TEST(error_called);
+        error_called = false;
+        hpx::basic_execution::set_value(std::move(rcv), 1);
+        HPX_TEST(value_called);
+        value_called = false;
```
msimberg, line 275: `clang-tidy` doesn't like the multiple moves of `rcv`: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/3682/workflows/68cf2de8-a809-4344-b23c-0abc3c4355c5/jobs/194028/steps (with good reason, it looks like).

hkaiser:
Note: the recent simplification does not fully reflect the original code as it does not verify `is_sender<>` on the first argument and does not enforce `is_operation_state<>` on the result. This still needs to be readded.

hkaiser:
This has all the changes from #4640 and #4642.

libs/core/execution_base/include/hpx/execution_base/operation_state.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+#include <hpx/config/constexpr.hpp>
+#include <hpx/functional/tag_fallback_invoke.hpp>
+#include <hpx/functional/tag_invoke.hpp>
+#include <hpx/functional/traits/is_invocable.hpp>
+
+#include <type_traits>
+#include <utility>
+
+namespace hpx { namespace execution_base { namespace experimental {
```
msimberg, line 17:
```suggestion
namespace hpx { namespace execution { namespace experimental {
```
? I'll happily do this change since! :D I wasn't very explicit when I first suggested the namespace change.

msimberg (dismissed):
Sorry for not looking at this earlier @hkaiser, but this looks good now. Thanks so much for bringing this up to date! I think having this in soon would be important so that we can start playing around with integration with our tasking.

hkaiser:
Thanks @msimberg for looking into this!

msimberg:
> Thanks @msimberg for looking into this!

Not much I did, but you're welcome :) Shall we go ahead with this? Assuming that CircleCI builder doesn't come back with any surprises it's ready to go according to CI at least.

hkaiser:
> Shall we go ahead with this?

Let's merge this.

PR #4687, opened 2020-05-28
Adding all_gather and fixing all_to_all 

hkaiser:
This essentially renames the existing `all_to_all` to `all_gather` and adds a proper implementation of `all_to_all`.

This relies on #4678 (only the last commit is significant).


hkaiser:
This should be good to go now as well (once the tests have cycled).

msimberg:
I'll merge this despite pycicle not having results. It looks like the only failure in older builds was due to using `hpx/collectives.hpp` rathern than `hpx/modules/collectives.hpp`. I'll fix it afterwards if something else sneaks in.

Issue #4594, opened 2020-05-05
Rename modules


msimberg:
Choosing `async` as the name for a module was not the best idea (because of potential conflicts between the namespace `hpx::async` and the real `hpx::async`).

I'd rename it `distributed_async`. Other ideas or preferences?

Modules that need to be renamed:

- [x] `local_async` (#4689)
- [x] `async` (#4689)
- [x] `mpi` (#4689)
- [x] `cuda_support` (#4385)
- [x] `local_lcos` (#4740)
- [x] `basic_execution` (#4740)

hkaiser:
In the light of our recent discussion to rename the mpi and cuda future modules to `async_mpi` and `async_cuda` I'd suggest to rename this one to `async_distributed` - even more as we already have `async_local`.

msimberg:
> In the light of our recent discussion to rename the mpi and cuda future modules to `async_mpi` and `async_cuda` I'd suggest to rename this one to `async_distributed` - even more as we already have `async_local`.

Good point. It is `local_async` at the moment though, so I'll rename that one as well. I can go and rename all of them once we have the CUDA PR in as well. Were there others (`async_base`, `async_local`, `async_distributed`, `async_mpi`, and `async_cuda`)?

hkaiser:
@msimberg thanks. In the end it does not matter how we name the modules. It would be (slightly) preferable to name them consistently, however.

msimberg:
All the renamings have been done.

PR #4689, opened 2020-05-29
Rename modules

aurianer:
Fixes partially #4594 

Rename async modules:
`local_async` => `async_local`
`async` => `async_distributed`
`mpi` => `async_mpi`


libs/async/include/hpx/async.hpp
```diff
@@ -1,15 +0,0 @@
...
-
-#pragma once
-
-#include <hpx/async/apply.hpp>
-#include <hpx/async/async.hpp>
-#include <hpx/async/async_callback.hpp>
-#include <hpx/async/async_continue.hpp>
-#include <hpx/async/async_continue_callback.hpp>
-#include <hpx/async/dataflow.hpp>
-#include <hpx/async/sync.hpp>
```
msimberg, line 15: This one we should keep. The `hpx/include` headers are one of the few things that we can 100% guarantee to be the same as in 1.4.1.

aurianer: I thought the one we keep is `libs/include/include/hpx/include/async.hpp` which I did not rename, does the one is `async_distributed` (`libs/async_distributed/include/hpx/async.hpp`) also is garanteed ?

msimberg: Yes, thanks for correcting me! I mentally skipped `async` in `libs/async/include` and thought this was in the `include` module. Is `hpx/async_distributed.hpp` generated now?

aurianer: I did not auto-generate it for now as there is 16 headers in the `async_distributed` module (excluding the detail and fwd ones) and only 7 in `hpx/async_distributed.hpp`, should I generate it ?

msimberg: Nope, let's keep it like this (again, I missed that this was moved, and not just deleted).

PR #4692, opened 2020-05-31
Changes to apply needed for cuda/cublas executor

biddisco:
Two patches:
#p1
    Use decltype(auto) for apply return type to fix cuda executor issue:
makes the `hpx::apply` dispatch match the `hpx::async` dispatch and allows the return type to be deduced by return from the executor instead of the function - which breaks when functions have things like cuda streams added by the executor

#p2 Allow apply to return a value:
Certain executors such as cuda, cublas, or networking, are able to
create asynchronous work without spawning a task and will usually return
a future, however the use of `apply` instead of `async` might be preferred
in certain cases to allow multiple kernels (for example) to be launched
on a stream (execution order guaranteed by the 3rd party library).

In these cases, the return value of the function invocation should
be passed back to the user so that a call to apply can be checked
before proceeding if an error code is returned from a function.

This allows constructions such as
    ```cudaError_t err = hpx::apply(cuda_exec, cudaMemcpyAsync,
        d_A, h_A.data(), size_A * sizeof(T), cudaMemcpyHostToDevice);```
to be used and the return checked directly.

### DIFF
--- a/libs/execution/include/hpx/execution/executors/execution.hpp
+++ b/libs/execution/include/hpx/execution/executors/execution.hpp
@@ -463,7 +463,7 @@ namespace hpx { namespace parallel { namespace execution {
                 std::forward<Ts>(ts)...))
             {
                 // use post, if exposed
-                exec.post(std::forward<F>(f), std::forward<Ts>(ts)...);
+                return exec.post(std::forward<F>(f), std::forward<Ts>(ts)...);
             }
 
             template <typename TwoWayExecutor, typename F, typename... Ts>
--- a/libs/executors/include/hpx/executors/apply.hpp
+++ b/libs/executors/include/hpx/executors/apply.hpp
@@ -49,9 +49,8 @@ namespace hpx { namespace detail {
             traits::is_threads_executor<Executor>::value>::type>
     {
         template <typename Executor_, typename F, typename... Ts>
-        HPX_FORCEINLINE static typename std::enable_if<
-            traits::detail::is_deferred_invocable<F, Ts...>::value, bool>::type
-        call(Executor_&& exec, F&& f, Ts&&... ts)
+        HPX_FORCEINLINE static decltype(auto) call(
+            Executor_&& exec, F&& f, Ts&&... ts)
         {
             parallel::execution::post(std::forward<Executor_>(exec),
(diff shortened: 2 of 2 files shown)


biddisco:
I see on CI that async_customization is failing. This is due to the suboptimal implementation of the shim executor - this can actually be replaced with the new polymorphic_executor (thanks @hkaiser  for implementing that). I shall redo the test to use that and push again once it is done.

hkaiser:
> Two patches:
> #p1
> Use decltype(auto) for apply return type to fix cuda executor issue:
> makes the hpx::apply dispatch match the hpx::async dispatch and allows the return type to be deduced by return from the executor instead of the function - which breaks when functions have things like cuda streams added by the executor

Sorry, but I don't think returning anything from `post`/`apply` makes any sense. These operations are explicitly fire & forget, so returning something from `post`/`apply` could be useful only to report errors that occurred while executing the `post` or `apply` itself (not errors caused by the task that is being posted). However, this kind of errors should be reported using exceptions.

> #p2 Allow apply to return a value:
> Certain executors such as cuda, cublas, or networking, are able to
> create asynchronous work without spawning a task and will usually return
> a future, however the use of apply instead of async might be preferred
> in certain cases to allow multiple kernels (for example) to be launched
> on a stream (execution order guaranteed by the 3rd party library).
> In these cases, the return value of the function invocation should
> be passed back to the user so that a call to apply can be checked
> before proceeding if an error code is returned from a function.
> This allows constructions such as
> cudaError_t err = hpx::apply(cuda_exec, cudaMemcpyAsync, d_A, h_A.data(), size_A * sizeof(T), cudaMemcpyHostToDevice);
> to be used and the return checked directly.

For this reason, I would vote against relying on returning an error code from `apply` like this. If there is a CUDA error generated while posting the kernel to the device, an exception should be thrown. This would be important for consistency as we usually employ exception handling to report errors. Returning an error code is very error prone in itself as people will forget to check for the result.


Please note that I'm not against removing the over-constraining SFINAE on the `post` customization point implementation. This is a valid change that fixes a problem in the current implementation.

biddisco:
I've reduced this PR to just the first patch that is needed for cuda async/apply. Please merge if it clears CI

PR #4693, opened 2020-06-01
Overhauling thread_mapper for public consumption

### LINKED ISSUE #4691: Suggestion to fix and enhance the thread_mapper API
The thread_mapper (`hpx::util::thread_mapper`) currently offers some limited
thread access functionality. It maintains a list of threads whose count you 
can get with the `get_thread_count()` member function.
You may then use indices 0 to count-1 and use (amongst others) these member functions:
```c++
long int get_thread_id(std::uint32_t tix) const;              // Returns -1  on Windows and is unusable there
std::string const& get_thread_label(std::uint32_t tix) const; // Get thread label from -index
std::uint32_t get_thread_index(std::string const&) const;     // Get thread index from -label
```

I would like to suggest to add a struct to the API which contains useful thread information:

```c++
struct os_thread_info {
    std::thread::id const  id;       // Portable type of the threads id on the system.
    std::string const      label;    // Label of the thread as before
    thread_type const      type;     // Enum class of threads type, like "worker", "timer", "io", ...    
    size_t                 userdata; // Thread Data, like in the case of tasks, could be size_t, void*, whatever.
}
```
The structure would be accessible by `os_thread_info& get_info(std::uint32_t tix)`
with an overload like `os_thread_info& get_info(std::thread::id id)`

Reasons:

1. The retrieval of thread IDs on Windows is not working.
2. The thread IDs being used are not C++ Standard conformant
3. There is currently no standardized way to associate data with an OS thread. 
E.g. one might use an object pool or allocator local to a thread to avoid contention.
(We are currently doing this for a pool of reusable Lua States and have one pool per OS thread)
4. After initial setup read access should be safe from any thread.

I'm sure there are many more possible improvements, but this is my first wish for this API.

Ideas? Suggestions?

hkaiser:
Fixes #4691


McKillroy:
I am now getting correct thread IDs from the mapper on windows, which was broken before.
With this the thread_mapper becomes useable for me.
I still would suggest on the long run to replace the long ints with correct standard std::thread::id types for standard conformance.
Custom thread data can now easily be implemented by the user.
Thanks for the quick fix - I hope this gets merged soonish!

hkaiser:
> I still would suggest on the long run to replace the long ints with correct standard std::thread::id types for standard conformance.

I'll see what I can do.

McKillroy:
I like this new change with the possibility to get the `std::thread::id` type. It makes some hacky conversions unnecessary.

hpx/util/thread_mapper.hpp
```diff
@@ -7,108 +8,132 @@
...
-    private:
-        // null callback
-        static bool null_cb(std::uint32_t);
+        // type for callback function invoked when thread is unregistered
+        using thread_mapper_callback_type =
+            util::function_nonser<bool(std::uint32_t)>;
 
         // thread-specific data
-        struct thread_data
+        class HPX_EXPORT thread_data
```
msimberg, line 38: I know this is in `detail` and the name doesn't matter too much, but would it be worth it to rename this to something like `os_thread_data` to avoid confusion with the HPX-thread `thread_data`?

libs/basic_execution/include/hpx/basic_execution/this_thread.hpp
```diff
@@ -20,47 +20,66 @@
...
+    enum class thread_type
+    {
+        unknown = -1,
+        main_thread = 0,
+        worker_thread,
+        io_thread,
+        timer_thread,
+        parcel_thread,
+        custom_thread
+    };
```
msimberg, line 36: I don't think this belongs here. This is purely a runtime concept and could live in e.g. the thread mapper files. Having `hpx::basic_execution` in the thread mapper API is a bit odd as well (in light of defining our public API; I'm aware that we have other existing "public" APIs that use the nested module namespaces, but that was from before our current API discussions).

hkaiser: These are needed by the `thread_executors` module (amongst other places). What module would you suggest I move that `enum` to instead?

hkaiser: Also, what namespace would you suggest we move the thread mapper to?

msimberg: Ah, right... Let's leave it here for now in that case, but since the thread executors are deprecated I would move this up to the `runtime_local` module from #4606 when we remove the thread executors. Or am I missing some other non-runtime place where the enum is used?

I would put `thread_mapper` in `hpx::runtime_local` (the `thread_mapper` files are already in that module in #4606). If this is really intended to be in the public API we should probably have free functions for the rest of the thread mapper member functions (just like `thread_mapper::(un)register_thread` already have `hpx::(un)register_thread`). If we're unsure let's not have the wrappers until we know better.

hkaiser: I have left the enum in the `basic_execution` module for now. However, it's now in a separate file and can easily be moved. I have also added an API function that allows to access the data that is stored by the runtime for each registered thread: `hpx::get_os_thread_data()`

msimberg:
What do we want to do with this one now? It seems like the removal of `bimap` from the thread mapper is a big win in any case. I also think that this is a welcome change even if @McKillroy didn't need these particular changes in the end.

PS. I think I will never be able to unsee std::thread::id.

hkaiser:
Let's go ahead with this as is.

McKillroy:
I think there is one lesson to be learned:
I had a requirement to initialize per-worker:thread: structures ahead of using these :thread:s. _[Hi Mikael - welcome to a_ :thread:_ed world_ 😉 _]_ 
But when I asked in IRC - actually it came up on several occasions - I didn't get a satisfying answer right away, which means even the people knowing HPX extremely well couldn't come up with the easy solution I had in the end spontaneously - it took time.
To me this means, that HPX has become so powerful in its functionality and broad in its applicability, that even specialists can not always overlook what is possible.
My conclusion would be, that we need a sort of knowledge base or collection of recipes, which is searchable and goes beyond the examples directory. 
I see it a bit like a bug, that now got resolved, but we should now have a test for it to prevent regressions. This would translate to having an entry with the problem and the solution somewhere, which can be indexed, tagged/categorized and is full-text searchable. 
The Wiki together with the examples? I'm not sure whats best and also maintainable.

hkaiser:
@msimberg this is good to go now (once the tests have cycled).

msimberg:
@hkaiser I hope I fixed the last deprecated headers. Feel free to merge once CI is happy.

Issue #4656, opened 2020-05-22
HEP 2 implementation tracker


msimberg:
This issue tracks how much of #4626 has been implemented. Since HEP 2 is still in progress, this is not necessarily a complete list of things that need to be done.

Headers:

- [ ] `hpx/algorithm.hpp`
- [ ] `hpx/any.hpp`
- [ ] `hpx/barrier.hpp`
- [ ] `hpx/chrono.hpp`
- [ ] `hpx/condition_variable.hpp`
- [ ] `hpx/exception.hpp`
- [ ] `hpx/execution.hpp`
- [ ] `hpx/future.hpp`
- [ ] `hpx/iostream.hpp`
- [x] `hpx/latch.hpp`
- [ ] `hpx/memory.hpp`
- [ ] `hpx/mutex.hpp`
- [ ] `hpx/numeric.hpp`
- [ ] `hpx/optional.hpp`
- [x] `hpx/semaphore.hpp`
- [ ] `hpx/shared_mutex.hpp`
- [ ] `hpx/system_error.hpp`
- [ ] `hpx/thread.hpp`
- [ ] `hpx/tuple.hpp`
- [ ] `hpx/runtime.hpp`
- [ ] `hpx/distributed/algorithm.hpp`
- [ ] `hpx/distributed/barrier.hpp`
- [ ] `hpx/distributed/execution.hpp`
- [ ] `hpx/distributed/future.hpp`
- [ ] `hpx/iostream.hpp`
- [ ] `hpx/distributed/latch.hpp`
- [ ] `hpx/distributed/memory.hpp`
- [ ] `hpx/distributed/numeric.hpp`
- [ ] `hpx/distributed/runtime.hpp`

- [ ] Add documentation page with public API
- [ ] update examples to use API headers (#4488)

PR #4695, opened 2020-06-02
API documentation

msimberg:
Once again very much WIP so don't worry about consistency etc. at this point. This is the documentation part of #4626.

I've restructured the public API documentation taking inspiration from the cppreference header docs (like https://en.cppreference.com/w/cpp/header/future). I can't currently upload a snapshot of the docs with these changes, but the changes are all in the "Reference" section of the documentation with the following hierarchy:
```
Reference
    Public API
        Utilities
        Concurrency and parallelism
        Runtime
    Full API
```

- The header pages are brief and group names according to functions, classes, and constants. The brevity is in the interest of avoiding duplication between the "Writing single-node HPX applications" section and the API docs at this point. I instead link to the latter to the former where it fits. However, I do quite like what the python standard library documentation does combining API docs and examples all in one page (e.g. https://docs.python.org/2/library/collections.html). 
- I link to e.g. `hpx::util::tuple` instead of `hpx::tuple` because 1. doxygen doesn't like `using hpx::util::tuple` and 2. it would link to just a typedef. I think we need to do
```
#if defined(DOXYGEN)
// simplified docs
#else
// implementation
#endif
```
for more classes (like `tuple`) for the docs to be readable. On top of that we should not just alias things into `hpx` but actually move them to `hpx`.
- I link to a corresponding standard library header on cppreference wherever it fits.
- I mention the namespace differences (`hpx` vs `hpx::util`) and encourage use of `hpx`.
- I've grouped things at a higher level into "Utilities", "Concurrency and parallelism", and "Runtime", but these should perhaps be the names of the libraries in #4649 instead (i.e. "Core", "Local ???", and "Runtime"?).
- I started adding a small blurb for each header about why we have a certain replacement for standard library functionality, but this is incomplete. @K-ballo perhaps you could help with some of these (datastructures mostly)? I think this is important for users to evaluate if they need our version of something or are fine using the standard library version.
- Not all API links work yet.

What else should we include? Should we remove anything? Or structure things differently?

Finally, I think #4626 is starting to outlive its usefulness. With this and the public API headers (`hpx/algorithms.hpp` and all the rest) a subset of the public API is already defined. I would suggest that we start with what we know now, and expand the public API as we go by updating docs and headers. Before 2.0.0 we'll need to do a double-check that we haven't forgotten anything essential. We will in any case not be able to get #4656 100% right by thinking hard about it. We'll need to have all the modules in place, and users complaining about things that are wrong or missing in our API before it'll be close to right.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -654,6 +654,7 @@ jobs:
                   tests.unit.modules.futures \
                   tests.unit.modules.hardware \
                   tests.unit.modules.hashing \
+                  tests.unit.modules.include \
                   tests.unit.modules.init_runtime \
                   tests.unit.modules.iterator_support \
                   tests.unit.modules.lcos_distributed \
--- a/cmake/templates/autodoc.doxy.in
+++ b/cmake/templates/autodoc.doxy.in
@@ -4,15 +4,16 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-EXTRACT_ALL = YES
+DIRECTORY_GRAPH = NO
+EXCLUDE_SYMBOLS = detail
 EXPAND_ONLY_PREDEF = YES
+EXTRACT_ALL = YES
+EXTRACT_PRIVATE = NO
 GENERATE_HTML = NO
+GENERATE_LATEX = NO
 GENERATE_XML = YES
 MACRO_EXPANSION = YES
 PREDEFINED = @doxygen_definitions@
 XML_OUTPUT = @doxygen_output_file@
 OUTPUT_DIRECTORY = @doxygen_output_dir@
-GENERATE_LATEX = NO
 INPUT = @doxygen_inputs@
-EXCLUDE_SYMBOLS = detail
-
--- a/cmake/templates/conf.py.in
+++ b/cmake/templates/conf.py.in
@@ -70,6 +70,7 @@ for subdir_full, dirs, files in os.walk(hpx_libs_dir):
     if len(subdir) == 0:
         for dir in dirs:
             breathe_projects_source[dir] = (subdir_full + '/' + dir, [])
+            breathe_projects[dir] = '../breathe/doxygen/' + dir + '/xml'
     # We are inside a library and need to add the files
     # to the breathe project
     else:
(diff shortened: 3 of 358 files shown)


libs/assertion/include/hpx/assertion.hpp
```diff
@@ -43,9 +43,9 @@ namespace hpx { namespace assertion {
 /// \param msg The optional message that is used to give further information if
 ///             the assert fails. This should be convertible to a std::string
 ///
-/// If \a expr evaluates to false, The source location and \a msg is being
+/// If \p expr evaluates to false, The source location and \p msg is being
 /// printed along with the expression and additional. Afterwards the program is
-/// being aborted. The assertion handler can be costumized by calling \a
+/// being aborted. The assertion handler can be custumized by calling
```
hkaiser, line 48: `customized`?

msimberg: Yep... talk about being blind. Thanks.

msimberg:
@sithhell did you have additional ideas for this?

msimberg:
I'll go ahead with the current structure next week unless there are comments by then. This need not (and will not) be the final structure anyway, but will serve well enough for 1.5.0.

One thing I want to add eventually is something like the example section here: https://en.cppreference.com/w/cpp/thread/future, except I'd do it per header, not per class. Something brief that shows basic usage of whatever is in the header. It would also serve as a small unit test to check that we've put everything in the right namespace.

msimberg:
I'm now fairly happy with the structure. I've included for each API header a list of

- Constants
- Macros
- Classes
- Functions
- Examples

in that order, following the convention from cppreference. The examples section is there only for the `hpx/future.hpp` header.

There are a few links that I cannot for the life of me get to work. These are for classes that are anyway not documented (e.g. `future`, `function`) so that needs to be fixed anyway, but I'd like to keep that separate from this PR. The class shows up in the documentation for the module but the link simply does not work. With a significantly stripped down class definition the links do work, so there's something in the definition that breathe or doxygen doesn't like. The documentation for those classes has to exclude some implementation details anyway, so I think that won't be a problem.

I removed the grouping into core/parallelism/etc. at this point since that split is not there yet in the libraries and the names are not clear yet. Instead I put all the headers on one page in alphabetical order.

msimberg:
Fixed inspect and formatting complaints... If CI comes back clean this should be ready.

PR #4696, opened 2020-06-02
Rename generated modules headers

### LINKED ISSUE #4647: Rename generated collective headers in modules
Since some headers in #4626 would clash with our current naming scheme of naming collective headers `hpx/modulename.hpp`, and #4626 should have precedence, we should rename the generated headers (e.g. `hpx/execution.hpp` should include much more than what the execution module contains, or the `uninitialized_X` functions need to be in `hpx/memory.hpp` but the memory module has already taken that name).

Since we've been calling them modules I propose `hpx/modules/modulename.hpp`. We've also chopped things up so that what we call modules could in fact be a one-to-one mapping to C++ modules eventually. We should also check that we're consistent with library/module terminology in our CMake and documentation.

aurianer:
Fixes #4647

Rename the modules main headers from `hpx/<modulename>.hpp` to `hpx/modules/<modulename>.hpp`

Note: it's on top of #4689 only the last 5 commits matter for this PR


aurianer:
Sorry I forgot to adapt the inspect tool to the new assertion header, this should be good now :)

libs/assertion/include_compatibility/hpx/util/assert.hpp
```diff
@@ -14,7 +14,7 @@
 
 #include <hpx/config.hpp>
 #include <hpx/assertion/config/defines.hpp>
-#include <hpx/assertion.hpp>
+#include <hpx/modules/assertion.hpp>
 
 #if defined(HPX_ASSERTION_HAVE_DEPRECATION_WARNINGS)
 #if defined(HPX_MSVC)
```
msimberg, line 20: The error messages in the compatibility headers need to be updated as well.

msimberg:
I'll merge this since the lint checks pass on CircleCI and the builds pass on GitHub actions and pycicle. I'm ignoring the spelling suggestions.

PR #4700, opened 2020-06-03
Removing obsolete memory component

hkaiser:
As far as I know, nobody has been using this for years. If necessary, we can easily re-add it as an example outside of the core library.


hkaiser:
@msimberg I think this was the only place where we used the `int128_t`/`uint128_t` emulation types. Should we remove those as well? Is anybody using them outside of HPX?

msimberg:
As far as I know at least noone at CSCS is using them, so from my side they can be removed.

PR #4704, opened 2020-06-03
Fix resource tests

### LINKED ISSUE #4661: Fix resource tests
The commit https://github.com/STEllAR-GROUP/hpx/commit/550950f3a0c52ba3ff59740195e15884e01deb72 has disabled three of the tests in `util/resource` in order to allow for CircleCI to generate a new stable tag. These tests need to be fixed.

msimberg:
Fixes #4661 ~, maybe someday... This doesn't actually fix anything yet. I'm struggling to reproduce this locally so I'll play around a bit more with this on CircleCI. It may be hitting some pathological case that only happens with limited resources. Not sure what's going on. Let's see.~

### DIFF
--- a/libs/executors/include/hpx/executors/parallel_executor.hpp
+++ b/libs/executors/include/hpx/executors/parallel_executor.hpp
@@ -216,9 +216,7 @@ namespace hpx { namespace parallel { namespace execution {
         template <typename F, typename... Ts>
         void post(F&& f, Ts&&... ts) const
         {
-            char const* annotation =
-                hpx::traits::get_function_annotation<F>::call(f);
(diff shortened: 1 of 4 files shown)


msimberg:
I think this fixes all the resource tests now. It certainly fixes one bug. In cases where there is no annotation for a function, no `altname`, and the thread is spawned from a non-HPX thread, there was no fallback, and the description was uninitialized. I've fixed that. I also changed the one occurrence of `<anonymous task>` to `<unknown>` (I don't mind `<anonymous task>`, but changing this one occurrence to what we were already use elsewhere is a smaller change...). Finally, I removed the unnecessary `get_function_annotation` from `parallel_executor::post`.

PR #4707, opened 2020-06-04
Use hpx::function_nonser instead of std::function internally

### LINKED ISSUE #4672: Consistently use hpx::function_* instead of std::function

msimberg:
Fixes #4672. I've only replaced uses of `std::function` in the main library. I've left `std::function` in tests, examples, and the `program_options` module.

### DIFF
--- a/hpx/runtime/threads/thread_pool_suspension_helpers.hpp
+++ b/hpx/runtime/threads/thread_pool_suspension_helpers.hpp
@@ -8,6 +8,7 @@
 
 #include <hpx/config.hpp>
 #include <hpx/futures/future.hpp>
+#include <hpx/modules/functional.hpp>
 #include <hpx/threading_base/thread_pool_base.hpp>
 
 #include <cstddef>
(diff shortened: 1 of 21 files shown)


hkaiser:
We might want to have:
```
namespace hpx {
    using function = hpx::util::function_nonser;
    namespace distributed {
        using function = hpx::util::function;
    }
}
```
(not sure if this is possible as we can't do `using` specializations , but conceptually this might be nice).

msimberg:
> We might want to have:
> 
> ```
> namespace hpx {
>     using function = hpx::util::function_nonser;
>     namespace distributed {
>         using function = hpx::util::function;
>     }
> }
> ```
> 
> (not sure if this is possible as we can't do `using` specializations , but conceptually this might be nice).

Hmm, not sure we need to make one "distributed" and one "local" (although it could be a nice split as well), i.e. we could just keep the names `function`, `function_nonser`, `unique_function`, `unique_function_nonser` when pulling them into `hpx`. I planned to add the `hpx/functional.hpp` header after the module header renaming, but will do it after #4710. Let's discuss on that PR what we do.

PR #4708, opened 2020-06-04
Add termination handler for parallel algorithms

msimberg:
I thought this was a circular dependency that `cpp-dependencies` did not see. It turns out it was fine, since the `init_runtime` module did not depend on algorithms. In any case, this gets rid of the dependency from `algorithms` to `init_runtime` and allows us to have `algorithms` in a separate library that does not depend on the runtime.

### DIFF
--- a/libs/algorithms/src/handle_exception_termination_handler.cpp
+++ b/libs/algorithms/src/handle_exception_termination_handler.cpp
@@ -1,0 +1,33 @@
+//  Copyright (c) 2020      ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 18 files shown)


libs/algorithms/src/handle_exception_termination_handler.cpp
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    HPX_NORETURN void parallel_exception_termination_handler()
+    {
+        if (get_parallel_exception_termination_handler())
+        {
+            get_parallel_exception_termination_handler()();
+        }
+
+        std::terminate();
```
hkaiser, line 31: `std::terminate` requires #include <exception>`. Looks like inspect is not catching that.

msimberg: But you did, thanks! ;) I'll add it to the inspect check.

PR #4710, opened 2020-06-04
Add compatibility headers and other fixes after module header renaming

msimberg:
`hpx/functional.hpp` and `hpx/async.hpp` were put back as normal headers. `hpx/functional.hpp` will soon be replaced with a header that forwards functional stuff into the `hpx` namespace.


msimberg:
@hkaiser I think that's it to at least get back compatibility with what was there before the renaming in #4696. As state on IRC, this might be slightly too much and we may still undeprecate some of these headers before 1.5.0.

msimberg:
It looks like the old comments got lost because I commented on a commit, rather than the whole PR. Continuing here:

> Ok, I will revert the removal of the forwarding headers. I however don't know what to replace hpx/format.hpp (and similar) with if the corresponding hpx/modulename/module.hpp is for internal use only?

The point I was trying to badly make is that if you need to include `hpx/modules/modulename.hpp` you do so. I think you won't have big problems in phylanx even if we make some incompatible changes along the way.

I just wouldn't want to recommend users to include `hpx/modules/modulename.hpp` headers. It's not so much the header location I'm worried about changing, but the contents of the header. I'd like for us to have the freedom to change things internally without having to worry about compatibility options. We just won't give the same guarantees about backward compatibility with `hpx/modules/modulename.hpp` headers. And in phylanx you can't anyway go back to older versions of HPX because you rely on new features on master.

msimberg:
Are there any other headers from this list: https://github.com/STEllAR-GROUP/hpx/pull/4710/commits/669a3530ef7cac545498bc93f99c5ceee739f6f4 that you think should be public?

aurianer:
Thanks a lot for this !

PR #4713, opened 2020-06-05
Remove hpx/runtime/threads_fwd.hpp and hpx/util_fwd.hpp

msimberg:
Removes two `_fwd.hpp` headers without compatibility options. Is someone actually relying on these? `hpx/lcos_fwd.hpp` is a bit trickier...


hpx/runtime/components/server/abstract_migration_support.hpp
```diff
@@ -135,5 +135,4 @@ namespace hpx { namespace components
             return this->base_type::on_migrated();
         }
     };
-}}
-
+}}    // namespace hpx::components
```
hkaiser, line 138: Should we format these files at this point?

hpx/runtime/components/server/migration_support.hpp
```diff
@@ -239,5 +243,4 @@ namespace hpx { namespace components
         hpx::lcos::local::promise<void> trigger_migration_;
         bool was_marked_for_migration_;
     };
-}}
-
+}}    // namespace hpx::components
```
hkaiser, line 246: Same here: should this be formatted now?

msimberg: No, this was accidental. I'll stick to only formatting/sorting includes in these files.

hpx/runtime/components/server/abstract_migration_support.hpp
msimberg: Removed these changes.

hpx/runtime/components/server/migration_support.hpp
msimberg: Removed these too.

PR #4714, opened 2020-06-05
Add hpx/future.hpp header

msimberg:
This is in the `include` module because it requires pulling in headers from quite a few different modules.

I've also added a little unit test/example that is roughly what I was thinking we could include alongside the public API listings in #4695.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -654,6 +654,7 @@ jobs:
                   tests.unit.modules.futures \
                   tests.unit.modules.hardware \
                   tests.unit.modules.hashing \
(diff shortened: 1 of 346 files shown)


msimberg:
In the last commit I changed all the `hpx/modules/assertion.hpp` includes to `hpx/assert.hpp` because the latter is part of the public API. This is a bit of an artifact from our inspect tool only allowing one header for a given symbol (it makes sense). I wouldn't actually mind including `hpx/modules/assertion.hpp` internally and only use `hpx/assert.hpp` in e.g. examples, but inspect will not allow that without changes. Does anybody mind this change?

PR #4716, opened 2020-06-05
Add hpx/runtime.hpp and hpx/distributed/runtime.hpp API headers

msimberg:
Adds the runtime headers according to #4626, but does not take any stance on what exactly is exposed through these headers. The `hpx/runtime.hpp` header exposed a bit more in 1.4.1 than it will now. Most existing users of HPX will want to include `hpx/distributed/runtime.hpp` instead of `hpx/runtime.hpp` after this, but we can't really add any warnings since `hpx/runtime.hpp` is still a valid header on its own.


libs/include/include/hpx/distributed/runtime.hpp
```diff
@@ -0,0 +1,10 @@
+//  Copyright (c) 2020 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/runtime/runtime_fwd.hpp>
+#include <hpx/runtime_fwd.hpp>
```
hkaiser, line 10: Just `#including` forwarding declarations might not be sufficient. Are you sure this is fine?

msimberg: Depends what we want to expose. Actually the first include might not even be necessary. The second includes all the `hpx::get_locality_name`, `hpx::find_locality` etc. functions (including the local functions). If accessing the runtime directly is public then this might not be enough, but I would've hoped we could get by without making that officially public.

PR #4718, opened 2020-06-05
Remove force linking functionality

msimberg:
Separates the force linking changes from #4649, to make that PR a bit smaller.


msimberg:
> LGTM, thanks!
> 
> I have not checked, but could you please make sure that _all_ of the dummy functions that were added for this are removed?

I've removed all that I can tell were for force linking. There were a number of `dummy` functions that had a comment saying they they were for force linking. I looked through everything else called dummy as well and they did not look like they were for force linking. Do you remember if there might be anything else I should search for?

hkaiser:
> Do you remember if there might be anything else I should search for?

Unfortunately, no. Let's go ahead with this and remove things if we happen to stumble across them.

msimberg:
I think this will conflict with some of the renamings (the file lists in CMake files have changed). Let's get those in first, then this.

msimberg:
> It shows that the two fo the `configure_file` invocations have been removed. As a result, this PR shaves off another 3s from the execution time for the configuration step. Nice.

Indeed very nice! Interestingly if you ignore the `add_hpx_source_group` calls, which don't happen on linux, way over 50% of the time is spent generating files (headers). I don't currently know of a way to improve that, but in the worst case those could just be generated and checked into the repo (but I'd like to avoid this if there's a CMake solution to speeding that up).

PR #4724, opened 2020-06-07
Fix 4723

### LINKED ISSUE #4723: Build fail with NETWORKING OFF
Looks like the usual problem of trying to use distributed stuff when it should be disabled
```
/home/biddisco/src/hpx/libs/init_runtime/src/hpx_init.cpp: In function ‘int hpx::disconnect(double, double, hpx::error_code&)’:
/home/biddisco/src/hpx/libs/init_runtime/src/hpx_init.cpp:991:17: error: ‘get_runtime_distributed’ was not declared in this scope; did you mean ‘runtime_distributed’?
  991 |                 get_runtime_distributed().get_runtime_support_lva());
      |                 ^~~~~~~~~~~~~~~~~~~~~~~
      |                 runtime_distributed
/home/biddisco/src/hpx/libs/init_runtime/src/hpx_init.cpp: In function ‘void hpx::terminate()’:
/home/biddisco/src/hpx/libs/init_runtime/src/hpx_init.cpp:1022:17: error: ‘get_runtime_distributed’ was not declared in this scope; did you mean ‘runtime_distributed’?
 1022 |                 get_runtime_distributed().get_runtime_support_lva());
      |                 ^~~~~~~~~~~~~~~~~~~~~~~
      |                 runtime_distributed
```

biddisco:
Fixes #4723 

Fix 1 - the module name is `MPI_BASE`, but various `#ifdef` checks in some files are using module `MPI` - this is not related to 4723, but needs to be addressed. Change `XXX_LIB_MPI` to `XXX_LIB_MPI_BASE` in several places

Fix 2 - when the user sets `HPX_WITH_XXX` it may be asked for, but it might not be possible if other contradicting options have been set. The code should set `HPX_HAVE_XXX` if and only if that option's requirements are fulfilled. A clear distinction between `WITH` and `HAVE` exists between features requested, and features delivered.

`HPX_WITH_DISTRIBUTED_RUNTIME` cannot be met if `HPX_WITH_NETWORKING=OFF` so the cmake code should not use `HPX_WITH_DISTRIBUTED_RUNTIME` directly in checks (in other modules etc), but rather a new variable `HPX_HAVE_DISTRIBUTED_RUNTIME` that is only set if all conditions are met.

### DIFF
--- a/libs/mpi_base/include/hpx/mpi_base/mpi.hpp
+++ b/libs/mpi_base/include/hpx/mpi_base/mpi.hpp
@@ -7,7 +7,7 @@
 #pragma once
 
 #if (defined(HPX_HAVE_NETWORKING) && defined(HPX_HAVE_PARCELPORT_MPI)) ||      \
-    defined(HPX_HAVE_LIB_MPI)
+    defined(HPX_HAVE_LIB_MPI_BASE)
 
 #if defined(__clang__)
 #pragma clang diagnostic push
--- a/libs/mpi_base/include/hpx/mpi_base/mpi_environment.hpp
+++ b/libs/mpi_base/include/hpx/mpi_base/mpi_environment.hpp
@@ -9,7 +9,7 @@
 #include <hpx/config.hpp>
 
 #if (defined(HPX_HAVE_NETWORKING) && defined(HPX_HAVE_PARCELPORT_MPI)) ||      \
-    defined(HPX_HAVE_LIB_MPI)
+    defined(HPX_HAVE_LIB_MPI_BASE)
 
 #include <hpx/modules/runtime_configuration.hpp>
 #include <hpx/mpi_base/mpi.hpp>
--- a/libs/mpi_base/src/mpi_environment.cpp
+++ b/libs/mpi_base/src/mpi_environment.cpp
@@ -66,7 +66,7 @@ namespace hpx { namespace util {
             return false;
         }
         return true;
-#elif defined(HPX_HAVE_LIB_MPI)
+#elif defined(HPX_HAVE_LIB_MPI_BASE)
         // if MPI futures are enabled while networking is off we need to
         // check whether we were run using mpirun
         return detail::detect_mpi_environment(cfg, HPX_HAVE_PARCELPORT_MPI_ENV);
@@ -77,7 +77,7 @@ namespace hpx { namespace util {
 }}    // namespace hpx::util
 
 #if (defined(HPX_HAVE_NETWORKING) && defined(HPX_HAVE_PARCELPORT_MPI)) ||      \
-    defined(HPX_HAVE_LIB_MPI)
+    defined(HPX_HAVE_LIB_MPI_BASE)
 
 namespace hpx { namespace util {
 
--- a/libs/resource_partitioner/src/detail_partitioner.cpp
+++ b/libs/resource_partitioner/src/detail_partitioner.cpp
@@ -894,7 +894,7 @@ namespace hpx { namespace resource { namespace detail {
         cfg_.hpx_main_f_ = f;
 
 #if (defined(HPX_HAVE_NETWORKING) && defined(HPX_HAVE_PARCELPORT_MPI)) ||      \
-    defined(HPX_HAVE_LIB_MPI)
+    defined(HPX_HAVE_LIB_MPI_BASE)
         // getting localities from MPI environment (support mpirun)
         if (util::mpi_environment::check_mpi_environment(cfg_.rtcfg_))
         {
--- a/libs/version/src/version.cpp
+++ b/libs/version/src/version.cpp
@@ -20,7 +20,7 @@
 #include <boost/version.hpp>
 
 #if (defined(HPX_HAVE_NETWORKING) && defined(HPX_HAVE_PARCELPORT_MPI)) ||      \
-    defined(HPX_HAVE_LIB_MPI)
+    defined(HPX_HAVE_LIB_MPI_BASE)
 #if defined(__clang__)
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wcast-qual"
@@ -86,7 +86,7 @@ namespace hpx {
     }
 
 #if (defined(HPX_HAVE_NETWORKING) && defined(HPX_HAVE_PARCELPORT_MPI)) ||      \
-    defined(HPX_HAVE_LIB_MPI)
+    defined(HPX_HAVE_LIB_MPI_BASE)
     std::string mpi_version()
     {
         std::ostringstream strm;
@@ -266,7 +266,7 @@ namespace hpx {
                                                 "  Boost: {}\n"
                                                 "  Hwloc: {}\n"
 #if (defined(HPX_HAVE_NETWORKING) && defined(HPX_HAVE_PARCELPORT_MPI)) ||      \
-    defined(HPX_HAVE_LIB_MPI)
+    defined(HPX_HAVE_LIB_MPI_BASE)
                                                 "  MPI: {}\n"
 #endif
                                                 "\n"
@@ -278,7 +278,7 @@ namespace hpx {
                                                 "  Standard Library: {}\n",
             build_string(), boost_version(), hwloc_version(),
 #if (defined(HPX_HAVE_NETWORKING) && defined(HPX_HAVE_PARCELPORT_MPI)) ||      \
-    defined(HPX_HAVE_LIB_MPI)
+    defined(HPX_HAVE_LIB_MPI_BASE)
             mpi_version(),
 #endif
             build_type(), build_date_time(), boost_platform(), boost_compiler(),


hkaiser:
Our current policy is to use `HPX_WITH_xxx` as configuration constants in CMake while `HPX_HAVE_xxx` is used as the name for the corresponding preprocessor constant in the actual code. Using both, the `HPX_WITH_xxx` and `HPX_HAVE_xxx` in CMake could be very confusing, I think. What do others say?

CMakeLists.txt
```diff
@@ -429,17 +429,19 @@ hpx_option(
   HPX_WITH_DISTRIBUTED_RUNTIME
   BOOL
   "Enable the distributed runtime (default: ON). Turning off the distributed runtime completely disallows the creation and use of components and actions. Turning this option off is experimental!"
-  ON
+  ${HPX_WITH_NETWORKING}
```
msimberg, line 432: I would keep these independent, simply because `HPX_WITH_DISTRIBUTED_RUNTIME=OFF` is still bound to have corner cases that are not handled in the configuration and command line options etc. (but please do test it as much as possible). There's a reason I've said that it's experimental here.

There may be a need for renaming these variables, but:
- `HPX_WITH_NETWORKING=OFF` means all code works unchanged (even components and actions), you just can't run HPX with more than one locality.
- `HPX_WITH_DISTRIBUTED_RUNTIME=OFF` means that there are no components, actions, performance counters, logging, and a few other things.

msimberg:
> Our current policy is to use `HPX_WITH_xxx` as configuration constants in CMake while `HPX_HAVE_xxx` is used as the name for the corresponding preprocessor constant in the actual code. Using both, the `HPX_WITH_xxx` and `HPX_HAVE_xxx` in CMake could be very confusing, I think. What do others say?

This has been my interpretation as well, even though it may never have been spelled out.

msimberg:
> Our current policy is to use `HPX_WITH_xxx` as configuration constants in CMake while `HPX_HAVE_xxx` is used as the name for the corresponding preprocessor constant in the actual code. Using both, the `HPX_WITH_xxx` and `HPX_HAVE_xxx` in CMake could be very confusing, I think. What do others say?

This has been my interpretation as well, even though it may never have been spelled out.

> `HPX_WITH_DISTRIBUTED_RUNTIME` cannot be met if `HPX_WITH_NETWORKING=OFF` so the cmake code should not use `HPX_WITH_DISTRIBUTED_RUNTIME` directly in checks (in other modules etc), but rather a new variable `HPX_HAVE_DISTRIBUTED_RUNTIME` that is only set if all conditions are met.

`HPX_WITH_DISTRIBUTED_RUNTIME` *can* be `ON` even when `HPX_WITH_NETWORKING` is `OFF`. The dependency is the other way around. As I said in https://github.com/STEllAR-GROUP/hpx/pull/4724#discussion_r436514992 a renaming may be appropriate.

CMakeLists.txt
biddisco: All this change does is set the default `HPX_WITH_DISTRIBUTED_RUNTIME` to OFF if you configure the very first time with `cmake -DHPX_WITH_NETWORKING=OFF` from a clean cache of hPX. it just makes the default off if you have disabled networking.

biddisco:
> Our current policy is to use `HPX_WITH_xxx` as configuration constants in CMake while `HPX_HAVE_xxx` is used as the name for the corresponding preprocessor constant in the actual code. Using both, the `HPX_WITH_xxx` and `HPX_HAVE_xxx` in CMake could be very confusing, I think. What do others say?

It's common practice to have WITH and HAVE as cmake vars. it just makes the scripts a bit simpler because you don't have so many if `this and that or that and those` type clauses in later cmake files as you can put soe checks under the umbrella of a single var.

CMakeLists.txt
msimberg: I know, and it means that users who have been using `HPX_WITH_NETWORKING=OFF` now get quite a lot of additional features disabled without them noticing (well, they'll notice the configuration errors because they need to disable other things as well when `HPX_WITH_DISTRIBUTED_RUNTIME=OFF`).

biddisco:
> `HPX_WITH_DISTRIBUTED_RUNTIME` _can_ be `ON` even when `HPX_WITH_NETWORKING` is `OFF` (edit: to clarify, `HPX_WITH_NETWORKING` cannot be `ON` if `HPX_WITH_DISTRIBUTED_RUNTIME` is `OFF`). The dependency is the other way around. As I said in [#4724 (comment)](https://github.com/STEllAR-GROUP/hpx/pull/4724#discussion_r436514992) a renaming may be appropriate.

If networking is off, what type of distributed runtime do you plan on using?

CMakeLists.txt
```diff
@@ -454,9 +456,9 @@ if(HPX_WITH_IO_COUNTERS)
       "HPX_WITH_IO_COUNTERS was set to ON, but IO counters are only available on Linux (this is \"${CMAKE_SYSTEM_NAME}\")"
     )
   endif()
-  if(NOT HPX_WITH_DISTRIBUTED_RUNTIME)
+  if(NOT HPX_HAVE_DISTRIBUTED_RUNTIME)
     hpx_error(
-      "HPX_WITH_IO_COUNTERS was set to ON, but performance counters cannot be used with HPX_WITH_DISTRIBUTED_RUNTIME=OFF"
+        "HPX_WITH_IO_COUNTERS was set to ON, but performance counters cannot be used with HPX_WITH_NETWORKING=OFF or HPX_HAVE_DISTRIBUTED_RUNTIME=OFF"
```
msimberg, line 461: This is not true. They can be used with `HPX_WITH_NETWORKING=OFF`. Same below for PAPI.

CMakeLists.txt
biddisco: Well, I don't think we need to have both any more then. Why not just have `HPX_WITH_DISTRIBUTED_RUNTIME` and not bother with the networking option any more. 
If distributed runtime now trumps networking, that's fine. we can just use that.

libs/async_mpi/include/hpx/async_mpi/mpi_future.hpp
```diff
@@ -14,6 +14,7 @@
 #include <hpx/modules/threading_base.hpp>
 #include <hpx/mpi_base/mpi.hpp>
 #include <hpx/runtime_local/thread_pool_helpers.hpp>
+#include <mpi.h>
```
msimberg, line 17:
```suggestion
#include <mpi_base/mpi.hpp>
```
`mpi.hpp` includes `mpi.h` but disables warnings that come otherwise from `mpi.h`.

CMakeLists.txt
msimberg: This would need a bit more discussion in that case. I would be happy with that, but I don't know if there are others who still see `HPX_WITH_NETWORKING=OFF` with components and actions as useful. Like I said `HPX_WITH_DISTRIBUTED_RUNTIME=OFF` at the moment also means disabling many other things that a `HPX_WITH_NETWORKING=OFF` user might still want. Naming is clearly unclear here, but I think it might be worth still having the two levels of options, whatever they are called.

biddisco:
The comments in this thread are all a bit jumbled up in order. Can I summarize by saying that the simplest option is to just remove hpx_with_networking - it doesn't seem to have any use any more.

CMakeLists.txt
biddisco: `HPX_WITH_NETWORKING=OFF` was added principally so that I could work on cholesky a few years back without having the distributed runtime. We were the ones asking for it.

msimberg: As I said, I support this but I think you'd be disappointed at the moment by what's supported by `HPX_WITH_DISTRIBUTED_RUNTIME=OFF`. APEX isn't currently even supported. Some things are easier to fix than others. This tracks things that could potentially be made to work with the local-only runtime: https://github.com/STEllAR-GROUP/hpx/projects/5?add_cards_query=is%3Aopen.

hkaiser:
> It's common practice to have WITH and HAVE as cmake vars. it just makes the scripts a bit simpler because you don't have so many if this and that or that and those type clauses in later cmake files as you can put soe checks under the umbrella of a single var.

I understand. So far we have not used both, `WITH` and `HAVE` as CMake variables. For this reason I'd like for others to voice their opinion on this.

biddisco:
I've reduced this PR to just the first commit that addresses the wrong lib name issue. 
#4727  solves the other problem

msimberg:
Hmm, looks like this was badly broken before. Would you mind changing all the `mpi.h` includes to `hpx/mpi_base/mpi.hpp` and adding `Mpi::mpi` as a (conditional) dependency to `hpx_core` (in `src/CMakeLists.txt`). We currently can't propagate external dependencies of modules nicely to the main library.

msimberg:
Something else is also wrong because none of the CI builds with MPI support but still the `mpi_base` module is built.

biddisco:
someone changed runtime modes and now it is broken anyway
```
pop-os:~/build/hpx-debug$ mpiexec -n 2 bin/mpi_ring_async_executor_test
  what():  Attempted to start the runtime in the mode "console", but HPX was compiled with HPX_WITH_DISTRIBUTED_RUNTIME=OFF, and "console" requires HPX_WITH_DISTRIBUTED_RUNTIME=ON. Recompile HPX with HPX_WITH_DISTRIBUTED_RUNTIME=ON or change the runtime mode.: HPX(invalid_status)
```
I tried changing it to use runtime mode = local, but it didn't help, so I will spend some time on this now.

biddisco:
I have fixed the runtime mode problem, and am looking at the other broader mpi dependencies.
I am going to close this PR and move the patches over to the mpi_fixes PR at #4663

PR #4726, opened 2020-06-08
Add more API headers

msimberg:
Adds the remaining local concurrency and parallelism headers in #4626 (except for `spmd_block.hpp` which I realized is not in the parallelism TS).


libs/functional/include/hpx/functional.hpp
```diff
@@ -6,4 +6,25 @@
...
+#include <hpx/functional/invoke.hpp>
+#include <hpx/functional/invoke_fused.hpp>
+#include <hpx/functional/mem_fn.hpp>
+#include <hpx/functional/traits/is_bind_expression.hpp>
+#include <hpx/functional/traits/is_placeholder.hpp>
+
+namespace hpx {
+    using hpx::util::bind_back;
+    using hpx::util::bind_front;
+    using hpx::util::function;
```
msimberg, line 22: https://github.com/STEllAR-GROUP/hpx/pull/4707#issuecomment-639445396 and https://github.com/STEllAR-GROUP/hpx/pull/4707#issuecomment-639447936 for reference.

I really don't mind too much which way we go here. Most other utility classes have nonintrusive serialization so there's no need to choose. `any` is in a similar situation as `function`, with intrusive serialization. I'm happy to go with the majority opinion here.

libs/functional/CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ set(functional_headers
     hpx/functional/traits/is_bind_expression.hpp
     hpx/functional/traits/is_callable.hpp
     hpx/functional/traits/is_placeholder.hpp
+    hpx/type_traits.hpp
```
msimberg, line 49: This doesn't need to live here forever. For now it only contains `result_of` so it's enough to keep it in the `functional` module.

msimberg:
Added `is_bind_expression` and `is_placeholder` to the `hpx/functional.hpp` header.

libs/serialization/include/hpx/any.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+#pragma once
+
+#include <hpx/datastructures/any.hpp>
+#include <hpx/serialization/serializable_any.hpp>
+
+namespace hpx {
+    using hpx::util::any;
+    using hpx::util::bad_any_cast;
+    using hpx::util::make_any;
```
hkaiser, line 15: Should the local-only (non-serializable) versions here as well? There are also things like `unique_any`.

msimberg: That's the question. The same applies to `function` as well. @sithhell, @biddisco, @K-ballo, @aurianer preferences? Do we put only one serializable and copyable name corresponding to the standard library name in `hpx` or all the variations as well? I think we can go with all of them in `hpx` since they mostly share implementations, and it's unlikely we'd want to remove any of the variants (except perhaps change names).

hkaiser: If we want to go for the 'standard' version only we should use `hpx::util::any_nonser<>`, if we decided to have all of them, well then `hpx::any` should still be an alias for the non-serializable variant, while `hpx::distributed::any` should refer to the serializable one.

msimberg: There's the same problem here as with checkpointing and resiliency. Serialization is usually used with distributed applications, but not in itself a distributed feature. There's also the option of making `hpx::any` the same as `hpx::util::any_nonser` and making `hpx::any_ser` (or `any_serializable`) the same as `hpx::util::any` i.e. flipping which variant gets the extra "annotation" in its name.

hkaiser: Ohh my... so many options. Like a kid in a candy shop ;-)

msimberg: Well, yes. I'd just like to avoid having to change the names in the next release again.

Is there actually a runtime cost to using the serializable variants of `any` or `function`, or is it purely compile-time? @K-ballo?

hkaiser: Should be pure compile-time costs, the only runtime cost is a slight increase in the size of the (manually constructed) vtable.

msimberg: Ok... so why don't we just do "no decision" and keep the same names as in `hpx::util` for both `any` and `function`. `tuple`, `optional` etc. are serializable as well (just nonintrusively). If we ever have something like a `HPX_WITH_SERIALIZATION` option we can make `hpx::any/function` be the nonserializable variants when the option is off.

hkaiser: @msimberg let's do that. Since we can't make up our minds, leaving things as is might be the best way forward...

PR #4729, opened 2020-06-08
Fix APEX build

msimberg:


tests/unit/apex/annotation_check.cpp
```diff
@@ -11,25 +11,25 @@
 
 #include <hpx/hpx_init.hpp>
 
-#include <hpx/modules/async_distributed.hpp>
 #include <hpx/include/parallel_execution.hpp>
 #include <hpx/include/threads.hpp>
-#include <hpx/lcos/when_all.hpp>
+#include <hpx/modules/async_base.hpp>
+#include <hpx/modules/async_combinators.hpp>
+#include <hpx/modules/async_distributed.hpp>
```
biddisco, line 18: why is distributed needed here? I think I removed any action tests, so what gets pulled in that the distributed include provides?

msimberg: If you say it's not needed, it's probably not needed. I just sorted the includes. I'll look through the file and see if it's needed for anything.

msimberg: Thanks John for pointing this out. It was indeed not needed. I also run the whole file through clang-format since the changes weren't so big.

PR #4733, opened 2020-06-09
Add experimental unity build option to CMake configuration

msimberg:
This adds a CMake option to use CMake's unity build functionality from 3.16 onwards (https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html). Essentially this batches compilation of source files to save a bit (or a lot) of header parsing. Internally CMake batches up to 8 (default) source files into one file that looks roughly like this:
```
#include "/path/to/source1.cpp"
#include "/path/to/source2.cpp"
#include "/path/to/source3.cpp"
...
```
and compiles that instead. The downside of this is that files that are batched together can't have duplicate static definitions. For `on_self_reset` in the `thread_executors` module I just lifted the definitions out into a separate header file. If needed the batching can be turned off for specific source files as well. This can also lead to slightly increased build times when only a few files need to be recompiled since the whole batch needs to be recompiled rather than just one or two files.

When `HPX_WITH_UNITY_BUILD=ON` the modules and main HPX library have the `UNITY_BUILD` property set to on. With the fixes here it works fine, and most importantly a single job build of the main library went from about 34 minutes to 16 minutes on my laptop.

Please give it a try. If this works well for a longer time we can consider always building with `UNITY_BUILD`.

### DIFF
--- a/libs/thread_executors/CMakeLists.txt
+++ b/libs/thread_executors/CMakeLists.txt
@@ -15,6 +15,7 @@ set(thread_executors_headers
     hpx/execution/executors/thread_pool_os_executors.hpp
     hpx/thread_executors/current_executor.hpp
     hpx/thread_executors/default_executor.hpp
+    hpx/thread_executors/detail/on_self_reset.hpp
     hpx/thread_executors/embedded_thread_pool_executors.hpp
     hpx/thread_executors/executors.hpp
     hpx/thread_executors/force_linking.hpp
--- a/src/runtime/agas/server/route.cpp
+++ b/src/runtime/agas/server/route.cpp
@@ -27,17 +27,17 @@
 #include <utility>
 #include <vector>
 
-
-namespace {
+namespace hpx { namespace detail {
     void update_agas_cache(hpx::naming::gid_type const& gid,
         hpx::naming::address const& addr, std::uint64_t count,
         std::uint64_t offset)
     {
-        hpx::naming::get_agas_client().update_cache_entry(gid, addr, count, offset);
+        hpx::naming::get_agas_client().update_cache_entry(
+            gid, addr, count, offset);
     }
-}
+}}    // namespace hpx::detail
 
-HPX_PLAIN_ACTION_ID(update_agas_cache, update_agas_cache_action,
+HPX_PLAIN_ACTION_ID(hpx::detail::update_agas_cache, update_agas_cache_action,
     hpx::actions::update_agas_cache_action_id)
 
 namespace hpx { namespace agas { namespace server
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -331,6 +331,17 @@ else()
   )
 endif()
 
+hpx_option(
+  HPX_WITH_UNITY_BUILD
+  BOOL
+  "Enable unity build for certain build targets (experimental, requires CMake 3.16 or newer) (default OFF)"
+  OFF
+  ADVANCED
+)
+if(HPX_WITH_UNITY_BUILD AND (CMAKE_VERSION VERSION_LESS "3.16"))
+  hpx_error("HPX_WITH_UNITY_BUILD=ON but this requires CMake 3.16 or newer.")
+endif()
+
 # ##############################################################################
 # Some platforms do not support dynamic linking. Enable this to link all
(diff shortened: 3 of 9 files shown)


src/runtime/agas/server/route.cpp
```diff
@@ -27,17 +27,17 @@
...
     void update_agas_cache(hpx::naming::gid_type const& gid,
         hpx::naming::address const& addr, std::uint64_t count,
         std::uint64_t offset)
     {
-        hpx::naming::get_agas_client().update_cache_entry(gid, addr, count, offset);
+        hpx::naming::get_agas_client().update_cache_entry(
+            gid, addr, count, offset);
     }
-}
+}}    // namespace hpx::detail
```
msimberg, line 38: Was there a particular reason for keeping this in the anonymous namespace? Without this change I get
```
/home/simbergm/src/hpx/src/runtime/agas/server/route.cpp:40:40: warning: 'update_agas_cache_action' has a base 'hpx::actions::action<void (*)(c
onst hpx::naming::gid_type&, const hpx::naming::address&, long unsigned int, long unsigned int), {anonymous}::update_agas_cache, update_agas_ca
che_action>' whose type uses the anonymous namespace [-Wsubobject-linkage]                                                                     
   40 | HPX_PLAIN_ACTION_ID(update_agas_cache, update_agas_cache_action,                                                                       
      |                                        ^~~~~~~~~~~~~~~~~~~~~~~~                                                                        
../../libs/preprocessor/include/hpx/preprocessor/expand.hpp:23:26: note: in definition of macro 'HPX_PP_EXPAND'                                
   23 | #define HPX_PP_EXPAND(X) X                                                                                                             
      |                          ^                                     
../../libs/preprocessor/include/hpx/preprocessor/cat.hpp:36:28: note: in expansion of macro 'HPX_DEFINE_PLAIN_ACTION_2'                        
   36 | #define HPX_PP_CAT_I(a, b) a##b                                
      |                            ^                                   
../../hpx/runtime/actions/plain_action.hpp:187:5: note: in expansion of macro 'HPX_DEFINE_PLAIN_ACTION_'                                       
  187 |     HPX_DEFINE_PLAIN_ACTION_(__VA_ARGS__)                                     \                                                        
      |     ^~~~~~~~~~~~~~~~~~~~~~~~                                   
../../hpx/runtime/actions/plain_action.hpp:342:5: note: in expansion of macro 'HPX_DEFINE_PLAIN_ACTION'                                        
  342 |     HPX_DEFINE_PLAIN_ACTION(func, name);                                      \
      |     ^~~~~~~~~~~~~~~~~~~~~~~                                    
/home/simbergm/src/hpx/src/runtime/agas/server/route.cpp:40:1: note: in expansion of macro 'HPX_PLAIN_ACTION_ID'                               
   40 | HPX_PLAIN_ACTION_ID(update_agas_cache, update_agas_cache_action,                                                                       
      | ^~~~~~~~~~~~~~~~~~~                                            
```
but I'm not sure if I've messed something up by doing this (it compiles, but I haven't tested it).

I can also instead of this change mark the file to be compiled alone even if `UNITY_BUILD` is enabled.

libs/thread_executors/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ set(thread_executors_headers
     hpx/execution/executors/thread_pool_os_executors.hpp
     hpx/thread_executors/current_executor.hpp
     hpx/thread_executors/default_executor.hpp
+    hpx/thread_executors/detail/on_self_reset.hpp
```
hkaiser, line 18: Did you forget to add this file?

msimberg: Yep... I'll add it.

hkaiser:
This works well for me! Thanks for doing this.

src/runtime/agas/server/route.cpp
hkaiser: There is no particular reason.

msimberg:
> This works well for me! Thanks for doing this.

Great! Let's give this a go then. It doesn't affect anything else since it's off by default (except for that one change to `update_agas_cache`).

By the way, it looks like codacy is now happy to run on this PR as well. Maybe it was there all along...

PR #4734, opened 2020-06-09
Fixing a bunch of codacy warnings

hkaiser:


examples/quickstart/factorial.cpp
```diff
@@ -26,8 +26,8 @@ HPX_PLAIN_ACTION(factorial, factorial_action);
 ///////////////////////////////////////////////////////////////////////////////
 std::uint64_t factorial(std::uint64_t n)
 {
-    if (0 >= n)
-        return 1;
+    if (n < 2)
+        return n;
```
msimberg, line 30: This isn't right if `n = 0`. It should still always return `1`.

examples/quickstart/hello_world_distributed.cpp
```diff
@@ -47,7 +47,7 @@ std::size_t hello_world_worker(std::size_t desired)
         char const* msg = "hello world from OS-thread {1} on locality {2}\n";
 
         hpx::util::format_to(hpx::cout, msg, desired, hpx::get_locality_id())
-            << hpx::flush;
+            << std::flush;
```
msimberg, line 50: Is this something we should enforce more generally (or maybe this is the last use of `hpx::flush`? I doubt it...)? And perhaps mark `hpx::flush` as deprecated?

libs/runtime_local/src/runtime_local.cpp
```diff
@@ -952,9 +952,8 @@ namespace hpx {
             return false;
         }
 
-        bool numa_sensitive = false;
         if (std::size_t(-1) != get_worker_thread_num())
-            return numa_sensitive;
+            return false;
```
msimberg, line 956: Hmm, does this function ever return anything but `false`? Either this function should query the default/current scheduler for the value or the function shouldn't exist at all. Does not need to be changed in this PR.

examples/quickstart/factorial.cpp
hkaiser: This should never be called with `n == 0`, I think. In any case, the old test was wrong, definitely.

hkaiser:
> Couple of questions: I've received more emails from codacy lately. Did something change in the codacy setup? 

I had to reinstall the codacy hook as they have changed things on their end. That could have caused the changes.

> Do you know if there's an option to check pull requests from forked repos as well? I think some of these have snuck in from my changes.

I don't know that, sorry.

examples/quickstart/hello_world_distributed.cpp
hkaiser: Yes, I think we should deprecate `hpx::flush` and all the other manipulators we have implemented. They are not needed anymore (and have not been for a long time).

libs/runtime_local/src/runtime_local.cpp
hkaiser: I didn't look at the semantics here. Just removed the variable that was not needed anymore.

hkaiser:
> > Do you know if there's an option to check pull requests from forked repos as well? I think some of these have snuck in from my changes.
> 
> I don't know that, sorry.

Update: you might be able to separately integrate your repository with codacy.

examples/quickstart/factorial.cpp
hkaiser: I also think we do the same for all the other fibonacci examples.

msimberg:
> > > Do you know if there's an option to check pull requests from forked repos as well? I think some of these have snuck in from my changes.
> > 
> > 
> > I don't know that, sorry.
> 
> Update: you might be able to separately integrate your repository with codacy.

No worries. Yes, I can definitely do that. That'll at least work for my PRs.

examples/quickstart/factorial.cpp
msimberg: `0! = 1` is perfectly valid. Not a very interesting input for a parallel `factorial` but it's definitely valid input. The zeroth fibonacci number is actually `0` so it makes sense there.

examples/quickstart/hello_world_distributed.cpp
msimberg: https://github.com/STEllAR-GROUP/hpx/issues/4736

libs/runtime_local/src/runtime_local.cpp
msimberg: No worries, let's not mix it into this PR.

examples/quickstart/factorial.cpp
hkaiser: > The zeroth fibonacci number is actually 0 so it makes sense there.

Ok, then `if (n < 2) return n;` is actually correct?

msimberg: > Ok, then `if (n < 2) return n;` is actually correct?

Well, yes, for fibonacci. This is factorial ;)

hkaiser: Uhh, right. Not sure what I was thinking...

hkaiser:

PR #4737, opened 2020-06-09
Allow hpxrun.py to run on remote nodes with tcp parcelport

stevenrbrandt:
hpxrun.py is a useful command, but does not allow running on remote localities.

## Proposed Changes

specify -n nodes to allow running on remote locations for tcp parcelport.

### DIFF
--- a/cmake/templates/hpxrun.py.in
+++ b/cmake/templates/hpxrun.py.in
@@ -26,6 +26,7 @@ import functools, signal
 import traceback
 import time
 import threading
+import re
 
 from optparse import OptionParser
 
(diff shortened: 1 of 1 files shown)


biddisco:
NB. If a small note/doc could be added somewhere to document hpxrun (or if it exists already - could it be extended?) - to briefly show an example command-line. Much of this stuff is not documented anywhere and when someone makes a change like this it would be great if someone familiar with the scripts could take 5 mins to add to the docs to save the next person a bit of time.

msimberg:
Thanks @stevenrbrandt! The pycicle clang-oldest failure is unrelated.

PR #4741, opened 2020-06-11
Fixing a warning generated for unity builds with msvc

hkaiser:
- apparently the build system doesn't like having files with the
  same names in the same module


msimberg:
Thanks! I think you'll need to update the file lists in `libs/format/CMakeLists.txt`...

PR #4744, opened 2020-06-12
Fork-join executor

msimberg:
Adds a fork-join executor `hpx::execution::experimental::fork_join_executor`. This implements what I had in mind in https://github.com/STEllAR-GROUP/hpx/issues/3348#issuecomment-624506798.

This is an always-blocking executor, that creates worker (HPX-)threads for the lifetime of the executor that never yield until the executor is destructed. This is the very extreme end of this idea and it does of course mean that no other work will run while this executor is alive.

There are a bunch of todos at the top of some of the new files that I still want to look into.

The good news is that this is enough to get performance on par with Kokkos, at least on stream. The real question will be if we can find something in between this and our current executors which has almost the same performance as this. Here's a graph of some stream benchmarks I did earlier this week: (best viewed in full-screen; Piz Daint dual 18-core Broadwells, as usual; no hyperthreading; compact thread placement; the `fork_join_executor` keeps up with Kokkos all the way down to 1024 elements, but the performance of both is obviously bad):
tmp


msimberg:
I've updated this to a state where I think it would already be useful for someone else. The executor now yields after a customizable delay of no work (1 ms by default) so that it can be used at the same time as other work is being scheduled. I'd especially appreciate a pair or two of eyes on the queue.

The big remaining issue I have is that the algorithms rely only on `bulk_async_execute` whereas this executor should only expose `bulk_sync_execute`. I've currently implemented `bulk_async_execute` with blocking semantics, and return a(n empty) vector of futures when the parallel region is done.

msimberg:
As discussed in our previous meeting, the fork-join executor is synchronous but is both `is_bulk_one_way_executor` and `is_bulk_two_way_executor` to work with the parallel algorithms. `bulk_async_execute` is blocking and returns an empty vector of futures (one exceptional future if there's an exception).

I think this is functionally complete, but I'd appreciate reviews of it.

aurianer (dismissed):
LGTM thanks!

hkaiser:
@msimberg thanks for moving this forward!

PR #4748, opened 2020-06-12
Create coverage reports and upload them to codecov.io

haampie:
Now that https://github.com/STEllAR-GROUP/docker_build_env/pull/25 is merged, this PR generates coverage reports and uploads them to codecov.io. I believe codecov will magically accept reports without any other config. 

On my fork of hpx circle ci timed out, so we have to see here if it works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,17 @@ move_debug_log: &move_debug_log
         mkdir -p ${CIRCLE_JOB}
         cp debug-log.txt ${CIRCLE_JOB} || true
 
+collect_coverage: &collect_coverage
+    name: Collecting coverage reports
+    command: |
+        mkdir -p /hpx/coverage
+        lcov -q --gcov-tool llvm-cov-wrapper --no-external --capture --base-directory /hpx/source/ --directory /hpx/build/ --output-file /hpx/coverage/${CIRCLE_JOB}.info
+
+persist_coverage: &persist_coverage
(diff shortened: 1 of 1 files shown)


hkaiser:
codacy accepts coverage reports as well and we already let it crawl our code base. Might be a good idea to make it aware of the new coverage results as well.

haampie:
Thanks, that sounds useful indeed.

Seems like circle ci is not triggered, maybe because this is a pull request opened from a fork?

hkaiser:
> Seems like circle ci is not triggered, maybe because this is a pull request opened from a fork?

Let's merge this, then. If something goes wrong we'll fix it.

hkaiser:
While there are unrelated problems, I'm at least seeing this:
https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/4083/workflows/68e0fecc-c6b7-4b90-b86f-6e56b3b7525c/jobs/201680

@haampie Could you have a look, please?

haampie:
Ah, sorry, it was not meant to be merged yet, I was just hoping that I could test CI by opening a pull request.

`lcov` seems to be a ridiculously slow tool for the job, probably we should use https://github.com/mozilla/grcov instead.

For now maybe best to revert the pr. I can test it more in my fork.

hkaiser:
FWIW, it might be better to create a dedicated GitHub action for this (instead of adding it to the CircleCI builder).

PR #4750, opened 2020-06-13
Fixing possible race condition during termination detection

hkaiser:
- flyby: excluding distributed termination detection from compilation
  if networking is disabled
- flyby: remove (obsolete) fast termination detection code


msimberg:
What was the possible race and which part of the changes actually fixes that?

hkaiser:
> What was the possible race and which part of the changes actually fixes that?

@msimberg it was probably a minor (and inconsequential) race here: https://github.com/STEllAR-GROUP/hpx/pull/4750/files#diff-82b7088989ee9c24570d6f0b34844db6R483. The flag `dijkstra_color_` could have been set to `true` after the lock was relinquished (L481), but before `send_dijkstra_termination_token` was invoked (L483).

PR #4759, opened 2020-06-16
Reduce memory requirements for our main shared state

hkaiser:
- flyby: add missing headers to CMakeLists.txt

### DIFF
--- a/libs/futures/CMakeLists.txt
+++ b/libs/futures/CMakeLists.txt
@@ -11,6 +11,8 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 set(futures_headers
     hpx/futures/future.hpp
     hpx/futures/futures_factory.hpp
+    hpx/futures/detail/future_data.hpp
+    hpx/futures/detail/future_transforms.hpp
     hpx/futures/packaged_continuation.hpp
     hpx/futures/traits/acquire_future.hpp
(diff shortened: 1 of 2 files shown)


hkaiser:
> The failure on CircleCI is unrelated (fixed in #4761), and the GH actions is due to the dependencies not downloading (I hope we didn't hit a bandwidth limit again?). I think we're safe to merge this.

I just retriggered those builders. Should go through this time.

msimberg:
By the way, I asked the linear algebra people at CSCS to check if this had any effect in their library (as they definitely have more than one continuation on most of their futures, *but* through dataflow which might make the comparison mostly useless). In any case, it seems like it had no negative effect and potentially small positive effect (but mostly within noise).

libs/futures/CMakeLists.txt
```diff
@@ -54,6 +56,8 @@ add_hpx_module(
   SOURCES ${futures_sources}
   HEADERS ${futures_headers}
   COMPAT_HEADERS ${futures_compat_headers}
+  EXCLUDE_FROM_GLOBAL_HEADER "hpx/futures/detail/future_data.hpp"
```
aurianer, line 59: Not necessary anymore since we exclude all detail headers from the global one in [HPX_AddModule](https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_AddModule.cmake#L134)

msimberg: Could you open a PR to remove that, please?

PR #4760, opened 2020-06-16
Storing hpx_options in global property to speed up summary report

### LINKED ISSUE #4758: CMake profiling results
CMake V3.18 can output profiling traces. The results show that about 40% of the time is spent in configuring and about 60% of the time is spent generating the make files. The interesting part is that during generation, about 50% of the time is spent in the function `create_configarion_summary`, and about 20-30% are spent during generating the force_linking header and source files (will go away soon). Here is an exemplar visualization for one of the HPX modules:

image

that demonstrates this nicely. I think we may have some low-hanging fruit here allowing to significantly reducing the generation phase of our build system.

hkaiser:
Fixes #4758 

As it turns out, CMake profiling reports only data for the configure step (not for the generation step). For this reason, the change proposed here improves only the configure step by reducing the amount of time required to create the configuration summary for each module. See here for the new trace:

image

which demonstrates the reduction nicely (if compared to #4758).

Overall timing: the configure step went down from 49s to 28s for me.

### DIFF
--- a/.cmake-format.py
+++ b/.cmake-format.py
@@ -493,7 +493,7 @@ with section("parse"):
     'hpx_include': {'pargs': {'nargs': 0}},
     'hpx_info': {'pargs': {'nargs': 0}},
     'hpx_message': {'pargs': {'nargs': 1}},
-    'hpx_option': { 'kwargs': {'CATEGORY': 1, 'STRINGS': '+'},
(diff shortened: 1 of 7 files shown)


msimberg:
I'm going to assume the header test failure on CircleCI is unrelated (https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/4132/workflows/cc886efe-0a0d-4a2a-badd-a0a5727153c9/jobs/203019/steps, it's a timeout...).

PR #4761, opened 2020-06-17
Fix running and building of execution module tests on CircleCI

msimberg:
The regexes didn't quite work after the renaming in #4740.


aurianer:
Thanks a lot!

PR #4762, opened 2020-06-17
Update pointees in compatibility headers

### LINKED ISSUE #4639: Update deprecation warnings in compatibility headers to point to collective headers
They should be updated to point to the headers defined in #4626 if possible, or `hpx/include/` headers.

msimberg:
Fixes #4639. Points the compatibility headers to the new public API headers where possible, next to `hpx/include/`, and last to `hpx/modules/`. Also removes `detail` compatibility headers. Goes in after #4385.

Split up into separate commits just to allow more easily dropping changes if someone doesn't like them. Will squash this into a just a few commits before merging.


msimberg:
Rebased and squashed most commits. Let's see what's broken...

msimberg:
By the way, I've ignored the compatibility headers in the circular dependency checks as many of them now point to the `include` module.

hkaiser:
> By the way, I've ignored the compatibility headers in the circular dependency checks as many of them now point to the include module.

Is it a good idea to point the user towards the include module? Is that module here to stay after all?

msimberg:
> > By the way, I've ignored the compatibility headers in the circular dependency checks as many of them now point to the include module.
> 
> Is it a good idea to point the user towards the include module? Is that module here to stay after all?

It might not be a good idea. And no, it's not here to stay. However, I think they might still be a better alternative than pointing directly to `hpx/modules` headers until 2.0.0 simply because their contents are very stable and work in a lot of versions of HPX. They mostly work as an intermediate until we have all the API headers in place to replace them. Modules that I point to `hpx/include`:

- `async_distributed`
- `collectives`
- `executors`
- `lcos_distributed`
- `lcos_local`
- `resource_partitioner`
- `thread_executors`
- `threadmanager`

The compatibility headers that have a new API header don't get forwarded to `hpx/include`, e.g. `hpx/parallel/executors/parallel_executor.hpp` in `executors` gets forwarded to `hpx/execution.hpp`.

So, yes, it's a temporary solution, but I think it's the gentlest way forward for users. We should train them off from including deeply nested headers anyway.

PR #4764, opened 2020-06-18
Force some tests to run serially

msimberg:
Forces some tests to run serially if `ctest` is given the `-j` option.

I've set all distributed tests, benchmarks, and tests that request `all` or `cores` threads. I've been a bit conservative and set the compile tests to run serially as well as I've seen problems before trying to build multiple header tests concurrently.

This would allow us to run tests much more quickly on hardware that has spare resources (most tests run with 1 or 4 threads).

@haampie I'm currently running the tests to see if there are any unexpected results. The option does work as advertised, I just have to make sure I haven't missed any other tests that require running serially.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1356,6 +1356,14 @@ hpx_option(
   CATEGORY "Debugging" ADVANCED
 )
 
+hpx_option(
+  HPX_WITH_PARALLEL_TESTS_BIND_NONE
+  BOOL
+  "Pass --hpx:bind=none to tests that may run in parallel (cmake -j flag) (default: OFF)"
+  OFF
+  CATEGORY "Debugging" ADVANCED
+)
+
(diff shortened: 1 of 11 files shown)


msimberg:
This should be ready now. I added an additional option which adds `--hpx:bind` to tests that are allowed to run in parallel, to avoid having all of them bind to the first few cores. I also reduced the running times of some of the longer running tests. #4772 also came up while doing this. I've forced the affected test (`parallel_fill_4132`) to run serially (i.e. without `--hpx:bind=none`) to work around the problem in this PR.

hkaiser:
Could this fix the transient test errors we see for partitioned_vector?

msimberg:
> Could this fix the transient test errors we see for partitioned_vector?

I very much doubt it. We don't actually run tests anywhere in parallel. This is meant to allow us to run some tests in parallel.

PR #4768, opened 2020-06-18
Add sized_sentinel_for concept and make distance dispatch according to that

gonidelis:
# Fixes 

- Added `sized_sentinel_for` on `is_sentinel_for.hpp`
- Added the `disable_sized_sentinel_for` feature check
- Adjusted `hpx::parallel::v1::detail::distance()` in order to decide for the type dispatching according to `sized_sentinel_for`
- Added test for the adapted `distance` implementation

## To do

  - [x] Add unit tests for `sized_sentinel_for`
  - [x] Add `remove_cv_t` trait on `disable_sized_sentinel_for` inside the `sized_sentinel_for` implementation (for some reason it would not compile when I add it)
  - [ ] Please inform me if there needs to be a test for `disable_sized_sentinel_for`
  - [x] Please inform me whether the test for `distance` is under the correct directory (`distance.hpp` lies on `algorithms/detail`)

### DIFF
--- a/cmake/HPX_AddConfigTest.cmake
+++ b/cmake/HPX_AddConfigTest.cmake
@@ -3,6 +3,7 @@
 # Copyright (c) 2017 Denis Blank
 # Copyright (c) 2017 Google
 # Copyright (c) 2017 Taeguk Kwon
+# Copyright (c) 2020 Giannis Gonidelis
 #
 # SPDX-License-Identifier: BSL-1.0
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
@@ -446,10 +447,10 @@ function(hpx_check_for_cxx20_coroutines)
 endfunction()
 
 # ##############################################################################
-function(hpx_check_for_cxx20_disable_sized_sentinel_for)
+function(hpx_check_for_cxx20_std_disable_sized_sentinel_for)
   add_hpx_config_test(
-    HPX_WITH_CXX20_DISABLE_SIZED_SENTINEL_FOR SOURCE cmake/tests/cxx20_disable_sized_sentinel_for.cpp FILE
-                                     ${ARGN}
+    HPX_WITH_CXX20_STD_DISABLE_SIZED_SENTINEL_FOR
+    SOURCE cmake/tests/cxx20_std_disable_sized_sentinel_for.cpp FILE ${ARGN}
   )
 endfunction()
 
--- a/cmake/HPX_AddConfigTest.cmake
+++ b/cmake/HPX_AddConfigTest.cmake
@@ -446,6 +446,14 @@ function(hpx_check_for_cxx20_coroutines)
 endfunction()
 
 # ##############################################################################
+function(hpx_check_for_cxx20_disable_sized_sentinel_for)
+  add_hpx_config_test(
+    HPX_WITH_CXX20_DISABLE_SIZED_SENTINEL_FOR SOURCE cmake/tests/cxx20_disable_sized_sentinel_for.cpp FILE
+                                     ${ARGN}
+  )
+endfunction()
+
+# ##############################################################################
 function(hpx_check_for_builtin_integer_pack)
   add_hpx_config_test(
     HPX_WITH_BUILTIN_INTEGER_PACK SOURCE cmake/tests/builtin_integer_pack.cpp
--- a/cmake/HPX_PerformCxxFeatureTests.cmake
+++ b/cmake/HPX_PerformCxxFeatureTests.cmake
@@ -2,6 +2,7 @@
 # Copyright (c) 2011-2014 Thomas Heller
 # Copyright (c) 2013-2016 Agustin Berge
 # Copyright (c)      2017 Taeguk Kwon
+# Copyright (c)      2020 Giannis Gonidelis
 #
 # SPDX-License-Identifier: BSL-1.0
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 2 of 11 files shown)


cmake/HPX_AddConfigTest.cmake
```diff
@@ -445,6 +446,14 @@ function(hpx_check_for_cxx20_coroutines)
   )
 endfunction()
 
+# ##############################################################################
+function(hpx_check_for_cxx20_disable_sized_sentinel_for)
+  add_hpx_config_test(
+    HPX_WITH_CXX20_DISABLE_SIZED_SENTINEL_FOR
+    SOURCE cmake/tests/cxx20_disable_sized_sentinel_for.cpp FILE ${ARGN}
```
hkaiser, line 453: Please name this file `cxx20_std_disable_sized_sentinel_for.cpp`.

cmake/HPX_PerformCxxFeatureTests.cmake
```diff
@@ -74,6 +75,8 @@ function(hpx_perform_cxx_feature_tests)
   # C++20 feature tests
   hpx_check_for_cxx20_coroutines(DEFINITIONS HPX_HAVE_CXX20_COROUTINES)
 
+  hpx_check_for_cxx20_disable_sized_sentinel_for(DEFINITIONS HPX_HAVE_CXX20_DISABLE_SIZED_SENTINEL_FOR)
```
hkaiser, line 78: Same here, please rename to `HPX_HAVE_CXX20_STD_DISABLE_SIZED_SENTINEL_FOR`. Also cmake-format this file.

libs/iterator_support/include/hpx/iterator_support/traits/is_sentinel_for.hpp
```diff
@@ -0,0 +1,61 @@
...
+    constexpr bool disable_sized_sentinel_for = false;
+#endif
+
+    template <typename Sent, typename Iter, typename Enable = void>
+    struct sized_sentinel_for : std::false_type
+    {
+    };
+
+    template <typename Sent, typename Iter>
+    struct sized_sentinel_for<Sent, Iter,
```
hkaiser, line 49: Please rename this to `is_sized_sentinel_for` to conform to the trait naming conventions.

libs/iterator_support/include/hpx/iterator_support/traits/is_sentinel_for.hpp
```diff
@@ -0,0 +1,61 @@
...
+        typename util::always_void<
+            typename hpx::traits::is_sentinel_for<Sent, Iter>::type,
```
hkaiser, line 51: `hpx::traits::is_sentinel_for<Sent, Iter>::type` is always value (as it evaluates to either `std::true_type` or `std::false_type`). You might want to include this into the `enable_if` condition below.

libs/algorithms/include/hpx/parallel/algorithms/detail/distance.hpp
```diff
@@ -36,6 +37,6 @@ namespace hpx { namespace parallel { inline namespace v1 { namespace detail {
     distance(InIterB first, InIterE last)
     {
         return distance(first, last,
-            typename std::iterator_traits<InIterB>::iterator_category{});
+            typename hpx::traits::sized_sentinel_for<InIterE, InIterB>::type{});
```
hkaiser, line 40: This must be `is_sized_sentinel_for` now as well.

cmake/tests/cxx20_disable_sized_sentinel_for.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+#include <iterator>
+
+int main()
+{
+    constexpr bool b = std::disable_sized_sentinel_for<void, void>;
+    (void) b;
+
+    return 0;
+}
```
hkaiser, line 17: This file is not needed anymore now.

libs/iterator_support/include/hpx/iterator_support/traits/is_sentinel_for.hpp
```diff
@@ -0,0 +1,59 @@
...
+        typename util::always_void<typename is_iterator<Iter>::type,
+            typename detail::equality_result<Iter, Sent>::type,
+            typename detail::equality_result<Sent, Iter>::type,
+            typename detail::inequality_result<Iter, Sent>::type,
+            typename detail::inequality_result<Sent, Iter>::type>::type>
+      : std::true_type
+    {
+    };
+
+#if defined(HPX_HAVE_CXX20_DISABLE_SIZED_SENTINEL_FOR)
```
hkaiser, line 32: Shouldn't be this `HPX_HAVE_CXX20_STD_DISABLE_SIZED_SENTINEL_FOR` now?

libs/iterator_support/include/hpx/iterator_support/traits/is_sentinel_for.hpp
```diff
@@ -0,0 +1,59 @@
...
+    template <typename Sent, typename Iter>
+    constexpr bool disable_sized_sentinel_for =
```
hkaiser, line 34: On C++17 or above this variable could be defined `inline`. Please use `HPX_INLINE_CONSTEXPR_VARIABLE` instead of the `constexpr`, this will take care of the difference for different standard versions.

cmake/tests/cxx20_disable_sized_sentinel_for.cpp
gonidelis: I thought I deleted that. My bad.

libs/iterator_support/include/hpx/iterator_support/traits/is_sentinel_for.hpp
hkaiser: OTOH, if `HPX_HAVE_CXX20_STD_DISABLE_SIZED_SENTINEL_FOR` is defined, we can probably assume that we're on C++20, so you could write `inline constexpr` to begin with. The `HPX_INLINE_CONSTEXPR_VARIABLE` would be needed for the `#else` below only.

libs/iterator_support/include/hpx/iterator_support/traits/is_sentinel_for.hpp
```diff
@@ -0,0 +1,59 @@
...
+namespace hpx { namespace traits {
+
+    template <typename Sent, typename Iter, typename Enable = void>
+    struct is_sentinel_for : std::false_type
+    {
+    };
+
+    template <typename Sent, typename Iter>
+    struct is_sentinel_for<Sent, Iter,
+        typename util::always_void<typename is_iterator<Iter>::type,
```
gonidelis, line 23: Got it! That should have been `typename std::enable_if<is_iterator<Iter>::value>::type, `. That may cause the problems. Must have missed it on some rebase.

PR #4769, opened 2020-06-18
Adds distributed, highly scalable 1d stencil

NK-Nikunj:
## Proposed Changes

This PR adds a distributed scalable 1D stencil for benchmarking purposes. Following are the highlights of the benchmark:
  - Uses NUMA aware allocators and executors (block allocator and executor)
  - The kernel iterates a complete-time step before checking for incoming left and right neighbors leading to latency hiding.
  - Uses `number of local partitions` to parallelize locally and distribution occurs by calling `hpx_main` on all localities.

One known issue: `Nlp` should be divisible by `Nx` for the code to work as expected. Best performances are achieved when `Nlp` is set to 2x or 4x the number of cores in a single node.

The benchmark can be further optimized by having `N` iterations in a single time step. A possible future work ;)


examples/distributed_stencil/1d_stencil.cpp
```diff
@@ -0,0 +1,169 @@
+// Copyright (c) 2020 Nikunj Gupta
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
```
msimberg, line 4: You could add a short description here saying what this does differently than the other stencil examples (something like this: https://github.com/STEllAR-GROUP/hpx/blob/6172cea92789ab5e30aa815673418aebbbd04456/examples/1d_stencil/1d_stencil_8.cpp#L7-L10).

examples/distributed_stencil/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+if(NOT HPX_WITH_DISTRIBUTED_RUNTIME)
+  return()
+endif()
+
+set(example_programs 1d_stencil)
```
msimberg, line 11: I think we have to rename this. How about `1d_stencil_channel` and putting it in the `examples/1d_stencil` directory? There are distributed 1d stencil implementations in that directory as well, and it might be useful to make it clear what this one does compared to those. Having them together would make that easier.

examples/distributed_stencil/1d_stencil.cpp
```diff
@@ -0,0 +1,169 @@
...
+            // consumed in the next timestep
+            comm.set(communicator_type::right, next[Nx-1], t+1); 
+        }
+    }
+    double elapsed = t.elapsed();
+    double telapsed = t_main.elapsed();
+
+    // for (double elem : U[(steps)%2])
+    //    std::cout << elem << " ";
+    // std::cout << std::endl;
```
msimberg, line 137: Remove? Or if you'd like a verbose mode that could be behind a command line flag.

examples/distributed_stencil/1d_stencil.cpp
```diff
@@ -0,0 +1,169 @@
...
+            // consumed in the next timestep
+            comm.set(communicator_type::left, next[0], t+1); 
+        }
+
+        if (comm.has_neighbor(communicator_type::right))
+        {
+            // Synchronization step. Wait for the right neighbor to arrive.
+            double right = comm.get(communicator_type::right, t).get();
+
+            // next[Nx-1] = (right + curr[Nx-2]) * 0.5f;
```
msimberg, line 124: Remove?

examples/distributed_stencil/stencil.hpp
```diff
@@ -0,0 +1,54 @@
...
+void stencil_update(std::array<data_type, 2>& U, const std::size_t& begin
+        , const std::size_t& end, const std::size_t t)
+{
+    data_type& curr = U[t % 2];
+    data_type& next = U[(t + 1) % 2];
+
+    #pragma unroll
+    for (std::size_t i = begin; i < end; ++i)
+    {
+        // next[i] = (curr[i-1] + curr[i+1]) * 0.5f;
```
msimberg, line 51: Remove?

examples/distributed_stencil/communicator.hpp
```diff
@@ -0,0 +1,84 @@
...
+        // Get our data from our neighbor, we return a future to allow the
+        // algorithm to synchronize.
+        return recv[n].get(hpx::launch::async, step);
+    }
+
+    std::array<hpx::lcos::channel<T>, 2> recv;
+    std::array<hpx::lcos::channel<T>, 2> send;
+};
+
+#endif
```
msimberg, line 84: Do we have a convention newlines at the end of files...?

examples/1d_stencil/stencil.hpp
```diff
@@ -0,0 +1,53 @@
...
+    }
+}
+
+void stencil_update(std::array<data_type, 2>& U, const std::size_t& begin
+        , const std::size_t& end, const std::size_t t)
+{
+    data_type& curr = U[t % 2];
+    data_type& next = U[(t + 1) % 2];
+
+    #pragma unroll
```
msimberg, line 48: @NK-Nikunj this might be interesting for you: https://cdash.cscs.ch//viewBuildError.php?buildid=112799. The loop definitely looks unrollable to me but I might be missing something. The build is in release mode, so loop unrolling shouldn't be disabled because of a debug build at least. Do you know if GCC manages to unroll this (it might just not be reporting that it can't like clang)?

NK-Nikunj: Interesting. I didn't know that it was unrollable for certain builds. GCC manages to unroll just fine for me on my system.

msimberg: I think besides this we could go ahead with this PR. You can always improve it at a later point with another pull request. Or did you have other improvements in the pipeline still?

Would it be acceptable to remove the `#pragma unroll` does performance change significantly without it? Some builders do not seem to recognize the pragma, and some (like clang) don't want to unroll it. That leaves two options: figure out which compilers support the pragma and which need to have the warning silenced and then put a bunch of `ifdef` to make the builds pass, or remove the pragma. The former would be better, but the latter is definitely simpler if the pragma doesn't actually make a big difference.

NK-Nikunj: I have one optimization in mind. The asynchronous nature of set with low grain size adds contention. I completely forgot to add the improved version here. I'll add the commit tonight.

Removing `#pragma unroll` should not cause significant performance loss but one can expect 100-200MLUPS worth of difference i.e. about 1-2% change in execution times.

msimberg:
@NK-Nikunj thanks for updating this. I think this looks ready to go except the inspect and build failures. That's is just clang complaining about unused lambda captures, and some other small issues. The codacy failure is real, but not very serious. You could try to silence it by moving things around though. Specifically, there's this: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/4301/workflows/42ee6888-0d93-486d-b843-99a4e160ac99/jobs/206284/steps, and this: https://206262-4455628-gh.circle-artifacts.com/0/hpx_inspect_report.html left.

NK-Nikunj:
Just saw the comment. @msimberg will correct them tomorrow.

NK-Nikunj:
@msimberg apologies for such a delay. I got engaged in other activities and totally forgot about this PR. I've corrected the inspect changes and the build errors. Let's hope it builds nicely this time ;-)

msimberg:
Thanks @NK-Nikunj! There were just the SPDX license tags missing. I've pushed a commit with those, hope you don't mind. This should be ready to go if nothing else comes up.

PR #4771, opened 2020-06-19
Fix compilation with HPX_WITH_NETWORKING=OFF

### LINKED ISSUE #4770: build error: HPX_WITH_DISTRIBUTED_RUNTIME=off is required for my configuration
## Expected Behavior

I tried building with the CMake setup I use for successfully build 1.4.1, so I expected it to work with master.

## Actual Behavior

It does not build (see [build-output.txt](https://github.com/STEllAR-GROUP/hpx/files/4802888/build-output.txt))

By disabling the module `-DHPX_WITH_DISTRIBUTED_RUNTIME=off`, that was not available in 1.4.1, it builds correctly (thanks @rasolca).

I built `master` 001c1bcc936c9bb73e54dbd3f8f8c95b3787cfda with a CMake configuration I used for HPX 1.4.1.

In particular, I think it is relevant the fact that I have `HPX_WITH_NETWORKING=off`.

## Specifications
  - HPX Version: master 001c1bcc936c9bb73e54dbd3f8f8c95b3787cfda
  - Platform (compiler, OS): GCC 8.3.0 on Linux

msimberg:
Fixes #4770 (at least for core, have not yet built the rest). Pycicle's `clang-oldest` now builds with `HPX_HAVE_NETWORKING=OFF` (`gcc-newest` has `HPX_WITH_DISTRIBUTED_RUNTIME=OFF` in addition, the rest are with full networking).

### DIFF
--- a/libs/collectives/tests/unit/CMakeLists.txt
+++ b/libs/collectives/tests/unit/CMakeLists.txt
@@ -9,31 +9,19 @@ set(tests
     all_reduce
     all_to_all
     barrier
-    broadcast
     broadcast_direct
     broadcast_apply
     broadcast_component
     fold
     gather
-    global_spmd_block
     reduce
     remote_latch
-    scatter
+    global_spmd_block
 )
 
-set(all_gather_PARAMETERS LOCALITIES 2)
-set(all_reduce_PARAMETERS LOCALITIES 2)
-set(all_to_all_PARAMETERS LOCALITIES 2)
-set(broadcast_PARAMETERS LOCALITIES 2)
-set(broadcast_direct_PARAMETERS LOCALITIES 2)
-set(broadcast_apply_PARAMETERS LOCALITIES 2)
-set(broadcast_component_PARAMETERS LOCALITIES 2)
-set(gather_PARAMETERS LOCALITIES 2)
-set(remote_latch_PARAMETERS LOCALITIES 2)
-set(reduce_PARAMETERS LOCALITIES 2)
-set(scatter_PARAMETERS LOCALITIES 2)
-
-set(global_spmd_block_PARAMETERS LOCALITIES 2)
+if(HPX_HAVE_NETWORKING)
+  set(tests ${tests} broadcast scatter)
+endif()
 
 foreach(test ${tests})
   set(sources ${test}.cpp)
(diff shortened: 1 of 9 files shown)


albestro:
I tested your branch and it builds successfully!

Thanks for your very quick support @msimberg !

libs/collectives/tests/unit/CMakeLists.txt
```diff
@@ -4,22 +4,25 @@
...
-    broadcast_apply
-    broadcast_component
-    fold
-    gather
-    global_spmd_block
-    reduce
-    remote_latch
-    scatter
-)
+set(tests barrier fold)
```
msimberg, line 7: @hkaiser should `barrier` and `fold` perhaps also run on two localities?

hkaiser: Yes, that might be a good idea.

hkaiser: However all of those should also run when networking is off as the facilities should be functional on one locality as well.

msimberg: Yeah, fair point. I changed it this way because those tests assert that they run with at least two localities. We can remove those asserts instead.

msimberg: I removed the checks for at least two localities, let's see how they run.

hkaiser: Thanks!

msimberg: @hkaiser it looks like broadcast and scatter time out when running only on one locality. Is that something you would expect to happen? Only `broadcast_to` gets called with one locality. Does it (or the completion of the future) block on someone receiving the broadcast with `broadcast_from`?

hkaiser: No, this is not what should happen. I'll have a look, thanks!

hkaiser: > Only broadcast_to gets called with one locality. Does it (or the completion of the future) block on someone receiving the broadcast with broadcast_from?

That's exactly what's happening. Let's disable those tests for one locality for now. This is for another PR, I presume.

msimberg: Thanks for confirming. Yes, I think that's best left for another PR. I opened https://github.com/STEllAR-GROUP/hpx/issues/4778 to keep track of that issue.

msimberg: I excluded scatter and broadcast again if networking isn't enabled. Let's see what CI says this time...

hkaiser: @msimberg: thanks for creating the ticket.

PR #4775, opened 2020-06-21
Checkpoint cleanup

hkaiser:
This PR does several things:

Cleaning up the existing checkpoint module

 - adding  `prepare_checkpoint` API function that calculates the amount of necessary 
    buffer space for a particular set of arguments checkpointed
 - adding nullary overloads for API functions (all of which do nothing), this is useful for 
    using the checkpointing facilities in generic contexts
 - making names more readable
 - modernize code by utilizing `=default` where possible
 - adding `noexcept` where possible

Separate the low-level checkpointing functionality into a new module `checkpoint_base`

 - The `checkpoint_base` module exposes three API functions: `save_checkpoint_data`, 
    `restore_checkpoint_data`, and `prepare_checkpoint_data`. These functions are essentially 
    convenience layers around serialization for a given set of (variadic) arguments.
 - The module does not support asynchrony, futures, or component-clients (this is left for the 
    `checkpoint` module). 
 - Please note that this lower-level functionality will be used (and is needed) in one of our 
    external projects

Move `hpx/runtime/serialization/prepare_container.hpp` to module `serialization`.


libs/serialization/include_compatibility/hpx/runtime/serialization/detail/preprocess_container.hpp
```diff
@@ -0,0 +1,23 @@
...
+#if defined(HPX_MSVC)
+#pragma message(                                                               \
+    "The header hpx/runtime/serialization/detail/preprocess_container.hpp is deprecated, \
+    please include hpx/serialization/detail/preprocess_container.hpp instead")
+#else
+#warning                                                                       \
+    "The header hpx/runtime/serialization/detail/preprocess_container.hpp is deprecated, \
+    please include hpx/serialization/detail/preprocess_container.hpp instead"
+#endif
+#endif
```
msimberg, line 23: We can skip this one if you like. I did remove all the detail compatibility headers in #4762.

hkaiser:
> Thanks, looks good to me. Could you expand a bit on where the nullary overloads are useful?

Mostly in generic contexts, e.g.
```
template <typename... Ts>
void print_and_checkpoint(Ts&&… ts)
{
    print(ts);
    save_checkpoint(std::forward<Ts>(ts));
}
```
As long as `save_checkpoint()` can be used without arguments one doesn't need to add a special case for it.

libs/serialization/include_compatibility/hpx/runtime/serialization/detail/preprocess_container.hpp
hkaiser: I have removed all of compatibility headers in serialization/detail now.

hkaiser:
This is ready to be merged now (once the tests have cycled), I think.

msimberg:
The `execution_base` module tests still have to be added to be explicitly built in the CircleCI config: https://circleci.com/gh/STEllAR-GROUP/hpx/204416?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link. Or... what do you think about simplifying the CircleCI config by just building and running most tests in a single step? Segmented algorithms would probably still have to be built sequentially. I don't think we're gaining much in feedback time right now by having the config split up into so many steps (and one has to manually add new modules to have their tests built). Most full pipelines take over 10 hours to finish right now because it looks like CircleCI's job scheduling isn't the smartest. (I volunteer to do the change of course.)

hkaiser:
> The execution_base module tests still have to be added to be explicitly built in the CircleCI config

I assume you meant the tests for the new `checkpoint_base` module. Yes, I always forget to change the CircleCI build (will do). Also, please feel free to merge the module tests into one step.

msimberg:
> > The execution_base module tests still have to be added to be explicitly built in the CircleCI config
> 
> I assume you meant the tests for the new `checkpoint_base` module. Yes, I always forget to change the CircleCI build (will do). Also, please feel free to merge the module tests into one step.

Yes, *checkpoint*, not execution... Think-o. Some base module :) Ok, I'll simplify the config separately from this PR.

msimberg:
@hkaiser I added the `fstream` header to `1d_stencil_checkpoint` because of this: https://cdash.cscs.ch//viewBuildError.php?buildid=112746. I'm guessing this is due to `fstream` being removed from one of the checkpointing files.

PR #4776, opened 2020-06-22
debug print improvements

biddisco:
This PR contains a load of debug printing improvements from several branches I'm working on, in particular latest libfabric changes. Pushing this to see if it causes any compilation problems on dashboards, as I will need it merged before other branches.

### DIFF
--- a/libs/debugging/include/hpx/debugging/print.hpp
+++ b/libs/debugging/include/hpx/debugging/print.hpp
@@ -11,8 +11,10 @@
 #include <array>
 #include <bitset>
 #include <chrono>
+#include <cmath>
 #include <cstddef>
 #include <cstdint>
+#include <cstring>
 #include <iomanip>
 #include <iostream>
 #include <iterator>
@@ -23,8 +25,9 @@
 #include <vector>
 
 #if defined(__linux) || defined(linux) || defined(__linux__)
-#include <linux/unistd.h>
+#include <unistd.h>
 #include <sys/mman.h>
+extern char **environ;
 #define DEBUGGING_PRINT_LINUX
 #endif
 
@@ -223,15 +226,21 @@ namespace hpx { namespace debug {
     {
         ipaddr(const void* a)
           : data_(reinterpret_cast<const uint8_t*>(a))
+          , ipdata_(0)
+        {
+        }
+        ipaddr(const uint32_t a)
+          : data_(reinterpret_cast<const uint8_t*>(&ipdata_))
+          , ipdata_(a)
         {
         }
         const uint8_t* data_;
+        const uint32_t ipdata_;
+
         friend std::ostream& operator<<(std::ostream& os, const ipaddr& p)
         {
-            os << std::dec << int((reinterpret_cast<const uint8_t*>(&p))[0])
-               << "." << int((reinterpret_cast<const uint8_t*>(&p))[1]) << "."
-               << int((reinterpret_cast<const uint8_t*>(&p))[2]) << "."
-               << int((reinterpret_cast<const uint8_t*>(&p))[3]);
+            os << std::dec << int(p.data_[0]) << "." << int(p.data_[1]) << "."
+               << int(p.data_[2]) << "." << int(p.data_[3]);
             return os;
         }
     };
@@ -239,7 +248,7 @@ namespace hpx { namespace debug {
     // ------------------------------------------------------------------
     // helper fuction for printing CRC32
     // ------------------------------------------------------------------
-    inline uint32_t crc32(const void* address, size_t length)
+    inline uint32_t crc32(const void* /*address*/, size_t /*length*/)
     {
         //        boost::crc_32_type result;
         //        result.process_bytes(address, length);
@@ -249,7 +258,7 @@ namespace hpx { namespace debug {
 
     // ------------------------------------------------------------------
     // helper fuction for printing short memory dump and crc32
-    // useful for debugging corruptions in buffers during parcelport
+    // useful for debugging corruptions in buffers during
     // rma or other transfers
     // ------------------------------------------------------------------
     struct mem_crc32
@@ -268,9 +277,12 @@ namespace hpx { namespace debug {
             const uint64_t* uintBuf = static_cast<const uint64_t*>(p.addr_);
             os << "Memory:";
             os << " address " << hpx::debug::ptr(p.addr_) << " length "
-               << hpx::debug::hex<8>(p.len_) << " CRC32:\n"
-               << hpx::debug::hex<8>(crc32(p.addr_, p.len_)) << " ";
-            for (size_t i = 0; i < (std::min)(p.len_ / 8, size_t(128)); i++)
+               << hpx::debug::hex<6>(p.len_)
+               << " CRC32:" << hpx::debug::hex<8>(crc32(p.addr_, p.len_))
+               << "\n";
+            for (size_t i = 0;
+                 i < (std::min)(size_t(std::ceil(p.len_ / 8.0)), size_t(128));
+                 i++)
             {
                 os << hpx::debug::hex<16>(*uintBuf++) << " ";
             }
@@ -281,7 +293,7 @@ namespace hpx { namespace debug {
 
     namespace detail {
 
-#ifdef HPX_HAVE_CXX17_FOLD_EXPRESSIONS
+#if defined(HPX_HAVE_CXX17_FOLD_EXPRESSIONS)
         template <typename TupleType, std::size_t... I>
         void tuple_print(
             std::ostream& os, const TupleType& tup, std::index_sequence<I...>)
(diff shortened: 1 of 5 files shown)


libs/debugging/include/hpx/debugging/print.hpp
```diff
@@ -225,71 +307,201 @@ namespace hpx { namespace debug {
...
+            {
+                os << std::get<0>(t);
+            }
+        };
+
+        template <class... Args>
+        void tuple_print(std::ostream& os, const std::tuple<Args...>& t)
+        {
+            tuple_printer<decltype(t), sizeof...(Args)>::print(os, t);
+        }
```
hkaiser, line 336: FWIW, something like the following would avoid the recursive expansion:
```suggestion
        // C++14 version
        // helper function to print a tuple of any size
        namespace detail {
            template <typename Tuple, std::size_t... Is>
            void tuple_print(std::ostream& os, Tuple const& t, 
                 std::index_sequence<Is...>)
            {
                using expander = int[];
                (void) expander{0, (void(os << ' ' << std::get<Is>(t)), 0)...};
            }
       }
        
        template <typename... Args> 
        void tuple_print( 
            std::ostream& os, std::tuple<Args...>const& t) 
        { 
            detail::tuple_print(os, t, std::make_index_sequence<sizeof...(Args)>{});
        }
```
```

File: libs/debugging/include/hpx/debugging/print.hpp [hunk 2]
```diff
@@ -225,71 +307,201 @@ namespace hpx { namespace debug {
...
+                    initialized = true;
+                    gethostname(hostname_, std::size_t(12));
+                    std::string temp = "(" + std::to_string(guess_rank()) + ")";
+                    std::strcat(hostname_, temp.c_str());
+                }
+                return hostname_;
+            }
+
+            int guess_rank() const {
+#ifdef DEBUGGING_PRINT_LINUX
```
hkaiser, line 377: WHat's the rationale of limiting this to `LINUX`?

libs/threading_base/include/hpx/threading_base/print.hpp
```diff
@@ -126,31 +125,48 @@ namespace hpx { namespace debug {
...
+            auto nowt = duration_cast<microseconds>(now - log_t_start).count();
             //
             os << debug::dec<10>(nowt) << " ";
             return os;
         }
 
+#ifdef HPX_HAVE_CXX17_FOLD_EXPRESSIONS
         template <typename... Args>
-        void display(const char* prefix, Args... args)
+        void display(const char* prefix, std::forward<Args>(args)... args)
```
hkaiser, line 141: This doesn't look right.

libs/debugging/include/hpx/debugging/print.hpp
biddisco: I'll try this out, thanks

libs/debugging/include/hpx/debugging/print.hpp
biddisco: I didn't want to add and env var checking things that might not compile on windows without testing them. I'm guessing that I'm the only person using this debug print stuff, so I didn't think anyone would notice.

libs/threading_base/include/hpx/threading_base/print.hpp
biddisco: It isn't right. not sure how that crept in, but I've removed it.

biddisco:
> Normally we use east-const notation in our code base... I'd appreciate it if you adapted to this. Thanks!

Is this something that clang-format can do fo us? I am not accustomed to writing consts on the right and I don't want to change all this by hand.

hkaiser:
> > Normally we use east-const notation in our code base... I'd appreciate it if you adapted to this. Thanks!
> 
> Is this something that clang-format can do fo us? I am not accustomed to writing consts on the right and I don't want to change all this by hand.

The clang-format people  say it's not the job of their tool to do that (clang-format never changes the sequence of tokens). They think this is something clang-tidy could provide. I'm not sure if somebody has actually implemented this there, however.

biddisco:
> > The clang-format people say it's not the job of their tool to do that (clang-format never changes the sequence of tokens). They think this is something clang-tidy could provide. I'm not sure if somebody has actually implemented this there, however.

There is work on-going in clang-format to add support for an east-west const fixer
https://reviews.llvm.org/D69764
I knocked up a quick regex search to find all the places where we are using west const instead of east const - it's a lot. I am going to push a cleanup patch to fix the cxx-14 variadic print (thanks, I've used your patch, plus some other improvments) and then a separate one with east const fixes for the debug printer.

hkaiser:
> There is work on-going in clang-format to add support for an east-west const fixer
> https://reviews.llvm.org/D69764

Cool, I was not aware of that.

> I knocked up a quick regex search to find all the places where we are using west const instead of east const - it's a lot. I am going to push a cleanup patch to fix the cxx-14 variadic print (thanks, I've used your patch, plus some other improvments) and then a separate one with east const fixes for the debug printer.

Thanks, much appreciated.

libs/debugging/include/hpx/debugging/print.hpp
```diff
@@ -23,11 +25,14 @@
...
 
 #if defined(__linux) || defined(linux) || defined(__linux__)
-#include <linux/unistd.h>
 #include <sys/mman.h>
+#include <unistd.h>
+extern char** environ;
 #define DEBUGGING_PRINT_LINUX
 #endif
 
+#undef HPX_HAVE_CXX17_FOLD_EXPRESSIONS
```
hkaiser, line 34: What's the rationale of always disabling the use of fold expressions here?

biddisco:
I made a bunch of cleanups and fixed some old stuff that wasn't updated. @hkaiser was correct in saying that debug arguments that are function returns are evaluated even in debug<false> paths, so I added a lazy `eval` wrapper (had to use a macro unfortunately) and added a unit test that checks that the functions/features are behaving as tested. Any other tweaks can wait for a new PR or the next release.

libs/debugging/include/hpx/debugging/print.hpp
hkaiser: The code you wrapped should be working fine on other platforms as long as you use:
```
#ifdef __APPLE__
#include <crt_externs.h>
#define environ (*_NSGetEnviron())
#elif defined(HPX_WINDOWS)
#define environ _environ
#else
extern char** environ;
#endif
```

libs/debugging/tests/unit/print.cpp
```diff
@@ -0,0 +1,129 @@
...
+    static hpx::debug::enable_print<true>  p_enabled("PRINT  ");
+    // this is disabled and we want it to have zero footprint
+    static hpx::debug::enable_print<false> p_disabled("PRINT  ");
+}   // namespace hpx
+
+int increment(std::atomic<int> &counter) {
+    return ++counter;
+}
+
+#define AB_LAZY(expr, debug) \
```
hkaiser, line 30: Is this is a typo? The code below uses `DB_LAZY`.

libs/debugging/include/hpx/debugging/print.hpp
```diff
@@ -61,6 +64,11 @@
 // The output will only be produced every N seconds
 // ------------------------------------------------------------
 
+// Used to wrap function call parameters to prevent evaluation
+// when debugging is disabled
+#define DP_LAZY(Expr, printer) \
```
hkaiser, line 69: Please prefix the macro name with `HPX_` to avoid unwanted name-clashes with macro names defined in other places.

biddisco: done

libs/debugging/tests/unit/print.cpp
biddisco: Sorry. Left over from an experiment. Removed

biddisco:
Add the environ check as suggested. I actually plan on removing that stuff and replacing it with something that uses the batch environment checks that already exist - but I will leave that until after the mpi branch is finished as I am actually using this in the mpi testing at the moment and it can wait as it is not urgent.

hkaiser:
Let's get this is asap.

biddisco:
After several rounds of fixing the windows/mac builds and sorting out some silly warnings etc - this now looks ok on dashboards - the fails being returned from tests are unrelated.

libs/debugging/include/hpx/debugging/print.hpp
```diff
@@ -23,9 +25,17 @@
...
-#include <linux/unistd.h>
 #include <sys/mman.h>
-#define DEBUGGING_PRINT_LINUX
+#include <unistd.h>
+#elif defined(__APPLE__)
+#include <crt_externs.h>
+#include <unistd.h>
+#define environ (*_NSGetEnviron())
+#elif defined(HPX_WINDOWS)
+#include <winsock2.h>
```
hkaiser, line 35: Why do we need to `#include <winsock2.h>` now?

hkaiser: The symbol `char** _environ` is defined in `#include <stdlib.h>` if this is what you were looking for.

biddisco: The CI windows build failed, so I replaced winsock with winsock2 and now it works. I haven't used windows in a while and don't really remember the details - the call to `gethostname` was the problem. It seems to be compiling without errors or warnings on all our test machines now.

PR #4782, opened 2020-06-24
Make header tests parallel again

haampie:
I did not see any test failures from running those tests in parallel


msimberg:
@hkaiser yes, but we discussed this already with @haampie. I set them to run serially because I had seen some spurious compilation failures (definitely not due to the actual code). The behaviour could depend on the generator as well. That's a roundabout way of saying that we should try this and if it shows problems later we can always go back again. This doesn't affect any of our current builders anyway. Thanks @haampie for changing this!

PR #4783, opened 2020-06-24
Add coveralls badge

msimberg:
Thanks to @haampie we now have coverage reports from master. Due to certain technical limitations this currently runs on CSCS's internal GitLab runners, so the setup is not the most transparent to others (this means no possibility to run this on pull requests from forked repos; GitHub actions ran out of disk space with the coverage reports). I think this is still very nice in its current form, and most importantly the coverage reports are public on https://coveralls.io/github/STEllAR-GROUP/hpx. I'm going to deflect technical questions to @haampie.

We're currently at around 80% coverage which I think is quite ok, considering that some of the lower coverage parts are not really necessary to cover in the first place (e.g. examples).


haampie:
@msimberg the issue I was talking about seems fixed now :)

hkaiser:
@haampie Thanks again for this! One question - why are there no concrete changes (between builds) shown even if there have been files added (see here for instance: https://coveralls.io/jobs/64326316)?

hkaiser:
> @haampie Thanks again for this! One question - why are there no concrete changes (between builds) shown even if there have been files added (see here for instance: https://coveralls.io/jobs/64326316)?

Never mind, I figured it out.

PR #4785, opened 2020-06-25
Fixing barrier test

hkaiser:
- flyby: making sure barrier is not released prematurely when used in
  asynchronous mode

### DIFF
--- a/libs/collectives/src/detail/barrier_node.cpp
+++ b/libs/collectives/src/detail/barrier_node.cpp
@@ -87,27 +87,36 @@ namespace hpx { namespace lcos { namespace detail {
 
     hpx::future<void> barrier_node::wait(bool async)
     {
+        hpx::intrusive_ptr<barrier_node> this_(this);
+        future<void> result;
+
         if (num_ < cut_off_)
         {
             if (rank_ != 0)
             {
                 HPX_ASSERT(children_.size() == 1);
                 hpx::lcos::base_lco::set_event_action action;
-                return hpx::async(action, children_[0]);
+                result = hpx::async(action, children_[0]);
             }
             else
             {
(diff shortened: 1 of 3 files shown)


libs/collectives/src/detail/barrier_node.cpp
```diff
@@ -196,13 +210,20 @@ namespace hpx { namespace lcos { namespace detail {
...
+        auto result = hpx::when_all(futures).then(
+            hpx::launch::sync, [this_](hpx::future<void> f) {
                 // Trigger possible errors...
                 f.get();
                 return this_->gather_promise_.get_future();
             });
+
+        // keep everything alive until future has become ready
+        traits::detail::get_shared_state(result)->set_on_completed(
+            [this_ = std::move(this_)] {});
```
msimberg, line 224: How much longer does this keep `this` alive than just capturing it above in the `then` continuation? Is the point that it has to live past the future actually being set (i.e. to the calling scope of the continuation), and why does that make a difference?

hkaiser: Fair point. I'll go back and check again. I might have been a bit overzealous here while I was doing the investigation.

hkaiser: My idea was to keep it alive as long as the returned future lives. The captured `this_` in the `.then()` expression lives as long as the future the continuation was attached to only.

hkaiser: I did look again. I think my changes are valid. We could remove the capture of `this_` for the continuations attached using `.then()`, however.

hkaiser:
The remaining errors are unrelated. Merging now.

PR #4788, opened 2020-06-25
Fixing inclusive transform scan algorithm to properly handle initial value

### LINKED ISSUE #4786: transform_inclusive_scan tries to implicitly convert between types, instead of using the provided `conv` function
## Expected Behavior

I am trying to use `hpx::parallel::transform_inclusive_scan`, taking a sequence of `T` as input, and producing a sequence of `R` as output. I provide a binary operator R+R->R, as well as a conversion function T->R. 
I expected that lack of the optional `init` argument cause the function to simply call the default constructor of type `R` where needed. Or alternatively, completely avoid calling the binary operator with `init` (which is possible, but needs some extra if-s and may be bad for performance).
When `init` is given, it works as intended.

## Actual Behavior

In the scenario, without the `init` argument, hpx tries to convert T to R through implicit conversion. When such conversion is impossible it leads to a compile error.

Example code:

```
#include <hpx/hpx_main.hpp>
#include <hpx/include/iostreams.hpp>
#include <hpx/parallel/algorithms/transform_inclusive_scan.hpp>

struct Integer {
    int integer;
    Integer()
        : integer(0) {
    }
    explicit Integer(int i)
        : integer(i) {
    }
};

int main(int argc, char *argv[]) {
    std::vector<int> test{1, 10, 100, 1000};
    std::vector<Integer> output(test.size());
    auto pol    = hpx::parallel::execution::par(hpx::parallel::execution::task);
    auto result = hpx::parallel::transform_inclusive_scan(
                      pol, test.cbegin(), test.cend(), output.begin(),
                      [](Integer acc, Integer xs) -> Integer {
                          return Integer{acc.integer + xs.integer};
                      },
                      [](int el) -> Integer { return Integer{el}; }/*, Integer{0}*/)
                      .get();
    for(auto el : output) {
        hpx::cout << el.integer << hpx::endl;
    }
    return 0;
}
```

## Specifications

Versions:
  HPX: V1.5.0-trunk (AGAS: V3.0), Git: unknown
  Boost: V1.72.0
  Hwloc: V1.11.0

Compiling with MSVC 2019, v.16.5.2 on Windows 10 x64

### LINKED ISSUE #4787: `transform_inclusive_scan` gives incorrect results for non-commutative operator
## Expected Behavior

`transform_inclusive_scan` should work correctly when the `op` is a non-commutative operator. 

Consider the following example:
Trying the perform an inclusive prefix scan of array: [1, 10, 100, 1000]
with a binary op(x,y) defined as follows:
if x<y then x+y else -10000

In this example, no matter how the operations are taken, the left argument will be smaller than the right. So, we should just get the regular prefix-sum result: [1, 11, 111, 1111].

## Actual Behavior

It seems to take the arguments in the inverted order.
The above example yields [-10000, -10000, -10000, -10000]

The code shows the problem:
```
#include <hpx/hpx_main.hpp>
#include <hpx/include/iostreams.hpp>
#include <hpx/parallel/algorithms/transform_inclusive_scan.hpp>

int main(int argc, char *argv[]) {
    std::vector<int> test{1, 10, 100, 1000};
    std::vector<int> output(test.size());
    auto pol    = hpx::parallel::execution::par(hpx::parallel::execution::task);
    auto result = hpx::parallel::transform_inclusive_scan(
                      pol, test.cbegin(), test.cend(), output.begin(),
                      [](int acc, int xs) -> int {
                          if(acc < xs) {
                              return acc + xs;
                          } else {
                              return -10000;
                          }
                      },
                      [](int el) -> int { return el; }, 0)
                      .get();
    for(auto el : output) {
        hpx::cout << el << hpx::endl;
    }
    return 0;
}
```

## Specifications

Versions:
  HPX: V1.5.0-trunk (AGAS: V3.0), Git: unknown
  Boost: V1.72.0
  Hwloc: V1.11.0

Compiling with MSVC 2019, v. 16.5.2 on Windows 10 x64.

hkaiser:
Fixes #4786
Fixes #4787


hkaiser:
This looks fine now. We can go ahead and merge this.

PR #4791, opened 2020-06-26
Create codeql-analysis.yml

msimberg:
Thanks @hkaiser for pointing out that this exists. I've no clue what it will actually do, but we'll hopefully find out soon enough (this config is unchanged from what GitHub suggested)!


msimberg:
I'll close this for now as it seems like it requires some additional setup and I'm not yet sure what we gain from this. This would still be interesting to look at in the future.

PR #4792, opened 2020-06-26
Replace enable_if with HPX_CONCEPT_REQUIRES_ and add is_sentinel_for constraint

gonidelis:
I replaced `enable_if` with the `HPX_CONCEPT_REQUIRES_` macro and added `is_sentinel_for` trait (except for the segmented algorithm part). All pre-existing tests seem to pass.

### DIFF
--- a/libs/algorithms/include/hpx/parallel/algorithms/reduce.hpp
+++ b/libs/algorithms/include/hpx/parallel/algorithms/reduce.hpp
@@ -9,8 +9,9 @@
 #pragma once
 
 #include <hpx/config.hpp>
+#include <hpx/concepts/concepts.hpp>
 #include <hpx/iterator_support/range.hpp>
-#include <hpx/iterator_support/traits/is_iterator.hpp>
+#include <hpx/iterator_support/traits/is_sentinel_for.hpp>
 #include <hpx/pack_traversal/unwrap.hpp>
 
 #include <hpx/executors/execution_policy.hpp>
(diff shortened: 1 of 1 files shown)


hkaiser:
@gonidelis I added one commit to this PR that improves the readability of the `HPX_CONCEPT_REQUIRES_()` expressions and disables formatting for them.

gonidelis:
> @gonidelis I added one commit to this PR that improves the readability of the `HPX_CONCEPT_REQUIRES_()` expressions and disables formatting for them.

Ok that's great! I didn't know about the clang-format on/off option. Looks better now for sure!

hkaiser:
I have still missed one place. Would you mind adding it here as well: https://github.com/STEllAR-GROUP/hpx/pull/4792/files#diff-ba8f23897c313b29d2dc80aa94393b40R87?

gonidelis:
> I have still missed one place. Would you mind adding it here as well: https://github.com/STEllAR-GROUP/hpx/pull/4792/files#diff-ba8f23897c313b29d2dc80aa94393b40R87?

You mean I should add `// clang-format on/of`, right?

libs/algorithms/include/hpx/parallel/algorithms/reduce.hpp
```diff
@@ -82,12 +83,17 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
-            execution::is_execution_policy<ExPolicy>::value,
-            typename util::detail::algorithm_result<ExPolicy, T>::type>::type
-        reduce_(ExPolicy&& policy, FwdIterB first, FwdIterE last, T init, F&& f,
+            typename T, typename F,
+            HPX_CONCEPT_REQUIRES_(
+                execution::is_execution_policy<ExPolicy>::value &&
+                hpx::traits::is_sentinel_for<FwdIterE, FwdIterB>::value
+            )>
+        // clang-format on
+        typename util::detail::algorithm_result<ExPolicy, T>::type reduce_(
```
aurianer, line 95: @hkaiser is the `inline` supposed to be gone ?

hkaiser: The `inline` does not make a big difference on template functions (those are `inline` implicitly). Removing it only removes a hint to the compiler to attempt inlining this function. But that's what it will do in any case.

aurianer: But just out of curiosity, isn't removing all those hints increasing our compile time ?

hkaiser: I wouldn't expect any measurable impact on the compilation process. If anything, compile times should go down as the compiler has to inline less code. I have no proof, however. In general I'm fine with adding the `inline` back, even if I think we don't do that consistently anyways.

PR #4793, opened 2020-06-29
Make sure global variable is handled properly

hkaiser:
Make sure global variable is

a) initialized on demand only
b) freed during shutdown

### DIFF
--- a/libs/lcos_local/src/composable_guard.cpp
+++ b/libs/lcos_local/src/composable_guard.cpp
@@ -225,12 +225,36 @@ namespace hpx { namespace lcos { namespace local {
         }
     };
 
+    namespace detail {
+        struct empty_helper
+        {
+            static guard_task*& get_empty_guard_task()
+            {
+                static guard_task* empty = new guard_task;
+                return empty;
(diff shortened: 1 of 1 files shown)


libs/lcos_local/src/composable_guard.cpp
```diff
@@ -225,12 +225,36 @@ namespace hpx { namespace lcos { namespace local {
...
+            empty_helper() = default;
+            ~empty_helper()
+            {
+                auto& empty = get_empty_guard_task();
+                delete empty;
+                empty = nullptr;
+            }
+        };
+
+        empty_helper empty_helper_{};
```
aurianer, line 246: Why is this needed ? (still working on my c++ skills) ^^'

hkaiser: This creates a global instance of `empty_helper`. We rely on the destructor of this (global) object to clean up the allocated memory.

stevenrbrandt:
Sorry, for whatever reason, I'm just seeing this ticket now.

PR #4794, opened 2020-06-29
Removing 128bit integer emulation

hkaiser:


hkaiser:
> LGTM.
> I'm curious - why do we need to remove this? is it just unused code or has it been superceded by compiler supported __int128 or something better?

These classes are simply not used anymore. We discussed this on IRC a while back and agreed to remove them.

PR #4795, opened 2020-06-29
Add CUDA flag to async_cuda CMakeLists

G-071:
Since PR #4385  got merged, HPX master could not be built with g++ and HPX_WITH_CUDA=ON as g++ was not able to find the cuda_runtime.h header file during the compilation.

This PR fixes this issue by simply adding the CUDA argument to the appropriate add_hpx_module call.

### DIFF
--- a/cmake/HPX_SetupCUDA.cmake
+++ b/cmake/HPX_SetupCUDA.cmake
@@ -20,6 +20,7 @@ if((HPX_WITH_CUDA_COMPUTE OR HPX_WITH_ASYNC_CUDA) AND NOT TARGET Cuda::cuda)
   endif()
 
   add_library(Cuda::cuda INTERFACE IMPORTED)
+  target_include_directories(Cuda::cuda INTERFACE ${CUDA_INCLUDE_DIRS})
 
   if(NOT HPX_WITH_CUDA_CLANG)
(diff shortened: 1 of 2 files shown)


G-071:
At least I do not think that this is caused by some cache problem. I used a fresh source copy and a fresh build directory to reproduce the issue. I also encounter this issue on two different machines, each with different g++ versions (7.5.0, 8.2.0) and cuda versions (10.0, 10.1).

biddisco:
hmmmm. I'm not happy. Adding `CUDA` to the module triggers the use of `cuda_add_library` in place of plain `add_library`, but this should not be necessary for the module. The correct fix is to make sure that the target `Cuda::cuda` has the cuda runtime includes set correctly. I will try to fix this today.

biddisco:
@G-071 could I ask you to try adding
`   target_include_directories(Cuda::cuda INTERFACE ${CUDA_INCLUDE_DIRS}) `
to `cmake/HPX_SetupCUDA.cmake` right after the line (22) containing
`add_library(Cuda::cuda INTERFACE IMPORTED)`
and tell me if this solves the problem. TIA

G-071:
@biddisco Yes, this fixes my issue! Thanks!

I have updated this PR accordingly!

PR #4796, opened 2020-06-30
Display relative paths in hook to make cut'n'paste easier

biddisco:
Just for ease of use - if you commit from a subdir instead of the root dir of the git repo ...


aurianer:
Ah actually it doesn't work when in the root of the repo apparently

PR #4797, opened 2020-06-30
Change cmake targets

aurianer:
- Change the default hpx target not to include `wrap_main` as discussed in the last PMC meeting, the `HPX::wrap_main` target is exposed to the user in case it's needed
- Add a gtest unit test to check we don't break anything during our changes to the cmake targets

### DIFF
--- a/.github/workflows/windows_debug.yml
+++ b/.github/workflows/windows_debug.yml
@@ -23,11 +23,15 @@ jobs:
     - name: Configure
       shell: bash
       run: |
-          cmake . -Bbuild -G'Visual Studio 16 2019' -DCMAKE_BUILD_TYPE=Debug -DCMAKE_TOOLCHAIN_FILE='C:/projects/vcpkg/scripts/buildsystems/vcpkg.cmake' -DHPX_WITH_PSEUDO_DEPENDENCIES=OFF -DHPX_WITH_EXAMPLES=ON -DHPX_WITH_TESTS=ON -DHPX_WITH_DEPRECATION_WARNINGS=OFF -DHPX_WITH_TESTS_MAX_THREADS_PER_LOCALITY=2
+          cmake . -Bbuild -G'Visual Studio 16 2019' -DCMAKE_BUILD_TYPE=Debug -DCMAKE_TOOLCHAIN_FILE='C:/projects/vcpkg/scripts/buildsystems/vcpkg.cmake' -DHPX_WITH_PSEUDO_DEPENDENCIES=OFF -DHPX_WITH_EXAMPLES=ON -DHPX_WITH_TESTS=ON -DHPX_WITH_TESTS_UNIT=ON -DHPX_WITH_DEPRECATION_WARNINGS=OFF -DHPX_WITH_TESTS_MAX_THREADS_PER_LOCALITY=2
     - name: Build
       shell: bash
       run: |
-          cmake --build build --config Debug --target ALL_BUILD -- -maxcpucount -verbosity:minimal -nologo
+          cmake --build build --config Debug \
+          --target ALL_BUILD \
+            cmake_build_dir_targets_test.make_compile \
+            cmake_build_dir_macros_test.make_compile \
+          -- -maxcpucount -verbosity:minimal -nologo
     - name: Test
       shell: bash
       run: |
--- a/.github/workflows/windows_release.yml
+++ b/.github/workflows/windows_release.yml
@@ -23,11 +23,26 @@ jobs:
     - name: Configure
       shell: bash
       run: |
-          cmake . -Bbuild -G'Visual Studio 16 2019' -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE='C:/projects/vcpkg/scripts/buildsystems/vcpkg.cmake' -DHPX_WITH_PSEUDO_DEPENDENCIES=OFF -DHPX_WITH_EXAMPLES=ON -DHPX_WITH_TESTS=ON -DHPX_WITH_DEPRECATION_WARNINGS=OFF -DHPX_WITH_TESTS_MAX_THREADS_PER_LOCALITY=2
+          cmake . -Bbuild -G'Visual Studio 16 2019' -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE='C:/projects/vcpkg/scripts/buildsystems/vcpkg.cmake' -DHPX_WITH_PSEUDO_DEPENDENCIES=OFF -DHPX_WITH_EXAMPLES=ON -DHPX_WITH_TESTS=ON -DHPX_WITH_TESTS_UNIT=ON -DHPX_WITH_DEPRECATION_WARNINGS=OFF -DHPX_WITH_TESTS_MAX_THREADS_PER_LOCALITY=2
     - name: Build
       shell: bash
       run: |
-          cmake --build build --config Release --target ALL_BUILD -- -maxcpucount -verbosity:minimal -nologo
+          cmake --build build --config Release \
+          --target ALL_BUILD \
+            cmake_build_dir_targets_test.make_compile \
+            cmake_build_dir_macros_test.make_compile \
+          -- -maxcpucount -verbosity:minimal -nologo
+    - name: Install
+      shell: bash
+      run: |
+          cmake --install build
+    - name: Build External
+      shell: bash
+      run: |
+          cmake --build build --config Release \
+            --target cmake_install_dir_targets_test.make_compile \
+              cmake_install_dir_macros_test.make_compile \
+            -- -maxcpucount -verbosity:minimal -nologo
     - name: Test
       shell: bash
       run: |
--- a/cmake/HPX_SetupTarget.cmake
+++ b/cmake/HPX_SetupTarget.cmake
@@ -172,9 +172,15 @@ function(hpx_setup_target target)
   endif()
 
   if(NOT target_NOLIBS)
+    set(_wrap_main_deps)
+    if("${_type}" STREQUAL "EXECUTABLE")
+      set(_wrap_main_deps $<TARGET_NAME_IF_EXISTS:wrap_main>
+                          $<TARGET_NAME_IF_EXISTS:HPX::wrap_main>
+      )
+    endif()
     target_link_libraries(
       ${target} ${__tll_public} $<TARGET_NAME_IF_EXISTS:hpx>
-      $<TARGET_NAME_IF_EXISTS:HPX::hpx>
+      $<TARGET_NAME_IF_EXISTS:HPX::hpx> ${_wrap_main_deps}
     )
     hpx_handle_component_dependencies(target_COMPONENT_DEPENDENCIES)
     target_link_libraries(
--- a/examples/gtest_emulation/CMakeLists.txt
+++ b/examples/gtest_emulation/CMakeLists.txt
@@ -28,6 +28,8 @@ if(EXISTS "${HPX_DIR}")
 else()
   message(
     WARNING
-      "HPX_DIR=${HPX_DIR} does not exist. Did you forget to run the install rule?"
+      "Skipping build test because HPX_DIR=${HPX_DIR} does not exist. This \
+      last test requires HPX to be installed.  Did you forget to run the \
+      install rule?"
   )
 endif()
--- a/examples/gtest_emulation/CMakeLists.txt
+++ b/examples/gtest_emulation/CMakeLists.txt
@@ -1,0 +1,33 @@
+# Copyright (c) 2020  ETH Zurich
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+cmake_minimum_required(VERSION 3.13)
+
+project(gtest CXX)
+
+if(EXISTS "${HPX_DIR}")
+
+  find_package(HPX REQUIRED)
+
+  # Add a static library which contains a main to emulate gtest_main
+  add_library(static_main_lib STATIC static_main.cpp)
+
+  # /!\ This helper interface is needed to keep the right linking order
+  add_library(hpx_helper_interface INTERFACE)
+  target_link_libraries(
+    hpx_helper_interface INTERFACE HPX::hpx HPX::wrap_main static_main_lib
+  )
+
+  # Test with the main function in a separate static library
+  add_executable(hpx_main_ext_main hpx_main_ext_main.cpp)
+  target_link_libraries(hpx_main_ext_main PRIVATE hpx_helper_interface)
+
+else()
+  message(
+    WARNING
+      "HPX_DIR=${HPX_DIR} does not exist. Did you forget to run the install rule?"
+  )
+endif()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -536,54 +536,51 @@ set(_library_types
 set(_is_executable "$<STREQUAL:$<TARGET_PROPERTY:TYPE>,EXECUTABLE>")
 set(_is_library "$<IN_LIST:$<TARGET_PROPERTY:TYPE>,${_library_types}>")
 
-add_library(hpx_no_wrap_main INTERFACE)
 add_library(hpx INTERFACE)
+add_library(wrap_main INTERFACE)
 
-target_link_libraries(hpx_no_wrap_main INTERFACE hpx_core)
 target_link_libraries(hpx INTERFACE hpx_core)
 
-# hpx_interface and hpx_interface_no_wrap_main contain additional interface
-# options to be passed to dependent targets. We create these as separate targets
-# to easily filter out the generator expressions that can't be handled by the
+# hpx_interface and hpx_interface_wrap_main contain additional interface options
+# to be passed to dependent targets. We create these as separate targets to
+# easily filter out the generator expressions that can't be handled by the
 # pkgconfig file generation.
-add_library(hpx_interface_no_wrap_main INTERFACE)
+add_library(hpx_interface INTERFACE)
 target_link_libraries(
-  hpx_interface_no_wrap_main
+  hpx_interface
   INTERFACE $<${_is_executable}:$<TARGET_NAME_IF_EXISTS:hpx_init>>
 )
 target_link_libraries(
-  hpx_interface_no_wrap_main
+  hpx_interface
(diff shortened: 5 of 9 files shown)


aurianer:
`fetchcontent_makeavailable` is available only from `CMake/3.14`, do we mind bumping the cmake minimum to 3.14 instead of 3.13?

aurianer:
> Should we also have a test verifying that using gtest's main wrapper works as expected?

Absolutely I'm working on it ;)

aurianer:
So in the end, the `gtest_minimal.cpp` requires an extra interface library to keep the linking order, the following does not keep the order and links `gtest_main` first which results in a `undefined reference to main` error linking error:
```
add_executable(gtest_minimal gtest_minimal.cpp)
target_link_libraries(gtest_minimal INTERFACE HPX::hpx HPX::wrap_main gtest_main)
```
And this works magically:
```
add_executable(gtest_minimal gtest_minimal.cpp)                               
                                                                                
add_library(gtest_main_hpx_interface INTERFACE)                               
target_link_libraries(gtest_main_hpx_interface INTERFACE HPX::hpx HPX::wrap_main gtest_main)              
                   
target_link_libraries(gtest_minimal PRIVATE gtest_main_hpx_interface)
```
I'm still researching how to make it our problem and not the user's problem

hkaiser:
@aurianer Even if we can't solve it on our side, we now have a recipe for the user on how to solve the issue. Thanks! If nothing else, we should add this to the documentation.

examples/gtest/gtest_minimal.cpp
```diff
@@ -0,0 +1,21 @@
+//  Copyright (c) 2020 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <gtest/gtest.h>
+
+#include <hpx/hpx_main.hpp>
```
hkaiser, line 9: Is the sequencing of the headers important (just curious)?

hkaiser:
@aurianer I have another question: Would all of this work if `HPX_WITH_DYNAMIC_HPX_MAIN` is disabled?

examples/gtest/gtest_minimal.cpp
aurianer: I've tried changing the order of the header before thinking of the interface target and it didn't solve my linking problem

aurianer:
> @aurianer I have another question: Would all of this work if `HPX_WITH_DYNAMIC_HPX_MAIN` is disabled?

Will try!

aurianer:
> > @aurianer I have another question: Would all of this work if `HPX_WITH_DYNAMIC_HPX_MAIN` is disabled?
> 
> Will try!
Seems to work with `HPX_WITH_DYNAMIC_HPX_MAIN` set to `OFF`

aurianer:
Ok looks like I still have some trouble with some configurations, will look into it!

msimberg:
> So in the end, the [gtest_minimal.cpp](https://github.com/STEllAR-GROUP/hpx/pull/4797/files#diff-18b2821f53b242c9db25e4fa3ee7e721R7-R21) requires an extra interface library to keep the linking order, the following does not keep the order and links `gtest_main` first which results in a `undefined reference to main` linking error:
> 
> ```
> add_executable(gtest_minimal gtest_minimal.cpp)
> target_link_libraries(gtest_minimal INTERFACE HPX::hpx HPX::wrap_main gtest_main)
> ```

Should that be `PRIVATE` (or `PUBLIC`) instead of `INTERFACE`? I'd expect `INTERFACE` on an executable to just not link anything at all... Probably explains why the below works but not the above.

> 
> And this works magically:
> 
> ```
> add_executable(gtest_minimal gtest_minimal.cpp)                               
>                                                                                 
> add_library(gtest_main_hpx_interface INTERFACE)                               
> target_link_libraries(gtest_main_hpx_interface INTERFACE HPX::hpx HPX::wrap_main gtest_main)              
>                    
> target_link_libraries(gtest_minimal PRIVATE gtest_main_hpx_interface)
> ```
> 
> I'm still researching how to make it our problem and not the user's problem

In any case, having this work even with explicit work from the user is still much better than the confusing errors one gets with the implicit behaviour on master.

msimberg:
> > > @aurianer I have another question: Would all of this work if `HPX_WITH_DYNAMIC_HPX_MAIN` is disabled?
> > 
> > 
> > Will try!
> > Seems to work with `HPX_WITH_DYNAMIC_HPX_MAIN` set to `OFF`

This whole business of actually having `gtest_main` and `wrap_main` linked and the runtime started implicitly can only ever work with `HPX_WITH_DYNAMIC_HPX_MAIN=ON` because when it's off the definition of `main` has to be where `hpx/hpx_main.hpp` is included. So these tests are primarily to test that there are no linking errors if one doesn't ask for `wrap_main`, and secondarily to test that it is possible to use something like `gtest_main` and `wrap_main` on supported platforms.

hkaiser:
> This whole business of actually having `gtest_main` and `wrap_main` linked and the runtime started implicitly can only ever work with `HPX_WITH_DYNAMIC_HPX_MAIN=ON` because when it's off the definition of `main` has to be where `hpx/hpx_main.hpp` is included. So these tests are primarily to test that there are no linking errors if one doesn't ask for `wrap_main`, and secondarily to test that it is possible to use something like `gtest_main` and `wrap_main` on supported platforms.

Sure, I however would like for things to work on Windows as well...

msimberg:
> > This whole business of actually having `gtest_main` and `wrap_main` linked and the runtime started implicitly can only ever work with `HPX_WITH_DYNAMIC_HPX_MAIN=ON` because when it's off the definition of `main` has to be where `hpx/hpx_main.hpp` is included. So these tests are primarily to test that there are no linking errors if one doesn't ask for `wrap_main`, and secondarily to test that it is possible to use something like `gtest_main` and `wrap_main` on supported platforms.
> 
> Sure, I however would like for things to work on Windows as well...

I hope they will, and they should keep working as before... :) It depends what you mean by "things" of course...

aurianer:
> > So in the end, the [gtest_minimal.cpp](https://github.com/STEllAR-GROUP/hpx/pull/4797/files#diff-18b2821f53b242c9db25e4fa3ee7e721R7-R21) requires an extra interface library to keep the linking order, the following does not keep the order and links `gtest_main` first which results in a `undefined reference to main` linking error:
> > ```
> > add_executable(gtest_minimal gtest_minimal.cpp)
> > target_link_libraries(gtest_minimal INTERFACE HPX::hpx HPX::wrap_main gtest_main)
> > ```
> 
> Should that be `PRIVATE` (or `PUBLIC`) instead of `INTERFACE`? I'd expect `INTERFACE` on an executable to just not link anything at all... Probably explains why the below works but not the above.
> 
> > And this works magically:
> > ```
> > add_executable(gtest_minimal gtest_minimal.cpp)                               
> >                                                                                 
> > add_library(gtest_main_hpx_interface INTERFACE)                               
> > target_link_libraries(gtest_main_hpx_interface INTERFACE HPX::hpx HPX::wrap_main gtest_main)              
> >                    
> > target_link_libraries(gtest_minimal PRIVATE gtest_main_hpx_interface)
> > ```
> > 
> > 
> > I'm still researching how to make it our problem and not the user's problem
> 
> In any case, having this work even with explicit work from the user is still much better than the confusing errors one gets with the implicit behaviour on master.

Yes sorry I modified this locally but didn't push it

aurianer:
So in the end, the gtest emulation is now working, thanks @msimberg for having a look, let me know what you think :)

cmake/HPX_SetupTarget.cmake
```diff
@@ -174,7 +174,8 @@ function(hpx_setup_target target)
   if(NOT target_NOLIBS)
     target_link_libraries(
       ${target} ${__tll_public} $<TARGET_NAME_IF_EXISTS:hpx>
-      $<TARGET_NAME_IF_EXISTS:HPX::hpx>
+      $<TARGET_NAME_IF_EXISTS:HPX::hpx> $<TARGET_NAME_IF_EXISTS:wrap_main>
+      $<TARGET_NAME_IF_EXISTS:HPX::wrap_main>
```
msimberg, line 178: Let's put linking the `wrap_main` targets behind `if("{_type}" STREQUAL "EXECUTABLE")`.

examples/gtest_emulation/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+
+if(EXISTS "${HPX_DIR}")
+
+  find_package(HPX REQUIRED)
+
+  # Add a static library which contains a main to emulate gtest_main
+  add_library(static_main_lib STATIC static_main.cpp)
+
+  # /!\ This helper interface is needed to keep the right linking order
+  add_library(hpx_helper_interface INTERFACE)
```
msimberg, line 19: So was this indirection through an interface library needed in the end or not?

src/CMakeLists.txt
```diff
@@ -628,8 +625,13 @@ target_compile_definitions(
...
 
-set(hpx_targets hpx hpx_no_wrap_main plugin component)
-set(hpx_internal_targets hpx_core hpx_interface hpx_interface_no_wrap_main)
+set(hpx_targets hpx wrap_main plugin component)
+set(hpx_internal_targets hpx_core hpx_interface hpx_interface_wrap_main)
+
+# Temporary for compatibility with projects like gridtools
+add_library(hpx_no_wrap_main INTERFACE)
+target_link_libraries(hpx_no_wrap_main INTERFACE hpx)
+list(APPEND hpx_targets hpx_no_wrap_main)
```
msimberg, line 634: I think we can remove this outright since the target didn't exist in 1.4.1.

examples/gtest_emulation/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+  target_link_libraries(
+    hpx_helper_interface INTERFACE HPX::hpx HPX::wrap_main static_main_lib
+  )
+
+  # Test with the main function in a separate static library
+  add_executable(hpx_main_ext_main hpx_main_ext_main.cpp)
+  target_link_libraries(hpx_main_ext_main PRIVATE hpx_helper_interface)
```
msimberg, line 26: This one would only work if `HPX_WITH_DYNAMIC_MAIN=ON`. I think the build tests don't get built on Windows GitHub actions... would you mind adding explicit commands to build the non-pkgconfig build tests to the Windows GitHub actions configs (there are two configurations) since pseudotargets don't work on Windows?

examples/gtest_emulation/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+
+  # Use of hpx_start so that it doesn't break at hpx/hpx_main.hpp not found but
+  # at the undefined reference to main
+  add_executable(hpx_start_ext_main hpx_start_ext_main.cpp)
+  target_link_libraries(hpx_start_ext_main PRIVATE HPX::hpx static_main_lib)
```
msimberg, line 31: What exactly does this one test?

tests/unit/build/CMakeLists.txt
```diff
@@ -137,6 +137,20 @@ create_cmake_test(
...
+  cmake_build_dir_gtest_emulation_test
+  "${PROJECT_BINARY_DIR}/lib/cmake/${HPX_PACKAGE_NAME}"
+  "examples/gtest_emulation" TARGETS
+)
+
+create_cmake_test(
+  cmake_install_dir_gtest_emulation_test
+  "${CMAKE_INSTALL_FULL_LIBDIR}/cmake/${HPX_PACKAGE_NAME}"
+  "examples/gtest_emulation" TARGETS
+)
```
msimberg, line 151: The `hpx_main` test can only work on Windows and I'm not sure the `hpx_start` test is useful (at least as it is). So we should disable the `hpx_main` test either here or in `examples/gtest_emulation/CMakeLists.txt`.

src/CMakeLists.txt
aurianer: It was more for projects like DLA / gridtools which currently depend on that, just for them to make the changes and after can be removed without deprecation protocol

examples/gtest_emulation/CMakeLists.txt
aurianer: will do

examples/gtest_emulation/CMakeLists.txt
aurianer: needed otherwise the link order is not the right one, cmake is weird on that point

src/CMakeLists.txt
msimberg: Fair enough, good thinking. https://github.com/STEllAR-GROUP/hpx/issues/4811 is there as a reminder to remove it before 1.5.0.

.github/workflows/windows_debug.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           ctest \
           --output-on-failure \
             --build-config Debug \
-            --tests-regex tests.examples \
+            --tests-regex "tests.examples|cmake_build_dir_gtest_emulation_test" \
```
msimberg, line 38: This will have to move to the build stage (there is no test, only a build target). We could of course make it into a proper compilation test but that doesn't need to happen in this PR.

msimberg: Actually, this will do nothing with the gtest test disabled on Windows. It would be good to test the regular build tests (with hello world component) on Windows though.

examples/gtest_emulation/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+
+  # Test with the main function in a separate static library
+  add_executable(hpx_main_ext_main hpx_main_ext_main.cpp)
+  target_link_libraries(hpx_main_ext_main PRIVATE hpx_helper_interface)
+
+else()
+  message(
+    WARNING
+      "HPX_DIR=${HPX_DIR} does not exist. Did you forget to run the install rule?"
+  )
```
hkaiser, line 32: Shouldn't things work from the build directory as well?

.github/workflows/windows_debug.yml
aurianer: Sorry did that too quickly.. So I add `HPX_WITH_UNIT_TESTS=ON` to the configure step and add `cmake_build_dir_targets_test` (`hello_world_component` test) in the build step?

examples/gtest_emulation/CMakeLists.txt
msimberg: This is unrelated to that... This just prints a warning if `HPX_DIR` (i.e. the install directory) is not found. The build directory is always found. We discussed this a long time ago and left it as a warning since that means one can run the build tests also without installing HPX.

.github/workflows/windows_debug.yml
msimberg: Yes. The tests should already be enabled (by default) so you can just add the targets to the build step.

examples/gtest_emulation/CMakeLists.txt
aurianer: So there is actually two targets constructed as for the hello world component, one for the build directory and one for the install directory. I would like for the test to be the closest to the real case (i.e. most probably a project using an installed  version of HPX), what do you think?

.github/workflows/windows_debug.yml
aurianer: thanks!

examples/gtest_emulation/CMakeLists.txt
msimberg: You can check what windows does with `cmake --target install`. It would indeed be good to test the installed version as well.

hkaiser: > This is unrelated to that... This just prints a warning if `HPX_DIR` (i.e. the install directory) is not found. The build directory is always found. We discussed this a long time ago and left it as a warning since that means one can run the build tests also without installing HPX.

Why do we need a warning if everything works?

msimberg: There are four options:
1. Silently ignore it if `HPX_DIR` does not exist (i.e. HPX has not been installed).
2. Print a warning.
3. Make it an error.
4. Make the test depend on the `install` target.
We currently do number 2 because it means that tests don't fail if you haven't installed HPX, but you still get a reminder that this test essentially was not run. We can change this if you think one of the other options would be better.

hkaiser: I'm fine with the warning, however Id suggest to make it more explanatory - e.g. saying that the test such-and-such was not run because... Or similar. 

BTW, can we use `hpx_warning()` instead of `message()`?

msimberg: > BTW, can we use `hpx_warning()` instead of `message()`?

Unfortunately not, because in this case HPX is not found. This CMakeLists.txt is for an "external project".

.github/workflows/windows_release.yml
```diff
@@ -23,11 +23,15 @@ jobs:
...
-          cmake . -Bbuild -G'Visual Studio 16 2019' -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE='C:/projects/vcpkg/scripts/buildsystems/vcpkg.cmake' -DHPX_WITH_PSEUDO_DEPENDENCIES=OFF -DHPX_WITH_EXAMPLES=ON -DHPX_WITH_TESTS=ON -DHPX_WITH_DEPRECATION_WARNINGS=OFF -DHPX_WITH_TESTS_MAX_THREADS_PER_LOCALITY=2
+          cmake . -Bbuild -G'Visual Studio 16 2019' -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE='C:/projects/vcpkg/scripts/buildsystems/vcpkg.cmake' -DHPX_WITH_PSEUDO_DEPENDENCIES=OFF -DHPX_WITH_EXAMPLES=ON -DHPX_WITH_TESTS=ON -DHPX_WITH_TESTS_UNIT=ON -DHPX_WITH_DEPRECATION_WARNINGS=OFF -DHPX_WITH_TESTS_MAX_THREADS_PER_LOCALITY=2
     - name: Build
       shell: bash
       run: |
-          cmake --build build --config Release --target ALL_BUILD -- -maxcpucount -verbosity:minimal -nologo
+          cmake --build build --config Release \
+          --target ALL_BUILD \
+            cmake_build_dir_targets_test cmake_install_dir_targets_test \
+            cmake_build_dir_macros_test cmake_install_dir_macros_test \
```
msimberg, line 33: I think these might have to be `cmake_build_dir_targets_test.make_compile` because of the lack of pseudo-dependencies on Windows. @auriane would you mind trying that out? If that doesn't work maybe @hkaiser could check what the targets are called on Windows?

aurianer: will try thanks!

PR #4799, opened 2020-06-30
Fixing #include paths in compatibility headers

hkaiser:


libs/async_combinators/include_compatibility/hpx/lcos/split_future.hpp
```diff
@@ -8,14 +8,14 @@
 
 #include <hpx/config.hpp>
 #include <hpx/async_combinators/config/defines.hpp>
-#include <hpx/future.hpp>
+#include <hpx/async_combinators/split_future.hpp>
```
msimberg, line 11: What's the motivation behind these changes? Most of these really are exposed through `hpx/future.hpp`. Some may be a bit overkill (I don't consider e.g. `futures_factory.hpp` to be part of the public API), but for most of them I would really like users to include `hpx/future.hpp` instead of exactly the header corresponding to the old location.

hkaiser: Ok, fair point. While creating the action modules I ran into `#include` not found errors (before fixing all headers) which made me look for the compatibility headers that were including `<hpx/future.hpp>` which I assumed were leftovers or copy&paste mistakes. I see your rationale now. Let me go back and see if now - after I have fixed all header paths - everything is ok if I revert the changes proposed here.

msimberg: Ok, I see. Yes, some of the headers we forward to were missing before #4726 (`hpx/future.hpp` was already there though... hmm). If there are more of them missing that's likely a bug.

hkaiser:
Closing this. Things work without this.

PR #4801, opened 2020-07-02
Replace boost::shared_array<T> with std::shared_ptr<T[]> if supported

hkaiser:


libs/serialization/include/hpx/serialization/serialize_buffer.hpp
```diff
@@ -95,7 +104,7 @@ namespace hpx { namespace serialization {
             else
```
msimberg, line 104: I guess the `boost::shared_array<T>` just two lines above should be `buffer_type` as well?

hkaiser: Excellent catch, thanks. Will fix.

msimberg:
@hkaiser I just rebased and resolved the one conflicting file so this is hopefully ready to go once it cycles through CI. Thanks a lot for doing this!

PR #4802, opened 2020-07-02
Fix datapar header paths

G-071:
Some of the datapar header include paths are wrong (```hpx/execution/parallel/traits/...``` instead of ```hpx/execution/traits/...```) causing issues during the compilation of Octo-Tiger using the latest HPX master.

This PR fixes this issue!


msimberg:
Thanks @G-071! You'll need to apply `clang-format` (version 9), or in this case it might be easier to just manually reorder the includes. Here's the diff for what you should have: https://207662-4455628-gh.circle-artifacts.com/0/hpx/artifacts/modified_clang_format_files.txt.

PR #4805, opened 2020-07-06
Applying changes that resolve time complexity issues in sort

### LINKED ISSUE #2836: The worst-case time complexity of parallel::sort seems to be O(N^2).
Currently, the worst-case time complexity of parallel::sort seems to be O(N^2).

https://github.com/STEllAR-GROUP/hpx/blob/d55dc1b220b6ba78a715a38779180f995b632ae2/hpx/parallel/algorithms/sort.hpp#L80-L89
https://github.com/STEllAR-GROUP/hpx/blob/d55dc1b220b6ba78a715a38779180f995b632ae2/hpx/parallel/algorithms/sort.hpp#L95-L97
Surely, due to the two code blocks above, the possibility is very rare.
But, theoretically it still seems to be able to be O(N^2) in the worst-case.
Normally, for ensuring O(N*logN), switch algorithm to heap sort when recursion depth is higher than specific threshold.
If we want to ensure O(N*logN), we should modify Line 80 to
```cpp
if (std::size_t(N) <= sort_limit_per_task || depth > recursion_depth_limit) 
```
But, I'm afraid of performance change from those modifications.

I want to hear your thoughts about this.

hkaiser:
Fixes #2836 

This is integrating changes as suggested by @fjtapia (see #2836). Thanks Francisco!

### DIFF
--- a/libs/algorithms/include/hpx/parallel/algorithms/sort.hpp
+++ b/libs/algorithms/include/hpx/parallel/algorithms/sort.hpp
@@ -1,6 +1,6 @@
 //  Copyright (c) 2015 John Biddiscombe
-//  Copyright (c) 2015-2017 Hartmut Kaiser
-//  Copyright (c) 2015 Francisco Jose Tapia
+//  Copyright (c) 2015-2020 Hartmut Kaiser
+//  Copyright (c) 2015-2019 Francisco Jose Tapia
 //  Copyright (c) 2018 Taeguk Kwon
 //
 //  SPDX-License-Identifier: BSL-1.0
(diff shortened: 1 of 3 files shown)


biddisco:
I looked at the code and I see what you've done, but my memory of the discussion is gone completely. Can you just explain in a short sentence, what you've done here - the change the pivot calculation - what's the significance?

hkaiser:
> I looked at the code and I see what you've done, but my memory of the discussion is gone completely. Can you just explain in a short sentence, what you've done here - the change the pivot calculation - what's the significance?

@fjtapia has summarized it here: https://github.com/STEllAR-GROUP/hpx/issues/2836#issuecomment-529115573

hkaiser:
@msimberg this is good to go now

msimberg:
Thanks @hkaiser! The CircleCI modules check is already fixed on master. Merging.

PR #4806, opened 2020-07-07
Update documentation links to new domain name

msimberg:


msimberg:
> LGTM, however - wouldn't it be easier to create a macro or somesuch?

Very good point :D If we end up changing them once more I'll do that. This time sed was once again easy enough.

PR #4807, opened 2020-07-07
Add Jenkins configuration for running on Piz Daint

msimberg:
This reproduces the configurations that we use with pycicle, but I will keep both running until I'm sure that this is working well. Some things I will try to update or fix separately:

- Some of the "oldest" configurations use somewhat outdated versions of Boost (they still build correctly though!).
- CDash may have a problem with too many builds trying to push updates at the same time (not sure if that's the real reason). In any case, when this fails the revision will not show up on CDash and the status link here on github will be wrong (it won't have a build id).
- This does *not* merge branches with master before building, unlike pycicle.

There are a couple of small code changes that popped up while setting this up, and I've included those changes here. I've also changed most tests that requested `all` threads to just use `4` (some tests specifically test that `all` works correctly and those I've left alone).

It's meant to build all pull requests from STEllAR-GROUP users automatically. For untrusted users jenkins will ask someone to verify the patch, to which an admin can reply "ok to test" to trigger one build or "add to whitelist" to permanently whitelist that user.

This now sets the github status link to point directly to the CDash build logs (modulo that potential CDash bug mentioned above). The jenkins "default" status points to the actual jenkins logs. These are unfortunately only available for CSCS users (e.g. with CSCS VPN).

Since the configurations are now in this repo, feel free to change build options if you think something is not being tested. To change dependencies you'll have to ask one of the CSCS people to install a particular version or package.

### DIFF
--- a/.jenkins/cscs/Jenkinsfile
+++ b/.jenkins/cscs/Jenkinsfile
@@ -1,0 +1,58 @@
+#!groovy
+
+// Copyright (c) 2020 ETH Zurich
+//
+// SPDX-License-Identifier: BSL-1.0
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+pipeline {
+    agent {
+        node {
+            label 'ssl_daintvm1'
+        }
+    }
+    environment {
+        SPACK_ROOT = '/apps/daint/SSL/HPX/spack'
+        GITHUB_TOKEN = credentials('97b30156-2f6f-47fc-94eb-68044d95ec50')
+    }
+    stages {
+        stage('checkout') {
+            steps {
+                dir('hpx') {
+                    checkout scm
+                    echo "Running ${env.BUILD_ID} on ${env.JENKINS_URL}"
+                }
+            }
+        }
+        stage('build') {
+            matrix {
+                axes {
+                    axis {
+                        name 'configuration_name'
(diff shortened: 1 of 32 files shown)


hkaiser:
> There are a couple of small code changes that popped up while setting this up, and I've included those changes here. I've also changed most tests that requested `all` threads to just use `4` (some tests specifically test that `all` works correctly and those I've left alone).

What's the rationale of this change? The tests pass `--hpx:threads=N` on the commandline anyways (which overwrites the config settings).

msimberg:
> > There are a couple of small code changes that popped up while setting this up, and I've included those changes here. I've also changed most tests that requested `all` threads to just use `4` (some tests specifically test that `all` works correctly and those I've left alone).
> 
> What's the rationale of this change? The tests pass `--hpx:threads=N` on the commandline anyways (which overwrites the config settings).

Hmm, indeed, you're right. Thanks for pointing that out. I was seeing a slightly higher number of timeouts and I assumed this was due to tests running in parallel while requesting `--hpx:threads=all`. It's probably unrelated in that case. I'll revert those changes.

@haampie by the way, this is the kind of failures I was seeing running compilation tests in parallel: https://cdash.cscs.ch/testDetails.php?test=23606055&build=115118. make vs ninja could also make a difference in this case. I think I'll make them serial again...

msimberg:
> > There are a couple of small code changes that popped up while setting this up, and I've included those changes here. I've also changed most tests that requested `all` threads to just use `4` (some tests specifically test that `all` works correctly and those I've left alone).
> 
> What's the rationale of this change? The tests pass `--hpx:threads=N` on the commandline anyways (which overwrites the config settings).

I think the actual issue was that I was running tests in parallel but did not use `--hpx:bind=none` for those tests.

This should be ready to go now. The rest are "regular" failures i.e. unrelated to the Jenkins configuration.

PR #4809, opened 2020-07-07
Foreach adapt

gonidelis:
Adapt for_each in order to support for the begin and end iterators to be of different type

### DIFF
--- a/libs/algorithms/include/hpx/parallel/algorithms/for_each.hpp
+++ b/libs/algorithms/include/hpx/parallel/algorithms/for_each.hpp
@@ -26,6 +26,7 @@
 #include <hpx/execution/algorithms/detail/is_negative.hpp>
 #include <hpx/executors/execution_policy.hpp>
 #include <hpx/parallel/algorithms/detail/dispatch.hpp>
+#include <hpx/parallel/algorithms/detail/distance.hpp>
 #include <hpx/parallel/util/detail/algorithm_result.hpp>
 #include <hpx/parallel/util/foreach_partitioner.hpp>
 #include <hpx/parallel/util/loop.hpp>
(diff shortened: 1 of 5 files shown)


libs/algorithms/include/hpx/parallel/algorithms/for_each.hpp
```diff
@@ -26,6 +26,7 @@
 #include <hpx/execution/algorithms/detail/is_negative.hpp>
 #include <hpx/executors/execution_policy.hpp>
 #include <hpx/parallel/algorithms/detail/dispatch.hpp>
```
gonidelis, line 28: Is this needed?

hkaiser: I think it's needed. This is the base class of the algorithm implementation (see `detail::for_each` below)

libs/algorithms/tests/unit/container_algorithms/foreach_adapt.cpp
```diff
@@ -0,0 +1,23 @@
+#include <hpx/hpx_main.hpp>
```
hkaiser, line 1: Please add copyright, license, and SPDX tag.

gonidelis: Sorry, missed it. Just fixed it and clang-formatted the code.

gonidelis:
Should I use the `is_sentinel_for` trait too?

libs/algorithms/include/hpx/parallel/algorithms/for_each.hpp
```diff
@@ -224,10 +224,10 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
         }
         // forward declare the segmented version of for_each_ algorithm
-        template <typename ExPolicy, typename SegIter, typename F,
-            typename Proj>
-        inline typename util::detail::algorithm_result<ExPolicy, SegIter>::type
-        for_each_(ExPolicy&& policy, SegIter first, SegIter last, F&& f,
+        template <typename ExPolicy, typename SegIterB, typename SegIterE,
+            typename F, typename Proj>
+        inline typename util::detail::algorithm_result<ExPolicy, SegIterB>::type
+        for_each_(ExPolicy&& policy, SegIterB first, SegIterE last, F&& f,
```
hkaiser, line 230: You might want to change the implementation of this function as well (just the prototype) ...

gonidelis: @hkaiser Isn't that the prototype? What other part should I change?

hkaiser: See here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/segmented_algorithms/include/hpx/parallel/segmented_algorithms/for_each.hpp#L183-L187

msimberg:
add to whitelist

gonidelis:
**Important Note**: Please don't merge yet. Need to add a couple of doxygen comments.

libs/segmented_algorithms/include/hpx/parallel/segmented_algorithms/for_each.hpp
```diff
@@ -180,23 +180,23 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
         {
             typedef parallel::execution::is_sequenced_execution_policy<ExPolicy>
                 is_seq;
 
             if (first == last)
             {
-                typedef util::detail::algorithm_result<ExPolicy, SegIter>
+                typedef util::detail::algorithm_result<ExPolicy, SegIterB>
                     result;
                 return result::get(std::move(last));
```
hkaiser, line 196: You might want to return `first` here.

PR #4819, opened 2020-07-10
Allow finding Boost components before finding HPX

### LINKED ISSUE #4816: Build of HPX fails when find_package(Boost) is called before FetchContent_MakeAvailable(hpx)
## Expected Behavior

Build succeeds

## Actual Behavior

Build fails because of misconfiguration of HPX's use of filesystem.

```
[  7%] Building CXX object _deps/hpx-build/libs/logging/CMakeFiles/hpx_logging.dir/src/logging.cpp.o                                             
  In file included from /home/kor/development/objects/Debug/lue/_deps/hpx-src/libs/logging/src/logging.cpp:11:0:                                   
  /home/kor/development/objects/Debug/lue/_deps/hpx-src/libs/filesystem/include/hpx/modules/filesystem.hpp:22:10: fatal error: filesystem: No such 
   #include <filesystem>                                                                                                                           
            ^~~~~~~~~~~~                                                                                                                           
  compilation terminated.                                                                                                                          
  _deps/hpx-build/libs/logging/CMakeFiles/hpx_logging.dir/build.make:75: recipe for target '_deps/hpx-build/libs/logging/CMakeFiles/hpx_logging.dir
  make[2]: *** [_deps/hpx-build/libs/logging/CMakeFiles/hpx_logging.dir/src/logging.cpp.o] Error 1                                                 
  make[2]: *** Waiting for unfinished jobs....
```

  1. Setup a project that builds HPX using CMake's FetchContent module
  1. Call `find_package(Boost REQUIRED COMPONENTS filesystem)` before calling the FetchContent stuff
  1. Build project

One way to do this is to clone https://github.com/kordejong/hello_hpx and add `find_package(Boost REQUIRED COMPONENTS filesystem)` to this line: https://github.com/kordejong/hello_hpx/blob/da5a4347bd4f036793b8ecf39bff58a4814b14d3/CMakeLists.txt#L87 and `bash ./debug_build.sh`.

## Specifications

  - HPX Version: Today's master, but I have seen this issue some weeks/months ago too
  - Platform (compiler, OS): Kubuntu 18.04, gcc-7

## Remarks
- In `HPX_SetupBoostFilesystem.cmake`, some configuration logic is conditionally executed, based on whether `Boost::filesystem` is defined. If the parent project has already found Boost filesystem, the target `Boost::filesystem` is defined and HPX is not configured correctly.
- As a user of HPX I can work around this issue by switching the find_package(Boost) and FetchContent stuff around. First build HPX, then find Boost filesystem ourselves.
- The issue is subtle and not easy to figure out. Maybe it would be good if `HPX_SetupBoostFilesystem.cmake` would work correctly even if find_package(Boost) has been called before the build of HPX. Maybe it is just a matter of moving the `hpx_add_config_define_namespace` outside of the conditional?

msimberg:
Attempt to fix #4816. The GitHub actions fetchcontent build should test this now. Let's see what it says. I have not tested the full build locally.

### DIFF
--- a/cmake/HPX_SetupBoostFilesystem.cmake
+++ b/cmake/HPX_SetupBoostFilesystem.cmake
@@ -7,13 +7,19 @@
 if(HPX_FILESYSTEM_WITH_BOOST_FILESYSTEM_COMPATIBILITY)
   # In case find_package(HPX) is called multiple times
   if(NOT TARGET Boost::filesystem)
-    hpx_add_config_define_namespace(
-      DEFINE HPX_FILESYSTEM_HAVE_BOOST_FILESYSTEM_COMPATIBILITY
-      NAMESPACE FILESYSTEM
-    )
-
     find_package(Boost ${Boost_MINIMUM_VERSION} MODULE COMPONENTS filesystem)
+
+    if(NOT Boost_FILESYSTEM_FOUND)
+      hpx_error(
+        "Could not find Boost.Filesystem Provide a boost installation including the filesystem library"
+      )
+    endif()
   endif()
+
+  hpx_add_config_define_namespace(
+    DEFINE HPX_FILESYSTEM_HAVE_BOOST_FILESYSTEM_COMPATIBILITY
+    NAMESPACE FILESYSTEM
+  )
 else()
   if(NOT HPX_WITH_CXX17_FILESYSTEM)
     hpx_error(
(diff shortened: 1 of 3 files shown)


msimberg:
@kordejong would you mind checking that this fixes the problem on your end as well?

kordejong:
> @kordejong would you mind checking that this fixes the problem on your end as well?

Your fix seems fine. Building HPX continues passed the point where it failed before. Great!

PR #4821, opened 2020-07-11
Implement customization point for hpx::copy and hpx::ranges::copy

hkaiser:
- This creates two customization points for the copy algorithm. This
  also makes `hpx::copy` and `hpx::ranges::copy` fully conforming to C++20
  by adding support for sentinels and algorithms that do not expect
  an execution policy as their first argument
- Replace all uses of `hpx::parallel::copy` with the appropriate new
  versions of the algorithms, deprecate `hpx::parallel::copy`
- Same for `copy_if` and `copy_n`
- flyby: add `HPX_NO_UNIQUE_ADDRESS` expanding to `[[no_unique_address]]`
  on systems that support it
- flyby: add check for support of non-type template parameter auto arguments (`HPX_HAVE_NONTYPE_TEMPLATE_PARAMATER_AUTO`)
- flyby: fixing `hpx::functional::tag_invoke` for MSVC and adding
  `hpx::functional::tag_t<>` helper template

This also adds a `hpx::functional::tag<>` template that implements the necessary `tag_invoke` functionality for customization point objects as a CRTP base class (@sithhell please review).

### DIFF
--- a/cmake/HPX_AddConfigTest.cmake
+++ b/cmake/HPX_AddConfigTest.cmake
@@ -466,7 +466,8 @@ endfunction()
 function(hpx_check_for_cxx17_std_nontype_template_parameter_auto)
   add_hpx_config_test(
     HPX_WITH_CXX17_NONTYPE_TEMPLATE_PARAMETER_AUTO
-    SOURCE cmake/tests/cxx17_std_nontype_template_parameter_auto.cpp FILE ${ARGN}
+    SOURCE cmake/tests/cxx17_std_nontype_template_parameter_auto.cpp FILE
+           ${ARGN}
   )
 endfunction()
 
--- a/cmake/HPX_AddConfigTest.cmake
+++ b/cmake/HPX_AddConfigTest.cmake
@@ -463,6 +463,14 @@ function(hpx_check_for_cxx17_shared_ptr_array)
 endfunction()
 
 # ##############################################################################
+function(hpx_check_for_cxx17_std_nontype_template_parameter_auto)
+  add_hpx_config_test(
+    HPX_WITH_CXX17_NONTYPE_TEMPLATE_PARAMETER_AUTO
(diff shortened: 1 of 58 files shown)


hkaiser:
@msimberg I think this is good to go now. The errors shown are from older commits.

msimberg:
This is very problematic: https://cdash.cscs.ch/buildSummary.php?buildid=116367... We'd have to decide if we want to keep trying to fix these issues or if we simply give up on nvcc. It's a big thing to drop support for nvcc but given that these issues keep coming up I don't see how we can reasonably keep supporting it.

hkaiser:
> This is very problematic: https://cdash.cscs.ch/buildSummary.php?buildid=116367... We'd have to decide if we want to keep trying to fix these issues or if we simply give up on nvcc. It's a big thing to drop support for nvcc but given that these issues keep coming up I don't see how we can reasonably keep supporting it.

Moreover, one of the now failing tests is completely unrelated to the changes at hand (I think so, at least). In general, I think that there is no need for us to support nvcc as long as we find a way to separate the compilation of the kernels from HPX. In any case, for now I suggest to disable the failing tests.

msimberg:
> > This is very problematic: https://cdash.cscs.ch/buildSummary.php?buildid=116367... We'd have to decide if we want to keep trying to fix these issues or if we simply give up on nvcc. It's a big thing to drop support for nvcc but given that these issues keep coming up I don't see how we can reasonably keep supporting it.
> 
> Moreover, one of the now failing tests is completely unrelated to the changes at hand (I think so, at least). In general, I think that there is no need for us to support nvcc as long as we find a way to separate the compilation of the kernels from HPX. In any case, for now I suggest to disable the failing tests.

I'm not sure that's so easy (the failures are likely caused by the new customization points and those we do need for compiling kernels), but I agree that we should disable the tests. Luckily the clang cuda builder builds almost all cuda tests and examples successfully so we'll still notice if we break them badly.

hkaiser:
The remaining test failures are unrelated, I think this could be merged now.

Issue #2235, opened 2016-07-05
Concurrent data structures support


biddisco:
Attempting to port some CMS (from CERN c.f. @felicepantaleo, @rovere) software to HPX from TBB has hit a problem - the CMS code makes fairly liberal use of tbb concurrent data structures that we do not have drop-in replacements for in hpx. These need to be created/included in HPX.

- [x] concurrent_queue (boost::lockfree::queue)
- [ ] concurrent_unordered_map
- [ ] concurrent_unordered_set
- [ ] concurrent_vector

also possibly required are replacements for
- [ ] tbb_hash
- [ ] tbb_hasher

For reference - some TBB related functions/classes that are used in the CMS software stack are as follows (this list may be incomplete).
```
$ git gr "tbb::" | grep -o -E "tbb::.*" | tr '<'  ' '  | awk '{print $1}' | sort | uniq -c
      2 tbb::",
      3 tbb::blocked_range
      6 tbb::concurrent_queue
     27 tbb::concurrent_unordered_map
      7 tbb::concurrent_unordered_set
      8 tbb::concurrent_vector
      3 tbb::empty_task*
      1 tbb::parallel_for(
      1 tbb::parallel_for(tbb::blocked_range
      1 tbb::parallel_for(0UL,collseed_size,1UL,theLoop);
      1 tbb::parallel_for_each(tasks.begin(),tasks.end(),invoker
      1 tbb::parallel_for_each(tasksD.begin(),tasksD.end(),invoker
      4 tbb::task
     35 tbb::task*
      1 tbb::task*).
      1 tbb::task*>
      1 tbb::task::allocate_additional_child_of(*pWaitTask))
      9 tbb::task::allocate_root())
      1 tbb::task::destroy(*eventLoopWaitTask);
      6 tbb::task::destroy(*iTask);}
      1 tbb::task::destroy(*iWait);
      1 tbb::task::enqueue(
      1 tbb::task::spawn(*iTask);
      1 tbb::task::spawn(*nextTask);
      3 tbb::task::spawn(*t);
      1 tbb::task::spawn(calc);
      4 tbb::task>
      5 tbb::task_scheduler_init
      2 tbb::task_scheduler_init::default_num_threads();
      1 tbb::task_scheduler_init>
      1 tbb::task_scheduler_init>&
      1 tbb::task_scheduler_init>(1);
      1 tbb::task_scheduler_init>(static_cast
      3 tbb::tbb_hash
[...]

PR #4823, opened 2020-07-13
initial hpx threading support for libcds data structure

weilewei:
This PR 
- might fixe #2235
- is part of work contributing to [Concurrent Data Structure project](https://github.com/STEllAR-GROUP/hpx/wiki/Google-Summer-of-Code-%28GSoC%29-2020#concurrent-data-structure-support) funded through [Google Summer of Code 2020](https://summerofcode.withgoogle.com/projects/#6740781068451840)

## Proposed Changes

  - Add additional hpx thread data structure for libcds thread data structure guarded by `#ifdef HPX_HAVE_LIBCDS`. The support is particularly targeting Hazard Pointer, and Dynamic HP. RCU is still work in progress, however, it does not impact HP integration and current PR. 
  - Enable cmake content fetch to libcds as a subdirectory
  - Add overhead test for libcds HP, which is located in tests/performance/local

To build libcds in HPX, one might need to follow these steps
```
cmake -DHPX_WITH_LIBCDS=ON /path/to/hpx
```
To run the overhead test, one might do

```
make tests.performance.local.libcds_hazard_pointer_overhead
cd bin
./tests.performance.local.libcds_hazard_pointer_overhead
```


msimberg:
add to whitelist

tests/performance/local/CMakeLists.txt
```diff
@@ -25,6 +25,11 @@ if(NOT HPX_WITH_SANITIZERS)
   set(start_stop_FLAGS DEPENDENCIES hpx_timing)
 endif()
 
+if(HPX_WITH_LIBCDS)
+  set(benchmarks ${benchmarks} libcds_hp_overhead)
+  set(libcds_hp_overhead_FLAGS DEPENDENCIES iostreams_component hpx_timing)
```
msimberg, line 30:
```suggestion
  set(libcds_hp_overhead_FLAGS DEPENDENCIES iostreams_component)
```

Linking to individual modules like that should only be done if `NOLIBS` is used.

tests/performance/local/CMakeLists.txt
```diff
@@ -216,6 +221,11 @@ foreach(benchmark ${benchmarks})
 
 endforeach()
 
+if(HPX_WITH_LIBCDS)
+  target_compile_definitions(cds PUBLIC CDS_THREADING_HPX)
```
msimberg, line 225: Should this be set here? Why not in `HPX_SetupLibCDS` or even in CDS itself?

tests/performance/local/CMakeLists.txt
```diff
@@ -216,6 +221,11 @@ foreach(benchmark ${benchmarks})
...
+  target_link_libraries(libcds_hp_overhead_test PRIVATE cds hpx_core)
```
msimberg, line 226:
```suggestion
  target_link_libraries(libcds_hp_overhead_test PRIVATE cds)
```

No need for `hpx_core`, it should get linked to HPX by `add_hpx_executable` above.

tests/performance/local/libcds_hp_overhead.cpp
```diff
@@ -0,0 +1,345 @@
+//  Copyright (c) 2020 Weile Wei
+//  Copyright (c) 2020 John Biddiscombe
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0.
+//  (See accompanying file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+
+#include <cds/gc/hp.h>    // for cds::HP (Hazard Pointer) SMR
+#include <cds/init.h>     // for cds::Initialize and cds::Terminate
```
msimberg, line 10: We usually put external headers after HPX headers and before standard library headers.

tests/performance/local/libcds_hp_overhead.cpp
```diff
@@ -0,0 +1,345 @@
...
+// we use globals here to prevent the delay from being optimized away
+double global_scratch = 0;
+std::uint64_t num_iterations = 0;
+
+///////////////////////////////////////////////////////////////////////////////
+double null_function(bool uselibcds) noexcept
+{
+#ifdef CDS_THREADING_HPX
+    if (uselibcds)
+        cds::threading::Manager::attachThread();
```
msimberg, line 123: Could this not be added to `future_overhead.cpp` instead?

libs/coroutines/include/hpx/coroutines/coroutine.hpp
```diff
@@ -96,6 +97,19 @@ namespace hpx { namespace threads { namespace coroutines {
             return impl_.set_thread_data(data);
         }
 
+#ifdef HPX_HAVE_LIBCDS
+        std::array<std::size_t, 3> get_libcds_data() const
```
msimberg, line 101: A typedef for `std::array<std::size_t, 3>` might be useful.

tests/performance/local/libcds_hp_overhead.cpp
weilewei: If we move libcds test to future_overhead.cpp, then this file might need `#ifdef HPX_HAVE_LIBCDS` and/or `#ifdef CDS_THREADING_HPX` everywhere, and also need add an extra bool input argument `bool libcds` to some measuring functions as well. I wonder will these actions become messy to future_overhead test? Just some thoughts.

libs/coroutines/include/hpx/coroutines/coroutine.hpp
weilewei: Sure, I will fix this later today.

weilewei: Ok, now it is fixed.

CMakeLists.txt
```diff
@@ -930,6 +930,14 @@ if(HPX_WITH_APEX)
     CATEGORY "Profiling"
   )
 endif()
+
+# LibCDS option
+hpx_option(HPX_WITH_LIBCDS BOOL "Enable LibCDS support." OFF)
+if(HPX_WITH_LIBCDS)
+  hpx_add_config_define(HPX_HAVE_LIBCDS) # tell HPX that we use LibCDS
+  include(HPX_SetupLibCDS)
+endif()
```
hkaiser, line 939: Is there some error handling?

libs/coroutines/include/hpx/coroutines/coroutine.hpp
```diff
@@ -96,6 +97,19 @@ namespace hpx { namespace threads { namespace coroutines {
...
         }
 
+#ifdef HPX_HAVE_LIBCDS
+        typedef std::array<std::size_t, 3> cds_type;
+        cds_type get_libcds_data() const
+        {
+            return impl_.get_libcds_data();
+        }
+
+        cds_type set_libcds_data(cds_type data)
```
hkaiser, line 107: You might want to pass the argument as `const&` to avoid unnecessary copies.

libs/coroutines/include/hpx/coroutines/detail/context_base.hpp
```diff
@@ -85,6 +88,12 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
           , m_thread_data(nullptr)
 #else
           , m_thread_data(0)
+#endif
+#ifdef HPX_HAVE_LIBCDS
+          , libcds_thread_data(cds_type{
+                { 0,
+                    0,
+                    0 }})
```
hkaiser, line 96: This formatting seems to be off.

libs/coroutines/include/hpx/coroutines/detail/context_base.hpp
```diff
@@ -98,6 +107,9 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
             delete_tss_storage(m_thread_data);
 #else
             m_thread_data = 0;
+#endif
+#ifdef HPX_HAVE_LIBCDS
+            libcds_thread_data = cds_type{{ 0, 0, 0 }};
```
hkaiser, line 112: Formatting off?

libs/coroutines/include/hpx/coroutines/detail/context_base.hpp
```diff
@@ -229,6 +244,20 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
 #endif
         }
 
+#ifdef HPX_HAVE_LIBCDS
+        cds_type get_libcds_data() const
+        {
+            return libcds_thread_data;
+        }
+
+        cds_type set_libcds_data(cds_type data)
```
hkaiser, line 253: Same as above, pass by `const&`.

libs/coroutines/include/hpx/coroutines/coroutine.hpp
```diff
@@ -96,6 +97,19 @@ namespace hpx { namespace threads { namespace coroutines {
             return impl_.set_thread_data(data);
         }
 
+#ifdef HPX_HAVE_LIBCDS
+        typedef std::array<std::size_t, 3> cds_type;
+        cds_type get_libcds_data() const
```
hkaiser, line 102: Return by `const&` to avoid copies?

libs/coroutines/include/hpx/coroutines/stackless_coroutine.hpp
```diff
@@ -121,6 +122,21 @@ namespace hpx { namespace threads { namespace coroutines {
 #endif
         }
 
+#ifdef HPX_HAVE_LIBCDS
+        typedef std::array<std::size_t, 3> cds_type;
```
hkaiser, line 126: Should all uses of `cds_type` refer t the same using declaration?

libs/coroutines/include/hpx/coroutines/stackless_coroutine.hpp
```diff
@@ -121,6 +122,21 @@ namespace hpx { namespace threads { namespace coroutines {
...
+            return libcds_data_;
+        }
+
+        cds_type set_libcds_data(cds_type data)
+        {
+            cds_type olddata = libcds_data_;
+            libcds_data_ = data;
+            return olddata;
+        }
+#endif
```
hkaiser, line 138: Could these functions go into the base class instead (avoids virtual functions and code duplication)?

libs/threading_base/src/thread_helpers.cpp
```diff
@@ -204,6 +205,33 @@ namespace hpx { namespace threads {
...
     }
 
+#ifdef HPX_HAVE_LIBCDS
+    cds_type get_libcds_data(thread_id_type const& id, error_code& ec)
+    {
+        if (HPX_UNLIKELY(!id))
+        {
+            HPX_THROWS_IF(ec, null_thread_id, "hpx::threads::get_libcds_data",
+                "null thread id encountered");
+            return cds_type{{0, 0, 0}};
```
weilewei, line 215: @hkaiser I agreed with passing by reference and return by reference, but what should I do for this return which returns a copy of cds data if `HPX_UNLIKELY(!id) == true`? Any suggestion?

hkaiser: What about:
```
static cds_type data({{0, 0, 0}});
return data;
```
?

libs/coroutines/include/hpx/coroutines/stackless_coroutine.hpp
weilewei: Would you mind telling which file specifically?

hkaiser: Please ignore this comment, I was mistaken.

weilewei:
I modified the underlying data structure for thread data for libcds, removed std::array<std::size_t, 3>, and now use three individual elements representing thread different thread-private data: thread manager, hazard pointer, dynamic hazard pointers. 

Because these three elements are independent of each other, when writing and reading one of the elements, we really do not need to write/read three elements together.

libs/coroutines/include/hpx/coroutines/coroutine.hpp
```diff
@@ -98,16 +97,35 @@ namespace hpx { namespace threads { namespace coroutines {
...
 #ifdef HPX_HAVE_LIBCDS
-        typedef std::array<std::size_t, 3> cds_type;
-        cds_type get_libcds_data() const
+        std::size_t& get_libcds_data() const
         {
             return impl_.get_libcds_data();
         }
 
-        cds_type set_libcds_data(cds_type data)
+        std::size_t& set_libcds_data(std::size_t& data)
```
hkaiser, line 105: What's your rationale of using references for the parameter and return value in this case?

hkaiser:
> I modified the underlying data structure for thread data for libcds, removed std::array<std::size_t, 3>, and now use three individual elements representing thread different thread-private data: thread manager, hazard pointer, dynamic hazard pointers.
> 
> Because these three elements are independent of each other, when writing and reading one of the elements, we really do not need to write/read three elements together. This is much easier for retrieving data, debugging, and providing straight-forward API for libcds end.

I like this change. Could we use more descriptive function names, however? `libcds_data`, `libcds_hazard_pointer_data`, and `libcds_dynamic_hazard_pointer_data` sounds good enough for me, but anything else descriptive would be fine as well.

libs/coroutines/include/hpx/coroutines/coroutine.hpp
weilewei: Oh, should be a typo, I meant const& everywhere, but I forgot to add it, will add soon.

hkaiser: That makes it only slightly better. Why do you think you need references in the first place?

weilewei: Oh, because you commented here saying adding const references to avoid copy: https://github.com/STEllAR-GROUP/hpx/pull/4823#discussion_r453971287 and https://github.com/STEllAR-GROUP/hpx/pull/4823#discussion_r453972807

hkaiser: @weilewei `sizeof(std::array<3, size_t>) == 3 * sizeof(std::array<3, size_t> const&)`, however `sizeof(size_t) == sizeof(size_t const&)`. Thus, if there is no size benefit of using references, then value semantics are preferable any day. Moreover, returning a reference from a function forces you to use a function-local static in the implementation which will create a data race as this variable would be shared in between threads.

weilewei: Thanks for the clarification, ok, then I will stick to value semantics then, pass by value and return by value.

weilewei:
Please review it again, thanks

CMakeLists.txt
weilewei: Added

libs/coroutines/include/hpx/coroutines/detail/context_base.hpp
```diff
@@ -48,6 +48,7 @@
 #include <hpx/coroutines/detail/tss.hpp>
 #include <hpx/coroutines/thread_id_type.hpp>
 
+#include <array>
```
hkaiser, line 51: Do we still need this `#include <array>`?

libs/coroutines/include/hpx/coroutines/detail/context_base.hpp
```diff
@@ -204,6 +214,11 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
             delete_tss_storage(m_thread_data);
 #else
             m_thread_data = 0;
+#ifdef HPX_HAVE_LIBCDS
+            libcds_data_ = 0;
+            libcds_hazard_pointer_data_ = 0;
+            libcds_dynamic_hazard_pointer_data_ = 0;
+#endif
```
hkaiser, line 221: I think this #if/#endif should be outside the `HPX_HAVE_THREAD_LOCAL_STORAGE` condition above.

libs/coroutines/include/hpx/coroutines/detail/context_base.hpp
```diff
@@ -229,6 +244,44 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
...
 
+#ifdef HPX_HAVE_LIBCDS
+        std::size_t get_libcds_data() const
+        {
+            return libcds_data_;
+        }
+
+        std::size_t set_libcds_data(std::size_t data)
+        {
+            static std::size_t olddata = libcds_data_;
```
hkaiser, line 255: This creates a data race as it is shared amongst threads. Same is true for the other functions below.

hkaiser: If in doubt, you could write it as:
```
std::size_t set_libcds_data(std::size_t data)
{
    std::swap(data, libcds_data_);
    return data;
}
```

libs/coroutines/include/hpx/coroutines/detail/context_base.hpp
```diff
@@ -345,6 +403,11 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
         mutable detail::tss_storage* m_thread_data;
 #else
         mutable std::size_t m_thread_data;
+#ifdef HPX_HAVE_LIBCDS
+        mutable std::size_t libcds_data_;
+        mutable std::size_t libcds_hazard_pointer_data_;
+        mutable std::size_t libcds_dynamic_hazard_pointer_data_;
+#endif
```
hkaiser, line 410: This should be outside the condition `HPX_HAVE_THREAD_LOCAL_STORAGE` above.

libs/coroutines/include/hpx/coroutines/stackless_coroutine.hpp
```diff
@@ -18,6 +18,7 @@
 #include <hpx/functional/unique_function.hpp>
 #include <hpx/type_support/unused.hpp>
 
+#include <array>
```
hkaiser, line 21: Do we still need this `#include <array>`?

libs/coroutines/include/hpx/coroutines/stackless_coroutine.hpp
```diff
@@ -121,6 +122,44 @@ namespace hpx { namespace threads { namespace coroutines {
...
 
+#ifdef HPX_HAVE_LIBCDS
+        std::size_t get_libcds_data() const
+        {
+            return libcds_data_;
+        }
+
+        std::size_t set_libcds_data(std::size_t data)
+        {
+            static std::size_t olddata = libcds_data_;
```
hkaiser, line 133: Same problem as above: the function-local static will create a data race.

libs/threading_base/include/hpx/threading_base/thread_helpers.hpp
```diff
@@ -20,6 +20,7 @@
 #include <hpx/threading_base/thread_pool_base.hpp>
 #include <hpx/timing/steady_clock.hpp>
 
+#include <array>
```
hkaiser, line 23: Still needed?

tests/performance/local/libcds_hp_overhead.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+// we use globals here to prevent the delay from being optimized away
+double global_scratch = 0;
+std::uint64_t num_iterations = 0;
+
+///////////////////////////////////////////////////////////////////////////////
+double null_function(bool uselibcds) noexcept
+{
+#ifdef CDS_THREADING_HPX
```
hkaiser, line 121: If libCDS is not configured to support HPX threads this test should not be enabled in the first place.

libs/coroutines/include/hpx/coroutines/detail/context_base.hpp
```diff
@@ -204,6 +214,11 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
             delete_tss_storage(m_thread_data);
 #else
             m_thread_data = 0;
+#endif
+#ifdef HPX_HAVE_LIBCDS
```
msimberg, line 218: Minor style detail: We tend to use `#if defined(X)` instead of `#ifdef X`.

msimberg:
retest

msimberg:
retest

libs/coroutines/include/hpx/coroutines/detail/context_base.hpp
weilewei: Thanks, I just adapted the style. Please check again.

PR #4826, opened 2020-07-14
Add Jenkins configuration for rostam

msimberg:


.jenkins/lsu/set_github_status.sh
```diff
@@ -0,0 +1,23 @@
...
+commit_status=${4}
+configuration_name=${5}
+build_id=${6}
+
+curl --verbose \
+    --request POST \
+    --url "https://api.github.com/repos/${commit_repo}/statuses/${commit_sha}" \
+    --header 'Content-Type: application/json' \
+    --header "authorization: Bearer ${github_token}" \
+    --data "{ \"state\": \"${commit_status}\", \"target_url\": \"https://cdash.cscs.ch/buildSummary.php?buildid=${build_id}\", \"description\": \"Jenkins\", \"context\": \"jenkins/cscs/${configuration_name}\" }"
```
hkaiser, line 23: `jenkins/lsu/${configuration_name}`?

msimberg: Yep... Thanks!

msimberg:
Jobs are now working by the way, but I only have one configuration in place. My plan is to fill the gaps between the configurations tested on Piz Daint, i.e. test:
- clang 6 to 9 with one Boost version each
- gcc 8 to 9 with one Boost version each
- cuda 11 with gcc 10
- all configurations with MPI (some with OpenMPI and some with MPICH) and release/debug
- at least one config with cuda on and distributed runtime off (as in #4839).

@hkaiser, @biddisco, @rori, @sithhell please comment if you think there are other "special" configurations that you'd like to see tested.

I had in mind that these configurations would be enough even for master, and that they would be run for each and every PR (we'll have to see how rostam keeps up with that; these are more configurations per PR, but fewer than before on master).

msimberg:
retest

msimberg:
This is now ready to go (at least the basic functionality works, we may tweak it over time). The CUDA tests don't seem to run correctly, but that is for a separate PR, if we decide to fix it...

biddisco:
looks like the test is failing on daint - 
https://cdash.cscs.ch/testDetails.php?test=24783653&build=118817
I wonder if I need to make sure that srun is used in place of mpi, or if the test is failing because the localities is wrong (it is using 1 instead of 2). I will check the test results of this PR when they come in ...

msimberg:
> looks like the test is failing on daint -
> https://cdash.cscs.ch/testDetails.php?test=24783653&build=118817
> I wonder if I need to make sure that srun is used in place of mpi, or if the test is failing because the localities is wrong (it is using 1 instead of 2). I will check the test results of this PR when they come in ...

I think you meant rostam ;) I think this result is before your fix and the number of localities were incorrectly set to first 2 and then 1. Let's wait for the next round to see if that fixes it.

biddisco:
Looks like rostam might be only allocating one node, then the test asks for 2 and an error is being triggered. Not completely sure actually, but can add a flag to the rostam setup to allow 2 ranks per node just in case?

msimberg:
@biddisco I was just looking at this and ended up disabling the test for now so that we can get this in, but I'd appreciate your help in figuring out if it's the configuration or the test that's not working. We should still re-enable it before the release. The build does get two nodes allocated and all other "normal" MPI parcelport tests do pass (although they may still only be using one node, but definitely two ranks since some of the tests require that and would fail otherwise). Does the test pass for you locally? Actually, it would probably be enough to allocate a single node and use two ranks/localities on the node for tests to avoid reserving two full nodes, but that should be unrelated to this.

biddisco:
I don't think disabling the test is a good idea. On rostam I see different invocations (mpich/openmpi) and I'm not sure which one to actually look at to diagnose what's wrong. is it possible to run the tests by hand from the jenkins build dir and see what's up? (clearly no, if the test is disabled).

biddisco:
looking at this page
https://cdash.cscs.ch/testSummary.php?project=2&name=tests.unit.modules.async_mpi.mpi_ring_async_executor&date=2020-07-22
it would seem that hpx compiled with mpich (bottom of page) is/was passing the test, but with openmpi failing and when the number of ranks was set to 1, it failed with both.  

All the gcc-10 builds appear to have passed, but builds with clang and earlier gcc failed. What I can't see is which ones explicitly enabled async_mpi and other options etc. the gcc-10 build were all launched with the correct N=2 mpi ranks, but others with the same hash appear to have N=1 - which may be the bug we have just addressed with the locality setting. 

Enabling the test again would help  to spot a pattern of fails/passes

msimberg:
@biddisco found a problem in the `async_mpi` test. I'll merge this as it is and he will re-enable the test in a later PR.

PR #4827, opened 2020-07-14
Update clang-newest configuration to use clang 10

msimberg:

### DIFF
--- a/libs/affinity/include/hpx/affinity/detail/partlit.hpp
+++ b/libs/affinity/include/hpx/affinity/detail/partlit.hpp
@@ -194,7 +194,7 @@ namespace hpx { namespace threads { namespace detail {
 
     private:
         // silence MSVC warning C4512: assignment operator could not be generated
-        partial_literal_string& operator=(partial_literal_string const&);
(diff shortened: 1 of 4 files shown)


libs/affinity/include/hpx/affinity/detail/partlit.hpp
```diff
@@ -194,7 +194,7 @@ namespace hpx { namespace threads { namespace detail {
 
     private:
         // silence MSVC warning C4512: assignment operator could not be generated
-        partial_literal_string& operator=(partial_literal_string const&);
+        partial_literal_string& operator=(partial_literal_string const&) = default;
```
hkaiser, line 197: This might have to be `=delete` if anything. I'll investigate.

msimberg:
test

libs/affinity/include/hpx/affinity/detail/partlit.hpp
hkaiser: I removed this declaration altogether as recent versions of the compiler don't generate the bogus warning anymore.

msimberg: Thank you!

PR #4830, opened 2020-07-15
Update documentation on using HPX::wrap_main

### LINKED ISSUE #4619: Fully document hpx_wrap behaviour and targets
Before I forget:

- Document all three targets `HPX::hpx`, `HPX::hpx_no_wrap_main`, and `HPX::wrap_main`
- Add troubleshooting entry for the common error messages resulting from not using the targets above correctly

msimberg:
Fixes #4619. I didn't mention anything about the fancier variations of linking with e.g. `gtest_main` since it's not portable. If it's going to be something more people want to do we can make it a bit more offical, and describe it in detail, but for now I'd like to not encourage more people to do so than the ones who have already tried to do so.

### DIFF
--- a/docs/sphinx/manual/creating_hpx_projects.rst
+++ b/docs/sphinx/manual/creating_hpx_projects.rst
@@ -236,11 +236,13 @@ Additionally, if you wish to require |hpx| for your project, replace the
 
 You can check if |hpx| was successfully found with the ``HPX_FOUND`` CMake variable.
 
-Using |hpx| targets
--------------------
+.. _using_hpx_cmake_targets:
+
+Using |cmake| targets
+---------------------
 
 The recommended way of setting up your targets to use |hpx| is to link to the
-``HPX::hpx`` target:
+``HPX::hpx`` |cmake| target:
 
 .. code-block:: cmake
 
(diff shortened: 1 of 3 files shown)


docs/sphinx/manual/creating_hpx_projects.rst
```diff
@@ -254,8 +256,14 @@ This requires that you have already created the target like this:
    target_include_directories(hello_world_component PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 
 When you link your library to ``HPX::hpx`` you will be able use |hpx|
```
hkaiser, line 258: Should we mention that `HPX::hpx` is a cmake target exported by the HPX cmake build system setup and that linking against it in this context means using a `target_link_libraries(<user_target> PRIVATE/PUBLIC HPX::hpx)`?

docs/sphinx/manual/starting_the_hpx_runtime.rst
```diff
@@ -49,15 +49,19 @@ The only change to your code you have to make is to include the file
 ``hpx/hpx_main.hpp``. In this case the function ``main()`` will be invoked as
 the first |hpx| thread of the application. The runtime system will be
 initialized behind the scenes before the function ``main()`` is executed and
-will automatically stop after ``main()`` has returned. All |hpx| API functions
-can be used from within this function now.
+will automatically stop after ``main()`` has returned. For this method to work
+you must link your application to ``HPX::wrap_main``. This is done automatically
```
hkaiser, line 53: We might want to be a bit more specific:
```suggestion
you must link your application to the cmake target ``HPX::wrap_main``. This is done automatically
```

docs/sphinx/quickstart.rst
```diff
@@ -179,14 +179,19 @@ build an executable using |cmake| and |hpx|:
...
 
 .. note::
 
    You will most likely have more than one ``main.cpp`` file in your project.
    See the section on :ref:`using_hpx_cmake` for more details on how to use
    ``add_hpx_executable``.
 
+.. note::
+
+   ``HPX::wrap_main`` is required if you are implicitly using ``main`` as the
```
hkaiser, line 192: This is 'required' on Linux/MacOS only. But I guess is doesn't do any arm on other systems

docs/sphinx/manual/creating_hpx_projects.rst
msimberg: Yes, good idea. I'll at least explicitly say `the CMake target ``HPX::hpx`` `.

docs/sphinx/quickstart.rst
msimberg: Yeah, I know, but I'd rather that users link to it anyway since that'll make things work everywhere. On top of that it would give us the freedom to move `hpx/hpx_main.hpp`/`hpx/wrap_main.hpp` to a separate include directory so that not linking to `HPX::wrap_main` would already be a compiler error rather than a linker error.

docs/sphinx/manual/creating_hpx_projects.rst
msimberg: Updated here and there. `CMake target` is mentioned many times now, including in the section heading ;)

PR #4832, opened 2020-07-16
Fixing assertion in split_gids and memory leaks in 1d_stencil_7

### LINKED ISSUE #4732: Failing assertion `split_gids_.empty()`, when running `1d_stencil_7`
## Expected Behavior

Successful execution

## Actual Behavior

When running the example on multiple localities an assertion fails. This happens in tests with my own code, but I can reproduce the issue with `1d_stencil_7`.

```
{stack-trace}: 4 frames:                                                                                       
0x2aaaab48351b  : ??? + 0x2aaaab48351b in /quanta1/home/jong0137/development/object/Debug/lue/_deps/hpx-build/lib/libhpxd.so.1
0x2aaaab4835e3  : ??? + 0x2aaaab4835e3 in /quanta1/home/jong0137/development/object/Debug/lue/_deps/hpx-build/lib/libhpxd.so.1
0x2aaaab480f47  : hpx::detail::custom_exception_info(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, long, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0x51 in /quanta1/home/jong0137/development/object/Debug/lue/_deps/hpx-build/lib/libhpxd.so.1
0x2aaaab4a9550  : ??? + 0x2aaaab4a9550 in /quanta1/home/jong0137/development/object/Debug/lue/_deps/hpx-build/lib/libhpxd.so.1
{locality-id}: 0                                                                                               
{hostname}: [ (mpi:0) (tcp:10.141.0.34:7910) ]                                                                 
{process-id}: 18915                                                                                            
{os-thread}: 1, locality#0/worker-thread#1                                                                     
{thread-id}: 00000000e40d4cb0                                                                                  
{thread-description}: hpx::parallel::execution::parallel_executor::post                                        
{state}: state_running                                                                                         
{auxinfo}:·                                                                                                    
{file}: /quanta1/home/jong0137/development/object/Debug/lue/_deps/hpx-src/hpx/runtime/serialization/detail/preprocess_gid_types.hpp
{line}: 48                                                                                                     
{function}: hpx::serialization::detail::preprocess_gid_types::~preprocess_gid_types()                          
{what}: Assertion 'split_gids_.empty()' failed: HPX(assertion_failure)
```

  1. Use a debug build of HPX 1.4.1
  1. Run `1d_stencil_7` on multiple localities (see example script attached, which I used)

## Specifications

  - HPX Version: 1.4.1
  - Platform (compiler, OS): gcc-7, Kubuntu 16.04

### LINKED ISSUE #4746: Memory leak when using dataflow icw components

hkaiser:
Fixes #4732
Fixes #4746

The two mentioned issues were connected. The cause for the memory leak (and as a side effect for the assertions) was that parcel preprocessing is done on a parcel-by-parcel basis, while the actual serialization may store several parcels using the same serialization archive. This caused (managed) `id_types` that were part of two or more parcels to have their credits split two or more times but subsequently they were serialized only once (because of pointer tracking). As a result the credits were leaked that were supposed to be sent over the wire by the second (or subsequent) instance of the same `id_type`.

This patch fixes this by giving back to AGAS all credits that end up not being sent. 

This patch also fixes a possible race condition between handling the credit splitting and the corresponding future during parcel preprocessing.

@kordejong please verify that the memory leaks are gone for you as well (for me their appear to be fixed).

### DIFF
--- a/examples/1d_stencil/1d_stencil_7.cpp
+++ b/examples/1d_stencil/1d_stencil_7.cpp
@@ -301,7 +301,7 @@ struct stepper
 
                     // The new partition_data will be allocated on the same locality
                     // as 'middle'.
-                    return partition(middle.get_id(), next);
+                    return partition(middle.get_id(), std::move(next));
                 }
             ),
             std::move(next_middle),
@@ -342,6 +342,10 @@ stepper::space stepper::do_work(std::size_t np, std::size_t nx, std::size_t nt)
     for (std::size_t i = 0; i != np; ++i)
         U[0][i] = partition(localities[locidx(i, np, nl)], nx, double(i));
 
+    // limit depth of dependency tree
+    std::size_t nd = 3;
+    hpx::lcos::local::sliding_semaphore sem(nd);
+
     heat_part_action act;
     for (std::size_t t = 0; t != nt; ++t)
     {
@@ -357,6 +361,20 @@ stepper::space stepper::do_work(std::size_t np, std::size_t nx, std::size_t nt)
                     current[idx(i, -1, np)], current[i], current[idx(i, +1, np)]
                 );
         }
+
+        if ((t % nd) == 0)
+        {
+            next[0].then(
+                [&sem, t](partition&&)
+                {
+                    // inform semaphore about new lower limit
+                    sem.signal(t);
+                });
+        }
+
+        // suspend if the tree has become too deep, the continuation above
+        // will resume this thread once the computation has caught up
+        sem.wait(t);
     }
 
     // Return the solution at time-step 'nt'.
--- a/hpx/runtime/naming/name.hpp
+++ b/hpx/runtime/naming/name.hpp
@@ -415,6 +415,8 @@ namespace hpx { namespace naming
         std::uint64_t id_msb_;
         std::uint64_t id_lsb_;
     };
+
+    HPX_EXPORT void decrement_refcnt(gid_type const& gid);
 }}
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/hpx/runtime/serialization/detail/preprocess_futures.hpp
+++ b/hpx/runtime/serialization/detail/preprocess_futures.hpp
@@ -1,4 +1,4 @@
-//  Copyright (c) 2015-2019 Hartmut Kaiser
+//  Copyright (c) 2015-2020 Hartmut Kaiser
 //  Copyright (c) 2015-2016 Thomas Heller
 //
 //  SPDX-License-Identifier: BSL-1.0
@@ -8,11 +8,11 @@
 #pragma once
 
 #include <hpx/assert.hpp>
-#include <hpx/modules/datastructures.hpp>
 #include <hpx/futures/future.hpp>
-#include <hpx/lcos_local/promise.hpp>
-#include <hpx/synchronization/spinlock.hpp>
 #include <hpx/lcos_fwd.hpp>
+#include <hpx/lcos_local/promise.hpp>
+#include <hpx/modules/datastructures.hpp>
+#include <hpx/synchronization/spinlock.hpp>
 
 #include <cstddef>
 #include <mutex>
@@ -80,6 +80,9 @@ namespace hpx { namespace serialization { namespace detail {
             {
                 std::lock_guard<mutex_type> l(mtx_);
                 ++triggered_futures_;
+
+                // trigger the promise only after the whole serialization
+                // operation is done and all futures have become ready
                 set_value = (done_ && num_futures_ == triggered_futures_);
             }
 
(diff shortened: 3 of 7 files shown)


kordejong:
> @kordejong please verify that the memory leaks are gone for you as well (for me their appear to be fixed).

- I cannot reproduce #4732 anymore. Great!
- I also cannot reproduce #4746 with patched stencil example 7 anymore. Memory usage is lower and constant now. Also great!
- I noticed that on Release builds, the (unpatched) stencil 7 example crashes (segmentation fault) when changing the tree depth argument of the sliding semaphore. You can reproduce this with using `nt` instead of 3 as an argument and the settings from the script below. Using `nd = 5000` doesn't work. Using `nd = 500` does work. I cannot reproduce this using a Debug build. And also not on a Release build with only a single locality.
- I noticed an imbalance in memory usage when using multiple localities. In the case of four localities one uses more memory than the others, which makes sense, I guess, given one executes main and handle tasks, and the others only handle tasks. But the other three don't use the same amount of memory. Given the script below and using 4 localities with one thread each, one locality (of the three non-root localities) uses almost 25% less memory than the other. I am not sure whether this is a result of a load imbalance or something else.

Thanks a lot for looking into this! This fix seems to be a big improvement for my use-case (and for HPX ;-) ). I will continue testing it with my own code now.

```
#!/usr/bin/env bash
set -e


# → Update this path to local conventions
install_prefix=$LUE_OBJECTS/_deps/hpx-build

nr_partitions=500
partition_size=16000
nr_time_steps=2000

nr_threads_per_locality=2
nr_localities=2

python $install_prefix/bin/hpxrun.py \
    --parcelport=tcp --verbose \
    --localities=$nr_localities --thread=$nr_threads_per_locality \
        $install_prefix/bin/1d_stencil_7 -- \
            --hpx:print-bind  \
            --np=$nr_partitions \
            --nx=$partition_size \
            --nt=$nr_time_steps
```

PR #4834, opened 2020-07-16
Fixing the calculation of the number of idle cores and the corresponding idle masks

hkaiser:
@scheherzade Please check whether this fixes the issue you encountered with calculating the idle masks


msimberg:
retest

scheherzade:
@hkaiser It works very well, thanks!

PR #4839, opened 2020-07-18
Fix async_cuda build problems when distributed runtime is disabled

biddisco:


libs/async_cuda/src/cuda_future.cpp
```diff
@@ -14,9 +14,10 @@
 #include <hpx/async_cuda/target.hpp>
 #include <hpx/futures/traits/future_access.hpp>
 #include <hpx/modules/errors.hpp>
-#include <hpx/modules/memory.hpp>
```
hkaiser, line 17: I think this is still needed as this file references the `internal_allocator` exposed by the memory module.

libs/async_cuda/src/cuda_target.cpp
```diff
@@ -11,9 +11,10 @@
 #include <hpx/async_cuda/target.hpp>
 #include <hpx/futures/traits/future_access.hpp>
 #include <hpx/modules/errors.hpp>
-#include <hpx/modules/memory.hpp>
+#if defined(HPX_HAVE_DISTRIBUTED_RUNTIME)
 #include <hpx/runtime/find_here.hpp>
 #include <hpx/runtime/naming/id_type_impl.hpp>
+#endif
 #include <hpx/runtime_fwd.hpp>
 #include <hpx/threading_base/thread_helpers.hpp>
```
hkaiser, line 20: Should the `#if`/`#endif` section below be adapted to `HPX_HAVE_DISTRIBUTED_RUNTIME` as well?

libs/async_cuda/src/cuda_future.cpp
biddisco: it's in 
```
#include <hpx/allocator_support/internal_allocator.hpp>
```
which is included on line 11, so I think it's safe to leave out the memory module.

libs/async_cuda/src/cuda_target.cpp
biddisco: I'm not sure which 'section below' you are referring to. Could you please elaborate. thanks

hkaiser: I'm referring to the `#if`/`#endif` starting at line 21. Github didn't allow me to add comments to unchanged code blocks.

libs/async_cuda/src/cuda_future.cpp
hkaiser: Ok, I didn't notice that. Thanks!

libs/async_cuda/src/cuda_target.cpp
biddisco: sorry - I was looking at the wrong file. I see it now. I think probably yes, but I didn't get any compilation problems, so I didn't look further at it. I will experiment tomorrow with it.

msimberg:
Thanks @biddisco for catching this. This is a good candidate for testing on rostam. Could you please comment on #4826 if there are other configurations that you'd like to see tested. I fear that `HPX_WITH_IO_POOL=OFF` would require significant changes or disabling of tests to test at the moment.

libs/async_cuda/src/cuda_target.cpp
msimberg: I don't think it needs to be adapted. Serialization is available even with the distributed runtime off.

PR #4840, opened 2020-07-19
Fixing TSS problems during thread termination

hkaiser:
Fixes the TSS test problems that found their way into master.

### DIFF
--- a/libs/coroutines/src/detail/coroutine_impl.cpp
+++ b/libs/coroutines/src/detail/coroutine_impl.cpp
@@ -74,9 +74,6 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
                     result_last = m_fun(*this->args());
                     HPX_ASSERT(
                         result_last.first == thread_state_enum::terminated);
-
-                    // Reset early as the destructors may still yield.
-                    this->reset();
                 }
                 catch (...)
                 {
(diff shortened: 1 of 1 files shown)


libs/coroutines/src/detail/coroutine_impl.cpp
```diff
@@ -74,17 +74,16 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
...
                 }
                 catch (...)
                 {
                     status = super_type::ctx_exited_abnormally;
                     tinfo = std::current_exception();
                 }
 
+                // Reset early as the destructors may still yield.
                 this->reset_tss();
+                this->reset();
```
msimberg, line 86: Thanks for looking into this! Do you have a stacktrace from the test that failed (`tests.unit.modules.threading.tss` I assume?)? I'm just wondering if both of these should be inside the `try` block. The comment that I added is actually understating things a bit. The destructors can call *arbitrary user code*, including throw exceptions. The question is if that's allowed as well. The second thing is if the destructors end up using tss that will now be reset before the destructors are called (but I assume that's what lead to the test failure in the first place). Could we call `reset_tss` inside `reset` just before the coroutine id gets reset?

hkaiser: Destructors shouldn't call exceptions, should they? In any case, I moved that out of the `try` block to avoid having the `reset` twice, once there and once inside the `catch` block.

The problem I was trying to fix was that the code was trying to use the `thread_id` during the execution of `reset_tss`.

PR #4841, opened 2020-07-20
Rearrange public API headers

msimberg:
As discussed with @hkaiser and @aurianer. We decided to change `hpx/x.hpp` headers to include both distributed and local headers wherever it fits (like `future.hpp` and `runtime.hpp`, more in the future), and the distributed header is conditionally included depending on if the distributed runtime is enabled in the configuration.

Users that really always only want the local functionality should include `hpx/local/x.hpp`. Not everything has a header in `hpx/local` or `hpx/distributed` as it doesn't make sense for everything. Should we add dummy headers `hpx/local/x.hpp`/`hpx/distributed/x.hpp` (that would just include the same as `hpx/x.hpp`) for some sort of symmetry? (I vote no, since it's just more to maintain and the cases where the split is needed will be listed in the documentation.)

Extra: I moved most of the API headers to the `include` module where it's a bit easier to have an overview of them.


msimberg:
Flyby: I had completely forgotten to add most of the synchronization primitives to the API page, now they're there.

Question: Should we already add the trio `channel`, `barrier`, `latch` to the `hpx::distributed` namespace (and headers)? I see no reason why not. These are most likely some of the more stable distributed LCOs we have. I can't judge how much else from the `collectives` modules should be added to a non-experimental namespace though.

biddisco:
I find this a bit strange. Why not have 
```
#include <hpx/x.hpp>
```
and inside the `x.hpp` add 
```
#if defined(HPX_HAVE_DISTRIBUTED_RUNTIME)
#include <hpx/distributed/x.hpp>
```
and not leave it to chance that the user 'knows' every module that has local and distributed versions?

msimberg:
> I find this a bit strange. Why not have
> 
> ```
> #include <hpx/x.hpp>
> ```
> 
> and inside the `x.hpp` add
> 
> ```
> #if defined(HPX_HAVE_DISTRIBUTED_RUNTIME)
> #include <hpx/distributed/x.hpp>
> ```
> 
> and not leave it to chance that the user 'knows' every module that has local and distributed versions?

I'm quite honestly unsure if we agree or disagree here. That sounds exactly like what I've done. See e.g. `hpx/future.hpp`: https://github.com/STEllAR-GROUP/hpx/blob/eed6733f835bfbc19242e419d0aed5832b43c824/libs/include/include/hpx/future.hpp. Are you saying that *every* `hpx/x.hpp` should also have a `hpx/local/x.hpp` and `hpx/distributed/x.hpp` no matter if there actually is a distinction? The users who don't want to care can include just `hpx/x.hpp`, and the users who absolutely want to avoid including distributed features would have to (currently) know which headers have a local variant (or they just check in the documentation).

biddisco:
Sorry. I completely misunderstood the PR and got the diffs the wrong way around. Please proceed and ignore my comments.

PR #4844, opened 2020-07-20
Add HPX-LibCDS example: flat combining priority queue

weilewei:
Adding flat combining priority queue (fc-pq) in HPX threads to demonstrate the fc-pq can sort a list of integers in order with hpx threading support.

- The example is at the draft stage just giving an idea how to use fc-pq, feel free to give suggestions on how to polish it
- These example files, as well as libcds performance files, might need to move under libcds modules @biddisco 

## How to use
```
cmake -DHPX_WITH_LIBCDS=ON -DHPX_WITH_EXAMPLES=ON /path/to/hpx
```
```
cd build/
make examples.libcds.libcds_flat_combining_priority_queue
./bin/examples.libcds.libcds_flat_combining_priority_queue
```


weilewei:
Not matured enough. Closed.

PR #4847, opened 2020-07-21
Fixing 'error prone' constructs as reported by Codacy

hkaiser:


msimberg:
Just added a missing header reported by inspect. Thanks @hkaiser!

PR #4848, opened 2020-07-21
Allow combined MPI and/or CUDA polling for futures

biddisco:
The polling functions will hardly ever change during normal
operation, so to avoid taking a lock, we use an function pointer
for each of the polling functions for MPI and CUDA separately.
This means they can be set/cleared by any thread and when unchanged
no cache lines are affected and overheads are minimized.

### DIFF
--- a/libs/threading_base/include/hpx/threading_base/scheduler_base.hpp
+++ b/libs/threading_base/include/hpx/threading_base/scheduler_base.hpp
@@ -276,15 +276,38 @@ namespace hpx { namespace threads { namespace policies {
             return thread_queue_init_.small_stacksize_;
         }
 
-        void set_user_polling_function(util::function_nonser<void()> const& f)
+        using PollingFunctionPtr = void (*)();
+
+        static void null_polling_function() {}
+
+        void set_mpi_polling_function(PollingFunctionPtr mpi_func)
         {
-            user_polling_function_ = f;
+            polling_function_mpi_ = mpi_func;
         }
 
-        void user_polling_function()
+        void clear_mpi_polling_function()
         {
-            if (user_polling_function_)
-                user_polling_function_();
+            polling_function_mpi_ = &null_polling_function;
+        }
+
+        void set_cuda_polling_function(PollingFunctionPtr cuda_func)
+        {
+            polling_function_cuda_ = cuda_func;
+        }
+
+        void clear_cuda_polling_function()
+        {
+            polling_function_cuda_ = &null_polling_function;
+        }
+
+        inline void custom_polling_function() const
+        {
+#if defined(HPX_HAVE_LIB_ASYNC_MPI)
+            (*polling_function_mpi_)();
+#endif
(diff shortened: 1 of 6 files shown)


biddisco:
Don't review yet - I will (force) push this again later. I just realized, I didn't delete the old user polling vector and forgot to remove the scheduler mode flag that we don't really need any more.

libs/threading_base/include/hpx/threading_base/scheduler_base.hpp
```diff
@@ -276,15 +276,38 @@ namespace hpx { namespace threads { namespace policies {
...
         {
-            if (user_polling_function_)
-                user_polling_function_();
+            polling_function_cuda_ = &null_polling_function;
+        }
+
+        inline void custom_polling_function() const
+        {
+#if defined(HPX_HAVE_LIB_ASYNC_MPI)
+            (*polling_function_mpi_)();
```
msimberg, line 306: Do you know if there's a difference in performance compared to `if (polling_function_mpi_ != nullptr) (*polling_function_mpi_)();`?

biddisco:
```
if (polling_function_mpi_ != nullptr) (*polling_function_mpi_)();
```
could be subject to a race if `polling_function_mpi` is set to `nullptr` between the test and the invoke. For this reason, I settled on a null function, or a valid function, with no `if ... then` check.

libs/threading_base/include/hpx/threading_base/scheduler_base.hpp
```diff
@@ -319,7 +342,11 @@ namespace hpx { namespace threads { namespace policies {
 
         std::atomic<std::int64_t> background_thread_count_;
 
-        util::function_nonser<void()> user_polling_function_;
+        // the scheduler mode, protected from false sharing
```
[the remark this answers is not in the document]
biddisco: The reason I wrote - please don't review this yet, is because I forgot to remove that and did so in another commit just now.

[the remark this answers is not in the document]
biddisco: oops. wrong line. I left an old comment in there. I will remove it

libs/threading_base/include/hpx/threading_base/scheduler_base.hpp
```diff
@@ -319,7 +342,8 @@ namespace hpx { namespace threads { namespace policies {
 
         std::atomic<std::int64_t> background_thread_count_;
 
-        util::function_nonser<void()> user_polling_function_;
+        polling_function_ptr polling_function_mpi_;
+        polling_function_ptr polling_function_cuda_;
```
hkaiser, line 346: Despite what I said while we talked last, I think these should be atomics to avoid partial reads/writes in concurrent scenarios (on weak-memory platforms). However, you should be able to read and write using `std::memory_order_relaxed` as the pointers themselves are stable.

PR #4852, opened 2020-07-23
Fix mpi_ring test in distributed mode by ensuring all ranks run hpx_main

biddisco:
Fixes : Test problem on rostam hopefully

Flyby. Remove the MPI_Barrier that was annoying @hkaiser


msimberg:
Thanks @biddisco for looking into this. Looks like the test is passing now! I just pushed a cmake-format fix. Merging.

PR #4855, opened 2020-07-24
Adapting transform to C++20

gonidelis:
This is an early adaptation for `transform`. It includes:

- adaptation for the end/begin iterators to be different in non-segmented transform
- adaptation for the end/begin iterators to be different in segmented transform (didn't touch `segmented_transform()` functions though)
- `transform_projected` specialization addition with `util::projection_utility`

## To do

- [ ] Add tests for the adaptations.

There is a compilation error when:

`make tests.unit.modules.segmented_algorithms.partitioned_vector_transform_binary2` and
`make tests.unit.modules.segmented_algorithms.partitioned_vector_transform_binary3`

which I believe stems from the `projection_identity` addition but I can't figure out why.
- [ ] Fix test compilation error for `partitioned_vector_transform_binary2/3`

# Important

This is an early adaptation which aims to facilitate the final adaptation that will utilize CPOs.


gonidelis:
I have adapted the result types for all three overloads of `transform`. There are lots of compilation errors that come from the way tests are implemented. The error stem from here:

https://github.com/gonidelis/hpx/blob/993acc55c4dc17914270e70cd6a50ffded8a722b/libs/algorithms/tests/unit/algorithms/transform_tests.hpp#L66-L67

and here:

https://github.com/gonidelis/hpx/blob/993acc55c4dc17914270e70cd6a50ffded8a722b/libs/algorithms/tests/unit/algorithms/transform_tests.hpp#L232

as there is no `get()` function implemented for `in_out_result` or `in_in_out_result` types. I would like to discuss with @hkaiser  on how to fix that. 

- [ ] I need to add a `projection_identity` specialization for `transform_binary_projected`.

libs/algorithms/include/hpx/parallel/algorithms/transform.hpp
```diff
@@ -42,6 +40,251 @@
...
+    /// \a in_in_out_result<FwdIter1, FwdIter2, FwdIter3>
+    ///           otherwise.
+    ///           The \a transform algorithm returns a tuple holding an iterator
+    ///           referring to the first element after the first input sequence,
+    ///           an iterator referring to the first element after the second
+    ///           input sequence, and the output iterator referring to the
+    ///           element in the destination range, one past the last element
+    ///           copied.
+    ///
+    template <typename ExPolicy, typename FwdIter1, typename FwdIter2, typename FwdIter3, typename F>
```
hkaiser, line 193: None of the overloads for `hpx::transform` should take 3 types of iterators

hkaiser: I take that back, sorry for the noise.

libs/algorithms/include/hpx/parallel/algorithms/transform.hpp
```diff
@@ -893,47 +1136,95 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+                hpx::traits::is_iterator<FwdIter1>::value &&
+                hpx::traits::is_iterator<FwdIter2>::value
+            )>
+        // clang-format on
+        friend FwdIter2 tag_invoke(hpx::transform_t, FwdIter1 first,
+            FwdIter1 last, FwdIter2 dest, F&& f)
         {
-            return get_function_address<
-                typename hpx::util::decay<F>::type>::call(f.f_);
+            return std::transform(first, last, dest);
```
hkaiser, line 1157: Calling the standard algorithms will have the correct semantics, but exposes inconsistent exception behavior. For the other algorithms we have settled on relying our own implementations even in those cases where semantically the standard version would be sufficient.

libs/algorithms/tests/unit/algorithms/transform_binary_tests.hpp
```diff
@@ -67,9 +67,9 @@ void test_transform_binary(ExPolicy policy, IteratorTag)
     auto result = hpx::parallel::transform(policy, iterator(std::begin(c1)),
```
hkaiser, line 67: Please change all uses of `hpx::parallel::transform` to `hpx::transform` or `hpx::ranges::transform` as appropriate.

libs/functional/include/hpx/functional/tag_invoke.hpp
```diff
@@ -193,7 +193,7 @@ namespace hpx { namespace functional {
         template <typename... Args>
         constexpr HPX_FORCEINLINE auto operator()(Args&&... args) const
             noexcept(is_nothrow_tag_invocable_v<Tag, Args...>)
-                -> tag_invoke_result_t<Tag, decltype(args)...>
+                -> tag_invoke_result_t<Tag, Args...>
```
hkaiser, line 196: Thanks!

libs/algorithms/tests/unit/algorithms/transform_binary_tests.hpp
gonidelis: I had let them like this in order to test if they run properely with the changed code. I will change them ASAP though.

libs/algorithms/include/hpx/parallel/util/transform_loop.hpp
```diff
@@ -53,7 +53,7 @@ namespace hpx { namespace parallel { namespace util {
         {
             template <typename InIter1, typename InIter2, typename OutIter,
                 typename F>
-            HPX_HOST_DEVICE HPX_FORCEINLINE static hpx::util::tuple<InIter1,
+            HPX_HOST_DEVICE HPX_FORCEINLINE static util::in_in_out_result<InIter1,
                 InIter2, OutIter>
             call(InIter1 first1, InIter1 last1, InIter2 first2, OutIter dest,
```
aurianer, line 60: You may also want to adapt the begin/end iter template arguments for `transform_loop` too

gonidelis: Got it! Thanks!

libs/algorithms/include/hpx/parallel/algorithms/transform.hpp
```diff
@@ -275,14 +275,19 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+        HPX_CONCEPT_REQUIRES_(
+            execution::is_execution_policy<ExPolicy>::value &&
+            hpx::traits::is_iterator<FwdIter1>::value &&
+            hpx::traits::is_iterator<FwdIter2>::value &&
+            traits::is_projected<Proj, FwdIter1>::value &&
+                traits::is_indirect_callable<ExPolicy, F,
+                    traits::projected<Proj, FwdIter1>>::value
+        )>
+    //clang-format on
+    HPX_DEPRECATED("hpx::parallel::transform is deprecated, use hpx::transform instead")
```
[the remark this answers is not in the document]
hkaiser: Yes, that's what we use everywhere else now.

libs/algorithms/tests/unit/algorithms/transform_binary_tests.hpp
hkaiser: Ahh, makes sense. Take your time with the changes as you see fit.

gonidelis:
This PR is closed and archieved. The `transform` C++20 adaptation will take place in a new PR.

PR #4857, opened 2020-07-24
Improve libCDS integration

hkaiser:
- make git repository and tag configurable
- place generated projects in 'Core' FOLDER

@weilewei please check
@aurianer could we add this to the release, still (doesn't have to go into RC1)?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -932,8 +932,20 @@ if(HPX_WITH_APEX)
 endif()
 
 # LibCDS option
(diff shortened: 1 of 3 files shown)


weilewei:
I modified the test a bit to directly use hazard pointer constructor `cds::gc::HP hpGC(nHazardPtrCount, nMaxThreadCount, nMaxRetiredPtrCount);` and direct usage of attach_thread interface provided by HP `cds::gc::hp::smr::attach_thread();`, this will not change the meaning of the test like before.

aurianer:
@hkaiser sure!

PR #4858, opened 2020-07-26
Extending resilience APIs to support distributed invocations

NK-Nikunj:
This PR will add distributed software level resilience to HPX. It is a work in progress and as of now only the async_replay APIs have been extended to a distributed setting.

Following are the changes proposed:
 - [x] Extend async_replay to distributed async_replay
 - [ ] Extend async_replay executor to distributed
 - [x] Extend async_replicate to distributed async_replicate
 - [ ] Extend async_replicate executor to distributed
 - [ ] Remove redundancy from executors
 - [ ] Remove redundancy in distributed code
 - [x] Extend to async replay and replicate to dataflow

The distributed APIs currently take a vector of ids instead of the number of replays/replications. It then uses these ids to invoke actions on other localities.

@hkaiser I think we can extend this approach to a distributed setting easily. What do you think?

PS: These are software level resiliency APIs and by no means do they tolerate node level faults.


NK-Nikunj:
If the current implementation of distributed replay and replicate looks fine, then I'll extend it to executors and dataflow as well. For now, I'm stopping here for reviews.

libs/resiliency/examples/async_replay.cpp
```diff
@@ -1,153 +1,85 @@
-//  Copyright (c) 2019 National Technology & Engineering Solutions of Sandia,
-//                     LLC (NTESS).
-//  Copyright (c) 2018-2019 Hartmut Kaiser
-//  Copyright (c) 2018-2019 Adrian Serio
-//  Copyright (c) 2019 Nikunj Gupta
+//  Copyright (c) 2019-2020 Nikunj Gupta
```
hkaiser, line 1: What is your rationale of removing the copyrights? I think we should leave those in place.

libs/resiliency/include/hpx/resiliency/async_replay.hpp
```diff
@@ -142,6 +143,71 @@ namespace hpx { namespace resiliency { namespace experimental {
                     });
             }
 
+            template <std::size_t... Is>
```
hkaiser, line 146: Can we keep the implementation of the distributed version independent from the local-only version?

libs/resiliency/include/hpx/resiliency/async_replay.hpp
```diff
@@ -142,6 +143,71 @@ namespace hpx { namespace resiliency { namespace experimental {
...
+            hpx::future<Result> invoke_distributed(
+                hpx::naming::id_type id, hpx::util::index_pack<Is...>)
+            {
+                return hpx::async(f_, id, std::get<Is>(t_)...);
+            }
+
+            hpx::future<Result> call(std::vector<hpx::naming::id_type> ids,
```
hkaiser, line 153: Should we use a distribution policy instead of the `vector<id_type>` here?

libs/resiliency/include/hpx/resiliency/async_replicate.hpp
```diff
@@ -186,6 +289,24 @@ namespace hpx { namespace resiliency { namespace experimental {
...
+    // Asynchronously launch given function \a f exactly \a n times on \a n
+    // different localities, where \a n is the size of the vector of \a ids.
+    // Verify the result of those invocations using the given predicate \a pred.
+    // Run all the valid results against a user provided voting function.
+    // Return the valid output.
+    template <typename Pred, typename Action, typename... Ts>
+    hpx::future<typename hpx::util::detail::invoke_deferred_result<Action,
+        hpx::naming::id_type, Ts...>::type>
+    tag_invoke(async_replicate_validate_t,
+        std::vector<hpx::naming::id_type> ids, Pred&& pred, Action&& action,
```
hkaiser, line 302: If we decided to keep the `vector<id_type>` it should be passed by `const&` to avoid copying.

libs/resiliency/tests/unit/async_replay_plain.cpp
```diff
@@ -1,141 +1,75 @@
...
+
+    {
+        universal_action our_action;
+        hpx::future<int> f =
+            hpx::resiliency::experimental::async_replay_validate(
+                locals, &validate, our_action);
+
+        auto result = f.get();
+
+        HPX_TEST(result == 42);
```
hkaiser, line 72: What's the rationale of removing the tests?

libs/resiliency/examples/async_replay.cpp
NK-Nikunj: Ohh I removed the whole example for a simplified version of it. Looks like I mistakenly removed all the copyrights as well. Will correct it.

libs/resiliency/include/hpx/resiliency/async_replay.hpp
NK-Nikunj: Like in a different namespace? (like hpx::resiliency::experimental::distributed)

libs/resiliency/include/hpx/resiliency/async_replay.hpp
NK-Nikunj: I initially thought of keeping a distribution policy. The idea behind not keeping one is because the user will be interested in invoking them on particular ids. For standard actions, there isn't any difference between using a policy vs giving an id. For components, one can get the ids list when invoking new_ with distribution policy. So he can again invoke async with those ids.

NK-Nikunj: I'm open to suggestions though. So if you have something else in mind, I'd like to hear it out.

libs/resiliency/include/hpx/resiliency/async_replicate.hpp
NK-Nikunj: Right. Forgot to make them const&.

libs/resiliency/tests/unit/async_replay_plain.cpp
NK-Nikunj: I reduced the complexity of the tests to do exactly what it should do and get result from it. I forgot that CI doesn't run on many localities so I'll have to fix the failing tests.

I'm fine bringing back the old test as well. Thought that the tests should express themselves clearly in cases where they start failing.

libs/resiliency/include/hpx/resiliency/async_replay.hpp
hkaiser: Yah, or just in a separate file?

NK-Nikunj: I'll add it in a separate file for now. I can't figure out how to get CPOs to work for different namespaces without an argument belonging to that namespace (to trigger ADL).
Let's do it one step at a time ;-)

libs/resiliency/include/hpx/resiliency/async_replay.hpp
hkaiser: I think we should follow the scheme we already have in place for `async` and friends. If there are additional use cases, those should be considered as a second step.

NK-Nikunj:
The idea behind keeping `vector<id_type>` as a first argument came as a replacement of number of replays/replications. Since we don't have it as an argument in distributed versions, I thought to replace it with the vector instead.

Should I keep an argument for number of replay/replications as first argument and the list of ids like we have in `async`? While this may be redundant as we can get the number from the size of vector, it will enable us to keep APIs that conform to both `async` family and the replay/replicate variations.

hkaiser:
> The idea behind keeping `vector<id_type>` as a first argument came as a replacement of number of replays/replications. Since we don't have it as an argument in distributed versions, I thought to replace it with the vector instead.
> 
> Should I keep an argument for number of replay/replications as first argument and the list of ids like we have in `async`? While this may be redundant as we can get the number from the size of vector, it will enable us to keep APIs that conform to both `async` family and the replay/replicate variations.

Ok, understood. However, I would suggest not to conflate the number of required replays/replicates and the number of localities to use.

NK-Nikunj:
Understood. I'll keep them separate.

NK-Nikunj:
@hkaiser sounds good. I'll create another module for it.

msimberg:
@NK-Nikunj I've just merged #4649 which conflicts heavily with this. Please let me know if you'd like help with rebasing on top of latest master. The modules directories have been rearranged a bit.

NK-Nikunj:
@msimberg  thanks for the heads up. I'll be creating a separate module for distributed resiliency APIs. So it should not be very difficult to rebase (I could certainly be wrong though).

hkaiser:
@NK-Nikunj do you still plan on working on this?

NK-Nikunj:
@hkaiser yes. I'll get it done before this weekend.

hkaiser:
@msimberg except for the inspect failure, this is good to go, I believe.

libs/full/resiliency_distributed/include/hpx/resiliency_distributed/async_replay_distributed.hpp
```diff
@@ -0,0 +1,184 @@
+//  Copyright (c) 2019-2020 Nikunj Gupta
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/resiliency/resiliency_cpos.hpp>
```
msimberg, line 9: The headers, tests, and examples in the `resiliency_distributed` module would still need to be wrapped in
```
#include <hpx/config.hpp>
...
#endif
```
in order for all of this to compile correctly with `hipcc`.

msimberg (changes requested):

libs/full/resiliency_distributed/docs/index.rst
```diff
@@ -0,0 +1,18 @@
...
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+.. _modules_resiliency_distributed:
+
+======================
+resiliency_distributed
+======================
+
+TODO: High-level description of the module.
```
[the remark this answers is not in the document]
NK-Nikunj: @msimberg I've added brief documentation about what's different in the distributed resiliency module and how to use it.

hkaiser:
@aurianer would you be able to assist with the HIP failures reported?

msimberg:
Those are unfortunately due to the rostam environment, and also appear on master. This should be good now from the HIP perspective.

hkaiser:
> Those are unfortunately due to the rostam environment, and also appear on master. This should be good now from the HIP perspective.

Ok, then this can go in now, I believe.

PR #4860, opened 2020-07-26
Making fill and fill_n compatible with C++20

hkaiser:
- converting algorithms to use CPOs


hkaiser:
I'll update the deprecation macros before merging.

PR #4861, opened 2020-07-27
Limiting executor improvements

biddisco:
#1 Rewrite of the limiting executor to use RAII and improve forwarding 
    to the wrapped executor

    #2 Allow limiting executor to use external counting mechanism
    
    The limiting executor does not work with some special executors
    like the MPI one since it returns immediately but leave a task
    that is 'detached' from the runtime (the counters count up and down
    almost immediately, but the task completes later).
    
    To solve this we add a specialization inside the limiting executor
    counting mechanism that allows the executor being wrapped to return
    a counter that can be used to throttle tasks.
    
    Any executor that implements the "in_flight_estimate" member can
    override the counting mechanism to control throttling.
    
    Currently we only handle a single counter so multiple mpi executors
    would reference the same internal mpi counter and it would not be
    possible to throttle different mpi executors spearately.
    
    Simplify the mpi_ring_async_test to use a limiting executor at
    the top level and thereby reduce resource consumption by only
    keeping N mpi requests in flight at once.
    
    Note that the limiting executor is only used for the outer iteration
    of sending, otherwise a deadlock would occur if a receive could


biddisco:
NB. I realied after pushing that I forgot to put anything useful into the unit test. Will fix that (or remove the test) and update the PR later

libs/async_mpi/include/hpx/async_mpi/mpi_future.hpp
```diff
@@ -165,14 +181,14 @@ namespace hpx { namespace mpi { namespace experimental {
...
-    // This is not completely safe as it will return when the request list is
+    // This is not completely safe as it will return when the request vector is
     // empty, but cannot guarantee that other communications are not about
     // to be launched in outstanding continuations etc.
     inline void wait()
     {
         hpx::util::yield_while([]() {
             std::unique_lock<detail::mutex_type> lk(
-                detail::get_list_mtx(), std::try_to_lock);
+                detail::get_vector_mtx(), std::try_to_lock);
```
msimberg, line 191: `get_requests_mtx`?

biddisco: It's not supposed to be exposed to the user and I renamed it really because the request-vector used to be a list, but then I switched to vector. There is a (lock free) request-queue as well and so I prefer `get_vector_mtx` to 'get_requests_mtx` as it reminds me exactly which object we are getting the lock for.

biddisco: mind you, `get_request_vector_mtx` would work...

biddisco:
> Should the executor move to `hpx::execution::experimental::limiting_executor` while you're at it?

I'll do that

libs/async_mpi/include/hpx/async_mpi/mpi_future.hpp
msimberg: Yeah, no problem. I wasn't thinking about users with this change, just ourselves. Feel free to keep the name if you feel like it. I just thought it would be appropriate to have `requests` in the name since that's what it protects.

libs/execution/tests/unit/limiting_executor.cpp
```diff
@@ -0,0 +1,143 @@
+//  Copyright (c) 2007-2016 Hartmut Kaiser
```
msimberg, line 1: This should probably move to the execut*ors* module.

biddisco:
Latest commit should address review comments. New unit test and some fixes make it more usable and robust. Should be good to go if dashboard comes back clean

biddisco:
just fixed the `/dev/shm/hpx/src/libs/executors/tests/unit/limiting_executor.cpp:7:10: fatal error: hpx/async.hpp: No such file or directory` error on the dashboard. Will let the builds cycle again.

libs/executors/include/hpx/executors/limiting_executor.hpp
```diff
@@ -29,166 +28,330 @@
...
-        limiting_executor(const Executor& ex, std::size_t lower,
-            std::size_t upper, bool block_on_destruction = true)
-          : executor_(ex)
-          , count_(0)
-          , lower_threshold_(lower)
-          , upper_threshold_(upper)
-          , block_(block_on_destruction)
-        {
-        }
+    namespace experimental {
```
msimberg, line 31: I had `hpx::execution::experimental` in mind (without `parallel`) since the execution functionality is slowly moving into `hpx::execution`. The namespace will anyway change once the limiting executor becomes non-experimental, so I don't mind if you prefer to keep this namespace. It would be slightly nicer to drop one namespace though.

msimberg (approved):
Thanks! Just one comment which you can do or ignore.

libs/executors/include/hpx/executors/limiting_executor.hpp
biddisco: done

msimberg: Thanks!

msimberg:
@biddisco FYI: https://217356-4455628-gh.circle-artifacts.com/0/hpx_inspect_report.html. Otherwise this can go in.

PR #4862, opened 2020-07-27
Adding version specific deprecation options

hkaiser:


hkaiser:
> Looks good, should we add the dependency `HPX_WITH_DEPRECATION_WARNINGS=ON` => `HPX_WITH_DEPRECATION_WARNINGS_V1_5=ON` ?

I can do that, sure. How do we handle that normally? Do we use CMake option dependencies?

hkaiser:
> > Looks good, should we add the dependency `HPX_WITH_DEPRECATION_WARNINGS=ON` => `HPX_WITH_DEPRECATION_WARNINGS_V1_5=ON` ?
> 
> I can do that, sure. How do we handle that normally? Do we use CMake option dependencies?

I've added that now by just guarding the definition of the new options by `if(HPX_WITH_DEPRECATION_WARNINGS)`. Not sure if this what you had in mind.

msimberg:
I'm not sure I understand the use case for this. @hkaiser would you mind summarizing how you would use this?

hkaiser:
> I'm not sure I understand the use case for this. @hkaiser would you mind summarizing how you would use this?

This allows to mark the deprated facility directly with a version, like:
```
HPX_REPRECATED_V1_5("foo is deprecated") using foo;
```
Which would produce a warning saying:
```
foo is deprecated (This functionality is deprecated starting HPX V1.5 and will be 
removed in the future. Disable this warning by adding this to your cmake invocation: 
-DHPX_WITH_DEPRECATION_WARNINGS_V1_5=Off")
```
This allows for controlling deprecation messages on a version-by-version bases by changing the CMake option's default values. It also adds a clear documentation to the functionality itself detailing with what version it was deprecated.

hkaiser:
@msimberg the use case for this are the algorithm changes we're working on which I would like to become 'active' starting V1.6 (i.e. I'd like to deprecate the old APIs with V1.6). These new macros allow me to mark the old APIs right now without them generating a warning in the upcoming release. I was planning to finish the algorithm work before we release V1.6 which would allow us to claim C++20 standards conformance on that release.

msimberg:
> @msimberg the use case for this are the algorithm changes we're working on which I would like to become 'active' starting V1.6 (i.e. I'd like to deprecate the old APIs with V1.6). These new macros allow me to mark the old APIs right now without them generating a warning in the upcoming release. I was planning to finish the algorithm work before we release V1.6 which would allow us to claim C++20 standards conformance on that release.

Thanks. This was the part I was missing. I'd have just deprecated them right away, but this makes sense as well so we can announce the standards conformance in one go, rather than spread out over multiple releases. In that case would it be enough to just have the deprecation macros, and not the CMake options (just thinking about not adding more options if not needed)? I'd even be in favor of getting rid of the module specific deprecation options and just always using the global one, but that's separate from this. Could the macro be a more generic `#define HPX_DEPRECATED(major, minor, message)` or will the implementation become nasty in that case?

hkaiser:
> > @msimberg the use case for this are the algorithm changes we're working on which I would like to become 'active' starting V1.6 (i.e. I'd like to deprecate the old APIs with V1.6). These new macros allow me to mark the old APIs right now without them generating a warning in the upcoming release. I was planning to finish the algorithm work before we release V1.6 which would allow us to claim C++20 standards conformance on that release.
> 
> Thanks. This was the part I was missing. I'd have just deprecated them right away, but this makes sense as well so we can announce the standards conformance in one go, rather than spread out over multiple releases. In that case would it be enough to just have the deprecation macros, and not the CMake options (just thinking about not adding more options if not needed)? 

I think that giving the user an option to silence the warnings temporarily would be important. There are time where one doesn't have the time to deal with upstream changes right away, but still wanting to have a clean build.

> I'd even be in favor of getting rid of the module specific deprecation options and just always using the global one, but that's separate from this. Could the macro be a more generic `#define HPX_DEPRECATED(major, minor, message)` or will the implementation become nasty in that case?

I'll try this. I like the idea.

hkaiser:
@msimberg better now?

PR #4867, opened 2020-07-28
Add CPOs for for_each

msimberg:
Adds CPOs for `hpx::for_each` and `hpx::ranges::for_each`. As discussed earlier, I've tried to follow the return types in the standard as closely as possible (leaving out the function when an execution policy is used). The return types are adjusted only in the `tag_invoke(for_each_t, ...)` overloads (unlike what I thought initially would need to be adjusted, namely `struct for_each(_n)` or the `foreach` partitioner).

The return types are as follows:

| policy | `hpx::for_each` | `hpx::ranges::for_each` | `hpx::for_each_n` | `hpx::ranges::for_each_n` |
|---|---|---|---|---|
| no policy | `F` | `in_fun_result` | `iter` | `in_fun_result` |
| non-task policy | `void` | `iter`* | `iter` | `iter`* |
| task policy | `future<void>`* | `future<iter>`* | `future<iter>`* | `future<iter>`* |

where `iter` is the iterator type of the input or corresponding to the input range. * is not in the standard. We could keep `iter` as the return type for the non-task policy `hpx::for_each`, even though this is different than the standard.

The iterators were restricted by `is_iterator` with SFINAE, but then in the body of `for_each` there was a `static_assert(is_input/forward_iterator)`. I've change this to only use `is_input/forward_iterator` in the template parameter list. Was the static assert there only for better error messages or also for some other reason?

### DIFF
--- a/examples/1d_stencil/1d_stencil_4.cpp
+++ b/examples/1d_stencil/1d_stencil_4.cpp
@@ -15,8 +15,8 @@
 // computation. This example is still fully local but demonstrates nice
 // scalability on SMP machines.
 
-#include <hpx/hpx_init.hpp>
 #include <hpx/hpx.hpp>
+#include <hpx/hpx_init.hpp>
 
 #include <hpx/algorithm.hpp>
 #include <boost/range/irange.hpp>
@@ -32,16 +32,16 @@
 
 ///////////////////////////////////////////////////////////////////////////////
 // Command-line variables
-bool header = true; // print csv heading
-double k = 0.5;     // heat transfer coefficient
-double dt = 1.;     // time step
-double dx = 1.;     // grid spacing
+bool header = true;    // print csv heading
+double k = 0.5;        // heat transfer coefficient
+double dt = 1.;        // time step
+double dx = 1.;        // grid spacing
 
 inline std::size_t idx(std::size_t i, int dir, std::size_t size)
 {
-    if(i == 0 && dir == -1)
-        return size-1;
-    if(i == size-1 && dir == +1)
+    if (i == 0 && dir == -1)
+        return size - 1;
+    if (i == size - 1 && dir == +1)
         return 0;
 
     HPX_ASSERT((i + dir) < size);
@@ -55,27 +55,39 @@ struct partition_data
 {
 public:
     explicit partition_data(std::size_t size)
-      : data_(new double[size]), size_(size)
-    {}
+      : data_(new double[size])
+      , size_(size)
+    {
+    }
 
     partition_data(std::size_t size, double initial_value)
-      : data_(new double[size]),
-        size_(size)
+      : data_(new double[size])
+      , size_(size)
     {
         double base_value = double(initial_value * size);
         for (std::size_t i = 0; i != size; ++i)
             data_[i] = base_value + double(i);
     }
 
-    partition_data(partition_data && other) noexcept
+    partition_data(partition_data&& other) noexcept
       : data_(std::move(other.data_))
       , size_(other.size_)
-    {}
+    {
+    }
 
-    double& operator[](std::size_t idx) { return data_[idx]; }
-    double operator[](std::size_t idx) const { return data_[idx]; }
+    double& operator[](std::size_t idx)
+    {
+        return data_[idx];
+    }
+    double operator[](std::size_t idx) const
+    {
+        return data_[idx];
+    }
 
-    std::size_t size() const { return size_; }
+    std::size_t size() const
+    {
+        return size_;
+    }
 
 private:
     std::unique_ptr<double[]> data_;
@@ -105,7 +117,7 @@ struct stepper
     // Our operator
     static double heat(double left, double middle, double right)
     {
-        return middle + (k*dt/(dx*dx)) * (left - 2*middle + right);
+        return middle + (k * dt / (dx * dx)) * (left - 2 * middle + right);
     }
 
     // The partitioned operator, it invokes the heat operator above on all
@@ -116,41 +128,38 @@ struct stepper
         std::size_t size = middle.size();
         partition_data next(size);
 
-        next[0] = heat(left[size-1], middle[0], middle[1]);
+        next[0] = heat(left[size - 1], middle[0], middle[1]);
 
-        for(std::size_t i = 1; i != size-1; ++i)
+        for (std::size_t i = 1; i != size - 1; ++i)
         {
-            next[i] = heat(middle[i-1], middle[i], middle[i+1]);
+            next[i] = heat(middle[i - 1], middle[i], middle[i + 1]);
         }
 
-        next[size-1] = heat(middle[size-2], middle[size-1], right[0]);
+        next[size - 1] = heat(middle[size - 2], middle[size - 1], right[0]);
 
         return next;
     }
 
     // do all the work on 'np' partitions, 'nx' data points each, for 'nt'
     // time steps, limit depth of dependency tree to 'nd'
-    hpx::future<space> do_work(std::size_t np, std::size_t nx, std::size_t nt,
-        std::uint64_t nd)
+    hpx::future<space> do_work(
+        std::size_t np, std::size_t nx, std::size_t nt, std::uint64_t nd)
     {
         using hpx::dataflow;
         using hpx::util::unwrapping;
 
         // U[t][i] is the state of position i at time t.
         std::vector<space> U(2);
-        for (space& s: U)
+        for (space& s : U)
             s.resize(np);
 
         // Initial conditions: f(0, i) = i
         std::size_t b = 0;
         auto range = boost::irange(b, np);
         using hpx::parallel::execution::par;
-        hpx::parallel::for_each(par, std::begin(range), std::end(range),
-            [&U, nx](std::size_t i)
-            {
-                U[0][i] = hpx::make_ready_future(partition_data(nx, double(i)));
-            }
-        );
+        hpx::ranges::for_each(par, range, [&U, nx](std::size_t i) {
+            U[0][i] = hpx::make_ready_future(partition_data(nx, double(i)));
+        });
 
         // limit depth of dependency tree
         hpx::lcos::local::sliding_semaphore sem(nd);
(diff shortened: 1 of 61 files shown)


hkaiser:
> The iterators were restricted by `is_iterator` with SFINAE, but then in the body of `for_each` there was a `static_assert(is_input/forward_iterator)`. I've change this to only use `is_input/forward_iterator` in the template parameter list. Was the static assert there only for better error messages or also for some other reason?

There were used for better error messages.

examples/1d_stencil/1d_stencil_4.cpp
```diff
@@ -116,41 +128,39 @@ struct stepper
...
 
         // Initial conditions: f(0, i) = i
         std::size_t b = 0;
         auto range = boost::irange(b, np);
         using hpx::parallel::execution::par;
-        hpx::parallel::for_each(par, std::begin(range), std::end(range),
-            [&U, nx](std::size_t i)
-            {
+        hpx::for_each(
+            par, std::begin(range), std::end(range), [&U, nx](std::size_t i) {
```
hkaiser, line 161: We should be able to use `hpx::ranges::for_each` now.

libs/algorithms/include/hpx/parallel/algorithms/copy.hpp
```diff
@@ -266,10 +266,10 @@ namespace hpx { namespace parallel { inline namespace v1 {
 
             template <typename ExPolicy, typename InIter, typename Sent,
                 typename OutIter>
-            static util::in_out_result<InIter, OutIter> sequential(
+            static hpx::ranges::in_out_result<InIter, OutIter> sequential(
```
hkaiser, line 269: I now remember why I left the `in_out_result` type in namespace util: I wanted to avoid mentioning the `namespace hpx::ranges` in the base algorithms. This is not a big deal, but seemed to be the right thing at that time.

Could we leave the type in `util` and just introduce a `using` directive in the ranges namespace to make in conform to the standard?

libs/algorithms/include/hpx/parallel/algorithms/for_each.hpp
```diff
@@ -8,23 +8,272 @@
...
+    ///                     implicitly converted to Type.
+    /// \param proj         Specifies the function (or function object) which
+    ///                     will be invoked for each of the elements as a
+    ///                     projection operation before the actual predicate
+    ///                     \a f is invoked.
+    ///
+    /// \returns            \a f.
+    template <typename InIter, typename F,
+        typename Proj = util::projection_identity>
+    typename F for_each(InIter first, InIter last, F&& f, Proj&& proj = Proj())
```
hkaiser, line 56: It's just doxygen, but the leading `typename` is superfluous.

hkaiser: Also, the base algorithm shouldn't have a projection.

libs/algorithms/include/hpx/parallel/algorithms/for_each.hpp
```diff
@@ -8,23 +8,272 @@
...
+        /// \returns  The \a for_each algorithm returns a
+        ///           \a hpx::future<void> if the execution policy is of
+        ///           type
+        ///           \a sequenced_task_policy or
+        ///           \a parallel_task_policy and returns void otherwise.
+        template <typename ExPolicy, typename FwdIterB, typename FwdIterE,
+            typename F, typename Proj = util::projection_identity>
+        typename util::detail::algorithm_result<ExPolicy, void>::type
+        for_each(ExPolicy&& policy, FwdIterB first, FwdIterE last, F&& f,
+            Proj&& proj = Proj())
```
hkaiser, line 131: The base algorithm does not support sentinels and does not expose a projection

libs/algorithms/include/hpx/parallel/algorithms/for_each.hpp
```diff
@@ -434,117 +608,175 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+            void>::type
+        tag_invoke(hpx::for_each_t, ExPolicy&& policy, FwdIter first,
+            FwdIter last, F&& f, Proj&& proj = Proj())
+        {
+            using is_segmented = hpx::traits::is_segmented_iterator<FwdIter>;
+
+            return parallel::v1::detail::get_for_each_void_result<>(
+                parallel::v1::detail::for_each_(std::forward<ExPolicy>(policy),
+                    first, last, std::forward<F>(f), std::forward<Proj>(proj),
+                    is_segmented()));
```
hkaiser, line 710:
```suggestion
            return parallel::util::detail::algorithm_result<ExPolicy>::get(
                parallel::v1::detail::for_each_(std::forward<ExPolicy>(policy),
                    first, last, std::forward<F>(f), std::forward<Proj>(proj),
                    is_segmented()));
```

libs/algorithms/include/hpx/parallel/algorithms/for_each.hpp
```diff
@@ -434,117 +608,175 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+#if (!defined(__NVCC__) && !defined(__CUDACC__)) || defined(__CUDA_ARCH__)
+                ,
+            HPX_CONCEPT_REQUIRES_(
+                hpx::parallel::traits::is_indirect_callable<hpx::parallel::execution::sequenced_policy, F,
+                    hpx::parallel::traits::projected<Proj, InIter>>::value)
+#endif
+            >
+        // clang-format on
+        friend F tag_invoke(hpx::for_each_t, InIter first, InIter last, F&& f,
+            Proj&& proj = Proj())
```
hkaiser, line 676: The base algorithm does not support projections.

libs/algorithms/include/hpx/parallel/algorithms/for_each.hpp
```diff
@@ -434,117 +608,175 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
-    ///           type
-    ///           \a sequenced_task_policy or
-    ///           \a parallel_task_policy and returns \a FwdIter
-    ///           otherwise.
-    ///           It returns \a last.
-    ///
+        template <typename R>
+        void get_for_each_void_result(R&& r)
+        {
+        }
```
hkaiser, line 615: This is not needed anymore (see below)

libs/algorithms/include/hpx/parallel/algorithms/for_each.hpp
```diff
@@ -434,117 +608,175 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+                hpx::traits::is_input_iterator<InIter>::value &&
+                hpx::parallel::traits::is_projected<Proj, InIter>::value &&
+                hpx::parallel::traits::is_indirect_callable<hpx::parallel::execution::sequenced_policy, F,
+                    hpx::parallel::traits::projected<Proj, InIter>>::value
+            )>
+        // clang-format on
+        friend typename parallel::util::detail::algorithm_result<
+            hpx::parallel::execution::sequenced_policy, InIter>::type
+        tag_invoke(hpx::for_each_n_t, InIter first, Size count, F&& f,
+            Proj&& proj = Proj())
```
hkaiser, line 731: The non-policy for_each/for_each_n can simply forward to `std::for_each` and `std::for_each_n`.

libs/algorithms/include/hpx/parallel/container_algorithms/for_each.hpp
```diff
@@ -8,19 +8,176 @@
...
+    ///           type
+    ///           \a sequenced_task_policy or
+    ///           \a parallel_task_policy and returns \a FwdIter
+    ///           otherwise.
+    ///           It returns \a last.
+    ///
+    template <typename ExPolicy, typename FwdIter, typename Sent, typename F,
+        typename Proj = util::projection_identity>
+    typename FwdIter for_each(
+        ExPolicy policy, FwdIter first, Sent last, F&& f, Proj&& proj = Proj());
```
hkaiser, line 129:
```suggestion
        ExPolicy&& policy, FwdIter first, Sent last, F&& f, Proj&& proj = Proj());
```

libs/algorithms/include/hpx/parallel/algorithms/copy.hpp
msimberg: Yes, I can definitely do that.

libs/algorithms/include/hpx/parallel/algorithms/for_each.hpp
msimberg: Good catch. That's a leftover from copying and pasting.

About the projection, I saw that it's not there in the standard, but should we explicitly leave it out? Are we not following the standard by having that as an extension (it's an optional parameter after all)?

hkaiser: I think we should follow the standard by the letter, I think. If somebody needs projections, they can use `hpx::ranges::for_each`.

msimberg: I've removed the projection parameter from these overloads. I've left the non-ranges tests that use projections to use the deprecated `hpx::parallel::for_each`.

libs/algorithms/include/hpx/parallel/algorithms/for_each.hpp
msimberg: I've left these to use `hpx::parallel::detail::for_each_` as `std::for_each` is C++17.

libs/algorithms/include/hpx/parallel/container_algorithms/for_each.hpp
```diff
@@ -84,24 +241,345 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+            for_each(ExPolicy&& policy, Rng&& rng, F&& f, Proj&& proj = Proj())
     {
         return for_each(std::forward<ExPolicy>(policy), hpx::util::begin(rng),
             hpx::util::end(rng), std::forward<F>(f), std::forward<Proj>(proj));
     }
 }}}    // namespace hpx::parallel::v1
+
+namespace hpx { namespace ranges {
+    template <typename I, typename F>
+    using for_each_result = in_fun_result<I, F>;
```
hkaiser, line 416: Shouldn't this be `util::in_fun_result`?

msimberg: Mmh, it could be either this or `parallel::util::in_fun_result` because of https://github.com/STEllAR-GROUP/hpx/blob/0bbfb3c5c22954b324dad2f43d1fb259998e9e11/libs/algorithms/include/hpx/parallel/util/result_types.hpp#L132. Do you have a preference?

hkaiser: Ahh, no - you're right. This is fine as it is.

msimberg:
retest

PR #4868, opened 2020-07-28
Fix miscellaneous header problems

msimberg:
- Moves `thread_pool_executor.hpp` to the `executors` module (it had for some reason been left behind in `execution`
- Fixes the guards for deprecation warnings in a few of the compatibility headers (`#if defined(X)` to `#if X`).


libs/executors/include_compatibility/hpx/parallel/executors/thread_pool_executor.hpp
```diff
@@ -10,7 +10,7 @@
 #include <hpx/execution/config/defines.hpp>
```
aurianer, line 10: Should this be `#include <hpx/executors/config/defines.hpp>` instead?

hkaiser: Good catch!

msimberg: Very good catch! There were a few other ones as well...

PR #4872, opened 2020-07-29
Force partitioned vector scan tests to run serially

msimberg:
I could not figure out why, but the `partitioned_vector_{in,ex}clusive_scan` tests run significantly slower (to the point of timing out like here: https://cdash.cscs.ch/viewTest.php?onlyfailed&buildid=120322, even though they do eventually finish) when run in parallel or even in parallel with `--hpx:bind=none`. This makes them run serially and makes other multi-locality tests use `--hpx:bind=none` when networking is off.


msimberg:
retest

PR #4873, opened 2020-07-29
Set CUDA compute capability on rostam Jenkins builds

msimberg:
This should fix the failed runs on rostam. As far as I can tell using `sm_35` will compile for that and newer architectures so this should work on the K40, K80, and V100.

Flyby: I increased the test timeouts on daint and rostam to 5 minutes. I think some of the timeouts on compile-only tests are simply due to occasionally slow filesystems.


msimberg:
There's a bit of a queue on rostam at the moment... so please don't merge this before we see that the CUDA builds actually improved over there.

PR #4876, opened 2020-07-30
Adding example demonstrating how to disable thread stealing during the execution of parallel algorithms

hkaiser:
@scheherzade This should demonstrate what you asked for. The new executor disables thread stealing while the parallel algorithm using it executes.


scheherzade:
@hkaiser That works nicely, thank you so much!

Issue #4877, opened 2020-07-30
hpx_main might not able to start hpx runtime properly


weilewei:
## Expected Behavior

Runs fine.

## Actual Behavior

Got error:

```
terminate called after throwing an instance of 'hpx::detail::exception_with_info<hpx::exception>'
  what():  Attempting to register a thread outside the HPX runtime and no default pool handler is installed. Did you mean to run this on an HPX thread?: HPX(invalid_status)
Aborted

```

## Steps to Reproduce the Problem

main.cpp:
```
#include <hpx/hpx_main.hpp>
#include <hpx/future.hpp>
#include <hpx/modules/threading.hpp>

int main()
{
  hpx::future<void> f = hpx::async([]() { auto id =  hpx::threads::get_self_id();});

  f.get();
}

```

CMakeLists.txt:
```
cmake_minimum_required(VERSION 3.13)

find_package(HPX REQUIRED)

add_executable(main main.cpp)
target_link_libraries(main HPX::hpx)

```

## Specifications

... Please describe your environment

  - HPX Version: latest master
  - Platform (compiler, OS):linux, gcc

NK-Nikunj:
@weilewei it is expected behavior. With the current build system if you want to run your HPX application from main, you need to link it to `wrap_main` as well. So you will need to update it to something like:

```
add_executable(main main.cpp)
target_link_libraries(main HPX::wrap_main HPX::hpx)
```

weilewei:
@NK-Nikunj ahh, ok, thanks. Now it works!

hkaiser:
@NK-Nikunj is there a way to detect this problem (at runtime)?

NK-Nikunj:
@hkaiser it will not be possible to have a check at runtime. The variable is set in `hpx_wrap.hpp`, so an exception handler has no way to know if a static library was linked to or not.
Having said that, I do believe that we can get a check at link time. This can be done by calling a random useless function from `hpx_main.hpp` that is present in `libhpx_wrap`. A link time error won't pop up if the static library is linked to. Otherwise, the linker will let the user know that it couldn't a certain function. Don't know if this solution is elegant, but that's the only course of action I can think of.

hkaiser:
@NK-Nikunj the link-time error diagnosis would be fine as well. How much effort would adding that be?

NK-Nikunj:
[...]

Issue #4880, opened 2020-07-31
Undefined reference to main build error when HPX_WITH_DYNAMIC_HPX_MAIN=OFF


scheherzade:
## Expected Behavior

I had been using HPX 1.3.0 for my [Blazemark](https://bitbucket.org/blaze-lib/blaze/wiki/Blazemark)(benchmark suite for Blaze) experiments but when I updated HPX I wasn't able to build a benchmark in Blazemark anymore.
I had always built HPX with `-DHPX_WITH_DYNAMIC_HPX_MAIN=OFF` for these runs. If I don't specify this flag and link the benchmark with hpx_wrap I am able to build it with the updated HPX, but I get an error when hpx is built with `-DHPX_WITH_DYNAMIC_HPX_MAIN=OFF` .

## Actual Behavior

I get this error message when I try to build `dmatdmatadd` benchmark from Blazemark.
```
/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../lib64/crt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
clang-6.0: error: linker command failed with exit code 1 (use -v to see invocation)

```

## Steps to Reproduce the Problem

... Please be as specific as possible while describing how to reproduce your problem.
  1. ```git clone git@bitbucket.org:blaze-lib/blaze.git```
  2. ```cd blazemark```
  3. change [Configfile](https://bitbucket.org/blaze-lib/blaze/src/master/blazemark/Configfile) as follows:
```
# Compiler selection
# This variable specifies the compiler used for the compilation of all benchmarks.
CXX=clang++

# Special compiler flags
# This variable specifies the compiler flags used for the compilation of all benchmarks.
CXXFLAGS="-O3 -std=c++17 -stdlib=libc++ -DNDEBUG -march=native -fpermissive -DBLAZE_USE_HPX_THREADS"

# Special include directives
# This variable can be used to specify special/additional include-related compiler directives.
INCLUDE_DIRECTIVES="-isystem  /home/sshirzad/lib/hpx/hpx_release_clang/include/"

# Special library directives
# This variable can be used to specify special/additional library-related compiler directives.
[...]

PR #4881, opened 2020-08-02
Adds different runtime exception when registering thread with the HPX runtime

NK-Nikunj:
## Proposed Changes

This PR adds a different runtime exception that will be thrown in scenarios where the user includes `hpx_main.hpp` but fails to link to `libhpx_wrap`. The error thrown is self sufficient and should let the user know that he/she is required to link to `libhpx_wrap`.

This fixes scenarios where the user faces thread registering exception that is orthogonal to the actual cause (see #4877 ).

### Following are two simple case scenarios:

Code 1:
```
#include <hpx/hpx_main.hpp>
#include <hpx/include/future.hpp>
#include <hpx/async_local/async.hpp>

int main(int argc, char * argv[])
{
    hpx::future<int> f = hpx::async([]() { return 42; });
    std::cout << f.get() << std::endl;

    return 0;
}
```

Code 2:
```
#include <hpx/include/future.hpp>
#include <hpx/async_local/async.hpp>

int main(int argc, char * argv[])
{
    hpx::future<int> f = hpx::async([]() { return 42; });
    std::cout << f.get() << std::endl;

    return 0;
}
```

CMakeLists.txt
```
cmake_minimum_required(VERSION 3.13)

find_package(HPX REQUIRED)

add_executable(main main.cpp)
target_link_libraries(main HPX::hpx)
```

All codes are written in correct standard C++ and will compile and link properly. The exception thrown by the first code is as follows:
```
  what():  Attempting to use hpx_main.hpp functionality without linking to libhpx_wrap. Did you link to HPX::wrap_main in your CMakeLists.txt?: HPX(invalid_status)
Aborted (core dumped)
```
This is a much better explanation of what is happening under the hood and should help the user to understand the cause.

The second throws the following exception:
```
  what():  Attempting to register a thread outside the HPX runtime and no default pool handler is installed. Did you mean to run this on an HPX thread?: HPX(invalid_status)
Aborted (core dumped)
```
This is the exception that needs to be thrown as the user forgot to choose between `hpx_init.hpp` or `hpx_main.hpp` and failed to initialize the runtime. This leads to the exception that we see above.

### DIFF
--- a/libs/threading_base/src/register_thread.cpp
+++ b/libs/threading_base/src/register_thread.cpp
@@ -70,7 +70,9 @@ namespace hpx { namespace threads { namespace detail {
                     "linking to libhpx_wrap. If you're using CMakeLists, make "
                     "sure to add HPX::wrap_main to target_link_libraries. "
                     "If you're using Makefile, make sure to link to "
-                    "libhpx_wrap when generating the executable.");
+                    "libhpx_wrap when generating the executable. If you're "
+                    "linking explicitly, consult the HPX docs for library "
+                    "link order and other subtle nuances.");
 
 #endif
 
--- a/libs/threading_base/src/register_thread.cpp
+++ b/libs/threading_base/src/register_thread.cpp
@@ -17,6 +17,11 @@
 #include <string>
 #include <utility>
 
+// The following implementation has been divided for Linux and Mac OSX
+#if defined(HPX_HAVE_DYNAMIC_HPX_MAIN) && \
+    (defined(__linux) || defined(__linux__) || defined(linux) || \
+    defined(__APPLE__))
+
 namespace hpx_start {
     // Redifining weak variables defined in hpx_main.hpp to facilitate error
     // checking and make sure correct errors are thrown. It is added again
@@ -25,6 +30,8 @@ namespace hpx_start {
     HPX_SYMBOL_EXPORT bool is_linked __attribute__((weak)) = false;
     HPX_SYMBOL_EXPORT bool include_libhpx_wrap __attribute__((weak)) = false;
 }   // namespace hpx_start
+
+#endif
 
 namespace hpx { namespace threads { namespace detail {
     static get_default_pool_type get_default_pool;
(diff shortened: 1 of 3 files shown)


NK-Nikunj:
@biddisco I've made sure to expose weak symbols only for apple and linux where they are supported. In that regard, we should be fine.

hkaiser:
Fixes #4877

libs/threading_base/src/register_thread.cpp
```diff
@@ -16,6 +17,22 @@
 #include <string>
 #include <utility>
 
+// The following implementation has been divided for Linux and Mac OSX
+#if defined(HPX_HAVE_DYNAMIC_HPX_MAIN) && \
+    (defined(__linux) || defined(__linux__) || defined(linux) || \
+    defined(__APPLE__))
+
+namespace hpx_start {
+    // Redifining weak variables defined in hpx_main.hpp to facilitate error
```
hkaiser, line 26:
```suggestion
    // Redefining weak variables defined in hpx_main.hpp to facilitate error
```

libs/threading_base/src/register_thread.cpp
```diff
@@ -39,6 +56,22 @@ namespace hpx { namespace threads { namespace detail {
...
+    (defined(__linux) || defined(__linux__) || defined(linux) || \
+    defined(__APPLE__))
+
+            // hpx_main.hpp is included but not linked to libhpx_wrap
+            if (!hpx_start::is_linked && hpx_start::include_libhpx_wrap)
+                HPX_THROW_EXCEPTION(invalid_status,
+                    "hpx::threads::detail::get_self_or_default_pool",
+                    "Attempting to use hpx_main.hpp functionality without "
+                    "linking to libhpx_wrap. Did you link to HPX::wrap_main "
+                    "in your CMakeLists.txt?");
```
hkaiser, line 71: Referring to `CMakeLists.txt` might be misleading as -- while this is the preferred way of linking HPX applications -- users might chose to directly link without using CMake.

NK-Nikunj: I've added a verbose exception message to cater to users using `CMakeLists.txt` and `Makefile`. `pkg-config` takes care of the linking part so we're good there.

hkaiser: What about users that take care of the linking explicitly (see #4880) and get it wrong anyways?

NK-Nikunj: @hkaiser don't think its a good idea to link explicitly (unless no other way). Should I just add a line that check HPX docs for instructions if linking manually?

hkaiser: That might be sufficient, indeed. Thanks!

NK-Nikunj: done!

hkaiser:
@NK-Nikunj once you have fixed the clang-format issues (https://214517-4455628-gh.circle-artifacts.com/0/hpx/artifacts/modified_clang_format_files.txt) we can merge.

PR #4882, opened 2020-08-03
Fixing exception handling for hpx::copy and adding missing tests

hkaiser:


hkaiser:
The test failures are unrelated

PR #4885, opened 2020-08-06
Adapting hpx::find and friends to C++20

hkaiser:
This adds compatibility to C++20 for

- [x] `find` `find_if` `find_if_not`
- [x] `find_end` `find_first_of`

This also adds range based overloads for those algorithms

- [x] `find` `find_if` `find_if_not`
- [x] `find_end` `find_first_of`

This also adds `constexpr` to a couple of facilities (as a flyby).


hkaiser:
This is ready for review now.

PR #4892, opened 2020-08-11
Adding replay_executor and replicate_executor

hkaiser:
- this allows for instance for the parallel algorithms to be resilient
  to failing iterations

### DIFF
--- a/libs/resiliency/include/hpx/resiliency/async_replicate_executor.hpp
+++ b/libs/resiliency/include/hpx/resiliency/async_replicate_executor.hpp
@@ -30,80 +30,150 @@ namespace hpx { namespace resiliency { namespace experimental {
     namespace detail {
 
         ///////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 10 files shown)


hkaiser:
@NK-Nikunj please have a look as it might be useful for the work you're doing as well.

hkaiser:
The error reported is unrelated. This is ready for review now.

libs/resiliency/include/hpx/resiliency/async_replicate_executor.hpp
```diff
@@ -30,80 +30,150 @@ namespace hpx { namespace resiliency { namespace experimental {
...
-                },
-                std::move(results));
-        }
+
+                        if (bool(ex))
+                        {
+                            std::rethrow_exception(ex);
+                        }
+
+                        // throw aborting exception no correct results ere produced
```
msimberg, line 171:
```suggestion
                        // throw aborting exception if no correct results were produced
```

PR #4893, opened 2020-08-11
Remove last use of std::result_of, removed in C++20

K-ballo:


msimberg:
Thanks @K-ballo, much appreciated!

PR #4894, opened 2020-08-12
Fix exception handling for tasks

msimberg:
Setting the exception on a future may yield (a lock is required), and thus lead to the task being resumed on another worker thread after setting the exception. Before this change the exception for a future was set directly in the catch block which could lead to the catch block being entered on one worker thread and exited on another thread. This seems to not be allowed and leads to
segfaults at least with gcc and clang. This change only gets the current exception in the catch block, but sets it after the catch block to avoid potentially yielding in the catch block.

I'm currently stress testing `partition` and `uninitialized_move` and they seem to not fail anymore locally. I'm assuming this will fix most of the other `tests.unit.modules.algorithms` tests that occasionally fail (it always seems to be the exception handling tests that fail, but time will tell if there were actually other ones lurking there as well).

### DIFF
--- a/hpx/lcos/detail/promise_base.hpp
+++ b/hpx/lcos/detail/promise_base.hpp
@@ -56,17 +56,26 @@ namespace lcos {
             void do_run()
             {
                 if (!f_)
-                    return;         // do nothing if no deferred task is given
+                    return;    // do nothing if no deferred task is given
+
+                std::exception_ptr p;
 
                 try
                 {
-                    f_();           // trigger action
-                    this->wait();   // wait for value to come back
+                    f_();            // trigger action
+                    this->wait();    // wait for value to come back
+                    return;
                 }
                 catch (...)
                 {
-                    this->set_exception(std::current_exception());
+                    p = std::current_exception();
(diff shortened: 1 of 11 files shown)


hkaiser:
After looking a bit closer, I would suggest to add a similar fix to other places as well. At least here:

- https://github.com/STEllAR-GROUP/hpx/blob/master/hpx/lcos/detail/promise_base.hpp#L68
- https://github.com/STEllAR-GROUP/hpx/blob/master/libs/async_combinators/include/hpx/async_combinators/split_future.hpp#L114
- https://github.com/STEllAR-GROUP/hpx/blob/master/libs/async_combinators/include/hpx/async_combinators/split_future.hpp#L241
- https://github.com/STEllAR-GROUP/hpx/blob/master/libs/async_cuda/include/hpx/async_cuda/cublas_executor.hpp#L230
- https://github.com/STEllAR-GROUP/hpx/blob/master/libs/async_cuda/include/hpx/async_cuda/cuda_executor.hpp#L178
- …

There are more, I presume

PR #4895, opened 2020-08-13
Update apex tag

aurianer:
Update the APEX tag on master instead of release as there are some failures due to this

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -953,7 +953,7 @@ if(HPX_WITH_APEX)
     "Do not update code from remote APEX repository." OFF CATEGORY "Profiling"
   )
   hpx_option(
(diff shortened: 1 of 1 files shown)


msimberg:
It looks like the standalone pool tests now work normally again. The partitioned vector failures do look new though. @aurianer are you able to reproduced the failure? A stack trace would be important to be able to figure out what's going wrong.

khuck:
Quick update - this failure happens with APEX v2.1.9, too...but only in the debugger.  And only occasionally.

aurianer:
retest cscs

msimberg:
retest cscs

msimberg:
Looks like that fixed it. Thanks @khuck!

msimberg:
The rostam builders are not 100% back to normal yet. Merging based on the daint results.

PR #4896, opened 2020-08-13
Making generate() and generate_n conforming to C++20

hkaiser:


plugins/parcelport/mpi/parcelport_mpi.cpp
```diff
@@ -284,7 +284,7 @@ namespace hpx { namespace traits
 #else
                 "env = ${HPX_HAVE_PARCELPORT_MPI_ENV:"
                         "MV2_COMM_WORLD_RANK,PMIX_RANK,PMI_RANK,OMPI_COMM_WORLD_SIZE,"
-                        "ALPS_APP_PE"
+                        "ALPS_APP_PE,PMIX_RANK"
```
msimberg, line 287: Did this one sneak in by mistake? Looks like `PMIX_RANK` is already in the list anyway...

libs/algorithms/tests/unit/container_algorithms/generate_range.cpp
```diff
@@ -52,7 +77,7 @@ void test_generate(ExPolicy policy, IteratorTag)
 }
 
 template <typename ExPolicy, typename IteratorTag>
-void test_generate_async(ExPolicy p, IteratorTag)
+void test_generate_async(ExPolicy&& p, IteratorTag)
```
gonidelis, line 80: @hkaiser  Could you please explain why the `&&`?

hkaiser: In this case it doesn't really make a big difference, I just thought that using a forwarding reference here would be more idiomatic and avoids copying the execution policy.

libs/algorithms/tests/unit/algorithms/generate.cpp
```diff
@@ -187,7 +245,7 @@ void generate_exception_test()
 
 //////////////////////////////////////////////////////////////////////////////
 template <typename ExPolicy, typename IteratorTag>
-void test_generate_bad_alloc(ExPolicy policy, IteratorTag)
+void test_generate_bad_alloc(ExPolicy&& policy, IteratorTag)
```
gonidelis, line 248: @hkaiser  Why no tests for bad_alloc without exec-policy?

hkaiser: I was too lazy to add those ;-) (I have not added those for any of the ported algorithms, sorry.

PR #4901, opened 2020-08-17
Update public API list with updated algorithms

msimberg:


docs/sphinx/api/public_api.rst
```diff
@@ -64,8 +64,8 @@ Functions
 - :cpp:func:`hpx::parallel::v1::min_element`
 - :cpp:func:`hpx::parallel::v1::minmax_element`
 - :cpp:func:`hpx::parallel::v1::mismatch`
-- :cpp:func:`hpx::parallel::v1::move`
-- :cpp:func:`hpx::parallel::v1::none_of`
+- :cpp:func:`hpx::move`
+- :cpp:func:`hpx::none_of`
 - :cpp:func:`hpx::parallel::v1::partition`
 - :cpp:func:`hpx::parallel::v1::partition_copy`
 - :cpp:func:`hpx::parallel::v1::remove`
```
hkaiser, line 71: I think we can already document all of the algorithms as being available in `namespace hpx` (those that have not been ported to CPOs yet still have a using directive to make this happen).

hkaiser: Ahh, ok - makes sense. Forget what I suggested, it's not worth the effort trying to change this.

PR #4902, opened 2020-08-17
Cancel previous pull requests builds before starting a new one with Jenkins

msimberg:


msimberg:
The cancelling is dependent on how quickly slurm cancels jobs, but it does work (https://lisone.cscs.ch/blue/organizations/jenkins/ssl%2FHPX/detail/HPX/492/pipeline/79 for those of you who have CSCS accounts).

PR #4904, opened 2020-08-17
Update boost to 1.74.0 for the newest jenkins configs

aurianer:
Update the CI to test boost 1.74.0


tests/performance/local/spinlock_overhead2.cpp
```diff
@@ -174,11 +178,20 @@ namespace test
...
+            __sync_lock_release(&v_);
+#  endif
 #else
+#  if !defined( BOOST_SP_HAS_SYNC )
+            HPX_COMPILER_FENCE;
+            *const_cast<std::uint64_t volatile*>(&v_) = 0;
+#  else
             __sync_lock_release(&v_);
+#  endif
 #endif
```
msimberg, line 195: How about this instead:
```suggestion
#if defined(BOOST_SP_HAS_SYNC_INTRINSICS) || defined(BOOST_SP_HAS_SYNC)
            __sync_lock_release(&v_);
#else
            HPX_COMPILER_FENCE;
            *const_cast<std::uint64_t volatile*>(&v_) = 0;
#endif
```
so we don't need to duplicate the two branches? The first branch can additionally be guarded by a version check if you like that.

tests/performance/local/spinlock_overhead2.cpp
```diff
@@ -153,7 +153,11 @@ namespace test
         {
             HPX_ITT_SYNC_PREPARE(this);
 
+#if defined(BOOST_VERSION) && BOOST_VERSION >= 107400
+#if !defined( BOOST_SP_HAS_SYNC_INTRINSICS )
+#else
 #if !defined( BOOST_SP_HAS_SYNC )
+#endif
```
hkaiser, line 160: This looks wrong, was this intended?

msimberg:
@aurianer I only clang-formatted one of the changed files. This should now be ready to go. Thanks for looking into this!

PR #4905, opened 2020-08-17
Adding specializations for simple for_loops

hkaiser:
- flyby: improve loop unrolling for algorithms

This requires some more thorough performance measurements, from my minimal tests this PR improves the picture slightly.


msimberg:
@hkaiser this looks good to go from the CI point-of-view. This looks like it should lead to strictly same or better performance in all cases, but how confident are you about the changes? Are there any tests you'd still like to run before we merge this?

hkaiser:
> @hkaiser this looks good to go from the CI point-of-view. This looks like it should lead to strictly same or better performance in all cases, but how confident are you about the changes? Are there any tests you'd still like to run before we merge this?

I think this can go in as it is.

PR #4909, opened 2020-08-18
Libcds module

biddisco:
GSoC project - merge `hpx` based hazard pointer and `std::thread` based libCDS work


CMakeLists.txt
```diff
@@ -967,17 +967,13 @@ if(HPX_WITH_LIBCDS)
...
     HPX_WITH_LIBCDS_GIT_REPOSITORY STRING
     "Define the LibCDS git repository to use."
-    https://github.com/STEllAR-GROUP/libcds CATEGORY "Thread Manager" ADVANCED
+    https://github.com/khizmax/libcds.git CATEGORY "Thread Manager" ADVANCED
   )
   hpx_option(
-    HPX_WITH_LIBCDS_GIT_TAG STRING "Define the LibCDS git tag to use." hpx-1.5
-    CATEGORY "Thread Manager" ADVANCED
+    HPX_WITH_LIBCDS_GIT_TAG STRING "Define the LibCDS git tag to use."
+    master CATEGORY "Thread Manager" ADVANCED
```
hkaiser, line 974: I would like to see a tag here instead of a reference to an ever moving branch.

cmake/HPX_SetupLibCDS.cmake
```diff
@@ -1,32 +1,23 @@
...
+# Copyright (c) 2020      John Biddiscombe
 #
 # SPDX-License-Identifier: BSL-1.0
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-if(HPX_WITH_LIBCDS AND NOT TARGET LibCDS::cds)
+# set(HPX_WITH_LIBCDS_GIT_REPOSITORY
+# https://github.com/STEllAR-GROUP/libcds.git) set(HPX_WITH_LIBCDS_GIT_TAG
+# hpx-module)
```
hkaiser, line 10: Do we still need these comments?

cmake/HPX_SetupLibCDS.cmake
```diff
@@ -44,9 +35,14 @@ if(HPX_WITH_LIBCDS AND NOT TARGET LibCDS::cds)
     fetchcontent_populate(libcds)
     set(LIBCDS_CXX_STANDARD ${HPX_CXX_STANDARD})
     add_subdirectory(${libcds_SOURCE_DIR} ${libcds_BINARY_DIR})
+    list(TRANSFORM LIBCDS_SOURCELIST PREPEND "${libcds_SOURCE_DIR}/")
+    set(LIBCDS_SOURCE_DIR ${libcds_SOURCE_DIR})
 
-    set_target_properties(cds PROPERTIES FOLDER "Core")
-    set_target_properties(cds-s PROPERTIES FOLDER "Core")
```
hkaiser, line 49: Could we leave the `FOLDER` properties in place, please (we might not need both of them now that we define our own imported target.

libs/libcds/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+  return()
+endif()
+
+include(HPX_SetupLibCDS)
+
+if(NOT libcds_POPULATED)
+  hpx_error("HPX_WITH_LIBCDS was set to ON, but HPX failed to fetch LibCDS")
+endif()
+
+hpx_add_config_define(HPX_HAVE_LIBCDS)
```
hkaiser, line 20: We already generate preprocessor constants for each of the modules. Do we need this additionally?

libs/libcds/include/hpx/libcds/hpx_tls_manager.hpp
```diff
@@ -0,0 +1,119 @@
...
+#include <hpx/config.hpp>
+#include <hpx/errors/exception.hpp>
+//
+#include <cds/gc/details/hp_common.h>
+#include <cds/gc/hp.h>
+#include <cds/threading/details/cxx11_manager.h>
+
+#include <atomic>
+
+//@cond
```
hkaiser, line 18: Not a big thing: we use the `\cond` doxygen keyword syntax everywhere else (instead of `@cond`). I appreciate it if you wouldn't mind changing this.

libs/libcds/include/hpx/libcds/hpx_tls_manager.hpp
```diff
@@ -0,0 +1,119 @@
...
+}}}}    // namespace cds::gc::hp::details
+
+namespace hpx { namespace cds {
+
+    // this wrapper will initialize an HPX thread/task for use with libCDS
+    // algorithms
+    struct hpxthread_manager_wrapper
+    {
+        // allow the std thread wrapper to use the same counter because
+        // the libCDS backen does not distinguihs between them yet.
```
hkaiser, line 37:
```suggestion
        // the libCDS backend does not distinguish between them yet.
```

libs/libcds/include/hpx/libcds/hpx_tls_manager.hpp
```diff
@@ -0,0 +1,119 @@
...
+                        "detaching more threads than number of attached "
+                        "threads");
+                }
+                ::cds::gc::hp::custom_smr<
+                    ::cds::gc::hp::details::HPXTLSManager>::detach_thread();
+            }
+        }
+
+        bool uselibcds_;
+        static std::atomic<std::size_t> max_concurrent_attach_thread_;
```
hkaiser, line 79: Should these be `private` as well?

libs/libcds/tests/performance/libcds_hazard_pointer_overhead.cpp
```diff
@@ -287,6 +264,18 @@ int hpx_main(variables_map& vm)
     libcds_wrapper wrapper;
 
     {
+        // Initialize Hazard Pointer singleton
+        const std::size_t nHazardPtrCount =
+            1;    // Hazard pointer count per thread
+        const std::size_t nMaxThreadCount = hpx::cds::hpxthread_manager_wrapper::
+            max_concurrent_attach_thread_;    // Max count of simultaneous working thread in the application
+        const std::size_t nMaxRetiredPtrCount =
+            16;    // Capacity of the array of retired objects for the thread
```
hkaiser, line 273: Could we stick to HPX naming conventions, please?

libs/libcds/tests/performance/libcds_hazard_pointer_overhead.cpp
```diff
@@ -309,11 +298,12 @@ int hpx_main(variables_map& vm)
...
             throw std::logic_error("error: count of 0 futures specified\n");
 
-        cds::gc::HP hpGC;
-
         hpx::parallel::execution::parallel_executor par;
         hpx::parallel::execution::parallel_executor_aggregated par_agg;
         hpx::parallel::execution::thread_pool_executor tpe;
+        //        hpx::parallel::execution::thread_pool_executor tpe_nostack(
+        //            hpx::threads::thread_priority_default,
+        //            hpx::threads::thread_stacksize_nostack);
```
hkaiser, line 306: Please remove code that is not needed anymore.

libs/libcds/tests/unit/libcds_feldman_map_hazard_pointer_stdthread.cpp
```diff
@@ -0,0 +1,89 @@
...
+    {
+        futures.push_back(std::async([&, ele]() {
+            // enable this thread/task to run using libcds support
+            cds::gc::hp::custom_smr<
+                cds::gc::hp::details::DefaultTLSManager>::attach_thread();
+
+            std::this_thread::sleep_for(std::chrono::seconds(rand() % 5));
+            map.insert(ele, std::to_string(ele));
+            cds::gc::hp::custom_smr<
+                cds::gc::hp::details::DefaultTLSManager>::detach_thread();
```
hkaiser, line 45: Can we use an RAII object instead, please?

libs/libcds/tests/unit/libcds_feldman_map_hazard_pointer_stdthread.cpp
```diff
@@ -0,0 +1,89 @@
...
+    {
+        using map_type =
+            cds::container::FeldmanHashMap<gc_type, key_type, value_type>;
+
+        cds::gc::hp::custom_smr<cds::gc::hp::details::DefaultTLSManager>::
+            construct(map_type::c_nHazardPtrCount + 1, 100, 16);
+
+        // enable this thread/task to run using libcds support
+        cds::gc::hp::custom_smr<
+            cds::gc::hp::details::DefaultTLSManager>::attach_thread();
```
hkaiser, line 77: Don't we need a corresponding `detach_thread` as well? Perhaps using an RAII object would simplify things?

libs/libcds/tests/unit/libcds_flat_combining_priority_queue_stdthread.cpp
```diff
@@ -0,0 +1,111 @@
...
+    {
+        pq.push(i);
+    }
+}
+
+// this test does not start the HPX runtime
+int main(int argc, char* argv[])
+{
+    // Initialize libcds
+    cds::Initialize();
```
hkaiser, line 49: Could we have an RAII object for the global libCDS initialization as well, please?

libs/libcds/tests/unit/libcds_michael_map_hazard_pointer.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    hpx::wait_all(futures);
+
+    std::size_t count = 0;
+
+    while (!map.empty())
+    {
+        auto guarded_ptr = map.extract(rand_vec[count]);
+        assert(guarded_ptr->first == rand_vec[count]);
+        assert(guarded_ptr->second == std::to_string(rand_vec[count]));
```
hkaiser, line 66: Please use `HPX_ASSERT` instead of `assert`.

biddisco:
> Thanks a lot for working on this! ... and sorry for the long list of comments.

No problem. I also want an RAII wrapper for the main init. I'm hoping @weilewei can take care of the issues here during this week and have everything ready before the end of GSoC

weilewei:
Yes, I will try my best to take care of issues this week, and please expect my fixes coming on Thursday or Friday. Thanks. @biddisco

msimberg:
I suggest we leave this out of 1.5.0 by the way, simply because it's very late in the release process and it's not intended to for public use yet. Let's discuss at the meeting if someone would still prefer to have this in 1.5.0.

CMakeLists.txt
weilewei: It needs efforts from LibCDS, is it correct?

libs/libcds/include/hpx/libcds/hpx_tls_manager.hpp
```diff
@@ -0,0 +1,172 @@
...
+#include <hpx/errors/exception.hpp>
+//
+#include <cds/gc/details/hp_common.h>
+#include <cds/gc/hp.h>
+#include <cds/init.h>
+#include <cds/threading/details/cxx11_manager.h>
+
+#include <atomic>
+
+/// \cond NODETAIL
```
weilewei, line 19: @hkaiser do you want to see something like this?

libs/libcds/tests/performance/libcds_hazard_pointer_overhead.cpp
weilewei: What is the style of HPX naming conventions?

weilewei:
I moved hazard pointer init into libcds init now using enum switch, please review it again. Thanks.

libs/libcds/tests/unit/libcds_feldman_map_hazard_pointer.cpp
```diff
@@ -23,39 +25,96 @@
...
+    }
 
-    while (!map.empty())
+    for (auto it = map.cbegin(); it != map.cend(); ++it)
     {
-        auto guarded_ptr = map.extract(rand_vec[count]);
-        HPX_ASSERT(guarded_ptr->first == rand_vec[count]);
-        HPX_ASSERT(guarded_ptr->second == std::to_string(rand_vec[count]));
-        count++;
+        HPX_TEST_EQ(counter_vec[it->first], (it->second)._a);
```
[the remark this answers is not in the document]
biddisco: This is much more like what I expected. Now the maps are tested more fully. 
One thing that confuses me - the sleep for random seconds - this happens once on task create and if I recall correctly, we wanted a sleep in the tasks to allow them to suspend and resume - possibly on a different kernel thread - this makes sure that the test covers the possibility that thread local storage is moving from one core to another correctly. However, putting a single sleep at the start of each task is not very useful.
Suggestion. Move the sleep into the main loop so that instead of this
```
    auto insert_val_and_increase_counter =
        [&](std::vector<std::size_t>& number_vec) {
            hpx::cds::hpxthread_manager_wrapper cds_hpx_wrap;
            hpx::this_thread::sleep_for(std::chrono::seconds(rand() % 5));
            for (auto val : number_vec)
            {
                typename Map::guarded_ptr gp; 

                gp = map.get(val);
                if (gp)
                {
                    (gp->second)._a++;
                    counter_vec[gp->first]++;
                }
            }                                                                                                                                                                                 
        };
```
we have   
```
    auto insert_val_and_increase_counter =
        [&](std::vector<std::size_t>& number_vec) {
            hpx::cds::hpxthread_manager_wrapper cds_hpx_wrap;
            for (auto val : number_vec)
            {
                if (rand() %10 == 0) hpx::this_thread::yield();
                typename Map::guarded_ptr gp;
                gp = map.get(val);
                if (gp)
                {
                    (gp->second)._a++;
                    counter_vec[gp->first]++;
                }
            }                                                                                                                                                                                 
        };
```
why? because now the task will suspend itself many times during the insertion phase and all the threads that are active will be hitting the map in random ordering as they suspend and resume constantly. This will really make sure that if any thread local data is corrupted, then we'll know about it.
I am not familiar with the `guarded_ptr`- should the `if` block also have an `else throw(std::runtime_error(...));` just to make sure we know if something fails? (I think that the map is initialized with all possible `val` values on start.

weilewei:
Now all map tests are in one single file, now only showing hpxthread and stdthread tests for map. I plan to add hpx::init around Wednesday as well as update doc.

cmake/HPX_SetupLibCDS.cmake
```diff
@@ -44,9 +32,17 @@ if(HPX_WITH_LIBCDS AND NOT TARGET LibCDS::cds)
...
+    list(TRANSFORM LIBCDS_SOURCELIST PREPEND "${libcds_SOURCE_DIR}/")
+    set(LIBCDS_SOURCE_DIR ${libcds_SOURCE_DIR})
+    # leave the FOLDER properties in place
+    #set_target_properties(cds PROPERTIES FOLDER "Core")
+    #set_target_properties(cds-s PROPERTIES FOLDER "Core")
 
-    set_target_properties(cds PROPERTIES FOLDER "Core")
-    set_target_properties(cds-s PROPERTIES FOLDER "Core")
+    # create an imported target that links to the real libcds so that when we
+    # link to the inported target, we don't get export X depends on cds that is
```
hkaiser, line 42:
```suggestion
    # link to the imported target, we don't get export X depends on cds that is
```

weilewei:
Please review it again, thanks. I have added command line argument option to update max attached concurrent thread for hp, for example: --hpx:ini=hpx.cds.num_concurrent_hazard_pointer_threads=256. Let me know what do you think.

cmake/HPX_SetupLibCDS.cmake
```diff
@@ -44,9 +32,16 @@ if(HPX_WITH_LIBCDS AND NOT TARGET LibCDS::cds)
...
+    list(TRANSFORM LIBCDS_SOURCELIST PREPEND "${libcds_SOURCE_DIR}/")
+    set(LIBCDS_SOURCE_DIR ${libcds_SOURCE_DIR})
+    # leave the FOLDER properties in place set_target_properties(cds PROPERTIES
+    # FOLDER "Core") set_target_properties(cds-s PROPERTIES FOLDER "Core")
 
-    set_target_properties(cds PROPERTIES FOLDER "Core")
-    set_target_properties(cds-s PROPERTIES FOLDER "Core")
+    # create an imported target that links to the real libcds so that when we
+    # link to the imported target, we don't get export X depends on cds that is
+    # not in the export set
```
msimberg, line 42: Is this warning just about export sets? Does libcds get installed alongside HPX?

Having libcds as a "normal" purely external dependency would remove the burden from us to do it correctly (we might need a `FindLibcds.cmake` module though).

libs/CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ set(HPX_CANDIDATE_LIBS
     itt_notify
     lcos_distributed
     lcos_local
+    libcds
```
msimberg, line 62: Would `cds` be enough for the module name?

libs/libcds/docs/index.rst
```diff
@@ -0,0 +1,145 @@
...
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+.. _libs_libcds:
+
+======
+libcds
+======
+
+LibCDS Overview
+###############
```
msimberg, line 16:
```suggestion
---------------
```
and the same for the other headings.

libs/libcds/docs/index.rst
```diff
@@ -0,0 +1,145 @@
...
+concurrent containers that don't require external (manual) synchronization
+for shared access, and safe memory reclamation (SMR) algorithms like
+Hazard Pointer and user-space RCU that is used as an epoch-based SMR.
+
+Supporting features
+###################
+Currently, this module **only** supports *Hazard Pointer*
+(a type of safe memory reclamation schemas/garbage collectors)
+based containers in LibCDS
+at HPX light-weight user-level thread. In the future, we might consider
```
msimberg, line 28:
```suggestion
at |hpx| light-weight user-level thread. In the future, we might consider
```
and elsewhere.

libs/libcds/docs/index.rst
```diff
@@ -0,0 +1,145 @@
...
+one might want to check
+`LibCDS Documentation <http://libcds.sourceforge.net/doc/cds-api/index.html>`_.
+For example, after clicking `cds > Modules > FeldmanHashMap <http://libcds.sourceforge.net/doc/cds-api/classcds_1_1container_1_1_feldman_hash_map.html>`_
+, one can find Template parameters in FeldmanHashMap class
+:cpp:class:`cds::container::FeldmanHashMap< GC, Key, T, Traits >`
+suggests *GC - safe memory reclamation schema. Can be gc::HP, gc::DHP or one of RCU type`*
+This means :cpp:type:`FeldmanHashMap` can be safely used with Hazard Pointer GC. However,
+again, current HPX does not support *gc:DHP or RCU*, so we cannot use these two types of garbage collectors.
+
+Build Hazard Pointer w/ HPX threads
```
msimberg, line 44:
```suggestion
Build hazard pointer with HPX threads
```

libs/libcds/docs/index.rst
```diff
@@ -0,0 +1,145 @@
...
+            //...
+        }
+
+        return hpx::finalize();
+    }
+
+    int main(int argc, char* argv[])
+    {
+        return hpx::init(argc, argv);
+    }
```
msimberg, line 77: Please move this to the examples directory, and add the file with `literalinclude` instead. Getting the path correct in the `literalinclude` is a bit tricky, so I can help with that part.

libs/libcds/src/hpx_tls_manager.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include <hpx/threading_base/thread_data.hpp>
+
+#include <cds/gc/details/hp_common.h>
+
+#include <atomic>
+#include <cstddef>
+
+namespace cds { namespace gc { namespace hp { namespace details {
+
+    /*static*/ CDS_EXPORT_API thread_data* HPXDataHolder::getTLS()
```
msimberg, line 19:
```suggestion
    CDS_EXPORT_API thread_data* HPXDataHolder::getTLS()
```
?

libs/libcds/include/hpx/libcds/hpx_tls_manager.hpp
```diff
@@ -0,0 +1,249 @@
...
+
+    // this wrapper will initialize a std::thread for use with libCDS
+    // algorithms
+    struct stdthread_manager_wrapper
+    {
+        explicit stdthread_manager_wrapper()
+        {
+            if (++hpxthread_manager_wrapper::thread_counter_ >
+                std::stoul(hpx::get_config_entry(
+                    "hpx.cds.num_concurrent_hazard_pointer_threads", "128")))
```
msimberg, line 208: This is not the end of the world, but I would prefer to see this be independent of the runtime configuration. If this really needs to have a default value set like this we can install a callback to get the configuration value instead. Otherwise what is done here: https://github.com/STEllAR-GROUP/hpx/blob/125de3907b128c4ae3972642a9b9796253961a9b/libs/libcds/tests/unit/libcds_map_hazard_pointer_hpxthread.cpp#L163-L167 is a fine alternative, but it does push the responsibility to the user.

cmake/HPX_SetupLibCDS.cmake
weilewei: This question goes to @biddisco

libs/CMakeLists.txt
weilewei: I will prefer `libcds` as it sounds like it is an external library and not implemented by hpx.

libs/libcds/docs/index.rst
weilewei: Can you help with this? I am not sure what `literalinclude` is. Thanks.

libs/libcds/include/hpx/libcds/hpx_tls_manager.hpp
weilewei: I can hardcode n_threads = 100 (it was true several commits old), would you prefer this way?

libs/libcds/src/hpx_tls_manager.cpp
hkaiser: We usually mark the declarations with the export macros, not the definition.

weilewei: Correct, it is done now.

libs/libcds/docs/index.rst
msimberg: Yeah, sure. I can push a change for this.

msimberg:
@weilewei could you mark the comments that you've already taken care of as resolved, please?

libs/libcds/include/hpx/libcds/hpx_tls_manager.hpp
```diff
@@ -0,0 +1,250 @@
...
+#include <iostream>
+#include <string>
+
+/// \cond NODETAIL
+namespace cds { namespace gc { namespace hp { namespace details {
+
+    class HPXDataHolder
+    {
+    public:
+        CDS_EXPORT_API static CDS_EXPORT_API thread_data* getTLS();
```
msimberg, line 34: Two questions:
1. Should this actually be using `HPX_EXPORT` instead?
2. Can the code be in `hpx::cds` instead of `cds::...`? This is in the end our code, not libcds code.

libs/libcds/tests/performance/libcds_hazard_pointer_overhead.cpp
```diff
@@ -342,7 +300,7 @@ int main(int argc, char* argv[])
 
     // clang-format off
     cmdline.add_options()
-        ("futures", value<std::uint64_t>()->default_value(500000),
+        ("futures", value<std::uint64_t>()->default_value(50000),
```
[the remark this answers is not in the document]
weilewei: @biddisco reduced the number in order to reduce testing time, maybe he can answer it. 

But to my understanding, launching more futures (i.e. 50k or 500k) than concurrently allowed attached threads (128 by default) is not a helpful testing, because the rest of threads (if there is already 128 threads running inside Hazard Pointer containers at one moment, and then the rest is 50k - 128) are not allowed to go into hazard pointer. So testing 128 futures might be good enough.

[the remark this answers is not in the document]
weilewei: It will be great for you to add a callback where appropriate, as I am not sure what does the callback looks like and how to add it.

libs/libcds/include/hpx/libcds/hpx_tls_manager.hpp
weilewei: For 1, in LibCDS code, it requires `CDS_EXPORT_API` so I was following their style.
For 2, @biddisco maybe have more ideas about it.

weilewei:
> @weilewei could you mark the comments that you've already taken care of as resolved, please?

Sure! Just marked

cmake/HPX_SetupLibCDS.cmake
biddisco: I was hoping nobody would ask about the install rules. libcds is not in the export set of hpx, so it won't get installed alongside it. If we are able to use libcds master, then I'm happy to simply have an external libcds that the user must point the build at using `cmake -DLIBCDS_ROOT=...` which removes the need to install ourselves - but if we require a dedicated branch (currently, we need Alex's `hp_generic` branch, then the in-source build is preferred. For now I'm happy to leave things as they are and ignore the install problem since only us developers are using it. If I can't solve it - then I'll switch to an external libcds install when latest stuff is merged to libcds master.

libs/libcds/include/hpx/libcds/hpx_tls_manager.hpp
```diff
@@ -0,0 +1,252 @@
...
+    struct libcds_wrapper
+    {
+        friend struct hpxthread_manager_wrapper;
+        friend struct stdthread_manager_wrapper;
+
+        libcds_wrapper(smr_t smr_type = smr_t::hazard_pointer_hpxthread,
+            std::size_t hazard_pointer_count = 1,
+            std::size_t max_thread_count =
+                detail::get_num_concurrent_hazard_pointer_threads(),
+            std::size_t max_retired_pointer_count = 16)
```
msimberg, line 70: I've added the callback here. My question now is if those other parameters should have configuration options as well or not? Is this the appropriate order for the default parameters (i.e. which ones are the user most likely to change)?

libs/libcds/include/hpx/libcds/hpx_tls_manager.hpp
```diff
@@ -0,0 +1,252 @@
...
+        CDS_EXPORT_API static CDS_EXPORT_API void setTLS(thread_data*);
+        CDS_EXPORT_API static CDS_EXPORT_API generic_smr<HPXDataHolder>*
+        getInstance();
+        CDS_EXPORT_API static CDS_EXPORT_API void setInstance(
+            generic_smr<HPXDataHolder>* new_instance);
+    };
+
+}}}}    // namespace cds::gc::hp::details
+
+namespace hpx { namespace cds {
```
msimberg, line 41:
```suggestion
namespace hpx { namespace cds { namespace experimental {
```
?

cmake/HPX_SetupLibCDS.cmake
msimberg: I'm glad I asked then! Yeah, I'm also ok with keeping this as they are at the moment, but I think long term having it as a normal external dependency is the only right thing to do.

libs/libcds/docs/index.rst
msimberg: I have a commit for this, but it's on another laptop. I'll push it next week.

libs/libcds/include/hpx/libcds/hpx_tls_manager.hpp
hkaiser: Do we need the macro twice? Also, whether we should be using `HPX_EXPORT` or `CDS_EXPORT_API` depends on what shared library the implementation of this function ends up being linked to.

libs/libcds/tests/performance/libcds_hazard_pointer_overhead.cpp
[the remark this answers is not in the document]
hkaiser: In other tests we use different number for Debug and Release configurations. One is for performance measurements, the other one for making sure things actually work without having to run for a long time.

libs/libcds/include/hpx/libcds/hpx_tls_manager.hpp
hkaiser: FWIW, having three `size_t` arguments is highly confusing. Nobody will be able to remember what sequence those have to be supplied in.

weilewei: @msimberg All these parameters should be changed according to the cds container you are using, and the number of threads you want to use. 

@hkaiser if not `size_t` what type should we use then?

hkaiser: @weilewei this is a tough question. `size_t` is the right type to use. The problem is that it's difficult to pass the values in a correct sequence. One possible way to enforce the user has to use names would be to define three structs, like:
```
struct hazard_pointer_count 
{ 
    explicit hazard_pointer_count(std::size_t value) : value_(value) {}

private:
    std::size_t value_;
};
```
and use those as:
```
struct libcds_wrapper
{
    libcds_wrapper(smr_t smr_type = smr_t::hazard_pointer_hpxthread,
        hazard_pointer_count count = hazard_pointer_count(1),
        max_thread_count max_threads = max_thread_count(
            detail::get_num_concurrent_hazard_pointer_threads()),
        max_retired_pointer_count retired_count = max_retired_pointer_count(16))
    {
        // ...
    }
};
```

hkaiser:
@biddisco, @weilewei: I have rebased this on current master and have changed a couple of things. Please verify.

hkaiser:
FWIW, the next step is to go through the review comments and apply necessary changes. I have not looked at the review comments and whether those have been addressed yet.

msimberg:
Closing as this is inactive.

Issue #4822, opened 2020-07-11
Adapt parallel algorithms to C++20


hkaiser:
Adapt our parallel algorithms to C++20 (http://eel.is/c++draft/#algorithms) 

Here is the list of algorithms mandated by the standard:

- [x] `adjacent_difference` (#5564)
- [x] `adjacent_find` (#4996)
- [x] `all_of` `any_of` `none_of` (#4859)
- [x] `copy` `copy_if` `copy_n` (#4821)
- [x] `move` (#4870)
- [x] `count` `count_if` (#4865)
- [x] `equal` (#4883) `mismatch` (#4884)
- [x] `exclusive_scan` (#5436)
- [x] `inclusive_scan` (#5439)
- [x] `reduce` (#4836) 
- [x] `transform` (#5051)
- [x] `fill` `fill_n` (#4860)
- [x] `find` `find_end` `find_first_of` `find_if` `find_if_not` (#4885)
- [x] `for_each` `for_each_n` (#4867)
- [x] `for_loop` `for_loop_n` `for_loop_strided` `for_loop_strided_n` (#4911)
- [x] `generate` `generate_n` (#4896)
- [x] `is_heap` `is_heap_until` (#4965)
- [x] `is_partitioned` (#5006)
- [x] `is_sorted` `is_sorted_until` (#5025)
- [x] `lexicographical_compare` (#5350)
- [x] `max_element` `min_element` `minmax_element` (#5241)
- [x] `make_heap` (#4964)
- [x] `partial_sort` `partial_sort_copy` (#5630)
- [x] `nth_element` (#5592)
- [x] `sort` (#5460)
- [x] `stable_sort` (#5470)
- [x] `partition` (#5540)
- [x] `partition_copy` (#5540)
- [x] `stable_partition`  (#5540)
- [x] `remove` `remove_if` (#5125)
- [x] `remove_copy` `remove_copy_if` (#5150)
- [x] `replace` `replace_copy` `replace_copy_if` `replace_if` (#5192)
- [x] `reverse` `reverse_copy` (#5225)
- [x] `rotate` (#5459)
- [x] `rotate_copy` (#5459)
- [x] `search` `search_n` (#5066)
- [x] `set_difference` `set_intersection` `set_symmetric_difference` `set_union` `includes` (#4970)
- [x] `inplace_merge`  (#4970)
- [x] `merge`  (#4970)
- [x] `swap_ranges` (#5449)
[...]

PR #4911, opened 2020-08-18
Add CPOs for for_loop algorithms

msimberg:
Can go in 1.5.0 or 1.6.0. This should be fairly straightforward (i.e. no big surprises with builds and tests), but let's see.

### DIFF
--- a/docs/sphinx/api/public_api.rst
+++ b/docs/sphinx/api/public_api.rst
@@ -92,10 +92,10 @@ Functions
 - :cpp:func:`hpx::parallel::v1::swap_ranges`
 - :cpp:func:`hpx::parallel::v1::unique`
 - :cpp:func:`hpx::parallel::v1::unique_copy`
-- :cpp:func:`hpx::parallel::v2::for_loop`
-- :cpp:func:`hpx::parallel::v2::for_loop_strided`
-- :cpp:func:`hpx::parallel::v2::for_loop_n`
-- :cpp:func:`hpx::parallel::v2::for_loop_n_strided`
+- :cpp:func:`hpx::for_loop`
+- :cpp:func:`hpx::for_loop_strided`
+- :cpp:func:`hpx::for_loop_n`
+- :cpp:func:`hpx::for_loop_n_strided`
 
 - :cpp:func:`hpx::ranges::all_of`
 - :cpp:func:`hpx::ranges::any_of`
--- a/docs/sphinx/manual/writing_single_node_hpx_applications.rst
+++ b/docs/sphinx/manual/writing_single_node_hpx_applications.rst
@@ -749,23 +749,23 @@ Parallel algorithms
      * ``<hpx/memory.hpp>``
      * :cppreference-memory:`uninitialized_value_construct_n`
 
-.. list-table:: Index-based for-loops (In Header: `<hpx/include/parallel_algorithm.hpp>`)
+.. list-table:: Index-based for-loops (In Header: `<hpx/algorithm.hpp>`)
 
    * * Name
      * Description
      * In header
-   * * :cpp:func:`hpx::parallel::v2::for_loop`
+   * * :cpp:func:`hpx::for_loop`
      * Implements loop functionality over a range specified by integral or iterator bounds.
-     * ``<hpx/include/parallel_for_loop.hpp>``
-   * * :cpp:func:`hpx::parallel::v2::for_loop_strided`
+     * ``<hpx/algorithm.hpp>``
+   * * :cpp:func:`hpx::for_loop_strided`
      * Implements loop functionality over a range specified by integral or iterator bounds.
-     * ``<hpx/include/parallel_for_loop.hpp>``
-   * * :cpp:func:`hpx::parallel::v2::for_loop_n`
+     * ``<hpx/algorithm.hpp>``
+   * * :cpp:func:`hpx::for_loop_n`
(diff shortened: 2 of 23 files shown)


gonidelis:
I like the doc fixes. @hkaiser Shouldn't the doxygen comments regarding the no-exec-policy overloads be added in the other algorithms that have already been adapted, too? Plus, should we link this to #4822?

msimberg:
@gonidelis does this address your comment about other algorithms needing updating: https://github.com/STEllAR-GROUP/hpx/pull/4908?

gonidelis:
@msimberg  Yeah. That's great! Plus, we need to fix these doxygen comments that concern the overloads that do not take an exec-policy argument like you did on this PR.

docs/sphinx/api/public_api.rst
```diff
@@ -92,10 +92,10 @@ Functions
...
 - :cpp:func:`hpx::parallel::v1::unique`
 - :cpp:func:`hpx::parallel::v1::unique_copy`
-- :cpp:func:`hpx::parallel::v2::for_loop`
-- :cpp:func:`hpx::parallel::v2::for_loop_strided`
-- :cpp:func:`hpx::parallel::v2::for_loop_n`
-- :cpp:func:`hpx::parallel::v2::for_loop_n_strided`
+- :cpp:func:`hpx::for_loop`
+- :cpp:func:`hpx::for_loop_strided`
+- :cpp:func:`hpx::for_loop_n`
+- :cpp:func:`hpx::for_loop_n_strided`
```
hkaiser, line 98: Should these go into `hpx::experimental` instead (it's not in the standard yet, after all)

docs/sphinx/manual/writing_single_node_hpx_applications.rst
```diff
@@ -749,23 +749,23 @@ Parallel algorithms
...
      * Implements loop functionality over a range specified by integral or iterator bounds.
-     * ``<hpx/include/parallel_for_loop.hpp>``
-   * * :cpp:func:`hpx::parallel::v2::for_loop_n`
+     * ``<hpx/algorithm.hpp>``
+   * * :cpp:func:`hpx::for_loop_n`
      * Implements loop functionality over a range specified by integral or iterator bounds.
-     * ``<hpx/include/parallel_for_loop.hpp>``
-   * * :cpp:func:`hpx::parallel::v2::for_loop_n_strided`
+     * ``<hpx/algorithm.hpp>``
+   * * :cpp:func:`hpx::for_loop_n_strided`
```
hkaiser, line 766: Thanks for changing it here as well... I think we have missed correcting all the other algorithms we have already ported to use CPOs here...

examples/quickstart/disable_thread_stealing_executor.cpp
```diff
@@ -120,10 +120,10 @@ int main()
 
     // The following for_loop will be executed while thread stealing is disabled
     auto exec = executor_example::make_disable_thread_stealing_executor(
-        hpx::parallel::execution::par.executor());
+        hpx::execution::par.executor());
 
-    hpx::parallel::for_loop(hpx::parallel::execution::par.on(exec), 0, v.size(),
-        [](std::size_t i) {});
+    hpx::for_loop(
+        hpx::execution::par.on(exec), 0, v.size(), [](std::size_t i) {});
```
hkaiser, line 126: Not to self: this change should be done in a separate PR for all occurrences.

docs/sphinx/manual/writing_single_node_hpx_applications.rst
hkaiser: See #4908.

hkaiser:
Do you plan to still add range-based versions here? We can add those in a separate PR as well.

docs/sphinx/api/public_api.rst
msimberg: Unsure. I'd be inclined to leave it here because they're stable (enough) in HPX, even though they are experimental in the standard. Our executors are also not in the standard, but not in the experimental namespace.

docs/sphinx/manual/writing_single_node_hpx_applications.rst
msimberg: I know you already saw the PR, but for anyone else wondering about this: #4908.

examples/quickstart/disable_thread_stealing_executor.cpp
msimberg: Yes, I only changed the ones close to the lines I changed. We'll eventually have to deprecate the names in `hpx::parallel::execution` which will help us change everything to the new names.

msimberg:
> Do you plan to still add range-based versions here? We can add those in a separate PR as well.

I was not planning on doing that here, but can definitely do it in a separate PR. I suppose there's no document for those, but they would just replace `first` and `last` with `range`?

hkaiser:
> I suppose there's no document for those, but they would just replace `first` and `last` with `range`?

Yes, that's what we've done for the other algorithms before C++20. This also might entail adding sentinel support to the machinery. But even that can be done separately.

msimberg:
@hkaiser anything against this still going in 1.5.0?

hkaiser:
> @hkaiser anything against this still going in 1.5.0?

No objections here. Please go ahead.

aurianer (approved):
LGTM, thanks!

PR #4912, opened 2020-08-19
Aggregate initialize CPOs

msimberg:
Potentially fixes some compilation problems seen by @G-071 (i.e. `/home/daissgr/external_packages/spack/opt/spack/linux-ubuntu18.04-broadwell/gcc-7.5.0/hpx-master-52nq7dny5j2onrsczqrnwazze6zduz6y/include/hpx/parallel/algorithms/all_any_none.hpp(628): error: const variable "hpx::none_of" requires an initializer -- class "hpx::none_of_t" has no user-provided default constructor`).

I'm a bit surprised by this, but it could be nvcc being dumb (gcc 8 and CUDA 10 is fine with this)? @G-071 would you mind posting the actual compiler version (gcc is apparently 7.5.0), CUDA version, and C++ standard that you're using?


G-071:
This seems to fix the issues I had!

For reference, I used gcc 7.5.0 with CUDA 10.0.130 and C++14.

PR #4914, opened 2020-08-19
Add alias targets with namespaces used for exporting

### LINKED ISSUE #4913: HPX CMake targets not available when using FetchContent
## Expected Behavior

When using CMake's `FetchContent` module to incorporate HPX in my project, I expect to be able to use the HPX targets.

## Actual Behavior

HPX CMake targets are not available:

```
CMake Error at CMakeLists.txt:96 (add_executable):
  Target "my_hpx_program" links to target "HPX::hpx" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?


CMake Error at CMakeLists.txt:96 (add_executable):
  Target "my_hpx_program" links to target "HPX::wrap_main" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?


CMake Error at CMakeLists.txt:96 (add_executable):
  Target "my_hpx_program" links to target "HPX::iostreams_component" but the
  target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?


-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

  1. Use CMake FetchContent to incorporate HPX in project
  1. Configure project, use HPX targets as described in manual

See `CMakeLists.txt` at https://github.com/kordejong/hello_hpx As a repro case, execute the first command in `debug_build.sh`, for example (comment out everything after line 8).

## Specifications

  - HPX Version: 1.5.0-rc2
  - Platform (compiler, OS): gcc-7, Linux

## Remarks
- I noticed a typo in the manual: https://hpx-docs.stellar-group.org/branches/release/html/quickstart.html#hello-world The include of iostream.hpp should be `hpx/distributed/iostream.hpp`, I think
- Besides the HPX CMake targets not being available, the `iostream.hpp` header is not available in `hpx/distributed/iostream.hpp`, but in `hpx/iostream.hpp`. Maybe some copying is done (only) at installation time?
- I am assuming I can use HPX as advertised by using CMake's FetchContent instead of installing HPX. This can be done in principle, I think. See also https://cmake.org/cmake/help/git-stage/module/FetchContent.html#examples It is mentioned there that the targets are available, after `FetchContent_MakeAvailable`.

msimberg:
Adds alias targets for internal use (i.e. within the HPX CMake project) which have the namespace used when exporting. This should fix #4913, and removes some ugly `$<TARGET_NAME_IF_EXISTS:...>` generator expressions.

@kordejong would you mind trying this out for yourself?

Let's try to get this in 1.5.0 still.

### DIFF
--- a/cmake/HPX_GeneratePackage.cmake
+++ b/cmake/HPX_GeneratePackage.cmake
@@ -41,6 +41,15 @@ export(
   FILE "${CMAKE_CURRENT_BINARY_DIR}/lib/cmake/${HPX_PACKAGE_NAME}/HPXTargets.cmake"
 )
 
+# Add aliases with the namespace for use within HPX
+foreach(export_target ${HPX_EXPORT_TARGETS})
+  add_library(HPX::${export_target} ALIAS ${export_target})
+endforeach()
+
+foreach(export_target ${HPX_EXPORT_INTERNAL_TARGETS})
+  add_library(HPXInternal::${export_target} ALIAS ${export_target})
+endforeach()
+
 # Export HPXTargets in the install directory
 install(
   EXPORT HPXTargets
--- a/cmake/HPX_SetupTarget.cmake
+++ b/cmake/HPX_SetupTarget.cmake
@@ -154,10 +154,7 @@ function(hpx_setup_target target)
 
   if("${_type}" STREQUAL "LIBRARY" AND target_PLUGIN)
     set(plugin_name "HPX_PLUGIN_NAME=hpx_${name}")
-    target_link_libraries(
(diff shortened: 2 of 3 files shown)


kordejong:
> @kordejong would you mind trying this out for yourself?

Yes, building now.

kordejong:
Hello world program using `HPX::hpx`, `HPX::wrap_main` and `HPX::iostreams_component`, when HPX is handled by CMake's `FetchContent` builds fine now. Cool!

PR #4916, opened 2020-08-20
HPX_HAVE_DEPRECATION_WARNINGS needs to be set even when disabled

biddisco:
if HPX_HAVE_DEPRECATION_WARNINGS is not defined, then some `#if HPX_HAVE_DEPRECATION_WARNINGS` checks are invalid, so set it to 0 when disabled instead of leaving it unset

### DIFF
--- a/components/iostreams/include/hpx/include/iostreams.hpp
+++ b/components/iostreams/include/hpx/include/iostreams.hpp
@@ -11,7 +11,7 @@
 #include <hpx/config.hpp>
 #include <hpx/iostream.hpp>
 
-#if (HPX_HAVE_DEPRECATION_WARNINGS != 0)
+#if HPX_HAVE_DEPRECATION_WARNINGS
 #if defined(HPX_MSVC)
 #pragma message("The header hpx/include/iostreams.hpp is deprecated, \
     please include hpx/iostream.hpp instead")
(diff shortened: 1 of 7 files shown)


aurianer (approved):
Thanks a lot!

msimberg:
Could you do the same here please: https://github.com/STEllAR-GROUP/hpx/blob/43fbe0378fb91d8ece011c8b8816a687f10d703f/cmake/HPX_AddModule.cmake#L62-L67? Thanks for fixing this!

biddisco:
> where `FOO` is undefined is valid and should do the correct thing as the preprocessor replaces all undefined symbols in conditional expressions with `0`. But the changes proposed don't do any harm.

The problems are where comparisons are made.

https://github.com/STEllAR-GROUP/hpx/blob/dac11daccfef1c3d98734b668dc09a1d0fb217db/components/iostreams/include/hpx/include/iostreams.hpp#L14

I'll check on the `HPX_${name_upper}_HAVE_DEPRECATION_WARNINGS` ones too

biddisco:
I just realized, you probably mean we should just change those lines that make the comparisons to simple `#ifdefs` 
I'll see about that too. I think there are some places where deprecation version was added that might need it ...

biddisco:
After looking more closely, I'm happy that we only need to change
```
#if (HPX_HAVE_DEPRECATION_WARNINGS != 0) 
```
to 
```
#if HPX_HAVE_DEPRECATION_WARNINGS
```
and the rest that use ```HPX_${name_upper}_HAVE_DEPRECATION_WARNINGS``` etc are all fine. It was only the comparison check that was causing trouble. I've changed this PR accordingly.

PR #4923, opened 2020-08-24
Don't acquire lock if not needed

hkaiser:
Second attempt at creating this PR


msimberg (approved):

biddisco:
This is good. Latch performance, now restores the behaviour of the hierarachical scheduling. Things look much better across the board.
 
image

biddisco:
NB. This should be merged in before the release if no tests break as it is a serious regression fix.

PR #4924, opened 2020-08-24
MPI init fixes

msimberg:
Attempts to fix MPI tests running on Rostam, but I would appreciate some input from @biddisco, @hkaiser on how we could do this better.

1. When using the parcelport option in `hpxrun` I added the priority configuration option to force the use of the chosen parcelport. The way the configuration is set up on master is that all available parcelports are enabled in the configuration, meaning that doing `--hpx:ini=hpx.parcelport.tcp.enable=1` has really no effect. If in an MPI environment (like slurm with MPI support) the TCP parcelport would not be used.
2. I moved the MPI initialization so that it happens later in the startup process. On master the MPI environment is checked and initialized before the command line options are parsed, meaning that options like `--hpx:ini=hpx.parcelport.mpi.enable=0` are not even considered when initializing MPI. Moving it even later leads to problems that I do not quite understand (it seems like all localities might think that they're locality 0, and try to do the full bootstrap procedure, not quite sure...).
3. OpenMPI may be broken on rostam. `MPI_Finalize` segfaults in an `hwloc` call... I'm changing the configurations to use only mpich for now.

I'm opening this as a draft to check if this even works on rostam. It seemed to work from my limited testing, but I'm 100% sure I've broken some other configuration like this.

Ideally I'd like to have the MPI parcelport only do the MPI initialization, instead of it happening early as it does now (with the exception for when networking is off and MPI really is a special case), but I'm not sure what that would require.


msimberg:
I think I'd prefer to leave this out of 1.5.0 since I'm not confident that this is correct, especially as the initialization is mostly unchanged from 1.4 and should keep working as before. We do need something like the priority changes, but I'm still unsure what's caused by the MPI setup on Rostam and what's HPX's fault. If I manage to find something before the release let's of course get it in, but I suspect that this is better left for a patch release afterwards.

hkaiser:
> I think I'd prefer to leave this out of 1.5.0 since I'm not confident that this is correct, especially as the initialization is mostly unchanged from 1.4 and should keep working as before. We do need something like the priority changes, but I'm still unsure what's caused by the MPI setup on Rostam and what's HPX's fault. If I manage to find something before the release let's of course get it in, but I suspect that this is better left for a patch release afterwards.

I think we should get Alireza (@akheir) involved here. He has done some MPI/SLURM integration work on rostam that might have the unneeded side effects we're seeing...

msimberg:
> > I think I'd prefer to leave this out of 1.5.0 since I'm not confident that this is correct, especially as the initialization is mostly unchanged from 1.4 and should keep working as before. We do need something like the priority changes, but I'm still unsure what's caused by the MPI setup on Rostam and what's HPX's fault. If I manage to find something before the release let's of course get it in, but I suspect that this is better left for a patch release afterwards.
> 
> I think we should get Alireza (@akheir) involved here. He has done some MPI/SLURM integration work on rostam that might have the unneeded side effects we're seeing...

Yes, he knows that we're having some difficulties with MPI, but I wanted to undestand things a bit better first before involving him further. I'll keep him in the loop when there's something to deal with.

msimberg:
I've rearranged things a bit again.

- OpenMPI seems to be fine now, I was mixing different hwloc versions.  (MPICH not, but this is due to the version of MPICH coming with CentOS.)
- I've kept the changes to parcelport priorities and hpxrun.
- From local testing the only failing test seems to be `thread_mapper`. Let's see if anything else fails on jenkins.
- I fixed the CUDA flags, so those should now finally be correct.

@hkaiser the `thread_mapper` test reports 4 parcelport threads, when it expects only 2. This is because both the MPI and TCP parcelport service pools get started (I don't think they're both actually used, only the one with higher priority). Is this actually intended or desired behaviour? This seems to be a result of the config entries `hpx.parcel.x.enable` all being 1 if the parcelport is there, and the service pool being started in that case. This is a new test, and jenkins on daint doesn't test the MPI parcelport, so that's why it never came up earlier.

hkaiser:
> @hkaiser the `thread_mapper` test reports 4 parcelport threads, when it expects only 2. This is because both the MPI and TCP parcelport service pools get started (I don't think they're both actually used, only the one with higher priority). Is this actually intended or desired behaviour? This seems to be a result of the config entries `hpx.parcel.x.enable` all being 1 if the parcelport is there, and the service pool being started in that case. This is a new test, and jenkins on daint doesn't test the MPI parcelport, so that's why it never came up earlier.

Ok, this needs fixing. I'll see what I can do.

msimberg:
> > @hkaiser the `thread_mapper` test reports 4 parcelport threads, when it expects only 2. This is because both the MPI and TCP parcelport service pools get started (I don't think they're both actually used, only the one with higher priority). Is this actually intended or desired behaviour? This seems to be a result of the config entries `hpx.parcel.x.enable` all being 1 if the parcelport is there, and the service pool being started in that case. This is a new test, and jenkins on daint doesn't test the MPI parcelport, so that's why it never came up earlier.
> 
> Ok, this needs fixing. I'll see what I can do.

Ok, thank you. Please let me know if I can help with that. Also feel free to push to this branch.

Unrelated, but I forgot to mention one more change: I've forced the `launch_process` test to run only with the TCP parcelport. Dynamically connecting MPI ranks is not possible, or am I mistaken?

msimberg:
retest lsu

msimberg:
retest lsu

msimberg:
@hkaiser I have a simple fix in place for the multiple parcelport service pools being started. I simply start the bootstrap parcelport and the highest priority parcelport (just one if they're the same). If this looks reasonable to you I can move out the logic of getting those parcelports into one place and use it for the other parcelport functionality as well. The minimal needed change is for starting and stopping the service pools, and the rest seems to gracefully do nothing or at least nothing harmful. It'd probably be good to filter out unnecessary parcelports from other operations as well. Or would there be a good reason to keep them initialized but doing nothing otherwise?

hkaiser:
> Unrelated, but I forgot to mention one more change: I've forced the launch_process test to run only with the TCP parcelport. Dynamically connecting MPI ranks is not possible, or am I mistaken?

It should work even if run using the MPI parcelport. The only caveat is that connecting processes will always use TCP, even if the main application runs using a different parcelport.

hkaiser:
@msimberg please  note, I have pushed a fix for the failing thread_mapper test here.

msimberg:
> > Unrelated, but I forgot to mention one more change: I've forced the launch_process test to run only with the TCP parcelport. Dynamically connecting MPI ranks is not possible, or am I mistaken?
> 
> It should work even if run using the MPI parcelport. The only caveat is that connecting processes will always use TCP, even if the main application runs using a different parcelport.

> @msimberg please note, I have pushed a fix for the failing thread_mapper test here.

Ah, please see the preceding commit as well. I'll leave it up to you to decide what we should do here. Your comment about the `launch_process` test makes it sound like we actually do want both the MPI and TCP parcelports to be started even if the MPI one has higher priority.

hkaiser:
> Ah, please see the preceding commit as well. I'll leave it up to you to decide what we should do here. Your comment about the `launch_process` test makes it sound like we actually do want both the MPI and TCP parcelports to be started even if the MPI one has higher priority.

To avoid doing this always, we could enable the TCP parcelport if it's the only one or only if the `--hpx:expect-connecting-localities` option (or similar) is provided (well, or if it's explicitly enabled using `hpx.parcel.tcp.enable=1`).

msimberg:
> > Ah, please see the preceding commit as well. I'll leave it up to you to decide what we should do here. Your comment about the `launch_process` test makes it sound like we actually do want both the MPI and TCP parcelports to be started even if the MPI one has higher priority.
> 
> To avoid doing this always, we could enable the TCP parcelport if it's the only one or only if the `--hpx:expect-connecting-localities` option (or similar) is provided (well, or if it's explicitly enabled using `hpx.parcel.tcp.enable=1`).

Currently all parcelports are enabled unless explicitly disabled, which is why the priorities are important. I'm fine with having them all enabled if that's expected behaviour. Especially as they don't really do anything with one locality (a few extra context switches for the pools maybe, but nothing significant). I'm really indifferent (and unsure) about what the right thing to do here is.

hkaiser:
> Currently all parcelports are enabled unless explicitly disabled, which is why the priorities are important. I'm fine with having them all enabled if that's expected behaviour. Especially as they don't really do anything with one locality (a few extra context switches for the pools maybe, but nothing significant). I'm really indifferent (and unsure) about what the right thing to do here is.

As long as you don't see any reason to change the current behavior I'd say, let's leave it unchanged. The user can always directly disable the TCP parcelport (for instance) if he/she wants to avoid instantiating it.

msimberg:
> > Currently all parcelports are enabled unless explicitly disabled, which is why the priorities are important. I'm fine with having them all enabled if that's expected behaviour. Especially as they don't really do anything with one locality (a few extra context switches for the pools maybe, but nothing significant). I'm really indifferent (and unsure) about what the right thing to do here is.
> 
> As long as you don't see any reason to change the current behavior I'd say, let's leave it unchanged. The user can always directly disable the TCP parcelport (for instance) if he/she wants to avoid instantiating it.

Yep, sounds sane. I'll have another look at the `launch_process` test to see if that still requires the changes it has now or not.

msimberg:
retest lsu

hkaiser (dismissed):
Let's merge this asap and see what breaks. Thanks @msimberg for looking into this!

PR #4927, opened 2020-08-25
Refactor synchronization primitives

hkaiser:
- reduce compile time by moving code to implementation files
- fix small performance regression allowing to avoid acquiring lock under
  limited circumstances
- fix #include issues


hkaiser:
@msimberg I'd need a second pair of eyes looking at my changes to `latch` to make sure I'm not introducing any new problems.

@biddisco would you be able to run your benchmarks for this PR, please?

biddisco:
I ran benchmarks, but they timed out before a full run was complete. It may have been a system issue on daint, or it might indicate some regression in the timings. The plot generated from the data that was collected looks ok. I have started a new benchmark run to see if it completes this time
image

libs/synchronization/include/hpx/synchronization/latch.hpp
```diff
@@ -128,28 +142,36 @@ namespace hpx { namespace lcos { namespace local {
...
-                notified_ = true;
-                cond_.data_.notify_all(std::move(l));    // release the threads
+                bool expected = false;
+                if (notified_.compare_exchange_strong(
+                        expected, true, std::memory_order_acq_rel))
+                {
+                    // release the threads
+                    std::unique_lock<mutex_type> l(mtx_.data_);
+                    cond_.data_.notify_all(std::move(l));
+                }
```
msimberg, line 157: This looks correct to me. My only question for the above block is: would `notified_` ever be `true` before a thread has decremented the count to 0? The thread that decrements the count to 0 is always the one that notifies as well, right? That is, should we assert that the `compare_exchange_strong` always returns `true`?

hkaiser: @msimberg Hmmm. I think you're right. There is no way for the `notified_` flag to be `true` without the `counter_` reaching zero first, which can happen on the same thread of execution only. I changed the implementation accordingly, please have another look.

hkaiser: Now the new assertion fires: https://cdash.cscs.ch/testDetails.php?test=26951590&build=124891. I guess that means that we should keep the `compare_exchange`. :/

msimberg: Or that the latch being used somehow inappropriately there? I'd like to at least check what the full stack trace for that failure is before we merge this. I'll try to do this today.

msimberg: The other segmentation faults also don't look very promising :/

biddisco:
I ran benchmarks last  night, but again, they were too slow to give me full coverage. A lot of core dumps in the benchmark folder this morning. Might well indicate a problem.

libs/synchronization/include/hpx/synchronization/latch.hpp
hkaiser: Thanks! The changes here are not very important, we can drop them on the floor without losing anything.

biddisco:
did a merge with the latest changes and ran again, performance looks ok - here's a different plot for a change
image

hkaiser:
Thanks John! From what I see, I think it's not worth the risk of merging this now as the effect is marginal. @msimberg should we just close this PR, what do you think?

msimberg:
> Thanks John! From what I see, I think it's not worth the risk of merging this now as the effect is marginal. @msimberg should we just close this PR, what do you think?

Yeah, let's not rush this into the release at least. Could you remove the functional changes from this PR? I don't think we need to dismiss the whole PR because of the functional changes failing at the moment. We can leave the potential optimizations for another day.

hkaiser:
@msimberg I reverted the changes to latch and rebased the refactorings onto master.

msimberg:
Thank you @hkaiser! Thinks look good now. I rebased once more because I had a mistake in the rostam jenkins configuration. Those green ticks were duds. Most things will be fine though.

PR #4928, opened 2020-08-26
Add HPX_NODISCARD to enable_user_polling structs

msimberg:
Avoid mistakes where user polling is enabled and disabled immediately, instead of at the end of a scope.

@auriane this is a nice extra for the release, but it does not have to be in 1.5.0-rc3.


biddisco (approved):
Good catch. I will do the same to the libcds on the next round of commits

aurianer (approved):

hkaiser:
This is a nice trick we could add to more RAII types we have. Thanks!

PR #4929, opened 2020-08-26
Add final version of the JOSS paper

diehlpk:
Add the final version of the JOSS paper


msimberg:
Thanks @diehlpk! I'm just curious, is it a requirement from JOSS that the paper be in the same repo as the software?

diehlpk:
> Thanks @diehlpk! I'm just curious, is it a requirement from JOSS that the paper be in the same repo as the software?

Yes, because the paper is part of the documentation of the software. It can be in any folder, if you like to move it around.

PR #4945, opened 2020-09-04
Enable HPX_WITH_UNITY_BUILD option on one of the Jenkins configurations

msimberg:
This just enables a build on Jenkins with the option turned on.


aurianer (dismissed):
Thanks!

hkaiser (dismissed):
LGTM, thanks!

Could we also enable unity builds for the other HPX core modules, please?

msimberg:
> LGTM, thanks!
> 
> Could we also enable unity builds for the other HPX core modules, please?

Yep, let me try that out on this branch.

cmake/HPX_SetupTarget.cmake
```diff
@@ -183,6 +183,10 @@ function(hpx_setup_target target)
     target_link_libraries(${target} ${__tll_private} hpx_private_flags)
   endif()
 
+  if(IN_HPX_BUILD AND HPX_WITH_UNITY_BUILD)
+    set_target_properties(${target} PROPERTIES UNITY_BUILD ON)
+  endif()
```
hkaiser, line 188: Wouldn't this enable unity builds for tests and examples as well? Does that make sense?

msimberg: Yes, but I figured it doesn't hurt. And you're right that it doesn't make much sense, since most of them only have use a single source file.

hkaiser: Is there a better solution (might not be as simple, but less - well - obnoxious)?

aurianer: We could pass an option to `add_hpx_executable` (used in tests) and `add_hpx_example_test` to disable unity build and forward it to `hpx_setup_target` which would enable unity builds by default if the disabling option is not specified

msimberg: Yes, the alternative is not much more complicated. I'd probably keep the default as non-unity still, and explicitly pass a flag to enable it for components.

hkaiser (dismissed):
LGTM, thanks!

cmake/HPX_SetupTarget.cmake
msimberg: I've rearranged things a bit. The modules and parcelports handle the option internally since they're not for external use. Parcelports, components, plugins have the flag passed explicitly.

PR #4946, opened 2020-09-04
Allow suspension on multiple localities

### LINKED ISSUE #4937: Allow suspension of runtime for multiple localities
## Expected Behavior

It should be possible to suspend and resume the HPX runtime when using any number of localities.

## Actual Behavior

When trying to suspend the runtime with more than one active locality, an exception is thrown indicating that suspending the runtime is [only allowed for a single locality](https://github.com/STEllAR-GROUP/hpx/blob/9a23fe98cd27c96013835f70f61743d50f92169b/src/runtime_distributed.cpp#L810).

  1. Start the runtime on more than one localities.
  2. Perform a call to `hpx::suspend()`.

severinstrobl:
The restriction of the suspend/resume mechanism of the runtime to a single locality seems unnecessary.

Fixes #4937

According to a short discussion with @msimberg it should be possible to suspend/resume the runtime also on multiple localities, so as a test I removed the corresponding checks and did some test runs on our cluster which showed no problems. Nevertheless, I would appreciate the input of someone with a deeper understanding of the code.

### DIFF
--- a/libs/full/runtime_local/src/runtime_local.cpp
+++ b/libs/full/runtime_local/src/runtime_local.cpp
@@ -1592,14 +1592,6 @@ namespace hpx {
 
     int runtime::resume()
     {
-        std::uint32_t initial_num_localities = get_initial_num_localities();
-        if (initial_num_localities > 1)
-        {
-            HPX_THROW_EXCEPTION(invalid_status, "runtime::resume",
-                "Can only suspend runtime when number of localities is 1");
-            return -1;
-        }
-
         LRT_(info) << "runtime_local: about to resume runtime";
 
         if (state_.load() == state_running)
--- a/src/runtime_distributed.cpp
+++ b/src/runtime_distributed.cpp
@@ -805,31 +805,11 @@ namespace hpx {
 
     int runtime_distributed::suspend()
     {
-#if defined(HPX_HAVE_NETWORKING)
-        std::uint32_t initial_num_localities = get_initial_num_localities();
-        if (initial_num_localities > 1)
-        {
-            HPX_THROW_EXCEPTION(invalid_status, "runtime_distributed::suspend",
-                "Can only suspend runtime when number of localities is 1");
-            return -1;
-        }
-#endif
-
         return runtime::suspend();
     }
 
     int runtime_distributed::resume()
     {
-#if defined(HPX_HAVE_NETWORKING)
-        std::uint32_t initial_num_localities = get_initial_num_localities();
-        if (initial_num_localities > 1)
-        {
-            HPX_THROW_EXCEPTION(invalid_status, "runtime_distributed::resume",
-                "Can only suspend runtime when number of localities is 1");
-            return -1;
-        }
-#endif
-
         return runtime::resume();
     }
 


msimberg:
Well, nice! Thanks @severinstrobl for testing this. The main concern I had for not allowing this without testing was that the suspension mechanism makes no effort to make sure other localities have finished their work. It only waits until the local thread pool is empty before suspending. That said, the current behaviour of throwing an exception is just an artificial limitation mainly to avoid nasty surprises, but as long as you're ok with the local-only synchronization I think going with your changes here is no problem at all. Most likely the way you use HPX ensures anyway that whatever work you need done is actually done before suspending. Does that sound fine for you?

 I might add something to the documentation about this behaviour, but otherwise it shouldn't require any other changes.

severinstrobl:
Yes, the global synchronization is done explicitly via other means, so at least for my use case this works nicely.

hkaiser:
Is there a way t avoid for the user to run into issues if there is _no_ explicit synchronization between localities done? Should we introduce a barrier in HPX before suspending (we could reuse our global predefined barrier for this: `hpx::lcos::barrier::synchronize()`)?

msimberg:
> Is there a way t avoid for the user to run into issues if there is _no_ explicit synchronization between localities done? Should we introduce a barrier in HPX before suspending (we could reuse our global predefined barrier for this: `hpx::lcos::barrier::synchronize()`)?

Is that enough though? Synchronizing with barrier before suspending doesn't mean that tasks can't be scheduled after the barrier. Synchronizing after waiting for queues to be empty also doesn't mean the queues still can't get tasks from remote localities. I guess this is the reason for the elaborate termination mechanism.

hkaiser:
> Is that enough though? Synchronizing with barrier before suspending doesn't mean that tasks can't be scheduled after the barrier. Synchronizing after waiting for queues to be empty also doesn't mean the queues still can't get tasks from remote localities. I guess this is the reason for the elaborate termination mechanism.

Yes, I agree. This is not as easy as it looks. We could expose the termination detection to the user and document it as one way of ensuring that distributed suspension works as expected. That would require a bit of additional work, though.

msimberg:
> > Is that enough though? Synchronizing with barrier before suspending doesn't mean that tasks can't be scheduled after the barrier. Synchronizing after waiting for queues to be empty also doesn't mean the queues still can't get tasks from remote localities. I guess this is the reason for the elaborate termination mechanism.
> 
> Yes, I agree. This is not as easy as it looks. We could expose the termination detection to the user and document it as one way of ensuring that distributed suspension works as expected. That would require a bit of additional work, though.

Yeah, that sounds doable. Let's not have that stop @severinstrobl from using this as is though?

msimberg:
retest cscs

msimberg:
@severinstrobl I just merged #4649 which creates conflicts with most PRs. Please let me know if you'd like me to rebase this one. We've split the `libs` directory into three intermediate directories with modules under those three directories instead of directly in `libs`.

severinstrobl:
@msimberg Sorry for not replying earlier, I was on vacation. I just rebased my request on the current master.

msimberg:
retest cscs

msimberg:
> @msimberg Sorry for not replying earlier, I was on vacation. I just rebased my request on the current master.

No problem at all and thanks for updating the PR. I'm merging this now since you seem to know what you're getting :) I'll try to add a note to the documentation about this behaviour (the lack of any additional synchronization when suspending). Please keep us updated about how this works in your use case.

PR #4952, opened 2020-09-08
Make parallel_executor use thread_pool_executor spawning mechanism

msimberg:
- Moves the hierarchical spawning helper functions to a separate file to make it easier to include elsewhere
- Makes the `parallel_executor` use the hierarchical spawning functions
- Typedefs `thread_pool_executor` to `parallel_executor` since the latter now behaves the same as the former
- The constructors are *mostly* compatible, except for the old `num_tasks`/`num_spread` and the new `hierarchical_threshold` options (I'm counting on this being enough compatibility, but if there are complaints we can make them more compatible)
- Change uses of `pool_executor` and `thread_pool_executor` to `parallel_executor`

This means that the `parallel_executor` now has a constructor that takes a `thread_pool_base` as well, and it now covers most of the use cases that `parallel_executor`, `thread_pool_executor`, and `pool_executor` previously covered. `block_executor`/`restricted_thread_pool_executor` are still a bit more flexible for cases when tasks need to be placed more precisely.


hkaiser (dismissed):
LGTM, thanks! Very nice!

biddisco:
I have a question. If I understand thing correctly - The hierarchical spawning creates a task bound to each core that in turn spawns tasks according to the parallel request. The advantage of this is that each core generates N tasks on it's own queues and helps greatly on benchmarks like the future overhead test. 
The future overhead test is something of a special case, as it tests the case when the queues are empty and working on tiny tasks - and in this situation, every core tries to steal from every other core and the task that generates the original parallel tasks is now split across all cores, so the actual task creation is accelerated (and not hindered by constant request for stealing).

In the more general case, when queues are not empty (we hope), is it not sufficient to simply bulk create N tasks per core on the queues of those cores without the extra initial task (per core) that creates the N tasks per core? (I may have misunderstood how the executor operates as I didn't yet look properly at the code, so please correct me if I've misunderstood the operation).

msimberg:
> I have a question. If I understand thing correctly - The hierarchical spawning creates a task bound to each core that in turn spawns tasks according to the parallel request. The advantage of this is that each core generates N tasks on it's own queues and helps greatly on benchmarks like the future overhead test.
> The future overhead test is something of a special case, as it tests the case when the queues are empty and working on tiny tasks - and in this situation, every core tries to steal from every other core and the task that generates the original parallel tasks is now split across all cores, so the actual task creation is accelerated (and not hindered by constant request for stealing).
> 
> In the more general case, when queues are not empty (we hope), is it not sufficient to simply bulk create N tasks per core on the queues of those cores without the extra initial task (per core) that creates the N tasks per core? (I may have misunderstood how the executor operates as I didn't yet look properly at the code, so please correct me if I've misunderstood the operation).

If I understand your question correctly, that's mostly an orthogonal concern, but you may be onto something. The main benefit of the hierarchical scheduling is that it reduces the number of tasks that get added from one worker thread to another one and as such hopefully reduces contention and cache misses. That said, the extra task that is used for hierarchical spawning is not for free, which is why there's the threshold for when to actually do hierarchical spawning (this can be adjusted to either force or inhibit hierarchical spawning). It could be that the threshold should be higher when queues already have a bunch of tasks in them, but I haven't made any measurements to test your hypothesis.

Somewhat unrelated, but I just had an idea from your comment. The "mid-level" tasks that spawn the actual work on each worker thread should actually participate in the work themselves. This would make the number of spawned tasks exactly the same as without the hierarchical spawning.

PR #4953, opened 2020-09-08
Fixing MSVC problems after recent restructurings

hkaiser:
@msimberg please check whether this is ok with you.


libs/CMakeLists.txt
```diff
@@ -69,6 +69,12 @@ foreach(lib ${HPX_LIBS})
 
     set_target_properties(hpx_${lib} PROPERTIES FOLDER "Core")
 
+    if((${lib} STREQUAL "core") AND MSVC)
+      if(HPX_WITH_SWAP_CONTEXT_EMULATION)
+        target_link_options(hpx_${lib} PRIVATE "/EXPORT:switch_to_fiber")
+      endif()
+    endif()
```
msimberg, line 76: This could go at the bottom together with the other `target_link_libraries` since it's only for `core` but this is ok as well.

hkaiser: Good point, I'll change that.

msimberg (dismissed):
Thanks @hkaiser for looking into this! Looks like CI was happy with this in the last round. I pushed one more commit to enable `HPX_WITH_SWAP_CONTEXT_EMULATION` on GitHub actions, but let's merge this asap if that comes back clean.

msimberg:
retest cscs

hkaiser:
I think this looks good now. The CI failures are unrelated, I believe.

PR #4954, opened 2020-09-10
Update Piz Daint Jenkins configurations

msimberg:
The Cray clang compiler was updated from being based on clang 9 to 10. One previously non-compiling example now compiles. The move from clang 9 to 10 also introduced a new warning which is fixed here.


aurianer (approved):
LGTM thanks!

msimberg:
The other configurations seem to need recompilation. Will do that and update this with new the new versions as well.

hkaiser (dismissed):
LGTM, thanks!

msimberg:
The configurations on should now all work, but let's see if my manual testing corresponds to what jenkins says. Surprise, surprise, the new nvcc (10.2) fails to compile yet another test (stream this time). I've disabled that one as well with nvcc.

hkaiser:
> Surprise, surprise, the new nvcc (10.2) fails to compile yet another test

It never happened before - and yet again .... *sigh*

msimberg:
> > Surprise, surprise, the new nvcc (10.2) fails to compile yet another test
> 
> It never happened before - and yet again .... _sigh_

Indeed. In this case it might actually be something we could fix, since it's not an internal compiler error. I'm hoping that with hip-clang (and cray's clang) we could slowly start getting more of HPX compilable in device mode as well, and since those are stricter we might inadvertently help nvcc not crash. But as it is I'm not going to investigate the nvcc error further until someone needs it to compile badly.

msimberg:
Actually... we're being a bit mean towards `nvcc`. I just tested two of the CUDA tests that were previously disabled, and they now compile. So that's a net +1 tests enabled with `nvcc` now.

PR #4955, opened 2020-09-10
Use object libraries for modules

### LINKED ISSUE #4135: Use object libraries for modules

msimberg:
Fixes #4135. Still a draft as I think this will have the same problems as in #4649 (i.e. long configuration times). At least I have not done anything about them. Profiling on Linux gives little useful information and the configuration times are mostly unchanged, so I'll first see what the GitHub actions builders say. @hkaiser at some point I'd need you to test this on Windows, but this is not urgent at all. I'm just opening this up to get this started again.

### DIFF
--- a/.cmake-format.py
+++ b/.cmake-format.py
@@ -350,6 +350,7 @@ with section("parse"):
                                     'EXCLUDE_FROM_GLOBAL_HEADER': '+',
                                     'GLOBAL_HEADER_GEN': 1,
                                     'HEADERS': '+',
+                                    'OBJECTS': '+',
                                     'MODULE_DEPENDENCIES': '+',
                                     'SOURCES': '+'},
                         'pargs': { 'flags': ['FORCE_LINKING_GEN',
--- a/cmake/HPX_AddCompileTest.cmake
+++ b/cmake/HPX_AddCompileTest.cmake
@@ -5,7 +5,7 @@
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 function(add_hpx_compile_test category name)
-  set(options FAILURE_EXPECTED NOLIBS)
+  set(options FAILURE_EXPECTED NOLIBS OBJECT)
   set(one_value_args SOURCE_ROOT FOLDER)
   set(multi_value_args SOURCES COMPONENT_DEPENDENCIES DEPENDENCIES)
 
@@ -17,18 +17,33 @@ function(add_hpx_compile_test category name)
   if(${name}_NOLIBS)
     set(_additional_flags ${_additional_flags} NOLIBS)
   endif()
+  if(${name}_OBJECT)
+    set(_additional_flags ${_additional_flags} OBJECT)
+  endif()
 
   string(REGEX REPLACE "\\." "_" test_name "${category}.${name}")
 
-  add_hpx_executable(
-    ${test_name}
-    SOURCE_ROOT ${${name}_SOURCE_ROOT}
-    SOURCES ${${name}_SOURCES}
-    EXCLUDE_FROM_ALL EXCLUDE_FROM_DEFAULT_BUILD
-    FOLDER ${${name}_FOLDER}
-    COMPONENT_DEPENDENCIES ${${name}_COMPONENT_DEPENDENCIES}
-    DEPENDENCIES ${${name}_DEPENDENCIES} ${_additional_flags}
-  )
+  if(${name}_OBJECT)
+    add_hpx_library(
+      ${test_name}
+      SOURCE_ROOT ${${name}_SOURCE_ROOT}
+      SOURCES ${${name}_SOURCES}
+      EXCLUDE_FROM_ALL EXCLUDE_FROM_DEFAULT_BUILD
+      FOLDER ${${name}_FOLDER}
+      COMPONENT_DEPENDENCIES ${${name}_COMPONENT_DEPENDENCIES}
+      DEPENDENCIES ${${name}_DEPENDENCIES} ${_additional_flags}
+    )
+  else()
+    add_hpx_executable(
+      ${test_name}
+      SOURCE_ROOT ${${name}_SOURCE_ROOT}
+      SOURCES ${${name}_SOURCES}
+      EXCLUDE_FROM_ALL EXCLUDE_FROM_DEFAULT_BUILD
+      FOLDER ${${name}_FOLDER}
+      COMPONENT_DEPENDENCIES ${${name}_COMPONENT_DEPENDENCIES}
+      DEPENDENCIES ${${name}_DEPENDENCIES} ${_additional_flags}
+    )
+  endif()
 
   add_test(
     NAME "${category}.${name}"
@@ -42,7 +57,6 @@ function(add_hpx_compile_test category name)
   endif()
 
   set_tests_properties("${category}.${name}" PROPERTIES RUN_SERIAL TRUE)
-
 endfunction(add_hpx_compile_test)
 
 function(add_hpx_compile_test_target_dependencies category name)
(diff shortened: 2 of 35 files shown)


msimberg:
@K-ballo since it seems you've been trying out this branch, have you noticed anything regarding the CMake configuration/generation times? I.e. is it still as bad as @hkaiser reported earlier or do you not notice much of a difference?

K-ballo:
> have you noticed anything regarding the CMake configuration/generation times?

With the default configuration flags, the generation phase is about 13 seconds longer on my machine (~33s vs ~20s).
There is no noticeable difference when configuring only core.

msimberg:
@hkaiser I've disabled object libraries for the `hpx_full` now so the circular dependencies don't matter at the moment. Would you mind trying this out again to see if it has any effect on the configuration times? Unfortunately I think it won't have much of a difference, but it'd be nice to get this in even in a limited form.

hkaiser:
@msimberg sorry, it took a while to get back to this. This seems to work just fine with VS, there are no significant timing differences to master, neither configuration, nor build times.

msimberg:
> @msimberg sorry, it took a while to get back to this. This seems to work just fine with VS, there are no significant timing differences to master, neither configuration, nor build times.

No problem, and thanks for testing again!

There are two remaining problems: pkgconfig file generation and switch_to_fiber. `switch_to_fiber.obj` does not seem to be built anymore now that the modules are object libraries. CMake is for some reason treating them differently. Did that work for you locally? Pkgconfig files are not generated correctly at the moment, and as I've said earlier I'd prefer to drop support for them. If someone would be interested in maintaining them that'd be helpful.

hkaiser:
> There are two remaining problems: pkgconfig file generation and switch_to_fiber. `switch_to_fiber.obj` does not seem to be built anymore now that the modules are object libraries. CMake is for some reason treating them differently. Did that work for you locally? 

I didn't try :/ Will check.

> Pkgconfig files are not generated correctly at the moment, and as I've said earlier I'd prefer to drop support for them. If someone would be interested in maintaining them that'd be helpful.

I'll ask @stevenrbrandt if he would be interested. Would there be an option to make those independent of the build system (i.e. static)?

stevenrbrandt:
@hkaiser @msimberg What is wrong with the pkgconfig files? They work for me. Regardless, they have been useful for me in a number of contexts.

msimberg:
> @hkaiser @msimberg What is wrong with the pkgconfig files? They work for me. Regardless, they have been useful for me in a number of contexts.

There's nothing wrong with them on master. The problem is that this breaks them. Object libraries end up in the linker flags as completely nonsense (CMake-specific) libraries. It may be possible to filter them out, but I don't know if that'd work with all CMake versions.

> I'll ask @stevenrbrandt if he would be interested. Would there be an option to make those independent of the build system (i.e. static)?

I'm not sure. I'll have a think about this.

stevenrbrandt:
I guess I'm not too worried about how they work, so long as I still have access to them. I can try to help maintain them if need be.

msimberg:
> I guess I'm not too worried about how they work, so long as I still have access to them. I can try to help maintain them if need be.

I've just pushed a fix which seems to work at least with CMake 3.19. I suppose you wouldn't be too bothered if this would require a pretty recent CMake (we're anyway considering bumping the minimum to 3.17 in #5165) and/or it wouldn't always work with every version of CMake? The fix currently filters out what seems to be a CMake implementation detail and that may change in future versions and stop this from working, but hopefully we can simply filter out something else if that does end up changing.

stevenrbrandt:
@msimberg recent cmakes are fine :)

hkaiser:
@msimberg I tried again on Windows, this time with `HPX_COROUTINES_WITH_SWAP_CONTEXT_EMULATION=On`. This results in linker errors for hpx_core:
```
45>------ Rebuild All started: Project: hpx_core, Configuration: Debug x64 ------
45>Building Custom Rule D:/Devel/hpx/worktree/feature1/libs/CMakeLists.txt
45>dummy.cpp
45>LINK : error LNK2001: unresolved external symbol switch_to_fiber
45>D:/Devel/hpx/worktree/feature1/build/Debug/lib/hpx_cored.lib : fatal error LNK1120: 1 unresolved externals
```
The corresponding object file was successfully compiled earlier, however:
```
33>------ Rebuild All started: Project: hpx_coroutines, Configuration: Debug x64 ------
33>Generating switch_to_fiber.obj
34>Building Custom Rule D:/Devel/hpx/worktree/feature1/libs/core/testing/CMakeLists.txt
33> Assembling: D:/Devel/hpx/worktree/feature1/libs/core/coroutines/src/switch_to_fiber.asm
33>Building Custom Rule D:/Devel/hpx/worktree/feature1/libs/core/coroutines/CMakeLists.txt
33>unity_0_cxx.cxx
33>coroutine_impl.cpp
33>coroutine_self.cpp
33>hpx_coroutines.vcxproj -> D:\Devel\hpx\worktree\feature1\build\libs\core\coroutines\hpx_coroutines.dir\Debug\hpx_coroutines.lib
```
(side note: it's not quite clear why the two `.cpp` files are compiled separately outside the unity build - but that's unrelated).

msimberg:
> @msimberg I tried again on Windows, this time with `HPX_COROUTINES_WITH_SWAP_CONTEXT_EMULATION=On`. This results in linker errors for hpx_core:
> 
> ```
> 45>------ Rebuild All started: Project: hpx_core, Configuration: Debug x64 ------
> 45>Building Custom Rule D:/Devel/hpx/worktree/feature1/libs/CMakeLists.txt
> 45>dummy.cpp
> 45>LINK : error LNK2001: unresolved external symbol switch_to_fiber
> 45>D:/Devel/hpx/worktree/feature1/build/Debug/lib/hpx_cored.lib : fatal error LNK1120: 1 unresolved externals
> ```
> 
> The corresponding object file was successfully compiled earlier, however:
> 
> ```
> 33>------ Rebuild All started: Project: hpx_coroutines, Configuration: Debug x64 ------
> 33>Generating switch_to_fiber.obj
> 33> Assembling: D:/Devel/hpx/worktree/feature1/libs/core/coroutines/src/switch_to_fiber.asm
> 33>Building Custom Rule D:/Devel/hpx/worktree/feature1/libs/core/coroutines/CMakeLists.txt
> 33>unity_0_cxx.cxx
> 33>coroutine_impl.cpp
> 33>coroutine_self.cpp
> 33>hpx_coroutines.vcxproj -> D:\Devel\hpx\worktree\feature1\build\libs\core\coroutines\hpx_coroutines.dir\Debug\hpx_coroutines.lib
> ```

Could you try what happens if you add a `target_link_libraries(hpx_${libname} PRIVATE ${${modulename}_OBJECTS})` after this line?

> (side note: it's not quite clear why the two `.cpp` files are compiled separately outside the unity build - but that's unrelated).

Not quite clear to me either, in the sense that I couldn't figure out what was going wrong when they were included in the unity build. When they were included there were random segfaults (if I remember correctly) or assertions firing.

hkaiser:
> > (side note: it's not quite clear why the two `.cpp` files are compiled separately outside the unity build - but that's unrelated).
> 
> Not quite clear to me either, in the sense that I couldn't figure out what was going wrong when they were included in the unity build. When they were included there were random segfaults (if I remember correctly) or assertions firing.

Ahh yes, I do remember now that you took those files explicitly out of the unity build.

hkaiser:
> Could you try what happens if you add a `target_link_libraries(hpx_${libname} PRIVATE ${${modulename}_OBJECTS})` after this line:
> 
> https://github.com/STEllAR-GROUP/hpx/blob/ccc155e33287d967d3d5f3c942e4ad761669794c/cmake/HPX_AddModule.cmake#L424
> 
> ?

Yep, that fixed it. Thanks!

hkaiser:
@msimberg thank you for looking into this!

PR #4959, opened 2020-09-14
Adding communications set

hkaiser:
A `communication_set` is a set of connected `communicator_node` objects that are connected using a binomial communication pattern. A `communication_set` can be created once and reused with any of the collective operations. This not only removes the overhead of establishing the connections for each instance of a collective operation, it also switches from the linear communication model to a tree based one.

Note that this PR does not change the collective operations (yet). These will be changed separately. Eventually all existing collective operations will be adapted to use the new `communication_set` mechanism.

### DIFF
--- a/libs/full/collectives/include/hpx/collectives/detail/communication_set_node.hpp
+++ b/libs/full/collectives/include/hpx/collectives/detail/communication_set_node.hpp
@@ -1,0 +1,172 @@
+//  Copyright (c) 2020 Hartmut Kaiser
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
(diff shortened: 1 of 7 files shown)


libs/full/collectives/include/hpx/collectives/detail/communication_set_node.hpp
```diff
@@ -0,0 +1,172 @@
...
+    constexpr std::size_t next_power_of_two(std::uint64_t v)
+    {
+        --v;
+        v |= v >> 1;
+        v |= v >> 2;
+        v |= v >> 4;
+        v |= v >> 8;
+        v |= v >> 16;
+        v |= v >> 32;
+        return static_cast<std::size_t>(++v);
```
msimberg, line 57: Don't know if this is where you got this from, but I just have to link it as it's a very useful resource for these kind of things: https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2.

msimberg (approved):
I have no objections to this, but also nothing useful to comment about it. @biddisco any comments?

PR #4963, opened 2020-09-15
Renaming files to avoid warnings from the vs build system

hkaiser:
Recent versions of VS have issued warnings from the build system that were caused by files having the same name in the same module. This PR disambiguates things.


aurianer (approved):
Thanks!

PR #4966, opened 2020-09-17
Add clang-format and cmake-format version prints

aurianer:
The docker image is installing the latest version, add print in circleci to know which version is currenly used.


hkaiser (approved):
LGTM, thanks!

PR #4969, opened 2020-09-18
Moving is_execution_policy and friends into namespace hpx

hkaiser:
- flyby: add `hpx::is_execution_policy_v` and friends

This improves conformance as the equivalent types are defined in `namespace std`

### DIFF
--- a/libs/parallelism/algorithms/include_compatibility/hpx/parallel/traits/projected.hpp
+++ b/libs/parallelism/algorithms/include_compatibility/hpx/parallel/traits/projected.hpp
@@ -10,12 +10,16 @@
 #include <hpx/algorithms/config/defines.hpp>
 #include <hpx/algorithm.hpp>
 
+// different versions of clang-format produce different results
+// clang-format off
 #if HPX_ALGORITHMS_HAVE_DEPRECATION_WARNINGS
 #if defined(HPX_MSVC)
-#pragma message("The header hpx/parallel/traits/projected.hpp is deprecated, \
+#pragma message(                                                               \
+    "The header hpx/parallel/traits/projected.hpp is deprecated, \
     please include hpx/algorithms/traits/projected.hpp")
 #else
 #warning "The header hpx/parallel/traits/projected.hpp is deprecated, \
     please include hpx/algorithms/traits/projected.hpp"
 #endif
 #endif
+// clang-format on
--- a/libs/parallelism/execution/include/hpx/execution/traits/is_execution_policy.hpp
+++ b/libs/parallelism/execution/include/hpx/execution/traits/is_execution_policy.hpp
@@ -1,4 +1,4 @@
-//  Copyright (c) 2016-2017 Hartmut Kaiser
+//  Copyright (c) 2016-2020 Hartmut Kaiser
 //
 //  SPDX-License-Identifier: BSL-1.0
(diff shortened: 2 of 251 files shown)


libs/parallelism/algorithms/include_compatibility/hpx/parallel/traits/projected.hpp
```diff
@@ -10,12 +10,16 @@
 #include <hpx/algorithms/config/defines.hpp>
 #include <hpx/algorithm.hpp>
 
+// different versions of clang-format produce different results
```
msimberg, line 13: Isn't this why we specify which version we use in CI?

libs/parallelism/execution/include/hpx/execution/traits/is_execution_policy.hpp
```diff
@@ -127,25 +138,77 @@ namespace hpx { namespace parallel { namespace execution {
...
+        is_vectorpack_execution_policy<T>::value;
+    /// \endcond
+}    // namespace hpx
+
+namespace hpx { namespace parallel { namespace execution {
+
+    template <typename T>
+    using is_execution_policy HPX_DEPRECATED_V(1, 6,
+        "hpx::parallel::execution::is_execution_policy is deprecated, use "
+        "hpx::is_execution_policy instead") = hpx::is_execution_policy<T>;
```
msimberg, line 181: My only question here is if these need a compatibility option. With this it's no longer possible to specialize the (alias) template in `hpx::parallel::execution`. The only case of a custom policy I know of right now is in `hpx-kokkos` but that's experimental enough that there's no point in a compatibility option for that. I also suspect that very few people will have added custom policies, so I think we're fine here. However, if we do the same to `is_executor` that would most likely need more boilerplate to keep specializations in `hpx::parallel::execution` working. Do you agree with this?

libs/parallelism/algorithms/include_compatibility/hpx/parallel/traits/projected.hpp
hkaiser: Yes, but it's difficult to always stay in sync with the version we use in CI.

libs/parallelism/execution/include/hpx/execution/traits/is_execution_policy.hpp
hkaiser: Fair points to both. My rationale was that it doesn't hurt to have the compatibility for a while. I'm happy to remove the compatibility layer. Also, I was not planning to touch the `is_executor` and friends traits (yet). We will have to reshuffle things there with the new executor interfaces anyways.

msimberg: Yeah, sorry, I didn't mean to say that we should remove these aliases. I also think we should keep them. I was thinking in the other direction, if these need a compatibility layer for *specializing* the traits, rather than just for getting the values of the traits, and I think I already answered my own question but wanted to know what you think. I think for this PR no change is needed here.

And yes, agreed on `is_executor` etc., not relevant quite yet.

hkaiser:
@msimberg can we go ahead with this?

msimberg (approved):

msimberg:
> @msimberg can we go ahead with this?

Yes. Sorry, didn't mean to keep this from going in. Thanks!

PR #4970, opened 2020-09-19
Making set algorithms conform to C++20

hkaiser:
This PR make sure the following algorithms conform to C++20

- [x] `set_intersection`
- [x] `set_difference`
- [x] `set_symmetric_difference`
- [x] `set_union`
- [x] `includes`
- [x] `merge`
- [x] `inplace_merge`

### DIFF
--- a/docs/sphinx/api/public_api.rst
+++ b/docs/sphinx/api/public_api.rst
@@ -51,7 +51,7 @@ Functions
 - :cpp:func:`hpx::for_each_n`
 - :cpp:func:`hpx::generate`
 - :cpp:func:`hpx::generate_n`
-- :cpp:func:`hpx::parallel::v1::includes`
+- :cpp:func:`hpx::includes`
 - :cpp:func:`hpx::parallel::v1::inplace_merge`
 - :cpp:func:`hpx::is_heap`
 - :cpp:func:`hpx::is_heap_until`
--- a/docs/sphinx/api/public_api.rst
+++ b/docs/sphinx/api/public_api.rst
@@ -83,10 +83,10 @@ Functions
 - :cpp:func:`hpx::parallel::v1::rotate_copy`
 - :cpp:func:`hpx::parallel::v1::search`
 - :cpp:func:`hpx::parallel::v1::search_n`
-- :cpp:func:`hpx::parallel::v1::set_difference`
+- :cpp:func:`hpx::set_difference`
 - :cpp:func:`hpx::set_intersection`
-- :cpp:func:`hpx::parallel::v1::set_symmetric_difference`
-- :cpp:func:`hpx::parallel::v1::set_union`
+- :cpp:func:`hpx::set_symmetric_difference`
+- :cpp:func:`hpx::set_union`
 - :cpp:func:`hpx::parallel::v1::sort`
 - :cpp:func:`hpx::parallel::v1::stable_partition`
 - :cpp:func:`hpx::parallel::v1::stable_sort`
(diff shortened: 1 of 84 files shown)


docs/sphinx/api/public_api.rst
```diff
@@ -83,10 +83,10 @@ Functions
...
 - :cpp:func:`hpx::parallel::v1::search`
 - :cpp:func:`hpx::parallel::v1::search_n`
-- :cpp:func:`hpx::parallel::v1::set_difference`
-- :cpp:func:`hpx::parallel::v1::set_intersection`
-- :cpp:func:`hpx::parallel::v1::set_symmetric_difference`
-- :cpp:func:`hpx::parallel::v1::set_union`
+- :cpp:func:`hpx::set_difference`
+- :cpp:func:`hpx::set_intersection`
+- :cpp:func:`hpx::set_symmetric_difference`
+- :cpp:func:`hpx::set_union`
```
msimberg, line 89: These can be added to the `hpx::ranges` list below as well.

msimberg (changes requested):

msimberg:
This looks like it might be related to the changes here: https://cdash.cscs.ch/testDetails.php?test=28421628&build=127919 (the same test fails with some other configurations as well)?

hkaiser:
> This looks like it might be related to the changes here: https://cdash.cscs.ch/testDetails.php?test=28421628&build=127919 (the same test fails with some other configurations as well)?

Yep, looks like I managed to break things. I'll investigate.

docs/sphinx/api/public_api.rst
hkaiser: Good point. Probably :cpp:func:`hpx::incudes` as well.

hkaiser:
> > This looks like it might be related to the changes here: https://cdash.cscs.ch/testDetails.php?test=28421628&build=127919 (the same test fails with some other configurations as well)?
> 
> Yep, looks like I managed to break things. I'll investigate.

@msimberg this should be fine now. I also managed to find a bug in `hpx::includes` that was there from the beginning \o/

hkaiser:
Please note that I will add another commit here that will make `merge` and `inplace_merge` conforming to c++20.

hkaiser:
@msimberg this should be fine now. Sorry for the large number of seemingly unrelated changes, apparently I did miss a couple of places when moving `is_execution_policy` to namespace hpx...

libs/parallelism/algorithms/tests/unit/container_algorithms/inplace_merge_range.cpp
```diff
@@ -177,16 +238,162 @@ void test_inplace_merge()
...
+
+    // Test projection.
+    {
+        typedef test::test_iterator<base_iterator, IteratorTag> iterator;
+
+        sol = res = org;
+
+        DataType val;
+        hpx::ranges::inplace_merge(
+            policy, iterator(res_first), iterator(res_middle),
```
msimberg, line 299:
```suggestion
            iterator(res_first), iterator(res_middle),
```
https://cdash.cscs.ch/viewBuildError.php?buildid=128319

msimberg:
> @msimberg this should be fine now. Sorry for the large number of seemingly unrelated changes, apparently I did miss a couple of places when moving `is_execution_policy` to namespace hpx...

No problem at all, they're not so easy to catch without the warnings. Thanks for doing this!

hkaiser:

PR #4974, opened 2020-09-20
Remove leftover format export hack

K-ballo:
Before there was a format module, this hack was necessary to have performance tests not depend on all of HPX. The hack was dropped once format became a module. The format export macro is a left-over.


hkaiser (approved):
LGTM, thanks!

PR #4975, opened 2020-09-21
Remove indirection layer in at_index_impl

K-ballo:
Remove indirection layer in at_index_impl

PR #4976, opened 2020-09-22
Remove indirection layer in at_index_impl

msimberg:
Retry of #4975 since I merged #4975 too early.

### DIFF
--- a/libs/core/type_support/include/hpx/type_support/pack.hpp
+++ b/libs/core/type_support/include/hpx/type_support/pack.hpp
@@ -147,25 +147,21 @@ namespace hpx { namespace util {
         };
 
 #if defined(HPX_HAVE_BUILTIN_TYPE_PACK_ELEMENT)
-        template <std::size_t I, typename Ts, bool C = (I < Ts::size)>
-        struct at_index_impl
+        template <std::size_t I, typename Ts, bool InBounds = (I < Ts::size)>
+        struct at_index_impl : empty
         {
-            using type = empty;
         };
 
         template <std::size_t I, typename... Ts>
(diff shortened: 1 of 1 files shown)


hkaiser (dismissed):
LGTM, thanks!

K-ballo:
That's a nice warning, this implementation is wrong.. was it always wrong?

msimberg:
> That's a nice warning, this implementation is wrong.. was it always wrong?

Feel free to ignore that last commit completely if necessary. It's most likely just me who just made it wrong...

K-ballo:
> It's most likely just me who just made it wrong...

No, it was always wrong... The empty overload was never even considered.

The original implementation had a selection function that wasn't templated on the index (see http://talesofcpp.fusionfenix.com/post-22/true-story-efficient-packing#the-case-of-the-%60tuple%60), that's why those were members of the implementation. For some reason, when I added this to HPX I chose to add a seemingly redundant template parameter `J` (I must have had _some_ reason), but failed to add one to the fallback overload.

The proper fix is to add `template <std::size_t J>` to the empty overload.

libs/core/type_support/include/hpx/type_support/pack.hpp
```diff
@@ -177,22 +173,23 @@ namespace hpx { namespace util {
         {
         };
 
+        template <std::size_t J>
+        static empty at_index_check(...);
```
msimberg, line 177: @K-ballo this is what you meant?

K-ballo: Yes, now this overload will be included in the `at_index_check<I>` call.

PR #4977, opened 2020-09-23
Cleaning up debug::print functionalities

hkaiser:
- move as much as possible into separate compilation units
- break circular dependency between threading_base and debugging modules
- flyby: disable debug output from numa allocator

### DIFF
--- a/libs/core/debugging/CMakeLists.txt
+++ b/libs/core/debugging/CMakeLists.txt
@@ -17,7 +17,7 @@ set(debugging_compat_headers
     hpx/util/debug/demangle_helper.hpp
 )
 
-set(debugging_sources attach_debugger.cpp backtrace.cpp)
+set(debugging_sources attach_debugger.cpp backtrace.cpp print.cpp)
 
 include(HPX_AddModule)
 add_hpx_module(
--- a/libs/core/debugging/include/hpx/debugging/print.hpp
+++ b/libs/core/debugging/include/hpx/debugging/print.hpp
@@ -9,34 +9,16 @@
 #include <hpx/config.hpp>
 
 #include <array>
-#include <bitset>
 #include <chrono>
-#include <cmath>
 #include <cstddef>
 #include <cstdint>
-#include <cstring>
-#include <iomanip>
 #include <iostream>
-#include <iterator>
 #include <sstream>
 #include <string>
+#include <tuple>
 #include <type_traits>
 #include <utility>
 #include <vector>
-
-#if defined(__linux) || defined(linux) || defined(__linux__)
-#include <sys/mman.h>
-#include <unistd.h>
-#elif defined(__APPLE__)
-#include <crt_externs.h>
-#include <unistd.h>
-#define environ (*_NSGetEnviron())
-#elif defined(HPX_WINDOWS)
-#include <winsock2.h>
-#define environ _environ
-#else
-extern char** environ;
-#endif
 
 // ------------------------------------------------------------
 // This file provides a simple to use printf style debugging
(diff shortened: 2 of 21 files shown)


msimberg:
@biddisco would you mind having a look at this?

biddisco:
Thanks for the cleanup. As @hkaiser mentioned some time ago, arguments to functions that are themselves function calls are still evaluated even when the debug is disabled - so I added a LAZY_XXX macro that can be used to wrap them. Since then I've started to worry that in fact even the simple calls like
```
    my_debug.debug(debug::str<8>(thing), debug::ptr<>(this), debug::int<4>(val), ...);
```
are actually triggering calls to the formatters themselves and the results are being discarded. My initial testing with godbolt was some time ago and I don't remember if I checked properly that these formatters that create temporary stringstream instances and then have their results discarded are completely elided and so I want to test further. 

If it turns out that the formatters are being called, then there doesn't seem to be any way better than using a macro to wrap the entire debug statement (which I strenuously wanted to avoid)

The existing 
```
#define HPX_DP_LAZY(Expr, printer) printer.eval([&] { return Expr; })
```
could be modified to something like 
```
#define HPX_DEBUG_PRINT(printer, Expr) printer.eval([&] { return printer.Expr; })
```
so that one could write
```
HPX_DEBUG_PRINT(my_debug, debug(debug::str<8>(thing), debug::ptr<>(this), debug::int<4>(val), ...));
HPX_DEBUG_PRINT(my_debug, trace(debug::str<8>(thing), debug::ptr<>(this), debug::int<4>(val), ...));
HPX_DEBUG_PRINT(my_debug, warning(debug::str<8>(thing), debug::ptr<>(this), debug::int<4>(val), ...));
// &etc
```
and have the macro fill in the printer object to make repetition a bit less. This would also mean that it would not be necessary to use the LAZY macro inside the args since the whole debug statement would be correctly wrapped already. 

I wish there was a way of doing it that didn't require a macro at all. I will do a few more tests with godbolt ASAP in any case to satisfy my anxiety.

hkaiser:
@biddisco with this cleanup I didn't look into the problems you mentioned in your comment. I was focusing on reducing the compile time (#inclusion) overheads for the existing code by moving things out of the header as much as possible. No functional or API changes were applied.

biddisco:
> @biddisco with this cleanup I didn't look into the problems you mentioned in your comment. I was focusing on reducing the compile time (#inclusion) overheads for the existing code by moving things out of the header as much as possible. No functional or API changes were applied.

Understood. I just wanted you to know that the debug print stuff is something that is on my mind and I am/have been thinking about and tweaking.

biddisco:
Something I wanted to mention, but forgot ...
When the declarations of some of the string formatters are in the header file, the compiler can see the definitions and possibly elide some of them when they produce redundant values. Moving them into the .cpp may well make this impossible, since the declarations are no longer visible in all the places where they are used. This is obviously moot if the compiler can't elide them anyway (whether they are in header or source). If this is merged, then I think it becomes necessary to use a lazy evaluation regardless and so I would like to have a bit of time to look at this before it is merged (and possibly replace some of the existing debug uses with wrapped ones).

hkaiser:
> Something I wanted to mention, but forgot ...
> When the declarations of some of the string formatters are in the header file, the compiler can see the definitions and possibly elide some of them when they produce redundant values. Moving them into the .cpp may well make this impossible, since the declarations are no longer visible in all the places where they are used. This is obviously moot if the compiler can't elide them anyway (whether they are in header or source). If this is merged, then I think it becomes necessary to use a lazy evaluation regardless and so I would like to have a bit of time to look at this before it is merged (and possibly replace some of the existing debug uses with wrapped ones).

Which of the facilities are you referring to? I believe the actual instantiation mechanisms have not changed. This PR mostly moves the non-template parts of the code to separate compilation units, nothing else (I hope).

hkaiser:
@msimberg, @biddisco can we go ahead with this now? Is there anything you would like to be changed?

PR #4979, opened 2020-09-24
Actually enable unity builds on Jenkins

msimberg:
No comment.

### DIFF
--- a/libs/core/coroutines/CMakeLists.txt
+++ b/libs/core/coroutines/CMakeLists.txt
@@ -100,6 +100,13 @@ elseif(HPX_COROUTINES_WITH_SWAP_CONTEXT_EMULATION)
   )
 endif()
 
+set_source_files_properties(
+  src/detail/coroutine_impl.cpp PROPERTIES SKIP_UNITY_BUILD_INCLUSION TRUE
+)
+set_source_files_properties(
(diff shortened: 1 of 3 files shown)


hkaiser (dismissed):
LGTM, thanks!

aurianer (dismissed):

libs/core/coroutines/CMakeLists.txt
```diff
@@ -96,6 +96,13 @@ elseif(HPX_COROUTINES_WITH_SWAP_CONTEXT_EMULATION)
   )
 endif()
 
+set_source_files_properties(
+  src/detail/coroutine_impl.cpp PROPERTIES SKIP_UNITY_BUILD_INCLUSION TRUE
+)
+set_source_files_properties(
+  src/detail/coroutine_self.cpp PROPERTIES SKIP_UNITY_BUILD_INCLUSION TRUE
+)
+
```
msimberg, line 105: These were causing problems in release builds. Basically what happens is that `local_self` returns a null thread id, presumably as a result of the `thread_local` self thread id possibly being instantiated multiple times or something like that. However, I don't see how something like that could be triggered by compiling these files together, and I don't see that we're doing anything wrong with the `thread_local`. Does anyone see what the problem could be? If not, this will work as a workaround, but I'd be interested in understanding this. I'm just out of ideas right now.

PR #4980, opened 2020-09-24
Add ranges overloads to for_loop (and variants)

msimberg:
This adds `for_loop` overloads to `hpx::ranges`. I've left out the `for_loop_n` and `for_loop_n_strided` overloads with sentinels, since the `size` parameter replaces the sentinel and they wouldn't actually be overloads with sentinels anymore. They would be identical to those in `hpx`.


libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_loop.hpp
```diff
@@ -0,0 +1,1369 @@
...
+        // clang-format on
+        friend void tag_invoke(
+            hpx::ranges::for_loop_t, Iter first, Sent last, Args&&... args)
+        {
+            static_assert(sizeof...(Args) >= 1,
+                "for_loop must be called with at least a function object");
+
+            return for_loop(parallel::execution::seq, first, last,
+                std::forward<Args>(args)...);
+        }
```
hkaiser, line 1141: I usually rather copy the code from the execution policy version instead of calling it to avoid to double tag-dispatch, but that's a minor issue.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_loop.hpp
```diff
@@ -0,0 +1,1369 @@
...
+        // clang-format off
+        template <typename ExPolicy, typename R, typename... Args,
+            HPX_CONCEPT_REQUIRES_(
+                hpx::is_execution_policy<ExPolicy>::value &&
+                hpx::traits::is_range<R>::value
+            )>
+        // clang-format on
+        friend typename parallel::util::detail::algorithm_result<ExPolicy>::type
+        tag_invoke(
+            hpx::ranges::for_loop_t, ExPolicy&& policy, R&& rng, Args&&... args)
```
hkaiser, line 1152: Is a pair of integers satisfying all requirement of a range?

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_loop.hpp
```diff
@@ -0,0 +1,1369 @@
...
+            typename... Args,
+            HPX_CONCEPT_REQUIRES_(
+                hpx::is_execution_policy<ExPolicy>::value &&
+                std::is_integral<Size>::value &&
+                hpx::traits::is_range<Rng>::value
+            )>
+        // clang-format on
+        friend typename parallel::util::detail::algorithm_result<ExPolicy>::type
+        tag_invoke(hpx::ranges::for_loop_n_t, ExPolicy&& policy, Rng&& rng,
+            Size size, Args&&... args)
```
hkaiser, line 1289: Wouldn't this be better represented by just taking a range (iter/sent)? Should we add such an overload in the first place?

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_loop.hpp
msimberg: Good point. I can update them to not do that (including the non-ranges versions).

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_loop.hpp
msimberg: Should it? This would have the same behaviour as `std::vector<T> c(...); hpx::for_loop(std::begin(c), std::end(c), ...);`, i.e. it just passes the iterator to the element-wise function.

msimberg: Or do you actually mean if we should accept `std::pair<Integral>` as a range?

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_loop.hpp
msimberg: There is an overload taking an explicit iterator and sentinel pair. This is overload would just be convenience for `for_loop(std::begin(c), std::end(c), ...)`.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_loop.hpp
msimberg: I updated them to avoid going through the tag dispatch twice.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_loop.hpp
hkaiser: The original `for_loop` supports being used with integral arguments: `for_loop(0, 10, [](int i) {});`. If the proposed `ranges::for_loop` does not support this (and as a pair of integers does not satisfy the requirements of a range, see http://eel.is/c++draft/range.range#3.3) I tend to think that we shouldn't add a corresponding overload.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_loop.hpp
hkaiser: Right. For this reason I'm advocating to not add this particular overload. There is no precedence (IIRC) where a range based algorithm uses the `begin(r)` of the given range `r` only. For instance, `ranges::for_each_n` has no overload that expects a range as its argument, just one expecting an iterator and a count (http://eel.is/c++draft/alg.foreach#lib:for_each_n__).

msimberg: Sorry, my bad. I missed that this was on the `_n` version. Yes, you're right that `for_each_n` doesn't have an equivalent overload and I'm indifferent about keeping this or not. It's a very minor convenience wrapper anyway. So remove?

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_loop.hpp
msimberg: I'm confused. Is your concern that `hpx::for_loop` accepts integral indices *or* iterators, and as such `hpx::ranges::for_loop` should as well support something analogous, but not only ranges?

`hpx::for_loop` supports iterators (and integral arguments). `hpx::ranges::for_loop` supports iterator/sentinel pairs and ranges. This to me is analogous to `hpx::for_each` and `hpx::ranges::for_each` except for the additional support for integral arguments on `hpx::for_loop`.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_loop.hpp
hkaiser: Yes, I'd be in favor of removing this overload.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_loop.hpp
hkaiser: > Is your concern that `hpx::for_loop` accepts integral indices _or_ iterators, and as such `hpx::ranges::for_loop` should as well support something analogous, but not only ranges?

Yes, that is my concern. OTOH, I understand where you're coming from. 

Originally `hpx::for_loop` (based on [N4755](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/n4755.pdf)) was meant to provide an alternative to OMP parallel for loops in C++, i.e. was meant to work with integral types most and foremost. It just turned out that iterators fulfill all of the requirements such an algorithm imposes onto its integral (loop boundary) arguments, which lead to the unwieldy thing we have today that accepts both, integral and iterator types. As such, reducing the `hpx::ranges::for_loop` to supporting just iterators does not 'feel' right.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_loop.hpp
msimberg: Done.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_loop.hpp
msimberg: Ok, I see your point of view as well. In that case wouldn't all of the ranges overloads for `for_loop` be slightly 'wrong', not just the range but also iterator/sentinel overloads...? Again, I'm mostly indifferent, but I think this makes `for_loop` a nice generalization of `for_each` so I'm slightly (but only slightly) in favor of keeping this (but then I opened the PR, so that shouldn't be a surprise).

There is no update to the parallelism ts after the ranges changes, right? I.e. currently there are no proposals for what, if any, overloads there should be for `hpx::ranges::for_loop`? Then it's a question of if we provide these overloads just because we can, or if we are conservative and stick to only what's proposed/standardized.

hkaiser: > There is no update to the parallelism ts after the ranges changes, right? 

None that I know of.

> Then it's a question of if we provide these overloads just because we can, or if we are conservative and stick to only what's proposed/standardized.

All of our ranges overloads that take an execution policy are 'because we can'. It's way too late to play the conservative card now ;-)

Let's have your overloads then. I don't want to be in the way of progress.

msimberg: No problem :P I don't *really* badly need these but I think they are nice enough to include. Then we go with this. Thanks! And thanks for being appropriately skeptical, I hadn't thought about these issues so it's good to be on the same page.

docs/sphinx/api/public_api.rst
```diff
@@ -129,6 +129,10 @@ Functions
 - :cpp:func:`hpx::ranges::set_intersection`
 - :cpp:func:`hpx::ranges::set_symmetric_difference`
 - :cpp:func:`hpx::ranges::set_union`
+- :cpp:func:`hpx::ranges::for_loop`
+- :cpp:func:`hpx::ranges::for_loop_strided`
+- :cpp:func:`hpx::ranges::for_loop_n`
+- :cpp:func:`hpx::ranges::for_loop_n_strided`
```
hkaiser, line 135: Should we remove mentioning `ranges::for_loop_n` here as well?

msimberg: Yes, good catch!

msimberg: Done.

PR #4982, opened 2020-09-25
Add page about citing HPX to documentation

msimberg:
I added a separate section in the documentation and link to it on the intro page. Prominent enough? I also left the links in place in the readme. I don't think it hurts to have it in both places.


hkaiser (approved):
LGTM, thanks! Very nice.

PR #4985, opened 2020-09-27
Compatibility header generation

K-ballo:
Automatically generates compatibility headers during configuration step. Arguments of the form `old/header.hpp => new/header.hpp` trigger compatibility header generation, while simple arguments `old/header.hpp` continue to be treated as before. Additionally, compatibility headers are now placed under "Generated files" or "Compatibility header files" source groups.

I have decided to only suggest a single replacement `#include` for each compatibility header. A functionality should only be defined by one header, even if multiple headers make the functionality available.

Only two compatibility headers introduce additional definitions, the ones for module `asnyc_cuda`, and were left as-is.

The overall state of compatibility headers looks frail. Not all headers under `include_compatibility` were being given to the corresponding module definition (someone please double check I got them all). The replacement `#include` sometimes suggests including the entire module while other times it suggests the actual new header, and the decision appears to be arbitrary.

Hopefully automatic generation will help keep better care of compatibility headers, by removing all the boilerplate and concentrating all the relevant information in a single concise definition.


libs/core/datastructures/CMakeLists.txt
```diff
@@ -34,15 +34,17 @@ set(datastructures_headers
...
+# cmake-format: off
 set(datastructures_compat_headers
-    hpx/datastructures.hpp
-    hpx/traits/supports_streaming_with_any.hpp
-    hpx/traits/is_tuple_like.hpp
-    hpx/util/any.hpp
-    hpx/util/optional.hpp
-    hpx/util/tagged.hpp
-    hpx/util/tuple.hpp
+    hpx/datastructures.hpp => hpx/modules/util.hpp
```
msimberg, line 39: I think this one was most probably a mistake... as in it should point to `hpx/modules/datastructures.hpp` instead.

msimberg (commented):
Thank you @K-ballo! This is very much needed. The reason for the inconsistencies is the manual process, and the fact that we started off with just forwarding to the equivalent header and then went to forwarding to the module header or API headers. This helps a lot in keeping things consistent. From what I can tell you've got all the compatibility headers updated correctly.

For the `async_cuda` headers we can move the `using` statements to the `hpx/async_cuda/target.hpp` file (with a proper deprecation warning). Let me know if you'd like me to do that.

APEX will need some changes as well to deal with the new deprecation warnings. I can take care of that.

msimberg:
@K-ballo I've pushed the changes that I think were relevant. Unless I've made a mistake somewhere I'm happy with this. Thanks again for doing this!

aurianer (approved):
LGTM thanks a lot!

K-ballo:
Now that the `async_cuda` compatibility headers are gone we could simplify the implementation by dropping support for "manual" compatibility headers.

msimberg:
> Now that the `async_cuda` compatibility headers are gone we could simplify the implementation by dropping support for "manual" compatibility headers.

I support that. I don't think we're missing out on much by not supporting those anymore.

msimberg (approved):
Thanks @K-ballo! Looks good to me.

PR #4986, opened 2020-09-28
Remove serialization to functional module dependency

msimberg:
Extracts the change from #4955.

### DIFF
--- a/libs/core/serialization/CMakeLists.txt
+++ b/libs/core/serialization/CMakeLists.txt
@@ -162,7 +162,6 @@ add_hpx_module(
     hpx_debugging
     hpx_errors
     hpx_format
-    hpx_functional
     hpx_hashing
     hpx_preprocessor
     hpx_type_support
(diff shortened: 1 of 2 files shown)


hkaiser (approved):
LGTM, thanks!

K-ballo:
We should probably be using `std::function` in all our user-facing interfaces, and use `hpx::function_nonser` for any internal erasure for a user-facing generic parameters.

msimberg:
> We should probably be using `std::function` in all our user-facing interfaces, and use `hpx::function_nonser` for any internal erasure for a user-facing generic parameters.

Good idea. I think the only "officially" public interface that would require a change in that case is for `hpx::init/start`. Luckily that's not even a breaking change.

hkaiser:
> > We should probably be using `std::function` in all our user-facing interfaces, and use `hpx::function_nonser` for any internal erasure for a user-facing generic parameters.
> 
> Good idea. I think the only "officially" public interface that would require a change in that case is for `hpx::init/start`. Luckily that's not even a breaking change.

There are also the APIs related to registering startup/shutdown functions...

PR #4990, opened 2020-09-30
Fixing some warnings from HPX complaining about use of obsolete types

hkaiser:
There are probably more warnings caused by the recent enabling of the deprecation macros. The changes in this PR make Phylanx compile without those.

### DIFF
--- a/libs/core/datastructures/include/hpx/datastructures/tagged.hpp
+++ b/libs/core/datastructures/include/hpx/datastructures/tagged.hpp
@@ -1,5 +1,5 @@
 //  Copyright Eric Niebler 2013-2015
-//  Copyright 2015 Hartmut Kaiser
+//  Copyright 2015-2020 Hartmut Kaiser
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
@@ -43,7 +43,7 @@ namespace hpx { namespace util {
             struct unpack_getter
             {
                 typedef typename Tag::template getter<T,
-                    typename tuple_element<I, T>::type, I>
+                    typename hpx::tuple_element<I, T>::type, I>
                     type;
 
             private:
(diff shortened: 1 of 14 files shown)


msimberg:
Thanks @hkaiser for fixing these. Any idea why these wouldn't have been triggered earlier? Or did I just not notice before merging?

hkaiser:
> Thanks @hkaiser for fixing these. Any idea why these wouldn't have been triggered earlier? Or did I just not notice before merging?

I have no idea. Also I'm sure that there are more of those. Especially the use of `execution::is_sequenced_execution_policy` by various algorithm implementations. I have not had the time to do a thorough check for those, however.

msimberg:
> > Thanks @hkaiser for fixing these. Any idea why these wouldn't have been triggered earlier? Or did I just not notice before merging?
> 
> I have no idea. Also I'm sure that there are more of those. Especially the use of `execution::is_sequenced_execution_policy` by various algorithm implementations. I have not had the time to do a thorough check for those, however.

You (or whoever found these) are not using gcc by any chance? It looks to me like clang doesn't warn about deprecated alias templates: https://godbolt.org/z/o7dYd5. I shall be enabling deprecation warnings on other builders as well (it's only enabled on CircleCI now).

hkaiser:
> You (or whoever found these) are not using gcc by any chance (edit: I suppose MSVC might warn about them?)? It looks to me like clang doesn't warn about deprecated alias templates: https://godbolt.org/z/o7dYd5. I shall be enabling deprecation warnings on other builders as well (it's only enabled on CircleCI now).

These warnings were produced by MSVC.

msimberg (changes requested):
Thanks! Looks good to me except for the clang formatting changes reported by CircleCI.

aurianer (approved):
LGTM thanks!

hkaiser:
> Thanks! Looks good to me except for the clang formatting changes reported by CircleCI.

Different version of clang-format again *sigh*. I'll fix it.

PR #4991, opened 2020-10-03
Avoid unnecessarily including entire modules

K-ballo:
Replace some internal inclusions of `hpx/module/` headers for the specific headers required for implementation. Reduces number of headers parsed (msvc, windows) from 479'032 to 432'162, roughly 10%. Only looks at replacing the fattest modules.

Fly-by changes:
- Replace some obsolete uses of `util::decay` by `std::decay`.
- Replace some internal uses of `util::invoke` by `HPX_INVOKE`.
- Simplify some expressions with superfluous binds.
- Add missing includes.

### DIFF
--- a/hpx/runtime/actions/continuation_impl.hpp
+++ b/hpx/runtime/actions/continuation_impl.hpp
@@ -6,13 +6,13 @@
 
 #pragma once
 
-#include <hpx/functional/invoke_result.hpp>
-#include <hpx/async_distributed/applier/detail/apply_implementations_fwd.hpp>
 #include <hpx/async_distributed/applier/apply.hpp>
+#include <hpx/async_distributed/applier/detail/apply_implementations_fwd.hpp>
+#include <hpx/functional/invoke_result.hpp>
 #include <hpx/runtime/naming/id_type.hpp>
 #include <hpx/serialization/access.hpp>
-#include <hpx/type_support/decay.hpp>
 
+#include <type_traits>
 #include <utility>
 
 namespace hpx { namespace actions {
@@ -21,7 +21,7 @@ namespace hpx { namespace actions {
     struct continuation_impl
     {
     private:
-        typedef typename util::decay<Cont>::type cont_type;
+        typedef typename std::decay<Cont>::type cont_type;
 
     public:
         continuation_impl() {}
--- a/hpx/lcos/packaged_action.hpp
+++ b/hpx/lcos/packaged_action.hpp
@@ -27,6 +27,7 @@
 
 #include <exception>
 #include <memory>
+#include <type_traits>
 #include <utility>
 
 ///////////////////////////////////////////////////////////////////////////////
@@ -199,7 +200,7 @@ namespace hpx { namespace lcos {
                         << hpx::actions::detail::get_action_name<action_type>()
                         << ", " << id << ") args(" << sizeof...(Ts) << ")";
 
-            using callback_type = typename util::decay<Callback>::type;
+            using callback_type = typename std::decay<Callback>::type;
 
 #if defined(HPX_HAVE_NETWORKING)
             auto&& f = detail::parcel_write_handler_cb<Result, callback_type>{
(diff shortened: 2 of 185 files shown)


hkaiser (approved):
LGTM, thanks!

hpx/runtime/actions/continuation_impl.hpp
```diff
@@ -54,11 +55,10 @@ namespace hpx { namespace actions {
         template <typename Archive>
         HPX_FORCEINLINE void serialize(Archive& ar, unsigned int const)
         {
-            ar & cont_ & target_;
+            ar& cont_& target_;
```
aurianer, line 58: I think this formatting was on purpose cause it's the & operator right?

aurianer (approved):
Thanks a lot!

hpx/runtime/actions/continuation_impl.hpp
hkaiser: clang-format always messes this up. We either live with it or surround this kind of code with `// clang-format off/on` comments.

K-ballo: I can manually revert the clang-formatting changes on these files.

K-ballo:
This is hitting some unfortunate compiler bugs https://gcc.godbolt.org/z/xoavv1

I will have to revert the use of `HPX_INVOKE` in lambdas for now.

hkaiser:
Should we add the `std::random` facilities to inspect to catch the missing #include in the future?

hpx/runtime/actions/continuation_impl.hpp
msimberg: Yes please, if it's not too much effort. They will need the `// clang-format` comments though.

msimberg:
> This is hitting some unfortunate compiler bugs https://gcc.godbolt.org/z/xoavv1
> 
> I will have to revert the use of `HPX_INVOKE` in lambdas for now.

Does that mean that we should use the macro everywhere else from now on?

K-ballo:
> Does that mean that we should use the macro everywhere else from now on?

The macro should be preferred for internal implementations, as it reduces the call depth for function objects, but not in examples or other user facing code.

hpx/runtime/actions/continuation_impl.hpp
K-ballo: I have manually reverted clang-formatting changes on these files.

msimberg: Thank you!

msimberg:
@K-ballo have you seen this: https://cdash.cscs.ch/viewBuildError.php?buildid=130585? That particular build has `HPX_WITH_NETWORKING=OFF` which probably leads to different includes. Let me know if you'd like me to have a look at that.

Also, we should be able to mark `hpx::util::decay` as deprecated now, right?

K-ballo:
> @K-ballo have you seen this: https://cdash.cscs.ch/viewBuildError.php?buildid=130585?

@msimberg I have not seen that one. It looks like all the relevant information has been stripped. Could I get a real build log?

K-ballo:
> > @K-ballo have you seen this: https://cdash.cscs.ch/viewBuildError.php?buildid=130585?
> 
> @msimberg I have not seen that one. It looks like all the relevant information has been stripped. Could I get a real build log?

@msimberg should be fine now

msimberg:
> > > @K-ballo have you seen this: https://cdash.cscs.ch/viewBuildError.php?buildid=130585?
> > 
> > 
> > @msimberg I have not seen that one. It looks like all the relevant information has been stripped. Could I get a real build log?
> 
> @msimberg should be fine now

Thanks! And sorry for not getting you the build logs faster...

msimberg (approved):
Thanks @K-ballo! This looks ready to go now (CI is happy). Anything you're still planning to change?

K-ballo:
I'm not planning any more changes.

PR #4994, opened 2020-10-05
Enable permissive- flag on Windows GitHub actions builders

msimberg:


aurianer (approved):
Thanks!

hkaiser:
I believe the problems @McKillroy ran into were caused by one (or both) of the following additional compiler options he used: `/permissive-` and `/Zc:preprocessor`. Should we add those to this builder?

msimberg:
> I believe the problems @McKillroy ran into were caused by one (or both) of the following additional compiler options he used: `/permissive-` and `/Zc:preprocessor` (most likely the former). Should we add those to this builder?

Ah, I didn't realize that, thanks! I've added both flags.

msimberg:
@hkaiser, @K-ballo are the Windows builder errors expected? I originally added both `/Zc:preprocessor` and `/permissive-` and got the same errors. This latest build is with only `/Zc:preprocessor`. Can we make MSVC ignore those errors/warnings for non-HPX files?

hkaiser:
> @hkaiser, @K-ballo are the Windows builder errors expected? I originally added both `/Zc:preprocessor` and `/permissive-` and got the same errors. This latest build is with only `/Zc:preprocessor`. Can we make MSVC ignore those errors/warnings for non-HPX files?

I think the key is to use `/permissive-`. The preprocessor flag is not relevant, even more as it breaks compilation of the windows headers.

msimberg:
@hkaiser, @K-ballo any clue about the current errors? Now there's just `-permissive-` (CMake didn't like `/permissive-` at all). Do you usually use `permissive-` locally?

hkaiser:
> @hkaiser, @K-ballo any clue about the current errors? Now there's just `-permissive-` (CMake didn't like `/permissive-` at all). Do you usually use `permissive-` locally?

None from looking at it briefly. This is very strange. I never used `/permissive-` for HPX locally, but will try.

K-ballo:
> @hkaiser, @K-ballo any clue about the current errors?

Boost issue: https://github.com/boostorg/proto/issues/20

Fixed in 1.72.0

msimberg:
@hkaiser do you think you could update the dependencies we use in the CI? It seems to be on 1.70 right now.

msimberg:
@hkaiser do you think you would have time to update the dependencies sometime between rostam being back in action and the 1.6.0 release?

hkaiser:
> @hkaiser do you think you would have time to update the dependencies sometime between rostam being back in action and the 1.6.0 release?

Will do.

msimberg:
@K-ballo, @hkaiser do the current Windows build errors tell you anything? I'm surprised by the linker error, since I had the impression Boost was ABI compatible between release and debug. This is only in debug mode: https://github.com/STEllAR-GROUP/hpx/pull/4994/checks?check_run_id=1694119642#step:5:1583. This is the other failure in release mode (using C++17): https://github.com/STEllAR-GROUP/hpx/pull/4994/checks?check_run_id=1694119612#step:5:557.

hkaiser:
> @K-ballo, @hkaiser do the current Windows build errors tell you anything? I'm surprised by the linker error, since I had the impression Boost was ABI compatible between release and debug. This is only in debug mode: https://github.com/STEllAR-GROUP/hpx/pull/4994/checks?check_run_id=1694119642#step:5:1583. This is the other failure in release mode (using C++17): https://github.com/STEllAR-GROUP/hpx/pull/4994/checks?check_run_id=1694119612#step:5:557.

I'm surprised by the errors as well. Let me investigate.

msimberg:
I'm closing this as I can't do much to test it. With #5283 we get C++17, but the other flags would still need checking. Please feel free to pick this up if it affects you.

hkaiser:
I have put that back on my stack. I'm definitely interested in getting this fixed.

msimberg:
@hkaiser did the sender/receiver build failures come up while looking at this?

hkaiser:
> @hkaiser did the sender/receiver build failures come up while looking at this?

No, I was trying to reproduce the issues we currently see here: #5441

msimberg:
@hkaiser I've rebased, let's see what it says. Note that since we merged #5283 this PR is of course no longer (only) about enabling C++17, but only about the `permissive-` flag.

hkaiser:
> @hkaiser I've rebased, let's see what it says. Note that since we merged #5283 this PR is of course no longer (only) about enabling C++17, but only about the `permissive-` flag.

It looks like that it is now failing for a different reason: the partitioned_vector actions are defined more than once (I've seen this on Windows before...). So I guess the ball is back in my court.

hkaiser:
This is superseded by #5821 (the target branch for this PR doesn't exist anymore).

PR #4995, opened 2020-10-05
Test all schedulers in cross_pool_injection test except shared_priority_queue_scheduler

msimberg:
I could get reliable hangs with the shared priority scheduler, and none with the other schedulers. This does not mean that the others are fault free, but at least I could not reproduce the issue with them. The other schedulers also don't use exactly the same flags as the shared priority scheduler, so it's not a 100% fair comparison. Time will tell if the test will keep failing.

### DIFF
--- a/cmake/HPX_AddTest.cmake
+++ b/cmake/HPX_AddTest.cmake
@@ -6,7 +6,9 @@
 
 function(add_hpx_test category name)
   set(options FAILURE_EXPECTED RUN_SERIAL)
-  set(one_value_args EXECUTABLE LOCALITIES THREADS_PER_LOCALITY RUNWRAPPER)
+  set(one_value_args EXECUTABLE LOCALITIES THREADS_PER_LOCALITY TIMEOUT
+                     RUNWRAPPER
+  )
   set(multi_value_args ARGS PARCELPORTS)
   cmake_parse_arguments(
     ${name} "${options}" "${one_value_args}" "${multi_value_args}" ${ARGN}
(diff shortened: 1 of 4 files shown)


hkaiser (approved):

hkaiser:
LGTM, thanks!

aurianer (approved):
LGTM thanks!

biddisco:
Fixing this scheduler has been on my list for a while and I will make time for it - it's the one last bug that stops it being a mainstream scheduler.

biddisco (approved):
LGTM

msimberg:
I've had to increase the test timeout on CircleCI because the test now takes a bit longer to complete due to testing more schedulers, but otherwise I haven't seen any hangs on the CI runs so far. @biddisco do you think the test could be made to run faster or would it lose its usefulness in that case? I don't really mind if it takes a minute or two, but if it can be made to run faster without affecting its usefulness we should make it run a bit faster.

hkaiser:
> I've had to increase the test timeout on CircleCI because the test now takes a bit longer to complete due to testing more schedulers, but otherwise I haven't seen any hangs on the CI runs so far. @biddisco do you think the test could be made to run faster or would it lose its usefulness in that case? I don't really mind if it takes a minute or two, but if it can be made to run faster without affecting its usefulness we should make it run a bit faster.

Would splitting up the test such that each scheduler is run in a separate binary help?

msimberg:
> > I've had to increase the test timeout on CircleCI because the test now takes a bit longer to complete due to testing more schedulers, but otherwise I haven't seen any hangs on the CI runs so far. @biddisco do you think the test could be made to run faster or would it lose its usefulness in that case? I don't really mind if it takes a minute or two, but if it can be made to run faster without affecting its usefulness we should make it run a bit faster.
> 
> Would splitting up the test such that each scheduler is run in a separate binary help?

I was mainly thinking about the total time, since most of our tests take a few seconds. To avoid having to increase the timeout for all tests I think I'd rather only increase it for this test in the CircleCI config to avoid the boilerplate required by having separate binaries (or same binary with different scheduler flags).

msimberg:
I've added a `TIMEOUT` option to `add_hpx_test` and increased the timeout only for the `cross_pool_injection` test. Will merge this if CircleCI is happy with this change (seems to work locally).

PR #4996, opened 2020-10-05
Adapt adjacent_find to C++20

msimberg:


libs/parallelism/algorithms/include/hpx/parallel/util/invoke_projected.hpp
```diff
@@ -36,7 +37,47 @@ namespace hpx { namespace parallel { namespace util {
                 pred_, hpx::util::invoke(proj_, std::forward<T>(t)));
         }
 
+        template <typename T>
+        auto operator()(T&& t,
+            T&& u) -> decltype(hpx::util::invoke(std::declval<pred_type>(),
+            hpx::util::invoke(std::declval<proj_type>(), std::forward<T>(t)),
+            hpx::util::invoke(std::declval<proj_type>(), std::forward<T>(u))))
```
hkaiser, line 44: Wouldn't a simple `decltype(auto)` be sufficient? Or do you need the SFINAE induced by the trailing return type calculation?

msimberg: Good catch. No, I don't think the SFINAE is needed.

msimberg:
This should now finally be ready to go.

hkaiser (dismissed):
Thanks for working on this! 
LGTM, thanks!

PR #4998, opened 2020-10-06
Fix deprecation warnings

msimberg:
Fixes more deprecation warnings and enables deprecation warnings on all the Jenkins builders. The main problem was that clang does not emit warnings for template aliases, whereas GCC does, and we only had deprecation warnings turned on with clang. GCC 10 also has a bug where using typedefs in deprecated functions triggers a deprecation warning (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95302) so I've worked around that.


libs/full/compute_cuda/examples/hello_compute.cu
```diff
@@ -48,40 +48,37 @@ int hpx_main(hpx::program_options::variables_map& vm)
     hpx::compute::vector<int, allocator_type> d_C(N, alloc);
 
     hpx::parallel::copy(
-        hpx::parallel::execution::seq,
-        h_A.begin(), h_A.end(), d_A.begin());
+        hpx::parallel::execution::seq, h_A.begin(), h_A.end(), d_A.begin());
```
hkaiser, line 51: Shouldn't this be `hpx::execution::seq`?

libs/full/compute_cuda/tests/unit/default_executor.cu
```diff
@@ -67,7 +67,7 @@ void test_bulk_sync()
 
     hpx::cuda::experimental::target target;
     executor exec(target);
-//    traits::bulk_execute(exec, hpx::util::bind(&bulk_test, _1), v);
+    //    traits::bulk_execute(exec, hpx::util::bind(&bulk_test, _1), v);
```
hkaiser, line 70: Let's remove this while we're at it...

libs/full/compute_cuda/tests/unit/for_loop_compute.cu
```diff
@@ -25,33 +25,26 @@ typedef hpx::cuda::experimental::default_executor executor_type;
...
     // FIXME : Lambda function given to for_loop_n() is momentarily defined as
     //         HPX_HOST_DEVICE in place of HPX_DEVICE to allow the host_side
     //         result_of<> (used inside for_each()) to get the return
     //         type
 
-    hpx::for_loop_n(
-        hpx::parallel::execution::par.on(exec),
-        d_A.data(), d_A.size(),
-        hpx::parallel::induction(d_B.data()),
+    hpx::for_loop_n(hpx::parallel::execution::par.on(exec), d_A.data(),
```
hkaiser, line 36: Same here: `hpx::execution::par`?

libs/full/compute_cuda/examples/hello_compute.cu
msimberg: It should :/ I suppose the flags don't get applied for nvcc... Will check if we can enable them. I'll fix this in any case. Very good catch!

hkaiser (commented):
Thanks for working on this!

libs/full/compute_cuda/examples/hello_compute.cu
msimberg: Actually they'll most likely trigger warnings. I just didn't test this with CUDA enabled.

aurianer (dismissed):
Thanks!

libs/full/compute_cuda/examples/data_copy.cu
```diff
@@ -45,11 +45,11 @@ int hpx_main(hpx::program_options::variables_map& vm)
 
     // copy data from host to device
     hpx::parallel::copy(
```
aurianer, line 47: Isn't this supposed to be `hpx::copy` now?

msimberg: Yep, good point. I still have to do a round of checks on the cuda builds.

msimberg:
I've combined this with #4993 since there was quite some overlap between the two PRs. Closing this one.

PR #4999, opened 2020-10-06
Attempt to fix MacOS Github action error

aurianer:
Add explicit template instanciation to fix undefined symbol when `print_bin` with `unsigned long` use

### DIFF
--- a/libs/core/debugging/src/print.cpp
+++ b/libs/core/debugging/src/print.cpp
@@ -137,6 +137,12 @@ namespace hpx { namespace debug {
 
         template HPX_CORE_EXPORT void print_bin(
             std::ostream&, std::uint64_t, int);
+
+#if defined(__APPLE__)
+        // Explicit instantiation necessary to solve undefined symbol for MacOS
+        template HPX_CORE_EXPORT void print_bin(
+            std::ostream&, unsigned long, int);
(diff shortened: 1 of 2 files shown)


hkaiser (approved):
LGTM, thanks!

libs/core/debugging/src/print.cpp
```diff
@@ -137,6 +137,10 @@ namespace hpx { namespace debug {
 
         template HPX_CORE_EXPORT void print_bin(
             std::ostream&, std::uint64_t, int);
+
+        // Explicit instanciation necessary to solve undefined symbol for MacOS
```
aurianer, line 141: I always forget that it is `instantiation` --'

aurianer:
So in the end, adding the explicit instantiation with `std::uint32_t` ends up with the same error, so I added a specific instantiation with `unsigned long` only for `__APPLE__` platforms, we are now back to the old error with pkgconfig that I'll try to fix

aurianer:
Couldn't find the second error for the pkgconfig tests, I temporarily disable the pkgconfig test for MacOS and will try to fix #5000 in another PR

msimberg (approved):
Thanks @aurianer for looking into this!

PR #5004, opened 2020-10-08
Fixing template specializations that make extra archive data types unique across module boundaries

hkaiser:
- this was broken on Windows, not sure if it was broken on other platforms as well

This fixes a serious issue and if it turns out that this was broken on non-Windows platforms as well then this would be a candidate for another point release.

### DIFF
--- a/hpx/runtime/serialization/detail/preprocess_gid_types.hpp
+++ b/hpx/runtime/serialization/detail/preprocess_gid_types.hpp
@@ -125,12 +125,10 @@ namespace hpx { namespace serialization { namespace detail {
     };
 
     // This is explicitly instantiated to ensure that the id is stable across
-    // shared libraries. MSVC and gcc/clang require different handling of
-    // exported explicitly instantiated templates.
-#if defined(HPX_MSVC)
-    extern template struct extra_archive_data_id_helper<preprocess_gid_types>;
-#else
-    extern template struct HPX_EXPORT
-        extra_archive_data_id_helper<preprocess_gid_types>;
-#endif
+    // shared libraries.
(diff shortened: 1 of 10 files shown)


msimberg:
Thanks @hkaiser! Interesting... how you notice this was broken on Windows? Some of our tests quite obviously broke on Linux when this was wrong so I'd expect them to fail if that was the case. Depends a bit on how you noticed that it was broken though.

This change went in after 1.5.0 so there shouldn't be a need for a patch release, unless you think the original implementation also had a bug.

hkaiser:
> Thanks @hkaiser! Interesting... how did you notice this was broken on Windows? Some of our tests quite obviously broke on Linux when this was wrong so I'd expect them to fail if that was the case. Depends a bit on how you noticed that it was broken though.

A future returned from a remote action was not ready during serialization. Some investigation revealed that the extra archive data handling the futures was not found. 

We did have different implementations for MSVC and gcc/clang, so it could be just a bug affecting MSVC. I hope that the solution now works everywhere (I believe it should).

> This change went in after 1.5.0 so there shouldn't be a need for a patch release, unless you think the original implementation also had a bug.

Ok, good.

hkaiser:
@msimberg also, I'd appreciate it if we merged this PR asap as it breaks Phylanx (on Windows).

msimberg:
> @msimberg also, I'd appreciate it if we merged this PR asap as it breaks Phylanx (on Windows).

Absolutely, let's go ahead with this. That looks like the same place that was failing on Linux as well. It's most likely a question of us not running enough tests on the Windows CI (we only run some examples).

PR #5008, opened 2020-10-11
Replace uses of boost::detail::spinlock

K-ballo:


libs/core/thread_support/include/hpx/thread_support/spinlock.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+        bool try_lock() noexcept
+        {
+            return !m.test_and_set(std::memory_order_acquire);
+        }
+
+        void lock() noexcept
+        {
+            for (unsigned k = 0; !try_lock(); ++k)
+                yield_k(k);
```
hkaiser, line 40: To reduce the pressure on the atomic flag this could be rewritten as:
```
while (!try_lock())
{
    util::yield_while([this] { return is_locked(); }, 
        "hpx::util::detail::spinlock::lock");
}
```

K-ballo: I'm only looking at cutting the dependency on boost's internal implementation details.

If you wish to make those changes feel free to push them to this branch.

hkaiser: I changed you PR slightly to use `std::atomic<bool>` and applied the modification mentioned above.

libs/core/thread_support/include/hpx/thread_support/spinlock.hpp
```diff
@@ -22,7 +23,7 @@ namespace hpx { namespace util { namespace detail {
         HPX_NON_COPYABLE(spinlock);
 
     private:
-        std::atomic_flag m = ATOMIC_FLAG_INIT;
+        std::atomic<bool> m = false;
 
         HPX_CORE_EXPORT void yield_k(unsigned) noexcept;
```
K-ballo, line 28: This is no longer needed.

hkaiser: We might still need it after all as my commit created a circular dependency.  It might not be worth applying the optimization, after all...

K-ballo: I peeked at that `yield_while` implementation, and it seems to be based in an older incarnation of `boost::smart_ptr::yield_k`. The one in this PR was based on the current `boost::smart_ptr::yield_k` implementation.

hkaiser: Sure. The point of my commit was not replacing the `yield_k` you added (I'd be fine with changing our existing implementation to match what you proposed). The change I would like to see is to reduce the pressure on the atomic itself by not constantly trying to write to the atomic flag, but instead have a separate loop that just reads from it before trying to write again.

hkaiser (approved):
Thanks a lot!

msimberg (changes requested):
Just the cmake formatting left... https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/5787/workflows/3f0c084b-e86a-468e-a000-2f655d6260b1/jobs/226053/artifacts. Thanks for doing this @K-ballo! Looks very good.

PR #5009, opened 2020-10-11
Replace boost::system::error_code with std::error_code

K-ballo:
Since Boost 1.65.0, `boost::system::error_code` is implicitly convertible to `std::error_code`. This takes care of compatibility when the error_code is taken by value or const ref; functions taking the error_code by mutable ref need to be manually mapped.

### DIFF
--- a/libs/core/errors/src/exception.cpp
+++ b/libs/core/errors/src/exception.cpp
@@ -5,6 +5,9 @@
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
+// hpxinspect:nodeprecatedinclude:boost/system/system_error.hpp
+// hpxinspect:nodeprecatedname:boost::system::system_error
+
 #include <hpx/config.hpp>
 #include <hpx/assert.hpp>
 #include <hpx/errors/error.hpp>
--- a/libs/core/errors/src/exception.cpp
+++ b/libs/core/errors/src/exception.cpp
@@ -20,6 +20,8 @@
 #include <unistd.h>
 #endif
 
+#include <boost/system/system_error.hpp>
+
 #include <algorithm>
 #include <atomic>
 #include <cstddef>
@@ -30,6 +32,7 @@
 #include <sstream>
 #include <stdexcept>
 #include <string>
+#include <system_error>
 #include <utility>
 #include <vector>
 
@@ -40,7 +43,7 @@ namespace hpx {
     /// \param e    The parameter \p e holds the hpx::error code the new
     ///             exception should encapsulate.
     exception::exception(error e)
-      : boost::system::system_error(make_error_code(e, plain))
+      : std::system_error(make_error_code(e, plain))
     {
         HPX_ASSERT((e >= success && e < last_error) || (e & system_error_flag));
         if (e != success)
@@ -50,16 +53,16 @@ namespace hpx {
     }
 
     /// Construct a hpx::exception from a boost#system_error.
-    exception::exception(boost::system::system_error const& e)
-      : boost::system::system_error(e)
+    exception::exception(std::system_error const& e)
+      : std::system_error(e)
     {
         LERR_(error) << "created exception: " << this->what();
     }
 
     /// Construct a hpx::exception from a boost#system#error_code (this is
     /// new for Boost V1.69).
-    exception::exception(boost::system::error_code const& e)
-      : boost::system::system_error(e)
+    exception::exception(std::error_code const& e)
+      : std::system_error(e)
     {
         LERR_(error) << "created exception: " << this->what();
     }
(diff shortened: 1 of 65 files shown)


K-ballo:
> > Do we need/want to retain boost serialization (possibly retain but guard by `HPX_SERIALIZATION_WITH_BOOST_TYPES`)?
> 
> I thought we had gotten rid of Boost serialization a long time ago. Anything still in the code base must be a leftover.

I meant serialization of `boost::system::system_error`.

It's unclear to me what the goal behind that support is, but it has been special cased alongside standard exception types, and it isn't guarded by `HPX_SERIALIZATION_WITH_BOOST_TYPES`, so it must have some foundational value. The question is whether that needs to be migrated to `std::system_error`, extended to `std::system_error`, or left as-is.

We are not explicitly throwing `system_error` ourselves (`hpx::exception` extends it, and it has migrated form `boost` to `std`, but that type is also special-cased). These exceptions may be originating from within Boost, likely ASIO (and possibly Filesystem too, before the switch). If that's the scenario the serialization of `boost::system::system_error` caters to then it needs to stay in its current form (ideally eventually wrapped in `hpx::exception` if those originate from core). That would suggest that support ought to have been extended to `std::system_error` already, but I'm not entirely sure.

hkaiser:
> It's unclear to me what the goal behind that support is, but it has been special cased alongside standard exception types, and it isn't guarded by `HPX_SERIALIZATION_WITH_BOOST_TYPES`, so it must have some foundational value. The question is whether that needs to be migrated to `std::system_error`, extended to `std::system_error`, or left as-is.

The safest way would be to leave it in and guard it with `HPX_SERIALIZATION_WITH_BOOST_TYPES`. As you said, the only reason it might be needed would be Boost.Asio and/or Boost/Filesystem.

msimberg:
> > It's unclear to me what the goal behind that support is, but it has been special cased alongside standard exception types, and it isn't guarded by `HPX_SERIALIZATION_WITH_BOOST_TYPES`, so it must have some foundational value. The question is whether that needs to be migrated to `std::system_error`, extended to `std::system_error`, or left as-is.
> 
> The safest way would be to leave it in and guard it with `HPX_SERIALIZATION_WITH_BOOST_TYPES`. As you said, the only reason it might be needed would be Boost.Asio and/or Boost/Filesystem.

I suppose the question is if we expect those exceptions to be thrown and then serialized from within HPX or if those are just there for convenience for users encountering boost exceptions. If the former I suppose they should not be guarded, but I'm not sure in what situation that would happen. Would e.g. parcel pool ASIO exceptions from a remote locality end up being serialized to the main locality? Writing that I realize that even if they don't they probably should. If it's only the latter, i.e. for exceptions from user code then guarding with that ifdef sounds appropriate.

K-ballo:
> Would e.g. parcel pool ASIO exceptions from a remote locality end up being serialized to the main locality?

That is indeed the question.

> Writing that I realize that even if they don't they probably should.

I think those should be transported as HPX exceptions (if they aren't already). The user needs not know to expect ASIO errors.

hkaiser (changes requested):
I believe this can be removed now as well: https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_SetupBoost.cmake#L58-L62

K-ballo:
> I believe this can be removed now as well: https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_SetupBoost.cmake#L58-L62

We still need to link to Boost.System if ASIO and/or Filesystem internally use it. Filesystem may already be handled at the cmake level, but ASIO doesn't exist as an imported target so it wouldn't pull the usage requirement.

hkaiser (approved):
LGTM, thanks!

libs/core/errors/src/exception.cpp
```diff
@@ -329,8 +333,7 @@ namespace hpx { namespace detail {
         hpx::exception const&, std::string const&, std::string const&, long);
 
     template HPX_CORE_EXPORT void throw_exception(
-        boost::system::system_error const&, std::string const&,
-        std::string const&, long);
+        std::system_error const&, std::string const&, std::string const&, long);
```
msimberg, line 336: I think you'll need to add an explicit template instantiation of `get_exception` above for `std::system_error` as well (or replace the one with `boost::system::system_error`?).

msimberg:
The build failure with `syscall` is quite mysterious... I'll see if I can figure out what's wrong there.

msimberg:
@K-ballo the `thread_mapper.cpp` file is missing an `#include <unistd.h>` next to the syscall include (https://github.com/STEllAR-GROUP/hpx/blob/5c06192914e787230f35be7ccd55b496f6f46ffa/libs/full/runtime_local/src/thread_mapper.cpp#L26). I suppose this must have come as a transitive include before these changes. I think you can include it unconditionally (we have a feature check for it) since the syscall include is already guarded by `__linux__` which we know is POSIX.

msimberg:
Thanks @K-ballo! The remaining test failure is definitely unrelated to these changes.

PR #5011, opened 2020-10-12
Inspect command line tweaks

K-ballo:
- Handle exceptions thrown from inspect's main
- Store parsed options into associated variables
- Change inspect check arguments into flags


hkaiser (approved):
Thanks for looking into this!

PR #5012, opened 2020-10-13
Properly reset pointer tracking data in output_archive

hkaiser:
- This fixes an issue that is caused by not cleaning out the pointer
  tracking data in output archives during serialization preprocessing
  while repeatedly serializing the data because of futures not being
  ready

### DIFF
--- a/hpx/runtime/serialization/detail/preprocess_gid_types.hpp
+++ b/hpx/runtime/serialization/detail/preprocess_gid_types.hpp
@@ -127,8 +127,9 @@ namespace hpx { namespace serialization { namespace detail {
     // This is explicitly instantiated to ensure that the id is stable across
     // shared libraries.
     template <>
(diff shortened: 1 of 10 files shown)


biddisco:
This fixes the serialization bug. Thank you very much. The size is now correctly computed in the final pass and matches the preprocessed 'estimate'.

I do still have one bug where the final size does not match the `preprocess_archive` estimate, but as it turns out, the final size is always (so far) **smaller** than the estimate, so I can for now ignore it and use the slightly larger estimated size when it happens (about once every N thousand parcels so far, so it is not a big deal). The difference is only 4 bytes or something AFAICT, so it isn't a problem. I will do some larger runs on daint with this fix and see if things are stable.

biddisco (approved):
This is good. Please merge if and when it passes the normal checks.

PR #5018, opened 2020-10-15
Update CSCS jenkins configuration to clang 11

msimberg:


libs/full/compute/include/hpx/compute/host/block_allocator.hpp
```diff
@@ -92,8 +92,7 @@ namespace hpx { namespace compute { namespace host {
             // topo.allocate(). The pointer hint may be used to provide locality of
             // reference: the allocator, if supported by the implementation, will
             // attempt to allocate the new memory block as close as possible to hint.
-            pointer allocate(
-                size_type n, std::allocator<void>::const_pointer hint = nullptr)
+            pointer allocate(size_type n, const void* hint = nullptr)
```
hkaiser, line 95: Should we change this to use `typename std::allocator_traits<...this_type...>::const_void_pointer` instead?

msimberg: I can do that. However, I made it a plain `const void*` since that's what the (deprecated) `allocate` in the standard library takes. Is there any case where `std::allocator_traits<T>::const_void_pointer` wouldn't be `const void*` or is there any value in making it that?

msimberg: @hkaiser would you still like to see this changed or are you ok with this as it is?

hkaiser: I think in our case using a plain pointer is good enough. Thanks!

Issue #5014, opened 2020-10-14
Rename all .cc to .cpp and .hh to .hpp


hkaiser:
For historical reasons some of our examples use their own file naming conventions. We should unify that.

msimberg:
Closed by #5022.

PR #5022, opened 2020-10-19
Remove C language from CMakeLists.txt

msimberg:
Doing a quick test to see if we actually need to specify C as the project language. I have the impression that we might be able to do without it but I've probably missed something... Most likely related to https://github.com/STEllAR-GROUP/hpx/issues/5014. If it's only the files in #5014 making us use C, I'll rename those file at the same time (I'll do it in any case).


msimberg:
This seems to work fine in the normal case, and saves a few CMake compiler flag checks during configuration. What I haven't checked is the fortran test as the compiler flag checks seem to use the C compiler instead of the fortran compiler (and then assumes they have the same flags available). The question is: do we see any value in keeping the fortran test?

msimberg:
I've taken the liberty of removing the fortran example that we have. With that all the support for C and fortran can be removed.

Please speak up if someone thinks it would be important to keep and revive that example.

CMakeLists.txt
```diff
@@ -1978,9 +1955,9 @@ endif()
 # Configure Warnings
 if(HPX_WITH_COMPILER_WARNINGS)
   if(MSVC) # Adding special warning settings for the MSVC compiler ...
-    hpx_add_compile_flag(-W3 LANGUAGES C CXX)
+    hpx_add_compile_flag(-W3 LANGUAGES CXX)
     # According to the ifort Windows manual, W3 isn't supported
-    hpx_add_compile_flag(-W1 LANGUAGES Fortran)
+    hpx_add_compile_flag(-W1 LANGUAGES)
```
hkaiser, line 1960: These two lines can go

examples/sheneos/sheneos_compare.cpp
```diff
@@ -24,7 +24,7 @@
 #include <vector>
 
 #include "sheneos/interpolator.hpp"
-#include "fname.h"
+#include "fname.hpp"
```
hkaiser, line 27: This shouldn't be needed anymore as it defines interface-compatibility of external symbols with fortran.

aurianer (dismissed):
This [cmake/TargetArch.cmake](https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/TargetArch.cmake) file also have things to change I believe :)
I also noticed some C sources variables in the libfabric parcelport when I was doing the cmake refactoring, you might want to have a look there too
Thanks for doing this!

examples/sheneos/sheneos_compare.cpp
msimberg: Good catch. This whole file is actually also unnecessary as it only serves as a shim to the fortran functionality. I've removed both.

CMakeLists.txt
msimberg: I've at the same time removed the `LANGUAGES CXX` arguments since that's the default. I've kept the option though, in case we need it in the future again.

msimberg:
> This [cmake/TargetArch.cmake](https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/TargetArch.cmake) file also have things to change I believe :)
> I also noticed some C sources variables in the libfabric parcelport when I was doing the cmake refactoring, you might want to have a look there too
> Thanks for doing this!

Thanks, that's a very good catch! The `TargetArch` one is straightforward enough, we can just use C++ there instead. I'll have to check the libfabric parts. @biddisco if you have time to comment on if we can get by with pure C++ there that'd be helpful, but please don't feel forced to reply.

cmake/TargetArch.cmake
```diff
@@ -113,9 +113,7 @@ function(target_architecture output_var)
       list(APPEND ARCH ppc64)
     endif()
   else()
-    file(WRITE "${PROJECT_BINARY_DIR}/arch.c" "${archdetect_c_code}")
-
-    enable_language(C)
+    file(WRITE "${PROJECT_BINARY_DIR}/arch.cpp" "${archdetect_cpp_code}")
 
     # Detect the architecture in a rather creative way... This compiles a small
     # C program which is a series of ifdefs that selects a particular #error
```
hkaiser, line 119:
```suggestion
    # C++ program which is a series of ifdefs that selects a particular #error
```

msimberg:
@hkaiser I see that the parts of the libfabric CMake configuration that reference C files are only enabled if `HPX_PARCELPORT_LIBFABRIC_WITH_HPXC` is on, and my impression is that hpxc is... dead? Do you think it should be kept around or can we clean that up?

msimberg:
> @hkaiser I see that the parts of the libfabric CMake configuration that reference C files are only enabled if `HPX_PARCELPORT_LIBFABRIC_WITH_HPXC` is on, and my impression is that hpxc is... dead? Do you think it should be kept around or can we clean that up?

I've tentatively removed the branches with hpxc enabled.

msimberg:
Given that the removal of the C parts in the libfabric config are ok, this should be ready to go now.

aurianer (approved):
LGTM thanks!

PR #5024, opened 2020-10-22
Moving naming_base to full modules

hkaiser:
- moving `gid_type` to `naming_base`
- moving credit handling to `naming`
- flyby: remove implicit dependency of `runtime_local` on `naming_base`


hkaiser:
@msimberg I think this is good to go now. There is a lot more we can do in this direction, but I'm trying to take small steps towards disentangling things.

libs/parallelism/executors/include/hpx/executors/dataflow.hpp
```diff
@@ -26,11 +26,13 @@
...
 #include <hpx/futures/traits/is_future.hpp>
 #include <hpx/modules/memory.hpp>
-#include <hpx/modules/naming.hpp>
 #include <hpx/pack_traversal/pack_traversal_async.hpp>
 #include <hpx/threading_base/annotated_function.hpp>
 #include <hpx/threading_base/thread_num_tss.hpp>
 #include <hpx/type_support/always_void.hpp>
+#if defined(HPX_HAVE_DISTRIBUTED_RUNTIME)
+#include <hpx/modules/naming.hpp>
+#endif
```
msimberg, line 35: I know this dependency was already there, but is there something we can do about it already now?

hkaiser: Let's leave this alone for now as it will involve (major) unrelated restructurings.

msimberg:
Thanks @hkaiser, let's go ahead with this as it is!

PR #5025, opened 2020-10-23
Adapt is_sorted and is_sorted_until to C++20

msimberg:


libs/parallelism/algorithms/include/hpx/parallel/algorithms/adjacent_find.hpp
```diff
@@ -159,7 +159,8 @@ namespace hpx { namespace parallel { inline namespace v1 {
                 ExPolicy, InIter first, InIter last, Pred&& pred, Proj&& proj)
             {
                 return std::adjacent_find(first, last,
-                    util::invoke_projected<Pred, Proj>(pred, proj));
+                    util::invoke_projected<Pred, Proj>(
+                        std::forward<Pred>(pred), std::forward<Proj>(proj)));
```
hkaiser, line 163: Excellent catch!

libs/parallelism/algorithms/include/hpx/parallel/algorithms/is_sorted.hpp
```diff
@@ -42,17 +257,19 @@ namespace hpx { namespace parallel { inline namespace v1 {
             {
             }
 
-            template <typename ExPolicy, typename Pred>
+            template <typename ExPolicy, typename Pred, typename Proj>
             static bool sequential(
-                ExPolicy, FwdIter first, FwdIter last, Pred&& pred)
+                ExPolicy, FwdIter first, FwdIter last, Pred&& pred, Proj&& proj)
```
hkaiser, line 262: Do you still plan to support iterator/sentinel pairs?

msimberg: Yes, very good catch here as well! I just forgot to update these.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/is_sorted.hpp
```diff
@@ -34,25 +249,28 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
-            template <typename ExPolicy, typename Pred>
+            template <typename ExPolicy, typename Pred, typename Proj>
             static bool sequential(
-                ExPolicy, FwdIter first, FwdIter last, Pred&& pred)
+                ExPolicy, FwdIter first, Sent last, Pred&& pred, Proj&& proj)
             {
-                return std::is_sorted(first, last, std::forward<Pred>(pred));
+                return is_sorted_sequential(first, last,
+                    util::invoke_projected<Pred, Proj>(
+                        std::forward<Pred>(pred), std::forward<Proj>(proj)));
```
hkaiser, line 267: Now, that you have your own implementations of `is_sorted`, may I suggest to directly pass the projection to that implementation instead of having to create another indirection layer (the `invoke_projected`)?

libs/parallelism/algorithms/include/hpx/parallel/algorithms/detail/is_sorted.hpp
```diff
@@ -13,15 +14,13 @@
 
 namespace hpx { namespace parallel { inline namespace v1 { namespace detail {
 
-    // provide implementation of std::distance supporting iterators/sentinels
-    // std::is_sorted is not available on all supported platforms yet
     template <typename Iter, typename Sent, typename Compare>
-    inline bool is_sorted_sequential(Iter first, Sent last, Compare const& comp)
+    inline bool is_sorted_sequential(Iter first, Sent last, Compare&& comp)
```
hkaiser, line 18: I think this could be marked as `constexpr`.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/is_sorted.hpp
msimberg: I've updated the base implementation to support this now as well.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/detail/is_sorted.hpp
msimberg: Yep, good idea!

libs/parallelism/algorithms/include/hpx/parallel/algorithms/is_sorted.hpp
msimberg: Also a good idea! I've changed that as well.

hkaiser (approved):
LGTM, thanks!

PR #5032, opened 2020-10-29
Do codespell comparison only on files changed from common ancestor

msimberg:
Only checks files changed from the common ancestor commit, instead of the tip of master, which may have unrelated changes if the PR is not recently rebased. `origin/master..` is shorthand for `origin/master..HEAD`, but `origin/master...` (three dots) is shorthand for `<common base of origin/master and HEAD>..HEAD`. This is still not 100% right, because master isn't always the merge base, but CircleCI does not give us an environment variable with the base branch (I also don't see it being worth the effort to get that manually right now).


aurianer (approved):
Thanks :ok_hand:

PR #5033, opened 2020-10-29
Implement P2220 properties module

msimberg:
Adds a `properties` module which provides the `hpx/property.hpp` header, containing the customization point:

- `hpx::experimental::prefer`

Compared to P1393 this only has a single generic customization point. Actual property customization points should be defined by the modules that have properties. I'm keeping the P1393 implementation in the history in case we need it in the future. P2220 for example lacks a nice way to require multiple properties in one go, so I expect the final properties module to be some sort of compromise between P1393 and P2220.

To get things started I've made `thread_schedule_hint` a property for `parallel_policy_executor` so it's possible to do `hpx::experimental::make_with_hint(parallel_executor{}, hpx::threads::thread_schedule_hint{42})` to get an executor with the hint `42`.

Flyby: remove a leftover modules file.


hkaiser:
> One thing that I've had to add is another `tag_invoke` helper struct, `tag_invoke_override`. It's not clear to me if `tag_invoke_fallback` is in fact supposed to be doing exactly what `tag_invoke_override` is doing. P1393 says that a member function `x.require(p)` (same for `prefer` etc.) should have precedence over a `require(x, p)`. This is not 100% the same since we're using `tag_invoke`, but I'm putting a `tag_invoke` overload in the place of a `require(x, p)` overload. `tag_invoke_fallback` does this backwards, i.e. `tag_invoke` has precedence over a member function. P0443 says the same about the customization points defined there, so I'm guessing `tag_invoke_fallback` should maybe just be what `tag_invoke_override` is right now? And in that case the customization points in `execution_base` should also be corrected. @hkaiser?

I wouldn't be surprised if I got it wrong in the first place. It would be definitely beneficial if we were able to unify `tag_invoke_fallback` and `tag_invoke_override`.

msimberg:
@hkaiser this has been updated to be closer to P2220, which in practice means leaving a single CPO, `prefer` in the module, since the rest of the CPOs in P2220 are CPOs related to executor properties which will come separately.

I've also removed the `tag_override_invoke` version for now. I'll see what's required by the execution CPOs and update `tag_invoke` as needed.

libs/parallelism/executors/include/hpx/executors/parallel_executor.hpp
```diff
@@ -165,6 +166,23 @@ namespace hpx { namespace execution {
...
+            return exec_with_hint;
+        }
+
+        friend hpx::threads::thread_schedule_hint tag_invoke(
+            hpx::execution::experimental::get_hint_t,
+            parallel_policy_executor const& exec)
+        {
+            return exec.schedulehint_;
+        }
+
```
hkaiser, line 185: Should we add overloads for arbitrary executors. Those obviously shouldn't do anything. It would make things generically work for all executors, however.

msimberg: Isn't that what `prefer` is for? I agree we should expand the support, but not with overloads that don't do anything.

msimberg:
@hkaiser can we go ahead with this?

hkaiser:
> @hkaiser can we go ahead with this?

Absolutely, please go ahead.

PR #5034, opened 2020-10-29
Use versioned container image on CircleCI

msimberg:
I've updated the base image for CI to Ubuntu 20.04 (20.10 is still a bit too recent, and not all packages worked there). At the same time I started tagging the pushed images with a simple incrementing number. That means that we can update the image without having all CI builds use the new image immediately. A PR on the HPX side (like this one) is required to actually update to the new version. The versioning is completely manual for the moment, as e.g. installing a single new package in the image doesn't necessarily require a new version. We can improve this later if needed. The `latest` tag always refers to the latest image.

With the move to Ubuntu 20.04 clang is updated to 10, but I've kept clang-format on 9 for now. Ubuntu helpfully provides a few versions of clang-format so we can in the future update the base image a bit more aggressively without changing the clang-format version.

@K-ballo this new tag also includes IWYU, so if you'd like to try things out on CircleCI you can base your IWYU work on this branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,8 +153,8 @@ jobs:
           name: Check that the modules are well clang-formatted
           command: |
               cd /hpx/source/libs && shopt -s globstar # to activate the ** globbing
-              clang-format --version
-              clang-format -i **/*.{cpp,hpp}
+              clang-format-9 --version
+              clang-format-9 -i **/*.{cpp,hpp}
               git diff --exit-code > /tmp/modified_clang_format_files.txt
       - store_artifacts:
           path: /tmp/modified_clang_format_files.txt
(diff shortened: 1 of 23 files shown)


msimberg:
The cmake-format step now fails. It seems like we have the same problem with cmake-format as we do with clang-format, i.e. that formatting can change between versions even with the same config. We can easily pin the version to an older version in the image, but if everyone is happy to go with the newer version we should do that. @hkaiser, @aurianer, @K-ballo, @biddisco?

hkaiser:
> We can easily pin the version to an older version in the image, but if everyone is happy to go with the newer version we should do that.

Let's go with the newest version. Things do not seem to change too often.

msimberg:
This is ready to go now.

aurianer (approved):
LGTM thanks!

PR #5037, opened 2020-10-30
Add is_invocable typedefs to top-level hpx namespace and public API list

msimberg:


hkaiser (commented):
Shouldn't we do this the other way around, i.e. move the implementation into `namespace hpx` and add a (deprecated) using statement into `namespace hpx::traits`?

msimberg:
> Shouldn't we do this the other way around, i.e. move the implementation into `namespace hpx` and add a (deprecated) using statement into `namespace hpx::traits`?

Yeah, I was being lazy... I'll update this to do that instead.

msimberg:
This should now be ready to go.

hkaiser (approved):
LGTM, thanks!

aurianer (approved):
LGTM thanks!

Issue #5035, opened 2020-10-30
Build with HPX_WITH_EXAMPLES fails  


scheherzade:
HPX build falis when ```HPX_WITH_EXAMPLES``` is on.

## Actual Behavior
This is the error message I get:
```
Scanning dependencies of target version
[ 65%] Building CXX object libs/full/resiliency/examples/CMakeFiles/dataflow_replicate.dir/dataflow_replicate.cpp.o
[ 65%] Building CXX object libs/full/resiliency/examples/CMakeFiles/version.dir/version.cpp.o
[ 65%] Linking CXX executable ../../../../bin/custom_syntax
[ 65%] Linking CXX executable ../../../../bin/env_options
/usr/bin/ld: ../../../../lib/libhpx.so.1.6.0: undefined reference to `bool hpx::detail::apply_impl<hpx::lcos::base_lco_with_value<hpx::naming::id_type, hpx::naming::gid_type, hpx::traits::detail::component_tag>::set_value_action, hpx::naming::gid_type>(hpx::naming::id_type const&, hpx::naming::address&&, hpx::threads::thread_priority, hpx::naming::gid_type&&)'
/usr/bin/ld: ../../../../lib/libhpx.so.1.6.0: undefined reference to `bool hpx::detail::apply_impl<hpx::lcos::base_lco_with_value<hpx::naming::id_type, hpx::naming::gid_type, hpx::traits::detail::managed_component_tag>::set_value_action, hpx::naming::gid_type>(hpx::naming::id_type const&, hpx::naming::address&&, hpx::threads::thread_priority, hpx::naming::gid_type&&)'
collect2: error: ld returned 1 exit status
gmake[2]: *** [libs/full/program_options/examples/CMakeFiles/custom_syntax.dir/build.make:114: bin/custom_syntax] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:42348: libs/full/program_options/examples/CMakeFiles/custom_syntax.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
[ 65%] Linking CXX executable ../../../../bin/real
[...]

PR #5039, opened 2020-10-31
Fix GCC 10 release builds

msimberg:
"Fixes" #5035. I know this isn't the way it should be since this leads to circular includes (which nonetheless work). The first change, `apply_implementations_fwd.hpp` to `apply_implementations.hpp` seems legit, as the `apply_impl` definitions are needed by the corresponding file (there are also other files that use `apply_impl` which should probably include the non-forwarding header). With this include the first error is gone, and what remains is undefined references to `apply_l_p` and `apply_r_p` which the second include "solves". I'm guessing maybe the proper fix would be to move some of the definitions from `apply.hpp` to `detail/apply_implementations.hpp`? The `apply_x_p` declarations are in `detail/apply_implementations_fwd.hpp`... @hkaiser let me know if you can take it from here or if I should try something else out.


hkaiser:
FWIW, this compiles for me.

msimberg:
> FWIW, this compiles for me.

Right, it does for me too ;) If you're not worried about the circular includes for the moment (they are at least within a module) we can go ahead with this.

hkaiser:
@msimberg can we merge this? I think it solves the issue for us.

aurianer (dismissed):
Thanks!

aurianer:
`test.headers` and `clang-format` CI failing though

msimberg:
> @msimberg can we merge this? I think it solves the issue for us.

> `test.headers` and `clang-format` CI failing though

I think that makes it a no... The header tests are most likely failing because of the circular includes.

msimberg:
@hkaiser I'd appreciate if you could look at this. My attempts at moving things around lead to the same problems with the header tests and I don't know them well enough to be able to say where things are going wrong. As for the original linker errors I think the main problem is simply that the full template function definitions should come before their use, and not just the forward declarations. It seems like some compilers are less strict about this though...

hkaiser:
@msimberg I believe this is ready for review now.

msimberg (commented):
It's not very pretty, but it does work, so let's go ahead with this. Thanks @hkaiser for doing this! (I can't approve this since I opened the PR.)

PR #5041, opened 2020-11-01
Adding Francisco Tapia's implementation of partial_sort

hkaiser:
@fjtapia thank you very much for contributing!

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/algorithms/partial_sort.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/algorithms/partial_sort.hpp
@@ -1,0 +1,537 @@
+//  Copyright (c) 2020 Francisco Jose Tapia
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+/// \file parallel/algorithms/partial_sort.hpp
+
+#pragma once
+
+#if defined(DOXYGEN)
+namespace hpx {
+
+    /// Places the first middle - first elements from the range [first, last)
+    /// as sorted with respect to comp into the range [first, middle). The rest
+    /// of the elements in the range [middle, last) are placed in an unspecified
+    /// order.
+    ///
+    /// \note   Complexity: Approximately (last - first) * log(middle - first)
+    ///         comparisons.
+    ///
+    /// \tparam ExPolicy    The type of the execution policy to use (deduced).
+    ///                     It describes the manner in which the execution
+    ///                     of the algorithm may be parallelized and the manner
+    ///                     in which it applies user-provided function objects.
+    /// \tparam RandIter    The type of the source begin, middle, and end
+    ///                     iterators used (deduced). This iterator type must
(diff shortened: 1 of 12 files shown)


libs/parallelism/algorithms/include/hpx/parallel/algorithms/sort.hpp
```diff
@@ -53,18 +54,20 @@ namespace hpx { namespace parallel { inline namespace v1 {
         /// \param iter_1 : iterator to the first value
         /// \param iter_2 : iterator to the second value
         /// \param iter_3 : iterator to the third value
-        /// \param comp : object for to compare two values
+        /// \param comp : object for to Comp two values
```
msimberg, line 57:
```suggestion
        /// \param comp : object for comparing two values
```

msimberg (approved):
Thanks, looks good to me!

libs/parallelism/algorithms/include/hpx/parallel/algorithms/partial_sort.hpp
```diff
@@ -0,0 +1,537 @@
...
+                num = (num >> 4);
+            }
+            if (num >= (1ull << 2))
+            {
+                nb += 2;
+                num = (num >> 2);
+            }
+            if (num >= (1ull << 1))
+            {
+                nb += 12;
```
aurianer, line 132: Shouldn't this be `1`?

aurianer (commented):
Sorry always late :woman_facepalming:

libs/parallelism/algorithms/include/hpx/parallel/algorithms/partial_sort.hpp
msimberg: That's a very good catch! It sure looks like it.

hkaiser: Will fix (straight to master). Thanks!

fjtapia:
It is an error, must add 1.
Sorry

fjtapia:
Hi Hartmut,
There is an error in the implementation of partial_sort_copy. The two first
iterators, the input range, are input iterator. With them only can
dereference and increment. The other two, the output range, are random
access.
If the input range would be random access iterators, the single and
parallel process can be done in place, without additional memory.
But in the specification talks about to remain unmodified the input range.
Then we have two scenarios
1.- When the input range is equal or small than the output range. We can
copy the input range in the output range and sort, in a single or parallel
mode.
2.- When the input range is greater than the output range, it needs an
auxiliary memory to copy the elements from the input range, and do a
partial_sort, single or parallel, and copy the results in the output range.

I am working on the modification of the code. In a few days, I will have
ready the new code.

About the fail in the function nbits, it is a consequence of a hurry.  I
use a meta programmed function, but is bigger, and wrote this function and,
against my custom. I didn't check and fail. My arrogance was defeated by
reality. I am sorry

Francisco

hkaiser:
> It is an error, must add 1. Sorry

No problem, @aurianer caught it.

hkaiser:
@fjtapia no worries, take your time. We have not added `partial_sort_copy` and `nth_element` to HPX yet. This PR adds `partial_sort` only.

fjtapia:
THANKS

Issue #4702, opened 2020-06-03
Prefer enum class over unscoped enums


msimberg:
We should gradually move all unscoped enums to enum classes.

Some of those listed below could possibly be hidden in some detail namespace...

- [x] `thread_state_enum`
- [x] `thread_state_ex_enum`
- [x] `thread_stacksize`
- [x] `thread_priority`
- [x] `scheduler_mode` (#5809)
- [x] `scheduler_state` (#5809)
- [x] `component_enum_type`
- [x] `id_type::management_type` (#5817)
- [x] `cache::statistics::method` (#5816)
- [x] `commandline_error_mode`
- [x] `hpx::error`
- [x] `hpx::throwmode` (#5820)
- [x] `future_state`
- [x] `performance_counters::counter_type` (#5850)
- [x] `performance_counters::counter_status`
- [x] `performance_counters::discover_counters_mode`
- [x] `agas::service_mode`
- [x] `serialization::archive_flags`
- [x] `serialization::chunk_type`
- [x] `util::exception_type`
- [ ] `hpx_hwloc_membind_policy`
- [x] `message_buffer_append_state` (parcel coalescing) (#6923)
- [x] `distribution_type` (affinity)
- [x] `deque_status_type` (boost lockfree deque)
- [x] `context_state`, `context_exit_state`, `context_exit_status` (coroutines)
- [x] `logging_destination` (logging)
- [x] `partitioner_mode`

… there are more, please add to this list, if needed.

gonidelis:
@hkaiser checked `throwmode`

PR #5046, opened 2020-11-06
Convert thread and future enums to class enums

msimberg:
Partially addresses #4702.

### DIFF
--- a/components/iostreams/src/server/output_stream.cpp
+++ b/components/iostreams/src/server/output_stream.cpp
@@ -80,7 +80,8 @@ namespace hpx { namespace iostreams { namespace server
         pending_output_.output(locality_id, count, in, write_f, mtx_);
 
         // Wake up caller.
-        threads::set_thread_state(caller, threads::thread_state_enum::pending);
+        threads::set_thread_state(
+            caller, threads::thread_schedule_state::pending);
     }
 
     void output_stream::write_sync(std::uint32_t locality_id,
@@ -93,6 +94,7 @@ namespace hpx { namespace iostreams { namespace server
                 locality_id, count, std::ref(in), threads::get_self_id()));
 
         // Sleep until the worker thread wakes us up.
-        this_thread::suspend(threads::thread_state_enum::suspended, "output_stream::write_sync");
+        this_thread::suspend(threads::thread_schedule_state::suspended,
+            "output_stream::write_sync");
     } // }}}
 }}}
--- a/components/iostreams/src/server/output_stream.cpp
+++ b/components/iostreams/src/server/output_stream.cpp
@@ -80,7 +80,7 @@ namespace hpx { namespace iostreams { namespace server
         pending_output_.output(locality_id, count, in, write_f, mtx_);
 
         // Wake up caller.
-        threads::set_thread_state(caller, threads::pending);
+        threads::set_thread_state(caller, threads::thread_state_enum::pending);
     }
 
     void output_stream::write_sync(std::uint32_t locality_id,
@@ -93,6 +93,6 @@ namespace hpx { namespace iostreams { namespace server
                 locality_id, count, std::ref(in), threads::get_self_id()));
 
         // Sleep until the worker thread wakes us up.
-        this_thread::suspend(threads::suspended, "output_stream::write_sync");
+        this_thread::suspend(threads::thread_state_enum::suspended, "output_stream::write_sync");
     } // }}}
 }}}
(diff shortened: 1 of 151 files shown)


components/iostreams/src/server/output_stream.cpp
```diff
@@ -81,7 +81,7 @@ namespace hpx { namespace iostreams { namespace server
         pending_output_.output(locality_id, count, in, write_f, mtx_);
 
         // Wake up caller.
-        threads::set_thread_state(caller, threads::pending);
+        threads::set_thread_state(caller, threads::thread_state_enum::pending);
```
hkaiser, line 84: Would it be appropriate to use this opportunity to rename the enum to `hpx::threads::thread_state`?

hkaiser (dismissed):
LGTM, thanks for looking into this!

components/iostreams/src/server/output_stream.cpp
msimberg: That's what I wanted to do first :P we already have a `thread_state` though, which is the combination of `thread_state_enum` and `thread_state_ex_enum` (I was very confused why they'd need `enum` in the name until I remembered that): https://github.com/STEllAR-GROUP/hpx/blob/e907e3227dd0b22e4e3c7c662a608490de8adb5b/libs/core/coroutines/include/hpx/coroutines/thread_enums.hpp#L139-L141.

I think the combined state is probably the rightful owner of the `thread_state` name (not very strongly, I'm open to alternatives here). That said, I realize now that `thread_state_ex_enum` definitely doesn't need the `enum` except for symmetry with the other enum. The main enum could be `thread_schedule_state`? And the extended one `thread_wait_state`? Or `thread_restart_state`? Or something else?

In any case, a rename would indeed be appropriate.

hkaiser: > That's what I wanted to do first :P we already have a `thread_state` though, which is the combination of `thread_state_enum` and `thread_state_ex_enum` (I was very confused why they'd need `enum` in the name until I remembered that):
> 
> https://github.com/STEllAR-GROUP/hpx/blob/e907e3227dd0b22e4e3c7c662a608490de8adb5b/libs/core/coroutines/include/hpx/coroutines/thread_enums.hpp#L139-L141
> 
> .
> I think the combined state is probably the rightful owner of the `thread_state` name (not very strongly, I'm open to alternatives here). That said, I realize now that `thread_state_ex_enum` definitely doesn't need the `enum` except for symmetry with the other enum. The main enum could be `thread_schedule_state`? And the extended one `thread_wait_state`? Or `thread_restart_state`? Or something else?
> 
> In any case, a rename would indeed be appropriate.

Fair points. I withdraw my suggestion. Yours are acceptable.

aurianer (dismissed):
LGTM thanks a lot!

msimberg:
@hkaiser, @K-ballo, any idea what MSVC is not happy about here: https://github.com/STEllAR-GROUP/hpx/pull/5046/checks?check_run_id=1374461029?

K-ballo:
> @hkaiser, @K-ballo, any idea what MSVC is not happy about here: https://github.com/STEllAR-GROUP/hpx/pull/5046/checks?check_run_id=1374461029?

Windows SDK defines `small` as `char`.

components/iostreams/src/server/output_stream.cpp
msimberg: I updated these to be `thread_schedule_state` and `thread_wait_state` now.

msimberg:
So, I've renamed `thread_stacksize::small` to `thread_stacksize::small_`. I don't particularly like it but we have precedent with `default_`. I like it more than alternative names (`little`, `tiny`?) though, which is why I've kept it like this. Can someone think of another better name, or another workaround?

hkaiser (approved):
LGTM, thanks!

One minor comment (doesn't need to go in here): we should add `small` to the list of identifiers not to use (similar to the apple macros we are flagging as prohibited: https://github.com/STEllAR-GROUP/hpx/blob/master/tools/inspect/apple_macro_check.cpp#L27)

msimberg:
> LGTM, thanks!
> 
> One minor comment (doesn't need to go in here): we should add `small` to the list of identifiers not to use (similar to the apple macros we are flagging as prohibited: https://github.com/STEllAR-GROUP/hpx/blob/master/tools/inspect/apple_macro_check.cpp#L27)

Good suggestion! I'll add it here, otherwise it'll never be done...

hkaiser (approved):
LGTM, thanks!

PR #5047, opened 2020-11-06
Limit cuda jenkins run to nodes with exclusively Nvidia GPUs 

aurianer:
Both nodes with AMD GPUs are used for jenkins, one for the cuda run and the other for the hip run.
This limit the cuda jenkins run to partitions which contains exclusively NVIDIA GPUs.


hkaiser (approved):
LGTM, thanks!

PR #5048, opened 2020-11-07
Limit jenkins hipcc to the reno node

aurianer:
Paired with #5047

Limit the hipcc jenkins to only the reno nodes, will revert it once the main work on hipcc is done.


hkaiser (approved):
LGTM, thanks!

PR #5050, opened 2020-11-09
Replace old init overloads in tests and examples

aurianer:
- Change the  old `hpx::init` overloads to the one using the `struct init_params`
- Add deprecated messages for the old `init` overloads
- Unify the indentation/name of variables for some `init` calls to facilitate the use of scripts

### DIFF
--- a/libs/full/init_runtime/include/hpx/hpx_init_impl.hpp
+++ b/libs/full/init_runtime/include/hpx/hpx_init_impl.hpp
@@ -147,6 +147,9 @@ namespace hpx {
     /// (or one of its overloads below) should be called from the users `main()`
     /// function. It will set up the HPX runtime environment and schedule the
     /// function given by \p f as a HPX thread.
+    HPX_DEPRECATED_V(1, 6,
+        "The init overload used is deprecated. Please use"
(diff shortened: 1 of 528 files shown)


libs/full/init_runtime/include/hpx/hpx_init_impl.hpp
```diff
@@ -147,6 +147,8 @@ namespace hpx {
     /// (or one of its overloads below) should be called from the users `main()`
     /// function. It will set up the HPX runtime environment and schedule the
     /// function given by \p f as a HPX thread.
+    HPX_DEPRECATED_V(1, 6, "The init overload used is deprecated please use"
+            "the new init overloads using the hpx::init_params struct")
```
msimberg, line 151:
```suggestion
    HPX_DEPRECATED_V(1, 6, "The init overload used is deprecated. Please use"
            "the init overloads using the hpx::init_params struct.")
```

msimberg (changes requested):
Thank you! Could you please do this for the `start` overloads as well?

aurianer:
I think this is good to go

msimberg (approved):
Nice, thanks for doing this!

PR #5051, opened 2020-11-09
Adapt transform to C++20

gonidelis:
This adapts `transform` to C++20 (all overloads according to [the Standard](http://eel.is/c++draft/alg.transform)) and adds unit tests. It deprecates the `hpx::parallel::transform` overloads.

flyby: it adds `get_in_in_out_result`.

### DIFF
--- a/libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/transform.hpp
+++ b/libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/transform.hpp
@@ -9,9 +9,9 @@
 #include <hpx/config.hpp>
 #include <hpx/algorithms/traits/segmented_iterator_traits.hpp>
 #include <hpx/datastructures/tuple.hpp>
+#include <hpx/parallel/util/result_types.hpp>
 #include <hpx/parallel/util/tagged_pair.hpp>
 #include <hpx/parallel/util/tagged_tuple.hpp>
-#include <hpx/parallel/util/result_types.hpp>
 
 #include <hpx/executors/execution_policy.hpp>
 #include <hpx/parallel/algorithms/detail/dispatch.hpp>
@@ -68,8 +68,9 @@ namespace hpx { namespace parallel { inline namespace v1 {
                 local_iterator_type2 ldest = traits2::begin(sdest);
                 if (beg != end)
                 {
-                    util::in_out_result<local_iterator_type1, local_iterator_type2> out =
-                        dispatch(traits2::get_id(sdest), algo, policy,
+                    util::in_out_result<local_iterator_type1,
+                        local_iterator_type2>
+                        out = dispatch(traits2::get_id(sdest), algo, policy,
                             std::true_type(), beg, end, ldest, f, proj);
                     last = traits1::compose(send, out.in);
                     dest = traits2::compose(sdest, out.out);
@@ -81,7 +82,8 @@ namespace hpx { namespace parallel { inline namespace v1 {
                 local_iterator_type1 beg = traits1::local(first);
                 local_iterator_type1 end = traits1::end(sit);
                 local_iterator_type2 ldest = traits2::begin(sdest);
-                util::in_out_result<local_iterator_type1, local_iterator_type2> out;
+                util::in_out_result<local_iterator_type1, local_iterator_type2>
+                    out;
                 if (beg != end)
                 {
                     out = dispatch(traits2::get_id(sdest), algo, policy,
@@ -113,7 +115,8 @@ namespace hpx { namespace parallel { inline namespace v1 {
                 last = traits1::compose(send, out.in);
                 dest = traits2::compose(sdest, out.out);
             }
-            return result::get(util::in_out_result<SegIter, OutIter>{std::move(last), std::move(dest)});
+            return result::get(util::in_out_result<SegIter, OutIter>{
+                std::move(last), std::move(dest)});
         }
 
         // parallel remote implementation
@@ -143,8 +146,8 @@ namespace hpx { namespace parallel { inline namespace v1 {
             segment_iterator1 send = traits1::segment(last);
             segment_iterator2 sdest = traits2::segment(dest);
 
-            typedef std::vector<
-                future<util::in_out_result<local_iterator_type1, local_iterator_type2>>>
+            typedef std::vector<future<util::in_out_result<local_iterator_type1,
+                local_iterator_type2>>>
                 segment_type;
             segment_type segments;
             segments.reserve(std::distance(sit, send));
@@ -227,7 +230,8 @@ namespace hpx { namespace parallel { inline namespace v1 {
 
             if (first == last)
             {
-                return result::get(util::in_out_result<SegIter, OutIter>{std::move(last), std::move(dest)});
+                return result::get(util::in_out_result<SegIter, OutIter>{
+                    std::move(last), std::move(dest)});
             }
 
             typedef hpx::traits::segmented_iterator_traits<SegIter>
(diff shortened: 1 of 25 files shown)


libs/parallelism/algorithms/tests/unit/container_algorithms/CMakeLists.txt
```diff
@@ -85,7 +85,8 @@ set(tests
     sort_range
     stable_sort_range
     transform_range
-    transform_range_binary
+    transform_range2
+    #transform_range_binary
```
msimberg, line 89: Why is this commented out?

msimberg (commented):
Thanks @gonidelis for updating this!

libs/parallelism/algorithms/tests/unit/container_algorithms/CMakeLists.txt
gonidelis: This test checks an overload that does not exist in the updated C++20 overloads. There is some discussion need to be made as far as tests are concerned. I added `transform_range2` and I think there are one or two tests that need to be added in order for all the overloads to be checked. Please tell me if you agree in order for me to completeley delete `transform_range_binary`.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform.hpp
```diff
@@ -681,121 +705,31 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+        HPX_CONCEPT_REQUIRES_(
+            hpx::is_execution_policy<
+                ExPolicy>::value&& hpx::traits::is_iterator<FwdIter1>::value&&
+            hpx::traits::is_iterator<FwdIter2>::value&&
+            hpx::traits::is_iterator<FwdIter3>::value&&
+            traits::is_projected<Proj1, FwdIter1>::value&&
+            traits::is_projected<Proj2, FwdIter2>::value&&
+                traits::is_indirect_callable<ExPolicy, F,
+                    traits::projected<Proj1, FwdIter1>,
+                    traits::projected<Proj2, FwdIter2>>::value
```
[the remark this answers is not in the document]
gonidelis: Actually I have already manually formated these, so please tell me if there is any problem with the identation and I will fix the `&&`s asap. Thanks :)

libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform.hpp
```diff
@@ -8,6 +8,170 @@
 
 #pragma once
 
+#if defined(DOXYGEN)
+namespace hpx {
+    // clang-format off
```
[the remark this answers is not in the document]
gonidelis: I will check it and update you soon. Thank you :)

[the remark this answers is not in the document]
gonidelis: @msimberg  Seems like it is redundant after all. It will be removed in the next commit.

libs/parallelism/algorithms/tests/unit/container_algorithms/CMakeLists.txt
msimberg: Sorry for not responding here earlier @gonidelis. I think it'd make sense to leave the test for the overloads that are not part of C++20. This means they'd still be using `hpx::parallel::transform`, `hpx::transform` or `hpx::ranges::transform`. This will lead to deprecation warnings so you'd have to add either `#define HPX_HAVE_DEPRECATION_WARNINGS_V1_6 0` to the test, or `target_compile_definitions(transform_range_binary_test HPX_HAVE_DEPRECATION_WARNINGS_V1_6=0)` to the CMake file. This will of course be deleted once we delete the deprecated overloads but it's unnecessary to delete tests right now that we already have.

gonidelis:
@hkaiser  Thank you very much for providing this!

libs/parallelism/algorithms/tests/unit/container_algorithms/CMakeLists.txt
gonidelis: @msimberg  sorry for responding so late to this, but since we are going to leave just this test as is and since `parallel:transform` invokes `transform_` which returns an `in_out_result` there is some inconsistency here. I could change the result type of `parallel::transform` but it wouldn't be deprecated then. What's the trick?

hkaiser:
@gonidelis apparently, tests/performance/local/CMakeFiles/stream_test.dir/stream.cpp relies on a now obsolete API. Would you mind updating this to the appropriate new API?

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/transform.hpp
```diff
@@ -9,6 +9,7 @@
 #include <hpx/config.hpp>
 #include <hpx/algorithms/traits/segmented_iterator_traits.hpp>
 #include <hpx/datastructures/tuple.hpp>
+#include <hpx/parallel/util/result_types.hpp>
 #include <hpx/parallel/util/tagged_pair.hpp>
 #include <hpx/parallel/util/tagged_tuple.hpp>
```
hkaiser, line 14: Are the `tagged_xxx.hpp` headers still needed?

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/transform.hpp
```diff
@@ -303,40 +358,357 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+#else    // DOXYGEN
+
+#include <hpx/config.hpp>
+#include <hpx/concepts/concepts.hpp>
+#include <hpx/iterator_support/range.hpp>
+#include <hpx/iterator_support/traits/is_iterator.hpp>
+#include <hpx/iterator_support/traits/is_range.hpp>
+#include <hpx/parallel/util/result_types.hpp>
+#include <hpx/parallel/util/tagged_pair.hpp>
+#include <hpx/parallel/util/tagged_tuple.hpp>
```
hkaiser, line 400: Are the `tagged_xxx.hpp` headers still needed?

hkaiser (dismissed):
Otherwise, LGTM, thanks!

msimberg:
@gonidelis I think this is the last remaining deprecated overload that should be updated: https://github.com/gonidelis/hpx/blob/09bff92fb2be501e89dd6437e524091d150d2bc0/libs/full/compute_cuda/tests/unit/transform_compute.cu#L48-L49.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/transform.hpp
gonidelis: No they are not. Please don't merge there is still some minor housekeeping left done. It will be ready within the day. Thanks :)

gonidelis:
@msimberg  Ok I will fix that. Is CI back again? I will do a `grep` just to be safe in any other case.

msimberg:
> @msimberg Ok I will fix that. Is CI up again? I will do a `grep` just to be safe in any other case.

The rostam jenkins CI is still down, but all the rest are up as usual. There are some additional configurations on rostam, but they're unlikely to uncover any additional issues on this PR.

gonidelis:
I am pretty sure that I fixed any remaining depracations and unnecessary `tagged_xxx` uses. The PR should be ready to merge now if CI is ok with it.

msimberg:
Looks like CI is (sort of) happy now, thanks @gonidelis! Let's merge this and deal with any remaining problems once rostam is back up (hopefully none).

Issue #5045, opened 2020-11-06
Implement P0443 senders and receivers


msimberg:
This is a tracking issue for the *new and improved* P0443 (previous issue was here: https://github.com/STEllAR-GROUP/hpx/issues/2439). As with the previous issue, things are still in progress, so we can't really go all in yet, but we can explore things and see what changes are required.

I'm starting this by collecting the relevant proposals here:
- A Unified Executors Proposal for C++: https://wg21.link/p0443
- One-Way execute is a Poor Basis Operation: https://wg21.link/p1525
- Towards C++23 executors: A proposal for an initial set of algorithms: https://wg21.link/p1897
- Disentangling schedulers and executors: https://wg21.link/p2235
- A General Property Customization Mechanism: https://wg21.link/p1393
- redefine properties in P0443: https://wg21.link/p2220

Related PRs:
- Sender, receiver, operation state: #4665 
- Properties: #5033

msimberg:
I have local changes to add the remaining CPOs defined in P0443, and some basic senders/receivers that work on our thread pools. I'll be opening a PR next week...

msimberg:
Added P2300 (https://isocpp.org/files/papers/P2300R0.html) to the list. It combines P0443 and P1897 into a single paper and cleans up some rough edges.

hkaiser:
Re P2300: here is a list of things that are changed compared to P0443 and P1897:

There is a new concept of completion-schedulers that may be associated with a particular sender. Thus
- [ ] `get_completion_scheduler` is a new customization point to be implemented for senders

General changes:
- [ ] `transform` has been renamed to `then`, `lazy_the` has been added
- [ ] `sync_wait` should go into the `namespace hpx::this_thread` (same as `std::this_thread_sync_wait`
- [ ] `on` has a new parameter sequence, `lazy_on` has been added
- [ ] `transfer_just` has been added to associate a completion scheduler with `just`
- [ ] added `upon_error` and `upon_done` to complement `then`
- [ ] added `lazy_let_value` and `lazy_let_error`
- [ ] added `transfer`, `lazy_transfer`
- [ ] added `into_variant`
- [ ] added `unschedule`
- [ ] added `bulk`
- [ ] added `transfer_when_all` and `lazy_transfer_when_all` and *_variant versions

msimberg:
[...]

PR #5053, opened 2020-11-13
Add P0443r14 executor and a a few P1897 algorithms

msimberg:
Adds the remaining CPOs and traits from P0443r14 (see #5045). I've used the fallback implementations from P0443r14 pretty much verbatim, so there might be room for changes here.

Also adds `transform` and `sync_wait` as the first algorithms (and `wait`, `get` for symmetry with the current `future` functionality, although these should proably not stay). Also adds a `sink_receiver` (should be called `void_receiver`?).

Adds a simple executor for our coroutines based on the new functionality. The `bulk_execute` functionality is incomplete but I'd like to keep it like this for now until the proposal evolves a bit more.

Extends `tag_fallback_invoke` with `tag_priority_invoke`, which gives three options: `tag_override_invoke` which always takes precedence over user-defined `tag_invoke`s, then uses user-defined `tag_invoke`s, and finally uses `tag_fallback_invoke` if nothing else matches.

@K-ballo (and others interested), I had a need for a type list `transform` and `unique`. I'm sure you have a better idea of how to do that efficiently (especially `unique`), so I'd appreciate your eyes on that (see https://github.com/msimberg/hpx/blob/1fa6ec165d410d0508183cb0e1d532e1840f0cf6/libs/core/type_support/include/hpx/type_support/pack.hpp#L196-L237).

Includes #5033.

To do:
- [x] Add documentation for the new CPOs
- [x] Update documentation for the existing CPOs
- [x] Move traits out from `traits` namespace (objections?)


msimberg:
retest

msimberg:
I welcome reviews on this one.

libs/core/functional/include/hpx/functional/tag_priority_invoke.hpp
```diff
@@ -0,0 +1,301 @@
...
+        constexpr HPX_FORCEINLINE auto operator()(Args&&... args) const noexcept
+            -> tag_fallback_invoke_result_t<Tag, Args&&...>
+        {
+            return hpx::functional::tag_fallback_invoke(
+                static_cast<Tag const&>(*this), std::forward<Args>(args)...);
+        }
+    };
+}}    // namespace hpx::functional
+
+#endif
```
hkaiser, line 301: I like this!

hkaiser (dismissed):
LGTM, thanks!

msimberg:
> LGTM, thanks!

Thanks @hkaiser for looking through this! I'm turning this back into a draft because it looks like gcc 7 and 8 are not happy with the changes. I'm not planning on making any functional changes.

hkaiser:
One question I had was whether we should use `tag_fallback_invoke` instead of `tag_invoke` where possible (based on our experience with segregating the segmented algorithms).

msimberg:
> One question I had was whether we should use `tag_fallback_invoke` instead of `tag_invoke` where possible (based on our experience with segregating the segmented algorithms).

Absolutely! The P0443 customization points use `tag_priority` which gives both `tag_fallback_invoke` and `tag_priority_invoke` (the latter because according to P0443 member functions should always take precedence over free functions, or tag_invoke overloads in our case). `sync_wait` and `transform` are not even CPOs here, but should be, and should use `tag_fallback_invoke` (I've already changed this in upcoming commits).

msimberg:
This *should* now be ok. Let's see if CI agrees. GCC 7 and 8 seemed to be more eager in instantiating templates compared to newer versions of GCC and clang.

PR #5056, opened 2020-11-19
Fix hpx::util::get

msimberg:
The deprecated forwarding definition was... not right.

### DIFF
--- a/libs/core/datastructures/include/hpx/datastructures/tuple.hpp
+++ b/libs/core/datastructures/include/hpx/datastructures/tuple.hpp
@@ -142,7 +142,7 @@ namespace hpx {
                                         UTuple>::type>::value ==
                 util::pack<Ts...>::size>::type>
           : util::all_of<std::is_convertible<
-                decltype(get<Is>(std::declval<UTuple>())), Ts>...>
+                decltype(hpx::get<Is>(std::declval<UTuple>())), Ts>...>
         {
         };
 
@@ -1072,13 +1072,13 @@ namespace hpx { namespace util {
         return hpx::tuple_cat(std::forward<Ts>(vs)...);
     }
 
-    template <typename... Ts>
+    template <std::size_t I, typename Tuple>
     HPX_DEPRECATED_V(1, 6,
         "hpx::util::get is deprecated. Use hpx::get "
         "instead.")
-    constexpr HPX_HOST_DEVICE HPX_FORCEINLINE auto get(Ts&&... vs)
+    constexpr HPX_HOST_DEVICE HPX_FORCEINLINE auto get(Tuple&& tuple)
     {
-        return hpx::get(std::forward<Ts>(vs)...);
+        return hpx::get<I>(std::forward<Tuple>(tuple));
     }
 }}    // namespace hpx::util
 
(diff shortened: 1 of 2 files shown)


hkaiser (commented):
Don't we need both overloads (i.e. the old one and the one you created)?

msimberg:
> Don't we need both overloads (i.e. the old one and the one you created)?

Mmh, do you mean a `template <typename T, ...> auto get(...)` overload? Yes, in principle, I didn't think about that one. However, I don't think we have an implementation of that, or can I just not find it? @K-ballo?

hkaiser:
> > Don't we need both overloads (i.e. the old one and the one you created)?
> 
> Mmh, do you mean a `template <typename T, ...> auto get(...)` overload? Yes, in principle, I didn't think about that one. However, I don't think we have an implementation of that, or can I just not find it? @K-ballo?

I take back what I said, sorry. Your implementation is fine.

hkaiser (dismissed):
LGTM, thanks!

hkaiser:
It was this code I got confused by: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/datastructures/include/hpx/datastructures/variant_helper.hpp#L23-L56. We need to add overloads there to support `get<T>(v)` n addition to `get<0>(v)` (assuming `std::variant<...> v;`). Once we have those overloads, we will need both versions forwarding to the new `hpx::get`.

msimberg:
> It was this code I got confused by: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/datastructures/include/hpx/datastructures/variant_helper.hpp#L23-L56. We need to add overloads there to support `get<T>(v)` in addition to `get<0>(v)` (assuming `std::variant<...> v;`). Once we have those overloads, we will need both versions forwarding to the new `hpx::get`.

Not sure if this is what you meant that we don't need *deprecated* forwarding definitions if we didn't have the typed versions in the first place. The namespace of those variant helpers was wrong as well. I've moved them in this PR.

hkaiser:
> > It was this code I got confused by: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/datastructures/include/hpx/datastructures/variant_helper.hpp#L23-L56. We need to add overloads there to support `get<T>(v)` in addition to `get<0>(v)` (assuming `std::variant<...> v;`). Once we have those overloads, we will need both versions forwarding to the new `hpx::get`.
> 
> Not sure if this is what you meant that we don't need _deprecated_ forwarding definitions if we didn't have the typed versions in the first place. The namespace of those variant helpers was wrong as well. I've moved them in this PR.

I know I'm talking in riddles :/ -- let me try again. The underlying issue is that we provide helpers to make our `hpx::get` work with `std::tuple` (and friends) and `std::variant`. We however provide only helpers for the `get<I>()` overloads and not for the `get<T>()` overloads (needed for `std::variant`). When I was writing the code, I was thinking that it would work for both cases. I was wrong on two accounts: a) we do need separate helpers, and b) we didn't even have the full set of proper helpers for `std::variant` in the first place. All of that muddled together with trying to get away with terse comments and little thinking has caused my confusion (and now yours). 

So what we should do in addition to this PR is to add `hpx::get<T>()` helpers for `std::variant`. We might get away with not providing compatibility functions for the new overloads in the old namespace.

I hope that things make more sense now.

PR #5061, opened 2020-11-22
Remove hipcc guards in examples and tests

aurianer:
Remove the hipcc guards in some of the examples and tests, following #5055

The remaining guards are due to the fact that the actions are undefined in device code so all tests using actions/components still need the guard. A potential solution would be to define the actions to be normal functions in device code but I'm not sure how much work it would require.


msimberg (approved):
Thanks!

PR #5063, opened 2020-11-24
Limit the device code guards to the distributed parts of the future_overhead bench

aurianer:
Instead of disabling the entire test we only disable the distributed parts (forgotten in #5061)


msimberg (approved):
Thanks!

PR #5064, opened 2020-11-25
Refactor runtime configuration, command line handling, and resource partitioner

msimberg:
Moves command line handling out of the resource partitioner and into `hpx_init.cpp`. This leaves the partitioner only needing the runtime configuration and affinity data to set things up. The command line handling is now also slightly dumber in the sense that it only parses command line options and populates the configuration, but after that configuration options should be accessed through the runtime configuration, not through the command line handling.

This does contain some potentially breaking changes (e.g. command line handling can't be accessed through the partitioner), and we'll have to see how this affects users. My guess is that it's again a very small fraction (like @biddisco) who are affected.

@K-ballo this should also get rid of program options leaking out from the partitioner. If you happen to have the time to check the effect of this I'd appreciate that very much!

Edit: I should add that it should also be possible to not have a partitioner singleton anymore, but I'm leaving that for a separate PR.

### DIFF
--- a/libs/full/threadmanager/src/threadmanager.cpp
+++ b/libs/full/threadmanager/src/threadmanager.cpp
@@ -11,7 +11,6 @@
 #include <hpx/config.hpp>
 #include <hpx/assert.hpp>
 #include <hpx/async_combinators/wait_all.hpp>
-#include <hpx/command_line_handling/command_line_handling.hpp>
 #include <hpx/execution_base/this_thread.hpp>
 #include <hpx/futures/future.hpp>
 #include <hpx/hardware/timestamp.hpp>
@@ -29,7 +28,7 @@
 #include <hpx/threading_base/thread_init_data.hpp>
 #include <hpx/threading_base/thread_queue_init_parameters.hpp>
 #include <hpx/topology/topology.hpp>
-#include <hpx/util/from_string.hpp>
+#include <hpx/util/get_entry_as.hpp>
 
 #include <cstddef>
 #include <cstdint>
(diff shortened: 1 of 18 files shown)


K-ballo:
> @K-ballo this should also get rid of program options leaking out from the partitioner. If you happen to have the time to check the effect of this I'd appreciate that very much!

There's no change in the number of `boost/program_options` headers being included.

libs/full/threadmanager/src/threadmanager.cpp
```diff
@@ -105,6 +82,7 @@ namespace hpx { namespace threads {
         notifier.add_on_stop_thread_callback(
             util::bind(&threadmanager::deinit_tss, this));
 
+        // TODO: Pass reference to partitioner through constructor.
```
aurianer, line 85: Should it be removed? Is this for another PR?

aurianer (approved):
LGTM, thanks a lot for doing this!

libs/full/threadmanager/src/threadmanager.cpp
msimberg: Removed, thanks! This is potentially for a future PR.

msimberg:
> > @K-ballo this should also get rid of program options leaking out from the partitioner. If you happen to have the time to check the effect of this I'd appreciate that very much!
> 
> There's no change in the number of `boost/program_options` headers being included.

For others following along, `command_line_handling` is still included in a few other places, leading to `boost/program_options` also still leaking. Getting rid of that will need further cleanup.

K-ballo:
> For others following along, `command_line_handling` is still included in a few other places, leading to `boost/program_options` also still leaking. Getting rid of that will need further cleanup.

The bulk of the includes (73/91) come from `hpx/libs/full/resource_partitioner/include/hpx/resource_partitioner/detail/partitioner.hpp`. (It's possible both partitioner and command_line_handling are leaking these, and partitioner happens to be first so it's the one reflected in the report).

msimberg:
> > For others following along, `command_line_handling` is still included in a few other places, leading to `boost/program_options` also still leaking. Getting rid of that will need further cleanup.
> 
> The bulk of the includes (73/91) come from `hpx/libs/full/resource_partitioner/include/hpx/resource_partitioner/detail/partitioner.hpp`. (It's possible both partitioner and command_line_handling are leaking these, and partitioner happens to be first so it's the one reflected in the report).

Urg, well spotted @K-ballo. I was too focused on `command_line_handling` including `program_options` transitively to notice that `program_options` was included directly. I've just pushed a commit which removes it from the resource partitioner module, which from quick testing seems to be ok. As far as I can tell there should be no need for it in that module anymore.

That still leaves program options being included in the init headers, but that may be unavoidable. Not sure... That would at least of the time be localized to a single translation unit. The distributed runtime headers also still include program options/command line handling.

msimberg:
I'm not planning any further changes to this. The program options dependency is gone from the partitioner, but any further cleanup is for separate PRs.

hkaiser (approved):
LGTM, thanks!

PR #5065, opened 2020-12-01
Fix hpx::range::adjacent_find doxygen function signatures

msimberg:
Thanks @aurianer for pointing this out!


aurianer (approved):
LGTM thanks!

PR #5066, opened 2020-12-01
Adapt `search` to C++20 and Range TS

aurianer:
- Move `search` and `search_n` in the `hpx` namespace the respective overloads in the `hpx::ranges` namespace.
- Add CPOs for `search` and `search_n`.
- Add missing overloads in line with the C++ working draft (without execution policy).
- Add cppreference sequential [possible implementation](https://en.cppreference.com/w/cpp/algorithm/ranges/search) of search to support sentinels + adapt the parallel version (remove assignment/return statement of the sentinel).
- Add overloads in the `ranges` namespace to support Iterator-Sentinel pair.
- Add tests for Iterator-Sentinel overloads (the test is quite dumb as the sentinel is constructed from an iterator, but it helps to check the validity of the overloads)

Will do the `search_n` sentinel adaptation in another PR


libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/search.hpp
```diff
@@ -21,7 +23,306 @@
 #include <type_traits>
 #include <utility>
 
-namespace hpx { namespace parallel { inline namespace v1 {
+#if defined(DOXYGEN)
+
+namespace hpx { namespace ranges {
+    // clang-format off
```
msimberg, line 29: This should not be necessary.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/search.hpp
```diff
@@ -21,7 +23,306 @@
...
+            hpx::traits::is_forward_iterator<FwdIter2>::value &&
+            hpx::traits::is_sentinel_for<Sent2, FwdIter2>::value &&
+            parallel::traits::is_projected<Proj2, FwdIter2>::value &&
+            parallel::traits::is_indirect_callable<
+                hpx::execution::sequenced_policy,
+                Pred,
+                parallel::traits::projected<Proj1, FwdIter>,
+                parallel::traits::projected<Proj2, FwdIter2>
+            >::value
+        )>
```
msimberg, line 125: We've tended to leave this out from the doxygen declaration. The comment above instead documents the constraints.

msimberg (commented):
Just a couple of quick comments...

libs/parallelism/algorithms/include/hpx/parallel/algorithms/search.hpp
```diff
@@ -364,110 +509,172 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+                parallel::traits::is_indirect_callable<
+                    hpx::execution::sequenced_policy,
+                    Pred, parallel::traits::projected<Proj1, FwdIter>,
+                    parallel::traits::projected<Proj2, FwdIter2>
+                >::value
+            )>
+        // clang-format on
+        friend FwdIter tag_invoke(hpx::search_t, FwdIter first, FwdIter last,
+            FwdIter2 s_first, FwdIter2 s_last, Pred&& op = Pred(),
+            Proj1&& proj1 = Proj1(), Proj2&& proj2 = Proj2())
```
hkaiser, line 571: I think `hpx::search` shouldn't expose projections.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/search.hpp
```diff
@@ -364,110 +509,172 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+                    Pred, parallel::traits::projected<Proj1, FwdIter>,
+                    parallel::traits::projected<Proj2, FwdIter2>
+                >::value
+            )>
+        // clang-format on
+        friend FwdIter tag_invoke(hpx::search_n_t, FwdIter first,
+            std::size_t count, FwdIter2 s_first, FwdIter2 s_last,
+            Pred&& op = Pred(), Proj1&& proj1 = Proj1(),
+            Proj2&& proj2 = Proj2())
+        {
```
hkaiser, line 637: `hpx::search_n` shouldn't expose projections either.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/detail/search.hpp
```diff
@@ -0,0 +1,179 @@
...
+// Project home: https://github.com/ericniebler/range-v3
+//
+//===----------------------------------------------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is dual licensed under the MIT and the University of Illinois Open
+// Source Licenses. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
```
hkaiser, line 19: Are we allowed to re-license this under the Boost license?

libs/parallelism/algorithms/tests/unit/container_algorithms/test_utils.hpp
```diff
@@ -40,6 +61,26 @@ namespace test {
           : base_type(base)
         {
         }
+
+        friend bool operator==(test_iterator i, Sentinel<BaseIterator> s)
+        {
+            return *i == *s;
```
hkaiser, line 67: Is this correct? Why do we compare the results from dereferencing the iterator and the sentinel?

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/detail/search.hpp
aurianer: So I checked that the MIT license allows us to "use, copy, modify, merge, publish, distribute", do I need to check specifically something else?

libs/parallelism/algorithms/tests/unit/container_algorithms/test_utils.hpp
aurianer: Can you be more precise? I'm still a newbie when it comes to C++ sorry

libs/parallelism/algorithms/include/hpx/parallel/algorithms/search.hpp
aurianer: Thanks!

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/detail/search.hpp
hkaiser: It also requires us to redistribute the original license, iiuc. I would suggest to avoid directly using code that originally was not published under the BSL. What part of the file did you take from LLVM?

libs/parallelism/algorithms/tests/unit/container_algorithms/test_utils.hpp
hkaiser: Sentinels can represent at least 3 different concepts: 
- iterators (the algorithm is called using a iterator pair)
- a type that encapsulates a 'sentinel value', e.g., when representing a C-style string with a pointer and the `'\0'` EOS 
- a type that encapsulates a range length, e.g. when representing a byte sequence with a pointer and a number of (consecutive) bytes

Your solution to overload the iterator's `operator==` comparing the dereferenced values will give correct answers only for the second case.

I think the correct solution would be to overload a `operator==` on the sentinel type (not the iterator) and `enable_if` it only for iterators the sentinel is valid for (`is_sentinel_for<Sent, Iter>::value == true`).

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/detail/search.hpp
aurianer: Ah okay didn't know that exactly, I took the sequential impl from the range-v3 repo.
I saw some "possible implementations" on cppreference but it is probably a bit slower, what do you suggest?

libs/parallelism/algorithms/tests/unit/container_algorithms/test_utils.hpp
aurianer: Thanks a lot for the explanations!

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/detail/search.hpp
aurianer: Would putting the original MIT license on top of the header taking only the range-v3 sequential adaptation be enough?

hkaiser: There are at least two ways out. Either - as you suggested:
- use the sequential implementation from cppreference and adapt it to support projections
- use the `std::search` as before, but pass our `compare_projected` helper function object 

Both are fine as a starting point, I believe.

hkaiser: BTW, once you have removed the license, please force-push to this branch such that the commit completely disappears from the git history.

aurianer: Yes no problem :+1:

libs/parallelism/algorithms/include/hpx/parallel/algorithms/search.hpp
```diff
@@ -15,7 +16,7 @@
 #include <hpx/execution/algorithms/detail/predicates.hpp>
 #include <hpx/executors/execution_policy.hpp>
 #include <hpx/parallel/algorithms/detail/dispatch.hpp>
-#include <hpx/parallel/algorithms/for_each.hpp>
+#include <hpx/parallel/container_algorithms/detail/search.hpp>
```
hkaiser, line 19: This is a bit unusual as normally the container versions depend on the non-container versions. Would it be possible to move this header here instead?

libs/parallelism/algorithms/include/hpx/parallel/algorithms/search.hpp
```diff
@@ -364,110 +453,138 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+    HPX_INLINE_CONSTEXPR_VARIABLE struct search_t final
+      : hpx::functional::tag<search_t>
+    {
+    private:
+        // clang-format off
+        template <typename FwdIter, typename FwdIter2,
+            typename Pred = parallel::v1::detail::equal_to,
+            HPX_CONCEPT_REQUIRES_(
+                traits::is_forward_iterator<FwdIter>::value &&
+                traits::is_forward_iterator<FwdIter2>::value
```
msimberg, line 501: Can we add `&& is_invocable_v<Pred, typename std::iterator_traits<FwdIter1>::value_type, typename std::iterator_traits<FwdIter1>::value_type>` (or do we have a more appropriate trait for that)? Same in the other `HPX_CONCEPT_REQUIRES_`.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/detail/search.hpp
```diff
@@ -0,0 +1,153 @@
...
+#include <hpx/parallel/util/partitioner.hpp>
+
+
+#include <cstddef>
+#include <iterator>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+namespace hpx { namespace ranges { namespace detail {
```
msimberg, line 27: Following @hkaiser's comment about moving this file, I'd also move this back to `hpx::parallel::v1::detail` even though it does also deal with ranges. The implementation is used for both the ranges and non-ranges versions though and if nothing else it'll stay more consistent with what we've done so far. Or do we want to start moving the algorithm *implementations* into `hpx::ranges`?

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/search.hpp
```diff
@@ -9,19 +10,281 @@
...
 #include <hpx/config.hpp>
+#include <hpx/execution/algorithms/detail/predicates.hpp>
 #include <hpx/execution/traits/is_execution_policy.hpp>
 #include <hpx/iterator_support/range.hpp>
 #include <hpx/iterator_support/traits/is_range.hpp>
 
 #include <hpx/algorithms/traits/projected.hpp>
 #include <hpx/algorithms/traits/projected_range.hpp>
+// Still need to include the non-range header since we don't have search_n
+// sequential implementation supporting sentinels yet
```
msimberg, line 21: Is this comment relevant? I think we include the non-range header for *all* the adapted algorithms. If anything we could start moving the implementations into something like `hpx/parallel/algorithms/detail`, but at the moment we always include the non-range header because it contains the actual algorithm implementations.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/search.hpp
```diff
@@ -198,24 +686,360 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+    HPX_DEPRECATED_V(1, 6, "Please use hpx::ranges::search instead.")
+    // clang-format on
     typename util::detail::algorithm_result<ExPolicy,
         typename hpx::traits::range_iterator<Rng1>::type>::type
-    search_n(ExPolicy&& policy, Rng1&& rng1, std::size_t count, Rng2&& rng2,
-        Pred&& op = Pred(), Proj1&& proj1 = Proj1(), Proj2&& proj2 = Proj2())
+        search(ExPolicy&& policy, Rng1&& rng1, Rng2&& rng2, Pred&& op = Pred(),
+            Proj1&& proj1 = Proj1(), Proj2&& proj2 = Proj2())
+    {
+        return search(std::forward<ExPolicy>(policy), hpx::util::begin(rng1),
```
msimberg, line 726: Let's call the algorithm implementation directly here instead of going through the CPO again. Same elsewhere.

libs/parallelism/algorithms/tests/unit/container_algorithms/search_range.cpp
```diff
@@ -85,6 +104,8 @@ template <typename IteratorTag>
 void test_search1()
 {
     using namespace hpx::execution;
+    test_search1_without_expolicy(IteratorTag());
```
msimberg, line 107: This is a matter of taste. I like to have this simply as an overload of `test_search1` with the execution policy argument missing, but feel free to leave it like this if you prefer that.

libs/parallelism/algorithms/tests/unit/container_algorithms/search_range.cpp
```diff
@@ -289,7 +375,7 @@ void test_search4()
 
 void search_test4()
 {
-    test_search4<std::random_access_iterator_tag>();
+    //test_search4<std::random_access_iterator_tag>();
```
msimberg, line 378: ?

libs/parallelism/algorithms/tests/unit/container_algorithms/test_utils.hpp
```diff
@@ -19,6 +21,46 @@
 #include <vector>
 
 namespace test {
+    ///////////////////////////////////////////////////////////////////////////
+    // Sentinel constructed from an Iterator just for the purpose of the
+    // overloads tests
+    template <typename IterType>
+    struct Sentinel
```
msimberg, line 28: Can we make this `sentinel` instead (i.e. snake case; I know it's capitalized in `iterator_traits`... you could change that one as well)?

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/detail/search.hpp
```diff
@@ -0,0 +1,153 @@
...
+            for (;; ++first)
+            {
+                FwdIter it1 = first;
+                for (FwdIter2 it2 = s_first;; ++it1, ++it2)
+                {
+                    if (it2 == s_last)
+                        return first;
+                    if (it1 == last)
+                        return it1;
+                    if (!hpx::util::invoke(op, hpx::util::invoke(proj1, *it1),
```
msimberg, line 54: Would you mind using `HPX_INVOKE` here instead (and in the other `search`-related files? As in https://github.com/STEllAR-GROUP/hpx/pull/4991.

msimberg:
This might need some looking into: https://cdash.cscs.ch/testDetails.php?test=33118718&build=137674 :/

libs/parallelism/algorithms/include/hpx/parallel/algorithms/search.hpp
aurianer: Yes no problem, I put it there as I had to do a sentinel adaptation and I thought it was related to container_algorithms.
Will do the change ;)

libs/parallelism/algorithms/include/hpx/parallel/algorithms/search.hpp
aurianer: Yes totally, forgot to add them back sorry

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/search.hpp
aurianer: When I used the range-v3 sequential impl, it made sense to put the detail implementation in another file because of the copyright. And I planned to adapt the search_n algorithm (other PR) in the same file and the container algos would then only depend on the detail file. It made sense for me as some other algos also have a detail file. But if you prefer having the detail impl in the main `algorithms/search.hpp` file I can move them back

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/detail/search.hpp
aurianer: I'll let @hkaiser reply, I don't really mind

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/search.hpp
aurianer: I didn't change this since it is the deprecated def. Should I change it?

libs/parallelism/algorithms/tests/unit/container_algorithms/search_range.cpp
aurianer: Oops

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/search.hpp
msimberg: I think `hpx/parallel/algorithms/detail/search.hpp` might make sense. We currently have the sequential, sentinel-supporting versions of algorithms in there.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/search.hpp
msimberg: Good point! That's up to you. The others could skip the CPO though (if they don't already).

libs/parallelism/algorithms/tests/unit/container_algorithms/test_utils.hpp
aurianer: Done for the sentinel in `test_utils`, will do the others including `Iterator` struct in `iter_sent.hpp` in another PR (see #5075)

aurianer:
> This might need some looking into: https://cdash.cscs.ch/testDetails.php?test=33118718&build=137674 :/

Yes will have a look!

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/search.hpp
msimberg: Could I suggest moving the `search_n` implementation to the detail header already now even if it doesn't support sentinels? It's seems to me that sentinel support is an orthogonal concern to having to include the non-range header.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/detail/search.hpp
hkaiser: There is no reason at this point to move this away from `parallelism::v1`, I think.

msimberg (approved):
Thanks!

hkaiser (approved):
LGTM, thanks!

PR #5068, opened 2020-12-02
Fix string split function

### LINKED ISSUE #5067: Bug string_util/split.hpp
## 1. split a string with `token_compress_mode::off`
call split a string  
```
std::vector<std::string> result;
hpx::string_util::split(result,
	"a-b-",
	[](auto ch) {return ch == '-'; },
	hpx::string_util::token_compress_mode::off);
```
* Expected size of `result`: 2
* Actual size of `result`: 3

## 2. split a string above with `token_compress_mode::on`
Error at ` pred(int(*token_begin))`.  
Cannot dereference string iterator because it is out of range (e.g. an end iterator)
```
while (token_end != std::end(str) && pred(int(*token_begin)))
{
	++token_begin;
}
```
## Specifications
  - HPX Version: master
  - Platform (compiler, OS): Windows

msimberg:
The implementation could dereference the end or past the end of the input. Fixes #5067.

### DIFF
--- a/libs/core/string_util/include/hpx/string_util/split.hpp
+++ b/libs/core/string_util/include/hpx/string_util/split.hpp
@@ -59,7 +59,7 @@ namespace hpx { namespace string_util {
             if (compress_mode == token_compress_mode::on)
             {
                 // Skip contiguous separators
-                while (token_end != std::end(str) && pred(int(*token_begin)))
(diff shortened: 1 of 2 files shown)


hkaiser (approved):
Nice catch! LGTM. Thanks.

aurianer (approved):
LGTM thanks!

PR #5069, opened 2020-12-02
Add support for sentinels to adjacent_find

msimberg:
Missed in #4996. Thanks again to @aurianer for going through the PRs!

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/algorithms/adjacent_find.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/algorithms/adjacent_find.hpp
@@ -144,36 +144,36 @@ namespace hpx { namespace parallel { inline namespace v1 {
     // adjacent_find
     namespace detail {
         /// \cond NOINTERNAL
-        template <typename Iter>
+        template <typename Iter, typename Sent>
         struct adjacent_find
-          : public detail::algorithm<adjacent_find<Iter>, Iter>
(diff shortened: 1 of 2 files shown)


libs/parallelism/algorithms/include/hpx/parallel/algorithms/adjacent_find.hpp
```diff
@@ -144,36 +144,36 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
 
-            template <typename ExPolicy, typename InIter, typename Pred,
-                typename Proj>
+            template <typename ExPolicy, typename InIter, typename Sent_,
+                typename Pred, typename Proj>
             static InIter sequential(
-                ExPolicy, InIter first, InIter last, Pred&& pred, Proj&& proj)
+                ExPolicy, InIter first, Sent_ last, Pred&& pred, Proj&& proj)
             {
                 return std::adjacent_find(first, last,
```
aurianer, line 161: I'm still wondering what to do here cause `std::adjacent_find` take `last` as being an `iterator`, `std::ranges::adjacent_find` takes a sentinel but is only available since C++20

msimberg: This is the underlying implementation, and it can be more general than the interface we present to users. `hpx::adjacent_find` does expect `first` and `last` to be the same type, and that is still allowed by this interface.

hkaiser (approved):
LGTM, thanks!

PR #5071, opened 2020-12-02
Fix unused parameters

aurianer:


msimberg (approved):
Good catch, thanks!

src/runtime/applier/applier.cpp
```diff
@@ -132,6 +141,8 @@ namespace hpx { namespace applier {
         prefixes.emplace_back(agas::get_console_locality());
 #endif
         return true;
+        HPX_UNUSED(prefixes);
+        HPX_UNUSED(ec);
```
msimberg, line 145: I would put these *before* the return statement. It also confuses at least GCC.

msimberg (changes requested):

PR #5073, opened 2020-12-07
Allow using noncopyable types with unwrapping

msimberg:
Avoiding decays at least makes the simple case work. CI will tell me if I've broken something else (unsure if some of the asynchronous cases depend on copies being made; should not if everything is set up correctly...).

Edit: this works for noncopyable lvalues but something breaks trying to make it work for rvalues. There's likely something that depends on having to make copies (e.g. containers). The current fix is enough for me at the moment.


hkaiser (approved):
LGTM, thanks!

aurianer (approved):
LGTM thanks

PR #5079, opened 2020-12-09
Add checks to make sure that MPI/CUDA polling is enabled/not disabled too early

msimberg:
Adds a check to make sure that polling has been registered at least once before getting MPI/CUDA futures, and another that checks that there are no more unprocessed events/requests/futures when polling is disabled. `HPX_NODISCARD` helps if one actually remembers to try to enable polling, but doesn't help one hasn't even remembered to do that. In those cases the application just hangs without any indication of what is wrong. These checks are intended to help debugging those situations.

The reason for the currently vague "registered at least once" is that the polling could be enabled on *any* pool, and I didn't want to look through every pool to see if at least one of them has polling enabled. This may be necessary in the end, but this should already catch most cases of not enabling polling.

I'm at the same time benchmarking the overhead of enabling polling (both CUDA and MPI). Another possibility is to simply enable polling whenever it's needed and leave it enabled. Perhaps simply using the normal `register_polling` functions would be easier than using the RAII wrappers. The RAII wrappers can also not be nested now. I'm leaving this PR as it is, but ideas on how to make this slightly less error prone are welcome.

### DIFF
--- a/libs/full/async_cuda/src/cuda_future.cpp
+++ b/libs/full/async_cuda/src/cuda_future.cpp
@@ -21,6 +21,7 @@
 #include <hpx/runtime_fwd.hpp>
 #include <hpx/threading_base/thread_helpers.hpp>
 
+#include <atomic>
 #include <cstddef>
 #include <memory>
 #include <string>
@@ -30,9 +31,6 @@
 #include <hpx/async_cuda/custom_gpu_api.hpp>
 
 namespace hpx { namespace cuda { namespace experimental { namespace detail {
-
-    HPX_EXPORT void register_polling(hpx::threads::thread_pool_base&);
-    HPX_EXPORT void unregister_polling(hpx::threads::thread_pool_base&);
 
     mutex_type& get_list_mtx()
     {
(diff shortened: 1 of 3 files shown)


libs/full/async_cuda/src/cuda_future.cpp
```diff
@@ -80,10 +77,17 @@ namespace hpx { namespace cuda { namespace experimental { namespace detail {
             hpx::util::internal_allocator<>{}, stream);
     }
 
+#if defined(HPX_DEBUG)
+    static std::atomic<std::size_t> register_polling_count{0};
```
hkaiser, line 81: In order to avoid initialization sequencing issues, would it make sense to wrap this as a function-local static instead? This would also have the benefit of not having any impact if not used.

libs/full/async_cuda/src/cuda_future.cpp
```diff
@@ -265,9 +269,22 @@ namespace hpx { namespace cuda { namespace experimental { namespace detail {
     // -------------------------------------------------------------
     void unregister_polling(hpx::threads::thread_pool_base& pool)
     {
+#if defined(HPX_DEBUG)
+        {
+            std::unique_lock<hpx::cuda::experimental::detail::mutex_type> lk(
+                detail::get_list_mtx());
+            HPX_ASSERT_MSG(get_event_queue().size_approx() == 0,
```
hkaiser, line 276: We shouldn't have locks active when calling ASSERTs as this will cause the lock registration to kick in otherwise.

libs/full/async_cuda/src/cuda_future.cpp
msimberg: Good point. I suppose it's unlikely to be used like that, but better safe than sorry.

libs/full/async_cuda/src/cuda_future.cpp
msimberg: Good catch! I've updated this to avoid that.

hkaiser (dismissed):
LGTM, thanks!

aurianer (approved):
LGTM thanks, just one remaining [CI failure](https://cdash.cscs.ch/viewBuildError.php?buildid=137964)

msimberg:
> LGTM thanks, just one remaining [CI failure](https://cdash.cscs.ch/viewBuildError.php?buildid=137964)

Thanks! Fixed.

msimberg:
I wasn't actually incrementing the counters. Small detail... This should now be ready to go.

PR #5080, opened 2020-12-09
1.6.0 release notes

msimberg:
Getting the process started for a 1.6.0 release.

The main question I have is how we want to deal with our compatibility options. There are a number of them that we "should" have turned off in the previous release (i.e. according to our release procedure). However, I consider a feature off by default the same as removed (pre-built packages are unlikely to turn those on just for users who need them).

My preference would be to outright remove the options and code guarded by the options (how many users' code will we actually break?). The other option is to properly add actual deprecation warnings to the removed features and remove the compatibility options. If a user has had a warning about a deprecated feature in a previous release we don't need to feel so bad about removing them quickly.

### DIFF
--- a/docs/sphinx/releases/whats_new_1_6_0.rst
+++ b/docs/sphinx/releases/whats_new_1_6_0.rst
@@ -1,0 +1,87 @@
+..
+    Copyright (C) 2007-2020 Hartmut Kaiser
+    Copyright (C)      2020 ETH Zurich
+
+    SPDX-License-Identifier: BSL-1.0
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+.. _hpx_1_6_0:
+
+==================
+|hpx| V1.6.0 (TBD)
+==================
+
+General changes
+===============
+
+This release continues the focus on C++20 conformance with multiple new
+algorithms adapted to be C++20 conformant and becoming customization point
+objects (CPOs). We have also added experimental support for HIP, allowing
+previous CUDA features to now be compiled with hipcc and run on AMD GPUs.
+
+* The following algorithms have been adapted to be C++20 conformant:
+  ``adjacent_find``, ``includes``, ``inplace_merge``, ``is_heap``,
+  ``is_heap_until``, ``is_partitioned``, ``is_sorted``, ``is_sorted_until``,
+  ``merge``, ``set_difference``, ``set_intersection``,
+  ``set_symmetric_difference``, ``set_union``.
+* Experimental HIP support can be used by compiling |hpx| with ``hipcc``. All
+  CUDA functionality in |hpx| can now be used with HIP. The HIP functionality is
+  for the time being exposed through the same API as the CUDA functionality,
+  i.e. no changes are required in user code. The CUDA, and now HIP,
+  functionality is in the ``hpx::cuda`` namespace.
+* We have added ``partial_sort`` based on Francisco Tapia's implementation.
+* ``hpx::init`` and ``hpx::start`` gained new overloads taking an
+  ``hpx::init_params`` struct in 1.5.0. All overloads not taking an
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1611,7 +1611,7 @@ endif()
 # HPX_WITH_ACTION_BASE_COMPATIBILITY: introduced in V1.4.0
 hpx_option(
   HPX_WITH_ACTION_BASE_COMPATIBILITY BOOL
-  "Enable deprecated action bases (default: ON)" ON ADVANCED
+  "Enable deprecated action bases (default: OFF)" OFF ADVANCED
```
msimberg, line 1614: Note to self. This is tentative.

msimberg:
For reference, the compatibility options we currently have are here: https://github.com/STEllAR-GROUP/hpx/blob/74a5b7935778c344ce3976a0c05438b729415646/CMakeLists.txt#L1599-L1668, and here: https://github.com/STEllAR-GROUP/hpx/blob/74a5b7935778c344ce3976a0c05438b729415646/CMakeLists.txt#L2193-L2257.

msimberg:
By the way, doing deprecations without the CMake options would help avoid situations like what CI shows right now (failed builds...) because we'd have a much smaller space of configurations to test.

hkaiser:
> By the way, doing deprecations without the CMake options would help avoid situations like what CI shows right now (failed builds...) because we'd have a much smaller space of configurations to test.

FWIW, some of the cmake options you listed above (`HPX_WITH_TUPLE_RVALUE_SWAP`, `HPX_PROGRAM_OPTIONS_WITH_BOOST_PROGRAM_OPTIONS_COMPATIBILITY`, etc.) should stay as those give options to the user to adapt to their environment. Not all of the options are deprecation related, I think.

msimberg:
Agreed on `HPX_WITH_TUPLE_RVALUE_SWAP`, I was a bit overzealous in marking the options. However, don't agree on the Boost program options compatibility. One of the main reasons to copy that over to HPX was so that we could eventually remove at least that particular Boost dependency. Essentially every CMake option with `_COMPATIBILITY` at the end is due for removal sooner or later.

hkaiser:
> Agreed on `HPX_WITH_TUPLE_RVALUE_SWAP`, I was a bit overzealous in marking the options. However, don't agree on the Boost program options compatibility. One of the main reasons to copy that over to HPX was so that we could eventually remove at least that particular Boost dependency. Essentially every CMake option with `_COMPATIBILITY` at the end is due for removal sooner or later.

The options that have `_COMPATIBILITY` in them are different by design from options that enable/disable Boost program_options or filesystem, for instance. The compatibility options help people to suppress warnings for a time if they are not ready to update their codes for some reason. The options related to Boost libraries help adapting to the user's environment. For instance, if the user already relies on Boost filesystem, then we want to enable HPX to use the same library to ensure compatibility. I would even argue, that we might not want to remove the Boost related options altogether at any point soon.

CMakeLists.txt
hkaiser: This is fine, I think.

msimberg:
Perhaps `HPX_FILESYSTEM_WITH_BOOST_FILESYSTEM_COMPATIBILITY` is wrongly named in that case. Currently we *need* this. We cannot remove this without requireing C++17 support. The same is not true for program options. Also, if we use std or Boost filesystem internally should not matter for users. If they wish to use Boost filesystem they are free to do so independent of our decisions. The same applies to program options *to some extent*. Removing Boost program options support means that users can't pass Boost options descriptions to init, but they can still use it to parse options outside of init.

CMakeLists.txt
msimberg: This one is perhaps less contentious, but my comment referred to all of the options below as well (which clearly do seem contentious!).

msimberg: Closing this as resolved. We can discuss on #5084 if needed.

msimberg:
Merging to have the notes in the 1.6.0-rc1 documentation. Additions and changes are still welcome and encouraged after merging though.

PR #5027, opened 2020-10-24
Lessen constraints on certain algorithm arguments

hkaiser:
- some sequential algorithms (copy, etc.) are fully functional if - when used sequentially - the destination is just an output iterator
- fixing is_output_iterator
- fixing `transform_reduce` implementation to properly handle constant input sequences
- flyby: minor formatting changes

PR #5082, opened 2020-12-10
Relax iterator requirements for sequential versions of set algorithms

hkaiser:
This is a partial duplicate of #5027 and has been merged over there.


hkaiser:
This is a partial duplicate of #5027 and has been merged over there.

PR #5090, opened 2020-12-23
Adopt C++20 std::endian

K-ballo:


hkaiser (approved):
Nice one! LGTM, thanks!

PR #5094, opened 2021-01-04
Attempt to remove macOS workaround for GH actions environment

msimberg:


msimberg:
This is now "working" in that the original workaround is gone, another workaround has been put in place, and the runner is extremely slow (https://github.com/actions/virtual-environments/issues/1336). However, at the moment there is not much else we can do about this (except try another provider again), so this is done for the moment.

hkaiser (approved):
LGTM, thanks!

PR #5097, opened 2021-01-05
Remove hpxMP support

### LINKED ISSUE #5087: Remove hpxMP support
I suggest we remove the hpxMP integration from our build system. It should still be possible to build it separately (perhaps not with the latest HPX, but still) if there is interest in using it. Objections?

msimberg:
Fixes #5087.


hkaiser (approved):
LGTM, thanks!

PR #5102, opened 2021-01-08
Untie deprecated thread enums from the CMake option

msimberg:
These were not meant to be disabled yet, and I'd prefer not to have a CMake option at all. They will simply be available but deprecated until they're removed.


aurianer (approved):
LGTM thanks

PR #5103, opened 2021-01-08
Remove default runtime mode argument from start/init overloads

msimberg:
These overloads may take precedence over the non-deprecated overloads
using `hpx::init_params` when the mode has a default value. Since these
overloads are deprecated, we don't want them to take precedence.


hkaiser (approved):
LGTM, thanks!

PR #5104, opened 2021-01-08
Fix regeneration of two files that trigger unnecessary rebuilds

biddisco:
The modules_enabled.hpp file was being wiped by a cleanup step
in the module creation script.

Architecture detection generated a file on every configure
step and only needs to be done once, so cache the result fo
reuse on subsequent runs.

### DIFF
--- a/cmake/HPX_AddModule.cmake
+++ b/cmake/HPX_AddModule.cmake
@@ -232,7 +232,8 @@ function(add_hpx_module libname modulename)
        ${CMAKE_CURRENT_BINARY_DIR}/include_compatibility/*.hpp
   )
   list(REMOVE_ITEM zombie_generated_headers ${generated_headers}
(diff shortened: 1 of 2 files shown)


biddisco:
Note that the architecture detection scripts are quite old and more recent cmake versions have upgraded tthe behaviour of [CMAKE_HOST_SYSTEM_PROCESSOR](https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html#variable:CMAKE_HOST_SYSTEM_PROCESSOR) and it might be worth removing the detection code.
I tested the `CMAKE_HOST_SYSTEM_PROCESSOR` on raspberry pi (arm), fujitsu A64FX, 
power, intel and AMD x64, and a couple of other nodes on our test machine and it seems to give sensible outputs.

hkaiser (approved):
LGTM, thanks!

msimberg:
Thanks @biddisco for fixing this. That was a very good catch! I've just formatted the CMake files since we require that in CI now as well.

PR #5106, opened 2021-01-12
Change initialization of property CPOs to satisfy older nvcc versions

msimberg:
Probably fixes compilation on older nvcc's. Not regression test (yet) as I couldn't reproduce this with gcc 8.3 and cuda 10.2. Likely only happens on older cuda versions. Same problem as in #4912.

@G-071 I'd appreciate you trying this out!


hkaiser (approved):
LGTM, thanks!

PR #5109, opened 2021-01-13
Remove default runtime mode argument from init overload, again

msimberg:
Apologies, I was off by one in #5103. On top of that `hpx::start` did not need any change...


hkaiser (approved):
LGTM, thanks!

hkaiser:
> On top of that `hpx::start` did not need any change...

Why's that? I always assumed for the two APIs to have similar overloads.

msimberg:
> > On top of that `hpx::start` did not need any change...
> 
> Why's that? I always assumed for the two APIs to have similar overloads.

Good question. I assumed so too, which is why I made the initial change on both sides instinctively. It must've been overlooked at some point...

PR #5114, opened 2021-01-15
Fix 5112 - make sure libatomic is used when needed

### LINKED ISSUE #5112: libatomic problems causing unexpected fails
Applications may crash on startup with the following message
```
  what():  big_boot_barrier has not been created yet: HPX(internal_server_error)
```
upon investigation, it turns out that the following exception is being thrown during initialization/startup
https://github.com/STEllAR-GROUP/hpx/blob/1527f3ba8055f795f701ddf89368f4a24199f79e/libs/full/threadmanager/src/threadmanager.cpp#L322-L329

which in turn causes the runtime destructor to be called, but the runtime has not yet started, so it throws the error that we actually see about the big boot barrier not being started. So the first exception needs to be caught somewhere and displayed, before the stack unwinds and throws the second.

Looking deeper, the problem is being caused by the following check failing during cmake configuration. 
```
-- Looking for __atomic_fetch_add_4 in atomic
-- Looking for __atomic_fetch_add_4 in atomic - not found
```
which in turn means that `HPX_WITH_CXX11_ATOMIC_128BIT` is not set. If I manually run the cmake feature check withoout linking libatomic, we see link errors
```
oryx2:~/build/hpx-master$ g++ ~/src/hpx/cmake/tests/cxx11_std_atomic_128bit.cpp
/usr/bin/ld: /tmp/ccduIFup.o: in function `std::atomic<uint128_type>::store(uint128_type, std::memory_order)':
cxx11_std_atomic_128bit.cpp:(.text._ZNSt6atomicI12uint128_typeE5storeES0_St12memory_order[_ZNSt6atomicI12uint128_typeE5storeES0_St12memory_order]+0x66): undefined reference to `__atomic_store_16'
/usr/bin/ld: /tmp/ccduIFup.o: in function `std::atomic<uint128_type>::load(std::memory_order) const':
cxx11_std_atomic_128bit.cpp:(.text._ZNKSt6atomicI12uint128_typeE4loadESt12memory_order[_ZNKSt6atomicI12uint128_typeE4loadESt12memory_order]+0x40): undefined reference to `__atomic_load_16'
collect2: error: ld returned 1 exit status
```
but if I add `-latomic` all is well.
```
oryx2:~/build/hpx-master$ g++ ~/src/hpx/cmake/tests/cxx11_std_atomic_128bit.cpp -latomic
oryx2:~/build/hpx-master$ <links correctly>
```

I have only run into this problem on my new laptop using gcc 10.2.0  

I'm not quite sure what the correct fix for the atomic problem is.

biddisco:
Fixes #5112 
We test a link to atomics and set the `HPX_WITH_CXX11_ATOMIC` if ok, but if it fails, we add lib atomic and try again. Then repeat the process with 128 bit atomics. If either needs lib atomic, then `HPX_CXX11_STD_ATOMIC_LIBRARIES` is set to `atomic` and we add that as a dependency to the main hpx_core library.

This fixes my problems (for now at least)

Edit : The problem seems to have been that the check_function_exists call was not delivering as expected.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1784,8 +1784,8 @@ if(NOT WIN32)
   # ############################################################################
   if(NOT MSVC)
     hpx_add_compile_flag_if_available(-pthread)
-    if(HPX_HAVE_LIBATOMIC)
-      target_link_libraries(hpx_base_libraries INTERFACE atomic)
+    if(HPX_CXX11_STD_ATOMIC_LIBRARIES)
+      target_link_libraries(hpx_base_libraries INTERFACE ${HPX_CXX11_STD_ATOMIC_LIBRARIES})
     endif()
   endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1785,7 +1785,9 @@ if(NOT WIN32)
   if(NOT MSVC)
     hpx_add_compile_flag_if_available(-pthread)
     if(HPX_CXX11_STD_ATOMIC_LIBRARIES)
-      target_link_libraries(hpx_base_libraries INTERFACE ${HPX_CXX11_STD_ATOMIC_LIBRARIES})
+      target_link_libraries(
+        hpx_base_libraries INTERFACE ${HPX_CXX11_STD_ATOMIC_LIBRARIES}
+      )
     endif()
   endif()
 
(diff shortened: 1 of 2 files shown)


hkaiser (dismissed):
LGTM, thanks!

biddisco:
Don't merge this yet. I seem to still have some libatomic related issues on another build. I need to check what isn't fixed.

msimberg:
@biddisco is this ready to go in?

PR #5116, opened 2021-01-20
Fix compilation with timer pool off

msimberg:


msimberg:
Merging right away since we do not test this case.

PR #5119, opened 2021-01-21
Use sequential_find and friends from separate detail header

gonidelis:
This makes `hpx::find` and friends to use the `sequential_find` functions that are located under the detail directory and thus removes the ones placed in the same header.

flyby: It adds the projection utility in the overloads under the separate header.

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
@@ -1,4 +1,5 @@
 //  Copyright (c) 2020 Hartmut Kaiser
+//  Copyright (c) 2021 Giannis Gonidelis
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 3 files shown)


libs/parallelism/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
```diff
@@ -27,13 +28,14 @@ namespace hpx { namespace parallel { inline namespace v1 { namespace detail {
     }
 
     // provide implementation of std::find_if supporting iterators/sentinels
-    template <typename Iterator, typename Sentinel, typename Pred>
+    template <typename Iterator, typename Sentinel, typename Pred,
+        typename Proj>
     inline constexpr Iterator sequential_find_if(
-        Iterator first, Sentinel last, Pred pred)
+        Iterator first, Sentinel last, Pred pred, Proj&& proj)
```
hkaiser, line 34: Could you make this consistent, please? Either pass both function objects by value of both using forwarding references (preferred).

libs/parallelism/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
```diff
@@ -42,13 +44,14 @@ namespace hpx { namespace parallel { inline namespace v1 { namespace detail {
     }
 
     // provide implementation of std::find_if supporting iterators/sentinels
-    template <typename Iterator, typename Sentinel, typename Pred>
+    template <typename Iterator, typename Sentinel, typename Pred,
+        typename Proj>
     inline constexpr Iterator sequential_find_if_not(
-        Iterator first, Sentinel last, Pred pred)
+        Iterator first, Sentinel last, Pred pred, Proj&& proj)
```
hkaiser, line 50: Same here, please make this consistent.

hkaiser (commented):
Please fix all other uses of these functions as well (see failing tests).

libs/parallelism/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
gonidelis: Sure. I am working on it right now ;)

gonidelis:
There are still changes to be made. Please don't review yet.

hkaiser (approved):
LGTM, thanks!

PR #5121, opened 2021-01-22
Working around hwloc limitation on certain platforms

### LINKED ISSUE #5115: Don't expect hwloc to always report Cores
Hello

I am the main hwloc developer. We recently got a bug report through FreeBSD because hwloc returned 0 cores, causing HPX tests to fail. https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=252643 and open-mpi/hwloc#442

This was indeed a hwloc bug that I am going to fix, but I think HPX shouldn't fail in this case. There are platforms where we cannot detect the cores. Of course it doesn't happen on Linux. However some operating systems such as *BSD don't have any native topology information. In these cases, hwloc tries alternative detection methods such as x86 CPUID assembly, but this isn't always supported.

HPX should have a fallback that uses PU instead of Cores in such cases. PU are guaranteed to always exist in hwloc.

Note that I didn't dig into the HPX code at all, I don't know if the failure is in the core HPX or in some tests mentioned in the freebsd bug report.

Thanks

hkaiser:
- on certain platforms (like BSD), hwloc doesn't correctly report the number of cores, use number of PUs instead

Fixes #5115

@bgoglin I have no way of actually testing this. Would you be able to verify if this fixes the issue by any chance?

### DIFF
--- a/.jenkins/lsu/env-clang-9.sh
+++ b/.jenkins/lsu/env-clang-9.sh
@@ -18,3 +18,7 @@ configure_extra_options+=" -DHPX_WITH_MALLOC=system"
 configure_extra_options+=" -DHPX_WITH_COMPILER_WARNINGS=ON"
 configure_extra_options+=" -DHPX_WITH_COMPILER_WARNINGS_AS_ERRORS=ON"
 configure_extra_options+=" -DHPX_WITH_PARCELPORT_MPI=ON"
+
+# Make sure HWLOC does not report 'cores'. This is purely an option to enable
+# testing the topology code under conditions close to those on FreeBSD.
+configure_extra_options+=" -DHPX_TOPOLOGY_WITH_ADDITIONAL_HWLOC_TESTING=ON"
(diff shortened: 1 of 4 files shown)


biddisco (approved):
LGTM

bgoglin:
@hkaiser github didn't notify me of this thread, sorry for delay. It's not easy for me to test this. But you should be able to test it by adding hwloc_topology_set_type_filter(topology, HWLOC_OBJ_CORE, HWLOC_TYPE_FILTER_KEEP_NONE) between hwloc_topology_init() and hwloc_topology_load() to tell hwloc that you don't want any core object.

hkaiser:
>  @hkaiser github didn't notify me of this thread, sorry for delay. It's not easy for me to test this. But you should be able to test it by adding hwloc_topology_set_type_filter(topology, HWLOC_OBJ_CORE, HWLOC_TYPE_FILTER_KEEP_NONE) between hwloc_topology_init() and hwloc_topology_load() to tell hwloc that you don't want any core object.

Thanks Brice. We'll set up a builder that is doing this.

hkaiser:
FWIW, the `jenkins/lsu/clang-9` builders should pick up the new option.

msimberg:
Thanks @hkaiser! Looks like the new builder is happy as well. Let's go ahead with this.

PR #5123, opened 2021-01-22
Reduce vector sizes in partial sort benchmarks when running in debug mode

msimberg:
The benchmarks otherwise take too long to run.


hkaiser (approved):
Good move, thanks!

PR #5128, opened 2021-01-26
Fixing environment handling for FreeBSD

### LINKED ISSUE #5115: Don't expect hwloc to always report Cores

hkaiser:
- flyby: disable memory perf counters for FreeBSD

Fixes #5115

### DIFF
--- a/components/performance_counters/memory/CMakeLists.txt
+++ b/components/performance_counters/memory/CMakeLists.txt
@@ -8,6 +8,11 @@ if(NOT HPX_WITH_DISTRIBUTED_RUNTIME)
   return()
 endif()
 
+# The memory counters are not supported on FreeBSD
+if(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
+  return()
+endif()
+
 set(HPX_COMPONENTS
     ${HPX_COMPONENTS} memory
     CACHE INTERNAL "list of HPX components"
--- a/components/process/include/hpx/components/process/util/posix/initializers/inherit_env.hpp
+++ b/components/process/include/hpx/components/process/util/posix/initializers/inherit_env.hpp
@@ -11,17 +11,10 @@
 #pragma once
 
 #include <hpx/config.hpp>
+#include <hpx/debugging/environ.hpp>
 
 #if !defined(HPX_WINDOWS)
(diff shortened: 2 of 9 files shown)


zao:
```
/home/zao/stellar/hpx/libs/core/debugging/src/print.cpp:251:17: error: use of undeclared identifier 'gethostname'
                gethostname(hostname_, std::size_t(12));
                ^
/home/zao/stellar/hpx/libs/core/debugging/src/print.cpp:261:26: error: cannot initialize a variable of type 'const char **' with an lvalue of type 'char **'
            char const** env = freebsd_environ;
                         ^     ~~~~~~~~~~~~~~~
2 errors generated.
```

hkaiser:
@zao should be fine now, thanks!

hkaiser:
@zao reported that this PR works for him on FreeBSD when using a patched hwloc 2.4 : https://gist.github.com/zao/c826f73531fccb5029061e06df9b94de (i.e. the hwloc patch that fixes the issue @bgoglin mentioned here: https://github.com/STEllAR-GROUP/hpx/issues/5115)

hkaiser:
@msimberg once the tests have cycled, this is good to go, I believe (I have fixed the cyclic dependency problem).

msimberg:
> @msimberg once the tests have cycled this is good to go, I believe (I have fixed the cyclic dependency problem).

Just the cmake_format check that's unhappy... Please feel free to merge without waiting for CI after that's updated.

PR #5136, opened 2021-01-29
Make sure codespell checks run on CircleCI

msimberg:
I *think* this fixes it. The CircleCI build will show if it's correct. I'll check if it does the right thing on master once merged.


hkaiser (approved):
LGTM, thanks!

PR #5138, opened 2021-01-29
Add program options variable map to resource partitioner init

biddisco:
Passing the variables map to the init callback for the resource
partitioner makes passing of options like "--pool-threads" more
straightforward than some of the alternatives such as parsing
the command-line twice.

Some of the examples/tests that use the resource partitioner
had copy/pasted a bunch of init from some very early tests and
some cleanup/simplication has been made, but not to all tests
so far.


hkaiser (approved):
LGTM, thanks (except for the west `const`)!

hkaiser:
@biddisco Thanks a lot! This is good to go.

PR #5139, opened 2021-01-30
Fix path problem in pre-commit hook, add summary commit line

biddisco:
To use this file execute
cp tools/pre-commit .git/hooks/pre-commit

The pre-commit hook failed to work properly if a file in the
root dir was changed as the relative path was passed an empty
string. Fix that by quoting the path.

Add an extra line to the output to summarize the file changes
into a new commit line so that after fixing format errors
it is easy to commit the fixed files.


hkaiser (approved):
LGTM, thanks!

PR #5140, opened 2021-01-30
Scheduler mode

biddisco:
I had problems with background idle flags being set even when I had set them to off - this PR fixes a wrong use of add_scheduler_mode - it should be set_scheduler_mode on initialization.

I also protected a couple of places that don't need to be run when the idle flag is unset.


hkaiser (approved):
LGTM, thanks!

msimberg:
Good catch!

PR #5141, opened 2021-01-30
Fix a link error when async_cuda is enabled, but not cuda_compute

biddisco:
Just replaces a couple of CUDA statements with ASYNC_CUDA in cmake - not 100% tested, but dashboard should reveal if it causes trouble

### DIFF
--- a/cmake/HPX_AddExecutable.cmake
+++ b/cmake/HPX_AddExecutable.cmake
@@ -194,10 +194,7 @@ function(add_hpx_executable name)
     endforeach()
   endif()
 
-  if(HPX_WITH_CUDA
-     AND (HPX_WITH_CUDA_COMPUTE OR HPX_WITH_ASYNC_CUDA)
-     AND NOT HPX_WITH_CUDA_CLANG
-  )
+  if((HPX_WITH_CUDA_COMPUTE OR HPX_WITH_ASYNC_CUDA) AND NOT HPX_WITH_CUDA_CLANG)
     cuda_add_executable(
       ${name} ${${name}_SOURCES} ${${name}_HEADERS} ${${name}_AUXILIARY}
     )
--- a/libs/full/async_cuda/CMakeLists.txt
+++ b/libs/full/async_cuda/CMakeLists.txt
@@ -51,7 +51,7 @@ endif()
 
 if(HPX_WITH_HIP AND TARGET roc::hipblas)
   set(async_cuda_extra_deps ${async_cuda_extra_deps} roc::hipblas)
-elseif(HPX_WITH_CUDA AND TARGET Cuda::cuda)
+elseif(HPX_WITH_ASYNC_CUDA AND TARGET Cuda::cuda)
   set(async_cuda_extra_deps ${async_cuda_extra_deps} Cuda::cuda
                             ${CUDA_CUBLAS_LIBRARIES}
   )
@@ -71,3 +71,6 @@ add_hpx_module(
   DEPENDENCIES hpx_core hpx_parallelism ${async_cuda_extra_deps}
   CMAKE_SUBDIRS examples tests
 )
+
+# if(HPX_WITH_ASYNC_CUDA AND TARGET Cuda::cuda)
+# CUDA_ADD_CUBLAS_TO_TARGET(hpx_async_cuda) endif()


hkaiser (commented):
Does `HPX_WITH_CUDA` imply `HPX_WITH_CUDA_COMPUTE`? Or is `HPX_WITH_CUDA` implicitly defined whenever one of `HPX_WITH_CUDA_COMPUTE` or `HPX_WITH_ASYNC_CUDA` is defined?

I think it should be the latter in which case the fix proposed here would break using cuda compute, IIUC.

biddisco:
> Does `HPX_WITH_CUDA` imply `HPX_WITH_CUDA_COMPUTE`? Or is `HPX_WITH_CUDA` implicitly defined whenever one of `HPX_WITH_CUDA_COMPUTE` or `HPX_WITH_ASYNC_CUDA` is defined?
> 
> I think it should be the latter in which case the fix proposed here would break using cuda compute, IIUC.

if `HPX_WITH_CUDA` is set - then both `HPX_WITH_ASYNC_CUDA` and `HPX_WITH_CUDA_COMPUTE` are turned on. If `HPX_WITH_CUDA_COMPUTE` is enabled, then `HPX_WITH_ASYNC_CUDA` is also enabled, but if ONLY `HPX_WITH_ASYNC_CUDA` is enabled, then the other two are not set. I mostly use `HPX_WITH_ASYNC_CUDA` and do not use the `CUDA_COMPUTE` - (I wanted to remove `HPX_WITH_CUDA` completely and only have the other two, but this was previously rejected).

msimberg:
When adding HIP support we decided to keep the `HPX_WITH_CUDA` option to distinguish it from `HPX_WITH_HIP`. These two options control whether support for either of these is generally enabled. The two other options, `HPX_WITH_ASYNC_CUDA` and `HPX_WITH_COMPUTE_CUDA`, are dependent on either of those two being set (`HPX_WITH_HIP` is set automatically when hipcc is the compiler) and control whether the two individual modules are enabled.

I think what you probably want to do is set `HPX_WITH_CUDA=ON` and `HPX_WITH_COMPUTE_CUDA=OFF` (with the exception that we seem to disagree on which flags you should be required to set).

Also, in my opinion, the fact that this is still so confusing tells me that the `HPX_WITH_{ASYNC,COMPUTE}_CUDA` options are not worth it, or that they really badly need to be renamed (as they now apply to both CUDA and HIP).

Finally, this breaks the HIP configuration.

biddisco:
OK. I obviously missed out of the change to support HIP. I'm happy to have 
`HPX_WITH_CUDA` - implies `HPX_WITH_ASYNC_CUDA` is enabled and we no longer need a specific `HPX_WITH_ASYNC_CUDA` cmake flag any more since `WITH_XXX` will imply `ASYNC_XXX` - where `XXX=CUDA / HIP`  
As long as we make sure that turning on `HPX_WITH_CUDA` does not forcefully turn on `HPX_WITH_COMPUTE_CUDA` - (because that's what it used to do). 

I will modify this PR and resubmit later.

msimberg:
> As long as we make sure that turning on `HPX_WITH_CUDA` does not forcefully turn on `HPX_WITH_COMPUTE_CUDA` - (because that's what it used to do).

Depends what you mean by forcefully. It enables `HPX_WITH_COMPUTE_CUDA` by default, but one is free to disable `HPX_WITH_COMPUTE_CUDA` explicitly.

hkaiser:
> > As long as we make sure that turning on `HPX_WITH_CUDA` does not forcefully turn on `HPX_WITH_COMPUTE_CUDA` - (because that's what it used to do).
> 
> Depends what you mean by forcefully. It enables `HPX_WITH_COMPUTE_CUDA` by default, but one is free to disable `HPX_WITH_COMPUTE_CUDA` explicitly.

@msimberg: could you please document here which flags need to be set to what for all of the combinations (CUDA on/off, HIP on/off, async CUDA on/off, etc.)? I lost track of that as well.

msimberg:
> @msimberg: could you please document here which flags need to be set to what for all of the combinations (CUDA on/off, HIP on/off, async CUDA on/off, etc.)? I lost track of that as well.

Apparently I can't. It doesn't behave the way I been trying to describe above. I apologize sincerely.

That said, what I've been trying to describe is how I would still like it to behave, given that we don't just remove the module-specific options, which would be my first choice.

Could we first try to agree on how we think the options should behave?

hkaiser:
> Could we first try to agree on how we think the options should behave?

@msimberg: That's what I was trying to get at by asking you how you think the options should behave. I think that `HPX_WITH_CUDA` and `HPX_WITH_HIP` should be the main options and should enable _all_ of the CUDA/HIP support. If we need to have more fine-grain control, those should be separate options that would have to be disabled explicitly (i.e. `HPX_WITH_COMPUTE_CUDA=OFF` etc.). I believe that is what you were suggesting earlier as well.  

Getting rid of the module-specific options would be fine with me as well. @biddisco if you have some additional rationale why those shouldn't/couldn't be removed, please let us know. I might not have the full picture.

This is however not set in stone for me and I'd be happy to change my mind, if there are better options.

msimberg:
> @msimberg: That's what I was trying to get at by asking you how you think the options should behave. I think that `HPX_WITH_CUDA` and `HPX_WITH_HIP` should be the main options and should enable _all_ of the CUDA/HIP support. If we need to have more fine-grain control, those should be separate options that would have to be disabled explicitly (i.e. `HPX_WITH_COMPUTE_CUDA=OFF` etc.). I believe that is what you were suggesting earlier as well.

Sorry. Yes, this sounds like how I expected it to behave currently as well. The discrepancy right now seems to be that if e.g. `HPX_WITH_CUDA=ON` then it forces `HPX_WITH_COMPUTE_CUDA=ON` even if the user sets `HPX_WITH_COMPUTE_CUDA=OFF` explicitly. I *think* this is the only discrepancy...

msimberg:
Closing in favour of #5145.

PR #5142, opened 2021-02-01
Store annotated_function annotations as std::strings

### LINKED ISSUE #5126: Use std::string for task annotations
Currently we use `char const*` strings for task annotations. This makes it impossible to use dynamically generated tasked names because the strings might be gone by the time they are used.

The main reason I can see for not using `std::string`s is to avoid dynamic allocation for task descriptions and annotations. However, I would guess that this is insignificant compared to overheads caused by having to e.g. log task data in the first place, and other overheads related to our tasks. @khuck does your experience tell us that we should in fact stick to C-style static strings for the annotations or do you see no problem with starting to use `std::string`.

@khuck a somewhat related question that came as a request from a user: what facilities does APEX have for storing metadata related to tasks and to write these to OTF2 files? It's related to this issue in the sense that one might want to write metadata to the task name dynamically, but if there were a separate place to put that metadata it might be more ergonomic than having to stick it in the task name. If there is such a facility in APEX it's of course another question if and how to expose that to users.

msimberg:
Storing the annotations in the `annotated_function` wrapper ensures that the annotations live at least until its `operator()` is called, by which point the annotation should already have been set to APEX.

The `annotation_check` test is updated to use dynamically allocated annotations which go out of scope before the tasks have completed (and it fails before the second commit in this PR).

Fixes #5126.

### DIFF
--- a/libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
+++ b/libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
@@ -33,11 +33,11 @@ namespace hpx { namespace util {
 #if defined(HPX_HAVE_THREAD_DESCRIPTION)
     ///////////////////////////////////////////////////////////////////////////
 #if defined(HPX_COMPUTE_DEVICE_CODE)
-    struct annotate_function
+    struct HPX_NODISCARD annotate_function
     {
         HPX_NON_COPYABLE(annotate_function);
 
-        explicit annotate_function(std::string) {}
+        explicit annotate_function(char const*) {}
         template <typename F>
         explicit HPX_HOST_DEVICE annotate_function(F&&)
         {
@@ -47,13 +47,12 @@ namespace hpx { namespace util {
         HPX_HOST_DEVICE ~annotate_function() {}
     };
 #elif HPX_HAVE_ITTNOTIFY != 0
-    struct annotate_function
+    struct HPX_NODISCARD annotate_function
     {
         HPX_NON_COPYABLE(annotate_function);
 
-        explicit annotate_function(std::string name)
-          : name_(std::move(name))
-          , task_(thread_domain_, hpx::util::itt::string_handle(name.c_str()))
+        explicit annotate_function(char const* name)
+          : task_(thread_domain_, hpx::util::itt::string_handle(name))
         {
         }
         template <typename F>
@@ -65,26 +64,24 @@ namespace hpx { namespace util {
         }
 
     private:
-        std::string name_;
         hpx::util::itt::thread_domain thread_domain_;
         hpx::util::itt::task task_;
     };
 #else
-    struct annotate_function
+    struct HPX_NODISCARD annotate_function
     {
         HPX_NON_COPYABLE(annotate_function);
 
-        explicit annotate_function(std::string name)
-          : name_(std::move(name))
-          , desc_(hpx::threads::get_self_ptr() ?
+        explicit annotate_function(char const* name)
+          : desc_(hpx::threads::get_self_ptr() ?
                     hpx::threads::set_thread_description(
-                        hpx::threads::get_self_id(), name_.c_str()) :
+                        hpx::threads::get_self_id(), name) :
                     nullptr)
         {
 #if defined(HPX_HAVE_APEX)
             /* update the task wrapper in APEX to use the specified name */
             threads::set_self_timer_data(external_timer::update_task(
-                threads::get_self_timer_data(), name_));
+                threads::get_self_timer_data(), std::string(name)));
 #endif
         }
 
@@ -111,7 +108,6 @@ namespace hpx { namespace util {
             }
         }
 
-        std::string name_;
         hpx::util::thread_description desc_;
     };
 #endif
@@ -121,19 +117,19 @@ namespace hpx { namespace util {
         struct annotated_function
         {
             annotated_function() noexcept
-              : name_("")
+              : name_(nullptr)
             {
             }
 
-            annotated_function(F const& f, std::string name)
+            annotated_function(F const& f, char const* name)
               : f_(f)
-              , name_(std::move(name))
+              , name_(name)
             {
             }
 
-            annotated_function(F&& f, std::string name)
+            annotated_function(F&& f, char const* name)
               : f_(std::move(f))
-              , name_(std::move(name))
+              , name_(name)
             {
             }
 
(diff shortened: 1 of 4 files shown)


libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
```diff
@@ -51,8 +51,9 @@ namespace hpx { namespace util {
     {
         HPX_NON_COPYABLE(annotate_function);
 
-        explicit annotate_function(char const* name)
-          : task_(thread_domain_, hpx::util::itt::string_handle(name))
+        explicit annotate_function(std::string const& name)
```
hkaiser, line 54: I think a `std::string name` might be better here as it would allow avoiding double allocations.

hkaiser (commented):
Is there a way to have both options? I.e. using a `std::string` only if needed?

libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
msimberg: Good idea. Will fix that.

msimberg:
> Is there a way to have both options? I.e. using a `std::string` only if needed?

Rhetorical question ;) I'll see what I can do.

msimberg:
I've pushed an update here which "fixes" the problem. A bit of explanation:

I thought that `annotated_function` would be sufficient to keep the strings alive long enough for APEX (keeping pointer stability with `shared_ptr`; heavy but sufficient for testing this). It turned out that this was not enough. That leaves two possibilities:

1. It's not enough to keep the strings in `annotated_function`. The `annotated_function` gets reset and the string freed just before the coroutine exits. @khuck do you know what is the last point in a task's lifetime that APEX accesses the HPX thread description? From what I could tell it's when the task is created, but I might have missed something which would invalidate the use of `annotated_function` for storing the annotations.
2. I just didn't do it right.

Given that I don't know currently know how long the strings need to be kept alive I've added thread local sets for storing them (only if a `std::string` is used, not for `char const*`). This is almost guarantees that they'll live long enough, modulo OS-threads exiting before a string is used (a global set would avoid this, but needs a lock).

msimberg:
After some more investigation I stand by the current state of this, using thread local sets for storing stable dynamically allocated strings. What I had ignored earlier was that tasks inherit the name of their parent task, meaning that a dynamically allocated description may need to live arbitrarily long (at least until the runtime is shut down). I think this is still the simplest solution. The main alternative I would still consider is just using `std::string`s as much as possible, but even that might not work well with e.g. ITT.

msimberg:
This should now be ready to go.

libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
```diff
@@ -33,11 +33,11 @@ namespace hpx { namespace util {
 #if defined(HPX_HAVE_THREAD_DESCRIPTION)
     ///////////////////////////////////////////////////////////////////////////
 #if defined(HPX_COMPUTE_DEVICE_CODE)
-    struct annotate_function
+    struct HPX_NODISCARD annotate_function
```
hkaiser, line 36: Good thinking!

libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
```diff
@@ -33,11 +33,11 @@ namespace hpx { namespace util {
...
     {
         HPX_NON_COPYABLE(annotate_function);
 
-        explicit annotate_function(char const*) {}
+        explicit annotate_function(std::string) {}
```
hkaiser, line 40: Shouldn't this be left as `char const*` to avoid allocations?

libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
```diff
@@ -214,7 +230,7 @@ namespace hpx { namespace util {
     ///
     /// \param function
     template <typename F>
-    F&& annotated_function(F&& f, char const* = nullptr)
+    F&& annotated_function(F&& f, std::string = "")
```
hkaiser, line 233: I think we should leave the overload that takes a `char const*` as otherwise the description will always cause an allocation (even if the string is discarded right away).

msimberg: Fair point! I'll change that.

libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
msimberg: Yes, and the change is an inconsistency in the first place. For `annotate_function` one can anyway use `std::string::c_str` since `annotate_function` is meant for the scope where it's defined.

libs/core/threading_base/src/annotated_function.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include <hpx/config.hpp>
+#if defined(HPX_HAVE_THREAD_DESCRIPTION) && defined(HPX_HAVE_APEX)
+#include <hpx/threading_base/annotated_function.hpp>
+
+#include <string>
+#include <unordered_set>
+#include <utility>
+
+namespace hpx { namespace util { namespace detail {
+    char const* store_function_annotation(std::string name)
```
hkaiser, line 16: Minor nitpick: if this function is always used with an rvalue (and seems it is), you might want to have it accept a `std::string&& name` instead. This is not a big deal however - it saves just one move operation.

hkaiser (dismissed):
LGTM, thanks!

libs/core/threading_base/src/annotated_function.cpp
msimberg: Another fair point. If we're going to do extra work to make the strings live longer, might as well do it as efficiently as possible. It's an internal API anyway so we can always change it later if there's another use case. I'll change that as well.

khuck:
@msimberg @hkaiser sorry, just getting caught up on this.  When APEX creates the task, it makes a local copy of the string.  As for parents, HPX maintains a pointer to the APEX "task description", which has the name copy in it, so I don't think APEX needs the name after task construction.  However, if `annotate_function` (not `annotated_function`!) is used, that causes a task to be renamed.  I don't know what happens in that case, other than APEX replaces the name of the task internally.

msimberg:
> @msimberg @hkaiser sorry, just getting caught up on this. When APEX creates the task, it makes a local copy of the string. As for parents, HPX maintains a pointer to the APEX "task description", which has the name copy in it, so I don't think APEX needs the name after task construction. However, if `annotate_function` (not `annotated_function`!) is used, that causes a task to be renamed. I don't know what happens in that case, other than APEX replaces the name of the task internally.

No worries, and thank you! Then that confirms that APEX itself is behaving well. My conclusion in the end was that HPX itself was simply passing stale C-string pointers to APEX, giving it no chance to do the right thing.

libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
```diff
@@ -218,6 +234,12 @@ namespace hpx { namespace util {
     {
         return std::forward<F>(f);
     }
+
+    template <typename F>
+    F&& annotated_function(F&& f, std::string const&)
```
msimberg, line 239: @hkaiser I added the `const&` here now. Thanks for pointing that out!

hkaiser (approved):
LGTM, thanks!

msimberg:
retest

PR #5144, opened 2021-02-04
AGAS module

hkaiser:
Moving more files to AGAS module


hkaiser:
retest

libs/full/runtime_local/include/hpx/runtime_local/runtime_local.hpp
```diff
@@ -388,6 +390,16 @@ namespace hpx {
...
+        virtual std::uint32_t assign_cores(std::string const&, std::uint32_t)
+        {
+            return std::uint32_t(-1);
+        }
+
+        virtual std::uint32_t assign_cores()
+        {
+            return std::uint32_t(-1);
+        }
+
```
msimberg, line 402: Do these need to be in the local runtime?

hkaiser: Yes, I know those are unrelated. I need to get back to this issue to try isolating the functionalities such that we don't create a dependency on the distributed runtime. Let me see what I can do,

msimberg: Where were these needed? Usually it should be enough to do `get_runtime_distributed().assign_cores()`, but was there something that forced these to be in the local one?

hkaiser: I didn't want to use this function as this would have created a dependency on the distributed runtime. AGAS shouldn't depend on it as the runtime depends on AGAS.

hkaiser: To summarize the discussion on IRC: for now we'll leave these virtual functions in place. Later we will move the `big_boot_barrier` into the module `runtime_distributed` (where it really belongs) which will allow to remove these functions.

PR #5149, opened 2021-02-04
Rename 'module' symbols to avoid keyword conflict

### LINKED ISSUE #5148: runtime_support.hpp does not work with newer cling
## Expected Behavior

Code including this file should compile.

## Actual Behavior

An error is triggered. The compiler claims a ')' is needed after the identifier "module" on line 300.

Changing "module" to "module_" on lines 300 and 303 allows the code to compile.

hkaiser:
Fixes #5148

('module' is a keyword starting with C++20)

### DIFF
--- a/hpx/runtime/components/server/runtime_support.hpp
+++ b/hpx/runtime/components/server/runtime_support.hpp
@@ -297,10 +297,10 @@ namespace hpx { namespace components { namespace server
             bool isdefault, bool isenabled,
             hpx::program_options::options_description& options,
             std::set<std::string>& startup_handled);
-        bool load_startup_shutdown_functions_static(std::string const& module,
+        bool load_startup_shutdown_functions_static(std::string const& mod,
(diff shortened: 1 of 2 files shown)


msimberg (approved):
Nasty... Thanks! How difficult would it be to add an inspect check for this? Let me know if you'd like me to add it.

aurianer (approved):
Thanks!

hkaiser:
> Nasty... Thanks! How difficult would it be to add an inspect check for this? Let me know if you'd like me to add it.

If we had a C++20 builder then we wouldn't need an inspect check, I think.

msimberg:
> > Nasty... Thanks! How difficult would it be to add an inspect check for this? Let me know if you'd like me to add it.
> 
> If we had a C++20 builder then we wouldn't need an inspect check, I think.

Good point! We can definitely add that.

PR #5154, opened 2021-02-07
Splitting unit tests on circleci to avoid timeouts

hkaiser:


msimberg (approved):
Thanks! I suspect the timeout is not just a problem on our side, but also on CircleCI's hardware. This should help avoid the worst of it though.

PR #5160, opened 2021-02-10
Adding checks ensuring modules are not cross-referenced from other module categories

hkaiser:
This makes sure that modules are not linked into more than one shared module library.


cmake/HPX_AddModule.cmake
```diff
@@ -267,10 +267,33 @@ function(add_hpx_module libname modulename)
...
-  target_link_libraries(
-    hpx_${modulename} PUBLIC ${${modulename}_MODULE_DEPENDENCIES}
-  )
-  target_link_libraries(hpx_${modulename} PUBLIC ${${modulename}_DEPENDENCIES})
+  if(${modulename}_MODULE_DEPENDENCIES)
+    # verify that all dependencies are from the same module category
+    foreach(dep ${${modulename}_MODULE_DEPENDENCIES})
+      # consider only module dependencies, not other targets
+      string(FIND ${dep} "hpx_" find_index)
+      string(FIND ${dep} "hpx_dependencies_" find_dep_index)
```
msimberg, line 275: This is a coincidence, but `hpx_dependencies_...` should in fact also not be in `MODULE_DEPENDENCIES`, would you mind changing the check to include that?

cmake/HPX_AddModule.cmake
```diff
@@ -267,10 +267,33 @@ function(add_hpx_module libname modulename)
     # cmake-format: on
   endif()
 
-  target_link_libraries(
-    hpx_${modulename} PUBLIC ${${modulename}_MODULE_DEPENDENCIES}
-  )
-  target_link_libraries(hpx_${modulename} PUBLIC ${${modulename}_DEPENDENCIES})
+  if(${modulename}_MODULE_DEPENDENCIES)
+    # verify that all dependencies are from the same module category
+    foreach(dep ${${modulename}_MODULE_DEPENDENCIES})
```
msimberg, line 272: Should we make it optional, and only enable the check on CI? I'd guess it's not terribly slow, but future similar checks may add up...

cmake/HPX_AddModule.cmake
```diff
@@ -267,10 +267,33 @@ function(add_hpx_module libname modulename)
     # cmake-format: on
   endif()
 
-  target_link_libraries(
-    hpx_${modulename} PUBLIC ${${modulename}_MODULE_DEPENDENCIES}
-  )
-  target_link_libraries(hpx_${modulename} PUBLIC ${${modulename}_DEPENDENCIES})
+  if(${modulename}_MODULE_DEPENDENCIES)
```
msimberg, line 270: Was this necessary to get things working?

msimberg (commented):
Thanks @hkaiser for doing this!

cmake/HPX_AddModule.cmake
hkaiser: No, just for consistency purposes. I'm not sure if `target_link_libraries` can be called without arguments.

cmake/HPX_AddModule.cmake
hkaiser: I can do that. I will also change the then failing incantations.

cmake/HPX_AddModule.cmake
hkaiser: How do I do that? Do we have a flag already in place?

cmake/HPX_AddModule.cmake
msimberg: It can indeed be empty, no need to check if it's empty.

cmake/HPX_AddModule.cmake
msimberg: No flag yet. We've done these kind of checks separately using shell scripts so far. What I had in mind was just adding an `ADVANCED` flag for this purpose.

cmake/HPX_AddModule.cmake
msimberg: Thank you!

hkaiser:
@msimberg I think I have addressed your comments.

msimberg (approved):
Looks good now, thank you!

PR #5165, opened 2021-02-10
Require cmake 3.17

### LINKED ISSUE #5161: Bump required CMake version to 3.17
3.17 is the first version of CMake to come with a `FindCUDAToolkit` module (also relevant for #4482, although the CUDA language is available since 3.10). If needed for something else we could consider bumping it even higher since it's quite a big jump in the first place.

aurianer:
Require CMake 3.17
Fixes #5161
Flyby: remove unused `add_hpx_in_framework_config_test`


hkaiser (dismissed):
LGTM, thanks!

msimberg:
> I'm not sure what to do about this:
> https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_AddConfigTest.cmake#L232-L237

Could you check what happens if you remove it?

> TODO: update CI image

Can you take care of the CI image? Looks like Ubuntu 20.10 unfortunately has CMake 3.16 so it might need manual building...

aurianer:
> > I'm not sure what to do about this:
> > https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_AddConfigTest.cmake#L232-L237
> 
> Could you check what happens if you remove it?
Probably nothing since `add_hpx_in_framework_config_test` is used nowhere
> 
> > TODO: update CI image
> 
> Can you take care of the CI image? Looks like Ubuntu 20.10 unfortunately has CMake 3.16 so it might need manual building...
Yes I did the modification locally, just checking that it works before opening a PR

hkaiser:
@aurianer I merged your PR to the base image, now ninja can't find cmake anymore (see for instance: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/6591/workflows/49f48bff-9c45-4597-9698-d881d42f0f66/jobs/236668/steps)

msimberg:
> @aurianer I merged your PR to the base image, now in CircleCI ninja can't find cmake anymore (see for instance: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/6591/workflows/49f48bff-9c45-4597-9698-d881d42f0f66/jobs/236668/steps)

I reverted the commit for now so that other builds can continue normally once the image is pushed again. Last time I made changes to the docker image I added a version tag (just an incrementing number) so that the new image can be bumped and tested in a separate PR first. I suggest we bump it to two for this change. The version is set here: https://github.com/STEllAR-GROUP/docker_build_env/blob/8b94110d3c81bcadbe648b382202059313ae90cd/.circleci/config.yml#L15, and then used in our various CI configs.

aurianer:
It works locally so I'll open a new PR and let circleci run to see if it shows the pb

aurianer:
Looks like I have more investigation to do cause looks like the circleci in place for the Dockerfile found cmake correctly:
https://app.circleci.com/pipelines/github/STEllAR-GROUP/docker_build_env/12/workflows/6aba159d-8285-48c5-9c94-3cfd462c4d3b/jobs/183

libs/core/allocator_support/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.17 FATAL_ERROR)
```
msimberg, line 7: By the way, should we perhaps get rid of these?

aurianer: Yes I think it was needed when the modules were sub-`project()`s but looks like it's not the case anymore :+1:

hkaiser: Didn't we plan to make modules available for external projects? I thought those would be needed for that.

msimberg: Yes, sometime in the very beginning of modularization. However, I'm still of the opinion that the modules are too fine-grained to expose directly to users (are we going to have a well-defined, public API for each module? versioning?). If that were to happen there would be a lot more work than just keeping the `cmake_minimum_required` calls in the modules, and adding those back when/if needed is not very difficult.

msimberg:
Thanks @aurianer! @hkaiser, others, note that the memory leak we saw with asan from topology has now reappeared with the updated image, and it seems to be coming from within topology (hwloc bitmaps?). @aurianer would you mind opening an issue for the leak?

aurianer:
I just dropped the last commit since CMake 3.19 has now been installed on rostam.

hkaiser:
retest lsu

docs/sphinx/manual/building_hpx.rst
```diff
@@ -71,7 +71,7 @@ Quick start
 
 Here, you will use the command-line, non-interactive CMake interface.
 
-#. Download and install CMake here: |cmake_download|_. Version 3.13 is the
+#. Download and install CMake here: |cmake_download|_. Version 3.17 is the
```
msimberg, line 74: There are a couple of tables in this file as well which still say 3.13. Would you mind updating those?

libs/create_module_skeleton.py
```diff
@@ -24,7 +24,7 @@
 header_str = '=' * len(module_name)
 
 # CMake minimum version
-cmake_version = '3.13'
+cmake_version = '3.17'
```
msimberg, line 27: This could in fact be removed (same in `create_component_skeleton.py`).

msimberg (changes requested):
Just a couple of minor comments still. This is good to go after that!

PR #5167, opened 2021-02-10
Fixing deprecation warnings that slipped through the net

hkaiser:


msimberg (approved):
Thank you! Now the question is why these didn't get caught in the net...

aurianer (approved):
Thanks

gonidelis:
@hkaiser  Thanks and sorry for missing those. @msimberg good question. I will check it out on my new adaptation ;)

PR #5168, opened 2021-02-11
Remove include of itself

aurianer:
This file includes itself


msimberg (approved):

hkaiser:
Excellent catch! Thanks!

gonidelis:
@aurianer  Thanks a lot!

gonidelis:
Just noticed that each and every parallel algorithm does include itself. I don't know where does this come from. Maybe it was part of the inital implementations.

PR #5169, opened 2021-02-11
Attempt to get Piz Daint jenkins setup running after maintenance

msimberg:


msimberg:
retest cscs

msimberg:
Note: this adds the account name to slurm invocations on the CSCS jenkins instance. It's required for normal users, but doesn't actually seem to be required on the jenkins instance. However, I'll merge this in case they change the requirements.

The actual issue was solved separately at CSCS and jobs start fine again now.

PR #5170, opened 2021-02-12
Add HPX_WITH_PRECOMPILED_HEADERS option

msimberg:
Seems to give a modest improvement (5-10% on top of a unity build, for `hpx_core`). The list of headers may need some tweaking, since there's a tradeoff between how many headers are included and how many are actually used by HPX.

Boost headers are not included since they need to come after `hpx/config.hpp` (what's the reason for this?).

I've no clue if this will work correctly on Windows, so please give it a try and see what happens!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -326,6 +326,83 @@ if(HPX_WITH_UNITY_BUILD)
   set(HPX_WITH_UNITY_BUILD_OPTION UNITY_BUILD)
 endif()
 
+hpx_option(
+  HPX_WITH_PRECOMPILED_HEADERS
+  BOOL
+  "Enable precompiled headers for certain build targets (experimental, requires CMake 3.16 or newer) (default OFF)"
+  OFF
+  ADVANCED
+)
+if(HPX_WITH_PRECOMPILED_HEADERS)
+  if(CMAKE_VERSION VERSION_LESS "3.16")
+    hpx_error("HPX_WITH_UNITY_BUILD=ON but this requires CMake 3.16 or newer.")
+  endif()
+
+  set(HPX_WITH_PRECOMPILED_HEADERS_INTERNAL ON)
+
+  add_library(hpx_precompiled_headers OBJECT libs/src/dummy.cpp)
+  target_link_libraries(hpx_precompiled_headers PRIVATE hpx_private_flags)
+  target_precompile_headers(
+    hpx_precompiled_headers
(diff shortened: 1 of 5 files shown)


hkaiser:
> Boost headers are not included since they need to come after `hpx/config.hpp` (what's the reason for this?).

We see a lot of stupid warnings if we #include std headers before hpx/config.hpp on windows. Those warnings are currently being disabled by some of the configuration #defines we add (e.g. `_SILENCE_ALL_CXX17_DEPRECATION_WARNINGS` and others). I think we can avoid those if we added `hpx/config.hpp` to the list of pre-compiled headers - this would work only if cmake retains the sequence of those headers when generating the pch file).

msimberg:
> > Boost headers are not included since they need to come after `hpx/config.hpp` (what's the reason for this?).
> 
> We see a lot of stupid warnings if we #include std headers before hpx/config.hpp on windows. Those warnings are currently being disabled by some of the configuration #defines we add (e.g. `_SILENCE_ALL_CXX17_DEPRECATION_WARNINGS` and others). I think we can avoid those if we added `hpx/config.hpp` to the list of pre-compiled headers - this would work only if cmake retains the sequence of those headers when generating the pch file).

Could those defines not be added to the compiler flags instead of having them in `hpx/config.hpp`? It sounds to me like that'd be much less error prone.

hkaiser:
> Could those defines not be added to the compiler flags instead of having them in `hpx/config.hpp`? It sounds to me like that'd be much less error prone.

We could do that (`PUBLIC`ly)

hkaiser:
Unfortunately this doesn't work for me. For some reason, the `REUSE_FROM` clause seems to be broken in some ways for VS. The PCH file gets properly generated first but then it is immediately deleted by the first target that tries to reuse it, causing the subsequent compilation to fail. This could be a cmake issue, but I have not found any other reports for it.

I'd say, let's go ahead with the new option in any case. Thanks for introducing it, @msimberg.

msimberg:
> Unfortunately this doesn't work for me. For some reason, the `REUSE_FROM` clause seems to be broken in some ways for VS. The PCH file gets properly generated first but then it is immediately deleted by the first target that tries to reuse it, causing the subsequent compilation to fail. This could be a cmake issue, but I have not found any other reports for it.

Could very well be. I may also be using it wrongly. It's a relatively new feature after all. @K-ballo do you have any experience with CMake's precompiled headers?

PR #5171, opened 2021-02-12
Remove HPX_WITH_THREAD_SCHEDULERS CMake option

msimberg:
Instead enables everything that is enable-able.

@biddisco I've added a fallback for the shared priority scheduler for when `HPX_WITH_MAX_CPU_COUNT` is undefined (HPX doesn't use fixed-size bitsets in this case). It seems to work ok, but I'd appreciate you having a look.


hkaiser (dismissed):
Thanks for looking into this! LGTM.

msimberg:
retest lsu

PR #5172, opened 2021-02-12
Add separate local-only entry point

msimberg:
Consider this more of a proposal than a finished thing.

In light of our discussions about splitting out a local-only core of HPX into a separate repository, some additional changes are needed in current HPX. The `HPX_WITH_DISTRIBUTED_RUNTIME`/`HPX_WITH_NETWORKING` options allow building only a subset of HPX. However, many places contain `#ifdef`s for those two definitions which makes it hard to split out a standalone local runtime.

Here I've removed any `#ifdef HPX_HAVE_DISTRIBUTED_RUNTIME/HPX_HAVE_NETWORKING` from the core/parallelism libraries and moved/split tests so that those tests that require the distributed runtime live in modules that are enabled with the distributed runtime. Currently I've somewhat crudely put them in e.g. the actions module, but they may have more appropriate homes.

I've also added a minimal local runtime initialization function `hpx::local::init`, together with the modules `init_runtime_local`, `command_line_handling_local`, and `runtime_configuration_local`. The `threadmanager` and `resource_partitioner` modules could only depend on `hpx::util::section` in the new `ini` module, and are fully shared between the local and distributed runtimes. These modules would be enough to have a separate local-only runtime without `#ifdef`s in the code. One test now uses the `hpx::local::init`, but the idea is that any test that belongs to "core" HPX would use that instead of `hpx::init`.

My open question is still how much to duplicate/share between the local and distributed runtimes. I think the local-only runtime is simple enough that duplicating a lot of the functionality is feasible, and it let's us change the local-only runtime independently of the distributed one. On the other hand, fixing e.g. bugs found in one runtime would likely need the same change in the other runtime. Note that by runtime I don't only refer to the runtime classes, but also the command line handling and configuration. Thoughts?


hkaiser:
> Move winsocket helper to init_runtime_local module

@msimberg why is that? I think this is really only needed if we use sockets, which we shouldn't in local-only mode.

msimberg:
> > Move winsocket helper to init_runtime_local module
> 
> @msimberg why is that? I think this is really only needed if we use sockets, which we shouldn't in local-only mode.

I wasn't aware that it wasn't required at all in that case. I'll happily revert that :)

hkaiser:
I think the new `ini` module can go to core, so can the plugin registry. I am planning to move all of the code related to parcel ports into a new sub-module `networking` with the goal that those will depend on core modules only. For this, some of the functionalities from full should (and can) be moved to core.

hkaiser:
@msimberg do you think the new local modules could be used as the base-implementation for the distributed modules, i.e. do you think we can remove (some of) the code duplication?

msimberg:
> I think the new `ini` module can go to core, so can the plugin registry. I am planning to move all of the code related to parcel ports into a new sub-module `networking` with the goal that those will depend on core modules only. For this, some of the functionalities from full should (and can) be moved to core.

Yep on `ini`, unsure about the plugin registry. There's already a base `plugin` module in `core` but the plugin registry helper functionality depends on the runtime configuration (I think fairly so).

> @msimberg do you think the new local modules could be used as the base-implementation for the distributed modules, i.e. do you think we can remove (some of) the code duplication?

To some extent, and they already do that to some extent. I can try to improve this a bit as well. I think I will revert the split of the `runtime_configuration` module for now as that needs some more thinking. Currently one of the reasons things like agas manage without depending on the distributed runtime is that the local runtime hands out the configuration for both the local and distributed runtimes. The distributed runtime should be handing out the configuration for things like agas, but that will create circular dependencies at the moment.

msimberg:
I've reverted the split of `runtime_configuration`. `command_line_handling` has a local and distributed module now, sharing utilities where reasonable. The logging initialization is split and most of the functionality is shared. Most of the other changes are just tests moving around and ifdefs being removed.

hkaiser (dismissed):
This in principle LGTM and can be merged. Some comments:

- late command line handling should be added to support `--hpx:print-bind` and friends
- this still has a bit too much code duplication, but is Good Enough (tm) as a first step

Thanks!

msimberg:
>     * late command line handling should be added to support `--hpx:print-bind` and friends

`--hpx:print-bind` is now handled (really, it was a matter of storing the command line in the config, the handling was already done in the runtime)

>     * this still has a bit too much code duplication, but is Good Enough (tm) as a first step

Yes, I'm still unsure how much the two should be tied together but time will tell which direction to go.

PR #5173, opened 2021-02-12
Fetching APEX using cmake FetchContent

hkaiser:
Can we remove the cmake/GitExternal.cmake script now?


hkaiser:
This is good to go.

Issue #5156, opened 2021-02-09
Disentangle segmented algorithms


hkaiser:
Currently the segmented algorithms are part and parcel of the implementation of the parallel algorithms. That unnecessarily complicates those. Now, that we have converted the parallel algorithm APIs into customization point (CPO) based implementations, the segmented algorithms could be separated and exposed through their own specializations of `tag_invoke` to be picked up by the CPOs. This would significantly simplify the base algorithms and allow to separate the segmented algorithms completely.

hkaiser:
Here is a list of algorithms that need to be touched:

- [ ] `adjacent_difference`
- [ ] `adjacent_find`
- [ ] `all_of`
- [ ] `any_of`
- [ ] `count`
- [ ] `count_if`
- [ ] `exclusive_scan`
- [ ] `fill`
- [ ] `find`
- [ ] `find_if`
- [ ] `find_if_not`
- [ ] `for_each` (#5176)
- [ ] `for_each_n`
- [ ] `generate`
- [ ] `inclusive_scan`
- [ ] `max_element`
- [ ] `min_element`
- [ ] `minmax_element`
- [ ] `none_of`
- [ ] `reduce`
- [ ] `transform`
- [ ] `transform_exclusive_scan`
- [ ] `transform_inclusive_scan`
- [ ] `transform_reduce`

hkaiser:
This can be closed now. Thanks to everybody helping with resolving the issue.

PR #5176, opened 2021-02-15
Starting to separate segmented algorithms: for_each

hkaiser:
This works towards solving #5156

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/algorithms/for_each.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/algorithms/for_each.hpp
@@ -240,7 +240,7 @@ namespace hpx {
 #include <hpx/algorithms/traits/segmented_iterator_traits.hpp>
 #include <hpx/concepts/concepts.hpp>
 #include <hpx/functional/detail/invoke.hpp>
-#include <hpx/functional/tag_invoke.hpp>
+#include <hpx/functional/tag_fallback_invoke.hpp>
 #include <hpx/iterator_support/traits/is_iterator.hpp>
 #include <hpx/modules/execution.hpp>
 #include <hpx/modules/executors.hpp>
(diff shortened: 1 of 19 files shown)


msimberg (commented):
This is very nice!

Would it be better to put the segmented iterators in `hpx::distributed` to avoid introducing another namespace? On the other hand `hpx::segmented` keeps them separate. Would compile times potentially be affected if there are other `tag_invoke` overloads in `hpx::distributed`? Given that the iterator names are rarely visible to the user `hpx::segmented` might be a good choice.

hkaiser:
> This is very nice!
> 
> Would it be better to put the segmented iterators in `hpx::distributed` to avoid introducing another namespace? On the other hand `hpx::segmented` keeps them separate. Would compile times potentially be affected if there are other `tag_invoke` overloads in `hpx::distributed`? Given that the iterator names are rarely visible to the user `hpx::segmented` might be a good choice.

I have left the iterators in `namespace hpx::segmented`. Yah, compile times would be affected if there will be many `tag_invoke` overloads in the same namespace. I'm not sure how this could be rectified, however.

I have also used `tag_invoke_fallback` for the non-segmented algorithms which allows to nicely specialize the segmented ones (and most likely your Kokkos specializations) without having to explicitly disable the non-segmented overloads.

msimberg:
> > This is very nice!
> > Would it be better to put the segmented iterators in `hpx::distributed` to avoid introducing another namespace? On the other hand `hpx::segmented` keeps them separate. Would compile times potentially be affected if there are other `tag_invoke` overloads in `hpx::distributed`? Given that the iterator names are rarely visible to the user `hpx::segmented` might be a good choice.
> 
> I have left the iterators in `namespace hpx::segmented`. Yah, compile times would be affected if there will be many `tag_invoke` overloads in the same namespace. I'm not sure how this could be rectified, however.
> 
> I have also used `tag_invoke_fallback` for the non-segmented algorithms which allows to nicely specialize the segmented ones (and most likely your Kokkos specializations) without having to explicitly disable the non-segmented overloads.

Thanks, this seems like the right thing to do (until we figure out the next restriction, but I think this will go a long way)!

We could probably do a sweep through the already converted algorithms to make them use `tag_invoke_fallback` in one go.

hkaiser:
> We could probably do a sweep through the already converted algorithms to make them use `tag_invoke_fallback` in one go.

@gonidelis would this be something you want to take on?

gonidelis:
> @gonidelis would this be something you want to take on?

Say no more

hkaiser:
@msimberg this is good to go, please review.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/for_each.hpp
```diff
@@ -560,30 +487,70 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
-        ///////////////////////////////////////////////////////////////////////
-        // non-segmented implementation
-        template <typename ExPolicy, typename FwdIterB, typename FwdIterE,
+        /// Non Segmented implementation
+        template <typename ExPolicy, typename FwdIter, typename Size,
             typename F, typename Proj>
-        inline typename util::detail::algorithm_result<ExPolicy, FwdIterB>::type
-        for_each_(ExPolicy&& policy, FwdIterB first, FwdIterE last, F&& f,
+        typename util::detail::algorithm_result<ExPolicy, FwdIter>::type
+        for_each_n_(ExPolicy&& policy, FwdIter first, Size count, F&& f,
```
msimberg, line 494: Just to check: this `for_each_n_` and the declaration for the segmented one below are only kept around for the deprecated `hpx::parallel::for_each_n`? Although I see you changed `hpx::parallel::for_each` to use `for_each::call` directly. Can we do the same here so that we can just remove these `for_each_n_`s? Am I missing some other place where they're required?

msimberg (commented):
Thanks @hkaiser! Looks good to me except for one spot that could possibly have more code deleted, but I'm unsure.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/for_each.hpp
hkaiser: My plan was to make this ready first and then to look through all other algorithms. `for_each_n` would be the next, I guess.

PR #5178, opened 2021-02-16
Fix parallel remove/remove_copy/transform namespace references in docs

gonidelis:
Fixing a minor documentation oversight on the parallel algos I adapted.


hkaiser (approved):
LGTM, thanks!

PR #5179, opened 2021-02-17
Make sure --hpx:local command line option is respected with networking is off but distributed runtime is on

msimberg:


aurianer (approved):
Thanks!

hkaiser (approved):
LGTM, thanks!

hkaiser:
Should `--hpx:local` be the assumed as the default if using the local HPX (if it isn't already?)

msimberg:
> Should `--hpx:local` be the assumed as the default if using the local HPX (if it isn't already?)

Yep, it's the default, and only, option when `HPX_WITH_DISTRIBUTED_RUNTIME=OFF`. When `HPX_WITH_NETWORKING=OFF` and `HPX_WITH_DISTRIBUTED_RUNTIME=ON` the default is still to use the distributed runtime (i.e. initialize whatever stubs of AGAS etc. still exist in that case).

PR #5180, opened 2021-02-17
Fix htts_v2 tests linking

aurianer:
Looks like there is an ABI problem with boost regex, this is an attempt to fix it


hkaiser (approved):
Thanks for looking into this!

PR #5183, opened 2021-02-19
Remove dependency on boost/iterators/iterator_categories.hpp

hkaiser:
- flyby: add `_t `/`_v` variations for `is_iterator` traits

### DIFF
--- a/libs/core/iterator_support/include/hpx/iterator_support/boost_iterator_categories.hpp
+++ b/libs/core/iterator_support/include/hpx/iterator_support/boost_iterator_categories.hpp
@@ -1,0 +1,209 @@
+//  Copyright (c) 2021 Hartmut Kaiser
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// (C) Copyright Jeremy Siek 2002.
+
+#pragma once
+
+#include <hpx/config.hpp>
+
+#if !defined(                                                                  \
+    HPX_ITERATOR_SUPPORT_HAVE_BOOST_ITERATOR_TRAVERSAL_TAG_COMPATIBILITY)
+
+#include <hpx/type_support/identity.hpp>
+#include <hpx/type_support/lazy_conditional.hpp>
+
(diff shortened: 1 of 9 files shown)


aurianer (approved):
LGTM thanks!

msimberg (approved):
Despite my slightly negative comment I'd be happy to go with this as is as well. Thank you @hkaiser for looking into this!

libs/core/iterator_support/include/hpx/iterator_support/boost_iterator_categories.hpp
```diff
@@ -0,0 +1,214 @@
...
+// (C) Copyright Jeremy Siek 2002.
+
+#pragma once
+
+#include <hpx/config.hpp>
+
+// we deliberately leave the header guards in place to protect against
+// #including this header and the original one from Boost at the same time
+#ifndef BOOST_ITERATOR_CATEGORIES_HPP
+#define BOOST_ITERATOR_CATEGORIES_HPP
```
msimberg, line 16: Hmm, looks like my comment was lost... I'll try to summarize: I don't love this, but this header would be the safest one to do this with since it hasn't changed in the last 4 years in Boost. I'm wondering if we could do the same here as what we've done with support for serialization of Boost types? That way when/if all remaining Boost compatibility is behind CMake options they could just be consolidated into a single `HPX_WITH_BOOST_COMPATIBILITY/SUPPORT` option or similar. What do you think?

hkaiser: That would work. Let me see what I can do. I think that using this option instead of the unwieldy header guard would be a possible solution (even if it isn't perfect either).

hkaiser: I introduced a compatibility option that I left `OFF` by default. So it shouldn't be a problem for people enable compatibility if needed.

hkaiser: Actually I had to enable it by default as the boost header is indirectly included somewhere.

hkaiser:
retest cscs

msimberg (approved):
Thanks @hkaiser, looks good to me! (I just force pushed a cmake-format fix, this is hopefully good to go after that.)

PR #5187, opened 2021-02-19
Force CMake to set the -std=c++XX flag

msimberg:
Fixes CUDA compilation with clang.

@NK-Nikunj would you mind trying this out?

Based on https://github.com/kokkos/kokkos/issues/2520 and https://github.com/masterleinad/kokkos/commit/2a562399fa273952b1910f4c690e4f1a50e9d0ff.


NK-Nikunj (commented):
It compiles just fine. Thanks for looking into this!

aurianer (approved):
Thanks!

hkaiser (approved):
This is a strange one. Thanks for looking into things!

PR #5092, opened 2020-12-24
Replace spirit.qi with x3

K-ballo:

PR #5188, opened 2021-02-19
Bump Boost minimum version to 1.71.0

aurianer:
This allows to replace spirit.qi with x3
https://github.com/STEllAR-GROUP/hpx/pull/5092#issuecomment-751365759

Any objections?


msimberg:
> This allows to replace spirit.qi with x3
> [#5092 (comment)](https://github.com/STEllAR-GROUP/hpx/pull/5092#issuecomment-751365759)
> 
> Any objections?

No objections at all, but CI will need some updates...

libs/full/program_options/docs/index.rst
```diff
@@ -12,8 +12,8 @@ program_options
 ===============
 
 The module program_options is a direct fork of the Boost.ProgramOptions library
-(Boost V1.70.0). For more information about this library please see `here
-<https://www.boost.org/doc/libs/1_70_0/doc/html/program_options.html>`__.
+(Boost V1.71.0). For more information about this library please see `here
+<https://www.boost.org/doc/libs/1_71_0/doc/html/program_options.html>`__.
```
msimberg, line 16: Hmm, it is still based on 1.70 even if we require 1.71. I'd either keep the note about 1.70, or remove it altogether.

hkaiser: I'd like to see this note left in place (while referring to the original V1.70.0).

aurianer:
Waiting to have the corresponding Boost modules/files on Daint

aurianer:
This PR will fail until https://github.com/eth-cscs/production/pull/2179 is merged

aurianer:
I think I've addressed all the CI related issues, let's see what's the next run says..
`env-icc` jenkins run is failing but this problem will be addressed in #5199.

msimberg:
Waiting for a reply to https://github.com/boostorg/math/issues/181#issuecomment-792766476 before doing anything with the Cray clang builder, but we may have to disable/change it for the moment.

msimberg:
I've added a custom Boost config which unconditionally disables `BOOST_HAVE_FLOAT128` on the Cray clang builder for now. There may be a better workaround/fix in the future. Assuming I managed to get the syntax right for the CMake scripts this is ready to go. I'm ignoring the ICC failures as they'll be fixed separately.

PR #5190, opened 2021-02-19
Set compatibility headers by default to on

### LINKED ISSUE #5163: Remove module-specific compatibility and deprecation options
I propose that we remove the module-specific options `HPX_WITH_<modulename>_COMPATIBILITY_HEADERS`, `HPX_WITH_<modulename>_DEPRECATION_WARNINGS`, and `HPX_WITH_<modulename>_TESTS`. We can instead use the global `HPX_WITH_DEPRECATION_WARNINGS` and `HPX_WITH_TESTS`. I see no harm in leaving compatiblity headers until they are outright removed. I find the module-specific options unnecessarily fine-grained, and they add unnecessary noise (and options) to users configuring HPX. The initial idea was good but I don't think it works as well when we have close to 100 modules. Objections?

aurianer:
Fixes #5163

- Remove the `COMPATIBILITY_HEADERS` option in the `add_hpx_module` cmake function as it is now set to `ON` by default
- Modify the print summary to print only the extraordinary options (if compatibility headers is off or other options)

### DIFF
--- a/cmake/HPX_AddModule.cmake
+++ b/cmake/HPX_AddModule.cmake
@@ -417,7 +417,7 @@ function(add_hpx_module libname modulename)
 
   include(HPX_PrintSummary)
   create_configuration_summary(
-    "  Module configuration summary (${modulename}):" "${modulename}"
+    "    Module configuration (${modulename}):" "${modulename}"
   )
 
 endfunction(add_hpx_module)
--- a/cmake/HPX_AddModule.cmake
+++ b/cmake/HPX_AddModule.cmake
@@ -9,7 +9,6 @@ include(HPX_ExportTargets)
 function(add_hpx_module libname modulename)
   # Retrieve arguments
   set(options CUDA CONFIG_FILES)
-  # Compatibility needs to be on/off to allow 3 states : ON/OFF and disabled
   set(one_value_args GLOBAL_HEADER_GEN)
   set(multi_value_args
       SOURCES
--- a/cmake/HPX_AddModule.cmake
+++ b/cmake/HPX_AddModule.cmake
@@ -10,7 +10,7 @@ function(add_hpx_module libname modulename)
   # Retrieve arguments
   set(options CUDA CONFIG_FILES)
   # Compatibility needs to be on/off to allow 3 states : ON/OFF and disabled
-  set(one_value_args COMPATIBILITY_HEADERS GLOBAL_HEADER_GEN)
+  set(one_value_args GLOBAL_HEADER_GEN)
   set(multi_value_args
       SOURCES
       HEADERS
@@ -35,12 +35,6 @@ function(add_hpx_module libname modulename)
   include(HPX_Message)
   include(HPX_Option)
 
-  if(NOT "${${modulename}_COMPATIBILITY_HEADERS}" STREQUAL "")
-    set(_have_compatibility_headers_option TRUE)
-  else()
-    set(_have_compatibility_headers_option FALSE)
-  endif()
-
   # Global headers should be always generated except if explicitly disabled
   if("${${modulename}_GLOBAL_HEADER_GEN}" STREQUAL "")
     set(${modulename}_GLOBAL_HEADER_GEN ON)
@@ -61,23 +55,6 @@ function(add_hpx_module libname modulename)
             FORCE
   )
 
-  # HPX options
-  if(${_have_compatibility_headers_option})
-    set(_compatibility_headers_default OFF)
-    if(${modulename}_COMPATIBILITY_HEADERS)
-      set(_compatibility_headers_default ON)
-    endif()
-    hpx_option(
-      HPX_${modulename_upper}_WITH_COMPATIBILITY_HEADERS
-      BOOL
-      "Enable compatibility headers for old headers. (default: ${_compatibility_headers_default})"
-      ${_compatibility_headers_default}
-      ADVANCED
-      CATEGORY "Modules"
-      MODULE ${modulename_upper}
-    )
-  endif()
-
   # Main directories of the module
   set(SOURCE_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/src")
   set(HEADER_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/include")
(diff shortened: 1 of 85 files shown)


cmake/HPX_AddModule.cmake
```diff
@@ -10,7 +10,7 @@ function(add_hpx_module libname modulename)
   # Retrieve arguments
   set(options CUDA CONFIG_FILES)
   # Compatibility needs to be on/off to allow 3 states : ON/OFF and disabled
```
hkaiser, line 12: This comment can go now, I believe

aurianer: Thanks :)

cmake/HPX_AddModule.cmake
```diff
@@ -62,21 +55,15 @@ function(add_hpx_module libname modulename)
...
-  endif()
+  hpx_option(
+    HPX_${modulename_upper}_WITH_COMPATIBILITY_HEADERS
+    BOOL
+    "Enable compatibility headers for old headers. (default: ON)"
+    ON
+    ADVANCED
+    CATEGORY "Modules"
+    MODULE ${modulename_upper}
+  )
```
msimberg, line 66: I'd go one step further and not have the option at all (just have the headers there until they're removed).

cmake/HPX_AddModule.cmake
```diff
@@ -85,9 +72,7 @@ function(add_hpx_module libname modulename)
   hpx_debug("Add module ${modulename}: SOURCE_ROOT: ${SOURCE_ROOT}")
   hpx_debug("Add module ${modulename}: HEADER_ROOT: ${HEADER_ROOT}")
 
-  if(${_have_compatibility_headers_option}
-     AND HPX_${modulename_upper}_WITH_COMPATIBILITY_HEADERS
-  )
+  if(HPX_${modulename_upper}_WITH_COMPATIBILITY_HEADERS)
```
msimberg, line 75: This probably needs to check if the list of compatibility headers is non-empty.

cmake/HPX_AddModule.cmake
aurianer: But if the warnings are annoying (too much of them) and the users wants to temporarily silence them because of a lack of time to fix them, doesn't it make sense to keep the option still?

msimberg: They still have the option to silence them by defining `HPX_HAVE_DEPRECATION_WARNINGS=0` (and they don't need to rebuild HPX for that to take effect).

msimberg: And more importantly, this option wouldn't have controlled that anyway. That would've been the one removed in #5177.

aurianer: True will remove them

cmake/HPX_AddModule.cmake
msimberg: Still there ;)

cmake/HPX_AddModule.cmake
```diff
@@ -61,33 +61,14 @@ function(add_hpx_module libname modulename)
...
   set(SOURCE_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/src")
   set(HEADER_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/include")
 
   hpx_debug("Add module ${modulename}: SOURCE_ROOT: ${SOURCE_ROOT}")
   hpx_debug("Add module ${modulename}: HEADER_ROOT: ${HEADER_ROOT}")
 
-  if(${_have_compatibility_headers_option}
-     AND HPX_${modulename_upper}_WITH_COMPATIBILITY_HEADERS
-  )
+  if(${_have_compatibility_headers})
```
msimberg, line 71:
```suggestion
  if(${modulename}_COMPAT_HEADERS)
```
?
Same below.

cmake/HPX_PrintSummary.cmake
```diff
@@ -31,22 +28,28 @@ function(create_configuration_summary message module_name)
   get_property(
     _variableNames GLOBAL PROPERTY HPX_MODULE_CONFIG_${module_name_uc}
   )
-  foreach(_variableName ${_variableNames})
-    if(${_variableName}Category)
 
-      # handle only options which start with HPX_WITH_
-      string(FIND ${_variableName} "${upper_cfg_name}_WITH_" __pos)
+  # If the only options in the module is the compatibility headers and this
+  # option is ON, then we don't print the module summary
```
msimberg, line 33: Comment probably needs updating?

docs/sphinx/contributing/modules.rst
```diff
@@ -76,10 +76,6 @@ header files in most cases.
 
 Optional single-value arguments are:
 
-* ``COMPATIBILITY_HEADERS``: Can be ``ON``, ``OFF``, or left out. Enables
-  compatibility headers. Creates a variable which can be turned on or off by the
-  user when set to ``ON`` or ``OFF``. If left out the option is completely
-  disabled.
```
msimberg, line 82: Good catch! Could you remove the `DEPRECATION_WARNINGS` point above as well?

hkaiser (dismissed):
LGTM, thanks!

cmake/HPX_AddModule.cmake
aurianer: If some compatibility headers are specified in the CMakeLists.txt of the module, it would install the compatibility headers etc., if none are specified then nothing is done. Did I miss something?
It was to match the modules like the config ones where there is no compatibility headers.

msimberg: The behaviour is correct, I was just proposing an alternative, but it might be subjective if it's clearer. Instead of
```
if(${modulename}_COMPAT_HEADERS)
  set(_have_compatibility_headers TRUE)
else()
  set(_have_compatibility_headers FALSE)
endif()

# later
if(${_have_compatibility_headers})
  # do something if compatibility headers are there
endif()
```
I would just do 
```
# later
if(${modulename}_COMPAT_HEADERS)
  # do something if compatibility headers are there
endif()
```
Please keep whichever one you find clearer though. I might be the odd one on this.

aurianer: Sure initially we just did that to keep the install steps together, etc. but we can totally do that!

msimberg: Note that the install steps can still be together. I just mean that the extra `_have_compatibility_headers` variable is unnecessary (to me).

aurianer: Yep okay I don't have any strong preference, but you're right that it's probably clearer. Sorry for the confusion

PR #5194, opened 2021-02-22
Another attempt at fixing the Fedora 35 problem

hkaiser:
On Fedora rawhide we get the following error on power and s390

```
In file included from /usr/include/signal.h:315,
                 from /usr/include/sys/param.h:28,
                 from /builddir/build/BUILD/hpx-1.6.0/libs/core/coroutines/include/hpx/coroutines/detail/posix_utility.hpp:55,
                 from /builddir/build/BUILD/hpx-1.6.0/libs/core/coroutines/include/hpx/coroutines/detail/context_posix.hpp:164,
                 from /builddir/build/BUILD/hpx-1.6.0/libs/core/coroutines/include/hpx/coroutines/detail/context_impl.hpp:129,
                 from /builddir/build/BUILD/hpx-1.6.0/libs/core/coroutines/include/hpx/coroutines/detail/context_base.hpp:44,
                 from /builddir/build/BUILD/hpx-1.6.0/libs/core/coroutines/src/detail/context_base.cpp:12:
/builddir/build/BUILD/hpx-1.6.0/libs/core/coroutines/include/hpx/coroutines/detail/context_posix.hpp:224:62: error: call to non-'constexpr' function 'long int sysconf(int)'
  224 |             static std::ptrdiff_t const default_stack_size = SIGSTKSZ;
```

More details

https://koji.fedoraproject.org/koji/watchlogs?taskID=62453133

PR #5197, opened 2021-02-22
Attempt to fix compilation of context implementations with unity build enabled

msimberg:
I merged #5194 prematurely. The unity build configuration did not work with the given changes, and this attempts to fix it (cscs/clang-newest tests this and is the one which should pass if this is correct).


hkaiser (approved):
LGTM, thanks! Excellent catch!

aurianer (approved):
Thanks!

PR #5199, opened 2021-02-23
Update ICC module name on Piz Daint Jenkins configuration

msimberg:


msimberg:
So the icc builds on Piz Daint were based on a lie... Which is hopefully fixed now. The intel module was unloaded because of later modules, meaning the builds were done with gcc. Now the build actually uses icc. @NK-Nikunj I've had to disable some resiliency examples because of what seems to be a dumb icc (and they're disabled only with icc). Just letting you know in case you ever bump into someone asking about using the resiliency stuff with icc.

NK-Nikunj:
@msimberg really curious what's breaking them on icc? Thanks for the heads up btw.

msimberg:
> @msimberg really curious what's breaking them on icc? Thanks for the heads up btw.

It claims there are no matching overloads of `std::forward` for the function pointers passed to `dataflow_replay/validate`. I did not look further into if it's fixable, but I'm pretty sure it's icc that is wrong, and not the other compilers. It could also be that it's fixed in newer iccs. This was with 19.1.

NK-Nikunj:
I see. Yeah, there's nothing much we can do about it there.

msimberg:
retest cscs

hkaiser (approved):
LGTM, thanks!

msimberg:
The ICC configuration now successfully runs with `HPX_WITH_NETWORKING=OFF`. It's not perfect, but better than before. Any comments on this?

PR #5203, opened 2021-02-25
Move copy algorithms to tag_fallback_invoke

srinivasyadav18:
## Proposed Changes

  - replaced tag_fallback_invoke header
  - replaced tag   with   tag_fallback
  - replaced tag_invoke    with   tag_fallback_invoke

From PR #5156

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/algorithms/copy.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/algorithms/copy.hpp
@@ -691,8 +691,8 @@ namespace hpx {
         // clang-format on
         friend typename hpx::parallel::util::detail::algorithm_result<ExPolicy,
             FwdIter2>::type
-        tag_fallback_invoke(hpx::copy_n_t, ExPolicy&& policy, FwdIter1 first, Size count,
-            FwdIter2 dest)
+        tag_fallback_invoke(hpx::copy_n_t, ExPolicy&& policy, FwdIter1 first,
+            Size count, FwdIter2 dest)
         {
             static_assert((hpx::traits::is_forward_iterator<FwdIter1>::value),
                 "Required at least forward iterator.");
--- a/libs/parallelism/algorithms/include/hpx/parallel/algorithms/copy.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/algorithms/copy.hpp
@@ -204,7 +204,7 @@ namespace hpx {
 #include <hpx/assert.hpp>
 #include <hpx/concepts/concepts.hpp>
 #include <hpx/functional/invoke.hpp>
-#include <hpx/functional/tag_invoke.hpp>
+#include <hpx/functional/tag_fallback_invoke.hpp>
 #include <hpx/iterator_support/traits/is_iterator.hpp>
 
 #include <hpx/algorithms/traits/projected.hpp>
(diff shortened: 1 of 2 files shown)


msimberg:
retest

msimberg:
@srinivasyadav18 would you mind doing the same in `container_algorithms/copy.hpp`?

srinivasyadav18:
> @srinivasyadav18 would you mind doing the same in `container_algorithms/copy.hpp`?

Yes, I would complete it soon.

hkaiser (approved):
LGTM, thanks!

libs/parallelism/algorithms/include/hpx/parallel/algorithms/copy.hpp
```diff
@@ -635,7 +635,7 @@ namespace hpx {
     ///////////////////////////////////////////////////////////////////////////
```
aurianer, line 635: @hkaiser Should we also add this comment here?
https://github.com/STEllAR-GROUP/hpx/pull/5176/commits/15868e69671312c08d0fa8cd8a37a0c86be86e7c#diff-c3460ebee73b757aa57b1f346ca7a8152792df29f292d6aaeb888af82c5fdb16R596-R602

aurianer (commented):
LGTM thanks!

libs/parallelism/algorithms/include/hpx/parallel/algorithms/copy.hpp
hkaiser: Good thinking!

msimberg: Would it be preferable to add an explanation in one place (`tag_fallback_invoke.hpp`) on why one would want to use `tag_fallback_invoke`, instead of copying that comment into every algorithm file?

gonidelis:
@srinivasyadav18 Thanks for that :)

libs/parallelism/algorithms/include/hpx/parallel/algorithms/copy.hpp
hkaiser: That would work as well. My initial rationale was that people look at the algorithms and wonder why in the world did they use `tag_fallback_invoke`. But I guess there are many places people would scratch their heads over when looking at our code ;-)

msimberg: I 100% agree with the rationale. I just suspect that that comment is at some point going to be modified in one place, but not in the 50 other algorithms. I can move that comment to the `tag_fallback_invoke.hpp` file (I'll add a bit more of an explanation to `tag_priority_invoke.hpp` as well). I don't feel very strongly about this in the end though. Preference?

msimberg:
retest

libs/parallelism/algorithms/include/hpx/parallel/algorithms/copy.hpp
hkaiser: Sure, please feel free to do that. No objections here.

msimberg: @srinivasyadav18 that means you're good with this PR ;)

srinivasyadav18: @msimberg Not completely, still need get a clear picture of tag_invoke and tag_fallback_invoke, I am trying to understand the change and build fail on macOS, but for the comment even I felt the same, adding it in tag_fallback_invoke.hpp would be better, instead of repeating in all other algorithms (as you mentioned above).

msimberg: The macos failures are unrelated to these changes. They're due to lack of macos knowledge/developers among us to actually fix those.

tag_invoke and friends are best understood with https://wg21.link/p1895, our implementation, a debugger, and lots of questions.

hkaiser:
@srinivasyadav18 thank you again for your contribution!

PR #5207, opened 2021-02-26
Always define HPX_COMPUTE_HOST_CODE for host code

msimberg:
Always define `HPX_COMPUTE_HOST_CODE` for host code, instead of only when compiling with nvcc or hipcc. This is a subtle change, which may in the worst case break someone's code, but I think it's mostly safe. This means that one doesn't have to do `#if !defined(HPX_COMPUTE_DEVICE_CODE)` when one wants something to be compiled only for the host, and `#if defined(HPX_COMPUTE_HOST_CODE)` will be enough (and the two are equivalent after with this PR). I've also cleaned up some other #ifdefs that I think didn't do right in the first place.


hkaiser (approved):
LGTM, thanks!

aurianer (approved):
Looks good thanks! Should we wait until hipcc CI is functioning again?
EDIT: actually not necessary as the build step is working

PR #5213, opened 2021-03-01
Removing ESSENTIAL option to the doc example

aurianer:
Looks like this option is not parsed in `add_hpx_executable` so simply removing it.


hkaiser (approved):
LGTM, thanks!

PR #5214, opened 2021-03-02
Fix clang-format error in sender.hpp

aurianer:


hkaiser:
@aurianer thanks for cleaning up my mess....

libs/core/execution_base/include/hpx/execution_base/sender.hpp
```diff
@@ -503,7 +503,7 @@ namespace hpx { namespace execution { namespace experimental {
 
                 template <typename E,
                     typename = std::enable_if_t<is_receiver_v<R, E>>>
-                void set_error(E&& e) && noexcept
+                    void set_error(E&& e) && noexcept
```
msimberg, line 506: That's an interesting formatting decision by clang-format... Let's keep it like this though, maybe a newer clang-format makes a more reasonable decision here.

And thank you @aurianer for fixing this!

PR #5226, opened 2021-03-08
Use new docker image in all CircleCI steps

msimberg:


.circleci/config.yml
```diff
@@ -5,8 +5,10 @@
 #  Distributed under the Boost Software License, Version 1.0. (See accompanying
 #  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-defaults: &defaults
+working_dir_default: &working_dir_default
```
aurianer, line 8: Isn't it better to keep the target `defaults` depending on `working_dir_default` and `docker_default`?
```
defaults: &defaults
    <<: *docker_default
    <<: *working_dir_default
```

hkaiser (approved):
LGTM, thanks!

aurianer:
Should we also update this one:
https://github.com/STEllAR-GROUP/hpx/pull/5226/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R83 ?

msimberg:
> Should we also update this one:
> https://github.com/STEllAR-GROUP/hpx/pull/5226/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R83 ?

Definitely, good catch!

aurianer (approved):
LGTM thanks!

msimberg:
@biddisco note that I've disabled the `scheduler_binding_check` on CircleCI as it was very consistently timing out. It may be that it just takes a long time to run (the timeout on CircleCI is now 120 seconds and the latest run I saw it pass it took 73 seconds). Would you mind having a look at it to see if it could be made to run a bit faster (possibly only in debug mode).

PR #5227, opened 2021-03-08
Update .codespell_whitelist

msimberg:
Changes the "Vill" entry to be in lower case. Codespell has an interesting way of matching whitelisted words... Since the dictionary contains "will" in lower case the whitelisted word also has to be in lower case.


hkaiser (approved):
LGTM, thanks!

PR #5228, opened 2021-03-08
Print module name only if it has custom configuration settings

hkaiser:


msimberg (approved):
Good idea, thanks!

aurianer (approved):
Thanks! Forgot that in my module options PR..

PR #5230, opened 2021-03-09
Fix C++20 replace algo adaptation misses

gonidelis:
MInor `hpx::replace` and `hpx::ranges::replace` misses from previous adaptation.

1. Docs signature matching.
2. Add tests for the no-exec-policy overloads.
3. Minor typo in deprecated overload.


aurianer:
retest lsu

PR #5231, opened 2021-03-09
Sort printed config options

aurianer:
Makes the config options more readable as alphabetically ordered, if nobody is against it..


hkaiser (approved):
Should we do the same while printing the configuration options at runtime (here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/version/src/version.cpp#L203-L217)?

hkaiser (dismissed):
LGTM, thanks!

msimberg:
Just needs formatting. Thanks for doing this!

PR #5233, opened 2021-03-10
Only find Boost.Iostreams if required for plugins

msimberg:


hkaiser (commented):
Don't we need Boost.Iostreams for our iostream component as well?

msimberg:
> Don't we need Boost.Iostreams for our iostream component as well?

Good question. Apparently not given that it hasn't been linked so far. It does need the headers, but it'd be better to link explicitly to the Boost::iostreams since it's mostly luck if the headers were enough. I'll change that.

hkaiser (dismissed):
LGTM, thanks!

aurianer (approved):
LGTM thanks!

msimberg:
> > Don't we need Boost.Iostreams for our iostream component as well?
> 
> Good question. Apparently not given that it hasn't been linked so far. It does need the headers, but it'd be better to link explicitly to the Boost::iostreams since it's mostly luck if the headers were enough. I'll change that.

I changed my mind on this. Given that iostreams pulls in an unnecessary regex dependency, and they document which headers can be used with header-only boost (https://www.boost.org/doc/libs/1_75_0/libs/iostreams/doc/index.html), I think it's ok to not link to the iostreams library. I take back my statement that it worked mostly through luck...

PR #5234, opened 2021-03-10
Fix small typo in --hpx:local option description

msimberg:


aurianer (approved):
LGTM thanks!

libs/full/command_line_handling/src/parse_command_line.cpp
```diff
@@ -436,7 +436,7 @@ namespace hpx { namespace util {
             // Always add the option to start the local runtime
             hpx_options.add_options()("hpx:local",
                 "run this instance in local mode (experimental; certain "
-                "functionality not available at runt-time)");
+                "functionality not available at run-time)");
```
hkaiser, line 439:
```suggestion
                "functionalities are not available at runtime)");
```

PR #5235, opened 2021-03-10
Attempting to fix datapar compilation issues

hkaiser:


msimberg:
I rebased this and added a few small changes (namespace and ifdefs). I also added vc to the docker image and enabled the datapar option on CircleCI (all the builds are currently failing because the image didn't get pushed yet, they'll have to be retriggered). Most tests seemed to compile fine after these changes. Some had longer error messages that I didn't attempt to fix. It looked like they were related to the ranges adaptations (error messages about things like `in_out_result`).

hkaiser:
retest

msimberg:
@hkaiser note that "retest" only triggers jenkins to restart jobs. I've now restarted the other builds (that actually use the new image). It's anyways only the CircleCI job that is interesting and that is likely to fail spectacularly since I think I didn't fix all the deprecation warnings (warnings as errors).

srinivasyadav18:
Why none of the builds still did not use vc?

hkaiser:
> Why none of the builds still did not use vc?

We have added Vc for the CircleCI builder only.

hkaiser:

msimberg:
> @msimberg I think this is good to go now.

Thanks @hkaiser and @srinivasyadav18 for looking into this. I'll push one more commit to fix some of the docker image tags.

msimberg:
@srinivasyadav18 is the segmentation fault we see here: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/7321/workflows/3ef4667c-eb7c-48c2-8fce-035925e0b12a/jobs/246486/steps something you can reproduce locally? It may simply be a buggy test, or an actual bug in the datapar implementation.

I'd be ok with having that test looked at in a separate PR given that this PR is already a huge improvement to what we had before, but I'd like an issue about it (and possibly a point in @srinivasyadav18's GSoC proposal?) in that case.

srinivasyadav18:
> SoC proposal?)



> @srinivasyadav18 is the segmentation fault we see here: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/7321/workflows/3ef4667c-eb7c-48c2-8fce-035925e0b12a/jobs/246486/steps something you can reproduce locally? It may simply be a buggy test, or an actual bug in the datapar implementation.
> 
> I'd be ok with having that test looked at in a separate PR given that this PR is already a huge improvement to what we had before, but I'd like an issue about it (and possibly a point in @srinivasyadav18's GSoC proposal?) in that case.

Ok, sure. I will try reproducing the bug locally and solve it and put a new PR for this.

hkaiser:
@msimberg I think this can be merged now. Any objections?

PR #5236, opened 2021-03-10
Add an example demonstrating the use of the invoke_function_action facility

hkaiser:
- also applies fixes to the facility itself.

### DIFF
--- a/examples/quickstart/invoke_function.cpp
+++ b/examples/quickstart/invoke_function.cpp
@@ -1,0 +1,52 @@
+//  Copyright (c) 2021 Hartmut Kaiser
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// This example demonstrates the use of the facility invoke_function_action that
+// allows to wrap arbitrary global functions into an action. Please note: this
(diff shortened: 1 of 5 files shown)


msimberg (dismissed):

examples/quickstart/invoke_function.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+        std::cout << "the action invocation returned: " << result.get() << "\n";
+    }
+
+    {
+        using action_type = hpx::components::server::invoke_function_action<
+            decltype(&void_call_me), int>;
+
+        hpx::future<void> result = hpx::async<action_type>(hpx::find_here(),
+            reinterpret_cast<std::ptrdiff_t>(&void_call_me), 42);
```
aurianer, line 46: Can you explain why do you use `ptrdiff_t`? I don't know this well..

aurianer (dismissed):
Thanks!

examples/quickstart/invoke_function.cpp
hkaiser: `std::size_t` would have worked as well. Both types are guaranteed to be large enough integral types to be used for roundtrip casting pointer->integral->pointer.

hkaiser: However it looks like that gcc and clang are not happy with this choice. I'll change this to `std::size_t`.

gonidelis:
Late comment: Thanks for that!

PR #5238, opened 2021-03-12
Make sure annotations are used in the binary transform

### LINKED ISSUE #5232: Annotation using hpx::util::annotated_function not working
## Expected Behavior

Annotating a function using `hpx::util::annotated_function` should have the same effect as using `hpx::util::annotate_function` inside of the function.

## Actual Behavior

Annotation via `hpx::util::annotated_function` does not seem to work at all, the information is lost at some point and not passed on to the instrumentation framework APEX.

An example using the two methods to annotate a lambda function and the respective output is available [here](https://gist.github.com/severinstrobl/23db96ccf205684e5d4145c8847cb166).

## Specifications

  - HPX Version: 1.6.0
  - Platform (compiler, OS): GCC 10.2.0, Linux
  - HPX Configuration:
  ```shell
-DHPX_WITH_CXX14:BOOL=ON -DHPX_WITH_MALLOC:STRING=system -DHPX_WITH_CUDA:BOOL=OFF -DHPX_WITH_TOOLS:BOOL=OFF -DHPX_WITH_EXAMPLES:BOOL=OFF -DHPX_WITH_ASYNC_MPI:BOOL=ON -DHPX_WITH_ASYNC_CUDA:BOOL=OFF -DHPX_WITH_TESTS:BOOL=OFF -DHPX_WITH_NETWORKING:BOOL=ON -DHPX_WITH_PARCELPORT_TCP:BOOL=OFF -DHPX_WITH_PARCELPORT_MPI:BOOL=ON -DHPX_WITH_MAX_CPU_COUNT:STRING=64 -DHPX_WITH_GENERIC_CONTEXT_COROUTINES:BOOL=OFF -DHPX_WITH_BOOST_ALL_DYNAMIC_LINK:BOOL=ON -DBUILD_SHARED_LIBS:BOOL=ON -DHPX_DATASTRUCTURES_WITH_ADAPT_STD_TUPLE:BOOL=OFF -DHPX_WITH_VALGRIND:BOOL=OFF -DHPX_WITH_APEX:BOOL=ON -DHPX_WITH_GOOGLE_PERFTOOLS:BOOL=OFF -DHPX_WITH_PAPI:BOOL=OFF -DAPEX_WITH_OTF2:BOOL=ON
  ```

msimberg:
Fixes #5232. @severinstrobl would you mind checking that this works for you as well?

Note: there may be other places that need this treatment, but I'm doing only this one for now.

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform.hpp
@@ -759,6 +759,32 @@ namespace hpx { namespace traits {
         }
     };
 
+    template <typename ExPolicy, typename F, typename Proj1, typename Proj2>
+    struct get_function_address<
+        parallel::v1::detail::transform_binary_iteration<ExPolicy, F, Proj1, Proj2>>
+    {
+        static std::size_t call(
+            parallel::v1::detail::transform_binary_iteration<ExPolicy, F, Proj1, Proj2> const&
+                f) noexcept
+        {
+            return get_function_address<typename std::decay<F>::type>::call(
+                f.f_);
+        }
+    };
+
+    template <typename ExPolicy, typename F, typename Proj1, typename Proj2>
+    struct get_function_annotation<
+        parallel::v1::detail::transform_binary_iteration<ExPolicy, F, Proj1, Proj2>>
+    {
+        static char const* call(
+            parallel::v1::detail::transform_binary_iteration<ExPolicy, F, Proj1, Proj2> const&
+                f) noexcept
+        {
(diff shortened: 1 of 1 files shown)


hkaiser (approved):
@msimberg thanks for looking into this. LGTM. Just a question, though - would it make sense to make those traits constexpr? I think most of the time they will be invoked with specific (const) functions, thus constexpr might actually reduce the runtime overheads...

severinstrobl:
Thanks for the quick fix, I'll recompile and give it a try.

severinstrobl:
This fully resolves the issue for my test case. :+1:

PR #5241, opened 2021-03-13
Adapt min_element, max_element and minmax_element to C++20

Jedi18:
Adapts min_element, max_element and minmax_element to C++20.
Adds min_max_result result type
Add segmented algorithm tests for min_element, max_element and minmax_element

https://github.com/STEllAR-GROUP/hpx/issues/4822
https://github.com/STEllAR-GROUP/hpx/issues/5156


msimberg:
retest

hkaiser:
retest

hkaiser (approved):
Sorry it took so long to look at this! This looks good to me and can go in. Thanks a lot, good work!

hkaiser:
@Jedi18 could you please rebase this onto master for a last time, we'll merge it afterwards.

Jedi18:
Thanks! Ok done, rebased onto master

hkaiser:
retest

hkaiser:
We now see more errors popping up (e.g. here: https://cdash.cscs.ch/viewBuildError.php?buildid=155130) I think this might be caused by those configurations testing things with HPX_NETWORKONG=OFF and HPX_WITHDISTRIBUTED_RUNTIME=OFF.

libs/full/segmented_algorithms/tests/unit/partitioned_vector_max_element2.cpp
```diff
@@ -0,0 +1,86 @@
...
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/config.hpp>
+#if !defined(HPX_COMPUTE_DEVICE_CODE)
+#include <hpx/hpx_main.hpp>
+#include <hpx/include/parallel_minmax.hpp>
+#include <hpx/include/partitioned_vector.hpp>
```
msimberg, line 11: For the errors about `hpx::find_all_localities` not being defined, please include `hpx/runtime.hpp` here (and in the other tests that need it).

msimberg (changes requested):

libs/full/segmented_algorithms/tests/unit/partitioned_vector_max_element2.cpp
Jedi18: Oh ok, thanks

msimberg:
retest

msimberg:
@Jedi18 is the failure here: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/7316/workflows/1cf271c4-1bf8-4ffe-9ce3-d45c4241a24a/jobs/246399/steps something you can reproduce locally?

Jedi18:
@msimberg No, it runs successfully locally.

Jedi18:
@msimberg I'm not sure if the fault lies with my changes or if the bug already existed (and since there were no tests before they were never discovered). What do you think could be the reason?

msimberg:
> @msimberg No, it runs successfully locally.

It seems to be nondeterministic, but quite easily reproducible. I can usually reproduce it after a few runs (`ctest --repeat-until-fail` is quite handy for checking if one can reproduce nondeterministic failures).

> @msimberg I'm not sure if the fault lies with my changes or if the bug already existed (and since there were no tests before they were never discovered). What do you think could be the reason?

It's indeed most likely something that has been buggy for a long time. Would you in any case mind giving it another go (don't sweat it)? If you can't find the problem right now we should disable the test and open an issue about it so that someone (tm) will look at it later.

Jedi18:
@msimberg Thanks! I could reproduce it locally too after using --repeat-until-fail. I'm not sure how to debug what's wrong since we don't know when it will fail. Anyway I tried stepping through it again but still couldn't find the bug.

aurianer:
I think the good practice here is to leave the original copyright and add yours to account for your modifications :)
https://github.com/Jedi18/hpx/commit/41eee62ceafc2a39e605e2d40d27eb75149efc5f#diff-9d4c0b9fca9871b099beb63e0578da31ca4928621d4c6411f9c450e30d79a5f0R1

Jedi18:
@aurianer Sorry about that, actually I'd copied an old test file when creating this new test file and made changes to it, but since this was a new file I figured I shouldn't add the old copyright.

hkaiser:
@Jedi18 what's the status of this PR? Can we go ahead with this?

Jedi18:
@hkaiser Yeah the PR should be ready. Couldn't find the bug so like msimberg said we could disable the test and open an issue for it. Apart from that, I've made all the requested changes.

hkaiser:
retest

hkaiser:
@Jedi18 sorry for the delay. Could you have a look at this again and resolve the conflict. This is good to go, so once you do that we can merge.

Jedi18:
@hkaiser No problem, resolved the conflicts. Don't we need to disable the segmented tests?

hkaiser:
> Don't we need to disable the segmented tests?

@Jedi18 I have pushed some changes that should fix the remaining issues. Let's see.

hkaiser:
retest lsu

hkaiser:
retest lsu

hkaiser:
retest lsu

hkaiser:
@msimberg everything looks good here except for random test failures on jenkins/cscs/clang-newest. I'm not able to reproduce these, would you have the time to have a look?

msimberg:
> @msimberg everything looks good here except for random test failures on jenkins/cscs/clang-newest. I'm not able to reproduce these, would you have the time to have a look?

No promises. Do you have any guesses about what might be going wrong? I see you pushed something but that doesn't fix it completely (or all of the failures).

hkaiser:
> > @msimberg everything looks good here except for random test failures on jenkins/cscs/clang-newest. I'm not able to reproduce these, would you have the time to have a look?
> 
> No promises. 

No worries. 

> Do you have any guesses about what might be going wrong? 

Since only the segmented algorithms fail I was assuming that there is an error in the functions that consolidate the results returned from the different partitions (i.e. any of the functions `sequential_minmax_element_ind`). OTOH, it may be a race condition somewhere as it fails inconsistently. The segmented minmax tests were added by this PR only, thus the problem might have been there from the start.

> I see you pushed something but that doesn't fix it completely (or all of the failures).

I was hoping that my changes would fix it, but hey - no luck.

msimberg:
@hkaiser, @Jedi18 I don't know how much this helps, but the problem is reproducible in debug as well. It just takes typically a bit longer for it to reproduce. It also works with the sequential version, so it's not a race in the local implementation. I haven't tried with other compilers. Perhaps this will give you something to look into.

Here's the output of some manual logging in `max_element` (it's not only `minmax_element` that's problematic) of a failed run (this happened after 55 repetitions):

```
invoking action                                                    
max_element::sequential                                                      
checking elements: 2 * 3 * 4 * 5 * 6 * 2 3 3 5 5 3 4 2 3 2 6 * 2 3 4 5 6 * 5 6 * 6 * 2 3 4 6 * 6 * 2 3                                                    
waited for result                                                            
invoking action                                                    
max_element::sequential                                                      
checking elements: 5 * 4 3 2 6 * 6 * 2 0 4 0 2 3 0 0 0 0 6 * 0 0 0 0 0 5 0 0 0 0 0 0 0 2 
waited for result
```

This runs on two nodes, so there are two calls to `max_element`. The local (first) partition is completely fine. The second partition contains partly garbage data (the zeros are wrong, and there should be an `8` there). The asterisks mark when the algorithm found a new max value.

There are at least two problems:
- The data is garbage on the second locality
- The algorithm should return the *first* element that is largest, if there are more than one; it now returns the last

@Jedi18 is the second something you'd be interested in fixing? @hkaiser does this tell you anything? I'd imagine this is possibly related to the various hangs we see in the partitioned vector tests (this one also hangs every now and then in addition to producing garbage results).

For comparison, a correct run looks like this:

```
invoking action                                                    
max_element::sequential                                                      
checking elements: 2 * 3 * 4 * 5 * 6 * 2 3 3 5 5 3 4 2 3 2 6 * 2 3 4 5 6 * 5 6 * 6 * 2 3 4 6 * 6 * 2 3                                                    
waited for result                                                            
invoking action                                                    
max_element::sequential                                                      
checking elements: 5 * 4 3 2 6 * 6 * 2 3 4 6 * 2 3 6 * 6 * 6 * 6 * 6 * 6 * 6 * 6 * 7 * 6 5 8 * 5 4 2 3 4 5 2                                              
waited for result
```

Jedi18:
@msimberg Thanks a lot! That is very helpful.
Sure I'll fix it to return the first element that is largest, and also add a test to check for this (and other similar edge cases)

hkaiser:
Thanks @msimberg, I'll try to have a look.

hkaiser:
> Sure I'll fix it to return the first element that is largest, and also add a test to check for this (and other similar edge cases)

@Jedi18 That was me breaking `max_element` (and probably `minmax_element` as well). I somehow assumed that it should return the last of the `max` elements (I think I read it somewhere in Stepanov's notes, but there was also something about `max` being broken in the standard...)

Jedi18:
@hkaiser Actually the minmax_element was already broken [Current minmax on master](https://github.com/STEllAR-GROUP/hpx/blob/master/libs/parallelism/algorithms/include/hpx/parallel/algorithms/minmax.hpp#L499) . I guess this explains some of the test failures.

hkaiser:
@Jedi18 FWIW, I have squashed all commits and rebased the branch onto master. This should simplify staying up to date in the future.

hkaiser:
I think this is fine now, the problems have been resolved. Let's wait for the last round of tests to come back and merge if all is green.

Jedi18:
Thanks a lot! I guess the LSU tests aren't running because rostam is down?

msimberg:
This just needs a clang-format round... @Jedi18 or @hkaiser could you apply that?

Incidentally, @hkaiser, I see changes to the channel communicator example. I assume that's otherwise unrelated to the changes here(...), but related to the timeouts that happen every now and then on that example?

hkaiser:
> This just needs a clang-format round... @Jedi18 or @hkaiser could you apply that?

Done.

> Incidentally, @hkaiser, I see changes to the channel communicator example. I assume that's otherwise unrelated to the changes here(...), but related to the timeouts that happen every now and then on that example?

Yes.

msimberg:
> > Incidentally, @hkaiser, I see changes to the channel communicator example. I assume that's otherwise unrelated to the changes here(...), but related to the timeouts that happen every now and then on that example?
> 
> Yes.

All right, thanks for looking into that!

bors merge

msimberg (approved):

msimberg:
bors merge

PR #5244, opened 2021-03-16
P1897 when_all

msimberg:
Adds the `when_all` algorithm from P1897. Contains #5237 and #5245.


msimberg:
Note that this is based on `std::variant` and `std::optional`. I'd prefer not to have to add our own `variant` and would be happy for this to be C++17-only, but I'm open to alternatives.

hkaiser:
> Note that this is based on `std::variant` and `std::optional`. I'd prefer not to have to add our own `variant` and would be happy for this to be C++17-only, but I'm open to alternatives.

We're planning to move to require C++17 anyways. let's have it C++17 only for now.

hkaiser (approved):
LGTM, thanks!

PR #5245, opened 2021-03-17
P1897 operator| overloads

msimberg:
Adds `operator|` overloads for `on`, `sync_wait`, and `transform`. The other algorithms don't really benefit from an `operator|` overload.

Also builds on #5237.


msimberg:
I'm going to merge this one directly instead of #5237 first as this PR fixes compilation problems on #5237 and I was going to merge them both at the same time in any case.

hkaiser:
Thanks for working on the sender/receiver facilities, Mikael!

PR #5247, opened 2021-03-18
Significantly speedup simple for_each, for_loop, and transform

hkaiser:
foreach_scaling_test on 10 cores:
```
----------------Parameters---------------------
Vector size:                       10000000
Number of tests                          10
Delay per iteration(nanoseconds)          0
Display time in:                    Seconds
-------------Average-(for)---------------------
Average execution time           : 0.00880673
Average execution time (iter)    : 0.00880363
-------------Average-(for_each)----------------
Average parallel execution time  : 0.00085571
Average task execution time      : 0.00067213
Average sequential execution time: 0.00537611
-----Execution Time Difference-(for_each)------
Parallel Scale:                     6.28263
Task Scale    :                     7.99862
-------------Average-(for_loop)----------------
Average parallel execution time  : 0.00059376
Average task execution time      : 0.00060955
Average sequential execution time: 0.00485463
-----Execution Time Difference-(for_loop)------
Parallel Scale:                     8.17608
Task Scale    :                     7.96429
```
Simple sequential `hpx::for_each` and `hpx::for_loop` are now almost twice as fast as plain equivalent `for()`

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/algorithms/for_each.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/algorithms/for_each.hpp
@@ -1,4 +1,4 @@
-//  Copyright (c) 2007-2017 Hartmut Kaiser
+//  Copyright (c) 2007-2021 Hartmut Kaiser
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
@@ -259,27 +259,27 @@ namespace hpx {
 #include <utility>
 
 namespace hpx { namespace parallel { inline namespace v1 {
+
     ///////////////////////////////////////////////////////////////////////////
     // for_each_n
     namespace detail {
         /// \cond NOINTERNAL
-        template <typename F, typename Proj>
+        template <typename F, typename Proj = util::projection_identity>
         struct invoke_projected
         {
             F& f_;
             Proj& proj_;
 
             template <typename T>
-            HPX_HOST_DEVICE HPX_FORCEINLINE typename std::enable_if<
+            HPX_HOST_DEVICE HPX_FORCEINLINE constexpr typename std::enable_if<
                 !hpx::traits::is_value_proxy<T>::value>::type
             call(T&& t)
             {
-                T&& tmp = std::forward<T>(t);
-                HPX_INVOKE(f_, HPX_INVOKE(proj_, tmp));
+                HPX_INVOKE(f_, HPX_INVOKE(proj_, std::forward<T>(t)));
             }
 
             template <typename T>
-            HPX_HOST_DEVICE HPX_FORCEINLINE typename std::enable_if<
+            HPX_HOST_DEVICE HPX_FORCEINLINE constexpr typename std::enable_if<
                 hpx::traits::is_value_proxy<T>::value>::type
             call(T&& t)
             {
@@ -288,7 +288,7 @@ namespace hpx { namespace parallel { inline namespace v1 {
             }
 
             template <typename Iter>
-            HPX_HOST_DEVICE HPX_FORCEINLINE void operator()(Iter curr)
+            HPX_HOST_DEVICE HPX_FORCEINLINE constexpr void operator()(Iter curr)
             {
                 call(*curr);
             }
@@ -297,7 +297,13 @@ namespace hpx { namespace parallel { inline namespace v1 {
         template <typename F>
         struct invoke_projected<F, util::projection_identity>
         {
-            HPX_HOST_DEVICE invoke_projected(F& f, util::projection_identity)
+            HPX_HOST_DEVICE constexpr invoke_projected(F& f) noexcept
+              : f_(f)
+            {
+            }
+
+            HPX_HOST_DEVICE constexpr invoke_projected(
+                F& f, util::projection_identity) noexcept
               : f_(f)
             {
             }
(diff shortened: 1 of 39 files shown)


msimberg:
Nice, thanks for looking into this @hkaiser! Do you know which part makes the biggest difference? Is it mainly the specializations for `projection_identity` or the manual loop unrolling? In any case, this is a nice improvement. I don't quite understand why this would be faster than the plain for loop, except for maybe the loop unrolling? There shouldn't be any vectorization going on.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/for_each.hpp
```diff
@@ -376,8 +376,59 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+            {
+                hpx::util::annotate_function annotate(f_);
+                execute(part_begin, part_size);
+            }
+        };
+
+        template <typename ExPolicy, typename F>
+        struct for_each_iteration<ExPolicy, F, util::projection_identity>
+        {
+            using execution_policy_type = typename std::decay<ExPolicy>::type;
```
msimberg, line 391: Not necessary to change for this PR, but since we're requiring C++14 support I've started using `decay_t` (and other `_t`s where available).

tests/performance/local/foreach_scaling.cpp
```diff
@@ -71,6 +71,24 @@ namespace hpx { namespace parallel { namespace execution {
...
 }}}    // namespace hpx::parallel::execution
 
+///////////////////////////////////////////////////////////////////////////////
+void measure_plain_for(std::vector<std::size_t> const& data_representation)
+{
+    std::size_t size = data_representation.size();
+    for (std::size_t i = 0; i != size; ++i)
+    {
+        worker_timed(delay);
+    }
```
msimberg, line 81: Might be worth trying? 
```suggestion
    while (size >= 4)
    {
        worker_timed(delay);
        worker_timed(delay);
        worker_timed(delay);
        worker_timed(delay);
        size -= 4;
    }
    // handle the rest like in the hpx for loops
```

hkaiser: Yes, true - this will bring things on par. Good catch.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/for_each.hpp
hkaiser: Yes, I think we should do that in a big sweep.

hkaiser:
> Nice, thanks for looking into this @hkaiser! Do you know which part makes the biggest difference? Is it mainly the specializations for `projection_identity` or the manual loop unrolling? 

The biggest effect came from specializing for `stride == 1` here: https://github.com/STEllAR-GROUP/hpx/pull/5247/files#diff-0ce47261fe8063baf12e10964c90a4ea64c484e38826ccb396c4a386cc25a980R932-R941 and moving the dereferencing of the iterator from the `invoke_projected` into the inner (unrolled) loop.

>I don't quite understand why this would be faster than the plain for loop, except for maybe the loop unrolling? There shouldn't be any vectorization going on.

Yes, it's the lop unrolling.

tests/performance/local/foreach_scaling.cpp
hkaiser: That's changed now and indeed the plain `for` loop is as fast as our algorithm, as expected:
```
----------------Parameters---------------------
Vector size                       : 10000000
Number of tests                   :       10
Delay per iteration (nanoseconds) :        0
Display time in                   : Seconds
-------------Average-(for)---------------------
Average execution time (unrolled) : 0.00547917
Average execution time (iter)     : 0.00877946
-------------Average-(for_each)----------------
Average parallel execution time   : 0.00158009
Average task execution time       : 0.00069507
Average sequential execution time : 0.00485273
-----Execution Time Difference-(for_each)------
Parallel Scale                    :  3.07117
Task Scale                        :  6.98164
-------------Average-(for_loop)----------------
Average parallel execution time   : 0.00060268
Average task execution time       : 0.000612
Average sequential execution time : 0.00490469
-----Execution Time Difference-(for_loop)------
Parallel Scale                    :  8.13813
Task Scale                        :   8.0142
```

hkaiser:
retest

hkaiser:
@msimberg once the test have cycled, this is good to go.

tests/performance/local/foreach_scaling.cpp
gonidelis: @hkaiser  Sorry for bringing up this late, but I think I am missing something: why is "Average parallel execution time" for `for_each` bigger than your initial (improved) measurements? Seems like `for_each` Parallel Scale is doubled down...

hkaiser: Good question, let me check this.

libs/parallelism/algorithms/include/hpx/parallel/util/loop.hpp
```diff
@@ -297,66 +352,258 @@ namespace hpx { namespace parallel { namespace util {
...
+                it, count, tok, std::forward<F>(f), pred());
+        }
+    };
 
-        return detail::loop_n<Iter>::call(
-            it, count, std::forward<F>(f), pred());
+#if !defined(HPX_COMPUTE_DEVICE_CODE)
+    template <typename ExPolicy>
+    HPX_INLINE_CONSTEXPR_VARIABLE loop_n_t<ExPolicy> loop_n =
+        loop_n_t<ExPolicy>{};
```
msimberg, line 390: @hkaiser this was the CPO I was talking about that is not in detail. I don't think it needs to be in a non-detail namespace.

hkaiser: This has been moved now. Please review.

libs/parallelism/algorithms/include/hpx/parallel/util/prefetching.hpp
```diff
@@ -31,7 +32,9 @@
 #endif
 
 namespace hpx { namespace parallel { namespace util {
-    namespace detail {
+
+    namespace prefetching {
```
msimberg, line 36: Did you have to change this because of name clashes?

msimberg:
Thanks @hkaiser, this looks good to go now! Unless you have more changes planned let's get this in.

libs/parallelism/algorithms/include/hpx/parallel/util/prefetching.hpp
hkaiser: This was changed to simplify supporting the new CPOs. The types are not directly used by the user anyways.

PR #5250, opened 2021-03-20
Try fixing vc circle ci

srinivasyadav18:
## Proposed Changes

  - enable HPX_WITH_DATAPAR=On for circle ci

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -205,6 +205,7 @@ jobs:
                 -G "Ninja" \
                 -DCMAKE_BUILD_TYPE=Debug \
                 -DHPX_WITH_MALLOC=system \
+                -DHPX_WITH_DATAPAR=On \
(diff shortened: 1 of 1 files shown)


hkaiser:
Currently compilation using Vc fails horribly on CircleCI (https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx?branch=fixing_datapar). For this patch: do we really need to enable this CMake option? I thought that it is enabled by default if we have `-DHPX_WITH_DATAPAR_VC=On` (see https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_SetupVc.cmake#L83).

srinivasyadav18:
Yes, actually, but the build still did not use vc, so i thought of enabling. It works as expected locally.

Closing this as I just was testing whether vc builds were working or not!

PR #5251, opened 2021-03-20
Banish <iostream> from library headers

K-ballo:
This header adds the global std::cin/cout/cerr and wide counterparts to the users' translation unit.


hkaiser (dismissed):
LGTM, thanks!

K-ballo:
Some leakage still remains, notably in the coroutine context implementations, and the debugging print functionality. With some redesign it should eventually be possible to fix those as well.

msimberg (approved):
@K-ballo FYI, this looks good, except for the clang-format step failing. Thanks for doing this!

PR #5253, opened 2021-03-20
Add utility to format ranges

K-ballo:


msimberg (dismissed):
Looks good to me. Did you already have a specific use-case in mind (and if yes, I guess that will be changed separately)?

K-ballo:
>  Did you already have a specific use-case in mind

I had some when I first wrote this, but I'm not sure those are still present today. Another PR which transitions more stuff from iostreams to format is in the works (branch `more-format`)

hkaiser (dismissed):
LGTM, thanks!

libs/core/format/include/hpx/modules/format.hpp
```diff
@@ -276,4 +276,36 @@ namespace hpx { namespace util {
...
+                    using value_type = std::decay_t<decltype(elem)>;
+                    detail::formatter<value_type>::call(os, spec, &elem);
+                }
+            }
+        };
+    }    // namespace detail
+
+    template <typename Range>
+    detail::format_join<Range> format_join(
+        Range const& range, boost::string_ref delimiter = "") noexcept
```
msimberg, line 307: Absolutely not a blocker, but would `","` be a more appropriate default delimiter? Or not having a default at all?

K-ballo: > would `","` be a more appropriate default delimiter?
no

> Or not having a default at all?
sure, go ahead and remove the default

PR #5255, opened 2021-03-22
Future/sender adapters

msimberg:
Makes futures senders by adding the required member typedefs and `connect`. Also adds a `hpx::execution::experimental::make_future` function for turning senders into futures. ~Especially the latter could likely be optimized as it's currently using an explicit promise/future pair for the implementation.~ Updated to a single heap allocation with a custom allocator.

I think these two additions will take care of most of the interoperability concerns with futures and senders, and doesn't require us to change the internals of futures (much... only what's in this PR).

### DIFF
--- a/libs/parallelism/execution/include/hpx/execution/sender_future.hpp
+++ b/libs/parallelism/execution/include/hpx/execution/sender_future.hpp
@@ -1,0 +1,139 @@
+//  Copyright (c) 2021 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/allocator_support/allocator_deleter.hpp>
+#include <hpx/allocator_support/internal_allocator.hpp>
+#include <hpx/execution/algorithms/detail/single_result.hpp>
+#include <hpx/execution_base/operation_state.hpp>
+#include <hpx/execution_base/sender.hpp>
+#include <hpx/functional/invoke_result.hpp>
+#include <hpx/futures/detail/future_data.hpp>
+#include <hpx/futures/promise.hpp>
+#include <hpx/modules/memory.hpp>
+#include <hpx/type_support/pack.hpp>
+#include <hpx/type_support/unused.hpp>
+
+#include <exception>
+#include <memory>
+#include <utility>
+
+namespace hpx { namespace execution { namespace experimental {
+    namespace detail {
+        template <typename T, typename Allocator>
+        struct make_future_receiver
+        {
+            hpx::intrusive_ptr<
+                hpx::lcos::detail::future_data_allocator<T, Allocator>>
+                data;
+
+            void set_error(std::exception_ptr e) noexcept
+            {
+                data->set_exception(e);
+                data.reset();
+            }
+
+            void set_done() noexcept
+            {
+                std::terminate();
+            }
+
+            template <typename U>
+            void set_value(U&& u) noexcept
+            {
+                data->set_value(std::forward<U>(u));
+                data.reset();
+            }
+        };
+
+        template <typename Allocator>
+        struct make_future_receiver<void, Allocator>
+        {
+            hpx::intrusive_ptr<
+                hpx::lcos::detail::future_data_allocator<void, Allocator>>
+                data;
+
+            void set_error(std::exception_ptr e) noexcept
+            {
+                data->set_exception(e);
+                data.reset();
+            }
+
+            void set_done() noexcept
+            {
+                std::terminate();
+            }
+
+            void set_value() noexcept
+            {
+                data->set_value(hpx::util::unused);
+                data.reset();
+            }
+        };
+
+        template <typename T, typename Allocator, typename OS>
+        struct future_data
+          : hpx::lcos::detail::future_data_allocator<T, Allocator>
+        {
+            HPX_NON_COPYABLE(future_data);
+
+            using operation_state_type = std::decay_t<OS>;
+            using init_no_addref =
+                typename hpx::lcos::detail::future_data_allocator<T,
+                    Allocator>::init_no_addref;
+            using other_allocator = typename std::allocator_traits<
+                Allocator>::template rebind_alloc<future_data>;
+
+            operation_state_type os;
+
+            template <typename S>
+            future_data(
+                init_no_addref no_addref, other_allocator const& alloc, S&& s)
+              : hpx::lcos::detail::future_data_allocator<T, Allocator>(
+                    no_addref, alloc)
+              , os(hpx::execution::experimental::connect(std::forward<S>(s),
+                    detail::make_future_receiver<T, Allocator>{this}))
+            {
+                hpx::execution::experimental::start(os);
+            }
+        };
+    }    // namespace detail
+
+    template <typename S, typename Allocator = hpx::util::internal_allocator<>,
+        typename = std::enable_if_t<is_sender_v<S>>>
+    auto make_future(S&& s, Allocator const& a = Allocator{})
+    {
+        using allocator_type = Allocator;
+
+        using value_types =
+            typename hpx::execution::experimental::sender_traits<std::decay_t<
+                S>>::template value_types<hpx::util::pack, hpx::util::pack>;
+        using result_type = std::decay_t<detail::single_result_t<value_types>>;
+        using operation_state_type = typename hpx::util::invoke_result<
+            hpx::execution::experimental::connect_t, S,
+            detail::make_future_receiver<result_type, allocator_type>>::type;
+
+        using shared_state = detail::future_data<result_type, allocator_type,
+            operation_state_type>;
+        using init_no_addref = typename shared_state::init_no_addref;
+        using other_allocator = typename std::allocator_traits<
+            allocator_type>::template rebind_alloc<shared_state>;
+        using allocator_traits = std::allocator_traits<other_allocator>;
+        using unique_ptr = std::unique_ptr<shared_state,
+            util::allocator_deleter<other_allocator>>;
+
+        other_allocator alloc(a);
+        unique_ptr p(allocator_traits::allocate(alloc, 1),
+            hpx::util::allocator_deleter<other_allocator>{alloc});
+
+        allocator_traits::construct(
+            alloc, p.get(), init_no_addref{}, alloc, std::forward<S>(s));
+
+        return hpx::traits::future_access<future<result_type>>::create(
+            p.release(), false);
+    }
+}}}    // namespace hpx::execution::experimental
--- a/libs/parallelism/futures/include/hpx/futures/future.hpp
+++ b/libs/parallelism/futures/include/hpx/futures/future.hpp
@@ -99,9 +99,8 @@ namespace hpx { namespace lcos { namespace detail {
             serialize_future_load(
                 ar, f, std::is_default_constructible<value_type>());
         }
-        else
-            // NOLINTNEXTLINE(bugprone-branch-clone)
-            if (state == future_state::has_exception)
+        // NOLINTNEXTLINE(bugprone-branch-clone)
+        else if (state == future_state::has_exception)
         {
             std::exception_ptr exception;
             ar >> exception;
@@ -112,9 +111,8 @@ namespace hpx { namespace lcos { namespace detail {
 
             f = hpx::traits::future_access<Future>::create(std::move(p));
         }
-        else
-            // NOLINTNEXTLINE(bugprone-branch-clone)
-            if (state == future_state::invalid)
+        // NOLINTNEXTLINE(bugprone-branch-clone)
+        else if (state == future_state::invalid)
         {
             f = Future();
         }
@@ -488,10 +486,9 @@ namespace hpx { namespace lcos { namespace detail {
     class future_base
     {
     public:
-        typedef R result_type;
-        typedef future_data_base<
-            typename traits::detail::shared_state_ptr_result<R>::type>
-            shared_state_type;
+        using result_type = R;
+        using shared_state_type = future_data_base<
+            typename traits::detail::shared_state_ptr_result<R>::type>;
 
     private:
         template <typename F>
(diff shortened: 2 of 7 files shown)


libs/parallelism/execution/include/hpx/execution/sender_future.hpp
```diff
@@ -0,0 +1,61 @@
...
+            hpx::lcos::local::promise<R> p;
+
+            void set_error(std::exception_ptr e) noexcept
+            {
+                p.set_exception(e);
+            }
+
+            void set_done() noexcept
+            {
+                std::terminate();
```
hkaiser, line 19: Could this be called for `future<void>` under certain circumstances?

hkaiser:
> Makes futures senders by adding the required member typedefs and `connect`. Also adds a `hpx::execution::experimental::make_future` function for turning senders into futures. Especially the latter could likely be optimized as it's currently using an explicit promise/future pair for the implementation.

Yes, a special future-state would do the trick. This would also avoid having to allocate a separate operation_state just in case as the returned future would keep it alive naturally.

> I think these two additions will take care of most of the interoperability concerns with futures and senders, and doesn't require us to change the internals of futures (much... only what's in this PR).

This is great! I like it.

libs/parallelism/execution/include/hpx/execution/sender_future.hpp
msimberg: The void-ness of the sender doesn't affect whether or not this can be called. This would be called if the sender is cancelled. We could set a special exception in that case (the terminate is mostly a placeholder right now) but it's not clear that's reasonable behaviour either.

In general I've done nothing so far to support cancellation for senders though so I might prefer to keep the terminate at least for the moment.

hkaiser: Ok, makes sense.

msimberg:
> This PR doesn't really allows to use a future as a 'sender', it's rather a 'receiver' facility. But this might be just in the eye of the beholder and I might misunderstand the terminology.
> 
> Along these lines, and not necessarily for this PR, I also would like to see:
> 
>     * `hpx::execution::experimental::make_future_alloc()`, same as proposed but allowing to specify the allocator to use for the shared state of the returned future.

Working on this. This comes more or less automatically with a custom shared state.

>     * a facility that allows to use futures as a 'sender', i.e. having a `connect` that attaches a continuation to an existing future to feed it's value into any other receiving facility

I may also be misunderstanding things, but that is *exactly* what this is meant to be. See: https://github.com/STEllAR-GROUP/hpx/blob/229ee42867c1cc834442d5916606b5a1c81418bb/libs/parallelism/executors/tests/unit/p0443_executor.cpp#L650-L651. `future` now fulfills the concept of sender, which includes having a `connect` operation which allows connecting a receiver ("the receiving facility").

Apologies that this contains unrelated commits, the point was maybe lost among the other changes.

hkaiser:
BTW: Is `shared_future` a sender/receiver now as well?

msimberg:
> BTW: Is `shared_future` a sender/receiver now as well?

A sender, yes.

I just pushed a change to use a custom shared state. I'd appreciate you having a look to see if it looks somewhat correct to you (especially the reference counting). I'll still be expanding the tests, but what's already there passes now.

hkaiser:
> > BTW: Is `shared_future` a sender/receiver now as well?
> 
> A sender, yes.

Promises would be the proper receivers, actually.

> I just pushed a change to use a custom shared state. I'd appreciate you having a look to see if it looks somewhat correct to you (especially the reference counting). I'll still be expanding the tests, but what's already there passes now.

Looks good! Thanks!

libs/parallelism/execution/include/hpx/execution/sender_future.hpp
```diff
@@ -0,0 +1,137 @@
...
+        using unique_ptr = std::unique_ptr<shared_state,
+            util::allocator_deleter<other_allocator>>;
+
+        other_allocator alloc(a);
+        unique_ptr p(allocator_traits::allocate(alloc, 1),
+            hpx::util::allocator_deleter<other_allocator>{alloc});
+
+        allocator_traits::construct(alloc, p.get(), init_no_addref{}, alloc,
+            std::forward<S>(s),
+            detail::make_future_receiver<result_type, allocator_type>{p.get()});
```
hkaiser, line 132: Shouldn't the `make_future_receiver` keep the shared state alive? If the returned future goes out of scope before the result has been received, the pointer to the shared state would have possibly become invalid. 

Hmmm. OTOH, this would create a cyclic dependency of the shared state to itself as the `make_future_receiver` is embedded in the new shared state, Can the operation_state break this cycle once it is triggered?

hkaiser: I saw just now that you already break the cycle in the `make_future_receiver` itself - so all what's needed is another addref (possibly pass `true` as the second argument to the `future_access`). The shared state in the returned future should have a reference `count == 2` (one from the future and one from the `make_future_receiver`).

libs/parallelism/execution/include/hpx/execution/sender_future.hpp
```diff
@@ -0,0 +1,137 @@
...
+                hpx::lcos::detail::intrusive_ptr_release(data);
+            }
+
+            void set_done() noexcept
+            {
+                std::terminate();
+            }
+
+            template <typename Ts>
+            void set_value() noexcept
```
hkaiser, line 92: Does this have to be a template, still?

libs/parallelism/execution/include/hpx/execution/sender_future.hpp
```diff
@@ -0,0 +1,137 @@
...
+            {
+                hpx::lcos::detail::intrusive_ptr_add_ref(this);
+                hpx::execution::experimental::start(os);
+            }
+        };
+
+        template <typename T, typename Allocator>
+        struct make_future_receiver
+        {
+            hpx::lcos::detail::future_data_allocator<T, Allocator>* data;
```
hkaiser, line 54: you could use a `hpx::intrusive_ptr<future_data_refcnt_base>` here instead (and a `data.release()` instead of the `intrusive_ptr_release` below. This would make things more robust wrt exception handling. Also, the `make_future_receiver` wouldn't have to be a template in this case.

libs/parallelism/execution/include/hpx/execution/sender_future.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+            template <typename S, typename R>
+            future_data(init_no_addref no_addref, other_allocator const& alloc,
+                S&& s, R&& r)
+              : hpx::lcos::detail::future_data_allocator<T, Allocator>(
+                    no_addref, alloc)
+              , os(hpx::execution::experimental::connect(
+                    std::forward<S>(s), std::forward<R>(r)))
+            {
+                hpx::lcos::detail::intrusive_ptr_add_ref(this);
```
hkaiser, line 46: Sorry, I have seen this just now (after writing all about the reference counting below). To summarize, I think you should replace the pointer in the `make_future_receiver` with an `intrusive_ptr<>`. That would simplify the reference counting by not having to deal with it in several places.

hkaiser:
@msimberg sorry for the flurry of comments on the reference counting where each subsequent comment replaces the previous one. It took me a while to understand what you have.

libs/parallelism/execution/include/hpx/execution/sender_future.hpp
msimberg: Nope (it can't be one), good catch! The void case is still untested, so I'll update this along with adding those tests.

libs/parallelism/execution/include/hpx/execution/sender_future.hpp
msimberg: That sounds like a nice simplification, thanks!

libs/parallelism/execution/include/hpx/execution/sender_future.hpp
msimberg: Yes, the cycle is a bit annoying, but there are similar situations in some of the other senders as well and I think the base class works reasonably well. That's why I went with the promise/future pair initially because it was the easy solution.

I'll check add some sanity checks to the reference counts.

msimberg:
> @msimberg sorry for the flurry of comments on the reference counting where each subsequent comment replaces the previous one. It took me a while to understand what you have.

No worries at all. Thanks for looking through this, this was exactly what I was hoping to get from you! If you have suggestions (besides the once you already gave) on how to make things more understandable that's also welcome.

Tangential, but I realized that given that futures are senders, `make_future` here essentially works as an operation that returns a type-erased, eagerly submitted sender, which we'll need anyway in some form. Whether *the* type-erased sender type should actually be `future` or another, similar, `any_sender` type we'll see later I guess. Multiple `value_types` might be the main benefit that a more generic `any_sender` could bring.

libs/parallelism/execution/include/hpx/execution/sender_future.hpp
msimberg: I pushed some more changes. It can't actually be only an intrusive pointer of `future_data_refcnt_base` since it needs to set the value, but an intrusive pointer of `future_data_allocator` still improves the exception handling.

libs/parallelism/execution/include/hpx/execution/sender_future.hpp
msimberg: I think the reference counts are a bit simpler now. The `init_no_addref` initializes the shared state to have a reference count of one (I think `init_no_addref` is very confusingly named), and then the intrusive pointer in the receiver increases it to two, meaning if I have the correct idea about how things are working the final access does not need to increase the reference count further.

hkaiser: `init_no_addref` is confusing, I agree. Any suggestions for a better name (I'd be happy to change that)?

libs/parallelism/futures/include/hpx/futures/future.hpp
```diff
@@ -782,8 +781,91 @@ namespace hpx { namespace lcos {
...
+        template <template <typename...> class Variant>
+        using error_types = Variant<std::exception_ptr>;
+
+        static constexpr bool sends_done = false;
+
+        template <typename Receiver>
+        struct operation_state
+        {
+            std::decay_t<Receiver> receiver;
+            std::decay_t<future> fut;
```
hkaiser, line 803: Why decay_t<future>?

libs/parallelism/futures/include/hpx/futures/future.hpp
```diff
@@ -782,8 +781,91 @@ namespace hpx { namespace lcos {
...
+
+            operation_state(operation_state&&) = delete;
+            operation_state(operation_state const&) = delete;
+
+            void start_helper(std::true_type) noexcept
+            {
+                try
+                {
+                    fut.then([receiver = std::move(receiver)](
```
hkaiser, line 812: Would using `shared->set_on_completed()` be sufficient (it will avoid allocating a new future that is discarded anyways)?

libs/parallelism/futures/include/hpx/futures/future.hpp
```diff
@@ -782,8 +781,91 @@ namespace hpx { namespace lcos {
...
+            }
+
+            void start() noexcept
+            {
+                start_helper(std::is_void<result_type>{});
+            }
+        };
+
+        template <typename Receiver>
+        operation_state<Receiver> connect(Receiver&& receiver) &&
```
hkaiser, line 865: Does `connect` have to be a member function? Making it a global function might allow to disentangle the `operations_state` defined here from the future itself (we might have to make it a friend, though).

libs/parallelism/execution/include/hpx/execution/sender_future.hpp
hkaiser: This looks good now, thanks for your work on this!

hkaiser:
> Tangential, but I realized that given that futures are senders, `make_future` here essentially works as an operation that returns a type-erased, eagerly submitted sender, which we'll need anyway in some form. Whether _the_ type-erased sender type should actually be `future` or another, similar, `any_sender` type we'll see later I guess. Multiple `value_types` might be the main benefit that a more generic `any_sender` could bring.

Sounds like a nice idea. FWIW, we can always use `future<tuple<Ts...>>` to represent multiple results, but that is probably overkill.

libs/parallelism/execution/include/hpx/execution/sender_future.hpp
msimberg: I don't know if `init_addref` is any better? Basically `init_no_addref` tells me exactly the opposite of what it does, but that's because I didn't know first what it's referring to.

libs/parallelism/futures/include/hpx/futures/future.hpp
msimberg: No reason, will change.

libs/parallelism/futures/include/hpx/futures/future.hpp
msimberg: Yeah, that'd be a lot better. Will also change that.

libs/parallelism/futures/include/hpx/futures/future.hpp
msimberg: No, it can be a tag_invoke overload. The only part that has to stay are the typedefs (`value_types` and `error_types`), but the rest can be outside the class.

msimberg:
> > Tangential, but I realized that given that futures are senders, `make_future` here essentially works as an operation that returns a type-erased, eagerly submitted sender, which we'll need anyway in some form. Whether _the_ type-erased sender type should actually be `future` or another, similar, `any_sender` type we'll see later I guess. Multiple `value_types` might be the main benefit that a more generic `any_sender` could bring.
> 
> Sounds like a nice idea. FWIW, we can always use `future<tuple<Ts...>>` to represent multiple results, but that is probably overkill.

True, or in fact `future<variant<tuple<...>,...>`. Let's add it if and when it becomes relevant. Right now I just require that the variant and tuple hold a single member.

libs/parallelism/execution/include/hpx/execution/sender_future.hpp
hkaiser: It meant to refer to the fact that the new object does _not_ need any addref when assigned to the `intrusive_ptr` eventually (which doesn't make too much sense now that I speak out loudly about it). I will change it to `init_addref`, thanks!

libs/parallelism/futures/include/hpx/futures/future.hpp
hkaiser: The only question I have is to on what execution context the continuation should execute. Do we have to do special things to enforce any possible constraints here?

libs/parallelism/execution/include/hpx/execution/sender_future.hpp
msimberg: Yep, in your defence there is a comment that explains exactly that, so I don't mind either way. If you do end up changing the name please update the comment in that case as well ;)

libs/parallelism/futures/include/hpx/futures/future.hpp
msimberg: I'm glad you asked! That reminds me that I wanted to make the continuations `sync`, i.e. use whatever context the future is completed on (`set_on_completed` should have the same effect if I understand correctly). `execution::on` is there so that one can explicitly change context if needed.

hkaiser: Yes, `set_on_completed` generally executes things `sync`, i.e. in the execution context of the thread that made the future ready.

msimberg: I changed this to use `set_on_completed`. I'd appreciate if you could have a look at it again.

libs/parallelism/futures/include/hpx/futures/future.hpp
msimberg: I moved the `operation_state` type outside the future classes, and there's only one (templated) type for `future` and `shared_future` now (instead of duplicated inside `future` and `shared_future`). I left it in this file though, because the `tag_invoke` overloads require the definition of the tag `connect_t` which would be problematic in terms of dependencies between modules.

libs/parallelism/futures/include/hpx/futures/future.hpp
hkaiser: LGTM now, thanks!

msimberg:
@hkaiser any further comments on this one?

hkaiser (approved):
LGTM, thanks!

PR #5258, opened 2021-03-22
Allow for HPX algorithms being invoked with std execution policies

hkaiser:
- flyby: significantly simplify algorithm dispatch helper

Should this be guarded by a special option (currently this will be enabled as soon as the standard execution policies are available)?

This also adds `hpx::execution::unsequenced_policy`, which currently maps to `hpx::execution::sequenced_policy`.

### DIFF
--- a/cmake/HPX_AddConfigTest.cmake
+++ b/cmake/HPX_AddConfigTest.cmake
@@ -537,6 +537,15 @@ function(hpx_check_for_cxx20_std_endian)
 endfunction()
 
 # ##############################################################################
+function(hpx_check_for_cxx20_std_execution_policies)
+  add_hpx_config_test(
+    HPX_WITH_CXX20_STD_EXECUTION_POLICES
+    SOURCE cmake/tests/cxx20_std_execution_policies.cpp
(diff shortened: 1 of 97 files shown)


msimberg:
> Should this be guarded by a special option (currently this will be enabled as soon as the standard execution policies are available)?

I don't think there's any harm in just enabling it if the policies are available (it shouldn't affect existing code).

Do I understand correctly that this effectively makes `std::execution::par` equivalent to `hpx::execution::par`, i.e. they will both dispatch to the same implementation (and likewise for the other policies)?

hkaiser:
> > Should this be guarded by a special option (currently this will be enabled as soon as the standard execution policies are available)?
> 
> I don't think there's any harm in just enabling it if the policies are available (it shouldn't affect existing code).

Ok, I agree.

> Do I understand correctly that this effectively makes `std::execution::par` equivalent to `hpx::execution::par`, i.e. they will both dispatch to the same implementation (and likewise for the other policies)?

Yes, that's exactly what this does.

msimberg (approved):
Looks good!

Issue #5257, opened 2021-03-22
Allow triggering the output of OS thread affinity from configuration settings


severinstrobl:
When initializing HPX solely via `hpx::init_params.cfg` it seems to not be possible to trigger certain functionality such as printing the OS thread affinity information. I found no way to pass the equivalent of the command line switch `--hpx:print-bind` using the configuration object, am I missing something here?

## Specifications

  - HPX Version: 1.6.0
  - Platform (compiler, OS): GCC 10.2.0, Linux

hkaiser:
Wouldn't this work:
```
int hpx_main(int argc, char* argv[])
{
    return hpx::finalize();
}

int main(int argc, char* argv[])
{
    std::vector<std::string> cfg = {"--hpx:print-bind" };
    hpx::init_params params;
    params.cfg = std::move(cfg);
    return hpx::init(argc, argv, params);
}
```

severinstrobl:
No, that seems to be just ignored, unless I'm doing something silly here. I think the problem might be that only the actual command line is taken into account [here](https://github.com/STEllAR-GROUP/hpx/blob/76a2ba5b266008be5e0d5a13e17ae0950099dde2/libs/full/runtime_distributed/src/server/runtime_support_server.cpp#L1049)?

hkaiser:
> No, that seems to be just ignored, unless I'm doing something silly here. I think the problem might be that only the actual command line is taken into account [here](https://github.com/STEllAR-GROUP/hpx/blob/76a2ba5b266008be5e0d5a13e17ae0950099dde2/libs/full/runtime_distributed/src/server/runtime_support_server.cpp#L1049)?

Indeed, excellent catch!

hkaiser:
@severinstrobl please have a look at #5259 to verify it fixes your problem

severinstrobl:
@hkaiser That does it, thanks for looking into this so quickly.

One more related question: How can one determine the correct way to pass the various options as configuration options? As an example: Many examples use `hpx.run_hpx_main!=1`, which seems to correspond to passing `--hpx:run-hpx-main` via the command line, but I cannot find that in the [docs](https://hpx-docs.stellar-group.org/latest/html/manual/launching_and_configuring_hpx_applications.html#the-hpx-ini-file-format). And for `--hpx:print-bind` and friends there does not seem to be a corresponding mapping.

hkaiser:
> @hkaiser That does it, thanks for looking into this so quickly.
> 
[...]

PR #5259, opened 2021-03-23
Making sure late command line options are recognized as configuration options

hkaiser:
- flyby: move tests into modules

Fixes: #5257


libs/full/command_line_handling/src/command_line_handling.cpp
```diff
@@ -1464,6 +1481,19 @@ namespace hpx { namespace util {
...
         ini_config_.erase(it, ini_config_.end());
 
+        // store the command line options that came from the configuration
+        // settings in the registry
+        if (!options.empty())
+        {
+            std::string config_options;
+            for (auto const& option : options)
+            {
+                config_options = " " + option;
```
severinstrobl, line 1491:
```suggestion
                config_options += " " + option;
```

libs/full/runtime_distributed/src/server/runtime_support_server.cpp
```diff
@@ -1046,7 +1046,12 @@ namespace hpx { namespace components { namespace server {
 
             // secondary command line handling, looking for --exit and other
             // options
-            std::string cmd_line(ini.get_entry("hpx.cmd_line", ""));
+            std::string cmd_line =
+                ini.get_entry("hpx.commandline.command", "") +
```
severinstrobl, line 1050: I think there should be an extra space at least after the command because the options don't all contain the leading space.

hkaiser:
Merging this now as all tests pass.

PR #5261, opened 2021-03-24
Limit Rostam Jenkins jobs to marvin partition temporarily

msimberg:
Hoping to get some insight into why some slurm jobs never start...


msimberg:
Well, the results already show that the partition doesn't make a difference. There are two failed jobs on marvin now, and I've seen it happen on medusa before. Closing without merging.

PR #5266, opened 2021-03-28
Making late command line options available for local runtime

hkaiser:
@msimberg please verify that this works as advertized for the local runtime.

### DIFF
--- a/libs/full/runtime_local/src/runtime_local.cpp
+++ b/libs/full/runtime_local/src/runtime_local.cpp
@@ -7,6 +7,7 @@
 
 #include <hpx/config.hpp>
 #include <hpx/assert.hpp>
+#include <hpx/command_line_handling/late_command_line_handling.hpp>
 #include <hpx/command_line_handling/parse_command_line.hpp>
 #include <hpx/coroutines/coroutine.hpp>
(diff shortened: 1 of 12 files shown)


msimberg:
> @msimberg please verify that this works as advertized for the local runtime.

@hkaiser thanks for looking into this! Only the finalization and return code were missing. I've added them and the few local tests seem to work just fine. Let's wait for CI to cycle but this should be good otherwise.

libs/full/runtime_local/src/runtime_local.cpp
```diff
@@ -1240,13 +1242,88 @@ namespace hpx {
...
+            {
+                hpx::program_options::options_description options;
+                if (util::handle_late_commandline_options(
+                        get_config(), options, &detail::handle_print_bind))
+                {
+                    lbt_ << "runtime_local::run_helper: bootstrap "
+                            "aborted, bailing out";
+                    set_state(state_running);
+                    finalize(-1.0);
+                    result = 1;
```
hkaiser, line 1320: I think this should be set to the value that was returned from `util::handle_late_commandline_options` above. I'll make that change.

PR #5267, opened 2021-03-29
Support annotations in hpx::transform_reduce

severinstrobl:
This is sort of a follow-up to #5232. Similar to the other parallel algorithms, `hpx::transform_reduce` should support annotations. I'm not sure this is the best way to handle the annotations, as really there are two functors involved, one for the reduction and one for the transformation/conversion. With the proposed implementation both can be annotated, but the annotation of the reduction takes precedence.

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform_reduce.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform_reduce.hpp
@@ -286,6 +286,68 @@ namespace hpx { namespace parallel { inline namespace v1 {
     // transform_reduce
     namespace detail {
 
+        ///////////////////////////////////////////////////////////////////////
+        template <typename T, typename ExPolicy, typename Reduce,
+            typename Convert>
+        struct transform_reduce_iteration
+        {
+            using execution_policy_type = typename std::decay<ExPolicy>::type;
+            using reduce_type = typename std::decay<Reduce>::type;
+            using convert_type = typename std::decay<Convert>::type;
+
+            reduce_type reduce_;
+            convert_type convert_;
+
+            template <typename Reduce_, typename Convert_>
+            HPX_HOST_DEVICE transform_reduce_iteration(
+                Reduce_&& reduce, Convert_&& convert)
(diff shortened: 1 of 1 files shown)


hkaiser (approved):
LGTM, thanks!

msimberg:
retest

libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform_reduce.hpp
```diff
@@ -326,25 +388,20 @@ namespace hpx { namespace parallel { inline namespace v1 {
                 using reference =
                     typename std::iterator_traits<Iter>::reference;
```
msimberg, line 389:
```suggestion
```
Would you mind removing this typedef? It's now unused (see https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/7241/workflows/7d3d3531-1e85-4225-bc4e-b46c0ef3bd4f/jobs/245396/steps).

msimberg (changes requested):
Thanks @severinstrobl! Looks good to me. There's just a now unused typedef that should be removed.

I think taking the annotation the way you've done it now is nice. For the future we might want to consider setting a label on the executor/policy directly as an alternative. Something like:
```
hpx::transform_reduce(hpx::prefer(hpx::execution::par, hpx::label("my transform_reduce"), ...);
```

libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform_reduce.hpp
severinstrobl: Sure, there's actually an unused one in the sequential version as well.

hkaiser:
> For the future we might want to consider setting a label on the executor/policy directly as an alternative. Something like:
> 
> ```
> hpx::transform_reduce(hpx::prefer(hpx::execution::par, hpx::label("my transform_reduce")), ...);
> ```

I like that idea!

msimberg:
Just for reference, I opened https://github.com/STEllAR-GROUP/hpx/issues/5269 for tracking the possibility of adding labels to policies/executors.

msimberg:
retest

PR #5272, opened 2021-04-01
Fix bogus deprecation warnings with GCC 10

msimberg:


gonidelis:
@msimberg  Thanks for that! Is that a gcc10 bug or is it our fault?

hkaiser (approved):
LGTM, thanks!

msimberg:
> @msimberg Thanks for that! Is that a gcc10 bug or is it our fault?

GCC bug, see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99318.

PR #5274, opened 2021-04-01
More P1897 algorithms

msimberg:
I've added the following missing algorithms from P1897:

* `ensure_started`: The returned sender acts similarly to a `shared_future`, i.e. can be connected and started multiple times, with the predecessor only being started once. I'm not 100% sure this is the intended use case, but the wording in P1897 is unclear enough (to me) that this could be intended.
* `let_error`: Adds an error handler to be used in cases of error. The handler returns a new sender and the error from the predecessor sender is kept alive for the duration of the new sender. Done and value signals are propagated normally without going through the handler.
* `let_value`: Adds a value handler. The handler returns a new sender and the value from the predecessor sender is kept alive for the duration of the new sender. Done and error signals are propagate normally withouth going through the handler.

I've added the following which is not in P1897, but which I think is a useful/necessary addition:
* `detach`: Submits a sender for execution and returns after calling start without waiting for completion (unlike `sync_wait`, which blocks until completion). Terminates on error signals, ignores done and value signals.

This also makes various smaller modifications to the existing algorithms.

Contains #5255.

This is still a draft but looking for early feedback on the implementations. I definitely still have some rearranging/renaming to do with the metaprogramming helpers, and most likely with other parts as well. This uses `std::variant` in more places, which will need some more guards until we require C++17 support.

### DIFF
--- a/libs/core/datastructures/include/hpx/datastructures/tuple.hpp
+++ b/libs/core/datastructures/include/hpx/datastructures/tuple.hpp
@@ -530,7 +530,7 @@ namespace hpx {
     };
 
     template <std::size_t I, typename... Ts>
-    struct tuple_element<I, tuple<Ts...>, void>
+    struct tuple_element<I, tuple<Ts...>>
     {
         using type = typename util::at_index<I, Ts...>::type;
 
@@ -548,7 +548,7 @@ namespace hpx {
     };
 
     template <typename T0, typename T1>
-    struct tuple_element<0, std::pair<T0, T1>, void>
+    struct tuple_element<0, std::pair<T0, T1>>
     {
         using type = T0;
 
@@ -566,7 +566,7 @@ namespace hpx {
     };
 
     template <typename T0, typename T1>
-    struct tuple_element<1, std::pair<T0, T1>, void>
+    struct tuple_element<1, std::pair<T0, T1>>
     {
         using type = T1;
 
@@ -584,7 +584,7 @@ namespace hpx {
     };
 
     template <std::size_t I, typename Type, std::size_t Size>
-    struct tuple_element<I, std::array<Type, Size>, void>
+    struct tuple_element<I, std::array<Type, Size>>
     {
         using type = Type;
 
@@ -610,7 +610,7 @@ namespace hpx {
 
 #if defined(HPX_DATASTRUCTURES_HAVE_ADAPT_STD_TUPLE)
     template <std::size_t I, typename... Ts>
-    struct tuple_element<I, std::tuple<Ts...>, void>
+    struct tuple_element<I, std::tuple<Ts...>>
     {
         using type = typename std::tuple_element<I, std::tuple<Ts...>>::type;
 
--- a/libs/core/datastructures/include/hpx/datastructures/tuple.hpp
+++ b/libs/core/datastructures/include/hpx/datastructures/tuple.hpp
@@ -37,7 +37,7 @@ namespace hpx {
     template <typename T>
     struct tuple_size;    // undefined
 
-    template <std::size_t I, typename T>
+    template <std::size_t I, typename T, typename Enable = void>
     struct tuple_element;    // undefined
 
     // Hide implementations of get<> inside an internal namespace to be able to
@@ -503,31 +503,34 @@ namespace hpx {
 
     // template <size_t I, class Tuple>
     // class tuple_element
-    template <std::size_t I, typename T>
+    template <std::size_t I, typename T, typename Enable>
     struct tuple_element
     {
     };
 
     template <std::size_t I, typename T>
-    struct tuple_element<I, const T>
+    struct tuple_element<I, const T,
+        typename util::always_void<typename tuple_element<I, T>::type>::type>
       : std::add_const<typename tuple_element<I, T>::type>
     {
     };
 
     template <std::size_t I, typename T>
-    struct tuple_element<I, volatile T>
+    struct tuple_element<I, volatile T,
+        typename util::always_void<typename tuple_element<I, T>::type>::type>
       : std::add_volatile<typename tuple_element<I, T>::type>
     {
     };
 
     template <std::size_t I, typename T>
-    struct tuple_element<I, const volatile T>
+    struct tuple_element<I, const volatile T,
+        typename util::always_void<typename tuple_element<I, T>::type>::type>
       : std::add_cv<typename tuple_element<I, T>::type>
     {
     };
 
     template <std::size_t I, typename... Ts>
-    struct tuple_element<I, tuple<Ts...>>
+    struct tuple_element<I, tuple<Ts...>, void>
     {
         using type = typename util::at_index<I, Ts...>::type;
 
@@ -545,7 +548,7 @@ namespace hpx {
     };
 
     template <typename T0, typename T1>
-    struct tuple_element<0, std::pair<T0, T1>>
+    struct tuple_element<0, std::pair<T0, T1>, void>
     {
         using type = T0;
 
@@ -563,7 +566,7 @@ namespace hpx {
     };
 
     template <typename T0, typename T1>
-    struct tuple_element<1, std::pair<T0, T1>>
+    struct tuple_element<1, std::pair<T0, T1>, void>
     {
         using type = T1;
 
@@ -581,7 +584,7 @@ namespace hpx {
     };
 
     template <std::size_t I, typename Type, std::size_t Size>
-    struct tuple_element<I, std::array<Type, Size>>
+    struct tuple_element<I, std::array<Type, Size>, void>
     {
         using type = Type;
 
@@ -607,7 +610,7 @@ namespace hpx {
 
 #if defined(HPX_DATASTRUCTURES_HAVE_ADAPT_STD_TUPLE)
     template <std::size_t I, typename... Ts>
-    struct tuple_element<I, std::tuple<Ts...>>
+    struct tuple_element<I, std::tuple<Ts...>, void>
     {
         using type = typename std::tuple_element<I, std::tuple<Ts...>>::type;
 
--- a/libs/core/type_support/include/hpx/type_support/detail/with_result_of.hpp
+++ b/libs/core/type_support/include/hpx/type_support/detail/with_result_of.hpp
@@ -1,0 +1,45 @@
+//  Copyright (c) 2021 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <utility>
+
+namespace hpx { namespace util { namespace detail {
+    // Based on
+    // https://quuxplusone.github.io/blog/2018/05/17/super-elider-round-2/ and
+    // https://akrzemi1.wordpress.com/2018/05/16/rvalues-redefined/.
+    //
+    // Useful for emplacing non-copyable and -movable types into e.g. variants.
+    // The wrapper delays the construction of the non-copyable and -movable type
+    // until it is required for conversion, and guaranteed copy elision in C++17
+    // ensures that a copy constructor is not called when returning from the
+    // conversion operator.
+    template <typename F>
+    class with_result_of_t
+    {
+        F&& f;
+
+    public:
+        using type = decltype(std::declval<F&&>()());
+
+        explicit with_result_of_t(F&& f)
+          : f(std::forward<F>(f))
+        {
+        }
+
+        operator type()
+        {
+            return f();
+        }
+    };
+
+    template <typename F>
+    inline with_result_of_t<F> with_result_of(F&& f)
+    {
+        return with_result_of_t<F>(std::forward<F>(f));
+    }
+}}}    // namespace hpx::util::detail
--- a/libs/core/type_support/include/hpx/type_support/detail/with_result_of.hpp
+++ b/libs/core/type_support/include/hpx/type_support/detail/with_result_of.hpp
@@ -33,7 +33,7 @@ namespace hpx { namespace util { namespace detail {
 
         operator type()
         {
-            return f();
+            return std::forward<F>(f)();
         }
     };
 
--- a/libs/full/async_cuda/include/hpx/async_cuda/cublas_executor.hpp
+++ b/libs/full/async_cuda/include/hpx/async_cuda/cublas_executor.hpp
@@ -13,6 +13,7 @@
 #include <hpx/async_cuda/cuda_executor.hpp>
 #include <hpx/async_cuda/cuda_future.hpp>
 #include <hpx/async_cuda/target.hpp>
+#include <hpx/execution_base/detail/try_catch_exception_ptr.hpp>
 #include <hpx/execution_base/execution.hpp>
 #include <hpx/execution_base/traits/is_executor.hpp>
 #include <hpx/futures/future.hpp>
@@ -223,32 +224,25 @@ namespace hpx { namespace cuda { namespace experimental {
             std::is_same<cublasStatus_t, R>::value, void>::type>
         async(R (*cublas_function)(Params...), Args&&... args)
         {
-            hpx::future<void> result;
-            std::exception_ptr p;
-            try
-            {
-                // make sue we run on the correct device
-                check_cuda_error(cudaSetDevice(device_));
-                // make sure this operation takes place on our stream
-                check_cublas_error(cublasSetStream(handle_.get(), stream_));
-                // insert the cublas handle in the arg list and call the cublas function
-                detail::dispatch_helper<R, Params...> helper;
-                helper(cublas_function, handle_.get(),
-                    std::forward<Args>(args)...);
-                return get_future();
-            }
-            catch (const hpx::exception& e)
-            {
-                p = std::current_exception();
-            }
-
-            // The exception is set outside the catch block since
-            // set_exception may yield. Ending the catch block on a
-            // different worker thread than where it was started may lead
-            // to segfaults.
-            auto state = traits::detail::get_shared_state(result);
-            state->set_exception(std::move(p));
-            return result;
+            return hpx::detail::try_catch_exception_ptr(
+                [&]() {
+                    // make sure we run on the correct device
+                    check_cuda_error(cudaSetDevice(device_));
+                    // make sure this operation takes place on our stream
+                    check_cublas_error(cublasSetStream(handle_.get(), stream_));
+                    // insert the cublas handle in the arg list and call the
+                    // cublas function
+                    detail::dispatch_helper<R, Params...> helper;
+                    helper(cublas_function, handle_.get(),
+                        std::forward<Args>(args)...);
+                    return get_future();
+                },
+                [&](std::exception_ptr ep) {
+                    hpx::future<void> result;
+                    auto state = traits::detail::get_shared_state(result);
+                    state->set_exception(std::move(ep));
+                    return result;
+                });
         }
 
         // -------------------------------------------------------------------------
--- a/libs/full/async_cuda/include/hpx/async_cuda/cublas_executor.hpp
+++ b/libs/full/async_cuda/include/hpx/async_cuda/cublas_executor.hpp
@@ -237,11 +237,8 @@ namespace hpx { namespace cuda { namespace experimental {
                         std::forward<Args>(args)...);
                     return get_future();
                 },
-                [&](std::exception_ptr ep) {
-                    hpx::future<void> result;
-                    auto state = traits::detail::get_shared_state(result);
-                    state->set_exception(std::move(ep));
-                    return result;
+                [&](std::exception_ptr&& ep) {
+                    return hpx::make_exceptional_future<void>(std::move(ep));
                 });
         }
 
--- a/libs/full/async_cuda/include/hpx/async_cuda/cuda_executor.hpp
+++ b/libs/full/async_cuda/include/hpx/async_cuda/cuda_executor.hpp
@@ -12,6 +12,7 @@
 #include <hpx/async_cuda/cuda_future.hpp>
 #include <hpx/async_cuda/target.hpp>
 #include <hpx/errors/exception.hpp>
+#include <hpx/execution_base/detail/try_catch_exception_ptr.hpp>
 #include <hpx/execution_base/execution.hpp>
 #include <hpx/execution_base/traits/is_executor.hpp>
 #include <hpx/futures/future.hpp>
@@ -163,29 +164,21 @@ namespace hpx { namespace cuda { namespace experimental {
         template <typename R, typename... Params, typename... Args>
         hpx::future<void> async(R (*cuda_kernel)(Params...), Args&&... args)
         {
-            hpx::future<void> result;
-            std::exception_ptr p;
-            try
-            {
-                // make sure we run on the correct device
-                check_cuda_error(cudaSetDevice(device_));
-                // insert the stream handle in the arg list and call the cuda function
-                detail::dispatch_helper<R, Params...> helper{};
-                helper(cuda_kernel, std::forward<Args>(args)..., stream_);
-                return get_future();
-            }
-            catch (const hpx::exception& e)
-            {
-                p = std::current_exception();
-            }
-
-            // The exception is set outside the catch block since
-            // set_exception may yield. Ending the catch block on a
-            // different worker thread than where it was started may lead
-            // to segfaults.
-            auto state = traits::detail::get_shared_state(result);
-            state->set_exception(std::move(p));
-            return result;
+            return hpx::detail::try_catch_exception_ptr(
+                [&]() {
+                    // make sure we run on the correct device
+                    check_cuda_error(cudaSetDevice(device_));
+                    // insert the stream handle in the arg list and call the cuda function
+                    detail::dispatch_helper<R, Params...> helper{};
+                    helper(cuda_kernel, std::forward<Args>(args)..., stream_);
+                    return get_future();
+                },
+                [&](std::exception_ptr ep) {
+                    hpx::future<void> result;
+                    auto state = traits::detail::get_shared_state(result);
+                    state->set_exception(std::move(ep));
+                    return result;
+                });
         }
     };
 
--- a/libs/full/async_cuda/include/hpx/async_cuda/cuda_executor.hpp
+++ b/libs/full/async_cuda/include/hpx/async_cuda/cuda_executor.hpp
@@ -173,11 +173,8 @@ namespace hpx { namespace cuda { namespace experimental {
                     helper(cuda_kernel, std::forward<Args>(args)..., stream_);
                     return get_future();
                 },
-                [&](std::exception_ptr ep) {
-                    hpx::future<void> result;
-                    auto state = traits::detail::get_shared_state(result);
-                    state->set_exception(std::move(ep));
-                    return result;
+                [&](std::exception_ptr&& ep) {
+                    return hpx::make_exceptional_future<void>(std::move(ep));
                 });
         }
     };
--- a/libs/parallelism/execution/include/hpx/execution/algorithms/detail/single_result.hpp
+++ b/libs/parallelism/execution/include/hpx/execution/algorithms/detail/single_result.hpp
@@ -6,9 +6,14 @@
 
 #pragma once
 
+#include <hpx/config.hpp>
 #include <hpx/type_support/pack.hpp>
 
 #include <type_traits>
+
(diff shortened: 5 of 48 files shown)


libs/parallelism/execution/include/hpx/execution/algorithms/on.hpp
```diff
@@ -84,12 +84,20 @@ namespace hpx { namespace execution { namespace experimental {
...
                     on_receiver<R, Scheduler>(
                         std::forward<R>(r), std::move(scheduler)));
             }
+
+            template <typename R>
+            auto connect(R&& r) &
+            {
+                return hpx::execution::experimental::connect(s,
+                    on_receiver<R, Scheduler>(
+                        std::forward<R>(r), std::move(scheduler)));
```
msimberg, line 99: Note to self, the scheduler should not be moved. Also check other `connect`s for similar mistakes.

msimberg:
The latest commit (https://github.com/STEllAR-GROUP/hpx/pull/5274/commits/0f2d8434ce6d59e75cfeec5cbfb5bd6fa501bd3e) adds SFINAE to our `tuple_element` so that it doesn't apply to types that it doesn't support. I'd like some feedback on if that's a reasonable thing to do; our adaptations for `std::tuple` are not quite clear to me.

This came up with the clang 11-based Cray compiler in C++17 mode compiling one of the new sender tests. There compilation failed when passing a `const std::variant<hpx::tuple<int>>` to `std::visit`, because it would attempt to use https://github.com/STEllAR-GROUP/hpx/blob/b01c5bc510674ff0464d86c5816a9ced86f73b09/libs/core/datastructures/include/hpx/datastructures/tuple.hpp#L48-L53 and subsequently our `tuple_element` which is not specialized for `std::variant`. What's the motivation for the specializations for `std::tuple`? Should I instead of adding the SFINAE add specializations for `std::variant`? It seems... unnecessarily roundabout.

msimberg:
retest

msimberg:
retest

msimberg:
This also now uses `execution::scheduler` in the default `execution::on` implementation instead of `execution::execute`, thus partly addressing #5294. However, I'm not changing the behaviour of on (yet) as I noticed the wording in P1897 is inconsistent about who it should behave.

msimberg:
@hkaiser, @biddisco, @aurianer any comments on this?

libs/core/datastructures/include/hpx/datastructures/tuple.hpp
```diff
@@ -503,31 +503,34 @@ namespace hpx {
...
-    struct tuple_element<I, const volatile T>
+    struct tuple_element<I, const volatile T,
+        typename util::always_void<typename tuple_element<I, T>::type>::type>
       : std::add_cv<typename tuple_element<I, T>::type>
     {
     };
 
     template <std::size_t I, typename... Ts>
-    struct tuple_element<I, tuple<Ts...>>
+    struct tuple_element<I, tuple<Ts...>, void>
```
hkaiser, line 533: Do we need the explicit `, void>` ?

libs/core/type_support/include/hpx/type_support/detail/with_result_of.hpp
```diff
@@ -0,0 +1,45 @@
...
+        using type = decltype(std::declval<F&&>()());
+
+        explicit with_result_of_t(F&& f)
+          : f(std::forward<F>(f))
+        {
+        }
+
+        operator type()
+        {
+            return f();
```
hkaiser, line 36: Should this be `return std::forward<F>(f)();` ?

libs/full/async_cuda/include/hpx/async_cuda/cublas_executor.hpp
```diff
@@ -222,32 +223,25 @@ namespace hpx { namespace cuda { namespace experimental {
...
+                    // make sure this operation takes place on our stream
+                    check_cublas_error(cublasSetStream(handle_.get(), stream_));
+                    // insert the cublas handle in the arg list and call the
+                    // cublas function
+                    detail::dispatch_helper<R, Params...> helper;
+                    helper(cublas_function, handle_.get(),
+                        std::forward<Args>(args)...);
+                    return get_future();
+                },
+                [&](std::exception_ptr ep) {
```
hkaiser, line 239:
```suggestion
                [&](std::exception_ptr&& ep) {
```
?

libs/full/async_cuda/include/hpx/async_cuda/cuda_executor.hpp
```diff
@@ -163,29 +164,21 @@ namespace hpx { namespace cuda { namespace experimental {
...
+                    // insert the stream handle in the arg list and call the cuda function
+                    detail::dispatch_helper<R, Params...> helper{};
+                    helper(cuda_kernel, std::forward<Args>(args)..., stream_);
+                    return get_future();
+                },
+                [&](std::exception_ptr ep) {
+                    hpx::future<void> result;
+                    auto state = traits::detail::get_shared_state(result);
+                    state->set_exception(std::move(ep));
+                    return result;
```
hkaiser, line 180: We have `hpx::make_exceptional_future<void>()`. That could be used instead.

libs/parallelism/execution/include/hpx/execution/algorithms/detail/single_result.hpp
```diff
@@ -6,10 +6,15 @@
...
 #pragma once
 
+#include <hpx/config.hpp>
 #include <hpx/type_support/pack.hpp>
 
 #include <type_traits>
 
+#if defined(HPX_HAVE_CXX17_STD_VARIANT)
+#include <variant>
+#endif
```
hkaiser, line 16: Probably unrelated to this PR, but should we add serialization support for `std::variant` (could be simply taken from here: https://github.com/STEllAR-GROUP/phylanx/blob/master/phylanx/util/serialization/variant.hpp)?

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
```diff
@@ -0,0 +1,377 @@
...
+        struct value_visitor
+        {
+            std::decay_t<R> r;
+
+            template <typename Ts>
+            void operator()(Ts const& ts)
+            {
+                hpx::util::invoke_fused(
+                    hpx::util::bind_front(
+                        hpx::execution::experimental::set_value, std::move(r)),
```
hkaiser, line 59: I like how the CPO instances can be directly used as functional versions of the API they represent.

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
```diff
@@ -0,0 +1,377 @@
...
+                    if (!continuations.empty())
+                    {
+                        l.unlock();
+                        for (auto const& continuation : continuations)
+                        {
+                            continuation();
+                        }
+
+                        continuations.clear();
+                    }
```
hkaiser, line 216:
```suggestion
                    boost::container::small_vector<continuation_type, 1> conts;
                    {
                        std::unique_lock<hpx::util::spinlock> l{mtx};
                        std::swap(continuations, conts);
                    }
                    for (auto const& continuation : conts)
                    {
                        continuation();
                    }
```

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
```diff
@@ -0,0 +1,377 @@
...
+            public:
+                template <typename R,
+                    typename =
+                        std::enable_if_t<std::is_rvalue_reference_v<R&&>>>
+                void add_continuation(R&& r)
+                {
+                    if (predecessor_done)
+                    {
+                        std::visit(
+                            done_error_value_visitor<R>{std::move(r)}, v);
```
hkaiser, line 228: Why moving the forwarding reference?

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
```diff
@@ -0,0 +1,377 @@
...
+                // Eagerly start the work received until this point.
+                //
+                // P1897r3 says "When start is called on os2 [the operation
+                // state resulting from connecting the ensure_started_sender to
+                // a receiver], call execution::start(os [the operation state
+                // resulting from connecting S to ensure_started_receiver])",
+                // which would indicate that start should be called later.
+                // However, to fulfill the promise that ensure_started actually
+                // eagerly submits the sender S we call start already here.
+                st->start();
```
hkaiser, line 305: If `st->start()` throws, then this will leak the memory `st` refers to.

hkaiser: I take this back, `st` is an intrusive_ptr. All is well.

hkaiser: However, since the memory is allocated using an arbitrary allocator, shouldn't it be deleted using the same allocator (i.e. shouldn't the shared_state store the allocator for later use)?

libs/parallelism/execution/include/hpx/execution/algorithms/let_error.hpp
```diff
@@ -0,0 +1,282 @@
...
+                hpx::util::detail::concat_pack_of_packs_t<hpx::util::detail::
+                        transform_t<successor_sender_types<Variant>,
+                            value_types<Tuple, Variant>::template apply
+#if defined(HPX_CLANG_VERSION) && HPX_CLANG_VERSION < 110000
+                            >
+                    //
+                    >>;
+#else
+                            >>>;
+#endif
```
hkaiser, line 84: This is strange! Does clang complain about `>>>`?

libs/parallelism/execution/include/hpx/execution/algorithms/let_error.hpp
```diff
@@ -0,0 +1,282 @@
...
+                        R_&& r, F_&& f, operation_state& os)
+                      : r(std::forward<R_>(r))
+                      , f(std::forward<F_>(f))
+                      , os(os)
+                    {
+                    }
+
+                    struct start_visitor
+                    {
+                        void operator()(std::monostate)
```
hkaiser, line 116:
```suggestion
                        void operator()(std::monostate) const
```

libs/parallelism/execution/include/hpx/execution/algorithms/let_error.hpp
```diff
@@ -0,0 +1,282 @@
...
+                        {
+                            std::terminate();
+                        }
+
+                        template <typename OS_>
+                        void operator()(OS_& os)
```
hkaiser, line 122:
```suggestion
                        void operator()(OS_& os) const
```

hkaiser: ... same in other places

libs/core/datastructures/include/hpx/datastructures/tuple.hpp
msimberg: No, I don't think so. Good catch. I've defaulted it (but not tested it).

libs/core/type_support/include/hpx/type_support/detail/with_result_of.hpp
msimberg: I think that would make sense. This isn't really intended to be used multiple times anyway, and if it needs to be used like that one should not move f into the wrapper.

libs/parallelism/execution/include/hpx/execution/algorithms/detail/single_result.hpp
msimberg: If you think it would be useful, sure.

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
msimberg: What's the motivation behind this suggestion?

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
msimberg: It's constrained with `is_rvalue_reference` because I really only want to move the receiver here. Is there a clearer idiom for not treating `R&&` as a universal reference, but really only as an rvalue reference?

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
msimberg: Very good point, I completely missed that! I'll fix that.

libs/parallelism/execution/include/hpx/execution/algorithms/let_error.hpp
msimberg: Exactly. It's fixed in 11 though.

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
msimberg: I think I've fixed these now (it was the same situation in `detach.hpp`).

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
hkaiser: Your code accesses `continuations` outside of the lock.

msimberg: It does, but no continuations will be added to the vector after this point. The lock around `continuations.empty()` is enough in this case (there's a separate atomic variable which is set to true when one of `set_done/error/value` has been called and that is checked when adding continuations; if one of the `set_` functions have been called the continuation will be called inline). I realize it might not be 100% clear so let me know if this doesn't make sense.

libs/parallelism/execution/include/hpx/execution/algorithms/detail/single_result.hpp
hkaiser: I'll create a separate PR, then. So far there was no need for this as we didn't use the type ourselves.

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
hkaiser: Understood. I still think the suggested change would make it more idiomatic without adding overhead.

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
hkaiser: FWIW, a simple `template <typename T> void add_continuation(T&) = delete;` would do the trick as well in which case we could avoid the explicit SFINAE. 

Also, while `std::is_rvalue_reference_v<R&&>` is doing the right thing, I'd would have expected to see `!std::is_lvalue_reference_v<R>` instead. But that's just a personal preference.

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
```diff
@@ -0,0 +1,379 @@
...
+            typename Allocator = hpx::util::internal_allocator<>>
+        friend constexpr HPX_FORCEINLINE auto tag_fallback_invoke(
+            ensure_started_t, S&& s, Allocator const& a = {})
+        {
+            return detail::ensure_started_sender<S, Allocator>{
+                std::forward<S>(s), a};
+        }
+
+        template <typename S, typename Allocator,
+            typename NewAllocator = hpx::util::internal_allocator<>>
```
hkaiser, line 363: What is the `NewAllocator` being used for?

libs/parallelism/execution/include/hpx/execution/algorithms/just.hpp
```diff
@@ -55,20 +57,37 @@ namespace hpx { namespace execution { namespace experimental {
...
+            auto connect(R&& r) &&
+            {
+                return operation_state<R>{std::forward<R>(r), std::move(ts)};
+            }
+
+            template <typename R>
+            auto connect(R&& r) &
             {
-                return operation_state<R>(std::forward<R>(r), std::move(ts));
+                return operation_state<R>{r, ts};
```
hkaiser, line 90:
```suggestion
                return operation_state<R>{std::forward<R>(r), ts};
```

libs/parallelism/execution/include/hpx/execution/algorithms/let_error.hpp
```diff
@@ -0,0 +1,284 @@
...
+                        R_&& r, F_&& f, operation_state& os)
+                      : r(std::forward<R_>(r))
+                      , f(std::forward<F_>(f))
+                      , os(os)
+                    {
+                    }
+
+                    struct start_visitor
+                    {
+                        void operator()(std::monostate) const
```
hkaiser, line 116: You might want to add `HPX_NORETURN` here.

libs/parallelism/execution/include/hpx/execution/algorithms/let_error.hpp
```diff
@@ -0,0 +1,284 @@
...
+                        {
+                            std::terminate();
+                        }
+
+                        template <typename OS_>
+                        void operator()(OS_& os) const
```
hkaiser, line 122: You have disabled this function for `E == std::monostate` below (https://github.com/STEllAR-GROUP/hpx/pull/5274/files#diff-99bae62aa9f979fbaf1b4aab04e665f35655abd25d684a45a23b11ebfb31f760R142). Would there be a need to do the same here?

libs/parallelism/execution/include/hpx/execution/algorithms/let_error.hpp
```diff
@@ -0,0 +1,284 @@
...
+                        std::forward<PS_>(ps),
+                        let_error_predecessor_receiver(
+                            std::forward<R_>(r), std::forward<F_>(f), *this))}
+                {
+                }
+
+                operation_state(operation_state&&) = delete;
+                operation_state& operator=(operation_state&&) = delete;
+                operation_state(operation_state const&) = delete;
+                operation_state& operator=(operation_state const&) = delete;
```
hkaiser, line 246: Sometimes the `operation_state` is made non-copyable, sometimes not. Is there a pattern?

libs/parallelism/execution/include/hpx/execution/algorithms/on.hpp
```diff
@@ -7,88 +7,233 @@
...
+                        hpx::util::invoke_fused(
+                            hpx::util::bind_front(
+                                hpx::execution::experimental::set_value,
+                                std::move(receiver)),
+                            std::forward<Ts>(ts));
+                    }
+                };
+
+                template <typename E>
+                void set_error_scheduler_sender(E&& e) noexcept
```
hkaiser, line 198: General question: would it be possible (make sense) to define all member functions on one-shot objects as `&&`?

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
msimberg: That's a very good question. The idea behind this overload is that `ensure_started(ensure_started(sender))` would be the same as `ensure_started(sender)`. `NewAllocator` was meant to be the allocator type for the second (outer) `ensure_started` but be ignored. However, I've used the wrong template parameter in the overload.

In general, I'll admit I'm not entirely sure if this is the right decision (to ignore the new allocator), but I'd find it weird to require them to be the same as well. Do you have a preference or prior art for something like this?

libs/parallelism/execution/include/hpx/execution/algorithms/let_error.hpp
msimberg: It's probably best to do so. Good catch!

libs/parallelism/execution/include/hpx/execution/algorithms/let_error.hpp
msimberg: In general they should be non-copyable, and non-movable (or rather, they can be copyable and movable, but to assume that they are would be an error). I see I've missed at least one spot. Did you notice many places where copy/move constructors/assignment operators are not deleted?

libs/parallelism/execution/include/hpx/execution/algorithms/on.hpp
msimberg: Possibly. The `connect`s on senders should already be marked `&&` where it makes sense. Operation state member functions could be marked `&`, but I'm not sure what it would gain us given that they should be non-copyable and non-movable. I don't really have a preference either way though. Do you think it would be preferable?

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
msimberg: Default constructor + swap compared to a call to `empty`? Probably not a huge difference, but not nothing. In what sense would the former be more idiomatic?

libs/parallelism/execution/include/hpx/execution/algorithms/on.hpp
hkaiser: Marking member functions as `&&` or `&` conveys intent, which is always a good thing as it might help reducing misuse. A good API is one that is easy to use correctly but difficult to use in unintended ways.

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
hkaiser: I'm not sure I fully understand the implications. It just struck me that there is a (defaulted) template argument that is not used anywhere.

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
msimberg: > FWIW, a simple `template <typename T> void add_continuation(T&) = delete;` would do the trick as well in which case we could avoid the explicit SFINAE.

That's a neat trick, I'll give it a shot!

> Also, while `std::is_rvalue_reference_v<R&&>` is doing the right thing, I would have expected to see `!std::is_lvalue_reference_v<R>` instead. But that's just a personal preference.

I prefer saying what I want it to be rather than what I don't want it to be. But with the previous suggestion I won't need either so hopefully we'll both be happy then!

hkaiser: > FWIW, a simple `template <typename T> void add_continuation(T&) = delete;` would do the trick as well in which case we could avoid the explicit SFINAE.

This might also improve error messages, BTW.

libs/parallelism/execution/include/hpx/execution/algorithms/on.hpp
msimberg: Absolutely fair! I'll make the change at least for the P0443 APIs, and I'll see where it makes sense for the internal ones.

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
```diff
@@ -0,0 +1,379 @@
...
+        friend constexpr HPX_FORCEINLINE auto tag_fallback_invoke(
+            ensure_started_t, detail::ensure_started_sender<S, Allocator> s,
+            Allocator const& = {})
```
msimberg, line 366: @hkaiser continuing from https://github.com/STEllAR-GROUP/hpx/pull/5274/files#r626815356 here.

Sorry that was unclear. There's a mistake here, and this is what I meant to have here:
```suggestion
            NewAllocator const& = {})
```
With that change, my previous explanation might make more sense. `NewAllocator` is the type of an allocator the the new `ensure_started_sender` would use, if it weren't for the fact that this overload simply returns the predecessor sender unchanged.

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
hkaiser: What I meant was that it's not obvious to me why calling `.empy()` needs a lock but actually accessing the container doesn't. Perhaps would adding a comment clarify the code. 

As for the slightly larger overhead of my suggestion: I'd prefer having correct over fast code.

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
msimberg: I changed this to what I originally intended. Does it make sense to you to simply ignore the new allocator?

libs/parallelism/execution/include/hpx/execution/algorithms/let_error.hpp
msimberg: I've updated these now to always be non-movable and non-copyable.

libs/parallelism/execution/include/hpx/execution/algorithms/let_error.hpp
msimberg: Updated. I also added a `HPX_UNREACHABLE` macro to replace the explicit `std::terminate` with `HPX_ASSERT_MSG(false, message to say that reaching this point is a bug); std::terminate();`. Does that make sense to you? It might be useful in some other places in the codebase as well, but I won't change those here (I know there are a few `HPX_ASSERT(false)`s e.g. in switch statements.).

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
msimberg: I fully agree with "I'd prefer having correct over fast code". However, to the best of my knowledge this is fine. Please let me know if you think it actually is racy! Being on the safe side is absolutely better than possibly wrong.

I've extensively commented the sections where the vector is accessed with hopefully useful explanations. The lock around the call to `empty()` was in fact a red herring. Not even that is needed, and I've changed it accordingly. The important thing is synchronizing with the lock so that threads adding to the vector see modifications to the `predecessor_done` flag correctly.

I also think the same change could be made to `future_data` to reduce the critical section there when adding/calling continuations.

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
hkaiser: Hmmm. Wouldn't it be better to return a new sender that relies on the new allocator (if the allocator is different, that is)? What would be the point of this API otherwise? Please note that I'm not sure I have the full picture here, so I'm likely wrong...

msimberg: That's what I'm not entirely sure about. It depends entirely on why the user specifies a different allocator here. I guess we can agree that if the allocator is the same there's no reason to create a new sender? That's the main purpose of this overload. I get a sender from somewhere that may or may not be started. I call `ensure_started` to make sure it's started, but if it's already been started there's no need for it to be started again (especially as this algorithm requires heap allocation).

It's the case when the allocators are different that things get interesting. I'd say no allocation is "better" than any custom allocator that they user might have in the sense that it's free, but then I might not see all the use cases of providing a custom allocator, other than for performance reasons.

libs/parallelism/execution/include/hpx/execution/algorithms/detail/single_result.hpp
hkaiser: NWM, we already have that :/

libs/parallelism/execution/include/hpx/execution/algorithms/ensure_started.hpp
hkaiser: Yes, if the allocator is the same and the provided sender was already started, then this function can simply return its argument. If its the same allocator but the argument was not started, simply start it and return the argument.

If the allocator is different I think you should create a new sender that uses the new allocator. Would going from the CPU realm to the GPU realm be a valid use case for this (I'm not sure)?

msimberg: Yeah, I think that's fair enough. I also don't know if there's a valid use case for it. I'd expect most of our initial use cases to be using the default allocator anyway and we know the behaviour there. For the the cases where the allocators are different we can come back and optimize it if it turns out that's necessary, but for now just creating a new sender sounds like the right thing to do.

Note: when I say "if the predecessor sender has been started" it really just means "if the predecessor sender is an ensure_started_sender". There may be other senders that eagerly submit their work which this algorithm doesn't care about. This algorithm ensures that the sender is connected and the resulting operation state is started (which is independent from whether or not the actual work was submitted).

msimberg:
retest lsu

hkaiser (approved):
LGTM, thanks!

msimberg:
I only pushed a formatting fix as the last thing. Merging now.

PR #5277, opened 2021-04-06
Small fixes and improvements to CUDA/MPI polling

msimberg:
- Return a status enum from the polling functions to indicate if the scheduler should call the idle callback (useful if polling is the only thing done on a separate pool, and the pool would otherwise call the idle callback even if there's polling to be done).
- The CUDA polling function used to spin waiting for events to complete. As far as I can tell this was unintentional (@biddisco? the iterator was not incremented on `cudaErrorNotReady`).
- Avoid quadratic worst-case complexity in the CUDA polling function in terms of the number of active futures (if every future is ready every iteration would shift the remaining elements to the front when calling `vector::erase`).
- Use CUDA host pinned memory in the `cuda_future` test to actually enable asynchronous data transfers.

### DIFF
--- a/libs/full/async_cuda/src/cuda_future.cpp
+++ b/libs/full/async_cuda/src/cuda_future.cpp
@@ -184,48 +184,49 @@ namespace hpx { namespace cuda { namespace experimental { namespace detail {
             hpx::cuda::experimental::cuda_event_pool::get_event_pool();
 
         // iterate over our list of events and see if any have completed
-        detail::future_data_ptr fdp;
-        using i_type = std::vector<future_data_ptr>::iterator;
-        for (i_type it = future_vec.begin(); it != future_vec.end();)
-        {
-            fdp = *it;
-            cudaError_t status = cudaEventQuery(fdp->event_);
-            if (status == cudaErrorNotReady)
-            {
-                // this event has not been triggered yet
-                ++it;
-                continue;
-            }
-            else if (status == cudaSuccess)
-            {
-                fdp->set_data(hpx::util::unused);
-                // clang-format off
+        future_vec.erase(
+            std::remove_if(future_vec.begin(), future_vec.end(),
+                [&](detail::future_data_ptr fdp) {
+                    cudaError_t status = cudaEventQuery(fdp->event_);
+                    if (status == cudaErrorNotReady)
+                    {
+                        // this event has not been triggered yet
+                        return false;
+                    }
+                    else if (status == cudaSuccess)
+                    {
+                        fdp->set_data(hpx::util::unused);
+                        // clang-format off
                 cud_debug.debug(debug::str<>("set ready vector")
                     , "event", debug::hex<8>(fdp->event_)
                     , "futures", debug::dec<3>(get_active_futures().size()));
-                // clang-format on
-                // drop future and reuse event
-                it = future_vec.erase(it);
-                pool.push(fdp->event_);
-            }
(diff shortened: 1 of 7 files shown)


hkaiser (commented):
Could you explain the rationale for these changes, please?

msimberg:
> Could you explain the rationale for these changes, please? Specifically, I would like to understand this: https://github.com/STEllAR-GROUP/hpx/pull/5277/files#diff-93e12b963ad29614127c32b33e22c54428ac32e8c435de5d9be363b284722c62R980-R984

Is it the particular implementation, the motivation, or both that you're wondering about?

hkaiser:
> > Could you explain the rationale for these changes, please? Specifically, I would like to understand this: https://github.com/STEllAR-GROUP/hpx/pull/5277/files#diff-93e12b963ad29614127c32b33e22c54428ac32e8c435de5d9be363b284722c62R980-R984
> 
> Is it the particular implementation, the motivation, or both that you're wondering about?

Mostly the motivation...

msimberg:
> > > Could you explain the rationale for these changes, please? Specifically, I would like to understand this: https://github.com/STEllAR-GROUP/hpx/pull/5277/files#diff-93e12b963ad29614127c32b33e22c54428ac32e8c435de5d9be363b284722c62R980-R984
> > 
> > 
> > Is it the particular implementation, the motivation, or both that you're wondering about?
> 
> Mostly the motivation...

This came up in DLA-Future where they're attempting to use the MPI polling integration for their communication. One configuration is to create a separate thread pool with a dedicated worker thread only for polling, and enabling polling only on that thread pool. However, if the thread pool is really dedicated only polling it means that it never gets any tasks to execute and with the default settings it starts suspending itself (for up to a second at a time) even though there are CUDA events/MPI requests to poll for since the scheduler knows nothing about what state the polling vectors/queues are in. With this change the scheduling loop won't sleep if the polling function claims that it still has something to poll for.

We've discussed an alternative solution as well where a dedicated polling thread would be created (without the scheduling loop) for polling, but that's a larger change, so this works as an intermediate solution until we look at that.

hkaiser:
@msimberg thanks for the explanation!

hkaiser (dismissed):
LGTM, thanks!

aurianer (dismissed):
LGTM thanks!

libs/full/async_cuda/src/cuda_future.cpp
```diff
@@ -193,6 +193,7 @@ namespace hpx { namespace cuda { namespace experimental { namespace detail {
             if (status == cudaErrorNotReady)
             {
                 // this event has not been triggered yet
+                ++it;
```
biddisco: That is a very disturbing mistake, and could have been responsible for some dramatic slowdowns overall.

biddisco (approved):
That's a bit nicer. Thanks

msimberg:
> I messed up my review, by looking at each commit individually. I'd like to rerun the octotiger/kokkkos tests hat we did with tthe dedicated polling pool to see if there's any noticable difference. The missing iterator increment might have caused a noticable slowdown on one stream if for some reason an event was held up.

Thanks @biddisco for having a look. Yes, agreed that it might make a difference. It may actually also improve timings in some situations by not returning to executing normal tasks, but only testing will tell... @G-071 this might be interesting for you as well, especially if you can easily rerun the tests we did earlier with this branch/master once this is merged.

PR #5279, opened 2021-04-07
Fix more GCC 10 deprecation warnings

msimberg:
Follow up to #5272. I wasn't very thorough the first time around. This time I've enabled warnings-as-errors on the GCC 10 build so this should take care of the rest.


aurianer (dismissed):
LGTM thanks!

PR #5281, opened 2021-04-07
Fix Segmentation fault on foreach_datapar_zipiter

srinivasyadav18:
## Proposed Changes

 - The len variable in initialised to count but did not use rightly in the loop.
  - replace **count** with **len**, must be a typo. 
 

https://github.com/STEllAR-GROUP/hpx/pull/5235#issuecomment-814671031

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/datapar/iterator_helpers.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/datapar/iterator_helpers.hpp
@@ -33,7 +33,7 @@ namespace hpx { namespace parallel { namespace util { namespace detail {
             typedef typename std::iterator_traits<Iter>::value_type value_type;
             return (reinterpret_cast<std::uintptr_t>(std::addressof(*it)) &
                        (traits::vector_pack_alignment<value_type>::value -
-                           1)) != 0;
+                           1)) == 0;
         }
     };
 
@@ -304,7 +304,7 @@ namespace hpx { namespace parallel { namespace util { namespace detail {
             typename hpx::util::invoke_result<F, V1*, V2*>::type
             callv(F&& f, Iter1& it1, Iter2& it2)
         {
-            if (is_data_aligned(it1) || is_data_aligned(it2))
+            if (!is_data_aligned(it1) || !is_data_aligned(it2))
             {
                 return invoke_vectorized_in2<V1, V2>::call_unaligned(
                     std::forward<F>(f), it1, it2);
@@ -452,7 +452,7 @@ namespace hpx { namespace parallel { namespace util { namespace detail {
 
             typedef typename traits::vector_pack_type<value_type>::type V;
 
-            if (is_data_aligned(it) || is_data_aligned(dest))
+            if (!is_data_aligned(it) || !is_data_aligned(dest))
             {
                 invoke_vectorized_inout1<V>::call_unaligned(
                     std::forward<F>(f), it, dest);
@@ -477,8 +477,8 @@ namespace hpx { namespace parallel { namespace util { namespace detail {
             typedef typename traits::vector_pack_type<value1_type>::type V1;
             typedef typename traits::vector_pack_type<value2_type>::type V2;
 
-            if (is_data_aligned(it1) || is_data_aligned(it2) ||
-                is_data_aligned(dest))
+            if (!is_data_aligned(it1) || !is_data_aligned(it2) ||
+                !is_data_aligned(dest))
             {
                 invoke_vectorized_inout2<V1, V2>::call_unaligned(
                     std::forward<F>(f), it1, it2, dest);
--- a/libs/parallelism/algorithms/include/hpx/parallel/datapar/loop.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/datapar/loop.hpp
@@ -124,7 +124,7 @@ namespace hpx { namespace parallel { namespace util {
                 iterator_datapar_compatible<Begin>::value, Begin>::type
             call(Begin first, End last, F&& f)
             {
-                while (is_data_aligned(first) && first != last)
+                while (!is_data_aligned(first) && first != last)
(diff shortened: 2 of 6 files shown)


libs/parallelism/algorithms/include/hpx/parallel/datapar/loop.hpp
```diff
@@ -252,7 +252,7 @@ namespace hpx { namespace parallel { namespace util {
                 static std::size_t constexpr size =
                     traits::vector_pack_size<V>::value;
 
-                for (std::int64_t lenV = std::int64_t(count - (size + 1));
+                for (std::int64_t lenV = std::int64_t(len - (size + 1));
```
hkaiser, line 255: Please adher to our naming conventions (see here: https://github.com/STEllAR-GROUP/hpx/wiki/HPX-Source-Code-Structure-and-Coding-Standards)

msimberg:
retest

srinivasyadav18:
Here https://github.com/STEllAR-GROUP/hpx/pull/5281/commits/a2d75115748193cb2c7c39f949e737a99e84f39d, is_data_aligned is logically correct and is used in correct sense, but the convention was reverse, it returns false when data is aligned. I now changed it to return true when is aligned.

msimberg:
retest

libs/parallelism/algorithms/include/hpx/parallel/datapar/iterator_helpers.hpp
```diff
@@ -33,7 +33,7 @@ namespace hpx { namespace parallel { namespace util { namespace detail {
             typedef typename std::iterator_traits<Iter>::value_type value_type;
             return (reinterpret_cast<std::uintptr_t>(std::addressof(*it)) &
                        (traits::vector_pack_alignment<value_type>::value -
-                           1)) != 0;
+                           1)) == 0;
```
msimberg, line 36: Well spotted!

libs/parallelism/algorithms/tests/regressions/for_each_datapar.cpp
```diff
@@ -0,0 +1,54 @@
+#include <hpx/algorithm.hpp>
```
msimberg, line 1: Does this test something that the existing tests didn't test (after all the existing tests segfaulted)? If not we can leave this out. If yes is would this test be useful to add to all the other `for_each` tests instead? And if yes, but the file will stay here it needs a license header.

srinivasyadav18: the existing tests starts vectorisation at only one iterator i.e begin(), and gets aligned with prologue loop, but this has very less chances to check segfaults, checking at more iterator positions assures that iterator at different positions can be adjusted for alignment and then go for vectorisation.

libs/parallelism/algorithms/include/hpx/parallel/datapar/iterator_helpers.hpp
srinivasyadav18: I still need to fix this https://github.com/STEllAR-GROUP/hpx/blob/fixing_datapar/libs/parallelism/algorithms/include/hpx/parallel/datapar/zip_iterator.hpp#L35-L45 as https://app.circleci.com/pipelines/github/srinivasyadav18/hpx/40/workflows/f86b0f5f-1086-4d1a-b716-fc630649b721/jobs/613/tests#failed-test-1 this is failing because now is_data_aligned changed.

libs/parallelism/algorithms/tests/regressions/for_each_datapar.cpp
srinivasyadav18: @msimberg Would you recommend adding this new test, or update the existing unit tests with these changes?. Now all the seg faults are fixed.

libs/parallelism/algorithms/include/hpx/parallel/datapar/zip_iterator.hpp
```diff
@@ -40,7 +40,7 @@ namespace hpx { namespace parallel { namespace util { namespace detail {
             auto const& t = it.get_iterator_tuple();
             bool const sequencer[] = {
                 false, is_data_aligned(hpx::get<Is>(t))...};
```
hkaiser, line 42:
```suggestion
                true, is_data_aligned(hpx::get<Is>(t))...};
```

hkaiser:
> why do we the need of first value (false/true) in sequencer, when we are using only from 1 to sizeof...(Is) + 1? in std::all_of(), why can't we directly use from 0 to sizeof...(Is) without the first value in sequencer?,

We definitely could.

srinivasyadav18:
> > why do we the need of first value (false/true) in sequencer, when we are using only from 1 to sizeof...(Is) + 1? in std::all_of(), why can't we directly use from 0 to sizeof...(Is) without the first value in sequencer?,
> 
> We definitely could. Be careful to account for a zero length sequence, however.

@hkaiser Okay, then the existing solution is a better one. I did not realise about zero length sequence.
And could you please suggest anything about this https://github.com/STEllAR-GROUP/hpx/pull/5281#discussion_r611003849
After adding or editing the tests, most probably there are no further issues related to datapar.

libs/parallelism/algorithms/tests/regressions/for_each_datapar.cpp
msimberg: I see, thanks for the explanation. Let's keep this as it is then! It does still need the license header though.

srinivasyadav18: @msimberg Done!

srinivasyadav18:
@msimberg @hkaiser I have made the changes! Please mention if any further changes are required :-).

msimberg:
@srinivasyadav18 I'll merge this into the other branch now, let's do further changes if needed there (I have nothing in mind right now).

PR #5283, opened 2021-04-07
Require minimum C++17 and change CUDA handling

### LINKED ISSUE #5317: Consider making hipblas optional
It may be beneficial to make hipblas (and cublas as well I suppose) optional, since it doesn't necessarily come with all hip installations. I'm not sure what consequences this would have for compatibility macros for hip/cuda though, and it may be more complicated than it sounds.

aurianer:
- Remove support for C++14
- Update CUDA cmake handling from find_package(CUDA) to enable_language(CUDA)
- Update CI configurations
- Disable icc CI
- Fixes #5317 (disabling cuBLAS hen not available) 

TODO:
- [x] Check NVCC builder (problematic)
- [x] " clang cuda builder
- [x] Fix hipcc warnings
- [x] Remove C++17 feature tests that are not needed anymore + corresponding macros checks


docs/sphinx/about_hpx/history.rst
```diff
@@ -37,7 +37,7 @@ project is to create a high quality, freely available, open source
 implementation of ParalleX concepts for conventional and future systems by
 building a modular and standards conforming runtime system for SMP and
 distributed application environments. The API exposed by |hpx| is conformant to
-the interfaces defined by the C++11/14 ISO standard and adheres to the
+the interfaces defined by the C++11/14/17/20 ISO standard and adheres to the
```
msimberg, line 40: Let's just change this to
```suggestion
the interfaces defined by the C++ ISO standard and adheres to the
```

docs/sphinx/about_hpx/people.rst
```diff
@@ -230,7 +230,7 @@ the project through discussions, pull requests, documentation patches, etc.
   specific patches for FreeBSD and MSVC12.
 * Agustin Berge, who contributed patches fixing some very nasty hidden template
   meta-programming issues. He rewrote large parts of the API elements ensuring
-  strict conformance with C++11/14.
+  strict conformance with C++11/14/17/20.
```
msimberg, line 233:
```suggestion
  strict conformance with the C++ ISO standard.
```

docs/sphinx/manual/building_hpx.rst
```diff
@@ -521,15 +521,14 @@ libraries, you need to specify them as link dependencies in your build (See
 :ref:`creating_hpx_projects`).
 
 As |hpx| is a modern C++ library, we require a certain minimum set of features
-from the C++11 standard. In addition, we make use of certain C++14 features if
+from the C++17 standard. In addition, we make use of certain C++20 features if
 the used compiler supports them. This means that the |hpx| build system will try
 to determine the highest support C++ standard flavor and check for availability
```
msimberg, line 526: My bad. This is outdated. This should say we use C++17 by default, and features from a more recent standard can be opted into explicitly.

msimberg (commented):
Thanks for looking into this!

msimberg:
We can leave this for a separate PR?
>     * [ ]  Update some old code with C++17 features

docs/sphinx/about_hpx/people.rst
aurianer: I just added an upper case to standard as it is appearing on cppreference, any objections?

hkaiser:
> We can leave this for a separate PR?
> 
> > ```
> > * [ ]  Update some old code with C++17 features
> > ```

I second this suggestion. Also @aurianer thanks for pushing this forward!

msimberg:
This tells me we're most likely not setting the `-std` flag for nvcc compilationn: https://cdash.cscs.ch/viewBuildError.php?buildid=155780. I think we can use the `cuda_std_XX` properties/`(CMAKE_)CUDA_STANDARD` variables the same way we use `cxx_std_XX`/`(CMAKE_)CXX_STANDARD` to set the standard when compiling with nvcc (see https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CUDA_KNOWN_FEATURES.html, https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_STANDARD.html?highlight=cmake_cuda_standard, available from CMake 3.17 onwards).

docs/sphinx/about_hpx/people.rst
msimberg: Forgotten push?

aurianer: yes it was on a temporary branch ;)

cmake/HPX_SetupCUDA.cmake
```diff
@@ -22,8 +22,20 @@ if(HPX_WITH_CUDA AND NOT TARGET Cuda::cuda)
...
+        "You've set CMAKE_CUDA_STANDARD to ${CMAKE_CUDA_STANDARD}, which is less than 17 (the minimum required by HPX)"
+      )
+    endif()
+  else()
+    set(CMAKE_CUDA_STANDARD 17)
+  endif()
+
   add_library(Cuda::cuda INTERFACE IMPORTED)
   target_include_directories(Cuda::cuda INTERFACE ${CUDA_INCLUDE_DIRS})
+  target_compile_options(Cuda::cuda INTERFACE cuda_std_${CMAKE_CUDA_STANDARD})
```
msimberg, line 38: This should be `target_compiler_features`.

aurianer: Thanks!

aurianer:
I'm encountering problems in the pkgconfig generation step because of the cmake generator expressions for languages. The error is the following:
```
  Evaluation file to be written multiple times with different content.  This
  is generally caused by the content evaluating the configuration type,
  language, or location of object files:
```
when generating `lib/pkgconfig/hpx_application_debug.pc` or one of the 3 others.
I thought about filtering those options out and make pkgconfig specific for CXX with a `list(FILTER ...` but I encountered another error, the flags being separated with a ; inside the generator expression. CMake considers therefore $<$<COMPILE_LANGUAGE:CUDA>:--extended-lambdas;--expt-relaxed-constexpr> as 2 items `$<$<COMPILE_LANGUAGE:CUDA>:--extended-lambdas` and `--expt-relaxed-constexpr>` and only the former is filtered out.

If anyone have a simple solution.. otherwise I would suggest we drop the support for pkgconfig, as it is not the first time we encounter problems with pkgconfig (object libraries) and that we have very few users using it.

biddisco:
I'd vote to drop support of pkgconfig

hkaiser:
@aurianer FWIW, you can use `$<SEMICOLON>` to place semicolons inside generator expressions. Not sure if this helps, though.

hkaiser:
> I'd vote to drop support of pkgconfig

@stevenrbrandt you said at some point that you'd be interested in maintaining the pkgconfig support in HPX as you're using it.

aurianer:
> @aurianer FWIW, you can use `$<SEMICOLON>` to place semicolons inside generator expressions. Not sure if this helps, though.

Thanks for the suggestion but I don't quite see how to replace only the semicolons inside the generator expressions without replacing the semicolons that separate the generator expressions, the pkgconfig flags are of type:
```
$<$<COMPILE_LANGUAGE:CUDA>:--extended-lambdas;--expt-relaxed-constexpr>;$<$<CONFIG:Debug>:HPX_DEBUG>; etc.
```

biddisco:
> Thanks for the suggestion but I don't quite see how to replace only the semicolons inside the generator expressions without replacing the semicolons that separate the generator expressions, the pkgconfig flags are of type:
> 
> ```
> $<$<COMPILE_LANGUAGE:CUDA>:--extended-lambdas;--expt-relaxed-constexpr>;$<$<CONFIG:Debug>:HPX_DEBUG>; etc.
> ```

string(REGEX REPLACE <match-regex> <replace-expr> <out-var> <input>...)

?

stevenrbrandt:
I don't have a solution as I don't have deep knowledge of cmake. I do think that having some way of knowing how to compile programs without cmake is important, be that pkgconfig or something else.

aurianer:
About @biddisco solution I still don't manage to make it work, my minimal example being:
```
message("Before flags !!!!!!!!\n ${_flags}")                                    
string(REGEX REPLACE "\\$<\\$<COMPILE_LANGUAGE:CUDA>:[^>]*>" "" _flags ${_flags})
message("After flags !!!!!!!!\n ${_flags}")
```
output
```
Before flags !!!!!!!!
$<$<CONFIG:Debug>:HPX_DEBUG>;$<$<COMPILE_LANGUAGE:CUDA>:;ASIO_DISABLE_CONSTEXPR;>;$<$<CONFIG:Debug>:HPX_DEBUG>
After flags !!!!!!!!
 $<$<CONFIG:Debug>:HPX_DEBUG>$<$<CONFIG:Debug>:HPX_DEBUG>
```
The important semicolon between the 2 generator expressions is also gone, maybe I missed something..
@stevenrbrandt in [this commit](https://github.com/STEllAR-GROUP/hpx/pull/5283/commits/fdbc83537fff763636a01171688e27e8ce7e998f) I just disable pkgconfig generation when `HPX_WITH_CUDA` is enabled, would that be sufficient for you?

.jenkins/cscs/env-clang-cuda.sh
```diff
@@ -14,7 +14,7 @@ export HWLOC_ROOT="${APPS_ROOT}/hwloc-2.0.3-gcc-8.3.0"
 module load daint-gpu
 module load cudatoolkit/10.2.89_3.29-7.0.2.1_3.5__g67354b4
 module load Boost/1.75.0-CrayCCE-20.11
-spack load cmake@3.17.3
+spack load cmake@3.18.6 # Necessary to handle cuda_std_17 compiler feature
```
msimberg, line 17: Do we need the comment here? Most likely we should just bump the minimum required version to 3.18, no?

.jenkins/cscs/env-clang-cuda.sh
```diff
@@ -29,6 +29,10 @@ configure_extra_options+=" -DHPX_WITH_CXX${CXX_STD}=ON"
 configure_extra_options+=" -DHPX_WITH_COMPILER_WARNINGS=ON"
 configure_extra_options+=" -DHPX_WITH_COMPILER_WARNINGS_AS_ERRORS=ON"
 configure_extra_options+=" -DHPX_WITH_SPINLOCK_DEADLOCK_DETECTION=ON"
+# Necessity to specify the cuda compiler when switching from FindCUDA to
+# enable_language(CUDA)
```
msimberg, line 33: Necessary?

.jenkins/cscs/env-icc.sh
```diff
@@ -19,6 +20,7 @@ export CXX=`which CC`
 export CC=`which cc`
 
 configure_extra_options="-DCMAKE_BUILD_TYPE=Debug"
+configure_extra_options+=" -DHPX_WITH_CXX${CXX_STD}=ON"
```
msimberg, line 23: Leave as default? I don't mind specifying it explicitly, but we should leave at least one builder that uses the default. We had trouble in the past with the default and explicitly specifying the standard behaving differently.

cmake/HPX_AddModule.cmake
```diff
@@ -198,26 +198,13 @@ function(add_hpx_module libname modulename)
   endif()
 
   # create library modules
-  if(${name}_CUDA
-     AND HPX_WITH_CUDA
-     AND NOT HPX_WITH_CUDA_CLANG
+  # cmake-format: off
```
msimberg, line 201: The `cmake-format: off` can probably be removed?

cmake/HPX_SetupCUDA.cmake
```diff
@@ -6,57 +6,118 @@
 
 if(HPX_WITH_CUDA AND NOT TARGET Cuda::cuda)
 
-  find_package(CUDA REQUIRED)
+  if(HPX_WITH_CUDA_CLANG AND NOT (CMAKE_CXX_COMPILER_ID STREQUAL "Clang"))
```
msimberg, line 9: The question now is: do we still need the `HPX_WITH_CUDA_CLANG` option?

cmake/HPX_SetupCUDA.cmake
```diff
@@ -6,57 +6,118 @@
...
+    hpx_error(
+      "To use Cuda Clang, please select Clang as your default C++ compiler"
+    )
+  endif()
+
+  # cuda_std_17 not recognized for previous versions
+  cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
```
msimberg, line 16: I think we should just require this for all configurations. However, it does mean another CMake version bump. @hkaiser, @biddisco, objections?

cmake/HPX_SetupCUDA.cmake
```diff
@@ -6,57 +6,118 @@
...
-    )
+  # CUDA libraries used
+  add_library(Cuda::cuda INTERFACE IMPORTED)
+  # Toolkit targets like CUDA::cudart, CUDA::cublas, CUDA::cufft, etc. available
+  find_package(CUDAToolkit MODULE REQUIRED)
+  if(CUDAToolkit_FOUND)
+    target_link_libraries(Cuda::cuda INTERFACE CUDA::cudart)
+    target_link_libraries(Cuda::cuda INTERFACE CUDA::cublas)
   endif()
+  # Flag not working for CLANG CUDA
```
msimberg, line 50: Still not working?

cmake/HPX_SetupCUDA.cmake
```diff
@@ -6,57 +6,118 @@
...
+  target_compile_features(Cuda::cuda INTERFACE cuda_std_${CMAKE_CUDA_STANDARD})
 
-  add_library(Cuda::cuda INTERFACE IMPORTED)
-  target_include_directories(Cuda::cuda INTERFACE ${CUDA_INCLUDE_DIRS})
+  # CUDA variables
+  set(CUDA_SEPARABLE_COMPILATION ON)
+  set(POSITION_INDEPENDENT_CODE ON)
```
msimberg, line 55: I think this is a target property, and we set it for pretty much every library target as far as I can tell.

cmake/HPX_SetupCUDA.cmake
```diff
@@ -6,57 +6,118 @@
...
+  # Add all the architectures to be targeted
+  if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
+    hpx_warn("CMAKE_CUDA_ARCHITECTURES not specified, using sm_60 by default.")
+    set(CMAKE_CUDA_ARCHITECTURES 60)
+  endif()
```
msimberg, line 60: I'd leave this to the user.

cmake/HPX_SetupCUDA.cmake
```diff
@@ -6,57 +6,118 @@
...
+  # keywords for target_link_libraries (cuda)
+  set(CUDA_LINK_LIBRARIES_KEYWORD "PRIVATE")
```
msimberg, line 62: I *think* this can go. As far as I can tell it shouldn't be required anymore since we don't use `FindCUDA`.

cmake/HPX_SetupCUDA.cmake
```diff
@@ -6,57 +6,118 @@
...
+        Cuda::cuda INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:-w>
+      )
     else()
+      # Windows
       set(CUDA_PROPAGATE_HOST_FLAGS OFF)
-      target_link_directories(
-        Cuda::cuda INTERFACE ${CUDA_TOOLKIT_ROOT_DIR}/lib/x64
+      target_compile_options(
+        Cuda::cuda
+        INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:$<$<CONFIG:Debug>:
```
msimberg, line 74: Note: this generator expression would not play well with the pkgconfig filtering that we have in place now. It's not a problem here since this is on Windows, and pkgconfig is disabled there anyway. I'm just pointing it out in case we need a nested generator expression like this on platforms that do have pkgconfig support enabled.

.jenkins/cscs/env-clang-cuda.sh
aurianer: It fails currently I was just trying things but looks like it does't help on jenkins whereas it works locally, so any help appreciated ^^'

.jenkins/cscs/env-icc.sh
aurianer: Agreed

cmake/HPX_SetupCUDA.cmake
aurianer: The compile options were not the same so I assumed so (but this could be replaced with generator expressions). Will tackle this as soon as clang-cuda jenkins CI passes ^^'

cmake/HPX_SetupCUDA.cmake
aurianer: Still trying to figure out what happens, cause it works locally, I'll investigate more

cmake/HPX_SetupCUDA.cmake
aurianer: yes sorry I made them both target properties then I had an error about `CUDA_SEPARABLE_COMPILATION` not being a whitelisted property so I reverted the commit, will do it for `position_independant_code` thanks

cmake/HPX_SetupCUDA.cmake
aurianer: Yes I noticed it but I figured we didn't need it for now, should I make a more generic regex in the pkgconfig generation?

msimberg: No, not for me at least. I also don't think you can construct a regex to match the generic case (there can be arbitrarily many levels of nesting). I just wanted to point it out for the benefit of others. A comment about the limitations next to the regex might be worth it though.

.jenkins/cscs/env-clang-cuda.sh
msimberg: Maybe I'm confused. I just meant if the comment is necessary? What fails?

aurianer: Ah understood will remove the comment :ok_hand:

aurianer: This fails:
https://cdash.cscs.ch/viewConfigure.php?buildid=157946

cmake/HPX_SetupCUDA.cmake
aurianer: CUDA Clang problem solved when specifying explicitly the `CMAKE_CUDA_COMPILER` and `CMAKE_CUDA_EXTENSIONS` for compiler flags specific to the compiler used

components/performance_counters/io/src/io_counters.cpp
```diff
@@ -28,6 +28,9 @@
 #include <fstream>
 #include <iterator>
 #include <string>
+#if defined(HPX_HAVE_UNISTD_H)
+#include <unistd.h>
+#endif
```
msimberg, line 33: Just curious, on which configuration was this not available?

aurianer: It failed with Clang CUDA

.jenkins/cscs/env-clang-cuda.sh
```diff
@@ -14,7 +14,7 @@ export HWLOC_ROOT="${APPS_ROOT}/hwloc-2.0.3-gcc-8.3.0"
 module load daint-gpu
 module load cudatoolkit/10.2.89_3.29-7.0.2.1_3.5__g67354b4
```
msimberg, line 15: This would need to be updated to cuda 11.

cmake/HPX_AddComponent.cmake
```diff
@@ -250,18 +250,11 @@ function(add_hpx_component name)
     endforeach()
   endif()
```
msimberg, line 251: Do we still need this part?

cmake/HPX_AddExecutable.cmake
```diff
@@ -193,15 +193,9 @@ function(add_hpx_executable name)
     endforeach()
   endif()
```
msimberg, line 194: Same here?

libs/full/compute_cuda/tests/unit/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ set(default_executor_CUDA On)
 set(for_each_compute_CUDA On)
 set(for_loop_compute_CUDA On)
 
-if(HPX_WITH_CUDA_CLANG OR HPX_WITH_HIP)
+if(HPX_WITH_CLANG_CUDA OR HPX_WITH_HIP)
```
msimberg, line 14: By the way, any particular reason for renaming `HPX_WITH_CUDA_CLANG` to `HPX_WITH_CLANG_CUDA`? Is its purpose only to internally signal that we have CUDA enabled with clang, or are users still expected to set it?

aurianer: The variable is not in the cache anymore so it is only internal, no user will have to set it, and I just renamed it as the environment is called  `env-clang-cuda` and it's the most intuitive name, I think..
But if anyone is opposed to that I can revert the change. I figured it was impacting nobody ^^'

msimberg: Sounds good, no objections.

cmake/HPX_AddComponent.cmake
aurianer: Removing this works when we add the boost workaround to the `CMAKE_CUDA_FLAGS` :ok_hand:

cmake/HPX_AddLibrary.cmake
```diff
@@ -213,26 +213,20 @@ function(add_hpx_library name)
   endif()
 
   # Manage files with .cu extension in case When Cuda Clang is used
-  if(HPX_WITH_CUDA_CLANG OR HPX_WITH_HIP)
+  if(HPX_WITH_CLANG_CUDA OR HPX_WITH_HIP)
     foreach(source ${${name}_SOURCES})
       get_filename_component(extension ${source} EXT)
       if(${extension} STREQUAL ".cu")
+        # TODO: Necessary for clang?
```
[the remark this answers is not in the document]
aurianer: Same works with the boost workaround with `CMAKE_CUDA_FLAGS`

.jenkins/cscs/env-clang-cuda.sh
```diff
@@ -12,23 +12,24 @@ export CXX_STD="17"
...
 spack load ninja@1.10.0
 
 export CXX=`which CC`
 export CC=`which cc`
 
 configure_extra_options="-DCMAKE_BUILD_TYPE=Debug"
 configure_extra_options+=" -DHPX_WITH_CUDA=ON"
-configure_extra_options+=" -DHPX_WITH_CUDA_CLANG=ON"
-configure_extra_options+=" -DHPX_CUDA_CLANG_FLAGS=\"--cuda-gpu-arch=sm_60\""
+configure_extra_options+=" -DCMAKE_CUDA_FLAGS=\"--cuda-gpu-arch=sm_60\""
```
msimberg, line 25: Watch out for the `CMAKE_CUDA_FLAGS` for the Boost workaround. They need to be combined into one.

cmake/HPX_SetupTarget.cmake
```diff
@@ -70,12 +70,12 @@ function(hpx_setup_target target)
...
 
   # Manage files with .cu extension in case When Cuda Clang is used
-  if(target_SOURCES AND HPX_WITH_CUDA_CLANG)
+  if(target_SOURCES AND HPX_WITH_CLANG_CUDA)
     foreach(source ${target_SOURCES})
       get_filename_component(extension ${source} EXT)
       if(${extension} STREQUAL ".cu")
         set_source_files_properties(
-          ${source} PROPERTIES COMPILE_FLAGS "${HPX_CUDA_CLANG_FLAGS}"
+          ${source} PROPERTIES COMPILE_FLAGS "${CMAKE_CUDA_FLAGS}"
```
msimberg, line 78: Isn't the point of `CMAKE_CUDA_FLAGS` that it gets applied automatically to all files that are of type `CUDA`?

tests/performance/local/CMakeLists.txt
```diff
@@ -162,9 +162,10 @@ if(HPX_WITH_CUDA_COMPUTE)
     stream.cpp PROPERTIES CUDA_SOURCE_PROPERTY_FORMAT OBJ
   )
 endif()
-if(HPX_WITH_CUDA_CLANG)
+# TODO: Necessary?
+if(HPX_WITH_CLANG_CUDA)
   set_source_files_properties(
-    stream.cpp PROPERTIES COMPILE_FLAGS "-xcuda ${HPX_CUDA_CLANG_FLAGS}"
+    stream.cpp PROPERTIES COMPILE_FLAGS "-xcuda ${CMAKE_CUDA_FLAGS}"
```
msimberg, line 168: I think the "proper" way to treat this as a cuda file is to set its type to `CUDA`. That should work for both nvcc and clang.

aurianer:
I've encountered a lot of errors with the Intel compiler when switching to C++17 so I'm planning on removing the support for it. If anyone wants to pick it up you are more than welcome!

aurianer:
launch perftests

msimberg:
@aurianer I just pushed the commits for the CUDA builds. Hoping I didn't break the non-CUDA builds too badly...

For others, there were at least a couple of new CUDA bugs revealed by this. The first is neither `std::invoke_result` nor `hpx::util::invoke_result` detect that a noexcept member function is a valid argument for `invoke_result`. CUDA 11.1 fixes this, but I don't know what the underlying bug was. Example: https://godbolt.org/z/7PPf6zKaf. 

The second bug is revealed by our `tag_dispatch` implementation. Specifically, this is triggered by having a templated base class for the CPOs. This is fixed in CUDA 11.2. Example: https://godbolt.org/z/43Mq5eh1E. As a workaround here I've made the nothrow detection always return true for the affected CUDA versions. This may lead to false positives, but I think those are better than false negatives if we make it always return false.

aurianer:
Thanks a lot for working on this!

aurianer:
I just rebased to solve the conflicts

.jenkins/cscs/env-clang-cuda.sh
msimberg: There are still two `CMAKE_CUDA_FLAGS`. One here and one for the boost workaround.

.jenkins/cscs/env-gcc-cuda.sh
```diff
@@ -5,14 +5,14 @@
...
-export CXX_STD="14"
+export CXX_STD="17"
 
 module load daint-gpu
 module switch PrgEnv-cray PrgEnv-gnu
 module load cudatoolkit
 module load Boost/1.75.0-CrayGNU-20.11
 module load hwloc/.2.0.3
-spack load cmake@3.17.3
+spack load cmake@3.18.6 # Necessary to handle cuda_std_17 compiler feature
```
msimberg, line 15: Do we need the comment? Should we not rather just require `3.18`?

libs/parallelism/execution/include/hpx/execution/algorithms/detail/single_result.hpp
```diff
@@ -11,9 +11,7 @@
 
 #include <type_traits>
```
msimberg, line 13:
```suggestion
```

tests/unit/build/CMakeLists.txt
```diff
@@ -49,15 +49,18 @@ function(
     )
   endif()
   set(test_dir "${PROJECT_SOURCE_DIR}/${test_dir}")
+  if(HPX_WITH_CLANG_CUDA)
+    set(cmake_cuda_compiler -DCMAKE_CUDA_COMPILER=$ENV{CXX})
```
msimberg, line 53:
```suggestion
    set(cmake_cuda_compiler -DCMAKE_CUDA_COMPILER=${CMAKE_CUDA_COMPILER})
```
?

`CXX` is not necessarily set.

msimberg (changes requested):

.jenkins/cscs/env-gcc-cuda.sh
aurianer: Was planning to do it but in another PR, cause this was originally just c++17 and then I added the `enable_language(CUDA)` refactoring (as it was necessary), I think doing this in this PR is a bit too much, what do you think?

msimberg: Up to you. It should be done as soon as possible, but a separate PR is also fine by me.

cmake/HPX_AddConfigTest.cmake
```diff
@@ -404,78 +380,6 @@ function(hpx_check_for_cxx17_hardware_destructive_interference_size)
...
-function(hpx_check_for_cxx17_inline_constexpr_variable)
-  add_hpx_config_test(
-    HPX_WITH_CXX17_INLINE_CONSTEXPR_VALUE
-    SOURCE cmake/tests/cxx17_inline_constexpr_variable.cpp
-    FILE ${ARGN}
-  )
-endfunction()
-
-# ##############################################################################
-function(hpx_check_for_cxx17_noexcept_functions_as_nontype_template_arguments)
```
hkaiser, line 462: I'd be a bit cautious with removing this check as I'm not convinced this is properly implemented everywhere.

cmake/HPX_AddConfigTest.cmake
```diff
@@ -503,15 +407,6 @@ function(hpx_check_for_cxx17_shared_ptr_array)
   )
 endfunction()
 
-# ##############################################################################
-function(hpx_check_for_cxx17_std_nontype_template_parameter_auto)
```
hkaiser, line 507: Same here. I know that MSVC is still struggling with this in certain scenarios. This might not be an issue for us, however.

cmake/HPX_SetupCUDA.cmake
hkaiser: No objections.

cmake/tests/builtin_make_integer_seq.cu
```diff
@@ -0,0 +1,34 @@
...
+        "");
+    static_assert(std::is_same<__make_integer_seq<pack_c, std::size_t, 2>,
+                      pack_c<std::size_t, 0, 1>>::value,
+        "");
+}
+
+int main()
+{
+    test();
+}
```
hkaiser, line 34: This is exactly the same as the corresponding .cpp file, isn't it? Could we simply create a file `builtin_make_integer_seq.cu` that does nothing more than `#include "builtin_make_integer_seq.cpp"`?

docs/sphinx/manual/building_hpx.rst
```diff
@@ -520,16 +520,12 @@ those are added to the ``Libs`` sections). In order to use the optional
...
-can use the following CMake variables:
+As |hpx| is a modern C++ library which relies on C++17 by default. The use of
+more recent standards can be opted into explicitly. If you want to force |hpx|
+to use a specific C++ standard version, you can use the following CMake
+variables:
 
-* ``HPX_WITH_CXX14``: Enables C++14 support (this is the minimum requirement)
-* ``HPX_WITH_CXX17``: Enables C++17 support
+* ``HPX_WITH_CXX17``: Enables C++17 support (this is the minimum requirement)
 * ``HPX_WITH_CXX2A``: Enables (experimental) C++20 support
```
hkaiser, line 529: Didn't we deprecate those? Should we change the docs reflecting this?

libs/core/execution_base/include/hpx/execution_base/operation_state.hpp
```diff
@@ -45,8 +45,12 @@ namespace hpx { namespace execution { namespace experimental {
     template <typename O>
     struct is_operation_state;
 
-    HPX_INLINE_CONSTEXPR_VARIABLE struct start_t
-      : hpx::functional::tag_priority_noexcept<start_t>
+#if defined(HPX_COMPUTE_DEVICE_CODE)
+    HPX_DEVICE static const
+#else
+    HPX_INLINE_CONSTEXPR_VARIABLE
+#endif
```
hkaiser, line 52: Should we hide this difference in the `HPX_INLINE_CONSTEXPR_VARIABLE` macro itself?

libs/core/functional/include/hpx/functional/tag_dispatch.hpp
```diff
@@ -184,13 +182,25 @@ namespace hpx { namespace functional {
         };
     }    // namespace detail
 
+    // CUDA versions less than 11.2 have a template instantiation bug which
+    // leaves out certain template arguments and leads to us not being able to
+    // correctly check this condition. We default to the more relaxed
+    // noexcept(true) to not falsely exclude correct overloads. However, this
+    // may lead to noexcept(false) overloads falsely being candidates.
+#if !defined(HPX_CUDA_VERSION) || (HPX_CUDA_VERSION >= 1102)
```
hkaiser, line 190: Can this be checked by a feature test instead of relying on a concrete version?

libs/full/async_cuda/include/hpx/async_cuda/cuda_executor.hpp
```diff
@@ -60,11 +60,9 @@ namespace hpx { namespace cuda { namespace experimental {
         template <typename... Args>
         struct dispatch_helper<cudaError_t, Args...>
         {
-            inline cudaError_t operator()(
-                cudaError_t (*f)(Args...), Args... args)
+            inline void operator()(cudaError_t (*f)(Args...), Args... args)
```
hkaiser, line 63: Leaving the return value doesn't do any harm. Also (unrelated) this `operator()()` can be `const`.

libs/parallelism/execution/include/hpx/execution/algorithms/bulk.hpp
```diff
@@ -115,6 +71,63 @@ namespace hpx { namespace execution { namespace experimental {
...
+                    typename hpx::execution::experimental::sender_traits<
+                        Sender>::template value_types<hpx::tuple, hpx::variant>;
+
+                template <typename... Ts>
+                    auto set_value(Ts&&... ts) &&
+                    noexcept -> decltype(
+                        std::declval<hpx::variant<hpx::monostate, value_type>>()
+                            .template emplace<value_type>(
+                                hpx::make_tuple<>(std::forward<Ts>(ts)...)),
+                        void())
```
hkaiser, line 114: What is the rationale of this change?

libs/parallelism/execution/tests/unit/algorithm_test_utils.hpp
```diff
@@ -132,6 +132,25 @@ struct error_callback_receiver
...
 };
 
+template <typename F>
+struct void_callback_helper
+{
+    std::decay_t<F> f;
+
+    // This overload is only used to satisfy tests that have a predecessor that
+    // can send void, but never does in practice.
+    void operator()()
```
hkaiser, line 142: Can this be `const`?

libs/parallelism/lcos_local/include/hpx/lcos_local/channel.hpp
```diff
@@ -252,7 +252,14 @@ namespace hpx { namespace lcos { namespace local {
                 empty_ = false;
                 push_active_ = false;
             }
-            void set_deferred(T&& val) noexcept
+            void set_deferred(T&& val)
+            // CUDA versions less than 11.1 don't compile push_pt correctly if
+            // this is noexcept. hpx::util::result_of (and std::result_of)
+            // inside deferred_call does not detect that the call is valid, and
+            // compilation fails.
+#if !defined(HPX_CUDA_VERSION) || (HPX_CUDA_VERSION >= 1101)
```
hkaiser, line 260: Could this be done using a feature test (see above)? We could define a `HPX_NOEXCEPT(...)` macro that hides all of this ugliness.

cmake/HPX_AddConfigTest.cmake
aurianer: I've checked this link to verify that it was supported everywhere:
https://en.cppreference.com/w/cpp/compiler_support/17
but I can have a second look

libs/parallelism/execution/include/hpx/execution/algorithms/bulk.hpp
aurianer: Some configurations were not tested with C++17, this is needed to sfinae out set_value for `gcc-oldest`

aurianer:
I've reported the comments in this issue #5463, I will address them as soon as this PR is merged

libs/full/async_cuda/include/hpx/async_cuda/cuda_executor.hpp
msimberg: I think making it void is actually very reasonable. `check_cuda_error` throws if the "error" is not success meaning it would in any case never be returned. On top of that, if I understood correctly, HIP actually complains about the return value being unused since many of the `f`'s are marked `nodiscard`. I'm undecided yet if I like HIP's behaviour there, but it certainly doesn't hurt in the cases we have.

libs/core/functional/include/hpx/functional/tag_dispatch.hpp
msimberg: My personal preference is the explicit CUDA version check, because this is not about a feature, this is a pure bug in nvcc. In addition, the workaround is very specific to `tag_dispatch` and does not need to be applied everywhere we check for noexcept.

libs/parallelism/lcos_local/include/hpx/lcos_local/channel.hpp
msimberg: See my comment about the other noexcept workaround, but I much prefer the explicit workaround here. There's no need to make `noexcept` conditional *everywhere* (even if it's just with certain nvcc versions) just because it doesn't work in this particular case. And if it wouldn't be `HPX_NOEXCEPT` everywhere `noexcept` is currently used I would at least find it confusing if only this member function would use `HPX_NOEXCEPT` and other places would use `noexcept`. This is just my preference though, and I'd be happy to consider the alternative if there's a good case for it.

cmake/HPX_AddConfigTest.cmake
msimberg: I think the main question for this would be: is code that relied on this feature check compiled on the msvc configuration? And does the feature test pass on master for msvc? Same goes for the other feature test below.

hkaiser: Fair enough. Thanks.

libs/core/functional/include/hpx/functional/tag_dispatch.hpp
hkaiser: Ok, fair enough.

libs/full/async_cuda/include/hpx/async_cuda/cuda_executor.hpp
hkaiser: Ok, makes sense. I misread the code.

aurianer:
@hkaiser @ms any hints about those failures of missing executables? Can it be related to the work on the GPU nodes at LSU?
https://cdash.cscs.ch/viewTest.php?onlyfailed&buildid=172230

Note: looks like the last 2 run on the perftests are green so it might have been due to daint in the end, I might increase the number of repetitions to have more stable results.

libs/parallelism/lcos_local/include/hpx/lcos_local/channel.hpp
```diff
@@ -252,7 +252,14 @@ namespace hpx { namespace lcos { namespace local {
                 empty_ = false;
                 push_active_ = false;
             }
-            void set_deferred(T&& val) noexcept
+            void set_deferred(T&& val)
+            // CUDA versions less than 11.3 don't compile push_pt correctly if
+            // this is noexcept. hpx::util::result_of (and std::result_of)
+            // inside deferred_call does not detect that the call is valid, and
+            // compilation fails.
+#if !defined(HPX_CUDA_VERSION) || (HPX_CUDA_VERSION >= 1130)
```
msimberg, line 260:
```suggestion
#if !defined(HPX_CUDA_VERSION) || (HPX_CUDA_VERSION >= 1103)
```
if the version check should indeed be changed.

aurianer: I checked on daint with cuda 11.3 and it's also failing so I'll keep the 1103 workaround instead of 1130

aurianer: Actually it fails on LSU with the workaround of 1103, I'll check more in details which version fail and which doesn't
https://cdash.cscs.ch/viewBuildError.php?buildid=172624

aurianer: The problem ended up being the version number handled as octal values, the problem is reproducible with cuda 11.4 so I extended the workaround for this version. The latest version being 11.4 I'll see if I can find any tickets about this

aurianer:
The failures are unrelated to this PR, I'm merging this :tada:

PR #5289, opened 2021-04-09
Allow splitting of futures holding std::tuple

severinstrobl:
Analogous to the case of std::pair, it should be possible to split a `hpx::future<std::tuple<Ts...>>` into `std::tuple<hpx::future<Ts>...>`.

### DIFF
--- a/libs/parallelism/async_combinators/include/hpx/async_combinators/split_future.hpp
+++ b/libs/parallelism/async_combinators/include/hpx/async_combinators/split_future.hpp
@@ -81,6 +81,7 @@ namespace hpx {
 #include <array>
 #include <cstddef>
 #include <exception>
+#include <tuple>
 #include <type_traits>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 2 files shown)


hkaiser (commented):
Will this work even if `HPX_DATASTRUCTURES_WITH_ADAPT_STD_TUPLE=Off`? Our support for `hpx::get` supporting `std::tuple` relies on having this enabled (see for instance https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/datastructures/include/hpx/datastructures/tuple.hpp#L107-L125)

severinstrobl:
Ah, completely missed that part. So would wrapping the respective sections using `#if defined()`/`#endif` be acceptable or is there a more elegant solution that would somehow allow the conversion between the tuple types (if enabled)?

hkaiser:
> Ah, completely missed that part. So would wrapping the respective sections using `#if defined()`/`#endif` be acceptable or is there a more elegant solution that would somehow allow the conversion between the tuple types (if enabled)?

That would be perfectly acceptable, thanks!

hkaiser (approved):
LGTM, thanks!

hkaiser:
retest

severinstrobl:
As far as I can tell, the failed CI jobs are unrelated to the changes introduced here, but maybe an admin can have a look?

msimberg:
> As far as I can tell, the failed CI jobs are unrelated to the changes introduced here, but maybe an admin can have a look?

This looks good, the remaining failures are unrelated to your changes. Thanks!

PR #5290, opened 2021-04-12
Fixed source file line statements in examples documentation

orgarten:
Updated the source code includes in the examples documentation, because the source files are not exactly the same anymore.
Worst case, source code blocks are empty: [accumulator examples](https://hpx-docs.stellar-group.org/latest/html/examples/accumulator.html)

### DIFF
--- a/docs/sphinx/examples/1d_stencil.rst
+++ b/docs/sphinx/examples/1d_stencil.rst
@@ -45,7 +45,9 @@ The first example is straight serial code. In this code we instantiate a vector
 ``stepper``.
 
 .. literalinclude:: ../../examples/1d_stencil/1d_stencil_1.cpp
-   :lines: 32-75
+   :language: c++
+   :start-after: //[stepper_1
+   :end-before: //]
 
 Each element in the vector of doubles represents a single grid point. To
 calculate the change in heat distribution, the temperature of each grid point,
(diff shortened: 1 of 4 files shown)


msimberg:
Thanks for noticing this! Would you mind updating those to use the tags that are in the examples instead? The `hello_world` section already uses that for example here: https://github.com/STEllAR-GROUP/hpx/blob/master/docs/sphinx/examples/hello_world.rst#walkthrough. I'm referring to the `start-after`/`end-before` part here: 
```
.. literalinclude:: ../../examples/quickstart/hello_world_distributed.cpp
   :language: c++
   :start-after: //[hello_world_action_wrapper
   :end-before: //]
```
With these we don't need to worry about the line numbers being out of date.

orgarten:
I updated the files and also noticed that the 1d_stencil example did not use the tags either and updated this as well. Unfortunately, I cannot build the documentation due to some sphinx errors which I do not have time to debug at the moment. I hope there are not typos within the tags.

msimberg:
> moment. I hope there are not typos within the tags.

It looks ok to me, thanks! The latest versions of breathe do not work with our documentation, which is why we have it pinned to 4.16 in our docker image.

orgarten:
So I just compiled the docs within the docker container and they look fine!

hkaiser (approved):
LGTM, thanks!

PR #5291, opened 2021-04-12
"Fix" termination detection

msimberg:
Fixes (or at least improves) the termination detection used in the occasionally timing-out `standalone_thread_pool_executor` test.

Using the thread-local task counts to decide if it's ok to stop a thread pool is racey since the total count can come out to zero even if there are tasks remaining (if stealing is enabled). This introduces a pool-global thread count that can be used instead to check if there are tasks remaining.

However, there are further problems which is why I'm leaving this as a draft:
- I think the interaction between the new task count and background task count is also racey (but less likely to fail).
- Scheduling tasks from one thread pool to another may lead to similar races when checking if the full runtime can be shutdown.
- MPI/CUDA polling is not taken into account when deciding whether to stop a pool (there should be no remaining events/requests when the pool is fully stopped, but some worker threads may be shut down before all events/requests have been processed).
- A global count increases cache misses (see plot below for timings). Without padding on the counts the hit is quite big (in the future overhead test). With padding the hit reduced quite significantly but the benchmark still runs noticeably slower than with the thread local counts. (`local-thread-count` is current master, `global-thread-count` is after the first commit on this PR, and `global-thread-count-padding` is after the second commit.)

termination_detection

Update:

I changed this to use the approach of checking the counts multiple times suggested by @hkaiser. With the additional requirements of all of the following needing to be non-racy: comparing the thread count to the background thread count, comparing thread counts across pools, checking CUDA and MPI polling work; I gave in and thought this would actually be the most sensible approach. It still allows for false positives to happen, but the rate at which they happen is tunable.

More concretely, I've added:
- `yield_while_count`: like `yield_while` but requires the predicate to return false n times before returning.
- `yield_while_count_timeout`: like the above but with a timeout (used in `runtime_support_server`)
- `wait`, `is_busy`, `is_idle` functions to `threadmanager` and the `thread_pool_base`/`scheduled_thread_pool`; `is_busy` and `is_idle` can return the opposite of the true state with a small probability (as before with comparing thread counts); `wait` uses `yield_while_count` to reduce the false positive probability
- a configuration option `hpx.shutdown_check_count` to adjust how many times `yield_while` should check the predicate before returning

I've removed the `x_thread_count` variants from the `future_overhead` and the libcds equivalent tests since we know that approach is racy.

### DIFF
--- a/libs/core/thread_pools/include/hpx/thread_pools/scheduled_thread_pool_impl.hpp
+++ b/libs/core/thread_pools/include/hpx/thread_pools/scheduled_thread_pool_impl.hpp
@@ -104,6 +104,7 @@ namespace hpx { namespace threads { namespace detail {
       , max_background_threads_(init.max_background_threads_)
       , max_idle_loop_count_(init.max_idle_loop_count_)
       , max_busy_loop_count_(init.max_busy_loop_count_)
+      , shutdown_check_count_(init.shutdown_check_count_)
     {
         sched_->set_parent_pool(this);
     }
@@ -178,6 +179,37 @@ namespace hpx { namespace threads { namespace detail {
     }
 
     template <typename Scheduler>
+    bool scheduled_thread_pool<Scheduler>::is_busy()
+    {
+        // If we are currently on an HPX thread, which runs on the current pool,
+        // we ignore it for the purposes of checking if the pool is busy (i.e.
+        // this returns true only if there is *other* work left on this pool).
+        std::int64_t hpx_thread_offset =
+            (threads::get_self_ptr() && this_thread::get_pool() == this) ? 1 :
+                                                                           0;
+        bool have_hpx_threads =
+            get_thread_count_unknown(std::size_t(-1), false) >
+            sched_->Scheduler::get_background_thread_count() +
+                hpx_thread_offset;
+        bool have_polling_work =
+            sched_->Scheduler::get_polling_work_count() > 0;
+        return have_hpx_threads || have_polling_work;
+    }
+
+    template <typename Scheduler>
+    bool scheduled_thread_pool<Scheduler>::is_idle()
+    {
+        return !is_busy();
+    }
+
+    template <typename Scheduler>
+    void scheduled_thread_pool<Scheduler>::wait()
+    {
+        hpx::util::detail::yield_while_count(
+            [this]() { return is_busy(); }, shutdown_check_count_);
+    }
+
+    template <typename Scheduler>
     template <typename Lock>
     void scheduled_thread_pool<Scheduler>::stop_locked(Lock& l, bool blocking)
     {
@@ -185,6 +217,13 @@ namespace hpx { namespace threads { namespace detail {
 
         if (!threads_.empty())
         {
+            // wait for all work to be done before requesting threads to shut
+            // down
+            if (blocking)
+            {
+                wait();
+            }
+
             // wake up if suspended
             resume_internal(blocking, throws);
 
--- a/examples/quickstart/init_globally.cpp
+++ b/examples/quickstart/init_globally.cpp
@@ -18,8 +18,8 @@
 // library and the main executable might not even be aware of this.
 
 #include <hpx/hpx.hpp>
-#include <hpx/include/run_as.hpp>
 #include <hpx/hpx_start.hpp>
+#include <hpx/include/run_as.hpp>
 
 #include <mutex>
 #include <string>
@@ -40,8 +40,8 @@ void set_argc_argv(int argc, char* argv[], char*[])
     __argv = argv;
 }
 
-__attribute__((section(".preinit_array")))
-    void (*set_global_argc_argv)(int, char*[], char*[]) = &set_argc_argv;
+__attribute__((section(".preinit_array"))) void (*set_global_argc_argv)(
+    int, char*[], char*[]) = &set_argc_argv;
 
 #elif defined(__APPLE__)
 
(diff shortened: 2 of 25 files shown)


hkaiser (commented):
Thanks for investigating this! Having two counters might not be a good idea though (might be racy as you mentioned), would having just one counter do the trick?

msimberg:
> Thanks for investigating this! Having two counters might not be a good idea though (might be racy as you mentioned), would having just one counter do the trick?

Sort of. One counter would indeed work. It's somewhat ugly for a single pool since it would have to ignore background threads, but correct and ugly is better than pretty and racey. There would however have to be a pool-level and runtime-level (shared among pools) counter to handle cross-pool scheduling as well, which is again ugly but doable. I'd really like to avoid the performance hit of changing an atomic counter (twice) for every task though.

hkaiser:
> > Thanks for investigating this! Having two counters might not be a good idea though (might be racy as you mentioned), would having just one counter do the trick?
> 
> Sort of. One counter would indeed work. It's somewhat ugly for a single pool since it would have to ignore background threads, but correct and ugly is better than pretty and racey. There would however have to be a pool-level and runtime-level (shared among pools) counter to handle cross-pool scheduling as well, which is again ugly but doable. I'd really like to avoid the performance hit of changing an atomic counter (twice) for every task though.

The whole issue arises because of -- as you say -- `since the total count can come out to zero even if there are tasks remaining (if stealing is enabled)`. Would there be a way of making sure that the counter will not go to zero instead of trying to find a workaround using global counters? Could you explain how the existing counter could go to zero?

msimberg:
> The whole issue arises because of -- as you say -- `since the total count can come out to zero even if there are tasks remaining (if stealing is enabled)`. Would there be a way of making sure that the counter will not go to zero instead of trying to find a workaround using global counters?

Not sure.

> Could you explain how the existing counter could go to zero?

`get_thread_count` reads and sums up the counters as follows in a simplified 2-thread scenario:
```
count += queue[0].thread_map_count;
count += queue[0].new_tasks_count;
count += queue[1].thread_map_count;
count += queue[1].new_tasks_count;
```
Then the following can happen:
```
# initial state: all counts zero except queue[1].new_tasks_count which is 1 (i.e. there's one staged task on the pool)
# main thread
read queue[0].thread_map_count
# worker thread 0
steal task from thread 1
increment queue[0].thread_map_count
decrement queue[1].new_tasks_count
# main thread
read the remaining counters, which are now all 0
# total count is 0, but queue[0].thread_map_count is 1
```

hkaiser:
Thanks for this explanation. This makes perfect sense. As an alternative solution to the issue, we could go over the counting loop again if the resulting thread-count is zero (possibly do that more than once). This is not fail-proof either, but should reduce the probability of a race significantly. Another option would be to count the overall number of new tasks first before looking at the thread map counts.

hkaiser:
Yet another option could be to have a special (global) counter that is incremented whenever a stealing operation is about to happen and is decremented afterwards.

msimberg:
> Another option would be to count the overall number of new tasks first before looking at the thread map counts.

This might work indeed work, but I'm not sure it accounts for races between the thread count and the background thread count. Another option would maybe also be to keep the thread local counts, and only keep track of how many worker threads are active on the pool by incrementing a pool-wide counter when a thread-local count goes from 0 to 1, and decrementing it when it goes back to zero. The same could be done for the runtime-wide counter. This would reduce the number of modifications on the shared counters quite significantly, but I'm not 100% sure it's not racey either (will likely depend a bit on the implementation).

msimberg:
I've updated the description.

hkaiser:
@msimberg thanks for looking into this again! Could you update the performance graph with your latest solution, please?

msimberg:
> @msimberg thanks for looking into this again! Could you update the performance graph with your latest solution, please?

Yeah, I'll do that! I expect it to be identical to master again, since I've removed the new counter that I initially added, but a sanity check would definitely be good.

msimberg:
This does seem to introduce timeouts to some tests though, so I'll investigate those first before attempting to benchmark the changes.

libs/core/thread_pools/include/hpx/thread_pools/scheduled_thread_pool_impl.hpp
```diff
@@ -177,6 +178,43 @@ namespace hpx { namespace threads { namespace detail {
         return sched_->Scheduler::get_state(num_thread).load();
     }
 
+    template <typename Scheduler>
+    bool scheduled_thread_pool<Scheduler>::is_busy()
+    {
+        // TODO: Take polling events and requests into account.
```
msimberg, line 184: Note to self, this comment is outdated.

msimberg:
This is almost there, but not quite. The init_globally example is broken because it interacts badly with the thread local execution agents. The execution agents are released before the global runtime object, and when the runtime wants to wait for the thread pools to be idle it attempts to yield the thread which fails because the execution agent is already gone.

`id_type_ref_counting_1032_4` and `async_continue_cb_colocated` seem to also time out on almost all builders (although the latter tends to time out on master as well). @hkaiser any guesses about why this would interact badly with those?

hkaiser:
> any guesses about why this would interact badly with those?

@msimberg AFAICT, there is nothing special about those :/

msimberg:
The `init_globally` example was fixed by making the runtime wrapper a `thread_local` and accessing the default agent before constructing the wrapper (there's a tiny bit more to it so please see the `init_globally` file for details). The `id_ref_type_counting_1032` test failure was completely unrelated to id types or refcounting. It was only caused by the sleeps triggering a corner case that was caused by me using `&&` instead of `||` `is_busy`.

Assuming I didn't make something else fail with these changes this should be ready to go. I'll still rerun the future overhead benchmark as the last thing.

msimberg:
Finally, the sanity check that this doesn't affect the future overhead benchmark:
termination_detection_before_after

hkaiser (approved):

PR #5292, opened 2021-04-13
Add async_rw_mutex based on senders

msimberg:
Adds an `async_rw_mutex` (name not finalized, but I think this describes it fairly well). Instead of directly locking a mutex, the async mutex returns senders that give access to a resource when scheduled. The `rw` part means that access can be read-only (shared, multiple concurrent readers allowed), or exclusively read-write.

This is based on an implementation used in DLA-Future where they protect access to tiles of a matrix in this fashion. That mechanism is based on futures/promises and custom for their matrix class. This is an attempt at a more generic version.

Contains #5274 so the diff is rather large.

### DIFF
--- a/libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
+++ b/libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
@@ -1,0 +1,695 @@
+//  Copyright (c) 2021 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/allocator_support/internal_allocator.hpp>
+#include <hpx/assert.hpp>
+#include <hpx/datastructures/optional.hpp>
+#include <hpx/execution_base/receiver.hpp>
+#include <hpx/functional/unique_function.hpp>
+#include <hpx/synchronization/mutex.hpp>
+
+#include <boost/container/small_vector.hpp>
+
+#include <exception>
+#include <memory>
+#include <mutex>
+#include <type_traits>
+#include <utility>
+
+namespace hpx { namespace experimental {
+    namespace detail {
+        enum class async_rw_mutex_access_type
+        {
+            read,
+            readwrite
+        };
+
+        template <typename T>
+        struct async_rw_mutex_shared_state
+        {
+            using shared_state_ptr_type =
+                std::shared_ptr<async_rw_mutex_shared_state>;
+            hpx::util::optional<T> value;
+            shared_state_ptr_type next_state;
+            hpx::lcos::local::mutex mtx;
+            boost::container::small_vector<
+                hpx::util::unique_function_nonser<void(shared_state_ptr_type)>,
+                1>
+                continuations;
+
+            async_rw_mutex_shared_state() = default;
+            async_rw_mutex_shared_state(async_rw_mutex_shared_state&&) = delete;
+            async_rw_mutex_shared_state& operator=(
+                async_rw_mutex_shared_state&&) = delete;
+            async_rw_mutex_shared_state(
+                async_rw_mutex_shared_state const&) = delete;
+            async_rw_mutex_shared_state& operator=(
+                async_rw_mutex_shared_state const&) = delete;
+
+            ~async_rw_mutex_shared_state()
+            {
+                // The last state does not have a next state. If this state has
+                // a next state it must always have continuations.
+                HPX_ASSERT((continuations.empty() && !next_state) ||
+                    (!continuations.empty() && next_state));
+
+                // This state must always have the value set by the time it is
+                // destructed. If there is no next state the value is destructed
+                // with this state.
+                HPX_ASSERT(value);
+
+                if (HPX_LIKELY(next_state))
+                {
+                    // The current state has now finished all accesses to the
+                    // wrapped value, so we move the value to the next state.
+                    next_state->set_value(std::move(value.value()));
+
+                    for (auto& continuation : continuations)
+                    {
+                        continuation(next_state);
+                    }
+                }
+            }
+
+            template <typename U>
+            void set_value(U&& u)
+            {
+                HPX_ASSERT(!value);
+                value.emplace(std::forward<U>(u));
+            }
+
+            void set_next_state(
+                std::shared_ptr<async_rw_mutex_shared_state> state)
+            {
+                // The next state should only be set once
+                HPX_ASSERT(!next_state);
+                next_state = std::move(state);
+            }
+
+            template <typename F>
+            void add_continuation(F&& continuation)
+            {
+                std::lock_guard<hpx::lcos::local::mutex> l(mtx);
+                continuations.emplace_back(std::forward<F>(continuation));
+            }
+        };
+
+        template <>
+        struct async_rw_mutex_shared_state<void>
+        {
+            using shared_state_ptr_type =
+                std::shared_ptr<async_rw_mutex_shared_state>;
+            shared_state_ptr_type next_state;
+            hpx::lcos::local::mutex mtx;
+            boost::container::small_vector<
+                hpx::util::unique_function_nonser<void(shared_state_ptr_type)>,
+                1>
+                continuations;
+
+            async_rw_mutex_shared_state() = default;
+            async_rw_mutex_shared_state(async_rw_mutex_shared_state&&) = delete;
+            async_rw_mutex_shared_state& operator=(
+                async_rw_mutex_shared_state&&) = delete;
+            async_rw_mutex_shared_state(
+                async_rw_mutex_shared_state const&) = delete;
+            async_rw_mutex_shared_state& operator=(
+                async_rw_mutex_shared_state const&) = delete;
+
+            ~async_rw_mutex_shared_state()
+            {
+                // The last state does not have a next state. If this state has
+                // a next state it must always have continuations.
+                HPX_ASSERT((continuations.empty() && !next_state) ||
+                    (!continuations.empty() && next_state));
+
+                for (auto& continuation : continuations)
+                {
+                    continuation(next_state);
+                }
+            }
+
+            void set_next_state(
+                std::shared_ptr<async_rw_mutex_shared_state> state)
+            {
+                // The next state should only be set once
+                HPX_ASSERT(!next_state);
+                next_state = std::move(state);
+            }
+
+            template <typename F>
+            void add_continuation(F&& continuation)
+            {
+                std::lock_guard<hpx::lcos::local::mutex> l(mtx);
+                continuations.emplace_back(std::forward<F>(continuation));
+            }
+        };
+
+        template <typename ReadWriteT, typename ReadT,
+            async_rw_mutex_access_type AccessType>
+        struct async_rw_mutex_access_wrapper;
+
+        template <typename ReadWriteT, typename ReadT>
+        struct async_rw_mutex_access_wrapper<ReadWriteT, ReadT,
+            async_rw_mutex_access_type::read>
+        {
+        private:
+            using shared_state_type =
+                std::shared_ptr<async_rw_mutex_shared_state<ReadWriteT>>;
+            shared_state_type state;
+
+        public:
+            async_rw_mutex_access_wrapper() = delete;
+            async_rw_mutex_access_wrapper(shared_state_type state)
+              : state(std::move(state))
+            {
+            }
+            async_rw_mutex_access_wrapper(
+                async_rw_mutex_access_wrapper&&) = default;
+            async_rw_mutex_access_wrapper& operator=(
+                async_rw_mutex_access_wrapper&&) = default;
+            async_rw_mutex_access_wrapper(
+                async_rw_mutex_access_wrapper const&) = default;
+            async_rw_mutex_access_wrapper& operator=(
+                async_rw_mutex_access_wrapper const&) = default;
+
+            operator ReadT&()
+            {
+                HPX_ASSERT(state);
+                HPX_ASSERT(state->value);
+                return state->value.value();
+            }
+        };
+
+        template <typename ReadWriteT, typename ReadT>
+        struct async_rw_mutex_access_wrapper<ReadWriteT, ReadT,
+            async_rw_mutex_access_type::readwrite>
+        {
+        private:
+            static_assert(!std::is_void<ReadWriteT>::value,
+                "Cannot mix void and non-void type in "
+                "async_rw_mutex_access_wrapper wrapper (ReadWriteT is void, "
+                "ReadT is non-void)");
+            static_assert(!std::is_void<ReadT>::value,
+                "Cannot mix void and non-void type in "
+                "async_rw_mutex_access_wrapper wrapper (ReadT is void, "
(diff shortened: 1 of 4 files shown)


libs/core/type_support/include/hpx/type_support/detail/with_result_of.hpp
```diff
@@ -0,0 +1,45 @@
...
+    // Based on
+    // https://quuxplusone.github.io/blog/2018/05/17/super-elider-round-2/ and
+    // https://akrzemi1.wordpress.com/2018/05/16/rvalues-redefined/.
+    //
+    // Useful for emplacing non-copyable and -movable types into e.g. variants.
+    // The wrapper delays the construction of the non-copyable and -movable type
+    // until it is required for conversion, and guaranteed copy elision in C++17
+    // ensures that a copy constructor is not called when returning from the
+    // conversion operator.
+    template <class F>
```
hkaiser, line 21: nitpick: `typename` ?

libs/core/type_support/include/hpx/type_support/detail/with_result_of.hpp
```diff
@@ -0,0 +1,45 @@
...
+        {
+            return f();
+        }
+    };
+
+    template <class F>
+    inline with_result_of_t<F> with_result_of(F&& f)
+    {
+        return with_result_of_t<F>(std::forward<F>(f));
+    }
```
hkaiser, line 44: Once we're in C++17 land we should replace this with a user defined deduction guideline to invoke CTAD.

libs/parallelism/execution/include/hpx/execution/algorithms/detach.hpp
```diff
@@ -0,0 +1,115 @@
...
+                };
+
+                template <typename... Ts>
+                void set_value(Ts&&...) noexcept
+                {
+                    os.reset();
+                }
+            };
+
+            std::atomic<std::size_t> count{0};
```
hkaiser, line 54: This should be slightly less heavyweight:
```suggestion
            hpx::util::atomic_count count{0};
```

hkaiser: You might want to consider making the `count` private to avoid anybody starts messing with it. This would also require for the `intrusive_ptr_addref/release` functions to become friends of this class.

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
```diff
@@ -0,0 +1,469 @@
...
+                {
+                    HPX_INVOKE(std::move(continuation));
+                }
+            }
+
+            template <typename F>
+            void add_continuation(F&& continuation)
+            {
+                std::lock_guard<hpx::lcos::local::mutex> l(mtx);
+                continuations.emplace_back(std::move(continuation));
```
hkaiser, line 57: Shouldn't this be:
```suggestion
                continuations.emplace_back(std::forward<F>(continuation));
```
?

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
```diff
@@ -0,0 +1,469 @@
...
+        template <typename ReadWriteT, typename ReadT,
+            async_rw_mutex_access_type AccessType>
+        struct async_rw_mutex_access_wrapper;
+
+        template <typename ReadWriteT, typename ReadT>
+        struct async_rw_mutex_access_wrapper<ReadWriteT, ReadT,
+            async_rw_mutex_access_type::read>
+        {
+            std::shared_ptr<ReadWriteT> value;
+            std::shared_ptr<detail::async_rw_mutex_shared_state> state;
```
hkaiser, line 70: Could these two be combined into one `shared_ptr`?

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
```diff
@@ -0,0 +1,469 @@
...
+            static_assert(!std::is_void<ReadWriteT>::value,
+                "Cannot mix void and non-void type in "
+                "async_rw_mutex_access_wrapper wrapper (ReadWriteT is void, "
+                "ReadT is non-void)");
+            static_assert(!std::is_void<ReadT>::value,
+                "Cannot mix void and non-void type in "
+                "async_rw_mutex_access_wrapper wrapper (ReadT is void, "
+                "ReadWriteT is non-void)");
+            std::shared_ptr<ReadWriteT> value;
+            std::shared_ptr<detail::async_rw_mutex_shared_state> state;
```
hkaiser, line 92: Same here, can those be combined?

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
```diff
@@ -0,0 +1,469 @@
...
+            detail::async_rw_mutex_access_wrapper<readwrite_type, read_type,
+                detail::async_rw_mutex_access_type::read>;
+        using readwrite_access_type =
+            detail::async_rw_mutex_access_wrapper<readwrite_type, read_type,
+                detail::async_rw_mutex_access_type::readwrite>;
+
+        using allocator_type = Allocator;
+
+        explicit async_rw_mutex(Allocator const& alloc = {})
+          : alloc(alloc){};
```
hkaiser, line 196: The is `;` superfluous here

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
```diff
@@ -0,0 +1,469 @@
...
+        async_rw_mutex(async_rw_mutex const&) = delete;
+        async_rw_mutex& operator=(async_rw_mutex const&) = delete;
+
+        sender<detail::async_rw_mutex_access_type::read> read()
+        {
+            HPX_ASSERT(state);
+
+            if (prev_access == detail::async_rw_mutex_access_type::readwrite)
+            {
+                prev_state = state;
```
hkaiser, line 208:
```suggestion
                prev_state = std::move(state);
```

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
```diff
@@ -0,0 +1,469 @@
...
+                state =
+                    std::allocate_shared<detail::async_rw_mutex_shared_state,
+                        Allocator>(alloc);
+                prev_access = detail::async_rw_mutex_access_type::read;
```
hkaiser, line 212: Can the disjoint testing and setting of `prev_access` lead to races?

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
```diff
@@ -0,0 +1,469 @@
...
+                            {
+                                hpx::execution::experimental::set_error(
+                                    std::move(r), std::current_exception());
+                            }
+                        });
+
+                    // We release prev_state here to allow continuations to run.
+                    // The operation state may otherwise keep it alive longer
+                    // than needed.
+                    prev_state.reset();
```
hkaiser, line 285: Concurrent invocations of `start` may lead to unreported inconsistencies such that the `HPX_ASSERT` does not fire even if another thread is trying to execute `start` at the same time. I'm not sure if this is a real problem, though.

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
```diff
@@ -0,0 +1,469 @@
...
+
+            template <typename R>
+            struct operation_state
+            {
+                std::decay_t<R> r;
+                std::shared_ptr<detail::async_rw_mutex_shared_state> prev_state;
+                std::shared_ptr<detail::async_rw_mutex_shared_state> state;
+
+                operation_state(operation_state&&) = delete;
+                operation_state(operation_state const&) = delete;
```
hkaiser, line 254: Should the assignment operators be deleted as well?

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
```diff
@@ -0,0 +1,469 @@
...
+        Allocator alloc;
+
+        detail::async_rw_mutex_access_type prev_access =
+            detail::async_rw_mutex_access_type::readwrite;
+
+        std::shared_ptr<detail::async_rw_mutex_shared_state> prev_state;
+        std::shared_ptr<detail::async_rw_mutex_shared_state> state =
+            std::allocate_shared<detail::async_rw_mutex_shared_state>(alloc);
+    };
+}}    // namespace hpx::experimental
```
hkaiser, line 469: General question: all of the types defined here are move-only at best. Would using `unique_ptr` instead of `shared_ptr` be an option? And if yes, do we have to use a pointer type in the first place?

libs/core/type_support/include/hpx/type_support/detail/with_result_of.hpp
msimberg: Yeah, good catch. This was copied from those links.

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
msimberg: Sort of. I wanted to constrain the signature to only take rvalue references, but this also works.

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
msimberg: Not with the current setup unless I'm missing something. The value should live separately from the shared states (which act as triggers to continuations when the reference counts go to zero) and should be kept alive across all read-only and read-write accesses.

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
msimberg: Same answer.

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
msimberg: Yes and no, meaning it's intentionally not threadsafe. It's mentioned in the docstring. Not sure if it's worth making it threadsafe.

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
msimberg: Can you explain how? `start` can only be called once on a (connected) sender.

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
msimberg: Yep!

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
msimberg: Read-only access requires the shared state to be a `shared_ptr`. Even in the read-write case, while only one sender sends the shared state in the wrapper, the following senders need to hold on to the same shared state to attach their continuations as well.

msimberg: That said, I realize now that the value could be in a `unique_ptr` instead of a `shared_ptr`. The reference wrappers can hold references rather than `shared_ptr`s and the last continuation can release the `unique_ptr`. Now I'll have to reexamine if the shared states can do the same (I don't think so).

hkaiser: If using a `unique_ptr<foo>` is possible, wouldn't be a plain `foo` possible as well? This should be viable at least for moveable or trivially copyable value types, shouldn't it?

msimberg: Maybe you see something I don't. I want heap allocation to have the value live past the end of the `async_rw_mutex` destructor (I'd prefer not to block there).

hkaiser: Don't pay attention to what I said, I was merely trying to understand the code base.

msimberg: I will do exactly the opposite, thanks for looking through this!

libs/parallelism/execution/include/hpx/execution/algorithms/detach.hpp
msimberg: Note: I'll apply this in #5274. Is the main difference that `atomic_count` uses more relaxed memory orderings by default?

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
msimberg: I've updated it to use a `unique_ptr` for the guarded value now.

msimberg:
I've removed the dependency on #5274, it was unnecessary.

msimberg:
retest

hkaiser:
@msimberg I still would like to come back to this, mostly to really understand how the facility is supposed to be used and what its semantics are. I still find it to be strange that two separate shared pointers are needed (four allocations!), for instance. My questions are: is the lifetime for those really such different, are those referred to from different places?.

msimberg:
> @msimberg I still would like to come back to this, mostly to really understand how the facility is supposed to be used and what its semantics are.

No problem, I appreciate you looking through this!

> I still find it to be strange that two separate shared pointers are needed (four allocations!), for instance.

Two allocations?

> My questions are: is the lifetime for those really such different, are those referred to from different places?.

There's one allocation for each read-write access, and one shared by consecutive read-only accesses. You can think of the shared states as roughly forming a linked list (where a state refers to its successor through the callbacks which hold hold the actual instances of the successor shared state). Some sort of queue would likely also work for representing the same structure, but also needs heap allocations. A custom allocator is likely also a possibility, but I'd like avoid overcomplicating it at this stage at least.

It might be easiest to discuss this in more detail offline (in a call...).

biddisco (commented):
I didn't look all the way through this yet, but would it not have been possible to use std::shared_mutex as the basic synchronization primitive - with try_lock giving exclusive access and try_shared_lock for shared access. A failure to acquire the lock being the trigger to swap from the current state to the next state. Just wondering out loud.

msimberg:
> A failure to acquire the lock being the trigger to swap from the current state to the next state. Just wondering out loud.

Could you expand on that last part? How would you set things up in that case?

One thing I'm trying to avoid in this setup is to unnecessarily start tasks before they get access to the resource (and thus avoid unnecessary yielding). It's not entirely clear to me if that would happen with the shared mutex approach (in a naive implementation in my head it would do that).

I realize now that I may have left out a critical piece of information, which I hope helps clarify things. The senders should be triggered in the order that they are accessed, not just in any order as the shared mutex would allow. This is a requirement from the use case in DLA-Future where this essentially helps create an implicit DAG of accesses to a tile of a matrix. The name might be worth changing part of the confusion comes from this.

msimberg:
@hkaiser the value is now moved from state to state instead of heap allocated. I'm pretty sure there are other optimizations that could be done here, but I'd like to leave those for the future if/when needed (please do point it out if you see something obviously bad though).

hkaiser (approved):
LGTM, thanks!

hkaiser:
@msimberg thanks for pushing this forward!

PR #5293, opened 2021-04-13
Fix Clang 11 cuda_future test bug

aurianer:
Updating clang cuda config to use Clang 11 resulted in a undefined symbol for `saxpy` in `cuda_future.cpp` unit test, `nm` shows:
```
Clang 11                                                                        
saxpy.cu.o:0000000000000000 T __device_stub__saxpy(int, float, float *, float *)                         
Clang 10                                                                        
saxpy.cu.o:0000000000000000 T saxpy(int, float, float *, float *)
```
Not sure if it's a bug in Clang / Clang cuda or our own cmake config.
The temporary fix is just moving the apply function in the `saxpy.cu` file.

### DIFF
--- a/libs/full/async_cuda/tests/unit/saxpy.cu
+++ b/libs/full/async_cuda/tests/unit/saxpy.cu
@@ -21,11 +21,13 @@ __global__ void saxpy(int n, float a, float* x, float* y)
 void launch_saxpy_kernel(hpx::cuda::experimental::cuda_executor& cudaexec,
         unsigned int& blocks, unsigned int& threads, void** args)
 {
+    // Invoking hpx::apply with cudaLaunchKernel<void> directly result in an
+    // error for NVCC with gcc configuration
 #ifdef HPX_HAVE_HIP
-    hpx::apply(cudaexec, cudaLaunchKernel,
+    auto launch_kernel = cudaLaunchKernel;
 #else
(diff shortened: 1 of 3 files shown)


libs/full/async_cuda/tests/unit/saxpy.cu
```diff
@@ -12,3 +15,15 @@ __global__ void saxpy(int n, float a, float* x, float* y)
     if (i < n)
         y[i] = a * x[i] + y[i];
 }
+
+void launch_saxpy_kernel(hpx::cuda::experimental::cuda_executor& cudaexec,
+        unsigned int& blocks, unsigned int& threads, void** args);
```
msimberg, line 20:
```suggestion
        unsigned int& blocks, unsigned int& threads, void** args)
```

libs/full/async_cuda/tests/unit/saxpy.cu
```diff
@@ -4,6 +4,9 @@
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
+#include <hpx/modules/async_base.hpp>
```
msimberg, line 7:
```suggestion
#include <hpx/local/future.hpp>
```

msimberg:
retest

PR #5300, opened 2021-04-28
Remove PRIVATE flag that was propagated into the LANGUAGES

aurianer:
The `PRIVATE` flag is not parsed by `hpx_add_target_compile_option` and was therefore propagated as a `LANGUAGE`.
The flag is private by default if `PUBLIC` is omitted, so removing it is sufficient.


msimberg (approved):
Nice catch, thanks!

hkaiser (approved):
Excellent catch! 
LGTM, thanks!

PR #5302, opened 2021-04-29
Fix cmake bug, now propagate flag with right language

aurianer:
Issue: `CXX` compile flags were also propagated to `nvcc`, which caused build problems with the new CUDA setup necessary for #5283
Solution: Add the flag with the compile language generator expression even when no config (Release/Debug...) is specified.


msimberg (approved):
Thanks!

biddisco (commented):
LGTM

PR #5303, opened 2021-04-29
split part of hpx_functional into hpx_serializable_function

biddisco:
Split serializable function code into smaller utility library

This removes a dependency of hpx_functional on hpx_serialization
which causes problems if customization point objects are needed
at lower layers of the software stack.

No new functionality or features are introduced and classes
are simply moved from one sub-dir to another to satisfy
compilation etc.

Fixes #
Problem caused by functional depending on serialization so it is not possible to use the CPO code in lower level parts of the library


hkaiser:
Is `function` the only type in the functional module that depends on serialization?

biddisco:
yes. It's really only the serializable_function/vtable stuff that actually pulls in any serialization headers. there are things like bind_front etc that implement hpx::serialization:: overloads, but they're purely template functions and take an archive object that is a template parameter, so they don't actually need a direct dependence on serialization - as long as the code that uses it has included the right stuff.

biddisco:
NB. the `hpx_functional` lib could almost be called `hpx_invoke`...

libs/core/serializable_function/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+    hpx/serializable_function/serialization/detail/vtable/serializable_vtable.hpp
+    hpx/serializable_function/serialization/serializable_function.hpp
+    hpx/serializable_function/serialization/serializable_unique_function.hpp
+)
+
+# Default location is $HPX_ROOT/libs/serializable_function/include_compatibility
+# cmake-format: off
+set(serializable_function_compat_headers
+    hpx/util/serializable_function.hpp => hpx/serializable_function.hpp
+    hpx/util/serializable_unique_function.hpp => hpx/serializable_function.hpp
```
msimberg, line 22: `hpx/serializable_function.hpp` doesn't actually exist, does it? `GLOBAL_HEADER_GEN` generates `hpx/modules/serializable_function.hpp`. So far `hpx/functional.hpp` only included the non-serializable `hpx::function`. Should it include both?

libs/core/serializable_function/README.rst
```diff
@@ -0,0 +1,16 @@
...
+..
+    Copyright (c) 2019 The STE||AR-Group
+
+    SPDX-License-Identifier: BSL-1.0
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+===
+serializable_function
+===
```
msimberg, line 11:
```suggestion
=====================
serializable_function
=====================
```

libs/core/serializable_function/docs/index.rst
```diff
@@ -0,0 +1,31 @@
...
+
+    SPDX-License-Identifier: BSL-1.0
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+.. _modules_serializable_function:
+
+==========
+serializable_function
+==========
```
msimberg, line 12:
```suggestion
=====================
serializable_function
=====================
```

libs/full/async_distributed/CMakeLists.txt
```diff
@@ -120,7 +120,8 @@ add_hpx_module(
   HEADERS ${async_headers}
   COMPAT_HEADERS ${async_compat_headers}
   DEPENDENCIES hpx_core hpx_parallelism
-  MODULE_DEPENDENCIES hpx_actions hpx_naming hpx_runtime_configuration
-                      hpx_runtime_local hpx_threadmanager
+  MODULE_DEPENDENCIES
+    hpx_actions hpx_naming hpx_runtime_configuration hpx_runtime_local
+    hpx_serializable_function hpx_threadmanager
```
msimberg, line 125: `hpx_serializable_function` does not need to (should not) be here. It comes indirectly through `hpx_core`, the shared library which `hpx_serializable_function` is part of.

libs/core/serializable_function/CMakeLists.txt
biddisco: I just copied and pasted the compat headers stuff from functional and changed the name accordingly.  I guess we can just get rid of this ass it's not really needed for the new module.

libs/core/serializable_function/docs/index.rst
```diff
@@ -0,0 +1,31 @@
...
+* :cpp:func:`hpx::util::bind`
+* :cpp:func:`hpx::util::bind_back`
+* :cpp:func:`hpx::util::bind_front`
+* :cpp:func:`hpx::util::deferred_call`
+* :cpp:func:`hpx::util::invoke`
+* :cpp:func:`hpx::util::invoke_fused`
+* :cpp:func:`hpx::util::mem_fn`
+* :cpp:func:`hpx::util::one_shot`
+* :cpp:func:`hpx::util::protect`
+* :cpp:class:`hpx::util::result_of`
```
hkaiser, line 28: A lot of those are not in this new module.

hkaiser:
I think we can close this once #5464 has been merged. @biddisco please try that PR to see if it helps solving the issue at hand.

biddisco:
Thanks @hkaiser #5464 solves my circular dependencies so I am closing this PR.

Issue #5162, opened 2021-02-10
Require C++17 support


msimberg:
We should strongly consider requiring C++17 support from compilers. The blocker is as usual `nvcc` although newer versions claim to support it.

msimberg:
We now have CUDA 11.0.2 on Piz Daint (rostam has had CUDA 11 for a while already), which in theory supports C++17. Assuming it actually supports C++17 well enough this could become a reality soon. Let's see how it goes...

msimberg:
Closed by #5283.

PR #5304, opened 2021-04-30
Fix compilation of sender CPOs with nvcc

msimberg:
Same problem as https://github.com/STEllAR-GROUP/hpx/pull/4912. I've set a reminder for myself to come back and add an actual regression test for this.


hkaiser (approved):
LGTM, thanks!

aurianer (approved):
LGTM thanks!

msimberg:
I went back to add the regression test, and while doing so I realized this is a problem up to and including CUDA 10.0, fixed in 10.1 onwards. After #5162 we'll be requiring CUDA 11 anyway, so I won't be adding the regression test.

PR #5310, opened 2021-05-02
Adding hpx::execution::experimental::task_group

hkaiser:
- this is similar to tbb:task_group

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/task_group.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/task_group.hpp
@@ -20,6 +20,7 @@
 #include <hpx/functional/invoke_fused.hpp>
 #include <hpx/synchronization/latch.hpp>
 
+#include <exception>
 #include <stdexcept>
 #include <type_traits>
 #include <utility>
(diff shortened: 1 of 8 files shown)


libs/parallelism/algorithms/include/hpx/parallel/task_group.hpp
```diff
@@ -0,0 +1,110 @@
...
+                [this, f = std::forward<F>(f),
+                    t = hpx::forward_as_tuple(
+                        std::forward<Ts>(ts)...)]() mutable {
+                    try
+                    {
+                        hpx::util::invoke_fused(std::move(f), std::move(t));
+                    }
+                    catch (...)
+                    {
+                        errors_.add(std::current_exception());
```
msimberg, line 65: This needs to do something like https://github.com/STEllAR-GROUP/hpx/blob/87c0d769cd627a70c1e1d81a63b408b68a952c26/libs/parallelism/futures/include/hpx/futures/detail/future_data.hpp#L531-L548 (because of the lock inside `exception_list::add`) or ideally use https://github.com/STEllAR-GROUP/hpx/blob/93714c3e77bdcccc0097a096631c7280cfe6bd31/libs/core/execution_base/include/hpx/execution_base/detail/try_catch_exception_ptr.hpp from #5274.

libs/parallelism/algorithms/include/hpx/parallel/task_group.hpp
```diff
@@ -0,0 +1,160 @@
...
+                    }
+
+                    // The exception is set outside the catch block since
+                    // set_exception may yield. Ending the catch block on a
+                    // different worker thread than where it was started may
+                    // lead to segfaults.
+                    add_exception(std::move(p));
+                });
+
+            l.latch_ = nullptr;
```
msimberg, line 121: Very minor, but is this needed?

msimberg (approved):
Looks good to me! I'd still like to see the rostam builds succeed though.

msimberg:
retest lsu

libs/parallelism/algorithms/include/hpx/parallel/task_group.hpp
hkaiser: Good catch, it's a left-over.

PR #5225, opened 2021-03-07
Adapt reverse to C++20

gonidelis:

PR #5312, opened 2021-05-03
Use local runtime and local headers in local-only modules and tests

msimberg:
- Adds a `include_local` module with `hpx/local/x.hpp` headers (even for cases where `hpx/x.hpp` is local-only. Having the local counterparts always available leaves out guesswork about when the local counterparts are available. This adds the local counterparts to what I needed for these changes, but I will separately go through them and see if any are missing (including updating docs).
- Uses the local runtime init (from `hpx/local/init.hpp`) in all of core and parallelism, and some of full.
- Uses the headers from `include_local` in all of core and parallelism, and some of full.
- Removes the HPX runtime completely from some tests that did not require it (e.g. in the cache module)
- Adds `NOLIBS` to all local-only module header tests.
- Flyby: reduce the duration of some sleeps in tests that were unnecessarily long.


aurianer (dismissed):
Whaouh thanks a lot! LGTM!

hkaiser (commented):
Just a question: initially we planned to have headers `<hpx/foo.hpp>` for each corresponding header from the standard library `<foo>` (where appropriate). Does this PR propose to introduce `<hpx/local/foo.hpp>` for those instead.?

hkaiser (commented):
Wouldn't the use of the `<hpx/local/...>` headers from inside the library itself (not the tests) create circular dependencies involving the `local` module?

msimberg:
> Just a question: initially we planned to have headers `<hpx/foo.hpp>` for each corresponding header from the standard library `<foo>` (where appropriate). Does this PR propose to introduce `<hpx/local/foo.hpp>` for those instead?

We've more or less started both of these (they are in the include module where headers from multiple modules need to be pulled in, otherwise in the corresponding modules). Since we decided that the hpx/foo.hpp headers should pull in both local and distributed headers I've added more of the local variants here.


> Wouldn't the use of the `<hpx/local/...>` headers from inside the library itself (not the tests) create circular dependencies involving the `local` module?

I've only changed them in the tests. Sorry if that was unclear from the description. They would definitely create circular dependencies.

hkaiser:
> > Wouldn't the use of the `<hpx/local/...>` headers from inside the library itself (not the tests) create circular dependencies involving the `local` module?
> 
> I've only changed them in the tests. Sorry if that was unclear from the description. They would definitely create circular dependencies.

I was asking because of this: https://github.com/STEllAR-GROUP/hpx/pull/5312/files#diff-3e6a30579010722efc976788e77b5e2a2208275784257b42a412f3cee9ccd4ddR9. That might be an oversight, however.

hkaiser:
> > Just a question: initially we planned to have headers `<hpx/foo.hpp>` for each corresponding header from the standard library `<foo>` (where appropriate). Does this PR propose to introduce `<hpx/local/foo.hpp>` for those instead?
> 
> We've more or less started both of these (they are in the include module where headers from multiple modules need to be pulled in, otherwise in the corresponding modules). Since we decided that the hpx/foo.hpp headers should pull in both local and distributed headers I've added more of the local variants here.

Makes sense, thanks.

msimberg:
I updated a few more tests to not unnecessarily start the runtime. If this comes back clean from CI it should be ready to go.

hkaiser:
> I updated a few more tests to not unnecessarily start the runtime. If this comes back clean from CI it should be ready to go.

Thanks for pushing this, Mikael! I think this is the right next step at this point.

msimberg:
Rebased on master after the changes from #5225. Hopefully everything is still ok...

msimberg:
retest lsu

PR #5313, opened 2021-05-03
Rearrange modules between libraries

msimberg:
Follow-up to #5312. 

Moves all local-only functionality to the core library (including the local runtime), and thus removes the parallelism library. With this hpx_core contains practically all local functionality (I've now included algorithms, async_cuda, and async_mpi as well), and hpx(_full) all distributed functionality. If noone has objections I'd rename hpx_core to hpx_local and document the target in the documentation. I would still personally like to see hpx_local/core in a separate repository to give it visibility as a separate project, but we'll have continue that discussion in person...

I've taken the liberty of removing the (incomplete) functionality to associate a locality with a CUDA (compute, not CMake) target (and with it the target distribution policy for CUDA targets). If someone really wants to see that survive they can speak up and take over maintenance.

Opinions?


hkaiser:
> I've taken the liberty of removing the (incomplete) functionality to associate a locality with a CUDA (compute, not CMake) target (and with it the target distribution policy for CUDA targets). If someone really wants to see that survive they can speak up and take over maintenance.

+1

msimberg:
This is now rebased and passing CI again.

As already mentioned in the description, this PR splits the library into two (from previously three somewhat arbitrarily chosen parts). core contains local functionality (and can be renamed), full contains distributed functionality. The split currently looks like this:

core:

- affinity
- algorithms
- allocator_support
- asio
- assertion
- async_base
- async_combinators
- async_cuda
- async_local
- async_mpi
- batch_environments
- cache
- command_line_handling_local
- concepts
- concurrency
- config
- config_registry
- coroutines
- datastructures
- debugging
- errors
- execution
- execution_base
- executors
- filesystem
- format
- functional
- futures
- hardware
- hashing
- include_local
- ini
- init_runtime_local
- io_service
- iterator_support
- itt_notify
- lcos_local
- logging
- memory
- mpi_base
- pack_traversal
- plugin
- prefix
- preprocessor
- program_options
- properties
- resiliency
- resource_partitioner
- runtime_configuration
- runtime_local
- schedulers
- serialization
- static_reinit
- string_util
- synchronization
- tag_dispatch
- testing
- thread_pool_util
- thread_pools
- thread_support
- threading
- threading_base
- threadmanager
- timed_execution
- timing
- topology
- type_support
- util
- version

full:

- actions
- actions_base
- agas
- agas_base
- async_colocated
- async_distributed
- checkpoint
- checkpoint_base
- collectives
- command_line_handling
- components
- components_base
- compute
- compute_cuda
- distribution_policies
- executors_distributed
- include
- init_runtime
- lcos_distributed
- naming
- naming_base
- performance_counters
- resiliency_distributed
- runtime_components
- runtime_distributed
- segmented_algorithms
- statistics

I don't know if there's a preference for how to proceed with this. I'd be ok with making this the initial split, and then further refining the local part. On the other hand leaving it for later may mean never doing it. If we were to split the local part into further libraries I see a few possible candidates:

- algorithms: simply the algorithms module on its own
- local runtime: the actual runtimey components, i.e. `init_runtime_local`, `threadmanager`, `resource_partitioner`, `runtime_config`, `command_line_handling_local`, `runtime_local` and maybe a few more
- everything `execution`, i.e. futures, senders, synchronization primitives
- thread pools and coroutines
- base utilities: everything that doesn't fit elsewhere

I'm only realistically considering algorithms and the local runtime as separate libraries, with the thread pools a possible candidate for a separate library. I consider everything `execution` (without the actual thread pools and runtime) so fundamental to HPX that I don't see much value in separating that from base utilities. I still consider `async_cuda/mpi` libraries that should eventually move out, but that requires changes in how the scheduler registers sets up polling so it can't be done that easily right now.

Are there other potential libraries? Opinions?

msimberg:
I take the silence here as acceptance. I'd like to move forward with this. There's still room to shift modules around after this is merged, but I'd like to start playing around with the setup of two separate repositories.

PR #5314, opened 2021-05-03
Fix a race condition in async mpi affecting limiting executor

biddisco:
The mpi_ring_async_executor_test occasionally failed because
the limiting executor was calling find_if on a vector that
was modified by the mpi polling thread.

Instead of allowing external checks (limiting executor) to
access the vector, we use an atomic counter that only the mpi
polling thread increments/decrements and stores the number
of futures/requests in flight. The limiting executor can safely
read this value from any thread.

### DIFF
--- a/libs/full/async_mpi/include/hpx/async_mpi/mpi_future.hpp
+++ b/libs/full/async_mpi/include/hpx/async_mpi/mpi_future.hpp
@@ -80,6 +80,17 @@ namespace hpx { namespace mpi { namespace experimental {
             bool error_handler_initialized_ = false;
             int rank_ = -1;
             int size_ = -1;
+            // active futures holds the size of the future vector
+            // which is always the same as the mpi request vector
+            std::atomic<std::uint32_t> active_futures_size_;
+            // active requests returns the size of the request queue
+            std::atomic<std::uint32_t> request_queue_size_;
+            //
+            mpi_info()
+              : active_futures_size_{0}
+              , request_queue_size_{0}
+            {
+            }
         };
 
         // an instance of mpi_info that we store data in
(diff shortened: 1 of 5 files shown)


libs/full/async_mpi/include/hpx/async_mpi/mpi_future.hpp
```diff
@@ -80,6 +80,17 @@ namespace hpx { namespace mpi { namespace experimental {
             bool error_handler_initialized_ = false;
             int rank_ = -1;
             int size_ = -1;
+            // active futures holds the size of the future vector
+            // which is always the same as the mpi request vector
+            std::atomic<std::uint32_t> active_futures_size_;
+            // active requests returns the size of the request queue
+            std::atomic<std::uint32_t> request_queue_size_;
```
msimberg, line 87:
```suggestion
            std::atomic<std::uint32_t> active_futures_size_{0};
            // active requests returns the size of the request queue
            std::atomic<std::uint32_t> request_queue_size_{0};
```
and `mpi_info() = default;` (or remove it)?

libs/full/async_mpi/include/hpx/async_mpi/mpi_future.hpp
```diff
@@ -126,7 +137,7 @@ namespace hpx { namespace mpi { namespace experimental {
         // too many mpi requests are being spawned at once
         // unfortunately, the lock-free queue can only return an estimate
         // of the queue size, so this is not guaranteed to be precise
-        HPX_EXPORT std::size_t get_number_of_enqueued_requests();
+        //        HPX_EXPORT std::size_t get_number_of_enqueued_requests();
```
msimberg, line 140: Still needed?

msimberg:
retest lsu

biddisco:
Should be good to merge now I think.

msimberg:
@biddisco thanks for updating this. I just pushed a commit to add some missing headers reported by inspect. This should be good to go now.

PR #5316, opened 2021-05-05
Attempt to fix checking for libatomic

msimberg:
In cases where libatomic does not actually exist we still attempted to link with libatomic later for 128-bit atomics. This tries to fix that.

@G-071 would you mind trying this?


G-071:
This solves the -latomic issue I have encountered using hipcc. 
Thanks!

aurianer (approved):
LGTM thanks!

hkaiser (approved):
LGTM, thanks!

PR #5319, opened 2021-05-06
gcc 11.1 c++20 build fixes

srinivasyadav18:
## Proposed Changes

  - Added c++20 [`Parenthesized initialization of aggregates`](https://wg21.link/P0960R3) cmake feature test.
  - Fixed traits in brace initialization test with macros.
  - Added c++20 [`Allow lambda-capture [=, this]` ](https://wg21.link/P0409R2)cmake feature test.
  - Added `[=, this]` capture at various places with macros.


msimberg:
retest

components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_impl.hpp
```diff
@@ -102,10 +102,17 @@ namespace hpx
         typedef typename components::server::distributed_metadata_base<
             server::partitioned_vector_config_data>::get_action act;
 
+#if !defined(HPX_HAVE_CXX20_STD_LAMBDA_CAPTURE)
         return async(act(), id).then(
             [=](future<server::partitioned_vector_config_data>&& f) -> void {
```
msimberg, line 107:
```suggestion
            [= HPX_CXX20_CAPTURE_THIS](future<server::partitioned_vector_config_data>&& f) -> void {
```
? Syntactically somewhat awkward but maybe nicer than the ifdefs? Alternatively I would place the if/else inside the capture list.

srinivasyadav18: @msimberg or else can i do this instead of if/else, i will use `HPX_CXX20_CAPTURE_THIS` as you suggested, but
define HPX_CXX20_CAPTURE_THIS as `, this` when `HPX_HAVE_CXX20_STD_LAMBDA_CAPTURE` is defined else to ` ` 
 empty space.

So instead of if/defs i just write  `HPX_CXX20_CAPTURE_THIS` in the captures as you shown above?

some thing like [this ](https://wandbox.org/nojs/gcc-head/permlink/ElKxIP2JLRaBM5tn).

msimberg: Yeah, that's exactly what I had in mind. @hkaiser any better ideas?

hkaiser: I think having the ',' as part of the macro is not good. How about we use it as:
```
[HPX_CXX20_CAPTURE_THIS(=)](future<server::partitioned_vector_config_data>&& f) -> void {...}
```
where the argument list could be potentially empty. I.e.
```
#if HPX_HAVE_CXX20_STD_LAMBDA_CAPTURE
#define HPX_CXX20_CAPTURE_THIS(...) __VA_ARGS__, this
#else
#define HPX_CXX20_CAPTURE_THIS(...) __VA_ARGS__
#endif
```

srinivasyadav18: should it be like this?
```
HPX_PP_IF(HPX_PP_NARGS(__VA_ARGS__) == 0, \
        this, HPX_PP_IDENTITY(__VA_ARGS__, this))
```
assuming that `HPX_PP_IF` works like ternary `HPX_PP_IF(CONDITION, TRUE, FALSE)`

And `HPX_PP_IDENTITY` is bit simpler to implement, as
`#define HPX_PP_IDENTITY __VA_ARGS__`

but  any clue how `HPX_PP_IF` would be implemented?
because we cannot do something like this `#define SOME_MACRO #if ...`

msimberg: Shouldn't that be `HPX_PP_IF(HPX_PP_NARGS(__VA_ARGS__) == 0, this, HPX_PP_IDENTITY(__VA_ARGS__, this))`? And in the else branch `[HPX_CXX20_CAPTURE_THIS()]` would not be right. I have the feeling any approach is going to have inputs that are just plain wrong, so I guess we'd just have to pick from the least bad one. The reason I like my suggestion is that it's anyway only meant to be used when there's a `=` capture and it would make it painfully obvious that `[HPX_CXX20_CAPTURE_THIS]` is not valid. But I admit it's far from perfect as well. In the end I don't have a particularly strong opinion about this.

srinivasyadav18: this `HPX_CXX20_CAPTURE_THIS(...)`,
expects at least one macro argument right? so whether `HPX_CXX20_CAPTURE_THIS` is either a macro function or plain macro they both would work only if there is more than one argument?

If yes, then the solution could be one of the following?

```
#define HPX_PP_IDENTITY(...) __VA_ARGS__

#if HPX_HAVE_CXX20_STD_LAMBDA_CAPTURE
#define HPX_CXX20_CAPTURE_THIS(...) \
    HPX_PP_IDENTITY(__VA_ARGS__, this)
#else
#define HPX_CXX20_CAPTURE_THIS(...) __VA_ARGS__
#endif
```

or

`#define HPX_CXX20_CAPTURE_THIS , this`

hkaiser: Yes, your suggestion of using identity is even simpler, nice. I would like to avoid having the `','` as part of the macro as this leads to 'invalidly' looking syntax: `[= HPX_CXX20_CAPTURE_THIS]`.

hkaiser: > Shouldn't that be `HPX_PP_IF(HPX_PP_NARGS(__VA_ARGS__) == 0, this, HPX_PP_IDENTITY(__VA_ARGS__, this))`?

Yes, my bad, that's what I meant.

> And in the else branch `[HPX_CXX20_CAPTURE_THIS()]` would not be right.

Yes to this as well, However, I would expect for this to be used as `[HPX_CXX20_CAPTURE_THIS(=)]` only (with possibly more arguments).

srinivasyadav18: @hkaiser  thanks for this!

hkaiser:
retest

srinivasyadav18:
can anyone please `retest` this?

aurianer:
retest

hkaiser:
retest

libs/core/config/include/hpx/config/lambda_capture_this.hpp
```diff
@@ -0,0 +1,16 @@
...
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/config/defines.hpp>
+
+#if defined(HPX_HAVE_CXX20_STD_LAMBDA_CAPTURE)
+#define HPX_PP_IDENTITY(...) __VA_ARGS__
```
hkaiser, line 12: I know it's a trivial macro, however - could we have the definition of `HPX_PP_IDENTITY()` in a separate header file in the preprocessor module?

srinivasyadav18: sure, i will change it soon.

msimberg:
retest

srinivasyadav18:
@msimberg @hkaiser I think now all bugs with gcc 11.1 (with c++20) are fixed. Please let me know if any further changes are required other than formatting ;-). If everthing is okay, i will apply formatting and push back the changes.

libs/core/config/include/hpx/config/lambda_capture_this.hpp
```diff
@@ -0,0 +1,16 @@
...
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/config/defines.hpp>
+#include <hpx/preprocessor/identity.hpp>
+
+#if defined(HPX_HAVE_CXX20_STD_LAMBDA_CAPTURE)
+#define HPX_CXX20_CAPTURE_THIS(...) HPX_PP_IDENTITY(__VA_ARGS__, this)
```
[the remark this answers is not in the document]
srinivasyadav18: @msimberg yes, that is actually enough.
So, do you suggest changing it to `__VA_ARGS__, this` ?.
If yes, shall I remove only use of `HPX_PP_IDENDTITY` or even the implementation of it in preprocessor module i.e this [file](https://github.com/STEllAR-GROUP/hpx/blob/d41b84d6d4fb0129fb181cbabfc573f77b8dd88f/libs/core/preprocessor/include/hpx/preprocessor/identity.hpp) as well?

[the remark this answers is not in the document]
hkaiser: Wouldn't just writing `__VA_ARGS__, this` leave the comma in place if `__VA_ARGS__` is empty (no arguments)?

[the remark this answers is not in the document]
msimberg: Good point. It probably does that. Then the question is if `HPX_CXX20_CAPTURE_THIS()` should be an error or not? It would be an error with C++17 and lower with either implementation. With C++20 using `HPX_PP_IDENTITY(__VA_ARGS__, this)` would compile. With C++20 using only `__VA_ARGS__, this` would be an error.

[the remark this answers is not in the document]
hkaiser: When using the identity macro
- a plain `HPX_CXX20_CAPTURE_THIS()` would expand to a simple `this`, which is fine for any C++ version
- `HPX_CXX20_CAPTURE_THIS(=)` would expand to `=, this` for C++20 and `=` for anything below - which is correct as well
- `HPX_CXX20_CAPTURE_THIS(something)` would expand to `something, this` for C++20 and just `something` everywhere else. 
 
The latter should be fine if `something` doesn't have `this` in it too. I think this is the best we can do.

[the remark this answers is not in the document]
msimberg: Given that the macro is called `HPX_CXX20_CAPTURE_THIS` I think it's definitely problematic if `this` is not captured. However, I also agree with you that there's no perfect solution for this. Enough bikeshedding ;) Let's keep this definition as it is. Our various CI configurations will catch at least catch the trivial cases anyway.

cmake/tests/cxx20_std_lambda_capture.cpp
```diff
@@ -0,0 +1,20 @@
+//  Copyright (c) 2021 Srinivas Yadav
```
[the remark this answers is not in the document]
srinivasyadav18: @msimberg I could not find any solution for gcc to enable the warning as error. So please let me know if you have any solution or fix for gcc.
Thank you.

cmake/HPX_AddConfigTest.cmake
```diff
@@ -118,16 +118,17 @@ function(add_hpx_config_test variable)
 
     if(${variable}_EXECUTE)
       if(NOT CMAKE_CROSSCOMPILING)
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
```
hkaiser, line 121: This will not work for MSVC, there the flag is `-WX`.

hkaiser (approved):
LGTM, thanks!

cmake/tests/cxx20_lambda_capture.cpp
```diff
@@ -0,0 +1,20 @@
...
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// test for availability of Allow lambda-capture [=, this] (C++20)
+
+struct S
+{
+    void f()
+    {
+        [=, this]{};
```
msimberg, line 13: @srinivasyadav18 please make sure that the test compiles without any other warnings as well. The test seems to be failing even with C++20 because this lambda is unused. Something like this should silence the warning, while keeping the warning related to the capture of `this`: `(void)[=, this]{}`.

srinivasyadav18: @msimberg I have made the changes, but can you please share the link where it is failing, it will be more clear for me. :).
Thank You.

msimberg: This is on the clang-newest configuration (which is clang 11 with C++20): https://cdash.cscs.ch/viewConfigure.php?buildid=159985. Note the failing `HPX_WITH_CXX20_LAMBDA_CAPTURE` test. Interestingly with gcc the test does actually pass, so I suppose it for some reason does not consider that an unused variable (at least with the default warnings).

msimberg:
retest

srinivasyadav18:
@msimberg do this check `jenkins/cscs/clang-oldest` fail because of changes in this PR or everything is fine here?

msimberg:
@srinivasyadav18 it looks unrelated to this PR. Thanks!

PR #5320, opened 2021-05-07
Fix is_sender trait and other small fixes to p0443 traits

msimberg:
Mainly fixes the `is_sender` trait. It was returning true for things like `int`. The reason was a missing constraint in the fallback which was meant to apply for executors. It was treating `int` as an executor. Purely my mistake, but in the interest of simplicity I've removed the fallback completely. In my understanding there's already been quite some discussion about how executors should not be treated as schedulers, and especially not as senders, implicitly.

I've also applied small fixes to the `is_executor` trait and the schedule CPO (which was using `is_executor` rather than `is_scheduler` as a constraint).

### DIFF
--- a/libs/core/execution_base/include/hpx/execution_base/sender.hpp
+++ b/libs/core/execution_base/include/hpx/execution_base/sender.hpp
@@ -341,8 +341,7 @@ namespace hpx { namespace execution { namespace experimental {
         template <typename Executor, typename F,
             typename =
                 std::enable_if_t<hpx::is_invocable<std::decay_t<F>&>::value &&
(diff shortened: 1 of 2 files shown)


hkaiser (dismissed):
LGTM, thanks!

aurianer (dismissed):
LGTM thanks!

msimberg:
I kept the implicit use of executors as schedulers etc. for now since I had to dig into some compilation failures on gcc 7 and found another workaround. However, I might still remove the executor-as-scheduler stuff at a later point. It's still unnecessarily complicated.

PR #5323, opened 2021-05-10
Add keep_future algorithm

msimberg:
This solves the problem of trying to use a `shared_future<noncopyable_type>` with e.g. `on` or `sync_wait` which fails because the algorithms have to copy the value internally (which obviously does not work with noncopyable types). This adds a wrapper algorithm `keep_future` (other name suggestions?) which calls `set_value` on receivers connected to futures with the future instead of the contained value.

Note: This could very well be a better default than the current implementation of futures as senders. There's always `unwrapping` for dealing with futures in continuations.

Builds on #5274 since it requires some fixes to `on`. Only the last commit is new.


hkaiser (approved):
Let's go ahead with this, LGTM. Thanks!

msimberg:
> Let's go ahead with this, LGTM. Thanks!

Note that this includes #5274. No further comments on that?

PR #5324, opened 2021-05-10
Add get member functions to async_rw_mutex proxy objects for explicitly getting the wrapped value

msimberg:
Just a little convenience for the cases when one doesn't want to/can't use the implicit conversion.

Edit: Flyby: initialize the `value` member in `async_rw_mutex` in the member initializer list instead of in the constructor body, to have it work for non-default constructible types.

### DIFF
--- a/libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
+++ b/libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
@@ -177,11 +177,16 @@ namespace hpx { namespace experimental {
             async_rw_mutex_access_wrapper& operator=(
                 async_rw_mutex_access_wrapper const&) = default;
 
-            operator ReadT&()
+            ReadT& get() const
             {
                 HPX_ASSERT(state);
                 HPX_ASSERT(state->value);
                 return state->value.value();
+            }
+
+            operator ReadT&() const
+            {
+                return get();
             }
         };
 
@@ -218,11 +223,16 @@ namespace hpx { namespace experimental {
             async_rw_mutex_access_wrapper& operator=(
                 async_rw_mutex_access_wrapper const&) = default;
 
-            operator ReadWriteT&()
+            ReadWriteT& get()
             {
                 HPX_ASSERT(state);
                 HPX_ASSERT(state->value);
                 return state->value.value();
+            }
+
+            operator ReadWriteT&()
+            {
+                return get();
             }
         };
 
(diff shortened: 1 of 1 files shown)


libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
```diff
@@ -177,12 +177,17 @@ namespace hpx { namespace experimental {
             async_rw_mutex_access_wrapper& operator=(
                 async_rw_mutex_access_wrapper const&) = default;
 
-            operator ReadT&()
+            ReadT& get()
```
hkaiser, line 180: Is the caller allowed to change the value stored in the object?

msimberg: No, `ReadT` will always be const (it's enforced by the `async_rw_mutex`, not the `async_rw_mutex_access_wrapper`).

hkaiser: Would changing it to  `ReadT& get() const` be an option in this case?

msimberg: Yes, good point!

msimberg: Done.

hkaiser (dismissed):
LGTM, thanks!

libs/core/synchronization/include/hpx/synchronization/async_rw_mutex.hpp
```diff
@@ -218,12 +223,17 @@ namespace hpx { namespace experimental {
             async_rw_mutex_access_wrapper& operator=(
                 async_rw_mutex_access_wrapper const&) = default;
 
-            operator ReadWriteT&()
+            ReadWriteT& get()
```
hkaiser, line 226: I assume that this function will return a non-const reference, will it? Am I right in my assumption, that the wrapper is 'valid' only while accessing (writing) the value is safe? Even then, is there a way to 'cheat' the API and to hold on to the returned reference even until after the lock has been released?

msimberg: Yes, this one will be non-const. You can definitely hold on to a reference even after the wrapper is gone, but it'll be an invalid reference since the value it points to will be moved once the wrapper (and the other wrappers pointing to the same value in the read-only case) goes out of scope. I don't know if that exact wording was intentional, but I'd flip your sentence and say: "Accessing the value is safe only while the wrapper is valid (moves are fine, copies increase the reference count)".

One more change I might do here is make the read-write wrapper non-copyable.

msimberg: Done, the read-write wrappers are now non-copyable. I've had to rebase on #5274 though, because it contains important fixes to on (on captures arguments by copy, rather than forwarding them correctly, on master; this is fixed in #5274).

PR #5326, opened 2021-05-11
Adding resize() to serialize_buffer allowing to shrink its size

hkaiser:
This is needed for @NanmiaoWu's work on the SHAD backend.

### DIFF
--- a/libs/core/serialization/include/hpx/serialization/serialize_buffer.hpp
+++ b/libs/core/serialization/include/hpx/serialization/serialize_buffer.hpp
@@ -8,6 +8,7 @@
 #pragma once
 
 #include <hpx/config.hpp>
+#include <hpx/assert.hpp>
 #include <hpx/datastructures/traits/supports_streaming_with_any.hpp>
 #include <hpx/modules/errors.hpp>
 
@@ -281,6 +282,17 @@ namespace hpx { namespace serialization {
         std::size_t size() const
         {
             return size_;
+        }
+
+        void resize_norealloc(std::size_t newsize)
+        {
+            HPX_ASSERT_MSG(newsize <= size_,
+                "serialize_buffer::resize_norealloc: new size shouldn't be "
+                "larger than current size");
+            if (newsize < size_)
+            {
+                size_ = newsize;
+            }
         }
 
     private:


libs/core/serialization/include/hpx/serialization/serialize_buffer.hpp
```diff
@@ -283,6 +283,14 @@ namespace hpx { namespace serialization {
...
         }
 
+        void resize(std::size_t newsize)
+        {
+            if (newsize < size_)
+            {
+                size_ = newsize;
+            }
+        }
+
```
msimberg, line 293: Please take this with a grain of salt as I know nothing of how the serialization works, but does this not require reallocation anywhere? And why is the size only decreased? What happens if `newsize` is bigger than `size_`?

hkaiser: > does this not require reallocation anywhere

No, I don't think so. `serialize_buffer` can be used in two modes
- reference memory managed by somebody else
- manage its own memory
In both cases the stored size is informational only. Reducing it does not do any harm. If we would allow for increasing the size, we'd need reallocation.

> What happens if newsize is bigger than size_?

This is an open question. Since I would like to avoid reallocation, we should probably throw an exception, or at least have an assert. Also, I was considering renaming the function to `shrink_to()` or similar to convey the idea.

msimberg: Thanks for the explanation! I have no opinion on this, except that an assert or a rename sounds like a good idea.

msimberg: I'll stop with the bikeshedding after this: Does one ever want to resize to a smaller size and then resize it back to a bigger size again? I.e. the actual buffer has a capacity like a vector, and can be resized multiple times to anything smaller or equal to the capacity without reallocation. Another name suggestion: `resize_noalloc` or `resize_norealloc`? That would leave space for a `resize` that can reallocate if needed.

hkaiser: Good points... I don't think howver that we should support a use case where the size is increasing in any scenario. The use case that triggered this PR is:
```
{
    serialize_buffer<uint8_t> buffer(SOME_MAX_SIZE);
    size_t size = buffer.size();
    some_user_supplied_callback(buffer.data(), &size);
    buffer.shrink_to(size);
    return buffer;
}
```
We have no control over the interface contract of calling a user supplied function with a max-sized buffer where the function is returning the actual size in its second argument after filling in its data into the supplied memory buffer. I personally believe this to be (very) bad design, but we're stuck with it, unfortunately...

I'm fine with renaming the function `resize_norealloc`.

msimberg (approved):

PR #5327, opened 2021-05-12
Fix missing ifdef HPX_SMT_PAUSE

jnvance:
Due to the missing definition of `HPX_SMT_PAUSE` in `compiler_fence.hpp`, source file `spinlock.cpp` cannot be compiled with the Intel compiler.

https://github.com/STEllAR-GROUP/hpx/blob/c320e11a3c1771c8accd446362d245da33016b15/libs/core/config/include/hpx/config/compiler_fence.hpp#L22-L26

### DIFF
--- a/libs/core/thread_support/src/spinlock.cpp
+++ b/libs/core/thread_support/src/spinlock.cpp
@@ -20,11 +20,13 @@ namespace hpx { namespace util { namespace detail {
         // Experiments on Windows and Fedora 32 show that a single pause,
         // followed by an immediate sleep, is best.
 
+#if defined(HPX_SMT_PAUSE)
         if (k == 0)
         {
             HPX_SMT_PAUSE;
         }
         else
+#endif
         {
             std::this_thread::sleep_for(std::chrono::microseconds(1));
         }


hkaiser:
restest

libs/core/thread_support/src/spinlock.cpp
```diff
@@ -20,11 +20,13 @@ namespace hpx { namespace util { namespace detail {
         // Experiments on Windows and Fedora 32 show that a single pause,
         // followed by an immediate sleep, is best.
 
+#if defined(HPX_SMT_PAUSE)
         if (k == 0)
         {
             HPX_SMT_PAUSE;
         }
         else
+#endif
```
hkaiser, line 29: Wouldn't it be better to actually define the macro instead of this conditional block?

jnvance: I agree. I guess [_mm_pause](https://software.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/intrinsics/intrinsics-for-intel-streaming-simd-extensions-2-intel-sse2/miscellaneous-functions-and-intrinsics/pause-intrinsic.html) would be the one for intel. 

I just followed how it's currently done in other parts of the code:
https://github.com/STEllAR-GROUP/hpx/blob/c320e11a3c1771c8accd446362d245da33016b15/libs/core/threading_base/src/execution_agent.cpp#L65-L70
https://github.com/STEllAR-GROUP/hpx/blob/c320e11a3c1771c8accd446362d245da33016b15/libs/core/execution_base/src/this_thread.cpp#L95-L97

msimberg: Please see https://github.com/STEllAR-GROUP/hpx/pull/5199/files#diff-94191e4d2264ae001231f8f358e44f0a855b33a427e48e41e8df445e5309fe5c as well. That defines `HPX_SMT_PAUSE` to be `_mm_pause`.

However, while doing those changes I was considering dropping support for ICC. Now that we have someone who seems to use ICC, I'd be curious to hear how hard of a requirement you have to use ICC? Do you find that ICC does a better job in whatever application you're using HPX?

jnvance: Then that PR should be able to fix this issue as well. The main reason for using ICC in our application (independent of HPX) is better automatic vectorization of some complex for-loops compared to GCC which leads to 2x speedup for that part of the code.

msimberg: Fair enough, thanks! That's a reasonable reason for using ICC. I'll see how we can keep the ICC CI going then (probably in a reduced form).

PR #5329, opened 2021-05-14
Sender overloads for parallel algorithms

msimberg:
The changes here are not meant to be *the* way to do this, but rather a place to discuss the alternatives in using P0443/1897 functionality with the parallel algorithms.

The last commit (https://github.com/STEllAR-GROUP/hpx/commit/131faa11ee48996c49398faf575999a9e1d6e8b6) of this adds the very simplest way I can think of currently integrating algorithms into task graphs, that also allows customization on the execution policy (not P0443 executors or schedulers). The overload is this (simplified):

```
auto tag_invoke(for_each_t, ExecutionPolicy p, PredecessorSender s) {
    return transform(s, bind_front(for_each_t{}, p));
}
```
where `PredecessorSender s` has `value_types` that are required by the plain synchronous `tag_invoke` overloads.

`for_each` can then be used in a task graph like this:
```
auto iter = sync_wait(ranges::for_each(some_exec_policy, when_all(range_sender, just(f))));
```
The sender from `for_each` can of course be used to chain further algorithms to it as well.

Some points to consider:
1. The `tag_invoke` overload could possibly be optimized by somehow taking advantage of the `task` policy. However, that would (as far as I can tell) involve roundtripping the `PredecessorSender` into a future to be able to use `dataflow` in place of `transform` (senders of senders are not collapsed like futures of futures are). It'd also be nice if it didn't need to rely on futures (in the long term).
2. The execution policy is not part of the `PredecessorSender` here to allow easier customization. If it were part of the sender it I'm not sure we would be able to optimally schedule e.g. CUDA versions of the algorithms (read: it would require unnecessary yielding instead of just attaching a continuation to an event).
3. Just directly doing `transform(predecessor, for_each)` in user code seems (to me) unlikely to ever happen in the standard library, at least for the non-ranges overloads, as long as they are not CPOs.
4. How should/can/might this interact with pipelining of ranges algorithms/views (@gonidelis)?
5. Note that something like `args_sender | on/via(std_thread_scheduler) | for_each(cuda_policy) | ...` would be fine and would mean the work should be *submitted* from a `std::thread` but *executed* according to the `cuda_policy` on a GPU, i.e. `on/via` would not control where the algorithm is executed.

Edit:
A possible alternative implementation that takes advantage of the task policy would look a bit like this:
```
auto tag_invoke(for_each_t, ExecutionPolicy p, PredecessorSender s) {
    return let_value(s, [policy](auto... ts) { return for_each_t{}(policy(task), ts...); });
}
```
I have no idea if there would be a noticeable difference in overheads.

To do:
- [x] rebase
- [x] add extensive tests for at least one of the algorithms

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_each.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_each.hpp
@@ -383,6 +383,7 @@ namespace hpx { namespace ranges {
 
 #include <hpx/algorithms/traits/projected_range.hpp>
 #include <hpx/parallel/algorithms/for_each.hpp>
+#include <hpx/parallel/util/detail/sender_util.hpp>
 #include <hpx/parallel/util/projection_identity.hpp>
 
 #include <type_traits>
@@ -421,7 +422,7 @@ namespace hpx { namespace ranges {
     ///////////////////////////////////////////////////////////////////////////
     // DPO for hpx::ranges::for_each
     HPX_INLINE_CONSTEXPR_VARIABLE struct for_each_t final
-      : hpx::functional::tag_fallback<for_each_t>
+      : hpx::detail::tag_parallel_algorithm<for_each_t>
     {
         // clang-format off
         template <typename InIter, typename Sent, typename F,
@@ -530,7 +531,7 @@ namespace hpx { namespace ranges {
     ///////////////////////////////////////////////////////////////////////////
     // DPO for hpx::ranges::for_each_n
     HPX_INLINE_CONSTEXPR_VARIABLE struct for_each_n_t final
-      : hpx::functional::tag_fallback<for_each_n_t>
+      : hpx::detail::tag_parallel_algorithm<for_each_n_t>
     {
         // clang-format off
         template <typename InIter, typename Size, typename F,
--- a/libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/merge.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/merge.hpp
@@ -384,7 +384,6 @@ namespace hpx { namespace ranges {
 
 #include <hpx/config.hpp>
 #include <hpx/concepts/concepts.hpp>
-#include <hpx/functional/tag_dispatch.hpp>
 #include <hpx/iterator_support/range.hpp>
 #include <hpx/iterator_support/traits/is_iterator.hpp>
 #include <hpx/iterator_support/traits/is_sentinel_for.hpp>
@@ -393,6 +392,7 @@ namespace hpx { namespace ranges {
 #include <hpx/algorithms/traits/projected_range.hpp>
 #include <hpx/executors/execution_policy.hpp>
 #include <hpx/parallel/algorithms/merge.hpp>
+#include <hpx/parallel/util/detail/sender_util.hpp>
 #include <hpx/parallel/util/detail/algorithm_result.hpp>
 #include <hpx/parallel/util/result_types.hpp>
 
@@ -518,7 +518,7 @@ namespace hpx { namespace ranges {
     ///////////////////////////////////////////////////////////////////////////
     // DPO for hpx::ranges::merge
     HPX_INLINE_CONSTEXPR_VARIABLE struct merge_t final
-      : hpx::functional::tag<merge_t>
+      : hpx::detail::tag_parallel_algorithm<merge_t>
     {
     private:
         // clang-format off
@@ -543,9 +543,9 @@ namespace hpx { namespace ranges {
             hpx::ranges::merge_result<
                 typename hpx::traits::range_iterator<Rng1>::type,
                 typename hpx::traits::range_iterator<Rng2>::type, Iter3>>::type
-        tag_dispatch(merge_t, ExPolicy&& policy, Rng1&& rng1, Rng2&& rng2,
-            Iter3 dest, Comp&& comp = Comp(), Proj1&& proj1 = Proj1(),
-            Proj2&& proj2 = Proj2())
+        tag_fallback_dispatch(merge_t, ExPolicy&& policy, Rng1&& rng1,
+            Rng2&& rng2, Iter3 dest, Comp&& comp = Comp(),
+            Proj1&& proj1 = Proj1(), Proj2&& proj2 = Proj2())
         {
             using iterator_type1 =
                 typename hpx::traits::range_iterator<Rng1>::type;
@@ -592,9 +592,10 @@ namespace hpx { namespace ranges {
         // clang-format on
         friend typename hpx::parallel::util::detail::algorithm_result<ExPolicy,
             hpx::ranges::merge_result<Iter1, Iter2, Iter3>>::type
-        tag_dispatch(merge_t, ExPolicy&& policy, Iter1 first1, Sent1 last1,
-            Iter2 first2, Sent2 last2, Iter3 dest, Comp&& comp = Comp(),
-            Proj1&& proj1 = Proj1(), Proj2&& proj2 = Proj2())
+        tag_fallback_dispatch(merge_t, ExPolicy&& policy, Iter1 first1,
+            Sent1 last1, Iter2 first2, Sent2 last2, Iter3 dest,
+            Comp&& comp = Comp(), Proj1&& proj1 = Proj1(),
+            Proj2&& proj2 = Proj2())
         {
             static_assert(hpx::traits::is_random_access_iterator<Iter1>::value,
                 "Required at least random access iterator.");
@@ -632,7 +633,7 @@ namespace hpx { namespace ranges {
         friend hpx::ranges::merge_result<
             typename hpx::traits::range_iterator<Rng1>::type,
             typename hpx::traits::range_iterator<Rng2>::type, Iter3>
-        tag_dispatch(merge_t, Rng1&& rng1, Rng2&& rng2, Iter3 dest,
+        tag_fallback_dispatch(merge_t, Rng1&& rng1, Rng2&& rng2, Iter3 dest,
             Comp&& comp = Comp(), Proj1&& proj1 = Proj1(),
             Proj2&& proj2 = Proj2())
         {
@@ -679,10 +680,10 @@ namespace hpx { namespace ranges {
                 >::value
             )>
         // clang-format on
-        friend hpx::ranges::merge_result<Iter1, Iter2, Iter3> tag_dispatch(
-            merge_t, Iter1 first1, Sent1 last1, Iter2 first2, Sent2 last2,
-            Iter3 dest, Comp&& comp = Comp(), Proj1&& proj1 = Proj1(),
-            Proj2&& proj2 = Proj2())
+        friend hpx::ranges::merge_result<Iter1, Iter2, Iter3>
+        tag_fallback_dispatch(merge_t, Iter1 first1, Sent1 last1, Iter2 first2,
+            Sent2 last2, Iter3 dest, Comp&& comp = Comp(),
+            Proj1&& proj1 = Proj1(), Proj2&& proj2 = Proj2())
         {
             static_assert(hpx::traits::is_random_access_iterator<Iter1>::value,
                 "Required at least random access iterator.");
@@ -703,7 +704,7 @@ namespace hpx { namespace ranges {
     ///////////////////////////////////////////////////////////////////////////
     // DPO for hpx::ranges::inplace_merge
     HPX_INLINE_CONSTEXPR_VARIABLE struct inplace_merge_t final
-      : hpx::functional::tag<inplace_merge_t>
+      : hpx::detail::tag_parallel_algorithm<inplace_merge_t>
     {
     private:
         // clang-format off
@@ -724,8 +725,8 @@ namespace hpx { namespace ranges {
         // clang-format on
         friend typename hpx::parallel::util::detail::algorithm_result<ExPolicy,
             Iter>::type
-        tag_dispatch(inplace_merge_t, ExPolicy&& policy, Rng&& rng, Iter middle,
-            Comp&& comp = Comp(), Proj&& proj = Proj())
+        tag_fallback_dispatch(inplace_merge_t, ExPolicy&& policy, Rng&& rng,
+            Iter middle, Comp&& comp = Comp(), Proj&& proj = Proj())
         {
             using iterator_type =
                 typename hpx::traits::range_iterator<Rng>::type;
@@ -758,7 +759,7 @@ namespace hpx { namespace ranges {
         // clang-format on
         friend typename hpx::parallel::util::detail::algorithm_result<ExPolicy,
             Iter>::type
-        tag_dispatch(inplace_merge_t, ExPolicy&& policy, Iter first,
+        tag_fallback_dispatch(inplace_merge_t, ExPolicy&& policy, Iter first,
             Iter middle, Sent last, Comp&& comp = Comp(), Proj&& proj = Proj())
         {
             static_assert(hpx::traits::is_random_access_iterator<Iter>::value,
@@ -785,8 +786,8 @@ namespace hpx { namespace ranges {
                 >::value
             )>
         // clang-format on
-        friend Iter tag_dispatch(inplace_merge_t, Rng&& rng, Iter middle,
-            Comp&& comp = Comp(), Proj&& proj = Proj())
+        friend Iter tag_fallback_dispatch(inplace_merge_t, Rng&& rng,
+            Iter middle, Comp&& comp = Comp(), Proj&& proj = Proj())
         {
             using iterator_type =
                 typename hpx::traits::range_iterator<Rng>::type;
@@ -817,8 +818,8 @@ namespace hpx { namespace ranges {
                 >::value
             )>
         // clang-format on
-        friend Iter tag_dispatch(inplace_merge_t, Iter first, Iter middle,
-            Sent last, Comp&& comp = Comp(), Proj&& proj = Proj())
+        friend Iter tag_fallback_dispatch(inplace_merge_t, Iter first,
+            Iter middle, Sent last, Comp&& comp = Comp(), Proj&& proj = Proj())
         {
             static_assert(hpx::traits::is_random_access_iterator<Iter>::value,
                 "Required at least random access iterator.");
--- a/libs/parallelism/algorithms/include/hpx/parallel/util/detail/sender_util.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/util/detail/sender_util.hpp
@@ -1,0 +1,125 @@
+//  Copyright (c) ETH Zurich 2021
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/execution/algorithms/detail/partial_algorithm.hpp>
+#include <hpx/execution/algorithms/let_value.hpp>
+#include <hpx/execution/algorithms/transform.hpp>
+#include <hpx/execution/traits/is_execution_policy.hpp>
+#include <hpx/execution_base/sender.hpp>
+#include <hpx/executors/execution_policy.hpp>
+
+#include <type_traits>
+#include <utility>
+
+namespace hpx { namespace detail {
+    // This is a lighter-weight alternative to bind for use in parallel
+    // algorithm overloads, where one needs to bind an execution policy to an
+    // algorithm for use in execution::transform. Typically used together with
+    // transform_with_bound_algorithm.
+    template <typename Tag, typename ExPolicy>
+    struct bound_algorithm
+    {
+        std::decay_t<ExPolicy> policy;
+
+        template <typename T1, typename... Ts>
+        auto operator()(T1&& t1, Ts&&... ts) -> decltype(Tag{}(
+            std::move(policy), std::forward<T1>(t1), std::forward<Ts>(ts)...))
+        {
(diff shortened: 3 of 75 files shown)


libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_each.hpp
```diff
@@ -525,6 +525,20 @@ namespace hpx { namespace ranges {
...
                 std::forward<Proj>(proj));
         }
+
+        // clang-format off
+        template <typename ExPolicy, typename Predecessor,
+            HPX_CONCEPT_REQUIRES_(
+                hpx::is_execution_policy<ExPolicy>::value)>
+        // clang-format on
+        friend auto tag_fallback_invoke(hpx::ranges::for_each_t,
+            ExPolicy&& policy, Predecessor&& predecessor)
```
hkaiser, line 535: There probably should be some SFINAE condition for the `Predecessor` as well.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_each.hpp
```diff
@@ -525,6 +525,20 @@ namespace hpx { namespace ranges {
...
+        {
+            return hpx::execution::experimental::transform(
+                std::forward<Predecessor>(predecessor),
+                hpx::util::bind_front(
+                    for_each_t{}, std::forward<ExPolicy>(policy)));
+        }
```
hkaiser, line 541: Nice!

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/for_each.hpp
msimberg: Absolutely. My first attempt at requiring it satisfy `is_sender` failed for some (worrying) reason, but given that this is mostly for the sake of the idea at this point I didn't look further into it.

hkaiser: > My first attempt at requiring it satisfy is_sender failed for some (worrying) reason

Did you forget to `decay` the argument?

msimberg: Possible... I'll have to give it another shot.

msimberg:
This in principle allows using all parallel algorithms in chains of senders, with the help of another helper base class for the CPOs. The helper functionality is in `libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/transform_reduce.hpp`.

I still haven't done any comparison of the `let_value` and `transform` versions. The unsequenced policies seem to be missing task versions which make the `let_value` version problematic at the moment. Is this intentional or just a matter of noone having had the time yet for that? Tests are obviously still missing. I'd also still want to explore rewriting the algorithms using custom senders (not futures), but that is a bigger task...

hkaiser:
> The unsequenced policies seem to be missing task versions which make the `let_value` version problematic at the moment.

Nobody has had time/interest to look into this yet. @srinivasyadav18 might be able to add those this summer.

libs/parallelism/algorithms/include/hpx/parallel/util/detail/sender_util.hpp
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        template <typename ExPolicy,
+            HPX_CONCEPT_REQUIRES_(hpx::is_execution_policy<ExPolicy>::value)>
+        friend auto tag_fallback_dispatch(Tag, ExPolicy&& policy)
+        {
+            return hpx::execution::experimental::detail::partial_algorithm<Tag,
+                ExPolicy>{std::forward<ExPolicy>(policy)};
+        }
+    };
```
hkaiser, line 112: I'm a bit concerned about putting this into `hpx::detail`, however I admit that I don't have the full understanding of the name lookup implications of doing so. For one, the `tag_dispatch`  friend functions here will be injected into that namespace (this is probably of no consequence). Also, how will the compiler find the overloads? Is it because this type is a base class for the algorithm DPOs?

libs/parallelism/algorithms/include/hpx/parallel/util/detail/sender_util.hpp
```diff
@@ -0,0 +1,113 @@
...
+    //      predecessor | partially_applied_parallel_algorithm.
+    template <typename Tag>
+    struct tag_parallel_algorithm : hpx::functional::tag_fallback<Tag>
+    {
+        template <typename Predecessor, typename ExPolicy,
+            typename Enable = std::enable_if_t<
+                std::conjunction_v<hpx::is_execution_policy<ExPolicy>,
+                    std::negation<detail::is_bound_algorithm<Predecessor>>,
+                    hpx::execution::experimental::is_sender<
+                        std::decay_t<Predecessor>>>>>
```
hkaiser, line 96: `HPX_CONCEPT_REQUIRES_()` might make this  a bit more readable.

libs/parallelism/algorithms/include/hpx/parallel/util/detail/sender_util.hpp
```diff
@@ -0,0 +1,113 @@
...
+    {
+        // Both versions below can work. The first makes use of task policies,
+        // and futures being senders. The first version may be slightly more
+        // efficient since a receiver will be connected directly to the returned
+        // future. The second version uses the blocking versions of the parallel
+        // algorithms, and waits on the future inside the transform until the
+        // algorithm has finished.
+
+        // TODO: task policies are mossing for unseq/par_unseq.
+        auto task_policy = std::move(policy)(hpx::execution::task);
```
hkaiser, line 63: Shouldn't this use `std::forward` - if at all?

libs/parallelism/algorithms/include/hpx/parallel/util/detail/sender_util.hpp
msimberg: Good point. Yes, because it's a base class. I'm also not sure if this will have unintended consequences. Might be better to move it (although that would require separate implementations in `hpx::` and `hpx::ranges::` if we really don't want additional lookup). Then again, the `tag_fallback` base class already pulls in `hpx::functional` for ADL.

msimberg:
Any preferences on how to proceed with this? I absolutely need to add proper tests for this, but with the current information we have this is about as done as it will be at the moment in terms of the actual overloads provided. I would consider the new overloads very much experimental. On one hand I'm sure that we'll tweak it over time and we might want to wait to find the right structure for them. On the other, getting feedback from users trying it out would be even nicer. Two more datapoints:

- I'm currently trying out something similar in DLA-Future with a custom policy specific for DLA-Future. I suppose we'll get some experience on pros and cons of this approach once we actually work with it a bit more. Example from DLA-Future: https://github.com/eth-cscs/DLA-Future/blob/3b7a3706337356d9c1829b5e297e775d175ee4e1/include/dlaf/factorization/cholesky/impl.h#L39-L49.
- In the P2300 presentation (I'll try to link to it when I find it/it's public) they *very tentatively* suggested something like this: `algorithm(executing_on(scheduler, policy), ...) -> T` and `async_algorithm(executing_async(sender, policy), ...) -> sender` for the blocking and non-blocking sender cases. I don't think it's very thought-through yet. E.g. in the second case I'm not sure what is supposed to go in the `sender` passed to `executing_async` and what goes in `...`.

hkaiser:
> Any preferences on how to proceed with this? I absolutely need to add proper tests for this, but with the current information we have this is about as done as it will be at the moment in terms of the actual overloads provided. I would consider the new overloads very much experimental. On one hand I'm sure that we'll tweak it over time and we might want to wait to find the right structure for them. On the other, getting feedback from users trying it out would be even nicer. 

Yes, I'm in favor of going ahead with what you have right now.

> Two more datapoints:
> 
> * I'm currently trying out something similar in DLA-Future with a custom policy specific for DLA-Future. I suppose we'll get some experience on pros and cons of this approach once we actually work with it a bit more. Example from DLA-Future: https://github.com/eth-cscs/DLA-Future/blob/3b7a3706337356d9c1829b5e297e775d175ee4e1/include/dlaf/factorization/cholesky/impl.h#L39-L49.

Cool!

> * In the P2300 presentation (I'll try to link to it when I find it/it's public) they _very tentatively_ suggested something like this: `algorithm(executing_on(scheduler, policy), ...) -> T` and `async_algorithm(executing_async(sender, policy), ...) -> sender` for the blocking and non-blocking sender cases. I don't think it's very thought-through yet. E.g. in the second case I'm not sure what is supposed to go in the `sender` passed to `executing_async` and what goes in `...`. Edit: the slides will be here eventually: https://wg21.link/P2403.

The `algorithm(executing_on(scheduler, policy), ...)` part is essentially equivalent to our `algorithm(policy.on(scheduler_executor(scheduler)), ...)`, so I like it! We should be able to easily implement that. Also, I like the fact that they agree on the need to have asynchronous algorithms ;-)

msimberg:
> The `algorithm(executing_on(scheduler, policy), ...)` part is essentially equivalent to our `algorithm(policy.on(scheduler_executor(scheduler)), ...)`, so I like it! We should be able to easily implement that. Also, I like the fact that they agree on the need to have asynchronous algorithms ;-)

Indeed! Unfortunately I fear that it's going to be out of scope for the first iteration, if any of this actually does make it into C++23, but that doesn't stop us from doing it already now.

msimberg:
I've now expanded the tests for the sender-and-ranges version of `for_each`, but I have not added tests for the sender overloads for all the other algorithms. I think we'll have to find some other, more sustainable way of adding tests for all the algorithms than copy-pasting tests for all the different overloads.

As a flyby I've updated some remaining algorithms which still used `tag` instead of `tag_fallback` as the base class. Or have I missed some reasoning for leaving them with `tag`? I assumed it was just a question of time until @Jedi18 would've gotten to them (note that I've only changed the base class, I haven't done any adaptations to support sentinels, if those were needed).

hkaiser:
> I've now expanded the tests for the sender-and-ranges version of `for_each`, but I have not added tests for the sender overloads for all the other algorithms. I think we'll have to find some other, more sustainable way of adding tests for all the algorithms than copy-pasting tests for all the different overloads.

Yah, that would be nice.

> As a flyby I've updated some remaining algorithms which still used `tag` instead of `tag_fallback` as the base class. Or have I missed some reasoning for leaving them with `tag`? I assumed it was just a question of time until @Jedi18 would've gotten to them (note that I've only changed the base class, I haven't done any adaptations to support sentinels, if those were needed).

Thanks for doing this. I think we either didn't touch those algorithms yet or simply missed them.

libs/parallelism/algorithms/include/hpx/parallel/util/detail/sender_util.hpp
```diff
@@ -0,0 +1,125 @@
...
+        }
+        // If the policy does not have a task policy, the algorithm can only be
+        // called synchronously. In this case we only use transform to chain the
+        // algorithm after the predecessor sender.
+        else
+        {
+            return hpx::execution::experimental::transform(
+                std::forward<Predecessor>(predecessor),
+                bound_algorithm<Tag, ExPolicy>{std::forward<ExPolicy>(policy)});
+        }
```
hkaiser, line 79: I think that we should create the task policies where missing. `unseq` and `par_unseq` simply map onto `par` and `seq` anyways... I have created a ticket for us not to forget: https://github.com/STEllAR-GROUP/hpx/issues/5492.

hkaiser:
Just a general note: I really like where this is going! One question, though: should the algorithms inherit the scheduler from their predecessor? This should be easily doable once #5487 and #5429 have been merged (i.e. `on(scheduler) | algorithm(policy, ...)`). Or should we go with `algorithm(executing_on(scheduler, policy), ...)` as mentioned in P2403 to begin with? Possibly both?

libs/parallelism/algorithms/include/hpx/parallel/util/detail/sender_util.hpp
msimberg: Thanks for opening the issue! I forgot to mention that I went the shorter route of removing the task overloads for those policies, but you noticed anyway ;)

msimberg:
> Just a general note: I really like where this is going! One question, though: should the algorithms inherit the scheduler from their predecessor? This should be easily doable once #5487 and #5429 have been merged (i.e. `on(scheduler) | algorithm(policy, ...)`). Or should we go with `algorithm(executing_on(scheduler, policy), ...)` as mentioned in P2403 to begin with? Possibly both?

I'm unsure about that. I prefer the explicitness of having the scheduler in the algorithm call rather than implicit from the predecessors, but I can see how both could work. I also wonder if user-defined execution policies will ever be a thing, and if yes, should e.g. CUDA execution be `executing_on(cuda_scheduler, par)` or `cuda_policy` (or either?). `executing_on(cuda_scheduler, seq)` would be a bit odd, though feasible. We of course extend the policies currently with the Kokkos execution policies, but that's not to say it's the right way. In the case of Kokkos it's also not obvious to me that there should be a "Kokkos scheduler". A "Kokkos policy" is a much narrower concept than a scheduler.

hkaiser:
> > Just a general note: I really like where this is going! One question, though: should the algorithms inherit the scheduler from their predecessor? This should be easily doable once #5487 and #5429 have been merged (i.e. `on(scheduler) | algorithm(policy, ...)`). Or should we go with `algorithm(executing_on(scheduler, policy), ...)` as mentioned in P2403 to begin with? Possibly both?
> 
> I'm unsure about that. I prefer the explicitness of having the scheduler in the algorithm call rather than implicit from the predecessors, but I can see how both could work. I also wonder if user-defined execution policies will ever be a thing, and if yes, should e.g. CUDA execution be `executing_on(cuda_scheduler, par)` or `cuda_policy` (or either?). `executing_on(cuda_scheduler, seq)` would be a bit odd, though feasible. We of course extend the policies currently with the Kokkos execution policies, but that's not to say it's the right way. In the case of Kokkos it's also not obvious to me that there should be a "Kokkos scheduler". A "Kokkos policy" is a much narrower concept than a scheduler.

I can see the implicit `on(scheduler) | ... | algorithm(policy, ...)` to be closer to what users would expect as otherwise one will have to specify the scheduler twice. This would leave the explicit `on(scheduler1) |... | algorithm(executing_on(scheduler2, policy), ...)` to cases where the algorithm should really run on a different scheduler compared to its predecessor (and possibly not even propagate the scheduler to its successor). But even in this case the first option would be as explicit: `... | on(scheduler2) | algorithm(policy, ...)`, which would be idiomatic and aligned with all the other senders and receivers.

The discussion about policies is a bit orthogonal and a very conceptual one. I wouldn't like to have a dedicated CUDA policy (or Kokkos policy for that matter) if I can avoid it. It is not the task of the policy to determine where or when to execute things. The mishmash we have is caused by the weak executor story that currently underlies things (and our desire to use overload resolution to pick the right algorithm implementation). `par` and `seq` are really just hints on how to execute things, thus I'd very much prefer  `executing_on(cuda_scheduler, par)` over a `cuda_policy`.

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/merge.hpp
```diff
@@ -384,7 +384,7 @@ namespace hpx { namespace ranges {
 
 #include <hpx/config.hpp>
 #include <hpx/concepts/concepts.hpp>
-#include <hpx/functional/tag_dispatch.hpp>
+#include <hpx/parallel/util/detail/sender_util.hpp>
```
msimberg, line 387: TODO: There are a few files which include `sender_util.hpp` twice after my find and replaces...

msimberg:
> I can see the implicit `on(scheduler) | ... | algorithm(policy, ...)` to be closer to what users would expect as otherwise one will have to specify the scheduler twice. This would leave the explicit `on(scheduler1) | ... | algorithm(executing_on(scheduler2, policy), ...)` to cases where the algorithm should really run on a different scheduler compared to its predecessor (and possibly not even propagate the scheduler to its successor). But even in this case the first option would be as explicit: `... | on(scheduler2) | algorithm(policy, ...)`, which would be idiomatic and aligned with all the other senders and receivers.

All good points. More aligned with the other sender adapters would actually be to prioritize `algorithm(policy, scheduler, sender)` if `sender` has a completion scheduler and `algorithm` has such an overload. This would allow implicitly using the completion scheduler, or explicitly passing a scheduler if one wants to scheduler. I kind of prefer this over `algorithm(executing_on(policy, scheduler), ...)` because you get both. That separates the policy from the scheduler which can be good or bad...

> EDIT: Let's not forget that there are the 'old' algorithms as well: `on(scheduler) | ... | for_each()`, i.e. the ones that do not take an execution policy - hmmm, can we even write this as `on(scheduler) | ... | for_each`?

We can't at least after this PR, but in general probably yes. We'll have to be careful about that implicit scheduler though: what should `on(scheduler) | ... | for_each(policy.on(executor))` do? Or the one without a policy for that matter. It has an implicit sequenced executor in the non-sender version. And if e.g. `for_each` would internally dispatch to `bulk`, `bulk` would have to know if the policy is `seq` or `par` (and `bulk` currently doesn't know or care about the policy).

The transition from our old executor/policy setup to scheduler/policy will be interesting. Ideally we'd avoid as much *implicit* conversion between the two different worlds as possible. I also still think the sender overloads introduced here could live in a completely separate namespace to avoid mixing the two worlds unnecessarily.

> The discussion about policies is a bit orthogonal and a very conceptual one. I wouldn't like to have a dedicated CUDA policy (or Kokkos policy for that matter) if I can avoid it. It is not the task of the policy to determine where or when to execute things. The mishmash we have is caused by the weak executor story that currently underlies things (and our desire to use overload resolution to pick the right algorithm implementation). `par` and `seq` are really just hints on how to execute things, thus I'd very much prefer `executing_on(cuda_scheduler, par)` over a `cuda_policy`.

Yeah, fair points.

msimberg:
I don't know how theoretical this concern is, but consider an FPGA which *only* does sorting. If one requires parallel algorithms be scheduled either with `on(scheduler)` or `executing_on(policy, scheduler)` then such an FPGA would need a dummy scheduler which isn't actually able to schedule anything. At the very least it should be possible to do `executing_on(policy, execution_context?)`, or something similar where `execution_context` doesn't *have* to be a full scheduler (i.e. implement `schedule`).

msimberg:
retest cscs

hkaiser:
> I don't know how theoretical this concern is, but consider an FPGA which _only_ does sorting. If one requires parallel algorithms be scheduled either with `on(scheduler)` or `executing_on(policy, scheduler)` then such an FPGA would need a dummy scheduler which isn't actually able to schedule anything. At the very least it should be possible to do `executing_on(policy, execution_context?)`, or something similar where `execution_context` doesn't _have_ to be a full scheduler (i.e. implement `schedule`).

Isn't that a general problem? All senders/receivers can be used on a particular execution-context only if they are being used with the appropriate scheduler that was derived from that context.

hkaiser:
> We'll have to be careful about that implicit scheduler though: what should `on(scheduler) | ... | for_each(policy.on(executor))` do?

I think we need to agree on what takes preference. In your example I would simply discard the executor set with `.on()`. IOW, the completion scheduler should take preference (we're in sender/receiver land after all). It might be possible to issue a warning/error, but I'm not sure.

msimberg:
> Isn't that a general problem? All senders/receivers can be used on a particular execution-context only if they are being used with the appropriate scheduler that was derived from that context.

I'm not quite sure what you mean by this so I don't what to respond before I understand this fully. What I mean is that a dummy scheduler is quite an ugly mechanism for dispatching to some special hardware. Additionally, what should `scheduler_for_hardware_that_can_only_sort | transform | sync_wait`? I can always delete the overload for the special scheduler, but should I do that for every possible algorithm that the scheduler might accidentally be used with? Maybe this is exactly what you mean that schedulers can only be used with certain sender adapters, but my point is that specifying the scheduler in the algorithm call (and only there) for such a special scheduler keeps those weird cases away.

In any case, it's quite clear that this will not be the final version, but at least this first iteration is done from my point of view.

hkaiser:
> > Isn't that a general problem? All senders/receivers can be used on a particular execution-context only if they are being used with the appropriate scheduler that was derived from that context.
> 
> I'm not quite sure what you mean by this so I don't what to respond before I understand this fully. 

What I meant is that each execution environment has it's own scheduling mechanisms (CPUs have our schedulers, CUDA has its own, so do FPGAs, etc.) It will be the responsibility of the user to make sure the sender/receiver chain is run on an appropriate scheduler (using `on()` or similar).

> What I mean is that a dummy scheduler is quite an ugly mechanism for dispatching to some special hardware. Additionally, what should `scheduler_for_hardware_that_can_only_sort | transform | sync_wait`? 

In the best case this should fail to compile, at least it should complain at runtime.

> I can always delete the overload for the special scheduler, but should I do that for every possible algorithm that the scheduler might accidentally be used with? Maybe this is exactly what you mean that schedulers can only be used with certain sender adapters, but my point is that specifying the scheduler in the algorithm call (and only there) for such a special scheduler keeps those weird cases away.

Yah, I'm not sure what the best way forward should be.

> In any case, it's quite clear that this will not be the final version, but at least this first iteration is done from my point of view.

Right. I'm fine with what you have, let's start playing with it to see what limitations we may hit.

hkaiser:
PRs merged concurrently need updating now, for instance `rotate` and friends. @hcq9102 could you please have a look into changing the `tag_fallback` base classes to `tag_parallel_algorithm` (see here for an example: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/parallelism/algorithms/include/hpx/parallel/algorithms/for_each.hpp#L681)?

hcq9102:
Sure, I will do it.
I need to creat a new PR to do this, right?
Thanks.

Hartmut Kaiser ***@***.***>于2021年8月17日 周二下午7:57写道：

hkaiser:
> Sure, I will do it. I need to creat a new PR to do this, right? Thanks.

Yes.

hcq9102:
Got it.
Thanks

Hartmut Kaiser ***@***.***>于2021年8月17日 周二下午8:30写道：

Issue #4942, opened 2020-09-03
1.5.0 breaks w/out CUDA: Error evaluating generator expression: $<COMPILE_LANG_AND_ID:CUDA,NVIDIA>


yurivict:
```
-- Configuring done
CMake Error at cmake/HPX_GeneratePackageUtils.cmake:365 (file):
  Error evaluating generator expression:

    $<COMPILE_LANG_AND_ID:CUDA,NVIDIA>

  $<COMPILE_LANG_AND_ID:lang,id> may only be used with binary targets to
  specify include directories, compile definitions, and compile options.  It
  may not be used with the add_custom_command, add_custom_target, or
  file(GENERATE) commands.
Call Stack (most recent call first):
  cmake/HPX_GeneratePackage.cmake:83 (hpx_generate_pkgconfig_from_target)
  CMakeLists.txt:2629 (include)


CMake Error at cmake/HPX_GeneratePackageUtils.cmake:365 (file):
  Error evaluating generator expression:

    $<COMPILE_LANG_AND_ID:CUDA,NVIDIA>

  $<COMPILE_LANG_AND_ID:lang,id> may only be used with binary targets to
  specify include directories, compile definitions, and compile options.  It
  may not be used with the add_custom_command, add_custom_target, or
  file(GENERATE) commands.
Call Stack (most recent call first):
  cmake/HPX_GeneratePackage.cmake:83 (hpx_generate_pkgconfig_from_target)
  CMakeLists.txt:2629 (include)
```

OS: FreeBSD 12.1
clang-9

msimberg:
@yurivict could you please post how you've configured HPX, including CMake version? That generator expression does not look like somthing we use internally in HPX.

yurivict:
cmake is invoked:
```
[...]

PR #5330, opened 2021-05-14
Add option to disable pkgconfig generation

msimberg:
@aurianer this is relevant to #5283. Also to https://github.com/STEllAR-GROUP/hpx/issues/4942 (as this would allow working around problems with generating pkgconfig files by simply disabling it).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -315,15 +315,11 @@ endif()
 hpx_option(
   HPX_WITH_PRECOMPILED_HEADERS
   BOOL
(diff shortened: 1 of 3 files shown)


msimberg:
I updated this to turn off pkgconfig file generation by default if on Windows, or if CUDA/HIP is enabled. It can still be turned on, especially for the CUDA/HIP cases, but we make no guarantees that it'll work with them turned on.

msimberg:
Any comments or ok to merge?

hkaiser (dismissed):
LGTM, thanks!

PR #5332, opened 2021-05-14
Simplify tag_invoke and friends

hkaiser:
The goal is to streamline the implementation of `tag_invoke` and friends. 

@msimberg, @K-ballo, if you could spare a minute, I'd appreciate a second pair of eyes as I'm not sure that the proposed changes will not subtly change overload resolution.

### DIFF
--- a/libs/core/functional/include/hpx/functional/tag_fallback_invoke.hpp
+++ b/libs/core/functional/include/hpx/functional/tag_fallback_invoke.hpp
@@ -116,7 +116,7 @@ namespace hpx { namespace functional {
     ///////////////////////////////////////////////////////////////////////////
     namespace tag_fallback_invoke_t_ns {
 
-        // MSVC needs this, don't ask
+        // poison pill
         void tag_fallback_invoke();
 
         struct tag_fallback_invoke_t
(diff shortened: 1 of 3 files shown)


hkaiser:
@msimberg not sure where the 443_executor test problem comes from, but otherwise things seem to be fine now.

hkaiser:
retest lsu

msimberg:
> @msimberg not sure where the 443_executor test problem comes from, but otherwise things seem to be fine now.

Not sure since it's probably from the previous CI run, but most likely it's the same failure that was mentioned earlier and is fixed by #5274?

msimberg:
@hkaiser would you mind summarizing the changes here. It looks like the `tag_X_invoke_ns` namespace became regular (non-inline) namespaces, the helper base classes moved into sub-namespaces, while being typedefed in separate inline namespaces. Is this to prevent certain accidental/unwanted overloads matching or similar? Is there some other change that I've missed?

hkaiser:
> @hkaiser would you mind summarizing the changes here. It looks like the `tag_X_invoke_ns` namespace became regular (non-inline) namespaces, the helper base classes moved into sub-namespaces, while being typedefed in separate inline namespaces. Is this to prevent certain accidental/unwanted overloads matching or similar? Is there some other change that I've missed?

Yes, this mostly sums it up. The main change is that the helper base classes now directly dispatch to the `tag_invoke` implementations instead of using `hpx::functional::tag_invoke` as before. This saves one function call for each tag_invoke invocation. The namespace changes were needed to avoid the roundtrip through the `hpx::functional::tag_invoke` facility.

msimberg:
Thanks @hkaiser! I *think* we can ignore the ICC failures. As far as I can tell they only happen because this is still using the configuration from before #5199 was merged.

PR #5335, opened 2021-05-17
Make partial_algorithm work with any number of arguments

msimberg:
I previously only specialized it for zero and one arguments, out of laziness. This should make it work with any number.

### DIFF
--- a/libs/parallelism/execution/include/hpx/execution/algorithms/detail/partial_algorithm.hpp
+++ b/libs/parallelism/execution/include/hpx/execution/algorithms/detail/partial_algorithm.hpp
@@ -6,6 +6,9 @@
 
 #pragma once
 
+#include <hpx/datastructures/member_pack.hpp>
+
+#include <cstddef>
 #include <type_traits>
 #include <utility>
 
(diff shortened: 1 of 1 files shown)


libs/parallelism/execution/include/hpx/execution/algorithms/detail/partial_algorithm.hpp
```diff
@@ -6,37 +6,47 @@
...
 #include <type_traits>
 #include <utility>
 
 namespace hpx {
     namespace execution {
         namespace experimental {
             namespace detail {
-    template <typename Tag, typename... Ts>
-    struct partial_algorithm;
+    template <typename Tag, typename Is, typename... Ts>
```
hkaiser, line 19: Nitpick: would you mind renaming `Is` to `Pack` (or something similar), please?

hkaiser (dismissed):
LGTM, very nice, thanks!

libs/parallelism/execution/include/hpx/execution/algorithms/detail/partial_algorithm.hpp
msimberg: I don't mind, but I named it `Is` to follow the convention in the functional module (and e.g. `just.hpp`, but that has the same motivation as here). At least I associate plain `Pack` with a pack of arguments like `Ts` so I'd prefer to keep the "index"-ness in the name somehow. `IsPack`? `IndexPack`?

aurianer (approved):
LGTM thanks!

libs/parallelism/execution/include/hpx/execution/algorithms/detail/partial_algorithm.hpp
hkaiser: I meant just changing the name of the parameter for the main template. I'd be absolutely fine with leaving `Is...` for where it refers to a sequence of indexes. But mind you - this is _really_ minor.

msimberg: Don't worry; details matter. I changed it to `IsPack`.

PR #5341, opened 2021-05-18
Adding make_with_annotation and get_annotation properties

### LINKED ISSUE #5269: Allow setting a label on executors/policies
Adding this as an issue so that we can come back to this idea: https://github.com/STEllAR-GROUP/hpx/pull/5267#pullrequestreview-623941366.

It may be preferable to allow setting labels on executors/properties instead of functions like this:
```
hpx::transform_reduce(hpx::prefer(hpx::execution::par, hpx::label("my transform_reduce")), ...);
```

I'm not sure what the best order of preference would be for the various annotations (i.e. if function annotations or executor annotations should take preference).

hkaiser:
- this adds an implementation of the new properties for all execution policies and for the parallel_executor
- adding thread_description to async_launch_policy_dispatch
- flyby: adding constexpr to get_function_address and get_function_name traits
- flyby: fixing annotated_function to rely on annotate_function
- flyby: moving executor tests from execution to executors module
- flyby: modernized implementation of execution policies

Fixes #5269

This enables the following syntax for setting an annotation for execution policies:

```
hpx::transform_reduce(
    hpx::experimental::prefer(
        hpx::execution::experimental::make_with_annotation, 
        hpx::execution::par, "my transform_reduce"), 
    ...);
```

Since almost all execution policies support this property, the following is possible as well:
```
hpx::transform_reduce(
    hpx::execution::experimental::make_with_annotation(
        hpx::execution::par, "my transform_reduce"), 
    ...);
```
Analogous operations are supported for executors. Currently only the `parallel_executor` supports this property, however.


libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
```diff
@@ -140,6 +140,7 @@ namespace hpx { namespace util {
                 Ts...>::type
             operator()(Ts&&... ts)
             {
+                annotate_function annotate(get_function_annotation());
```
msimberg, line 143: I *think* the reason there was no `annotate_function` here is that typically the annotation of an `annotated_function` will already be applied when the task is spawned, meaning that using `annotate_function` here would create a new "task" in the trace with the same name. Leaving it out here of course means that just calling `operator()` on an `annotated_function` not as a task function will not apply the annotation.

Could `annotate_function` only apply the new annotation if it's different (either simply comparing the pointers, or actually comparing the strings)? Maybe it already does that?

libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform.hpp
```diff
@@ -787,7 +787,7 @@ namespace hpx { namespace traits {
     struct get_function_annotation<
         parallel::v1::detail::transform_iteration<ExPolicy, F, Proj>>
     {
-        static char const* call(
+        static constexpr char const* call(
```
msimberg, line 790: Do we know by the way if this actually has an effect on inlining or something like that? I doubt that we ever actually use the string in a constexpr context, or am I missing some other reason to make these `constexpr`?

libs/parallelism/execution/tests/unit/CMakeLists.txt
```diff
@@ -1,25 +1,18 @@
...
-    created_executor
     executor_parameters
     executor_parameters_dispatching
     executor_parameters_timer_hooks
     future_then_executor
     minimal_async_executor
     minimal_sync_executor
-    parallel_executor
-    parallel_fork_executor
-    parallel_policy_executor
```
msimberg, line 18: Thanks for taking care of this, I had definitely missed them.

libs/parallelism/executors/include/hpx/executors/execution_policy_annotation.hpp
```diff
@@ -0,0 +1,49 @@
...
+    template <typename ExPolicy,
+        typename Enable =
+            std::enable_if_t<hpx::is_execution_policy<ExPolicy>::value>>
+    constexpr decltype(auto) tag_invoke(
+        hpx::execution::experimental::make_with_annotation_t, ExPolicy&& policy,
+        char const* annotation)
+    {
+        auto exec = hpx::experimental::prefer(
+            hpx::execution::experimental::make_with_annotation,
+            policy.executor(), annotation);
```
msimberg, line 33: Should this not instead sfinae out cases when `make_with_annotation` doesn't work with the executor? That way one can either prefer the annotation in user code and it'll be silently ignored if one can't set the annotation, but it'll be a compile error if one tries to require it and it can't be set.

msimberg (changes requested):
Nice!

Very much a side-note, but I'm starting to dislike the verbosity of those `make_with_x` functions... Not sure there's much we can do about it right now though (and definitely not in this PR). The `make_` prefix is slightly unnecessary as well.

hkaiser:
> Very much a side-note, but I'm starting to dislike the verbosity of those `make_with_x` functions... Not sure there's much we can do about it right now though (and definitely not in this PR). The `make_` prefix is slightly unnecessary as well.

I agree. I don't like this either. Any ideas are welcome.

libs/parallelism/executors/include/hpx/executors/execution_policy_annotation.hpp
hkaiser: I was thinking about this as well. Do we have a `can_prefer` trait or similar (checking whether prefer would 'succeed')?

libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform.hpp
hkaiser: I was hoping that under certain circumstances the annotation for a function could be evaluated at compile time, for instance:
```
void foo();
void bar()
{
    char const* annotation = hpx::traits::get_function_annotation<&foo>();
}
```
But I have not checked whether this is actually happening.

libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
hkaiser: > I _think_ the reason there was no `annotate_function` here is that typically the annotation of an `annotated_function` will already be applied when the task is spawned, meaning that using `annotate_function` here would create a new "task" in the trace with the same name. Leaving it out here of course means that just calling `operator()` on an `annotated_function` not as a task function will not apply the annotation.

Yes, that was the intend. However, IIRC, `annotated_function` had all kind of problems in the past that APEX didn't properly pick up the annotation. Adding the `annotate_function` in the function operator should fix this. 

> Could `annotate_function` only apply the new annotation if it's different (either simply comparing the pointers, or actually comparing the strings)? Maybe it already does that?

I can certainly do this.

hkaiser:
> > Very much a side-note, but I'm starting to dislike the verbosity of those `make_with_x` functions... Not sure there's much we can do about it right now though (and definitely not in this PR). The `make_` prefix is slightly unnecessary as well.
> 
> I agree. I don't like this either. Any ideas are welcome.

Perhaps dropping the `make_` prefix would already make things better.

libs/parallelism/executors/include/hpx/executors/execution_policy_annotation.hpp
msimberg, line 33: What I had in mind was something like this:

```suggestion
    template <typename ExPolicy,
        typename Enable =
            std::enable_if_t<hpx::is_execution_policy<ExPolicy>::value &&
            hpx::is_invocable<hpx::execution::experimental::make_with_annotation_t,
                typename ExPolicy::executor_type, char const*>::value>>
    constexpr decltype(auto) tag_invoke(
        hpx::execution::experimental::make_with_annotation_t, ExPolicy&& policy,
        char const* annotation)
    {
        auto exec = hpx::execution::experimental::make_with_annotation(
            policy.executor(), annotation);
```

`can_prefer` could probably be useful in some contexts (we don't have it), but would almost always be true (I don't know yet which properties would be useful to never ignore if they can't be applied). I also don't think `can_prefer` would be right in this context.

`can_require` is doable, but kind of weird, because it would essentially be `template <typename Tag, typename... Args> can_require = hpx::is_invocable<Tag, Args...>::value`. I'm not sure if there's any sensible way to constrain it, but if not, I don't see a point in that kind of an alias.

I might be overseeing something though...

msimberg: See my follow-up comment: https://github.com/STEllAR-GROUP/hpx/pull/5341#discussion_r634457296.

libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
msimberg: Thanks! Yep, agreed, it was kind of odd before so probably applying the annotation (even if it's unconditional) is probably the lesser evil.

msimberg:
> > > Very much a side-note, but I'm starting to dislike the verbosity of those `make_with_x` functions... Not sure there's much we can do about it right now though (and definitely not in this PR). The `make_` prefix is slightly unnecessary as well.
> > 
> > 
> > I agree. I don't like this either. Any ideas are welcome.
> 
> Perhaps dropping the `make_` prefix would already make things better.

Agreed. I'll open a separate issue for this though, maybe we'll come up with something better.

libs/parallelism/executors/include/hpx/executors/execution_policy_annotation.hpp
hkaiser: Makes sense. I'll go with `is_tag_invocable` instead of the plain `is_invocable`, though

msimberg: Makes sense! Forgot about that one.

hkaiser:
retest lsu

libs/parallelism/execution/include/hpx/execution/executors/rebind_executor.hpp
```diff
@@ -49,23 +49,42 @@ namespace hpx { namespace parallel { namespace execution {
...
+            using rebound_type =
+                typename rebind_executor<ExPolicy, Executor, Parameters>::type;
+
+            return rebound_type(std::forward<Executor>(exec),
+                std::forward<Parameters>(parameters));
+        }
+    };
+
+    HPX_INLINE_CONSTEXPR_VARIABLE create_rebound_policy_t
+        create_rebound_policy{};
```
msimberg, line 89: I have to ask only because this looks like it could be a plain function template: is this an object instead of a function template to make the friend declarations shorter?

msimberg (approved):
Looks good to me (except for the CI complaints)! Just a minor question, but that's not blocking. Thanks @hkaiser!

aurianer (dismissed):
LGTM thanks!

libs/parallelism/execution/include/hpx/execution/executors/rebind_executor.hpp
hkaiser: > I have to ask only because this looks like it could be a plain function template: is this an object instead of a function template to make the friend declarations shorter?

Yes.

libs/parallelism/executors/include/hpx/executors/execution_policy_annotation.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <utility>
+
+namespace hpx { namespace execution { namespace experimental {
+
+    // make_with_annotation property implementation for execution policies
+    // that simply forwards to the embedded executor
+    template <typename ExPolicy,
+        typename Enable =
+            std::enable_if_t<hpx::is_execution_policy<ExPolicy>::value &&
+                hpx::functional::is_tag_invocable<
```
msimberg, line 27: Considering @K-ballo's comment: https://github.com/STEllAR-GROUP/hpx/issues/5344#issuecomment-844361465, I should note that I think our `is_tag_invocable` does not currently take into account fallback/override implementations. In this particular case it doesn't matter because `make_with_annotation` has neither a fallback nor an override.

msimberg: Depends what you mean by sufficient. If you're ok with checking those separately, yes. If you expect `is_tag_invocable` on its own to say (in this case) if `make_with_x` is valid, it's not. For that one should currently use just `is_invocable`.

hkaiser: In light of #5347 we should leave it as. In order to forcing a sensible compile error I fell back to `tag_fallback`. I agree that this SFINAE here check relies on properties always being implemented in terms of `tag_invoke`.

hkaiser: Adding to the previous... Assuming properties are implemented in terms of `tag_invoke` is not an actual limitation as `prefer` uses it to do the dispatching. However what I think we should do is to introduce a trait that allows to check whether invoking a property (i.e. `require` semantics) would succeed compiling. This would allow to separate concerns without having to expose implementation details of `prefer` to user code as currently done above.

msimberg: Good point about `prefer`! Let's keep it like this here and see what we can do separately. Also thanks for #5347, looks good.

msimberg:
I'd merge this if it weren't for this one: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/7667/workflows/9b28f480-e152-441a-85af-4614df085660/jobs/251189/tests#failed-test-1. CircleCI is the only configuration where it fails, but it's probably just not deterministic. Are you able to reproduce it locally?

hkaiser:
retest lsu

msimberg:
@hkaiser it's `--hpx:debug-hpx-log` that is triggering the segfault, because it wants to print a thread description when destroying a thread. One of those descriptions is not kept alive long enough... Can you reproduce it on your end? In particular, those `desc.c_str()`s look dangerous.

hkaiser:
> @hkaiser it's `--hpx:debug-hpx-log` that is triggering the segfault, because it wants to print a thread description when destroying a thread. One of those descriptions is not kept alive long enough... Can you reproduce it on your end? In particular, those `desc.c_str()`s look dangerous.

Thanks, this is an excellent remark that explains why I wasn't able to reproduce the issue. I'll investigate.

hkaiser:
@msimberg this should be fine now, thanks again for diagnosing the issue.

hkaiser:
retest lsu

libs/core/threading_base/include/hpx/threading_base/register_thread.hpp
```diff
@@ -58,6 +58,12 @@ namespace hpx { namespace threads {
 
                 // run and free all registered exit functions for this thread
                 auto* p = get_self_id_data();
+
+                // reset thread descriptions to avoid dangling pointers
+                util::thread_description desc;
+                p->set_description(desc);
+                p->set_lco_description(desc);
+
```
msimberg, line 66: My only question with this: is the description now just not going to be printed when a thread is destroyed (this should happen before thread destruction as far as I can tell)? With `annotate(d)_function` we assume that c-strings are static strings (without a way to enforce it of course) and for `std::string` we add them to a map to keep them alive. Would the same mechanism be appropriate here?

hkaiser: Where is that map? We can certainly do this. My rationale was that maybe having the thread name available during destruction (which usually is long after the thread has exited) might not be necessary. But I'd be happy to change it to storing `std::string`s in a map, somewhere.

msimberg: It's declared and used a couple of times here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp, and defined here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/threading_base/src/annotated_function.cpp.

hkaiser: Thanks. I have added support for `std::string` for executor annotations. This however only adds it for `with_annotation`, not for `get_annotation`. However this shouldn't be an issue.

msimberg: Thanks! Yes, getting them as c-strings should be fine for the same reason it should be fine to use them indefinitely as annotations.

msimberg:
Thanks again @hkaiser for doing this, let's go ahead with this once CI passes.

hkaiser:
I'll close this as it is fully subsumed by #5347 now.

PR #5342, opened 2021-05-18
Refactor CUDA event polling

msimberg:
Needs #5335.

This PR is mainly intended to refactor the CUDA event polling so that it can be used with senders. On master the polling queues contain a future data pointer, which is not generic enough. I've:
- Refactored the polling to store a generic `unique_function_nonser<void(cudaError_t)>`. The callback is triggered when the status is no longer `cudaErrorNotReady`, and the status is passed to the callback.
  - This works for the old future based polling by capturing the future data pointer in the callback
  - Senders simply add a callback which captures a receiver and calls set_value in the callback
  - This seems to have no impact on the performance of the `cuda_executor_throughput` test
- General refactoring in the CUDA event polling files
  - Separate functionality into a few different headers, clean up unneeded includes
  - Specialize the CUDA `future_data` for the two modes (no need to store the runtime pointer nor the event for event-based futures (the event is stored separately now), no need to store an event for callback-based futures)
- Added `hpx::cuda::experimental::transform_stream` which is a different customization point from `hpx::execution::experimental::transform`. It inserts a stream as the last argument when calling the callable. It synchronizes with events, and skips synchronization if the next receiver is also from `transform_stream` and the stream is the same. This is for sure not going to be enough (cuBLAS support, cuSOLVER support, stream pools, etc. could be added), but this is the minimum to test the event callbacks with senders.
- Changed the member functions of `hpx::lcos::local::condition_variable` to take any type of lock, not just `hpx::lcos::local::mutex`. I think this is fine (as long as the lock is the same, obviously, but that's a problem even when the type is fixed), but I'd be grateful for any comments on if this might be dangerous.
- Changed `sync_wait` to use a `hpx::lcos::local::spinlock` internally for the condition variable. Again, I think this is fine since the condition variable stores internally an `agent_ref` which should dispatch to the correct functionality even if HPX and std threads are mixed, but again, I might be missing something.
  - I did this to not need to do `transform_stream | on(executor) | sync_wait` where the `on` starts an HPX thread context (`transform_stream` calls `set_value` directly in the scheduler)
- Expanded the `synchronize` benchmark to test the sender-based approach.

Typical run below. All runs spawn 1024 dummy kernels and synchronize either immediately, or after 10 spawns (the "batched" variants).
```
native + synchronize (warmup):                                 0.00689671
native + synchronize:                                          0.00653416
native + synchronize batched:                                  0.00255409
executor.sync_execute([](){}):                                 0.00672425
executor.post([](){}) + synchronize:                           0.00309613
executor.post([](){}) + synchronize batched:                   0.00222514
executor.post([](){}) + get_future() callback:                 0.0969046
executor.post([](){}) + get_future() callback batched:         0.0103812
executor.post([](){}) + get_future() event:                    0.00887968
executor.post([](){}) + get_future() event batched:            0.00289038
executor.async_execute([](){}).get():                          0.088615
executor.async_execute([](){}).get() batched:                  0.0842923
transform_stream:                                              0.0106874
transform_stream batched:                                      0.00306984
transform_stream force synchronize batched:                    0.0089994
transform_stream with on:                                      0.0226031
transform_stream with on batched:                              0.00442085
```
`native + synchronize` and `post synchronize` are unsurprisingly almost identical and set the baseline. Callback-based futures are unsurprisingly slower than event-based futures. Surprisingly, `get_future` with events is slightly faster than the sender-based version with events. I don't have an explanation for this yet.


msimberg:
bors try-

libs/full/async_cuda/src/cuda_event_callback.cpp
```diff
@@ -0,0 +1,231 @@
...
+        using hpx::threads::policies::detail::polling_status;
+
+        auto& event_callback_vector = detail::get_event_callback_vector();
+
+        // Don't poll if another thread is already polling
+        std::unique_lock<hpx::cuda::experimental::detail::mutex_type> lk(
+            detail::get_vector_mtx(), std::try_to_lock);
+        if (!lk.owns_lock())
+        {
+            if (cud_debug.is_enabled())
```
aurianer, line 126: can it be turn to a if constexpr? not sure

aurianer (approved):
LGTM thanks!

libs/full/async_cuda/src/cuda_event_callback.cpp
msimberg: Could with C++17, good point! As far as I understood @biddisco verified that this gets optimized out in any case, but constexpr if would guarantee it.

libs/full/async_cuda/src/cuda_event_callback.cpp
```diff
@@ -0,0 +1,257 @@
...
+    void add_to_event_callback_queue(event_callback&& continuation)
+    {
+        HPX_ASSERT_MSG(get_register_polling_count() != 0,
+            "CUDA event polling has not been enabled on any pool. Make sure "
+            "that CUDA event polling is enabled on at least one thread pool.");
+
+        get_event_callback_queue().enqueue(std::move(continuation));
+
+        cud_debug.debug(debug::str<>("event queued"), "event",
+            debug::hex<8>(continuation.event), "callbacks in vector",
```
aurianer, line 80: callbacks in queue?

msimberg: It's using `get_event_callack_vector`, so seems ok to me?

aurianer: Okay I thought you would want to print the size of the queue here as it's the queue that is modified but I might be mistaken

msimberg: I think you're right, it looks like it would make more sense to print the size of the queue!

PR #5338, opened 2021-05-17
Fixing timed_executor specializations of customization points

hkaiser:
This is a leftover from #5331

PR #5343, opened 2021-05-19
Fix timed_executor static cast conversion

srinivasyadav18:
Fixes #5338

## Proposed Changes

  - static cast [result](https://github.com/STEllAR-GROUP/hpx/blob/timed_executors/libs/parallelism/execution/include/hpx/execution/executors/execution_parameters.hpp#L675) to `const*`


hkaiser (approved):
LGTM, thanks!

hkaiser:
add to whitelist

PR #5347, opened 2021-05-20
Remove `make_` prefix for executor properties

### LINKED ISSUE #5345: Rename P2220 property CPOs?
Related to #5341.

https://wg21.link/p2220 proposes to use `tag_invoke` for properties. One of the naming conventions it uses to require a property on something is `make_with_x`. This can unfortunately be quite verbose. E.g.:
```
// p1393
auto new_exec = require(exec, hint);
// p2220
auto new_exec = make_with_hint(exec, hint);
```
The difference is bigger when using prefer:
```
// p1393
auto new_exec = prefer(exec, hint);
// p2220
auto new_exec = prefer(make_with_hint, exec, hint);
```

This is very much a bikeshedding topic, and it's possible that renaming these properties won't make much of a difference, but it's worth at least considering. The only alternatives I can think of at the moment are:
- Drop the `make_` prefix to get e.g. `prefer(with_hint, exec, hint)`
- Use `add` as a prefix, e.g. `prefer(add_hint, exec, hint)`

Renaming is more of a partial fix though. Using `tag_invoke` is inherently more verbose than the P1393 approach. Or?

hkaiser:
Fixes #5345, relies on #5341

- flyby: improve generated (compiler) error messages in case a property is not supported but required.


msimberg:
The `annotation_property` test actually seems to have failed here as well on CircleCI: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/7672/workflows/4a271390-0963-4d10-9754-167f765c8987/jobs/251176, so it may in fact be deterministic. Not sure what could be going wrong though.

hkaiser:
The `annotation_property` test succeeds now. I think this is good to go.

Issue #5344, opened 2021-05-19
Further simplify tag_invoke helpers


msimberg:
Related to #5332. @hkaiser, @K-ballo, I went back to wg21.link/p1895, and @K-ballo is very correct in saying that they did consider the fallback and override cases in the paper. The "Design Discussion" section contains the relevant parts. With #5283 we get a very viable alternative with `if constexpr`. The `tag*` base classes do help reduce boilerplate currently, so it'd be a shame to completely get rid of them, but being able to use `if constexpr` would quite significantly simplify things without having to keep the `fallback`/`override` versions. Until we require C++17 I think they are a reasonable compromise though.

hkaiser:
The `if constexpr` alternative is good only in cases where we want for the base implementation to 'be aware' of all the alternative implementations. The fallback facilities we have is still the only way for cases where the base implementation should be extended by alternatives without changing it (like with the segmented algorithms or the specializations related to `datapar`). 

The fallback and override facilities we have created go beyond what was intended for `std::tag_invoke`. Those are not only helpers to create customization points, but provide viable tools for flexible and extensible overload resolution. 

The main problem I see (as also pointed out by @K-ballo) is that having a `tag_invoke` in HPX might (ADL) conflict with a future `std::tag_invoke`. 

That said, I still believe that our facilities (`tag_invoke`, `tag_fallback_invoke`, and `tag_override_invoke`) are useful and should stay. We should consider finding better names for those, however. That may help avoiding conflicts with a future `std::tag_invoke` and also may also circumvent possible misconceptions from users that assume semantics as described by wg21.link/p1895.

K-ballo:
The uses of `tag_invoke` for the purposes of segment algorithm separation have no relation whatsoever to the proposed `std::tag_invoke` design, none of them are customization points, so in that regard it is unsurprising that extensions were needed to take it in different directions. A less misleading name would definitely help.

In the future we can look into approaching the underlying problem differently.

hkaiser:
As a first step I'd suggest to add detection of `std::tag_invoke` (now or in the future once some libraries actually implement it) and -- if it is available -- disable our `hpx::functional::tag_invoke` API facility. The CPO implementations we have should be usable with a `std::tag_invoke` without change (it's based on ADL anyways).

[...]

PR #5348, opened 2021-05-20
Rename tag_invoke and related facilities to tag_dispatch

hkaiser:
In the light of the recent discussions (see #5344) this proposes to rename `tag_invoke` and friends to `tag_dispatch`, mainly to avoid potential conflicts with a possible future `std::tag_invoke` and to prevent users from misunderstanding the concepts they represent. I think we should make this change now (as in as quickly as possible) to avoid breaking code that users may write in the future. 

This PR is really a proposal and I'm open to discussions (and bike-shedding).


hkaiser:
The quicker we get this merged, the better, I think.

aurianer (dismissed):
LGTM thanks!
Should we also rename the `tag_invoke_*_calls` or `tag_invoke_*_called` in the sender receivers unit tests?

msimberg:
> > LGTM thanks!
> > Should we also rename the `tag_invoke_*_calls` or `tag_invoke_*_called` in the sender receivers unit tests?
> 
> In general we should leave `tag_invoke` wherever we want for a future `std::tag_invoke` to be usable with our functions.

I'm not sure I understand this comment. Have you left any `tag_invoke` overloads unchanged? I think @aurianer's question was simply about the variable names we use in the tests. I think it makes sense to rename them as well as long as we call our "thing" `tag_dispatch` and not `tag_invoke`. Am I missing something?

hkaiser:
> I'm not sure I understand this comment. Have you left any `tag_invoke` overloads unchanged? I think @aurianer's question was simply about the variable names we use in the tests. I think it makes sense to rename them as well as long as we call our "thing" `tag_dispatch` and not `tag_invoke`.

Apparently I misunderstood, sorry. I will make the change.

hkaiser:
@aurianer I have now renamed the uses of `tag_invoke_...` in the tests as well. @msimberg This should be good to go.

hkaiser:
> Thanks @hkaiser. What about the discussion from yesterday about keeping the `tag_invoke` name for the main facility? Did you get some clarity on that?

I think now that we should rename `tag_invoke` as well. The rationale is that if we renamed `tag_fallback_invoke` (and friends) only, then we would have to implement base functionalities by overloading `tag_fallback_dispatch` but specializations for the same DPO would have to be done using `tag_invoke`. That doesn't sound too appealing to me and would be difficult to explain.

msimberg:
> I think now that we should rename `tag_invoke` as well. The rationale is that if we renamed `tag_fallback_invoke` (and friends) only, then we would have to implement base functionalities by overloading `tag_fallback_dispatch` but specializations for the same DPO would have to be done using `tag_invoke`. That doesn't sound too appealing to me and would be difficult to explain.

Good point, and agreed that the names should be consistent if possible. I suppose the question was more about why we need to rename `tag_invoke`, but as I said I don't fully understand the issue, so let's go ahead with this if you think it's the right thing to do.

hkaiser:
> > I think now that we should rename `tag_invoke` as well. The rationale is that if we renamed `tag_fallback_invoke` (and friends) only, then we would have to implement base functionalities by overloading `tag_fallback_dispatch` but specializations for the same DPO would have to be done using `tag_invoke`. That doesn't sound too appealing to me and would be difficult to explain.
> 
> Good point, and agreed that the names should be consistent if possible. I suppose the question was more about why we need to rename `tag_invoke`, but as I said I don't fully understand the issue, so let's go ahead with this if you think it's the right thing to do.

It does not hurt... , at least I can't see it getting in the way.

msimberg:
> It does not hurt... , at least I can't see it getting in the way.

Yes, the only potential downside is that we'd have rename things back if we get a std::tag_invoke and we want to make things work with that. However, that's far off in the future, if it ever happens. Let's do this. This just needs a formatting round as well, but let's merge it once that's done.

msimberg:
Thanks @hkaiser for updating this as well. Since this, #5341, and #5347 will conflict (or will at least need changes) after one of them is merged, I'll let you merge them in whichever order you see as easiest.

PR #5353, opened 2021-05-26
Add CUDA timestamp support to HPX Hardware Clock

NK-Nikunj:
This PR provides support for timestamps in CUDA kernels. The output is received in nanosecond duration and more information can be found [here](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#special-registers-globaltimer).

I'm adding CUDA support so that I can use timestamp instead of relying on asm directly within my hpx-kokkos-resilience project.

### DIFF
--- a/libs/core/hardware/CMakeLists.txt
+++ b/libs/core/hardware/CMakeLists.txt
@@ -27,7 +27,6 @@ set(hardware_compat_headers
     hpx/util/hardware/cpuid/linux_x86.hpp => hpx/modules/hardware.hpp
     hpx/util/hardware/cpuid/msvc.hpp => hpx/modules/hardware.hpp
     hpx/util/hardware/timestamp/bgq.hpp => hpx/modules/hardware.hpp
-    hpx/util/hardware/timestamp/cuda.hpp => hpx/modules/hardware.hpp
     hpx/util/hardware/timestamp/linux_generic.hpp => hpx/modules/hardware.hpp
     hpx/util/hardware/timestamp/linux_x86_32.hpp => hpx/modules/hardware.hpp
     hpx/util/hardware/timestamp/linux_x86_64.hpp => hpx/modules/hardware.hpp
--- a/libs/core/hardware/CMakeLists.txt
+++ b/libs/core/hardware/CMakeLists.txt
@@ -11,6 +11,7 @@ set(hardware_headers
     hpx/hardware/cpuid/linux_x86.hpp
     hpx/hardware/cpuid/msvc.hpp
     hpx/hardware/timestamp/bgq.hpp
+    hpx/hardware/timestamp/cuda.hpp
     hpx/hardware/timestamp/linux_generic.hpp
     hpx/hardware/timestamp/linux_x86_32.hpp
     hpx/hardware/timestamp/linux_x86_64.hpp
@@ -26,6 +27,7 @@ set(hardware_compat_headers
     hpx/util/hardware/cpuid/linux_x86.hpp => hpx/modules/hardware.hpp
     hpx/util/hardware/cpuid/msvc.hpp => hpx/modules/hardware.hpp
     hpx/util/hardware/timestamp/bgq.hpp => hpx/modules/hardware.hpp
+    hpx/util/hardware/timestamp/cuda.hpp => hpx/modules/hardware.hpp
     hpx/util/hardware/timestamp/linux_generic.hpp => hpx/modules/hardware.hpp
     hpx/util/hardware/timestamp/linux_x86_32.hpp => hpx/modules/hardware.hpp
     hpx/util/hardware/timestamp/linux_x86_64.hpp => hpx/modules/hardware.hpp
@@ -47,6 +49,7 @@ add_hpx_module(
     "hpx/hardware/cpuid/linux_x86.hpp"
     "hpx/hardware/cpuid/msvc.hpp"
     "hpx/hardware/timestamp/bgq.hpp"
+    "hpx/hardware/timestamp/cuda.hpp"
     "hpx/hardware/timestamp/linux_generic.hpp"
     "hpx/hardware/timestamp/linux_x86_32.hpp"
     "hpx/hardware/timestamp/linux_x86_64.hpp"
--- a/libs/core/hardware/include/hpx/hardware/timestamp.hpp
+++ b/libs/core/hardware/include/hpx/hardware/timestamp.hpp
@@ -10,12 +10,7 @@
 
 #include <hpx/config.hpp>
 
-// clang-format off
-// nvcc complains about multiple definition error for same function prototype
-// irrespective of __host__, __device__, and __host__ __device__ attributes. 
-#if defined(HPX_HAVE_CUDA) && defined(__CUDA_ARCH__)
-    #include <hpx/hardware/timestamp/cuda.hpp>
-#elif defined(HPX_MSVC)
+#if defined(HPX_MSVC)
   #include <hpx/hardware/timestamp/msvc.hpp>
 #elif defined(__amd64__) || defined(__amd64) || defined(__x86_64__) ||         \
     defined(__x86_64) || defined(_M_X64)
--- a/libs/core/hardware/include/hpx/hardware/timestamp.hpp
+++ b/libs/core/hardware/include/hpx/hardware/timestamp.hpp
@@ -10,6 +10,7 @@
 
 #include <hpx/config.hpp>
 
+// clang-format off
 #if defined(HPX_MSVC)
   #include <hpx/hardware/timestamp/msvc.hpp>
 #elif defined(__amd64__) || defined(__amd64) || defined(__x86_64__) ||         \
--- a/libs/core/hardware/include/hpx/hardware/timestamp.hpp
+++ b/libs/core/hardware/include/hpx/hardware/timestamp.hpp
@@ -11,7 +11,11 @@
 #include <hpx/config.hpp>
 
 // clang-format off
-#if defined(HPX_MSVC)
+// nvcc complains about multiple definition error for same function prototype
+// irrespective of __host__, __device__, and __host__ __device__ attributes. 
+#if defined(HPX_HAVE_CUDA) && defined(__CUDA_ARCH__)
+    #include <hpx/hardware/timestamp/cuda.hpp>
+#elif defined(HPX_MSVC)
   #include <hpx/hardware/timestamp/msvc.hpp>
 #elif defined(__amd64__) || defined(__amd64) || defined(__x86_64__) ||         \
     defined(__x86_64) || defined(_M_X64)
--- a/libs/core/hardware/include/hpx/hardware/timestamp/bgq.hpp
+++ b/libs/core/hardware/include/hpx/hardware/timestamp/bgq.hpp
@@ -17,11 +17,20 @@
 
 #include <cstdint>
 
+#include <hpx/config.hpp>
+#if defined(HPX_HAVE_CUDA)
+#include <hpx/hardware/timestamp/cuda.hpp>
+#endif
+
 namespace hpx { namespace util { namespace hardware {
 
-    inline std::uint64_t timestamp()
+    HPX_HOST_DEVICE std::uint64_t timestamp()
     {
+#if defined(HPX_HAVE_CUDA) && defined(__CUDA_ARCH__)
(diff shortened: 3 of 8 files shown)


libs/core/hardware/include/hpx/hardware/timestamp/linux_generic.hpp
```diff
@@ -14,11 +14,20 @@
 
 namespace hpx { namespace util { namespace hardware {
 
+#if defined(HPX_HAVE_CUDA)
+    HPX_HOST_DEVICE
+#endif
```
msimberg, line 19: The `#if defined` around the macro is not necessary. That's why `HPX_HOST_DEVICE` is a macro.

msimberg (commented):
Do you have an example of in which kinds of situations this is useful?

libs/core/hardware/include/hpx/hardware/timestamp/linux_generic.hpp
NK-Nikunj: Aah, I thought it appends it all situations. I'll correct that.

NK-Nikunj:
would you want a different function, perhaps `timestamp_cuda` or something?

hkaiser:
> would you want a different function, perhaps `timestamp_cuda` or something?

I think we need an alternative implementation for CUDA integrated here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/hardware/include/hpx/hardware/timestamp.hpp

libs/core/hardware/include/hpx/hardware/timestamp.hpp
```diff
@@ -11,7 +11,11 @@
 #include <hpx/config.hpp>
 
 // clang-format off
-#if defined(HPX_MSVC)
+// nvcc complains about multiple definition error for same function prototype
+// irrespective of __host__, __device__, and __host__ __device__ attributes.
```
msimberg, line 15: Could you explain what the problem is here?

libs/core/hardware/include/hpx/hardware/timestamp/cuda.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+#pragma once
+
+#include <cuda.h>
+
+#include <hpx/config.hpp>
+
+namespace hpx { namespace util { namespace hardware {
+
+    HPX_HOST_DEVICE std::uint64_t timestamp()
```
msimberg, line 17: This `HPX_HOST_DEVICE` doesn't look right, it should just be `HPX_DEVICE`, but then I guess it's related to the comment above?

libs/core/hardware/include/hpx/hardware/timestamp.hpp
NK-Nikunj: I just realized that this implementation won't work either. Let me clarify what I meant:
clang and nvcc views same function prototypes with different attributes (__host__, __device__ or __host__ __device__) differently. For clang, `__host__ void foo()` and `__device__ void foo()` are 2 different function prototypes. But nvcc views them as the same function despite having 2 completely different attributes. nvcc in these cases throws multiple definition errors and the `#if defined(__CUDA_ARCH__)` is used to achieve a similar scenario as clang. So, with nvcc, you would defined one function foo and use the macro to check whether the current execution path is on GPU. I used this technique in the previous commit where I define the function HPX_HOST_DEVICE and use the macro to define 2 separate implementation paths depending on GPU/CPU execution. I realized that the macro won't help here as it's not used within a function.
I'll have to come up with something that integrates well with the current system and doesn't lead to multiple declaration error with nvcc. I'll update the PR again with that.

libs/core/hardware/include/hpx/hardware/timestamp/cuda.hpp
NK-Nikunj: In theory, it should work for both but HPX_DEVICE is more appropriate here. Let me update the whole function again so that it caters to the problem I described above.

NK-Nikunj:
@hkaiser I don't think we can add an implementation directly to [timestamp header](https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/hardware/include/hpx/hardware/timestamp.hpp).
nvcc treats same function signatures with different attributes (__host__ or __device__) as a single definition. So, I can't have a different timestamp function implementation to cater specifically for function calls from __device__ kernels. What I could do (and what I did in the first commit) is to have a different implementation inside of the timestamp function using __CUDA_ARCH__ macro. Otherwise, we'll have to rename the variant to something like timestamp_cuda or send it to some other namespace so that the compiler doesn't throw multiple definition errors.

msimberg:
@NK-Nikunj in that case maybe something like this would make sense, to at least not have to duplicate the CUDA implementation:

```
#include <.../timestamp_cuda.hpp>
/...
HPX_HOST_DEVICE std::uint64_t timestamp() {
#ifdef __CUDA__
    return timestamp_cuda();
#else
    // CPU implementation
#endif
}
```
This would have to be repeated for the different platforms but at least the body won't have to be repeated. I might be missing some corner case with this though. The platform specific functions could also be renamed and you would instead do
```
HPX_HOST_DEVICE std::uint64_t timestamp() {
#ifdef __CUDA__
    return timestamp_cuda();
#elifdef PLATFORM1
    // CPU implementation
    return timestamp_platform1();
#elifdef PLATFORM2
    // Another CPU implementation
#endif
}
```

NK-Nikunj:
I think I'll go with the first approach. The 2nd one would simply make the function unreadable IMO.

If it's fine with by you, I'll make a commit.

NK-Nikunj:
@msimberg could you re-trigger the tests?

libs/core/hardware/CMakeLists.txt
```diff
@@ -26,6 +27,7 @@ set(hardware_compat_headers
     hpx/util/hardware/cpuid/linux_x86.hpp => hpx/modules/hardware.hpp
     hpx/util/hardware/cpuid/msvc.hpp => hpx/modules/hardware.hpp
     hpx/util/hardware/timestamp/bgq.hpp => hpx/modules/hardware.hpp
+    hpx/util/hardware/timestamp/cuda.hpp => hpx/modules/hardware.hpp
```
msimberg, line 30: This is not necessary (the file is new so there's no need for a compatibility header).

libs/core/hardware/include/hpx/hardware/timestamp/linux_generic.hpp
```diff
@@ -12,13 +12,23 @@
 
 #include <time.h>
 
+#include <hpx/config.hpp>
+
+#if define(HPX_HAVE_CUDA)
```
msimberg, line 17:
```suggestion
#if defined(HPX_HAVE_CUDA)
```

libs/core/hardware/include/hpx/hardware/timestamp/linux_generic.hpp
```diff
@@ -12,13 +12,23 @@
...
+#include <hpx/hardware/timestamp/cuda.hpp>
+#endif
+
 namespace hpx { namespace util { namespace hardware {
 
-    inline std::uint64_t timestamp()
+    HPX_HOST_DEVICE std::uint64_t timestamp()
     {
+#if defined(HPX_HAVE_CUDA) && defined(__CUDA_ARCH__)
+        return timestamp_cuda();
```
msimberg, line 26: Is there any reason not to add this to the bgq and msvc files? We don't test them with CUDA, but I don't see any reason why it shouldn't work there (modulo compiler bugs).

msimberg (changes requested):

libs/core/hardware/include/hpx/hardware/timestamp/linux_generic.hpp
NK-Nikunj: I haven't tested them with bgq or msvc. So, I decided to play safe and not add it there.

NK-Nikunj:
@hkaiser any idea what's causing the linker error?

libs/core/hardware/CMakeLists.txt
hkaiser: @NK-Nikunj Would you be able to fix this before the release?

hkaiser:
retest lsu

libs/core/hardware/CMakeLists.txt
NK-Nikunj: yes, I'm currently working on it. I believe we can get it into the release.

NK-Nikunj:
Looks like all the tests are passing now (I'll fix the clang-format issue in a bit). The error message on sc15 nodes was a bit misleading. The issue arose because HPX_HAVE_CUDA checks if building cuda modules is enabled. Therefore, if I use that to include a file that uses Cuda hardware registers (through inline assembly; like in our case) but is compiled for host (i.e. doesn't use `-x cuda` or similar nvcc flags) then the compiler is unable to identify the said register. That's why we were observing those errors.

Adding `__CUDACC__` ensures that the code is included only if the source is compiled for Cuda.

libs/core/hardware/include/hpx/hardware/timestamp.hpp
```diff
@@ -10,7 +10,6 @@
 
 #include <hpx/config.hpp>
 
-// clang-format off
 #if defined(HPX_MSVC)
```
msimberg, line 13: This should probably stay.

libs/core/hardware/include/hpx/hardware/timestamp/bgq.hpp
```diff
@@ -17,11 +17,21 @@
 
 #include <cstdint>
 
+#include <hpx/config.hpp>
```
msimberg, line 20: Could you put this include above the standard library includes. This is consistent with how we do it elsewhere.

libs/core/hardware/include/hpx/hardware/timestamp/bgq.hpp
```diff
@@ -17,11 +17,21 @@
...
+
+#if defined(HPX_HAVE_CUDA) && defined(__CUDACC__)
```
msimberg, line 22:
```suggestion
#if defined(HPX_HAVE_CUDA) && defined(HPX_COMPUTE_CODE)
```

msimberg: I would also maybe move the ifdef inside the header file.

msimberg: And move this include also above the `<cstdint>` include.

libs/core/hardware/include/hpx/hardware/timestamp/bgq.hpp
```diff
@@ -17,11 +17,21 @@
...
+#include <hpx/hardware/timestamp/cuda.hpp>
+#endif
+
 namespace hpx { namespace util { namespace hardware {
 
-    inline std::uint64_t timestamp()
+    HPX_HOST_DEVICE inline std::uint64_t timestamp()
     {
+#if defined(HPX_HAVE_CUDA) && defined(__CUDA_ARCH__)
```
msimberg, line 30:
```suggestion
#if defined(HPX_HAVE_CUDA) && defined(HPX_COMPUTE_DEVICE_CODE)
```

libs/core/hardware/include/hpx/hardware/timestamp/cuda.hpp
```diff
@@ -0,0 +1,23 @@
...
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#include <hpx/config.hpp>
+#include <cstdint>
```
msimberg, line 12:
```suggestion

#include <cstdint>
```

libs/core/hardware/include/hpx/hardware/timestamp/bgq.hpp
msimberg: And the same comments apply to the other files...

msimberg (changes requested):
Thanks @NK-Nikunj for looking into this.

Sorry for the additional comments, but they are just cleanup. Let's get this in after these final changes.

PR #5351, opened 2021-05-24
Adding enable_logging/disable_logging API functions

hkaiser:
- flyby: simplify logging initialization, moving code to source files

@diehlpk you should be able to use this to enable logging from the command line and disable it first thing in `hpx_main` (add `--hpx:debug-hpx-log=<filename>` on the command line, and insert a `hpx::util::disable_logging(hpx::destination_hpx);` in `hpx_main`). This might help diagnosing the startup hangs on summit.

PR #5354, opened 2021-05-26
Making sure default_stack_size is defined whenever declared

hkaiser:
Fixes problems as described in #5351

### DIFF
--- a/libs/core/coroutines/include/hpx/coroutines/detail/context_posix.hpp
+++ b/libs/core/coroutines/include/hpx/coroutines/detail/context_posix.hpp
@@ -54,7 +54,9 @@
 #endif
 
 #if defined(__FreeBSD__) ||                                                    \
-    (defined(_XOPEN_UNIX) && defined(_XOPEN_VERSION) && _XOPEN_VERSION >= 500)
+    (defined(_XOPEN_UNIX) && defined(_XOPEN_VERSION) &&                        \
+        _XOPEN_VERSION >= 500) ||                                              \
+    defined(__bgq__) || defined(__powerpc__) || defined(__s390x__)
 
 // OS X 10.4 -- despite passing the test above -- doesn't support
 // swapcontext() et al. Use GNU Pth workalike functions.
--- a/libs/core/coroutines/src/detail/context_posix.cpp
+++ b/libs/core/coroutines/src/detail/context_posix.cpp
@@ -5,12 +5,22 @@
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 #include <hpx/config.hpp>
+#include <hpx/coroutines/detail/context_impl.hpp>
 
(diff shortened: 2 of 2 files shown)


diehlpk:
@hkaiser That fixed the linking error reported in #5351

libs/core/coroutines/src/detail/context_posix.cpp
```diff
@@ -7,9 +7,10 @@
 #include <hpx/config.hpp>
 
 #if !defined(HPX_HAVE_GENERIC_CONTEXT_COROUTINES) &&                           \
-    !(defined(__linux) || defined(linux) || defined(__linux__)) &&             \
+    !defined(HPX_HAVE_FIBER_BASED_COROUTINES) &&                               \
```
msimberg, line 10: Don't the checks for the linux definitions have to stay?

hkaiser: On Summit both macros, `__linux__` and `__powerpc__` are defined...

hkaiser: @msimberg I have changed the conditions to match the ones used in context_impl.hpp. Things should work everywhere now.

hkaiser: @msimberg yep, clang-newest is happy now as well. @diehlpk would you mind testing this one more time on summit?

msimberg (dismissed):
Looks good, thanks!

libs/core/coroutines/include/hpx/coroutines/detail/context_posix.hpp
```diff
@@ -340,7 +342,7 @@ namespace hpx { namespace threads { namespace coroutines {
 #if defined(HPX_HAVE_THREADS_GET_STACK_POINTER)
                 return get_stack_ptr() - reinterpret_cast<std::size_t>(m_stack);
 #else
-                return (std::numeric_limits<std::ptrdiff_t>::max)();
+                return (std::numeric_limits<std::ptrdiff_t>::max) ();
```
msimberg, line 345: This just needs reverting.

hkaiser: Darn, different versions of clang-format still do different things...

msimberg: Are you using 11? 9 has some questionable formatting choices at least in the sender/receiver files, and I wouldn't mind bumping that up at least to 10 (and I don't mind 11 either).

hkaiser: Yes, I'm using V11 here. I'd be fine with bumping the version. This will cause various reformattings, however.

msimberg: Ok, let's look into bumping it, but separately from this PR.

PR #5356, opened 2021-05-27
Update API documentation

msimberg:
- Adds `hpx/unwrap.hpp` to the public API page and moves `hpx::util::unwrap*` to `hpx::unwrap*` because it's very commonly used functionality.
- Moves all API headers that were in (non-include) modules to `include` or `include_local`. I previously eagerly put the API headers in modules where possible, but I think it's much easier to have an overview if they're exclusively in the `include` modules.
- Adds the `hpx/local/x.hpp` headers to the API page. These exclusively come from `include_local`. I already added these headers in #5312, and I think they make sense even if `hpx/x.hpp` would be local only, because then local-only HPX users don't have to wonder in which cases there is an equivalent `hpx/local/x.hpp` header.
- Adds missing `hpx/distributed/x.hpp` headers. These I'm less sure about. I had added a few of these earlier, so I completed the set here, but I'm actually thinking we can remove them. There's almost no case where `hpx/distributed/x.hpp` wouldn't include `hpx/local/x.hpp` (directly or indirectly), and in that case `hpx/x.hpp` would be practically identical to `hpx/distributed/x.hpp`. Opinions? Is there a case for leaving `hpx/distributed/x.hpp` (in the API, they may still be useful for internal organization, but that's a separate issue).
- I tentatively made the `include` module be enabled only with the distributed runtime enabled. I've not tested it thoroughly so something is likely to break on the first try. However, this makes it easier to check that local tests and examples don't include headers from the `include` module.

Edit: There are four API headers that I didn't do any changes to here: `hpx/wrap_main.hpp` (there's no equivalent functionality for the local runtime, and I'm not planning to add it), `hpx/assert.hpp` (so fundamental that I don't know if separate headers make sense; then again `hpx/local/assert.hpp` would be consistent with the rest), `hpx/version.hpp` (only one version, might make sense to have `hpx/local/version.hpp` with a split), `hpx/init.hpp` (I'm not yet sure if `hpx/local/init.hpp` is the right move; in particular I don't like that the functionality is in `hpx::local` whereas it's the other way around in other cases, i.e. distributed functionality is in `hpx::distributed`; then again the header itself does make sense).


msimberg:
Any opinions on the `hpx/distributed/x.hpp` headers? If not, I'll go ahead and remove them (before this PR there were only around three of them).

This also includes some fixes to init/start that this revealed. `hpx::start` was still using `hpx::util::function_nonser` in the API. Additionally, the local init/start and the normal init overloads were not skipping the entry point when passing `nullptr_t`.

aurianer (dismissed):
LGTM thanks!
No opposition to drop the hpx/distributed/.. headers as they are unlikely to be included on their own :+1:

msimberg:
> No opposition to drop the hpx/distributed/.. headers as they are unlikely to be included on their own +1

I've removed the hpx/distributed headers now. I also attempted to revert or prettify some of the slightly ugly `options_description` formatting that I previously applied.

msimberg:
If there are no further comments on this I'd very much like to have this in 1.7.0-rc2. Are there further comments?

PR #5357, opened 2021-05-27
Making sure collective operations can be reused by preallocating communicator

hkaiser:
- fixing issues with `client`/`client_base` allowing to propagate names registered with AGAS that are associated with a client, fixing lifetime management
- adding `transfer_result` to `future_access` trait
- change `set()` API exposed by `and_gate`, it now forces lock to be moved into the function
- removing overloads of collective operations that rely on futures as those can easily be emulated using `future::then` (this is a breaking change without compatibility option)
- moving all collective operations into namespace `hpx::collectives` (this is a breaking change without compatibility option)
- adding `hpx::collectives::reduce`, `hpx::collectives::inclusive_scan`, `hpx::collectives::exclusive_scan`
- adding new `hpx::collectives::create_communicator` usable with all collective operations
- flyby: remove obsolete `reset_registered_name()` API function from `client_base`
- flyby: working around an #include conflict on recent Windows toolkits
- flyby: modernize future access traits and packaged_continuation facilities

### DIFF
--- a/examples/1d_stencil/1d_stencil_8.cpp
+++ b/examples/1d_stencil/1d_stencil_8.cpp
@@ -11,11 +11,11 @@
 
 #include <hpx/config.hpp>
 #if !defined(HPX_COMPUTE_DEVICE_CODE)
-#include <hpx/hpx_init.hpp>
 #include <hpx/hpx.hpp>
+#include <hpx/hpx_init.hpp>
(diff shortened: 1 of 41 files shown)


hkaiser:
retest lsu

hkaiser:
If there aren't any objections, I'd like to go ahead and merge this PR.

PR #5359, opened 2021-05-28
Fix check for pu-step in single-core case

### LINKED ISSUE #5358: HPX init fails for single-core environments
## Expected Behavior

It should be possible to launch a HPX application on systems with only one CPU core/PU available.

## Actual Behavior

The initialization of HPX fails with:
> hpx::init: std::exception caught: Invalid command line option --hpx:pu-step, value must be non-zero and smaller than number of available processing units.

The problem can be reproduced by launching a HPX application after taking all but one CPU offline or by limiting the number of PUs in an batch environment, which is how I encountered the issue in the first place. I used the `os_thread_num` example in conjunction with `srun -N 1 -n 1 -c 1` to reproduce the issue.

## Specifications

  - HPX Version: 1.7.0/master
  - Platform (compiler, OS): GCC 9.3, Linux, SLURM

severinstrobl:
Fixes #5358

### DIFF
--- a/libs/full/command_line_handling/src/command_line_handling.cpp
+++ b/libs/full/command_line_handling/src/command_line_handling.cpp
@@ -611,7 +611,8 @@ namespace hpx { namespace util {
 
     void command_line_handling::check_pu_step() const
     {
-        if (pu_step_ == 0 || pu_step_ >= hpx::threads::hardware_concurrency())
+        if (hpx::threads::hardware_concurrency() > 1 &&
+            (pu_step_ == 0 || pu_step_ >= hpx::threads::hardware_concurrency()))
         {
             throw hpx::detail::command_line_error(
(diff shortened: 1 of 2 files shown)


hkaiser (dismissed):
LGTM, thanks!

hkaiser:
add to whitelist

hkaiser:
@severinstrobl Thanks a lot!

hkaiser:
@severinstrobl could you please apply the same fix here as well: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/full/command_line_handling_local/src/command_line_handling_local.cpp#L425

severinstrobl:
@hkaiser Certainly, I completely missed that.

hkaiser (approved):
LGTM, thanks! This is good to go now - this time for sure.

PR #5361, opened 2021-06-01
Fix compilation with MPI enabled

msimberg:
This was missed due to the tests not running on rostam. Let's wait for those to actually run correctly before merging this.


aurianer (approved):
LGTM thanks!

PR #5362, opened 2021-06-02
Adding channel_communicator 

hkaiser:
- ~adding distributed one_element_channel~
- flyby: Modernize channel implementations
- flyby: Modernize promise, packaged_task, and various traits
- flyby: hpx::wait_all: modernize and fix problem

### DIFF
--- a/libs/full/collectives/include/hpx/collectives/all_to_all.hpp
+++ b/libs/full/collectives/include/hpx/collectives/all_to_all.hpp
@@ -42,8 +42,9 @@ namespace hpx { namespace collectives {
     hpx::future<std::vector<std::decay_t<T>>>
     all_to_all(char const* basename, T&& result,
         std::size_t num_sites = std::size_t(-1),
+        std::size_t this_site = std::size_t(-1),
         std::size_t generation = std::size_t(-1),
-        std::size_t this_site = std::size_t(-1), std::size_t root_site = 0);
+        std::size_t root_site = 0);
 
     /// AllToAll a set of values from different call sites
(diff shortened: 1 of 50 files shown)


libs/full/collectives/include/hpx/collectives/all_to_all.hpp
```diff
@@ -42,8 +42,9 @@ namespace hpx { namespace collectives {
     hpx::future<std::vector<std::decay_t<T>>>
     all_to_all(char const* basename, T&& result,
         std::size_t num_sites = std::size_t(-1),
+        std::size_t this_site = std::size_t(-1),
         std::size_t generation = std::size_t(-1),
-        std::size_t this_site = std::size_t(-1), std::size_t root_site = 0);
+        std::size_t root_site = 0);
```
msimberg, line 47: This looks very easy to get wrong at the call site. Would it be worth considering some types for those parameters (not for this PR)? Just putting the idea out there.

msimberg (approved):

libs/full/collectives/include/hpx/collectives/all_to_all.hpp
hkaiser: Yes, I agree. I did think of this as well. Let me introduce such types now (before the release), so we don't have to change it later on.

hkaiser: That's in place now. Please re-review.

msimberg (approved):
Thanks!

PR #5365, opened 2021-06-02
Use ccache for Jenkins builds on Piz Daint

msimberg:


msimberg:
This actually already seems to be working. Based on the three builds so far I'm not sure this is a win, but I'd like to give it a try on master for a while. It's possible that a cache hit on a slow scratch system is about as fast as rebuilding everything in `/dev/shm`... Check "Show build time graph" here for the builds so far: https://cdash.cscs.ch/buildSummary.php?buildid=163514.

aurianer (approved):
LGTM thanks!

PR #5367, opened 2021-06-02
Disable pkgconfig by default again on macOS

msimberg:
I unintentionally re-enabled the build unit test with pkgbuild on macOS in #5330. This turns it off by default on macOS.


hkaiser (approved):
LGTM, thanks!

PR #5373, opened 2021-06-07
More changes to clang-cuda Jenkins configuration

msimberg:
On the clang-cuda configuration the build unit test hangs when mixing Debug and... Debug. Other mixes are fine. I have no idea why this is the case. The non-Cray clang 12 configuration has no problems with that particular mix.


aurianer (approved):
LGTM thanks!

PR #5374, opened 2021-06-07
Remove unnecessary moves from future sender implementations

msimberg:


libs/parallelism/execution/include/hpx/execution/algorithms/keep_future.hpp
```diff
@@ -40,11 +40,13 @@ namespace hpx { namespace execution { namespace experimental {
...
-                            hpx::execution::experimental::set_value(
-                                std::move(receiver), std::move(future));
-                        });
+                    // The operation state has to be kept alive until set_value
+                    // is called, which means that we don't need to move
+                    // receiver and future into the on_completed callback.
+                    state->set_on_completed([this]() mutable {
+                        hpx::execution::experimental::set_value(
+                            std::move(receiver), std::move(future));
+                    });
```
hkaiser, line 49: Hmmm. Please excuse that I may not fully understand how the lifetime of the objects is controlled. However I think that this change has made things worse. Is there a guarantee that the object referenced by `this` is still alive when the lambda is eventually executed? Would it be better to move the `sender` and `receiver` objects into the lambda capture? This would ensure that those are still around at the point when it is run.

msimberg: The two choices are what's there now, and what you propose. In any other situation I would go with the explicit move, but in this case it is guaranteed that this will be alive and my preference is to leave out the moves (but the comment is a must!). The comment above is the very brief version of what's in section 2.2.5 in http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p0443r14.html. I can certainly expand the comment and add an explicit reference to p0443 as well.

hkaiser: Just the fact that the `operation_state` instance is kept alive until `set_value` is being called doesn't guarantee that it will be still alive whenever the lambda is actually being run (the future might become ready much later), does it?

msimberg: How so? I feel like I must be missing something obvious... `receiver` and `future` (members of `this`) are moved *before* `set_value` is called.

msimberg: I should perhaps explicitly state that I mean `hpx::execution::experimental::set_value`, not `set_value` on the future shared state...

hkaiser: > How so? I feel like I must be missing something obvious... `receiver` and `future` (members of `this`) are moved _before_ `set_value` is called.

Yes, but they are moved inside a lambda which might run only long after `state->set_on_completed` has returned. What I'm concerned about is that the captured `this` pointer might not be valid anymore at that point.

msimberg: If `this` stays alive at least until `set_value` inside the lambda is called, `this` *must* stay alive until the lambda is called since the lambda is called before `set_value` is called. I must still be missing something...

hkaiser: Ok, understand. Sorry for being dense here.

msimberg: No worries, it wouldn't be the first time I've missed something. Thanks!

aurianer (approved):
LGTM thanks!

PR #5375, opened 2021-06-08
Slight performance improvement for hpx::copy and hpx::move et.al.

hkaiser:
- this improves the detection of contiguous trivially copyable sequences
  for which copy and move can fall back to memcpy
- flyby: adding traits::is_contiguous_iterator
- flyby: modernize transfer algorithm implementations

The sequential version gives me about 40% speedup when copying `vector<>`s of trivially copyable types.

### DIFF
--- a/libs/core/iterator_support/include/hpx/iterator_support/traits/is_iterator.hpp
+++ b/libs/core/iterator_support/include/hpx/iterator_support/traits/is_iterator.hpp
@@ -15,6 +15,7 @@
 #include <iterator>
 #include <type_traits>
 #include <utility>
+#include <vector>
 
 namespace hpx { namespace traits {
     ///////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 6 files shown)


msimberg (dismissed):
Looks good to me! My only question is: would we be better off actually using memcpy for trivially copyable types (apologies if I missed it in these changes)?

hkaiser:
> Looks good to me! My only question is: would we be better off actually using memcpy for trivially copyable types (apologies if I missed it in these changes)?

We do use memmove here: https://github.com/STEllAR-GROUP/hpx/pull/5375/files#diff-f3ce576e728f2a80f45976bad564e6f3bcc59596662509a07a05f370c9b5db1dR91. We did use it before, just the detection logic was flawed. That is what this PR fixes, actually - I should have been more concise when describing the changes.

hkaiser:
rebuild lsu

PR #5377, opened 2021-06-09
More precompiled headers

msimberg:
- Adds Asio and Boost headers to the precompiled headers target
- Does some small rearranging to reduce Asio includes
- ifdefs out more of thread local storage when it's disabled

Regarding the first point, `boost/config.hpp` was problematic. We force `hpx/config.hpp` to be included before `boost/config.hpp`. However, I can't find any reasoning about *why* that's the case. It looks like the current check was introduced in https://github.com/STEllAR-GROUP/hpx/commit/a370734ce21e700aa5e7a7e9a119865178b74f47, and the accompanying intel compiler config is gone since then. Does anyone know? Is it also a problem on Windows?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -322,69 +322,95 @@ hpx_option(
 if(HPX_WITH_PRECOMPILED_HEADERS)
   set(HPX_WITH_PRECOMPILED_HEADERS_INTERNAL ON)
 
+  set(system_precompiled_headers
+      <algorithm>
+      <array>
+      <atomic>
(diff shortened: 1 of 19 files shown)


aurianer (dismissed):
LGTM thanks!

msimberg:
@hkaiser, @sithhell anything you know about that requirement to include `hpx/config.hpp` before `boost/config.hpp`? If not, I'd like to go ahead with this and deal with the consequences if and when they come up again. Reverting this change is not very complicated.

Issue #5378, opened 2021-06-09
HPX 1.7.0 stopped building on Fedora


diehlpk:
```
-- Found Boost: /usr/include (found suitable version "1.75.0", minimum required is "1.71")  
-- Found Boost: /usr/include (found suitable version "1.75.0", minimum required is "1.71") found components: filesystem 
[ 11%] Creating directories for 'asio-populate'
[ 22%] Performing download step (git clone) for 'asio-populate'
Cloning into 'asio-src'...
fatal: unable to access 'https://github.com/chriskohlhoff/asio.git/': Could not resolve host: github.com
Cloning into 'asio-src'...
fatal: unable to access 'https://github.com/chriskohlhoff/asio.git/': Could not resolve host: github.com
Cloning into 'asio-src'...
fatal: unable to access 'https://github.com/chriskohlhoff/asio.git/': Could not resolve host: github.com
-- Had to git clone more than once:
          3 times.
CMake Error at asio-subbuild/asio-populate-prefix/tmp/asio-populate-gitclone.cmake:31 (message):
  Failed to clone repository: 'https://github.com/chriskohlhoff/asio.git'
gmake[2]: *** [CMakeFiles/asio-populate.dir/build.make:102: asio-populate-prefix/src/asio-populate-stamp/asio-populate-download] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/asio-populate.dir/all] Error 2
gmake: *** [Makefile:91: all] Error 2
```

Fedora's build system does not allow to have any access to the internet and got clone any repo.

hkaiser:
The easy solution for this is to have asio pre-downloaded as part of the tarball submitted to Fedora and to add `-DFETCHCONTENT_FULLY_DISCONNECTED=ON` and `-DFETCHCONTENT_SOURCE_DIR_ASIO=<asio_dir>` to cmake.

diehlpk:
https://koji.fedoraproject.org/koji/taskinfo?taskID=69713089

diehlpk:
For now, we can use the Fedora asio-devel package and x86 built successfully.

PR #5379, opened 2021-06-09
Allow using pre-downloaded (not installed) versions of Asio and/or Apex

hkaiser:
- this adds ASIO_SOURCE_DIR and APEX_SOURCE_DIR cmake options that will prevent
  cmake from downloading the source code but to use it from the specified
  directories

This is an alternative option for resolving #5378

### DIFF
--- a/cmake/HPX_SetupApex.cmake
+++ b/cmake/HPX_SetupApex.cmake
@@ -28,13 +28,22 @@ if(HPX_WITH_APEX AND NOT TARGET APEX::apex)
       set(HPX_APEX_ROOT ${APEX_ROOT})
 
     else()
-      # If APEX_ROOT not specified, local clone into hpx source dir
+      # If APEX_ROOT not specified, local clone into hpx source dir or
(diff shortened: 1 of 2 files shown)


msimberg:
I find this unnecessary, but if it really is necessary, it's necessary. I tried again with `FETCHCONTENT_SOURCE_DIR_ASIO=...` and `FETCHONTENT_FULLY_DISCONNECTED=OFF` and it seems to still use the directory specified (i.e. not clone). There's still a message saying "ASIO_ROOT is not set. Cloning Asio into <path set in FETCHCONTENT_SOURCE_DIR_ASIO>", but that's on us. It should possibly say "ASIO_ROOT not set. Using fetchcontent for Asio." or something like that.

hkaiser:
I'm fine with dropping this on the floor, it indeed smells like overkill. I'll see what I can do wrt the message printed.

msimberg:
Should we go ahead and close this?

PR #5382, opened 2021-06-11
Prefer installed Asio before considering FetchContent

### LINKED ISSUE #5380: CMake should search for asio package installed on the system
It would be good if the default is that cmake search for some asio installed on the system and the fall back if not found could be to download it and use that install

msimberg:
A system-installed Asio will be looked for first. If none is found an error is raised. Users have to opt-in to using FetchContent for Asio with HPX_WITH_FETCH_ASIO=ON. @hkaiser I also tried to change the messages for the different cases that we discussed earlier. Please have a look if they match your expectations. @diehlpk would you be able to try this branch on the Fedora build system? It's meant to find Asio without any additional configuration, so if it doesn't I may have missed something.

Fixes #5380.


aurianer (dismissed):
LGTM thanks!
Only Circle CI is failing, you probably needs to change the cmake options in the yml file

diehlpk:
@msimberg Yes, I can use a patch next week.

cmake/FindAsio.cmake
```diff
@@ -0,0 +1,41 @@
...
+
+hpx_warn("finding asio")
+if(NOT TARGET Asio::asio)
+  find_path(
+    ASIO_INCLUDE_DIR asio.hpp
+    HINTS "${ASIO_ROOT}" ENV ASIO_ROOT "${HPX_ASIO_ROOT}"
+    PATH_SUFFIXES include
+  )
+
+  if(NOT ASIO_INCLUDE_DIR)
```
hkaiser, line 18: Couldn't we silently fetch Asio automatically if we don't find any installed versions?

msimberg: I thought we concluded in the meeting that we don't want a silent fallback. I also prefer loudly stopping and telling users to explicitly opt-in to fetching Asio.

hkaiser: My concern is that one of the arguments in favor of splitting our repository was to reduce the necessary configuration options for the users. This in fact increases the necessary options for a large number of users, however. However, in the end I am impartial to this change.

hkaiser (approved):
LGTM, thanks!

msimberg:
@diehlpk did you get a chance to try this out on fedora or should we go ahead with it and you try 1.7.0-rc2 instead?

diehlpk:
@msimberg I will test it tomorrow on my desktop.

diehlpk:
@msimberg I applied this patch to 1.7.0-rc1 and cmake could configure HPX and I could build it.

Issue #5366, opened 2021-06-02
Performance regression in CUDA executor’s callback mode


DarkDeepBlue:
There is a performance regression for the _callback_ mode of the CUDA executor.

The regression is introduced by commit
`b5a7913aad` (Fix check in unregister_thread and add register bool, 2020-10-14).

## Expected Behavior

Using preceding commit `02cbcf5413` (Fix hipcc bug of returning const ref when accessing non-const array, 2020-09-30), the results of the executor’s performance benchmark are as follows.

```
$ bin/cuda_executor_throughput_test
[HPX CUBLAS executor benchmark] - Starting...
GPU Device 0: "GeForce RTX 3080" with compute capability 8
MatrixA(4,4), MatrixB(4,4), MatrixC(4,4)

Async host->device copy operation completed

Small matrix multiply tests using CUBLAS...

us per iteration 40.23 : Warmup

us per iteration 32.3667 : Callback based executor

us per iteration 27.0667 : Event polling based executor
```

Note that the runtime of each iteration does not depend on the total iteration count.

```
$ bin/cuda_executor_throughput_test --iterations=10000
[HPX CUBLAS executor benchmark] - Starting...
GPU Device 0: "GeForce RTX 3080" with compute capability 8
MatrixA(4,4), MatrixB(4,4), MatrixC(4,4)

Async host->device copy operation completed

Small matrix multiply tests using CUBLAS...

us per iteration 35.19 : Warmup

us per iteration 29.3334 : Callback based executor

us per iteration 16.5664 : Event polling based executor
```

## Actual Behavior

Starting with commit `b5a7913aad` (Fix check in unregister_thread and add register bool, 2020-10-14), the runtime of the callback based executor is about three times as much. The runtime of the event based executor does not seem to be affected.

```
$ bin/cuda_executor_throughput_test
[HPX CUBLAS executor benchmark] - Starting...
GPU Device 0: "GeForce RTX 3080" with compute capability 8
MatrixA(4,4), MatrixB(4,4), MatrixC(4,4)

Async host->device copy operation completed

Small matrix multiply tests using CUBLAS...

us per iteration 112.14 : Warmup

us per iteration 99.7333 : Callback based executor

us per iteration 24.1333 : Event polling based executor
```

Note that the runtime of the callback mode depends on the total iteration count. With more iterations, the average runtime increases for the executor’s callback mode.

```
$ bin/cuda_executor_throughput_test --iterations=10000
[...]

PR #5383, opened 2021-06-11
Tentatively remove runtime_registration_wrapper from cuda futures

msimberg:
Tentatively removing it because removing it should also fix the regression in #5366.

This is untested and I want to see what the different CUDA and HIP builders say about this change. According to @DarkDeepBlue things seem to be fine without the wrapper.

I think the wrapper is mostly unneeded, but I continuations are probably still problematic without the wrapper. When run on a non-HPX thread they would fall back to trying to get the default pool which is done through the runtime pointer.

One possible solution to this is to not have the thread-local runtime pointer be thread-local anymore. @hkaiser, @sithhell what was the original motivation for making it thread-local? The pointer itself is only ever read (except when initialized or reset during init and finalize), so I don't see any performance benefits of having it be thread-local. Is there some other reason for it?

### DIFF
--- a/libs/full/async_cuda/include/hpx/async_cuda/cuda_future.hpp
+++ b/libs/full/async_cuda/include/hpx/async_cuda/cuda_future.hpp
@@ -45,18 +45,6 @@ namespace hpx { namespace cuda { namespace experimental {
         struct future_data;
 
         // -------------------------------------------------------------
-        // a callback on an NVidia cuda thread should be registered with
-        // hpx to ensure any thread local operations are valid
-        // @TODO - get rid of this
-        struct runtime_registration_wrapper
-        {
-            runtime_registration_wrapper(hpx::runtime* rt);
-            ~runtime_registration_wrapper();
-            hpx::runtime* rt_;
-            bool registered_;
-        };
-
-        // -------------------------------------------------------------
         // helper struct to delete future data in destructor
         template <typename Allocator>
         struct release_on_exit
@@ -169,7 +157,6 @@ namespace hpx { namespace cuda { namespace experimental {
             {
                 future_data* this_ = static_cast<future_data*>(user_data);
 
-                runtime_registration_wrapper wrap(this_->rt_);
                 release_on_exit<Allocator> on_exit(this_);
 
                 if (error != cudaSuccess)
--- a/libs/full/async_cuda/src/cuda_future.cpp
+++ b/libs/full/async_cuda/src/cuda_future.cpp
@@ -12,38 +12,6 @@
 #include <hpx/async_cuda/custom_gpu_api.hpp>
 
 namespace hpx { namespace cuda { namespace experimental { namespace detail {
-    runtime_registration_wrapper::runtime_registration_wrapper(hpx::runtime* rt)
-      : rt_(rt)
-      , registered_(false)
-    {
-        if (nullptr != hpx::get_runtime_ptr())
-        {
-            return;
-        }
-
-        HPX_ASSERT(rt);
-
-        // Register this thread with HPX, this should be done once for
-        // each external OS-thread intended to invoke HPX functionality.
-        // Calling this function more than once on the same thread will
-        // report an error.
-        hpx::error_code ec(hpx::lightweight);    // ignore errors
-        hpx::register_thread(rt_, "cuda", ec);
-        registered_ = true;
-    }
-
-    runtime_registration_wrapper::~runtime_registration_wrapper()
-    {
-        // Unregister the thread from HPX, this should be done once in the end
-        // before the external thread exits, if the runtime registration
-        // wrapper actually registered the thread (it may not do so if the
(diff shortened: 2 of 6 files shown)


hkaiser:
> One possible solution to this is to not have the thread-local runtime pointer be thread-local anymore. @hkaiser, @sithhell what was the original motivation for making it thread-local? The pointer itself is only ever read (except when initialized or reset during init and finalize), so I don't see any performance benefits of having it be thread-local. Is there some other reason for it?

From what I can see, the runtime_pointer could be made global without any issues nowadays. Same is probably true for some of the other thread-local variables we have.

hkaiser:
> Tentatively removing it because removing it should also fix the regression in #5366.

Not sure why this should fix the performance regression. The runtime_wrapper was in there forever.

> This is untested and I want to see what the different CUDA and HIP builders say about this change. According to @DarkDeepBlue things seem to be fine without the wrapper.

Things should be fine as long as the future's shared state does not suspend.

> I think the wrapper is mostly unneeded, but I continuations are probably still problematic without the wrapper. When run on a non-HPX thread they would fall back to trying to get the default pool which is done through the runtime pointer.

Yes, I agree.

msimberg:
> > Tentatively removing it because removing it should also fix the regression in #5366.
> 
> Not sure why this should fix the performance regression. The runtime_wrapper was in there forever.

The regression was introduced by changes to the wrapper. Removing it is of course the heavy-handed way of fixing the regression, but it should still fix the regression. And if we really don't need the wrapper then fixing the regression by removing it seems fair to me.

msimberg:
@hkaiser I've pushed some preliminary changes to make the runtime pointers non-thread local. The CUDA tests at least seem to be running fine with and without the distributed runtime.

One thing that annoyed me was that with the distributed runtime enabled the runtime pointer has to be reset early in `stop`/`stop_helper`, because if it's reset in the destructor it seems like at least some of the performance counter destruction/deinitialization/whatever is run (again?) in the `runtime_distributed` destructor and it eventually hits an assertion. For symmetry it'd be nice if the pointers could be reset in the `runtime`/`runtime_distributed` destructors (they're initialized in the constructors). Do you think that would be possible (I realize this is superficial, I can explain it in more detail if you'd like)?

hkaiser:
> One thing that annoyed me was that with the distributed runtime enabled the runtime pointer has to be reset early in `stop`/`stop_helper`, because if it's reset in the destructor it seems like at least some of the performance counter destruction/deinitialization/whatever is run (again?) in the `runtime_distributed` destructor and it eventually hits an assertion. For symmetry it'd be nice if the pointers could be reset in the `runtime`/`runtime_distributed` destructors (they're initialized in the constructors). Do you think that would be possible (I realize this is superficial, I can explain it in more detail if you'd like)?

Have you been able to resolve this? I can have a look otherwise.

msimberg:
> > One thing that annoyed me was that with the distributed runtime enabled the runtime pointer has to be reset early in `stop`/`stop_helper`, because if it's reset in the destructor it seems like at least some of the performance counter destruction/deinitialization/whatever is run (again?) in the `runtime_distributed` destructor and it eventually hits an assertion. For symmetry it'd be nice if the pointers could be reset in the `runtime`/`runtime_distributed` destructors (they're initialized in the constructors). Do you think that would be possible (I realize this is superficial, I can explain it in more detail if you'd like)?
> 
> Have you been able to resolve this? I can have a look otherwise.

I did not attempt to resolve it (meaning it's working as it is, but the pointer is not reset in the destructor). I'd definitely appreciate if you could have a look. Would you like a broken version to show the problem?

hkaiser:
> Would you like a broken version to show the problem?

That might simplify diagnosing the issue, much appreciated.

msimberg:
> > Would you like a broken version to show the problem?
> 
> That might simplify diagnosing the issue, much appreciated.

I've pushed another commit which moves the deinit to the destructors. Specifically, what happens is that `hpx::performance_counters::registry::remove_counter_type` checks if the (local) runtime pointer is still available, and proceeds if it is. It then fails further down on the stack. On master the runtime pointer (both of them) is already reset at this point. Since the distributed runtime destructor happens before the local runtime destructor the local runtime pointer is still there. It may very well be that the fix is as simple as checking for the distributed runtime pointer instead of the local one, but I can't judge if that's actually the correct thing to do.

msimberg:
@hkaiser have you possibly had a chance to look at this? If not, should we go ahead without moving the deinit for the moment (i.e. keep the same behaviour as currently on master)?

msimberg:
I've rebased and dropped the bad commit. If this comes back clean this should be ready to go. I've kept the breaking commit here: https://github.com/msimberg/hpx/commit/67c40f8a4f5dab1ba8bdb18bb225ad98747d65c4 in case someone wants to come back to this.

msimberg:
Reformatted. Looks like CI was happy with this except for the formatting. So this time it should really be ready to go.

Issue #4558, opened 2020-04-26
Prevent work-stealing from stalling


hkaiser:
Work-stealing relies on the assumption that all running HPX-tasks eventually give back control to the scheduler at some point. This can lead to 'lock-ups' if 

- one (or more) of the HPX tasks keep running without ever yielding control to the scheduler
- the scheduler of the core that is running that HPX task has staged work
- the amount of staged work is smaller than the configuration setting for `min_tasks_to_steal_staged`

In this situation no other scheduler/core will be able to steal the staged tasks while the core that owns them will never execute them either.

We should enable the scheduler to detect that it has run out of work and is 'desperate' for more work, overruling the configuration minimum for `min_tasks_to_steal_staged` and to steal the staged tasks eventually (after several unsuccessful attempts, perhaps).

msimberg:
This was "fixed" by https://github.com/STEllAR-GROUP/hpx/pull/5387 since the default for `min_tasks_to_steal_staged` is now `0`. The underlying problem still exists of course if someone changes that default to a non-zero value.

hkaiser:
I think this was resolved by defaulting to zero the minimal number of threads that need to be available before allowing stealing.

PR #5387, opened 2021-06-14
Extend thread state logging and change default stealing parameters

msimberg:
This PR is the result of some debuggin in DLA-Future of a deadlock caused by a combination of our scheduler and a blocking MPI call being made in an HPX thread. We know that the latter is very naughty, but at the same time it can lead to performance benefits in this particular case (lots of small tasks if one does not do the blocking call). We're looking into alternatives, but one of the changes here also helps for the particular situation. I'm happy to drop it though as it's only a question of default configuration values. That said, the changes here are:

- Add more extensive logging to the scheduling loop. Previously the state of an HPX thread was logged only once the context was changed back to the scheduler. This PR adds logging both before and after threads are scheduled so that one can see when they start and stop. This helps with recognizing if a task has been started but never finishes. I've also added more information to the messages (thread pool, scheduler). I also log when attmpting to steal threads, but the number of threads is lower than the threshold.
- Change the `idle_loop_count` to count up again like the other counters. The counter was decremented in most places, but incremented in another (https://github.com/STEllAR-GROUP/hpx/blob/bb51c2c8cbd0dc68026f3bd92b0b746b7a273d86/libs/core/schedulers/include/hpx/schedulers/queue_helpers.hpp#L53). This alone did not cause a deadlock, but it was part of it. The `idle_loop_count` never reached a value where `enable_stealing_staged` would be true. @hkaiser I know you changed this to count down in https://github.com/STEllAR-GROUP/hpx/pull/3768. However, I don't see which cases become more cumbersome when incrementing the counter (https://github.com/STEllAR-GROUP/hpx/pull/3768#discussion_r272145021). Do you remember what those were? Does `idle_loop_count` need to be reset to `max_idle_loop_count` somewhere outside the scheduling loop?
- Finally, change default value of `min_tasks_to_steal_staged` to 0. This is not a must, mostly convenience.

### DIFF
--- a/libs/core/schedulers/include/hpx/schedulers/thread_queue.hpp
+++ b/libs/core/schedulers/include/hpx/schedulers/thread_queue.hpp
@@ -1004,11 +1004,24 @@ namespace hpx { namespace threads { namespace policies {
 
                 // don't try to steal if there are only a few tasks left on
                 // this queue
-                if (running &&
-                    parameters_.min_tasks_to_steal_staged_ >
-                        addfrom->new_tasks_count_.data_.load(
-                            std::memory_order_relaxed))
+                std::int64_t new_tasks_count =
+                    addfrom->new_tasks_count_.data_.load(
+                        std::memory_order_relaxed);
+                bool enough_threads =
+                    new_tasks_count >= parameters_.min_tasks_to_steal_staged_;
+
+                if (running && !enough_threads)
                 {
+                    if (new_tasks_count != 0)
+                    {
+                        LTM_(debug).format(
+                            "thread_queue::wait_or_add_new: not enough threads "
+                            "to steal from queue {} to queue {}, have {} but "
+                            "need at least {}",
+                            addfrom, this, new_tasks_count,
+                            parameters_.min_tasks_to_steal_staged_);
+                    }
+
                     return false;
                 }
 
--- a/docs/sphinx/manual/launching_and_configuring_hpx_applications.rst
+++ b/docs/sphinx/manual/launching_and_configuring_hpx_applications.rst
@@ -346,7 +346,7 @@ The ``hpx.thread_queue`` configuration section
 
    [hpx.thread_queue]
    min_tasks_to_steal_pending = ${HPX_THREAD_QUEUE_MIN_TASKS_TO_STEAL_PENDING:0}
-   min_tasks_to_steal_staged = ${HPX_THREAD_QUEUE_MIN_TASKS_TO_STEAL_STAGED:10}
+   min_tasks_to_steal_staged = ${HPX_THREAD_QUEUE_MIN_TASKS_TO_STEAL_STAGED:0}
    min_add_new_count = ${HPX_THREAD_QUEUE_MIN_ADD_NEW_COUNT:10}
    max_add_new_count = ${HPX_THREAD_QUEUE_MAX_ADD_NEW_COUNT:10}
    max_delete_count = ${HPX_THREAD_QUEUE_MAX_DELETE_COUNT:1000}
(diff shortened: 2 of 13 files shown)


hkaiser:
> This PR is the result of some debuggin in DLA-Future of a deadlock caused by a combination of our scheduler and a blocking MPI call being made in an HPX thread. We know that the latter is very naughty, but at the same time it can lead to performance benefits in this particular case (lots of small tasks if one does not do the blocking call). We're looking into alternatives, but one of the changes here also helps for the particular situation. I'm happy to drop it though as it's only a question of default configuration values. That said, the changes here are:
> 
> * Add more extensive logging to the scheduling loop. Previously the state of an HPX thread was logged only once the context was changed back to the scheduler. This PR adds logging both before and after threads are scheduled so that one can see when they start and stop. This helps with recognizing if a task has been started but never finishes. I've also added more information to the messages (thread pool, scheduler). I also log when attempting to steal threads, but the number of threads is lower than the threshold.

Very nice. Thanks!

> * Change the `idle_loop_count` to count up again like the other counters. The counter was decremented in most places, but incremented in another (https://github.com/STEllAR-GROUP/hpx/blob/bb51c2c8cbd0dc68026f3bd92b0b746b7a273d86/libs/core/schedulers/include/hpx/schedulers/queue_helpers.hpp#L53). 

This is a bug. I shouldn't ever have been incremented.

> This alone did not cause a deadlock, but it was part of it. The `idle_loop_count` never reached a value where `enable_stealing_staged` would be true. @hkaiser I know you changed this to count down in #3768. However, I don't see which cases become more cumbersome when incrementing the counter ([#3768 (comment)](https://github.com/STEllAR-GROUP/hpx/pull/3768#discussion_r272145021)). Do you remember what those were? Does `idle_loop_count` need to be reset to `max_idle_loop_count` somewhere outside the scheduling loop?

IIRC, I changed it to count down as it needed to be tested some place where the `max_idle_loop_count` wasn't readily available. But this might have been changed since then (sorry I don't remember the details). I'm fine with changing it back to count up if there is no problem doing so.

> * Finally, change default value of `min_tasks_to_steal_staged` to 0. This is not a must, mostly convenience.

I admit that at some point I did run into a deadlock issue caused by this value not being zero as well and forgot to change the default value as this would have prevented my deadlock (I did create an issue, though: #4558). Thanks for this.

libs/core/schedulers/include/hpx/schedulers/thread_queue.hpp
```diff
@@ -1004,11 +1004,24 @@ namespace hpx { namespace threads { namespace policies {
...
-                            std::memory_order_relaxed))
+                std::int64_t new_tasks_count =
+                    addfrom->new_tasks_count_.data_.load(
+                        std::memory_order_relaxed);
+                bool enough_threads =
+                    new_tasks_count >= parameters_.min_tasks_to_steal_staged_;
+
+                if (running && !enough_threads)
                 {
+                    if (!enough_threads && new_tasks_count != 0)
```
hkaiser, line 1015:
```suggestion
                    if (new_tasks_count != 0)
```

hkaiser (approved):
LGTM, thanks!

albestro:
Just tested again, and it works! and the new debug log is super useful.

thanks @msimberg!

msimberg:
> This is a bug. I shouldn't ever have been incremented.
> IIRC, I changed it to count down as it needed to be tested some place where the `max_idle_loop_count` wasn't readily available. But this might have been changed since then (sorry I don't remember the details). I'm fine with changing it back to count up if there is no problem doing so.

No worries. I am slightly worried that I'll introduce the same bug in the opposite direction with this change, so I'll have another look through the code to see if I find a case like this. My reasoning for changing the direction is that it's consistent with the counts elsewhere and thus it'll hopefully avoid a similar bug in the future, but that doesn't mean I can't mess it up while doing this change...

> I admit that at some point I did run into a deadlock issue caused by this value not being zero as well and forgot to change the default value as this would have prevented my deadlock (I did create an issue, though: #4558). Thanks for this.

Ok, sounds good. Users will still have the choice of increasing it if they know it's not worth it because of too small tasks, but this sounds like a better default in that case.

msimberg:
> No worries. I am slightly worried that I'll introduce the same bug in the opposite direction with this change, so I'll have another look through the code to see if I find a case like this. My reasoning for changing the direction is that it's consistent with the counts elsewhere and thus it'll hopefully avoid a similar bug in the future, but that doesn't mean I can't mess it up while doing this change...

Well, to the best of my knowledge there are no cases like this (anymore at least). I'll go ahead and merge this now.

PR #5390, opened 2021-06-15
Make sure shared priority scheduler steals tasks on the current NUMA domain when (core) stealing is enabled

msimberg:
I've re-enabled the cross pool injection test with the shared priority scheduler in the hopes that this maybe fixes the hangs there, but have not tested it myself. Edit: this does not seem to fix that particular problem, so I'm leaving it disabled.


msimberg:
@biddisco do you think this is ok to merge?

biddisco (approved):
This looks like a silly mistake in the scheduler and the fix should go in.

PR #5392, opened 2021-06-15
Fixing wrong usage of std::forward

hkaiser:
@msimberg sorry for this oversight... Could we add this to the release, please?

### DIFF
--- a/libs/full/collectives/include/hpx/collectives/gather.hpp
+++ b/libs/full/collectives/include/hpx/collectives/gather.hpp
@@ -260,8 +260,8 @@ namespace hpx { namespace collectives {
 
             // make sure id is kept alive as long as the returned future,
             // explicitly unwrap returned future
-            hpx::future<std::vector<arg_type>> result = async(
-                action_type(), c, this_site, std::forward<T>(local_result));
+            hpx::future<std::vector<arg_type>> result =
+                async(action_type(), c, this_site, std::move(local_result));
 
             traits::detail::get_shared_state(result)->set_on_completed(
                 [client = std::move(c)]() { HPX_UNUSED(client); });
@@ -307,8 +307,8 @@ namespace hpx { namespace collectives {
 
             // make sure id is kept alive as long as the returned future,
             // explicitly unwrap returned future
-            hpx::future<void> result = async(
-                action_type(), c, this_site, std::forward<T>(local_result));
+            hpx::future<void> result =
+                async(action_type(), c, this_site, std::move(local_result));
 
             traits::detail::get_shared_state(result)->set_on_completed(
                 [client = std::move(c)]() { HPX_UNUSED(client); });


msimberg (approved):
Thanks @hkaiser! (I force pushed a formatting change.)

libs/full/collectives/include/hpx/collectives/gather.hpp
```diff
@@ -260,8 +260,8 @@ namespace hpx { namespace collectives {
 
             // make sure id is kept alive as long as the returned future,
             // explicitly unwrap returned future
-            hpx::future<std::vector<arg_type>> result = async(
-                action_type(), c, this_site, std::forward<T>(local_result));
+            hpx::future<std::vector<arg_type>> result =
+                async(action_type(), c, this_site, std::move(local_result));
```
aurianer, line 264: Could you explain the `move` here and the `forward<T>(local_result)` in dataflow? My C++ knowledge is quite lacking..

hkaiser: Generally, `std::forward` is a tool enabling perfect forwarding of forwarding reference (i.e. `T&&` arguments, where `T` is an immediate template argument on the function). OTOH, `std::move` can be used on rvalue reference arguments (i.e. `T&&` where `T` in not a template parameter but a real type) in order to enable moving the value.

In this case, I initially changed the lambda to accept an rvalue reference instead of a forwarding reference. This PR fixes my oversight of not correcting the handling of said argument as well.

msimberg: Follow-up question: shouldn't the `local_result` parameter to the `gather_data_direct` lambda be `auto&&` (and then forwarded) rather than an rvalue reference and then moved?

On line 272 `local_result` is forwarded into `dataflow`, not moved, so should the lambda also not accept lvalue references?

hkaiser: > Follow-up question: shouldn't the `local_result` parameter to the `gather_data_direct` lambda be auto&& (and then forwarded) rather than an rvalue reference and then moved?

`dataflow` always moves the arguments into the function (same as `async`, btw) as the thread executing it binds them using a `util::one_shot`. For this reason, using a forwarding reference as the lambda argument would break for certain scenarios (for instance as reported by Kevin).

> On line 272 `local_result` is forwarded into `dataflow`, not moved, so should the lambda also not accept lvalue references?

Yes, this perfectly forwards the argument down into the `util::one_shot` binding it. If the argument is an lvalue, then it will be copied _there_, if it's an rvalue, it will be simply moved into the binder.

msimberg: Thanks. I had completely forgotten about dataflow copying references internally.

PR #5400, opened 2021-06-18
Fix sliding semaphore test

msimberg:
The semaphore has to be kept alive long enough for all tasks to get to signal it. The task that releases the waiting thread may not be the last task to signal the semaphore, and if that happens the semaphore will go
out of scope too early. (Explanation and diagram in the test.)


hkaiser (approved):
Excellent catch! Thanks!

aurianer (approved):
Thanks!

PR #5401, opened 2021-06-19
Modernize a variety of facilities related to parallel algorithms

### LINKED ISSUE #5397: Performance regression in thread annotations
Some change introduced between 84cf823 and the current master seems to result in a serious performance regression when using the parallel algorithms, e.g., `hpx::transform_reduce`, with an annotated lambda, where the performance deteriorates by a factor of 5 in my application. It seems also other loop constructs are affected, although not quite as drastically.

@khuck: Any idea if there was a change in APEX that could manifest in such a dramatic way?

## Specifications
  - HPX Version:1.7.0-rc1/master
  - APEX Version: 2.4.1
  - Platform (compiler, OS): GCC 10.2.0, Linux

hkaiser:
- flyby: modernize `transform` algorithms, `annotated_function`
- flyby: various optimizations for transform algorithms
- flyby: move `parallel::util::detail::loop_n` and `parallel::util::detail::loop_n_ind`
  into `namespace parallel::util`
- flyby: enable `datapar` for `for_loop` with stride == 1 and no additional reducers
- flyby: add `transform_binary_loop_ind` and `transform_binary_loop_ind_n` helper facilities (these need datapar counterparts, still)

Fixes #5397

If we decided to add the lifting of annotated function from the algorithm iteration to the chunk (as currently proposed), then other algorithms need to be adapted. If we decided not to add the lifting (but perhaps add a generic implementation for executors to support annotations), then I'll change this PR, leaving only all the flyby changes, which on their own improve things in various ways.

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform.hpp
@@ -205,6 +205,7 @@ namespace hpx {
 #include <utility>
 
 namespace hpx { namespace parallel { inline namespace v1 {
+
     ///////////////////////////////////////////////////////////////////////////
     // transform
     namespace detail {
@@ -253,9 +254,9 @@ namespace hpx { namespace parallel { inline namespace v1 {
         template <typename ExPolicy, typename F, typename Proj>
         struct transform_iteration
         {
-            typedef typename std::decay<ExPolicy>::type execution_policy_type;
-            typedef typename std::decay<F>::type fun_type;
-            typedef typename std::decay<Proj>::type proj_type;
+            using execution_policy_type = std::decay_t<ExPolicy>;
+            using fun_type = std::decay_t<F>;
+            using proj_type = std::decay_t<Proj>;
 
             fun_type f_;
             proj_type proj_;
@@ -283,44 +284,32 @@ namespace hpx { namespace parallel { inline namespace v1 {
             {
             }
 #endif
-
             transform_iteration& operator=(
                 transform_iteration const&) = default;
             transform_iteration& operator=(transform_iteration&&) = default;
 
-            template <typename Iter>
+            template <typename Iter, typename F_ = fun_type>
             HPX_HOST_DEVICE HPX_FORCEINLINE
                 std::pair<typename hpx::tuple_element<0,
                               typename Iter::iterator_tuple_type>::type,
                     typename hpx::tuple_element<1,
                         typename Iter::iterator_tuple_type>::type>
-                execute(Iter part_begin, std::size_t part_size)
+                operator()(Iter part_begin, std::size_t part_size, std::size_t)
             {
+                hpx::util::annotate_function _(f_);
+
                 auto iters = part_begin.get_iterator_tuple();
                 return util::transform_loop_n<execution_policy_type>(
                     hpx::get<0>(iters), part_size, hpx::get<1>(iters),
                     transform_projected<F, Proj>(f_, proj_));
-            }
-
-            template <typename Iter>
-            HPX_HOST_DEVICE HPX_FORCEINLINE
-                std::pair<typename hpx::tuple_element<0,
-                              typename Iter::iterator_tuple_type>::type,
-                    typename hpx::tuple_element<1,
-                        typename Iter::iterator_tuple_type>::type>
-                operator()(Iter part_begin, std::size_t part_size,
-                    std::size_t /*part_index*/)
-            {
-                hpx::util::annotate_function annotate(f_);
-                return execute(part_begin, part_size);
             }
         };
 
         template <typename ExPolicy, typename F>
         struct transform_iteration<ExPolicy, F, util::projection_identity>
         {
-            typedef typename std::decay<ExPolicy>::type execution_policy_type;
-            typedef typename std::decay<F>::type fun_type;
+            using execution_policy_type = std::decay_t<ExPolicy>;
+            using fun_type = std::decay_t<F>;
 
             fun_type f_;
 
@@ -345,35 +334,23 @@ namespace hpx { namespace parallel { inline namespace v1 {
             {
             }
 #endif
-
             transform_iteration& operator=(
                 transform_iteration const&) = default;
             transform_iteration& operator=(transform_iteration&&) = default;
 
-            template <typename Iter>
+            template <typename Iter, typename F_ = fun_type>
             HPX_HOST_DEVICE HPX_FORCEINLINE
                 std::pair<typename hpx::tuple_element<0,
                               typename Iter::iterator_tuple_type>::type,
                     typename hpx::tuple_element<1,
                         typename Iter::iterator_tuple_type>::type>
-                execute(Iter part_begin, std::size_t part_size)
+                operator()(Iter part_begin, std::size_t part_size, std::size_t)
             {
+                hpx::util::annotate_function _(f_);
+
                 auto iters = part_begin.get_iterator_tuple();
                 return util::transform_loop_n_ind<execution_policy_type>(
                     hpx::get<0>(iters), part_size, hpx::get<1>(iters), f_);
-            }
-
-            template <typename Iter>
-            HPX_HOST_DEVICE HPX_FORCEINLINE
-                std::pair<typename hpx::tuple_element<0,
-                              typename Iter::iterator_tuple_type>::type,
-                    typename hpx::tuple_element<1,
-                        typename Iter::iterator_tuple_type>::type>
-                operator()(Iter part_begin, std::size_t part_size,
-                    std::size_t /*part_index*/)
-            {
-                hpx::util::annotate_function annotate(f_);
-                return execute(part_begin, part_size);
             }
         };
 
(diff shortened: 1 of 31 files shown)


hkaiser:
@severinstrobl could you please verify if this fixes the issue you reported?

severinstrobl:
@hkaiser Nice, thanks for fixing that so quickly! With your changes, performance is back to the original level for my simple testcase. I'll do some more testing with a more complex workload on our cluster later, just in case.

hkaiser:
> @hkaiser Nice, thanks for fixing that so quickly! With your changes, performance is back to the original level for my simple testcase. I'll do some more testing with a more complex workload on our cluster later, just in case.

@severinstrobl Nice, thanks for checking! I would even hope that the performance improved compared to before - can you confirm that? 

The solution @msimberg mentioned is preferable, it requires some more work on our end, however.

msimberg:
As a short-term solution for @severinstrobl and `transform` specifically I don't mind the changes in this PR. Long-term I definitely think adding this same boilerplate to all the algorithms is unnecessary, error-prone, and as I mentioned earlier semantically wrong (perhaps a bit strong, but the only cases where this is even applicable is when there is at most a single callable passed to the algorithm; there are also algorithms which take no callables where putting the annotation on the policy is the only reasonable option).

I can try to add support for annotations to `block_executor` on Mon/Tue. I don't think the changes would be any more complicated than here, and possibly even simpler, especially as @hkaiser already has added support to policies in general and the `parallel_executor`.

hkaiser:
> As a short-term solution for @severinstrobl and `transform` specifically I don't mind the changes in this PR. Long-term I definitely think adding this same boilerplate to all the algorithms is unnecessary, error-prone, and as I mentioned earlier semantically wrong (perhaps a bit strong, but the only cases where this is even applicable is when there is at most a single callable passed to the algorithm; there are also algorithms which take no callables where putting the annotation on the policy is the only reasonable option).
> 
> I can try to add support for annotations to `block_executor` on Mon/Tue. I don't think the changes would be any more complicated than here, and possibly even simpler, especially as @hkaiser already has added support to policies in general and the `parallel_executor`.

I fully agree. I would however advocate for creating a wrapping executor that can add annotations to any wrapped executor, instead. This executor could then silently be injected by `with_annotation` if the executor passed into doesn't already support those. That has the other benefit that `with_annotation` would never fail.

hkaiser:
> I fully agree. I would however advocate for creating a wrapping executor that can add annotations to any wrapped executor, instead. This executor could then silently be injected by `with_annotation` if the executor passed into doesn't already support those. That has the other benefit that `with_annotation` would never fail.

@severinstrobl with #5403 you should be able to remove the `annotated_function` from the lambda and use `with_annotation(policy, "...")` as described by @msimberg instead. Could you please verify that this would solve your performance problem as well? If yes, then I will rollback the changes specific to transform here.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform.hpp
```diff
@@ -530,32 +508,89 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
                                typename Iter::iterator_tuple_type>::type,
                     typename hpx::tuple_element<1,
                         typename Iter::iterator_tuple_type>::type,
                     typename hpx::tuple_element<2,
                         typename Iter::iterator_tuple_type>::type>
-                operator()(Iter part_begin, std::size_t part_size,
-                    std::size_t /*part_index*/)
+                operator()(Iter part_begin, std::size_t part_size, std::size_t)
             {
+                hpx::util::annotate_function _(f_);
```
msimberg, line 526: Is this supposed to stay here?

libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform.hpp
```diff
@@ -530,32 +508,89 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+            HPX_HOST_DEVICE HPX_FORCEINLINE
+                hpx::tuple<typename hpx::tuple_element<0,
+                               typename Iter::iterator_tuple_type>::type,
+                    typename hpx::tuple_element<1,
+                        typename Iter::iterator_tuple_type>::type,
+                    typename hpx::tuple_element<2,
+                        typename Iter::iterator_tuple_type>::type>
+                operator()(Iter part_begin, std::size_t part_size, std::size_t)
+            {
+                hpx::util::annotate_function _(f_);
```
msimberg, line 585: Same here? And a few other places...

msimberg (changes requested):
Looks good to me in general, thanks!

Could you perhaps squash the commits? Right now the commit messages might make one think that this PR has no changes at all...

severinstrobl:
**Note**: The statement below are still regarding the changes done in f32c8d94, so with the lifting of the annotated function, which was reverted in the meantime.

> @severinstrobl Nice, thanks for checking! I would even hope that the performance improved compared to before - can you confirm that?

@hkaiser  Sorry for not replying earlier. I do not see a (measurable) increase in performance compared to the older version, but this might just be specific to my test case.

severinstrobl:
> @severinstrobl with #5403 you should be able to remove the `annotated_function` from the lambda and use `with_annotation(policy, "...")` as described by @msimberg instead. Could you please verify that this would solve your performance problem as well? If yes, then I will rollback the changes specific to transform here.

Attaching the annotation to the execution policy using the changes from #5403 works as well and gives me identical performance results compared to the original version or f32c8d94 with the lifted annotations.

While I can see that attaching the annotation to the execution policy is a neater way of handling this, I'm unclear on whether (in the long run) this should be the only way to handle annotations. For example, when using a `hpx::transform_reduce` having just one annotation on the execution policy does not allow me to have a separate annotation for the transformation and the reduction operation, unless I'm missing something here. Assuming my reduction is less trivial than a `std::plus<>`, I might want to still be able to have separate annotations for the two operations, especially if one is interested in generating task graphs via APEX. So I guess my question is, will `hpx::util::annotated_function` be removed eventually, or will it remain available, potentially incurring some runtime overhead?

libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform.hpp
hkaiser: It was there before... I made it consistent, only.

hkaiser:
> So I guess my question is, will `hpx::util::annotated_function` be removed eventually, or will it remain available, potentially incurring some runtime overhead?

It will stay available. We have no plans of removing it.

hkaiser:
> Could you perhaps squash the commits?

I can certainly do that - even if I specifically left the rollback in a separate commit to make it easier to understand what has been rolled back.

msimberg:
> > Could you perhaps squash the commits?
> 
> I can certainly do that - even if I specifically left the rollback in a separate commit to make it easier to understand what has been rolled back.

Just editing the commit message on the first one would also be sufficient to describe the changes that were *not* reverted in the second commit.

hkaiser:
@severinstrobl After all that has been said and done - currently we don't have a 'fix' for the performance regression you reported. All we can provide you with are workarounds. This is something that needs more thorough investigation and experimentation. I will keep it on my list of things to look into.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/transform.hpp
msimberg: Hrm, indeed. Now the question is, is this going to overwrite the annotation set by the policy? I think it will. For the sequential cases it makes sense to set the annotation explicitly, but for the parallel cases I'm not so sure.

hkaiser: You're right. The sequential cases of all algorithms ought to be changed to go through the executor anyways (can be done in one place), so I'd say let's remove the annotation there as well.

hkaiser: I have removed the annotations here and from for_each now.

severinstrobl:
Understood, thanks a lot for putting all this effort into resolving the problem. For the time being I'll stay on an older snapshot. In case the more general solution for the  performance regression does not make it into 1.7.0, you might want to include a corresponding warning in the release notes.

msimberg (dismissed):
Thanks @hkaiser for this! Assuming CircleCI comes back clean let's merge this.

libs/parallelism/algorithms/include/hpx/parallel/datapar/loop.hpp
```diff
@@ -475,28 +475,25 @@ namespace hpx { namespace parallel { namespace util {
...
-        HPX_HOST_DEVICE HPX_FORCEINLINE constexpr typename std::enable_if<
-            hpx::is_vectorpack_execution_policy<ExPolicy>::value, Iter>::type
-        tag_dispatch(hpx::parallel::util::detail::loop_n_t<ExPolicy>, Iter it,
-            std::size_t count, F&& f)
-        {
-            return hpx::parallel::util::detail::datapar_loop_n<Iter>::call(
-                it, count, std::forward<F>(f));
-        }
+    template <typename ExPolicy, typename Iter, typename F>
+    HPX_HOST_DEVICE HPX_FORCEINLINE constexpr typename std::enable_if<
```
aurianer, line 479: `std::enable_if_t`? as you have been updating the others? (not a must)

libs/parallelism/algorithms/include/hpx/parallel/datapar/transform_loop.hpp
```diff
@@ -460,8 +460,11 @@ namespace hpx { namespace parallel { namespace util {
             call(InIter1 first1, InIter1 last1, InIter2 first2, InIter2 last2,
                 OutIter dest, F&& f)
             {
-                std::size_t count = (std::min)(
-                    std::distance(first1, last1), std::distance(first2, last2));
+                // different version of clang-formt do different things
```
aurianer, line 463: There is the `SpaceAfterCStyleCast` option that we can add to our `.clang-format` configuration

aurianer (approved):
LGTM thanks a lot!

libs/parallelism/algorithms/include/hpx/parallel/datapar/loop.hpp
hkaiser: @aurianer I have not looked at this file too much, just did the edits to conform to the namespace changes of the main facility. I'll try to have a look at this file separately.

libs/parallelism/algorithms/include/hpx/parallel/datapar/transform_loop.hpp
hkaiser: I was not aware that clang-format applies the `SpaceAfterCStyleCast` setting in this case. It might be worth a try adding it to see if this makes versions more conforming to each other. Could you try this setting with your version of clang-format, I'll verify it with v11 here.

hkaiser:
The hipcc problems were caused by an issue on master that has been fixed (see https://github.com/STEllAR-GROUP/hpx/commit/258d146d4708300f785d63da8c8a6c4fa5903c2d).

libs/parallelism/algorithms/include/hpx/parallel/datapar/transform_loop.hpp
hkaiser: Currently `SpaceAfterCStyleCast` is enabled, if I disable it, the space after the `(std::min)` is gone. So it looks like they changed the behavior in v11. Should we change the setting to `false` to make newer versions of clang-format behave like the older ones?

hkaiser: Sadly, changing it to `false` touches on a lot of other spots...

PR #5044, opened 2020-11-04
Add performance test in jenkins and reports

aurianer:
Very much WIP, I plan to work on that on the side

Start work on performance regression:
- Reuse gridtools python scripts for the moment (might use another tool provided by cscs in the future)
- Will create a performance regression report with history plots (static)

TODO unit testing:
- [x] Check that the jenkins pipeline is working just by configuring, building and running `future_overhead_report_test` (`future_overhead` outputing a json report) with the python scripts provided
- [x] Check the comparison with the reference time + report
- [x] Update the cdash status to fail if the performance has dropped

PR #5405, opened 2021-06-21
Fetch Asio automatically in perftests CI

msimberg:
This just got missed with #5382 and #5044 being merged at roughly the same time.


hkaiser (approved):

hkaiser:
BTW, is there a way to make the perftest results publicly available? What would we need for this? Just any web-hosting?

msimberg:
> BTW, is there a way to make the perftest results publicly available? What would we need for this? Just any web-hosting?

That was left as an open TODO on #5044. We very much want it but left it out of scope from the first PR to at least get it running even though not everyone can see them.

Uploading to CDash was the first option, but CDash will only display the html report as plaintext, which makes it somewhat hard to read... A second option could be to comment on PRs with the GitHub API. A third option would be to upload them *anywhere* (webhost, git repo, etc. etc.).

PR #5403, opened 2021-06-19
Create generic executor that adds annotations to any other executor

hkaiser:
- flyby: removed obsolete #if defined(HPX_HAVE_CXX17_VARIABLE_TEMPLATES)
- flyby: constexpr variables can be inline
- flyby: modernize execution traits

PR #5406, opened 2021-06-21
Simplify dispatching to annotation handlers

hkaiser:
This is mostly a cleanup PR for #5403 simplifying the selection conditions for the various `with_annotation` tag_dispatch overloads.

### DIFF
--- a/libs/parallelism/executors/include/hpx/executors/annotating_executor.hpp
+++ b/libs/parallelism/executors/include/hpx/executors/annotating_executor.hpp
@@ -139,8 +139,6 @@ namespace hpx { namespace execution { namespace experimental {
         }
 
         // support with_annotation property
-        using supports_annotations = void;
-
         friend constexpr annotating_executor tag_dispatch(
             hpx::execution::experimental::with_annotation_t,
             annotating_executor const& exec, char const* annotation)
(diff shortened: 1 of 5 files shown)


libs/parallelism/executors/include/hpx/executors/annotating_executor.hpp
```diff
@@ -188,59 +186,32 @@ namespace hpx { namespace execution { namespace experimental {
...
     // to `with_annotation`.
-
+    //
     // clang-format off
-    template <typename Executor>
+    template <typename Executor,
+        HPX_CONCEPT_REQUIRES_(
+            hpx::traits::is_executor_any_v<Executor>
+        )>
+    // clang-format on
```
msimberg, line 199: What's stopping this from being more specific than https://github.com/STEllAR-GROUP/hpx/blob/258d146d4708300f785d63da8c8a6c4fa5903c2d/libs/parallelism/executors/include/hpx/executors/parallel_executor.hpp#L190-L192? In fact, what makes this overload visible to e.g. `block_executor` (should've asked already on the previous PR...)?

hkaiser: This overload is visible to the `block_executor` as it is in the same namespace as `with_annotation_t` and the `block_executor` makes `is_executor_any_v` true. It is always less specific than any overload that concretely refers to an executor (here `parallel_policy_executor`). 

We can always go back and add the `using supports_annotation = void;` hack If we come across a situation where the code is ambiguous. For now everything seems to work as expected.

msimberg: Ah, yes, makes sense about finding the correct overload.

You'll probably know the rules better than anyone so please correct me here, but surely there are cases where `template <typename T> void f(T&&)` is a better (exact) match than `void f(sometype const&)`, no? I assumed that was the reason for originally introducing `supports_annotation`? Or is there some other rule that is kicking in in this case?

Actually, is this not exactly a case again for `tag_fallback_dispatch`?

hkaiser: > You'll probably know the rules better than anyone so please correct me here, but surely there are cases where `template <typename T> void f(T&&)` is a better (exact) match than `void f(sometype const&)`, no? I assumed that was the reason for originally introducing `supports_annotation`? Or is there some other rule that is kicking in in this case?

From how I understand the rules (and I'm by no means an expert here), the second will always be more specific than the first (for an argument of type `sometype`). I also believe, that friend functions are considered first (but I again I might be wrong).

The `supports_annotation` I originally introduced was a fool-proof implementation. After some experimentation I convinced myself that we should be able to go without it. But please correct me if I'm mistaken here. I would like to avoid it as I believe it adds unnecessary verboseness.

> Actually, is this not exactly a case again for `tag_fallback_dispatch`?

Possibly - what did you have in mind?

msimberg: @hkaiser see the last two commits.

I added some new tests for `annotating_executor` in the first commit. The annotating executor was applied for rvalue `parallel_executor`s. The test fails on that commit. The second commit makes some changes to `prefer` and the `with_annotation` fallback. I'm not 100% sure this is the right way to do it, and I may have missed some fun, new corner case, but I think this is in the right direction at least. Please have a look and see what you think.

hkaiser: Looks good to me. Thanks!

msimberg: Thanks @hkaiser for the fixes! I also replaced the `is_same_v`s with `is_same<>::value`s. Let's see what CI says now.

msimberg:
retest

msimberg:
I've only rebased this (squashed two commits together). This should be ready to go otherwise. Thanks @hkaiser again for looking into this.

PR #5409, opened 2021-06-22
Fundamentals module

hkaiser:
This is based on @K-ballo's initial work on a new fundamentals module. This also adds `tag_dispatch` and friends here, which makes #5303 obsolete.


libs/core/fundamentals/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+    hpx/functional/traits/is_invocable.hpp
+    hpx/iterator_support/range.hpp
+    hpx/type_support/detail/wrap_int.hpp
+    hpx/type_support/always_void.hpp
+    hpx/type_support/decay.hpp
+    hpx/type_support/lazy_conditional.hpp
+    hpx/type_support/lazy_enable_if.hpp
+    hpx/type_support/pack.hpp
+    hpx/type_support/unwrap_ref.hpp
+    hpx/type_support/void_guard.hpp
```
msimberg, line 28: While I'm not opposed to this at all (I did thumbs-up the issue for this after all), what is not clear to me is why specifically these headers? There are many other modules that are meant for internal use only, so what qualifies a functionality as "fundamentals"?

msimberg: Also, with #5283 around the corner, would it be worth trying to remove/replace what can be removed/replaced first?

hkaiser: The selection was made by @K-ballo, essentially (except for the `tag_dispatch` headers). My motivation was to resolve #5303 and leverage his work while doing so, I'm fine with adding more facilities over time, if sensible.

hkaiser: > Also, with #5283 around the corner, would it be worth trying to remove/replace what can be removed/replaced first?

How much of this would be replaced? I can see `always_void` being a candidate - are there more? Frankly, I don't think that #5283 should do any actual replacing (like `always_void` -> `void_t`), so I would like to decouple this from the current PR.

msimberg: All of the invoke functionality could be replaced, right? `noncopyable`, `decay`, `unwrap_ref` as @K-ballo pointed out? Possibly range? I agree that #5283 does not need to do any of this; I was referring to after it's merged.

If all of those are gone that would leave:
```
    hpx/concepts/concepts.hpp
    hpx/config/lambda_capture.hpp
    hpx/functional/tag_dispatch.hpp
    hpx/functional/tag_fallback_dispatch.hpp
    hpx/functional/tag_priority_dispatch.hpp
    hpx/type_support/detail/wrap_int.hpp
    hpx/type_support/lazy_conditional.hpp
    hpx/type_support/lazy_enable_if.hpp
    hpx/type_support/pack.hpp
    hpx/type_support/void_guard.hpp
```

I suppose the question that I'm trying to figure out is: what benefit do we have right now from putting these in one module? What's the long term goal for a "fundamentals" module?

I'm leaning towards just splitting out the `tag_dispatch` functionality into a separate module, and at least #5303 would be solved by that, but I may just not get it yet.

hkaiser: > I'm leaning towards just splitting out the `tag_dispatch` functionality into a separate module, and at least #5303 would be solved by that, but I may just not get it yet.

No objections here. As said, I wanted to leverage the work @K-ballo had invested. Just isolating `tag_dispatch` would be equally fine to resolve #5303. I'll have a look.

hkaiser:
This is superseded by #5464

Issue #5381, opened 2021-06-10
Add starts_with and ends_with algorithms


hkaiser:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p1659r3.html proposes the `ranges::starts_with` and `ranges::ends_with` which have been added to the C++23 draft. We should add those algorithms to our parallel algorithms.

hkaiser:
This was fixed by merging #5410

PR #5410, opened 2021-06-23
Add ranges starts_with and ends_with algorithms

Jedi18:
Added the ranges::starts_with and ranges::ends_with algorithms. 

Issue #5381


libs/parallelism/algorithms/include/hpx/parallel/algorithms/starts_ends_with.hpp
```diff
@@ -0,0 +1,132 @@
...
+#include <hpx/parallel/util/projection_identity.hpp>
+
+#include <algorithm>
+#include <cstddef>
+#include <iterator>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+namespace hpx { namespace parallel { inline namespace v1 {
```
Jedi18, line 31: Should the base implementation be here or should they be in their respective `container_algorithms/starts_with.hpp` and `container_algorithms/ends_with.hpp`?

hkaiser: Let's have them in separate headers to go along with the conventions applied for the other algorithms.

hkaiser (commented):
Do you plan to provide parallelized implementations?

libs/parallelism/algorithms/include/hpx/parallel/algorithms/starts_ends_with.hpp
hkaiser: > Let's have them in separate headers to go along with the conventions applied for the other algorithms.

Ok, I misunderstood what you were asking. Let's have the base implementations here. Could those be made independent of the ranges headers, though? I would like to avoid for the algorithms headers to depend on ranges headers.

Jedi18:
@hkaiser Sure, you mean instead of relying on the mismatch and equal algorithms, write our own implementation?

libs/parallelism/algorithms/include/hpx/parallel/algorithms/starts_ends_with.hpp
Jedi18: Oh ok sure, can I use the parallel headers of mismatch and equal?

Unless as mentioned in the other comment, I'll add our own implementation which dosen't use mismatch and equal.

hkaiser:
> @hkaiser Sure, you mean instead of relying on the mismatch and equal algorithms, write our own implementation?

No, relying on those algorithms is fine. I just would like to avoid for the iterator algorithms to rely on the range-algorithms.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/starts_ends_with.hpp
hkaiser: > Oh ok sure, can I use the parallel headers of mismatch and equal?

Sure. I don't see any problems with this. We do this for other algorithms as well (e.g. copy uses for_each, etc.)

Jedi18:
@hkaiser I've added the iterator based overloads as discussed. Are these okay or did you have some others in mind?

hkaiser:
The new algorithms should be added to the file `#include <hpx/parallel/algorithm.hpp>`

libs/parallelism/algorithms/include/hpx/parallel/algorithms/ends_with.hpp
```diff
@@ -0,0 +1,301 @@
...
+                hpx::parallel::traits::is_indirect_callable<
+                    hpx::execution::sequenced_policy, Pred,
+                    hpx::parallel::traits::projected<Proj1, InIter1>,
+                    hpx::parallel::traits::projected<Proj2, InIter2>
+                >::value
+            )>
+        // clang-format on
+        friend bool tag_fallback_dispatch(hpx::ends_with_t, InIter1 first1,
+            InIter1 last1, InIter2 first2, InIter2 last2, Pred&& pred = Pred(),
+            Proj1&& proj1 = Proj1(), Proj2&& proj2 = Proj2())
```
hkaiser, line 249: The algorithms in `std` namespace usually don't have projections, only the ones in `std::ranges` do have those.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/starts_with.hpp
```diff
@@ -0,0 +1,332 @@
...
+                hpx::parallel::traits::is_indirect_callable<
+                    hpx::execution::sequenced_policy, Pred,
+                    hpx::parallel::traits::projected<Proj1, InIter1>,
+                    hpx::parallel::traits::projected<Proj2, InIter2>
+                >::value
+            )>
+        // clang-format on
+        friend bool tag_fallback_dispatch(hpx::starts_with_t, InIter1 first1,
+            InIter1 last1, InIter2 first2, InIter2 last2, Pred&& pred = Pred(),
+            Proj1&& proj1 = Proj1(), Proj2&& proj2 = Proj2())
```
hkaiser, line 280: Same here, I'd suggest leaving off the projections from the iterator-based overloads of the new algorithms.

hkaiser (changes requested):
Otherwise, LGTM, thanks!

hkaiser (dismissed):
LGTM, thanks! Sorry it took so long.

hkaiser:
@Jedi18 could you please fix the conflict? Once that's done this can be merged.

Jedi18:
Could you hold on the merging for a bit? starts_with range test seems to be failing, I'll have a look at it.

Jedi18:
@msimberg Ok so the starts_with range test crashes. I looked into it and it seems the error is not with this PR code but with the mismatch parallel implementation. I ran the mismatch_binary range test and that crashes the same way too.
This test was running earlier (before the rebase) so I guess the crash is due to some recent changes. Any idea what it could be?

msimberg:
> @msimberg Ok so the starts_with range test crashes. I looked into it and it seems the error is not with this PR code but with the mismatch parallel implementation. I ran the mismatch_binary range test and that crashes the same way too.
> This test was running earlier (before the rebase) so I guess the crash is due to some recent changes. Any idea what it could be?

Nothing relevant comes to mind. What do you mean by crashing? The test fails on CI, but are you seeing something worse (segfault or similar)?

Jedi18:
> > @msimberg Ok so the starts_with range test crashes. I looked into it and it seems the error is not with this PR code but with the mismatch parallel implementation. I ran the mismatch_binary range test and that crashes the same way too.
> > This test was running earlier (before the rebase) so I guess the crash is due to some recent changes. Any idea what it could be?
> 
> Nothing relevant comes to mind. What do you mean by crashing? The test fails on CI, but are you seeing something worse (segfault or similar)?

Yes it segfaults when running locally (when passing in par(task))

msimberg:
> Yes it segfaults when running locally (when passing in par(task))

Would you mind looking a bit deeper into it? I wonder if that only happens on windows by coincidence or if there's some more fundamental reason to it.

Jedi18:
Sure I'll look into it! It happens on master too btw.

Jedi18:
I think it's a visual studio thing because it happens even if I go back to a commit from three months ago. Also it only happens in Debug. Something related to the destruction of the futures I think though it's hard to tell. Do you want me to dig deeper?

hkaiser:
> I think it's a visual studio thing because it happens even if I go back to a commit from three months ago. Also it only happens in Debug. Something related to the destruction of the futures I think though it's hard to tell. Do you want me to dig deeper?

Well, yes, we need to fix it. How can I reproduce the issues you're investigating?

Jedi18:
@hkaiser Ok I will. Run the mismatch_binary test (on master or this branch, dosen't matter) in Debug mode. Does it show a read access violation for you too?

hkaiser:
> @hkaiser Ok I will. Run the mismatch_binary test (on master or this branch, dosen't matter) in Debug mode. Does it show a read access violation for you too?

Yes, I can reproduce this problem. Let me see what I can do.

hkaiser:
@Jedi18 sorry for the delay. Could you please rebase this so we can finally merge it?

aurianer:
Looks good! I just reran build-and-test cause it timed out

PR #5418, opened 2021-06-28
Reduce max size and number of files in ccache cache

msimberg:
Reducing the cache size since the files count towards a shared quota. Hoping that it won't affect build speeds too dramatically.


hkaiser (approved):
LGTM, thanks!

aurianer (approved):
LGTM thanks!

PR #5420, opened 2021-06-29
Update logic to find threading library

### LINKED ISSUE #5412: Link error
## Expected Behavior

Build without link error.

## Actual Behavior

When using HPX CMake targets it seems pthread is not in the list of libs to link against.

  1. Configure HPX without any arguments.
  1. Compile a simple Hello World program
  1. Link against hpx, wrap_main, iostreams_component

## Specifications

  - HPX Version: 1.7.0-rc2
  - Platform (compiler, OS): gcc / linux (kubuntu 20.04)

It seems relevant that the libasio-dev package is used to provide asio. When configuring HPX with `HPX_WITH_FETCH_ASIO=ON` all is well.

When passing `pthread` to CMake's `target_link_libraries` (in addition to HPX' targets), the link error is gone.

msimberg:
Fixes #5412 and #4942.

Removes the custom logic for adding pthread flags and uses `FindThreads` exclusively. The manually added pthread flags were interfering with CMake's detection for if `-pthread` should be added. However, the manually added flags were also not in the public interface, leading to linker errors in certain configurations.

Removing the manually added flags also lead to a different path in `FindThreads.cmake` which was triggering #4942. I've added a workaround for that, which is to filter out the `Threads::Threads` target from the pkgconfig file generation. *This may mean that pkgconfig file users have to add `-pthread` manually*. @stevenrbrandt is this an acceptable workaround for you?

@kordejong this is mostly the same changes that you tested earlier, but since I did reshuffle things a bit I'd be grateful if you could give these changes a try as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1730,8 +1730,6 @@ if(NOT WIN32)
   # System libraries
   # ############################################################################
   if(NOT MSVC)
-    hpx_add_compile_flag_if_available(-pthread)
-    hpx_add_link_flag_if_available(-pthread TARGETS EXE SHARED)
     if(HPX_CXX11_STD_ATOMIC_LIBRARIES)
       target_link_libraries(
         hpx_base_libraries INTERFACE ${HPX_CXX11_STD_ATOMIC_LIBRARIES}
@@ -2146,6 +2144,8 @@ endif()
 # Find Our dependencies: These are all dependencies needed to build the core
 # library. Dependencies that are only needed by plugins, examples or tests
 # should be found separately in the appropriate subdirectory.
+
+include(HPX_SetupThreads)
 
 # Setup our required Boost libraries.
 include(HPX_SetupBoost)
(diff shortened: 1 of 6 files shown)


kordejong:
@msimberg I installed the Ubuntu libasio-dev package, built HPX and my hello-world target now links fine, as before.

BTW, do you recommend using the system's Asio or fetching a (newer?) version of Asio?

msimberg:
> @msimberg I installed the Ubuntu libasio-dev package, built HPX and my hello-world target now links fine, as before.
> 
> BTW, do you recommend using the system's Asio or fetching a (newer?) version of Asio?

Thanks for checking!

I think all relatively recent versions of Asio will be fine. I've tested 1.16 myself and it works fine. We have 1.18.X as the tagged version when it's fetched. However, we use only a small subset of Asio that changes relatively raraely, so older versions are likely to work as well.

hkaiser (approved):
LGTM, thanks!

PR #5425, opened 2021-06-30
Adding the missing defined for HPX_HAVE_DEPRECATION_WARNINGS

shahrzad:


hkaiser:
add to whitelist

hkaiser (approved):
LGTM, thanks!

aurianer (approved):
Thanks!

libs/core/config/include/hpx/config/attributes.hpp
```diff
@@ -63,7 +63,7 @@
 
 ///////////////////////////////////////////////////////////////////////////////
 // handle [[deprecated]]
-#if HPX_HAVE_DEPRECATION_WARNINGS && !defined(HPX_INTEL_VERSION)
+#if defined(HPX_HAVE_DEPRECATION_WARNINGS) && !defined(HPX_INTEL_VERSION)
```
msimberg, line 66: Um, I think it was not using `defined` on purpose. `HPX_HAVE_DEPRECATION_WARNINGS` is added with `hpx_add_config_cond_define` which means that one can add `-DHPX_HAVE_DEPRECATION_WARNINGS=0` to manually disable deprecation warnings even if HPX was built with deprecation warnings on.

Did not having `defined(HPX_HAVE_DEPRECATION_WARNINGS)` cause problems?

hkaiser: @msimberg You are certainly right. I asked @shahrzad to create this PR but I missed that we left it that way on purpose :/ Should we roll it back?

msimberg: I would, yes, unless this really fixed something that was broken? And if it fixed something that was broken, the fix should probably look different.

hkaiser: See #5430

PR #5429, opened 2021-07-03
Scheduler executor

hkaiser:
This implements an executor that relies on any sender/receiver scheduler

### DIFF
--- a/libs/parallelism/executors/CMakeLists.txt
+++ b/libs/parallelism/executors/CMakeLists.txt
@@ -26,6 +26,7 @@ set(executors_headers
     hpx/executors/parallel_executor_aggregated.hpp
     hpx/executors/parallel_executor.hpp
     hpx/executors/restricted_thread_pool_executor.hpp
+    hpx/executors/scheduler_executor.hpp
     hpx/executors/sequenced_executor.hpp
     hpx/executors/service_executors.hpp
     hpx/executors/std_execution_policy.hpp
--- a/libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
+++ b/libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
@@ -1,0 +1,256 @@
+//  Copyright (c) 2021 Hartmut Kaiser
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+/// \file parallel/executors/scheduler_executor.hpp
+
+#pragma once
+
+#include <hpx/config.hpp>
+#include <hpx/datastructures/tuple.hpp>
+#include <hpx/execution/algorithms/bulk.hpp>
+#include <hpx/execution/algorithms/detach.hpp>
+#include <hpx/execution/algorithms/keep_future.hpp>
+#include <hpx/execution/algorithms/make_future.hpp>
+#include <hpx/execution/algorithms/on.hpp>
+#include <hpx/execution/algorithms/sync_wait.hpp>
+#include <hpx/execution/algorithms/transform.hpp>
+#include <hpx/execution/executors/execution.hpp>
+#include <hpx/execution/executors/execution_parameters.hpp>
+#include <hpx/execution_base/execution.hpp>
+#include <hpx/execution_base/sender.hpp>
+#include <hpx/execution_base/traits/is_executor.hpp>
+#include <hpx/functional/bind_back.hpp>
+#include <hpx/functional/bind_front.hpp>
+#include <hpx/functional/deferred_call.hpp>
+#include <hpx/functional/invoke_fused.hpp>
+
+#include <exception>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+namespace hpx { namespace execution { namespace experimental {
+
+    namespace detail {
+#if defined(HPX_HAVE_CXX20_PERFECT_PACK_CAPTURE)
+        template <typename F, typename... Ts>
+        auto captured_args_then(F&& f, Ts&&... ts)
+        {
+            return [f = std::forward<F>(f), ... ts = std::forward<Ts>(ts)](
+                       auto i, auto&& predecessor, auto& v) mutable {
+                v[i] = HPX_INVOKE(std::forward<F>(f), i,
+                    std::forward<decltype(predecessor)>(predecessor),
+                    std::forward<Ts>(ts)...);
+            };
+        }
+#else
+        template <typename F, typename... Ts>
+        auto captured_args_then(F&& f, Ts&&... ts)
+        {
+            return [f = std::forward<F>(f),
+                       t = hpx::make_tuple(std::forward<Ts>(ts)...)](
+                       auto i, auto&& predecessor, auto& v) mutable {
+                v[i] = hpx::util::invoke_fused(
+                    hpx::util::bind_front(std::forward<F>(f), i,
+                        std::forward<decltype(predecessor)>(predecessor)),
+                    std::move(t));
+            };
+        }
+#endif
+    }    // namespace detail
+
+    ///////////////////////////////////////////////////////////////////////////
+    // A scheduler_executor wraps any P2300 scheduler and implements the
+    // executor functionalities for those.
+    template <typename BaseScheduler>
+    struct scheduler_executor
+    {
+        template <typename Scheduler,
+            typename Enable = std::enable_if_t<
+                hpx::execution::experimental::is_scheduler_v<Scheduler>>>
+        constexpr explicit scheduler_executor(Scheduler&& sched)
+          : sched_(std::forward<Scheduler>(sched))
+        {
+        }
+
+        /// \cond NOINTERNAL
+        constexpr bool operator==(scheduler_executor const& rhs) const noexcept
+        {
+            return sched_ == rhs.sched_;
+        }
+
+        constexpr bool operator!=(scheduler_executor const& rhs) const noexcept
+        {
+            return sched_ != rhs.sched_;
+        }
+
+        constexpr auto const& context() const noexcept
+        {
+            return *this;
+        }
+        /// \endcond
+
+        /// \cond NOINTERNAL
+        // Associate the parallel_execution_tag executor tag type as a default
+        // with this executor.
+        using execution_category = parallel_execution_tag;
+
+        // Associate the static_chunk_size executor parameters type as a default
+        // with this executor.
+        using executor_parameters_type = static_chunk_size;
+
+        template <typename T, typename... Ts>
+        using future_type = hpx::future<T>;
+
+        // NonBlockingOneWayExecutor interface
+        template <typename F, typename... Ts>
+        void post(F&& f, Ts&&... ts)
+        {
+            detach(transform(schedule(sched_),
+                hpx::util::deferred_call(
+                    std::forward<F>(f), std::forward<Ts>(ts)...)));
+        }
+
+        // OneWayExecutor interface
+        template <typename F, typename... Ts>
+        decltype(auto) sync_execute(F&& f, Ts&&... ts)
+        {
+            return sync_wait(transform(schedule(sched_),
+                hpx::util::deferred_call(
+                    std::forward<F>(f), std::forward<Ts>(ts)...)));
+        }
+
+        // TwoWayExecutor interface
+        template <typename F, typename... Ts>
+        decltype(auto) async_execute(F&& f, Ts&&... ts)
+        {
+            return make_future(transform(schedule(sched_),
+                hpx::util::deferred_call(
+                    std::forward<F>(f), std::forward<Ts>(ts)...)));
+        }
+
+        template <typename F, typename Future, typename... Ts>
+        decltype(auto) then_execute(F&& f, Future&& predecessor, Ts&&... ts)
+        {
+            auto&& predecessor_on_sched =
+                on(keep_future(std::forward<Future>(predecessor)), sched_);
+
+            return make_future(transform(std::move(predecessor_on_sched),
+                hpx::util::bind_back(
+                    std::forward<F>(f), std::forward<Ts>(ts)...)));
+        }
+
+        // BulkTwoWayExecutor interface
+        template <typename F, typename S, typename... Ts>
+        decltype(auto) bulk_async_execute(F&& f, S const& shape, Ts&&... ts)
+        {
+            using shape_element =
+                typename hpx::traits::range_traits<S>::value_type;
+            using result_type = hpx::util::detail::invoke_deferred_result_t<F,
+                shape_element, Ts...>;
+
+            std::vector<hpx::future<result_type>> results;
+            results.reserve(hpx::util::size(shape));
+
+            for (auto const& s : shape)
+            {
+                results.push_back(async_execute(f, s, ts...));
+            }
+
+            return results;
+        }
+
+        template <typename F, typename S, typename... Ts>
+        decltype(auto) bulk_sync_execute(F&& f, S const& shape, Ts&&... ts)
+        {
+            return sync_wait(bulk(schedule(sched_), shape,
+                hpx::util::bind_back(
+                    std::forward<F>(f), std::forward<Ts>(ts)...)));
+        }
+
+        template <typename F, typename S, typename Future, typename... Ts>
+        decltype(auto) bulk_then_execute(
+            F&& f, S const& shape, Future&& predecessor, Ts&&... ts)
+        {
+            using result_type =
+                parallel::execution::detail::then_bulk_function_result_t<F, S,
+                    Future, Ts...>;
+
+            if constexpr (std::is_void<result_type>::value)
+            {
+                // the overall return value is future<void>
+                auto prereq =
+                    when_all(keep_future(std::forward<Future>(predecessor)));
+
+                auto loop = bulk(on(std::move(prereq), sched_), shape,
+                    hpx::util::bind_back(
+                        std::forward<F>(f), std::forward<Ts>(ts)...));
+
+                return make_future(std::move(loop));
+            }
+            else
+            {
+                // the overall return value is future<std::vector<result_type>>
+                auto prereq =
+                    when_all(keep_future(std::forward<Future>(predecessor)),
+                        just(std::vector<result_type>(hpx::util::size(shape))));
+
+                auto loop = bulk(on(std::move(prereq), sched_), shape,
+                    detail::captured_args_then(
+                        std::forward<F>(f), std::forward<Ts>(ts)...));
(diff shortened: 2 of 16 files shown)


libs/parallelism/executors/include/hpx/executors/sender_receiver_executor.hpp
```diff
@@ -0,0 +1,244 @@
...
+                hpx::util::invoke_result_t<F, shape_element, Ts...>;
+
+            std::vector<hpx::future<result_type>> results;
+            results.reserve(hpx::util::size(shape));
+
+            for (auto const& s : shape)
+            {
+                results.push_back(schedule(sched_) |
+                    transform(detail::captured_args(f, s, ts...)) |
+                    make_future());
```
msimberg, line 181:
```suggestion
                results.push_back(async_execute(f, s, ts...);
```
?

msimberg (commented):
Cool!

libs/parallelism/executors/include/hpx/executors/sender_receiver_executor.hpp
```diff
@@ -0,0 +1,244 @@
...
+                        std::forward<F>(f), std::forward<decltype(arg)>(arg)),
+                    std::move(t));
+            };
+        }
+#endif
+    }    // namespace detail
+
+    ///////////////////////////////////////////////////////////////////////////
+    /// A \a sender_receiver_executor wraps any other executor and adds the
+    /// capability to add annotations to the launched threads.
```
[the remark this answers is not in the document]
hkaiser: uhh, copy/paste snafu.

hkaiser:
Note: `the `then_bulk_execute` implementation is still missing. I think we would need something like `bulk_schedule`/`bulk_join` to express this functionality any better than what the emulation is already doing.

libs/parallelism/executors/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ set(executors_headers
     hpx/executors/parallel_executor.hpp
     hpx/executors/restricted_thread_pool_executor.hpp
     hpx/executors/sequenced_executor.hpp
+    hpx/executors/sender_receiver_executor.hpp
```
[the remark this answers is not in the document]
hkaiser: I don't like `sender_receiver_executor` either, `scheduler_executor` is better.

libs/parallelism/executors/include/hpx/executors/sender_receiver_executor.hpp
```diff
@@ -0,0 +1,244 @@
...
+        decltype(auto) bulk_sync_execute(F&& f, S const& shape, Ts&&... ts)
+        {
+            return schedule(sched_) |
+                bulk(shape,
+                    detail::captured_args_unary(
+                        std::forward<F>(f), std::forward<Ts>(ts)...)) |
+                sync_wait();
+        }
+
+        // FIXME: implement bulk_then_execute
```
msimberg, line 197: This is supposed to return a future of a vector (container) of the results of each invocation of `f`, correct? If yes, would this do:

```
return just_on(sched_, std::vector<result_type>(size(shape))) |
  bulk(shape, [=](auto i, std::vector<result_type> v) {
    v[i] = HPX_INVOKE(f, ts...);
  }) | 
  make_future();
```
(ignoring forwarding, iterators vs indices etc.)?

If no, then maybe not... Could you expand on what you had in mind for what we could do differently/better than the current emulation?

hkaiser: What about running things concurrently? Also, it should depend on the future passed in. Otherwise, yes - the function should return a `future<vector<result_type>>`, where `result_type` is whatever `F` returns (or simply a `future<void>` if `F` returns `void`)..

msimberg: > What about running things concurrently?

That should be taken care of by `bulk`, no? There are of course two routes for this, 1. use the emulation fallback which uses `async_execute` and 2. use `bulk`. Since this is `(P2300/P0443)scheduler_executor` I think it makes most sense to use `bulk` and if there's no customization of `bulk` for the scheduler and it does a serial for-loop then so be it.

> Also, it should depend on the future passed in.

Uh, of course... something like this:

```
return when_all(keep_future(predecessor), just(std::vector<result_type>(size(shape)))) |
  on(sched_) |
  bulk(shape, [=](auto i, auto predecessor, std::vector<result_type> v) {
    v[i] = HPX_INVOKE(f, predecessor.get(), ts...);
  }) |
  transform([](auto future, std::vector<result_type> v) {
    return v;
  }) |
  make_future();
```
?

That should:
- run when the predecessor future is ready
- pass the future value and the additional arguments to f
- return only the vector of results as a future

hkaiser: > > What about running things concurrently?
> 
> That should be taken care of by `bulk`, no? There are of course two routes for this, 1. use the emulation fallback which uses `async_execute` and 2. use `bulk`. Since this is `(P2300/P0443)scheduler_executor` I think it makes most sense to use `bulk` and if there's no customization of `bulk` for the scheduler and it does a serial for-loop then so be it.

I assumed that `bulk()` would simply run things sequentially as it does not take a scheduler as one of its argument (at least now). Once we introduce the `get_completion_scheduler` CPO that could be used for this purpose, however.

hkaiser: Allowing `bulk()` to run things concurrently would also assume that it is fork/join. Not sure if using it in this case would be a good idea....

msimberg: > I assumed that bulk() would simply run things sequentially as it does not take a scheduler as one of its arguments (at least now). Once we introduce the get_completion_scheduler CPO that could be used for this purpose, however.

Yes, the dispatching has to be changed to allow a scheduler overload to be used as well, but that's exactly the plan in my understanding. Running things sequentially is just meant to be the default.

> Allowing `bulk()` to run things concurrently would also assume that it is fork/join. Not sure if using it in this case would be a good idea....

What other option is there (ok, running things sequentially...)? Or why would it not be a good idea? If one doesn't want fork/join or concurrency one should pick a scheduler that doesn't allow concurrency (`inline_scheduler` or something like that?).

hkaiser:
I have added the `bulk_then_execute` implementation now. This PR is ready to be reviews.

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
```diff
@@ -0,0 +1,319 @@
...
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+/// \file parallel/executors/scheduler_executor.hpp
+
+#pragma once
+
+#include <hpx/config.hpp>
+#if defined(HPX_HAVE_CXX17_STD_VARIANT)
```
msimberg, line 12: We should be able to do without this now that #5283 is merged.

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
```diff
@@ -0,0 +1,319 @@
...
+        constexpr auto const& context() const noexcept
+        {
+            return *this;
+        }
+        /// \endcond
+
+        /// \cond NOINTERNAL
+        // Associate the parallel_execution_tag executor tag type as a default
+        // with this executor.
+        using execution_category = parallel_execution_tag;
```
msimberg, line 163: This is somewhat questionable, but probably the right default since the scheduler *can* run things in parallel but doesn't have to. We don't have a way to query the underlying scheduler if it will actually run `bulk` in parallel. Not really expecting a change here, but just wanted to point it out.

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
```diff
@@ -0,0 +1,319 @@
...
+        template <typename T, typename... Ts>
+        using future_type = hpx::future<T>;
+
+        // NonBlockingOneWayExecutor interface
+        template <typename F, typename... Ts>
+        void post(F&& f, Ts&&... ts)
+        {
+            detach(transform(schedule(sched_),
+                detail::captured_args(
+                    std::forward<F>(f), std::forward<Ts>(ts)...)));
```
msimberg, line 178: Is `capture_args` different than `deferred_call`? This question applies to all the captures below as well.

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
```diff
@@ -0,0 +1,319 @@
...
+        }
+
+        template <typename F, typename Future, typename... Ts>
+        decltype(auto) then_execute(F&& f, Future&& predecessor, Ts&&... ts)
+        {
+            auto&& sched =
+                on(keep_future(std::forward<Future>(predecessor)), sched_);
+
+            return make_future(transform(std::move(sched),
```
msimberg, line 205:
```suggestion
            auto&& predecessor_on_sched =
                on(keep_future(std::forward<Future>(predecessor)), sched_);

            return make_future(transform(std::move(predecessor_on_sched),
```
or something in that direction? I don't particularly like that `sched_` and `sched` are completely different things with almost the same name.

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
```diff
@@ -0,0 +1,319 @@
...
+            using result_type = hpx::util::detail::invoke_deferred_result_t<F,
+                shape_element, Ts...>;
+
+            std::vector<hpx::future<result_type>> results;
+            results.reserve(hpx::util::size(shape));
+
+            for (auto const& s : shape)
+            {
+                results.push_back(async_execute(f, s, ts...));
+            }
```
msimberg, line 225: Should this not also be based on `bulk`? Something in the direction of this:
```
return make_future(bulk(schedule(sched_), size(shape), something_to_bind_ts_and_possibly_a_vector_of_results);
```
Or if we're lazy:
```
return bulk_then_execute(std::forward<F>(f), shape, make_ready_future(), std::forward<Ts>(ts)...);
```
?
```

File: libs/parallelism/executors/tests/unit/scheduler_executor.cpp [hunk 10]
```diff
@@ -0,0 +1,300 @@
...
+
+        auto result = hpx::parallel::execution::bulk_then_execute(
+            exec, hpx::util::bind(&bulk_test_f, _1, _2, _3), 107, f, 42)
+                          .get();
+
+        HPX_TEST(executed);
+        HPX_TEST(result.size() == 107);
+
+        int expected = 0;
+        for (auto i: result)
```
msimberg, line 192: And as a side-note, this needs formatting:
```suggestion
        for (auto i : result)
```

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
hkaiser: Yes, I wasn't sure what to do here. However I believe we don't really use this tag for anything. We mostly rely on functions being available (or not). Will the schedulers expose their capabilities in any way?

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
hkaiser: The `captured_args` helper is very similar to `deferred_call`, indeed. It however inserts an extra argument into the embedded function call. Those could be replaced with something like `bind(f, _1, ts...)` - I'll try that.

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
hkaiser: Nod, I'll try that.

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
msimberg: There's  currently nothing in P2300 similar to this tag. There have been questions about if `bulk` should support something like execution policies or if that should just be up to the scheduler (and not communicated outside of the scheduler), but no concensus as far as I can tell. I think it's rather far down on the list of open questions about senders/receivers at the moment.

I think the only place we rely on the tag is e.g. if you do `seq.on(exec)`, where the tag will be checked. `seq` requires `sequenced_execution_tag` whereas `par` requires either of the two. If `exec` is a `scheduler_executor` which only does a sequential `bulk_*` then `seq.on(exec)` will wrongly not compile, I think. However, I'm not too bothered by this.

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
hkaiser: The second option wouldn't really work as the return types of `bulk_async_execute` and `bulk_then_execute` are slightly different. `bulk_async_execute` returns a `vector<future<result_type>>`, while `bulk_then_execute` returns a `future<vector<result_type>>`. I will wait for `bulk` to support concurrent scheduling before changing the implementation here.

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
hkaiser: FWIW, in order to use bind() we need to merge #5489 first.

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
msimberg: > The second option wouldn't really work as the return types of bulk_async_execute and bulk_then_execute are slightly different. bulk_async_execute returns a vector<future<result_type>>, while bulk_then_execute returns a future<vector<result_type>>.
I missed that, good point.

> I will wait for bulk to support concurrent scheduling before changing the implementation here.

Does it make a difference? The implementation here should look the same before and after #5487.

hkaiser: > Does it make a difference? The implementation here should look the same before and after #5487.

*grumble* I was hoping to get away with it,  but you caught me... :-)

msimberg: I don't mind doing it if you'd like :P I just wanted to understand if there was a technical reason for not changing it now.

msimberg:
retest cscs

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
hkaiser: @msimberg I give up, I can't see how this could possibly work. I have removed most of the custom binders now, please feel free to take a stab at `bulk_async_execute` if you like.

msimberg: I'll have a look. Perhaps I was too optimistic and I'll get stuck as well...

msimberg: I've pushed an attempt at it, but I can see now why you didn't like it. I don't particularly like it either as it's far from a portable definition (this won't work with a GPU scheduler).

The problem is really the return type. If we were to accept the same relaxation `parallel_executor_aggregated` has on its return type from `bulk_async_execute` things would be significantly simpler, but it does of course lose the futures to the individual tasks (which is anyway something that mainly makes sense on the host, not on a GPU).

msimberg:
As a flyby I added a CTAD guide for `scheduler_executor`, and opened #5497 to list things we can change now that we require C++17.

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
hkaiser: As discussed, I think we can safely return a `std::vector<hpx::future<void>>`  with one element in case the return type is `void`. Otherwise, I think we should go with what you suggest. For the former, we might even get away with returning a `std::array<hpx::future<void>, 1>`, this might require (I'm not sure) slight changes to the algorithms, however.

msimberg: Updated. I did not make it a `std::array<hpx::future<void>, 1>` since that would have required larger changes to the algorithms.

As another flyby, I did what should've been done in #5455: remove the `as_sender` fallback for `schedule`. It turned out that the `scheduler_executor` was identifying as a scheduler, meaning that the one explicit constructor (https://github.com/STEllAR-GROUP/hpx/blob/c5465617297b1cf8189941c82d93e0c5b315f154/libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp#L70-L76) was being considered instead of the real copy constructor.

libs/parallelism/executors/include/hpx/executors/scheduler_executor.hpp
```diff
@@ -0,0 +1,298 @@
...
+        template <typename F, typename S, typename... Ts>
+        decltype(auto) bulk_async_execute(F&& f, S const& shape, Ts&&... ts)
+        {
+            using shape_element =
+                typename hpx::traits::range_traits<S>::value_type;
+            using result_type = hpx::util::detail::invoke_deferred_result_t<F,
+                shape_element, Ts...>;
+
+            using size_type = decltype(hpx::util::size(shape));
+            size_type const n = hpx::util::size(shape);
```
hkaiser, line 155: `n` is needed for the non-void case only.

msimberg: Good catch, fixed.

hkaiser:
@msimberg I can't approve my own PRs, but I think this is ready to go now.

msimberg:
> @msimberg I can't approve my own PRs, but I think this is ready to go now.

Ah yes... Well, I approve of your changes if you approve of mine.

msimberg (approved):

msimberg:
I force-pushed one last change because of codacy. It found one unused variable in the test.

hkaiser:
This looks good to go now, merging.

msimberg:
@hkaiser what happened to the last changes...? It looks like you may have force-pushed an older version. What were the last changes you wanted to force push? I can restore the other changes if you let me know which of your changes I should keep.

hkaiser:
@msimberg any force push I did was `with_lease`, so I'm not aware of losing any changes - sorry if I did. What is missing?

msimberg:
@hkaiser no worries. It's the `bulk_async_execute` changes (and some other stuff) that is missing. The commits are still there so recovering them is no problem (https://github.com/STEllAR-GROUP/hpx/commit/16e4eb0fecc878e5856fd6ce207bcc3625c0dff8 and a few commits up from that). I was just wondering if the last force push did any functional changes or if it was actually just a rebase?

PR #5431, opened 2021-07-06
Rename the new P0443/P2300 executor to thread_pool_scheduler

msimberg:
I suggest `thread_pool_scheduler` to mirror `thread_pool_executor`, both of which allow scheduling/executing tasks on a given HPX `thread_pool_base`. Other suggestions are welcome.

I'm leaving the "executor" functionality for now in `thread_pool_scheduler` but it's likely that the change to make `execute` a plain algorithm in P2300 will likely be supported. I'll remove it separately (there's a meeting today on P2300, and unless the feedback is very negative I think we can start applying changes from it more widely).

Edit: I also wanted to add that I'm not sure if the thread pool attached to the scheduler should be a "property" or not, so it's currently not one (it's also not a property at the moment for `thread_pool_executor`/`parallel_executor`). If someone has a good argument for either case I'd love to hear it.

### DIFF
--- a/libs/core/synchronization/tests/unit/async_rw_mutex.cpp
+++ b/libs/core/synchronization/tests/unit/async_rw_mutex.cpp
@@ -19,9 +19,9 @@
 #include <vector>
 
 using hpx::execution::experimental::execute;
-using hpx::execution::experimental::executor;
 using hpx::execution::experimental::on;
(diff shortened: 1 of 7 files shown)


hkaiser (dismissed):
LGTM, thanks!

msimberg:
retest cscs

PR #5432, opened 2021-07-06
Add deprecation warnings for v1.8

Jedi18:
Add deprecation warnings for the next version (i.e. v1.8).


Jedi18:
Shouldn't the values compared to HPX_VERSION_FULL in deprecation.hpp be in hexadecimal(0x010700) instead of the octal values they're currently in?

hkaiser:
> Shouldn't the values compared to HPX_VERSION_FULL in deprecation.hpp be in hexadecimal(0x010700) instead of the octal values they're currently in?

Yes, good catch. That's definitely a bug.

Jedi18:
Ok, I'll change the rest of the values too then

msimberg (approved):

PR #5434, opened 2021-07-06
Update CUDA polling logging to be more verbose

msimberg:
@aurianer I think this fixes the inconsistency you noticed on #5342. I just made the logging statements print both the queue and vector sizes each time as both may be interesting. Could you please have a look?


aurianer (approved):
Thanks! I've done a similar renaming for the MPI refactoring as I was a bit lost with the names, will make it match with this one! :ok_hand:

PR #5435, opened 2021-07-06
This adds support for systems that assume all types are bitwise serializable by default

hkaiser:
- this also allows for certain types to be normally serialized, still

More explanations of the rationale for this:

HPX so far assumes that only types that are specifically marked as such are bitwise serializable. This is fine in normal situations. Certain systems (like [SHAD](https://github.com/pnnl/SHAD)) however assume the opposite: all types are bitwise serializable by default, and only few require proper serialization. This PR adds an option to HPX that allows to support such systems by defining the new configuration option `HPX_SERIALIZATION_WITH_ALL_TYPES_ARE_BITWISE_SERIALIZABLE=On`. This also adds a new helper macro `HPX_IS_NOT_BITWISE_SERIALIZABLE(T)` which marks `T` as not bitwise serializable.

@msimberg I'd appreciate it if this PR could still make it into the release.

### DIFF
--- a/libs/core/serialization/CMakeLists.txt
+++ b/libs/core/serialization/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2019-2020 The STE||AR-Group
+# Copyright (c) 2019-2021 The STE||AR-Group
 #
 # SPDX-License-Identifier: BSL-1.0
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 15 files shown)


msimberg (dismissed):
Thanks @hkaiser, let's add this to the release. I think only `cmake-format` needs addressing before merging.

hkaiser:
> Thanks @hkaiser, let's add this to the release. I think only `cmake-format` needs addressing before merging.

I'll take care of that. There is also one more things I would like to add here, namely make sure our own types are handled correctly in this mode.

hkaiser:
@msimberg I'm done with this. If the CI comes back ok we can merge.

PR #5437, opened 2021-07-07
Add back explicit -pthread flags when available

msimberg:
It turns out that we do actually need the `-pthread` flag when using the Cray compiler, so I'm adding it back here for both compilation and linking (#5420 is responsible for the timeouts on the clang-cuda configuration and I missed this before merging). The flag also now gets passed on to consuming projects unlike before #5420.

@kordejong I know you may be tired of rebuilding HPX, but if you have the time I'd be very grateful if you could also test this on your system to ensure that I haven't brought things back to before #5420...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2015,7 +2015,9 @@ else()
      AND NOT WIN32
      AND NOT HPX_WITH_SANITIZERS
   )
-    hpx_add_link_flag_if_available(-Wl,-z,defs TARGETS SHARED EXE)
+    hpx_add_link_flag_if_available(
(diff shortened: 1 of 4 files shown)


hkaiser (approved):
LGTM, thanks!

msimberg:
This will be the last regular PR (I'm not counting #5417) for 1.7.0 unless there's a really critical regression or bug that still pops up.

kordejong:
> @kordejong I know you may be tired of rebuilding HPX, but if you have the time I'd be very grateful if you could also test this on your system to ensure that I haven't brought things back to before #5420...

@msimberg No problem. My builds succeeded. Your updates seem fine.

msimberg:
retest cscs

PR #5441, opened 2021-07-11
Adding reference counting to thread_data

hkaiser:
This is the first step towards enabling scoped execution of threads

This PR now also has changes that are unrelated to the reference counting of threads but that have been necessary because of ASAN problems that started showing up. These changes are more pervasive than I was hoping for and are related to the initialization of the runtime, AGAS, parcelport, thread-manager, and io_pool classes. These changes are all in one commit (https://github.com/STEllAR-GROUP/hpx/pull/5441/commits/c2965e6bfee08b87582f46c6b0521b15f1cd6738). If needed, I could try separating those into their own PR that would have to be merged first. @msimberg please let me know how you would like to handle this.

### DIFF
--- a/examples/balancing/hpx_thread_phase.cpp
+++ b/examples/balancing/hpx_thread_phase.cpp
@@ -6,9 +6,9 @@
 
 #include <hpx/init.hpp>
 #include <hpx/local/barrier.hpp>
+#include <hpx/local/functional.hpp>
 #include <hpx/local/mutex.hpp>
 #include <hpx/local/thread.hpp>
-#include <hpx/modules/functional.hpp>
 
 #include <boost/lockfree/queue.hpp>
 
@@ -22,37 +22,38 @@
 
 using boost::lockfree::queue;
 
-using hpx::program_options::variables_map;
 using hpx::program_options::options_description;
 using hpx::program_options::value;
+using hpx::program_options::variables_map;
 
 using std::chrono::milliseconds;
 
 using hpx::lcos::local::barrier;
 using hpx::lcos::local::mutex;
 
+using hpx::threads::make_thread_function_nullary;
 using hpx::threads::register_thread;
 using hpx::threads::thread_init_data;
-using hpx::threads::make_thread_function_nullary;
 
-using hpx::threads::get_thread_phase;
-using hpx::threads::get_self_id;
 using hpx::threads::get_self;
-using hpx::threads::thread_id_type;
+using hpx::threads::get_self_id;
+using hpx::threads::get_thread_phase;
 using hpx::threads::set_thread_state;
+using hpx::threads::thread_id_ref_type;
+using hpx::threads::thread_id_type;
 
 typedef std::pair<thread_id_type, std::size_t> value_type;
 typedef std::vector<value_type> fifo_type;
 
 ///////////////////////////////////////////////////////////////////////////////
 void lock_and_wait(mutex& m, barrier& b0, barrier& b1, value_type& entry,
-    std::size_t /* wait */
-)
+    std::size_t /* wait */)
 {
     // Wait for all hpxthreads in this iteration to be created.
     b0.wait();
 
-    const thread_id_type this_ = get_self_id();
+    // keep this thread alive while being suspended
+    thread_id_ref_type this_ = get_self_id();
 
     while (true)
     {
@@ -61,12 +62,12 @@ void lock_and_wait(mutex& m, barrier& b0, barrier& b1, value_type& entry,
 
         if (l.owns_lock())
         {
-            entry = value_type(this_, get_thread_phase(this_));
+            entry = value_type(this_.noref(), get_thread_phase(this_.noref()));
             break;
         }
 
         // Schedule a wakeup.
-        set_thread_state(this_, milliseconds(30),
+        set_thread_state(this_.noref(), milliseconds(30),
             hpx::threads::thread_schedule_state::pending);
 
         // Suspend this HPX thread.
@@ -112,20 +113,19 @@ int hpx_main(variables_map& vm)
         std::vector<mutex> m(mutex_count);
         barrier b0(hpxthread_count + 1), b1(hpxthread_count + 1);
 
+        // keep created threads alive while they are suspended
+        std::vector<thread_id_ref_type> ids;
         for (std::size_t j = 0; j < hpxthread_count; ++j)
         {
             // Compute the mutex to be used for this thread.
             const std::size_t index = j % mutex_count;
 
-            thread_init_data data(make_thread_function_nullary(
-                hpx::util::bind
-                (&lock_and_wait, std::ref(m[index])
-                               , std::ref(b0)
-                               , std::ref(b1)
-                               , std::ref(hpxthreads[j])
-                               , wait))
-              , "lock_and_wait");
-            register_thread(data);
+            thread_init_data data(
+                make_thread_function_nullary(hpx::util::bind(&lock_and_wait,
+                    std::ref(m[index]), std::ref(b0), std::ref(b1),
+                    std::ref(hpxthreads[j]), wait)),
+                "lock_and_wait");
+            ids.push_back(register_thread(data));
         }
 
         // Tell all hpxthreads that they can start running.
@@ -135,7 +135,7 @@ int hpx_main(variables_map& vm)
         b1.wait();
 
         // {{{ Print results for this iteration.
-        for(value_type &entry: hpxthreads)
+        for (value_type& entry : hpxthreads)
         {
             std::cout << "  " << entry.first << "," << entry.second << "\n";
         }
@@ -154,6 +154,7 @@ int main(int argc, char* argv[])
     options_description
        desc_commandline("Usage: " HPX_APPLICATION_STRING " [options]");
 
+    // clang-format off
     desc_commandline.add_options()
         ("hpxthreads,T", value<std::size_t>()->default_value(128),
             "the number of PX threads to invoke")
@@ -164,6 +165,7 @@ int main(int argc, char* argv[])
         ("iterations", value<std::size_t>()->default_value(1),
             "the number of times to repeat the test")
         ;
+    // clang-format on
 
     // Initialize and run HPX.
     hpx::init_params init_args;
@@ -171,4 +173,3 @@ int main(int argc, char* argv[])
 
     return hpx::init(argc, argv, init_args);
 }
-
--- a/examples/balancing/hpx_thread_phase.cpp
+++ b/examples/balancing/hpx_thread_phase.cpp
@@ -8,6 +8,7 @@
 #include <hpx/local/barrier.hpp>
 #include <hpx/local/mutex.hpp>
 #include <hpx/local/thread.hpp>
+#include <hpx/modules/functional.hpp>
 
 #include <boost/lockfree/queue.hpp>
 
--- a/examples/cancelable_action/cancelable_action/server/cancelable_action.hpp
+++ b/examples/cancelable_action/cancelable_action/server/cancelable_action.hpp
@@ -1,4 +1,4 @@
-//  Copyright (c) 2007-2012 Hartmut Kaiser
+//  Copyright (c) 2007-2021 Hartmut Kaiser
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
@@ -11,14 +11,14 @@
 #include <hpx/hpx.hpp>
 #include <hpx/include/actions.hpp>
 #include <hpx/include/components.hpp>
+#include <hpx/include/lcos.hpp>
 #include <hpx/include/threads.hpp>
 #include <hpx/include/util.hpp>
 
-#include <atomic>
+#include <mutex>
 
 ///////////////////////////////////////////////////////////////////////////////
-namespace examples { namespace server
-{
+namespace examples { namespace server {
     ///////////////////////////////////////////////////////////////////////////
     inline void delay(int c)
     {
@@ -50,19 +50,17 @@ namespace examples { namespace server
             explicit reset_id(cancelable_action& this_)
               : outer_(this_)
             {
-                {
-                    hpx::thread::id old_value =
-                        outer_.id_.exchange(hpx::this_thread::get_id());
-                    HPX_ASSERT(old_value == hpx::thread::id());
-                    HPX_UNUSED(old_value);
-                }
+                std::lock_guard<hpx::lcos::local::mutex> l(outer_.mtx_);
+                hpx::thread::id old_value = outer_.id_;
+                outer_.id_ = hpx::this_thread::get_id();
+                HPX_ASSERT(old_value == hpx::thread::id());
+                HPX_UNUSED(old_value);
             }
             ~reset_id()
             {
-                hpx::thread::id old_value =
-                    outer_.id_.exchange(hpx::thread::id());
+                hpx::thread::id old_value = outer_.id_;
+                outer_.id_ = hpx::thread::id();
                 HPX_ASSERT(old_value != hpx::thread::id());
-                HPX_ASSERT(outer_.id_ == hpx::thread::id());
                 HPX_UNUSED(old_value);
             }
 
@@ -73,15 +71,16 @@ namespace examples { namespace server
         // Do some lengthy work
         void do_it()
         {
-            reset_id r(*this);      // manage thread id
+            reset_id r(*this);    // manage thread id
 
-            while (true) {
+            while (true)
+            {
                 // do something useful ;-)
                 delay(1000);
 
                 // check whether this thread was interrupted and throw a special
                 // exception if it was interrupted
-                hpx::this_thread::suspend(); // interruption_point();
+                hpx::this_thread::suspend();    // interruption_point();
             }
         }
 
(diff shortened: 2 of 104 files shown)


msimberg:
@hkaiser I planned to take a look at the `thread_pool_scheduler` test hanging/failling, but I see that it currently doesn't build. Should I wait a bit? Do you have a possible fix for it already?

hkaiser:
> @hkaiser I planned to take a look at the `thread_pool_scheduler` test hanging/failling, but I see that it currently doesn't build. Should I wait a bit? Do you have a possible fix for it already?

Sorry for this. I'll fix it right away. You can go back one commit in the meantime (https://github.com/STEllAR-GROUP/hpx/pull/5441/commits/a3d371b2505fb655e14e04582171b2cf2b86914b), that should compile and will expose the same issue.

msimberg:
@hkaiser this is keeping the `thread_pool_scheduler` test from exiting: https://github.com/STEllAR-GROUP/hpx/blob/877c2be8229a44810d0017ddd24ab8303b3570d2/libs/parallelism/executors/tests/unit/thread_pool_scheduler.cpp#L123. With the changes in this PR, is the recommendation to simply explicitly reset ids like that or is there another way? I see a `thread_id_noref_type`, but that looks like an internal type. To be clear, having a global `hpx::thread::id` like that is very test-specific and I'd hope noone does that in real code so I'd be fine with just explicitly requiring that it be reset if the plan is to keep an id like that alive for a longer time. The documentation for `hpx::this_thread::get_id()` and/or `hpx::thread::id` could probably do with a note about reference counting.

hkaiser:
> @hkaiser this is keeping the `thread_pool_scheduler` test from exiting:
> 
> https://github.com/STEllAR-GROUP/hpx/blob/877c2be8229a44810d0017ddd24ab8303b3570d2/libs/parallelism/executors/tests/unit/thread_pool_scheduler.cpp#L123
> 
> . With the changes in this PR, is the recommendation to simply explicitly reset ids like that or is there another way? I see a `thread_id_noref_type`, but that looks like an internal type. To be clear, having a global `hpx::thread::id` like that is very test-specific and I'd hope noone does that in real code so I'd be fine with just explicitly requiring that it be reset if the plan is to keep an id like that alive for a longer time. The documentation for `hpx::this_thread::get_id()` and/or `hpx::thread::id` could probably do with a note about reference counting.

Thanks for diagnosing this! Not sure why it doesn't show up on all platforms, though... I think we can make the `hpx::thread::id` non-reference counting, however. That should resolve the issue.

msimberg:
> > @hkaiser this is keeping the `thread_pool_scheduler` test from exiting:
> > https://github.com/STEllAR-GROUP/hpx/blob/877c2be8229a44810d0017ddd24ab8303b3570d2/libs/parallelism/executors/tests/unit/thread_pool_scheduler.cpp#L123
> > 
> > . With the changes in this PR, is the recommendation to simply explicitly reset ids like that or is there another way? I see a `thread_id_noref_type`, but that looks like an internal type. To be clear, having a global `hpx::thread::id` like that is very test-specific and I'd hope noone does that in real code so I'd be fine with just explicitly requiring that it be reset if the plan is to keep an id like that alive for a longer time. The documentation for `hpx::this_thread::get_id()` and/or `hpx::thread::id` could probably do with a note about reference counting.
> 
> Thanks for diagnosing this! Not sure why it doesn't show up on all platforms, though... I think we can make the `hpx::thread::id` non-reference counting, however. That should resolve the issue.

If by platform, you mean CI configuration, then most likely because that test is only built and run on those that are using C++17 or higher.

examples/balancing/hpx_thread_phase.cpp
```diff
@@ -8,6 +8,7 @@
 #include <hpx/local/barrier.hpp>
 #include <hpx/local/mutex.hpp>
 #include <hpx/local/thread.hpp>
+#include <hpx/modules/functional.hpp>
```
msimberg, line 11:
```suggestion
#include <hpx/local/functional.hpp>
```

examples/cancelable_action/cancelable_action/server/cancelable_action.hpp
```diff
@@ -50,19 +50,17 @@ namespace examples { namespace server
...
               : outer_(this_)
             {
-                {
-                    hpx::thread::id old_value =
-                        outer_.id_.exchange(hpx::this_thread::get_id());
-                    HPX_ASSERT(old_value == hpx::thread::id());
-                    HPX_UNUSED(old_value);
-                }
+                std::lock_guard<hpx::lcos::local::mutex> l(outer_.mtx_);
+                hpx::thread::id old_value = outer_.id_;
```
msimberg, line 54: Just curious, why does this now require a lock rather than just being atomic?

libs/core/concurrency/include/hpx/concurrency/deque.hpp
```diff
@@ -29,6 +29,7 @@
 #include <memory>
```
msimberg, line 29: Are these changes related to the thread id changes. If yes, how?

libs/core/coroutines/include/hpx/coroutines/detail/coroutine_accessor.hpp
```diff
@@ -29,12 +29,17 @@
...
+
+#include <utility>
+
 namespace hpx { namespace threads { namespace coroutines { namespace detail {
 
     struct coroutine_accessor
     {
         template <typename Coroutine>
-        static typename Coroutine::impl_ptr get_impl(Coroutine& x)
+        HPX_FORCEINLINE static typename Coroutine::impl_ptr get_impl(
```
msimberg, line 41: Not your change, but why is this static?

libs/core/schedulers/include/hpx/schedulers/thread_queue.hpp
```diff
@@ -188,7 +189,7 @@ namespace hpx { namespace threads { namespace policies {
                     p = threads::thread_data_stackful::create(
                         data, this, stacksize);
                 }
-                thrd = thread_id_type(p);
+                thrd = thread_id_type(p, false);    // do not addref
```
msimberg, line 192: Since we've talked about adding and not adding references before, how about an `enum class addref { yes, no }` or something similar in place of that bool?

libs/full/resiliency_distributed/include/hpx/resiliency_distributed/async_replay_distributed.hpp
```diff
@@ -66,8 +66,7 @@ namespace hpx { namespace resiliency { namespace experimental {
                 // necessary
                 auto this_ = this->shared_from_this();
                 return f.then(hpx::launch::sync,
-                    [this_ = std::move(this_), ids, iteration](
-                        hpx::future<Result>&& f) {
+                    [this_, ids, iteration](hpx::future<Result>&& f) {
```
msimberg, line 69: What's the reason for copying `this_` here rather than moving it? Or just capturing it directly as `[this_ = this->shared_from_this(), ...]`?

libs/core/coroutines/include/hpx/coroutines/thread_id_type.hpp
```diff
@@ -56,60 +74,277 @@ namespace hpx { namespace threads {
...
+            return lhs.thrd_ == rhs.thrd_;
+        }
+
+        friend bool operator!=(
+            thread_id const& lhs, thread_id const& rhs) noexcept
+        {
+            return lhs.thrd_ != rhs.thrd_;
+        }
+
+        friend bool operator<(
```
msimberg, line 327: I've nothing against this, but is this useful?

msimberg (commented):
Thanks @hkaiser, this looks good to me. Just some minor comments, mainly out of curiousity.

examples/cancelable_action/cancelable_action/server/cancelable_action.hpp
hkaiser: I might be able to roll back this change now that `thread::id` is not reference counting anymore. The reference counted `thread::id` can't be an atomic.

libs/core/concurrency/include/hpx/concurrency/deque.hpp
hkaiser: No, it was a missing #include.

libs/core/coroutines/include/hpx/coroutines/detail/coroutine_accessor.hpp
hkaiser: Because it does not need to access any members of that struct.

libs/core/schedulers/include/hpx/schedulers/thread_queue.hpp
hkaiser: Good thinking!

libs/core/coroutines/include/hpx/coroutines/thread_id_type.hpp
hkaiser: It's needed if somebody wants to hold this type in a `std::set` or `std::map`, etc.

libs/full/resiliency_distributed/include/hpx/resiliency_distributed/async_replay_distributed.hpp
hkaiser: Good catch, that's a left-over from previous changes. I will roll this back.

libs/core/coroutines/include/hpx/coroutines/detail/coroutine_accessor.hpp
msimberg: D'oh. I missed that it was inside a struct... Makes absolute sense.

examples/cancelable_action/cancelable_action/server/cancelable_action.hpp
msimberg: Was it not possible in the end or you just didn't get to it yet (or forget to push ;))?

hkaiser: It wasn't possible as even the non-ref-counted `thread;;id` ended up being not trivially copyable. I have not investigated whether this can be achieved yet, however.

msimberg: Ok, thanks. What makes the new non-ref-counted id different from the old non-ref-counted id on master? I imagined they'd be practically the same.

hkaiser: As said, I have not investigated yet. It could be the explicit move constructor/move assignment I added. We might be able to remove those again.

hkaiser:
@msimberg this is finally ready for review. All what's left is to rebase it one more time. Please, if you have the time, have a closer look again.

libs/core/schedulers/include/hpx/schedulers/thread_queue.hpp
```diff
@@ -249,21 +255,14 @@ namespace hpx { namespace threads { namespace policies {
...
-                if (schedule_now)
-                {
-                    // pushing the new thread into the pending queue of the
-                    // specified thread_queue
-                    ++added;
-                    schedule_thread(get_thread_id_data(thrd));
-                }
+                // insert the thread into the work-items queue assuming it is
+                // in pending state, thread would go out of scope otherwise
+                HPX_ASSERT(schedule_now);
```
msimberg, line 260: This assert seems odd to me, but I'm probably missing the bigger context. Why can we assume that `schedule_now` is always true here? The comment above makes it sound like it can still be false sometimes, but the assert doesn't match that.

libs/full/runtime_local/src/runtime_local.cpp
```diff
@@ -329,39 +319,55 @@ namespace hpx {
...
+
+        thread_manager_.reset(new hpx::threads::threadmanager(rtcfg_,
+#ifdef HPX_HAVE_TIMER_POOL
+            timer_pool_,
+#endif
+            notifier_, network_background_callback));
+
+        // initialize our TLS
         runtime::init_tss();
         util::reinit_construct();    // call only after TLS was initialized
```
msimberg, line 370: `init_tss` and `reinit_construct` don't seem to belong in `set_notification_policies`? Can they be moved to after the call to `set_notification_policies`?

msimberg:
Sorry @hkaiser for taking so long with this. Two final comments, possibly non-issues.

> This PR now also has changes that are unrelated to the reference counting of threads but that have been necessary because of ASAN problems that started showing up. These changes are more pervasive than I was hoping for and are related to the initialization of the runtime, AGAS, parcelport, thread-manager, and io_pool classes. These changes are all in one commit ([c2965e6](https://github.com/STEllAR-GROUP/hpx/commit/c2965e6bfee08b87582f46c6b0521b15f1cd6738)). If needed, I could try separating those into their own PR that would have to be merged first. @msimberg please let me know how you would like to handle this.

As mentioned in last week's call, now that I understand why the "unrelated" changes are here, let's keep them in this PR.

libs/core/schedulers/include/hpx/schedulers/thread_queue.hpp
hkaiser: In general, the scheduling system now keeps threads alive if they are `pending` (sitting in a queue), `active`/running (the scheduling loop holds them), are `suspended` (there is a reference count sitting on the thread's stack created before the actual suspension), or explicitly by the user. 

Therefore, staged tasks must have their initial state set to `pending`. They are converted to real threads without user intervention by the scheduler, thus the only way to keep them alive (keep the reference count up) is to schedule them (add them to a queue) once they have been created.

This assert should be always true because of this: https://github.com/STEllAR-GROUP/hpx/pull/5441/files/c2965e6bfee08b87582f46c6b0521b15f1cd6738#diff-34057d72d801cb0ee81de2f396edd2f1de7bd0d6309445a06df1d1e58630de8eR733. I should however turn the second assert into an exception to enforce the first.

libs/full/runtime_local/src/runtime_local.cpp
hkaiser: Yes, definitely.

libs/core/schedulers/include/hpx/schedulers/thread_queue.hpp
msimberg: Do we then even need the `staged` state anymore?

hkaiser: The staged state represents a thread to be delay-created only once the scheduler runs out of work,  so this question is an orthogonal one. The new limitation really just requires for staged threads to be scheduled right away once they are created eventually. I think this is what I would expect anyways. What use is a staged task if it should not be run once turned into a 'real' thread?

msimberg: Fully agreed. The `run_now` flag represents whether or not a thread should start as `staged` or be scheduled immediately, not the `staged` state. As far as I can tell, and this seems to have been true before this PR as well, the only places we use `thread_schedule_state::staged` is to query the sizes of the staged tasks queues. Neither `thread_init_data` nor `thread_data` should ever be in the `staged` state (there's an `initial_state_` in `thread_init_data` but that's for when the thread actually is scheduled).

With that in mind should we also add an assertion that `initial_state_` is never `staged` to the `thread_init_data` constructors: https://github.com/STEllAR-GROUP/hpx/blob/fa6e4df704bf5f81ddc0f4e09716bd11ff8724c8/libs/core/threading_base/include/hpx/threading_base/thread_init_data.hpp#L29-L133? In addition to the assertions that you've already added and proposed, that is.

hkaiser:
> @hkaiser this is keeping the `thread_pool_scheduler` test from exiting:
> 
> https://github.com/STEllAR-GROUP/hpx/blob/877c2be8229a44810d0017ddd24ab8303b3570d2/libs/parallelism/executors/tests/unit/thread_pool_scheduler.cpp#L123
> 
> . With the changes in this PR, is the recommendation to simply explicitly reset ids like that or is there another way? I see a `thread_id_noref_type`, but that looks like an internal type. To be clear, having a global `hpx::thread::id` like that is very test-specific and I'd hope noone does that in real code so I'd be fine with just explicitly requiring that it be reset if the plan is to keep an id like that alive for a longer time. The documentation for `hpx::this_thread::get_id()` and/or `hpx::thread::id` could probably do with a note about reference counting.

`hpx::thread::id` is non-reference counting now. I think that there are no changes for the user.

libs/core/schedulers/include/hpx/schedulers/thread_queue.hpp
hkaiser: I agree that the initial state should not be `staged`, ever. It's an internal state users shouldn't have to worry about. I will add an error check for this.

aurianer (approved):
Thanks!

hkaiser:
> Uh, sorry @hkaiser. I merged #5377 which conflicts with this. It may be easiest to ignore the changes to the `set_thread_state` files that I did in #5377, and I can try to redo them on top of your changes. Basically they boiled down to separating the timed and non-timed functionality to separate headers to avoid pulling in Asio in as many places. In any case, if you need a hand with this let me know.

@msimberg I rebased and tried to re-apply your changes. Could you please verify I didn't mess anything up?

msimberg:
> @msimberg I rebased and tried to re-apply your changes. Could you please verify I didn't mess anything up?

The changes look very sane, thank you! The move of the implementations to cpp files is also two steps better than the changes I had in #5377. :+1:

msimberg (dismissed):
Thanks @hkaiser. Let's go ahead with this if it comes back clean (we might have to wait for at least the Piz Daint builders to be back in action for that though...).

msimberg:
retest cscs

hkaiser:
This should be good to go once the tests have cycled.

hkaiser:
During the PMC meeting today we agreed that this is good to go.

PR #5445, opened 2021-07-13
Improve performance of Spinlocks

NK-Nikunj:
The current implementation of spinlocks uses atomic exchanges on every loop iteration that can cause excess contention due to cache coherence policies within the CPU. This will lead to overwhelming traffic on the bus reducing the performance of the locks. The backoff through yield does reduce contention but we can still do better in performance.

I have mitigated this by first checking if the lock can be acquired by loading the value to cache and then spinning on it every iteration. This reduces the contention from every iteration to iterations where the lock has been released by another thread. The improvements will be insignificant for low contention situations (probably an improvement of a couple of nanoseconds) but should be observable for very high contention situations.

### DIFF
--- a/libs/core/synchronization/include/hpx/synchronization/spinlock.hpp
+++ b/libs/core/synchronization/include/hpx/synchronization/spinlock.hpp
@@ -49,7 +49,30 @@ namespace hpx { namespace lcos { namespace local {
         {
             HPX_ITT_SYNC_PREPARE(this);
 
-            while (!acquire_lock())
+            // Checking for the value in is_locked() ensures that
+            // acquire_lock is only called when is_locked computes
+            // to false. This way we spin only on a load operation
+            // which minimizes false sharing that comes with an
+            // exchange operation.
+            // Consider the following cases:
+            // 1. Only one thread wants access critical section:
+            //      is_locked() -> false; computes acquire_lock()
+            //      acquire_lock() -> false (new value set to true)
+            //      Thread acquires the lock and moves to critical
+            //      section.
+            // 2. Two threads simultaneously access critical section:
+            //      Thread 1: is_locked() || acquire_lock() -> false
+            //      Thread 1 acquires the lock and moves to critical
+            //      section.
+            //      Thread 2: is_locked() -> true; execution enters
+            //      inside while without computing acquire_lock().
+            //      Thread 2 yields while is_locked() computes to
+            //      false. Then it retries doing is_locked() -> false
+            //      followed by an acquire_lock() operation.
+            //      The above order can be changed arbitrarily but
+            //      the nature of execution will still remain the
+            //      same.
+            while (is_locked() || !acquire_lock())
             {
                 util::yield_while([this] { return is_locked(); },
                     "hpx::lcos::local::spinlock::lock");
@@ -88,18 +111,18 @@ namespace hpx { namespace lcos { namespace local {
 
     private:
         // returns whether the mutex has been acquired
-        bool acquire_lock()
+        HPX_FORCEINLINE bool acquire_lock()
         {
             return !v_.exchange(true, std::memory_order_acquire);
         }
 
         // relinquish lock
-        void relinquish_lock()
+        HPX_FORCEINLINE void relinquish_lock()
         {
             v_.store(false, std::memory_order_release);
         }
 
-        bool is_locked() const
+        HPX_FORCEINLINE bool is_locked() const
         {
             return v_.load(std::memory_order_relaxed);
         }
--- a/libs/core/synchronization/include/hpx/synchronization/spinlock.hpp
+++ b/libs/core/synchronization/include/hpx/synchronization/spinlock.hpp
@@ -72,11 +72,11 @@ namespace hpx { namespace lcos { namespace local {
             //      The above order can be changed arbitrarily but
             //      the nature of execution will still remain the
             //      same.
-            while (is_locked() || !acquire_lock())
+            do
             {
                 util::yield_while([this] { return is_locked(); },
                     "hpx::lcos::local::spinlock::lock");
-            }
+            } while (!acquire_lock());
 
             HPX_ITT_SYNC_ACQUIRED(this);
             util::register_lock(this);
--- a/libs/core/synchronization/include/hpx/synchronization/spinlock.hpp
+++ b/libs/core/synchronization/include/hpx/synchronization/spinlock.hpp
@@ -4,6 +4,7 @@
 //  Copyright (c) 2014 Thomas Heller
 //  Copyright (c) 2008 Peter Dimov
 //  Copyright (c) 2018 Patrick Diehl
+//  Copyright (c) 2021 Nikunj Gupta
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
--- a/libs/core/thread_support/include/hpx/thread_support/spinlock.hpp
+++ b/libs/core/thread_support/include/hpx/thread_support/spinlock.hpp
@@ -45,17 +45,17 @@ namespace hpx { namespace util { namespace detail {
 
         void lock() noexcept
         {
+            // Optimistically assume the lock is free on the first try
+            if (try_lock())
+                return;
+
+            // Wait for lock to be released without generating cache misses
+            // Similar implementation to hpx::lcos::local::spinlock
             unsigned k = 0;
             do
             {
-                // Wait for lock to be released without generating cache misses
-                // Similar implementation to hpx::lcos::local::spinlock
-                if (try_lock())
-                    return;
-
                 yield_k(k++);
-
-            } while (true);
+            } while (!try_lock());
         }
 
         HPX_FORCEINLINE void unlock() noexcept
--- a/libs/core/thread_support/include/hpx/thread_support/spinlock.hpp
+++ b/libs/core/thread_support/include/hpx/thread_support/spinlock.hpp
@@ -1,6 +1,7 @@
 ////////////////////////////////////////////////////////////////////////////////
 //  Copyright (c) 2014 Peter Dimov
 //  Copyright (c) 2020 Agustin Berge
+//  Copyright (c) 2021 Nikunj Gupta
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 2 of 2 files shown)


libs/core/synchronization/include/hpx/synchronization/spinlock.hpp
```diff
@@ -49,7 +49,30 @@ namespace hpx { namespace lcos { namespace local {
...
+            //      Thread 2 yields while is_locked() computes to
+            //      false. Then it retries doing is_locked() -> false
+            //      followed by an acquire_lock() operation.
+            //      The above order can be changed arbitrarily but
+            //      the nature of execution will still remain the
+            //      same.
+            while (is_locked() || !acquire_lock())
             {
                 util::yield_while([this] { return is_locked(); },
                     "hpx::lcos::local::spinlock::lock");
```
msimberg, line 78: Would this be equivalent, and possibly simpler?

```suggestion
            do 
            {
                util::yield_while([this] { return is_locked(); },
                    "hpx::lcos::local::spinlock::lock");
            } while (!acquire_lock())
```
(ignore the mismatched lines, github doesn't let me comment on lines below the diff).

Can we construct a small test to verify the effect of this? If nothing else, running the `future_overhead` benchmark, which relies on (mostly uncontended) spinlocks.

Thanks for looking into this!

NK-Nikunj: Good catch! We can certainly do it this way to make it look simpler. This will also ensure that there is a single load operation when leaving the loop (instead of the current 2). I'll correct it right away.

NK-Nikunj: @msimberg Let me write a small high contention test to test the performance. I do not suspect any significant speedup but decreasing the CPU bus traffic may potentially help us with performance somewhere else.

hkaiser: This change will take effect in the uncontended case mostly. With higher contention false sharing is still an issue. But this can't be resolved that easily. Thanks for looking into this.

Would you mind fixing the `util::spinlock` in a similar way? This needs to be changed to try a load first as well: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/thread_support/include/hpx/thread_support/spinlock.hpp#L48.

NK-Nikunj: @hkaiser the changes in the commit also include changes to the link you directed to, or am I getting something wrong?

hkaiser: Perfect, I didn't see that. Thanks!

libs/core/thread_support/include/hpx/thread_support/spinlock.hpp
```diff
@@ -45,15 +45,18 @@ namespace hpx { namespace util { namespace detail {
...
             // Wait for lock to be released without generating cache misses
+            // Similar implementation to hpx::lcos::local::spinlock
             unsigned k = 0;
-            do
+            while (m.load(std::memory_order_relaxed) ||
+                m.exchange(true, std::memory_order_acquire))
             {
                 yield_k(k++);
-            } while (!try_lock());
+            }
```
hkaiser, line 59: A `do{} while()` would be more appropriate as we know that the lock can't be acquired if `k == 0`. Also, wy not call `try_lock()` instead?

NK-Nikunj: `while()` vs `do{} while()` : `yield_k` for `k == 0` is a no-op so I felt both ways of representing are equivalent. This can be changed however if you'd like it to be a `do{} while()` without any change in performance.

why not call `try_lock`? The nature of the return statement means that `!m.load(std::memory_order_relaxed)` will always compute to false in cases where the lock has been previously acquired. This means that `!try_lock()` will essentially trigger an exchange statement every iteration. This can be mitigated by doing something like:

```
while (true)
{
    if (try_lock())
        return;
    yield_k(k++);
}
```

Would you want me to change the code to look like above?

NK-Nikunj: It could also be turned into a `do{} while()` using which we can get rid of the above optimistic try_lock. I think I'll convert it to `do{} while()` for better code clarity.

libs/core/thread_support/include/hpx/thread_support/spinlock.hpp
```diff
@@ -44,19 +45,17 @@ namespace hpx { namespace util { namespace detail {
...
             {
+                // Wait for lock to be released without generating cache misses
+                // Similar implementation to hpx::lcos::local::spinlock
+                if (try_lock())
+                    return;
+
                 yield_k(k++);
-            }
+
+            } while (true);
```
hkaiser: What's wrong with simply writing:
```
do {
    yield_k(k++);
} while (!try_lock());
```

NK-Nikunj: OK. I just tried compiler explorer to check how compiler evaluates `!try_lock()` and it worked as well. I was thinking that it will lead to the evaluation of exchange on each iteration and thus avoided writing that way.

NK-Nikunj:
@hkaiser @msimberg Not the most extensive of tests, but I wrote a small [benchmark](https://gist.github.com/NK-Nikunj/7bf4641ff77b4db1ee3aaf841aa52bfb) and here are the results:

For the unimproved implementation, I got the following results:
```
Minimum Time: 0.862814
Maximum Time: 0.923062
Avg Time: 0.885118
```

For the improved implementation, I got the following results:
```
Minimum Time: 0.847032
Maximum Time: 0.89873
Avg Time: 0.87345
```

As I suspected, the improvements aren't anything big. 1.3% improvements are small enough to even hide within noise but hey... An improvement is still an improvement!

hkaiser:
@NK-Nikunj the numbers you showed are for sequential execution (i.e. low contention). Would you able to do the same for a parallel loop (i.e. high contention)?

NK-Nikunj:
@hkaiser I think I updated the code after I did parallel execution. The execution times above are for the parallel (contended) cases. The sequential results are as follows:

For the unimproved implementation:
```
Minimum Time: 10.0699
Maximum Time; 10.093
Avg Time: 10.0811
```

For the improved implementation:
```
Minimum Time: 10.0697
Maximum Time; 10.093
Avg Time: 10.079
```

The difference in execution times is barely anything over 100 iterations. I suspected a slight increase in execution times due to the additional load operation alongside exchange that we already do but seems like there isn't any difference in execution times.

msimberg:
> As I suspected, the improvements aren't anything big. 1.3% improvements are small enough to even hide within noise but hey... An improvement is still an improvement!

I feel obliged to post this since I just rewatched it yesterday: https://www.youtube.com/watch?v=koTf7u0v41o. Even with that, I really wouldn't be surprised if that's a real improvement, no matter how small.

libs/core/thread_support/include/hpx/thread_support/spinlock.hpp
```diff
@@ -45,10 +46,11 @@ namespace hpx { namespace util { namespace detail {
         void lock() noexcept
         {
             // Optimistically assume the lock is free on the first try
-            if (!m.exchange(true, std::memory_order_acquire))
+            if (try_lock())
```
msimberg, line 49: Is there now any reason why this spinlock should first do `try_lock` and the other one would first check `is_locked`? What's different?

NK-Nikunj: There's no difference. The first one does a `try_lock` in hopes of getting the lock. If the `load` operation returns false it continues executing without doing an exchange (that we are trying to reduce). The second one does exactly the same. It looks different because `yield_k` functions are different in both implementations. The first one takes a counter k and implements a backoff strategy. The second one takes a predicate and does what the first one does but only until the predicate returns true. So, the design for both of them is essentially the same. The function signature causes them to look different.

msimberg: Let me rephrase: Why not structure this exactly the same way as the spinlock in `synchronization`, with `is_locked` etc.?

Additionally, with the change to use `try_lock` in the `if` these two are identical:
```
if (try_lock) return;
unsigned k = 0;
do { yield_k(k++); } while (!try_lock());
```
and:
```
unsigned k = 0;
while (!try_lock) { yield_k(k++); }
```

NK-Nikunj: Aah, right! I got what you're saying now. Yes, it can be structured this way and I should do it do, apologies.

hkaiser (approved):
LGTM, thanks!

libs/core/thread_support/include/hpx/thread_support/spinlock.hpp
msimberg: Sorry @NK-Nikunj for beating this topic to death, but:

`synchronization/spinlock.hpp` was changed from
```
            while (!acquire_lock())
            {
                util::yield_while([this] { return is_locked(); },
                    "hpx::lcos::local::spinlock::lock");
            };
```
to
```
            do
            {
                util::yield_while([this] { return is_locked(); },
                    "hpx::lcos::local::spinlock::lock");
            } while (!acquire_lock());
```
because of `<reasons>`.

`thread_support/spinlock.hpp` is now:
```
            unsigned k = 0;
            while (!try_lock())
            {
                yield_k(k++);
            };
```

Now my original question in this thread becomes a bit more obvious: Are these two still equivalent, and if not, why should they not be?

In https://github.com/STEllAR-GROUP/hpx/pull/5445#discussion_r669503863 I merely wanted to point out that if there is going to be an `if (x) return; do { z; } while (!x);` it's better written as `while (!x) { z; };` not necessarily suggest that it should be a `while (!x) { z; };`

NK-Nikunj: > Are these two still equivalent, and if not, why should they not be?

Yes. I think the confusion arises from the behavior of yield functions. Let's consider both cases one by one. For `thread_support/spinlock.hpp`, we do:

```
unsigned k = 0;
while (!try_lock())
{
    yield_k(k++);
};
```

This does one atomic load which when computed to true does a yield of 0 following which it does another atomic load. If that computes to true as well then we do yield of 1 and so on. In essence, we are doing atomic loads as long as yield either suspends the thread or the atomic loads computes to false. Finally, if it computes to false, it does an atomic exchange to try and acquire the lock.

Now consider `synchronization/spinlock.hpp` file with `do {} while` implementation. Here, it goes into yield first (the internal yield variable k is still 0) and tries for an atomic load (defined by `is_locked()`). It continues to yield with k = 1, 2, ... until `is_locked` returns false following which it does an atomic exchange to acquire the lock (defined by `!acquire_lock()`). So, it has the same functioning as the above case.

Now consider the case with `while` instead of `do {} while`. Here we do an extra atomic load in the while loop. So, for k = 0, we do one atomic load when we enter the loop followed by another atomic load through `is_locked()`. Now, this wouldn't matter for scenarios where the first thread acquires the lock (as we never get into the while loop) but when the thread has to spin on while, we end up doing an extra load.

Again, the cost of a single atomic load might not be substantial, but I won't call them equivalent implementations then :P

hkaiser: We should avoid unneeded atomic operations in any case. They might create non-measurable overheads, but they are still atomics (even if if they use relaxed semantics), i.e. they force the core's execution pipeline to flush/stall.

hkaiser:
retest lsu

hkaiser:
The remaining issues are unrelated.

libs/core/thread_support/include/hpx/thread_support/spinlock.hpp
msimberg: For posterity, we decided to attempt the unification of the two spinlock implementations separately. This is resolved for now.

PR #5449, opened 2021-07-15
Adapt swap_ranges to C++ 20

Jedi18:
Adapted swap_ranges to C++ 20 and added container overloads. (range and sentinel). Added container tests.

Issue #4822
Issue #3646
Issue #1668

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/algorithms/swap_ranges.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/algorithms/swap_ranges.hpp
@@ -13,6 +13,7 @@
 
 #include <hpx/executors/execution_policy.hpp>
 #include <hpx/parallel/algorithms/detail/dispatch.hpp>
+#include <hpx/parallel/algorithms/detail/distance.hpp>
 #include <hpx/parallel/algorithms/for_each.hpp>
 #include <hpx/parallel/util/detail/algorithm_result.hpp>
 #include <hpx/parallel/util/projection_identity.hpp>
@@ -37,18 +38,24 @@ namespace hpx { namespace parallel { inline namespace v1 {
             {
             }
 
-            template <typename ExPolicy, typename FwdIter1>
+            template <typename ExPolicy, typename FwdIter1, typename Sent>
             static FwdIter2 sequential(
-                ExPolicy, FwdIter1 first1, FwdIter1 last1, FwdIter2 first2)
+                ExPolicy, FwdIter1 first1, Sent last1, FwdIter2 first2)
             {
-                return std::swap_ranges(first1, last1, first2);
+                while (first1 != last1)
+                {
+                    std::swap(*first1, *first2);
+                    first1++;
+                    first2++;
+                }
+                return first2;
             }
 
-            template <typename ExPolicy, typename FwdIter1>
+            template <typename ExPolicy, typename FwdIter1, typename Sent>
             static typename util::detail::algorithm_result<ExPolicy,
                 FwdIter2>::type
-            parallel(ExPolicy&& policy, FwdIter1 first1, FwdIter1 last1,
-                FwdIter2 first2)
+            parallel(
+                ExPolicy&& policy, FwdIter1 first1, Sent last1, FwdIter2 first2)
             {
                 typedef hpx::util::zip_iterator<FwdIter1, FwdIter2>
                     zip_iterator;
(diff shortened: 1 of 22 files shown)


libs/parallelism/algorithms/include/hpx/parallel/algorithms/swap_ranges.hpp
```diff
@@ -117,17 +94,199 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+            }
+
+            template <typename ExPolicy, typename FwdIter1, typename Sent1,
+                typename FwdIter2, typename Sent2>
+            static swap_ranges_result<FwdIter1, FwdIter2> sequential(ExPolicy,
+                FwdIter1 first1, Sent1 last1, FwdIter2 first2, Sent2 last2)
+            {
+                while (first1 != last1 && first2 != last2)
+                {
+                    std::swap(*first1, *first2);
```
hkaiser, line 182: Shouldn't this use `std::iter_swap` instead?

Jedi18: Yep, I've used the implementation of iter_swap, probably better to use iter_swap itself. Ok I'll make the change, thanks!

hkaiser: To be even more precise, we actually should use `std::ranges::iter_swap`, but that's available with C++20 only. The benefit would be that it properly handles iterators that return proxy types (`std::iter_swap` does not). I have created https://github.com/STEllAR-GROUP/hpx/issues/5456 for us to remember touching all algorithms, if needed.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/swap_ranges.hpp
```diff
@@ -117,17 +94,201 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+          : public detail::algorithm<swap_ranges<IterPair>, IterPair>
+        {
+            swap_ranges()
+              : swap_ranges::algorithm("swap_ranges")
+            {
+            }
+
+            template <typename ExPolicy, typename FwdIter1, typename Sent,
+                typename FwdIter2>
+            static FwdIter2 sequential(
```
hkaiser, line 163: This can be `constexpr`, I believe.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/swap_ranges.hpp
hkaiser: This should still use `iter_swap`.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/swap_ranges.hpp
```diff
@@ -117,17 +94,201 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+
+            return hpx::parallel::v1::detail::swap_ranges<FwdIter2>().call(
+                hpx::execution::seq, first1, last1, first2);
+        }
+
+        // clang-format off
+        template <typename ExPolicy, typename FwdIter1, typename FwdIter2,
+            HPX_CONCEPT_REQUIRES_(
+                hpx::is_execution_policy<ExPolicy>::value &&
+                hpx::traits::is_forward_iterator<FwdIter1>::value &&
```
hkaiser, line 274: If you check for `is_forward_iterator` here, the `static_assert` below is superfluous (and btw wrong in this case anyways).

libs/parallelism/algorithms/include/hpx/parallel/container_algorithms/swap_ranges.hpp
```diff
@@ -0,0 +1,355 @@
...
+    ///           \a util::in_in_result<InIter1, InIter2>.
+    ///           The \a swap_ranges algorithm returns in_in_result with the
+    ///           first element as the iterator to the element past the last
+    ///           element exchanged in range beginning with \a first1 and the
+    ///           second element as the iterator to the element past the last
+    ///           element exchanged in the range beginning with \a first2.
+    ///
+    template <typename InIter1, typename Sent1, typename InIter2,
+        typename Sent2>
+    parallel::util::in_in_result<InIter1, InIter2>
```
hkaiser, line 54: IIRC, in other places we import `in_in_result` (and friends) into the namespace `hpx::ranges`. We might have to consolidate this, though.

hkaiser: Hmmm, we should rather explicitly reference `hpx::ranges::swap_ranges_result` here.

aurianer:
add to whitelist

hkaiser (approved):
LGTM, thanks!

hkaiser:
retest cscs

PR #5450, opened 2021-07-15
Change the print-bind output to be more precise

aurianer:
The "Node" word can be confusing

### DIFF
--- a/libs/core/topology/src/topology.cpp
+++ b/libs/core/topology/src/topology.cpp
@@ -951,7 +951,7 @@ namespace hpx { namespace threads {
                 break;
 
             case HWLOC_OBJ_NODE:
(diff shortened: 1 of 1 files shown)


libs/core/topology/src/topology.cpp
```diff
@@ -951,7 +951,7 @@ namespace hpx { namespace threads {
                 break;
 
             case HWLOC_OBJ_NODE:
-                print_info(os, obj, "Node ", comma);
+                print_info(os, obj, "NumaNode ", comma);
```
msimberg, line 954: Nitpick: `NUMANode`? Only because `NUMA` is an acronym and that's the capitalization hwloc uses as well. Either is fine by me though.

msimberg (approved):
Thanks!

PR #5452, opened 2021-07-15
Fix find_end parallel overload

srinivasyadav18:
## Fixes find_end parallel overload
The current version just checks the first element in the subsequence and returns the iterator if matches.
It does not check the whole subsequence.

## Proposed Changes
- Replaces typedef reference with auto.
- Removes unused `curr` Iterator variable.
- Uses `mid` and `mid2` to check the whole subsequence.


hkaiser (approved):
LGTM, thanks!

Do you have a test going with this that makes sure the change is correct?

srinivasyadav18:
Yup, will do it soon and update the PR.

hkaiser (approved):
LGTM, thanks!

PR #5460, opened 2021-07-20
Adapt sort to C++ 20

Jedi18:
Adapted sort to C++ 20 and added container overloads. (range and sentinel). Added sentinel tests.

Issue #4822
Issue #3646

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/algorithms/sort.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/algorithms/sort.hpp
@@ -2,12 +2,237 @@
 //  Copyright (c) 2015-2020 Hartmut Kaiser
 //  Copyright (c) 2015-2019 Francisco Jose Tapia
 //  Copyright (c) 2018 Taeguk Kwon
+//  Copyright (c) 2021 Akhil J Nair
 //
 //  SPDX-License-Identifier: BSL-1.0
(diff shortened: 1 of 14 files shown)


libs/parallelism/algorithms/include/hpx/parallel/algorithms/sort.hpp
```diff
@@ -376,14 +541,87 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+                hpx::traits::is_iterator<RandomIt>::value &&
+                parallel::traits::is_projected<Proj, RandomIt>::value &&
+                parallel::traits::is_indirect_callable<
+                    hpx::execution::sequenced_policy, Comp,
+                    parallel::traits::projected<Proj, RandomIt>,
+                    parallel::traits::projected<Proj, RandomIt>
+                >::value
+            )>
+        // clang-format on
+        friend RandomIt tag_fallback_dispatch(hpx::sort_t, RandomIt first,
```
hkaiser, line 585: The algorithms in `namespace std` don't return anything (http://eel.is/c++draft/sort)

Jedi18: Oh right my bad, I'll make the change. Thanks!

hkaiser:
retest cscs

PR #5467, opened 2021-07-25
Direct scoped execution

hkaiser:
This PR adds code that allows for futures to execute the associate thread directly if that thread has not started executing through 'normal' means yet (i.e. through the scheduler). Currently, this mode is disabled by default (see configuration option `HPX_COROUTINES_WITH_THREAD_SCHEDULE_HINT_RUNS_AS_CHILD_DEFAULT`, which is `OFF` by default). Direct execution can be enabled/disabled through an additional flag in the `thread_schedule_hint` structure.

In the future, this mode could be enabled by default as many futures could potentially take advantage of the reduced overheads. For the fibonacci example this reduces the overall execution time to 30%.

This PR relies on #5441.

While working on this PR I have applied a number of flyby changes that are not directly related to the PR itself:

- flyby: local execution of async<Action>() now properly sets thread description (#5551)
- flyby: fixing iostream shutdown (#5546)
- flyby: fixing implementation of agent_ref::yield_k (#5549, #5567)
- flyby: improve shutdown operation (#5546)
- flyby: making get_action_name<Action>() noexcept (#5548)
- flyby: modernize async_launch_policy_dispatch (#5552)
- flyby: fixing communication_set_test (#5547)
- flyby: fixing problem in hierarchical spawner (#5552)

~~Once everything works as expected I will separate those into their own PRs.~~ - done.


libs/parallelism/futures/src/future_data.cpp
```diff
@@ -81,7 +83,64 @@ namespace hpx { namespace lcos { namespace detail {
     }
 
     ///////////////////////////////////////////////////////////////////////////
-    future_data_base<traits::detail::future_data_void>::~future_data_base() {}
+    future_data_base<traits::detail::future_data_void>::~future_data_base()
+    {
+        if (runs_child_ != threads::invalid_thread_id)
```
msimberg, line 88: What exactly does `runs_child_` represent?

libs/parallelism/futures/src/future_data.cpp
```diff
@@ -81,7 +83,64 @@ namespace hpx { namespace lcos { namespace detail {
...
+        {
+            auto* thrd = get_thread_id_data(runs_child_);
+            (void) thrd;
```
msimberg, line 91:
```suggestion
            HPX_UNUSED(thrd);
```

libs/parallelism/futures/tests/unit/direct_scoped_execution.cpp
```diff
@@ -0,0 +1,95 @@
...
+    std::uint64_t n2 = fibonacci(n - 2);
+
+    // wait for the Futures to return their values
+    return n1.get() + n2;
+}
+
+///////////////////////////////////////////////////////////////////////////////
+int hpx_main(hpx::program_options::variables_map& vm)
+{
+    hpx::threads::add_scheduler_mode(hpx::threads::policies::fast_idle_mode);
```
msimberg, line 50: Is this necessary for the scoped execution to work? Does it have an effect on the performance?

libs/core/concurrency/include/hpx/concurrency/deque.hpp
```diff
@@ -55,15 +55,15 @@ namespace boost { namespace lockfree {
         T data;
 
         deque_node()
```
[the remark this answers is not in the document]
hkaiser: Yes, these changes are all unrelated and improve general performance. I will separate those into another PR.

libs/core/coroutines/include/hpx/coroutines/detail/coroutine_stackful_self_direct.hpp
```diff
@@ -0,0 +1,141 @@
...
+#include <hpx/coroutines/thread_enums.hpp>
+#include <hpx/coroutines/thread_id_type.hpp>
+
+#include <cstddef>
+#include <limits>
+#include <utility>
+
+namespace hpx { namespace threads { namespace coroutines { namespace detail {
+
+    class coroutine_stackful_self_direct : public coroutine_self
```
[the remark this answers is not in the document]
hkaiser: Excellent question, I'm not sure however if this would reuse any of the functionality of `coroutine_stackful_self`. I'll investigate,

libs/core/coroutines/src/detail/coroutine_impl.cpp
```diff
@@ -96,4 +97,52 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
         // should not get here, never
         HPX_ASSERT(this->m_state == super_type::ctx_running);
     }
+
+    // execute the coroutine function directly in the context of the calling
+    // thread
+    coroutine_impl::result_type coroutine_impl::invoke_directly(
+        coroutine_impl::arg_type arg)
+    {
+        using context_exit_status = super_type::context_exit_status;
```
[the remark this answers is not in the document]
hkaiser: The only HPX thread specifics would be suspension (which suspends the executing (HPX) thread), but since the execution agent handles suspension generically, we should be able to execute things directly on any thread.

I mostly kept the execution here as similar to the 'normal' thread execution for now. We might be able to remove things, however.

libs/core/schedulers/include/hpx/schedulers/lockfree_queue_backends.hpp
```diff
@@ -56,28 +56,34 @@ namespace hpx { namespace threads { namespace policies {
         {
         }
 
-        bool push(const_reference val, bool /*other_end*/ = false)
```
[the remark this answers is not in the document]
hkaiser: I think this change is unrelated. The code simply looked wrong to me. I will investigate

libs/core/threading_base/src/create_work.cpp
```diff
@@ -88,9 +88,10 @@ namespace hpx { namespace threads { namespace detail {
         if (data.priority == thread_priority::default_)
             data.priority = thread_priority::normal;
 
-        data.run_now = (thread_priority::high == data.priority ||
-            thread_priority::high_recursive == data.priority ||
-            thread_priority::boost == data.priority);
+        data.run_now =
+            (data.run_now || thread_priority::high == data.priority ||
+                thread_priority::high_recursive == data.priority ||
+                thread_priority::boost == data.priority);
```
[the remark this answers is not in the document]
hkaiser: Changing this to an `assert` would also be an option. Either way, I made this change to close an obvious logic-hole in the existing code. This is unrelated to the current PR as well, however.

libs/full/runtime_configuration/src/runtime_configuration.cpp
```diff
@@ -695,6 +695,7 @@ namespace hpx { namespace util {
         char const* argv0_, runtime_mode mode)
       : mode_(mode)
       , num_localities(0)
+      , num_os_threads(0)
```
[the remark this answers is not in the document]
hkaiser: This is an unrelated change that gave me 10% performance gain on the fibonacci benchmark (this avoids acquiring the mutex during the ini handling), As such it is unrelated to the current PR, I will separate it.

libs/full/runtime_configuration/src/runtime_configuration.cpp
```diff
@@ -1110,16 +1111,20 @@ namespace hpx { namespace util {
...
             {
-                return hpx::util::get_entry_as<std::size_t>(
-                    *sec, "os_threads", 1);
+                util::section const* sec = get_section("hpx");
+                if (nullptr != sec)
+                {
+                    num_os_threads = hpx::util::get_entry_as<std::uint32_t>(
+                        *sec, "os_threads", 1);
+                }
             }
```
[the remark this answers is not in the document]
hkaiser: I agree, and assert would be sufficient. We should do that for all similar functions in this file, however.

libs/parallelism/execution/include/hpx/execution/detail/future_exec.hpp
```diff
@@ -151,9 +151,10 @@ namespace hpx { namespace lcos { namespace detail {
     struct post_policy_spawner
     {
         template <typename F>
-        void operator()(F&& f, hpx::util::thread_description desc)
+        threads::thread_id_ref_type operator()(
```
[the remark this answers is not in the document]
hkaiser: The need to be able to keep the created thread alive (under certain circumstances). Is this a problem?

libs/parallelism/futures/src/detail/execute_thread.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+        // launch a new thread executing the given function, preferably on the
+        // same core
+        threads::thread_schedule_hint hint(
+            static_cast<std::int16_t>(get_worker_thread_num()));
+
+        // wait for the task to run
+        threads::thread_id_ref_type tid =
+            p.apply(threads::detail::get_self_or_default_pool(),
+                "hpx::threads::detail::execute_thread_new_thread", launch::fork,
```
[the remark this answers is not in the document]
hkaiser: I most likely will remove the option to execute things on a new thread if the stack is insufficiently large as we could simply let the thread be executed the normal way in this case.

libs/parallelism/futures/src/future_data.cpp
hkaiser: It refers to the thread that is supposed to deliver the value that will make the future ready. We could probably come up with a better name for this ;-) Any suggestions?

libs/parallelism/futures/tests/unit/direct_scoped_execution.cpp
hkaiser: Right, I'll remove this. I copied the code from another example...

hkaiser:
@msimberg thanks very much for reviewing this PR. As a start, I have removed all unrelated changes (and I will create a separate PR for those).

libs/parallelism/execution/include/hpx/execution/detail/post_policy_dispatch.hpp
```diff
@@ -69,19 +72,21 @@ namespace hpx { namespace parallel { namespace execution { namespace detail {
...
 
             threads::thread_init_data data(
                 threads::make_thread_function_nullary(hpx::util::deferred_call(
                     std::forward<F>(f), std::forward<Ts>(ts)...)),
                 desc, priority, hint, stacksize,
                 threads::thread_schedule_state::pending);
-            threads::register_work(data);
+
+            threads::thread_id_ref_type thrd = threads::register_work(data);
+            return hint.runs_as_child ? thrd : threads::invalid_thread_id;
```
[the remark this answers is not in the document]
hkaiser: I have reverted this particular change, post will now return `void` as before.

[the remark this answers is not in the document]
msimberg: What consequences did reverting this have? What did you have to change to make up for it?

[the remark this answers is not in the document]
hkaiser: No big consequences. I have changed the implementation of these function to support returning the thread_id: https://github.com/STEllAR-GROUP/hpx/pull/5467/files#diff-6abd893418c71fd43f8d01620d9c601f3f12f4d1e987e9a78144669e3ff5f54aR154-R183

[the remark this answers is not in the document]
msimberg: The main thing to be aware of with these changes is that threads will be created eagerly (though not necessarily the stack) for . I don't know if that's a good, neutral, or bad thing.

If I understand correctly, this would apply to `async`, `then`, and `dataflow` *without* an explicit executor, correct?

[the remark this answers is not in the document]
hkaiser: Yes, you're right on both accounts. There is also the issue that the scoped execution is currently enabled through the `bool` in the thread-hint, but we don't have any way to pass that to the non-executor APIs, but OTOH, the scoped execution should be applied to the non-executor APIs only ... More design work is needed, apparently.

msimberg:
On the topic of of returning or not returning thread ids from `create/register_thread/work`, it occurred to me yesterday that we may in fact be able to return a valid thread id for all threads with some restructuring. The important distinction between staged and pending work was whether or not a stack was allocated. With stacks lazily allocated the difference is not so clear anymore. On my system `sizeof(thread_init_data) == 112` and `sizeof(thread_data) == 168` which hardly warrants the two-stage process. Additionally, we currently do two allocations per task (ignoring the stack): one for putting a `thread_init_data`/`thread_description` in a queue, and another for `thread_data` (only one if the thread is created eagerly). A change like this would require changing `thread_data` recycling to stack recycling.

Assuming I haven't missed anything critical, I think this could simplify the scheduler quite a bit. Note that I would still keep the distinction of staged/pending work since it's useful to prioritize already started work, but they'd both be represented by `thread_data`.

I know this would be another very big change, so I'm not asking you @hkaiser to do it here (or at all), but I wanted put the idea here before thinking about trying to do this change.

hkaiser:
@msimberg interestingly enough, the changes in this PR make the cross_pool_injection test fail reliably (even on platforms that have the new features disabled). Do you know anything about this test?

libs/core/execution/include/hpx/execution/detail/async_launch_policy_dispatch.hpp
```diff
@@ -83,6 +84,17 @@ namespace hpx { namespace detail {
                 std::forward<F>(f), std::forward<Ts>(ts)...));
             if (hpx::detail::has_async_policy(policy))
             {
+                // if one of the arguments is a future we play it conservatively
+                if constexpr (hpx::traits::is_future_any_v<std::decay_t<Ts>...>)
```
msimberg, line 88: Are there situations where this makes a difference? A future in `Ts` at this point should act as any type of argument, right, i.e. have no special handling? But I guess that's not the case if we need this here...

libs/core/resource_partitioner/tests/unit/scheduler_binding_check.cpp
```diff
@@ -68,7 +68,9 @@ void threadLoop()
         auto exec = hpx::execution::parallel_executor(
             hpx::threads::thread_priority::bound,
             hpx::threads::thread_stacksize::default_,
-            hpx::threads::thread_schedule_hint(std::int16_t(i % threads)));
+            hpx::threads::thread_schedule_hint(
+                hpx::threads::thread_schedule_hint_mode::thread,
+                std::int16_t(i % threads), true));
```
msimberg, line 73: Is the runs-as-child hint true for testing or is it meant to stay?

libs/core/runtime_local/src/runtime_local.cpp
```diff
@@ -1387,12 +1387,14 @@ namespace hpx {
         threads::thread_init_data data(util::bind(&runtime::run_helper, this,
                                            func, std::ref(result_), true),
             "run_helper", threads::thread_priority::normal,
-            threads::thread_schedule_hint(0), threads::thread_stacksize::large);
+            // prevent main from being executed directly
+            threads::thread_schedule_hint(
+                threads::thread_schedule_hint_mode::thread, 0, false),
+            threads::thread_stacksize::large);
```
msimberg, line 1393: It's good to be explicit here, so I like that. However, is there any situation where this could be executed directly? This is just creating an HPX thread that will never be associated with a future, right?

libs/core/schedulers/include/hpx/schedulers/queue_holder_thread.hpp
```diff
@@ -404,7 +404,8 @@ namespace hpx { namespace threads { namespace policies {
         void create_thread(thread_init_data& data, thread_id_ref_type* tid,
             std::size_t thread_num, error_code& ec)
         {
-            if (thread_num != thread_num_)
+            if (thread_num != thread_num_ &&
+                data.initial_state == thread_schedule_state::pending)
```
msimberg, line 408: Is this related to the changes here? If yes, how?

libs/core/threading_base/include/hpx/threading_base/scheduler_base.hpp
```diff
@@ -369,6 +369,12 @@ namespace hpx { namespace threads { namespace policies {
             return work_count;
         }
 
+        // almost all schedulers support direct execution
+        virtual bool supports_direct_execution() const
+        {
+            return true;
```
msimberg, line 375: Would it be better to make this opt-in?

libs/core/executors/include/hpx/executors/detail/hierarchical_spawning.hpp
```diff
@@ -63,7 +62,8 @@ namespace hpx { namespace parallel { namespace execution { namespace detail {
         auto it = std::begin(shape);
         for (std::size_t t = 0; t < num_threads; ++t)
         {
-            std::size_t const part_end = ((t + 1) * size) / num_threads;
+            std::size_t const part_end =
+                (std::min)(((t + 1) * size) / num_threads, size);
```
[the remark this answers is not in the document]
hkaiser: Hmmm, I don't remember why I made this change :/ I'll revisit it.

libs/core/threading_base/include/hpx/threading_base/thread_data.hpp
```diff
@@ -700,6 +718,13 @@ namespace hpx { namespace threads {
     /// thread (or zero if the current thread is not a HPX thread).
     HPX_CORE_EXPORT thread_id_type get_self_id();
 
+    /// The function \a get_outer_self_id returns the HPX thread id of
+    /// the current outer thread (or zero if the current thread is not a HPX
+    /// thread). This usually returns the same as \a get_self_id, except for
+    /// directly executed threads, in which case this returns the thread id
+    /// of the outermost HPX thread.
+    HPX_CORE_EXPORT thread_id_type get_outer_self_id();
```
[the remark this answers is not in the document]
hkaiser: `get_self_id` and `get_outer_self_id` will return the same only if no direct execution is happening. Any thread that is directly executed returns its own id with `get_self_id` and from get_out_self_id` it returns the id of the outermost executing thread (the thread that actually is rescheduled when the whole thing is suspended).

examples/quickstart/1d_wave_equation.cpp
```diff
@@ -182,8 +182,9 @@ double wave(std::uint64_t t, std::uint64_t x)
     }
     else
     {
-        std::lock_guard<hpx::lcos::local::mutex> l(u[t][x].mtx);
         double u_tminus_x = async<wave_action>(here, t - 2, x).get();
+
+        std::lock_guard<hpx::lcos::local::mutex> l(u[t][x].mtx);
```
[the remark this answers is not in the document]
hkaiser: Those are problems revealed by the PR.

libs/core/async_combinators/tests/unit/when_all.cpp
```diff
@@ -264,19 +264,19 @@ using hpx::lcos::future;
...
+        test_when_all_from_list_iterators<std::vector<future<int>>>();
+        test_when_all_from_list_iterators<std::list<future<int>>>();
+        test_when_all_from_list_iterators<std::deque<future<int>>>();
+        test_when_all_one_future();
+        test_when_all_two_futures();
+        test_when_all_three_futures();
+        test_when_all_four_futures();
+        test_when_all_five_futures();
+        test_when_all_late_futures();
+        test_when_all_deferred_futures();
```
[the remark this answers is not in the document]
hkaiser: Yes indeed. This is an unrelated change to the PR and I can separate it, if needed.

libs/core/execution/include/hpx/execution/detail/async_launch_policy_dispatch.hpp
hkaiser: This is a tricky one. Generally, futures that are used as arguments to other threads can lead to deadlocks if the thread that calls async is the one that will make the future ready (assuming the new thread happens to be executed directly). For this reason I made the decision to explicitly inhibit direct execution for threads that take futures as their arguments.

libs/core/resource_partitioner/tests/unit/scheduler_binding_check.cpp
hkaiser: I have disabled direct execution for the shared_priority_scheduler, this is a test that things still work even if a thread has a corresponding hint.

libs/core/runtime_local/src/runtime_local.cpp
hkaiser: I made this explicit as things were failing when running on the shared_priority_scheduler otherwise.

libs/core/threading_base/include/hpx/threading_base/scheduler_base.hpp
hkaiser: I didn't think of this option. This is mostly a mechanism to disable direct execution on the shared_priority_scheduler.

hkaiser:
@msimberg many thanks for closely looking at this PR.

libs/core/threading_base/include/hpx/threading_base/thread_data.hpp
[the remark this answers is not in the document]
msimberg: All right, that makes more sense now. Thanks!

libs/core/async_combinators/tests/unit/when_all.cpp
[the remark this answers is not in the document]
msimberg: Unless this turns out to drag out much longer, let's keep this here.

libs/core/execution/include/hpx/execution/detail/async_launch_policy_dispatch.hpp
msimberg: Hmm, do you have a small example to demonstrate this? Not necessarily compiling code, just the gist of it.

libs/core/resource_partitioner/tests/unit/scheduler_binding_check.cpp
msimberg: The hint should essentially just be ignored by the shared priority scheduler, correct?

libs/core/runtime_local/src/runtime_local.cpp
msimberg: That sounds worrying. I hope that was before you made the shared priority scheduler not support direct execution? But even then, if this made things hang the hint itself must've affected something that it shouldn't have affected, since the direct execution should only come in effect with futures, right?

libs/core/threading_base/include/hpx/threading_base/scheduler_base.hpp
msimberg: I think it's the only correct default since a scheduler has to explicitly support direct execution. Or is there some other reason for the shared priority scheduler not working with direct execution?

libs/core/execution/include/hpx/execution/detail/async_launch_policy_dispatch.hpp
hkaiser:
```
promise<void> p;
auto f = async([](future<void> f2){ f2.get(); }, p.get_future());
f.then(...);    // <-- might cause lambda to be executed directly
p.set_value();
``` 
If the lambda is executed directly, `f2.get()` will block indefinitely as the promise `p` will never be triggered.

libs/core/resource_partitioner/tests/unit/scheduler_binding_check.cpp
hkaiser: yes, indeed.

libs/core/runtime_local/src/runtime_local.cpp
hkaiser: I'll revisit this again. This change might not be needed anymore.

libs/core/threading_base/include/hpx/threading_base/scheduler_base.hpp
hkaiser: The shared priority scheduler relies on exactly placing threads onto particular queues (i.e. cores). Direct execution might interfere, so I decided not to support it on that scheduler.

msimberg: Relies? Or supports? The default scheduler also allows placing threads on particular queues (worker threads), so that alone can't be the reason for not supporting direct execution.

libs/core/execution/include/hpx/execution/detail/async_launch_policy_dispatch.hpp
hkaiser: > I don't have ideas to avoid this though...

The only thing we can do is to be aware of this. A sudden deadlock is a plain giveaway for this kind of situation. It might be difficult to identify it in real code, however.

msimberg: > A sudden deadlock is a plain giveaway for this kind of situation.

It's plain giveaway for something being broken somewhere... But on the whole I agree with your comment.

libs/core/threading_base/include/hpx/threading_base/scheduler_base.hpp
hkaiser: Well, the whole point of this scheduler is to try as hard as possible to run threads on the core as specified, would you agree? Other schedulers are quite loose about the hint.

msimberg: I'm not sure it's useful to start quantifying how hard a scheduler tries to run a thread on a specified core. The default scheduler tries pretty hard as well. It schedules things on a core given by the hint if there is a hint. It just doesn't know what to do with numa hints.

It sounds like `runs_as_child = true` should only be allowed if the hint mode is `thread_schedule_hint_mode::none`, independently of the scheduler. Do you think that sounds reasonable?

hkaiser: Frankly, there was more to it than just the thread placement, I just can't recall what it was. Let me get back to it and figure out why I thought that the only way was to disable direct execution.

libs/core/executors/include/hpx/executors/detail/hierarchical_spawning.hpp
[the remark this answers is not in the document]
hkaiser: You're right, I'll revert this change

libs/core/schedulers/include/hpx/schedulers/queue_holder_thread.hpp
hkaiser: If the thread is not to be run right away (`thread_num != thread_num_`) we should delay scheduling only if the initial state is pending. This is a follow-on change required by adding thread reference counting (and was missing from #5441). Alternatively I could have thrown an exception here, but I was not sure if this wouldn't break existing code relying on the `shared_priority_scheduler`. Possibly not delaying thread creation in this case (if `initial_state != thread_schedule_state::pending`) seemed to be an acceptable compromise.

libs/core/runtime_local/src/runtime_local.cpp
hkaiser: Looks like this change is not needed anymore (because of: https://github.com/STEllAR-GROUP/hpx/pull/5467/files#diff-31312dbe78048647c045b2467636f2995665d13ba2ec641c0554ca1ede9bddb0R613-R618). I'll revert it.

libs/core/schedulers/include/hpx/schedulers/queue_holder_thread.hpp
msimberg: I think I don't see the connection to reference counting. If this needs to change then so does https://github.com/STEllAR-GROUP/hpx/blob/eb83655fdf307d4f172082092adaa0bf7f513e93/libs/core/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp#L417-L449. What about `pending_boost`?

hkaiser: > I think I don't see the connection to reference counting. If this needs to change then so does

When introducing reference counting we realized that any thread created with `run_now == true` have to be created with `pending` as their initial state as otherwise there wouldn't be any way to keep the newly (delay-)created thread alive. With that PR we introduced an exception that would throw if `!run_now && state != pending`. Without the discussed change we would see that exception. As said, I didn't want to risk possibly breaking existing code at your end, even more as I don't know anything about the `shared_priority_queue_scheduler`.

> https://github.com/STEllAR-GROUP/hpx/blob/eb83655fdf307d4f172082092adaa0bf7f513e93/libs/core/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp#L417-L449

Excellent catch. The same change needs to be applied here.

> . What about `pending_boost`?

Another excellent catch. This needs to be applied retroactively to the reference counting PR as well.

gonidelis:
@hkaiser What is the status of this work? Should we rebase and try to wrap it up?

hkaiser:
> @hkaiser What is the status of this work? Should we rebase and try to wrap it up?

Let's close this without merging. I'm stuck here and was not able to make work the way I wanted.

Issue #5472, opened 2021-07-28
Compilation error with cuda/11.3 


aurianer:
This error appeared on master with the `lsu/env-gcc-9-cuda-11.sh` configuration, it takes `cuda/11` and cuda/11.3 must have been set recently as the default
https://cdash.cscs.ch/viewBuildError.php?buildid=172746

PR #5473, opened 2021-07-28
Update cuda arch and change cuda version

aurianer:
<s>- Downgrade the cuda version on rostam to avoid the errors currently on master
https://cdash.cscs.ch/viewBuildError.php?buildid=172746
- Update at the same time the cuda architecture as the sc15 nodes seems to have V100 now

Created this ticket #5472 to fix the problem with cuda/11.3</s>

EDIT:
- Workaround another bug in using noexcept with variadic templates in CUDA 11.4, @msimberg confirmed that it compiles with CUDA 11.3
- Update at the same time the cuda architecture as the sc15 nodes seems to have V100 now
- Pin CUDA version to 11.4 to avoid breaking master after a rostam default cuda package update
- Workaround bug in mismatch between pointers type which appeared with CUDA 11.3 and CUDA 11.4 but not in older versions


aurianer:
launch perftests

aurianer:
Looks like it only solves the `gcc-9-cuda-11-release` build

hkaiser (dismissed):
LGTM, thanks!

aurianer:
retest lsu

.jenkins/lsu/slurm-configuration-gcc-9-cuda-11.sh
```diff
@@ -4,5 +4,5 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-configuration_slurm_partition="sc15"
+configuration_slurm_nodes="geev"
```
msimberg, line 7: Do we need this *if* we can compile for multiple architectures at the same time? It'd be nice to keep two nodes for CUDA builds if possible.

aurianer: not at all that's why I called the commit [Tmp] I was just trying it out as I couldn't get a manual allocation on rostam

msimberg:
retest lsu

msimberg:
retest lsu

msimberg:
Thanks @aurianer for updating this! It looks good except that there were odd failures on the release build with nvcc (executables not found even though they were built). My guess is that it was unfortunate timing with the maintenance on rostam, but let's see. I've retriggered the builds on rostam.

hkaiser (approved):
LGTM, thanks!

aurianer:
Still have to investigate a compilation error which triggers the CI error (executables do not exist) so not ready to merge

aurianer:
So I downgraded temporarily the CUDA version to 11.2 and will investigate the error which appeared for CUDA 11.3 and CUDA 11.4 and fix it in another PR

msimberg:
retest lsu

libs/full/compute_cuda/include/hpx/compute/cuda/detail/launch.hpp
```diff
@@ -109,6 +107,7 @@ namespace hpx { namespace cuda { namespace experimental { namespace detail {
                     std::string("kernel launch failed: ") +
                         cudaGetErrorString(error));
             }
+            HPX_UNUSED(launcher);
```
msimberg, line 110: What's the reasoning to duplicate this in both branches?

aurianer: Cause in one of them it's used, I was finding weird that we declared it unused and then used it ^^ but I don't mind reverting the change either

msimberg: Ah, makes sense. I missed that there are three branches.

msimberg (approved):
This looks good to me, and works nicely on rostam!

I'd just like to wait for the CUDA configuration on CSCS to verify that it doesn't break with older CUDA versions. Piz Daint maintenance is almost over and should be available again later today, but it may first require changes to the CUDA configuration on Piz Daint.

msimberg:
retest cscs

PR #5477, opened 2021-07-29
Require cxx17 minor comments

### LINKED ISSUE #5463: Address minor comments of the C++17 PR bump
- [x] Signaled that HPX_WITH_CXX17 is deprecated in the docs and explain the current policy in place to find the standard.
- [x] Recheck that the c++17 features check removed [are supported](https://en.cppreference.com/w/cpp/compiler_support/17)
- [x] Includes the .cpp feature tests for the .cu ones
- [x] Hide the device code difference in the `HPX_INLINE_CONSTEXPR_VARIABLE`
- [x] Make the operator () const in the algorithm_test_utils

aurianer:
Fixes #5463

### DIFF
--- a/libs/core/config/include/hpx/config/constexpr.hpp
+++ b/libs/core/config/include/hpx/config/constexpr.hpp
@@ -9,6 +9,7 @@
 
 #pragma once
 
+#include <hpx/config/compiler_specific.hpp>
 #include <hpx/config/defines.hpp>
 
 /// This macro evaluates to ``constexpr`` if the compiler supports it.
@@ -24,12 +25,15 @@
 /// keyword. Prefer using ``constexpr`` directly instead.
 #define HPX_CONSTEXPR_OR_CONST constexpr
 
-/// This macro evaluates to ``inline constexpr`` if the compiler supports it,
-/// ``constexpr`` otherwise.
-#ifdef HPX_HAVE_CXX17_INLINE_CONSTEXPR_VARIABLE
+/// This macro evaluates to ``inline constexpr``
 #define HPX_INLINE_CONSTEXPR_VARIABLE inline constexpr
+
+/// This macro evaluates to ``inline constexpr`` for host code and
+// ``device static const`` for device code
+#if defined(HPX_COMPUTE_DEVICE_CODE)
+#define HPX_HOST_DEVICE_INLINE_CONSTEXPR_VARIABLE HPX_DEVICE static const
 #else
-#define HPX_INLINE_CONSTEXPR_VARIABLE constexpr
+#define HPX_HOST_DEVICE_INLINE_CONSTEXPR_VARIABLE HPX_INLINE_CONSTEXPR_VARIABLE
 #endif
 
 /// This macro evaluates to ``static constexpr`` if the compiler supports it,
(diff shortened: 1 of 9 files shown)


libs/core/config/include/hpx/config/constexpr.hpp
```diff
@@ -26,10 +27,10 @@
 
 /// This macro evaluates to ``inline constexpr`` if the compiler supports it,
 /// ``constexpr`` otherwise.
-#ifdef HPX_HAVE_CXX17_INLINE_CONSTEXPR_VARIABLE
-#define HPX_INLINE_CONSTEXPR_VARIABLE inline constexpr
+#if defined(HPX_COMPUTE_DEVICE_CODE)
+#define HPX_INLINE_CONSTEXPR_VARIABLE HPX_DEVICE static const
 #else
-#define HPX_INLINE_CONSTEXPR_VARIABLE constexpr
+#define HPX_INLINE_CONSTEXPR_VARIABLE inline constexpr
```
msimberg, line 33: What I'm worried about with this change is that it makes *everything* that is an `HPX_INLINE_CONSTEXPR_VARIABLE` also a device variable, which we don't necessarily want. Perhaps we want a separate `HPX_HOST_DEVICE_INLINE_CONSTEXPR_VARIABLE` as an alternative to `HPX_HOST_DEVICE HPX_INLINE_CONSTEXPR_VARIABLE` so that one can opt-in to this where needed?

msimberg (dismissed):
Looks good, thanks!

aurianer:
I just rebased, this is ready to go once CI has run

aurianer:
I just rebased after the clang-format fix has been merged but still good to go afterwards

msimberg:
This looks good, but it looks like jenkins isn't running on neither rostam nor daint (jenkins is not running currently on rostam, the daint jenkins has run out of disk quota), so we'll have to wait a bit.

msimberg:
retest

libs/core/config/include/hpx/config/constexpr.hpp
```diff
@@ -24,12 +25,15 @@
...
+/// This macro evaluates to ``inline constexpr``
 #define HPX_INLINE_CONSTEXPR_VARIABLE inline constexpr
+
+/// This macro evaluates to ``inline constexpr`` for host code and
+// ``device static const`` for device code
+#if defined(HPX_COMPUTE_DEVICE_CODE)
+#define HPX_HOST_DEVICE_INLINE_CONSTEXPR_VARIABLE HPX_DEVICE static const
 #else
-#define HPX_INLINE_CONSTEXPR_VARIABLE constexpr
+#define HPX_HOST_DEVICE_INLINE_CONSTEXPR_VARIABLE HPX_INLINE_CONSTEXPR_VARIABLE
```
hkaiser, line 36: I don't want to be picky here (so please ignore me if you feel I am), however I'd like to suggest to name the new variable `HPX_HOST_INLINE_CONSTEXPR_VARIABLE` as it represents an inline constexpr variable on the host only.

msimberg: `HPX_INLINE_CONSTEXPR_VARIABLE` is already `HPX_HOST_INLINE_CONSTEXPR_VARIABLE`. The point here is that it's a host and device variable so `DEVICE` really has to stay in the name. On the device it would be `HPX_DEVICE_STATIC_CONST_VARIABLE` but that's wrong on the host. So `HPX_HOST_DEVICE_VARIABLE`? That's pretty vague. `HPX_HOST_DEVICE_CONST_VARIABLE`? Not sure... Some day `inline constexpr` might actually be allowed on the device, which is why I'm not too bothered by the name.

hkaiser: Fine by me, just as I said, please ignore my pickiness.

PR #5479, opened 2021-07-29
Fix version check for CUDA noexcept/result_of bug

msimberg:
The version check was not accurate. It turns out that the behaviour is also dependent on the GCC version, and the new check is what I *think* is closer to the truth. I have not exhaustively tested the combinations though. This seems to hold for the versions on rostam and on compiler explorer. What I had missed earlier is that the different nvcc versions on compiler explorer of course implicitly change the GCC version as well, leading me to incorrectly think that the bug was dependent on the CUDA version (and I seem to have had the version wrong even if the GCC version would not have changed implicitly).


aurianer (dismissed):
LGTM thanks!

PR #5482, opened 2021-07-29
Change extra archive data to rely on uniqueness of a variable address, not a function address

### LINKED ISSUE #5478: hpx.dll built with vcpkg got functions pointing to the same location
The problem arises when experimenting with `1d_stencil_6.cpp`, running with only one locality is fine, but with more than one locality (thus needs serialization) AND in release mode (some optimization happens), the program throws at location
https://github.com/STEllAR-GROUP/hpx/blob/76a2ba5b266008be5e0d5a13e17ae0950099dde2/libs/full/naming/src/credit_handling.cpp#L495-L502

I am just getting started with hpx. I found this exception weird (checkpoint-related functions are never used and do not even have function address due to no template instantiation) and debugged a bit, I found that multiple no-op functions pointing to the same location, which may cause the problem. However, I'm not confident it's an issue caused by hpx library or cmake or vcpkg port or msvc compiler.

## Expected Behavior

`hpx::serialization::detail::extra_archive_data_helper<hpx::serialization::detail::preprocess_gid_types>::id(void)` and `hpx::serialization::detail::extra_archive_data_helper<hpx::util::checkpointing_tag>::id(void)` should have different address. Otherwise, it will break `id_ == id` checking in `hpx::serialization::detail::extra_archive_data_node::get()`.

https://github.com/STEllAR-GROUP/hpx/blob/76a2ba5b266008be5e0d5a13e17ae0950099dde2/libs/core/serialization/include/hpx/serialization/detail/extra_archive_data.hpp#L117-L136

## Actual Behavior

A bunch of functions got same address, including two mentioned above.

Part of hpx.dll exports with same address (using IDA), ordinal 5311 and 5312 are the two functions mentioned above.
image

  1. Get vcpkg and install hpx with `vcpkg install hpx`
  2. Check function address in hpx.dll (with a Disassembler or print with iostream)

## Specifications

  - HPX Version: 1.7.0
  - Compiler: MSVC cl 16.10.2+857e5a733
  - OS: Windows 10 19043.1110
  - Arch: amd64

hkaiser:
Fixes #5478

### DIFF
--- a/libs/core/serialization/include/hpx/serialization/detail/extra_archive_data.hpp
+++ b/libs/core/serialization/include/hpx/serialization/detail/extra_archive_data.hpp
@@ -14,24 +14,24 @@
 
 namespace hpx { namespace serialization { namespace detail {
 
+    using extra_archive_data_id_type = void*;
+
     template <typename T>
     struct extra_archive_data_helper
     {
         // this is intentionally left unimplemented, will lead to linker errors
         // if used with unknown data type
-        static void id() noexcept;
+        static extra_archive_data_id_type id() noexcept;
 
-        // this is a function that should e implemented in order to reset the
+        // this is a function that should be implemented in order to reset the
         // extra archive data item
-        static constexpr void reset(T* data) noexcept;
+        static void reset(T* data) noexcept;
     };
 
-    using extra_archive_data_id_type = void (*)();
-
     template <typename T>
-    constexpr extra_archive_data_id_type extra_archive_data_id() noexcept
+    extra_archive_data_id_type extra_archive_data_id() noexcept
     {
-        return &extra_archive_data_helper<T>::id;
+        return extra_archive_data_helper<T>::id();
     }
 
     template <typename T>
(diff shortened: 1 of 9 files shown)


jjyyxx:
This fix work as expected (static variable's address uniqueness is guaranteed). But I doubt the id will change run-to-run, will this affect loading dumped checkpoint (I'm not sure whether dumped checkpoint contains such id), or is just OK?

wrt the implementation, I see this static variable technique most used in template definition, where multiple template instantiation got unique static variables automatically. In this implementation, the template function is left unimplemented to forbid unknown data types, so why not just return a hard-coded number constant in each template instantiation (though not very elegant)?

hkaiser:
> This fix work as expected (static variable's address uniqueness is guaranteed). 

Thanks for checking. I'll fix the inspect errors and merge afterwards.

> But I doubt the id will change run-to-run, will this affect loading dumped checkpoint (I'm not sure whether dumped checkpoint contains such id), or is just OK?

The id is not used anywhere except for identifying the data type at runtime. So it will not affect the checkpointing data.

> wrt the implementation, I see this static variable technique most used in template definition, where multiple template instantiation got unique static variables automatically. In this implementation, the template function is left unimplemented to forbid unknown data types, so why not just return a hard-coded number constant in each template instantiation (though not very elegant)?

The point of this solution is to provide an extensible solution (possibly even by the user) that guarantees uniqueness of the id's at runtime.

aurianer (approved):
Thanks!

hkaiser:
retest lsu

msimberg:
> retest lsu

Rostam is still down...

hkaiser:
retest lsu

PR #5484, opened 2021-07-30
Add missing header to jacobi_component/server/solver.hpp

msimberg:
With networking on, `find_all_localities` is included implicitly. In certain configurations with networking off, `find_all_localities` is no longer included. This explicitly includes `hpx/runtime.hpp`.

Flyby: format the file.


aurianer (approved):
LGTM thanks!

PR #5487, opened 2021-07-30
Add get_completion_scheduler CPO and customize bulk for thread_pool_scheduler

msimberg:
Customizes `bulk` for `thread_pool_scheduler`. The customization currently does *not* do any chunking or hierarchical spawning like the `parallel_executor` does. Hierarchical spawning will be a must that I will do separately. I also think the chunking should be the responsibility of the scheduler, rather than the algorithms, but there may be trade-offs I'm missing (and things like this should not be necessary: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/full/compute_cuda/include/hpx/compute/cuda/default_executor_parameters.hpp).

The `get_completion_scheduler` customizations are somewhat verbose, but as far as I can tell it's all mostly necessary. If there are ideas on simplifying them I'd be happy to hear them. Also, all senders don't have a customization since I'm not yet sure it makes sense in those cases. E.g. `when_all` can complete on many different schedulers. `let_value` can also in principle use different senders and thus different schedulers. `just` doesn't have one inherently.

Flybys: leftover renaming from `exec` to `scheduler`. Some utility type traits and constexpr variables (i.e. `_t` and `_v` versions of existing facilities). Moved `make_counting_shape` to `iterator_support` module and `detail` namespace. Fix some typos.

### DIFF
--- a/libs/core/execution_base/include/hpx/execution_base/completion_scheduler.hpp
+++ b/libs/core/execution_base/include/hpx/execution_base/completion_scheduler.hpp
@@ -1,0 +1,92 @@
+//  Copyright (c) 2021 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/config.hpp>
+#include <hpx/execution_base/sender.hpp>
+#include <hpx/functional/tag_dispatch.hpp>
+
+#include <type_traits>
+
+namespace hpx::execution::experimental {
+    template <typename Scheduler>
+    struct get_completion_scheduler_t final
+      : hpx::functional::tag<get_completion_scheduler_t<Scheduler>>
+    {
+    };
+
+    template <typename Scheduler>
+    HPX_INLINE_CONSTEXPR_VARIABLE get_completion_scheduler_t<Scheduler>
+        get_completion_scheduler{};
+
+    namespace detail {
+        template <bool TagDispatchable, typename CPO, typename Sender>
+        struct has_completion_scheduler_impl : std::false_type
+        {
+        };
+
+        template <typename CPO, typename Sender>
+        struct has_completion_scheduler_impl<true, CPO, Sender>
+          : hpx::execution::experimental::is_scheduler<hpx::functional::
+                    tag_dispatch_result_t<get_completion_scheduler_t<CPO>,
+                        std::decay_t<Sender> const&>>
+        {
+        };
+
+        template <typename CPO, typename Sender>
+        struct has_completion_scheduler
+          : has_completion_scheduler_impl<
+                hpx::functional::is_tag_dispatchable_v<
+                    get_completion_scheduler_t<CPO>,
+                    std::decay_t<Sender> const&>,
+                CPO, Sender>
+        {
+        };
+
+        template <typename CPO, typename Sender>
+        HPX_INLINE_CONSTEXPR_VARIABLE bool has_completion_scheduler_v =
+            has_completion_scheduler<CPO, Sender>::value;
+
+        template <bool HasCompletionScheduler, typename ReceiverCPO,
+            typename Sender, typename AlgorithmCPO, typename... Ts>
+        struct is_completion_scheduler_tag_dispatchable_impl : std::false_type
+        {
+        };
+
+        template <typename ReceiverCPO, typename Sender, typename AlgorithmCPO,
+            typename... Ts>
+        struct is_completion_scheduler_tag_dispatchable_impl<true, ReceiverCPO,
+            Sender, AlgorithmCPO, Ts...>
+          : std::integral_constant<bool,
+                hpx::functional::is_tag_dispatchable_v<AlgorithmCPO,
+                    hpx::functional::tag_dispatch_result_t<
+                        hpx::execution::experimental::
+                            get_completion_scheduler_t<ReceiverCPO>,
+                        Sender>,
+                    Sender, Ts...>>
+        {
+        };
+
+        template <typename ReceiverCPO, typename Sender, typename AlgorithmCPO,
+            typename... Ts>
+        struct is_completion_scheduler_tag_dispatchable
+          : is_completion_scheduler_tag_dispatchable_impl<
+                hpx::execution::experimental::detail::
+                    has_completion_scheduler_v<ReceiverCPO, Sender>,
+                ReceiverCPO, Sender, AlgorithmCPO, Ts...>
+        {
+        };
+
+        template <typename ReceiverCPO, typename Sender, typename AlgorithmCPO,
+            typename... Ts>
+        HPX_INLINE_CONSTEXPR_VARIABLE bool
+            is_completion_scheduler_tag_dispatchable_v =
+                is_completion_scheduler_tag_dispatchable<ReceiverCPO, Sender,
+                    AlgorithmCPO, Ts...>::value;
+
+    }    // namespace detail
+}    // namespace hpx::execution::experimental
--- a/libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp
+++ b/libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp
@@ -11,6 +11,9 @@
 
 #include <hpx/config.hpp>
 #include <hpx/iterator_support/iterator_adaptor.hpp>
+#include <hpx/iterator_support/iterator_range.hpp>
+#include <hpx/iterator_support/range.hpp>
+#include <hpx/iterator_support/traits/is_range.hpp>
 #include <hpx/type_support/identity.hpp>
 #include <hpx/type_support/lazy_conditional.hpp>
 
--- a/libs/parallelism/execution/include/hpx/execution/algorithms/bulk.hpp
+++ b/libs/parallelism/execution/include/hpx/execution/algorithms/bulk.hpp
@@ -14,6 +14,7 @@
 #include <hpx/errors/try_catch_exception_ptr.hpp>
 #include <hpx/execution/algorithms/detail/partial_algorithm.hpp>
 #include <hpx/execution/algorithms/transform.hpp>
+#include <hpx/execution_base/completion_scheduler.hpp>
 #include <hpx/execution_base/receiver.hpp>
 #include <hpx/execution_base/sender.hpp>
 #include <hpx/functional/invoke_result.hpp>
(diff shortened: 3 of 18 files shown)


hkaiser:
Great stuff, thanks! I'll look closer a bit later.

> Customizes `bulk` for `thread_pool_scheduler`. The customization currently does _not_ do any chunking or hierarchical spawning like the `parallel_executor` does. Hierarchical spawning will be a must that I will do separately. I also think the chunking should be the responsibility of the scheduler, rather than the algorithms, but there may be trade-offs I'm missing (and things like this should not be necessary: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/full/compute_cuda/include/hpx/compute/cuda/default_executor_parameters.hpp).

I agree that chunking is a property of executors/schedulers. My plan was to slowly phase out the separate parameters objects anyways. All of the things that can be done with parameters can already be done using executors today, so this is not too far off.

> The `get_completion_scheduler` customizations are somewhat verbose, but as far as I can tell it's all mostly necessary. If there are ideas on simplifying them I'd be happy to hear them. Also, all senders don't have a customization since I'm not yet sure it makes sense in those cases. E.g. `when_all` can complete on many different schedulers. `let_value` can also in principle use different senders and thus different schedulers. `just` doesn't have one inherently.

P2304 says that `algorithm(snd, args…)` is equivalent to

- `tag_invoke(<algorithm>, get_completion_scheduler<cpo>(snd), snd, args...)` if valid
- `tag_invoke(<algorithm>, snd, args...)` otherwise

so I assume that every sender should at least propagate the completion scheduler from its predecessor

> Flybys: leftover renaming from `exec` to `scheduler`. Some utility type traits and constexpr variables (i.e. `_t` and `_v` versions of existing facilities). Moved `make_counting_shape` to `iterator_support` module and `detail` namespace. Fix some typos.

+1

msimberg:
> Great stuff, thanks! I'll look closer a bit later.

Thanks, much appreciated! This clearly still needs a bit more cleanup, but the overall picture should be more or less there.

> I agree that chunking is a property of executors/schedulers. My plan was to slowly phase out the separate parameters objects anyways. All of the things that can be done with parameters can already be done using executors today, so this is not too far off.

:+1: I should add that we don't necessarily need to get rid of chunking in the algorithms right away, since the schedulers/executors can do their own chunking on top of it. If the algorithm has already chunked things up the scheduler will most likely choose not to chunk things up further (although there would likely be additional unnecessary wrapping, so we do still want to get rid of possible double chunking when we can).

> so I assume that every sender should at least propagate the completion scheduler from its predecessor

Not every sender, but those that have one unambigous predecessor sender. To make it a bit more explicit, the following forward the scheduler:

- `bulk`
- `transform`

The following customize `get_completion_scheduler` but don't need to do any forwarding:

- `on` (it forwards the `set_done` and `set_error` specializations, though I'm not yet sure if this is the right behaviour either; the completion scheduler depends on where the error comes from)
- `just_on` since `on` does so

`sync_wait` and `detach` don't produce senders. `just` doesn't have a scheduler. `when_all`, `let_value`, and `let_error` are ambiguous. However, `let_error` could actually forward the `set_value` and `set_done` schedulers, and `let_value` could forward the `set_error` and `set_done` schedulers. `ensure_started` could in fact also forward the scheduler. `make_future` and `keep_future` are type-erased and don't have a scheduler. I'll add the missing ones I've listed here.

libs/core/execution_base/include/hpx/execution_base/completion_scheduler.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include <hpx/config.hpp>
+#include <hpx/execution_base/sender.hpp>
+#include <hpx/functional/tag_dispatch.hpp>
+
+#include <type_traits>
+
+namespace hpx::execution::experimental {
+    template <typename Scheduler>
+    struct get_completion_scheduler_t
```
hkaiser, line 17: I think it is good practice to mark the CPOs as `final`

libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp
```diff
@@ -183,4 +186,19 @@ namespace hpx { namespace util {
...
+        template <typename Incrementable>
+        HPX_HOST_DEVICE inline counting_shape_type<Incrementable>
+        make_counting_shape(Incrementable n)
+        {
+            return hpx::util::make_iterator_range(
+                hpx::util::make_counting_iterator(Incrementable(0)),
+                hpx::util::make_counting_iterator(n));
+        }
+    }    // namespace detail
+}}       // namespace hpx::util
```
hkaiser, line 204: #5429 moves this facility to it's own file, see https://github.com/STEllAR-GROUP/hpx/pull/5429/files#diff-c3c03e6827ab3b06f782481b097c19238540e0a89ef723a46982430a2944a7ef.

libs/parallelism/execution/include/hpx/execution/algorithms/bulk.hpp
```diff
@@ -127,28 +124,54 @@ namespace hpx { namespace execution { namespace experimental {
...
     HPX_INLINE_CONSTEXPR_VARIABLE struct bulk_t final
-      : hpx::functional::tag_fallback<bulk_t>
+      : hpx::functional::tag_priority<bulk_t>
     {
     private:
+        // clang-format off
+        template <typename Sender, typename Shape, typename F,
+            HPX_CONCEPT_REQUIRES_(
+                is_sender_v<Sender> &&
+                std::is_integral<Shape>::value &&
```
hkaiser, line 155: This overload should work for both, integral shapes, and iterative ones.

libs/parallelism/execution/include/hpx/execution/algorithms/bulk.hpp
```diff
@@ -127,28 +124,54 @@ namespace hpx { namespace execution { namespace experimental {
...
+                hpx::execution::experimental::detail::has_completion_scheduler_v<
+                    hpx::execution::experimental::set_value_t, Sender> &&
+                hpx::functional::is_tag_dispatchable_v<bulk_t,
+                    hpx::functional::tag_dispatch_result_t<
+                        hpx::execution::experimental::get_completion_scheduler_t<
+                            hpx::execution::experimental::set_value_t>, Sender>,
+                    Sender, Shape, F>
```
hkaiser, line 162: Should we combine these two conditions into a separate facility, I would think that can be used in many places.

libs/parallelism/execution/include/hpx/execution/algorithms/on.hpp
```diff
@@ -59,6 +59,36 @@ namespace hpx { namespace execution { namespace experimental {
...
+                HPX_CONCEPT_REQUIRES_(
+                    hpx::execution::experimental::detail::is_receiver_cpo_v<CPO> &&
+                    (std::is_same_v<CPO, hpx::execution::experimental::set_value_t> ||
+                        hpx::execution::experimental::detail::has_completion_scheduler_v<
+                                hpx::execution::experimental::set_done_t,
+                                std::decay_t<Sender>> ||
+                        hpx::execution::experimental::detail::has_completion_scheduler_v<
+                                hpx::execution::experimental::set_done_t,
+                                std::decay_t<Sender>>))
+                // clang-format on
```
hkaiser, line 73: The last two conditions are equivalent, typo?

libs/parallelism/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -0,0 +1,301 @@
...
+    template <typename Sender, typename Shape, typename F,
+        HPX_CONCEPT_REQUIRES_(!std::is_integral_v<std::decay_t<Shape>>)>
+    constexpr auto tag_dispatch(bulk_t, thread_pool_scheduler scheduler,
+        Sender&& sender, Shape&& shape, F&& f)
+    {
+        return detail::thread_pool_bulk_sender<std::decay_t<Sender>,
+            std::decay_t<Shape>, std::decay_t<F>>{std::move(scheduler),
+            std::forward<Sender>(sender), std::forward<Shape>(shape),
+            std::forward<F>(f)};
+    }
```
hkaiser, line 300: Nice solution!

libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp
msimberg: We're on the same page then :) Do you mind the move to `detail`? I'd go with your PR first. I can rebase this on yours.

libs/parallelism/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
hkaiser: Nah, I just liked the idea of creating a wrapper that implements the bulk part without replicating the rest.

libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp
hkaiser: Having it in namespace detail is fine.

msimberg:
retest cscs

msimberg:
I'm hoping this is now ready to go. The last problem only showed up on release builds with clang. I was capturing `this` instead of `op_state` here https://github.com/STEllAR-GROUP/hpx/blob/b316e223427c1689c4eff6a4d57f1f6c8517a8af/libs/parallelism/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp#L182. That was correct-by-luck in most situations, but would fail weirdly with optimizations.

aurianer (dismissed):
LGTM thanks!
Codacy is unhappy but that should be easily fixable ;)

msimberg:
> Codacy is unhappy but that should be easily fixable ;)

I think that warning is bogus :/ at least I can't see a problem with it. It's not actually an assignment to itself.

hkaiser:
> > Codacy is unhappy but that should be easily fixable ;)
> 
> I think that warning is bogus :/ at least I can't see a problem with it. It's not actually an assignment to itself.

I usually just tell codacy to ignore these things.

aurianer:
> > Codacy is unhappy but that should be easily fixable ;)
> 
> I think that warning is bogus :/ at least I can't see a problem with it. It's not actually an assignment to itself.

I was just guessing adding an underscore to op_state would avoid the warning but I'm fine with ignoring the warning too

msimberg:
> I was just guessing adding an underscore to op_state would avoid the warning but I'm fine with ignoring the warning too

I'll ignore it for now in the hope that Codacy/cppcheck will improve and not warn about it in the future, but if it keeps coming up let's rename it.

PR #5489, opened 2021-08-01
Adding bound-checking for bind placeholders

### LINKED ISSUE #5488: `hpx::util::bind` doesn't bounds-check placeholders
The following should compile, but doesn't:
```
template <typename F,
    std::enable_if_t<hpx::is_invocable_v<F>, int> = 0>
auto test(F&& f)
{
    return f();
}

template <typename F,
    std::enable_if_t<hpx::is_invocable_v<F, int>, int> = 0>
auto test(F&& f)
{
    return f(42);
}

void foo(int) {}

int main()
{
    test(hpx::util::bind(foo,  hpx::util::placeholders::_1));
    return 0;
}
```

hkaiser:
Fixes #5488

- flyby: modernize bind and related facilities

### DIFF
--- a/libs/core/functional/include/hpx/functional/bind.hpp
+++ b/libs/core/functional/include/hpx/functional/bind.hpp
@@ -26,6 +26,7 @@
 #include <utility>
 
 namespace hpx { namespace util {
+
     ///////////////////////////////////////////////////////////////////////////
     namespace detail {
         template <std::size_t I>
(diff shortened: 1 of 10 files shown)


K-ballo (approved):
Looks good. A conditionally deleted overload would provide significantly better error messages when not passing enough bound arguments, but that can be added later on.

hkaiser:
retest

hkaiser:
@msimberg can we go ahead with this? I think it's ready.

PR #5491, opened 2021-08-02
Remove Boost.Filesystem compatibility layer

msimberg:
Keeps the utility functions that are in Boost.Filesystem but not in std::filesystem.

Just checking if this even compiles on all our CI configurations first.


hkaiser:
Are we sure all platforms we're interested in support std filesystem now? What about MacOS? Also, cpp-reference says: 
> Using this library may require additional compiler/linker options. GNU implementation prior to 9.1 requires linking with -lstdc++fs and LLVM implementation prior to LLVM 9.0 requires linking with -lc++fs.

Should we drop support for those?

msimberg:
> Are we sure all platforms we're interested in support std filesystem now? What about MacOS? Also, cpp-reference says:

No, we're not sure (hence the draft). All I know is that the GitHub actions (the one for macOS) build detects filesystem and builds fine, but we'll need to see how many versions we'd have to drop if we actually go ahead with this.

> Should we drop support for those?

Maybe? :)

aurianer:
> > Are we sure all platforms we're interested in support std filesystem now? What about MacOS? Also, cpp-reference says:
> 
> No, we're not sure (hence the draft). All I know is that the GitHub actions (the one for macOS) build detects filesystem and builds fine, but we'll need to see how many versions we'd have to drop if we actually go ahead with this.
> 
> > Should we drop support for those?
> 
> Maybe? :)

Dropping support for those would also allow us to get rid of the [sfinaes](https://github.com/STEllAR-GROUP/hpx/pull/5283/commits/3b9b3cdb59e807a51a5be7759703008a59d8fcbd) necessary in the senders/receivers code :)

hkaiser:
What about if we enabled the std filesystem by default (i.e. leaving the boost filesystem code in place, but disabled). That would allow users to fall back for the time being.

msimberg:
> What about if we enabled the std filesystem by default (i.e. leaving the boost filesystem code in place, but disabled). That would allow users to fall back for the time being.

That's essentially what we do now on master. We use `std::filesystem` if it's available, and fall back (implicitly) to Boost filesystem otherwise. I don't know if you meant that they'd have to explicitly fall back to Boost filesystem, but I don't think that makes much of a difference.

Note that I'm not in a huge rush to go ahead with this, especially if it looks like a lot of our currently supported compilers don't support `std::filesystem`. I just wanted to start by checking exactly how many don't support it.

msimberg:
retest cscs

msimberg:
I'll close this in its current form and come back to it at a later point.

PR #5493, opened 2021-08-03
Update Jenkins GitHub token to use StellarBot GitHub account

msimberg:
To do:
- [x] Wait for Rostam and Piz Daint Jenkins instances to start working again
- [x] Check that new token works on Piz Daint
- [x] Check that new token works on Rostam

### DIFF
--- a/.jenkins/cscs/entry.sh
+++ b/.jenkins/cscs/entry.sh
@@ -61,12 +61,9 @@ if [[ -n "${ghprbPullId:-}" ]]; then
     # Get the CDash dashboard build id
     cdash_build_id="$(cat jenkins-hpx-${configuration_name}-cdash-build-id.txt)"
 
-    # Extract actual token from GITHUB_TOKEN (in the form "username:token")
-    github_token=$(echo ${GITHUB_TOKEN} | cut -f2 -d':')
-
     # Set GitHub status with CDash url
     .jenkins/common/set_github_status.sh \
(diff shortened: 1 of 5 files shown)


.jenkins/cscs/entry.sh
```diff
@@ -61,12 +61,9 @@ if [[ -n "${ghprbPullId:-}" ]]; then
     # Get the CDash dashboard build id
     cdash_build_id="$(cat jenkins-hpx-${configuration_name}-cdash-build-id.txt)"
 
-    # Extract actual token from GITHUB_TOKEN (in the form "username:token")
-    github_token=$(echo ${GITHUB_TOKEN} | cut -f2 -d':')
-
     # Set GitHub status with CDash url
     .jenkins/common/set_github_status.sh \
-        "${github_token}" \
+        "${GITHUB_TOKEN}" \
```
hkaiser, line 66: How did that ever work before? ;-)

hkaiser (approved):
LGTM, thanks!

.jenkins/cscs/entry.sh
msimberg: See the removed lines above this. I've changed how the token is stored in Jenkins from `username:token` to just `token`, so the intermediate step of extracting the token itself shouldn't be necessary anymore.

aurianer (dismissed):
Thanks!

PR #5499, opened 2021-08-06
Rename simdpar to par_simd

### LINKED ISSUE #5494: Rename `simdpar` execution policy to `par_simd`
This is for consistency with the `unseq` and `par_unseq` policies.

msimberg:
Fixes #5494. I just did a find-and-replace, hoping there are no surprises.

### DIFF
--- a/libs/parallelism/algorithms/include/hpx/parallel/datapar/transform_loop.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/datapar/transform_loop.hpp
@@ -421,7 +421,7 @@ namespace hpx { namespace parallel { namespace util {
                 F&& f)
             {
                 auto ret = util::transform_binary_loop_n<
-                    hpx::execution::simdpar_policy>(first1,
+                    hpx::execution::par_simd_policy>(first1,
                     std::distance(first1, last1), first2, dest,
                     std::forward<F>(f));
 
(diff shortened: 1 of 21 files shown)


hkaiser (approved):
LGTM, thanks!

hkaiser:
@srinivasyadav18 could you please verify this doesn't break anything?

msimberg:
Am I right in my understanding that if `HPX_WITH_CXX20_EXPERIMENTAL_SIMD` is detected then there should be tests for this? `gcc-newest` (i.e. GCC 11 with C++20) seems to be running some tests with `datapar` in their name.

srinivasyadav18:
> Am I right in my understanding that if `HPX_WITH_CXX20_EXPERIMENTAL_SIMD` is detected then there should be tests for this? `gcc-newest` (i.e. GCC 11 with C++20) seems to be running some tests with `datapar` in their name.

Yup, we use the existing datapar tests defined [here](https://github.com/STEllAR-GROUP/hpx/blob/master/libs/parallelism/algorithms/tests/unit/datapar_algorithms/CMakeLists.txt#L9). As it works for both `Vc` backend and `std-simd` backend.

srinivasyadav18 (approved):
LGTM, thanks! ;-)

PR #5500, opened 2021-08-06
Minor bug fix in transform exclusive and inclusive scan tests

Jedi18:
Will lead to a segmentation fault if end_len happens to be the vector size value.


msimberg (approved):
Looks good, thanks!

hkaiser (approved):
LGTM, thanks!

PR #5501, opened 2021-08-09
Attempt to fix thread_pool_scheduler test

msimberg:
Added predicates to the `wait_for`s. Without them the waits may fail even if the the tasks were executed (and since https://github.com/STEllAR-GROUP/hpx/blob/0bfa445daefecb5ab4f8a4bb1873c3b92ceed468/libs/parallelism/executors/tests/unit/thread_pool_scheduler.cpp#L1229 seems to not fail, it must be the case here).

### DIFF
--- a/libs/parallelism/executors/tests/unit/thread_pool_scheduler.cpp
+++ b/libs/parallelism/executors/tests/unit/thread_pool_scheduler.cpp
@@ -1223,8 +1223,8 @@ void test_detach()
 
         {
             std::unique_lock<hpx::mutex> l(m);
-            hpx::cv_status st = cv.wait_for(l, std::chrono::seconds(1));
-            HPX_TEST(st == hpx::cv_status::no_timeout);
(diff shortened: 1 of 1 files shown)


hkaiser (dismissed):
LGTM, even if I was not aware that we have spurious wakeups.... Thanks, I would have never thought of applying this change!

msimberg:
> LGTM, even if I was not aware that we have spurious wakeups.... Thanks, I would have never thought of applying this change!

The failure doesn't require spurious wakeups (and I don't think they happen in the runtime). It just requires that the task notifies before the wait happens (either the task happens to be scheduled immediately, or more likely that the waiting thread happens to give up its timeslice just before it waits).

Note: I wasn't able to reproduce this locally, so I'm only guessing that it fixes the problem. The symptom and fix does fit CircleCI being heavily oversubscribed though (and the fix is needed in any case).

PR #5506, opened 2021-08-13
Fix condition for checking if we should use our own variant

msimberg:
There's no longer a feature check for `std::variant`. We assume it is always available.

### DIFF
--- a/libs/core/datastructures/include/hpx/datastructures/variant.hpp
+++ b/libs/core/datastructures/include/hpx/datastructures/variant.hpp
@@ -13,6 +13,7 @@
 
 namespace hpx {
 
+    using std::get;
     using std::monostate;
     using std::variant;
     using std::visit;
@@ -24,6 +25,7 @@ namespace hpx {
 
 namespace hpx {
 
+    using hpx::variant_ns::get;
     using hpx::variant_ns::monostate;
     using hpx::variant_ns::variant;
     using hpx::variant_ns::visit;
--- a/libs/core/datastructures/include/hpx/datastructures/variant.hpp
+++ b/libs/core/datastructures/include/hpx/datastructures/variant.hpp
@@ -8,7 +8,7 @@
 
 #include <hpx/config.hpp>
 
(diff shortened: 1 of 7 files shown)


hkaiser (approved):
LGTM, thanks!

msimberg:
retest cscs

aurianer (approved):
LGTM thanks a lot I probably messed up when rebasing :/

msimberg:
> LGTM thanks a lot I probably messed up when rebasing :/

I doubt it. There were enough changes to try to keep track of.

Note that this does not compile with CUDA for some mysterious reason. Changing to draft so noone accidentally merges.

msimberg:
This now compiles again with CUDA on Piz Daint (there's another CUDA problem with the CUDA version on Rostam, but I'm going to look at that separately). The original problem appeared when enabling `namespace hpx { using std::variant; }`. The three character change to not trigger the problem was to change `hpx::tuple` to `std::tuple` here: https://github.com/STEllAR-GROUP/hpx/blob/0db6fc565cc24a1356d0d87ae3728b1f0b69d399/libs/parallelism/execution/include/hpx/execution/algorithms/bulk.hpp#L101-L103. I'm assuming this is a result of some sort of name lookup bug in nvcc. In any case, since https://github.com/STEllAR-GROUP/hpx/blob/0db6fc565cc24a1356d0d87ae3728b1f0b69d399/libs/parallelism/execution/include/hpx/execution/algorithms/bulk.hpp#L101-L103 was nonsense anyway (I've copy-pasted it from another sender algorithm) I changed the code anyway and the problem no longer appears. I'm hoping it won't come back to bite us. I've also changed all uses of `std::variant` to `hpx::variant` (and the same for `tuple` and `optional`) in the sender algorithms. I'm somewhat worried about what this will do https://github.com/STEllAR-GROUP/hpx/blob/0724a93172c114093feaeb73ca4a64d5e4ce04d6/libs/core/datastructures/include/hpx/datastructures/variant.hpp#L16, but at least CI seems to be happy with it.

aurianer (approved):
LGTM thanks!

PR #5509, opened 2021-08-16
Test with Boost 1.77 on gcc/clang-newest configurations

msimberg:


msimberg:
retest cscs

msimberg:
Merging since this seems to work without any changes. According to @K-ballo it also builds cleanly on Windows.

PR #5510, opened 2021-08-17
Add any_sender and unique_any_sender

msimberg:
Adds type-erased senders, `any_sender` and `unique_any_sender` (of which the latter is move-only). I'd appreciate feedback on this as it's the first time I do any serious type-erasing. The type-erasure mechanisms don't have to stay this way if the techniques used elsewhere are preferable.

The senders are currently templated on the value types that the wrapped senders send, and only on that. The wrapped senders must send only one set of values (i.e. no multiple variants in `value_types`). The senders are currently not templated on the error type at all, and assume `std::exception_ptr`. This could easily be changed if needed.

The buffers used for SBO are quite big right now and may be self-defeating. I guess experience will show what are good sizes for the buffers. It can of course also become a template parameter but I'd prefer not putting it there unless there's a strong need for it.

There are still a few TODOs in the code, but the overall structure is there, and the tests pass.

### DIFF
--- a/libs/core/execution/include/hpx/execution/algorithms/just.hpp
+++ b/libs/core/execution/include/hpx/execution/algorithms/just.hpp
@@ -29,11 +29,27 @@ namespace hpx { namespace execution { namespace experimental {
         {
             hpx::util::member_pack_for<std::decay_t<Ts>...> ts;
 
-            template <typename... Ts_>
-            explicit constexpr just_sender(Ts_&&... ts)
-              : ts(std::piecewise_construct, std::forward<Ts_>(ts)...)
+            constexpr just_sender() = default;
+
+            template <typename T,
+                typename = std::enable_if_t<
+                    !std::is_same_v<std::decay_t<T>, just_sender>>>
+            explicit constexpr just_sender(T&& t)
+              : ts(std::piecewise_construct, std::forward<T>(t))
             {
             }
+
+            template <typename T0, typename T1, typename... Ts_>
+            explicit constexpr just_sender(T0&& t0, T1&& t1, Ts_&&... ts)
+              : ts(std::piecewise_construct, std::forward<T0>(t0),
+                    std::forward<T1>(t1), std::forward<Ts_>(ts)...)
+            {
+            }
+
+            just_sender(just_sender&&) = default;
+            just_sender(just_sender const&) = default;
+            just_sender& operator=(just_sender&&) = default;
+            just_sender& operator=(just_sender const&) = default;
 
             template <template <typename...> class Tuple,
                 template <typename...> class Variant>
--- a/libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
+++ b/libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
@@ -1,0 +1,811 @@
+//  Copyright (c) 2021 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/assert.hpp>
+#include <hpx/errors/error.hpp>
+#include <hpx/errors/throw_exception.hpp>
+#include <hpx/execution_base/sender.hpp>
+
+#include <cstddef>
+#include <cstring>
+#include <exception>
+#include <memory>
+#include <type_traits>
+#include <utility>
+
+namespace hpx::detail {
+    template <typename T>
+    struct empty_vtable
+    {
+        static_assert(
+            sizeof(T) == 0, "No empty vtable type defined for given type T");
+    };
+
+    template <typename T>
+    struct get_empty_vtable
+    {
+        using empty_vtable_type = typename empty_vtable<T>::type;
+        static_assert(std::is_base_of_v<T, empty_vtable_type>,
+            "Given empty vtable type should be a base of T");
+
+        static T* call()
+        {
+            static empty_vtable_type empty;
+            return &empty;
+        }
+    };
+
+    template <typename Base, std::size_t EmbeddedStorageSize,
+        std::size_t AlignmentSize = sizeof(void*)>
+    class movable_sbo_storage
+    {
+    protected:
+        using base_type = Base;
+        static constexpr std::size_t embedded_storage_size =
+            EmbeddedStorageSize;
+        static constexpr std::size_t alignment_size = AlignmentSize;
+
+        // The union has two members:
+        // - embedded_storage: Embedded storage size array used for types that
+        //   are at most embedded_storage_size bytes large, and require at most
+        //   alignment_size alignment.
+        // - heap_storage: A pointer to base_type that is used when objects
+        //   don't fit in the embedded storage.
+        union
+        {
+            std::aligned_storage_t<embedded_storage_size, alignment_size>
+                embedded_storage;
+            base_type* heap_storage;
+        };
+        base_type* object;
+
+        // Returns true when it's safe to use the embedded storage, i.e.
+        // when the size and alignment of Impl are small enough.
+        template <typename Impl>
+        static constexpr bool can_use_embedded_storage()
+        {
+            constexpr bool fits_storage =
+                sizeof(std::decay_t<Impl>) <= embedded_storage_size;
+            constexpr bool sufficiently_aligned =
+                alignof(std::decay_t<Impl>) <= alignment_size;
+            return fits_storage && sufficiently_aligned;
+        }
+
+        bool using_embedded_storage() const noexcept
+        {
+            return object ==
+                reinterpret_cast<base_type const*>(&embedded_storage);
+        }
+
+        bool empty() const noexcept
+        {
+            return get().empty();
+        }
+
+        void release()
+        {
+            HPX_ASSERT(!empty());
+
+            if (using_embedded_storage())
+            {
+                get().~base_type();
+            }
+            else
+            {
+                delete heap_storage;
+                heap_storage = nullptr;
+            }
+
+            object = get_empty_vtable<base_type>::call();
+        }
+
+        void move_assign(movable_sbo_storage&& other) &
+
+        {
+            HPX_ASSERT(&other != this);
+            HPX_ASSERT(empty());
+
+            if (!other.empty())
+            {
+                if (other.using_embedded_storage())
+                {
+                    auto p = reinterpret_cast<base_type*>(&embedded_storage);
+                    other.get().move_into(p);
+                    object = p;
+                }
+                else
+                {
+                    heap_storage = other.heap_storage;
+                    other.heap_storage = nullptr;
+                    object = heap_storage;
+                }
+
+                other.object = get_empty_vtable<base_type>::call();
+            }
+        }
+
+    public:
+        movable_sbo_storage()
+          : heap_storage(nullptr)
+          , object(get_empty_vtable<base_type>::call())
+        {
+        }
+
+        ~movable_sbo_storage()
+        {
+            if (!empty())
+            {
+                release();
+            }
+        }
+
+        movable_sbo_storage(movable_sbo_storage&& other)
+          : heap_storage(nullptr)
+          , object(get_empty_vtable<base_type>::call())
+        {
+            move_assign(std::move(other));
+        }
+
+        movable_sbo_storage& operator=(movable_sbo_storage&& other)
+        {
+            if (&other != this)
+            {
+                if (!empty())
+                {
+                    release();
+                }
+
+                move_assign(std::move(other));
+            }
+            return *this;
+        }
+
+        movable_sbo_storage(movable_sbo_storage const&) = delete;
+        movable_sbo_storage& operator=(movable_sbo_storage const&) = delete;
+
+        base_type const& get() const noexcept
+        {
+            return *object;
+        }
+
+        base_type& get() noexcept
+        {
+            return *object;
+        }
+
+        template <typename Impl, typename... Ts>
+        void store(Ts&&... ts)
+        {
+            if (!empty())
+            {
+                release();
+            }
+
+            if constexpr (can_use_embedded_storage<Impl>())
+            {
+                Impl* p = reinterpret_cast<Impl*>(&embedded_storage);
+                new (p) Impl(std::forward<Ts>(ts)...);
+                object = p;
+            }
+            else
+            {
+                heap_storage = new Impl(std::forward<Ts>(ts)...);
+                object = heap_storage;
+            }
+        }
+    };
+
+    template <typename Base, std::size_t EmbeddedStorageSize,
+        std::size_t AlignmentSize = sizeof(void*)>
+    class copyable_sbo_storage
+      : public movable_sbo_storage<Base, EmbeddedStorageSize, AlignmentSize>
+    {
+        using storage_base_type =
+            movable_sbo_storage<Base, EmbeddedStorageSize, AlignmentSize>;
+
+        using typename storage_base_type::base_type;
+
+        using storage_base_type::embedded_storage;
+        using storage_base_type::empty;
+        using storage_base_type::heap_storage;
+        using storage_base_type::object;
+        using storage_base_type::release;
+        using storage_base_type::using_embedded_storage;
+
+        void copy_assign(copyable_sbo_storage const& other) &
+        {
+            HPX_ASSERT(&other != this);
+            HPX_ASSERT(empty());
+
+            if (!other.empty())
+            {
+                if (other.using_embedded_storage())
+                {
+                    base_type* p =
+                        reinterpret_cast<base_type*>(&embedded_storage);
+                    other.get().clone_into(p);
+                    object = p;
+                }
+                else
+                {
+                    heap_storage = other.get().clone();
+                    object = heap_storage;
+                }
+            }
+        }
+
+    public:
+        using storage_base_type::get;
+        using storage_base_type::store;
+
+        copyable_sbo_storage() = default;
+        copyable_sbo_storage(copyable_sbo_storage&&) = default;
+        copyable_sbo_storage& operator=(copyable_sbo_storage&&) = default;
+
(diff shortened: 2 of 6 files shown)


msimberg:
I want to add that a complementary option to this (in some situations) would be a "`variant_sender`" when the set of possible sender types is known ahead of time. It *may* lead to better code generation and is a lot simpler in implementation. However, I'm leaving that out for the moment (if someone is interested please comment, I have a somewhat working version of that as well).

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
```diff
@@ -0,0 +1,730 @@
...
+    struct get_empty_vtable
+    {
+        using empty_vtable_type = typename empty_vtable<T>::type;
+        static_assert(std::is_base_of_v<T, empty_vtable_type>,
+            "Given empty vtable type should be a base of T");
+
+        static T* call()
+        {
+            static empty_vtable_type empty;
+            return &empty;
```
hkaiser, line 39: This is dangerous on platforms that instantiate global and function-local static variables for every module they are visible to (e.g. Windows). In this case, the address returned from `call()` may not be unique.

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
```diff
@@ -0,0 +1,730 @@
...
+        // Returns true when it's safe to use the embedded storage, i.e.
+        // when the size and alignment of Impl are small enough.
+        template <typename Impl>
+        static constexpr bool can_use_embedded_storage()
+        {
+            bool fits_storage =
+                sizeof(std::decay_t<Impl>) <= embedded_storage_size;
+            bool sufficiently_aligned =
+                alignof(std::decay_t<Impl>) <= alignment_size;
+            return fits_storage && sufficiently_aligned;
```
hkaiser, line 82: The variables can be `constexpr`, I believe.

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
msimberg: I feared that. Do we need to do the same thing we do for the extra archive data stuff? Explicitly instantiate this for types we need?

hkaiser: Does the `empty_table` have to be a template?

msimberg: With the current setup yes, unless I'm misunderstanding the question. What alternatives do you see?

hkaiser: I thought that all `any_sender` instances could share the same empty table as in the end all you need is its address. Making `empty_table` a non-template would allow to separate the `call` function into a source file.

msimberg: I see now. There's an empty vtable for `any_sender`, `unique_any_sender`, `any_receiver`, and `any_operation_state`, and I access it generically through this template in the base classes. That's why it's templated. I think I'll just have to explicitly instantiate them in a cpp file.

hkaiser: But, but... do we really have to use different instance of the empty table for those?

msimberg: Well... I don't know? I suppose this is where my inexperience with type-erasure shows. With the current setup they do need different instances. `empty_any_operation_state` does not have the same member functions as `empty_any_receiver`. How is this typically dealt with?

In either case, the function static empty vtable(s) have(/has) to go in a cpp file.

hkaiser: Hmmm, I might miss something, sorry. If the set of types the empty table has to be instantiated for is confined, then yes, simply explicitly instantiate those in a cpp file and export the corresponding `call()` function.

hkaiser: Alternatively you could have separate empty table types (non-templates) for each of the variants and explicitly instantiate the `get_empty_vtable<T>::call()` function

msimberg: > Hmmm, I might miss something, sorry.

So may I.

The set of types is confined at least in this context. If `sbo_storage` is to be useful elsewhere it's of course not confined anymore, though I wouldn't put that in our public API in any case. The empty vtable types themselves are not templated. It's just that I access the (four) different empty vtable types (their instances) through `get_empty_vtable` in `sbo_storage`. It could actually also be a function template, but that doesn't matter too much.

msimberg: Hmm, I take back the part about the set of types being confined. The base types for senders and receivers (not operation state) are templated on the types sent by the wrapped sender. Just to understand the restrictions we're dealing with, this is also problematic, right? https://github.com/STEllAR-GROUP/hpx/blob/65280c318d97c53bee082b8ce87ce78922660070/libs/parallelism/execution/include/hpx/execution/executors/polymorphic_executor.hpp#L592-L599 (it's similarly templated on a function signature which is provided by the user). If that's fine then this would be fine as well? If it's not, then whatever fix I end up with here should be applied there as well.

hkaiser: > Hmm, I take back the part about the set of types being confined. The base types for senders and receivers (not operation state) are templated on the types sent by the wrapped sender. Just to understand the restrictions we're dealing with, this is also problematic, right?
> 
> https://github.com/STEllAR-GROUP/hpx/blob/65280c318d97c53bee082b8ce87ce78922660070/libs/parallelism/execution/include/hpx/execution/executors/polymorphic_executor.hpp#L592-L599
> 
> (it's similarly templated on a function signature which is provided by the user). If that's fine then this would be fine as well? If it's not, then whatever fix I end up with here should be applied there as well.

Yes, this is equally problematic. Good catch.

msimberg: Currently the only reasonable solution I can think of is representing the empty state by `nullptr` rather than a pointer to a base class, and hope that `if (HPX_LIKELY(ptr)) ptr->member(...); else empty{}.member(...);` is not too expensive. Branch prediction *should* make it cheap in any case, but it's not quite the same as avoiding the branch in the first place.

msimberg:
Interestingly there seems to be a trade-off between eager and lazy execution.

Laziness helps avoid allocations in the non-type-erased case. However, in the type-erased case (as implemented here right now) in the worst case there are more allocations than with async/then/dataflow (potentially one for each of the type-erased sender, receiver, and operation state).

I haven't tried this yet, but I think eagerly submitting the work wrapped in an `any_sender` would help remove at least some of those allocations. I'm not sure how this would play with any laziness/eagerness guarantees that P2300 might make (currently there's no type-erasure in it).

It's also possible that my implementation here is just dumb and it would be possible to keep the laziness and reduce potential allocations, but I don't see how to do that right now.

msimberg:
I'd like to keep this in its current v0.1 state, and explore the performance implications (if any) of those allocations and the branching when dispatching separately inside DLA-Future.

I currently don't see any way to avoid that branch when dispatching, though I'm most likely missing something clever.

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
msimberg: We discussed this in the meeting and there are essentially two approaches around this:
- represent the empty state with nullptr and pay for a branch on dispatch
- store a boolean to represent the empty state, and don't pay for branches on dispatch
I've implemented the second here. I think the polymorphic executor is actually fine after looking at it again. It also uses extra storage to represent the empty state, and keeps dispatch branch-free.

hkaiser: A third option would be to add another function to the vtable that returns a bool. This safes on memory for the `any_sender` but adds a slight overhead to the check for emptyness (indirect function call instead of simple memory read).

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
```diff
@@ -0,0 +1,746 @@
...
+
+        copyable_sbo_storage() = default;
+        copyable_sbo_storage(copyable_sbo_storage&&) = default;
+        copyable_sbo_storage& operator=(copyable_sbo_storage&&) = default;
+
+        copyable_sbo_storage(copyable_sbo_storage const& other)
+          : movable_sbo_storage()
+        {
+            *this = other;
+        }
```
hkaiser, line 233:
```suggestion
        copyable_sbo_storage(copyable_sbo_storage const& other)
          : movable_sbo_storage(other)
        {
        }
```

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
```diff
@@ -0,0 +1,746 @@
...
+        //   don't fit in the embedded storage.
+        // - embedded_storage: Fixed size array used for types that are at
+        //   most embedded_storage_size bytes large, and require at most
+        //   alignment_size alignment.
+        union
+        {
+            std::aligned_storage_t<embedded_storage_size, alignment_size>
+                alignment_storage;
+            base_type* heap_storage;
+            unsigned char embedded_storage[embedded_storage_size];
```
hkaiser, line 66: Wouldn't it be sufficient to have the `alignment_storage` and the `heap_storage` members only?

msimberg: Yeah, it would. I mostly separated it because I thought it was clearer like this, at least when I wrote this initially. Now I'm not sure. I could go either way. Do you have a preference?

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
msimberg: Those are not the same though. `sbo_storage` (the alias `movable_sbo_storage`) does not have a copy-constructor.

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
msimberg: I've updated it to use the third option.

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
msimberg: I've removed the separate char array.

libs/core/execution/include/hpx/execution/algorithms/just.hpp
```diff
@@ -29,12 +29,31 @@ namespace hpx { namespace execution { namespace experimental {
         {
             hpx::util::member_pack_for<std::decay_t<Ts>...> ts;
 
-            template <typename... Ts_>
-            explicit constexpr just_sender(Ts_&&... ts)
-              : ts(std::piecewise_construct, std::forward<Ts_>(ts)...)
+            constexpr just_sender()
```
hkaiser, line 32: Wouldn't `just_sender() = default` be sufficient?

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
```diff
@@ -0,0 +1,769 @@
...
+                if (!other.empty())
+                {
+                    if (other.using_embedded_storage())
+                    {
+                        auto p =
+                            reinterpret_cast<base_type*>(&embedded_storage);
+                        auto otherp = reinterpret_cast<base_type*>(
+                            &other.embedded_storage);
+                        // We cast the pointers to void* to silence warnings
+                        // about copying a dynamic class (which don't apply
```
hkaiser, line 147:
```suggestion
                        // about copying a dynamic class (which doesn't apply
```

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
```diff
@@ -0,0 +1,769 @@
...
+                        // here)
+                        std::memcpy(static_cast<void*>(p),
+                            static_cast<void*>(otherp), embedded_storage_size);
```
hkaiser, line 150: Uhhh. This assumes that the embedded object is bitwise moveable. Shouldn't we explicitly `std::move` the rhs here?

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
```diff
@@ -0,0 +1,769 @@
...
+        void store(Ts&&... ts)
+        {
+            if (!empty())
+            {
+                release();
+            }
+
+            if constexpr (can_use_embedded_storage<Impl>())
+            {
+                base_type* p = reinterpret_cast<base_type*>(&embedded_storage);
```
hkaiser, line 187: Should this be `Impl* p = ...; `?

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
hkaiser: Urgs.

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
msimberg: I'm not sure actually. Do you know if it makes a difference? `Impl *` is definitely not less correct at least, so I can change.

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
msimberg: How should I interpret this comment? ;)

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
msimberg: "...warning*s* about copying..."

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
msimberg: My understanding is that this is in fact ok, since we're managing the construction and destruction manually, but I'm definitely not 100% sure. If it's not ok, I'd again like to understand the difference to the polymorphic executor ;) What are the alternatives?

```
*p = std::move(*otherp)
```
without `p` holding a necessarily valid object seems problematic, and
```
new (p) {};
*p = std::move(*otherp);
```
requires a default constructor (though if that's required for this to be correct then so be it).

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
hkaiser: The placement `new` expects a `void*`, so it probably doesn't make a difference in this case. It just looks more consistent to have the right pointer type. You probably would get away with even writing: `new (&embedded_storage) ...;`

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
hkaiser: I simply find it to be unusual that a class has a `foo& operator=(foo const&)` without having a `foo(foo const&)`

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
hkaiser: What about `new (p) Impl(std::move(*otherp));`?

The code you have will probably work for 99.9% of all types. It will however break for instance for types that have a member pointing to another member of the same instance.

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
msimberg: Unfortunately I can't use `&embedded_storage` directly, as it's a completely unrelated type (`aligned_storage`). I updated it to be `Impl*`.

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
msimberg: Good, thanks for explaining! That's a corner case I did not consider. I can anyway use memcpy in the cases that the type is trivially copyable.

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
hkaiser: I was halfway joking anyways ;-)

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
msimberg: But `copyable_sbo_storage` has both the copy constructor and copy assignment operator defined... what am I missing? What did I mess up? ;) And `movable_sbo_storage` has neither (though I haven't explicitly deleted them; will do so now).

hkaiser: Right, I misspoke. I guess what bothered me then was that you implemented the copy constructor in terms of the assignment operator. `*this = std::move(other)` first destructs `*this` and then constructs a copy of the rhs.

msimberg: I see, that makes more sense. Thanks. I'll have another look at this.

msimberg: I actually still don't see it, and this might be due to gaps in my understanding of constructors and assignment operators. There's at least one mixup here. You're `move`-ing `other`, but talking about the copy constructor, so I assume you meant `*this = other`, correct?

At what point is `*this = other` going to do destruct `*this`? Is it because `copyable_sbo_storage& operator=(copyable_sbo_storage const& other)` is not ref qualified? Can or should the copy assignment operator generally actually be ref qualified? In practice it looks like the extra constructor and destructor calls are elided, but I guess you're saying this is not guaranteed?kkk

Whatever the case is, I've moved the copy and move assignment into helper functions and call them from the copy/move constructors/assignment operators instead.

libs/core/execution_base/include/hpx/execution_base/any_sender.hpp
msimberg: I'm now placement-new-ing this. I'm leaving the memcpy for a maybe-sometime-in-the-future optimization.

PR #5512, opened 2021-08-17
Update Rostam HIP configuration

msimberg:
~Explicitly loads clang and a Boost module that was build with clang for the HIP configuration. Without this a Boost module built with GCC (and thus incompatible with HIP) would be used.~ Ignore that. There's this: https://github.com/boostorg/config/issues/392. Which has been fixed on Boost develop. I've put a version check that only defines `BOOST_NOINLINE` definition with HIP for versions older than Boost 1.78. 1.78 is obviously not out yet, but I'm counting on that fix being in 1.78.


msimberg:
retest lsu

msimberg:
retest lsu

msimberg:
I've kept HIP at 4.0.1. The 4.3.0 install is broken somehow and we'll have to look into that separately. The HIP builds are now clean again.

aurianer (dismissed):
LGTM thanks!

msimberg:
Reverted back to 4.0.1 after testing 4.3.0 briefly.

msimberg:
retest lsu

PR #5514, opened 2021-08-18
Add split sender adapter

msimberg:
Adds `split` by reusing the `ensure_started` implementation. Currently, the only distinction between the two is that `ensure_started` eagerly starts the work represented by the given sender (it connects and starts it when calling `ensure_started`), and `split` does not (it starts it when connect and start is called).

Added a TODO for the behaviour of the completion scheduler. Currently, if the given work has already completed by the time more work is connected to an `ensure_started` sender, the continuation is called inline. This may of course start a new execution agent if the continuation is `via`, but it would have to be explicitly requested by the caller. Another option is that in the case that the given work is already complete and we would be in a position to run work inline, we instead `schedule` the continuation given that the sender has a completion scheduler.


libs/parallelism/execution/include/hpx/execution/algorithms/split.hpp
```diff
@@ -0,0 +1,500 @@
...
+                        //   the lock they can safely write to the vector. This
+                        //   thread will not proceed past the lock until they
+                        //   have finished writing to the vector.
+                        //
+                        // Importantly, once this thread has taken and released
+                        // this lock, threads attempting to add continuations to
+                        // the vector must see predecessor_done = true after
+                        // taking the lock in their threads and will not add
+                        // continuations to the vector.
+                        std::unique_lock<mutex_type> l{mtx};
```
hkaiser, line 276: @msimberg I know that this code is not new. However, I'm still puzzled why this strange construct is needed. The `predecessor_done` is an atomic bool - shouldn't that be visible to other threads immediately?

msimberg: The lock is needed to avoid accessing the callbacks while another thread goes into this section: https://github.com/STEllAR-GROUP/hpx/blob/438db62fe90053d493e4709b09c56f192e3d342a/libs/parallelism/execution/include/hpx/execution/algorithms/split.hpp#L331-L346. The lock can earliest be released as it is right now and latest after checking if the vector of continuations is empty.

msimberg: @hkaiser do you still have concerns about this? I'm not sure how to explain it differently but if you do still have concerns it's probably easier to take it up at the next meeting.

msimberg:
retest cscs

libs/parallelism/execution/include/hpx/execution/algorithms/split.hpp
msimberg: @hkaiser let's discuss this separately if needed.

Issue #1485, opened 2015-04-29
`ignore_while_locked` doesn't support all Lockable types


sithhell:
The current implementation of `ignore_while_locked` only supports locks. This gets problematic when it is used in places where anything *Lockable* can be used, for example `condtion_variable::wait`. If a unsupported *Lockable* is passed we get a compile error. This change was introduced with b6d4e0ca5a62a66422419990de4fd319a6f50bb0. Before that we just silently ignored unsupported lockables leading to runtime failures when lock detection is turned on.

A suggestion to fix that situation would a default template that merely tries to have the passed ignored.

PR #5515, opened 2021-08-20
Avoid deadlock in ignore_while_locked_1485 test

msimberg:
Fixes deadlocks that appear every now and then on `ignore_while_locked_1485` (`notify` called before `wait` again). However, the test is either unnecessary or wrong and could possibly be removed? The description in the test claims that it's testing `ignore_while_locked` but doesn't actually use it. @hkaiser do you know if the test is even needed anymore?

### DIFF
--- a/libs/core/synchronization/tests/regressions/ignore_while_locked_1485.cpp
+++ b/libs/core/synchronization/tests/regressions/ignore_while_locked_1485.cpp
@@ -4,8 +4,10 @@
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-// This test case demonstrates the issue described in #1485:
-// `ignore_while_locked` doesn't support all Lockable types
+// This test case tests the workaround for the issue described in #1485:
+// `ignore_while_checking` doesn't support all Lockable types.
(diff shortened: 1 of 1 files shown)


hkaiser:
@msimberg it uses `ignore_all_while_checking` here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/synchronization/tests/regressions/ignore_while_locked_1485.cpp#L82.

Thanks for fixing the test.

hkaiser (dismissed):
LGTM, thanks!

msimberg:
> @msimberg it uses `ignore_all_while_checking` here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/synchronization/tests/regressions/ignore_while_locked_1485.cpp#L82.

Which is not the same as `ignore_while_locked`. The latter doesn't even exist anymore. Is it the same as `ignore_while_checking`?

hkaiser:
> Which is not the same as `ignore_while_locked`. The latter doesn't even exist anymore. Is it the same as `ignore_while_checking`?

Yes.

msimberg:
I've updated the description of the test in the file to better match my understanding of the situation. If I understand correctly #1485 was "fixed" by introducing `ignore_all_while_checking`, i.e. `ignore_while_checking` still doesn't support all lockable types, but `ignore_all_while_checking` can be used as a workaround, and that is why this test was using `ignore_all_while_checking`.

PR #5468, opened 2021-07-26
Experimentation to make the perftest report public

aurianer:
Comment on the PR with the performance report in case of a failure

Note: I will work on uploading the corresponding image of the report on CDash as it is not allowed through the Github API, only through manual drag and drop

PR #5516, opened 2021-08-20
Add comment script to jenkins pipeline + force execution

aurianer:
The report has to be uploaded even on failures.


aurianer:
Nope I was just testing things locally and I accidentally opened a PR against STE||AR-GROUP instead of my fork, the real important one is #5468 but not ready at all yet..

PR #5517, opened 2021-08-20
Fix coroutine integration

msimberg:
Haven't tested the actual build yet. `gcc-newest` should have coroutines enabled if this works as intended.


cmake/HPX_AddConfigTest.cmake
```diff
@@ -126,7 +127,9 @@ function(add_hpx_config_test variable)
 
     if(${variable}_EXECUTE)
       if(NOT CMAKE_CROSSCOMPILING)
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${additional_cmake_flags}")
+        set(CMAKE_CXX_FLAGS
+            "${CMAKE_CXX_FLAGS} ${additional_cmake_flags} ${CXXFLAGS}"
```
hkaiser, line 131:
```suggestion
            "${CMAKE_CXX_FLAGS} ${additional_cmake_flags} ${${variable}_CXXFLAGS}"
```

libs/parallelism/futures/include/hpx/futures/traits/detail/future_await_traits.hpp
```diff
@@ -343,6 +346,8 @@ namespace std {
                     this->base_type::set_value(std::forward<U>(value));
                 }
 
+                void unhandled_exception() const noexcept {}
+
```
hkaiser, line 350: Shouldn't this be added to the specialization for `hpx::lcos::future<void>` above and `hpx::lcos::shared_future<void>` below as well?

msimberg: Definitely. The tests also don't compile yet. This is just playing around. If you happen to know what's missing feel free to push here (though I know you might not have the time). I need significantly more time to understand the implementation as I've never looked at it properly before.

prlw1:
Why did you change the test condition which was correct in my patch? You appear to be adding the flag for all gcc, and not just version 10, even though you proved it wasn't necessary for version 11, and didn't exist in version 9.

msimberg:
> Why did you change the test condition which was correct in my patch? You appear to be adding the flag for all gcc, and not just version 10, even though you proved it wasn't necessary for version 11, and didn't exist in version 9.

Simply because it doesn't matter, but I can indeed make the condition more specific as well.

Spongman:
I believe the handlers should look like this:
```c++
                void unhandled_exception() const noexcept
                {
                    this->base_type::set_exception(std::current_exception());
                }
```

msimberg:
@Spongman indeed, thanks. I thought I mentioned this on some issue or PR, but I currently don't have the time to work on this. I'm wondering, would you be interested in finishing this off?

msimberg:
I'd appreciate reviews on this now. I've made a few more configuration changes.

GCC 10 "supports" coroutines with the `-fcoroutines` flag. However, coroutines like these:
```
shared_future<void> f() {
    co_return co_await make_ready_future<void>();
}
```
fail to compile with GCC 10 because the compiler attempts to construct a shared future with an lvalue reference of the future returned from `make_ready_future`. We could work around this in the test itself by explicitly making the future returned from `make_ready_future` a shared future, but I don't see much benefit in "fixing" it for GCC 10 when we anyway can't support this for all GCC versions. I.e. requiring GCC 11 seems to me like an acceptable solution, especially since it doesn't come with funny restrictions. Any objections?

The second change is related to GCC making C++20 features available even with `-std=c++17`. This came up while trying the changes here accidentally with the `-fcoroutines` flag and `HPX_WITH_CXX17`. The feature test would detect coroutines support but then the feature macro would not be defined since it's guarded by a `__cplusplus` check (https://github.com/STEllAR-GROUP/hpx/blob/35c7a2d752d7b6d25299198788ab8646b3609bed/cmake/HPX_AddDefinitions.cmake#L123-L138) and tests expecting coroutines would fail to compile. For coroutines this isn't relevant anymore since I removed the `-fcoroutines` fallback, but the same happens for simd support. I've removed the conditional definition of those feature test macros. It's a quirk of GCC, but I think it's reasonable to always define the feature test macros if the compiler seems to support the feature, even if it's with the wrong standard. The alternative is going back to not even checking for C++XX features if HPX_WITH_CXXXX isn't enabled.

msimberg:
Note: I have not tested the "non-standard" coroutines options that we have (emulation, clang's `-fcoroutines-ts`, MSVC's `-await`) and will not do so in this PR.

hkaiser:
> requiring GCC 11 seems to me like an acceptable solution, especially since it doesn't come with funny restrictions.

That's reasonable.

hkaiser:
> The alternative is going back to not even checking for C++XX features if HPX_WITH_CXXXX isn't enabled.

What's wrong with that? I think this would avoid funny mistakes.

hkaiser:
> Note: I have not tested the "non-standard" coroutines options that we have (emulation, clang's `-fcoroutines-ts`, MSVC's `-await`) and will not do so in this PR.

We might be able to completely get rid of those. Let's stick with enabling coroutines for HPX_WITH_CXX20 only.

msimberg:
> > The alternative is going back to not even checking for C++XX features if HPX_WITH_CXXXX isn't enabled.
> 
> What's wrong with that? I think this would avoid funny mistakes.

We had problems with it in the past which is why we enabled the feature checks unconditionally. However, I don't remember what the exact problem was. I'll try to dig a bit to see if I can find out what the problem was.



> > Note: I have not tested the "non-standard" coroutines options that we have (emulation, clang's `-fcoroutines-ts`, MSVC's `-await`) and will not do so in this PR.
> 
> We might be able to completely get rid of those. Let's stick with enabling coroutines for HPX_WITH_CXX20 only.

Yes, that sounds reasonable. I can do that here. Does MSVC also support coroutines without any special flags nowadays (except probably `/std:c++20` or whatever the equivalent is there)?

hkaiser:
> Yes, that sounds reasonable. I can do that here. Does MSVC also support coroutines without any special flags nowadays (except probably `/std:c++20` or whatever the equivalent is there)?

I would think so, yes.

msimberg:
> > > The alternative is going back to not even checking for C++XX features if HPX_WITH_CXXXX isn't enabled.
> > 
> > 
> > What's wrong with that? I think this would avoid funny mistakes.
> 
> We had problems with it in the past which is why we enabled the feature checks unconditionally. However, I don't remember what the exact problem was. I'll try to dig a bit to see if I can find out what the problem was.

I made them unconditional here: https://github.com/STEllAR-GROUP/hpx/pull/4378, but I wasn't smart enough to write why. I would prefer keeping them unconditional though, but I don't feel very strongly about this as long as I can't remember the original reason for making them unconditional.

msimberg:
> Aren't these two orthogonal? One is how we specify the C++ standard to rely on, the other is to enable/disable coroutines. Did I misunderstand something?

I fear that at least one of us is misunderstanding the other, if not both... I would call the current behaviour on master orthogonal, i.e. the feature checks do not care about which C++ standard is set (and definitely not vice-versa, but that was never the case and makes little sense). When I say "unconditional" I see that as the same as your "orthogonal". Does this match what you're saying?

Note: there are probably two separate "unconditional" involved here:
1. Doing the feature checks independently of the C++ standard set.
2. Defining the feature check macros without `__cplusplus` guards.
I take the second for granted if we do the first since things are definitely broken otherwise.

hkaiser:
> > Aren't these two orthogonal? One is how we specify the C++ standard to rely on, the other is to enable/disable coroutines. Did I misunderstand something?
> 
> I fear that at least one of us is misunderstanding the other, if not both... I would call the current behaviour on master orthogonal, i.e. the feature checks do not care about which C++ standard is set (and definitely not vice-versa, but that was never the case and makes little sense). When I say "unconditional" I see that as the same as your "orthogonal". Does this match what you're saying?

Possibly ;-)

But jokes apart - I think I agree with what you said.

> Note: there are probably two separate "unconditional" involved here:
> 
> 1. Doing the feature checks independently of the C++ standard set.
> 2. Defining the feature check macros without `__cplusplus` guards.
> 
> I take the second for granted if we do the first since things are definitely broken otherwise.

The second is important for depending projects that possibly use a different compiler (version). The guards allow for a minimum of safety if the compiler used for a depending project is too old. We've ran into these issues before, which triggered guarding the macros in the first place.

msimberg:
> The second is important for depending projects that possibly use a different compiler (version). The guards allow for a minimum of safety if the compiler used for a depending project is too old. We've ran into these issues before, which triggered guarding the macros in the first place.

What sort of safety does it provide? Compiling applications with a lower C++ version than HPX was compiled with is bound to be broken anyway (and enforced when building with CMake). Compiling applications with the same standard as HPX but a different compiler that doesn't define `__cplusplus` correctly is broken. Compiling applications with a different compiler with the same or higher standard but that doesn't support the feature that HPX detected was available seems weird. In what situations is that desireable? If one compiler was available to compile HPX, why would one not use that, or a newer compiler, to compile the application?

hkaiser:
> What sort of safety does it provide? Compiling applications with a lower C++ version than HPX was compiled with is bound to be broken anyway (and enforced when building with CMake). Compiling applications with the same standard as HPX but a different compiler that doesn't define `__cplusplus` correctly is broken. Compiling applications with a different compiler with the same or higher standard but that doesn't support the feature that HPX detected was available seems weird. In what situations is that desireable? If one compiler was available to compile HPX, why would one not use that, or a newer compiler, to compile the application?

It might have been caused by compilers that claim to be C++yy conforming but in fact were not. Otherwise I agree that this shouldn't happen - however it did happen multiple times. Otherwise we wouldn't have added the guards...

msimberg:
> It might have been caused by compilers that claim to be C++yy conforming but in fact were not. Otherwise I agree that this shouldn't happen - however it did happen multiple times. Otherwise we wouldn't have added the guards...

Hrm, so we've had problems with both solutions. And we can't remember what the problems were...

Fair enough, let's go with conditional defines and conditional feature checks again then. We'll find out sooner or later what the problem was, if any. And like you said, it's reasonable to not allow C++20 features with C++17. That's just a GCC quirk after all.

msimberg:
Let's see if I broke something with the latest change. I reverted back to conditionally defining the macros, and made the feature checks conditional on the C++ standard set in the CMake configuration.

hkaiser:
> > It might have been caused by compilers that claim to be C++yy conforming but in fact were not. Otherwise I agree that this shouldn't happen - however it did happen multiple times. Otherwise we wouldn't have added the guards...
> 
> Hrm, so we've had problems with both solutions. And we can't remember what the problems were...
> 
> Fair enough, let's go with conditional defines and conditional feature checks again then. We'll find out sooner or later what the problem was, if any. And like you said, it's reasonable to not allow C++20 features with C++17. That's just a GCC quirk after all.

I think this situation occurred for instance when HPX was built with C++17, but the consuming project was C++14 only (or similar).

msimberg:
Rebased again. This should be ready to go unless there are comments on the changes to the `coroutine_traits`.

CMakeLists.txt
```diff
@@ -1651,10 +1616,10 @@ if(WIN32)
...
 
     # /RTC1 is incompatible with /await
-    if((NOT HPX_WITH_AWAIT) AND (NOT HPX_WITH_CXX20_COROUTINES))
-      hpx_add_target_compile_option(/RTC1 PUBLIC CONFIGURATIONS Debug)
-    else()
+    if(HPX_WITH_CXX20_COROUTINES)
       hpx_remove_target_compile_option(/RTC1 PUBLIC CONFIGURATIONS Debug)
+    else()
+      hpx_add_target_compile_option(/RTC1 PUBLIC CONFIGURATIONS Debug)
     endif()
```
hkaiser, line 1623: I will verify if this is still needed for the C++20 coroutines. I think this was a workaround for a limitation of the pre-C++20 implementation in MSVC.

hkaiser (dismissed):
Otherwise this LGTM, thanks!

CMakeLists.txt
msimberg: Thanks!

aurianer (dismissed):
LGTM thanks!

CMakeLists.txt
msimberg: Thanks for checking! I can remove them. However, should the `/RTC1` flag still be added, just not conditionally? Because it is added if coroutines aren't available. Or is there no need for the flag at all anymore?

hkaiser: The flag is automatically added by CMake, we don't have to do anything. Originally we have had it in here to make sure the flag was re-added whenever the await support was disabled (without deleting the CMake cache).

msimberg: Ah, gotcha. I've removed it all.

msimberg:
bors merge

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

msimberg:
Following up on the question about if the C++ standard feature checks should be conditional on the standard that is set or not, the answer is: sort of. The changes introduced in this PR are problematic when reconfiguring with a different standard since the variables may stay `ON` when going down to e.g. 17. Always checking for the features works around that problem. However, in https://github.com/STEllAR-GROUP/hpx-local/pull/2 I've changed the behaviour to allow the feature check to depend on a standard, and if the set standard is too low it will explicitly unset the corresponding variable. See here for the check: https://github.com/STEllAR-GROUP/hpx-local/blob/7acbdc78c6898ed9ec149feea4530708d1bbff54/cmake/HPXLocal_AddConfigTest.cmake#L34-L44.

PR #5520, opened 2021-08-23
Update reference as perftests failing regularly

aurianer:


msimberg (approved):
Thanks @aurianer.

For reference (heh) @aurianer reran the benchmark on the commit that she used for the original reference, and it shows the same 1-3% performance drop that we're seeing regularly on PRs, indicating that it's something outside of our control that has changed on Piz Daint.

aurianer:
Update:
Misunderstanding (probably because of me) the performance drop of 4% were not there on the first commit merging the perftests. I opened this PR to stop the perftests from failing on the upcoming PRs but still plan on doing a bisect to determine where the performance drop comes from.
```
master (254db57ab5e998619e49744062d14609f1b466da):
mean time    0.169334
min time    0.160787
max time    0.214315

merge commit of the perftests (0faa1f473a141a166dc6e611b8efaa52d7fa5431):
mean time    0.162544
min time    0.157797
max time    0.210743
```

PR #5521, opened 2021-08-23
Minor changes that improve performance

hkaiser:
- flyby: fix minor inconsistency while thread stealing
- flyby: fixing warnings on Windows

This separates some minor improvements from #5467

### DIFF
--- a/libs/full/runtime_configuration/src/runtime_configuration.cpp
+++ b/libs/full/runtime_configuration/src/runtime_configuration.cpp
@@ -695,6 +695,7 @@ namespace hpx { namespace util {
         char const* argv0_, runtime_mode mode)
       : mode_(mode)
       , num_localities(0)
+      , num_os_threads(0)
       , small_stacksize(HPX_SMALL_STACK_SIZE)
       , medium_stacksize(HPX_MEDIUM_STACK_SIZE)
       , large_stacksize(HPX_LARGE_STACK_SIZE)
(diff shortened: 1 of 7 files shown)


libs/full/runtime_configuration/src/runtime_configuration.cpp
```diff
@@ -1299,7 +1274,7 @@ namespace hpx { namespace util {
             return huge_stacksize;
 
         case threads::thread_stacksize::nostack:
-            return (std::numeric_limits<std::ptrdiff_t>::max)();
+            return (std::numeric_limits<std::ptrdiff_t>::max) ();
```
msimberg, line 1277: This needs to be reverted.

msimberg (changes requested):
Nice, thanks @hkaiser! Looks good to me except the one formatting change that should be reverted.

Is that due clang-format 12 this time...? Or manually applied?

hkaiser:
> Is that due clang-format 12 this time...? Or manually applied?

Yep, it's clang-format v12. I'll revert this.

libs/full/runtime_configuration/src/runtime_configuration.cpp
```diff
@@ -753,40 +754,34 @@ namespace hpx { namespace util {
 
     std::size_t runtime_configuration::get_ipc_data_buffer_cache_size() const
     {
-        if (has_section("hpx.parcel"))
+        util::section const* sec = get_section("hpx.parcel.ipc");
+        if (nullptr != sec)
```
msimberg, line 758: Purely subjective, and I don't mind to keep it like this, but what about:
```suggestion
        if (util::section const* sec = get_section("hpx.parcel.ipc")
```
If you prefer this I don't mind doing the change.

aurianer (dismissed):
LGTM thanks!

libs/full/runtime_configuration/src/runtime_configuration.cpp
hkaiser: I can do that.

msimberg:
@hkaiser same thing here... I've only applied a formatting change on the one commit. I think this should be ready to be merged unless any surprises come up from CI.

hkaiser:
Thanks!

PR #5525, opened 2021-08-28
Fix Building HPX in Quick start

dimitraka:
Change instructions for building HPX from paragraphs to structured information (numbered and bulleted lists, code blocks etc.)

[skip ci]

### DIFF
--- a/docs/sphinx/quickstart.rst
+++ b/docs/sphinx/quickstart.rst
@@ -32,8 +32,9 @@ unpack it or clone the repository directly using ``git``:
 It is also recommended that you check out the latest stable tag:
 
 .. code-block:: sh
-
-    git checkout 1.7.1
+    
+    cd hpx
+    git checkout 1.7.1 
 
 |hpx| dependencies
 ==================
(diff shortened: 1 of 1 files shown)


hkaiser:
add to whitelist

hkaiser (approved):
Very nice! LGTM. Thanks!

docs/sphinx/quickstart.rst
```diff
@@ -28,12 +28,13 @@ unpack it or clone the repository directly using ``git``:
...
 
     git clone https://github.com/STEllAR-GROUP/hpx.git
+    cd hpx
 
 It is also recommended that you check out the latest stable tag:
 
 .. code-block:: sh
 
-    git checkout 1.7.1
+    git checkout branch_name  # 1.7.0, 1.7.1, etc.
```
msimberg, line 37: My preference would be to keep `1.7.1` explicitly in the command. I've seen people literally copy-paste commands and wonder why `git checkout branch_name` doesn't work. It's an extreme case, but I think there's no need to make things more difficult if we can actually have a command that one can copy-paste. What do you think?

docs/sphinx/quickstart.rst
```diff
@@ -53,59 +54,107 @@ versions, is available at :ref:`prerequisites`.
...
-    mkdir build && cd build
-    cmake -DCMAKE_INSTALL_PREFIX=/install/path ..
-    make install
+  .. code-block:: sh
 
-This will build the core |hpx| libraries and examples, and install them to your
-chosen location. If you want to install |hpx| to system folders, simply leave out
-the ``CMAKE_INSTALL_PREFIX`` option. This may take a while. To speed up the
-process, launch more jobs by passing the ``-jN`` option to ``make``.
+      cd ..
```
msimberg, line 64: `build` can actually also be inside the source directory, and we even ignore it in `.gitignore`. I for example have a `~/src/` directory with all my cloned repos, and if I put `build` inside the source directory for all of those I can just call them all `build`. If I go up to the source directory I'd have to start calling them `hpx-build` etc. Do you think it's clearer to have the build directory completely outside the source directory? It might also be a good idea to just add a (bash) comment at the top of the code block saying in which directory you assume the user is in at the moment?

docs/sphinx/quickstart.rst
```diff
@@ -53,59 +54,107 @@ versions, is available at :ref:`prerequisites`.
...
+        cmake -DCMAKE_INSTALL_PREFIX=/install/path ..
+
+    .. tip::
+
+       If you want to change |cmake| variables for your build, it is usually a good
+       idea to start with a clean build directory to avoid configuration problems.
+       It is especially important that you use a clean build directory when changing
+       between ``Release`` and ``Debug`` modes.
+
+  * To install |hpx| to system folders, simply leave out the ``CMAKE_INSTALL_PREFIX`` option:
```
msimberg, line 83:
```suggestion
  * To install |hpx| to the default system folders, simply leave out the ``CMAKE_INSTALL_PREFIX`` option:
```
?

docs/sphinx/quickstart.rst
```diff
@@ -53,59 +54,107 @@ versions, is available at :ref:`prerequisites`.
...
-   It is especially important that you use a clean build directory when changing
-   between ``Release`` and ``Debug`` modes.
+#. Once |cmake| is complete, to build the project you run:
 
-If your dependencies are in custom locations, you may need to tell |cmake| where
-to find them by passing one or more of the following options to |cmake|:
+  .. code-block:: sh
+
+      make -jN # where N is the number of jobs
+      make install -jN # where N is the number of jobs
```
msimberg, line 117:
```suggestion
      make install
```
simply because there's no benefit to running this in parallel (it's a single job that doesn't run in parallel).

docs/sphinx/quickstart.rst
```diff
@@ -53,59 +54,107 @@ versions, is available at :ref:`prerequisites`.
...
+
+* To run single tests, for example a test for ``for_loop``:
 
 .. code-block:: sh
 
-    -DBOOST_ROOT=/path/to/boost
-    -DHWLOC_ROOT=/path/to/hwloc
-    -DTCMALLOC_ROOT=/path/to/tcmalloc
-    -DJEMALLOC_ROOT=/path/to/jemalloc
+    ctest --output-on-failure -R tests.unit.parallel.algorithms.for_loop
```
msimberg, line 148: Another one of those things that are outdated...
```suggestion
    ctest --output-on-failure -R tests.unit.modules.algorithms.for_loop
```

docs/sphinx/quickstart.rst
```diff
@@ -53,59 +54,107 @@ versions, is available at :ref:`prerequisites`.
...
+
+    make tests
+
+To run the tests:
+
+.. code-block:: sh
+
+    make test
+
+To control over which tests to run use ``ctest``:
```
msimberg, line 142:
```suggestion
To control which tests to run use ``ctest``:
```

docs/sphinx/quickstart.rst
msimberg: Also, should the `cd hpx` command move into this code block instead?

docs/sphinx/quickstart.rst
```diff
@@ -53,59 +54,107 @@ versions, is available at :ref:`prerequisites`.
...
+      cd ..
+      mkdir build && cd build
 
-.. tip::
+#. To configure the project you have the following options:
 
-   Do not set only ``-j`` (i.e. ``-j`` without an explicit number of jobs)
-   unless you have a lot of memory available on your machine.
+  * To build the core |hpx| libraries and examples, and install them to your
+  chosen location (suggested):
```
msimberg, line 70:
```suggestion
  chosen location (recommended):
```
? I may just be missing what `suggested` is referring to.

msimberg (changes requested):

docs/sphinx/quickstart.rst
dimitraka: Sure, we can do it this way!

docs/sphinx/quickstart.rst
dimitraka: I wanted to mention that this step is the recommended one. Do you think I should remove it because it's not clear?

docs/sphinx/quickstart.rst
dimitraka: Which do you think is the most proper way to do it? Having `build` inside or outside the source directory? Since there is always another way to do it, I think we can chose the optimal way for the user, so that they would just copy-paste the commands without wondering which directory is the proper one. What do you think?

docs/sphinx/quickstart.rst
msimberg: Nope, you can leave it. I would just change the word to `recommended` rather than `suggested`.

docs/sphinx/quickstart.rst
msimberg: I think having a `build` directory inside the source directory is a very common convention, but obviously not universal. At the very least having `build` inside the source directory is pretty much guaranteed to not conflict with a `build` directory that the user may have created in the parent directory for other purposes. I'd keep it inside the source directory.

docs/sphinx/quickstart.rst
aurianer: or `make -jN install` which does both in one step?

docs/sphinx/quickstart.rst
```diff
@@ -53,59 +54,107 @@ versions, is available at :ref:`prerequisites`.
...
+
+       If you are building |hpx| for a system with more than 64 processing units,
+       you must change the |cmake| variable ``HPX_WITH_MAX_CPU_COUNT`` (to a value at least as big as the
+       number of (virtual) cores on your system).
 
-   If you want to change |cmake| variables for your build, it is usually a good
-   idea to start with a clean build directory to avoid configuration problems.
-   It is especially important that you use a clean build directory when changing
-   between ``Release`` and ``Debug`` modes.
+#. Once |cmake| is complete, to build the project you run:
```
aurianer, line 112: `Once the |cmake| step is complete` or `Once the configuration is complete`? As cmake is a tool

docs/sphinx/quickstart.rst
```diff
@@ -53,59 +54,107 @@ versions, is available at :ref:`prerequisites`.
...
+  ``-DHPX_WITH_MALLOC=system`` to |cmake|:
+
+    .. code-block:: sh 
+
+        cmake -DCMAKE_INSTALL_PREFIX=/install/path -DHPX_WITH_MALLOC=system ..
+
+    .. important::
+
+       If you are building |hpx| for a system with more than 64 processing units,
+       you must change the |cmake| variable ``HPX_WITH_MAX_CPU_COUNT`` (to a value at least as big as the
```
aurianer, line 109: Maybe add that the default is 64?

aurianer (approved):
LGTM thanks a lot!

docs/sphinx/quickstart.rst
dimitraka: Done!

msimberg (approved):
Looks good, thanks @dimitraka for the updates!

PR #5526, opened 2021-08-30
Update Rostam HIP configuration to use 4.3.0

msimberg:


msimberg:
Thanks @akheir for the smooth upgrade on rostam!

PR #5530, opened 2021-08-30
Eagerly initialize a configurable number of threads on scheduler/thread queue init

msimberg:
This adds a configuration option `hpx.thread_queue.init_threads_count` (default `10`) which controls how many threads to allocate and initialize (including allocating a stack) with the default stack size in each `thread_queue`. The thread queue also first attempts to clean up terminated threads when it runs out of recycled threads to reuse existing threads if possible. Only if it can't reuse an old thread will it actually allocate a new thread. Checking for this is cheap, and we are anyway holding the thread queue lock at that point.

This reduces the need to "warm up" the scheduler and thread queues when benchmarking (see benchmarks below), and I don't see the harm in doing this generally either since threads will need to be allocated sooner or later.

To achieve this, I've added a way to create `thread_data` with a reference count of zero so that it can be added to the "heaps" of recycled threads immediately without having to go through `terminated_threads`. I've also added a way to eagerly initialize a `thread_data`, which in practice means allocating a stack. `on_start_thread` creates the configured number of threads, initializes them, and adds them to the default stack size heap.

Stream benchmark before these changes (note the very irregular lines for all executors except the fork-join executor (which is not affected by these changes since it starts its threads up front and then reuses them)):
stream_lazy_thread_init

Stream after these changes (the dips are mostly missing in these results):
stream_eager_thread_init

The dips also mostly disappear with enough warmup iterations of the stream benchmark, or by increasing the number of iterations. The benchmark defaults to 10 iterations, and the plots show the best result from those 10 iterations. With the changes in this PR the best performance is easily achieved within the default 10 iterations.

An open question is whether 10 is a good default, and if the eager initialization should be more granular (per stacksize/priority), but I figured this was a reasonable number for most systems.

### DIFF
--- a/libs/core/coroutines/include/hpx/coroutines/detail/context_base.hpp
+++ b/libs/core/coroutines/include/hpx/coroutines/detail/context_base.hpp
@@ -68,12 +68,14 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
     template <typename CoroutineImpl>
     class context_base : public default_context_impl<CoroutineImpl>
     {
+        using base_type = default_context_impl<CoroutineImpl>;
+
     public:
         using deleter_type = void(context_base const*);
         using thread_id_type = hpx::threads::thread_id;
 
         context_base(std::ptrdiff_t stack_size, thread_id_type id)
-          : default_context_impl<CoroutineImpl>(stack_size)
+          : base_type(stack_size)
           , m_caller()
           , m_state(ctx_ready)
           , m_exit_state(ctx_exit_not_requested)
@@ -149,6 +151,11 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
             return m_state == ctx_exited;
         }
 
+        void init()
+        {
+            base_type::init();
+        }
+
         // Resume coroutine.
         // Pre:  The coroutine must be ready.
         // Post: The coroutine relinquished control. It might be ready
@@ -159,7 +166,7 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
         // on return.
         void invoke()
         {
-            this->init();
+            base_type::init();
             HPX_ASSERT(is_ready());
             do_invoke();
 
@@ -208,7 +215,7 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
         ~context_base() noexcept
         {
             HPX_ASSERT(!running());
-            HPX_ASSERT(exited());
+            HPX_ASSERT(exited() || (is_ready() && m_phase == 0));
             m_thread_id.reset();
 #if defined(HPX_HAVE_THREAD_LOCAL_STORAGE)
             delete_tss_storage(m_thread_data);
@@ -321,7 +328,7 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
 
         void rebind_base(thread_id_type id)
         {
-            HPX_ASSERT(exited());
+            HPX_ASSERT(!running());
 
             m_thread_id = id;
             m_state = ctx_ready;
@@ -387,8 +394,7 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
 #endif
         }
 
-        typedef typename default_context_impl<CoroutineImpl>::context_impl_base
-            ctx_type;
+        using ctx_type = typename base_type::context_impl_base;
         ctx_type m_caller;
 
         context_state m_state;
--- a/libs/core/coroutines/include/hpx/coroutines/thread_id_type.hpp
+++ b/libs/core/coroutines/include/hpx/coroutines/thread_id_type.hpp
@@ -10,6 +10,7 @@
 #pragma once
 
 #include <hpx/config.hpp>
+#include <hpx/assert.hpp>
 #include <hpx/modules/format.hpp>
 #include <hpx/modules/memory.hpp>
 #include <hpx/thread_support/atomic_count.hpp>
@@ -156,6 +157,12 @@ namespace hpx { namespace threads {
     };
 
     ///////////////////////////////////////////////////////////////////////////
+    enum class thread_id_addref
+    {
+        yes,
+        no
+    };
+
     namespace detail {
 
         struct thread_data_reference_counting;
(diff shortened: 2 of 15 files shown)


libs/core/coroutines/include/hpx/coroutines/detail/context_base.hpp
```diff
@@ -149,6 +151,11 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
             return m_state == ctx_exited;
         }
 
+        void init()
+        {
+            base_type::init();
```
hkaiser, line 156: Don't we still need a `this->`, this type is a template after all?

libs/core/coroutines/include/hpx/coroutines/detail/context_base.hpp
```diff
@@ -208,7 +215,7 @@ namespace hpx { namespace threads { namespace coroutines { namespace detail {
         ~context_base() noexcept
         {
             HPX_ASSERT(!running());
-            HPX_ASSERT(exited());
+            HPX_ASSERT(exited() || m_phase == 0);
```
hkaiser, line 218: Can you explain the rationale of this change, please?

libs/core/coroutines/include/hpx/coroutines/thread_id_type.hpp
```diff
@@ -165,11 +171,12 @@ namespace hpx { namespace threads {
...
-            // will be held alive by the variable returned from the creation
-            // function
-            thread_data_reference_counting()
-              : count_(1)
+            // the initial reference count is one by default as each newly
+            // created thread will be held alive by the variable returned from
+            // the creation function;
+            explicit thread_data_reference_counting(
+                thread_id_addref addref = thread_id_addref::yes)
+              : count_(addref == thread_id_addref::yes)
```
hkaiser, line 179: Excellent catch, thanks!

hkaiser (dismissed):
Excellent thinking! LGTM, thanks!

libs/core/coroutines/include/hpx/coroutines/detail/context_base.hpp
msimberg: Apparently not.

libs/core/coroutines/include/hpx/coroutines/detail/context_base.hpp
msimberg: This and the other changes to the assertions in this file were an attempt to loosen the pre- and post-conditions enough that we can eagerly initialize the stack on a thread that hasn't been used yet, but still keep most of the sanity checks that we have in place.

This particular change is because with the eagerly initialized threads it's possible to have a thread that is initialized but never used. In that case the thread would have `m_state = `ctx_ready` when the thread is destructed, which is fine. A more accurate assertion would probably be `HPX_ASSERT(exited() || (ready() && m_phase == 0))`, i.e. at thread destruction a thread should:
- not be running, and
- have run and exited, or
- be ready but not have run at all (i.e. `m_phase == 0`)

msimberg:
The last commit fixes a "bug" that I've seen come up every now and then on the `thread_pool_scheduler` test on master and now consistently on the same and `thread_stacksize_current_test`. The test would sometimes expect that a thread had a medium stack size, but would instead get a small stack size. This is not a clear bug because in debug mode on linux small and medium stack sizes are the same. However, it's nice if a thread reports the correct logical size that was used when creating it. The last commit explicitly sets the logical stack size again when rebinding a thread (which can be different than before) but still requires that the physical stack size is the same as before.

msimberg:
launch perftests

msimberg:
I've rebased and pushed one more time to retrigger the perf tests. If They come back ok I'm going to assume the drop in that first perf test was unrelated to these changes.

libs/core/coroutines/include/hpx/coroutines/detail/context_base.hpp
msimberg: I've updated the assertion.

msimberg:
I'll retest this a few times. The `thread_pool_scheduler` test has timed out worryingly on a few of the configurations and I want to see if it's due to the changes here.

msimberg:
launch perftests

msimberg:
retest cscs

msimberg:
The `thread_pool_scheduler` test timeouts were an existing problem, but I think the eager creation of stacks made it more likely to happen (the particular test that was hanging is one of the earlier tests in `thread_pool_scheduler_test` where it most likely can't reuse old threads and thread creation is slower). See https://github.com/STEllAR-GROUP/hpx/pull/5530/commits/8e766eabb373c048a381c58642b6d58ff26dcd47 for details.

The more eager reuse of threads also revealed that checks like these are not absolute: https://github.com/msimberg/hpx/blob/8e766eabb373c048a381c58642b6d58ff26dcd47/libs/core/executors/tests/unit/thread_pool_scheduler.cpp#L378. That's because a new thread started by `on` may reuse the thread by the previous `on` context. I'm leaving the test in there for now knowing that there's a small chance of false positives, but in principle that test could be removed.

With that I think this is ready to go. The performance report is something I want to keep an eye on after merging. I suspect that the eager recycling may actually reduce throughput slightly. I think the eager thread creation at startup should never negatively affect performance.

msimberg:
@hkaiser unrelated to this PR, but do you know what is going on here: https://github.com/STEllAR-GROUP/hpx/pull/5530/checks?check_run_id=3688557988?

hkaiser:
> @hkaiser unrelated to this PR, but do you know what is going on here: https://github.com/STEllAR-GROUP/hpx/pull/5530/checks?check_run_id=3688557988?

This is probably some failure while downloading the vcpkg binaries caused by problems on rostam. Retrying should make it work again...

msimberg:
launch perftests

msimberg:
I've dropped the commit which eagerly recycles threads. The two perftest runs since then have had the same performance as the reference or even better than the reference, so it seems like this is the right choice. Thanks @aurianer again for pointing this out and for working on the perf testing infrastructure!

PR #5529, opened 2021-08-30
Update benchmarks and add support for scheduler_executor

msimberg:
- Add support for `scheduler_executor<thread_pool_scheduler>` to the stream and future overhead benchmarks.
- Add default constructor (and copy/move constructors/assignment operators) to `scheduler_executor`.
- Add support for setting warmup iterations in stream benchmark.
- Allow using `scheduler_executor` with forward iterators.

PR #5531, opened 2021-08-30
Use contiguous_index_queue in thread_pool_scheduler

msimberg:
Uses `contiguous_index_queue` as in `fork_join_executor`, with the addition that `scheduler_executor` also does its own chunking. This means that only one HPX thread is spawned per worker thread, instead of 4-8 per worker thread. The load balancing is instead done by stealing from the `contiguous_index_queue`s of the other worker threads.

Still a draft, as I will rebase on #5530 and #5529, but initial benchmarks show consistently better performance than the hierarchical spawning in `parallel_executor`. The changes to the stream benchmarks will be replaced by those in #5529.


msimberg:
Rebased on #5530 and cleaned up. This should now be ready to go. Stream benchmark comparing `scheduler_executor<thread_pool_scheduler>` before (`scheduler_executor_old`) and after (`scheduler_executor`) to the other executors:

stream-scheduler-executor-new

aurianer (dismissed):
LGTM thanks!

PR #5533, opened 2021-09-01
Fix `annotate_function` for `std::string`

msimberg:
`annotate_function` did not have a constructor taking `std::string`s which means that passing a `std::string` to it would try to use `get_function_annotation` for getting an annotation from a `std::string`. Since there's no specialization, this returns `<unknown>`. This adds an explicit constructor taking a `std::string`.

This also updates the `annotation_check` test and adds a test for `annotate_function`.

Flyby: restore the thread description correctly in the `annotate_function` destructor. @hkaiser after https://github.com/STEllAR-GROUP/hpx/pull/5401/commits/21f87b24ba8ecef89ac89807b1ecb99ca7fb36e1#diff-9d331166da570e7469e476d468b684dfd42590305a752f337fd6f9aa3f3bff86 there was no description stored in the `annotate_function` constructors, meaning that it was always restored to a default constructed thread description. Was this an oversight or intentional?

@albestro would you mind trying this?

### DIFF
--- a/libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
+++ b/libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
@@ -30,6 +30,10 @@
 #include <utility>
 
 namespace hpx { namespace util {
+    namespace detail {
+        HPX_CORE_EXPORT char const* store_function_annotation(std::string name);
+    }    // namespace detail
+
 #if defined(HPX_HAVE_THREAD_DESCRIPTION)
     ///////////////////////////////////////////////////////////////////////////
 #if defined(HPX_COMPUTE_DEVICE_CODE)
@@ -78,8 +82,8 @@ namespace hpx { namespace util {
             auto* self = hpx::threads::get_self_ptr();
             if (self != nullptr)
             {
-                threads::get_thread_id_data(self->get_thread_id())
-                    ->set_description(name);
+                desc_ = threads::get_thread_id_data(self->get_thread_id())
+                            ->set_description(name);
             }
 
 #if defined(HPX_HAVE_APEX)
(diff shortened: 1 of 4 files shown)


msimberg:
The test says that this is fixed. However, the annotations don't work in DLA-Future where @albestro noticed the problem. This is back to a draft until I understand if the fix here is unsufficient or if there's some other unrelated problem there.

albestro:
@msimberg just tested the branch and it works as expected with a simple example I wrote.

It does not solve the other problem (missing annotations), but you already know about it and it is up to you evaluating if it is worth waiting for seeing what's going on on that side too or not. 😉

libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
```diff
@@ -89,13 +93,33 @@ namespace hpx { namespace util {
 #endif
         }
 
-        template <typename F>
+        explicit annotate_function(std::string const& name)
```
hkaiser, line 96: May I suggest to either provide two overloads (`const&` and `&&`) or simply take the argument by value? This may allow to reduce the number of allocations.

libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
```diff
@@ -78,8 +82,8 @@ namespace hpx { namespace util {
             auto* self = hpx::threads::get_self_ptr();
             if (self != nullptr)
             {
-                threads::get_thread_id_data(self->get_thread_id())
-                    ->set_description(name);
+                desc_ = threads::get_thread_id_data(self->get_thread_id())
+                            ->set_description(name);
```
hkaiser, line 86: Thanks for this fix.

libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp
msimberg: Yeah, good idea.

msimberg:
All is good here. The missing annotations were due to something completely different...

hkaiser (dismissed):
LGTM, thanks!

aurianer (approved):
LGTM thanks!

aurianer:
Just a `cmake-format` error to fix

PR #5536, opened 2021-09-02
This fixes referencing stale iterators during the execution of binary mismatch

hkaiser:
- flyby: modernize the implementation of mismatch
- flyby: modernize `advance_to_sentinel`, `algorithm_result`

for more details, see https://github.com/STEllAR-GROUP/hpx/pull/5410#issuecomment-911790179

@Jedi18 please verify whether this fixes the issues you were seeing.

### DIFF
--- a/libs/core/iterator_support/include/hpx/iterator_support/traits/is_sentinel_for.hpp
+++ b/libs/core/iterator_support/include/hpx/iterator_support/traits/is_sentinel_for.hpp
@@ -60,4 +60,8 @@ namespace hpx { namespace traits {
     {
     };
 
+    template <typename Sent, typename Iter>
+    HPX_INLINE_CONSTEXPR_VARIABLE bool is_sized_sentinel_for_v =
+        is_sized_sentinel_for<Sent, Iter>::value;
+
 }}    // namespace hpx::traits
--- a/libs/parallelism/algorithms/include/hpx/parallel/algorithms/adjacent_difference.hpp
+++ b/libs/parallelism/algorithms/include/hpx/parallel/algorithms/adjacent_difference.hpp
@@ -89,9 +89,12 @@ namespace hpx { namespace parallel { inline namespace v1 {
                         });
                 };
 
-                auto f2 =
-                    [dest, count](
-                        std::vector<hpx::future<void>>&&) mutable -> FwdIter2 {
+                auto f2 = [dest, count](
+                              std::vector<hpx::future<void>>&& data) mutable
+                    -> FwdIter2 {
+                    // make sure iterators embedded in function object that is
+                    // attached to futures are invalidated
+                    data.clear();
                     std::advance(dest, count);
                     return dest;
                 };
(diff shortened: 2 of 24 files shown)


hkaiser:
@Jedi18 while working on this I realized that the non-binary version of `mismatch` has not been adapted to sentinels yet. Would you be able to look into that?

Jedi18:
@hkaiser Yes it does fix it, thanks a lot! Oh ok sure, I'll adapt mismatch to sentinels.

msimberg:
@hkaiser could you please describe the problem in more detail and why this fixes it? Can we do something more with our tests to catch similar problems in the future?

hkaiser:
> @hkaiser could you please describe the problem in more detail and why this fixes it? Can we do something more with our tests to catch similar problems in the future?

Certainly, sorry for being skimpy with information. Debug builds with MSVC use a special version of the standard library that tracks iterator validity. We were seeing debug failures from this iterator validation during destruction of the function objects that are executed by the threads that make ready the futures that represent the chunk execution. At the point of destruction of said function objects the iterators that are being captured by the lambdas went dangling as the sequences they referred to did already go out of scope.

My fix makes ensure that the futures are released earlier, which releases the iterators that are bound to the associated function objects (thread functions).

Please note that I have added a second commit to this PR that applies a similar fix to all other algorithms, where needed.

msimberg (approved):

msimberg:
This is unhappy for some reason: https://cdash.cscs.ch/viewBuildError.php?buildid=178740. Not sure why...

msimberg:
> > This is unhappy for some reason: https://cdash.cscs.ch/viewBuildError.php?buildid=178740. Not sure why...
> 
> I think I've encountered this before, this happens when using HPX_INVOKE instead of hpx::util::invoke with projections for some reason.

Did you find a workaround when you saw this before?

hkaiser:
> > > This is unhappy for some reason: https://cdash.cscs.ch/viewBuildError.php?buildid=178740. Not sure why...
> > 
> > 
> > I think I've encountered this before, this happens when using HPX_INVOKE instead of hpx::util::invoke with projections for some reason.
> 
> Did you find a workaround when you saw this before?

I now have changed the capture mode for that lambda, let's see if it helps. This is definitely a compiler problem.

Jedi18:
> > > This is unhappy for some reason: https://cdash.cscs.ch/viewBuildError.php?buildid=178740. Not sure why...
> > 
> > 
> > I think I've encountered this before, this happens when using HPX_INVOKE instead of hpx::util::invoke with projections for some reason.
> 
> Did you find a workaround when you saw this before?

Nope, I just changed it back to hpx::util::invoke

hkaiser:
> > > > This is unhappy for some reason: https://cdash.cscs.ch/viewBuildError.php?buildid=178740. Not sure why...
> > > 
> > > 
> > > I think I've encountered this before, this happens when using HPX_INVOKE instead of hpx::util::invoke with projections for some reason.
> > 
> > 
> > Did you find a workaround when you saw this before?
> 
> I now have changed the capture mode for that lambda, let's see if it helps. This is definitely a compiler problem.

Looks like changing the capture mode fixed this.

PR #5540, opened 2021-09-06
Adapt partition, partition_copy and stable_partition to C++ 20

Jedi18:
Adapted partition, partition_copy and stable_partition to C++ 20 and added container overloads. (range and sentinel). Added container tests for stable_partition.

Issue #4822
Issue #3646
Issue #1668


libs/parallelism/algorithms/include/hpx/parallel/algorithms/partition.hpp
```diff
@@ -136,21 +585,22 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
-                typename Proj>
+            template <typename ExPolicy, typename BidirIter, typename Sent,
+                typename F, typename Proj>
             static BidirIter sequential(
-                ExPolicy&&, BidirIter first, BidirIter last, F&& f, Proj&& proj)
+                ExPolicy&&, BidirIter first, Sent last, F&& f, Proj&& proj)
             {
-                return std::stable_partition(first, last,
+                auto last_iter = detail::advance_to_sentinel(first, last);
+                return std::stable_partition(first, last_iter,
```
hkaiser, line 594: This is a simple, but sub-optimal solution. For non-random-access iterators it requires to go over the sequence twice.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/partition.hpp
```diff
@@ -159,14 +609,15 @@ namespace hpx { namespace parallel { inline namespace v1 {
                     typename std::iterator_traits<RandIter>::difference_type;
 
                 future<RandIter> result;
+                auto last_iter = detail::advance_to_sentinel(first, last);
 
                 try
                 {
-                    difference_type size = std::distance(first, last);
+                    difference_type size = detail::distance(first, last);
```
hkaiser, line 616: We talked about this for other algorithms. We should probably create a facility that does both in one loop: advancing the iterator _and_ returning the distance.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/partition.hpp
```diff
@@ -422,8 +818,8 @@ namespace hpx { namespace parallel { inline namespace v1 {
             // block manager for random access iterator.
             template <typename RandIter>
             class block_manager<RandIter,
-                typename std::enable_if<hpx::traits::is_random_access_iterator<
-                    RandIter>::value>::type>
+                typename std::enable_if_t<
```
hkaiser, line 821: The `typename` is superflous now.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/partition.hpp
```diff
@@ -497,10 +893,9 @@ namespace hpx { namespace parallel { inline namespace v1 {
             // block manager for forward access iterator.
             template <typename FwdIter>
             class block_manager<FwdIter,
-                typename std::enable_if<
-                    hpx::traits::is_forward_iterator<FwdIter>::value &&
-                    !hpx::traits::is_random_access_iterator<FwdIter>::value>::
-                    type>
+                typename std::enable_if_t<
```
hkaiser, line 896: Same here, `typename` can go.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/partition.hpp
```diff
@@ -1017,84 +1420,36 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+            hpx::traits::is_iterator_v<FwdIter> &&
+            traits::is_projected<Proj, FwdIter>::value &&
+            traits::is_indirect_callable<ExPolicy,
+                    Pred, traits::projected<Proj, FwdIter>>::value
+        )>
+    // clang-format on
+    HPX_DEPRECATED_V(1, 8,
+        "hpx::parallel::partition is deprecated, use "
+        "hpx::partition instead")
+        typename util::detail::algorithm_result<ExPolicy, FwdIter>::type
```
hkaiser, line 1437: IIRC, `util::detail::algorithm_result_t` is a thing now as well.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/partition.hpp
```diff
@@ -1265,120 +1622,241 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
-                                    traits::projected<Proj, FwdIter1>>::value)>
+        HPX_CONCEPT_REQUIRES_(
+            hpx::is_execution_policy<ExPolicy>::value &&
+            hpx::traits::is_iterator_v<FwdIter1> &&
+            hpx::traits::is_iterator_v<FwdIter2> &&
+            hpx::traits::is_iterator_v<FwdIter3> &&
+            traits::is_projected<Proj, FwdIter1>::value &&
+            traits::is_indirect_callable<ExPolicy, Pred,
+                traits::projected<Proj, FwdIter1>>::value
+        )>
```
hkaiser, line 1637: All of the above have `_v` versions now. If not, let's add those.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/partition.hpp
Jedi18: Since the stable_partition implementation isn't trivial, would it be okay if I add it in a seperate PR? Or would you prefer I add it in this PR itself?

hkaiser: Sure, np.

libs/parallelism/algorithms/include/hpx/parallel/algorithms/partition.hpp
Jedi18: Is this okay? https://github.com/STEllAR-GROUP/hpx/blob/b9a3820071ecf47286c58661a5086772933a5214/libs/parallelism/algorithms/include/hpx/parallel/algorithms/detail/advance_and_get_distance.hpp

libs/parallelism/algorithms/include/hpx/parallel/algorithms/partition.hpp
```diff
@@ -118,14 +569,42 @@ namespace hpx { namespace parallel { inline namespace v1 {
...
+                        // does not return the new middle point
+                        std::advance(first, std::distance(mid, last));
+                        return first;
+                    },
+                    std::move(left), std::move(right));
+            }
+        };
+
+        template <typename BidirIter, typename Sent, typename F, typename Proj>
+        static BidirIter stable_partition_seq(
```
Jedi18, line 582: @hkaiser This algorithm has a O(n) space complexity since we need to store the false values. The standard library implementations also have a less efficient O(nlogn) algorithm that they use if sufficient space is not available.

Should we add the other algorithm too? If so, how do we check if sufficient space is available?

libs/parallelism/algorithms/include/hpx/parallel/algorithms/detail/advance_and_get_distance.hpp
```diff
@@ -0,0 +1,53 @@
+//  Copyright (c) 2019 Hartmut Kaiser
```
Jedi18, line 1: @hkaiser Is this okay?

libs/parallelism/algorithms/include/hpx/parallel/algorithms/partition.hpp
hkaiser: We can do that, I'd suggest leaving it to a separate PR, however. 

Re how to check: try and catch `std::bad_alloc` exceptions ;-) How is that handled in the std libraries?

hkaiser (approved):
LGTM, thanks!

Are the cosmetic changes to the unrelated algorithms separable? Might be better to create a different PR for those...

libs/parallelism/algorithms/include/hpx/parallel/algorithms/partition.hpp
Jedi18: `We can do that, I'd suggest leaving it to a separate PR, however.`
Oh ok sure.
`Re how to check: try and catch std::bad_alloc exceptions ;-)`
Thanks!
`How is that handled in the std libraries?`
They try and allocate a temporary buffer, if the length of the returned buffer is not enough then they revert to the inplace implementation.

hkaiser (approved):
LGTM, thanks!

PR #5541, opened 2021-09-07
Downgrade CI on rostam to CUDA 11.2

aurianer:
I didn't find yet a workaround for all CI configuration for the second `nvcc` bug introduced in CUDA 11.3, this temporarily downgrade to 11.2 to fix master

### DIFF
--- a/.jenkins/lsu/env-gcc-9-cuda-11.sh
+++ b/.jenkins/lsu/env-gcc-9-cuda-11.sh
@@ -9,7 +9,7 @@ module load cmake
 module load gcc/9
 module load boost/1.73.0-${build_type,,}
 module load hwloc
-module load cuda/11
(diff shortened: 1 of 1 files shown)


aurianer:
Looks like there is a problem of access on the compute nodes, it builds fine on the login node. So I'll wait for a fix to the cuda 11.2 installation

msimberg:
Since #5473 looks like it's almost ready to go, we can hopefully skip this PR.

aurianer:
Closing since #5473 seems to solve the problem with CUDA 11.3 and CUDA 11.4

aurianer:
Closing this since #5473 seems to solve the problem with CUDA 11.3 and CUDA 11.4, will reopen if necessary

PR #5542, opened 2021-09-09
Simplify execution CPOs

msimberg:
In P0443 member functions were given priority over free functions in the various CPOs. In P2300 tag_dispatch is the only way to implement various CPOs, i.e. member functions are not taken into account. This updates our execution CPOs to behave like P2300. This means that `schedule`, `set_value`, `set_error`, `set_done`, `start`, `connect` inherit `hpx::functional::tag`, not `hpx::functional::tag_priority`. This is slightly more verbose (`friend void tag_dispatch(start_t, ...)` instead of `void start(...)`) but is simpler to reason about. Additionally, the `thread_pool_scheduler` test compiles 10-15% faster with this change.

As a flyby I've:
- Removed the leftover scheduler is a sender functionality (i.e. one has to use `schedule(scheduler)` to get a sender; `thread_pool_scheduler` was previously a sender in itself)
- `just` is no longer a CPO (customization makes very little sense, this is aligned with P2300)
- `keep_future` is no longer a CPO (it only makes sense for futures)

Algorithms that have a reasonable default implementation still use `tag_fallback_dispatch` as before.

I think we should also revisit the `tag_dispatch`/`tag_invoke` question. The way we are now using `tag_dispatch` for the execution CPOs is really what `tag_invoke` is supposed to be. For the case of the segmented algorithms I believe we could put the dispatching one level down into the algorithm implementations as an internal implementation detail, but it should not use the "public" customization point.

### DIFF
--- a/libs/core/async_cuda/include/hpx/async_cuda/transform_stream.hpp
+++ b/libs/core/async_cuda/include/hpx/async_cuda/transform_stream.hpp
@@ -137,8 +137,8 @@ namespace hpx { namespace cuda { namespace experimental {
 
             template <typename R_, typename F_>
             transform_stream_receiver(R_&& r, F_&& f, cudaStream_t stream)
-              : r(std::forward<R>(r))
-              , f(std::forward<F>(f))
+              : r(std::forward<R_>(r))
+              , f(std::forward<F_>(f))
               , stream(stream)
             {
             }
--- a/libs/core/async_cuda/include/hpx/async_cuda/transform_stream.hpp
+++ b/libs/core/async_cuda/include/hpx/async_cuda/transform_stream.hpp
@@ -155,11 +155,10 @@ namespace hpx { namespace cuda { namespace experimental {
                 transform_stream_receiver&& r) noexcept
             {
                 hpx::execution::experimental::set_done(std::move(r.r));
-            };
+            }
 
             template <typename... Ts>
-            friend void tag_dispatch(hpx::execution::experimental::set_value_t,
-                transform_stream_receiver&& r, Ts&&... ts) noexcept
+            void set_value(Ts&&... ts) noexcept
             {
                 hpx::detail::try_catch_exception_ptr(
                     [&]() {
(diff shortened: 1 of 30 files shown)


msimberg:
> That's exactly what we just removed. The goal was to completely separate the segmented implementations from the non-segmented (normal) ones. I think the base implementations shouldn't be aware of a possible segmented overload.

Right. I suppose the problem is that now the "odd" name `tag_dispatch` (there because we want to do dispatching for things like the segmented algorithms) leaks into things that are legitimately `tag_invoke`-y. One solution could be to have both `tag_dispatch` and `tag_invoke`. In any case, I'm not proposing to change that in this PR, it's just something that came up while doing these changes.

hkaiser:
> > That's exactly what we just removed. The goal was to completely separate the segmented implementations from the non-segmented (normal) ones. I think the base implementations shouldn't be aware of a possible segmented overload.
> 
> Right. I suppose the problem is that now the "odd" name `tag_dispatch` (there because we want to do dispatching for things like the segmented algorithms) leaks into things that are legitimately `tag_invoke`-y. One solution could be to have both `tag_dispatch` and `tag_invoke`. In any case, I'm not proposing to change that in this PR, it's just something that came up while doing these changes.

I was thinking about possible interactions between our `tag_invoke`  and a potential `std::tag_invoke`. From what I see, even if we would name our facility `tag_invoke`, there wouldn't be any bad situations, where a `std::tag_invoke` could accidentally consider one of our APIs (the used tags would prevent that). That leaves us only with conceptual reasons to name our facilities `dispatch` instead of `invoke`.

I still think that `tag_dispatch` is much more than just `tag_invoke` as it provides convenient ways to manage overloading. We should consider all of our uses before starting to rename `tag_dispatch` back into `tag_invoke`.

hkaiser:
Also: what about the algorithm overloads for Kokkos? Wouldn't that require `tag_fallback_dispatch` to function as well?

msimberg:
Responded here https://github.com/STEllAR-GROUP/hpx/pull/5348#issuecomment-919174774, since the `tag_invoke/dispatch` discussion doesn't affect this PR.

aurianer (approved):
LGTM thanks!

PR #5543, opened 2021-09-10
Extend launch policy to carry stack size and scheduling hint in addition to priority

hkaiser:
This PR extends `hpx::launch` and related types (i.e. `hpx::launch::async`, etc.) such that they contain the scheduling hint and the stack size in addition to the thread priority.

This will be needed for the direct scoped execution to be able to control whether threads should be executed directly or not. This also simplifies a couple of internal interfaces.

As a flyby, this PR changes all thread enums to be of size `std::int8_t`, which allows for the launch policy to fit into a 64 bit register.

### DIFF
--- a/libs/core/async_base/CMakeLists.txt
+++ b/libs/core/async_base/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2020 The STE||AR-Group
+# Copyright (c) 2020-2021 The STE||AR-Group
 #
 # SPDX-License-Identifier: BSL-1.0
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
@@ -7,9 +7,13 @@
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 set(async_base_headers
-    hpx/async_base/apply.hpp hpx/async_base/async.hpp
-    hpx/async_base/dataflow.hpp hpx/async_base/launch_policy.hpp
-    hpx/async_base/sync.hpp hpx/async_base/traits/is_launch_policy.hpp
+    hpx/async_base/apply.hpp
+    hpx/async_base/async.hpp
+    hpx/async_base/dataflow.hpp
+    hpx/async_base/launch_policy.hpp
+    hpx/async_base/sync.hpp
+    hpx/async_base/traits/is_launch_policy.hpp
+    hpx/async_base/scheduling_properties.hpp
 )
 
 set(async_base_sources launch_policy.cpp)
@@ -29,5 +33,6 @@ add_hpx_module(
   COMPAT_HEADERS ${async_base_compat_headers}
   SOURCES ${async_base_sources}
   MODULE_DEPENDENCIES hpx_allocator_support hpx_config hpx_coroutines
+                      hpx_tag_dispatch
   CMAKE_SUBDIRS examples tests
 )
--- a/libs/core/async_base/include/hpx/async_base/launch_policy.hpp
+++ b/libs/core/async_base/include/hpx/async_base/launch_policy.hpp
@@ -9,16 +9,20 @@
 #pragma once
 
 #include <hpx/config.hpp>
+#include <hpx/async_base/scheduling_properties.hpp>
+#include <hpx/async_base/traits/is_launch_policy.hpp>
 #include <hpx/coroutines/thread_enums.hpp>
 #include <hpx/serialization/serialization_fwd.hpp>
 
+#include <cstdint>
 #include <type_traits>
 #include <utility>
 
 namespace hpx {
     /// \cond NOINTERNAL
     namespace detail {
-        enum class launch_policy
+
+        enum class launch_policy : std::int8_t
         {
             async = 0x01,
             deferred = 0x02,
@@ -37,9 +41,14 @@ namespace hpx {
         {
             constexpr explicit policy_holder_base(launch_policy p,
                 threads::thread_priority priority =
-                    threads::thread_priority::default_) noexcept
+                    threads::thread_priority::default_,
+                threads::thread_stacksize stacksize =
+                    threads::thread_stacksize::default_,
+                threads::thread_schedule_hint hint = {}) noexcept
               : policy_(p)
               , priority_(priority)
+              , stacksize_(stacksize)
+              , hint_(hint)
             {
             }
 
@@ -58,13 +67,41 @@ namespace hpx {
                 return static_cast<int>(policy_) != 0;
             }
 
-            constexpr threads::thread_priority get_priority() const
+            constexpr threads::thread_priority get_priority() const noexcept
             {
                 return priority_;
             }
 
+            constexpr threads::thread_stacksize get_stacksize() const noexcept
+            {
+                return stacksize_;
+            }
+
+            constexpr threads::thread_schedule_hint get_hint() const noexcept
+            {
+                return hint_;
+            }
+
+            void set_priority(threads::thread_priority priority) noexcept
+            {
+                priority_ = priority;
+            }
+
+            void set_stacksize(threads::thread_stacksize stacksize) noexcept
+            {
+                stacksize_ = stacksize;
+            }
+
+            void set_hint(threads::thread_schedule_hint hint) noexcept
+            {
+                hint_ = hint;
+            }
+
+        protected:
             launch_policy policy_;
             threads::thread_priority priority_;
+            threads::thread_stacksize stacksize_;
+            threads::thread_schedule_hint hint_;
 
         private:
             friend class serialization::access;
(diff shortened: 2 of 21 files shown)


libs/parallelism/async_base/include/hpx/async_base/launch_policy.hpp
```diff
@@ -138,90 +172,235 @@ namespace hpx {
...
             {
-                return async_policy(priority);
+                return async_policy(priority, stacksize, hint);
+            }
+
+            async_policy& operator=(threads::thread_priority priority) noexcept
+            {
+                this->priority_ = priority;
+                return *this;
+            }
```
msimberg, line 219: I find this a bit questionable. I think I'd expect
```
policy = priority;
```
to behave the same as
```
policy = {priority};
```
If anything maybe
```
policy += priority;
```
could change only the priority, but even that is very questionable.

Should we not use `with_priority/stacksize/hint` as we do with the executors instead?

libs/parallelism/async_base/tests/unit/launch_policy.cpp
```diff
@@ -0,0 +1,58 @@
...
+    HPX_TEST_EQ(policy.hint().hint, hint.hint);
+}
+
+int main()
+{
+    HPX_TEST_LTE(sizeof(hpx::launch::async_policy), sizeof(std::int64_t));
+    HPX_TEST_LTE(sizeof(hpx::launch::sync_policy), sizeof(std::int64_t));
+    HPX_TEST_LTE(sizeof(hpx::launch::deferred_policy), sizeof(std::int64_t));
+    HPX_TEST_LTE(sizeof(hpx::launch::fork_policy), sizeof(std::int64_t));
+    HPX_TEST_LTE(sizeof(hpx::launch::apply_policy), sizeof(std::int64_t));
```
msimberg, line 43:
```suggestion
    static_assert(sizeof(hpx::launch::async_policy) == sizeof(std::int64_t));
    static_assert(sizeof(hpx::launch::sync_policy) == sizeof(std::int64_t));
    static_assert(sizeof(hpx::launch::deferred_policy) == sizeof(std::int64_t));
    static_assert(sizeof(hpx::launch::fork_policy) == sizeof(std::int64_t));
    static_assert(sizeof(hpx::launch::apply_policy) == sizeof(std::int64_t));
```

libs/parallelism/async_base/include/hpx/async_base/launch_policy.hpp
hkaiser: Yes, I was already planning to change this. I will replace the `operator=()` with equivalent `operator()()` overloads, which will make things more `constexpr` friendly as well.

libs/parallelism/async_base/tests/unit/launch_policy.cpp
hkaiser: Should we rely on `==`, I think `<=` would be more appropriate.

msimberg: Uh, brainfart. I only meant the change from `HPX_TEST_X` to `static_assert`. The change to `==` was unintentional.

libs/parallelism/async_base/include/hpx/async_base/launch_policy.hpp
msimberg: Do you mean yop'd change it to have
```
policy(priority);
```
or, possibly return a new policy:
```
policy = policy(priority);
```
?

As I said, why not `with_x` as we do elsewhere?

hkaiser: > Do you mean yop'd change it to have
> 
> ```
> policy(priority);
> ```
> 
> or, possibly return a new policy:
> 
> ```
> policy = policy(priority);
> ```
> 
> ?

yes

> As I said, why not `with_x` as we do elsewhere?

Isn't that a bit overkill?

msimberg: Is it overkill for executors? If no, I don't think it is for launch policies either. If this is indeed meant to as a user-facing facility (which it is, right?) then I'd rather not have to explain to users that to set a priority for an executor you use `with_priority(exec, prio)` but for launch policies you do `policy(prio)`.

msimberg: > yes, even more as we've already had `launch::async(priority)`

we also have `parallel_policy_executor(priority)` so should have `exec(priority)` as well? I think no. Those are constructors where it's much more explicit that you're creating a new instance. Overloading the call operator for this purpose is more confusing than helpful, to me at least. Plus it has the downside that getting the priority (or any property like that) still needs a plain member function which is to me a weirdly asymmetric situation.

hkaiser: > Is it overkill for executors? If no, I don't think it is for launch policies either. If this is indeed meant to as a user-facing facility (which it is, right?) then I'd rather not have to explain to users that to set a priority for an executor you use `with_priority(exec, prio)` but for launch policies you do `policy(prio)`.

Fair point. Should we deprecate `hpx::launch::async(priority)` in this case?

libs/core/properties/include/hpx/properties/scheduling_properties.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+    HPX_INLINE_CONSTEXPR_VARIABLE struct with_annotation_t final
+      : detail::property_base<with_annotation_t>
+    {
+    } with_annotation{};
+
+    HPX_INLINE_CONSTEXPR_VARIABLE struct get_annotation_t final
+      : detail::property_base<get_annotation_t>
+    {
+    } get_annotation{};
```
msimberg, line 81: I don't think these belong in the properties module. `execution` or `execution_base` seems more appropriate.

libs/parallelism/executors/include/hpx/executors/detail/hierarchical_spawning.hpp
```diff
@@ -56,6 +55,9 @@ namespace hpx { namespace parallel { namespace execution { namespace detail {
         std::size_t const size = hpx::util::size(shape);
         results.resize(size);
 
+        auto post_policy = policy;
+        post_policy.set_stacksize(threads::thread_stacksize::small_);
```
msimberg, line 59: Do you count this as "internal use", hence `set_stacksize` and not `with_stacksize` (if yes, :+1:)?

msimberg (commented):
Nice! This seems to have the nice side-effect of simplifying all the functions that take priorities, stacksizes, and hints.

libs/parallelism/async_base/include/hpx/async_base/launch_policy.hpp
msimberg: > Fair point. Should we deprecate `hpx::launch::async(priority)` in this case?

I don't think that's necessary. It's still useful just like the constructors for the parallel executor are useful. I was mainly concerned with having redundant ways of setting and getting those properties.

libs/core/properties/include/hpx/properties/scheduling_properties.hpp
hkaiser: Ok, will try moving the file. I was afraid of this creating a circular dependency between the execution and async_base modules.

libs/parallelism/executors/include/hpx/executors/detail/hierarchical_spawning.hpp
hkaiser: Yes, internally I have used the `set_` versions, everywhere.

msimberg (dismissed):
Looks good to me. Thanks @hkaiser!

Except that most CI configurations don't like the changes... This can go in once they are happy.

hkaiser:
This looks good now, I believe.

msimberg:
I suppose these would be fitting to change in this PR: https://app.codacy.com/gh/STEllAR-GROUP/hpx/pullRequest?prid=8056581. Are you able to reproduce this: https://cdash.cscs.ch/testDetails.php?test=51703242&build=180874?

hkaiser:
> I suppose these would be fitting to change in this PR: https://app.codacy.com/gh/STEllAR-GROUP/hpx/pullRequest?prid=8056581. 

These are deliberate, I'll ignore the tests for codacy.

> Are you able to reproduce this: https://cdash.cscs.ch/testDetails.php?test=51703242&build=180874?

Not yet, looks like something compiler specific, though.

msimberg:
@hkaiser looks like this is ready to go. Did you find out what was causing https://cdash.cscs.ch/testDetails.php?test=51703242&build=180874?

hkaiser:
> @hkaiser looks like this is ready to go. Did you find out what was causing https://cdash.cscs.ch/testDetails.php?test=51703242&build=180874?

Not yet, I wasn't even able to reproduce it locally...

hkaiser:
> > @hkaiser looks like this is ready to go. Did you find out what was causing https://cdash.cscs.ch/testDetails.php?test=51703242&build=180874?
> 
> Not yet, I wasn't even able to reproduce it locally...

@msimberg do you recall whether this particular platform has any special configurations?

msimberg:
@hkaiser I think C++20 is the only possibly relevant configuration option: https://github.com/STEllAR-GROUP/hpx/blob/5f9505f5312d2d538b64039c862154244f98d33f/.jenkins/cscs/env-gcc-newest.sh#L10.

PR #5546, opened 2021-09-10
Fixing shutdown problems caused by missing ref-counting

hkaiser:
- flyby: handle terminate threads while waiting for activities to cease

### DIFF
--- a/libs/core/thread_pools/include/hpx/thread_pools/scheduled_thread_pool_impl.hpp
+++ b/libs/core/thread_pools/include/hpx/thread_pools/scheduled_thread_pool_impl.hpp
@@ -194,6 +194,7 @@ namespace hpx { namespace threads { namespace detail {
                 hpx_thread_offset;
         bool have_polling_work =
             sched_->Scheduler::get_polling_work_count() > 0;
+
         return have_hpx_threads || have_polling_work;
     }
 
--- a/libs/core/threading_base/src/set_thread_state.cpp
+++ b/libs/core/threading_base/src/set_thread_state.cpp
@@ -25,7 +25,7 @@
 namespace hpx { namespace threads { namespace detail {
 
     ///////////////////////////////////////////////////////////////////////////
-    thread_result_type set_active_state(thread_id_type const& thrd,
+    thread_result_type set_active_state(thread_id_ref_type thrd,
         thread_schedule_state newstate, thread_restart_state newstate_ex,
         thread_priority priority, thread_state previous_state)
     {
@@ -59,7 +59,7 @@ namespace hpx { namespace threads { namespace detail {
 
         // just retry, set_state will create new thread if target is still active
         error_code ec(lightweight);    // do not throw
-        detail::set_thread_state(thrd, newstate, newstate_ex, priority,
+        detail::set_thread_state(thrd.noref(), newstate, newstate_ex, priority,
             thread_schedule_hint(), true, ec);
 
         return thread_result_type(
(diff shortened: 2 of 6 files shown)


libs/core/thread_pools/include/hpx/thread_pools/scheduled_thread_pool_impl.hpp
```diff
@@ -194,7 +194,13 @@ namespace hpx { namespace threads { namespace detail {
...
         bool have_polling_work =
             sched_->Scheduler::get_polling_work_count() > 0;
-        return have_hpx_threads || have_polling_work;
+
+        if (have_hpx_threads || have_polling_work)
+        {
+            sched_->Scheduler::cleanup_terminated(true);
+            return true;
+        }
+        return false;
```
msimberg, line 203: Is this essential for fixing whatever shutdown problems this is fixing? Note that `threadmanager::cleanup_terminated` is called inside the `yield_while` loops in `runtime_support_server.cpp`.

libs/core/threading_base/src/set_thread_state.cpp
```diff
@@ -134,8 +134,8 @@ namespace hpx { namespace threads { namespace detail {
                         get_thread_state_name(new_state));
 
                     thread_init_data data(
-                        util::bind(&set_active_state, thrd, new_state,
-                            new_state_ex, priority, previous_state),
+                        util::bind(&set_active_state, thread_id_ref_type(thrd),
+                            new_state, new_state_ex, priority, previous_state),
```
msimberg, line 138: Is there a difference here compared to the old code? This is explicitly making a copy, whereas the old code would implicitly make a copy, but the result is the same, right?

libs/core/thread_pools/include/hpx/thread_pools/scheduled_thread_pool_impl.hpp
hkaiser: It was not necessary, but helped speeding up the shutdown process as terminated threads were cleaned up right away.

aurianer (approved):
LGTM thanks!

libs/core/threading_base/src/set_thread_state.cpp
hkaiser: The old code did make a copy of a non-reference counted thread-id.

libs/core/thread_pools/include/hpx/thread_pools/scheduled_thread_pool_impl.hpp
msimberg: I guess I just don't see how this helps speed things up. Is there a place where `threadmanager::is_busy` is called repeatedly without a corresponding `threadmanager::cleanup_terminated` right next to it? I think I'd prefer it if `is_busy` really did just check if the pool is busy or not.

libs/core/threading_base/src/set_thread_state.cpp
msimberg: With `thrd` a `thread_id_ref_type`, couldn't you just pass `thrd` to `bind` and it would behave exactly the same? Sorry if I'm being thick. I just want to understand if I'm missing something and there's a functional difference between the two.

libs/core/thread_pools/include/hpx/thread_pools/scheduled_thread_pool_impl.hpp
hkaiser: Yes, there is a place where this happens. I'll go back and change things there.

libs/core/threading_base/src/set_thread_state.cpp
hkaiser: AFAICT, `thrd` is a non-reference counted thread-id (see https://github.com/STEllAR-GROUP/hpx/pull/5546/files/70bbe3eab3988c43977b636e4cb329f3c4ee645a#diff-93346ad9af6b7251c1205b63dcfe560916d8df9b783850442dc6bedecd52b71cR70)

msimberg: Sorry for the noise. You're absolutely right. I just don't know how to read the diff. It makes absolute sense now.

libs/core/thread_pools/include/hpx/thread_pools/scheduled_thread_pool_impl.hpp
hkaiser: I was wrong, everything is ok here. I have reverted this change.

msimberg (approved):

PR #5547, opened 2021-09-10
Fixing communication set

hkaiser:
This fixes occasional hangs of the communication set test

### DIFF
--- a/libs/full/collectives/src/detail/communication_set_node.cpp
+++ b/libs/full/collectives/src/detail/communication_set_node.cpp
@@ -105,12 +105,12 @@ namespace hpx { namespace lcos { namespace detail {
     std::size_t calculate_num_connected(
         std::size_t num_sites, std::size_t site, std::size_t arity)
     {
-        std::size_t num_children = hpx::ranges::count(hpx::execution::par,
-            hpx::util::make_counting_iterator(site + 1),
-            hpx::util::make_counting_iterator(num_sites), site,
-            [&](std::size_t node) {
-                return calculate_connected_node(node, arity);
-            });
+        std::size_t num_children =
+            hpx::ranges::count(hpx::util::make_counting_iterator(site + 1),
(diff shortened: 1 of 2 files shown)


libs/full/collectives/src/detail/communication_set_node.cpp
```diff
@@ -105,12 +105,12 @@ namespace hpx { namespace lcos { namespace detail {
...
-            hpx::util::make_counting_iterator(num_sites), site,
-            [&](std::size_t node) {
-                return calculate_connected_node(node, arity);
-            });
+        std::size_t num_children =
+            hpx::ranges::count(hpx::util::make_counting_iterator(site + 1),
+                hpx::util::make_counting_iterator(num_sites), site,
+                [&](std::size_t node) {
+                    return calculate_connected_node(node, arity);
+                });
```
msimberg, line 113: I guess this is unrelated to the hangs? (We can keep the change here, just trying to understand.)

libs/full/collectives/tests/unit/communication_set.cpp
```diff
@@ -86,12 +86,13 @@ namespace hpx { namespace traits {
...
-        Result get(std::size_t, test::get_connected_to_zero)
+        Result get(std::size_t which, test::get_connected_to_zero)
         {
             // first forward request to parent
-            if (communicator_.connect_to_ != communicator_.site_)
+            if (communicator_.connect_to_ != communicator_.site_ &&
+                which == communicator_.site_)
             {
-                hpx::sync(test::get_connected_to_zero_action{},
+                hpx::apply(test::get_connected_to_zero_action{},
```
msimberg, line 95: Does this synchronize with the code after this through the mutex? I guess this is somehow the main trigger to potential deadlocks?

hkaiser: The test does not depend on the synchronization. And yes, this was possibly causing deadlocks.

PR #5549, opened 2021-09-11
Fixing agent_ref::yield_k to actually call yield_k

hkaiser:


hkaiser:
This might require for #5546 to be merged first in order to avoid hangs at shutdown.

libs/core/execution_base/src/agent_ref.cpp
```diff
@@ -27,14 +27,14 @@ namespace hpx { namespace execution_base {
...
-    void agent_ref::yield_k(std::size_t /* k */, const char* desc)
+    void agent_ref::yield_k(std::size_t k, const char* desc)
     {
         HPX_ASSERT(*this == hpx::execution_base::this_thread::agent());
         // verify that there are no more registered locks for this OS-thread
 #ifdef HPX_HAVE_VERIFY_LOCKS
         util::verify_no_locks();
 #endif
-        impl_->yield(desc);
+        impl_->yield_k(k, desc);
```
msimberg, line 37: Good catch!

How does this affect shutdown? Calling `yield` instead of `yield_k` should as far as I can tell maybe slow down shutdown (or speed it up) in some cases, but eventually `yield_k` also yields like `yield`.

hkaiser: `yield` never actually suspends (at least in the context where this was triggered. `yield_k` does suspend. This caused occasional shutdown hangs as threads couldn't run.

msimberg: Are you referring to this: https://github.com/STEllAR-GROUP/hpx/blob/c1923e5d1d6edef8883e99c2ee2fc07cad6876ec/libs/core/execution_base/src/this_thread.cpp#L93-L104? It looks like that should yield, and not just do `HPX_SMT_PAUSE`, no?

hkaiser: Yes, that's the spot.

msimberg: Do you agree that we should remove `HPX_SMT_PAUSE` from there? It's orthogonal to the fix here, so I can do it in a separate PR.

msimberg: Yep, I meant leaving only the non-smt-pause case.

hkaiser: No objections here.

msimberg (approved):

msimberg:
I suppose we really have to wait for #5546. `launch_process` really doesn't like this change. Do you have any guesses why?

hkaiser:
> I suppose we really have to wait for #5546. `launch_process` really doesn't like this change. Do you have any guesses why?

That's because the shutdown is fixed only in #5546. It's waiting for threads to be cleaned up -- which never happens as `yield_k` is never yielding....

aurianer (dismissed):
LGTM thanks!

hkaiser:
retest

hkaiser:
@msimberg the remaining problems seem to be unrelated. I think this can be merged now.

PR #5551, opened 2021-09-13
Fixing for local executions of actions to properly set task names

hkaiser:


msimberg (approved):
Looks good, thanks!

msimberg:
I suppose the `apex_action_count` test might need updating after this fix: https://cdash.cscs.ch/testDetails.php?test=51364877&build=180058?

hkaiser:
> I suppose the `apex_action_count` test might need updating after this fix: https://cdash.cscs.ch/testDetails.php?test=51364877&build=180058?

Uhhh, would you be able to do that? I have no APEX setup available easily. Thanks for noticing.

msimberg:
@hkaiser I've just updated the count in the test. I suppose the fix here was part of the "some reason" (https://github.com/STEllAR-GROUP/hpx/blob/30f47ed5ea8622bf526ee0080d276f73142f2be2/tests/unit/apex/apex_action_count.cpp#L108) for the count being off. It's still off by one, so is it possible that some other action names aren't being set?

hkaiser:
> @hkaiser I've just updated the count in the test. I suppose the fix here was part of the "some reason" (
> 
> https://github.com/STEllAR-GROUP/hpx/blob/30f47ed5ea8622bf526ee0080d276f73142f2be2/tests/unit/apex/apex_action_count.cpp#L108
> 
> ) for the count being off. It's still off by one, so is it possible that some other action names aren't being set?

Not sure why, I'll try to have a look.

msimberg:
We decided to leave the final missing task name for a future investigation. This is already an improvement so merging.

PR #5557, opened 2021-09-15
Adding an example for the new channel_communicator API

hkaiser:
@NanmiaoWu Please check out this example and let us know if it helps you with the issues you reported


NanmiaoWu:
> @NanmiaoWu Please check out this example and let us know if it helps you with the issues you reported

@hkaiser Thank you! The example could work when using 3 or more localities. But it will fail when using 2 localities. The failing info is [error_2loc](https://gist.github.com/NanmiaoWu/6849c9c66ea9b21b135ca5e7258b4593).

I write a example code for 2 localities, it could work. see here [2_loc_example](https://gist.github.com/NanmiaoWu/d46d545222af08e4046b8b19b20b6095)

PR #5560, opened 2021-09-15
Update tests and examples

msimberg:
- Make more tests and examples compile with the local runtime
- Update the OpenMP examples and enable them in one CI configuration
- Move some more tests to modules (functional)


examples/1d_stencil/1d_stencil_1.cpp
```diff
@@ -105,32 +106,32 @@ int hpx_main(hpx::program_options::variables_map& vm)
...
+        "Local x dimension") ("nt",
+        po::value<std::uint64_t>()->default_value(45),
+        "Number of time steps") ("k", po::value<double>(&k)->default_value(0.5),
+        "Heat transfer coefficient (default: 0.5)") ("dt",
+        po::value<double>(&dt)->default_value(1.0),
+        "Timestep unit (default: 1.0[s])") ("dx",
+        po::value<double>(&dx)->default_value(1.0),
+        "Local x dimension") ("no-header",
+        "do not print out the csv header row");
+    // clang-format on
```
hkaiser, line 130: Can we avoid reformatting the program options, please?

examples/jacobi_smp/jacobi.cpp
```diff
@@ -15,6 +15,7 @@
 #include <iostream>
 #include <string>
 
+using hpx::program_options::command_line_parser;
 using hpx::program_options::options_description;
 using hpx::program_options::parse_command_line;
```
hkaiser, line 20: This is not needed anymore, I think.

examples/quickstart/customize_async.cpp
```diff
@@ -79,5 +79,10 @@ int main()
         f.wait();
     }
 
-    return 0;
+    return hpx::local::finalize();
+}
+
+int main(int argc, char* argv[])
```
hkaiser, line 85: Unrelated to this PR: should we add support for `#include <hpx/local/hpx_main.hpp>`?

examples/1d_stencil/1d_stencil_1.cpp
msimberg: Yes. I've tried to them from clang-formatting but missed this one. Thanks!

examples/quickstart/customize_async.cpp
msimberg: My preference is not to add it.

hkaiser: What's your rationale for this? Wouldn't it simplify porting codes?

msimberg: Not noticeably. While it's convenient on the surface, the implementation is complex, error prone and not really portable. That's of course only my opinion and I can see how others might think that's a fine tradeoff.

aurianer (dismissed):
LGTM thanks!

PR #5562, opened 2021-09-15
More changes to overcome build problems on Windows after recent module rearrangements

hkaiser:
- this is a somewhat hacky solution as it marks functions that are declared in the core module with `HPX_EXPORT` (those are defined in full, currently)

### DIFF
--- a/libs/core/mpi_base/include/hpx/mpi_base/mpi_environment.hpp
+++ b/libs/core/mpi_base/include/hpx/mpi_base/mpi_environment.hpp
@@ -41,7 +41,7 @@ namespace hpx { namespace util {
 
         static std::string get_processor_name();
 
-        struct scoped_lock
+        struct HPX_CORE_EXPORT scoped_lock
         {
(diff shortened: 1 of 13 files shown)


msimberg:
@hkaiser I've pushed a separate commit which:
- Removes unnecessary forward declarations (the functinos had correctly exported declarations in core, but additional wrongly exported declarations in full)
- Moves the static factory data functionality to the core module (where the declarations were anyway)

Could you please verify if this still works on windows or if I've broken things even more for you.

hkaiser:
> @hkaiser I've pushed a separate commit which:
> 
> * Removes unnecessary forward declarations (the functinos had correctly exported declarations in core, but additional wrongly exported declarations in full)
> * Moves the static factory data functionality to the core module (where the declarations were anyway)
> 
> Could you please verify if this still works on windows or if I've broken things even more for you.

Thanks everything works as expected. We can go ahead and merge this.

PR #5567, opened 2021-09-17
Make default_agent::yield actually yield

msimberg:
When `HPX_SMT_PAUSE` is available, `default_agent::yield` would never actually yield. This changes it to actually yield (`HPX_SMT_PAUSE` is still used in the first iterations of `yield_k`). See discussion at https://github.com/STEllAR-GROUP/hpx/pull/5549#discussion_r707106060 for context.


hkaiser (approved):
LGTM, thanks!

aurianer (approved):
LGTM thanks!

PR #5569, opened 2021-09-17
Use no_unique_address in sender adaptors

msimberg:


hkaiser (approved):
Good thinking! LGTM, thanks!

We can probably do the same in many other places...

aurianer (approved):
LGTM thanks!

PR #5572, opened 2021-09-20
Remove more guards for the distributed runtime

msimberg:
These guards are no longer necessary since the modules themselves are not built/installed unless the distributed runtime is enabled. There are some more ifdefs that I'll remove when splitting things up. After that the `HPX_WITH_DISTRIBUTED_RUNTIME` option won't be necessary anymore.


hkaiser (approved):
Very nice, thanks a lot!

aurianer (approved):
LGTM thanks!

PR #5573, opened 2021-09-20
Enable hidden visibility by default with clang

msimberg:
Tested only with the local runtime and clang 12.

Pushing to see what the other clang builders think about this change.

@hkaiser any memory of why this was disabled in the first place?


msimberg:
I suppose all the clang setups failing is part of the reason this was off... We'll have to come back to this at a later point. If someone feels like picking this up to investigate what the problem is feel free to do so.

Issue #3733, opened 2019-03-05
exclusive_scan is faster in sequential than parallel


jefftrull:
## Expected Behavior

Parallel should be slower than sequential for small data sizes, then faster after some crossover point.

## Actual Behavior

Parallel is consistently slower than sequential for data sizes from 10 to 10000000

## Steps to Reproduce the Problem

Compile and run the attached code with Google Benchmark.

## Specifications

gbench hardware report:
```
Run on (8 X 3800 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 256K (x4)
  L3 Unified 6144K (x1)
```
  - HPX Version: 1.2.1
  - Platform (compiler, OS): gcc 8.2.0, Ubuntu 18.10

## Benchmark results:
```
-----------------------------------------------------------
Benchmark                    Time           CPU Iterations
-----------------------------------------------------------
Sequential/10                4 ns          4 ns  151589055
Sequential/64               27 ns         27 ns   26228755
Sequential/512             215 ns        215 ns    3314081
Sequential/4096           1719 ns       1719 ns     426286
Sequential/32768         13536 ns      13535 ns      50881
Sequential/262144       111051 ns     111045 ns       4988
Sequential/2097152     1108953 ns    1108894 ns        652
Sequential/10000000    5183736 ns    5183256 ns        131
Parallel/10              60798 ns    2172331 ns       3530
Parallel/64             149103 ns     172653 ns      10000
Parallel/512            151103 ns     312761 ns       1841
Parallel/4096           130873 ns 7155866560 ns          1
Parallel/32768          158808 ns     151704 ns       5030
Parallel/262144         227019 ns     161383 ns       3287
Parallel/2097152       1642302 ns 7508294482 ns          1
Parallel/10000000      7278212 ns    7715326 ns        182
```

jefftrull:
Looks like I can't simply attach code to this issue, so here's the MWE as inline text:

```c++
#include <random>

#include <hpx/hpx_init.hpp>
#include <hpx/hpx.hpp>
#include <hpx/include/parallel_numeric.hpp>
#include <hpx/include/parallel_algorithm.hpp>

#include <benchmark/benchmark.h>

int main(int argc, char* argv[])
{
    // process and remove gbench arguments
    benchmark::Initialize(&argc, argv);

    // By default this should run on all available cores
    std::vector<std::string> const cfg = {
[...]

PR #5575, opened 2021-09-21
Change scan partitioner implementation

Jedi18:
Modify scan partitioner implementation to execute f2 sequentially once f1 tasks are done, so that f3 tasks can all be executed in parallel.
Minor changes to algorithms using the scan partitioner.
Added benchmark for comparing sequential and parallel execution times of scan algorithms.

Issue #3733

### DIFF
--- a/libs/core/algorithms/tests/performance/benchmark_scan_algorithms.cpp
+++ b/libs/core/algorithms/tests/performance/benchmark_scan_algorithms.cpp
@@ -4,11 +4,23 @@
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-#include <hpx/algorithm.hpp>
-#include <hpx/hpx_init.hpp>
+#include <hpx/local/init.hpp>
+#include <hpx/modules/program_options.hpp>
+#include <hpx/modules/testing.hpp>
+#include <hpx/parallel/algorithms/copy.hpp>
+#include <hpx/parallel/algorithms/exclusive_scan.hpp>
+#include <hpx/parallel/algorithms/inclusive_scan.hpp>
+#include <hpx/parallel/algorithms/transform_exclusive_scan.hpp>
+#include <hpx/parallel/algorithms/transform_inclusive_scan.hpp>
+#include <hpx/parallel/algorithms/unique.hpp>
 
+#include <array>
+#include <cstddef>
 #include <fstream>
 #include <functional>
+#include <iostream>
+#include <map>
+#include <string>
 #include <vector>
 
 //#define OUTPUT_TO_CSV
@@ -177,8 +189,25 @@ void measureScanAlgorithms()
     }
 }
 
-int hpx_main(int, char**)
+int hpx_main(hpx::program_options::variables_map&)
 {
-    measureScanAlgorithms();
-    return hpx::finalize();
+    {
+        measureScanAlgorithms();
+    };
+    // Initiate shutdown of the runtime systems on all localities.
+    return hpx::local::finalize();
+};
+
+int main(int argc, char* argv[])
+{
+    std::vector<std::string> cfg;
+    cfg.push_back("hpx.os_threads=all");
+    hpx::local::init_params init_args;
+    init_args.cfg = cfg;
+
+    // Initialize and run HPX.
+    HPX_TEST_EQ_MSG(hpx::local::init(hpx_main, argc, argv, init_args), 0,
+        "HPX main exited with non-zero status");
+
+    return hpx::util::report_errors();
 }
--- a/libs/core/algorithms/tests/performance/benchmark_scan_algorithms.cpp
+++ b/libs/core/algorithms/tests/performance/benchmark_scan_algorithms.cpp
@@ -1,0 +1,184 @@
+//  Copyright (c) 2021 Akhil J Nair
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/algorithm.hpp>
+#include <hpx/hpx_init.hpp>
+
+#include <fstream>
+#include <functional>
+#include <vector>
+
+//#define OUTPUT_TO_CSV
+
+enum class ALGORITHM
+{
+    EXCLUSIVE_SCAN,
+    INCLUSIVE_SCAN,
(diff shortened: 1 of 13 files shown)


hkaiser:
Thanks @Jedi18! Could you have a look to fix the compilation problems, please? Most likely just a missing #include...

Jedi18:
Not a missing include, looks like I missed partition_copy which also uses the scan partitioner. I'll run the benchmark on it too to be sure that it is also showing performance improvements.

hkaiser:
> Not a missing include, looks like I missed partition_copy which also uses the scan partitioner. I'll run the benchmark on it too to be sure that it is also showing performance improvements.

Could you please add the performance results to this PR for us to see?

Jedi18:
@hkaiser Btw in the advance_and_get_sentinel, the [is_sized_sentinel_for](https://github.com/STEllAR-GROUP/hpx/blob/6b2b5d1cf04a654311ec6d6da200d86ad20ce896/libs/core/algorithms/include/hpx/parallel/algorithms/detail/advance_and_get_distance.hpp#L37) check is satisfied by the test_iterator wrapped around a bidirectional iterator (in the partition tests) and it fails to compile.

Do I change the advance_and_get_sentinel implementation or is there a way to make the check test for the base iterator of the test_iterator?

Sure, the performance results of all the algorithms or only partition_copy?

hkaiser:
> Sure, the performance results of all the algorithms or only partition_copy?

The more the merrier ;-)

hkaiser:
> @hkaiser Btw in the advance_and_get_sentinel, the [is_sized_sentinel_for](https://github.com/STEllAR-GROUP/hpx/blob/6b2b5d1cf04a654311ec6d6da200d86ad20ce896/libs/core/algorithms/include/hpx/parallel/algorithms/detail/advance_and_get_distance.hpp#L37) check is satisfied by the test_iterator wrapped around a bidirectional iterator (in the partition_copy tests) and it fails to compile.
> 
> Do I change the advance_and_get_sentinel implementation or is there a way to make the check test for the base iterator of the test_iterator?

Is this a bug in `is_sized_sentinel_for`?

Jedi18:
> Is this a bug in `is_sized_sentinel_for`?

The problem is that [here](https://github.com/STEllAR-GROUP/hpx/blob/3f12aca17be855dc211c3d91efe31641c31cd731/libs/core/iterator_support/include/hpx/iterator_support/traits/is_iterator.hpp#L172) T and U are test::test_iterator<base_iterator,std::bidirectional_iterator_tag> instead of the base_iterator and so the sentinel check thinks '-' is a valid operator for the iterator.

However when actually subtracting, the base iterator is checked and the static assertion fails and so the compile error.

hkaiser:
> > Is this a bug in `is_sized_sentinel_for`?
> 
> The problem is that [here](https://github.com/STEllAR-GROUP/hpx/blob/3f12aca17be855dc211c3d91efe31641c31cd731/libs/core/iterator_support/include/hpx/iterator_support/traits/is_iterator.hpp#L172) T and U are test::test_iterator<base_iterator,std::bidirectional_iterator_tag> instead of the base_iterator and so the sentinel check thinks '-' is a valid operator for the iterator.
> 
> However when actually subtracting, the base iterator is [checked](https://github.com/STEllAR-GROUP/hpx/blob/3f12aca17be855dc211c3d91efe31641c31cd731/libs/core/iterator_support/include/hpx/iterator_support/iterator_facade.hpp#L590) and the static assertion fails and so the compile error.

Thanks for diagnosing this! Looks like it is a bug in our `test_iterator`, then.

Jedi18:
@hkaiser For the time being, should I use a normal iterator instead of the test iterator in the partition_copy tests?

libs/core/algorithms/tests/performance/benchmark_scan_algorithms.cpp
```diff
@@ -0,0 +1,213 @@
...
+        {ALGORITHM::COPY_IF, "copyIfCompare.csv"},
+        {ALGORITHM::UNIQUE_COPY, "uniqueCopyCompare.csv"},
+    };
+#endif
+
+    for (int alg = (int) ALGORITHM::EXCLUSIVE_SCAN;
+         alg <= (int) ALGORITHM::UNIQUE_COPY; alg++)
+    {
+        std::size_t start = 32;
+        std::size_t till = 1 << 28;
```
msimberg, line 57: @Jedi18 this benchmark is timing out in CI, and I'm guessing it's because of this being too large in debug mode. Could you check how long a debug run of this test runs, and adjust this (in debug mode only) to something smaller. Try to target having the test run no longer than 30 seconds.

Jedi18: Thanks yeah it runs for too long, I've reduced the NUM_ITERATIONS and till variable, forgot to push the changes.

This PR is held up on a bug with copy_if that is causing the reduce_by_key test to fail, I'll look into it this week.

Jedi18:
@msimberg @hkaiser Ok so the issue with the reduce_by_key test is that it the destination iterator belongs to the same range as the input, which is an [undefined behaviour](https://en.cppreference.com/w/cpp/algorithm/copy).

https://github.com/STEllAR-GROUP/hpx/blob/35c7a2d752d7b6d25299198788ab8646b3609bed/libs/core/algorithms/include/hpx/parallel/algorithms/reduce_by_key.hpp#L403

hkaiser:
> @msimberg @hkaiser Ok so the issue with the reduce_by_key test is that it the destination iterator belongs to the same range as the input, which is an [undefined behaviour](https://en.cppreference.com/w/cpp/algorithm/copy) (_The behavior is undefined if d_first is within the range [first, last)_), so it dosen't work with the new scan partitioner implementation (and as it turns out also our sequential copy_if implementation)
> 
> https://github.com/STEllAR-GROUP/hpx/blob/35c7a2d752d7b6d25299198788ab8646b3609bed/libs/core/algorithms/include/hpx/parallel/algorithms/reduce_by_key.hpp#L403

Excellent catch.

msimberg:
> https://github.com/STEllAR-GROUP/hpx/blob/35c7a2d752d7b6d25299198788ab8646b3609bed/libs/core/algorithms/include/hpx/parallel/algorithms/reduce_by_key.hpp#L403

Do I see correctly that you seem to have fixed the problem? Is this ready to go from your side?

Jedi18:
@msimberg Yes the PR is ready.

hkaiser (approved):
LGTM, thanks!

jefftrull:
Does this mean issue #3733 is resolved :)

hkaiser:
> Does this mean issue #3733 is resolved :)

Yep, it's at least an improvement.

PR #5579, opened 2021-09-25
Split absolute paths for generated pkg-config files into -L/-l parts

severinstrobl:
Absolute paths to shared libraries in the generated pkg-config files are not fully supported by waf-based build systems and are especially problematic in conjunction with `-Wl,--as-needed`. While the pkg-config format does not explicitly mention absolute paths, the 'Libs' entry should contain the 'Required linking flags for this package'.


hkaiser (approved):
LGTM, thanks!

aurianer (approved):
LGTM thanks!

PR #5583, opened 2021-09-28
fix coroutine_traits allocate calls, add unhandled_exception() implementation.

### LINKED ISSUE #5502: C++20 std::coroutine cmake detection
Configuring HPX head on a box with GNU g++ 10.3.0, I am trying to get
```
-- Performing Test HPX_WITH_CXX20_COROUTINES - Failed
```
to pass, but am getting lost in the cmake files.

According to g++'s error message, it seems that a `promise_type` also needs to define `unhandled_exception`, so I applied
```
index 619fb73205..bba9190fac 100644
--- a/cmake/tests/cxx20_coroutines.cpp
+++ b/cmake/tests/cxx20_coroutines.cpp
@@ -30,6 +30,7 @@ struct resumable
         }
 
         void return_value(int value) const noexcept {}
+        void unhandled_exception() const noexcept {}
     };
 
     bool await_ready() const noexcept
```

Copying the patched `cmake/tests/cxx20_coroutines.cpp` to another directory, and adding
```
$ cat CMakeLists.txt 
cmake_minimum_required(VERSION 3.00)
project(coro)
set(CMAKE_CXX_STANDARD 20)
set(CMAKE_CXX_STANDARD_REQUIRED True)
add_executable(coro cxx20_coroutines.cpp)
target_compile_options(coro PUBLIC "-fcoroutines")
```
showed that it compiles successfully.

I tried configuring hpx with
```
cmake \
-DCMAKE_BUILD_TYPE=Release \
-DHPX_WITH_DOCUMENTATION=OFf \
-DHPX_WITH_EXAMPLES=OFF \
-DHPX_WITH_TESTS=OFF \
-DHPX_WITH_NETWORKING=OFF \
-DHPX_WITH_DISTRIBUTED_RUNTIME=OFF \
-DCMAKE_CXX_STANDARD=20 -DHPX_USE_CMAKE_CXX_STANDARD=ON \
..
```
and
```
index fd7ab0dad2..3b5a79bcaf 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1623,6 +1623,8 @@ if(CMAKE_COMPILER_IS_GNUCXX AND HPX_WITH_GCC_VERSION_CHECK)
     hpx_error(
       "GCC 7.0 or higher is required. Specify HPX_WITH_GCC_VERSION_CHECK=OFF to ignore this error."
     )
+  elif(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 10.0)
+      hpx_add_target_compile_option(-fcoroutines PUBLIC)
   endif()
 endif()
 
```
but that didn't do the trick...

Spongman:
re: #5502, #5517

these are the changes required to get my coroutine-using application to build with gcc 10.2 & hpx.

it doesn't fix #5502 (I had to force `HPX_HAVE_CXX20_COROUTINES` in my non-cmake build).

### DIFF
--- a/libs/core/futures/include/hpx/futures/traits/detail/future_await_traits.hpp
+++ b/libs/core/futures/include/hpx/futures/traits/detail/future_await_traits.hpp
@@ -258,6 +258,11 @@ namespace std {
                     this->base_type::set_value(std::forward<U>(value));
                 }
 
+                void unhandled_exception() noexcept
+                {
+                    this->base_type::set_exception(std::current_exception());
+                }
+
                 HPX_NODISCARD HPX_FORCEINLINE static void* operator new(
                     std::size_t size)
                 {
@@ -299,10 +304,15 @@ namespace std {
                     this->base_type::set_value();
                 }
 
+                void unhandled_exception() noexcept
+                {
+                    this->base_type::set_exception(std::current_exception());
+                }
+
                 HPX_NODISCARD HPX_FORCEINLINE static void* operator new(
                     std::size_t size)
                 {
-                    return base_type::allocate();
+                    return base_type::allocate(size);
                 }
 
                 HPX_FORCEINLINE static void operator delete(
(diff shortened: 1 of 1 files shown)


hkaiser (approved):
LGTM, thanks!

PR #5584, opened 2021-09-28
Serialization of hpx::tuple must use hpx::get

severinstrobl:
As HPX can be configured without the compatibility of `hpx::tuple` and `std::tuple`, `hpx::get` must be used for accessing the elements of a `hpx::tuple`.  
This fixes a compilation error with the current master when configured with `-DHPX_DATASTRUCTURES_WITH_ADAPT_STD_TUPLE=OFF`


hkaiser (approved):
Excellent catch!

aurianer (approved):
LGTM thanks a lot!

PR #5588, opened 2021-09-30
Fixing the output formatting for id_types

hkaiser:
This fixes the wrong output reported here: https://gist.github.com/severinstrobl/bfba712aa4a03454e5fbab65bd70046e


severinstrobl:
Thanks, with this change i get the expected output, where different IDs in the format `{0000000x00000000, 0000000000000000}` are printed.

msimberg (dismissed):
Thanks @hkaiser. Would a regression test for this be possible? I suppose we don't expect to at least be all zeros?

aurianer (dismissed):
LGTM thanks!

hkaiser:
@msimberg I have added the test now.

PR #5589, opened 2021-09-30
Add HPX_FORWARD/HPX_MOVE macros

msimberg:
Inspired by https://github.com/boostorg/hana/commit/540f665e5132d75bbf6eda704638622727c0c01c and https://www.foonathan.net/2020/09/move-forward/, this replaces `std::forward` and `std::move` with `static_cast`s. In principle I hate this change, but this gives an approximately 5% speedup in compilation (tested on `for_loop` and `thread_pool_scheduler` tests, which both use forward/move quite heavily). To me this makes the change worth it, but if there are objections please speak up. Especially @hkaiser and @K-ballo, I'd appreciate if you know of any clear downsides to doing this. This was tested with GCC 10. I'd imagine other compilers might see different effects from this.

To do:
- [x] fix all three-argument calls to std::move (should not use `HPX_MOVE`)
- [x] don't use `HPX_MOVE` or `HPX_FORWARD` in tests and examples (this is a purely internal utility)
- [x] decide between `#define HPX_FORWARD(x) static_cast<decltype(x)>(x)` and `#define HPX_FORWARD(T, x) static_cast<T&&>(x)`
  - [x] use the `decltype` version


K-ballo:
The forwarding macro should not take a type. If the type differs from the one deduced from the expression, then the operation is not forwarding but just an arbitrary cast.

msimberg:
> The forwarding macro should not take a type. If the type differs from the one deduced from the expression, then the operation is not forwarding but just an arbitrary cast.

Right, you would prefer `#define HPX_FORWARD(x) static_cast<decltype(x)&&>(x)`, correct? I measured a small improvement staying with explciitly specifying the type, but *that* difference is perhaps not worth it anymore. The main benefit I see with explicitly specifying the type is that if for some magical reason `std::forward` becomes faster in the future, changing back to it will be painless. Then again, if that happens, `std::forward` might not require specifying the type anymore either, so it's not a huge concern. `#define HPX_FORWARD(T, x) static_cast<decltype(x)&&>(x)` is also an option, but a bit weird.

hkaiser:
> Right, you would prefer `#define HPX_FORWARD(x) static_cast<decltype(x)&&>(x)`, correct?

I think just `static_cast<decltype(x)>(x)` would be fine.

msimberg:
> > if someone (@K-ballo?) feels strongly
> 
> I feel strongly

Thanks @K-ballo, I'll go with that then!

msimberg:
One side-effect of this change is that clang-tidy gives false positive warnings of this type:

```
/hpx/source/libs/core/algorithms/include/hpx/parallel/algorithms/partition.hpp:973:37: error: side effects in the 1st macro argument '__VA_ARGS__' are repeated in macro expansion [bugprone-macro-repeated-side-effects,-warnings-as-errors]
                    return HPX_MOVE(blocks_[left_++]);
                                    ^
```
As far as I can tell this is harmless since the first "repetition" is not evaluated, but is inside `decltype`. I've silenced those warnings locally.

msimberg:
This is slowly coming back together. Just fixing the remaining compilation errors, but otherwise this is ready.

msimberg:
retest cscs

msimberg:
@hkaiser do you know if these errors are due to MSVC limitations or me expecting macros to do something they won't do everywhere: https://github.com/STEllAR-GROUP/hpx/runs/4141784764?check_suite_focus=true#step:6:390? Triggered by this: https://github.com/msimberg/hpx/blob/68cf3328cee63edefbdfa51d4eb56c2f6c38c051/libs/core/execution/include/hpx/execution/algorithms/transform.hpp#L55-L61.

msimberg:
> @hkaiser do you know if these errors are due to MSVC limitations or me expecting macros to do something they won't do everywhere: https://github.com/STEllAR-GROUP/hpx/runs/4141784764?check_suite_focus=true#step:6:390? Triggered by this: https://github.com/msimberg/hpx/blob/68cf3328cee63edefbdfa51d4eb56c2f6c38c051/libs/core/execution/include/hpx/execution/algorithms/transform.hpp#L55-L61.

Answering my own question, I was nesting preprocessor directives inside a macro call, which is no good (and I've attempted to do that before as well... so I should've known :/). However, it seems like clang-tidy actually gives a nice warning about it: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/9360/workflows/6dbce8f1-bfe0-4aee-be28-d275bcb4a09a/jobs/273593.

msimberg:
bors merge

aurianer (approved):
LGTM thanks!

msimberg:
bors merge

PR #5592, opened 2021-10-04
Add nth_element

Jedi18:
Thanks to @fjtapia for his sequential nth_element implementation.

Added tag_fallback_dispatch overloads for parallel and container overloads. Added normal and range tests.
Added simple parallel nth_element implementation (till Francisco works on his parallel implementation)

Issue #4822
Issue #3646 
Issue #1668


msimberg:
@Jedi18 FYI, I think the test failures are only due to #5593. I'll get that in and hopefully CI will be a lot happier after that.

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
```diff
@@ -0,0 +1,411 @@
...
+    ///                     The signature of this
+    ///                     comparison function should be equivalent to:
+    ///                     \code
+    ///                     bool cmp(const Type1 &a, const Type2 &b);
+    ///                     \endcode \n
+    ///                     The signature does not need to have const&, but
+    ///                     the function must not modify the objects passed to
+    ///                     it. The type must be such that an object of
+    ///                     type \a randomIt can be dereferenced and then
+    ///                     implicitly converted to Type.
```
hkaiser, line 53: Please mention that `Pred` defaults to `std::less<>`

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
```diff
@@ -0,0 +1,411 @@
...
+        ///
+        template <class RandomIt, typename Compare, typename Proj>
+        void nth_element_seq(RandomIt first, RandomIt nth, RandomIt end,
+            uint32_t level, Compare&& comp, Proj&& proj)
+        {
+            const uint32_t nmin_sort = 24;
+
+            // Check if the iterators are corrects
+            auto nelem = end - first;
+            HPX_ASSERT(nelem >= 0 and (nth - first + 1) > 0 and
```
hkaiser, line 177: Please don't use alternative keywords (`and`).

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
```diff
@@ -0,0 +1,411 @@
...
+        /// \param end : iterator to the element after the last in the range
+        /// \param level : level of depth in the call from the root
+        /// \param comp : object for to Compare elements
+        /// \param proj : projection
+        ///
+        template <class RandomIt, typename Compare, typename Proj>
+        void nth_element_seq(RandomIt first, RandomIt nth, RandomIt end,
+            uint32_t level, Compare&& comp, Proj&& proj)
+        {
+            const uint32_t nmin_sort = 24;
```
hkaiser, line 173: Please use `std::uint32_t` instead. Also please note that we generally try to use 'east const' notation (i.e. `std::uint32_t const`)

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
```diff
@@ -0,0 +1,411 @@
...
+            // Check if the iterators are corrects
+            auto nelem = end - first;
+            HPX_ASSERT(nelem >= 0 and (nth - first + 1) > 0 and
+                (nth - first + 1) <= nelem);
+
+            // Check  the special conditions
+            if (nth == first)
+            {
+                RandomIt it = detail::min_element<RandomIt>().call(
+                    hpx::execution::seq, first, end, comp, proj);
```
hkaiser, line 184: Should we `std::forward` the predicate and the projection here?

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
```diff
@@ -0,0 +1,411 @@
...
+                if (it != first)
+                    std::swap(*it, *first);
```
hkaiser, line 186: `std::iter_swap`?

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
hkaiser: Also, if the assert was moved one level up, this function could be `constexpr`.

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
```diff
@@ -0,0 +1,411 @@
...
+                    RandomIt last_iter =
+                        detail::advance_to_sentinel(first, last);
+                    return_last = last_iter;
+
+                    while (first != last_iter)
+                    {
+                        auto n = detail::distance(first, last_iter);
+
+                        // get random pivot index
+                        auto pivotIndex = std::rand() % n;
```
hkaiser, line 278: Please note that `std::rand()` is not thread-safe. Could we use more sophisticated pivoting techniques instead (like in `hpx::stable_sort`)?

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
```diff
@@ -0,0 +1,411 @@
...
+#if defined(HPX_HAVE_CXX20_STD_RANGES_ITER_SWAP)
+                        std::ranges::iter_swap(first, first + pivotIndex);
+#else
+                        std::iter_swap(first, first + pivotIndex);
+#endif
+
+                        auto partitionResult =
+                            hpx::parallel::v1::detail::partition<RandomIt>()
+                                .call(
+                                    std::forward<ExPolicy>(policy), first + 1,
```
hkaiser, line 289: Wouldn't it be better to always use a synchronous policy here (since we wait for it below anyways)?

libs/core/algorithms/include/hpx/parallel/container_algorithms/nth_element.hpp
```diff
@@ -0,0 +1,421 @@
...
+    ///                     \endcode \n
+    ///                     The signature does not need to have const&, but
+    ///                     the function must not modify the objects passed to
+    ///                     it. The type must be such that an object of
+    ///                     type \a randomIt can be dereferenced and then
+    ///                     implicitly converted to Type.
+    /// \param proj         Specifies the function (or function object) which
+    ///                     will be invoked for each of the elements as a
+    ///                     projection operation before the actual predicate
+    ///                     \a is invoked.
```
hkaiser, line 60: Please mention that `Proj` defaults to identity.

hkaiser (commented):
Very nice!

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
Jedi18: Yeah, so do I always pass in an hpx::execution::par(but then what if the user passes in a par_unseq)? Is there a way to "convert" a parallel asynchronous policy to it's corresponding synchronous one?

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
Jedi18: Oh right yeah, line 193 too since we are returning immediately after there too.

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
Jedi18: Oh ok sure, thanks!

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
hkaiser: We don't have anything ready for this, I believe. One way would be to do:
```
auto non_task_policy = par.on(policy.executor()).with(policy.parameters());
```
This will break with `par_unseq` and similar, however, So it might be better to create a trait that can be specialized for the task policies and that returns the given policy unchanged for all non-task policies.

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
hkaiser: Yes, this is needed in several spots. I have mentioned it once only.

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
Jedi18: Oh ok got it, thanks!

libs/core/executors/include/hpx/executors/execution_policy.hpp
```diff
@@ -1622,5 +1622,19 @@ namespace hpx { namespace detail {
...
     {
     };
+
+    template <typename ExPolicy>
+    constexpr decltype(auto) async_to_sync_policy(ExPolicy&& policy)
+    {
+        if constexpr (hpx::is_async_execution_policy_v<ExPolicy>)
+        {
+            return hpx::execution::par.on(policy.executor())
+                .with(policy.parameters());
```
hkaiser, line 1632: This will always use `par` as the non-task policy. Is that acceptable?

Jedi18: Ok yeah I should also consider if it's called somewhere else using a sequential policy. Btw are there other tasks policies besides seq(task) and par(task)? Also is this location okay or should I place this function somewhere else?

hkaiser: Another option would be to add a `to_non_task()` member (or a `operator()(non_task)`) to all execution policies (similar to the member function that returns a corresponding task policy). Whatever is easier for you.

Jedi18:
This PR should be ready. Also I'm not able to locally reproduce the unique test failure (not related to this PR).

PR #5593, opened 2021-10-04
Update any_sender to use tag_dispatch for execution customizations

msimberg:
Apologies for breaking master. #5510 passed CI before #5332 was merged, but I did not update #5510 to use `tag_dispatch` instead of member functions before merging. This should fix things.


hkaiser (dismissed):
LGTM, thanks!

PR #5594, opened 2021-10-05
Adapt datapar generate

### LINKED ISSUE #2333: Implement datapar for parallel algorithms
- [ ] `adjacent_find` 
- [x] `copy` `copy_n` (#5428)
- [ ] `copy_if` 
- [ ] `move`
- [ ] `equal` `mismatch` 
- [ ] `reduce` 
- [x] `transform` (see #2330)
- [x] `fill` `fill_n` (#5465)
- [ ] `remove` `remove_if` 
- [ ] `rotate` `rotate_copy`
- [ ] `search` `search_n`
- [ ] `inplace_merge` `merge`
- [ ] `swap_ranges`
- [ ] `transform_reduce`
- [ ] `unique` `unique_copy`

srinivasyadav18:
Fixes #2333 

## Proposed Changes

  - Moves sequential generate to detail
  - Adapts datapar generate
  - Splits existing units tests for generate into header and source
  - Adapts unit tests for datapar generate


hkaiser (dismissed):
Otherwise, LGTM, thanks!

hkaiser:
We've had some discussion on how to resolve the problem that the user supplied function (object) has to be generic (i.e. callable for a single value and a vector register) without expecting an argument that could be used to deduce the required return type. We have two possible solutions

- always call the vectorized version of the function object and manually take apart the returned value if needed (during execution of prefix/postfix loops). 
- rely on specifying the required template type explicitly

The former solution is non-conforming, however, as the generator function might be called more often than expected. The latter solution requires for the user to supply a function (object) that was written in a specific way, for instance:

```
// either provide a function object with a templated `operator()()`
struct f
{
    template <typename T>
    T operator()() const
    {
        return T();
    }
};

// or a lambda with an explicit template parameter (note: C++20)
auto l = []<typename T>() -> T { return T(); };

int main() {

    f foo;

    foo.operator()<int>();
    l.operator()<int>();
    
    return 0;
}
```
Also, for the latter, global functions are not usable.

While the second solution is still preferable (it is conforming), I'm uncertain if the limitations imposed on the function object are too restrictive or if the solution is acceptable in terms of usability.

msimberg:
> While the second solution is still preferable (it is conforming), I'm uncertain if the limitations imposed on the function object are too restrictive or if the solution is acceptable in terms of usability.

I also think the second option is preferable. One can always wrap a global function in a lambda/struct if needed.

srinivasyadav18:
> o, for th



> We've had some discussion on how to resolve the problem that the user supplied function (object) has to be generic (i.e. callable for a single value and a vector register) without expecting an argument that could be used to deduce the required return type. We have two possible solutions
> 
> * always call the vectorized version of the function object and manually take apart the returned value if needed (during execution of prefix/postfix loops).
> * rely on specifying the required template type explicitly
> 
> The former solution is non-conforming, however, as the generator function might be called more often than expected. The latter solution requires for the user to supply a function (object) that was written in a specific way, for instance:
> 
> ```
> // either provide a function object with a templated `operator()()`
> struct f
> {
>     template <typename T>
>     T operator()() const
>     {
>         return T();
>     }
> };
> 
> // or a lambda with an explicit template parameter (note: C++20)
> auto l = []<typename T>() -> T { return T(); };
> 
> int main() {
> 
>     f foo;
> 
>     foo.operator()<int>();
>     l.operator()<int>();
>     
>     return 0;
> }
> ```
> 
> Also, for the latter, global functions are not usable.
> 
> While the second solution is still preferable (it is conforming), I'm uncertain if the limitations imposed on the function object are too restrictive or if the solution is acceptable in terms of usability.

Oops!, Did not read this sorry!, I think i pushed the changes for first method. I will look into second option and will try to implement it.

aurianer (dismissed):
LGTM thanks a lot!

libs/core/algorithms/include/hpx/parallel/algorithms/detail/generate.hpp
```diff
@@ -0,0 +1,87 @@
...
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/config.hpp>
+#include <hpx/executors/execution_policy.hpp>
+#include <hpx/functional/tag_fallback_dispatch.hpp>
```
msimberg, line 11: We renamed it to `tag_invoke` again :upside_down_face: sorry...

srinivasyadav18: Np ;) , I will make the required changes and push!

hkaiser:
@srinivasyadav18 which of the builders is actually running the vectorized tests?

hkaiser (approved):
LGTM, thanks!

srinivasyadav18:
> g the vectorized t

gcc-11-debug and gcc-11-debug

hkaiser:
> > g the vectorized t
> 
> gcc-11-debug and gcc-11-release

Thanks! I'm wondering how the test can actually pass on that platform now. The tests use a non-templated lambda (e.g. https://github.com/STEllAR-GROUP/hpx/pull/5594/files#diff-94177c5266a2c92106d9f6e2640962f9bf6289153a24edee615122d7b10dbc4aR35), but the code is invoking the templated operator() (e.g. https://github.com/STEllAR-GROUP/hpx/pull/5594/files#diff-c8d8a6871e54681264b783160ea771babc0a46e669dafa47c5419841dda7a09bR49). Are you sure the tests are running?

srinivasyadav18:
> > > g the vectorized t
> > 
> > 
> > gcc-11-debug and gcc-11-release
> 
> Thanks! I'm wondering how the test can actually pass on that platform now. The tests use a non-templated lambda (e.g. https://github.com/STEllAR-GROUP/hpx/pull/5594/files#diff-94177c5266a2c92106d9f6e2640962f9bf6289153a24edee615122d7b10dbc4aR35), but the code is invoking the templated operator() (e.g. https://github.com/STEllAR-GROUP/hpx/pull/5594/files#diff-c8d8a6871e54681264b783160ea771babc0a46e669dafa47c5419841dda7a09bR49). Are you sure the tests are running?

I updated with the changes to test files in this https://github.com/STEllAR-GROUP/hpx/pull/5594/commits/12a28536c76fdef341fd48547a755a6b2eedecc7 commit.

hkaiser:
> > > > g the vectorized t
> > > 
> > > 
> > > gcc-11-debug and gcc-11-release
> > 
> > 
> > Thanks! I'm wondering how the test can actually pass on that platform now. The tests use a non-templated lambda (e.g. https://github.com/STEllAR-GROUP/hpx/pull/5594/files#diff-94177c5266a2c92106d9f6e2640962f9bf6289153a24edee615122d7b10dbc4aR35), but the code is invoking the templated operator() (e.g. https://github.com/STEllAR-GROUP/hpx/pull/5594/files#diff-c8d8a6871e54681264b783160ea771babc0a46e669dafa47c5419841dda7a09bR49). Are you sure the tests are running?
> 
> I updated with the changes to test files in this [12a2853](https://github.com/STEllAR-GROUP/hpx/commit/12a28536c76fdef341fd48547a755a6b2eedecc7) commit.

Hmmm. From what I can see those changes are not on the PR, are they?

srinivasyadav18:
> > > > > g the vectorized t
> > > > 
> > > > 
> > > > gcc-11-debug and gcc-11-release
> > > 
> > > 
> > > Thanks! I'm wondering how the test can actually pass on that platform now. The tests use a non-templated lambda (e.g. https://github.com/STEllAR-GROUP/hpx/pull/5594/files#diff-94177c5266a2c92106d9f6e2640962f9bf6289153a24edee615122d7b10dbc4aR35), but the code is invoking the templated operator() (e.g. https://github.com/STEllAR-GROUP/hpx/pull/5594/files#diff-c8d8a6871e54681264b783160ea771babc0a46e669dafa47c5419841dda7a09bR49). Are you sure the tests are running?
> > 
> > 
> > I updated with the changes to test files in this [12a2853](https://github.com/STEllAR-GROUP/hpx/commit/12a28536c76fdef341fd48547a755a6b2eedecc7) commit.
> 
> Hmmm. From what I can see those changes are not on the PR, are they?

The changes are present. 
Previously there was only test source files generate.cpp and generate_n.cpp in `tests/unit/algorithms`.
I had split these 2 ` .cpp` source files into `.hpp` and `.cpp` i.e generate.cpp and generate.hpp in https://github.com/STEllAR-GROUP/hpx/pull/5594/commits/bbd2ebc5ebadd8532c914c3fd1f1c4cb146897b6.
Later in this https://github.com/STEllAR-GROUP/hpx/pull/5594/commits/12a28536c76fdef341fd48547a755a6b2eedecc7 commit I created seperate generate.cpp and generate.hpp in `tests/unit/datapar_algorithms` because we need seperate implementations for datapar tests.

You can find the templated functions for datapar tests in then end of commit differences from all commits i.e in tests/unit/datapar_algorithms/.

hkaiser:
Let's get this merged, then. Thanks!

msimberg:
> > g the vectorized t
> 
> gcc-11-debug and gcc-11-release

In principle also CircleCI (`build-and-test`). However, for some reason which I still don't understand that never runs for @srinivasyadav18... It runs for other forked repositories, so it's not that. @srinivasyadav18 have you done something to upset CircleCI? ;)

PR #5595, opened 2021-10-05
Drop semicolons for macro calls

severinstrobl:
Depending on the compiler flags, superfluous semicolons after macro calls result in compiler warnings.


msimberg:
Nice, thanks! Is this from `-Wextra-semi`? If yes, would you mind adding it e.g. here: https://github.com/STEllAR-GROUP/hpx/blob/92ab22eb6efc9215fa57092b3304ee1330493b29/CMakeLists.txt#L1918?

severinstrobl:
The code actually uses the flag `-Wpedantic`, do you want me to add it or is this too much?

msimberg:
> The main issue with this is that so far we always had the explicit semicolon after using one of those macros, e.g.:
> 
> ```
> HPX_REGISTER_COMPONENT_MODULE();
> ```
> 
> The idea was to make it look more consistent with other (non-macro based) C++ code. I have no real problem with removing those semicolons consistently, we should however agree on such a (mostly cosmetic) change.

@severinstrobl do you have an opinion on this? When you were making the changes, do you remember how straightforward it would've been to make the opposite change, i.e. remove semicolons from the macro definitions? Or would that have been difficult for some reason?

severinstrobl:
Please excuse the delayed response. I agree that making the macro calls look more like regular function calls has a certain appeal. Unfortunately adding/enforcing semicolons for all macro calls is a bit tricky, as can be seen at the example of [`HPX_SERIALIZATION_SPLIT_FREE` ](https://github.com/STEllAR-GROUP/hpx/blob/00634053054af6141567210db5de4fbc849ddfd7/libs/core/serialization/include/hpx/serialization/serialization_fwd.hpp#L50). One (horribly ugly) approach would be to add a forward declaration for a dummy struct at the end of the inserted code to enforce the usage of a semicolon. But as this would break existing code just for the benefit of getting rid of some warnings, I don't think this is a reasonable path. So unless there are other ideas, I guess there is currently no easy way to resolve the inconsistency in the macro calls and the warnings triggered by the stray semicolons.

msimberg:
@severinstrobl I'm just checking in to see if you'll have time soon to finish this off? If not, no problem at all. I'd like to see this in soon and can try to finish off the remaining warnings if you don't mind.

severinstrobl:
> @severinstrobl I'm just checking in to see if you'll have time soon to finish this off? If not, no problem at all. I'd like to see this in soon and can try to finish off the remaining warnings if you don't mind.

It is still on my to-do list, sorry for letting this one sit around for so long. I'll try to sort this out by the end of the week.

msimberg:
> > @severinstrobl I'm just checking in to see if you'll have time soon to finish this off? If not, no problem at all. I'd like to see this in soon and can try to finish off the remaining warnings if you don't mind.
> 
> It is still on my to-do list, sorry for letting this one sit around for so long. I'll try to sort this out by the end of the week.

No worries, and thanks! Just let us know if you won't have time for it and we'll happily take over.

severinstrobl:
Hopefully this time the regex got all of the remaining stray semicolons :crossed_fingers:

severinstrobl:
Not sure about the failing checks as I cannot access the logs, are they related to the changes here?

msimberg:
> Not sure about the failing checks as I cannot access the logs, are they related to the changes here?

Most likely nothing to do with these changes. I'm still waiting for `build-and-test`, but if that comes back ok this should be ready to merge. Thanks again for all the work on this! It's really nice to have `-Wpedantic` there :)

msimberg (approved):
Thanks @severinstrobl!

PR #5601, opened 2021-10-08
Test case for any_sender should use hpx::tuple

severinstrobl:
According to the documentation, the tuple type passed to `hpx::util::invoke_fused` must be compatible to `hpx::get`.
Also disabled unit test for structured binding in case compatibility between `hpx::tuple` and `std::tuple` is disabled.


msimberg:
Thanks, good catch! I've also opened https://github.com/STEllAR-GROUP/hpx/pull/5602 to make sure we test this in the future.

PR #5604, opened 2021-10-08
Make use of shell-session to allow non-copyable $

aurianer:
Add "$" to shell commands to differentiate them from the output displayed in the terminal.

Note: the $ is not copied if the user wants to copy the code block

@dimitraka what do you think?

docs


dimitraka:
Exactly what we wanted, seems fine!

msimberg:
Really neat, thanks @aurianer!

PR #5605, opened 2021-10-11
Fix usage of HPX_INVOKE macro

severinstrobl:
`HPX_INVOKE` is a variadic macro and as such requires at least one argument for the '...' part to be strictly standard-compliant, at least pre-C++20.


severinstrobl:
> @severinstrobl could you please fix the clang-format errors?

Absolutely, completely forgot about running `clang-format` again.

PR #5611, opened 2021-10-16
Increase parallelism ctest if possible

aurianer:


aurianer:
Looks like this speeds things up

PR #5580, opened 2021-09-25
Adapt simd adj diff

srinivasyadav18:
Fixes #2333 

## Proposed Changes

  - Moves sequential implementation of adjacent_difference to detail which enables overload for simd policy.
  - Adds new header in datapar for adjacent difference algorithm.
  - Splits the existing adjacent_difference into header and source files and merges exception and bad alloc tests into one.
  - Adds adjacent difference datapar unit test.

PR #5613, opened 2021-10-18
Fix Vc datapar adjacent_difference

srinivasyadav18:
Vc fails with `std::size_t `when using `std::minus<>` in adjacent difference algorithm.

Fixes #5580 
## Proposed Changes

  - Replaces `std::size_t` with `int`


aurianer:
Thanks a lot!

PR #5616, opened 2021-10-19
Remove separately building segmented algorithms on CircleCI

msimberg:
They were being built as part of the tests.unit target already, so it's unnecessary. They were originally separated to reduce memory requirements, but that seems to no longer be a problem.


msimberg:
> Have we been building those twice all the time - silly us!

No no! Once it's built it's built. The second call to ninja just doesn't do anything since `tests.unit.modules.segmented_algorithms`
is a subset of `tests.unit`.

PR #5617, opened 2021-10-19
Use advance_and_get_distance where required

Jedi18:
Use advance_and_get_distance utility to get the distance and last iterator instead of using two seperate calls to distance and advance_to_sentinel.


hkaiser:
Looks like this is ok now. Please review.

PR #5625, opened 2021-10-26
Revert "Use advance_and_get_distance where required"

msimberg:
Opening primarily to verify that #5617 actually does break the CUDA examples/tests.


Jedi18:
The `tests.unit.modules.compute_cuda.for_each_compute` and `tests.unit.modules.compute_cuda.for_loop_compute` are the ones failing right? Are there any other failing tests?

The for_loop_compute test fail seems to be linked to hpx::copy (replacing it with std::copy fixed the error), so I'll try to figure out what's wrong there.

msimberg:
Closing this since #5677 "fixes" the failing tests after #5617 by removing them.

PR #5628, opened 2021-11-01
Add Clang 13 CI configuration

msimberg:
- Fixes unused variable warnings reported by Clang 13 (it now detects transitively unused variables).
- Don't use `std::iterator` in `prefetching_iterator`.
- Explicitly default `counting_iterator` constructors.
- Removed `oldest/newest` CI configuration names and replaced them with explicit version numbers (`newest/oldest` were originally intended to be the only configurations; now there are many more so it makes more sense to use the explicit version number).


hkaiser (approved):
LGTM, thanks!

PR #5632, opened 2021-11-03
Tell Asio to use std::aligned_new only if available

hkaiser:


msimberg (approved):
Does Asio misdetect it by itself (it wouldn't be the only thing it misdetects...)?

hkaiser:
> Does Asio misdetect it by itself (it wouldn't be the only thing it misdetects...)?

I'm not sure. We ran into this when building Phylanx... I think the change proposed does not hurt in any case.

msimberg:
Possibly this: https://github.com/chriskohlhoff/asio/issues/922, or at least related?

PR #5633, opened 2021-11-03
Moving packaged_task to module futures

hkaiser:
- flyby: move tests related to local futures to futures module

### DIFF
--- a/libs/core/futures/tests/unit/local_use_allocator.cpp
+++ b/libs/core/futures/tests/unit/local_use_allocator.cpp
@@ -1,4 +1,4 @@
-//  Copyright (c) 2007-2017 Hartmut Kaiser
+//  Copyright (c) 2007-2021 Hartmut Kaiser
 //  Copyright (C) 2011 Vicente J. Botet Escriba
 //
 //  SPDX-License-Identifier: BSL-1.0
@@ -32,4 +32,6 @@ int main()
                       test_allocator<int>>::value,
         "std::uses_allocator<local::packaged_task<int()>, "
         "test_allocator<int> >::value");
+
+    return 0;
 }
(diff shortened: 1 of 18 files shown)


libs/core/futures/tests/unit/local_use_allocator.cpp
```diff
@@ -32,4 +32,6 @@ int main()
                       test_allocator<int>>::value,
         "std::uses_allocator<local::packaged_task<int()>, "
         "test_allocator<int> >::value");
+
+    return 0;
```
msimberg, line 36: Not really necessary.

msimberg (commented):
Is the move just a preference or is there a more fundamental reason to put it in the futures module?

libs/core/futures/tests/unit/local_use_allocator.cpp
hkaiser: I know :/ I usually add it for consistency anyways.

msimberg: I suspected you knew :P Maybe someone who didn't know before knows now... I don't mind keeping the explicit return.

hkaiser:
This is ready to go in.

msimberg:
I've nothing against this changed (it's merged), but I'm still looking for an answer to this.

> Is the move just a preference or is there a more fundamental reason to put it in the futures module?

hkaiser:
> I've nothing against these changes but I'm still looking for an answer to this:
> 
> > Is the move just a preference or is there a more fundamental reason to put it in the futures module?

Sorry, I didn't notice your question. There are two reasons that motivated the change:
- packaged_task belongs into the same group of types as future and promise, so I thought we should have it moved
- during some experimentation with executors I wanted to use `packaged_task`, but that created circular dependencies with the executor module - moving the file helped resolving those

PR #5641, opened 2021-11-10
Factorize perftests scripts

aurianer:


.jenkins/cscs-perftests/batch.sh
```diff
@@ -49,48 +47,11 @@ configure_build_errors=0
...
-
-  # Run performance tests
-  ${perftests_dir}/driver.py -v -l $logfile_tmp perftest run --local True \
-      --run_output $result --targets-and-opts "${run_command[@]}" \
-      || { echo 'Running failed'; test_errors=1; exit 1; }
-
-  index=$((index+1))
-done
+# Synchronize after the asynchronous copy from the source dir
+wait
```
msimberg, line 51: I learned about `wait` from here, very cool! I wonder if it would be clearer to put this closer to the backgrounded copies? Or does the virtualenv `activate` take a long time so it's worth it to put it after it?

msimberg (commented):
This is very nice, thanks!

.jenkins/cscs-perftests/batch.sh
aurianer: yes it's quite long in my experience, but I don't mind so much if you prefer having it before

aurianer:
I think this is good to go :)

msimberg (approved):
Thanks!

msimberg:
bors merge

PR #5637, opened 2021-11-09
Use bors for merging pull requests

msimberg:
Bors is meant to avoid this problem: https://bors.tech/essay/2017/02/02/pitch/, which happened (again) yesterday.

This will not be very useful until CI actually passes reliably, but I'm starting by setting it up for a subset, i.e. one single GitHub action on Linux that builds and runs the examples (i.e. the current "Linux CI (Debug)" configuration). Eventually all CI should just report to bors, but I'm keeping it separate for now until we get some experience with how well this works.

PR #5642, opened 2021-11-10
Adapting adjacent_difference to work with proxy iterators

hkaiser:
- fixing assignment operator for proxy objects returned from segmented iterators
- adding proxy_value trait allowing to extract type returned by a proxy
- flyby: modernize adjacent_difference algorithm, adding missing concept checks

### DIFF
--- a/components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_segmented_iterator.hpp
+++ b/components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_segmented_iterator.hpp
@@ -157,7 +157,9 @@ namespace hpx { namespace segmented {
                 return *(it_.get_data()->begin() + it_.get_local_index());
             }
 
-            template <typename T_>
+            template <typename T_,
+                typename Enable = std::enable_if_t<!std::is_same_v<
+                    std::decay_t<T_>, local_vector_value_proxy>>>
             local_vector_value_proxy& operator=(T_&& value)
             {
                 if (!it_.get_data())
@@ -214,7 +216,9 @@ namespace hpx { namespace segmented {
                 return v_.get_value(launch::sync, index_);
             }
 
-            template <typename T_>
+            template <typename T_,
+                typename Enable = std::enable_if_t<
+                    !std::is_same_v<std::decay_t<T_>, vector_value_proxy>>>
             vector_value_proxy& operator=(T_&& value)
             {
                 v_.set_value(launch::sync, index_, std::forward<T_>(value));
@@ -1230,8 +1234,21 @@ namespace hpx { namespace traits {
     };
 
     template <typename T, typename Data>
+    struct proxy_value<
+        hpx::segmented::detail::local_vector_value_proxy<T, Data>>
+    {
+        using type = T;
+    };
+
+    template <typename T, typename Data>
     struct is_value_proxy<hpx::segmented::detail::vector_value_proxy<T, Data>>
       : std::true_type
     {
+    };
+
+    template <typename T, typename Data>
+    struct proxy_value<hpx::segmented::detail::vector_value_proxy<T, Data>>
+    {
+        using type = T;
     };
 }}    // namespace hpx::traits
--- a/libs/core/algorithms/include/hpx/algorithms/traits/is_value_proxy.hpp
+++ b/libs/core/algorithms/include/hpx/algorithms/traits/is_value_proxy.hpp
@@ -8,11 +8,27 @@
 
 #pragma once
 
+#include <hpx/config.hpp>
+
 #include <type_traits>
 
 namespace hpx { namespace traits {
+
     template <typename T>
     struct is_value_proxy : std::false_type
     {
     };
+
+    template <typename T>
+    HPX_INLINE_CONSTEXPR_VARIABLE bool is_value_proxy_v =
+        is_value_proxy<T>::value;
+
+    template <typename T>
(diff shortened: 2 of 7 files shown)


components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_segmented_iterator.hpp
```diff
@@ -1230,8 +1234,21 @@ namespace hpx { namespace traits {
...
+    struct is_value_proxy<hpx::segmented::detail::vector_value_proxy<T, Data>>
+    {
+        using type = T;
+    };
+
+    template <typename T, typename Data>
+    struct proxy_value<hpx::segmented::detail::vector_value_proxy<T, Data>>
+    {
+        using type = T;
+    };
```
msimberg, line 1253: Am I misunderstanding these traits or should this be:
```suggestion
    struct proxy_value<hpx::segmented::detail::local_vector_value_proxy<T, Data>>
    {
        using type = T;
    };

    template <typename T, typename Data>
    struct is_value_proxy<hpx::segmented::detail::vector_value_proxy<T, Data>>
      : std::true_type
    {
    };

    template <typename T, typename Data>
    struct proxy_value<hpx::segmented::detail::vector_value_proxy<T, Data>>
    {
        using type = T;
    };
```
?

Or should `is_value_proxy` simply be `true_type` for `T` and `Data` if and only if `proxy_value<T, Data>::type` exists?

msimberg (commented):
One question about the traits. Otherwise this looks good to me. Thanks for looking into this!

components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_segmented_iterator.hpp
hkaiser: `is_value_proxy<T>` checks whether `T` is a type that is used as a proxy. `proxy_value<T>` allows to deduce the type that is returned by a proxy (`is_value_proxy<T>` is true). If `is_value_proxy<T>`  evaluates to false, then `proxy_value<T>` returns `T`.

msimberg: All right, that sounds like what I imagined. But then my first question still stands. Currently `struct proxy_value< hpx::segmented::detail::local_vector_value_proxy<T, Data>>` inherits from `true_type` when I would expect it to have a nested typedef `using type = T`. Likewise, `struct is_value_proxy<hpx::segmented::detail::vector_value_proxy<T, Data>>` has a nested typedef `using type = T` when I would expect it to inherit from `true_type`. In other words, I think there's been some sort mistaken transposition...

hkaiser: I'm confused. The idea is for `is_value_proxy<T>` to be a boolean trait usable in an enable_if (or similar). For this reason it is derived from `true_type` or `false_type` depending on whether `T` is a proxy type or not. Note that `true_type` and `false_type` have an embedded `using type = ...;` referring to themselves.

The type `proxy_value<T>` is not meant to be used as a boolean trait, rather to extract the type returned by the proxy. It 'returns' the type that is provided by the proxy (if `is_value_proxy<T>` would return true). For usability reasons, it defaults to simply 'returning' `T` for non-proxy types (if `is_value_proxy<T>` would return false).

hkaiser: Uhh, I think I see now what you're referring to. Yes - good catch, I messed it up. Will fix.

hkaiser: That's fixed now. Thanks again for catching the issue.

hkaiser:
@msimberg we should go ahead and merge this asap as other PRs are failing because of the problems fixed here.

msimberg (approved):

hkaiser:
bors merge

msimberg:
> This PR was included in a batch that successfully built, but then failed to merge into master. It will not be retried.
> 
> Additional information:
> 
> ```json
> {"message":"2 of 2 required status checks are expected.","documentation_url":"https://docs.github.com/articles/about-protected-branches"}
> ```

Hrm :/ Those should've been done already (I've only required `cscs/gcc-10` for now). Not quite sure what's happening here. Retrying...

bors merge

msimberg:
> This PR was included in a batch that successfully built, but then failed to merge into master. It will not be retried.
> 
> Additional information:
> 
> ```json
> {"message":"2 of 2 required status checks are expected.","documentation_url":"https://docs.github.com/articles/about-protected-branches"}
> ```

Boo, not quite what I was expecting... Merging manually. Maybe it's trying to tell us something useful, but if it is I don't know what it is. Merging manually...

hkaiser:
@msimberg I think it happens because we protected the master branch and bors first merges things into the branch `staged` and then tries to directly commit to master. We might have to add bors to the admins of this repo (not sure if this is a good idea, however, but worth a try).

msimberg:
That should not be the case anymore (see my failed attempts here: https://github.com/STEllAR-GROUP/hpx/pull/5637/#issuecomment-964971164). It's already merged two PRs...

PR #5643, opened 2021-11-11
Update performance test references

msimberg:
I don't think there's been an actual regression, so I'm updating these to what will hopefully be more representative references. There seems to be enough variability that I may simply have been "lucky" with the previous references.


msimberg:
bors merge

aurianer (approved):
LGTM thanks

aurianer:
bors merge

PR #5645, opened 2021-11-11
Add matrix multiplication example

dimitraka:
## Proposed Changes

  - matrix multiplication example
  - matrix multiplication explanation

### DIFF
--- a/docs/sphinx/examples.rst
+++ b/docs/sphinx/examples.rst
@@ -23,6 +23,7 @@ powerful |hpx| concepts.
    :maxdepth: 1
 
    examples/fibonacci_local
+   examples/matrix_multiplication
    examples/fibonacci
    examples/hello_world
    examples/accumulator
--- a/docs/sphinx/examples/matrix_multiplication.rst
+++ b/docs/sphinx/examples/matrix_multiplication.rst
@@ -1,0 +1,119 @@
+..
+    Copyright (C) 2021 Dimitra Karatza
+
+    SPDX-License-Identifier: BSL-1.0
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+.. _examples_matrix_multiplication:
+
+===================
+Parallel algorithms
+===================
+
+This program will perform a matrix multiplication in parallel. The output will look something like this:
+
+.. code-block:: text
+
+   Matrix A is : 
+   4 9 6 
+   1 9 8 
+
+   Matrix B is : 
+   4 9 
+   6 1 
+   9 8 
+
+   Resultant Matrix is : 
+   124 93 
+   111 127 
+
+Setup
+=====
+
+The source code for this example can be found here:
+:download:`matrix_multiplication.cpp
+<../../examples/quickstart/matrix_multiplication.cpp>`.
+
+To compile this program, go to your |hpx| build directory (see
+:ref:`hpx_build_system` for information on configuring and building |hpx|) and
+enter:
+
+.. code-block:: shell-session
+
+   $ make examples.quickstart.matrix_multiplication
+
+To run the program type:
+
+.. code-block:: shell-session
+
+   $ ./bin/matrix_multiplication
+
+or:
+
+.. code-block:: shell-session
+
+   $ ./bin/matrix_multiplication --n 2 --m 3 --k 2 --s 100 --l 0 --u 10 
+
+where the first matrix is `n` x `m` and the second `m` x `k`, s is the seed for creating the random values of 
+the matrices and the range of these values is [l,u] 
+
+This should print:
+
+.. code-block:: text
+
+   Matrix A is : 
+   4 9 6 
+   1 9 8 
+
+   Matrix B is : 
+   4 9 
+   6 1 
+   9 8 
+   
+   Resultant Matrix is : 
+   124 93 
+   111 127 
+
+Notice that the numbers may be different because of the random initialization of the matrices.
+
+Walkthrough
+===========
+
+Now that you have compiled and run the code, let's look at how the code works.
+
+First, ``main()`` is used to initialize the runtime system and pass the command line arguments to the program. 
+``hpx::init`` calls ``hpx_main()`` after setting up HPX, which is where our program is implemented.
+
+.. literalinclude:: ../../examples/quickstart/matrix_multiplication.cpp
+   :language: c++
+   :start-after: //[mul_main
+   :end-before: //]
+
+Proceeding to the ``hpx_main()`` function, we can see that matrix multiplication can be done very easily. 
+
+.. literalinclude:: ../../examples/quickstart/matrix_multiplication.cpp
+   :language: c++
+   :start-after: //[mul_hpx_main
+   :end-before: //]
+
+First, the dimensions of the matrices are defined. If they were not given as command-line arguments, their default 
+values are `2` x `3` for the first matrix and `3` x `2` for the second. We use standard vectors to define the matrices 
+to be multiplied as well as the resultant matrix. 
+
+To give some random initial values to our matrices, we use `std::uniform_int_distribution
+<https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution>`_. Then, ``std::bind()`` is used
+along with ``hpx::ranges::generate()`` to yield two matrices A and B, which contain values in the range of [0, 10] or in
+the range defined by the user at the command-line arguments. The seed to generate the values can also be defined by the user.
+
+The next step is to perform the matrix multiplication in parallel. This can be done by just using an :cpp:func:`\hpx::for_loop` 
+combined with a parallel execution policy ``hpx::execution::par`` as the outer loop of the multiplication. Note that the execution 
+of :cpp:func:`\hpx::for_loop` without specifying an execution policy is equivalent to specifying ``hpx::execution::seq`` 
+as the execution policy.
+
+Finally, the matrices A, B that are multiplied as well as the resultant matrix R are printed using the following function.
+
+.. literalinclude:: ../../examples/quickstart/matrix_multiplication.cpp
+   :language: c++
+   :start-after: //[mul_print_matrix
+   :end-before: //]
--- a/examples/quickstart/matrix_multiplication.cpp
+++ b/examples/quickstart/matrix_multiplication.cpp
@@ -1,0 +1,126 @@
+////////////////////////////////////////////////////////////////////////////////
+//  Copyright (c) 2021 Dimitra Karatza
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+////////////////////////////////////////////////////////////////////////////////
+
+// Parallel implementation of matrix multiplication
+
+#include <hpx/local/algorithm.hpp>
+#include <hpx/local/execution.hpp>
+#include <hpx/local/init.hpp>
+
+#include <cstddef>
+#include <iostream>
+#include <random>
+#include <vector>
+
+std::mt19937 gen;
+
+///////////////////////////////////////////////////////////////////////////////
+//[mul_print_matrix
+void print_matrix(
+    std::vector<int> M, std::size_t rows, std::size_t cols, const char* message)
+{
+    std::cout << "\nMatrix " << message << " is:" << std::endl;
+    for (std::size_t i = 0; i < rows; i++)
+    {
+        for (std::size_t j = 0; j < cols; j++)
+            std::cout << M[i * cols + j] << " ";
+        std::cout << "\n";
+    }
+}
+//]
+
+///////////////////////////////////////////////////////////////////////////////
+//[mul_hpx_main
+int hpx_main(hpx::program_options::variables_map& vm)
+{
+    using element_type = int;
+
+    // Define matrix sizes
+    std::size_t rowsA = vm["n"].as<std::size_t>();
+    std::size_t colsA = vm["m"].as<std::size_t>();
+    std::size_t rowsB = colsA;
+    std::size_t colsB = vm["k"].as<std::size_t>();
+    std::size_t rowsR = rowsA;
+    std::size_t colsR = colsB;
+
+    // Initialize matrices A and B
+    std::vector<int> A(rowsA * colsA);
+    std::vector<int> B(rowsB * colsB);
+    std::vector<int> R(rowsR * colsR);
+
+    // Define seed
+    unsigned int seed = std::random_device{}();
+    if (vm.count("seed"))
+        seed = vm["seed"].as<unsigned int>();
+
+    gen.seed(seed);
+    std::cout << "using seed: " << seed << std::endl;
+
+    // Define range of values
+    std::size_t lower = vm["l"].as<std::size_t>();
+    std::size_t upper = vm["u"].as<std::size_t>();
+
+    // Matrices have random values in the range [lower, upper]
+    std::uniform_int_distribution<element_type> dis(lower, upper);
+    auto generator = std::bind(dis, gen);
+    hpx::ranges::generate(A, generator);
+    hpx::ranges::generate(B, generator);
+
+    // Perform matrix multiplication
+    hpx::for_loop(hpx::execution::par, 0, rowsA, [&](auto i) {
+        hpx::for_loop(0, colsB, [&](auto j) {
+            R[i * colsR + j] = 0;
+            hpx::for_loop(0, rowsB, [&](auto k) {
+                R[i * colsR + j] += A[i * colsA + k] * B[k * colsB + j];
+            });
+        });
+    });
+
+    // Print all 3 matrices
+    print_matrix(A, rowsA, colsA, "A");
+    print_matrix(B, rowsB, colsB, "B");
+    print_matrix(R, rowsR, colsR, "R");
+
+    return hpx::local::finalize();
+}
+//]
+
+///////////////////////////////////////////////////////////////////////////////
+//[mul_main
+int main(int argc, char* argv[])
+{
+    using namespace hpx::program_options;
+    options_description cmdline("usage: " HPX_APPLICATION_STRING " [options]");
+    // clang-format off
+    cmdline.add_options()
+        ("n",
+        hpx::program_options::value<std::size_t>()->default_value(2),
+        "Number of rows of first matrix")
+        ("m",
+        hpx::program_options::value<std::size_t>()->default_value(3),
+        "Number of columns of first matrix (equal to the number of rows of "
+        "second matrix)")
+        ("k",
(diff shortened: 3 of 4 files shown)


docs/sphinx/examples.rst
```diff
@@ -22,6 +22,7 @@ powerful |hpx| concepts.
 .. toctree::
    :maxdepth: 1
 
+   examples/matrix_multiplication
    examples/fibonacci_local
```
msimberg, line 26:
```suggestion
   examples/fibonacci_local
   examples/matrix_multiplication
```

Putting plain async first would be my preference, but I can see the case for the other way around as well. @hkaiser, @aurianer, @gonidelis, @NK-Nikunj preferences?

docs/sphinx/examples/matrix_multiplication.rst
```diff
@@ -0,0 +1,107 @@
...
+
+.. code-block:: shell-session
+
+   $ make examples.quickstart.matrix_multiplication
+
+To run the program type:
+
+.. code-block:: shell-session
+
+   $ ./bin/matrix_multiplication
```
msimberg, line 50: I think it'd be nice if this took the size of the matrices as command line options. The defaults can very well be 2x3 and 3x2 so you can leave the line above as it is, but if someone wants to play with the example it'd be nice to be able to change the sizes.

Defining three command-line options, `--n`, `--m`, `--k`, for the dimensions should be ok, where the first matrix is `n` x `m` and the second `m` x `k`.

docs/sphinx/examples/matrix_multiplication.rst
```diff
@@ -0,0 +1,107 @@
...
+
+First, ``main()`` is used to initialize the runtime system and pass the command line arguments to the program. 
+``hpx::init`` calls ``hpx_main()`` after setting up HPX, which is where our program is implemented.
+
+.. literalinclude:: ../../examples/quickstart/matrix_multiplication.cpp
+   :language: c++
+   :start-after: //[mul_main
+   :end-before: //]
+
+Going on with the ``hpx_main()`` function, we can see that matrix multiplication can be done very easily.
```
msimberg, line 84:
```suggestion
Continuing with the ``hpx_main()`` function, we can see that matrix multiplication can be done very easily. 
```

docs/sphinx/examples/matrix_multiplication.rst
```diff
@@ -0,0 +1,107 @@
...
+   :end-before: //]
+
+Going on with the ``hpx_main()`` function, we can see that matrix multiplication can be done very easily. 
+
+.. literalinclude:: ../../examples/quickstart/matrix_multiplication.cpp
+   :language: c++
+   :start-after: //[mul_hpx_main
+   :end-before: //]
+
+First, the dimentions of the matrices is defined, so as that the number of columns of matrix A is equal
```
msimberg, line 91:
```suggestion
First, the dimensions of the matrices is defined, so as that the number of columns of matrix A is equal 
```

though this might need a small rewrite if the dimensions are given as command-line options.

docs/sphinx/examples/matrix_multiplication.rst
```diff
@@ -0,0 +1,107 @@
...
+   :end-before: //]
+
+First, the dimentions of the matrices is defined, so as that the number of columns of matrix A is equal 
+to the number of rows of matrix B. You can freely change the dimentions as you wish. We use standard vectors 
+to define the matrices to be multipled as well as the resultant matrix.
+
+To give some random initial values to our matrices, we use `std::uniform_real_distribution
+<https://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution>`_. Then, ``hpx::generate()`` is used
+along with ``std::bind()`` as an argument to yield two matrices A and B, which contain values in the range of [0, 10]. 
+You can also freely change the range of the values, if you wish.
```
msimberg, line 98: Continuing with the command-line options, at least the seed should also be a command-line option, possibly also the range of values. You can seed the random number generator with something like this: https://github.com/STEllAR-GROUP/hpx/blob/dd3f3d9ae87b93c383877f4f69dc97842c2f8387/libs/core/async_cuda/tests/unit/cublas_matmul.cpp#L269-L273.

docs/sphinx/examples/matrix_multiplication.rst
```diff
@@ -0,0 +1,107 @@
...
+
+The next step is to perform the matriux multiplication in parallel. This can be done very easily by just using an
```
msimberg, line 100:
```suggestion
The next step is to perform the matrix multiplication in parallel. This can be done very easily by just using an 
```

msimberg:
```suggestion
The next step is to perform the matriux multiplication in parallel. This can be done by just using an 
```
only my preference... I'd hope that the simplicity speaks for itself :P

examples/quickstart/matrix_multiplication.cpp
```diff
@@ -0,0 +1,87 @@
...
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+////////////////////////////////////////////////////////////////////////////////
+
+// Parallel implementation of matrix multiplication
+
+#include <hpx/local/algorithm.hpp>
+#include <hpx/local/future.hpp>
```
msimberg, line 12:
```suggestion
#include <hpx/local/execution.hpp>
```
for `hpx::execution::seq/par`. `hpx/local/future.hpp` is not needed.

examples/quickstart/matrix_multiplication.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+//[mul_hpx_main
+int hpx_main()
+{   
+    int rowsA, colsA, rowsB, colsB;
+    rowsA = 2;
+    colsA = 3;
+    rowsB = colsA;
+    colsB = 2;
```
msimberg, line 27: See earlier comments about command line options.

examples/quickstart/matrix_multiplication.cpp
```diff
@@ -0,0 +1,87 @@
...
+    
+    // Initialize matrices A and B
+    std::vector<int> A(rowsA * colsA);
+    std::vector<int> B(rowsB * colsB);
+    std::vector<int> R(rowsA * colsB);
+
+    // Matrices have random values in the range [0, 10]
+    std::random_device rd;
+    std::mt19937 gen(rd());
```
msimberg, line 36: See earlier comments about command line options.

examples/quickstart/matrix_multiplication.cpp
```diff
@@ -0,0 +1,87 @@
...
+    });
+
+    // Print all 3 matrices
+    std::cout << "Matrix A is : \n";
+    for (std::size_t i = 0; i < rowsA; i++) {
+        for (std::size_t j = 0; j < colsA; j++)
+            std::cout << A[i * colsA + j] << " ";
+        std::cout << "\n";
+    }
+    std::cout << "\n";
```
msimberg, line 58: Could you factor this out into a separate function?

docs/sphinx/examples/matrix_multiplication.rst
```diff
@@ -0,0 +1,107 @@
+..
+    Copyright (C) 2021 Dimitra Karatza
+
+    SPDX-License-Identifier: BSL-1.0
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+.. _examples_matrix_multiplication:
```
msimberg, line 8: Somewhere in the example we should mention that this example is for demonstration purposes only, and that users should prefer to use optimized libraries like BLAS or blaze (https://bitbucket.org/blaze-lib/blaze, it even has an HPX backend!).

examples/quickstart/matrix_multiplication.cpp
```diff
@@ -0,0 +1,87 @@
...
+    std::vector<int> A(rowsA * colsA);
+    std::vector<int> B(rowsB * colsB);
+    std::vector<int> R(rowsA * colsB);
+
+    // Matrices have random values in the range [0, 10]
+    std::random_device rd;
+    std::mt19937 gen(rd());
+    std::uniform_real_distribution<double> dis(0.0, 10.0);
+    hpx::generate(hpx::execution::seq, std::begin(A), std::end(A), std::bind(dis, gen));
+    hpx::generate(hpx::execution::seq, std::begin(B), std::end(B), std::bind(dis, gen));
```
msimberg, line 39: Maybe?
```suggestion
    auto generator = std::bind(dis, gen);
    hpx::generate(A, generator);
    hpx::generate(B, generator);
```

msimberg (changes requested):

examples/quickstart/matrix_multiplication.cpp
dimitraka: This is throwing an error "no match for call" at line 39 for some reason :/ 
any ideas what I did wrong?

docs/sphinx/examples/matrix_multiplication.rst
dimitraka: what is `vm`?

examples/quickstart/matrix_multiplication.cpp
aurianer: Maybe with hpx::ranges::generate?

docs/sphinx/examples/matrix_multiplication.rst
msimberg: `vm` is an object of type `variables_map`. Instead of `hpx_main` taking `argc/argv` you can have it take a `variables_map` instead. To make HPX aware of your custom command line options you have to tell it about them with an `options_description` like here: https://github.com/STEllAR-GROUP/hpx/blob/dd3f3d9ae87b93c383877f4f69dc97842c2f8387/libs/core/async_cuda/tests/unit/cublas_matmul.cpp#L330-L353.

examples/quickstart/matrix_multiplication.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+#include <iostream>
+#include <random>
+#include <vector>
+
+///////////////////////////////////////////////////////////////////////////////
+//[mul_hpx_main
+int hpx_main()
+{   
+    int rowsA, colsA, rowsB, colsB;
```
msimberg, line 23:
```suggestion
    std::size_t rowsA, colsA, rowsB, colsB;
```

examples/quickstart/matrix_multiplication.cpp
```diff
@@ -0,0 +1,87 @@
...
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+////////////////////////////////////////////////////////////////////////////////
+
+// Parallel implementation of matrix multiplication
+
+#include <hpx/local/algorithm.hpp>
+#include <hpx/local/future.hpp>
+#include <hpx/local/init.hpp>
+
+#include <iostream>
```
msimberg, line 15:
```suggestion
#include <cstddef>
#include <iostream>
```

examples/quickstart/matrix_multiplication.cpp
```diff
@@ -0,0 +1,87 @@
...
+    
+    // Initialize matrices A and B
+    std::vector<int> A(rowsA * colsA);
+    std::vector<int> B(rowsB * colsB);
+    std::vector<int> R(rowsA * colsB);
+
+    // Matrices have random values in the range [0, 10]
+    std::random_device rd;
+    std::mt19937 gen(rd());
+    std::uniform_real_distribution<double> dis(0.0, 10.0);
```
msimberg, line 37: By the way, this should use the same type as the matrix. It might be nice to put a `using element_type = int` at the top of the function and use that alias in place of the concrete type.

```suggestion
    std::uniform_real_distribution<element_type> dis(0, 10);
```

examples/quickstart/matrix_multiplication.cpp
```diff
@@ -0,0 +1,87 @@
...
+    hpx::generate(hpx::execution::seq, std::begin(A), std::end(A), std::bind(dis, gen));
+    hpx::generate(hpx::execution::seq, std::begin(B), std::end(B), std::bind(dis, gen));
+
+    // Perform matrix multiplication
+    hpx::for_loop(hpx::execution::par, 0, rowsA, [&](auto i){
+        hpx::for_loop(0, colsB, [&](auto j){
+            R[i * rowsA + j] = 0;
+            hpx::for_loop(0, rowsB, [&](auto k){
+                R[i * rowsA + j] += A[i * rowsA + k] * B[k * rowsA + j];
```
msimberg, line 46: The stride for each matrix is currently not consistent. Later when you print the matrices you use `colsX` for some and `rowsX` for some, leading to wrong results. On the above line, the stride of A and B is also not necessarily the same, although they both use `rowsA`. It doesn't matter which one you use, as long as it's consistent (one is row-major, the other column-major).

examples/quickstart/matrix_multiplication.cpp
dimitraka: that works, thanks!

docs/sphinx/examples/matrix_multiplication.rst
```diff
@@ -0,0 +1,123 @@
...
+
+To give some random initial values to our matrices, we use `std::uniform_int_distribution
+<https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution>`_. Then, ``std::bind()`` is used
+along with ``hpx::ranges::generate()`` to yield two matrices A and B, which contain values in the range of [0, 10] or in
+the range defined by the user at the command-line arguments. The seed to generate the values can also be defined by the user.
+
+The next step is to perform the matrix multiplication in parallel. This can be done by just using an ``hpx::for_loop()`` 
+combined with a parallel execution policy ``hpx::execution::par`` as the outer loop of the multiplication. Note that the execution 
+of ``hpx::for_loop()`` without specifying an execution policy is equivalent to specifying hpx::execution::seq 
+as the execution policy. More information about ``hpx::for_loop()`` can be found in :cpp:func:`\hpx::for_loop`.
```
msimberg, line 116: Do you think it would make more sense to just make all the mentions of `hpx::for_loop` links to the API reference (i.e. `:cpp:func:` links)? Like this it's going to render as "More information about `hpx::for_loop` can be found in `hpx::for_loop`" with the latter being a link, which to me looks a bit funny and redundant.

examples/quickstart/matrix_multiplication.cpp
msimberg: This one still needs to be resolved.

examples/quickstart/matrix_multiplication.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+#include <iostream>
+#include <random>
+#include <vector>
+
+std::mt19937 gen;
+
+///////////////////////////////////////////////////////////////////////////////
+//[mul_print_matrix
+void print_matrix(std::vector<int> M, int rows, int cols, const char* message){
```
msimberg, line 23:
```suggestion
void print_matrix(std::vector<int> M, std::size_t rows, std::size_t cols, const char* message){
```
I think the examples are not checked for formatting, but inspect will complain about the longer line length. If you haven't used the formatter before you'll need `clang-format` version 9 (can be installed with `clang-format-9` at least on Ubuntu), and then you run `clang-format -i path/to/file.cpp` to format it.

examples/quickstart/matrix_multiplication.cpp
msimberg: This one still needs fixing (the access to `B` uses `rowsA`). I'd perhaps also define
```
    std::size_t rowsR = rowsA;
    std::size_t colsR = colsB;
```
for indexing into `R`.

msimberg (changes requested):
This is looking good. Just a few minor changes still needed. Thanks!

examples/quickstart/matrix_multiplication.cpp
```diff
@@ -0,0 +1,135 @@
...
+    using namespace hpx::program_options;
+    options_description cmdline("usage: " HPX_APPLICATION_STRING " [options]");
+    // clang-format off
+    cmdline.add_options()
+        ("n",
+        hpx::program_options::value<std::size_t>()->default_value(2),
+        "Number of rows of first matrix")
+        ("m",
+        hpx::program_options::value<std::size_t>()->default_value(3),
+        "Number of columns of first matrix (equal to the number of rows of second matrix)")
```
msimberg, line 116: This line should be shorter... Since it's inside `// clang-format off` you'll have to manually wrap it to 80 characters.

docs/sphinx/examples/matrix_multiplication.rst
```diff
@@ -0,0 +1,119 @@
...
+Proceeding to the ``hpx_main()`` function, we can see that matrix multiplication can be done very easily. 
+
+.. literalinclude:: ../../examples/quickstart/matrix_multiplication.cpp
+   :language: c++
+   :start-after: //[mul_hpx_main
+   :end-before: //]
+
+First, the dimensions of the matrices are defined. If they were not given as command-line arguments, their default 
+values are `2` x `3` for the first matrix and `3` x `2` for the second. We use standard vectors to define the matrices 
+to be multipled as well as the resultant matrix.
```
msimberg, line 102:
```suggestion
to be multiplied as well as the resultant matrix. 
```
as reported by our spell checker.

examples/quickstart/matrix_multiplication.cpp
```diff
@@ -0,0 +1,135 @@
...
+    });
+
+    // Print all 3 matrices
+    std::cout << "Matrix A is : \n";
+    for (std::size_t i = 0; i < rowsA; i++) {
+        for (std::size_t j = 0; j < colsA; j++)
+            std::cout << A[i * colsA + j] << " ";
+        std::cout << "\n";
+    }
+
```
msimberg, line 91:
```suggestion
```

examples/quickstart/matrix_multiplication.cpp
```diff
@@ -0,0 +1,135 @@
...
+    auto generator = std::bind(dis, gen);
+    hpx::ranges::generate(A, generator);
+    hpx::ranges::generate(B, generator);
+
+    // Perform matrix multiplication
+    hpx::for_loop(hpx::execution::par, 0, rowsA, [&](auto i) {
+        hpx::for_loop(0, colsB, [&](auto j) {
+            R[i * rowsR + j] = 0;
+            hpx::for_loop(0, rowsB, [&](auto k) {
+                R[i * rowsR + j] += A[i * rowsA + k] * B[k * rowsB + j];
```
msimberg, line 79: I think the indexing here is still wrong: https://github.com/STEllAR-GROUP/hpx/runs/4464305113?check_suite_focus=true#step:6:137. Could you have another look at this? I may backtrack on saying that we should keep it simple. A `matrix` wrapper would probably be a good idea after all.

msimberg (changes requested):

msimberg (approved):
Thanks @dimitraka for this work! Now let's merge it.

msimberg:
bors merge

PR #5646, opened 2021-11-12
Fix ends_with test

msimberg:
With certain seeds/random numbers it would fail with a floating point exception because of `x % 0` (like this: https://cdash.cscs.ch/testDetails.php?test=57251016&build=196753). I've updated it to use `uniform_int_distribution` to avoid the modulo in the first place. `starts_with` was not buggy in the same way, but I've still updated it to use `uniform_int_distribution` as I find it much easier to see what range of values can be generated.

Finally, I've added a "Checklist" section to the pull request template. Anything that can be checked in CI should be checked there, but some things are hard or impossible to check automatically. For those cases it's sometimes enough to have a gentle reminder to check things manually. Please feel free to expand the list in the future (or remove things; needs change), or suggest additions already now.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,3 +7,11 @@ Fixes #
   -
 
 ## Any background context you want to provide?
+
+## Checklist
+
+Not all points below apply to all pull requests.
+
+- [ ] I have added a new feature and have added tests to go along with it.
+- [ ] I have fixed a bug and have added a regression test.
+- [ ] I have added a test using random numbers; I have made sure it uses a seed, and that random numbers generated are valid inputs for the tests.
--- a/libs/core/algorithms/tests/unit/algorithms/ends_with.cpp
+++ b/libs/core/algorithms/tests/unit/algorithms/ends_with.cpp
@@ -14,17 +14,23 @@
 #include <iostream>
 #include <iterator>
 #include <numeric>
+#include <random>
 #include <string>
 #include <vector>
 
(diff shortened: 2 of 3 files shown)


Jedi18:
Thanks! I'll remember to use uniform_int_distribution in the tests I add from now on.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,3 +7,11 @@ Fixes #
...
 
 ## Any background context you want to provide?
+
+## Checklist
+
+Not all points below apply to all pull requests.
+
+- [ ] I have added a new feature and have added tests to go along with it.
+- [ ] I have fixed a bug and have added a regression test.
+- [ ] I have added a test using random numbers; I have made sure it uses a seed, and that random numbers generated are valid inputs for the tests.
```
hkaiser, line 17: NIce!

hkaiser (approved):
We have cleaned many of our tests to fix similar issues already. Would you mind checking if there are still other tests having similar problems?

Otherwise, LGTM, thanks!

.github/PULL_REQUEST_TEMPLATE.md
msimberg: Note, codacy warns about these checkboxes as it's an extension of github flavoured markdown, but given that the template is only for use here on github I'm going to ignore the warnings.

msimberg:
> We have cleaned many of our tests to fix similar issues already. Would you mind checking if there are still other tests having similar problems?

I'll try to do so in a separate PR.

PR #5655, opened 2021-11-15
Avoid for module lists to grow indefinitely in cmake cache

hkaiser:
The CMake variables `HPX_ENABLED_MODULES` and `HPX_[CORE|FULL]_ENABLED_MODULES` have to be made unique as otherwise they grow indefinitely in the CMake cache.


msimberg (approved):
Good catch, thank you!

msimberg:
bors merge

PR #5658, opened 2021-11-16
Replace HPX_INLINE_CONSTEXPR_VARIABLE with inline constexpr

hkaiser:


libs/core/config/include/hpx/config/constexpr.hpp
```diff
@@ -33,7 +33,7 @@
 #if defined(HPX_COMPUTE_DEVICE_CODE)
```
msimberg, line 33: Could you also remove the macro itself a few lines up?

msimberg (changes requested):
Looks good to me! I'd just remove the macro itself as well.

libs/core/config/include/hpx/config/constexpr.hpp
hkaiser: Good catch, will do.

aurianer (approved):
LGTM thanks!

msimberg:
Unrelated to this PR, but the partitioned vector minmax failure is unfortunately not gone: https://cdash.cscs.ch/testDetails.php?test=57780895&build=198648.

msimberg:
bors merge

msimberg (approved):

msimberg:
bors merge

msimberg:
If bors ends up being more annoying than useful we can revise the conditions it requires for a merge. However, I'd try it like this still for a while. At least this forces someone to check that requests for changes have actually been resolved.

PR #5659, opened 2021-11-17
Modernize hpx::future and related facilities

hkaiser:
- move future (and facilities) to namespace hpx
- deprecate those types/functions in namespace hpx::lcos

### DIFF
--- a/libs/core/futures/include/hpx/futures/packaged_continuation.hpp
+++ b/libs/core/futures/include/hpx/futures/packaged_continuation.hpp
@@ -121,19 +121,6 @@ namespace hpx { namespace lcos { namespace detail {
     }
 
     ///////////////////////////////////////////////////////////////////////////
-    template <typename ContResult>
-    struct continuation_result
-    {
-        using type = ContResult;
-    };
-
-    template <typename ContResult>
-    struct continuation_result<future<ContResult>>
-    {
-        using type = ContResult;
-    };
-
-    ///////////////////////////////////////////////////////////////////////////
     template <typename Future, typename F, typename ContResult>
     class continuation : public detail::future_data<ContResult>
     {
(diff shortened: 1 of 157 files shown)


msimberg (dismissed):
Thanks @hkaiser!

msimberg:
Looks like one more `lcos::future` over here: https://cdash.cscs.ch/viewBuildError.php?buildid=198602. But let's get this in once CI is happy.

aurianer (dismissed):
LGTM thanks!

libs/core/futures/include/hpx/futures/packaged_continuation.hpp
```diff
@@ -120,38 +120,24 @@ namespace hpx { namespace lcos { namespace detail {
...
         using base_type = future_data<ContResult>;
 
-        using mutex_type = typename base_type::mutex_type;
         using result_type = typename base_type::result_type;
 
     protected:
         threads::thread_id_type get_id() const
         {
-            std::lock_guard<mutex_type> l(this->mtx_);
+            std::lock_guard l(this->mtx_);
```
msimberg, line 135: As much as I'd like for this to work, `nvcc` seems to choke on this: https://cdash.cscs.ch/viewBuildError.php?buildid=199806...

hkaiser: Yes, I did revert this in other places already :/  It's the `this->mtx_` that is throwing the compiler off.

msimberg: I wonder if it would be happier with `std::lock_guard l(mtx_)`? Might be worth a try if you feel like it. Otherwise let's just keep the explicit template parameter.

hkaiser: Yes, `std::lock_guard l(mtx_)` works, however not in this context as the `mtx_` member is defined in the base class and the lock guard is needed in the derived class (which is a template).

hkaiser: I hope I found a way to make it work, let's see.

msimberg:
bors merge

msimberg (approved):

msimberg:
bors merge

PR #5663, opened 2021-11-18
Use dynamic bitsets by default for CPU masks

### LINKED ISSUE #5661: HPX should detect the number of cores on the build machine
## Expected Behavior

HPX should build for the number of cores on the machine you're on by default

## Actual Behavior

It builds for 64 or less cores

msimberg:
Maybe fixes #5661.

Opening this as a draft only to see the performance test CI results.

### DIFF
--- a/.jenkins/cscs/env-clang-10.sh
+++ b/.jenkins/cscs/env-clang-10.sh
@@ -23,7 +23,6 @@ module load daint-mc
 spack load cmake@3.18.6
 spack load ninja@1.10.0
 
-configure_extra_options+=" -DHPX_WITH_MAX_CPU_COUNT=128"
 configure_extra_options+=" -DHPX_WITH_MALLOC=system"
 configure_extra_options+=" -DHPX_WITH_FETCH_ASIO=ON"
 configure_extra_options+=" -DHPX_WITH_COMPILER_WARNINGS=ON"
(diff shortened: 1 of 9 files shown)


msimberg:
The first runs looks good both for `future_overhead` and `stream`. However, that's only with 4 threads. I'll try to get some more comprehensive runs in later.

msimberg:
Below are some more thorough benchmarks. All them are with the default scheduler. I'd say the differences are within noise levels.

Future overhead benchmark with the `parallel_executor` (since it spawns more tasks):
future_overhead_static_dynamic

Stream benchmark:
stream_static_dynamic

hkaiser:
> Below are some more thorough benchmarks. All them are with the default scheduler. I'd say the differences are within noise levels.

Ok, let's switch to dynamic bitsets, then.

msimberg:
In yesterday's meeting we decided to use vectors in the `shared_priority_queue_scheduler` to make it work with dynamic bitsets. In my infinite wisdom I already made that change in #5171 (https://github.com/STEllAR-GROUP/hpx/blob/7ae910fcc1180b48d775791fb1941a74ab3be584/libs/core/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp#L1362-L1368) and promptly forgot about it. It even uses arrays if `HPX_WITH_MAX_CPU_COUNT` is defined and falls back to vectors otherwise.

msimberg:
bors merge

PR #5669, opened 2021-11-22
Use HPX_WITH_CXX_STANDARD for controlling C++ version 

### LINKED ISSUE #5668: HPX does not detect the C++ 20 standard using gcc 11.2
## Expected Behavior

gcc 11.2 has experimental support for C++ 20, especially coroutines. So, 

`cmake -DHPX_CXX_STANDARD=20 `

should switch to the C++ 20 standard. 

## Actual Behavior

However, HPX is still using `-- Using C++17` even by requesting `-DHPX_CXX_STANDARD=20`.

The workaround is to add `-DHPX_WITH_CXX20=ON `and the correct C++ standard is used. 

  1. Build HPX master with gcc 11.1.2 

## Specifications

  - HPX Version: Master
  - Platform (compiler, OS): Fedora, gcc 11.12

hkaiser:
Fixes #5668

### DIFF
--- a/cmake/HPX_CheckCXXStandard.cmake
+++ b/cmake/HPX_CheckCXXStandard.cmake
@@ -5,29 +5,43 @@
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 # We require at least C++17. However, if a higher standard is set by the user in
-# CMAKE_CXX_STANDARD that requirement has to be propagated to users of HPX as
-# well (i.e. HPX can't be compiled with C++20 and applications with C++17; the
-# other way around is allowed). Ideally, users should not set CMAKE_CXX_STANDARD
-# when building HPX.
-set(HPX_CXX_STANDARD 17)
+# HPX_WITH_CXX_STANDARD (or CMAKE_CXX_STANDARD in conjunction with
+# HPX_USE_CMAKE_CXX_STANDARD=ON) that requirement has to be propagated to users
+# of HPX as well (i.e. HPX can't be compiled with C++20 and applications with
+# C++17; the other way around is allowed). Ideally, users should not set
+# CMAKE_CXX_STANDARD when building HPX.
+
+set(HPX_CXX_STANDARD_DEFAULT 17)
+if(CMAKE_CXX_STANDARD AND HPX_USE_CMAKE_CXX_STANDARD)
+  set(HPX_CXX_STANDARD_DEFAULT ${CMAKE_CXX_STANDARD})
+endif()
+
+hpx_option(
+  HPX_WITH_CXX_STANDARD
+  STRING
+  "Set the C++ standard to use when compiling HPX itself. (default: ${HPX_CXX_STANDARD_DEFAULT})"
+  "${HPX_CXX_STANDARD_DEFAULT}"
+)
+
+set(HPX_CXX_STANDARD ${HPX_WITH_CXX_STANDARD})
 
 # Compatibility for old HPX_WITH_CXXAB options.
 if(HPX_WITH_CXX11)
   hpx_error(
-    "HPX_WITH_CXX11 is deprecated and the minimum C++ standard required by HPX is 17. Avoid setting the standard explicitly or use CMAKE_CXX_STANDARD and HPX_USE_CMAKE_CXX_STANDARD if you must set it."
+    "HPX_WITH_CXX11 is deprecated and the minimum C++ standard required by HPX is 17. Use HPX_WITH_CXX_STANDARD instead."
   )
 elseif(HPX_WITH_CXX14)
   hpx_error(
-    "HPX_WITH_CXX14 is deprecated and the minimum C++ standard required by HPX is 17. Avoid setting the standard explicitly or use CMAKE_CXX_STANDARD and HPX_USE_CMAKE_CXX_STANDARD if you must set it."
+    "HPX_WITH_CXX14 is deprecated and the minimum C++ standard required by HPX is 17. Use HPX_WITH_CXX_STANDARD instead."
   )
 elseif(HPX_WITH_CXX17)
   hpx_warn(
-    "HPX_WITH_CXX17 is deprecated. Avoid setting the standard explicitly or use CMAKE_CXX_STANDARD and HPX_USE_CMAKE_CXX_STANDARD if you must set it."
+    "HPX_WITH_CXX17 is deprecated. Use HPX_WITH_CXX_STANDARD=17 instead."
   )
   set(HPX_CXX_STANDARD 17)
 elseif(HPX_WITH_CXX20)
   hpx_warn(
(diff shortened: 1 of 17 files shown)


msimberg (changes requested):
`HPX_CXX_STANDARD` is not a CMake option and is not meant to be used directly. That's why setting it doesn't work. However, we should change it to be one in favour of setting `CMAKE_CXX_STANDARD`. Could you update this to make a CMake option?

hkaiser:
> `HPX_CXX_STANDARD` is not a CMake option and is not meant to be used directly. That's why setting it doesn't work. However, we should change it to be one in favour of setting `CMAKE_CXX_STANDARD`. 

The options we currently have are
- use `CMAKE_CXX_STANDARD=20` and `HPX_USE_CMAKE_CXX_STANDARD=ON`
- use `HPX_WITH_CXX20=ON`

However, both issue a warning ([here](https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_CheckCXXStandard.cmake#L42-L44) and [here](https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_CheckCXXStandard.cmake#L29-L31)). So what is the preferred way of compiling HPX using C++20?

> Could you update this to make a CMake option?

Certainly, I'd suggest to rename it to `HPX_WITH_CXX_STANDARD` for naming consistency in this case.

msimberg:
> The options we currently have are
> 
>     * use `CMAKE_CXX_STANDARD=20` and `HPX_USE_CMAKE_CXX_STANDARD=ON`
> 
>     * use `HPX_WITH_CXX20=ON`
> 
> 
> However, both issue a warning ([here](https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_CheckCXXStandard.cmake#L42-L44) and [here](https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_CheckCXXStandard.cmake#L29-L31)). So what is the preferred way of compiling HPX using C++20?

The former.

> Certainly, I'd suggest to rename it to `HPX_WITH_CXX_STANDARD` for naming consistency in this case.

Yeah, absolutely. Since it hasn't been an option so far `HPX_CXX_STANDARD` fit much better alongside `CMAKE_CXX_STANDARD`, but as a proper option `HPX_WITH_CXX_STANDARD` fits better.

hkaiser:
> > The options we currently have are
> > ```
> > * use `CMAKE_CXX_STANDARD=20` and `HPX_USE_CMAKE_CXX_STANDARD=ON`
> > 
> > * use `HPX_WITH_CXX20=ON`
> > ```
> > However, both issue a warning ([here](https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_CheckCXXStandard.cmake#L42-L44) and [here](https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_CheckCXXStandard.cmake#L29-L31)). So what is the preferred way of compiling HPX using C++20?
> 
> The former.

Why the warning in this case, then? Rather, I'd expect to see a warning if only `CMAKE_CXX_STANDARD=20` is given, while using `HPX_USE_CMAKE_CXX_STANDARD=ON` to silence it.

msimberg:
> Why the warning in this case, then? Rather, I'd expect to see a warning if only `CMAKE_CXX_STANDARD=20` is given, while using `HPX_USE_CMAKE_CXX_STANDARD=ON` to silence it.

That's a good question. I think the idea was simply to encourage users to use the default standard, but I think that was probably a bit misguided. It is just a warning after all, but I don't think it's necessary.

cmake/HPX_CheckCXXStandard.cmake
```diff
@@ -37,17 +50,10 @@ if(CMAKE_CXX_STANDARD)
...
-      set(HPX_CXX_STANDARD ${CMAKE_CXX_STANDARD})
-    else()
-      hpx_error(
-        "You've set CMAKE_CXX_STANDARD manually, which is not recommended. If you really want to set CMAKE_CXX_STANDARD, set HPX_USE_CMAKE_CXX_STANDARD=ON."
-      )
-    endif()
+  elseif(NOT HPX_USE_CMAKE_CXX_STANDARD)
+    hpx_error(
+      "You've set CMAKE_CXX_STANDARD manually, which is not recommended. If you really want to set CMAKE_CXX_STANDARD, set HPX_USE_CMAKE_CXX_STANDARD=ON."
+    )
```
msimberg, line 56: If we have `HPX_WITH_CXX_STANDARD` as the normal way to set the standard, we should just set `CMAKE_CXX_STANDARD` unconditionally to `HPX_WITH_CXX_STANDARD`.

hkaiser: Hmmm. I thought the idea was not to set the CMAKE_CXX_STANDARD directly to avoid conflicts with depending projects. But I might miss part of the picture.

msimberg: You're probably right... I don't remember the full reason for that. In any case:
1. The error here should probably tell users to use `HPX_WITH_CXX_STANDARD` instead.
2. Since we set `cxx_std_XX` on the targets I don't think we need to set `CMAKE_CXX_STANDARD`. I forgot about this one earlier.

cmake/HPX_CheckCXXStandard.cmake
```diff
@@ -5,11 +5,24 @@
...
+  set(HPX_CXX_STANDARD_DEFAULT ${CMAKE_CXX_STANDARD})
+endif()
+
+hpx_option(
+  HPX_WITH_CXX_STANDARD STRING
+  "Set the C++ standard to use when compiling HPX itself. (default: ${HPX_CXX_STANDARD_DEFAULT})"
+  "${HPX_CXX_STANDARD_DEFAULT}"
+)
+
+set(HPX_CXX_STANDARD=${HPX_WITH_CXX_STANDARD})
```
aurianer, line 25:
```suggestion
set(HPX_CXX_STANDARD ${HPX_WITH_CXX_STANDARD})
```
?

cmake/HPX_CheckCXXStandard.cmake
aurianer: Maybe also suggest `HPX_WITH_CXX_STANDARD` in the warning messages of the `HPX_WITH_CXX<xx>` options?

aurianer (dismissed):
2 comments but otherwise LGTM thanks!

cmake/HPX_CheckCXXStandard.cmake
hkaiser: Silly me! Thanks!

hkaiser:

hkaiser:
bors merge

aurianer (approved):
LGTM thanks!

msimberg:
bors merge

msimberg (approved):

msimberg:
bors merge

PR #5671, opened 2021-11-23
Making sure task_group can be reused

hkaiser:


libs/core/synchronization/include/hpx/synchronization/latch.hpp
```diff
@@ -285,9 +284,40 @@ namespace hpx { namespace lcos { namespace local {
...
+            std::scoped_lock l(mtx_.data_);
             HPX_ASSERT(notified_);
             notified_ = false;
         }
+
+        /// Effects: Equivalent to:
+        ///             if (is_ready())
+        ///                 reset(count);
+        ///             count_up(n);
+        void count_up_or_reset(std::ptrdiff_t n, std::ptrdiff_t count)
```
msimberg, line 296: I suppose this is slightly better than the "manual" version in the comment above, because it avoids potentially taking the lock twice?

My only concern is the name, since it's `count_up_and_reset` when the `latch` is ready, and otherwise it's `count_up`. However, I don't have a suggestion for a better name so I'm only mentioning this in case you or someone else does.

hkaiser: > I suppose this is slightly better than the "manual" version in the comment above, because it avoids potentially taking the lock twice?

Yes, this also avoids associated race conditions.

> My only concern is the name, since it's count_up_and_reset when the latch is ready, and otherwise it's count_up. However, I don't have a suggestion for a better name so I'm only mentioning this in case you or someone else does.

It's the best name I was able to come up with ;-)

msimberg:
Doesn't `has_arrived_` have to be updated to `false` before reusing the group? As it is the first `task_group::wait` will update `has_arrived_` to `true` and subsequent uses won't bother waiting in `task_group::wait`, or am I missing something? This would have to be baked into the `count_up_or_reset` operation, and I don't see a simple way of doing that.

Outside of that I don't see any obvious race conditions, but then it's not obvious to me that there aren't any either (what race conditions are obvious...). I wonder if an explicit `task_group::reset` might not be easier to reason about?

libs/core/synchronization/include/hpx/synchronization/latch.hpp
msimberg: I like the suggestion you mentioned yesterday better than the current option, i.e. something like `reset_if_needed_and_count_up`.

hkaiser:
> Doesn't `has_arrived_` have to be updated to `false` before reusing the group? As it is the first `task_group::wait` will update `has_arrived_` to `true` and subsequent uses won't bother waiting in `task_group::wait`, or am I missing something? This would have to be baked into the `count_up_or_reset` operation, and I don't see a simple way of doing that.

Thanks, I'll have a second look.

> Outside of that I don't see any obvious race conditions, but then it's not obvious to me that there aren't any either (what race conditions are obvious...). I wonder if an explicit `task_group::reset` might not be easier to reason about?

Sure we can add that for different use cases. But in the discussed use case - wouldn't it be prone to races again as we would have to write:
```
    if (tg.is_ready())
        tg.reset();
    tg.run(...);
```
?

msimberg:
> Sure we can add that for different use cases. But in the discussed use case - wouldn't it be prone to races again as we would have to write:
> 
> ```
>     if (tg.is_ready())
>         tg.reset();
>     tg.run(...);
> ```
> 
> ?

Wouldn't you rather do

```
tg.wait();
tg.reset();
tg.run(...);
tg.wait();
```
?
And require in `task_group::reset` that `has_arrived_` is true before resetting. Likewise `task_group::run` should require that `has_arrived_` is false. `task_group` is also move-only, right (or should be)? You can only `task_group::wait` in one thread, and the `task_group::reset` would have to happen after `wait`. I admit I haven't thought this all the way to the end so I may not see the intended use case. Is the test you've added representative of the real thing?

hkaiser:
> Edit: of course there are references... What guarantees does `task_group` give about concurrent access? Don't all `run` calls have to be either in the same thread that calls `wait` or nested within other `run` calls?

The idea was that `task_group` is fully thread safe (it fully relies on `latch`, so all corresponding promises hold). At least I would like for it to be as thread safe as possible (mostly as I do not have a full picture of how it is being used in the SHAD context).

msimberg:
> The idea was that `task_group` is fully thread safe (it fully relies on `latch`, so all corresponding promises hold). At least I would like for it to be as thread safe as possible (mostly as I do not have a full picture of how it is being used in the SHAD context).

Can you define thread safe in this context? ;) The following is definitely problematic (and that's probably ok):

```
task_group tg;
hpx::async([&]() { tg.run(...); });
tg.wait();
```

hence my comment about nested tasks or in the same thread as `wait`. That's why I would expect the `wait` and reset to be called sequentially in the same thread. But again, don't let my comments distract you too much since I don't know how it's going to be used.

hkaiser:
I would like to go ahead with the currently proposed solution as this will cover the use cases we see in SHAD (after renaming the function). Would you support that? Let's solve issues as we go - as said, I'm not 100% sure what other use cases there might be.

msimberg:
> I would like to go ahead with the currently proposed solution as this will cover the use cases we see in SHAD (after renaming the function). Would you support that? Let's solve issues as we go - as said, I'm not 100% sure what other use cases there might be.

Yeah, ok by me, but the issue with `has_arrived_` is still there as far as I can tell.

hkaiser:
> > I would like to go ahead with the currently proposed solution as this will cover the use cases we see in SHAD (after renaming the function). Would you support that? Let's solve issues as we go - as said, I'm not 100% sure what other use cases there might be.
> 
> Yeah, ok by me, but the issue with `has_arrived_` is still there as far as I can tell.

Working on it...

msimberg (approved):

msimberg:
bors merge

PR #5674, opened 2021-11-24
Allow using  `any_sender` in global variables

msimberg:
## Proposed Changes

This allows storing `(unique_)any_sender` in global variables (or even function-local statics). The problem is that the destructor of the `any_sender` may hold e.g. hold an `any_operation_state/receiver` from a type-erased predecessor sender. When the global `any_sender` is destroyed it may try to access an empty vtable that was destroyed before the global `any_sender` was destroyed. The failure case in the added test is a segfault. The `empty` method on the (supposed) empty vtable happens to return `false` and the destructor tries to delete the empty vtable even though it was never dynamically allocated.

~This changes the empty vtables to be `inline constexpr` variables.~ This adds a helper base class which accesses the `any_operation_state` and `any_receiver` empty vtables when constructing `(unique_)any_sender`.

Flyby:
- Remove `explicit` from the `(unique_)any_sender` constructors. It's very convenient to be able to simply assign any sender to a `(unique_)any_sender` without spelling out the type.

## Checklist

### DIFF
--- a/cmake/HPX_AddConfigTest.cmake
+++ b/cmake/HPX_AddConfigTest.cmake
@@ -531,6 +531,15 @@ function(hpx_check_for_cxx20_std_ranges_iter_swap)
 endfunction()
 
 # ##############################################################################
+function(hpx_check_for_cxx20_trivial_virtual_destructor)
+  add_hpx_config_test(
+    HPX_WITH_CXX20_TRIVIAL_VIRTUAL_DESTRUCTOR
+    SOURCE cmake/tests/cxx20_trivial_virtual_destructor.cpp
+    FILE ${ARGN}
+  )
+endfunction()
+
+# ##############################################################################
 function(hpx_check_for_cxx_lambda_capture_decltype)
   add_hpx_config_test(
     HPX_WITH_CXX_LAMBDA_CAPTURE_DECLTYPE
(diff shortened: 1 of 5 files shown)


msimberg:
The current approach only works with C++20, and even then apparently only with GCC. Back to the drawing board...

msimberg:
This now has another workaround in place. The `(unique_)any_sender` constructors now access the `any_operation_state` and `any_receiver` empty vtables to make sure they're created early enough. This seems to work on the test case, but I'm not sure if there are edge cases not covered by this (not that this has to cover all of them). @hkaiser do you see any glaring problems with this approach. Apologies for the diff, it looks worse than it is because I've had to shuffle around some of the functionality. This is the gist of the change: https://github.com/STEllAR-GROUP/hpx/blob/af5d87cab466694ac66b213b10efb788ab090d6e/libs/core/execution_base/include/hpx/execution_base/any_sender.hpp#L674-L682.

msimberg:
I've added a feature test for trivial types with virtual destructors and a cleaner implementation based on that feature. The cleaner implementation is the original one I tried in this PR, i.e. using `inline constexpr` variables for the empty vtables. Interestingly GCC would actually compile the cleaner implementation, but does not compile the feature test because of a bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93413. However, I'm leaving the stricter test in place meaning GCC will still use the fallback implementation with C++20.

msimberg:
bors merge

PR #5679, opened 2021-11-25
Fix `fork_join_executor` with dynamic schedule

msimberg:
I've expanded the `fork_join_executor` to also test the dynamic schedule. I've also added tests for normal and high priority threads.

The constructor now throws an exception if the executor is constructed with `thread_stacksize::nostack` since that would prevent yielding. For an unclear reason `thread_priority::low` does not work with the executor. I'm guessing it has to do with how we decide to steal threads, and low priority threads having a single queue, but I don't have a full explanation. I'm leaving it out of the test at the moment, but I'm not adding a check for it in the constructor without understanding the reason for it.

@gonidelis could you please try this on your test cases as well? 

## Checklist

### DIFF
--- a/libs/core/executors/CMakeLists.txt
+++ b/libs/core/executors/CMakeLists.txt
@@ -64,7 +64,9 @@ set(executors_compat_headers
 )
 # cmake-format: on
 
-set(executors_sources current_executor.cpp exception_list_callbacks.cpp)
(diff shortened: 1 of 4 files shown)


hkaiser (approved):
LGTM, thanks!

msimberg:
bors merge

msimberg:
bors merge

PR #5683, opened 2021-11-26
Make hpx::wait_all etc. throw exceptions when waited futures hold exceptions and deprecate hpx::lcos::wait_all[_n] in favor of hpx::wait_all[_n]

hkaiser:
- Attention: hpx::wait_all[_n] now throws if one of the futures is exceptional
  this is a change in semantics.
- Introduce hpx::wait_all[_n]_nothrow that exhibits the (non-throwing) old
  semantics of wait_all[_n]
- Extending existing tests, add new for nothrow variants

The change in semantics proposed overcome a problem of the long-standing practice that users most of the time didn't realize the old versions of `wait_all` wouldn't throw if one of the futures was exceptional. This was prone to exceptions being dropped on the floor, leading to obscure runtime issues.

Similar changes have been applied for `hpx::wait_any[_n]`, `hpx::wait_some[_n]`, and `hpx::wait_each[_n]`.

This also deprecates `hpx::lcos::wait` and friends. This should be removed completely in the future. Its functionality is fully subsumed by `hpx::wait_each`.

### DIFF
--- a/components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_impl.hpp
+++ b/components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_impl.hpp
@@ -87,7 +87,7 @@ namespace hpx {
                         .then(get_ptr_helper{l, partitions_}));
             }
         }
-        wait_all(ptrs);
+        hpx::wait_all(ptrs);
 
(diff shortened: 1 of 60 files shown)


msimberg (dismissed):
Thanks! Looks good to me besides the build errors.

msimberg:
bors merge

PR #5684, opened 2021-11-30
Move lock registration to separate module and remove global lock registration

msimberg:
If someone has a better suggestion for where to put it I'm not at all set on the `init_runtime` module.


hkaiser:
Should we move all of the lock handling into its own module?

msimberg:
> Should we move all of the lock handling into its own module?

Yeah, I'll do that. I'll make it `lock_registration_local` for the "regular" locks and `lock_registration` for the global locks. Feel free to suggest other names as well.

aurianer (approved):
LGTM thanks

hkaiser:
@msimberg could you please explain what was the rationale to separate the functionalities of register_lock and register_lock_globally into separate modules?

msimberg:
> @msimberg could you please explain what was the rationale to separate the functionalities of register_lock and register_lock_globally into separate modules?

Because I _thought_ global lock registration was only relevant for the distributed runtime. However, I was clearly wrong. But what _is_ the purpose of `register_lock_globally` actually? `(un)register_lock_globally` are never called anywhere in the codebase, unless I've messed up my search. Is there a use for users (but there's no way to "verify" the registered locks)?

hkaiser:
`register_lock` works on a (kernel-)thread basis, while `register_lock_globally` applies to all (kernel-)threads. I'm surprised we don't use it anymore, however - I'd say, let's remove the global version altogether. I'm not even sure of what use case this was introduced for :/

msimberg:
> `register_lock` works on a (kernel-)thread basis, while `register_lock_globally` applies to all (kernel-)threads. I'm surprised we don't use it anymore, however - I'd say, let's remove the global version altogether. I'm not even sure of what use case this was introduced for :/

Mystery... in any case, I've removed the global lock registration now (along with the CMake option). Thanks @hkaiser!

msimberg:
This should now be ready to go. With only one module, the "local" one obviously doesn't need to have local in the name anymore. There's now only `lock_registration`.

msimberg:
I only fixed formatting in the last force push.

bors merge

msimberg:
bors merge

PR #5690, opened 2021-12-02
Replace tag_dispatch with tag_invoke in minmax segmented

aurianer:


hkaiser (dismissed):
Excellent catch! LGTM, thanks!

msimberg:
@hkaiser/@Jedi18 would you have time to take over here? There's more than just the function names to be done here. There's also the mystery of how on earth things compiled with `tag_dispatch` and what was being called instead...

hkaiser:
> @hkaiser / @Jedi18 would you have time to take over here? There's more than just the function names to be done here. There's also the mystery of how on earth things compiled with `tag_dispatch` and what was being called instead...

Yah, mystery indeed. This might also explain why we've seen these strange test failures related to min/max.

hkaiser:
Besides the min/max test failures (that are unrelated to the proposed changes), this looks good now. It's interesting to see that those test failures have manifested themselves for non-CUDA runners now. That should simplify diagnosing the issue...

msimberg:
Thanks @hkaiser!

The min/max were failing occasionally also before this PR on master, e.g. https://cdash.cscs.ch/testDetails.php?test=57125541&build=205178. But given that they were using the wrong `tag_invoke` overloads the previous failures may have been caused by something else.

msimberg:
bors merge

PR #5691, opened 2021-12-02
Rename `annotate_function` to `scoped_annotation`

### LINKED ISSUE #5686: Rename `annotate_function`
The distinction between `annotate_function` and `annotated_function` is often confusing to users. I think the current `annotate_function` could be renamed e.g. `scoped_annotation` as it

1. clearly distinguishes it from `annotated_function` and
2. at least in my opinion describes its purpose much better than `annotate_function`

I'm sure there would be other alternative names as well so feel free to suggest some if you can think of any. The main reason for opening this issue is that I think it should be renamed.

Bonus change that I'm adding as a reminder:
- move `annotate(d)_function` from `hpx::util` to `hpx`

msimberg:
Fixes #5686. Also moves the utilities to the top-level `hpx` namespace. Adds them to the public API list.

Open question: which header should provide these? I've now added both to `hpx/local/functional.hpp` (and by extension `hpx/functional.hpp`). `scoped_annotation` could also fit in `hpx/local/execution.hpp` in my opinion (`annotated_function` could sort of also fit there). 

Some (in my opinion worse) alternative names:

- `annotation`
- `annotate_scope`
- `annotation_scoped`
- `raii_annotation`

### DIFF
--- a/docs/sphinx/api/public_api.rst
+++ b/docs/sphinx/api/public_api.rst
@@ -402,10 +402,12 @@ Classes
 - :cpp:class:`hpx::util::unique_function_nonser`
 - :cpp:struct:`hpx::traits::is_bind_expression`
 - :cpp:struct:`hpx::traits::is_placeholder`
+- :cpp:struct:`hpx::scoped_annotation`
 
 Functions
 ---------
 
+- :cpp:func:`hpx::annotated_function`
 - :cpp:func:`hpx::util::bind`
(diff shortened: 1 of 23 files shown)


hkaiser:
`scoped_annotation` is fine, thanks!

hkaiser (approved):
LGTM, thanks!

msimberg:
The codacy warning is a false positive...

msimberg:
bors merge

Issue #5544, opened 2021-09-10
Use _GLIBCXX_DEBUG in CI


msimberg:
Defining `_GLIBCXX_DEBUG` enables additional checks in libstdc++, which may be useful for catching bugs.

PR #5693, opened 2021-12-03
Fix test errors with `_GLIBCXX_DEBUG` defined

msimberg:
Fixes part of #5544.

The `is_heap` change is due to `greater_equal` not satisfying https://en.cppreference.com/w/cpp/named_req/Compare. With `_GLIBCXX_DEBUG` libstdc++ actually gives a surprisingly helpful error message:

```
 57: In function:
 57:     _RAIter std::is_heap_until(_RAIter, _RAIter, _Compare) [with _RAIter =
 57:     __gnu_debug::_Safe_iterator<__gnu_cxx::__normal_iterator<int*,
 57:     std::__cxx1998::vector<int, std::allocator<int> > >,
 57:     std::__debug::vector<int>, std::random_access_iterator_tag>; _Compare =
 57:     std::greater_equal<int>]
 57:
 57: Error: comparison doesn't meet irreflexive requirements, assert(!(a < a)).
 57:
 57: Objects involved in the operation:
 57:     instance "functor" @ 0x0x2b53cad26267 {
 57:       type = std::greater_equal<int>;
 57:     }
 57:     iterator::value_type "ordered type" {
 57:       type = int;
 57:     }
```

The change to `is_contiguous_iterator` is because vector iterators are different with `_GLIBCXX_DEBUG` (the pointers are wrapped). 

The last change to `incplace_merge` tests is due to confused use of variables. With `_GLIBCXX_DEBUG` libstdc++ reported use of moved-from iterators.

There are still a few tests that fail with `_GLIBCXX_DEBUG`, so I'm not enabling it in CI. If anyone feels like looking into them they're at least `uninitialized_copyn` and `uninitialized_move`. They both segfault.


hkaiser (approved):
LGTM, thanks!

msimberg:
Force-push only fixed formatting.

bors merge

PR #5694, opened 2021-12-03
Change default `get_function_annotation` implementation

msimberg:
Note, this is a proposal. There are pros and cons to this change. I'd like to hear some opinions from others about this.

Currently on master `get_function_annotation` returns `nullptr` unless there's a specialization (like when using `annotated_function`). In practice this means that unannotated functions typically inherit the annotation from the spawning thread (like `hpx_main`). This is nice in some situations because you know what spawned the task.

On the other hand it would be nice if the default annotations were actually related to the actual task being spawned. Instead of returning `nullptr` `get_function_annotation` can fall back to returning `typeid(F).name()`. Even if this a mangled name it will usually be more descriptive (in my opinion...). This also means that all the other fallback names that we use will never be taken into account because a function will always have a non-`nullptr` name.

Ideally we'd be able to have both. APEX does keep track of parent task ids which indirectly gives the task name of the parent. However, tooling for viewing this data is still quite bad.

Are there other pros/cons to using `typeid` instead?

### DIFF
--- a/libs/core/functional/include/hpx/functional/traits/get_function_annotation.hpp
+++ b/libs/core/functional/include/hpx/functional/traits/get_function_annotation.hpp
@@ -19,7 +19,7 @@ namespace hpx { namespace traits {
     {
         static constexpr char const* call(F const& /*f*/) noexcept
         {
-            return nullptr;
+            return typeid(F).name();
         }
     };
 


hkaiser (approved):
LGTM, thanks!

aurianer (approved):
LGTM thanks!

msimberg:
The tests showed the change in behaviour resulting from this change. The annotation in the following changes (assuming `f` does not have an explicit annotation):
```
async(with_annotation("executor-annotation"), f);
```
Previously `f` would have a `nullptr` annotation, and the annotation from the executor would be used as a fallback. With the change in this PR `f` will always have an annotation and that will be used instead. A related question is which annotation should take precedence in:
```
async(with_annotation("executor-annotation"), annotated_function("function-annotation", f));
```
? It's not obvious to me that one or the other is clearly better. Thoughts?

biddisco:
> Are there other pros/cons to using `typeid` instead?

```
#include <hpx/debugging/demangle_helper.hpp>
...
hpx::util::debug::print_type<F>()
```
just in case you were not aware of this.

IMHO, if a task has been annotated, this should be returned first, then the executor annotation, and if nothing is avail, then the typeid?

hkaiser:
> The tests showed the change in behaviour resulting from this change. The annotation in the following changes (assuming `f` does not have an explicit annotation):
> 
> ```
> async(with_annotation("executor-annotation"), f);
> ```
>
> Previously `f` would have a `nullptr` annotation, and the annotation from the executor would be used as a fallback. With the change in this PR `f` will always have an annotation and that will be used instead. 

The correct behavior for this would be the equivalent of:
```
async(exec, [&](){
    scoped_annotation _("executor-annotation");
    f();
});
```

> A related question is which annotation should take precedence in:
> 
> ```
> async(with_annotation("executor-annotation"), annotated_function("function-annotation", f));
> ```
> 
> ? It's not obvious to me that one or the other is clearly better. Thoughts?

This should be equivalent to writing:
```
async(exec, [&](){
    scoped_annotation _("executor-annotation");
    {
        scoped_annotation __("function-annotation");
        f();
    }
}
```
In both examples `exec` has no annotation itself.

hkaiser:
FWIW, just as an additional data point, we already use `typeid(F).name()` here: 
https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/threading_base/include/hpx/threading_base/annotated_function.hpp#L208

msimberg:
Replaced by https://github.com/STEllAR-GROUP/hpx-local/pull/12.

PR #5697, opened 2021-12-05
Adding fork_join_executor to foreach_benchmark

hkaiser:
- minor tweaks to `fork_join_executor`
- ~~disable lock verification for RelWithDebInfo builds~~
- more `constexpr` and `noexcept`
- fixing issue when `HPX_WITH_THREAD_PHASE_INFORMATION=On`
- remove duplicate macro definition
- remove bogus assert in latch
- fixing problem with `small_vector`
- minor optimization for `wait_all` and `handle_local_exceptions`
- making chunk iterators random access
- always `#define HPX_SMT_PAUSE`

This is a collection of minor optimizations and fixes applied during doing some performance analysis of the fork-join executor. Some of the changes are unrelated, I can separate those, if needed.


libs/core/config/include/hpx/config.hpp
```diff
@@ -202,13 +202,13 @@
 ///////////////////////////////////////////////////////////////////////////////
 /// By default, enable minimal thread deadlock detection in debug builds only.
 #if !defined(HPX_HAVE_THREAD_MINIMAL_DEADLOCK_DETECTION)
-#  if defined(HPX_DEBUG)
+#  if defined(HPX_DEBUG) && !defined(NDEBUG)
```
msimberg, line 205: Are you sure this is a good idea? This means that everything that depends on these defines (`HPX_HAVE_THREAD_MINIMAL_DEADLOCK_DETECTION`, `HPX_HAVE_SPINLOCK_DEADLOCK_DETECTION`, etc.) will depend on both `HPX_DEBUG` (which we control, as much as we can) and `NDEBUG` which the user is free to set as they wish in their application. I'm worried this will lead to ABI problems.

If this is for "disable lock verification for RelWithDebInfo builds", then `HPX_DEBUG` _should_ not be defined with `RelWithDebInfo`. If it is, something else is likely missing/wrong.

libs/core/datastructures/include/hpx/datastructures/detail/small_vector.hpp
```diff
@@ -78,22 +78,25 @@ namespace hpx::detail {
             noexcept(Allocator()))
           : memory_()
           , resource_(alloc)
-          , pool_(std::data(memory_), std::size(memory_), &resource_)
+          , pool_(
+                std::data(memory_), std::size(memory_) * sizeof(T), &resource_)
```
msimberg, line 82: Oops, good catch! Luckily this is not a memory issue? The missing `sizeof(T)` "only" meant that the actual embedded capacity was `sizeof(T)` smaller, right?

libs/core/executors/include/hpx/executors/fork_join_executor.hpp
```diff
@@ -208,13 +229,16 @@ namespace hpx { namespace execution { namespace experimental {
...
             {
                 for (std::size_t t = 0; t < num_threads_; ++t)
                 {
                     while (thread_states_[t].data_.load(
-                               std::memory_order_acquire) != state)
+                               std::memory_order_relaxed) != state)
                     {
+#if defined(HPX_SMT_PAUSE)
+                        HPX_SMT_PAUSE;
+#endif
```
msimberg, line 241: Should we maybe always define `HPX_SMT_PAUSE` so that we can remove the `#if defined(...` from here?

libs/core/executors/include/hpx/executors/fork_join_executor.hpp
```diff
@@ -428,28 +450,27 @@ namespace hpx { namespace execution { namespace experimental {
...
+                thread_function_helper<F, S, decltype(argument_pack)>::call(
+                    element_function_, shape_, argument_pack_, main_thread_,
+                    num_threads_, schedule_, queues_, thread_states_,
+                    exception_mutex_, exception_);
 
+                // Wait for all threads to finish their work assigned to
+                // them in this parallel region.
                 wait_state_all(thread_state::idle);
 
+                std::lock_guard l(exception_mutex_);
```
msimberg, line 473: I think this lock should not be needed _if_ the states are read/written with acquire/release instead of relaxed.

libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size_iterator.hpp
```diff
@@ -71,20 +75,58 @@ namespace hpx { namespace parallel { namespace util { namespace detail {
...
             count_ -= chunk();
 
             iterator() = parallel::v1::detail::next(iterator(), chunk());
             chunk() = (hpx::detail::min)(chunk_size_, count_);
         }
 
+        HPX_HOST_DEVICE void decrement()
+        {
+            // prepare next value
+            count_ += chunk();
```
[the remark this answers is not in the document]
hkaiser: I might not be right. I'm aware of this potential issue, but since all we do with this iterator is to possibly `std::advance` it, I decided to ignore it. I'll take a note and will try to work on it next.

libs/core/config/include/hpx/config.hpp
hkaiser: Yes, I agree, this is a tricky one. On one hand, we want for the asserts to be in place for RelWithDebInfo (`HPX_DEBUG` is defined), otoh, the options related to thread debugging might not be desired by default for this build type. We could simply remove the conditional `#define ...` here and enable it through the build system for `Debug` builds instead.

libs/core/datastructures/include/hpx/datastructures/detail/small_vector.hpp
hkaiser: Yep, it should have been just that.

libs/core/executors/include/hpx/executors/fork_join_executor.hpp
hkaiser: Yes, I understand. Except if it's not ;-) (assuming that the executor is always used from the same main thread, and that it is never invoked concurrently...) Does it hurt to have the lock here?

libs/core/config/include/hpx/config.hpp
msimberg: Where is `HPX_DEBUG` defined for `RelWithDebInfo`?

libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size_iterator.hpp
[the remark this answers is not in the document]
msimberg: If we only `advance` it do we need `decrement` for something?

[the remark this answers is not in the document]
hkaiser: It's there only for completeness's sake.

libs/core/executors/include/hpx/executors/fork_join_executor.hpp
msimberg: Yeah, I realize you changed some of them to `relaxed` :P 

> Does it hurt to have the lock here?

Would it be nice not to have it there? Yes. Will I (or anyone else) notice that the lock is there? I doubt it ;)

The question is what we gain with the `relaxed` vs what we lose with the lock here.

libs/core/config/include/hpx/config.hpp
hkaiser: Uhh, you got me there. I was _assuming_ it was defined (I believe to remember it was defined at some point). But if it's not (as it appears to be the case), I will revert the related changes.

libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size_iterator.hpp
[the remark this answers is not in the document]
msimberg: Can we check for the bad case somehow? This will be a hard-to-debug issue for someone in the future way after we've forgotten about this behaviour and it'd be nice to fail early here if it's ever used.

libs/core/algorithms/tests/performance/foreach_scaling.cpp
```diff
@@ -522,6 +522,38 @@ int hpx_main(hpx::program_options::variables_map& vm)
                 seq_time_forloop = averageout_sequential_forloop(vector_size);
             }
         }
+        else if (vm.count("forkjoin") != 0)
```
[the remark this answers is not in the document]
hkaiser: The code currently supports running one of the executors at a time only, I think the `else if` is correct. I can change the code to report results for more than one executor, however. Should this be done in this PR?

libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size_iterator.hpp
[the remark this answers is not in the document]
hkaiser: Ok, I'll try to make it right with this PR. I'll also add a test.

libs/core/executors/include/hpx/executors/fork_join_executor.hpp
hkaiser: The `relaxed` is still an atomic operation; `release`/`acquire` just additionally prevent operation reordering across it. Thus, we most likely do not gain anything (measurable) from using `relaxed`. The lock will also most likely not be contended, so all we do is adding another two atomic load/store operations.

The `fork_join_executor` is quite sophisticated and not easy to understand. Not having the lock might make understanding it even more difficult.

libs/core/algorithms/tests/performance/foreach_scaling.cpp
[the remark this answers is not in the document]
msimberg: I understand why the `if-else` is (was) there. There used to be two choices and no `--aggregated` and `--aggregated` used to select between the two options. Now there are three options but adding a `--forkjoin` command line option gives you four possible combinations. I'm saying it should either be possible to switch on/off all three individually or it should be possible to select only one of them using a single command line option. I don't mind which one but I think it very much belongs in this PR.

libs/core/executors/include/hpx/executors/fork_join_executor.hpp
msimberg: I'm saying I don't mind if the lock is there so if you prefer to have it there let's have it there.

But outside of that I doubt the lock helps in understanding the executor ;) And if `relaxed` doesn't help then why change it?

libs/core/algorithms/tests/performance/foreach_scaling.cpp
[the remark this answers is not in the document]
hkaiser: Ok, I'll  try to address your comment.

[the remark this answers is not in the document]
hkaiser: FWIW, I went for `--executor=...`.

libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size_iterator.hpp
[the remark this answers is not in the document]
hkaiser: The iterators should be fine now.

libs/core/algorithms/tests/unit/algorithms/detail/chunk_size_idx_iterator.cpp
```diff
@@ -0,0 +1,176 @@
...
+    HPX_TEST(std::binary_search(start, finish, *internal));
+
+    std::pair<Iterator, Iterator> xy(
+        std::equal_range(start, finish, *internal));
+    Iterator x = xy.first, y = xy.second;
+
+    HPX_TEST(std::distance(x, y) == 1);
+
+// disable warning: unary minus operator applied to unsigned type, result
+// still unsigned
```
msimberg, line 59: What triggers this warning and why can it be ignored?

libs/core/config/include/hpx/config.hpp
```diff
@@ -208,7 +208,7 @@
 #endif
 #if !defined(HPX_HAVE_SPINLOCK_DEADLOCK_DETECTION)
 #  if defined(HPX_DEBUG)
-//#    define HPX_HAVE_SPINLOCK_DEADLOCK_DETECTION
+#    define HPX_HAVE_SPINLOCK_DEADLOCK_DETECTION
```
msimberg, line 211: I think this has been disabled, unless explicitly asked for in the CMake configuration, because it may lead to false positives (if spinning just takes a very long time). I'm not sure it's a problem in 99.9% of the cases, but I just want to mention this since this will always enable it for debug builds with no way to turn it off.

libs/core/functional/include/hpx/functional/invoke_fused.hpp
```diff
@@ -80,7 +81,7 @@ namespace hpx { namespace util {
     ///
     /// \note This function is similar to `std::apply` (C++17)
     template <typename F, typename Tuple>
-    constexpr HPX_HOST_DEVICE
+    constexpr HPX_HOST_DEVICE HPX_FORCEINLINE
```
msimberg, line 84: Do we have any evidence that using forceinline helps here (actually, in general)?

libs/core/algorithms/tests/unit/algorithms/detail/chunk_size_idx_iterator.cpp
hkaiser: No idea, this is copied straight from the Boost iterator test suite - which reminds me that I need to add their copyright...

libs/core/config/include/hpx/config.hpp
hkaiser: I have removed the silent configuration change in Debug mode now and have moved this to explicit options in the build system. I think this is the correct solution...

libs/core/functional/include/hpx/functional/invoke_fused.hpp
hkaiser: I do have evidence from looking at performance results in vtune showing that (at least for MSVC) this option helps reducing the stack noise.

libs/core/algorithms/tests/unit/algorithms/detail/chunk_size_idx_iterator.cpp
msimberg: Could you give it a try without ignoring the warning? I suppose it might be a warning inside the MSVC `std::binary_search` implementation :/

libs/core/execution/tests/unit/executor_parameters.cpp
```diff
@@ -187,9 +187,9 @@ int hpx_main(hpx::program_options::variables_map& vm)
     std::cout << "using seed: " << seed << std::endl;
     std::srand(seed);
 
-    test_dynamic_chunk_size();
-    test_static_chunk_size();
-    test_guided_chunk_size();
+    //test_dynamic_chunk_size();
+    //test_static_chunk_size();
+    //test_guided_chunk_size();
```
msimberg, line 192: Temporary, I hope? Just commenting as a reminder to uncomment these before merging.

libs/core/executors/include/hpx/executors/fork_join_executor.hpp
```diff
@@ -87,12 +89,38 @@ namespace hpx { namespace execution { namespace experimental {
...
+                // The helper function that does the actual work for a single
+                // parallel region.
+                std::atomic<thread_function_helper_type*>
+                    thread_function_helper_;
+
+                // Pointers to inputs to bulk_sync_execute.
+                void* element_function_;
+                void const* shape_;
+                void* argument_pack_;
+            };
```
msimberg, line 113: Did you separate these out for performance or readability? I'm asking because of the potential for false sharing, but as far as I can tell this should be ok. A worker thread only changes its own `state_` at the end of a region.

libs/core/executors/include/hpx/executors/fork_join_executor.hpp
```diff
@@ -87,12 +89,38 @@ namespace hpx { namespace execution { namespace experimental {
...
+
+            // Can't apply 'using' here as the type needs to be forward
+            // declared
+            struct region_data_type
+              : std::vector<hpx::util::cache_aligned_data<region_data>>
+            {
+                using base_type =
+                    std::vector<hpx::util::cache_aligned_data<region_data>>;
+                using base_type::base_type;
+            };
```
msimberg, line 123: I think I don't understand the need for this. Could you explain it?

libs/core/config/include/hpx/config.hpp
msimberg: Thanks! The downside of this is that the default will persist when changing the build type which is a bit unfortunate. I don't think there's a right solution here, but I think your current change is ok.

libs/core/functional/include/hpx/functional/invoke_fused.hpp
msimberg: Ok, thanks, I was thinking in terms of performance but this is another benefit.

libs/core/executors/include/hpx/executors/fork_join_executor.hpp
hkaiser: I wanted for those variables to always be on the same cache line. The core always loads the state, which makes those other values available right away. This change has improved the performance significantly. I will post a performance comparison plot once we're done with this.

libs/core/executors/include/hpx/executors/fork_join_executor.hpp
hkaiser: There is a circular dependency between the region_data_type and the thread function type embedded in it. This requires for the region_data_type to be forward declared, which is not possible if it's introduced with a `using` directive.

libs/core/algorithms/tests/unit/algorithms/detail/chunk_size_idx_iterator.cpp
hkaiser: I have removed the #pragmas now as I don't see the warning locally. It might have been that Boost was using these functions in a different context, causing the warning to show up.

hkaiser:
retest cscs

hkaiser:
@msimberg this seems to be ready to go now. Also for good measure, the promised performance plot:
image

This graph shows the achieved performance comparing different versions of the taskbench benchmark. The important part for this PR are the yellow and lilac lines (old vs. new fork_join executor).

libs/core/executors/include/hpx/executors/fork_join_executor.hpp
```diff
@@ -133,108 +178,101 @@ namespace hpx { namespace execution { namespace experimental {
...
-                    {
-                        wait_not_state_this_thread(thread_state::idle);
-                        if (thread_state_ == thread_state::stopping)
-                        {
-                            break;
-                        }
+                    region_data& data = region_data_[thread_index_].data_;
+                    auto state = shared_data::wait_state_this_thread(
+                        data.state_, thread_state::idle, yield_delay_,
+                        std::not_equal_to<>());
```
msimberg, line 210: Final comment. Does the region data not have to be loaded _after_ waiting for the state to change and the load should be `acquire`? I would expect the above code to be the "inverse" of https://github.com/STEllAR-GROUP/hpx/blob/29fc9d457587fecd6d40af7f8a3e98ac2fdc0162/libs/core/executors/include/hpx/executors/fork_join_executor.hpp#L515-L523, i.e.
```
... load(state, acquire) ...;
region_data& data = ...;
```
Is there some other memory barrier that ensures that the load can actually be `relaxed`?

msimberg: I see now that `data` is a reference so loading that before the wait should not be an issue. The question about `relaxed` vs `acquire` still stands.

hkaiser: Sorry, but which `load()` do you refer to?

msimberg: The one(s) inside `wait_state_this_thread`.

hkaiser: Here is my reasoning: the function `wait_state_this_thread` is called in order to wait for a given thread's state to change: either for it to become equal to or to be changed away from the current state. In both cases there are two options: either the condition is already true whenever the `load()` is called (in which case the function returns), or the condition is false (in which case the function will try again). In both scenarios, a `load(relaxed)` is fine - if the condition is true, we know that we can stop waiting and no possible reordering of operations will make it false again. If the condition is false the operation will be retried, even if because of some reordering the actual state was already changed and our `load()` didn't see the state change yet. The next iteration will probably see the change. Do I miss something?

hkaiser: Let me add that I probably should flip the use of `equal_to` and `not_equal_to` and change the test conditions inside the function to the opposite. This would better reflect my reasoning laid out above.

msimberg: I actually liked the inverted condition :P but I like your update as well! In any case, my comment had nothing to do with that.

The comment about the `relaxed` load refers to the fact that it doesn't provide any guarantees about reordering other loads around the load of the state. In particular I'm worried about loading the region data here https://github.com/STEllAR-GROUP/hpx/blob/29fc9d457587fecd6d40af7f8a3e98ac2fdc0162/libs/core/executors/include/hpx/executors/fork_join_executor.hpp#L383-L388. An `acquire` load would guarantee that the region data is not reordered to be loaded before reaching the required state. My question is: is there something _else_ that gives the same guarantee if it's using a `relaxed` load?

In practice it's very unlikely that the compiler will rewrite it like that, but I wouldn't want to rely on "very unlikely"...

hkaiser: Well, all of those variables are on the same cache line as the state that is loaded atomically. If this isn't guarantee enough, then we will have turn those into atomics as well.

msimberg: Thanks! I may be off here, but I still don't think that's a guarantee. I suppose in the end this function call being behind a function pointer would prevent any chance of the loads being inlined and reordered: https://github.com/STEllAR-GROUP/hpx/blob/29fc9d457587fecd6d40af7f8a3e98ac2fdc0162/libs/core/executors/include/hpx/executors/fork_join_executor.hpp#L214-L217. However, that's again not a guarantee.

Let me ask a different question: did you see a speed-up with `relaxed` vs. `acquire`? I have a vague memory that you mentioned that in some comment or conversation, but can't find it anywhere now.

msimberg: I just wanted to add that it might be easier to figure this out together in Thursday's call...

hkaiser: > However, that's again not a guarantee.

Indeed. 

> Let me ask a different question: did you see a speed-up with `relaxed` vs. `acquire`? I have a vague memory that you mentioned that in some comment or conversation, but can't find it anywhere now.

I did not see any speedup (on x86 those are equivalent anyways). I'll change it to `acquire`, then.

msimberg: Thank you. I may be wrong in the end but especially if the performance is the same I'm a lot more comfortable with an `acquire`.

libs/core/iterator_support/tests/include/hpx/iterator_support/tests/iterator_tests.hpp
```diff
@@ -344,7 +344,8 @@ namespace tests {
     template <typename Iterator, typename TrueVals>
     void random_access_readable_iterator_test(Iterator i, int N, TrueVals vals)
     {
-        bidirectional_readable_iterator_test(i, vals[0], vals[1]);
+        bidirectional_readable_iterator_test(
+            i, vals[0], N > 1 ? vals[1] : vals[0]);
```
msimberg, line 348: Huh, why? https://github.com/STEllAR-GROUP/hpx/pull/5703 asserts that `N` is at least 2 because the iterator `i` is advanced and compared to `v2` in the tests.

hkaiser: Uhh, merge snafu, I selected the wrong change here. Will fix.

libs/core/executors/include/hpx/executors/fork_join_executor.hpp
msimberg: Wait, I fear that we've badly misunderstood each other. In https://github.com/STEllAR-GROUP/hpx/pull/5697/commits/020deb82058ad8aa3cde3901fccd409af4082d73 you've changed some `relaxed` memory orderings. I was referring to the loads of the thread state... When setting the state it's set with `release` after writing to other variables. When reading the state it should be read with `acquire` so that reads of the other variables definitely see the writes to them that happened before writing the state.

msimberg: To expand a bit more, this is what I'm going for with this: https://en.cppreference.com/w/cpp/atomic/memory_order#Release-Acquire_ordering. The reads/writes of other variables are ordered appropriately with respect to changes to the thread states (when using `release`/`acquire`).

hkaiser: @msimberg I do agree with you now. Thanks for being persistent! This also allows us to make the function pointer in the region data to be non-atomic.

msimberg: Thanks for bearing with me, and more importantly for doing this PR in the first place!

msimberg (approved):

msimberg:
bors merge

PR #5702, opened 2021-12-10
Attempting to fix CircleCI timeouts

hkaiser:
- [x] Split header tests
- [x] Split unit tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -449,17 +449,6 @@ jobs:
       - store_artifacts:
           path: tests.unit
 
-      - run:
-          <<: *convert_xml
-      - run:
-          <<: *move_core_dump
-      - run:
-          <<: *move_debug_log
-      - store_test_results:
-          path: tests.unit2
-      - store_artifacts:
-          path: tests.unit2
-
   tests.regressions:
     <<: *defaults
     steps:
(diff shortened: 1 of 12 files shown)


libs/core/algorithms/include/hpx/parallel/datapar/transform_loop.hpp
```diff
@@ -11,6 +11,7 @@
 #if defined(HPX_HAVE_DATAPAR)
 #include <hpx/datastructures/tuple.hpp>
 #include <hpx/execution/traits/is_execution_policy.hpp>
+#include <hpx/executors/datapar/execution_policy.hpp>
```
msimberg, line 14: I think `hpx/parallel/datapar/transfer.hpp` should also have this header included. It doesn't show up on the header tests because it includes `hpx/parallel/datapar/transform_loop.hpp` but it uses `simd_policy` directly so it should have the include as well.

.circleci/tests.unit1.targets
```diff
@@ -0,0 +1,29 @@
+#  Copyright (c) 2021 Hartmut Kaiser
```
msimberg, line 1: Would a simple split into algorithms and non-algorithms be enough? This seems very error prone as it requires adding a new module to one of these lists explicitly.

.circleci/config.yml
```diff
@@ -408,15 +408,18 @@ jobs:
...
           at: /hpx
       - run:
-          name: Building Unit Tests
+          name: Building Unit Tests (Algorithms)
           command: |
-              ninja -j2 -k 0 tests.unit
+              ninja -j2 -k 0 \
+                tests.unit.modules.algorithms.algorithms \
+                tests.unit.modules.algorithms.block \
+                tests.unit.modules.algorithms.datapar_algorithms
```
msimberg, line 422: One of the reasons to wait for the split to fix these was that the set of tests will be smaller. None of these tests will run for this repository after the split anymore.

.circleci/tests.unit1.targets
hkaiser: No, that still times out. I tried.

.circleci/config.yml
hkaiser: Sure. However, I already had to fix several things we've overlooked since CircleCI limited the runtime.

hkaiser:

msimberg (approved):
Thanks!

msimberg:
bors merge

PR #5707, opened 2021-12-13
Better diagnose parcel bootstrap problems

hkaiser:
- flyby: add PALS_NODEID env variable to the list of recognized MPI environments

This fixes problems related to MPI on the (Cray) Perlmutter system


msimberg (approved):
Looks good to me, thanks!

Possible addition: would it be a good idea to add a debug logging message to `detect_mpi_environment` saying something along the lines of "Trying to detect MPI environment with environment variables A, B, C." and then a failure/success message? That could make it a bit easier to detect if these kind of failures are a problem in the environment or something internal to HPX.

hkaiser:
> Looks good to me, thanks!
> 
> Possible addition: would it be a good idea to add a debug logging message to `detect_mpi_environment` saying something along the lines of "Trying to detect MPI environment with environment variables A, B, C." and then a failure/success message? That could make it a bit easier to detect if these kind of failures are a problem in the environment or something internal to HPX.

Not sure. This would generate misleading error messages in cases where the MPI parcelport is built (or MPI async), but the application doesn't run in an MPI environment.

msimberg:
> Not sure. This would generate misleading error messages in cases where the MPI parcelport is built (or MPI async), but the application doesn't run in an MPI environment.

I was only thinking of an `LRT(debug, "...")` or something like that. Not something that would be printed without asking for it.

libs/core/mpi_base/src/mpi_environment.cpp
```diff
@@ -33,16 +34,23 @@ namespace hpx { namespace util {
...
                 if (env)
+                {
+                    LBT_(debug)
+                        << "Found MPI environment variable: " << tok << "="
+                        << std::string(env) << ", enabling MPI support\n";
                     return true;
+                }
             }
+
+            LBT_(info) << "No know MPI environment variable found, disabling "
```
msimberg, line 52:
```suggestion
            LBT_(info) << "No known MPI environment variable found, disabling "
```
sorry... looks good otherwise. Thank you!

msimberg (approved):

msimberg:
bors merge

PR #5708, opened 2021-12-14
Use `/dev/shm` for Jenkins checkout on Piz Daint

msimberg:
Apparently we're taxing the filesystem too heavily. I will try some workarounds here. This might need a few iterations...


msimberg:
This is useless since `/dev/shm` is node-specific. Closing and revisiting if and when things go wrong again.

PR #5713, opened 2021-12-17
Significantly reduce CI jobs run on Piz Daint

msimberg:
We've again hit the quota for the month. We've received more hours until the end of the month (quarter), but we won't be allowed to run as many jobs in the future.

I'm unsure if this is possible, but I'm trying if the hours can be accounted to the HPX development project instead (d69). It has 9000 node hours per quarter.

The split should also significantly reduce the node hours we need.

Finally, not triggering jobs automatically should significantly help.


hkaiser:
Now that rostam is back into production, we can move part of the builders over to LSU. That will help reducing the burden on PizDaint as well.

msimberg:
bors merge

msimberg:
I'm going ahead with this while we wait for a response about accounting the CI jobs to our development project.

PR #5716, opened 2021-12-18
Avoid duplicate definition error

hkaiser:
`OMPI_IMPORTS` is already defined in hpx-local if ASYNC_MPI is enabled

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -598,15 +598,6 @@ if(HPX_WITH_NETWORKING)
   endif()
 endif(HPX_WITH_NETWORKING)
 
-if((HPX_WITH_NETWORKING AND HPX_WITH_PARCELPORT_MPI) OR HPXLocal_WITH_ASYNC_MPI)
-  if(MSVC)
-    # FIXME: add OpenMPI specific flag here for now as the
-    # hpx_add_compile_flag() below does not add the extra options to the top
-    # level directory
-    hpx_add_config_define(OMPI_IMPORTS)
-  endif()
-endif()
-
 # External libraries/frameworks used by sme of the examples and benchmarks
 hpx_option(
   HPX_WITH_EXAMPLES_HDF5 BOOL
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -598,7 +598,9 @@ if(HPX_WITH_NETWORKING)
   endif()
 endif(HPX_WITH_NETWORKING)
 
-if((HPX_WITH_NETWORKING AND HPX_WITH_PARCELPORT_MPI) OR HPXLocal_WITH_ASYNC_MPI)
+if((HPX_WITH_NETWORKING AND HPX_WITH_PARCELPORT_MPI)
+   AND NOT HPXLocal_WITH_ASYNC_MPI
+)
```
msimberg, line 603: Note that this is an oxymoron. This will never be true.

I think the correct thing would be to add `OMPI_IMPORTS` as a compile flag in HPXLocal, not as a config define. I'm guessing the `FIXME` is trying to say this as well... Does that sound reasonable?

hkaiser: > Note that this is an oxymoron. This will never be true.

In this case we can remove it from this repository alltogether. 

> I think the correct thing would be to add `OMPI_IMPORTS` as a compile flag in HPXLocal, not as a config define. I'm guessing the `FIXME` is trying to say this as well... Does that sound reasonable?

What would that change? In the end it needs to be defined whenever mpi.h (from OpenMPI) is included.

msimberg: > > Note that this is an oxymoron. This will never be true.
> 
> In this case we can remove it from this repository alltogether.

Yes, it's enough to have it in HPXLocal.
 
> > I think the correct thing would be to add `OMPI_IMPORTS` as a compile flag in HPXLocal, not as a config define. I'm guessing the `FIXME` is trying to say this as well... Does that sound reasonable?
> 
> What would that change? In the end it needs to be defined whenever mpi.h (from OpenMPI) is included.

Multiple `-DOMPI_IMPORTS` on the command line will not conflict (as long as they don't try to set a value to it of course...). Alternatively, the define in the config should be conditional. Defining it with `-DOMPI_IMPORTS` has the benefit of not requiring `hpx/config.hpp` to be included before `mpi.h` though.

In any case, I think the define can be removed from here in all its possible forms.

PR #5729, opened 2022-01-11
Workaround for CMake/Ninja generator OOM problem

hkaiser:
CMake runs out of memory when using the Ninja generator if the modules are being built as OBJECT libraries. In this case the fallback is to build STATIC (whole-archive) libraries instead.


hkaiser:
retest lsu

PR #5730, opened 2022-01-12
fix returning non default constructable objects from action

kleinhenz:
Refixes #2847

## Proposed Changes

The ability to return a non default constructable object from an action was broken by ba7153c7fcc9cb63b9d8398314595a00cf6a6991 because of a missing negation of `std::is_default_constructible<T>`.
This wasn't caught by CI because the current test doesn't actually attempt to save or load any construction data.

## Checklist

### DIFF
--- a/libs/core/serialization/include/hpx/serialization/tuple.hpp
+++ b/libs/core/serialization/include/hpx/serialization/tuple.hpp
@@ -106,15 +106,15 @@ namespace hpx { namespace util { namespace detail {
         template <typename T>
         static void save_element_helper(std::true_type, Archive& ar, T& t)
         {
-            using serialization::detail::save_construct_data;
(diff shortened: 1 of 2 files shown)


kleinhenz:
This looks like it was fixed on master by 23c910c08ee616394f4adeed639462deb2691a55 but opening this to fix the 1.7.X branch.

hkaiser:
@kleinhenz We do not plan to make a point release for V1.7 at this point. How important is that fix for you for this version? We plan to release V1.8 soon, would that be sufficient for you?

kleinhenz:
ok I will close if there won't be another point release. I can wait for 1.8.

hkaiser:
@kleinhenz thanks for submitting the patch in any case! Much appreciated.

PR #5731, opened 2022-01-12
Randomly delay execution of bash scripts launched by Jenkins

hkaiser:
This is an attempt to overcome the Slurm issues we're seeing on rostam (sporadic failures to launch the builders that do not seem to be reproducible).


hkaiser:
Looks like this really improves the picture, merging.

PR #5734, opened 2022-01-13
Revert #5586

msimberg:
@hkaiser this is everything "obvious" reverted, plus non-Jenkins configuration commits cherry picked from HPXLocal. I have not made any sanity checks yet so this is a draft but I'm opening this to let CI run.

Things that will likely have to be updated/checked:
- [x] Update reference for newly added foreach benchmark
- [x] Check that all tests are still enabled correctly
- [x] Check that there are no remaining uses of `HPXLocal*`/`HPX_LOCAL*`/`hpx_local*` anywhere

@hkaiser please feel free to expand the list above with things you know or think should be checked and I'll take care of it.


hkaiser (approved):
@msimberg For me everything looks good so far. Thanks a lot for working on this!

msimberg:
@hkaiser I've added two fixes that are not directly related to reverting #5586.

First is that when modules are static libraries they are linked privately to keep them from leaking out to consumers, but that also means that their dependencies also don't "leak" to consumers. They have to be added manually to the top-level library targets in that case: https://github.com/msimberg/hpx/blob/416ddfb437d8949ad354fd095907b515fc371b38/libs/CMakeLists.txt#L113-L124.

Second is something that came up with the cray clang compiler regarding `get`. The use of `visit` on a `std::variant` in some of the sender facilities eventually calls an unqualified `get` on the variant in the standard library. If the variant holds a `hpx::tuple` this leads to two possible overloads: one from the standard library, and one from here: https://github.com/msimberg/hpx/blob/416ddfb437d8949ad354fd095907b515fc371b38/libs/core/datastructures/include/hpx/datastructures/variant_helper.hpp#L24-L54. I don't know why this isn't triggered elsewhere, but if I understand correctly this: https://github.com/msimberg/hpx/blob/3257865c8418d4c39ff38b36ccdfafbe35079ce5/libs/core/datastructures/include/hpx/datastructures/variant_helper.hpp#L57 does not act as an ADL barrier (I don't know if the name is just a copy-paste or if it really should act as a barrier). This does actually act as a barrier: https://github.com/msimberg/hpx/blob/416ddfb437d8949ad354fd095907b515fc371b38/libs/core/datastructures/include/hpx/datastructures/variant_helper.hpp#L57.

This fixes all build failures that I've seen so far on this PR, ignoring test failures that were there already before. I've yet to update the performance CI reference.

hkaiser:
> This does actually act as a barrier: https://github.com/msimberg/hpx/blob/416ddfb437d8949ad354fd095907b515fc371b38/libs/core/datastructures/include/hpx/datastructures/variant_helper.hpp#L57.

Nice catch! Yes, you're right. In order to be an ADL barrier we need to write `using namespace ...;`

hkaiser:
@msimberg I'd be fine with merging this now. The spellcheck problems can be solved separately.

hkaiser:
bors merge

hkaiser (approved):
LGTM, thanks!

hkaiser:
@FunMiles This PR has merged hpx-local back into this repository. You should be able to work on your code again.

PR #5736, opened 2022-01-17
Set commit status from Jenkins also for special branches 

msimberg:
@hkaiser this is required if you decide to try to use bors more extensively. Otherwise it's not necessary and can be closed without merging.

### DIFF
--- a/.jenkins/cscs/entry.sh
+++ b/.jenkins/cscs/entry.sh
@@ -56,12 +56,21 @@ else
     github_commit_status="failure"
 fi
 
-if [[ -n "${ghprbPullId:-}" ]]; then
+# Get the CDash dashboard build id
+cdash_build_id="$(cat jenkins-hpx-${configuration_name_with_build_type}-cdash-build-id.txt)"
+
+if [[ -z "${ghprbPullId:-}" ]]; then
+    .jenkins/common/set_github_status.sh \
+        "${GITHUB_TOKEN}" \
+        "STEllAR-GROUP/hpx" \
+        "${GIT_COMMIT}" \
+        "${github_commit_status}" \
+        "${configuration_name_with_build_type}" \
+        "${cdash_build_id}" \
+        "jenkins/cscs"
+else
     # Extract just the organization and repo names "org/repo" from the full URL
     github_commit_repo="$(echo $ghprbPullLink | sed -n 's/https:\/\/github.com\/\(.*\)\/pull\/[0-9]*/\1/p')"
-
-    # Get the CDash dashboard build id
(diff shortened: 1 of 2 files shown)


hkaiser:
> @hkaiser this is required if you decide to try to use bors more extensively. Otherwise it's not necessary and can be closed without merging.

I'd be more than happy to use bors. We might have to reduce the review requirements for this, however.

hkaiser:
@khuck: while unrelated to this particular PR, APEX now seems to depend on zlib without checking if it's available (see: https://github.com/STEllAR-GROUP/hpx/runs/4838798276?check_suite_focus=true). Is that a known issue?

hkaiser:
bors merge

khuck:
> @khuck: while unrelated to this particular PR, APEX now seems to depend on zlib without checking if it's available (see: https://github.com/STEllAR-GROUP/hpx/runs/4838798276?check_suite_focus=true). Is that a known issue?
> 
> Edit: please see [UO-OACISS/apex#163](https://github.com/UO-OACISS/apex/issues/163) as well.

@hkaiser HPX / APEX _does_ check if zlib is available.  See https://github.com/UO-OACISS/apex/blob/develop/cmake/Modules/APEX_SetupZlib.cmake for the check, and `gzstream.hpp` should only get included if APEX_HAVE_ZLIB is set (https://github.com/UO-OACISS/apex/blob/b09eb7d94c9e329ce74aff29375ee9b7a384561b/src/apex/trace_event_listener.hpp#L15-L17).  Does the cmake `find_package(ZLIB)` step only check for the library, and not the development headers?

hkaiser:
> > @khuck: while unrelated to this particular PR, APEX now seems to depend on zlib without checking if it's available (see: https://github.com/STEllAR-GROUP/hpx/runs/4838798276?check_suite_focus=true). Is that a known issue?
> > Edit: please see [UO-OACISS/apex#163](https://github.com/UO-OACISS/apex/issues/163) as well.
> 
> @hkaiser HPX / APEX _does_ check if zlib is available. See https://github.com/UO-OACISS/apex/blob/develop/cmake/Modules/APEX_SetupZlib.cmake for the check, and `gzstream.hpp` should only get included if APEX_HAVE_ZLIB is set (https://github.com/UO-OACISS/apex/blob/b09eb7d94c9e329ce74aff29375ee9b7a384561b/src/apex/trace_event_listener.hpp#L15-L17). Does the cmake `find_package(ZLIB)` step only check for the library, and not the development headers?

@khuck honestly, I don't know. But this would explain things.

PR #5739, opened 2022-01-19
Enable the option to attach a debugger on a segmentation fault (linux)

hkaiser:
- flyby: rename HPX_LOCAL_ option to HPX_


hkaiser:
This is good to go. Merging...

PR #5743, opened 2022-01-20
Add check for MPICH and set the correct env to suport multi-threaded

diehlpk:
## Proposed Changes

  - If MPICH is used we set `export MPICH_MAX_THREAD_SAFETY=multiple` to activate multithreaded MPI.

This might solve some hanging, we experienced with MPI recently.

Note, this has to happen before `MPI_Init` is being called.


hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #5745, opened 2022-01-21
Modernize serialization module

hkaiser:
- adding a test combining serialization of all-types-assumed-to-be-bitwise-serializable, with
  pointer-serialization and shared_ptr serialization with pointer tracking


hkaiser:
This looks good now, merging...

PR #5750, opened 2022-01-24
Make sure required MPI initialization flags are properly applied and supported

hkaiser:


hkaiser:
This can be merged now

PR #5751, opened 2022-01-24
Propagate MPI dependencies to command line handling

hkaiser:
This allows to compile the command_line_handling odule in case MPI is not installed in system directories.


hkaiser:
This should be good to be merged now.

PR #5757, opened 2022-01-25
Fix missing links in Public API 

dimitraka:
Fix the missing links in API documentation


hkaiser (approved):
LGTM, thanks! Very nice!

PR #5760, opened 2022-01-31
Update cdash build url

msimberg:
The url has changed from `/buildSummary.php?buildid=<buildid>` to just `/build/<buildid>`. The old urls still work and get forwarded to the new ones.


hkaiser (approved):
LGTM, thanks a lot!

PR #5765, opened 2022-02-05
Adding a sphinx role allowing to link to a file directly in github

hkaiser:
@dimitraka: this adds a new role (a first version of it, at least) that enables linking a specific file to it's page on github:
```
:hpx-header:`libs/core/include_local/include,hpx/local/algorithm.hpp`
```
will link for instance to https://github.com/STEllAR-GROUP/hpx/blob/20bcde189e62bd617f6403a8a9c95a0945b2fffa/libs/core/include_local/include/hpx/local/algorithm.hpp (if the git-hash is '20bcde189e62bd617f6403a8a9c95a0945b2fffa') and show `hpx/local/algorithm.hpp` in the generated html (please note the comma in the `base_path,file_name` argument to the new role).

### DIFF
--- a/cmake/templates/conf.py.in
+++ b/cmake/templates/conf.py.in
@@ -571,8 +571,8 @@ html_theme_options = {'logo_only': True}
 # of the sidebar.
 html_logo = '_static/images/HPX_STELLAR_blue.png'
 
-#  The name of an image file (relative to this directory) to place at the top
-# of the brower page.
+# The name of an image file (relative to this directory) to place at the top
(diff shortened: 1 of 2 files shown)


hkaiser:
@dimitraka I have added another sphinx role to conveniently link to any cppreference page. For instance, 
```
:cppreference:`utilities/functional,function` 
```
will now link to https://en.cppreference.com/w/cpp/utility/functional/function and show `std::function` in the generated page.

PR #5773, opened 2022-02-09
Moving barrier to consolidated namespaces

hkaiser:
- make cpp20 barrier default local barrier
- moving local barrier to namespace hpx
- moving distributed barrier to namespace hpx::distributed


gonidelis:
What about:

https://github.com/STEllAR-GROUP/hpx/blob/f4890bc01f18beb4a95eadbc0e55d68fbf43bd63/libs/core/synchronization/tests/unit/local_barrier_reset.cpp#L30

and

https://github.com/STEllAR-GROUP/hpx/blob/f4890bc01f18beb4a95eadbc0e55d68fbf43bd63/libs/core/synchronization/tests/unit/local_barrier_count_up.cpp#L30

?

hkaiser:
> What about:
> 
> https://github.com/STEllAR-GROUP/hpx/blob/f4890bc01f18beb4a95eadbc0e55d68fbf43bd63/libs/core/synchronization/tests/unit/local_barrier_reset.cpp#L30
> 
> and
> 
> https://github.com/STEllAR-GROUP/hpx/blob/f4890bc01f18beb4a95eadbc0e55d68fbf43bd63/libs/core/synchronization/tests/unit/local_barrier_count_up.cpp#L30
> 
> ?
> 
> and possibly: https://github.com/STEllAR-GROUP/hpx/blame/barrier/libs/core/synchronization/docs/index.rst#L17

These two tests use functionalities that go beyond `std::barrier`, so they use a different barrier type.

PR #5780, opened 2022-02-11
Update Piz Daint Jenkins configurations from gcc/clang 7 to 8

msimberg:
GCC 8 and Boost/hwloc built with GCC 8 are still building as I write this so that'll fail at the moment. I moved the CMake configuration options from the clang 7 configuration to the clang 8 configuration (`HPX_WITH_MAX_CPU_COUNT=128` is the most important one).


msimberg:
retest cscs

gonidelis:
@msimberg  I think you need to rebase to current master cause of the Visual Studio from 16 to 17 version change.

hkaiser (dismissed):
LGTM, thanks!

PR #5781, opened 2022-02-11
Updating used version of Visual Studio

hkaiser:
Github has removed VS2019 from the used Windows image

### DIFF
--- a/.github/workflows/windows_debug.yml
+++ b/.github/workflows/windows_debug.yml
@@ -16,17 +16,17 @@ jobs:
     - uses: actions/checkout@v2
     - uses: jwlawson/actions-setup-cmake@v1.9
       with:
-        cmake-version: '3.18.x'
+        cmake-version: '3.22.x'
     - name: Install dependencies
       run: |
         md C:\projects
         $client = new-object System.Net.WebClient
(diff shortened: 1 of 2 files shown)


hkaiser:
> @hkaiser I guess since segmented algos are off we shouldn't expect `build-and-test` to succeed?
> 
> Beyond that, should `tests.regressions.modules.async_distributed.distributed.mpi.*` targets fail too? I think they are not related to our segmented algos errors, right?

Yes, please ignore the segmented algorithm failures, those are known issues and will be addressed separately.

hkaiser:
> > > @hkaiser I guess since segmented algos are off we shouldn't expect `build-and-test` to succeed?
> > > Beyond that, should `tests.regressions.modules.async_distributed.distributed.mpi.*` targets fail too? I think they are not related to our segmented algos errors, right?
> > 
> > 
> > Yes, please ignore the segmented algorithm failures, those are known issues and will be addressed separately.
> 
> What about `async_distributed.distributed.mpi.` though?

All of the test errors seen here are known and are either already have been addressed by other PRs or will be addressed separately. This PR really just changes the github action Windows builder, updating the used version of visual studio.

PR #5786, opened 2022-02-14
Attempting to fix hangs in distributed barrier

hkaiser:
@biddisco could you please try whether this improves the situation?

### DIFF
--- a/libs/full/collectives/include/hpx/collectives/detail/barrier_node.hpp
+++ b/libs/full/collectives/include/hpx/collectives/detail/barrier_node.hpp
@@ -75,8 +75,7 @@ namespace hpx { namespace distributed { namespace detail {
         hpx::lcos::local::promise<void> broadcast_promise_;
         hpx::barrier<> local_barrier_;
 
-        template <typename This>
-        hpx::future<void> do_wait(This this_, hpx::future<void> future);
+        hpx::future<void> do_wait(hpx::future<void> future);
 
         template <typename>
         friend struct components::detail_adl_barrier::init;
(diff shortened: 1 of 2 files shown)


biddisco:
> @biddisco could you please try whether this improves the situation?

Out of 72, 19 completed correctly, so I would say that this patch made no difference. I see that the branch has been force pushed overnight, so I will run again with `-DHPX_WITH_THREAD_DEBUG_INFO=On` using this version.

biddisco:
https://cloud.cscs.ch/owncloud/index.php/s/lmrMKszQANOv2E8 
contains some log data from a new run, but it does not look terribly interesting.

hkaiser:
> https://cloud.cscs.ch/owncloud/index.php/s/lmrMKszQANOv2E8 contains some log data from a new run, but it does not look terribly interesting.

Thanks John! Would it be possible to collect another set of logs using `--hpx:0:debug-hpx-log` and `--hpx:debug-app-log` (don't restrict the latter for locality zero, please) and with `-DHPX_WITH_THREAD_DEBUG_INFO=On`?

biddisco:
https://cloud.cscs.ch/owncloud/index.php/s/yHCfKOZj6dh8AY8
contains new trace with settings as requested

hkaiser:
@biddisco thanks for the logs. Also, after merging #5773 it might be worth trying master again as this PR completely replaces the barrier implementation (amongst other things). We can be almost sure that it's not the barrier that is causing the problems if we still see the same hangs now.

PR #5793, opened 2022-02-20
Remove features that are deprecated since V1.6

hkaiser:


hkaiser:
The remaining issues are unrelated

PR #5794, opened 2022-02-21
Update Piz Daint Jenkins configuration

msimberg:
Attempt to update the configurations... One of the Boost modules has been removed, hence I'm using one of our Boost installs directly. Since it's header-only the used compiler shouldn't matter.

@hkaiser I'll update the performance CI references separately.

### DIFF
--- a/.jenkins/cscs/env-clang-cuda.sh
+++ b/.jenkins/cscs/env-clang-cuda.sh
@@ -8,10 +8,10 @@ export CRAYPE_LINK_TYPE=dynamic
 export APPS_ROOT="/apps/daint/SSL/HPX/packages"
 export CXX_STD="17"
 export HWLOC_ROOT="${APPS_ROOT}/hwloc-2.0.3-gcc-8.3.0"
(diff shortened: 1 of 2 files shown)


msimberg:
retest cscs

msimberg:
retest cscs

msimberg:
The two missing CUDA tests fail to build with `nvcc error   : 'cicc' died due to signal 11 (Invalid memory reference)`. I've retriggered the builds once more, but it looks like it might be an actual problem in nvcc. If you'd like to merge this already now to get "almost everything" working again that's ok, but I'll still try to see if I can do something about the build failure today.

msimberg:
Actually, I take that back. https://github.com/STEllAR-GROUP/hpx/pull/5758 seems to be the first one that fails to build those two CUDA tests also on rostam. This PR is ready to go in that case.

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

Issue #5774, opened 2022-02-09
CMake can't find HPXCacheVariables.cmake 


karlfroldan:
## Expected Behavior

CMake should be able to find `HPXCacheVariables.cmake` as listed in `HPXConfig.cmake`.

## Actual Behavior

Error during generating Makefile with cmake with the following message:
```
Executing command: /usr/local/bin/cmake --no-warn-unused-cli -DCMAKE_TOOLCHAIN_FILE=/home/karl/.local/vcpkg/scripts/buildsystems/vcpkg.cmake -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -DCMAKE_BUILD_TYPE:STRING=Debug -DCMAKE_C_COMPILER:FILEPATH=/bin/clang-12 -DCMAKE_CXX_COMPILER:FILEPATH=/bin/clang++-12 -H/home/karl/dev/cpp/hpx -B/home/karl/dev/cpp/hpx/build -G "Unix Makefiles"
CMake Error at /home/karl/.local/vcpkg/installed/x64-linux/share/hpx/HPXConfig.cmake:13 (include):
  include could not find requested file:
 
     /home/karl/.local/vcpkg/installed/x64-linux/share/hpx/HPXCacheVariables.cmake
 Call Stack (most recent call first):
   /home/karl/.local/vcpkg/scripts/buildsystems/vcpkg.cmake:793 (_find_package)
   CMakeLists.txt:10 (find_package)
 
 
 Not searching for unused variables given on the command line.
 -- Using  allocator.
 -- Configuring incomplete, errors occurred!
 See also "/home/karl/dev/cpp/hpx/build/CMakeFiles/CMakeOutput.log".
 CMake Error at /home/karl/.local/vcpkg/installed/x64-linux/share/hpx/HPX_SetupAllocator.cmake:150 (if):
   if given arguments:
 
     "STREQUAL" "jemalloc"
 
   Unknown arguments specified
 Call Stack (most recent call first):
   /home/karl/.local/vcpkg/installed/x64-linux/share/hpx/HPXConfig.cmake:83 (include)
   /home/karl/.local/vcpkg/scripts/buildsystems/vcpkg.cmake:793 (_find_package)
   CMakeLists.txt:10 (find_package)
```

Furthermore,  `HPXCacheVariables.cmake` is not located in `/path/to/vcpkg/installed/x64-linux/share/hpx` which I believe is the root of the problem.

[...]

PR #5795, opened 2022-02-22
Avoid writing path and directories into HPXCacheVariables.cmake

hkaiser:
- this is in support of https://github.com/STEllAR-GROUP/hpx/issues/5774


hkaiser:
The failures are unrelated

gonidelis:
Thanks for that!

PR #5797, opened 2022-02-23
Extend Sphinx role to support arbitrary text to display on a link

hkaiser:
This extends the Spinx role `:cppreference-generic:` to support an optional thrid argument that will be used for the displayed text. for instance:
```
:cppreference-generic:`algorithm,copy,copy_if`
```
will generate `std::copy_if` referring to https://en.cppreference.com/w/cpp/algorithm/copy.

@dimitraka please try this out for the links that require customization as discussed. Please note that the third 'argument' is optional, all links that don't require it will still work as before.

### DIFF
--- a/docs/sphinx/extensions/sphinx-hpx.py
+++ b/docs/sphinx/extensions/sphinx-hpx.py
@@ -38,14 +38,20 @@ def autolink_hpx_file(pattern):
     return role
 
 # The text in the rst file should be:
-# :cppreference-generic:`base_path,typename`, for instance `thread,barrier`
+# :cppreference-generic:`base_path,typename[,shown]`, for instance `thread,barrier`
 def autolink_generic(pattern):
     def role(name, rawtext, text, lineno, inliner, options={}, content=[]):
         text_parts = [p.strip() for p in text.split(',')]
-        if len(text_parts) >= 2:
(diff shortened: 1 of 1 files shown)


dimitraka:
> This extends the Sphinx role `:cppreference-generic:` to support an optional thrid argument that will be used for the displayed text. for instance:
> 
> ```
> :cppreference-generic:`algorithm,copy,copy_if`
> ```
> 
> will generate `std::copy_if` referring to https://en.cppreference.com/w/cpp/algorithm/copy.
> 
> @dimitraka please try this out for the links that require customization as discussed. Please note that the third 'argument' is optional, all links that don't require it will still work as before.



> This extends the Sphinx role `:cppreference-generic:` to support an optional thrid argument that will be used for the displayed text. for instance:
> 
> ```
> :cppreference-generic:`algorithm,copy,copy_if`
> ```
> 
> will generate `std::copy_if` referring to https://en.cppreference.com/w/cpp/algorithm/copy.
> 
> @dimitraka please try this out for the links that require customization as discussed. Please note that the third 'argument' is optional, all links that don't require it will still work as before.

It works fine! Also, very handy that the argument is optional and we don't have to modify all the links. Thank you!

PR #5798, opened 2022-02-23
Adding MPI as a dependency to APEX

hkaiser:
This attempts to fix build problems in Phylanx. 

@khuck, @stevenrbrandt: please verify whether this overcomes the issues we're seeing.

### DIFF
--- a/cmake/HPX_SetupApex.cmake
+++ b/cmake/HPX_SetupApex.cmake
@@ -74,6 +74,15 @@ if(HPX_WITH_APEX AND NOT TARGET APEX::apex)
     target_link_options(APEX::apex INTERFACE "-Wl,-no-as-needed")
   endif()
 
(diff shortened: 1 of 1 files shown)


khuck:
@hkaiser @stevenrbrandt this worked for me on POWER9 with GCC 10.3.  Failed with GCC 7.3, but for different reasons.

hkaiser:
> @hkaiser @stevenrbrandt this worked for me on POWER9 with GCC 10.3. Failed with GCC 7.3, but for different reasons.

@khuck, I have refined the conditions for which MPI will be used as a dependency for APEX (see https://github.com/STEllAR-GROUP/hpx/pull/5798/files#diff-19aec7660b6277fa703aa116684b3c58ffcc046296e303bf0450ed3d263f41eaR77). Could you please verify this is correct?

khuck:
sorry, late to the party... yes, that should do it (as you probably already determined)

PR #5800, opened 2022-02-28
Add option to restrict MPI background work to subset of cores

hkaiser:
This PR adds an option to the MPI parcelport that allows specifying the maximum number of cores that will participate in the MPI parcelport background work: `--hpx:ini=hpx.parcel.mpi.background_threads!=N`, where `N` is the number of cores that will participate (default is `-1`, meaning all cores participate). Alternatively, this can be set through the environment variable `HPX_HAVE_PARCELPORT_MPI_BACKGROUND_THREADS`.

@diehlpk please verify whether this has any effect on the performance issues you're investigating.

### DIFF
--- a/libs/core/config/include/hpx/config.hpp
+++ b/libs/core/config/include/hpx/config.hpp
@@ -114,6 +114,18 @@
 #  define HPX_PARCEL_MPI_MAX_REQUESTS 2147483647
 #endif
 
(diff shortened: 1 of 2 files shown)


diehlpk:
Thanks, @hkaiser. Using only one thread per node reduced the computation time by 102 seconds. I will play around with the number of threads to find the sweet spot.

PR #5819, opened 2022-03-20
Marking customization points for intrusive_ptr as noexcept

hkaiser:
- moving hpx::memory::intrusive_ptr to hpx::intrusive_ptr
- flyby: mark all deallocate() functions as noexcept


hkaiser:
This looks to be good to go.

PR #5822, opened 2022-03-21
Clang-format remaining files

### LINKED ISSUE #4992: Clang-format the rest of the files
As the modularization is mainly in place, we could format the rest of the files of the repo (components, examples, tests, etc.) and add this to circleci, that would avoid formatting problems like on #4989

hkaiser:
Fixes #4992


hkaiser:
The remain clang-format issue was fixed on master

PR #5826, opened 2022-03-23
Reduce number of allocations while calling async_bulk_execute

hkaiser:
- reduce allocations by reducing the number of created futures and by circumventing using `std::vector` if not needed
- return a single `hpx::future<void>` for bulk execution of chunks that return `void`
- modernized chunk size computation


hkaiser:
> Performance test report

This performance regression has been fixed now.

PR #5837, opened 2022-04-08
Moving latch types to hpx and hpx::distributed namespaces

hkaiser:


gonidelis:
@hkaiser Should I added to the namespace changes or is it just for internal use?

PR #5842, opened 2022-04-09
change the default value of HPX_WITH_LCI_TAG to v1.7

JiakunYan:
Change the default value of HPX_WITH_LCI_TAG to the tag "v1.7".

Previously, it is a hardcoded commit hash value, which will be hard to update if we fix a bug in LCI in the future.


hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #5843, opened 2022-04-11
Remove staged threads, immediately create thread objects

hkaiser:
- flyby: more noexcept for combined_tagged_state


hkaiser:
This might not be viable after all. All threads that are created immediately either use a cached thread object (with an allocated stack) or creates a new thread object (without an associated stack) otherwise. This way, even threads that do not require immediate execution might end up with an allocated stack segment (if the thread object was retrieved from the cache).

This is prone to causing memory issues as too many threads/stacks may be created (the tests here expose exactly that problem). 

For now, I don't have any fool-proof idea on how to decide whether a thread needs to run right away, so it might benefit from being initialized from the thread object cache.

The only solution I see is more intrusive: in addition (or instead of) caching the thread objects, we will have to cache the stack segments. That would allow to minimize the number of created stack segments to the minimal number of concurrently active threads.

For now, I will convert this PR to 'daft' and get back to it after V1.8 has been released.

gonidelis:
What is the status of this PR? Should we push it for 1.10 instead?

hkaiser:
> What is the status of this PR? Should we push it for 1.10 instead?

Yes, let's move this to the next milestone.

hkaiser:
Closing as invalid.

PR #5844, opened 2022-04-11
Attempting to fix timeouts of segmented iterator tests

hkaiser:
- flyby: add missing dependency to segmented_algorithm module


hkaiser:
This indeed fixes all the timeouts. More problems with the segmented algorithms persist, however.

PR #5845, opened 2022-04-11
Adding local work requesting scheduler that is based on message passing internally

hkaiser:
This adds a new experimental work-requesting scheduler to the list of existing schedulers

- Using uniform_int_distribution with proper bounds
- Removing queue index from thread_queues as it was unused
- Renaming workstealing --> workrequesting
- Adding adaptive work stealing (steal half/steal one)
  - this makes this scheduler consistently (albeit only slightly) faster than the (default) local-priority scheduler
- Adding LIFO and FIFO variations of local work-stealing scheduler
  - flyby: fixing HPX_WITH_SWAP_CONTEXT_EMULATION
  - flyby: minor changes to fibonacci_local example
- Adding high- and low- priority queues
  - flyby: cache_line_data now does not generate warnings errors if padding is not needed
- Adding bound queues
- flyby: using cache_line_data for scheduler states

### DIFF
--- a/examples/1d_stencil/1d_stencil_8.cpp
+++ b/examples/1d_stencil/1d_stencil_8.cpp
@@ -24,6 +24,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <iostream>
+#include <memory>
 #include <mutex>
 #include <stack>
 #include <string>
@@ -575,7 +576,7 @@ stepper_server::space stepper_server::do_work(
     }
 
     // limit depth of dependency tree
-    hpx::sliding_semaphore sem(nd);
+    auto sem = std::make_shared<hpx::sliding_semaphore>(nd);
 
     for (std::size_t t = 0; t != nt; ++t)
     {
@@ -626,15 +627,15 @@ stepper_server::space stepper_server::do_work(
         // trigger the semaphore once computation has reached this point
         if ((t % nd) == 0)
         {
-            next[0].then([&sem, t](partition&&) {
+            next[0].then([sem, t](partition&&) {
                 // inform semaphore about new lower limit
-                sem.signal(t);
+                sem->signal(t);
             });
         }
 
         // suspend if the tree has become too deep, the continuation above
         // will resume this thread once the computation has caught up
-        sem.wait(t);
+        sem->wait(t);
     }
 
     return U_[nt % 2];
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -250,7 +250,8 @@ jobs:
                 -DHPX_WITH_FETCH_JSON=On \
                 -DCMAKE_EXPORT_COMPILE_COMMANDS=On \
                 -DHPX_WITH_DOCUMENTATION=On \
-                -DHPX_WITH_DOCUMENTATION_OUTPUT_FORMATS="${DOCUMENTATION_OUTPUT_FORMATS}"
+                -DHPX_WITH_DOCUMENTATION_OUTPUT_FORMATS="${DOCUMENTATION_OUTPUT_FORMATS}" \
+                -DHPX_WITH_TESTS_COMMAND_LINE=--hpx:queuing=local-workrequesting-fifo
       - persist_to_workspace:
           root: /hpx
           paths:
--- a/.jenkins/lsu/env-clang-13.sh
+++ b/.jenkins/lsu/env-clang-13.sh
@@ -25,6 +25,8 @@ configure_extra_options+=" -DHPX_WITH_PARCELPORT_MPI=ON"
 configure_extra_options+=" -DHPX_WITH_PARCELPORT_LCI=ON"
 configure_extra_options+=" -DHPX_WITH_FETCH_LCI=ON"
 configure_extra_options+=" -DHPX_WITH_LOGGING=OFF"
+configure_extra_options+=" -DHPX_WITH_TESTS_COMMAND_LINE=--hpx:queuing=local-workrequesting-fifo"
+
 
 # The pwrapi library still needs to be set up properly on rostam
 # configure_extra_options+=" -DHPX_WITH_POWER_COUNTER=ON"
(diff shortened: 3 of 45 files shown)


hkaiser:
First performance measurements show an overall improvement of up to 10%. Very promising!

Pansysk75:
@hkaiser 
Sorry I deleted your feedback with force-push, I remember it though :P
I'll let the tests run without the additional logging I had in the previous push, let's see what we get

hkaiser:
@Pansysk75 could you please fix the clang-format issues as well?

examples/1d_stencil/1d_stencil_8.cpp
```diff
@@ -575,7 +575,8 @@ stepper_server::space stepper_server::do_work(
     }
 
     // limit depth of dependency tree
-    hpx::sliding_semaphore sem(nd);
+    std::shared_ptr<hpx::sliding_semaphore> sem =
+        std::make_shared<hpx::sliding_semaphore>(nd);
```
hkaiser, line 579: Wouldn' it be more appropriate to fix the `sliding_semaphore` instead of working around it's deficiencies in user code?

Pansysk75: Sure! Although, technically, we could demand that the user keeps it alive, I suppose this is exactly the issue you've been fixing on other facilities recently, making sure the relevant data remains alive.

Mind if I take a stab at it? :)

hkaiser:
> I started looking at the performance of this scheduler a bit (on the single-NUMA-domain case)
> 
> Directly replacing our current scheduler, it performs worse in the general case (at least on the algorithm benchmarks, which I had handy), I also tried using the `do_not_combine_tasks` hint to bypass the index queues we use on bulk_execute, because that is technically another form of stealing, which would skew the comparison. It is still a bit worse.
> 
> In the case of a bulk-execution of relatively uniform work, stealing is very limited, because our scheduling is already quite balanced anyways. So this would explain the performance deficit, as we are possibly taking on a larger overhead (more complex scheduler) for small benefit (a few non-cache-disrupting steals).
> 
> And then, there is the fact that we are only able to poll the steal requests in between thread execution, which introduces some latency in responding to steal requests.
> 
> I will try to produce a best-case scenario for this scheduler though, even if it is just as a proof of concept.
> 
> Edit: it seems to perform much better on few cores, which could suggest a large number of failed stealing attempts when we have many cores. We can experiment with sending the steal request where there is actual work.

I did see improvements on tests that ran a large amount of separate tasks (like fibonacci). For uniform iterative parallelism the benefit would probably be small.

Pansysk75:
@hkaiser You'll still have to add the fix in 1d_stencil_8
https://github.com/STEllAR-GROUP/hpx/commit/451bbd02756155e94f2ca04cdfe3fb3695dc6043#diff-126ef56cea5c727b92b9ec59b26d8c429db7f6655405270d63adda430a4a5091
The sem->wait() is not guaranteed to block, so the lambda ends up with a reference to a deleted object.

hkaiser:
> @hkaiser You'll still have to add the fix in 1d_stencil_8 [451bbd0#diff-126ef56cea5c727b92b9ec59b26d8c429db7f6655405270d63adda430a4a5091](https://github.com/STEllAR-GROUP/hpx/commit/451bbd02756155e94f2ca04cdfe3fb3695dc6043#diff-126ef56cea5c727b92b9ec59b26d8c429db7f6655405270d63adda430a4a5091) The sem->wait() is not guaranteed to block, so the lambda ends up with a reference to a deleted object.

I thought that was fixed by: https://github.com/STEllAR-GROUP/hpx/pull/6294

Pansysk75:
> 
> I thought that was fixed by: #6294

@hkaiser  No, I think it cannot be fixed that way (that's why I had gotten a bit confused there). It's a plain old out-of-scope situation.
Just think of what happens if the semaphore doesn't need to wait for the sem.signal() (which will often happen, if the semaphore is not "saturated").
The created lambda (probably not executed yet) has a reference to the semaphore, which gets deleted if we go out of scope.
There is no way for the semaphore to know it is being referenced this way.

hkaiser:
@Pansysk75 I have applied the change to the use of the `sliding_semaphore` everywhere. This however doesn't seem to be sufficient to make things work: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/15639/workflows/eb35fe70-6509-4d54-80a1-3a4454e93d06/jobs/360130

Pansysk75:
> @Pansysk75 I have applied the change to the use of the `sliding_semaphore` everywhere. This however doesn't seem to be sufficient to make things work: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/15639/workflows/eb35fe70-6509-4d54-80a1-3a4454e93d06/jobs/360130

Seems like tasks get stuck in the stealing queue, re-applying this fix solves the issue:
https://github.com/STEllAR-GROUP/hpx/commit/451bbd02756155e94f2ca04cdfe3fb3695dc6043 

Did you do something else to mitigate that issue, or was it accidentally left behind?

hkaiser:
> > @Pansysk75 I have applied the change to the use of the `sliding_semaphore` everywhere. This however doesn't seem to be sufficient to make things work: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/15639/workflows/eb35fe70-6509-4d54-80a1-3a4454e93d06/jobs/360130
> 
> Seems like tasks get stuck in the stealing queue, re-applying this fix solves the issue: [451bbd0](https://github.com/STEllAR-GROUP/hpx/commit/451bbd02756155e94f2ca04cdfe3fb3695dc6043)
> 
> Did you do something else to try solve that issue, or was this fix accidentally left behind?

Thanks a  lot - not sure how that got lost. Much appreciated!

hkaiser:
I think this is good to go now. Thanks again @Pansysk75!

hkaiser:
bors merge

PR #5850, opened 2022-04-14
Convert perf_counters::counter_type enum to enum class.

gonidelis:
Convert `performance_counters::counter_type` according to #4702.

todo: convert `discover_counters_mode` and `counter_status`.


libs/full/performance_counters/src/counters.cpp
```diff
@@ -506,7 +506,7 @@ namespace hpx { namespace performance_counters {
     {
         if (type < counter_text || type > counter_raw_values)
```
gonidelis, line 507: @hkaiser Shouldn't the unscoped `counter_raw_values` here provide a deprecation warning? I don't see any when I run locally.

libs/full/performance_counters/include/hpx/performance_counters/counters_fwd.hpp
```diff
@@ -28,31 +28,31 @@ namespace hpx { namespace performance_counters {
...
+    enum class counter_type
     {
         // \a counter_text shows a variable-length text string. It does not
         // deliver calculated values.
         //
         // Formula:  None
         // Average:  None
         // Type:     Text
-        counter_text,
+        counter_text = 0,
```
hkaiser, line 39: I'd suggest renaming the enumeration values to reduce the naming noise, e.g. `counter_type::raw`, etc..

libs/full/performance_counters/src/counters.cpp
hkaiser: You'd have to touch on all code that uses the enumeration values. Otherwise, the CI builders will fail (as soon as we bump the version to V1.9).

libs/full/performance_counters/include/hpx/performance_counters/counters_fwd.hpp
gonidelis: done

libs/full/performance_counters/src/counters.cpp
gonidelis: Great! Done on my last commit. Let's resolve and wait for the tests.

libs/full/performance_counters/include/hpx/performance_counters/counters_fwd.hpp
```diff
@@ -28,44 +28,44 @@ namespace hpx { namespace performance_counters {
...
     {
-        // \a counter_text shows a variable-length text string. It does not
+        // \a text counter shows a variable-length text string. It does not
         // deliver calculated values.
         //
         // Formula:  None
         // Average:  None
         // Type:     Text
-        counter_text,
+        text = 0,
```
hkaiser, line 39: Do we need the explicit `= N`? I don't think we care about particular values here.

gonidelis: Let me try it. Thanks for the tip.

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #5851, opened 2022-04-14
Forking dynamic_bitset from Boost

hkaiser:
- moving serialization of optional and any to datastructures module


hkaiser:
retest cscs

PR #5852, opened 2022-04-17
Optimize LCI parcelport

JiakunYan:
Optimize LCI parcelport.

Done:
- only use a single progress thread to call LCI_progress;
- remove all the locks around LCI calls;
- use LCI_sendm or LCI_sendl based on message size. Previously, we only use LCI_sendl.
- simultaneously post multiple send/recv for the same connection. Only synchronize in the end.


gonidelis:
Thanks for the amazing work!

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #5854, opened 2022-04-18
Fixing left-overs from changing counter_type to enum class

hkaiser:


gonidelis (approved):
LGTM Thanks for the clean-up.

gonidelis:
@hkaiser Seems unrelated but why is thread stacksize failing?

hkaiser:
> @hkaiser Seems unrelated but why is thread stacksize failing?

I don't know. Let's find out!

PR #5855, opened 2022-04-18
Adding performance test Jenkins builder at LSU

hkaiser:


hkaiser:
@Pansysk75 perfect! This seems to work as expected now! Thanks a lot!

Pansysk75:
> @Pansysk75 perfect! This seems to work as expected now! Thanks a lot!

I think the above StellarBot comment was from the Daint cluster and not from Rostam. The Rostam Jenkins has't picked up on it yet, maybe it needs a " retest "?

Pansysk75:
retest

Pansysk75:
retest lsu

Pansysk75:
@hkaiser 
Ok, it seems to work all the way through. The report results seem a bit off, I'll need to check whether that's due to general inconsistency or due to a bad performance reference.

hkaiser:
> Ok, it seems to work all the way through. 

Great!

> The report results seem a bit off, I'll need to check whether that's due to general inconsistency or due to a bad performance reference.

The current reference contained in this PR was copied from the Daint performance tests. We need to create a new reference set based on the data collected on Rostam.

hkaiser:
@Pansysk75 also, please let me know once you think that this can be merged.

hkaiser:
Let's merge this now. It might need more work, however.

hkaiser:
bors merge

PR #5857, opened 2022-04-19
Convert enum counter_status to enum class

gonidelis:
Convert `enum counter_status` to `enum class counter_status` and remove `status_` prefixes from status values.

### DIFF
--- a/libs/full/performance_counters/include/hpx/performance_counters/counters_fwd.hpp
+++ b/libs/full/performance_counters/include/hpx/performance_counters/counters_fwd.hpp
@@ -193,16 +193,49 @@ namespace hpx { namespace performance_counters {
     ///////////////////////////////////////////////////////////////////////////
     // Status and error codes used by the functions related to
     // performance counters.
-    enum counter_status
+    enum class counter_status
     {
-        status_valid_data,         // No error occurred, data is valid
-        status_new_data,           // Data is valid and different from last call
-        status_invalid_data,       // Some error occurred, data is not value
-        status_already_defined,    // The type or instance already has been defined
-        status_counter_unknown,         // The counter instance is unknown
-        status_counter_type_unknown,    // The counter type is unknown
-        status_generic_error            // A unknown error occurred
+        valid_data,         // No error occurred, data is valid
+        new_data,           // Data is valid and different from last call
+        invalid_data,       // Some error occurred, data is not value
+        already_defined,    // The type or instance already has been defined
+        counter_unknown,    // The counter instance is unknown
+        counter_type_unknown,    // The counter type is unknown
+        generic_error            // A unknown error occurred
     };
+
+    inline std::ostream& operator<<(
+        std::ostream& os, counter_status const& rhs) noexcept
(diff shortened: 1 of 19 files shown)


gonidelis:
@hkaiser Ah... I wanted to open a draft because I get an error here:

https://github.com/STEllAR-GROUP/hpx/pull/5857/files#diff-f3eaea65580273671882d4ec78512dc450029c8886cdeaa6cd88ba744ed7be8dR205

`‘enum class hpx::performance_counters::counter_status’ is not a class or namespace`

hkaiser:
> @hkaiser Ah... I wanted to open a draft because I get an error here:
> 
> https://github.com/STEllAR-GROUP/hpx/blob/9ca801475733aad99410614b3a38c140ff735487/libs/full/performance_counters/include/hpx/performance_counters/counters_fwd.hpp#L205
> 
> `‘enum class hpx::performance_counters::counter_status’ is not a class or namespace`
> 
> I tried both and without return by ref.

Sure, enum classes can't have function members. You will need to create a free function to do what you want.

gonidelis:
> > @hkaiser Ah... I wanted to open a draft because I get an error here:
> > https://github.com/STEllAR-GROUP/hpx/blob/9ca801475733aad99410614b3a38c140ff735487/libs/full/performance_counters/include/hpx/performance_counters/counters_fwd.hpp#L205
> > 
> > `‘enum class hpx::performance_counters::counter_status’ is not a class or namespace`
> > I tried both and without return by ref.
> 
> Sure, enum classes can't have function members. You will need to create a free function to do what you want.

Turns out the operator overloading wasn't needed after all. Overloading `<<` was needed instead. All tests pass locally. Please verify.

libs/full/performance_counters/include/hpx/performance_counters/counters_fwd.hpp
```diff
@@ -191,17 +191,50 @@ namespace hpx { namespace performance_counters {
...
+        valid_data,         // No error occurred, data is valid
+        new_data,           // Data is valid and different from last call
+        invalid_data,       // Some error occurred, data is not value
+        already_defined,    // The type or instance already has been defined
+        counter_unknown,    // The counter instance is unknown
+        counter_type_unknown,    // The counter type is unknown
+        generic_error            // A unknown error occurred
     };
 
+    inline constexpr std::ostream& operator<<(
```
gonidelis, line 205: @K-ballo Any ideas on how to tackle the `all paths through this function will call itself` error here? If `counter_status` was just a class and not an enum class I would just provide a `friend` overload for the operator inside the class body. But now...?

hkaiser: You need to cast `rhs` to an integer, then it will do what you want.

gonidelis: `static_cast` or just `int`? and why?

hkaiser:
bors merge

PR #5859, opened 2022-04-20
Splitting documentation generation steps on CircleCI

hkaiser:
CircleCI documentation generation times out for master and release branches (after recent changes). This PR slits the generation of the html, singlehtml, and latexpdf docs into separate steps in the hope that those will not time out separately anymore.


hkaiser:
This looks better. Generating HTML docs doesn't time out anymore.

PR #5860, opened 2022-04-21
Updating APEX tag

hkaiser:


hkaiser:
@khuck this moves HPX to the APEX release v.2.5.0. We however see problems now (e.g.: https://github.com/STEllAR-GROUP/hpx/runs/6104634398?check_suite_focus=true). Any clues?

khuck:
looks like you resolved it... yes, the JSON trace output writes out compressed text, so zlib is required.

hkaiser:
> looks like you resolved it... yes, the JSON trace output writes out compressed text, so zlib is required.

I will add zlib to our docker images, that will resolve the issue, indeed. Thanks!

hkaiser:
@khuck: with your latest changes, this test started to fail: https://cdash.cscs.ch/test/12823993. Would you be able to have a look?

msimberg:
@hkaiser, @khuck I noticed the same problem in pika while briefly trying APEX develop. The output format with `APEX_SCREEN_OUTPUT` has changed and this affects the validation of the test output. Most likely everything is still fine with the APEX integration.

The tasks used to be sorted alphabetically which made it easy-ish to have a fixed regex to match the expected output. The tasks are now sorted by time spent which is not the same as the fixed output and nondeterministic.

@khuck is there any way to control the way the output is formatted?

hkaiser:
> @msimberg @hkaiser No, there's no way to control the output... I can confirm that I did make that change, to make it easier to find actual performance issues.

@khuck: what should we do with the test now?

khuck:
Can't you search for the regular expressions individually, and not in order?

hkaiser:
> Can't you search for the regular expressions individually, and not in order?

@khuck: I don't think there is a way to achieve what you suggested, at least not without completely redoing the whole test setup. For now I have disabled the test. If you happen to add an option that controls the output format generated at some point, we can go back an re-enable it.

khuck:
I can add an option to leave the timers sorted alphabetically, if that's what you mean...

hkaiser:
> I can add an option to leave the timers sorted alphabetically, if that's what you mean...

Yep, that's what I meant - just didn't mean to rush you in any way.

PR #5866, opened 2022-04-24
add cmake option HPX_WITH_PARCELPORT_COUNTERS

JiakunYan:
Turn on/off the performance counters related to parcelport
Those counters can be expensive due to the locks when gathering data

### DIFF
--- a/docs/sphinx/manual/optimizing_hpx_applications.rst
+++ b/docs/sphinx/manual/optimizing_hpx_applications.rst
@@ -1008,10 +1008,15 @@ system and application performance.
        (see ``<operation``, e.g. ``en`` or ``eceived``) for the specified
        ``<connection_type>``.
 
+       The performance counters are available only if the compile time constant
+       ``HPX_HAVE_PARCELPORT_COUNTERS`` was defined while compiling the |hpx|
+       core library (which is not defined by default). The corresponding cmake
+       configuration constant is ``HPX_WITH_PARCELPORT_COUNTERS``.
+
        The performance counters for the connection type ``mpi`` are available
        only if the compile time constant ``HPX_HAVE_PARCELPORT_MPI`` was defined
-       while compiling the |hpx| core library (which is not defined by default,
-       the corresponding cmake configuration constant is
+       while compiling the |hpx| core library (which is not defined by default).
+       The corresponding cmake configuration constant is
        ``HPX_WITH_PARCELPORT_MPI``.
 
        Please see :ref:`cmake_variables` for more details.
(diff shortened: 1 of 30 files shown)


libs/full/parcelport_lci/include/hpx/parcelport_lci/sender_connection.hpp
```diff
@@ -291,7 +291,9 @@ namespace hpx::parcelset::policies::lci {
             buffer_.data_point_.time_ =
```
hkaiser, line 291: Should this be disabled as well?

JiakunYan: I am hesitant because if we disable all the `buffer_.data_point_` related lines, there will be too many `ifdef` all over the code. I am kind of hoping the compiler's dead code elimination is smart enough to remove these lines.

hkaiser: > I am hesitant because if we disable all the `buffer_.data_point_` related lines, there will be too many `ifdef` all over the code. I am kind of hoping the compiler's dead code elimination is smart enough to remove these lines.

Right. I would at least suggest removing the invocations of the timers. Those introduce a bit of avoidable overhead. In general, I wouldn't rely on the compiler to be smart enough to eliminate unneeded code.

JiakunYan: Okay, let me think about it. Maybe a better way is to create a set method for each field of the data point so that we can directly replace the methods with NULL methods.

JiakunYan: Eventually, I fell back to the trivial solution. It's hard to find an easy way to encapsulate the use of timer within a single method of data_point.

hkaiser (dismissed):
LGTM, thanks!

docs/sphinx/manual/optimizing_hpx_applications.rst
```diff
@@ -1008,6 +1008,11 @@ system and application performance.
        (see ``<operation``, e.g. ``en`` or ``eceived``) for the specified
        ``<connection_type>``.
 
+       The performance counters are available only if the compile time constant
+       ``HPX_HAVE_PARCELPORT_COUNTERS`` was defined while compiling the |hpx|
+       core library (which is not defined by default, the corresponding cmake
+       configuration constant is ``HPX_WITH_PARCELPORT_COUNTERS``.
```
hkaiser, line 1014: This misses a closing ')'. It looks like we missed that in other places as well. Could you add it everywhere while you're at it, please?

JiakunYan: Done. Also fix the same issues on the HPX_HAVE_PARCELPORT_MPI comments.

hkaiser (dismissed):
LGTM, thanks!

hkaiser:
Several of the tests are now failing as those did depend on querying one of the disabled counters. Could you please either disable  (#ifdef/#endif) those parts of the tests that depend on them? Also, please enable the counters for some of the builders.

JiakunYan:
> Several of the tests are now failing as those did depend on querying one of the disabled counters. Could you please either disable (#ifdef/#endif) those parts of the tests that depend on them? Also, please enable the counters for some of the builders.

I enabled the counters for all Debug tests. Is this what you want?

JiakunYan:
@hkaiser I think this PR is ready to merge. Do you have any more comments?

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #5867, opened 2022-04-25
Update CMakeLists.txt

khuck:
Updating to APEX version 2.5.1


hkaiser:
@khuck thanks for this PR. Would it be ok if I merged this with the already existing #5860?

hkaiser:
@khuck, I retargeted this PR to be based on #5860, this way merging it will do the right thing.

khuck:
> 

no problem, sounds good to me!

PR #5868, opened 2022-04-26
Separating local parts of compute module into core libraries

hkaiser:
Fixes #4502

PR #5872, opened 2022-04-30
Block fork join executor

hkaiser:
The `block_fork_join_executor` is building on top of a set of plain `fork_join_executor`s such that those build a hierarchy of executors allowing to restrict the necessary synchronization to a subset of cores for each of those executors. This allows to build for instance a NUMA-aware `fork_join_executor`, where each of the NUMA-domains has its own executor and those are combined into a higher-level executor that performs synchronization between the NUMA-domains.

The hope is that this executor helps reducing cross-NUMA domain synchronization. 

This PR contains a couple of flyby changes:

- adding support for `thread_priority::bound` to the default scheduler
- replace some of the uses of `boost::irange` with our `hpx::util::detail::make_counting_shape`
- added `thread_description(std::string desc)` constructor
- added the use of the new executor to the stream benchmark

This PR relies on #5868.

### DIFF
--- a/libs/core/affinity/src/affinity_data.cpp
+++ b/libs/core/affinity/src/affinity_data.cpp
@@ -206,7 +206,7 @@ namespace hpx { namespace threads { namespace policies { namespace detail {
             return ret;
         }
 
-        for (std::size_t thread_num = 0; thread_num < num_threads_;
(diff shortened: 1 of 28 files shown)


hkaiser:
bors merge

PR #5864, opened 2022-04-22
ATOMIC_INIT_FLAG is deprecated starting C++20

hkaiser:
- this disables the use of ATOMIC_INIT_FLAG if needed (clang 14 complains about its use)

PR #5881, opened 2022-05-02
Fixing recent patch that disables ATOMIC_FLAG_INIT for C++20 and up

hkaiser:
The original PR #5864 introduced problems on older platforms. This hopefully fixes those.


hkaiser:
bors merge

PR #5883, opened 2022-05-04
Optimization on LCI parcelport: uses LCI_putva

JiakunYan:
### Background

LCI_putva: put iovec allocate

- LCI_putva transfers an `iovec`, which includes one medium message and multiple long messages. 
- It is a one-sided put operation, which means the target side doesn't need to post a `recv`. Instead, a completion entry will be pushed to a preconfigured LCI completion queue when all messages arrive. 
- Unlike RDMA, the sender side doesn't need to specify a remote registered buffer. LCI runtime will allocate and register the receive buffer.

### Changes on HPX parcelport layer
- No tag management on HPX layer, so no mutex lock.
- Only one "posting send and testing" on the sender side. (It can be further optimized.)
- No "posting receive and testing" on the receive side, so receiver_connection can be removed.


hkaiser:
@JiakunYan since we have finally finished the release, is there anything you'd still like to do here before merging it?

JiakunYan:
@hkaiser Do you have any idea why the test "jenkins/lsu/gcc-12-release" failed?

libs/full/parcelport_lci/include/hpx/parcelport_lci/receiver.hpp
```diff
@@ -17,15 +17,57 @@
...
 #include <iterator>
 #include <list>
 #include <memory>
 #include <mutex>
 #include <set>
 #include <utility>
+#include <vector>
 
 namespace hpx::parcelset::policies::lci {
+    struct BufferWrapper
```
hkaiser, line 32: Could you please conform to the HPX naming conventions (not `CamelCase`, but `snake_case`) (see: https://github.com/STEllAR-GROUP/hpx/blob/master/.github/CONTRIBUTING.md#general-info)?

libs/full/parcelport_lci/include/hpx/parcelport_lci/receiver.hpp
```diff
@@ -46,54 +90,157 @@ namespace hpx::parcelset::policies::lci {
         bool background_work() noexcept
         {
             // We first try to accept a new connection
-            connection_ptr connection = accept();
+            LCI_request_t request = accept();
```
hkaiser, line 93: Wrapping the `LCI_request_t` into a RAII type that releases it at destruction time would improve exception safety.

libs/full/parcelport_lci/include/hpx/parcelport_lci/receiver.hpp
```diff
@@ -46,54 +90,157 @@ namespace hpx::parcelset::policies::lci {
...
-            {
-                receive_messages(HPX_MOVE(connection));
+                buffer_type buffer = decode_request(request);
+                decode_parcels(pp_, HPX_MOVE(buffer), -1);
+                free_request(request);
                 return true;
             }
-
             return false;
+            //            // If we don't have a new connection, try to handle one of the
```
hkaiser, line 103: Can this (commented) code be removed?

hkaiser (changes requested):

hkaiser (approved):
LGTM, thanks!

hkaiser (approved):

hkaiser:
bors merge

PR #5884, opened 2022-05-04
Fix miscellaneous doc sections

dimitraka:
- Remove HPX main library from Full API
- New additions to troubleshooting
- Update contributors
- Acknowledgements to GSoC and GSoD
- Add dependency report 
- Skip generation of files lacking documentation in Full API


gonidelis:
> Why is this necessary? `` Your application’s build type should align with the HPX build type. For example, if you specified ``-DCMAKE_BUILD_TYPE=Debug`` during the |hpx| compilation, then your application needs to be compiled ``
> 
> If I trust hpx, I could debug my code, while benefitting from hpx being compiled in release mode?

We have witnessed inconsistencies and compilation failures when the two flags do not align. At the end of the day your error might stem from HPX and you might wanna find out what part in the HPX code is to blame. Also, for perf. analysis reasons we just want to highlight that Release mode is on every side in order for the results to be trustworthy.

HTH

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

hkaiser:
@gonidelis could you add this PR to the release, please?

gonidelis:
> @gonidelis could you add this PR to the release, please?

done

PR #5897, opened 2022-05-26
Add missing ""

Neumann-A:
Fixes pkg-config file generation on windows where CMAKE_SHARED_LIBRARY_PREFIX is empty. 

Always quote string like inputs to functions because they can be empty.

## Checklist

- [x] I have fixed a bug ~~and have added a regression test~~. (just run with HPX_WITH_PKGCONFIG=ON on windows)


hkaiser:
What problem does this patch solve?

hkaiser (approved):
LGTM, thanks!

hkaiser:

PR #5769, opened 2022-02-08
Slight improvement of small_vector performance

hkaiser:
- adding benchmark

PR #5899, opened 2022-05-27
Replace PMR based version of small_vector

hkaiser:
- this significantly improves its performance

This is a follow-up to #5769


msimberg:
@hkaiser I pushed the updated references for two of the tests. Unsure if the stream reference also needs updating (it was slower on the second run until now), so leaving that out for now. Let me know if the new references are no good.

hkaiser:
retest cscs

hkaiser:
The HIP and CUDA failures are caused by configuration problems on rostam and will be fixed separately.

hkaiser:
bors merge

PR #5904, opened 2022-05-31
Delete FindBZip2.cmake

Neumann-A:
Superior:
https://cmake.org/cmake/help/latest/module/FindBZip2.html
https://github.com/Kitware/CMake/blob/master/Modules/FindBZip2.cmake

Also currently misses the case of bz2d for the debug library.....
Furthermore hpx requires cmake 3.18 so why go with outdated modules?
https://github.com/STEllAR-GROUP/hpx/blob/c2e2e1b2a0671d8afcdeecb4c2de87f38e8770c8/CMakeLists.txt#L17


hkaiser:
bors merge

PR #5905, opened 2022-05-31
Fix incorrect usage of generator expressions

Neumann-A:
`$<TARGET_NAME:...>` is required for the install interface

Don't yet know what all the code does but it generates wrong stuff in the INSTALL_INTERFACE which needs to be informed about target names. 

see
https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:TARGET_NAME

Fixes using static exports of HPX on windows.


hkaiser:
The remaining test errors are unrelated.

hkaiser (approved):
LGTM, thanks!

PR #5906, opened 2022-05-31
Use BUILD_SHARED_LIBS correctly

Neumann-A:
see https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html

- [x] I have fixed a bug

`CMAKE_SHARED_LIBS` doesn't exist


hkaiser (approved):
LGTM, thanks!

PR #5910, opened 2022-06-03
Improve MPI initialization

hkaiser:
- explicitly set number of MPI background threads to one if MPI is not multi-threaded
- improve generated error message if required threading mode is not supported

This makes it simpler to control the used threading level for MPI


hkaiser:
this is now subsumed by #6229. Closing.

Issue #199, opened 2012-07-10
--dump-config --exit should be supported


brycelelbach:
**[reported by blelbach]** _[Trac time Thu Oct 13 02:05:28 2011]_ Currently, doing --dump-config --exit doesn't work. I added an explicit throw for this the other day, as I didn't want to risk screwing up the command line logic. This would be nice to have at some point, though.

PR #5916, opened 2022-06-09
Fix minor doxygen parsing typo

gonidelis:
Fixes minor typo that might cause hyperlink to break in Doxygen. The warning stems from line #199 and everything below that fails to link to our docs page. Let's merge it as it is a typo anyways and see if it also fixes the linking problem.

I don't think the `\a` clause has to be an one-liner, does it?

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/container_algorithms/partial_sort.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/container_algorithms/partial_sort.hpp
@@ -174,11 +174,11 @@ namespace hpx { namespace ranges {
     /// calling thread.
     ///
     /// \returns  The \a partial_sort algorithm returns \a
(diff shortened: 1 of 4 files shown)


gonidelis:
I summon @bhumitattarde @dimitraka on that.

hkaiser (changes requested):
A cursory grep for `'   ::type>'` brings up at least one more similar problem in sort.hpp. Could you please fix that as well?

gonidelis:
> A cursory grep for `' ::type>'` brings up at least one more similar problem in sort.hpp. Could you please fix that as well?

Ok I will track them down all

gonidelis:
@hkaiser  Let's see if that fixes the problem.

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #5917, opened 2022-06-10
Fixing various 'held lock while suspending' problems

hkaiser:
- flyby: introduce HPX__ASSERT_LOCKED() et.al. facilities that unlock a given lock
  before invoking the assert


hkaiser:
The LSU-CUDA and LSU-HIP builder issues are unrelated (those are rostam configuration problems that will be fixed separately)

hkaiser:
bors merge

PR #5920, opened 2022-06-12
[P2300] enhancements: receiver_of, sender_of improvements

SAtacker:

## Proposed Changes

  - The `receiver_of` concept takes a receiver and an instance of the `completion_signatures<>` class template.
  - The `sender_of` concept defines the requirements for a sender type that on successful completion sends the specified set of value types.

## Checklist

- [x] receiver_of tests
- [x] sender_of tests

### DIFF
--- a/libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
+++ b/libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
@@ -383,7 +383,7 @@ namespace hpx::execution::experimental {
 
     // The sender_of concept defines the requirements for a sender type that on
     // successful completion sends the specified set of value types.
-    template <class S, class R>
+    template <class S, class E = no_env, class... Ts>
     struct is_sender_of;
 
     namespace detail {
@@ -607,56 +607,35 @@ namespace hpx::execution::experimental {
     inline constexpr bool is_sender_v = is_sender<Sender, Env>::value;
 
     namespace detail {
-
-        // Sender may or may not support sends_stopped
-        // In the false case only the value_types and error_types are checked
-        template <class S, class R, class CS,
-            bool sends_stopped = CS::sends_stopped>
-        struct sender_support_sends_stopped;
-
-        template <class S, class R, class CS>
-        struct sender_support_sends_stopped<S, R, CS, true>
-          : std::integral_constant<bool,
-                is_receiver_of_v<R, completion_signatures_of_t<S, env_of_t<R>>>>
-        {
-        };
-
-        template <class S, class R, class CS>
-        struct sender_support_sends_stopped<S, R, CS, false>
-          : std::integral_constant<bool,
-                is_invocable_variant_of_tuples<set_value_t, R,
-                    typename CS::template value_types<hpx::tuple,
-                        hpx::variant>>::value &&
-                    is_invocable_variant<set_error_t, R,
-                        typename CS::template error_types<hpx::variant>>::value>
-        {
-        };
-
-        template <bool IsSenderOf, class S, class R>
+        template <bool IsSenderOf, typename S, typename E, typename... Ts>
         struct is_sender_of_impl;
 
-        template <class S, class R>
-        struct is_sender_of_impl<false, S, R> : std::false_type
+        template <typename S, typename E, typename... Ts>
+        struct is_sender_of_impl<false, S, E, Ts...> : std::false_type
         {
         };
 
-        template <class S, class R>
-        struct is_sender_of_impl<true, S, R>
+        template <typename CS, typename... Ts>
+        inline bool constexpr is_same_types = std::same_as<
+            typename CS::template value_types<meta::pack, meta::pack>,
+            meta::pack<meta::pack<Ts...>>>;
+
+        template <class S, class E, class... Ts>
+        struct is_sender_of_impl<true, S, E, Ts...>
           : std::integral_constant<bool,
-                sender_support_sends_stopped<S, R,
-                    completion_signatures_of_t<S, env_of_t<R>>>::value>
+                is_same_types<completion_signatures_of_t<S, E>, Ts...>>
         {
         };
     }    // namespace detail
 
-    template <class S, class R>
+    template <class S, class E, class... Ts>
     struct is_sender_of
-      : detail::is_sender_of_impl<is_sender_v<S, env_of_t<R>>, S, R>
+      : detail::is_sender_of_impl<is_sender_v<S, E>, S, E, Ts...>
     {
     };
 
-    template <class S, class R>
-    inline constexpr bool is_sender_of_v = is_sender_of<S, R>::value;
+    template <typename S, typename E, typename... Ts>
+    inline constexpr bool is_sender_of_v = is_sender_of<S, E, Ts...>::value;
 
     template <typename Sender, typename Receiver>
     struct is_sender_to
--- a/libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
+++ b/libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
@@ -10,6 +10,7 @@
 #include <hpx/datastructures/tuple.hpp>
 #include <hpx/datastructures/variant.hpp>
 #include <hpx/execution_base/get_env.hpp>
+#include <hpx/execution_base/receiver.hpp>
 #include <hpx/execution_base/sender.hpp>
 #include <hpx/functional/detail/tag_fallback_invoke.hpp>
 #include <hpx/functional/invoke_result.hpp>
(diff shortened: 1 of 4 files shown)


libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
```diff
@@ -600,6 +606,58 @@ namespace hpx::execution::experimental {
     template <typename Sender, typename Env = empty_env>
     inline constexpr bool is_sender_v = is_sender<Sender, Env>::value;
 
+    namespace detail {
+
+        // Sender may or may not support sends_stopped
+        // In the false case only the value_types and error_types are checked
+        template <class S, class R, class CS,
+            bool sends_stopped = CS::sends_stopped>
```
hkaiser, line 614: What happens if the sender's completion signatures don't expose a `sends_stopped`?

libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
```diff
@@ -600,6 +606,58 @@ namespace hpx::execution::experimental {
...
+        struct sender_support_sends_stopped;
+
+        template <class S, class R, class CS>
+        struct sender_support_sends_stopped<S, R, CS, true>
+          : std::integral_constant<bool,
+                is_receiver_of_v<R, completion_signatures_of_t<S, env_of_t<R>>>>
```
hkaiser, line 620: Did you mean:
```suggestion
                is_receiver_of_v<R, CS>>
```

libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
```diff
@@ -600,6 +606,58 @@ namespace hpx::execution::experimental {
...
+        {
+        };
+
+        template <class S, class R, class CS>
+        struct sender_support_sends_stopped<S, R, CS, false>
+          : std::integral_constant<bool,
+                is_invocable_variant_of_tuples<set_value_t, R,
+                    typename CS::template value_types<hpx::tuple,
```
hkaiser, line 628: What if the sender's completion signatures don't expose a dependent type `value_types`?

libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
```diff
@@ -600,6 +606,58 @@ namespace hpx::execution::experimental {
...
+                        hpx::variant>>::value &&
+                    is_invocable_variant<set_error_t, R,
+                        typename CS::template error_types<hpx::variant>>::value>
```
hkaiser, line 631: Same here, what if the sender's completion signatures don't expose a dependent type `error_types`?

hkaiser, line 631: I think this would be sufficient:
```suggestion
                        typename CS::template error_types<meta::pack>>::value>
```

libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
```diff
@@ -600,6 +606,58 @@ namespace hpx::execution::experimental {
...
+                is_receiver_of_v<R, completion_signatures_of_t<S, env_of_t<R>>>>
+        {
+        };
+
+        template <class S, class R, class CS>
+        struct sender_support_sends_stopped<S, R, CS, false>
+          : std::integral_constant<bool,
+                is_invocable_variant_of_tuples<set_value_t, R,
+                    typename CS::template value_types<hpx::tuple,
+                        hpx::variant>>::value &&
```
hkaiser, line 629: Can we simplify this?
```suggestion
                    typename CS::template value_types<meta::pack,
                        meta::pack>>::value &&
```

libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
```diff
@@ -600,6 +606,58 @@ namespace hpx::execution::experimental {
     template <typename Sender, typename Env = empty_env>
     inline constexpr bool is_sender_v = is_sender<Sender, Env>::value;
 
+    namespace detail {
+
+        // Sender may or may not support sends_stopped
+        // In the false case only the value_types and error_types are checked
+        template <class S, class R, class CS,
```
hkaiser, line 613: Not critical, but more of a consistency issue: in HPX we generally use `typename` instead of `class` for template parameter definitions.

libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
```diff
@@ -600,6 +606,58 @@ namespace hpx::execution::experimental {
...
+                    completion_signatures_of_t<S, env_of_t<R>>>::value>
+        {
+        };
+    }    // namespace detail
+
+    template <class S, class R>
+    struct is_sender_of
+      : detail::is_sender_of_impl<is_sender_v<S, env_of_t<R>>, S, R>
+    {
+    };
```
hkaiser, line 656: Apparently, we have overdone this. P2300 defines `sender_of` as:
```
template<class S, class E = no_env, class... Ts>
  concept sender_of =
    sender<S, E> &&
    same_as<
      type-list<Ts...>,
      value_types_of_t<S, E, type-list, type_identity_t>
    >;
```
IOW, all we need to check is whether the possible arguments to `set_value()` match the given list of arguments. The confusion was caused (well, my confusion) by the fact that there are two different concepts: `sender_of` and `sender_to`. The first checks the value-types only while the second verifies that `connect()` can be called on a particular pair of a sender and a receiver:
```
template<class S, class R>
  concept sender_to =
    sender<S, env_of_t<R>> &&
    receiver_of<R, completion_signatures_of_t<S, env_of_t<R>>> &&
    requires (S&& s, R&& r) {
      execution::connect(std::forward<S>(s), std::forward<R>(r));
    };
```

To make the picture complete, there is a `receiver_of` concept that verifies whether a given receiver type can handle the given completion signatures:
```
template <class T, class Completions>
  concept receiver_of =
    receiver<T> &&
    requires (Completions* completions) {
        []<valid-completion-for<T>...Sigs>(completion_signatures<Sigs...>*)
        {}(completions);
    };
```

I apologize for the confusion but it seems you will have to revisit your implementation of `is_sender_of` one more time.

SAtacker: Hi, thanks for clarifying, I was going to do it https://github.com/STEllAR-GROUP/hpx/commit/86cca5e08514ce457bbdab472a51b728cadfa4f3 but then I thought maybe it's supposed to be this way. I could have mentioned it to you, sorry.
Let me redo this and also think over the other comments.
Thanks

libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
SAtacker: Noted, will change it.

libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
SAtacker: Yes, got it.

libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
SAtacker: It will not reach this point as `is_sender_v` will evaluate as false. Thus `is_sender_of` is false

libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
SAtacker: Yes, thanks. got it.

libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
SAtacker: [This](https://github.com/SAtacker/hpx/blob/p2300_enhancements/libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp#L654) will evaluate as false

SAtacker: Yes, thanks

libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
SAtacker: > IOW, all we need to check is whether the possible arguments to set_value() match the given list of arguments.

IOW, we need to check if set_value is invocable with the given list of arguments?

SAtacker: If it is invocable then it implies that the types match, is this implementation okay or I should go by with some other mechanism?

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #5923, opened 2022-06-13
Fixing git checkout to reflect latest version tag

hkaiser:
Makes sure that the 'Getting HPX' documentation page always refers to the latest version tag


hkaiser:
This will be solved by merging #5926. Closing...

PR #5926, opened 2022-06-15
Fix doxygen warnings when building documentation

dimitraka:
Fix warnings including missing arguments, deprecations, link requests etc.

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_default_construct.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_default_construct.hpp
@@ -11,8 +11,8 @@
 #if defined(DOXYGEN)
 namespace hpx {
 
-    /// Constructs objects of type typename iterator_traits<ForwardIt>
-    /// ::value_type in the uninitialized storage designated by the range
+    /// Constructs objects of type typename iterator_traits<ForwardIt> 
+    /// in the uninitialized storage designated by the range
     /// by default-initialization. If an exception is thrown during the
     /// initialization, the function has no effects.
     ///
@@ -38,7 +38,7 @@ namespace hpx {
     void uninitialized_default_construct(FwdIter first, FwdIter last);
 
     /// Constructs objects of type typename iterator_traits<ForwardIt>
-    /// ::value_type in the uninitialized storage designated by the range
+    /// in the uninitialized storage designated by the range
     /// by default-initialization. If an exception is thrown during the
     /// initialization, the function has no effects.
     ///
@@ -81,7 +81,7 @@ namespace hpx {
         ExPolicy&& policy, FwdIter first, FwdIter last);
 
     /// Constructs objects of type typename iterator_traits<ForwardIt>
-    /// ::value_type in the uninitialized storage designated by the range
+    /// in the uninitialized storage designated by the range
     /// [first, first + count) by default-initialization. If an exception
     /// is thrown during the initialization, the function has no effects.
     ///
@@ -113,7 +113,7 @@ namespace hpx {
     FwdIter uninitialized_default_construct_n(FwdIter first, Size count);
 
     /// Constructs objects of type typename iterator_traits<ForwardIt>
-    /// ::value_type in the uninitialized storage designated by the range
+    /// in the uninitialized storage designated by the range
     /// [first, first + count) by default-initialization. If an exception
     /// is thrown during the initialization, the function has no effects.
     ///
--- a/libs/core/algorithms/include/hpx/parallel/container_algorithms/reverse.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/container_algorithms/reverse.hpp
@@ -232,7 +232,7 @@ namespace hpx { namespace ranges {
     ///
     /// \returns  The \a reverse_copy algorithm returns a
     ///           \a ranges::reverse_copy_result<
-    ///           typename hpx::traits::range_iterator<Rng>::type, OutIter>>::type.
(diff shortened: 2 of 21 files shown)


libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_default_construct.hpp
```diff
@@ -11,8 +11,8 @@
 #if defined(DOXYGEN)
 namespace hpx {
 
-    /// Constructs objects of type typename iterator_traits<ForwardIt>
-    /// ::value_type in the uninitialized storage designated by the range
```
hkaiser, line 15: The `::value_type` part needs to stay, but the whole `typename...` should be on a single line to appease doxygen.

libs/full/performance_counters/include/hpx/performance_counters/counters.hpp
```diff
@@ -242,8 +242,8 @@ namespace hpx { namespace performance_counters {
     /// instance. Generally, a full name of a counter instance has the
     /// structure:
     ///
-    ///    /objectname{parentinstancename#parentindex/instancename#instanceindex}
-    ///      /countername#parameters
+    ///    /objectname{parentinstancename#parentindex/instancename}
+    ///      /countername
```
hkaiser, line 246: What was doxygen complaining about here?

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_default_construct.hpp
dimitraka: Ok!

libs/full/performance_counters/include/hpx/performance_counters/counters.hpp
dimitraka: The warnings are:

- `explicit link request to 'instanceindex' could not be resolved`

- `explicit link request to 'parameters' could not be resolved`

libs/core/algorithms/include/hpx/parallel/algorithms/reverse.hpp
```diff
@@ -146,7 +146,7 @@ namespace hpx {
     ///                     the algorithm will be applied to.
     /// \param last         Refers to the end of the sequence of elements the
     ///                     algorithm will be applied to.
-    /// \param dest         Refers to the begin of the destination range.
+    /// \param dest_first   Refers to the begin of the destination range.
```
hkaiser, line 149: I think the parameter name used here needs to correspond to the name used in the function signature below. 
```suggestion
    /// \param first        Refers to the begin of the destination range.
```

libs/core/algorithms/include/hpx/parallel/container_algorithms/for_each.hpp
```diff
@@ -281,8 +281,8 @@ namespace hpx { namespace ranges {
     ///
     /// \param first        Refers to the beginning of the sequence of elements
     ///                     the algorithm will be applied to.
-    /// \param count        Refers to the number of elements starting at
-    ///                     \a first the algorithm will be applied to.
+    /// \param last         Refers to the end of the sequence of elements the
```
hkaiser, line 284: The `\param Size` a couple of line above should be `\tparam Sent` instead

libs/core/algorithms/include/hpx/parallel/container_algorithms/reverse.hpp
```diff
@@ -232,7 +232,7 @@ namespace hpx { namespace ranges {
     ///
     /// \returns  The \a reverse_copy algorithm returns a
     ///           \a ranges::reverse_copy_result<
-    ///           typename hpx::traits::range_iterator<Rng>::type, OutIter>>::type.
+    ///           typename hpx::traits::range_iterator<Rng>::type, OutIter>.
```
hkaiser, line 235: The `>::type` should not go away, I rather think that the two line above should be combined into one as doxygen is confused by the line split, not the `::type`

libs/core/algorithms/include/hpx/parallel/container_algorithms/uninitialized_copy.hpp
```diff
@@ -136,9 +136,8 @@ namespace hpx { namespace ranges {
     /// the calling thread.
     ///
     /// \returns  The \a uninitialized_copy algorithm returns an
-    ///           \a in_out_result<typename hpx::traits::range_traits<Rng1>
-    ///           ::iterator_type, typename hpx::traits::range_traits<Rng2>
-    ///           ::iterator_type>.
+    ///           \a in_out_result<typename hpx::traits::range_traits<Rng1>, 
+    ///           typename hpx::traits::range_traits<Rng2>>.
```
hkaiser, line 140: Same here, the `>::iterator_type` needs to stay. We might want to find a different way of shortening the type. In this case we could write:
```
in_out_result<range_iterator_t<Rng1>, range_iterator_t<Rng2>>
```
instead.

libs/core/algorithms/include/hpx/parallel/container_algorithms/uninitialized_default_construct.hpp
```diff
@@ -15,7 +15,7 @@ namespace hpx { namespace ranges {
     // clang-format off
 
     /// Constructs objects of type typename iterator_traits<ForwardIt>
-    /// ::value_type in the uninitialized storage designated by the range
+    /// in the uninitialized storage designated by the range
```
hkaiser, line 18: Let's not remove the `>::value_type` part, rather have the whole `typename iterator_traits<ForwardIt>::value_type` on its own line.

PR #5932, opened 2022-06-23
Remove mimalloc version constraint

msimberg:
At least on linux there is no need to constrain the version since mimalloc is only being linked, no mimalloc APIs are used by HPX. I don't know if this works on Windows, but I assume it does.


hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

msimberg:
For reference, these two environment variables _may_ significantly improve performance when using mimalloc:
- `MIMALLOC_EAGER_COMMIT_DELAY=0`
- `MIMALLOC_LARGE_OS_PAGES=1`

PR #5933, opened 2022-06-23
Fix documentation for ranges algorithms

dimitraka:
Fix parallel container algorithms documentation


libs/core/algorithms/include/hpx/parallel/container_algorithms/adjacent_difference.hpp
```diff
@@ -68,187 +76,258 @@ namespace hpx { namespace ranges {
...
 
-    /// Searches the range rng for two consecutive identical elements.
+    /// Searches the \a rng for two consecutive identical elements.
     ///
-    /// \note   Complexity: Exactly the smaller of (result - std::begin(rng)) + 1
-    ///                     and (std::begin(rng) - std::end(rng)) - 1 applications
-    ///                     of the predicate where \a result is the value returned
+    /// \note   Complexity: Exactly the smaller of (result - first) + 1 and
+    ///                     (last - first) - 1 application of the predicate
+    ///                     where \a result is the value returned
```
hkaiser, line 191: This change is not necessary for this overload of the algorithm.

PR #5936, opened 2022-06-27
Adding mapping functions between execution policies

### LINKED ISSUE #5925: Simplify implementation of execution policies
The following could be done to simplify the implementation of our execution policies:

- We could turn the non-shim policies into `using` directives referring to the corresponding shim policies. This would allow to reduce the amount of code duplication significantly, i.e.
```
    using sequenced_policy = sequenced_shim_policy<sequenced_executor, sequential_executor_parameters>;
```
- The association between execution policies of similar categories is currently done using the `operator()([non]task_policy_tag)`. This is repetitive, error-prone, and non-extensible. We have a need to expose more of those 'conversion' operations. For instance to retrieve the corresponding non-vectorizing policy from a vectorizing one, or v.v. We should add a more flexible and extensible way of creating mappings between various execution policy categories based on more category tags like `task_policy_tag`. I'm envisioning something along the lines of:
```
    template <typename Tag, 
        typename Enable = std::enable_if_t<is_policy_mapping_v<Tag>>>
    auto operator()(Tag tag)
    {
        return tag(*this);    // this needs to be specialized (tag_invoke)
    }
```

hkaiser:
- simplify implementation of execution policies (these are now derived from a CRTP base class)
- simplify `handle_local_exceptions` implementation

Fixes #5925


hkaiser:
bors merge

PR #5937, opened 2022-06-28
Remove leftover files from PMR based small_vector

hkaiser:
- this also gets rid of depending on Boost `small_vector`


hkaiser:
bors merge

PR #5940, opened 2022-07-03
Protect collective operations against std::vector<bool> idiosyncrasies

hkaiser:
- std::vector<T> is internally used to store intermediate results for the collective operations. This change makes sure that automatic return type deduction does not get fooled by the `vector<bool>`'s proxy value_type.


hkaiser:
bors merge

PR #5941, opened 2022-07-04
Fix sync_wait 

hcq9102:
Issue:
_sync_wait implementation assumes the senders send only one single value. (Actually, senders can pass several values);
sync_wait looks at the variant of the predecessor and makes sure that it contains one tuple with one element._
which is not conform to the description of sync_wait in P2300.

revamp :
The result_type should be optional of the variant of the tuples.
The sync_wait works when the variant has one tuple( no matter how many elements in the tuple)


hcq9102:
please ignore this pr.

PR #5942, opened 2022-07-05
Fix doc warnings

dimitraka:
Fix doc warnings (follow up from previous PR) and `git checkout` version


libs/core/algorithms/include/hpx/parallel/algorithms/for_loop_induction.hpp
```diff
@@ -232,7 +232,7 @@ namespace hpx::experimental {
 }    // namespace hpx::experimental
 
 namespace hpx::parallel { inline namespace v2 {
-
+    /// \cond NOINTERNAL
```
hkaiser, line 235: These will most likely create conflicts with #5939 (once it's merged). Could you resolve those conflicts, please?

dimitraka: Done!

hkaiser:
bors merge

PR #5943, opened 2022-07-05
Fix sync_wait

hcq9102:
Issue:
_sync_wait implementation assumes the senders send only one single value. (Actually, senders can pass several values);
sync_wait looks at the variant of the predecessor and makes sure that it contains one tuple with one element._
which is not conform to the description of sync_wait in P2300.

revamp :
The result_type should be optional of the variant of the tuples;
The sync_wait works when the variant has one tuple( no matter how many elements in the tuple)

### DIFF
--- a/libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
+++ b/libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
@@ -348,9 +348,9 @@ void test_for_each_sender(ExPolicy&& p, IteratorTag)
     auto rng = hpx::util::make_iterator_range(
         iterator(std::begin(c)), iterator(std::end(c)));
     auto f = [](std::size_t& v) { v = 42; };
-    auto result = ex::just(rng, f) |
-        hpx::ranges::for_each(std::forward<ExPolicy>(p)) | tt::sync_wait();
-    HPX_TEST(*result == iterator(std::end(c)));
+    auto result = hpx::get<0>(*(ex::just(rng, f) |
+        hpx::ranges::for_each(std::forward<ExPolicy>(p)) | tt::sync_wait()));
+    HPX_TEST(result == iterator(std::end(c)));
 
     // verify values
     std::size_t count = 0;
--- a/libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
+++ b/libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
@@ -1,4 +1,4 @@
-//  Copyright (c) 2014-2015 Hartmut Kaiser
+//  Copyright (c) 2014-2022 Hartmut Kaiser
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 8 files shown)


libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
```diff
@@ -348,8 +348,8 @@ void test_for_each_sender(ExPolicy&& p, IteratorTag)
     auto rng = hpx::util::make_iterator_range(
         iterator(std::begin(c)), iterator(std::end(c)));
     auto f = [](std::size_t& v) { v = 42; };
-    auto result = ex::just(rng, f) |
-        hpx::ranges::for_each(std::forward<ExPolicy>(p)) | tt::sync_wait();
+    auto result = hpx::get<0>(*(ex::just(rng, f) |
+        hpx::ranges::for_each(std::forward<ExPolicy>(p)) | tt::sync_wait()));
     HPX_TEST(result == iterator(std::end(c)));
```
hcq9102, line 353: Adding * in HPX_TEST(*result == iterator(std::end(c))) doesn't get rid of the errors. 
I think for the sync_wait(), * is not enough to get the one tuple.
So I tried to pull out the value first as above(with get<0>(*...)).
Please correct me if I did it wrong.

libs/core/async_mpi/tests/unit/algorithm_transform_mpi.cpp
```diff
@@ -9,6 +9,7 @@
 #include <hpx/modules/errors.hpp>
 #include <hpx/modules/execution.hpp>
 #include <hpx/modules/testing.hpp>
+#include <hpx/tuple.hpp>
```
hkaiser, line 12: From inside HPX, please use `#include <hpx/datastructures/tuple.hpp>` instead.

hcq9102: got it.

libs/core/execution/include/hpx/execution/algorithms/sync_wait.hpp
```diff
@@ -110,26 +100,23 @@ namespace hpx::execution::experimental::detail {
...
-                        return;
-                    }
-                    else
-                    {
-                        return HPX_MOVE(hpx::get<value_type>(value));
-                    }
+                    // pull the tuple out of the variant and wrap it into an
+                    // optional
+                    return hpx::make_optional(
+                        hpx::get<0>(hpx::get<result_type>(HPX_MOVE(value))));
```
hcq9102, line 108: The return issue here: 
`/dev/shm/hpx/src/libs/core/execution/include/hpx/execution/algorithms/sync_wait.hpp:108:36: error: no matching function for call to ‘get<0>(std::variant<>)’
                         hpx::get<0>(hpx::get<result_type>(HPX_MOVE(value))));`

--> I think hpx::make_optional() trys to construct an optional object, but we should get_value() here from optional, so I think we should do 
`return hpx::get<0>(hpx::get<hpx::make_optional(result_type)>(HPX_MOVE(value))));` ?

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #5945, opened 2022-07-07
[P2300] Added fundamental coroutine_traits for S/R

SAtacker:

## Proposed Changes

  - Awaiter traits
  - Awaitable traits
  - `get_awaiter`

## Checklist

- [X] Awaiter traits tests
- [x] awaitable traits tests
- [x] get waiter tests


libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
```diff
@@ -0,0 +1,231 @@
...
+        {
+            return await.await_suspend(std::coroutine_handle<Promise>{});
+        }
+        return;
+    }
+
+    namespace detail {
+
+        template <typename T, typename... Ts>
+        bool inline constexpr one_of = (std::is_same_v<T, Ts> || ...);
```
hkaiser, line 25: We already have a similar facility here: https://github.com/STEllAR-GROUP/hpx/blob/84801de6a0003336287f2e4dd3da7138352c419f/libs/core/type_support/include/hpx/type_support/pack.hpp#L122-L140

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
```diff
@@ -0,0 +1,231 @@
...
+        constexpr bool is_instance_of_ = false;
+
+        template <class... _As, template <class...> class _T>
+        constexpr bool is_instance_of_<_T<_As...>, _T> = true;
+
+        template <class _Ty, template <class...> class _T>
+        constexpr bool is_instance_of = is_instance_of_<_Ty, _T>;
+
+        template <typename T>
+        inline constexpr bool is_await_suspend_result_t =
```
hkaiser, line 37: We usually use the `_v` postfix for variables, not `_t`.

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+    // Returns the result of applying operator co_await() to the function’s
+    // argument, if the operator is defined, otherwise returns a reference
+    // to the input argument.
+    template <typename Awaitable>
+    decltype(auto) get_awaiter(Awaitable&& await, void*)
+    {
+        if constexpr (detail::has_member_operator_co_await_v<Awaitable>)
+        {
+            return ((Awaitable &&) await).operator co_await();
```
hkaiser, line 152: Could you use `HPX_FORWARD(Awaitable, await).operator co_await()` instead, please? I know, it's the same ... just for consistency's sake again.

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
SAtacker: Okay, noted

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
```diff
@@ -0,0 +1,231 @@
...
+        }
+        else
+        {
+            return (Awaitable &&) await;
+        }
+    }
+
+    template <typename Awaitable, typename Promise,
+        typename = std::enable_if_t<
+            detail::has_await_transform<Promise, Awaitable>, void>>
```
hkaiser, line 166: the `void` is not needed (it's the default for `enable_if`)

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+        template <class _Ty, template <class...> class _T>
+        constexpr bool is_instance_of = is_instance_of_<_Ty, _T>;
+
+        template <typename T>
+        inline constexpr bool is_await_suspend_result_t =
+            one_of<T, void, bool> || is_instance_of<T, std::coroutine_handle>;
+
+        template <typename, typename = void>
+        constexpr bool has_await_ready{};
```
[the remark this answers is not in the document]
SAtacker: Yes, I think these should be `inline constexpr`. I shall make the change

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
```diff
@@ -0,0 +1,231 @@
...
+      : detail::is_awaiter_impl<detail::has_await_ready<Awaiter> &&
+                detail::has_await_resume<Awaiter> &&
+                (detail::has_await_suspend<Awaiter> ||
+                    detail::has_await_suspend_coro_handle<Awaiter, Promise>),
+            Awaiter, Promise>
+    {
+    };
+
+    template <typename Awaiter, typename Promise = void>
+    bool inline constexpr is_awaiter_v = is_awaiter<Awaiter, Promise>::value;
```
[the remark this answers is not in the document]
SAtacker: Right, oops. Need to be careful with these. Especially the template/class, sorry for making you remind me.

hkaiser (commented):
Very nice! Thanks!

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
SAtacker: Yes, definitely, I have seen `HPX_FORWARD` but was not sure if it did the same. Great!

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
SAtacker: Noted

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
SAtacker: I think it's different
one_of is true if any one of the `Ts` matches with the given `T`. `any_of` only checks whether `is_false<Ts>` right?
Context HPX
```
util::any_of<is_one_way_executor<T>,
            is_never_blocking_one_way_executor<T>, is_bulk_one_way_executor<T>,
            is_two_way_executor<T>, is_bulk_two_way_executor<T>>
```
Where as I want to check if `is_same<int,bool,void>` (false) `is_same<bool, bool,void>` (true) `is_same<void, bool, void>` (true) ...

hkaiser: FWIW, there also is `is_any_executor<T>` that resolves to true if any of the above conditions are true.

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
```diff
@@ -0,0 +1,223 @@
+//  Copyright (c) 2022 Shreyas Atre
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
```
hkaiser, line 6: I think all of this file should be wrapped into a 
```
#include <hpx/config.hpp>

#if defined(HPX_HAVE_CXX20_COROUTINES)
...
#endif
```

libs/core/execution_base/tests/unit/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ set(tests
     execution_context
     get_env
     execute_may_block_caller
+    coroutine_traits
```
hkaiser, line 18: This should be added to the tests only if `HPX_WITH_CXX20_COROUTINES` is On. E.g. somethig like:
```suggestion
)
if(HPX_WITH_CXX20_COROUTINES)
  set(tests ${tests} coroutine_traits)
endif()
```

libs/core/execution_base/tests/unit/coroutine_traits.cpp
```diff
@@ -0,0 +1,288 @@
+//  Copyright (c) 2022 Shreyas Atre
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include "hpx/execution_base/traits/coroutine_traits.hpp"
```
hkaiser, line 7: Please use `#include <...>` for including HPX headers.

hkaiser (changes requested):

libs/core/execution_base/tests/unit/CMakeLists.txt
SAtacker: I have tried `HPX_HAVE_CXX20_COROUTINES` but it doesn't seem to work. All config being unchanged now suddenly it found no target `coroutine_traits_test`

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
SAtacker: I had experimented with it but it didn't work. I have coroutines support yet, `HPX_HAVE_CXX20_COROUTINES` is not defined.

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
SAtacker: I tried to find `is_any_executor` but could not.

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
SAtacker:
```
-- Performing Test HPX_WITH_CXX20_COROUTINES - Failed
-- Performing Test HPX_WITH_CXX20_EXPERIMENTAL_SIMD - Failed
-- Performing Test HPX_WITH_CXX20_LAMBDA_CAPTURE - Success
```

SAtacker: I have loaded the module llvm/12.0.1

SAtacker:
```
[satacker@medusa01 build]$ git diff
diff --git a/cmake/tests/cxx20_coroutines.cpp b/cmake/tests/cxx20_coroutines.cpp
index bba9190fac..460dee3609 100644
--- a/cmake/tests/cxx20_coroutines.cpp
+++ b/cmake/tests/cxx20_coroutines.cpp
@@ -5,7 +5,13 @@
 
 // This tests whether C++20 coroutines are supported
+#if __has_include(<coroutine>)
 #include <coroutine>
+namespace coro = std;
+#else
+#include <experimental/coroutine>
+namespace coro = std::experimental;
+#endif
 
 struct resumable
 {
@@ -19,12 +25,12 @@ struct resumable
             return {};
         }
 
-        std::suspend_never initial_suspend() const noexcept
+        coro::suspend_never initial_suspend() const noexcept
         {
             return {};
         }
 
-        std::suspend_never final_suspend() const noexcept
+        coro::suspend_never final_suspend() const noexcept
         {
             return {};
         }
@@ -41,7 +47,7 @@ struct resumable
     {
         return 42;
     }
-    void await_suspend(std::coroutine_handle<promise_type>) const noexcept {}
+    void await_suspend(coro::coroutine_handle<promise_type>) const noexcept {}
 };
 
 resumable test()
```
Made it work

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
hkaiser: Sorry, it's `hpx::traits::is_executor_any<T>`.

libs/core/futures/include/hpx/futures/traits/detail/future_await_traits.hpp
```diff
@@ -194,7 +201,7 @@ namespace hpx { namespace lcos { namespace detail {
 }}}    // namespace hpx::lcos::detail
 
 ///////////////////////////////////////////////////////////////////////////////
-namespace std {
+namespace coro {
```
SAtacker, line 204: This has issues if the `coro` is `std::experimental`, any suggestions?

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
hkaiser: You might want to make this safer by saying:
```
#if defined(__has_include)
#if __has_include(<coroutine>)
...
#endif
#endif
```
as some older compilers don't support `__has_include()`

libs/core/execution_base/tests/unit/CMakeLists.txt
hkaiser: Note, we generally use `HPX_WITH_XXX` during configuration time and a corresponding `HPX_HAVE_XXX` as preprocessor constants to be used during compilation.

libs/core/futures/include/hpx/futures/traits/detail/future_await_traits.hpp
hkaiser: What issues?

libs/core/execution_base/tests/unit/CMakeLists.txt
SAtacker: https://github.com/STEllAR-GROUP/hpx/pull/5945/commits/c8090feefb12261d618a345e9d7302f03acf7584 fixed it for llvm/12.0.1 module

libs/core/futures/include/hpx/futures/traits/detail/future_await_traits.hpp
SAtacker: https://pastebin.com/2hmJjAEU @hkaiser

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
```diff
@@ -0,0 +1,227 @@
...
+#if defined(HPX_HAVE_CXX20_COROUTINES)
+
+#if __has_include(<coroutine>)
+#include <coroutine>
+namespace coro = std;
+#else
+#include <experimental/coroutine>
+namespace coro = std::experimental;
+#endif
```
hkaiser, line 18: I'd suggest aliasing the namespaces to `namespace hpx::coro` to avoid name clashes for users. E.g. something like:
```suggestion
#if __has_include(<coroutine>)
#include <coroutine>
namespace hpx {
    namespace coro = std;
}
#else
#include <experimental/coroutine>
namespace hpx {
    namespace coro = std::experimental;
}
#endif
```
or even better import the necessary names into `hpx::coro` instead of aliasing the whole namespace `std`. This should also go into a separate header file to avoid duplicating the declarations

SAtacker: Thanks :+1:

libs/core/futures/include/hpx/futures/traits/detail/future_await_traits.hpp
hkaiser: We currently specialize `coroutine_traits` in `namespace std` (here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/futures/include/hpx/futures/traits/detail/future_await_traits.hpp#L197). If you add support for coroutines that are in `std::experimental`, this specialization will need to go into that namespace as well.

hkaiser: What I meant is that you need to explicitly say what 'original' namespace the coroutine_traits base template is defined:
```
namespace HPX_CORO_NAMESPACE {
...
}
```
where `HPX_CORO_NAMESPACE` is either `std` or `std::experimental`.

hkaiser:
@SAtacker I think you need to wrap the new configuration header in an #if defined(...) to circumvent the build errors. There is also some inspect errors that needs fixing.

SAtacker:
> @SAtacker I think you need to wrap the new configuration header in an #if defined(...) to circumvent the build errors. There is also some inspect errors that needs fixing.

https://github.com/STEllAR-GROUP/hpx/pull/5945/commits/8e63d9348ad7efa8732df211b8e8fd9ac0278fab#diff-c87c94ab3db419c43094f1852b046cd3200ae0bfb7a66a482f51be42bf76803bR36-R38
and https://github.com/STEllAR-GROUP/hpx/pull/5945/commits/8e63d9348ad7efa8732df211b8e8fd9ac0278fab#diff-73bb5d75ca22afdf27a16580a7a5a51d476e49c6b4357c8e591fb1485397b094R38-R40 should have taken care of it, I had thought so

SAtacker:
https://github.com/STEllAR-GROUP/hpx/runs/7414352657?check_suite_focus=true#step:6:230 the coroutine test itself is failing

hkaiser:
> https://github.com/STEllAR-GROUP/hpx/runs/7414352657?check_suite_focus=true#step:6:230 the coroutine test itself is failing

What I meant is that you either have to avoid #including the new coroutine_traits.hpp if coroutines support is not detected or wrap the content of the header in a corresponding #ifdef/#endif.

hkaiser:
> > https://github.com/STEllAR-GROUP/hpx/runs/7414352657?check_suite_focus=true#step:6:230 the coroutine test itself is failing
> 
> What I meant is that you either have to avoid #including the new coroutine_traits.hpp if coroutines support is not detected or wrap the content of the header in a corresponding #ifdef/#endif.

The second option seems to be better, however.

SAtacker:
> > > https://github.com/STEllAR-GROUP/hpx/runs/7414352657?check_suite_focus=true#step:6:230 the coroutine test itself is failing
> > 
> > 
> > What I meant is that you either have to avoid #including the new coroutine_traits.hpp if coroutines support is not detected or wrap the content of the header in a corresponding #ifdef/#endif.
> 
> The second option seems to be better, however.

Did you mean https://github.com/STEllAR-GROUP/hpx/pull/5945/files#diff-73bb5d75ca22afdf27a16580a7a5a51d476e49c6b4357c8e591fb1485397b094R38-R40 ?

hkaiser:
> > > > https://github.com/STEllAR-GROUP/hpx/runs/7414352657?check_suite_focus=true#step:6:230 the coroutine test itself is failing
> > > 
> > > 
> > > What I meant is that you either have to avoid #including the new coroutine_traits.hpp if coroutines support is not detected or wrap the content of the header in a corresponding #ifdef/#endif.
> > 
> > 
> > The second option seems to be better, however.
> 
> Did you mean https://github.com/STEllAR-GROUP/hpx/pull/5945/files#diff-73bb5d75ca22afdf27a16580a7a5a51d476e49c6b4357c8e591fb1485397b094R38-R40 ?

I meant adding a wrapper here: [libs/core/config/include/hpx/config/coroutines_support.hpp](https://github.com/STEllAR-GROUP/hpx/pull/5945/files#diff-573797b1dcfd340cd0b4669fbea646a9f42626a0da2615198f7e3cc4337466fc)

SAtacker:
> > > > > https://github.com/STEllAR-GROUP/hpx/runs/7414352657?check_suite_focus=true#step:6:230 the coroutine test itself is failing
> > > > 
> > > > 
> > > > What I meant is that you either have to avoid #including the new coroutine_traits.hpp if coroutines support is not detected or wrap the content of the header in a corresponding #ifdef/#endif.
> > > 
> > > 
> > > The second option seems to be better, however.
> > 
> > 
> > Did you mean https://github.com/STEllAR-GROUP/hpx/pull/5945/files#diff-73bb5d75ca22afdf27a16580a7a5a51d476e49c6b4357c8e591fb1485397b094R38-R40 ?
> 
> I meant adding a wrapper here: [libs/core/config/include/hpx/config/coroutines_support.hpp](https://github.com/STEllAR-GROUP/hpx/pull/5945/files#diff-573797b1dcfd340cd0b4669fbea646a9f42626a0da2615198f7e3cc4337466fc)

Then `hpx::coro` is not found by it.
```diff
diff --git a/libs/core/config/include/hpx/config/coroutines_support.hpp b/libs/core/config/include/hpx/config/coroutines_support.hpp
index fb7be8d728..08b1adb20a 100644
--- a/libs/core/config/include/hpx/config/coroutines_support.hpp
+++ b/libs/core/config/include/hpx/config/coroutines_support.hpp
@@ -6,6 +6,8 @@
 
 #pragma once
 
+#if defined(HPX_HAVE_CXX20_COROUTINES)
+
 #if defined(__has_include)
 #if __has_include(<coroutine>)
 #include <coroutine>
@@ -25,3 +27,5 @@ namespace hpx { namespace coro {
 #define HPX_COROUTINE_NAMESPACE_STD std::experimental
 #endif
 #endif
+
+#endif // HPX_HAVE_CXX20_COROUTINES
\ No newline at end of file
```

SAtacker:
I know it is supposed to work, but building systems always make it harder to believe.

hkaiser:
> > > > > > https://github.com/STEllAR-GROUP/hpx/runs/7414352657?check_suite_focus=true#step:6:230 the coroutine test itself is failing
> > > > > 
> > > > > 
> > > > > What I meant is that you either have to avoid #including the new coroutine_traits.hpp if coroutines support is not detected or wrap the content of the header in a corresponding #ifdef/#endif.
> > > > 
> > > > 
> > > > The second option seems to be better, however.
> > > 
> > > 
> > > Did you mean https://github.com/STEllAR-GROUP/hpx/pull/5945/files#diff-73bb5d75ca22afdf27a16580a7a5a51d476e49c6b4357c8e591fb1485397b094R38-R40 ?
> > 
> > 
> > I meant adding a wrapper here: [libs/core/config/include/hpx/config/coroutines_support.hpp](https://github.com/STEllAR-GROUP/hpx/pull/5945/files#diff-573797b1dcfd340cd0b4669fbea646a9f42626a0da2615198f7e3cc4337466fc)
> 
> Then `hpx::coro` is not found by it.
> 
> ```diff
> diff --git a/libs/core/config/include/hpx/config/coroutines_support.hpp b/libs/core/config/include/hpx/config/coroutines_support.hpp
> index fb7be8d728..08b1adb20a 100644
> --- a/libs/core/config/include/hpx/config/coroutines_support.hpp
> +++ b/libs/core/config/include/hpx/config/coroutines_support.hpp
> @@ -6,6 +6,8 @@
>  
>  #pragma once
>  
> +#if defined(HPX_HAVE_CXX20_COROUTINES)
> +
>  #if defined(__has_include)
>  #if __has_include(<coroutine>)
>  #include <coroutine>
> @@ -25,3 +27,5 @@ namespace hpx { namespace coro {
>  #define HPX_COROUTINE_NAMESPACE_STD std::experimental
>  #endif
>  #endif
> +
> +#endif // HPX_HAVE_CXX20_COROUTINES
> \ No newline at end of file
> ```

You need to add a `#include <hpx/config/defines.hpp>` before using the `HPX_HAVE_CXX20_COROUTINES` constant (it's defined in that header).

libs/core/config/include/hpx/config.hpp
```diff
@@ -35,6 +35,10 @@
 #include <hpx/config/version.hpp>
 #include <hpx/config/weak_symbol.hpp>
 
+#if defined(HPX_HAVE_CXX20_COROUTINES)
```
hkaiser, line 38: This #ifdef is not needed anymore.

libs/core/config/include/hpx/config/coroutines_support.hpp
```diff
@@ -0,0 +1,33 @@
+//  Copyright (c) 2022 Shreyas Atre
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/config/defines.hpp>
+
+#pragma once
```
hkaiser, line 9: Nitpick: we usually place the #pragma once first in the file.

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
```diff
@@ -0,0 +1,220 @@
+//  Copyright (c) 2022 Shreyas Atre
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/config/coroutines_support.hpp>
+#include <hpx/config/forward.hpp>
```
hkaiser, line 8: This change would simply things:
```suggestion
#include <hpx/config.hpp>
```

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
```diff
@@ -0,0 +1,220 @@
...
+#include <hpx/concepts/has_member_xxx.hpp>
+#include <type_traits>
+
+#if defined(HPX_HAVE_CXX20_COROUTINES)
```
hkaiser, line 12:
```suggestion
#if defined(HPX_HAVE_CXX20_COROUTINES)

#include <hpx/concepts/has_member_xxx.hpp>
#include <type_traits>
```

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
```diff
@@ -0,0 +1,220 @@
...
+namespace hpx { namespace execution { namespace experimental {
+
+    template <typename Promise, typename Awaiter>
+    decltype(auto) await_suspend(Awaiter&& await)
+    {
+        if constexpr (!std::is_same_v<Promise, void>)
+        {
+            return await.await_suspend(hpx::coro::coroutine_handle<Promise>{});
+        }
+        return;
```
hkaiser, line 23: Should this be
```suggestion
    else
    {
        return;
    }
```
? IOW, what's the return type if `std::is_same_v<Promise, void>`?

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+    // Returns the result of applying operator co_await() to the function’s
+    // argument, if the operator is defined, otherwise returns a reference
+    // to the input argument.
+    template <typename Awaitable>
+    decltype(auto) get_awaiter(Awaitable&& await, void*)
+    {
+        if constexpr (detail::has_member_operator_co_await_v<Awaitable>)
+        {
+            return ((HPX_FORWARD(Awaitable, await))).operator co_await();
```
hkaiser, line 150:
```suggestion
            return HPX_FORWARD(Awaitable, await).operator co_await();
```

libs/core/config/include/hpx/config/coroutines_support.hpp
```diff
@@ -0,0 +1,33 @@
...
+namespace hpx { namespace coro {
+    using std::experimental::coroutine_handle;
+    using std::experimental::suspend_always;
+    using std::experimental::suspend_never;
+}}    // namespace hpx::coro
+#define HPX_COROUTINE_NAMESPACE_STD std::experimental
+#endif
+#endif
+
+#endif    // HPX_HAVE_CXX20_COROUTINES
```
hkaiser, line 33: This misses an EOL character.

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+    template <typename Awaitable, typename Promise = void,
+        typename = std::enable_if_t<is_awaitable_v<Awaitable, Promise>>>
+    using await_result_t = decltype(
+        (get_awaiter(std::declval<Awaitable>(), static_cast<Promise*>(nullptr))
+                .await_resume()));
+
+}}}    // namespace hpx::execution::experimental
+
+#endif    // HPX_HAVE_CXX20_COROUTINES
```
hkaiser, line 220: This misses an EOL character.

libs/core/futures/include/hpx/futures/traits/detail/future_await_traits.hpp
```diff
@@ -26,8 +25,8 @@
 namespace hpx { namespace lcos { namespace detail {
 
     template <typename Promise = void>
-    using coroutine_handle = std::coroutine_handle<Promise>;
-    using suspend_never = std::suspend_never;
+    using coroutine_handle = coro::coroutine_handle<Promise>;
+    using suspend_never = coro::suspend_never;
```
hkaiser, line 29: Does this file miss a `#include <hpx/config/coroutines_support.hpp>`?

hkaiser (changes requested):

libs/core/futures/include/hpx/futures/traits/detail/future_await_traits.hpp
SAtacker: It has https://github.com/STEllAR-GROUP/hpx/blob/ce7b0096f7a9354dfb1f6219ec5d2ecd9b8b76da/libs/core/futures/include/hpx/futures/traits/detail/future_await_traits.hpp#L9

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
SAtacker: It should be void

hkaiser: > It should be void

In this case I'm surprised it compiles as is. Or does `await.await_suspend(hpx::coro::coroutine_handle<Promise>{});` always return `void` as well?

libs/core/futures/include/hpx/futures/traits/detail/future_await_traits.hpp
hkaiser: Yes, I know. We always try to make every file self-contained in terms of the `#include`s it requires. Otherwise removing `#include <hpx/config.hpp>` from the other file (for whatever reason) may break the compilation of this file.

libs/core/execution_base/include/hpx/execution_base/traits/coroutine_traits.hpp
SAtacker: Thanks, this should have some tests. Also, await_suspend should always accept a coroutine handle --> [p2880](https://www.open-std.org/JTC1/SC22/wg21/docs/papers/2018/p1288r0.pdf) page 6

hkaiser: Since `await_suspend` should either return `void` or `bool`, this code needs more attention. In the case that `await_suspend` returns a `bool` this will not compile as is (the different return paths would return different types). Thus the patch I suggested will be necessary.

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #5948, opened 2022-07-09
Adding support for pure sender/receiver based executors to parallel algorithms

hkaiser:
- thread_pool_scheduler now takes an execution policy type
- bulk with a sync scheduler falls back to default bulk implementation
- flyby: add workaround for MSVC failing to compile scheduler_executor test
- flyby: modernize executor customization point implementations
- flyby: add unary and nullary wait_all(future<T>)
- flyby: fixing size_t -> uint32_t conversion warnings
- flyby: making policy mapping CPOs final and tag_invoke friend members private
- flyby: adding missing #include to datapar.hpp
- flyby: fix value_type calculation for let_value
- adding support for s/r for partitioner (and for_loop)
- adding direct s/r support to for_each_scaling benchmark
- adding tests
- added default for get_annotation property (returns nullptr)

### DIFF
--- a/cmake/HPX_AddConfigTest.cmake
+++ b/cmake/HPX_AddConfigTest.cmake
@@ -440,6 +440,15 @@ function(hpx_check_for_cxx17_copy_elision)
 endfunction()
 
 # ##############################################################################
+function(hpx_check_for_cxx17_optional_copy_elision)
+  add_hpx_config_test(
+    HPX_WITH_CXX17_OPTIONAL_COPY_ELISION
+    SOURCE cmake/tests/cxx17_optional_copy_elision.cpp
+    FILE ${ARGN}
+  )
+endfunction()
+
+# ##############################################################################
 function(hpx_check_for_cxx20_coroutines)
   if(NOT HPX_WITH_CXX20_COROUTINES)
     unset(HPX_CXX20_COROUTINES_FLAGS)
--- a/cmake/HPX_PerformCxxFeatureTests.cmake
+++ b/cmake/HPX_PerformCxxFeatureTests.cmake
@@ -77,6 +77,10 @@ function(hpx_perform_cxx_feature_tests)
 
   hpx_check_for_cxx17_copy_elision(DEFINITIONS HPX_HAVE_CXX17_COPY_ELISION)
 
+  hpx_check_for_cxx17_optional_copy_elision(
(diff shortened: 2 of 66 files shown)


hkaiser:
@bhumitattarde could you please have a look why the docs-html build step has failed for this PR?

bhumitattarde:
> @bhumitattarde could you please have a look why the docs-html build step has failed for this PR?

Doxygen/Sphinx seem to be thinking

```cpp
template <typename BaseExecutor>
    explicit annotating_executor(BaseExecutor&& sched, std::string annotation)
        -> annotating_executor<std::decay_t<BaseExecutor>>;

    template <typename BaseExecutor>
    explicit annotating_executor(
        BaseExecutor&& sched, char const* annotation = nullptr)
        -> annotating_executor<std::decay_t<BaseExecutor>>;
```

are the same declarations. Telling Doxygen to not consider one of the declarations makes the build job go through, though at the expense of incomplete documentation. Something like this can be done as a temporary fix:

```cpp
template <typename BaseExecutor>
    explicit annotating_executor(BaseExecutor&& sched, std::string annotation)
        -> annotating_executor<std::decay_t<BaseExecutor>>;

    /// \cond BREAKS_DOXYGEN
    template <typename BaseExecutor>
    explicit annotating_executor(
        BaseExecutor&& sched, char const* annotation = nullptr)
        -> annotating_executor<std::decay_t<BaseExecutor>>;
    /// \endcond BREAKS_DOXYGEN
```

I'm not sure why this is happening, but some guesses  -- 

1. Doxygen version might not be compatible with the Sphinx/Breathe version. We can try playing around with the versions a bit.
2. Doxygen has an open issue where trailing return type declarations are not parsed correctly. I tried converting trailing return statements to traditional return statements but interestingly, that didn't help.
3. Bug in Doxygen/Breathe/Sphinx (that might have been fixed in the latest versions).

hkaiser:
> 1. Doxygen version might not be compatible with the Sphinx/Breathe version. We can try playing around with the versions a bit.
> 2. Doxygen has an open issue where trailing return type declarations are not parsed correctly. I tried converting trailing return statements to traditional return statements but interestingly, that didn't help.
> 3. Bug in Doxygen/Breathe/Sphinx (that might have been fixed in the latest versions).

Thanks for the analysis. I think that there is no harm done if one of the CTAD deduction rules would not be documented. Documenting the constructors will be fine. I'll add that `\cond` as you suggested.

hkaiser:
bors merge

PR #5956, opened 2022-07-17
Adding sync_wait_with_variant

hcq9102:
P2300R5: [sync_wait_with_variant](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2300r5.html#design-typed)
sync_wait_with_variant, which accepts any sender, but returns an optional whose value type is the variant of all the possible tuples sent by the input sender:
## Checklist

### DIFF
--- a/libs/core/execution/CMakeLists.txt
+++ b/libs/core/execution/CMakeLists.txt
@@ -26,6 +26,7 @@ set(execution_headers
     hpx/execution/algorithms/split.hpp
     hpx/execution/algorithms/start_detached.hpp
     hpx/execution/algorithms/sync_wait.hpp
+    hpx/execution/algorithms/sync_wait_with_variant.hpp
     hpx/execution/algorithms/then.hpp
     hpx/execution/algorithms/transfer.hpp
     hpx/execution/algorithms/transfer_just.hpp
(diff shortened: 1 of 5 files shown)


hkaiser (changes requested):
Very nice! Please add the new file here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/execution/CMakeLists.txt#L26, and add the new test to the build system (here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/execution/tests/unit/CMakeLists.txt#L19).

Also, now that you have a working version that duplicates 95% of the `sync_wait` code, would you see a way to combine the implementations of `sync_wait` and `sync_wait_with_variant` to reduce the amount of code repetition?

hcq9102:
> Also, now that you have a working version that duplicates 95% of the `sync_wait` code, would you see a way to combine the implementations of `sync_wait` and `sync_wait_with_variant` to reduce the amount of code repetition?

Sure, will do.

hkaiser:
@hcq9102 things should be fine now. Please note that I also have rebased this branch onto master.

hkaiser:
This can be closed as it is subsumed by #6015

hcq9102:
#6015 got merged. We don't need this one anymore.(sync_wait_with_variant integrating with sync_wait)

PR #5961, opened 2022-07-21
Prevent stalling during shutdown when running hello_world_distributed

hkaiser:
- flyby: fix dll-export warnigs


hkaiser:
bors merge

PR #5970, opened 2022-07-29
Fixing component multiple inheritance

### LINKED ISSUE #5964: component with multiple inheritance
## Expected Behavior

Previously I have successfully used components with multiple inheritance but this appears to cause problems in HPX 1.8.0.

## Actual Behavior

Components which inherit from another class cause segfaults (the backtraces seems to generally point back to the class destructor). 

See this test case which works in HPX 1.7.0 but fails in HPX 1.8.0. This test case passes if I inherit from `component_base` first but this makes me feel like I am relying on some fragile behavior which may not really be working properly. Is this meant to be supported?

```
#include <hpx/config.hpp>
#include <hpx/hpx_init.hpp>
#include <hpx/include/actions.hpp>
#include <hpx/include/async.hpp>
#include <hpx/include/components.hpp>
#include <hpx/include/runtime.hpp>
#include <hpx/include/util.hpp>

#include <atomic>
#include <cstddef>
#include <vector>

struct foo_t {
  int a;
  int b;
  foo_t() = default;
  foo_t(int a, int b) : a(a), b(b) {}

  friend class hpx::serialization::access;
  template<typename Archive> inline void serialize(Archive &ar, const unsigned int) {
    ar & a & b;
  }
};

struct component_server : foo_t, hpx::components::component_base<component_server> {
  component_server(foo_t foo) : foo_t(std::move(foo)) {}
};

HPX_REGISTER_COMPONENT(
    hpx::components::component<component_server>, component_server_component)

int hpx_main()
{
    foo_t in {1,2};
    hpx::id_type id = hpx::new_<component_server>(hpx::find_here(), in).get();
    const component_server& out = *hpx::get_ptr<component_server>(hpx::launch::sync_policy(), id);
    HPX_TEST_EQ(out.a, in.a);
    HPX_TEST_EQ(out.b, in.b);

    return hpx::finalize();
}

int main(int argc, char** argv)
{
    HPX_TEST_EQ(hpx::init(argc, argv), 0);
    return hpx::util::report_errors();
}
#endif
```

## Specifications

Please describe your environment:

  - HPX Version: 1.8.0
  - Platform (compiler, OS): gcc 12.1.1, opensuse tumbleweed

hkaiser:
- flyby: fixing get_ptr<>(sync)

Fixes #5964

@kleinhenz would you be able to verify whether this fixes your problem?

### DIFF
--- a/libs/full/components/include/hpx/components/get_ptr.hpp
+++ b/libs/full/components/include/hpx/components/get_ptr.hpp
@@ -253,7 +253,8 @@ namespace hpx {
             naming::get_locality_id_from_gid(gid) == agas::get_locality_id(ec))
         {
             return std::shared_ptr<Component>(
-                get_lva<Component>::call(gid.get_lsb()),
+                get_lva<Component>::call(
+                    reinterpret_cast<hpx::naming::address_type>(gid.get_lsb())),
                 detail::get_ptr_no_unpin_deleter(id));
         }
 
--- a/libs/full/components/tests/regressions/CMakeLists.txt
+++ b/libs/full/components/tests/regressions/CMakeLists.txt
@@ -1,5 +1,25 @@
-# Copyright (c) 2020-2021 The STE||AR-Group
+# Copyright (c) 2022 Hartmut Kaiser
 #
 # SPDX-License-Identifier: BSL-1.0
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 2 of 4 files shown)


kleinhenz:
Yes this fixes the problem for me. Thanks! Any chance this can make it into 1.8.1?

hkaiser:
> Yes this fixes the problem for me. Thanks! Any chance this can make it into 1.8.1?

Yes, that's the plan.

hkaiser:
bors merge

PR #5971, opened 2022-07-29
added logic to detect riscv compiler configured for 64 bit target

ct-clmsn:
## Fixes #

Modification to cmake target detection script. Enforces 64-bit target constraint on risc-v builds.

## Proposed Changes

  - cmake/TargetArch.cmake has 1 modification

Current risc-v support does not check to see if the compiler is configured for the 32 bit or 64 bit build target. This PR enforces a check, in cmake, making sure the 64-bit build target is configured in the compiler.

## Checklist

### DIFF
--- a/cmake/TargetArch.cmake
+++ b/cmake/TargetArch.cmake
@@ -39,7 +39,7 @@ set(archdetect_cpp_code
         #else
             #error cmake_ARCH arm
         #endif
-    #elif defined(__riscv)
+    #elif defined(__riscv) && (__riscv_xlen == 64)
         #error cmake_ARCH riscv64
     #elif defined(__i386) || defined(__i386__) || defined(_M_IX86)
         #error cmake_ARCH i386
--- a/libs/core/hardware/include/hpx/hardware/timestamp/linux_riscv_64.hpp
+++ b/libs/core/hardware/include/hpx/hardware/timestamp/linux_riscv_64.hpp
@@ -10,8 +10,6 @@
 
 #include <hpx/config.hpp>
 
-#if defined(__riscv)
-
 #include <cstdint>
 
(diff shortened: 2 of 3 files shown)


hkaiser:
@ct-clmsn do you expect more changes for this?

ct-clmsn:
@hkaiser yeah, is it cool if I continue to push on this branch for the context switching implementation or would ya'll prefer a new branch?

hkaiser:
> @hkaiser is it cool if I continue to push on this branch for the context switching implementation or would ya'll prefer a new branch? as for cmake scripts, this will be the last modification.

As you prefer... there is no problem in doing it either way. 

What difference will a new context switching implementation do compared to the one based on Boost.Context?

ct-clmsn:
@hkaiser, great question! This is probably a misunderstanding on my part; figured you all had a need/preference for a non-boost implementation.

hkaiser:
> @hkaiser, great question! This is probably a misunderstanding on my part; figured you all had a need/preference for both boost and non-boost implementations.

So far, we have relied on the Boost.Context implementation for platforms for which we don't have a 'native' implementation (those were created long before Boost.Context existed, and we have kept them). The only advantage of not using the Boost.Context version is that it removes one external dependency. 

That said, if you feel that you would like to create a 'native' one (independent of Boost) because that might be beneficial in one way or another (or just because you like the challenge), please go ahead.

hkaiser (approved):
LGTM, thanks!

ct-clmsn:
@hkaiser fair enough! In that case, I'm good with closing this branch and PR out with this update.

hkaiser:
This can go in now, thanks again!

hkaiser:
bors merge

PR #5972, opened 2022-08-02
Fix documentation for container algorithms

dimitraka:
Fix:
- the final part of the documentation of the container algorithms
- the warnings caused by the container algorithms when building the documentation

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/container_algorithms/transform.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/container_algorithms/transform.hpp
@@ -1,5 +1,6 @@
 //  Copyright (c) 2015 Hartmut Kaiser
 //  Copyright (c) 2021 Giannis Gonidelis
+//  Copyright (c) 2022 Dimitra Karatza
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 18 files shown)


libs/core/algorithms/include/hpx/parallel/container_algorithms/transform.hpp
```diff
@@ -21,13 +23,17 @@ namespace hpx { namespace ranges {
...
-    ///                     meet the requirements of an input iterator.
-    /// \tparam OutIter     The type of the iterator representing the
-    ///                     destination range (deduced).
+    /// \tparam FwdIter1    The type of the source iterators for the first
+    ///                     range used (deduced).
     ///                     This iterator type must meet the requirements of an
-    ///                     output iterator.
+    ///                     forward iterator.
+    /// \tparam Sent1       The type of the end source iterators used (deduced).
+    ///                     This iterator type must meet the requirements of an
```
gonidelis, line 31: a* sentinel?

libs/core/algorithms/include/hpx/parallel/container_algorithms/transform.hpp
```diff
@@ -21,13 +23,17 @@ namespace hpx { namespace ranges {
...
+    ///                     sentinel for FwdIter1.
+    /// \tparam FwdIter2    The type of the source iterators for the first
+    ///                     range used (deduced).
+    ///                     This iterator type must meet the requirements of an
```
gonidelis, line 35: a* forward...?

libs/core/algorithms/include/hpx/parallel/container_algorithms/transform.hpp
```diff
@@ -21,13 +23,17 @@ namespace hpx { namespace ranges {
...
-    ///                     The iterators extracted from this range type must
-    ///                     meet the requirements of an input iterator.
-    /// \tparam OutIter     The type of the iterator representing the
-    ///                     destination range (deduced).
+    /// \tparam FwdIter1    The type of the source iterators for the first
+    ///                     range used (deduced).
     ///                     This iterator type must meet the requirements of an
-    ///                     output iterator.
+    ///                     forward iterator.
+    /// \tparam Sent1       The type of the end source iterators used (deduced).
```
gonidelis, line 30: Is Sent1 always an iterator? Could be value, no?

dimitraka: Something like " The type of the end source **elements** used (deduced)." sounds better to you?

gonidelis: nai

gonidelis: Thought you copy-pasted from smwr. If it's from the non-ranges algos then we should come up with a different wording. Cause the ranges-algos have a different end type.

dimitraka: It is a copy-paste, so maybe we should discuss about it today because I am not sure what the proper description is.

gonidelis: True.

hkaiser: Sentinels are a special type of 'end-iterator'. I think it is ok to use the term 'end source iterator' as in the above text.

dimitraka: Alright, I'll leave it as it is then

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #5981, opened 2022-08-18
docs: add docs for condition_variable.hpp

bhumitattarde:
@hkaiser Adding docs for this header is causing duplicate declarations issue that we encountered before, If this is happening for `condition_variable.hpp`, it means the same issue will show up later for other headers as we start documenting them,
Additionally, Doxygen is not able to resolve some symbols.
I'll be debugging these, hence the draft PR.


libs/core/synchronization/include/hpx/synchronization/condition_variable.hpp
```diff
@@ -43,43 +105,109 @@ namespace hpx {
...
+         *                      the \a wait functions in a loop or using the
+         *                      overloads of \a wait, \a wait_for, or
+         *                      \a wait_until that take a predicate. end note]
+         *
+         * IOW, \a ~condition_variable() can execute before a signaled thread
+         * returns from a wait. If this happens with \a condition_variable,
+         * that waiting thread will attempt to lock the destructed mutex.
+         * To fix this, there must be shared ownership of the data members
+         * FIXME(bhumit->hkaiser) shouldn't this be condition_variable_any?
+         * between the \a condition_variable_any object and the member
```
bhumitattarde, line 138: @hkaiser shouldn't this be `condition_variable_any`? I have a feeling I'm wrong, however

libs/core/synchronization/include/hpx/synchronization/condition_variable.hpp
```diff
@@ -43,43 +105,109 @@ namespace hpx {
...
+         * functions \a wait (\a wait_for, etc.).
+         */
         ~condition_variable() = default;
 
+        /**
+         * \brief If any threads are waiting on \a *this, calling \a notify_one
+         * unblocks one of the waiting threads. 
+         * 
+         * FIXME(bhumit->hkaiser) is this correct?
+         * \param ec Used to hold error code value originted during the
```
bhumitattarde, line 148: @hkaiser is this correct?

libs/core/synchronization/include/hpx/synchronization/condition_variable.hpp
```diff
@@ -43,43 +105,109 @@ namespace hpx {
...
+         * Atomically unlocks lock, blocks the current executing thread, and
+         * adds it to the list of threads waiting on \a *this. The thread will
+         * be unblocked when \a notify_all() or \a notify_one() is executed.
+         * It may also be unblocked spuriously. When unblocked, regardless of
+         * the reason, lock is reacquired and wait exits.
+         * 
+         * If this function fails to meet the postconditions
+         * (\a lock.owns_lock()==true and \a lock.mutex() is locked by the
+         * FIXME(bhumit->hkaiser) what does hpx call?
+         * calling thread), \a std::terminate is called. For example, this
```
bhumitattarde, line 191: @hkaiser does hpx call `std::terminate` as well?

libs/core/synchronization/include/hpx/synchronization/condition_variable.hpp
```diff
@@ -181,7 +540,22 @@ namespace hpx {
...
+     *
+     * See \a hpx::condition_variable for the description of the semantics of
+     * condition variables.
+     * The class \a hpx::condition_variable_any is a \a StandardLayoutType. It
+     * is not \a CopyConstructible, \a MoveConstructible, \a CopyAssignable,
+     * or \a MoveAssignable.
+     *
+     * FIXME(bhumit->hkaiser) true for std, what about hpx?
+     * If the lock is \a std::unique_lock<std::mutex>,
+     * \a hpx::condition_variable may provide better performance.
```
bhumitattarde, line 557: @hkaiser is this true for hpx as well?

libs/core/synchronization/include/hpx/synchronization/condition_variable.hpp
hkaiser: This should be `condition_variable`, I think.

libs/core/synchronization/include/hpx/synchronization/condition_variable.hpp
hkaiser: I think so, yes (except for the typo in `originated`).

libs/core/synchronization/include/hpx/synchronization/condition_variable.hpp
hkaiser: Good question. I don't think we handle this case ATM. We should revisit the code.

libs/core/synchronization/include/hpx/synchronization/condition_variable.hpp
hkaiser: I don't think so. HPX does not implement any special optimizations for `condition_variable` (which is the only reason why there are two different classes in the first place). Implementation wise in HPX both, `condition_variable` and `condition_variable_any` are almost identical.

hkaiser (changes requested):
A general comment: would you mind using the `///` comments for doxygen comments instead of the `/** */`? This is mostly for consistency, nothing else.

bhumitattarde:
> A general comment: would you mind using the `///` comments for doxygen comments instead of the `/** */`? This is mostly for consistency, nothing else.

Sure, should I change the style for this PR, and should I use `///` hereon?

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #5984, opened 2022-08-21
Reimplement distributed::barrier on top of existing collectives infrastructure

hkaiser:
This is meant to fix possible communication tearing during HPX startup


hkaiser:
This can be closed as #7210 has solved the issue.

PR #5985, opened 2022-08-23
Re-add deprecated tag_policy_tag et.al. types that were removed in V1.8.1

hkaiser:
This fixes oversight that introduces an incompatibility with HPX-Kokkos

### DIFF
--- a/libs/core/executors/include/hpx/executors/execution_policy.hpp
+++ b/libs/core/executors/include/hpx/executors/execution_policy.hpp
@@ -34,7 +34,17 @@ namespace hpx { namespace execution {
     /// Default sequential execution policy object.
     inline constexpr hpx::execution::experimental::to_task_t task{};
 
+    using task_policy_tag HPX_DEPRECATED_V(1, 9,
+        "hpx::execution::task_policy_tag is deprecated, use "
+        "hpx::execution::experimental::to_task_t instead") =
+        hpx::execution::experimental::to_task_t;
+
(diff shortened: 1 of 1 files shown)


msimberg:
Is the required change literally just changing `task_policy_tag` to `to_task_t` or did more than that change? hpx-kokkos was using it in these `operator()`'s: https://github.com/STEllAR-GROUP/hpx-kokkos/blob/20a44967c742f5a7670b4dff9658d9973bf849f2/src/hpx/kokkos/policy.hpp#L176-L178.

hkaiser:
> Is the required change literally just changing `task_policy_tag` to `to_task_t` or did more than that change? hpx-kokkos was using it in these `operator()`'s: https://github.com/STEllAR-GROUP/hpx-kokkos/blob/20a44967c742f5a7670b4dff9658d9973bf849f2/src/hpx/kokkos/policy.hpp#L176-L178.

Short answer: yes, those are equivalent.

Longer answer: this was changed when we introduced to CPOs that allow to 'convert' one type of execution policy to another type (i.e. `to_non_simd(par_simd) === par`. In this context the `to_task()` CPO was introduced as well. for backwards compatibility, we kept `policy::operator()(to_task_t)` (which is equivalent to the previous `task_policy_tag`). What this change missed was to keep the old names in place to avoid immediate breakage of existing code.

hkaiser:
bors merge

PR #5994, opened 2022-08-29
Fix CTRL+C on windows

Neumann-A:
According to https://docs.microsoft.com/en-us/windows/console/setconsolectrlhandler
returning TRUE will stop further execution of installed handlers and stop the normal handler from running.

- [x] I have fixed a bug

### DIFF
--- a/libs/core/runtime_local/src/runtime_local.cpp
+++ b/libs/core/runtime_local/src/runtime_local.cpp
@@ -103,23 +103,23 @@ namespace hpx {
         {
         case CTRL_C_EVENT:
             handle_termination("Ctrl-C");
-            return TRUE;
+            break;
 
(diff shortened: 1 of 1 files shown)


hkaiser (approved):
Apparently, we had misunderstood the docs, then. 
LGTM, thanks!

hkaiser:
@Neumann-A could you please rebase this onto master and then remove the std::terminate I added here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/runtime_local/src/runtime_local.cpp#L97? It's not needed anymore as your solution is better.

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6002, opened 2022-09-04
Making sure that default constructble arguments are properly constructed during deserialization

### LINKED ISSUE #5998: Serializing multiple arguments when applying distributed action results in segfault
## Expected Behavior

Call a remote action regardless of the number of parameters as long as they are all serializable. 

## Actual Behavior

A very specific combination of parameters seems to systematically result in a segfault in a distributed environment.

I have a [reproducible example](https://github.com/esseivaju/HPXDistributed/blob/action_2args_error/src/demo.cpp). Configuring it with ```cmake . -DSINGLE_ARG=OFF -DDEFAULT_CONSTRUCTIBLE=OFF -DVECTOR_2ND_ARG=ON``` will result in a segfault when executed (tested on Cori haswell  and Perlmutter CPU). Any other options combination will work.  

## Specifications

  - HPX Version: 1.8.1 (compiled with ```-DHPX_WITH_PARCELPORT_TCP=OFF -DHPX_WITH_PARCELPORT_MPI=ON```)
  - Platform (compiler, OS): 
    - Cori, gcc 11.2.0, cray-mpich 7.7.19 
    - Perlmutter, gcc 11.2.0, cray-mpich 8.1.17

hkaiser:
- if one or more argument to an action was non-default constructible, then the other (default constructible) arguments were not always properly constructed

Fixes #5998

### DIFF
--- a/libs/core/datastructures/include/hpx/datastructures/serialization/tuple.hpp
+++ b/libs/core/datastructures/include/hpx/datastructures/serialization/tuple.hpp
@@ -90,8 +90,14 @@ namespace hpx::util::detail {
             {
                 if constexpr (!std::is_default_constructible_v<T>)
                 {
+                    // non-default constructible types are handled by the user
                     using serialization::detail::load_construct_data;
                     load_construct_data(ar, &t, 0);
+                }
+                else
+                {
+                    // default constructible types still need to be constructed
+                    ::new (&t) T;
                 }
                 ar >> t;
             }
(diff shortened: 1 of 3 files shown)


hkaiser:
@esseivaju please verify that this patch solves the issue you reported

esseivaju:
@hkaiser Thanks for the quick fix! I just tested the patch which does fix the issue on my side.

PR #6003, opened 2022-09-05
Always exporting get_component_name implementations

### LINKED ISSUE #5995: Linker error with hpx::new_<T[]>()
## Expected Behavior

[Documentation](https://hpx-docs.stellar-group.org/latest/html/manual/writing_distributed_hpx_applications.html#defining-components) and [examples](https://github.com/STEllAR-GROUP/hpx/blob/a50602d97bf709b7128164d0d59d82bcb4afe678/examples/hello_world_component/hello_world_component.cpp#L27) suggest to place the macro ```HPX_REGISTER_COMPONENT``` in a source file. The expected behavior is to be able to instantiate the component in other translation units.

## Actual Behavior

If we try to instantiate the component in a different translation unit using ```hpx::new_<T[]>()```, the linker won't find the definition of ```char const* hpx::components::get_component_name<T, void>()``` as it is a function template defined in a source file.

[This commit](https://github.com/esseivaju/HPXDistributed/commit/4c71bd32f635574c9742556b58a7fa326671ea54) shows the diff between using ```hpx::new_<T>()``` and ```hpx::new_<T[]>()```, it will fail to link when setting the Cmake option ```-DSINGLE_WORKER=OFF``` 

## Specifications

  - HPX Version: 1.8.1
  - Platform (compiler, OS): Cori, GCC 11.2.0

hkaiser:
Fixes #5995

### DIFF
--- a/libs/full/components_base/include/hpx/components_base/component_type.hpp
+++ b/libs/full/components_base/include/hpx/components_base/component_type.hpp
@@ -185,12 +185,12 @@ namespace hpx { namespace components {
 
     // Returns the (unique) name for a given component
     template <typename Component, typename Enable = void>
-    constexpr char const* get_component_name() noexcept;
+    HPX_ALWAYS_EXPORT char const* get_component_name() noexcept;
 
     // Returns the (unique) name of the base component. If there is none,
     // nullptr is returned
     template <typename Component, typename Enable = void>
-    constexpr const char* get_component_base_name() noexcept;
+    HPX_ALWAYS_EXPORT const char* get_component_base_name() noexcept;
 
     template <typename Component>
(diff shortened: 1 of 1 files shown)


hkaiser:
@esseivaju: please verify that this patch solves the issue you reported.

esseivaju:
@hkaiser I tried compiling my example with the patch but still get the same linker error

hkaiser:
@esseivaju could you please try one more time?

esseivaju:
@hkaiser The last commit did fix the issue on my side.

hkaiser:
> @hkaiser The last commit did fix the issue on my side.

Yay! Thanks!

hkaiser:
bors merge

PR #6007, opened 2022-09-07
docs: add docs for mutex.hpp

bhumitattarde:


libs/core/synchronization/include/hpx/synchronization/mutex.hpp
```diff
@@ -32,75 +33,398 @@ namespace hpx { namespace threads {
...
         typedef hpx::spinlock mutex_type;
+        /// \endcond NOPROTECTED
 
     public:
+        ///
+        /// \brief Constructs the \a mutex. The \a mutex is in unlocked state
+        ///        after the constructor completes.
+        ///
+        ///       FIXME(bhumit->hkaiser): true for hpx::mutex?
+        /// \note Because the default constructor is \a constexpr, static
```
bhumitattarde, line 76: Is this true for hpx::mutex as well?

libs/core/synchronization/include/hpx/synchronization/mutex.hpp
```diff
@@ -32,75 +33,398 @@ namespace hpx { namespace threads {
...
-            char const* description, error_code& ec = throws);
+        ///
+        /// \brief Locks the mutex. If another thread has already locked the
+        ///        mutex, a call to lock will block execution until the lock is
+        ///        acquired.
+        ///        If lock is called by a thread that already owns the mutex,
+        ///        the behavior is undefined: for example, the program may
+        ///        deadlock.
+        ///        FIXME(bhumit->hkaiser) do we follow either of the following?
+        ///        An implementation that can detect the invalid
```
bhumitattarde, line 102: @hkaiser does our implementation follow either of these?

libs/core/synchronization/include/hpx/synchronization/mutex.hpp
hkaiser: Good catch, we should apply the corresponding changes. Could you please create a ticket for us not to forget? In the end, all we need is to add a constexpr default constructor and leave the one that takes the (currently optional) string in place without a default value.

libs/core/synchronization/include/hpx/synchronization/mutex.hpp
hkaiser: Yes, this is true for HPX as well.

libs/core/synchronization/include/hpx/synchronization/mutex.hpp
```diff
@@ -32,75 +33,398 @@ namespace hpx { namespace threads {
...
+        /// \brief Tries to lock the mutex. Blocks until specified \a rel_time
+        ///        has elapsed or the lock is acquired, whichever comes first.
+        ///        On successful lock acquisition returns \a true, otherwise
+        ///        returns \a false.
+        ///        If \a rel_time is less or equal \a rel_time.zero(), the
+        ///        function behaves like \a try_lock().
+        ///        This function may block for longer than \a rel_time due to
+        ///        scheduling or resource contention delays.
+        ///        FIXME(bhumit->hkaiser) what does our implementation use?
+        ///        The standard recommends that a \a steady_clock is used to
```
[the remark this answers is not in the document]
hkaiser: Yes, we use chrono::steady_clock as well.

libs/core/synchronization/include/hpx/synchronization/mutex.hpp
bhumitattarde: Will do.

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

Issue #2271, opened 2016-07-29
Add vectorization to the par_vec (aka par_unseq) implementations of the parallel algorithms


brycelelbach:
The par_vec (aka par_unseq) policy allows interleaving of element access functions, e.g. it is safe to the iterations of the algorithm.

Explicit engagement of compiler vectorizers through pragmas is probably the best way to ensure this occurs (e.g. `#pragma simd`, `#pragma omp simd`).

I will probably take a look into doing this myself while preparing my CppCon talk on parallel algorithms.

diehlpk:
@brycelelbach @hkaiser Could you please add a project description here https://github.com/STEllAR-GROUP/hpx/wiki/GSoC-2017-Project-Ideas

Johan511:
I am interested in working on this project. I have seen that in the previous PRs we have added openMP pragmas for vectorization and parallelisation of a loop. Can someone guide me on how I can start out with working on this issue?

hkaiser:
> I am interested in working on this project. I have seen that in the previous PRs we have added openMP pragmas for vectorization and parallelisation of a loop. Can someone guide me on how I can start out with working on this issue?

Yes, we have implemented this for the first batch of algorithms. There are still algorithms left that have not been touched, though. Also, we would need a thorough performance analysis of the existing implementation, combined with improvements, if needed.

Johan511:
par_unseq implementation for algorithms

* [x]  `adjacent_difference`
* [ ]  `inner_product`  does it support any execution policy, could not find doc. Do we Implement using transform reduce?
* [x]  `adjacent_find` 
* [x]  `all_of` `any_of` `none_of` 
------------------------------------
* [ ]  `copy` `copy_if` `copy_n`
* [ ]  `move`
* [ ]  `count` `count_if`
* [ ]  `equal` `mismatch` 
* [ ]  `exclusive_scan` `inclusive_scan` 
* [ ]  `reduce` `transform`
* [ ]  `fill` `fill_n`
* [ ]  `find` `find_end` `find_first_of` `find_if` `find_if_not`
* [x]  `for_each` `for_each_n`
* [ ]  `generate` `generate_n`
* [ ]  `is_heap` `is_heap_until` 
* [ ]  `is_partitioned` `is_sorted` `is_sorted_until`
* [ ]  `lexicographical_compare`
* [ ]  `max_element` `min_element` `minmax_element`
* [ ]  `make_heap`
* [ ]  `partial_sort`
[...]

PR #6016, opened 2022-09-18
Config and structural updates to support unseq implementation

akcube:
## Proposed Changes
- Config file for auto-vectorization user provided hints
- Structural changes to account for unseq implementations

#2271

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/unseq.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/unseq.hpp
@@ -1,0 +1,9 @@
+//  Copyright (c) 2022 A Kishore Kumar
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/config.hpp>
--- a/libs/core/algorithms/tests/unit/unseq_algorithms/CMakeLists.txt
+++ b/libs/core/algorithms/tests/unit/unseq_algorithms/CMakeLists.txt
@@ -1,0 +1,22 @@
+set(tests "")
+
+foreach(test ${tests})
+  set(sources ${test}.cpp)
+
+  set(${test}_PARAMETERS THREADS_PER_LOCALITY 4)
+
+  source_group("Source Files" FILES ${sources})
+
+  # add example executable
+  add_hpx_executable(
+    ${test}_test INTERNAL_FLAGS
+    SOURCES ${sources} ${${test}_FLAGS}
+    EXCLUDE_FROM_ALL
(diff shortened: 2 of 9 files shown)


libs/core/algorithms/include/hpx/parallel/unseq.hpp
```diff
@@ -0,0 +1,9 @@
+//  Copyright (c) 2022 A Kishore Kumar
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/config.hpp>
```
hkaiser, line 9: Why do you need this file? Wouldn't #including <hpx/config.hpp> instead of this file be sufficient?

libs/core/algorithms/tests/unit/unseq_algorithms/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+set(tests "")
```
hkaiser, line 1: This file misses Copyright, License, and SPDX-License-Identifier. It also misses an endline on the last line.

libs/core/config/include/hpx/config/auto_vectorization.hpp
```diff
@@ -0,0 +1,96 @@
...
+#define HPX_LANE_SIZE 64
+#elifdef __AVX2__
+#define HPX_LANE_SIZE 32
+#elifdef __SSE3__
+#define HPX_LANE_SIZE 16
+#else
+#define HPX_LANE_SIZE 64
+#endif
+
+#endif
```
hkaiser, line 96: Endline is missing here.

libs/core/config/include/hpx/config/auto_vectorization.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+#ifdef __AVX512F__
+#define HPX_LANE_SIZE 64
+#elifdef __AVX2__
+#define HPX_LANE_SIZE 32
+#elifdef __SSE3__
+#define HPX_LANE_SIZE 16
+#else
+#define HPX_LANE_SIZE 64
+#endif
```
hkaiser, line 94: I think the default should be no vectorization, e.g. if this is compiled on an ARM/SVE platform or similar. Alternatively, let's cause the compilation to fail (have an `#error`.

libs/core/config/include/hpx/config/auto_vectorization.hpp
hkaiser, line 96: All of your new files miss endline characters on the last lines.

hkaiser (changes requested):

hkaiser:
This PR is superseded by #6018. Closing without merging. All comments here should be addressed in #6018.

PR #6017, opened 2022-09-20
loop and transform_loop unseq adaptation

akcube:
## Proposed Changes
- Implements the unseq adaptation of `util::loop` and `transform_loop`. Will affect `transform`, `for_each` and `for_loop`.
- Added tests for both.

This is based on #6016

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/unseq/loop.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/unseq/loop.hpp
@@ -1,0 +1,356 @@
+//  Copyright (c) 2022 A Kishore Kumar
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/config.hpp>
+
+#include <hpx/executors/execution_policy.hpp>
+#include <hpx/functional/tag_invoke.hpp>
+#include <hpx/parallel/util/loop.hpp>
+
+#include <algorithm>
+#include <cstddef>
+#include <cstdint>
+#include <iterator>
+#include <type_traits>
+#include <utility>
+
+namespace hpx { namespace parallel { namespace util {
+
+    ///////////////////////////////////////////////////////////////////////////
+    namespace detail {
+        ///////////////////////////////////////////////////////////////////////
+        struct unseq_loop_n
+        {
+            template <typename InIter, typename F>
+            HPX_HOST_DEVICE HPX_FORCEINLINE static InIter call(
+                InIter it, std::size_t count, F&& f)
+            {
+                // clang-format off
+                HPX_IVDEP HPX_UNROLL HPX_VECTORIZE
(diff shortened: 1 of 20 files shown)


libs/core/algorithms/include/hpx/parallel/unseq/loop.hpp
```diff
@@ -0,0 +1,356 @@
...
+            template <typename Begin, typename End, typename F>
+            HPX_HOST_DEVICE HPX_FORCEINLINE static typename std::enable_if<
+                !hpx::traits::is_random_access_iterator_v<Begin>, Begin>::type
+            call(Begin it, End end, F&& f)
+            {
+                for (/* */; it != end; it++)
+                {
+                    HPX_INVOKE(f, it);
+                }
+            }
```
hkaiser, line 102: Could we make this fall back to the existing non-vectorized implementation?

libs/core/algorithms/include/hpx/parallel/unseq/loop.hpp
```diff
@@ -0,0 +1,356 @@
...
+                F&& f)
+            {
+                for (/* */; it != end; ++it)
+                {
+                    HPX_INVOKE(f, it);
+                    if (tok.was_cancelled())
+                        return it;
+                }
+                return it;
+            }
```
hkaiser, line 129: Same here: Could we make this fall back to the existing non-vectorized implementation?

libs/core/algorithms/include/hpx/parallel/unseq/loop.hpp
```diff
@@ -0,0 +1,356 @@
...
+            HPX_HOST_DEVICE HPX_FORCEINLINE static typename std::enable_if<
+                !hpx::traits::is_random_access_iterator_v<Begin>, Begin>::type
+            call(Begin it, End end, F&& f)
+            {
+                for (/* */; it != end; ++it)
+                {
+                    HPX_INVOKE(f, *it);
+                }
+                return it;
+            }
```
hkaiser, line 153: Could we make this fall back to the existing non-vectorized implementation? This comment applies to many more places in this file below.

hkaiser (commented):
I think we could reduce the amount of code duplication in loop.hpp et.al. by enabling the tag_invoke overloads for unseq  for random_access iterators only.

Johan511:
@akcube  can you please help me out with how I can run the algorithms using the par_unseq execution policy?  I am not able to be unseq/loop.hpp in my hpx installation

`sudo find . -name "*loop.hpp*" -print`

`./include/parallel_for_loop.hpp`
`./thread_pools/scheduling_loop.hpp`
`./execution/algorithms/run_loop.hpp`
`./parallel/container_algorithms/for_loop.hpp`
`./parallel/util/transform_loop.hpp`
`./parallel/util/loop.hpp`
`./parallel/algorithms/for_loop.hpp`
`./parallel/datapar/transform_loop.hpp`
`./parallel/datapar/loop.hpp`

also I did place a breakpoint at loop.hpp and did find the execution breaking over [here](https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/algorithms/include/hpx/parallel/util/loop.hpp#L302). Was this the expected behaviour?

PR #6024, opened 2022-09-27
Add --start-group and --end-group before libraries are added for linking

stevenrbrandt:
Recently, I attempted to compile simple_resource_partitioner.cpp using hpxcxx and got some undefined symbols, but these symbols were defined in libhpx_init.a, a library that is already on the link line.

## Proposed Changes
Add -Wl,--start-group before linking libs and -Wl,--end-group after linking libs.


hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6026, opened 2022-09-28
add option hpx:force_ipv4 to force resolving hostnames to ipv4 adresses

Neumann-A:
## Proposed Changes

  - add an option to force resolving hostnames to ipv4 adresses

Local hostnames on windows are commonly resolved to ipv6 addresses while remote hostnames resolved to ipv4 addresses. This leads to hpx applications to listen on the wrong addresses if using hostnames on windows. 

The following code might also be related to this since pinging localhost returns a ipv6 `::1` instead of ipv4. 

https://github.com/STEllAR-GROUP/hpx/blob/435ae68f04084df3fa652ef76571b78db1980e8d/libs/core/asio/src/map_hostnames.cpp#L26-L36

## Checklist

- [x] I have added a new feature ~~and have added tests to go along with it.~~ (I don't even see where to add a test here;somewhere in CI?)
- [ ] ~~I have fixed a bug and have added a regression test~~.
- [ ] ~~I have added a test using random numbers; I have made sure it uses a seed, and that random numbers generated are valid inputs for the tests~~.

## Other comment.

Maybe HPX should detect ipv4/ipv6 mismatch on its own since it has to lookup the endpoints any way?

### DIFF
--- a/libs/core/asio/include/hpx/asio/map_hostnames.hpp
+++ b/libs/core/asio/include/hpx/asio/map_hostnames.hpp
@@ -27,7 +27,7 @@ namespace hpx { namespace util {
             transform_function_type;
 
         map_hostnames(bool debug = false)
-          : debug_(debug)
+          : ipv4_(false), debug_(debug)
         {
         }
 
@@ -46,12 +46,18 @@ namespace hpx { namespace util {
             transform_ = f;
         }
 
+        void force_ipv4(bool const& f)
+        {
+            ipv4_ = f;
+        }
+
         std::string map(std::string host_name, std::uint16_t port) const;
 
     private:
         transform_function_type transform_;
         std::string suffix_;
         std::string prefix_;
+        bool ipv4_;
         bool debug_;
     };
 }}    // namespace hpx::util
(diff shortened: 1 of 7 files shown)


libs/core/asio/include/hpx/asio/map_hostnames.hpp
```diff
@@ -46,12 +46,18 @@ namespace hpx { namespace util {
             transform_ = f;
         }
 
+        void force_ipv4(bool const& f)
```
hkaiser, line 49: Why is this a `const&`?

hkaiser (approved):
Otherwise, LGTM, thanks!

libs/core/asio/include/hpx/asio/map_hostnames.hpp
Neumann-A, line 49:
```suggestion
        void force_ipv4(bool f)
```

Neumann-A: just c&p from the function above. Change it to only be bool

hkaiser (approved):

hkaiser:
bors merge

PR #6030, opened 2022-09-29
Adding parcelport initialization hook for resource partitioner operation

hkaiser:
This adds a hook to all parcelports that will be invoked during initialization giving each parcelport a chance to create thread pools or perform other initialization work involving the resource partitioner.


hkaiser:
The problem with this PR (ATM) is that a parcelport will possibly create a thread pool regardless whether it will be actually used in the end or not. Some refinement is still required.

hkaiser:
bors merge

PR #6033, opened 2022-10-04
Support for data-parallelism for replace, replace_if, replace_copy, replace_copy_if algorithms

srinivasyadav18:
Contributes to Fixing https://github.com/STEllAR-GROUP/hpx/issues/2333

## Proposed Changes

  - Moved replace algorithm implementation to detail/replace.hpp
  - Added datapar replace implementation
  - Added unit tests for datapar replace

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/datapar/replace.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/datapar/replace.hpp
@@ -1,0 +1,306 @@
+//  Copyright (c) 2022 Srinivas Yadav
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/config.hpp>
+
+#if defined(HPX_HAVE_DATAPAR)
+#include <hpx/execution/traits/is_execution_policy.hpp>
+#include <hpx/execution/traits/vector_pack_conditionals.hpp>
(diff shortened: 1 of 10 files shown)


libs/core/algorithms/include/hpx/parallel/datapar/replace.hpp
```diff
@@ -0,0 +1,306 @@
...
+
+    template <typename ExPolicy, typename InIter, typename T1, typename T2,
+        typename Proj,
+        HPX_CONCEPT_REQUIRES_(
+            hpx::is_vectorpack_execution_policy<ExPolicy>::value)>
+    HPX_HOST_DEVICE HPX_FORCEINLINE auto tag_invoke(
+        sequential_replace_t<ExPolicy>, ExPolicy&& policy, InIter first,
+        InIter last, T1 const& old_value, T2 const& new_value, Proj&& proj)
+    {
+        if constexpr (hpx::parallel::util::detail::iterator_datapar_compatible<
```
hkaiser, line 72: Could you add this to the `HPX_CONCEPT_REQUIRES_` above, thus avoiding duplicating the `else` case below? This might also require adding the `to_non_simd` to the non-datapar case, though.

hkaiser (changes requested):

libs/core/algorithms/include/hpx/parallel/datapar/replace.hpp
hkaiser: Let's leave this PR as is and get back to the mentioned issue later.

hkaiser (approved):

hkaiser:
bors merge

PR #6035, opened 2022-10-06
Simplify datapar replace copy if

srinivasyadav18:
Fixes #6033 

## Proposed Changes

  - To use util::in_out_result to return the result from zip_iterator.


hkaiser (approved):
LGTM, thanks!

PR #6037, opened 2022-10-07
Avoid performing parcel related background work if networking is disabled

hkaiser:
This fixes a performance regression if running on one locality if a parcel port is enabled that performs background work. @diehlpk this should resolve the issue you reported.


diehlpk:
@G-071 Can you please check that on Ookami, since I do not have access to Fugaku.

PR #6038, opened 2022-10-07
Clean up command line handling

hkaiser:
- flyby: add purely local hello world example
- flyby: fix APPLICATION name for local runtime


hkaiser:
bors merge

PR #6042, opened 2022-10-11
Modernize dataflow API

hkaiser:
- introduce `hpx::detail::dataflow_t` CPO
- deprecate `dataflow_alloc()` and `dataflow<Action>()`


hkaiser:
bors merge

PR #6044, opened 2022-10-15
Generalize the use of then() and dataflow

hkaiser:
- futures are not senders anymore (by default), this introduces `as_sender()` that can be used for this purpose
- fixed reverse and rotate algorithms to properly work with sender/receiver based executors
- flyby: modernize reverse and rotate algorithm implementations
- flyby: adding hpx::tuple_size_v (C++17)


hkaiser:
bors merge

PR #6046, opened 2022-10-19
Support using unwrapping with nullary function objects

### LINKED ISSUE #6045: Can't call nullary callables wrapped with `hpx::unwrapping`
## Expected Behavior

`hpx::unwrapping` with a nullary lambda should be callable, e.g. `hpx::unwrapping([]{})()` (even though the unwrapping isn't needed in this particular case, it's useful for generic contexts).

Note that this isn't something we rely on, but we happened to bump into this accidentally. I thought you might be interested.

## Actual Behavior

It fails to compile. Note that simply creating the `unwrapping` callable with `hpx::unwrapping([]{})` compiles just fine. Calling it does not work.

Attempt to compile the following:
```
#include <hpx/unwrap.hpp>

int main() {
    hpx::unwrapping([]{})();
}
```

## Specifications

  - HPX Version: latest master (https://github.com/STEllAR-GROUP/hpx/commit/1e4b87375d907e88c57c50e90a26179689f46485)
  - Platform (compiler, OS): GCC 11.3.0

hkaiser:
Fixes #6045

### DIFF
--- a/libs/core/pack_traversal/include/hpx/pack_traversal/detail/pack_traversal_impl.hpp
+++ b/libs/core/pack_traversal/include/hpx/pack_traversal/detail/pack_traversal_impl.hpp
@@ -868,6 +868,11 @@ namespace hpx { namespace util { namespace detail {
         }
 
         /// \copybrief try_traverse
+        template <typename Tag>
+        constexpr void init_traverse(Tag) const noexcept
+        {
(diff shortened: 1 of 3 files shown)


hkaiser:
retest lsu

hkaiser:
bors merge

PR #6051, opened 2022-10-25
Add non-contiguous index queue and use it in thread_pool_bulk_scheduler

hkaiser:
This changes the sequence of partitions scheduled on the same core.

@hcq9102 Could you please run your performance scripts for the S/R executor using PR to see if this behaves differently for larger numbers of cores?

### DIFF
--- a/libs/core/algorithms/CMakeLists.txt
+++ b/libs/core/algorithms/CMakeLists.txt
@@ -178,6 +178,7 @@ set(algorithms_headers
     hpx/parallel/spmd_block.hpp
     hpx/parallel/task_block.hpp
     hpx/parallel/task_group.hpp
+    hpx/parallel/util/adapt_placement_mode.hpp
     hpx/parallel/util/cancellation_token.hpp
     hpx/parallel/util/compare_projected.hpp
     hpx/parallel/util/detail/algorithm_result.hpp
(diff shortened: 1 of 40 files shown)


hkaiser:
Closing this as the change makes things worse, performance-wise.

hkaiser:
Re-opening this as after all this might have some uses (for instance for find() and friends).

hkaiser:
@Pansysk75 this PR enables controlling the placement strategies (breadth-first or depth-first) in the scheduler (currently supported for S/R only). A couple of algorithms are being adapted to take advantage of breadth-first placement. This also enables using S/R with `hpx::find`.

So what this already could be used for is to see whether `hpx::find` performance really improves when changing the placement strategies.

Pansysk75:
> @Pansysk75 this PR enables controlling the placement strategies (breadth-first or depth-first) in the scheduler (currently supported for S/R only). A couple of algorithms are being adapted to take advantage of breadth-first placement. This also enables using S/R with `hpx::find`.

I'll check it out, nice! Just a quick thought:
I made a list of algorithms that are expected to benefit from this scheduling (find, mismatch, search, etc. that return an iterator to the first occurance). 
I would not expect algorithms that return a bool (all_of, equal, etc.) to benefit from this scheduling, since for these we don't care about "first occurance" of something, so we aren't as pressed to immediately schedule the first chunks of the container. Those should cancel as soon as there's any "match", anywhere.

hkaiser:
@Pansysk75 There is also `find_end` (and similar) that would require scheduling things backwards as we're interested in finding the last occurence...

hkaiser:
bors merge

PR #6055, opened 2022-10-28
Patches needed to be able to build HPX 1.8.1 on various platforms

kordejong:
This PR contains a subset of patches I use to be able to build HPX-1.8.1 on all platforms I use.
Not all patches I use are needed anymore with the current state of the master branch. I left those out.

I did not keep track of which platform (OS, compiler) required which patch, unfortunately. Will do that in future cases.

The second patch, in `parcelset_base_fwd.hpp` is weird, but required in some case(s). IIRC, the `locality` class is already annotated with export somewhere else and the compiler didn't like the second occurrence. I could be wrong.

### DIFF
--- a/libs/full/parcelset_base/include/hpx/parcelset_base/parcelset_base_fwd.hpp
+++ b/libs/full/parcelset_base/include/hpx/parcelset_base/parcelset_base_fwd.hpp
@@ -14,7 +14,7 @@
 namespace hpx::parcelset {
 
     class HPX_EXPORT parcelport;
-    class /* HPX_EXPORT */ locality;
+    class HPX_EXPORT locality;
     class HPX_EXPORT parcel;
 
     /// The type of a function that can be registered as a parcel write handler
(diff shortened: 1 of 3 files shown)


hkaiser:
add to whitelist

libs/full/parcelset_base/include/hpx/parcelset_base/parcelset_base_fwd.hpp
```diff
@@ -14,7 +14,7 @@
 namespace hpx::parcelset {
 
     class HPX_EXPORT parcelport;
-    class HPX_EXPORT locality;
+    class /* HPX_EXPORT */ locality;
```
hkaiser, line 17: I think that the `HPX_EXPORT` should stay here (at the declaration). Instead, we should remove it here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/full/parcelset_base/include/hpx/parcelset_base/locality.hpp#L30 (at the definition). That also requires adding `#include <hpx/parcelset_base/parcelset_base_fwd.hpp>` to this file.

kordejong: I have applied this update.

hkaiser:
The test failures are unrelated.

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6057, opened 2022-10-28
Create a dedicated thread pool to run LCI_progress.

JiakunYan:
The `LCI_progress` function is not thread-safe and must be called frequently. Previously, we created a dedicated progress thread using pthread when initializing LCI. However, HPX doesn't know this progress thread and thus cannot bind threads to cores correctly.

This PR uses the new resource partitioner hook in the parcelport to create a thread pool "lci-progress-pool" (which only contains one core) to run LCI_progress.

Fixed:
- The thread pool will be allocated even if users are not using LCI parcelport.

### DIFF
--- a/libs/core/lci_base/src/lci_environment.cpp
+++ b/libs/core/lci_base/src/lci_environment.cpp
@@ -18,6 +18,7 @@
 #include <atomic>
 #include <cstddef>
 #include <cstdlib>
+#include <memory>
 #include <string>
 
 ///////////////////////////////////////////////////////////////////////////////
@@ -103,7 +104,8 @@ namespace hpx { namespace util {
     LCI_comp_t lci_environment::rt_cq_r_;
     LCI_endpoint_t lci_environment::h_ep_;
     LCI_comp_t lci_environment::h_cq_r_;
-    std::thread* lci_environment::prg_thread_p = nullptr;
+    // We need this progress thread to send early parcels
+    std::unique_ptr<std::thread> lci_environment::prg_thread_p = nullptr;
     std::atomic<bool> lci_environment::prg_thread_flag = false;
 
     ///////////////////////////////////////////////////////////////////////////
@@ -142,12 +144,10 @@ namespace hpx { namespace util {
         LCI_endpoint_init(&h_ep_, LCI_UR_DEVICE, h_plist_);
         LCI_plist_free(&h_plist_);
         // DEBUG("Rank %d: Init lci env", LCI_RANK);
-
         HPX_ASSERT(prg_thread_flag == false);
         HPX_ASSERT(prg_thread_p == nullptr);
         prg_thread_flag = true;
-        prg_thread_p = new std::thread(progress_fn);
-
+        prg_thread_p = std::make_unique<std::thread>(progress_fn);
         return LCI_OK;
     }
 
@@ -214,14 +214,19 @@ namespace hpx { namespace util {
             LCI_initialized(&lci_init);
             if (lci_init)
             {
-                HPX_ASSERT(prg_thread_flag.load() == true);
-                HPX_ASSERT(prg_thread_p != nullptr);
-                prg_thread_flag = false;
-                prg_thread_p->join();
-                delete prg_thread_p;
-
+                join_prg_thread_if_running();
                 LCI_finalize();
             }
+        }
+    }
+
+    void lci_environment::join_prg_thread_if_running()
+    {
+        if (prg_thread_flag.load())
+        {
+            HPX_ASSERT(prg_thread_p != nullptr);
+            prg_thread_flag = false;
+            prg_thread_p->join();
         }
     }
 
@@ -231,6 +236,13 @@ namespace hpx { namespace util {
         {
             LCI_progress(LCI_UR_DEVICE);
         }
+    }
+
+    bool lci_environment::do_progress()
+    {
+        LCI_error_t ret = LCI_progress(LCI_UR_DEVICE);
+        HPX_ASSERT(ret == LCI_OK || ret == LCI_ERR_RETRY);
+        return ret == LCI_OK;
     }
 
     bool lci_environment::enabled()
--- a/libs/core/lci_base/src/lci_environment.cpp
+++ b/libs/core/lci_base/src/lci_environment.cpp
@@ -107,6 +107,7 @@ namespace hpx { namespace util {
     // We need this progress thread to send early parcels
     std::unique_ptr<std::thread> lci_environment::prg_thread_p = nullptr;
     std::atomic<bool> lci_environment::prg_thread_flag = false;
+    hpx::spinlock prg_thread_mtx;
 
     ///////////////////////////////////////////////////////////////////////////
     LCI_error_t lci_environment::init_lci()
@@ -222,11 +223,14 @@ namespace hpx { namespace util {
 
     void lci_environment::join_prg_thread_if_running()
     {
-        if (prg_thread_flag.load())
+        if (prg_thread_p)
         {
-            HPX_ASSERT(prg_thread_p != nullptr);
-            prg_thread_flag = false;
-            prg_thread_p->join();
+            std::lock_guard<hpx::spinlock> l(prg_thread_mtx);
+            if (prg_thread_p) {
+                prg_thread_flag = false;
+                prg_thread_p->join();
+                prg_thread_p.reset();
+            }
         }
     }
 
--- a/libs/core/lci_base/src/lci_environment.cpp
+++ b/libs/core/lci_base/src/lci_environment.cpp
@@ -107,7 +107,6 @@ namespace hpx { namespace util {
     // We need this progress thread to send early parcels
     std::unique_ptr<std::thread> lci_environment::prg_thread_p = nullptr;
     std::atomic<bool> lci_environment::prg_thread_flag = false;
-    hpx::spinlock prg_thread_mtx;
 
     ///////////////////////////////////////////////////////////////////////////
     LCI_error_t lci_environment::init_lci()
@@ -225,12 +224,9 @@ namespace hpx { namespace util {
     {
         if (prg_thread_p)
         {
-            std::lock_guard<hpx::spinlock> l(prg_thread_mtx);
-            if (prg_thread_p) {
-                prg_thread_flag = false;
-                prg_thread_p->join();
-                prg_thread_p.reset();
-            }
+            prg_thread_flag = false;
+            prg_thread_p->join();
+            prg_thread_p.reset();
         }
     }
 
--- a/libs/full/parcelport_lci/src/parcelport_lci.cpp
+++ b/libs/full/parcelport_lci/src/parcelport_lci.cpp
@@ -175,7 +175,7 @@ namespace hpx::parcelset {
                 if (stopped_)
                     return false;
 
-                static __thread int do_lci_progress = -1;
+                static thread_local int do_lci_progress = -1;
                 if (do_lci_progress == -1)
                 {
                     if (enable_lci_progress_pool &&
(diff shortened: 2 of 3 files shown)


libs/core/lci_base/src/lci_environment.cpp
```diff
@@ -214,17 +214,22 @@ namespace hpx { namespace util {
...
-
+                join_prg_thread_if_running();
                 LCI_finalize();
             }
         }
     }
 
+    void lci_environment::join_prg_thread_if_running()
+    {
+        if (prg_thread_flag.load())
```
hkaiser, line 225: Should this be a `compare_exchange` instead? The way it's coded now the atomic doesn't protect from races.

libs/core/lci_base/src/lci_environment.cpp
```diff
@@ -214,17 +214,22 @@ namespace hpx { namespace util {
...
+        {
+            HPX_ASSERT(prg_thread_p != nullptr);
+            prg_thread_flag = false;
+            prg_thread_p->join();
```
hkaiser, line 229: Can the thread (-object) be deleted after joining?

libs/core/lci_base/src/lci_environment.cpp
```diff
@@ -233,6 +238,13 @@ namespace hpx { namespace util {
         }
     }
 
+    bool lci_environment::do_progress()
+    {
+        LCI_error_t ret = LCI_progress(LCI_UR_DEVICE);
+        HPX_ASSERT(ret == LCI_OK || ret == LCI_ERR_RETRY);
```
hkaiser, line 244: Note that asserts are disabled (compiled out) in release mode. Should this throw an exception instead?

libs/core/lci_base/src/lci_environment.cpp
JiakunYan: For this specific case, only the thread belonging to the `lci-progress-pool` will call this function, so it will not be called by multiple threads at the same time. However, I admit the code here can be tricky. It would be much better if there is a parcelport function that is called after all the early parcels are sent/received and before the scheduling loop of worker threads starts. Is it possible to have this function?

libs/core/lci_base/src/lci_environment.cpp
JiakunYan: Yes, will do.

libs/core/lci_base/src/lci_environment.cpp
JiakunYan: I think it is fine. I don't want to add an additional branch here in release mode. Currently, LCI_progress will not return a value other than LCI_OK and LCI_ERR_RETRY in any case.

libs/core/lci_base/src/lci_environment.cpp
JiakunYan: I used a lock to make it thread-safe for now, but a cleaner way would be to provide a method in the parcelport interface to join the progress thread after all the early parcels are sent/received and before any worker threads can call the do_background_work function.

libs/core/lci_base/src/lci_environment.cpp
JiakunYan: done.

libs/core/lci_base/src/lci_environment.cpp
hkaiser: > For this specific case, only the thread belonging to the lci-progress-pool will call this function, so it will not be called by multiple threads at the same time.

Why did you use an atomic, then?

> It would be much better if there is a parcelport function that is called after all the early parcels are sent/received and before the scheduling loop of worker threads starts. Is it possible to have this function?

Unfortunately, that cannot be done as the early parcels are being sent only after the schedulers have been brought up. The handling of the early parcels requires for the threading to be functional. I can however notify the parcelports that the early parcel handling is done and normal operation is about to start. Is that what you need?

JiakunYan: > Why did you use an atomic, then?

This variable is written by the thread in lci-progress-pool and read by the thread pointed by prg_thread_p, so I make it an atomic.

> Unfortunately, that cannot be done as the early parcels are being sent only after the schedulers have been brought up. The handling of the early parcels requires for the threading to be functional.

By "the schedulers have been brought up", do you mean the worker threads can already call the do_background_work function? If that is the case, why the thread in lci-progress-pool is not functioning (no LCI_progress is called) when early parcels are sent?

hkaiser: > By "the schedulers have been brought up", do you mean the worker threads can already call the do_background_work function? 

I think so, yes.

> If that is the case, why the thread in lci-progress-pool is not functioning (no LCI_progress is called) when early parcels are sent?

I'm not sure, will investigate.

libs/core/lci_base/src/lci_environment.cpp
```diff
@@ -214,13 +215,21 @@ namespace hpx { namespace util {
...
-                HPX_ASSERT(prg_thread_p != nullptr);
+                join_prg_thread_if_running();
+                LCI_finalize();
+            }
+        }
+    }
+
+    void lci_environment::join_prg_thread_if_running()
+    {
+        if (prg_thread_p)
```
hkaiser, line 226: This change doesn't look right to me. The purpose is to make sure that the thread is joined once. I think the best solution would be to utilize the atomic and write:
```
    bool expected = false;
    if (prg_thread_p.compare_exchange_strong(expected, true))
    {
        prg_thread_p->join();
        prg_thread_p.reset();
    }
```
or even
```
    if (!prg_thread_p.exchange(true))
    {
        prg_thread_p->join();
        prg_thread_p.reset();
    }
```

JiakunYan: I think they both work. Due to the lock, `prg_thread_p` will only be joined once and then immediately reset to nullptr. After that, no other threads can join the thread again.

libs/core/lci_base/src/lci_environment.cpp
hkaiser: @JiakunYan you were right. The pools start running only after early parcel handling (that's also the reason why you need your own thread for those). I will add the notification function as you requested.

JiakunYan: I figured out early parcels are sent in `big_boot_barrier`. Could you point me to the function where schedulers are brought up?

JiakunYan: > I will add the notification function as you requested.

That's great! Thank you!

libs/core/lci_base/src/lci_environment.cpp
hkaiser: I think you don't need the lock as you're using that atomic variable.

JiakunYan:
@hkaiser  The new scheduler runs into errors. The `create_background_thread` function has a line `scheduler.SchedulingPolicy::create_thread`, but it leads to an exception in the new scheduler.

hkaiser:
> @hkaiser The new scheduler runs into errors. The `create_background_thread` function has a line `scheduler.SchedulingPolicy::create_thread`, but it leads to an exception in the new scheduler.

I did not consider the fact that the background threads themselves are handled by the scheduler as well. I think just removing the overrides for `create_thread` https://github.com/STEllAR-GROUP/hpx/blob/d65a7bfd6d226647cf758371b3dffdf94c122aa2/libs/core/schedulers/include/hpx/schedulers/background_scheduler.hpp#L97-L103 and `destroy_thread` https://github.com/STEllAR-GROUP/hpx/blob/d65a7bfd6d226647cf758371b3dffdf94c122aa2/libs/core/schedulers/include/hpx/schedulers/background_scheduler.hpp#L123-L128 for the new scheduler should help. Could you try that, please?

JiakunYan:
@hkaiser I have to removing `create_thread`, `destro_thread`, and `schedule_thread` since there are two lines of
```
scheduler.SchedulingPolicy::schedule_thread(
                                        HPX_MOVE(background_thread),
                                        threads::thread_schedule_hint(
                                            static_cast<std::int16_t>(
                                                num_thread)),
                                        true, priority);
```
in the scheduling loop. However, after removing them, the program never exits.

JiakunYan:
Removing the overload of `get_next_thread` fix this hanging, but I don't know why.

hkaiser:
> @hkaiser I have to removing `create_thread`, `destro_thread`, and `schedule_thread` since there are two lines of
> 
> ```
> scheduler.SchedulingPolicy::schedule_thread(
>                                         HPX_MOVE(background_thread),
>                                         threads::thread_schedule_hint(
>                                             static_cast<std::int16_t>(
>                                                 num_thread)),
>                                         true, priority);
> ```
> 
> in the scheduling loop. However, after removing them, the program never exits.

Ok, I'll concoct a test and will try to make it work.

JiakunYan:
> Ok, I'll concoct a test and will try to make it work.

Thanks! For your information, I was just using the lci pp in this branch with the `fibonacci_futures_distributed` example.

hkaiser:
@JiakunYan please see https://github.com/STEllAR-GROUP/hpx/pull/6140 for a fix.

JiakunYan:
@hkaiser Some tests are running with only 1 thread per locality. If we enable the lci-progress-pool, the default pool will have no threads. We can either run all tests with at least 2 threads or disable lci-progress-pool for all tests. Which way do you prefer?

hkaiser:
> @hkaiser Some tests are running with only 1 thread per locality. If we enable the lci-progress-pool, the default pool will have no threads. We can either run all tests with at least 2 threads or disable lci-progress-pool for all tests. Which way do you prefer?

Let's do different things on different platforms (if possible).

hkaiser:
> > @hkaiser Some tests are running with only 1 thread per locality. If we enable the lci-progress-pool, the default pool will have no threads. We can either run all tests with at least 2 threads or disable lci-progress-pool for all tests. Which way do you prefer?
> 
> Let's do different things on different platforms (if possible).

Orthogonal comment: can you additionally make that decision in the parcelport as well? E.g. disable the background pool if the application runs using one thread.

JiakunYan:
> Let's do different things on different platforms (if possible).
What do you mean by different platforms? Run two threads on Rostam and one thread on the cscs cluster?

> Orthogonal comment: can you additionally make that decision in the parcelport as well? E.g. disable the background pool if the application runs using one thread.

Okay, how can I get the thread number during initialization?

hkaiser:
> > Let's do different things on different platforms (if possible).
> 
> What do you mean by different platforms? Run two threads on Rostam and one thread on the cscs cluster?

What I meant is to have different configurations on different builders.

> > Orthogonal comment: can you additionally make that decision in the parcelport as well? E.g. disable the background pool if the application runs using one thread.
> 
> Okay, how can I get the thread number during initialization?

During startup the best would be to query the configuration registry: 
```
std::size_t num_cores = hpx::util::from_string<std::size_t>(hpx::get_config_entry("hpx.os_threads"));
```

libs/full/parcelport_lci/src/parcelport_lci.cpp
```diff
@@ -313,15 +313,27 @@ namespace hpx::traits {
...
+                        hpx::util::get_entry_as<bool>(cfg.rtcfg_,
+                            "hpx.parcel.lci.background_only_scheduler",
+                            false /* Does not matter*/);
+            }
+            else
+            {
+                hpx::parcelset::policies::lci::parcelport::
+                    enable_background_only_scheduler = false;
+                hpx::parcelset::policies::lci::parcelport::
+                    enable_background_only_scheduler = false; // doesn't matter
```
hkaiser, line 335: This seems to be duplicated.

JiakunYan: Fixed!

hkaiser (dismissed):
LGTM, thanks!

hkaiser:
@JiakunYan I think this can be merged once you fix the clang-format complaints.

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6060, opened 2022-11-01
Allow to disable registering signal handlers

hkaiser:
This adds the `hpx.handle_signals` configuration option (defaults to `1`), which if set to `0` will disable handling signal handlers. This allows to generate core dumps to support debugging.

This is limited to non-windows systems.

@srinivasyadav18 this PR enables using the `--hpx:ini=hpx.handle_signals!=0` command line option instead of commenting out the signal registration. Please verify.


hkaiser:
bors merge

PR #6062, opened 2022-11-02
Make sure wrapping executor does not go out of scope prematurely

hkaiser:
- when used with a seq(task) execution policy
- flyby: modernize executor_with_thread_hooks example


hkaiser:
bors merge

PR #6063, opened 2022-11-02
Modernize iterator_support module

hkaiser:
- deprecate make_xxx functions, replace those with CTAD deduction guides


hkaiser:
bors merge

PR #6068, opened 2022-11-09
Prevent using task and non_task as a CPO

hkaiser:
This prevents using constructs like `task(par)` while still allowing writing `par(task)` (and similar).


hkaiser:
bors merge

PR #6069, opened 2022-11-09
Do not check for cancellation during each iteration but only once per partition

hkaiser:
The goal is to improve the performance of the algorithms that rely on optional cancellation (like `hpx::find`, et.al.)

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/datapar/loop.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/datapar/loop.hpp
@@ -124,11 +124,11 @@ namespace hpx { namespace parallel { namespace util {
                     traits::vector_pack_size<V>::value;
 
                 End const lastV = last - (size + 1);
-                int offset = 0;
 
                 while (first < lastV)
                 {
-                    offset = datapar_loop_pred_step<Begin>::callv(pred, first);
+                    int offset =
+                        datapar_loop_pred_step<Begin>::callv(pred, first);
                     if (offset != -1)
                     {
                         std::advance(first, offset);
--- a/libs/core/algorithms/include/hpx/parallel/datapar/loop.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/datapar/loop.hpp
@@ -98,6 +98,57 @@ namespace hpx { namespace parallel { namespace util {
         };
 
         ///////////////////////////////////////////////////////////////////////
+        // Helper class to repeatedly call a function starting from a given
+        // iterator position till the predicate returns true.
+        template <typename Iterator>
+        struct datapar_loop_pred
+        {
+            typedef typename std::decay<Iterator>::type iterator_type;
+            typedef typename std::iterator_traits<iterator_type>::value_type
+                value_type;
+
+            typedef typename traits::vector_pack_type<value_type>::type V;
+
+            template <typename Begin, typename End, typename Pred>
+            HPX_HOST_DEVICE HPX_FORCEINLINE static Begin call(
+                Begin first, End last, Pred&& pred)
+            {
+                while (!is_data_aligned(first) && first != last)
+                {
+                    if (datapar_loop_pred_step<Begin>::call1(pred, first) != -1)
+                        return first;
+                    ++first;
+                }
+
+                static std::size_t constexpr size =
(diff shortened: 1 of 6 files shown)


Pansysk75:
@hkaiser  I think you missed **loop_idx_n** and **datapar_loop_idx_n**, could you have a look?

hkaiser:
> @hkaiser I think you missed **loop_idx_n** and **datapar_loop_idx_n**, could you have a look?

I left those out on purpose as the cancellation function expects to receive the current element index. I was not sure if changing this would break anything. I will look again.

Pansysk75:
> > @hkaiser I think you missed **loop_idx_n** and **datapar_loop_idx_n**, could you have a look?
> 
> I left those out on purpose as the cancellation function expects to receive the current element index. I was not sure if changing this would break anything. I will look again.

Aha yes, didn't think it through, thanks.
It's used here for example, which is the default fallback when using hpx::find.
Probably in other places as well.

https://github.com/STEllAR-GROUP/hpx/blob/018de710acc5dc3d6ff6f55b8aed3b35ee610885/libs/core/algorithms/include/hpx/parallel/algorithms/detail/find.hpp#L55

hkaiser:
@Pansysk75 I think this is different. You referred to the iteration function itself, what I meant is the cancellation function receiving the index, e.g. here: 

https://github.com/STEllAR-GROUP/hpx/blob/018de710acc5dc3d6ff6f55b8aed3b35ee610885/libs/core/algorithms/include/hpx/parallel/util/loop.hpp#LL932

hkaiser:
> @Pansysk75 I think this is different. You referred to the iteration function itself, what I meant is the cancellation function receiving the index, e.g. here:
> 
> https://github.com/STEllAR-GROUP/hpx/blob/018de710acc5dc3d6ff6f55b8aed3b35ee610885/libs/core/algorithms/include/hpx/parallel/util/loop.hpp#L932

To answer my own question: I think we can apply the same change to the facilities you mentioned as the index that is passed to the cancellation function is really used just to verify whether a better match was already found. I will apply the change for those as well.

Pansysk75:
I run a few test cases on a Medusa node (20 cores), on representative algorithms that use cancellation tokens (find/find_if, search, all_of).

The performance impact of this PR depends heavily on two factors:

1. Likelyhood of cancellable event occuring (eg "find" finding a match)
2. Amount of work per chunk

- If the cancelable event never occurs, minimizing the cancelation-checking overhead is always better, since we never benefit from the canceling anyways. Below, I set up a case where the cancellable event never occurs.
 
find_0

- On the other hand, if the cancelable event is likely, the benefit of checking every loop iteration becomes noticable when the chunk size (or, the amount of work per chunk) gets large.  

find_70  

 Below is find_if with a predicate that requires non-trivial work. Checking the token every loop iteration is now better, because a. 
  the relative overhead of the token becomes smaller and b.  we skip as much of the heavy work as we can.

find_if_70

Similar results were obtained on the other algorithms tested.
That's all for now, let me now what you think.

hkaiser:
@Pansysk75 thanks a lot! Your results make perfect sense! The conclusion seems to be that we should leave the decision to the user after all.

Pansysk75:
> @Pansysk75 I think we should go ahead with this PR as even in the worst case, the new code is close in performance to the old code. In contrast, the new code has the potential to improve performance significantly.

Unfortunately, one can create scenarios where each of the two choices performs atrociously, so I didn't make a good job of representing both sides fairly here. In fact, I can more easily think of worse worst-case scenarios for this PR, than without it. 
I'd love to discuss this when we are able to. 
(Sorry for the late response here, still getting used to checking my GitHub notifications)

libs/core/algorithms/include/hpx/parallel/datapar/loop.hpp
```diff
@@ -102,41 +124,28 @@ namespace hpx { namespace parallel { namespace util {
                     traits::vector_pack_size<V>::value;
 
                 End const lastV = last - (size + 1);
+                int offset = 0;
```
hkaiser, line 127: Please remove this variable, it's not used anymore.

hkaiser:
bors merge

PR #6085, opened 2022-11-24
Add experimental sycl integration/executor

G-071:
## This pull request adds a basic, experimental HPX-SYCL integration
In a nutshell: This allows us to get HPX futures from asynchronous SYCL calls. 

### The main features added are:
1. Allow compilation with dpcpp (Intel OneAPI) and [hipSYCL](https://github.com/illuhad/hipSYCL). For dpcpp we use the compiler arguments ```-fsycl``` (for sycl related files) and ```-fno-sycl``` for normal HPX files. Due to problems with the compilation using hipSYCL's syclcc, we rely on its cmake integration here instead.
2. Added event polling to create HPX futures from SYCL events. Effectively, this allows to write code like: 
```cpp
        // Submit kernel into sycl queue [pure SYCL]
        cl::sycl::event my_kernel_event = my_sycl_queue.submit([&](cl::sycl::handler& h) {
            cl::sycl::accessor a(a_buf, h, cl::sycl::read_only);
            cl::sycl::accessor b(b_buf, h, cl::sycl::read_only);
            cl::sycl::accessor results(results_buf, h, cl::sycl::write_only, cl::sycl::no_init);
            h.parallel_for(num_items, [=](auto i) { /* insert numeric code here */ });
        });
        // Get future from event [HPX-SYCL]
        hpx::future<void> my_kernel_future =
            hpx::sycl::experimental::detail::get_future(my_kernel_event);
```
3. Bolted on top of this basic get_future functionality, the PR also adds an experimental SYCL executor. The executor encapsulates a SYCL queue and allows calling queue member functions using ```async_execute``` and ```post``` by specifying the desired member function, as well as passing all required arguments like this:
```cpp
        // Create executor
        hpx::sycl::experimental::sycl_executor exec(cl::sycl::default_selector{});
        // Launch kernel
        hpx::future<void> fut = exec.async_execute(
            &cl::sycl::queue::submit, [&](cl::sycl::handler& h) {
                cl::sycl::accessor a(a_buf, h, cl::sycl::read_only);
                cl::sycl::accessor b(b_buf, h, cl::sycl::read_only);
                cl::sycl::accessor results(results_buf, h, cl::sycl::write_only, cl::sycl::no_init);
                h.parallel_for(num_items, [=](auto i) { /* insert numeric code here */ });
            });
```
4. Added three simple vector_add tests/examples: One just using the basic get_future integration, one using the executor and a last one subdividing the problem into chunks and using multiple executors, launching kernels from multiple threads.

### Some implementation details:
I used the CUDA HPX integration as a template, so overall it is rather similar to the event polling implemented there (see PR  #4385). However, I had to adapt things for a) differences in the SYCL specification (lack of normal callbacks, lack of the ability to just create and record sycl events without any kernel/memcpy calls) and b) for differences between two used SYCL implementations (oneapi, hipsycl). Notable SYCL-implementation-specific workarounds for said differences:
- For hipSYCL, we need to flush hipSYCL's internal DAG after launching a kernel and creating a future, as it is not guaranteed that hipSYCL is actually going to start the kernel before we request its results (resulting in a deadlock).
- Also for hipSYCL: We need to keep the hipSYCL runtime alive whilst the event polling is activated - if we don't we might recreate and destroy the entire hipSYCL runtime each time we call poll() as long as no other SYCL object is alive. This can result in significant overhead during phases where we don't have a kernel/executor active (which was already noticeable in the concurrent executor test and the beginning and ending of the test). This can be prevented by simply keeping an extra queue alive as long as the polling is active, which is done internally (so the user won't have to worry about it).
- For oneAPI: Here the sycl queue member functions have a different interface (including a code_location parameter), requiring adaptations in the sycl executor ```async_execute``` and ```post``` implementation to find the correct member function given the passed function pointer.
- Compilation differences: For oneAPI we could simply compile with dpcpp (given the correct flags for sycl/none-sycl files), for hipSYCL the compilation with its syclcc caused too many compilation issues, so I ended up using its cmake integration instead.

### How to build/test:
I tested this PR on my local laptop which contains an integrated Intel GPU (Iris Xe) and a discrete NVIDIA Geforce GTX 1650). I used the Intel GPU with oneAPI, the NVIDIA GPU with hipSYCL. Hence there are two build configurations:

1. With OneAPI (using dpcpp version 2022.2.1.20221020 as installed from the ubuntu 20.04 PPA) I used:
```
cmake -DCMAKE_CXX_COMPILER=dpcpp -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DHPX_WITH_EXAMPLES=ON -DHPX_WITH_TESTS=ON -DHPX_WITH_TESTS_UNIT=ON -DHPX_WITH_NETWORKING=OFF -DHPX_WITH_FETCH_ASIO=ON -DHPX_WITH_CXX17=ON -DHPX_WITH_CUDA=OFF -DHPX_WITH_SYCL=ON -DHPX_WITH_MALLOC=system -DCMAKE_INSTALL_PREFIX=/where/ever/you/want/to/have/this /hpx-with-PR-checked-out
```

2. With hipSYCL (I used hipSYCL commit ee450740dade99b9174cbc0845cf881ec64bdd10 from early October 2022), first configure hipSYCL like this (assuming you use clang 12 like I did and have an NVIDIA GPU):
```
 cmake -DCMAKE_INSTALL_PREFIX=/hipsycl/install/path -DCMAKE_CXX_COMPILER=clang++-12 -DCMAKE_C_COMPILER=clang-12  -DWITH_CUDA_BACKEND=ON ../hipSYCL/
```
then make and install hipsycl.
Using this, configure HPX like this:
```
cmake -DCMAKE_CXX_COMPILER=clang++-12 -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DHPX_WITH_EXAMPLES=ON -DHPX_WITH_TESTS=ON -DHPX_WITH_NETWORKING=OFF -DHPX_WITH_FETCH_ASIO=ON -DHPX_WITH_CXX17=ON -DHPX_WITH_CUDA=ON -DHPX_WITH_MALLOC=system -DCMAKE_INSTALL_PREFIX=/where/ever/you/want/to/have/this-DhipSYCL_DIR=/hipsycl/install/path/lib/cmake/hipSYCL -DHIPSYCL_TARGETS="cuda.integrated-multipass:sm_75" -DCMAKE_CXX_FLAGS="-Wno-unknown-cuda-version " -DHPX_WITH_SYCL=ON -DHPX_WITH_HIPSYCL=ON /hpx-with-PR-checked-out
```
(change HIPSYCL_TARGETS depending on what GPU you have available -- consult the hipSYCL documentation for this).

Once all of this is done, you can simply build the HPX SYCL test with 
```
make tests.unit.modules.async_sycl 
```
You can run them with
``` ctest -R sycl  ```
or, to run the tests/examples individually from the build directory:
```
./bin/sycl_vector_add_get_future_test
./bin/sycl_vector_add_executor_test 
./bin/sycl_vector_add_concurrent_executors_test -t 4
```

### Possible next steps:
- While I am considering the SYCL executor itself to be rather experimental (basically a proof-of-concept), the basic get_future functionality seems to work well. In theory, this can already be used in HPX-Kokkos for getting futures from Kokkos kernels that run on the SYCL execution space. This could in turn allow applications like Octo-Tiger to use the Kokkos SYCL execution space.
- More examples/polish and testing on machines other than my laptop.

### CODE REVIEW

**Comment by hkaiser** [comment]
@G-071 Thanks for doing this!

Does this also enable writing:
```
hpx::sycl::experimental::sycl_executor exec(cl::sycl::default_selector{});
hpx::async(exec, ...);
```
(in theory it should)?

File: cmake/HPX_SetupSYCL.cmake [hunk 1]
```diff
@@ -0,0 +1,17 @@
+if(HPX_WITH_SYCL)
```
hkaiser, line 1: Please add a copyright, license, and SPDX-License to the file.

CMakeLists.txt
```diff
@@ -545,6 +545,14 @@ if(HPX_WITH_CUDA AND HPX_WITH_HIP)
   )
 endif()
 
+# ##############################################################################
+# HPX SYCL configuration
+# ##############################################################################
+set(SYCL_OPTION_STRING "Enable support for Sycl (default: OFF)")
+hpx_option(HPX_WITH_SYCL BOOL "${SYCL_OPTION_STRING}" OFF ADVANCED)
+set(HIPSYCL_OPTION_STRING "Use hipsycl cmake integration (default: OFF)")
+hpx_option(HPX_WITH_HIPSYCL BOOL "${HIPSYCL_OPTION_STRING}" OFF ADVANCED)
```
hkaiser, line 554: Enabling `HIPSYCL` without enabling `SYCL` should produce a warning (or an error). Alternatively, it should enable SYCL (perhaps with a warning also).

libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
```diff
@@ -0,0 +1,182 @@
...
+        // To find the correct overload (or any at all actually) we need to add the
+        // code_location argument which is the last argument in every queue member
+        // function in the intel oneapi sycl implementation.  As far as I can tell
+        // it is usually invisible from the user-side since it is using a default
+        // argument (code_location::current())
+
+        /// sycl::queue::member_function type with code_location parameter
+        template <typename... Params>
+        using queue_function_ptr_t = cl::sycl::event (cl::sycl::queue::*)(
+            Params..., const cl::sycl::detail::code_location&);
```
hkaiser, line 71: Would you mind using 'east-const' notation, if possible? This would make the overall codebase more uniform.

libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
```diff
@@ -0,0 +1,182 @@
...
+        /// code_location argument added automatically.
+        template <typename... Params>
+        void post(queue_function_ptr_t<Params...>&& queue_member_function,
+            Params&&... args)
+        {
+            // for the intel version we need to actually pass the code location.
+            // Within the intel sycl api this is usually a default argument,
+            // but for invoke we need to pass it manually 
+            cl::sycl::event e =
+                std::invoke(std::forward<queue_function_ptr_t<Params...>>(
```
hkaiser, line 82: Would you mind using `HPX_INVOKE()` instead of `std::invoke()` (both are functionally fully compatible)? Our implementation is much less heavyweight in terms of compilation overheads. (please note, that we use `HPX_INVOKE` only inside the library, not for examples or tests).

libs/core/async_sycl/include/hpx/async_sycl/sycl_future.hpp
```diff
@@ -0,0 +1,100 @@
...
+            HPX_NON_COPYABLE(future_data);
+
+            using init_no_addref =
+                typename lcos::detail::future_data_allocator<void,
+                    Allocator>::init_no_addref;
+
+            using other_allocator = typename std::allocator_traits<
+                Allocator>::template rebind_alloc<future_data>;
+
+            future_data() {}
```
hkaiser, line 51:
```suggestion
            future_data() = default;
```

libs/core/async_sycl/tests/unit/sycl_vector_add_get_future.cpp
```diff
@@ -0,0 +1,205 @@
...
+// Uncomment the pragma message commands for more information about the compile passes 
+// (2 passes for hipsycl, 3 for dpcpp)
+#if defined(SYCL_LANGUAGE_VERSION)
+#if !defined(__INTEL_LLVM_COMPILER) && !defined(__HIPSYCL__)
+#warning "HPX-SYCL integration only tested with Intel oneapi and HipSYCL. \
+Utilized compiler appears to be neither of those!"
+#endif
+#else
+#error "Compiler does not seem to support SYCL! SYCL_LANGUAGE_VERSION is undefined!"
+#endif
```
hkaiser, line 37: Should this be part of the library (instead of being part of the test)? Users might not run the test and still use the code in their own applications.

libs/core/async_sycl/tests/unit/sycl_vector_add_executor.cpp
```diff
@@ -0,0 +1,181 @@
...
+    check_vector_add_results(a, b, add_parallel);
+    print_vector_results(a, b, add_parallel);
+
+    // Cleanup
+    std::cout << "Disabling SYCL future polling.\n";
+    hpx::sycl::experimental::detail::unregister_polling(hpx::resource::get_thread_pool(0));
+    return hpx::finalize();
+}
+
+int main(int argc, char* argv[])
```
hkaiser, line 177: Shouldn't there be an empty main() implementation in case `HPX_HAVE_SYCL` is not defined?

libs/core/async_sycl/tests/unit/common/sycl_vector_add_test_utils.hpp
```diff
@@ -0,0 +1,57 @@
+#include <vector>
```
hkaiser, line 1: Please add a copyright and license header to this file.

libs/core/async_sycl/src/sycl_future.cpp
```diff
@@ -0,0 +1,18 @@
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/allocator_support/internal_allocator.hpp>
+#include <hpx/assert.hpp>
+#include <hpx/async_sycl/sycl_future.hpp>
+#ifndef __SYCL_DEVICE_ONLY__
```
hkaiser, line 8:
```suggestion
#ifndef HPX_COMPUTE_DEVICE_CODE
```

G-071:
@hkaiser Thanks for the feedback!

It's good that you mentioned whether it works with ```hpx::async``` -- turns out it did not! So I took care of that first:

The tag_invoke for it was still missing. After adding that it worked but only when specifying the exact type of the sycl queue function which was rather ugly:
```cpp
auto my_sycl_kernel_to_submit = [&](cl::sycl::handler& h) {
                cl::sycl::accessor a(a_buf, h, cl::sycl::read_only);
                cl::sycl::accessor b(b_buf, h, cl::sycl::read_only);
                cl::sycl::accessor results(results_buf, h, cl::sycl::write_only, cl::sycl::no_init);
                h.parallel_for(num_items, [=](auto i) { /* insert numeric code here */ });
            };
auto fut = hpx::async<decltype(exec), my_queue_function_t<decltype(my_sycl_kernel_to_submit)>>(exec,
            &cl::sycl::queue::submit, my_sycl_kernel_to_submit);
```
Essentially, the issue here was that the compiler could not figure out the type of the sycl queue member function pointer ```&cl::sycl::queue::submit``` on its own as it depends on other templates [(most of the queue functions do)](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:interface.queue.class) so it needed some help!

To work around this, I added an overload to hpx::async (and one for hpx::apply) for the sycl executors. These specify the member function itself not as a template, but simply as a pointer depending on the type of the following arguments -- this seems to be enough for clang to figure out the rest! [Link to said overloads](https://github.com/G-071/hpx/blob/798db79f7b38c078e7d14c856f36385a9eafc86a/libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp#L228).

With all that done, we can now write this as expected:
```cpp
auto fut = hpx::async(exec,
            &cl::sycl::queue::submit, [&](cl::sycl::handler& h) {
                cl::sycl::accessor a(a_buf, h, cl::sycl::read_only);
                cl::sycl::accessor b(b_buf, h, cl::sycl::read_only);
                cl::sycl::accessor results(results_buf, h, cl::sycl::write_only, cl::sycl::no_init);
                h.parallel_for(num_items, [=](auto i) { /* insert numeric code here */ });
            });
```
I added two additional tests for this to the single executor vector_add file and changed the concurrent vector_add test  to use ```hpx::async``` and ```hpx::apply``` altogether.

Overall, I am now happy with this and will move on to the rest of the comments (though I am still open to other solutions besides having the async/apply overloads).

CMakeLists.txt
G-071: This is done in [HPX_SetupSYCL.cmake](https://github.com/G-071/hpx/blob/b4f1484d253fd514b66edc1a90441ca3537839ae/cmake/HPX_SetupSYCL.cmake#L19)

libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
```diff
@@ -0,0 +1,182 @@
...
+        /// Create a SYCL executor (based on a sycl queue)
+        explicit sycl_executor(cl::sycl::default_selector selector)
+          : command_queue(selector, cl::sycl::property::queue::in_order{})
+        {
+        }
+
+        // TODO Future work: Add more constructors as needed (in case we need different queue constructors)
+        // -------------------------------------------------------------------------
+        // Queue will be cleaned up by its own destructor
+        ~sycl_executor() {}
```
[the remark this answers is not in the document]
G-071: Added

libs/core/async_sycl/include/hpx/async_sycl/sycl_future.hpp
G-071: Added

libs/core/async_sycl/tests/unit/common/sycl_vector_add_test_utils.hpp
G-071: Added

libs/core/async_sycl/src/sycl_future.cpp
G-071: Added

libs/core/async_sycl/tests/unit/sycl_vector_add_get_future.cpp
G-071: Hmm, good point!

I briefly considered putting this into compiler_specific.hpp, however, even when building HPX for SYCL not necessarily all files are build with a SYCL compiler (for instance, with HIPSYCL we still build most things with clang++).

That being said, I think sycl_event_callback.hpp is a good place to put it. This file will be included by sycl_future.hpp and sycl_executor.hpp and is thus guaranteed to be included when using SYCL (with HPX) somewhere. I move the check to this header in c4df54708e5e4560e19b3ebdea390b45a9fe05de

libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
G-071: I have done so in e280fd14c3ab236ed640c3b58dac7236308c821d

I also changed std::forward to HPX_FORWARD while I was at it! Is the reasoning behind HPX_FORWARD the same? (better compiletime)

libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
G-071: Sure, not a problem! I have changed to east-const in 951382f48d56c1b90d460ed996e57fa3c670cbdd

libs/core/async_sycl/tests/unit/sycl_vector_add_executor.cpp
G-071: Hmm, good question. Currently, the tests are not being built when HPX_WITH_SYCL=OFF (and thus HPX_HAVE_SYCL is undefined) -- this makes the entire HPX_HAVE_SYCL ifdef in there superfluous. 

That being said, I put it in there in case someone just wanted to quickly look up how to use HPX with SYCL (and that one is good to know). Since I don't really want to remove it for that reason (at least until we have some more examples, but that's for a different PR), I went ahead and added a main implementation for when HPX_HAVE_SYCL is not defined to the unit tests (making it more complete as an example). This also serves the second purpose of checking that HPX_HAVE_SYCL got defined correctly (as those dummy mains return 1 and thus fail the test). Added in f4844b513b61903dc32b436413ebee0e0b551e59

libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
```diff
@@ -0,0 +1,182 @@
...
+            Params&&... args)
+        {
+            // for the intel version we need to actually pass the code location.
+            // Within the intel sycl api this is usually a default argument,
+            // but for invoke we need to pass it manually 
+            cl::sycl::event e =
+                std::invoke(std::forward<queue_function_ptr_t<Params...>>(
+                                queue_member_function),
+                    command_queue, std::forward<Params...>(args...),
+                    cl::sycl::detail::code_location::current());
```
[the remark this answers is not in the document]
G-071: The code location is not part of the SYCL specification itself, but something implementation-specific to OneAPI. From what I can tell, in their API, it is usually "hidden" from the user by the default parameter value ```cl::sycl::detail::code_location::current()```. We had to deal with it for the correct function resolution when passing queue functions to the executor, however, I *think* for normal SYCL usage it should not be necessary to know about (or use) this parameter. 

If someone really requires it, we can easily expose this with an extra parameter (and a default value to maintain compatibility with other SYCL implementations/runtimes) in a later PR. However, right now, I am not sure for which use-case this is needed. Hence, for now, I would simply suggest keeping it as is, until we encounter such a use-case.

G-071:
@hkaiser I think I addressed all of your feedback (I took the liberty of marking the minor points as resolved already)!

libs/core/async_sycl/tests/unit/sycl_vector_add_executor.cpp
hkaiser: > Hmm, good question. Currently, the tests are not being built when HPX_WITH_SYCL=OFF (and thus HPX_HAVE_SYCL is undefined) -- this makes the entire HPX_HAVE_SYCL ifdef in there superfluous.

Yes, that was what I was wondering about... Either always compile and run the test (in which case you need the empty `main()`) or remove the HPX_HAVE_SYCL ifdef.

libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
hkaiser: Yes, it's just all about compile time overheads. Thanks!

hkaiser:
> ```c++
> auto fut = hpx::async(exec,
>             &cl::sycl::queue::submit, [&](cl::sycl::handler& h) {
>                 cl::sycl::accessor a(a_buf, h, cl::sycl::read_only);
>                 cl::sycl::accessor b(b_buf, h, cl::sycl::read_only);
>                 cl::sycl::accessor results(results_buf, h, cl::sycl::write_only, cl::sycl::no_init);
>                 h.parallel_for(num_items, [=](auto i) { /* insert numeric code here */ });
>             });
> ```

Now I'm wondering whether this could be simplified further:
```
auto fut = hpx::async(exec, as_result(results_buf), std::ref(a_buf), std::ref(b_buf));
```

G-071:
Using SYCL USM, I should be able to add an example like (or at least close to this). I think that will be worth having as an example. I will add that to the PR this week when I find some time for! I still need to clean up the circleci-inspect warnings anyways.

hkaiser:
> SYCL USM

I wasn't thinking in terms of universal memory. I think that it is possible to rewrap things such to achieve what I suggested without adding more capabilities...

hkaiser:
@G-071 would you mind resolving the conflicts, please?

G-071:
@hkaiser Done! I just merged master into the branch!

I've also added some additional changes to this PR:
- It now works with the Intel LLVM instead of just the ubuntu-PPA-version of dpcpp (handy for having one single toolchain able to target Intel, NVIDIA and AMD GPUs).
- I added support and a test for USM memory. 
- Additionally, I added support for the SYCL queue shortcut methods, allowing us to call kernels like this (for example):
```cpp
auto fut = hpx::async(exec, &cl::sycl::queue::parallel_for, cl::sycl::range<1>{number_workitems}, kernel);
```



I was also able to get both HPX-Kokkos and Octo-Tiger itself working with SYCL (using the HPX-SYCL integration added  in this PR together with the Kokkos-SYCL execution space). See referenced hpx-kokkos/octotiger PRs for more details.

For the lack of a FP64-capable Intel GPU (my own iGPU only seems to supports FP64 via emulation which I haven't tried yet), I've tested Octo-Tiger on both a NVIDIA A100 and an AMD MI100 GPU using the SYCL execution space and it seems to work well. From my first impression on these GPUs, there is a slight performance price to pay, though it did not seem too bad. Though, my next step would be some more thorough performance tests with Octo-Tiger and SYCL to get more data.

libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
```diff
@@ -0,0 +1,324 @@
...
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// hpxinspect:noascii
+
+#pragma once
+
+#include <hpx/config.hpp>
+//#include <hpx/async_base/apply.hpp>
```
hkaiser, line 12: Please remove unneeded code

libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
```diff
@@ -0,0 +1,324 @@
...
+        }
+
+        /// Default Implementation without the extra intel code_location
+        /// parameter. Removes the reference for trivial types to make the
+        /// function matching easier (see sycl_stream.cpp test)
+        template <typename... Params>
+        using queue_function_ptr_t = cl::sycl::event (cl::sycl::queue::*)(
+            typename std::conditional<
+                std::is_trivial<std::remove_reference_t<Params>>::value,
+                std::decay_t<Params>, Params>::type...);
```
hkaiser, line 69:
```suggestion
            std::conditional_t<
                std::is_trivial_v<std::remove_reference_t<Params>>,
                std::decay_t<Params>, Params>...);
```

libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
```diff
@@ -0,0 +1,324 @@
...
+    hpx::sycl::experimental::sycl_executor>::value);
+// Use the same apply_dispatch than the normal apply otherwise
+return detail::post_dispatch<typename std::decay<Executor>::type>::call(
+    HPX_FORWARD(Executor, exec),
+    HPX_FORWARD(
+        hpx::sycl::experimental::sycl_executor::queue_function_code_loc_ptr_t<Ts>, f),
+    HPX_FORWARD(Ts, ts));
+}
+#endif
+}    // namespace hpx
```
hkaiser, line 324: Something is off with the formatting in this file, does it need processing with clang-format?

hkaiser:
@G-071 What's the status of this?

G-071:
I have added some additional modifications to this PR. It now offers two versions of the HPX-SYCL integration: The event-polling that I originally added and a second version using SYCL host tasks to set the future data. In the end, the event polling version performs a lot better, but I wanted to have the comparison data for both, so now we have also have both implemented!

I also re-tested everything with dpcpp/icpx and hipSYCL (now called Open SYCL). With dpcpp/icpx everything works, both the SYCL-related tests within HPX , as well as the Octo-Tiger tests.
With hipSYCL the event-polling integration works as well and passes the HPX tests. However, hipSYCL lacks support for the SYCL host tasks and the required oneapi SYCL extensions that we would need for the Kokkos SYCL execution space (and thus for Octo-Tiger), so I couldn't test that. Though support for SYCL host tasks is completely optional for this HPX-SYCL integration itself anyway, as long as one sticks with the event polling version (which I would also recommend for performance reasons). Hence, hipSYCL can still be used together with HPX as long as one does not need any of the oneapi SYCL extensions.

Other than that, I think I addressed the issues mentioned in January @hkaiser. Can you have another look at it? Please let me know if I missed anything. From my end, this looks good to be merged, though!

In the future, I would like to add asynchronous exception handling for the SYCL executor and some tests for that. However, I would put this in a future PR and rather get this one in for the 1.9.0 release.

hkaiser (dismissed):
Let's go ahead with this if the tests are passing. LGTM, Thanks!

hkaiser:
@G-071 one last thing - could you fix the inspect and cmake-format issues, please?

G-071:
> @G-071 one last thing - could you fix the inspect and cmake-format issues, please?

The last two commits resolved those issues! The remaining test failures ("tests.unit.segmented_algorithms") look like they are unrelated to this PR (I think).

G-071:
@hkaiser Is there anything else that should be fixed before merging this?

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6086, opened 2022-11-24
Deprecate `hpx::util::unlock_guard` in favor of `hpx::unlock_guard`

dimitraka:
- Deprecate `hpx::util::unlock_guard` in favor of `hpx::unlock_guard`
- Add corresponding documentation

### DIFF
--- a/libs/core/include_local/include/hpx/local/mutex.hpp
+++ b/libs/core/include_local/include/hpx/local/mutex.hpp
@@ -13,5 +13,5 @@
 #include <hpx/thread_support/unlock_guard.hpp>
 
 namespace hpx {
-    using hpx::util::unlock_guard;
+    using hpx::unlock_guard;
 }    // namespace hpx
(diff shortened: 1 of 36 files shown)


libs/core/include_local/include/hpx/local/mutex.hpp
```diff
@@ -13,5 +13,5 @@
 #include <hpx/thread_support/unlock_guard.hpp>
 
 namespace hpx {
-    using hpx::util::unlock_guard;
+    using hpx::unlock_guard;
 }    // namespace hpx
```
hkaiser, line 17: This can go away.

dimitraka: Done!

hkaiser: That's not what I meant. You now already have `hpx::unlock_guard` defined, there is no need to reimport it into the HPX namespace again. You can remove the whole section (L15-17).

dimitraka: yes, this is what I have done, please check here https://github.com/STEllAR-GROUP/hpx/pull/6086/commits/fc73f749c0a949e67e780dd3b9c003a7a18292e7 because this is the outdated version of the file

hkaiser: Yah, sorry, I was looking at an older commit. All is well.

PR #6089, opened 2022-11-29
Implementing p2079

### LINKED ISSUE #5753: Implement p2079 on top of our senders receivers
As we are in the midst of implementing [p2300](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2300r4.html) and new means of asynchronous execution (senders/receivers) with #5045, [p2079](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2079r2.html) adds on top of it by presenting facilities that default the access to system thread pool. The main puprose is for the senders receivers to carry a default execution context and scheduler.

The two facilites to be implemented are:

- `system_context`
- `system_scheduler`

and need to bind properly with our existing senders/receivers implementations.

hkaiser:
- flyby: fixing minor issues with run_loop test

Fixes #5753

### DIFF
--- a/libs/core/execution/CMakeLists.txt
+++ b/libs/core/execution/CMakeLists.txt
@@ -160,12 +160,13 @@ add_hpx_module(
     hpx_async_base
     hpx_async_combinators
     hpx_config
-    hpx_threading
-    hpx_pack_traversal
     hpx_errors
+    hpx_iterator_support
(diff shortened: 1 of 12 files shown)


hkaiser:
I'm putting this on ice for the time being. From what I see, p2079 is not implementable the way it's specified.

Issue #5497, opened 2021-08-05
Start using C++17 features unconditionally


msimberg:
Now that we require C++17 we can start using C++17 features more liberally without having to worry about fallbacks for older standards. Below is a grab bag of things we could start changing (please feel free to add to the list, or even better simply make the change if there's something missing). Keep in mind that we don't need to change every instance possible in one go, so don't be afraid to just make a local change while you're working on some piece of code (as we do right now with e.g. `typedef` to `using`). Small steps are better than no steps.

- [ ] Replace the `HPX_INLINE_CONSTEXPR_VARIABLE` macro with `inline constexpr`
- [ ] Add class template argument deduction guides
- [ ] Use `if constexpr` for dispatching where appropriate (e.g. `std::true/false_type` overloads: https://github.com/STEllAR-GROUP/hpx/blob/03309bea58bcc18aff2864900c77e31116158b1b/libs/parallelism/futures/include/hpx/futures/future.hpp#L51-L84)
- [ ] Fold expressions
- [ ] Replace various type traits with `std` counterparts (e.g. `always_void` -> `void_t`)
- [ ] Replace various utilities with `std` counterparts (e.g. `hpx::invoke` -> `std::invoke`? @K-ballo I suspect you know best which ones are actually reasonable to change and for which ones we actually would want to keep our own implementation)
- [ ] Nested namespace definitions (e.g. `namespace hpx::execution::experimental { }`)
- [ ] Replace various attributes hidden behind macros directly with the attributes themselves (e.g. `[[nodiscard]]`)
- [ ] Remove empty messages from `static_assert`s that don't have a meaningful message
- [ ] Use `_t` and `_v` variants of various type traits where appropriate

hkaiser:
What about the `hpx::apply` <--> `std::apply` semantic mismatch? Should we start deprecating `hpx::apply` for some other name (`hpx::fire_n_forget`/`hpx::post`?) so we can substitute our `hpx::util::invoke_fused` with `std::apply` without causing undue confusion?

K-ballo:
> * Replace various utilities with `std` counterparts (e.g. hpx::invoke -> std::invoke)

`std::invoke` implementations are considerably heavier at build time. It makes sense to replace `hpx::invoke` in examples and other user facing bits. In implementation code `HPX_INVOKE` should be used instead, to additionally reduce the stack noise in debug builds.

msimberg:
[...]

PR #6091, opened 2022-12-01
Replace artificial sequencing with fold expressions

hkaiser:
- flyby: fix left-over problems from namespace change

working towards resolving #5497


hkaiser:
bors merge

PR #6099, opened 2022-12-07
Futures attempt to execute threads directly if those have not started executing

hkaiser:
- adding new API function hpx::threads::get_outer_self_id

### DIFF
--- a/libs/core/futures/src/future_data.cpp
+++ b/libs/core/futures/src/future_data.cpp
@@ -122,7 +122,7 @@ namespace hpx::lcos::detail {
             return false;
         }
 
-        auto state = this->state_.load(std::memory_order_acquire);
+        auto const state = this->state_.load(std::memory_order_acquire);
         if (state != this->empty)
         {
             return false;
--- a/libs/core/futures/src/future_data.cpp
+++ b/libs/core/futures/src/future_data.cpp
@@ -117,7 +117,8 @@ namespace hpx::lcos::detail {
     bool future_data_base<traits::detail::future_data_void>::execute_thread()
     {
         // we try to directly execute the thread exactly once
-        if (!runs_child_)
+        threads::thread_id_ref_type runs_child = runs_child_;
+        if (!runs_child)
         {
             return false;
         }
@@ -129,7 +130,6 @@ namespace hpx::lcos::detail {
         }
 
         // this thread would block on the future
-        threads::thread_id_ref_type runs_child = runs_child_;
 
         auto* thrd = get_thread_id_data(runs_child);
         HPX_UNUSED(thrd);    // might be unused
--- a/libs/core/futures/src/future_data.cpp
+++ b/libs/core/futures/src/future_data.cpp
@@ -8,13 +8,16 @@
 #include <hpx/futures/future.hpp>
 
 #include <hpx/config.hpp>
+#include <hpx/assert.hpp>
 #include <hpx/async_base/launch_policy.hpp>
 #include <hpx/errors/try_catch_exception_ptr.hpp>
 #include <hpx/execution_base/this_thread.hpp>
 #include <hpx/functional/deferred_call.hpp>
 #include <hpx/functional/move_only_function.hpp>
+#include <hpx/futures/detail/execute_thread.hpp>
 #include <hpx/futures/futures_factory.hpp>
 #include <hpx/modules/errors.hpp>
+#include <hpx/modules/logging.hpp>
 #include <hpx/modules/memory.hpp>
 
 #include <cstddef>
@@ -95,8 +98,68 @@ namespace hpx::lcos::detail {
     }
 
     ///////////////////////////////////////////////////////////////////////////
-    future_data_base<traits::detail::future_data_void>::~future_data_base() =
-        default;
+    future_data_base<traits::detail::future_data_void>::~future_data_base()
+    {
+        if (runs_child_ != threads::invalid_thread_id)
+        {
+            auto* thrd = get_thread_id_data(runs_child_);
+            (void) thrd;
+            LTM_(debug).format(
+                "task_object::~task_object({}), description({}): "
+                "destroy runs_as_child thread",
+                thrd, thrd->get_description(), thrd->get_thread_phase());
+
+            runs_child_ = threads::invalid_thread_id;
+        }
+    }
+
+    // try to performed scoped execution of the associated thread (if any)
+    bool future_data_base<traits::detail::future_data_void>::execute_thread()
+    {
+        // we try to directly execute the thread exactly once
+        if (!runs_child_)
+        {
+            return false;
+        }
+
+        auto state = this->state_.load(std::memory_order_acquire);
+        if (state != this->empty)
+        {
+            return false;
+        }
+
+        // this thread would block on the future
+        threads::thread_id_ref_type runs_child = runs_child_;
+
+        auto* thrd = get_thread_id_data(runs_child);
+        HPX_UNUSED(thrd);    // might be unused
+
+        LTM_(debug).format("task_object::get_result_void: attempting to "
+                           "directly execute child({}), description({})",
+            thrd, thrd->get_description());
+
+        if (threads::detail::execute_thread(HPX_MOVE(runs_child)))
+        {
+            // don't try running this twice
+            runs_child_.reset();
+
+            LTM_(debug).format("task_object::get_result_void: successfully "
+                               "directly executed child({}), description({})",
+                thrd, thrd->get_description());
+
+            // thread terminated, mark as being destroyed
+            HPX_ASSERT(thrd->get_state().state() ==
+                threads::thread_schedule_state::deleted);
+
+            return true;
+        }
+
+        LTM_(debug).format("task_object::get_result_void: failed to "
+                           "directly execute child({}), description({})",
+            thrd, thrd->get_description());
+
+        return false;
+    }
 
     static util::unused_type unused_;
 
@@ -111,8 +174,8 @@ namespace hpx::lcos::detail {
             return nullptr;
         }
 
-        // No locking is required. Once a future has been made ready, which
-        // is a postcondition of wait, either:
+        // No locking is required. Once a future has been made ready, which is a
+        // postcondition of wait, either:
         //
         // - there is only one writer (future), or
         // - there are multiple readers only (shared_future, lock hurts
@@ -139,16 +202,15 @@ namespace hpx::lcos::detail {
             return nullptr;
         }
 
-        // the thread has been re-activated by one of the actions
-        // supported by this promise (see promise::set_event
-        // and promise::set_exception).
+        // the thread has been re-activated by one of the actions supported by
+        // this promise (see promise::set_event and promise::set_exception).
         if (s == exception)
         {
             auto const* exception_ptr =
                 static_cast<std::exception_ptr const*>(storage);
 
-            // an error has been reported in the meantime, throw or set
-            // the error code
+            // an error has been reported in the meantime, throw or set the
+            // error code
             if (&ec == &throws)
             {
                 std::rethrow_exception(*exception_ptr);
@@ -171,8 +233,8 @@ namespace hpx::lcos::detail {
                 on_completed();
             },
             [&](std::exception_ptr ep) {
-                // If the completion handler throws an exception, there's nothing
-                // we can do, report the exception and terminate.
+                // If the completion handler throws an exception, there's
+                // nothing we can do, report the exception and terminate.
                 if (run_on_completed_error_handler)
                 {
                     run_on_completed_error_handler(HPX_MOVE(ep));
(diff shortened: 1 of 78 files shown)


Pansysk75:
The latest commit addresses the following race contition on the runs_child_ variable:

Here is `future_data::set_value()`. The modification (resetting) of `runs_child_` happens under a lock. This is alright:
https://github.com/STEllAR-GROUP/hpx/blob/0ad30ef89b666f557de3a91902e745a08a3abe6f/libs/core/futures/include/hpx/futures/detail/future_data.hpp#L559-L561

However, there was a race condition which happened when a scheduled thread was executed very rapidly. Here's the code that schedules a thread:
https://github.com/STEllAR-GROUP/hpx/blob/3236b2fd7c82e56917c0e37d1ecaa92a3097f6fd/libs/core/futures/include/hpx/futures/packaged_continuation.hpp#L233-L237

`this->runs_child_ = spawner(...)` is supposed to keep a reference of the scheduled thread.
When the scheduled thread was executed very quickly, `this->runs_child_ = spawner(...)` could run at the same time (or after) `future_data::set_value()`.
Thus, there was a race condition between `runs_child_.reset()` and `runs_child_ = spawner(...)` . 

Here's the fix I applied:

https://github.com/STEllAR-GROUP/hpx/blob/0ad30ef89b666f557de3a91902e745a08a3abe6f/libs/core/futures/include/hpx/futures/packaged_continuation.hpp#L239-L243

Firstly, I added a lock to prevent concurrent modification of `runs_child_`.
I also added a check for future_data::is_ready(), ensuring we don't keep an unneeded reference to the scheduled thread (in case it has been already executed).

Let's hope this makes things better

Pansysk75:
I also noticed that the code below does not check whether there is a `thread_execution_hint::run_as_child` hint.
 https://github.com/STEllAR-GROUP/hpx/blob/3236b2fd7c82e56917c0e37d1ecaa92a3097f6fd/libs/core/futures/include/hpx/futures/packaged_continuation.hpp#L233-L237
I'm not sure if that was intentional, but my instinct is that we would also want to enable/disable it here depending on the hint.

hkaiser:
> I also noticed that the code below does not check whether there is a `thread_execution_hint::run_as_child` hint.
> 
> https://github.com/STEllAR-GROUP/hpx/blob/3236b2fd7c82e56917c0e37d1ecaa92a3097f6fd/libs/core/futures/include/hpx/futures/packaged_continuation.hpp#L233-L237
> 
> 
> I'm not sure if that was intentional, but my instinct is that we would also want to enable/disable it here depending on the hint.

I'd need to look closer, but I believe this is only called after the check has happened before in other places.

Pansysk75:
> 
> I'd need to look closer, but I believe this is only called after the check has happened before in other places.

I did my learning/testing on the cross_pool_injection example. The `.then` would lead to those lines executing, even with no hint placed anywhere.

hkaiser:
> Firstly, I added a lock to prevent concurrent modification of `runs_child_`.
> I also added a check for future_data::is_ready(), ensuring we don't keep an unneeded reference to the scheduled thread (in case it has been already executed).

@Pansysk75 please check my commit. I think I was able to remove this lock.

hkaiser:
> > > I'd need to look closer, but I believe this is only called after the check has happened before in other places.
> > 
> > 
> > I did my learning/testing on the cross_pool_injection example. The `.then` would lead to those lines executing, even with no hint placed anywhere.
> > Also now that I think about it, the lock could be avoided if the thread was first "registered" (ie created in a suspended state so that we get a thread_id) and then executed/resumed separately. I think I saw that someplace in this PR
> 
> In that case we will need to add the check here as well.

I'm not quite sure what we should do here. If we left this as is we would loose the direct execution optimization in the worst case...

hkaiser:
@Pansysk75 I enabled the new feature on gcc10, clang11, and clang12 and now we see hangs: `tests.unit.modules.resource_partitioner.suspend_thread` and `tests.unit.modules.resource_partitioner.suspend_thread_external` hang :/

Pansysk75:
> @Pansysk75 I enabled the new feature on gcc10, clang11, and clang12 and now we see hangs: `tests.unit.modules.resource_partitioner.suspend_thread` and `tests.unit.modules.resource_partitioner.suspend_thread_external` hang :/

I'll have a look

hkaiser:
@Pansysk75 FWIW, my last commit has improved things, but not entirely fixed the hangs.

libs/core/futures/src/future_data.cpp
```diff
@@ -131,6 +131,11 @@ namespace hpx::lcos::detail {
         // this thread would block on the future
         threads::thread_id_ref_type runs_child = runs_child_;
 
+        if (!runs_child)
+        {
+            return false;
+        }
+
```
hkaiser, line 138: Good catch, that also mean we can remove these lines: https://github.com/STEllAR-GROUP/hpx/blob/3a5a47a0a5fcc853971742bf853b92eb16239ded/libs/core/futures/src/future_data.cpp#L120-L123

Pansysk75: Sure, just left them in the case that taking a thread_id_ref_type was expensive. Give a thumbs up if you think it's ok to change it

hkaiser: I have looked again and I have convinced myself that there is no need to protect the `runs_child_` intrusive pointer needs to be protected in any way. The issue has to be somewhere else... we need to keep looking.

Pansysk75: Ok, I didn't understand exactly where the issue was supposed to occur, so I'll just take your word for it.

Currently I can barely reproduce the current assert failure (where a tid from terminated_threads_ does not appear in the thread_map_, when it should), and the issue goes away when logging... I'll give it another go, might find a more reliable failure case

hkaiser: I think this assert is wrong to begin with, I'm looking at it as we speak.

hkaiser:
@Pansysk75 the next error (segfault) happens in `suspend_pool_external`, however I'm not able to reproduce it locally. Could you try that, please?

Pansysk75:
> @Pansysk75 the next error (segfault) happens in `suspend_pool_external`, however I'm not able to reproduce it locally. Could you try that, please?

Copying the exact setup on Rostam did it for me.
That means doing
`source src/.jenkins/lsu/env-gcc-10.sh`
building with 
`-DHPX_WITH_PARALLEL_TESTS_BIND_NONE=ON -DCMAKE_BUILD_TYPE=Debug -DHPX_WITH_CHECK_MODULE_DEPENDENCIES=ON -DHPX_WITH_PARCELPORT_COUNTERS=ON -DLCI_DEBUG=ON -DHPX_WITH_CXX_STANDARD=17 -DHPX_WITH_MALLOC=system -DHPX_WITH_FETCH_ASIO=ON -DHPX_WITH_ASIO_TAG=asio-1-12-0 -DHPX_WITH_COMPILER_WARNINGS=ON -DHPX_WITH_COMPILER_WARNINGS_AS_ERRORS=ON -DHPX_WITH_PARCELPORT_MPI=ON -DHPX_WITH_PARCELPORT_LCI=ON -DHPX_WITH_FETCH_LCI=ON -DHPX_COROUTINES_WITH_THREAD_SCHEDULE_HINT_RUNS_AS_CHILD=ON`
and executing
`bin/suspend_thread_external_test --hpx:ini=hpx.stacks.use_guard_pages=0 --hpx:bind=none`


Now, the segfault stack trace points at the catch part of the following try-catch block:
https://github.com/STEllAR-GROUP/hpx/blob/09eabc71da8f2a150b2e26efe3d52d4163e49072/libs/core/thread_pools/include/hpx/thread_pools/scheduled_thread_pool_impl.hpp#L314-L362

It seems like `std::thread()` (inside of`add_processing_unit_internal()` call,  L338) occasionally throws a "Resource temporarily unavailable".

I guess we are not handling that properly, as stop_locked(l) (L358) leads to Scheduler::get_thread_count() being called, and it trying to read uninitialized data.

That's all for now, hope it helps!

hkaiser:
@Pansysk75 thanks a lot for this analysis. Nw the next question is why does it throw that exception to begin with? There has to be a reason.

Pansysk75:
> @Pansysk75 thanks a lot for this analysis. Now the next question is why does it throw that exception to begin with? There has to be a reason.

I wrote a script that monitors the amount of memory mappings (using the pmap utility).
Looks like std::thread fails when we have more mappings than specified by `/proc/sys/vm/max_map_count` (currently 65530 on Rostam).

Increasing  `/proc/sys/vm/max_map_count` or decreasing the amount of allocations would allow the tests to complete successfully. The failing tests spawn an absurd amount of futures, so it seems sensible to just limit the amount of futures spawned.

hkaiser:
> > @Pansysk75 thanks a lot for this analysis. Now the next question is why does it throw that exception to begin with? There has to be a reason.
> 
> I wrote a script that monitors the amount of memory mappings (using the pmap utility). Looks like std::thread fails when we have more mappings than specified by `/proc/sys/vm/max_map_count` (currently 65530 on Rostam).
> 
> Increasing `/proc/sys/vm/max_map_count` or decreasing the amount of allocations would allow the tests to complete successfully. The failing tests spawn an absurd amount of futures, so it seems sensible to just limit the amount of futures spawned.
> 
> (Also `--hpx:ini=hpx.stacks.use_guard_pages=0` results in the accumulation of a higher number of (but smaller in individual size) mappings. But I do not know enough of what's happening to interpret that observation.)

Thanks! I have at least managed to avoid the segfault now and properly report the exception. ;-)

Pansysk75:
@hkaiser Here's what I see happening:

Work started without the hint, gets scheduled through `register_work`, which does not directly create a thread object, but only adds the thread descripition to Scheduler.new_data_. That keeps the amount of allocations low, because the allocated stack memory can be reused. 

Work started with the runs_as_child hint gets handled through `register_thread`, which leads to Scheduler::create_thread always creating the thread object (and allocating the memory for it). So if a bulk of tasks (ie through hpx::async) are created, we immediately do as many stack allocations as there are threads created, and there is no chance for memory recycling to happen

So, the exessive memory usage is not a bug, rather it's a design issue. I remember you telling me about how you wanted to delay the thread allocation, now I get why that could be a good idea :)

hkaiser:
> @hkaiser Here's what I see happening:
> 
> Work started without the runs_as_child hint, gets scheduled through `register_work`, which does not directly create a thread object, but only adds the thread descripition to Scheduler.new_data_. The actual thread will be allocated/created when the Scheduler starts running out of threads in the queues. That keeps the amount of allocations low, because, since the threads are created on an as-needed basis, the allocated thread-stack memory can be reused.
> 
> Work started with the runs_as_child hint gets handled through `register_thread`, which leads to Scheduler::create_thread always creating the thread object (and allocating the memory for it). So if a bulk of tasks are created (ie through a ton of calls of hpx::async), we immediately do as many stack allocations as there are threads created, and there is no chance for memory recycling to happen
> 
> So, the exessive memory usage is not a bug, rather it's a design issue. I remember you telling me about how you wanted to delay the thread allocation, now I get why that could be a good idea :)

Yes, that's it! Very nice analysis! The creation of a thread with the hint uses `register_thread` instead of `register_work` as we need the thread id. A possible solution would be (as you said) to recycle the stacks separately from the thread objects, possibly we can even avoid recycling the thread objects themselves, just the stacks. Another solution could be to delay storing the `thread_id` until the thread is actually created.

Pansysk75:
Although disabling run_as_child decreases allocations by a lot, here is still a memory-leak-like issue arising with hpx:ini=hpx.stacks.use_guard_pages=0. I'm investigating it rn.

hkaiser:
That would explain why the suspend_runtime et.al. now fail even for builders that don't enable the run_as_child hint by default.

Pansysk75:
> That would explain why the suspend_runtime et.al. now fail even for builders that don't enable the run_as_child hint by default.

Here, does mmap-ing memory of (size + EXEC_PAGESIZE)  make sense even if we are not using guard pages?

https://github.com/STEllAR-GROUP/hpx/blob/dd17010adf02cdab681d96f10f96626d476459c0/libs/core/coroutines/include/hpx/coroutines/detail/posix_utility.hpp#L84-L96

Deallocation seems to not take EXEC_PAGESIZE into account if we don't use guard pages:

https://github.com/STEllAR-GROUP/hpx/blob/dd17010adf02cdab681d96f10f96626d476459c0/libs/core/coroutines/include/hpx/coroutines/detail/posix_utility.hpp#L156-L172

hkaiser:
Excellent catch! Could you patch that, please?

Pansysk75:
> Excellent catch! Could you patch that, please?

Thank you, I'll do that :)

hkaiser:
I'm also preparing a change right now that starts deleting stack segments of recycled threads once more then a certain number of threads have been recycled. That should keep the amount of memory allocated down a bit as well.

hkaiser:
@Pansysk75 your fix did the trick! I will roll back my change that limits the numer of recycled stacks as it made the performance tests fail. Let's see if after that change things will be still green.

Pansysk75:
> @Pansysk75 your fix did the trick! I will roll back my change that limits the numer of recycled stacks as it made the performance tests fail. Let's see if after that change things will be still green.

Yes, things are much better.

The memory mappings now barely go above a few hundred, as the OS is able to merge the mapped areas more efficiently. Even with guard_pages=on, they are in the hundreds now (an order of magnitude less than before).
That seems very significant, so keep an eye out for any macroscopic effects it may have.

The only memory "concern" is the slightly higher memory usage when using direct execution (because of the immediate thread spawining), but it's not that significant. So the thread stack allocation / recycling change is not as crucial.

hkaiser:
@Pansysk75 I think this is finally good to be merged. What do you think?

hkaiser:
> @hkaiser There is a single timed-out test on daint (I had a look yesterday but didn't get anywhere), but other than that, nothing further to add.

I believe this is unrelated. Could be another case of the synchronization object going out of scope too early.

hkaiser:
bors merge

PR #6103, opened 2022-12-09
Document `task_group` and include in Public API

dimitraka:
- Add header file `task_group.hpp` to the Public API page
- Add documentation for `task_group`

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/task_group.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/task_group.hpp
@@ -30,9 +30,12 @@
 #include <type_traits>
 #include <utility>
 
+/// Top-level namespace
 namespace hpx { namespace execution { namespace experimental {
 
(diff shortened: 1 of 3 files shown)


libs/core/algorithms/include/hpx/parallel/task_group.hpp
```diff
@@ -55,7 +58,7 @@ namespace hpx { namespace execution { namespace experimental {
         };
 
     public:
-        // Spawns a task to compute f() and returns immediately.
+        /// \brief Adds a task to compute \c f() and returns immediately.
```
hkaiser, line 61: Could you add documentation for the parameters as well, please?

dimitraka: Done!

hkaiser:
bors merge

PR #6106, opened 2022-12-16
Modernizing modules of levels 0 to 5

hkaiser:
- flyby: HPX_FORWARD/HPX_MOVE now expand to std::forward and std::move if those are implemented as builtin functions

working towards #5497


hkaiser:
bors merge

PR #6116, opened 2022-12-24
Add new option loopback_network

JiakunYan:
Option `--hpx:loopback_network=0|1`. If set to 1, the network (at least the initialization phase) will be enabled even if the locality number is 1.

It is useful for debugging networks.


libs/core/runtime_configuration/src/runtime_configuration.cpp
```diff
@@ -160,6 +160,7 @@ namespace hpx { namespace util {
             "pu_step = 1",
             "pu_offset = 0",
             "numa_sensitive = 0",
+            "loopback_network = 0"
```
hkaiser, line 163: This is missing a comma at the end of the line.

hkaiser (commented):
Does this add a command line option?

JiakunYan:
> Does this add a command line option?

Yes.

libs/core/runtime_configuration/src/runtime_configuration.cpp
JiakunYan: Fixed.

hkaiser:
bors merge

PR #6120, opened 2022-12-28
Use index_queue for parallel executors bulk_async_execute

hkaiser:
This should significantly improve the performance of many parallel algorithms


hkaiser:
bors merge

PR #6123, opened 2023-01-05
Fixed issues found by PVS-Studio

hkaiser:
- added a lot of suppression comments that prevent PVS-warnings from being generated


hkaiser:
bors merge

PR #6130, opened 2023-01-09
Remove the mutex lock in the critical path of get_partitioner.

JiakunYan:
Remove the mutex lock in the critical path of hpx::resource::detail::get_partitioner.

The protected variable `partitioner_ref` is only set once during initialization.


hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6134, opened 2023-01-11
Adding notification function for parcelports to be called after early parcel handling

hkaiser:
Parcelports now can override a new function `void initialized()` that will be called after early parcel handling is finished and before the thread pools are operational (i.e. before background work starts).

@JiakunYan please let me know if this is what you requested.

### DIFF
--- a/libs/full/parcelset/src/parcelhandler.cpp
+++ b/libs/full/parcelset/src/parcelhandler.cpp
@@ -59,9 +59,10 @@
 #endif
 
 ///////////////////////////////////////////////////////////////////////////////
-namespace hpx { namespace detail {
+namespace hpx::detail {
+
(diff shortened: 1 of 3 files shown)


JiakunYan:
Yes, it is what I want. Thank you!

JiakunYan:
Could you merge this PR? After that, I can rebase #6057.

hkaiser:
@JiakunYan this change still causes many (random) tests to hang. I'd like to first figure out what's wrong.

JiakunYan:
@hkaiser I see. This is weird.

hkaiser:
> @hkaiser I see. This is weird.

I have convinced myself that the errors are being caused by unrelated issues introduced by earlier commits. I think we can merge this now.

hkaiser:
bors merge

PR #6135, opened 2023-01-12
Fixing warnings reported by MSVC analysis

hkaiser:
- adding MSVC specific #pragma's to suppress the benign warnings


hkaiser:
bors merge

PR #6140, opened 2023-01-14
Fixing background scheduler to properly exit in the end

hkaiser:
- flyby: add API to add background work to any thread pool


hkaiser:
the test failures are caused by hangs in the LCI backend

hkaiser:
bors merge

PR #6147, opened 2023-01-20
Make posix co-routine stacks non-executable

jtsylve:
This reduces the exploitation surface area and resolves an issue with macOS hardened runtime that does not allow executable memory without an entitlement.


gonidelis:
Nice! Could you explain how this does increase safety?

jtsylve:
> Could you please explain how this does increase safety?

Allocating memory without the "execute" protection, allows the operating system to set the "NX" bit in the page tables that map the virtual address.  This allows for hardware-enforced guarantees that anything copied to that memory can not be executed as code.  This reduces the attack surface for several exploitation techniques. It's a common practice in secure programming to only allocate executable memory when absolutely necessary.

hkaiser (approved):
@jtsylve thanks a lot. I fixed the clang-format issue. Otherwise, LGTM!

hkaiser:
bors merge

hkaiser:

PR #6148, opened 2023-01-22
Investigate the failure of the LCI parcelport.

JiakunYan:


JiakunYan:
@hkaiser I think I have fixed the LCI parcelport.

hkaiser:
@JiakunYan very nice, thanks! What was the issue?

JiakunYan:
@hkaiser A problem on the LCI side. A dumb bug. When you asked me to fix the compilation error "non-void function doesn't have return value", I fixed it in the wrong way. It leads to some function calls being skipped when compilation. The reason it took me so long to reproduce the bug is it only went wrong when LCI_DEBUG=OFF.

hkaiser:
> @hkaiser A problem on the LCI side. A dumb bug. When you asked me to fix the compilation error "non-void function doesn't have return value", I fixed it in the wrong way. It leads to some function calls being skipped when compilation. The reason it took me so long to reproduce the bug is it only went wrong when LCI_DEBUG=OFF.

I'm glad you caught it!

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6152, opened 2023-01-23
The -rd and -mr options didn't work, and they should have been --rd and --mr

stevenrbrandt:
The arg parsing logic was broken in a number of ways. For example, if any option was provided after -rd (RelWithDebInfo), it automatically reverted to -r (Release). The same for -g, etc. In addition "-rd" is non-standard for a multi-letter unix arg. It was changed to --rd.


hkaiser:
bors merge

PR #6154, opened 2023-01-25
Avoid performing late command line handling twice in distributed runtime

hkaiser:
@weilewei this should fix the issue you were seeing (`--hpx:print-bind` being invoked twice)


hkaiser:
bors merge

PR #6168, opened 2023-02-09
Remove the mutex lock in parcelhandler::invoke_write_handler

JiakunYan:
The mutex lock is very expensive as the program needs to acquire it whenever a parcel is sent.

It appears the original mutex lock is used to protect the handler invocation from hpx::set_parcel_write_handler, but I think the code would be ill-formed anyway if the user wants to set the handler when there are parcels on the fly.


hkaiser (approved):
Ok, fair points. Let's watch the space if anything breaks. LGTM, thanks!

hkaiser:
bors merge

PR #6169, opened 2023-02-09
LCI parcelport: bypass the parcel queue and connection cache.

JiakunYan:
Originally, when hpx tries to send a parcel, it will (1) push the parcel into a parcel queue corresponding to its destination (2) try to get a "connection" of the destination from the connection cache (3) dequeue multiple parcels from the parcel queue (4) encode the parcels into a message and pass it to the parcelport layer. This process contains multiple attempts of acquiring mutex locks.

This PR optimizes the LCI parcelport so that it could bypass the parcel queue and connection cache. The parcelport layer offers a "send" function. Whenever the upper layer wants to send a parcel, it can directly invoke this function.

This PR shares the same idea as the "send_immediate" feature of the libfabric parcelport (#2379).

### DIFF
--- a/libs/full/parcelset/include/hpx/parcelset/parcelport_impl.hpp
+++ b/libs/full/parcelset/include/hpx/parcelset/parcelport_impl.hpp
@@ -488,75 +488,151 @@ namespace hpx::parcelset {
         }
 
     protected:
+        void send_immediate_impl_connectionless(locality const& dest_,
+            write_handler_type* fs, parcel* ps, std::size_t num_parcels)
+        {
+            static_assert(connection_handler_traits<
+                              ConnectionHandler>::is_connectionless::value &&
+                    connection_handler_traits<
+                        ConnectionHandler>::send_immediate_parcels::value,
+                "This parcelport is not suitable for the connectionless "
+                "version of send_immediate!");
+            std::size_t encoded_parcels = 0;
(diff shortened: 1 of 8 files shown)


libs/full/parcelset/include/hpx/parcelset/parcelport_impl.hpp
```diff
@@ -492,7 +492,65 @@ namespace hpx::parcelset {
...
+                {
+                    std::vector<parcel> overflow_parcels(
+                        std::make_move_iterator(ps + encoded_parcels),
+                        std::make_move_iterator(ps + num_parcels));
+                    std::vector<write_handler_type> overflow_handlers(
+                        std::make_move_iterator(fs + encoded_parcels),
+                        std::make_move_iterator(fs + num_parcels));
+                    enqueue_parcels(dest_, HPX_MOVE(overflow_parcels),
+                        HPX_MOVE(overflow_handlers));
+                }
```
hkaiser, line 550: Would you mind extracting the above into a separate function?

JiakunYan: Done!

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6173, opened 2023-02-12
C++ modules

Johan511:


hkaiser:
I think I have found a solution to this problem that does not require extensive code changes. Stay tuned.

PR #6195, opened 2023-03-13
Change the default tag of autofetch LCI to v1.7.3.

JiakunYan:
LCI v1.7.3 provides more ways of performance profiling and tuning. It also has some performance improvement.

[v1.7.3 change log](https://github.com/uiuc-hpc/LC/blob/dev-v1.7/doc/changelog_1_7_3.md)


hkaiser:
bors merge

PR #6199, opened 2023-03-16
added support for unseq, par_unseq for hpx::make_heap algorithm

Johan511:
Added unseq and par_unseq to hpx::make_heap algorithm

par_unseq => existing function calls sift_down_range in parallel. Overloaded sift_down_range function to call appropriate function based on value of hpx::is_unsequenced_execution_policy<ExPolicy>

unseq => current sequential implementation calls sequential_make_heap(), overloaded it to call appropriate function for seq vs unseq.

Should we consider using sift_down_range for sequential implementation too?


libs/core/algorithms/include/hpx/parallel/algorithms/make_heap.hpp
```diff
@@ -293,6 +293,33 @@ namespace hpx::parallel {
             }
         }
```
hkaiser, line 294: Is the original implementation of `sift_down_range()` still needed?

libs/core/algorithms/include/hpx/parallel/algorithms/make_heap.hpp
```diff
@@ -293,6 +293,33 @@ namespace hpx::parallel {
...
+            HPX_IVDEP HPX_UNROLL HPX_VECTORIZE
+            for (std::size_t i = 0; i != count; ++i)
+            {
+                sift_down(first, comp, proj, len, start - i);
+            }
+            // clang-format on
+        }
+
         template <typename Iter, typename Sent, typename Comp, typename Proj>
         constexpr Iter sequential_make_heap(
```
hkaiser, line 324: Same here, is this still needed?

hkaiser (changes requested):
Have you done some performance analysis to make sure the changes have the desired effect?

libs/core/algorithms/include/hpx/parallel/algorithms/make_heap.hpp
Johan511: I don't think so, the only place where sift_down_range() is called has been overloaded. We can delete the original implementation.

libs/core/algorithms/include/hpx/parallel/algorithms/make_heap.hpp
Johan511: I don't think so, the only place where sequential_make_heap() is called has been overloaded. We can delete the original implementation.

Johan511:
Haven't done any performance analysis as of yet, but have confirmed (using gdb) that the desired overload is begin called.
Any advice on how formally do a performance analysis?

Johan511:
Following are the performance test graphs


hpx::make_heap(hpx::execution::par, vec1.begin(), vec1.end());
par


hpx::make_heap(hpx::execution::par_unseq, vec1.begin(), vec1.end());
par_unseq

hkaiser (commented):
From your graphs, your patch has no noticeable effect on the algorithm's performance. Are you sure the compiler is able to see through the `sift_down` function to vectorize the loop?

Johan511:
Yes, I have gone through the assembly generated, sift down is not being vectorized. I have talked to few people with more experience than me in this field and they believe the function is too complex and might not be vectorized, they have advised me to look into std::experimental::simd and also into breaking the loop into parts and vectorize a simpler part of it. I will try doing it and benchmarking them. It is probably better to close this issue as the PR is unlikely to work.

hkaiser:
> Yes, I have gone through the assembly generated, sift down is not being vectorized. I have talked to few people with more experience than me in this field and they believe the function is too complex and might not be vectorized, they have advised me to look into std::experimental::simd and also into breaking the loop into parts and vectorize a simpler part of it. I will try doing it and benchmarking them. It is probably better to close this issue as the PR is unlikely to work.

I'd be happy with leaving this PR open. You can always (force-) push more changes here to modify the patch.

hkaiser:
@Johan511 could you please rebase this onto master, now that the release is out?

PR #6200, opened 2023-03-16
Par link jobs

Johan511:
Added cmake Option HPX_PARALLEL_LINK_JOBS to limit number of parallel link jobs if using ninja as generator.

Usage: 
-DHPX_PARALLEL_LINK_JOBS=2

defaults to 2 if not set (and generator is ninja)

How do I test if it works?

Referenced and motivated from https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/HandleLLVMOptions.cmake#L46

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -444,7 +444,7 @@ hpx_option(
 )
 hpx_option(
   HPX_WITH_PARALLEL_LINK_JOBS STRING
-  "Number of Parallel link jobs while building hpx (only for Ninja as generator) (default Any)" "2"
+  "Number of Parallel link jobs while building hpx (only for Ninja as generator) (default 2)" "2"
   CATEGORY "Build Targets"
 )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -444,7 +444,7 @@ hpx_option(
 )
 hpx_option(
   HPX_WITH_PARALLEL_LINK_JOBS STRING
-  "Number of Parallel link jobs while building hpx (only for Ninja as generator) (default Any)" "Any"
+  "Number of Parallel link jobs while building hpx (only for Ninja as generator) (default Any)" "2"
   CATEGORY "Build Targets"
 )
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -442,6 +442,23 @@ hpx_option(
   "Create build system support for fail compile HPX tests (default ON)" ON
   CATEGORY "Build Targets"
 )
+hpx_option(
+  HPX_PARALLEL_LINK_JOBS STRING
```
hkaiser, line 446:
```suggestion
  HPX_WITH_PARALLEL_LINK_JOBS STRING
```

CMakeLists.txt
```diff
@@ -442,6 +442,23 @@ hpx_option(
...
+if(CMAKE_GENERATOR MATCHES "Ninja")
+  if(NOT HPX_PARALLEL_LINK_JOBS)
+    set(HPX_PARALLEL_LINK_JOBS "2")
+  endif()
+  if(HPX_PARALLEL_LINK_JOBS)
+    set_property(GLOBAL APPEND PROPERTY JOB_POOLS link_job_pool=${HPX_PARALLEL_LINK_JOBS})
+    set(CMAKE_JOB_POOL_LINK link_job_pool)
+  endif()
+elseif(HPX_PARALLEL_LINK_JOBS)
+  MESSAGE(WARNING "Job pooling is only available with Ninja generators.")
```
hkaiser, line 460:
```suggestion
  hpx_warn("Job pooling is only available with Ninja generators.")
```

libs/core/algorithms/include/hpx/parallel/algorithms/make_heap.hpp
hkaiser, line 1: Please separate these changes into an independent PR.

hkaiser (changes requested):

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6205, opened 2023-03-21
Do not generate hpxcxx if support for pkgconfig was disabled

### LINKED ISSUE #6156: hpxcxx does not work if HPX_WITH_PKGCONFIG=OFF
`hpxcxx` depends on the generation of `pkgconfig` directory to locate HPX.  If `HPX_WITH_PKGCONFIG` is set to off (which for some reason is the default on macOS), `hpxcxx` will not be able to locate HPX. 

If `HPX_WITH_PKGCONFIG` is not enabled, `hpxcxx` should not be generated.

hkaiser:
Fixes #6156


hkaiser:
bors merge

PR #6208, opened 2023-03-23
Fix the compilation warning in the MPI parcelport with gcc 11.2

JiakunYan:
Without this change, it will give me (with HPX_WITH_COMPILER_WARNINGS_AS_ERRORS=ON)

/XXX/hpx/libs/full/parcelport_mpi/include/hpx/parcelport_mpi/header.hpp: In member function ‘bool hpx::parcelset::policies::mpi::header::valid() const’: /XXX/hpx/libs/full/parcelport_mpi/include/hpx/parcelport_mpi/header.hpp:85:29: error: comparison is always true due to limited range of data type [-Werror=type-limits]
   85 |             return data_[0] != -1;


hkaiser:
bors merge

PR #6215, opened 2023-03-29
added vectorization to generate_n

Johan511:
added vectorization to generate_n

hpx::generate_n calls std::generate_n, if in parallel mode it splits up the work into chunks and calls generate_n on each chunk.
Previously no execution policy was specified for std::generate_n (defaulted to seq), this PR changes it and mentions seq or unseq based on hpx::execution policy mentioned by user 

par_unseq: 
scale

par:
add


hkaiser:
@Johan511 could you create graphs that use the same y-axis limits, please?

hkaiser:
The code you proposed touches on sequential operations only. Could you measure the sequential speedup as well?

Johan511:
The change works for par_unseq too as parallel version of generate_n works by calling sequential generate on chunks. Will post speedups for unseq soon.

Johan511:
Disposed runs which took more than 0.3ms

unseq : 
scale
mean : 0.1488

seq
add
mean : 0.2056

Johan511:
Please note that the performance gains are actually not very significant.
Reason for minimal performance gains is because std::generate_n has very similar performance when compiled with -O3 flag. Google bench results are attached.

Without -O3 flag
BM_gen_n_par          6889328 ns      6888372 ns           76
BM_gen_n_par_unseq    2665548 ns      2665325 ns          257



With -O3 flag
-------------------------------------------------------------
Benchmark                   Time             CPU   Iterations
-------------------------------------------------------------
BM_gen_n_par           124210 ns       124210 ns         4038
BM_gen_n_par_unseq     159027 ns       159020 ns         4949

hkaiser:
> Please note that the performance gains are actually not very significant. Reason for minimal performance gains is because std::generate_n has very similar performance when compiled with -O3 flag. Google bench results are attached.
> 
> ## Without -O3 flag
> Benchmark Time CPU Iterations BM_gen_n_par 6889328 ns 6888372 ns 76 BM_gen_n_par_unseq 2665548 ns 2665325 ns 257
> 
> ## With -O3 flag
> Benchmark Time CPU Iterations BM_gen_n_par 124210 ns 124210 ns 4038 BM_gen_n_par_unseq 159027 ns 159020 ns 4949

You should always enable all optimizations for performance measurements.

Johan511:
-O3 flag seems to tries vectorize most loops. Should I try compiling HPX with O2 flag and compare performance of vectorized vs non vectorized?

Often times the performance on vectorization gains seem to be minimal as -O3 seems to already vectorize loops.

hkaiser:
@Johan511 could you please rebase this onto master, now that the release is out?

hkaiser:
inspect was reporting:
```
/libs/core/algorithms/include/hpx/parallel/algorithms/detail/generate.hpp

*I* missing #include (type_traits) for symbol std::true_type on line 57
```

Please rebase one more time to pull in all changes from master.

srinivasyadav18:
@hkaiser I have rebased and added unit tests to ensure everything is working with **generate_n** algorithm with **unseq** and **par_unseq** execution policies. 

And for the performance, both **seq** and **unseq** are almost generating same assembly with **Release Mode** (uses -O3). Because compiler is able to vectorize the loops in **seq** mode also, there is seems to be no extra gains using **unseq**.
However, if compared with `-fno-tree-vectorize -O3` which disables auto vectorization, there is a 3-5x speed up.

hkaiser:
retest lsu

PR #6216, opened 2023-04-02
added tests for set_difference, updated set_operation.hpp to fix #6198

### LINKED ISSUE #6198: set_intersection/set_difference fails when run with execution::par
## Expected Behavior

The function should return the same result independent of the # of threads it uses. It should return an iterator to the end of the range with the common elements. This happens with set_difference as well.

## Actual Behavior

Running the algorithm with `hpx::execution::par` gives out different results depending on how many threads it is run with.

For input std::vectors: 
{-1, 1, 2, 3, 4, 5} and 
{0, 1, 2, 3, 8, 10} 

it gives 3 common elements with 1-3 threads, 2 for 4 threads, 1 for 5 threads, 3 again for 6 threads, and 2, 1, 0, -1 (?!), -2, -3 for 7...12 threads. (negatives are result of pointer distance, not something explicitly returned by the algorithm)

For input:
int v1[30] = {0, 1 ... 29},
int v2[29] = {15, 16 ... 43}

it gives 14 common elements for 7, 9, 11 threads, 13 common elements for 12 threads,

For input: (according to the documentation this is well defined)
{ 1, 1, 1, 4, 5 }
{ 1, 1, 1 }
It crashes with a `double free` error when run with 5 or more threads
 
Here is the code i used:
```c++   
  std::vector<int> v = { 1, 2, 3, 4, 5 };
  std::vector<int> w = { 3, 4, 5 };
  std::vector<int> x(v.size() + w.size());

  auto s = hpx::set_intersection(hpx::execution::par, v.begin(), v.end(), w.begin(), w.end(), x.begin());

  std::cout << "Common elements amount: " << std::distance(x.begin(), s) << std::endl;
```

## Specifications

I am using

  HPX: V1.9.0-trunk (AGAS: V3.0), Git: 248a14c80a
  Boost: V1.81.0
  Hwloc: V2.8.0

Build:
  Type: release
  Date: Mar  3 2023 17:15:21
  Platform: linux
  Compiler: GNU C++ version 11.3.0
  Standard Library: GNU libstdc++ version 20220421
  Allocator: system

Johan511:
Fixes #6198 

Error : when number of chunks < number of cores => error occurred as chunk by default length of chunk was -1 (changed to 1)

added randomised tests which assume std::set_difference is implemented correctly

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/detail/set_operation.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/detail/set_operation.hpp
@@ -74,12 +74,12 @@ namespace hpx::parallel::detail {
 
     struct set_chunk_data
     {
-        static const std::size_t uninit_start = static_cast<std::size_t>(-1);
-        static const std::size_t uninit_len = static_cast<std::size_t>(0);
-        static const std::size_t uninit_start_index =
+        static constexpr std::size_t uninit_start = static_cast<std::size_t>(-1);
+        static constexpr std::size_t uninit_len = static_cast<std::size_t>(0);
+        static constexpr std::size_t uninit_start_index =
             static_cast<std::size_t>(-1);
-        static const std::size_t uninit_first1 = static_cast<std::size_t>(-1);
-        static const std::size_t uninit_first2 = static_cast<std::size_t>(-1);
+        static constexpr std::size_t uninit_first1 = static_cast<std::size_t>(-1);
+        static constexpr std::size_t uninit_first2 = static_cast<std::size_t>(-1);
 
         std::size_t start = uninit_start;
         std::size_t len = uninit_len;
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/detail/set_operation.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/detail/set_operation.hpp
@@ -74,11 +74,18 @@ namespace hpx::parallel::detail {
 
     struct set_chunk_data
     {
-        std::size_t start = static_cast<std::size_t>(-1);
-        std::size_t len = static_cast<std::size_t>(-1);
-        std::size_t start_index = static_cast<std::size_t>(-1);
-        std::size_t first1 = static_cast<std::size_t>(-1);
-        std::size_t first2 = static_cast<std::size_t>(-1);
+        static const std::size_t uninit_start = static_cast<std::size_t>(-1);
+        static const std::size_t uninit_len = static_cast<std::size_t>(0);
+        static const std::size_t uninit_start_index =
+            static_cast<std::size_t>(-1);
+        static const std::size_t uninit_first1 = static_cast<std::size_t>(-1);
+        static const std::size_t uninit_first2 = static_cast<std::size_t>(-1);
+
+        std::size_t start = uninit_start;
+        std::size_t len = uninit_len;
+        std::size_t start_index = uninit_start_index;
+        std::size_t first1 = uninit_first1;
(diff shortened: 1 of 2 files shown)


Johan511:
[Following](https://github.com/Johan511/HPX-SetOperationsTest) is a repository to test the changes independently of HPX test suite

libs/core/algorithms/include/hpx/parallel/algorithms/detail/set_operation.hpp
```diff
@@ -74,11 +74,18 @@ namespace hpx::parallel::detail {
...
+        static const std::size_t uninit_start_index =
+            static_cast<std::size_t>(-1);
+        static const std::size_t uninit_first1 = static_cast<std::size_t>(-1);
+        static const std::size_t uninit_first2 = static_cast<std::size_t>(-1);
+
+        std::size_t start = uninit_start;
+        std::size_t len = uninit_len;
+        std::size_t start_index = uninit_start_index;
+        std::size_t first1 = uninit_first1;
+        std::size_t first2 = uninit_first2;
```
hkaiser, line 88: What is the rationale for this change?

hkaiser:
There are a couple of inspect test failures: https://output.circle-artifacts.com/output/job/0c3f30ab-fc2d-4301-9bd2-8601014bd5c4/artifacts/0/hpx_inspect_report.html. 

Also, does this help diagnosing the problems reported by #6198?

libs/core/algorithms/include/hpx/parallel/algorithms/detail/set_operation.hpp
Johan511: After the async execution, we have an if conditional which checks if the chunk values are same an default values, in which case it does not process the chunk. I felt defining the default values was a better alternative to simply checking is start_index == -1. Note that the bug was because a default value was incorrect.

Johan511:
The number of chunks is always same as number of cores, but in cases when the sets are small we run into errors. The randomised tests which I have added are expected to catch them but as they are randomised, there is an minuscule possibility they would not be caught (more likely when number of threads being used is small)

Johan511:
I am not sure if the test case mentioned in the issue will always give an wrong answer, it would depend on the number of cores on the user's system. I will try to design a test case which would work for the majority of situations. I still do believe the randomised test cases are a very good choice for catching such an error.

libs/core/algorithms/include/hpx/parallel/algorithms/detail/set_operation.hpp
hkaiser: Hmmm. What change fixed the issue? From what I can see, this PR does not introduce any changes that would modify behavior. I might miss something, however.

Johan511: len used to default to -1, similar to remaining variables. It is supposed to default to 0. There was also a out of bounds memory access I believe.

hkaiser: Ok, now I see the change. If you want to separate the initial values from the initialization, I'd suggest using `static constexpr` variables, however.

Johan511: updated, missed that point earlier.

hkaiser (approved):
LGTM, thanks!

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6218, opened 2023-04-08
Creating INSTALL component 'runtime' to enable installing binaries only

### LINKED ISSUE #6188: Request for adding support for installing only runtime targets
I maintain the HPX Conda package (https://anaconda.org/conda-forge/hpx). Conda packages get built for various platforms (https://github.com/conda-forge/hpx-feedstock). After being built, they are available for people to install in their Conda environment, for example to be able to build their own software that depends on HPX and which provides a Python API.

To be able to **build a Conda package** that depends on HPX, it is necessary that there is a HPX Conda package containing all the HPX headers, libraries, CMake scripts, etc.

To be able to **use a Conda package** that depends on HPX, it is useful that there is an HPX Conda package containing only the shared libraries and likely some other files (e.g. `hpxrun.py` is useful). This makes the package smaller. These users don't need the static libraries, header files, etc.

I am submitting this issue because I think it is currently not possible to install only the runtime targets of HPX. If this is indeed the case, is support for installing only runtime targets be something that the HPX team is willing to add?

Once I can create seperate development and runtime packages I will update the HPX Conda packaging logic to create a `hpx-dev` and `hpx` package.

hkaiser:
Fixes #6188

### DIFF
--- a/.cmake-format.py
+++ b/.cmake-format.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2022 Hartmut Kaiser
+# Copyright (c) 2020-2023 Hartmut Kaiser
 #
 # SPDX-License-Identifier: BSL-1.0
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
@@ -238,6 +238,7 @@ with section("parse"):
                                        'HEADER_ROOT': 1,
                                        'INI': 1,
                                        'INSTALL_SUFFIX': 1,
+                                       'INSTALL_COMPONENT': 1,
                                        'LANGUAGE': 1,
                                        'LINK_FLAGS': '+',
                                        'OUTPUT_SUFFIX': 1,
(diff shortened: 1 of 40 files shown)


hkaiser:
@kordejong could you please verify if this fixes the issue you reported? The installation component to use for this is `runtime`.

kordejong:
> @kordejong could you please verify if this fixes the issue you reported? The installation component to use for this is `runtime`.

Thanks, yes, the `runtime` install target works. I did notice two remaining static libraries in de runtime install directory:
`./lib/libhpx_init.a`
`./lib/libhpx_wrap.a`
Other than that it looks good to me.

hkaiser:
> > @kordejong could you please verify if this fixes the issue you reported? The installation component to use for this is `runtime`.
> 
> Thanks, yes, the `runtime` install target works. I did notice two remaining static libraries in de runtime install directory: `./lib/libhpx_init.a` `./lib/libhpx_wrap.a` Other than that it looks good to me.

Ok, good point. I will remove those from the 'runtime' target.

hkaiser:
bors merge

PR #6221, opened 2023-04-12
Fix segmented algorithms tests

Pansysk75:
Fixes a group of tests for HPX segmented algorithms, which were occasionally failing due to improper initialization of partitioned_vector. Initialization using partitioned_vector::set_value() is asynchronous and returns a future, but we never waited for it to complete.

### DIFF
--- a/libs/full/segmented_algorithms/tests/unit/partitioned_vector_adjacent_find1.cpp
+++ b/libs/full/segmented_algorithms/tests/unit/partitioned_vector_adjacent_find1.cpp
@@ -34,9 +34,10 @@ void initialize(hpx::partitioned_vector<T>& xvalues)
     T init_array[SIZE] = {1, 2, 3, 4, 5, 1, 2, 3, 1, 5, 2, 3, 4, 2, 3, 2, 1, 2,
         3, 4, 5, 6, 5, 6, 1, 2, 3, 4, 2, 1, 2, 3, 3, 5, 4, 3, 2, 1, 1, 2, 3, 4,
         1, 2, 3, 1, 1, 1, 1, 1, 1, 1, 1, 7, 6, 5, 7, 5, 4, 2, 3, 4, 5, 2};
-    for (int i = 0; i < SIZE; i++)
+    typename hpx::partitioned_vector<T>::iterator it = xvalues.begin();
+    for (int i = 0; i < SIZE; i++, it++)
     {
-        xvalues.set_value(i, init_array[i]);
+        *it = init_array[i];
     }
 }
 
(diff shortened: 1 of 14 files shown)


hkaiser (changes requested):
Very nice! Thanks a lot! Could you please fix the clang-format issues? Otherwise LGTM.

hkaiser (approved):
LGTM, thanks!

Pansysk75:
> @Pansysk75 We now could revert the change here [e427e0e](https://github.com/STEllAR-GROUP/hpx/commit/e427e0e50919956b8a095f55401b8a1fdec1a6b7). This was introduced to ignore errors when running the segmented algorithm tests.

I saw some "Timeout" failures on some segmented algorithms on some past CircleCI runs. This fix does not address those failures. For example: [https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/14417/workflows/69d7cda3-1ae3-4492-b881-a39d963dd6b5/jobs/340881/tests](url)

I would like to try fixing that as well, but I'm not sure how to approach that issue, or what it's cause might be. The CircleCI error message is not very informative.
Well... there is also a 2Gb debug-log.txt, which is saved as an artifact, but I haven't gotten around to trying to parse that yet.

Let me know if you have any input on that.

Pansysk75:
Then there is also` tests.unit.modules.runtime_components.distributed.tcp.migrate_polymorphic_component` failing.
I'll use this PR as an opportunity to look into that as well.

hkaiser:
> Then there is also` tests.unit.modules.runtime_components.distributed.tcp.migrate_polymorphic_component` failing. I'll use this PR as an opportunity to look into that as well.

I appreciate that. Let's do this on a separate PR, however.

hkaiser:
> Well... there is also a 2Gb debug-log.txt, which is saved as an artifact, but I haven't gotten around to trying to parse that yet.

Short of reproducing this locally, the log file is all we have.

hkaiser:
@Pansysk75 While this fixes one issue, it seems that there are more [problems](https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/14455/workflows/50cca2fb-0577-41bb-9c4a-8524226a232d/jobs/341638). Should we go ahead with merging this PR anyways?

Pansysk75:
> @Pansysk75 While this fixes one issue, it seems that there are more [problems](https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/14455/workflows/50cca2fb-0577-41bb-9c4a-8524226a232d/jobs/341638). Should we go ahead with merging this PR anyways?

I agree, let's merge this PR, since I don't have any insight on the "Timeout" issue for the time beign.

PR #6225, opened 2023-04-15
Make heap fails in par for small sized heaps #6217

### LINKED ISSUE #6217: hpx::make_heap does not terminate when exPolicy is par (or par_unseq) and size of vector is 4
hpx::make_heap does not terminate when exPolicy is par (or par_unseq) and size of vector is 4

## Expected Behavior

function returns

## Actual Behavior

function does not return

`std::vector<int> v(4, 0);`
`hpx::make_heap(hpx::execution::par, v.begin(), v.end());`

Johan511:
Fixes #6217

added tests, fixed bug. We try to find last element index of parent level (end_exclusive). If child is element of index 1, last element of parent level becomes -1 and there is negative overflow.

fixed it by enforcing end_exclusive to be >=0

## Checklist

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/make_heap.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/make_heap.hpp
@@ -394,16 +394,15 @@ namespace hpx::parallel {
                     while (start > 0)
                     {
                         // Index of start of level, and amount of items in level above start
-                        std::size_t const end_exclusive =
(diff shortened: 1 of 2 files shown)


hkaiser (dismissed):
LGTM, thanks!

hkaiser:
@Johan511 could you please take care of the inspect errors?

hkaiser:
bors merge

PR #6227, opened 2023-04-18
Reliably disable background work when no networking is enabled

hkaiser:


hkaiser:
@diehlpk this should resolve the performance regression you reported when running on one locality.

PR #6228, opened 2023-04-18
Fixing algorithms for zero length sequences when run with s/r scheduler

hkaiser:
@Pansysk75 this should fix the issue we discussed. Please verify.


hkaiser:
bors merge

Pansysk75:
@hkaiser I didn't follow up with the regressive tests I promised here, because I couldn't reproduce the issue in a minimal example (ie by directly invoking a zero-length for_each using S/R), at least not in a multi-threaded setup.

However, this fixed the crash of the S/R rotate algorithm with the midpoint at the last element of the container (which was where this bug was discovered).

PR #6229, opened 2023-04-30
Adding zero-copy support on the receiving end of the TCP and MPI parcel ports

hkaiser:
- flyby: cleaning up and modernizing TCP parcel port

@JiakunYan this implements what we discussed recently by de-serializing received parcels once the chunk information is available. This de-serialization however does not assume that the chunk data has been received, but merely allocates the memory for the subsequent networking operations to place received chunk data directly into the internal memory buffers.


libs/full/parcelport_mpi/include/hpx/parcelport_mpi/receiver_connection.hpp
```diff
@@ -152,19 +154,74 @@ namespace hpx::parcelset::policies::mpi {
...
+
+            while (chunks_idx_ != buffer_.chunks_.size())
             {
+                auto& c = buffer_.chunks_[chunks_idx_++];
+                if (c.type_ == serialization::chunk_type::chunk_type_index)
+                {
+                    continue;    // skip non-zero-copy chunks
+                }
+
                 if (!request_done())
```
JiakunYan, line 208: I think you should put this `if (!request_done())` in front of `chunks_idx_++` of line 202. Otherwise, you will skip this zero-copy chunk if the receive is not completed.

JiakunYan:
Do you think it is better to add a command line option controlling this zero-copy behavior so we can figure out how much performance improvement we get from this optimization?

hkaiser:
> Do you think it is better to add a command line option controlling this zero-copy behavior so we can figure out how much performance improvement we get from this optimization?

This is a good suggestion. I will add a configuration variable for this.

libs/full/parcelport_mpi/include/hpx/parcelport_mpi/receiver_connection.hpp
hkaiser: Good catch! This is what was causing the assertion failures you mentioned above.

hkaiser:
> > Do you think it is better to add a command line option controlling this zero-copy behavior so we can figure out how much performance improvement we get from this optimization?
> 
> This is a good suggestion. I will add a configuration variable for this.

This will require some acrobatics as the two versions rely on different types representing the chunking data (the old uses a `std::vector<char>`, the new `serialization::detail::chunk_data`. I'll see if using a `variant<>` isn't making things unwieldy.

JiakunYan:
> This will require some acrobatics as the two versions rely on different types representing the chunking data (the old uses a std::vector<char>, the new uses serialization::serialization_chunk). I'll see if using a variant<> isn't making things unwieldy.

`variant<>` might complicate things. I think you can use chunk_data for both cases. I guess you can just create a chunk_data from a vector<char>?
I can do this later once you merge this PR.

hkaiser:
@JiakunYan this now supports the `--hpx:ini=hpx.parcel.zero_copy_receive_optimization=0` key to disable the new zero-copy support on the receiving end (default is `on`)

docs/sphinx/manual/launching_and_configuring_hpx_applications.rst
```diff
@@ -465,6 +465,7 @@ The ``hpx.parcel`` configuration section
     max_outbound_message_size = ${HPX_PARCEL_MAX_OUTBOUND_MESSAGE_SIZE:<hpx_parcel_max_outbound_message_size>}
     array_optimization = ${HPX_PARCEL_ARRAY_OPTIMIZATION:1}
     zero_copy_optimization = ${HPX_PARCEL_ZERO_COPY_OPTIMIZATION:$[hpx.parcel.array_optimization]}
+    zero_copy_receive_optimization = ${HPX_PARCEL_ZERO_COPY_RECEIVE_OPTIMIZATION:$[hpx.parcel.array_optimization]}
```
JiakunYan, line 468: Should the default value be `$[hpx.parcel.array_optimization]`? I think it can just be `1`?

hkaiser: We can't logically support zero-copy networking with array optimization is off.  However, I think the default should be `$[hpx.parcel.zero_copy_optimization]`. I'll make that change.

hkaiser:
bors merge

PR #6233, opened 2023-05-05
Moving is_relocatable to namespace hpx

hkaiser:
Fixes compilation problems on master

### DIFF
--- a/libs/core/algorithms/include/hpx/algorithms/traits/is_relocatable.hpp
+++ b/libs/core/algorithms/include/hpx/algorithms/traits/is_relocatable.hpp
@@ -8,7 +8,7 @@
 
 #include <type_traits>
 
-namespace hpx::traits {
+namespace hpx {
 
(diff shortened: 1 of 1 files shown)


hkaiser:
@isidorostsa please verify that this is ok.

isidorostsa:
It looks good to me! Should I also change the unit test to look for hpx::traits::is_relocatable<>?

hkaiser:
> It looks good to me! Should I also change the unit test to look for hpx::traits::is_relocatable<>?

I think we should keep the trait as `hpx::is_relocatable` for compatibility with the standard.

isidorostsa:
> > It looks good to me! Should I also change the unit test to look for hpx::traits::is_relocatable<>?
> 
> I think we should keep the trait as `hpx::is_relocatable` for compatibility with the standard.

I completely agree with that

hkaiser:
bors merge

PR #6236, opened 2023-05-08
Attempt to resolve occasional shutdown hangs in distributed operation

hkaiser:
@Pansysk75 could you please try this patch to see if it resolves the issues we were discussing?


Pansysk75:
@hkaiser Still getting hangs at the same spot occasionally.

In the 2 node setup, node 0 initiates the dijkstra termination algorithm and successfully invokes dijkstra_termination_action on node 1, but the invocation of dijkstra_termination_action from node 1 with destination node 0 does not seem to be picked up by node 0 (note that the newly added latch gets decremented all the way to 0).

Thus, node 0 hangs on the following step:
https://github.com/STEllAR-GROUP/hpx/blob/4647d1931cd3902f4e9f674f8f51cc28e8756409/libs/full/runtime_distributed/src/server/runtime_support_server.cpp#L377-L382

I confirmed this with additional debug logging.

Is there a flag for additional parcelport logging messages?

hkaiser:
@Pansysk75 thanks! I have added some more changes, this time on the receiving end of the final parcel.

> Is there a flag for additional parcelport logging messages?

This should be enabled by the settings you already use. Look for log messages that are tagged with `[PT]`. I would consider it a bug if those were not emitted.

Pansysk75:
Now the issue does not appear, at least not in the way that it did before.
There is another occasional hang, however. 

There seems to be a non-terminating base_set_event_action thread in the queue of the node initiating the shutdown.
This prevents the thread manager from reporting as idle, so we get stuck here, at the first run of the dijkstra routine:
https://github.com/STEllAR-GROUP/hpx/blob/bae1f089cbec2fc752cf193a375a71106702aecd/libs/full/runtime_distributed/src/server/runtime_support_server.cpp#L248

In the meanwhile, I'll check how these correlate with including the partitioned_vector facilities.

hkaiser:
> @hkaiser Now the issue does not appear, at least not in the way that it did before. There is another occasional hang, however.
> 
> There seems to be a non-terminating base_set_event_action thread in the queue of the node initiating the shutdown. This prevents the thread manager from reporting as idle, so we get stuck here, at the first run of the dijkstra routine:
> 
> https://github.com/STEllAR-GROUP/hpx/blob/bae1f089cbec2fc752cf193a375a71106702aecd/libs/full/runtime_distributed/src/server/runtime_support_server.cpp#L248
> 
> In the meanwhile, I'll check how these issues correlate with including the partitioned_vector facilities, which is how I've reproducing this issue all along.
> 
> Edit: The issues still appear without the partitioned_vector facilities, they just surface less often.

Thanks! I'll have some ideas what could be done about this. Also, the segmented algorithms do not hang on the CI anymore. Some progress, it seems.

hkaiser:
@Pansysk75 I added some more possible fixes. Could you please try again?

Pansysk75:
@hkaiser The hangs are very sparse now. Here's some clues about the latest issue I mentioned:

The non-terminating `base_set_event_action` which occasionally prevents shutdown seems to be a remnant from startup, lanuched (I think) around here:
https://github.com/STEllAR-GROUP/hpx/blob/242eb290e7f5f83ce1c15dc7c4ad430203094504/libs/full/init_runtime/src/pre_main.cpp#L188C7-L198

Also in the logs, there are 8 "entering local barrier" messages, but only 7 "exiting local barrier" on startup, I assume it's related.

Here are the logs from node 0. I suppose around line 27070 is where the interesting stuff happens (regex for `create_work.*base_set` and `base_set_event_action.*terminated` was also useful):

[debug_log_node0.txt](https://github.com/STEllAR-GROUP/hpx/files/11456208/debug_log_node0.txt)

Hope this helps!

hkaiser:
bors merge

PR #6238, opened 2023-05-10
Update lsu perftest baseline

Pansysk75:


hkaiser:
add to whitelist

Pansysk75:
retest lsu

Pansysk75:
@hkaiser Well that's not great... Let me think what I can do about this

Pansysk75:
retest lsu

Pansysk75:
@hkaiser The final performance measurements are now produced by merging together data from multiple executions.
Based on the last time we looked into this, this should make things more stable.

This only affects the tests running on Rostam.
(Probably not the most elegant way to implement this, but we can make it prettier if it turns out to work)

hkaiser:
> @hkaiser The final performance measurements are now produced by merging together data from multiple executions. Based on the last time we looked into this, this should make things more stable.
> 
> This only affects the tests running on Rostam. (Probably not the most elegant way to implement this, but we can make it prettier if it turns out to work)

Looks like the patch fixies the issue, let's merge it.

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6239, opened 2023-05-11
Fix `Optimizing HPX applications` page of Manual

dimitraka:
The content of the tables used to overlap with the side bar of the documentation making it impossible to actually read the information. These extended tables were split to smaller tables which are now clear and readable. 

With this change there is no need to include anchors to the table rows. For each counter there is a separate table, so one can just use the reference to the table in order to refer to a specific counter.


hkaiser:
bors merge

Issue #5799, opened 2022-02-23
Investigate CUDA compilation problems


hkaiser:
Two of our CUDA tests fail compiling by NVCC unceremoniously stopping without error message. For now, these tests have been disabled (see: https://github.com/STEllAR-GROUP/hpx/pull/5796). We should investigate whether this can be worked around.

G-071:
NVCC apparently got some improvement/fixes with the CUDA/12.3 release meaning we might be able to finally fix this old issue (which is still very much relevant as the same problem prevents us from using/compiling Kokkos >= 4.1 with gcc and the HPX and CUDA execution space both enabled)

With previous CUDA releases, it simply crashed when compiling transform_stream.cu:
```nvcc error   : 'cudafe++' died due to signal 11 (Invalid memory reference)```

With CUDA 12.3 we get some actual compilation errors instead:

```
[100%] Building CUDA object libs/core/async_cuda/tests/unit/CMakeFiles/transform_stream_test.dir/transform_stream.cu.o
cd /work/gdaiss/hpx/build/libs/core/async_cuda/tests/unit && /usr/local/cuda-12.3/bin/nvcc -forward-unknown-to-host-compiler -DASIO_DISABLE_CONSTEXPR -DHPX_APPLICATION_EXPORTS -DHPX_APPLICATION_NAME=transform_stream_test -DHPX_APPLICATION_NAME_DEFAULT=transform_stream_test -DHPX_APPLICATION_STRING=\"transform_stream_test\" -DHPX_DEBUG -DHPX_PREFIX=\"/work/gdaiss/hpx/build\" -DHPX_PREFIX_DEFAULT=\"/usr/local\" -D_GNU_SOURCE --options-file CMakeFiles/transform_stream_test.dir/includes_CUDA.rsp -g -std=c++17 --generate-code=arch=compute_70,code=[compute_70,sm_70] --generate-code=arch=compute_80,code=[compute_80,sm_80] -Xcompiler=-fPIE
[...]

PR #6248, opened 2023-05-18
Fix CUDA/HIP Jenkins pipelines 

G-071:
This PR aims to fix the unstable CUDA/HIP tests.

Changes so far:
- Fixes CUDA architectures for Jenkins nodes
- Exclude faulty Jenkins node for now (bahram: listed as V100 GPU node but does not list any GPUs according to nvidia-smi)
- Allow Jenkins HIP builds with warnings (in order to run tests despite of the warnings, just like the CUDA tests already do)
- I further looked into #5799 again to see if the new compilers and cuda versions made any difference. Unfortunately not, using gcc/12 and cuda/12 compilation still fails with:
`'cudafe++' died due to signal 11 (Invalid memory reference)` 
I adapted the guards in the failing examples accordingly (to also skip the tests with newer cuda versions)

The PR is a bit of a work-in-progress as I imagine it will take a few iterations to get everything right. That being said, let's see if the current changes already help and trigger the Jenkins tests!


hkaiser (dismissed):
@G-071 many thanks! The clang-format error has been fixed on master, we can ignore it here. 
LGTM, thanks!

G-071:
There are a few more things I would like to test before merging:

- I *think* I fixed the underlying issue with the HIP tests, however, both the rocm/4 and rocm/5 modules on the nodes are currently not working properly so we have to wait until this is fixed to actually test it (I already contacted Al about it).
- The cuda gcc-10-cuda-11 tests seem to work fine on the V100 node, but I'd also like to test on the A100 node just to be safe (which was not available yesterday).

PR #6249, opened 2023-05-19
Implement the send immediate optimization for the MPI parcelport.

JiakunYan:
Users can use the option `hpx.parcel.mpi.sendimm` to enable and disable it.

When HPX wants to send a parcel, the default behavior is to enqueue this parcel into the parcel queue of the target locality, dequeue all the parcels from that parcel queue, aggregate them, and pass them in one `parcel_buffer` structure. It will also try to get an existing connection from the connection cache. These two data structures are protected by locks and can be a source of thread contention.

With the send immediate optimization, the parcelport will bypass the connection cache and parcel queues and directly send the parcel.

This option is mainly for research purposes. The default is off since it will slow down the MPI parcelport in most cases.


hkaiser:
@JiakunYan Please let me know if this is ready.

JiakunYan:
@hkaiser Sorry forgot to let you know. It is ready!

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6253, opened 2023-05-20
Modernize set_thread_name on Windows

hkaiser:
This changes setting thread names on Windows such that the name is visible in the debugger even if the process was attached and not started by VS.


hkaiser:
bors merge

PR #6255, opened 2023-05-23
#6224, fold algorithms

### LINKED ISSUE #6224: Add C++23 fold algorithms
C++23 added the following `fold` algorithms:
- `ranges::fold_left` (https://en.cppreference.com/w/cpp/algorithm/ranges/fold_left)
- `ranges::fold_left_first` (https://en.cppreference.com/w/cpp/algorithm/ranges/fold_left_first)
- `ranges::fold_right` (https://en.cppreference.com/w/cpp/algorithm/ranges/fold_right)
- `ranges::fold_right_first` (https://en.cppreference.com/w/cpp/algorithm/ranges/fold_right_first)
- `ranges::fold_left_with_iter` (https://en.cppreference.com/w/cpp/algorithm/ranges/fold_left_with_iter)
- `ranges::fold_left_first_with_iter` (https://en.cppreference.com/w/cpp/algorithm/ranges/fold_left_first_with_iter)

These algorithms are specified as enforcing sequential execution. We should discuss whether those should be added to HPX anyways.

See also: http://eel.is/c++draft/alg.fold

Johan511:
Fixes #6224

## Proposed Changes

  - fold algorithms, as proposed in C++23

## Checklist

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/fold.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/fold.hpp
@@ -1,0 +1,78 @@
+#pragma once
+
+#include <hpx/concepts/concepts.hpp>
+#include <hpx/parallel/algorithms/reduce.hpp>
+
+namespace hpx::parallel { namespace detail {
+
+    template <typename _T>
+    struct fold_left : public algorithm<fold_left<_T>, _T>
+    {
+        constexpr fold_left() noexcept
+          : algorithm<fold_left, _T>("fold_left")
+        {
+        }
+
+        template <typename ExPolicy, typename FwdIter, typename Sent,
+            typename T, typename F>
+        HPX_HOST_DEVICE static constexpr T sequential(
+            ExPolicy&& policy, FwdIter first, Sent last, T&& init, F&& f)
+        {
+            if (first == last)
+                return HPX_MOVE(init);
+            T acc = HPX_MOVE(init);
+            while (first != last)
+                acc = HPX_MOVE(f(HPX_MOVE(acc), *first++));
+            return HPX_MOVE(acc);
+        }
+
+        template <typename ExPolicy, typename FwdIter, typename Sent,
+            typename T>
+        static util::detail::algorithm_result_t<ExPolicy, FwdIter> parallel(
+            ExPolicy&& policy, FwdIter first, Sent last, T const& val)
+        {
+            return first;
+        }
+    };
+
+}}    // namespace hpx::parallel::detail
+
+namespace hpx {
+inline constexpr struct fold_left_t final
+  : hpx::detail::tag_parallel_algorithm<fold_left_t>
+{
+private:
+    template <typename ExPolicy, typename FwdIter, typename T,
+        typename F>    // TODO : add concept
+    friend
+        typename hpx::parallel::util::detail::algorithm_result<ExPolicy>::type
+        tag_fallback_invoke(fold_left_t, ExPolicy&& policy, FwdIter first,
+            FwdIter last, T&& init, F&& f)
+    {
+        static_assert(hpx::traits::is_forward_iterator_v<FwdIter>,
+            "Requires at least forward iterator.");
+
+        using result_type =
+            typename hpx::parallel::util::detail::algorithm_result<
+                ExPolicy>::type;
+
+        return hpx::util::void_guard<result_type>(),
+               hpx::parallel::detail::fold_left<FwdIter>().call(
+                   HPX_FORWARD(ExPolicy, policy), first, last,
+                   HPX_FORWARD(T, init), HPX_FORWARD(F, f));
+    }
+
+    template <typename FwdIter, typename T,
+        typename F>    // TODO : add concept
+    // clang-format on
+    friend T tag_fallback_invoke(
+        fold_left_t, FwdIter first, FwdIter last, T&& init, F&& f)
+    {
+        static_assert(hpx::traits::is_forward_iterator_v<FwdIter>,
+            "Requires at least forward iterator.");
+
+        return hpx::parallel::detail::fold_left<T>().call(hpx::execution::seq,
+            first, last, HPX_FORWARD(T, init), HPX_FORWARD(F, f));
+    }
+} fold_left{};
+}    // namespace hpx
--- a/libs/core/algorithms/CMakeLists.txt
+++ b/libs/core/algorithms/CMakeLists.txt
@@ -50,6 +50,7 @@ set(algorithms_headers
     hpx/parallel/algorithms/exclusive_scan.hpp
     hpx/parallel/algorithms/fill.hpp
     hpx/parallel/algorithms/find.hpp
+    hpx/parallel/algorithms/fold.hpp
     hpx/parallel/algorithms/for_each.hpp
     hpx/parallel/algorithms/for_loop.hpp
     hpx/parallel/algorithms/for_loop_induction.hpp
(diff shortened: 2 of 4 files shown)


libs/core/algorithms/include/hpx/parallel/algorithms/fold.hpp
```diff
@@ -0,0 +1,74 @@
+#pragma once
+
+#include <hpx/concepts/concepts.hpp>
+#include <hpx/parallel/algorithms/reduce.hpp>
+
+namespace hpx::parallel { namespace detail {
+
+    template <typename _T>
```
hkaiser, line 8: Please do not use symbols that start with an underscore

libs/core/algorithms/include/hpx/parallel/algorithms/fold.hpp
```diff
@@ -0,0 +1,74 @@
...
+inline constexpr struct fold_left_t final
+  : hpx::detail::tag_parallel_algorithm<fold_left_t>
+{
+private:
+    template <typename ExPolicy, typename FwdIter, typename T,
+        typename F>    // TODO : add concept
+    friend
+        typename hpx::parallel::util::detail::algorithm_result<ExPolicy>::type
+        tag_fallback_invoke(fold_left_t, ExPolicy&& policy, FwdIter first,
+            FwdIter last, T&& init, F&& f)
```
hkaiser, line 46: The initial value `init` and the function `f` should be taken by value.

hkaiser (changes requested):

libs/core/algorithms/include/hpx/parallel/algorithms/fold.hpp
Johan511: Is there any reason we can't pass f by reference?

hkaiser: a) the standard says to take a copy
b) this is to make sure that the lifetime of any state `F` may hold is decoupled from the invocation site

libs/core/algorithms/include/hpx/parallel/algorithms/fold.hpp
```diff
@@ -18,57 +22,156 @@ namespace hpx::parallel { namespace detail {
         HPX_HOST_DEVICE static constexpr T sequential(
             ExPolicy&&, FwdIter first, Sent last, T&& init, F&& f)
         {
-            hpx::reduce(first, last, HPX_FORWARD(T, init), HPX_FORWARD(F, f));
+            if (first == last)
+                return HPX_MOVE(init);
+            T acc = HPX_MOVE(init);
+            while (first != last)
+                acc = HPX_MOVE(f(HPX_MOVE(acc), *first++));
+            return HPX_MOVE(acc);
```
hkaiser, line 30: There is no need to move returned values from functions, those are rvalues to begin with (e.g. writing things like `a = HPX_MOVE(f());`. 

There is usually also no need to move the values in the return statement. Those are treated as rvalues as well. writing `return HPX_MOVE(bar);` does more harm than good as it disables RTVO.

But in general, what is your rationale for reimplementing the sequential fold? We do have that already somewhere for implementing the sequential `reduce()`.

libs/core/algorithms/include/hpx/parallel/algorithms/fold.hpp
```diff
@@ -18,57 +22,156 @@ namespace hpx::parallel { namespace detail {
         HPX_HOST_DEVICE static constexpr T sequential(
             ExPolicy&&, FwdIter first, Sent last, T&& init, F&& f)
         {
-            hpx::reduce(first, last, HPX_FORWARD(T, init), HPX_FORWARD(F, f));
+            if (first == last)
+                return HPX_MOVE(init);
+            T acc = HPX_MOVE(init);
```
hkaiser, line 27: This should be:
```suggestion
            T acc = HPX_FORWARD(T, init);
```

libs/core/algorithms/include/hpx/parallel/algorithms/fold.hpp
Johan511: > There is no need to move returned values from functions, those are rvalues to begin with (e.g. writing things like a = HPX_MOVE(f());.

sure will change it.


> There is usually also no need to move the values in the return statement. Those are treated as rvalues as well. writing return HPX_MOVE(bar); does more harm than good as it disables RTVO.
The [possible implementation](https://en.cppreference.com/w/cpp/algorithm/ranges/fold_left) in the standard seems to return moved values


> But in general, what is your rationale for reimplementing the sequential fold? We do have that already somewhere for implementing the sequential reduce().
I can change the implementation to hpx::reduce if that's preferred.

libs/core/algorithms/include/hpx/parallel/algorithms/fold.hpp
Johan511: Sure, will make that change

libs/core/algorithms/include/hpx/parallel/algorithms/fold.hpp
hkaiser: > I can change the implementation to hpx::reduce if that's preferred.

I think you could directly call our sequential_reduce implementation.

libs/core/datastructures/include/hpx/datastructures/in_value_result.hpp
```diff
@@ -0,0 +1,24 @@
...
+        }
+
+        template <class I2, class T2>
+        // hpx::convertible?
+        constexpr operator in_value_result<I2, T2>() &&
+        {
+            return {std::move(in), std::move(value)};
+        }
+    };
+}}    // namespace hpx::experimental::ranges
```
hkaiser, line 24: We do already have a list of similar types defined here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/algorithms/include/hpx/parallel/util/result_types.hpp

libs/core/algorithms/include/hpx/parallel/algorithms/fold.hpp
Johan511: To use the fold algorithm, the user needs to `#include <hpx/algorithm/parallel/fold.hpp>`, should we force them to include another entire .hpp file to skip writing one loop?

hkaiser: I'm not sure what you mean by that. Could you elaborate, please?

libs/core/datastructures/include/hpx/datastructures/in_value_result.hpp
Johan511: https://en.cppreference.com/w/cpp/algorithm/ranges/fold_left_with_iter

The examples here seem to suggest that the member variables are to be named as `in` and `value`

vivekd01:
i want to work on this issues please assign me this issue.

Johan511:
Hi @vivekd01, thank you for the initiative but all the fold algorithms have been implemented in the PR.

hkaiser (commented):
Could you please add tests for all the algorithms?

Johan511:
@hkaiser basic tests for all the algorithms have been added

hkaiser:
@Johan511 thanks for adding the tests. The last thing missing here is to add the documentation for the algorithms. All other algorithms have extensive documentation. Please coordinate with @dimitraka on what's necessary for this.

hkaiser:
This can be closed as the algorithms were added by https://github.com/STEllAR-GROUP/hpx/pull/6884

PR #6256, opened 2023-05-24
Fix nvcc/gcc-10 (Octo-Tiger) compilation issue 

G-071:
Using HPX 1.9.0, the Octo-Tiger compilation (using just CUDA, not Kokkos) with gcc+nvcc fails due to an internal nvcc error pointing to loop.hpp within HPX:

```
build/hpx/include/hpx/parallel/util/loop.hpp(216): internal error: assertion failed at: "cp_gen_be.c", line 22617 in gen_variable_decl                                                                                                                 

1 catastrophic error detected in the compilation of "/src/octotiger/src/unitiger/hydro_impl/flux_cuda_kernel.cpp".                                                                                                                                      
Compilation aborted.
nvcc error   : 'cudafe++' died due to signal 6
nvcc error   : 'cudafe++' core dumped
```

That is with using g++10 and CUDA 11.5.
While using newer versions of g++/CUDA fix this, it would be nice to not lose g++-10 support over this.

Interestingly, simply providing nvcc with more information about the type seems to help as well, hence this PR to resolve the issue without losing support for the older compiler.

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/util/loop.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/util/loop.hpp
@@ -214,7 +214,7 @@ namespace hpx::parallel::util {
 
 #if !defined(HPX_COMPUTE_DEVICE_CODE)
     template <typename ExPolicy>
-    inline constexpr loop_ind_t loop_ind = loop_ind_t<ExPolicy>{};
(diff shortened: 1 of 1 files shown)


hkaiser (approved):
Would it make sense to apply this change to all similar constructs in this file? Otherwise, LGTM, thanks!

hkaiser:
bors merge

Issue #6246, opened 2023-05-17
HPX fails to compile under cxx 20 (fresh system)


SAtacker:
## Expected Behavior

HPX should be able to compile with CXX 20 standard (CXX 17 compiles fine). However it gave me this error 2nd time on a fresh system.

## Actual Behavior

Please find the below error log: [complete log](https://pastebin.com/RecxFDma)
~~~
In file included from /home/satacker/workspace/hpx/libs/core/affinity/src/parse_affinity_options.cpp:10:
In file included from /home/satacker/workspace/hpx/libs/core/datastructures/include/hpx/datastructures/tuple.hpp:100:
In file included from /home/satacker/workspace/hpx/build/libs/core/type_support/include/hpx/modules/type_support.hpp:32:
/home/satacker/workspace/hpx/libs/core/type_support/include/hpx/type_support/generator.hpp:327:21: error: no matching function for call to 'operator delete[]'
                    ::operator delete[](
                    ^~~~~~~~~~~~~~~~~~~
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/new:146:6: note: candidate function not viable: no known conversion from 'unsigned long' to 'const std::nothrow_t' for 2nd argument
void operator delete[](void*, const std::nothrow_t&) _GLIBCXX_USE_NOEXCEPT
     ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/new:161:6: note: candidate function not viable: no known conversion from 'unsigned long' to 'std::align_val_t' for 2nd argument
void operator delete[](void*, std::align_val_t)
     ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/new:181:13: note: candidate function not viable: no known conversion from 'unsigned long' to 'void *' for 2nd argument
inline void operator delete[](void*, void*) _GLIBCXX_USE_NOEXCEPT { }
            ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/new:132:6: note: candidate function not viable: requires 1 argument, but 2 were provided
void operator delete[](void*) _GLIBCXX_USE_NOEXCEPT
     ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/new:163:6: note: candidate function not viable: requires 3 arguments, but 2 were provided
[...]

PR #6257, opened 2023-05-25
Cmake Tests: Delete operator check for size_t arg

SAtacker:
- Clang fails to recognize the global delete operator which has 2 arguments the second one being std::size_t
- Reference - https://godbolt.org/z/q1hMrfKjv
- Fixes https://github.com/STEllAR-GROUP/hpx/issues/6246


libs/core/type_support/include/hpx/type_support/generator.hpp
```diff
@@ -324,8 +324,13 @@ namespace hpx {
 
                 dealloc_fn const dealloc = [](void* const p,
                                                std::size_t const s) {
+#if defined(HPX_WITH_CXX14_DELETE_OPERATOR_WITH_SIZE)
                     ::operator delete[](
                         p, s + sizeof(std::size_t) + sizeof(dealloc_fn));
+#else
+                    ::operator delete[](p);
+                    (void) s;
+#endif
```
hkaiser, line 333:
```suggestion
                dealloc_fn const dealloc = [](void* const p,
                    [[maybe_unused]] std::size_t const s) {
#if defined(HPX_WITH_CXX14_DELETE_OPERATOR_WITH_SIZE)
                    ::operator delete[](
                        p, s + sizeof(std::size_t) + sizeof(dealloc_fn));
#else
                    ::operator delete[](p);
#endif
```

SAtacker: @hkaiser maybe_unused is since c++17, what do you suggest?

SAtacker: Ohkay, it'll work out because of the previous checks. Thanks!

hkaiser (commented):
@SAtacker please fix the clang-format and inspect issues. Otherwise, LGTM! Thanks!

hkaiser:
bors merge

PR #6259, opened 2023-05-25
Be consistent about weak symbols

hkaiser:


hkaiser:
This apparently causes segmentation faults, closing...

PR #6264, opened 2023-05-28
hpx::is_trivially_relocatable trait implementation

isidorostsa:
This PR deals with the implementation of the `is_trivially_relocatable` trait. It offers the user an interface to declare their custom class as `trivially_relocatable` using the macro:

`HPX_DECLARE_TRIVIALLY_RELOCATABLE(<class name>)`

This assumes that trivially_copyable types are also trivially relocatable, following Folly's convention: https://github.com/facebook/folly/blob/ec297b748575e8ab86333899295715e6e85f909d/folly/Traits.h#L542


libs/core/algorithms/include/hpx/algorithms/traits/is_relocatable.hpp
isidorostsa, line 1: This was removed as move_constructible implies destructible

libs/core/algorithms/include/hpx/algorithms/traits/is_trivially_relocatable.hpp
```diff
@@ -0,0 +1,53 @@
...
+    // All pointer types are trivially relocatable
+    template <typename T>
+    struct is_trivially_relocatable<T*> : std::true_type
+    {
+    };
+
+    // All iterator types are trivially relocatable
+    template <typename T>
+    struct is_trivially_relocatable<T,
+        typename std::enable_if<traits::is_iterator<T>::value>::type>
```
hkaiser, line 37: Please use the `_t` and `_v` variations of the traits, wherever possible. I.e.
```suggestion
        std::enable_if_t<traits::is_iterator_v<T>>>
```

libs/core/algorithms/include/hpx/algorithms/traits/is_trivially_relocatable.hpp
```diff
@@ -0,0 +1,53 @@
...
+    struct is_trivially_relocatable<T,
+        std::enable_if_t<traits::is_iterator_v<T>>>
+      : std::true_type
+    {
+    };
+
+    // All trivially copyable types are trivially relocatable
+    template <typename T>
+    struct is_trivially_relocatable<T,
+        typename std::enable_if<std::is_trivially_copyable<T>::value>::type>
```
hkaiser, line 45:
```suggestion
        std::enable_if_t<std::is_trivially_copyable_v<T>>>
```
?

isidorostsa: Oops, missed that!

libs/core/algorithms/include/hpx/algorithms/traits/is_trivially_relocatable.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+    template <typename T, typename = void>
+    struct is_trivially_relocatable : std::false_type
+    {
+    };
+
+    // All trivially copyable types are trivially relocatable
+    template <typename T>
+    struct is_trivially_relocatable<T,
+        typename std::enable_if_t<std::is_trivially_copyable_v<T>>>
```
hkaiser, line 31: You don't need the `typename` anymore:
```suggestion
        std::enable_if_t<std::is_trivially_copyable_v<T>>>
```
;-)

libs/core/algorithms/tests/unit/algorithms/util/test_is_trivially_relocatable.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+struct NotTriviallyCopyable_2
+{
+    NotTriviallyCopyable_2(NotTriviallyCopyable_2 const&){};
+};
+HPX_DECLARE_TRIVIALLY_RELOCATABLE(NotTriviallyCopyable_2)
+
+static_assert(hpx::is_trivially_relocatable_v<NotTriviallyCopyable_2>,
+    "Not Trivially Copyable but declared Trivially Relocatable type should "
+    "be Trivially Relocatable");
```
Quuxplusone, line 37: Here I recommend
```
struct ExplicitlyTriviallyRelocatable1 {
    ExplicitlyTriviallyRelocatable1();
    ExplicitlyTriviallyRelocatable1(const ExplicitlyTriviallyRelocatable1&);
    ExplicitlyTriviallyRelocatable1& operator=(const ExplicitlyTriviallyRelocatable1&);
    ~ExplicitlyTriviallyRelocatable1();
};
HPX_DECLARE_TRIVIALLY_RELOCATABLE(ExplicitlyTriviallyRelocatable1);

struct ExplicitlyTriviallyRelocatable2 {
    ExplicitlyTriviallyRelocatable2();
    ExplicitlyTriviallyRelocatable2(ExplicitlyTriviallyRelocatable2&&);
    ExplicitlyTriviallyRelocatable2& operator=(ExplicitlyTriviallyRelocatable2&&);
    ~ExplicitlyTriviallyRelocatable2();
};
HPX_DECLARE_TRIVIALLY_RELOCATABLE(ExplicitlyTriviallyRelocatable2);

struct DerivedFromExplicitlyTriviallyRelocatable : ExplicitlyTriviallyRelocatable1 {
    ~DerivedFromExplicitlyTriviallyRelocatable();
};

static_assert(hpx::is_trivially_relocatable_v<ExplicitlyTriviallyRelocatable1>);
static_assert(hpx::is_trivially_relocatable_v<ExplicitlyTriviallyRelocatable2>);
static_assert(!hpx::is_trivially_relocatable_v<DerivedFromExplicitlyTriviallyRelocatable>);
```

libs/core/algorithms/tests/unit/algorithms/util/test_is_trivially_relocatable.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+// Standard library types are trivially relocatable
```
Quuxplusone, line 39: nit: this comment says "standard library types" but you don't test any; you test only primitive (core-language) types

libs/core/algorithms/tests/unit/algorithms/util/test_is_trivially_relocatable.cpp
```diff
@@ -0,0 +1,55 @@
...
+static_assert(hpx::is_trivially_relocatable_v<void*>,
+    "void* should be Trivially Relocatable");
+static_assert(hpx::is_trivially_relocatable_v<int*>,
+    "int* should be Trivially Relocatable");
+static_assert(hpx::is_trivially_relocatable_v<double*>,
+    "double* should be Trivially Relocatable");
+static_assert(hpx::is_trivially_relocatable_v<char*>,
+    "char* should be Trivially Relocatable");
+
+int main(int, char*[]) {}
```
Quuxplusone, line 55: I'd also add
```
struct Polymorphic {
    virtual int f();
};
static_assert(!hpx::is_trivially_relocatable_v<Polymorphic>);
```

libs/core/algorithms/include/hpx/algorithms/traits/is_relocatable.hpp
```diff
@@ -13,8 +13,7 @@ namespace hpx {
     template <typename T>
     struct is_relocatable
     {
-        static constexpr bool value =
-            std::is_move_constructible_v<T> && std::is_destructible_v<T>;
+        static constexpr bool value = std::is_move_constructible_v<T>;
     };
```
Quuxplusone, line 17: Almost certainly you should make this
```
template <class T>
struct is_relocatable : std::is_move_constructible<T> {};
```
or
```
template <class T>
struct is_relocatable : std::bool_constant<std::is_move_constructible_v<T>> {};
```
so that `is_relocatable<X>` can be used for tag dispatch.

hkaiser:
@Quuxplusone thanks for your comments! @isidorostsa if you apply the suggested changes, please stick with our naming conventions, though.

libs/core/algorithms/include/hpx/algorithms/traits/is_trivially_relocatable.hpp
isidorostsa: Thanks :D

libs/core/algorithms/include/hpx/algorithms/traits/is_relocatable.hpp
isidorostsa: makes sense

libs/core/algorithms/tests/unit/algorithms/util/test_is_trivially_relocatable.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+// Non trivially copyable types are not trivially relocatable
+// Unless they are explicitly declared as such
+struct NotTriviallyCopyable_1
+{
+    NotTriviallyCopyable_1(NotTriviallyCopyable_1 const&){};
+};
+static_assert(!hpx::is_trivially_relocatable_v<NotTriviallyCopyable_1>,
+    "Not Trivially Copyable and Not declared Trivially Relocatable type should "
+    "not be Trivially Relocatable");
```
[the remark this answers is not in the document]
isidorostsa: I see, what could a trivially relocatable type do in the assignment operator to break swap?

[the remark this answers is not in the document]
Quuxplusone: Literally anything other than "assign the value of the right-hand side to the left-hand side." I'm actually planning a blog post for this week or next, on all the things that potentially go wrong when we treat `std::pmr::string` as trivially relocatable.

```
struct Widget {
    std::pmr::string ps;
};

std::vector<Widget> w;
w.push_back(Widget{ std::pmr::string("hello"), &mr1) });
w.push_back(Widget{ std::pmr::string("world"), &mr2) });
w.erase(w.begin()); // ideally should destroy w[0] and relocate downward, right?
assert(w[0].ps.get_allocator().resource() == &mr1); // but today this is required to be true

auto v1 = std::pmr::vector<std::pmr::string>({"hello", "hello"}, &mr1);
auto v2 = std::pmr::vector<std::pmr::string>({"world", "world"}, &mr2);
std::swap_ranges(v1.begin(), v1.end(), v2.begin()); // ideally should swap the bytes, right?
assert(v1[0].get_allocator().resource() == &mr1); // but today this is required to be true
```

There are several theories of what to do here:
- Don't call `std::pmr::string` trivially relocatable at all. (This is the default, implied by the current wording of P1144R8.)
- Call `std::pmr::string` trivially relocatable, and put a precondition on _all_ standard library facilities that says you're not allowed to pass them objects from different arenas; if you do, that's library UB. (This is my own preferred theory.)
- Call `std::pmr::string` trivially relocatable, and put a precondition on _some_ standard library facilities that says you're not allowed to pass them objects from different arenas; if you do, that's library UB. For example, let's say `vector<Widget>::erase` gets to assume that all the `Widget`s in the vector have the same arena, but `std::swap<Widget&>` does not, and thus simply cannot be optimized for trivially relocatable types. (This is the Bloomberg/P2786R1 theory, as of last week, although it might change in the future.)

isidorostsa:
> @Quuxplusone thanks for your comments! @isidorostsa if you apply the suggested changes, please stick with our naming conventions, though.

Thanks for bringing that up, I looked and found that we require snake_case C++ STL style, is there something else I should keep in mind?

hkaiser:
> > @Quuxplusone thanks for your comments! @isidorostsa if you apply the suggested changes, please stick with our naming conventions, though.
> 
> Thanks for bringing that up, I looked and found that we require snake_case C++ STL style, is there something else I should keep in mind?

Please see [here](https://github.com/STEllAR-GROUP/hpx/blob/master/.github/CONTRIBUTING.md#general-info) for the whole story.

libs/core/algorithms/tests/unit/algorithms/util/test_is_trivially_relocatable.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include <cassert>
+
+// Trivially Copyable types are trivially relocatable
+struct TriviallyCopyable
+{
+    TriviallyCopyable(TriviallyCopyable const&) = default;
+};
+static_assert(hpx::is_trivially_relocatable_v<TriviallyCopyable>,
+    "Trivially Copyable type should be Trivially Relocatable");
```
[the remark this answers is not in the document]
isidorostsa: That's a nice convention, is there some significance in declaring the assignment operator void vs the class's type in the last example?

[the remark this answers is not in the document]
Quuxplusone: > is there some significance in declaring the assignment operator void vs the class's type

No, just laziness on my part. I don't care about that part either way. :)

libs/core/algorithms/tests/unit/algorithms/util/test_is_trivially_relocatable.cpp
isidorostsa: From my understanding, the primary difference between 1 and 2 is that 1 is only **copy** assignable/constructible while 2 is only **move** assignable/constructible, with non-trivial functions for everything. Does each structure represent a specific edge case scenario that the tests need to cover? I'm asking this to better understand the principles guiding the selection process for future test cases.

Quuxplusone: 1 is both copyable and movable; it's a classic C++98 "Rule of Three" type, which you'll probably see a lot of in real code. The fact that when you move it you get a call to the copy constructor (not a separate move constructor) doesn't really matter to me. The point is just that it's not Rule-of-Zero and it isn't move-only.
2 is move-only, exactly isomorphic to `unique_ptr`.

The point of `DerivedFromExplicitlyTriviallyRelocatable` is that back when I was first thinking about "The Best Type Traits C++ Doesn't Have," I considered the idea of opting into things like trivial relocatability and trivial comparability by using a member typedef:
```
struct S {
    using is_trivially_relocatable = void;
};
```
(The STL uses this technique for [`is_transparent`](https://stackoverflow.com/questions/20317413/what-are-transparent-comparators).) The downside to this is that then anything _derived_ from `S` _inherits_ that member typedef — i.e., anything derived from a trivially relocatable type ends up being marked as trivially relocatable — which obviously isn't going to work.

Now, neither P1144 nor HPX are actually doing anything at all with member typedefs, so we don't expect anything at all to go wrong with `DerivedFromExplicitlyTriviallyRelocatable`. But I figured it couldn't hurt to add the test case for it anyway.

isidorostsa: I see, both the tests and the rationale behind not using typedefs make sense, thank you :)

libs/core/algorithms/include/hpx/algorithms/traits/is_trivially_relocatable.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <hpx/iterator_support/traits/is_iterator.hpp>
+#include <type_traits>
+
+// Macro to specialize template for given type
+#define HPX_DECLARE_TRIVIALLY_RELOCATABLE(T)                                   \
+    namespace hpx {                                                            \
+        template <>                                                            \
+        struct is_trivially_relocatable<T> : std::true_type                    \
+        {                                                                      \
```
isidorostsa, line 17: last minute thought: adding static_assert(is_relocatable<T>) ; as it is impossible to be trivially_relocatable but not relocatable.

Quuxplusone: That's what I thought as of P1144R6; but the current revision (P1144R8) adopts the viewpoint that "trivially relocatable" is like "trivially copyable." It implies, "I promise that **hypothetically when** you (copy|relocate|compare) this thing, you can do it trivially." It doesn't imply anything about whether it is **actually okay** to (copy|relocate|compare) the thing; that's up to the algorithm's author to check.

E.g.
```
struct A { int i; A(A&&) = delete; A& operator=(A&&) = delete; };
static_assert(std::is_trivially_copyable_v<A>);  // C++ today
static_assert(std::is_trivially_relocatable_v<A>);  // in P1144R8
A src[10];
A dst[10];
alignas(A) char rawdst[10*sizeof(A)];
std::copy(src, src+10, dst);  // invalid C++ today
std::uninitialized_move(src, src+10, (A*)rawdst);  // also invalid C++ today
std::uninitialized_relocate(src, src+10, (A*)rawdst);  // also invalid in P1144R8
```

...however, I see that I need to update my libc++ implementation to actually reject the latter. ;)

Anyway, I don't have a strong opinion on which way HPX should define their trait. If you want to make it false for non-relocatable types, that's not crazy; it won't explode or anything. But I did want to point out _that_ P1144R8 chooses to do it differently, and _why_ P1144R8 chooses to do it differently.

isidorostsa: I see, I am struggling to think of a case where using std::relocate on a trivially relocatable but not relocatable type would be correct. I think I must be missing something

Quuxplusone: > I am struggling to think of a case where using [std::uninitialized_relocate] on a trivially relocatable but not relocatable type would be correct.

That's the point: it'd never be correct. And neither would it be correct to use `std::copy` on a trivially copyable but not copy-assignable type. That doesn't mean such a type is _not_ trivially copyable (or _not_ trivially relocatable); it just means that you can't use it with those algorithms because they require copy-assignability resp. relocatability.

The  _interface constraint_ on `std::copy` is that the type you pass in needs to be `is_copy_assignable`. Internally, the _optimization into a memcpy_ kicks in when the type happens to be `is_trivially_copyable`.

The _interface constraint_ on `std::uninitialized_relocate` is that the type you pass in needs to be `is_relocatable`. Internally, the _optimization into a memcpy_ kicks in when the type happens to be `is_trivially_relocatable`.

isidorostsa: Ohhh I see, so it's just a matter of where this check should be performed, R8 suggests it should not throw a compile error when you give a non-relocatable type this trait, but it should when we try to relocate it, right? And this is checked with a static assertion inside the relocation algorithms

Quuxplusone: Yes, I believe you've got it now. Here's a concrete example with libc++: https://godbolt.org/z/Yr9MdncoT
```
struct S {
    const std::unique_ptr<int> m;
};

static_assert(std::is_trivially_relocatable_v<S>);
static_assert(!std::is_relocatable_v<S>);

void f(S *p) {
    std::relocate_at(p, p); // ill-formed
}

__memory/relocate_at.h:91:19: error: static assertion failed [...]
```
If you _could_ relocate `S`, then you could do so trivially. But in fact you can't relocate `S`.

libs/core/algorithms/include/hpx/algorithms/traits/is_relocatable.hpp
```diff
@@ -11,10 +11,8 @@
 namespace hpx {
 
     template <typename T>
-    struct is_relocatable
+    struct is_relocatable : std::bool_constant<std::is_move_constructible_v<T>>
```
hkaiser, line 14:
```suggestion
    struct is_relocatable : std::is_move_constructible<T>
```

hkaiser (changes requested):
Let's have these low-level facilities in `core/type_support` after all (as discussed). Otherwise we will not be able to use this for the implementation of our own data types (`small_vector`, etc.).

libs/core/type_support/include/hpx/type_support/is_trivially_relocatable.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+    template <typename T, typename = void>
+    struct is_trivially_relocatable : std::false_type
+    {
+    };
+
+    // All trivially copyable types are trivially relocatable
+    template <typename T>
+    struct is_trivially_relocatable<T,
+        std::enable_if_t<std::is_trivially_copyable_v<T>>> : std::true_type
```
hkaiser, line 30: Do we have to disable this specialization if `T` was marked as trivially relocatable (by specializing `is_trivially_relocatable<T>` where `T` is also trivially copyable? 

Alterntively, we could change the macro above to:
```
#define HPX_DECLARE_TRIVIALLY_RELOCATABLE(T)                                   \
    namespace hpx {                                                            \
        template <>                                                            \
        struct is_trivially_relocatable<T,                                     \
            std::enable_if_t<!std::is_trivially_copyable_v<T>>>                \
          : std::true_type                                                     \
        {                                                                      \
        };                                                                     \
    }
```

libs/core/type_support/tests/unit/is_trivially_relocatable.cpp
```diff
@@ -0,0 +1,155 @@
...
+    explicitly_trivially_relocatable_2() = default;
+    explicitly_trivially_relocatable_2(explicitly_trivially_relocatable_2&&);
+    explicitly_trivially_relocatable_2& operator=(
+        explicitly_trivially_relocatable_2&&);
+    ~explicitly_trivially_relocatable_2() = default;
+};
+
+HPX_DECLARE_TRIVIALLY_RELOCATABLE(explicitly_trivially_relocatable_1);
+HPX_DECLARE_TRIVIALLY_RELOCATABLE(explicitly_trivially_relocatable_2);
+
```
hkaiser, line 116: Please add a test for a trivially copyable type that is also explicitly marked as trivially relocatable.

libs/core/type_support/include/hpx/type_support/is_trivially_relocatable.hpp
Quuxplusone: Alternatively, just change the base case to
```
    template <typename T, typename = void>
    struct is_trivially_relocatable : std::is_trivially_copyable<T>
    {
    };
```

isidorostsa: Thank you that's very elegant, I will apply this. Would 

std::enable_if_t<!std::**is_trivially_relocatable_v**<T>>> ...

work as a more general solution? It does look kind of suspicious...

Quuxplusone: No, you don't want to define `is_trivially_relocatable` in terms of itself. ;)
Notice that P1144 can't help with this part of the HPX design, because P1144 proposes that the core language should just know which types are trivially relocatable — P1144's raison d'etre is specifically to _avoid_ this kind of opt-in macro. (But HPX needs this part _because_ P1144's core-language changes aren't in the language yet.)

@isidorostsa: You should definitely take a look at what BSL and Folly are doing with their opt-in macros, and try to evaluate what advantages and disadvantages each approach might have. I think what you're doing with just letting the user specialize `hpx::is_trivially_relocatable<T, enable_if_t<...>>` is fine, but I don't think it is isomorphic to either BSL's or Folly's approach, so it would be good to evaluate, justify, and document.

* [`bslmf::IsBitwiseMoveable`](https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/bslmf__isbitwisemoveable_8h_source.html)
* [`folly::IsRelocatable`](https://github.com/facebook/folly/blob/5c227af84c1379f6f196932ca6b2fcfc7863c8ca/folly/Traits.h#L535)

isidorostsa: Thanks for this note, having examined `bslmf` and `folly` we decided to avoid using a nested trait, due to it being invasive to existing code to work. We did implement more generic macros to compensate for the lack of template and conditional support by the previous macro.

new macros: https://github.com/STEllAR-GROUP/hpx/pull/6264/commits/7139cb18df90897ecf39a1345c30faeb950b9751

libs/core/algorithms/include/hpx/algorithms/traits/is_trivially_relocatable.hpp
isidorostsa: | If you could relocate `S`, then you could do so trivially. But in fact you can't relocate `S`.

Just for a sanity check; arrays of trivially relocatable types `T[]` should be considered trivially relocatable, but not relocatable (as not move constructable)?

Quuxplusone: > Just for a sanity check; arrays of trivially relocatable types T[] should be considered trivially relocatable, but not relocatable (as not move constructable)?

Yes, that's exactly right. https://godbolt.org/z/9n4YGTjfd  (P1144R8 has actually removed `std::is_relocatable_v`, but kept `std::relocatable`. My Godbolt implementation keeps both, at least for now, since they're trivial to implement.)

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6267, opened 2023-05-30
Attempting to fix migration tests

hkaiser:
@Pansysk75 would you mind verifying if this fixes the issues?

### DIFF
--- a/components/component_storage/include/hpx/components/component_storage/server/component_storage.hpp
+++ b/components/component_storage/include/hpx/components/component_storage/server/component_storage.hpp
@@ -23,7 +23,8 @@
 #include <vector>
 
 ///////////////////////////////////////////////////////////////////////////////
-namespace hpx { namespace components { namespace server {
+namespace hpx::components::server {
+
     ///////////////////////////////////////////////////////////////////////////
     class HPX_MIGRATE_TO_STORAGE_EXPORT component_storage
       : public component_base<component_storage>
@@ -48,7 +49,7 @@ namespace hpx { namespace components { namespace server {
     private:
         hpx::unordered_map<naming::gid_type, std::vector<char>> data_;
     };
-}}}    // namespace hpx::components::server
+}    // namespace hpx::components::server
 
 HPX_REGISTER_ACTION_DECLARATION(
     hpx::components::server::component_storage::migrate_to_here_action,
--- a/components/component_storage/include/hpx/components/component_storage/server/migrate_from_storage.hpp
+++ b/components/component_storage/include/hpx/components/component_storage/server/migrate_from_storage.hpp
@@ -23,7 +23,8 @@
 #include <utility>
 #include <vector>
 
-namespace hpx { namespace components { namespace server {
+namespace hpx::components::server {
+
     ///////////////////////////////////////////////////////////////////////////
     // Migrate given component from the specified storage component
     //
@@ -58,6 +59,7 @@ namespace hpx { namespace components { namespace server {
     //       scope.
     //
     namespace detail {
+
         ///////////////////////////////////////////////////////////////////////
         template <typename Component>
         future<hpx::id_type> migrate_from_storage_here_id(
@@ -101,9 +103,6 @@ namespace hpx { namespace components { namespace server {
                 archive >> ptr;
             }
 
-            // make sure the migration code works properly
-            traits::component_pin_support<Component>::pin(ptr.get());
-
             // if target locality is not specified, use the address of the last
             // locality where the object was living before
             if (target_locality == hpx::invalid_id)
@@ -126,13 +125,12 @@ namespace hpx { namespace components { namespace server {
     future<hpx::id_type> trigger_migrate_from_storage_here(
         hpx::id_type const& to_resurrect, hpx::id_type const& target_locality)
     {
-        if (!traits::component_supports_migration<Component>::call())
+        if constexpr (!traits::component_supports_migration<Component>::call())
         {
             HPX_THROW_EXCEPTION(hpx::error::invalid_status,
                 "hpx::components::server::trigger_migrate_from_storage_here",
                 "attempting to migrate an instance of a component which "
                 "does not support migration");
-            return make_ready_future(hpx::invalid_id);
         }
 
         if (naming::get_locality_id_from_id(to_resurrect) != get_locality_id())
@@ -141,22 +139,21 @@ namespace hpx { namespace components { namespace server {
                 "hpx::components::server::trigger_migrate_from_storage_here",
                 "this function has to be executed on the locality responsible "
                 "for managing the address of the given object");
-            return make_ready_future(hpx::invalid_id);
         }
 
-        auto r = agas::begin_migration(to_resurrect).get();
+        auto const r = agas::begin_migration(to_resurrect).get();
 
         // retrieve the data from the given storage
-        typedef typename server::component_storage::migrate_from_here_action
-            action_type;
+        using action_type =
+            typename server::component_storage::migrate_from_here_action;
+
         return async<action_type>(r.first, to_resurrect.get_gid())
             .then(hpx::bind_back(&detail::migrate_from_storage_here<Component>,
                 to_resurrect, r.second, target_locality))
-            .then(
-                [to_resurrect](future<hpx::id_type>&& f) -> hpx::id_type {
-                    agas::end_migration(to_resurrect);
-                    return f.get();
-                });
+            .then([to_resurrect](future<hpx::id_type>&& f) -> hpx::id_type {
+                agas::end_migration(to_resurrect);
+                return f.get();
+            });
     }
 
     template <typename Component>
@@ -167,4 +164,4 @@ namespace hpx { namespace components { namespace server {
             trigger_migrate_from_storage_here_action<Component>>
     {
     };
-}}}    // namespace hpx::components::server
+}    // namespace hpx::components::server
--- a/components/component_storage/include/hpx/components/component_storage/server/migrate_to_storage.hpp
+++ b/components/component_storage/include/hpx/components/component_storage/server/migrate_to_storage.hpp
@@ -12,7 +12,6 @@
 #include <hpx/modules/errors.hpp>
 #include <hpx/naming_base/address.hpp>
 #include <hpx/naming_base/id_type.hpp>
-#include <hpx/type_support/unused.hpp>
 
 #include <hpx/components/component_storage/export_definitions.hpp>
 #include <hpx/components/component_storage/server/component_storage.hpp>
@@ -22,7 +21,8 @@
 #include <utility>
 #include <vector>
 
-namespace hpx { namespace components { namespace server {
+namespace hpx::components::server {
+
     ///////////////////////////////////////////////////////////////////////////
     // Migrate given component to the component storage
     //
@@ -57,6 +57,7 @@ namespace hpx { namespace components { namespace server {
     //       scope.
     //
     namespace detail {
+
         // clean up (source) memory of migrated object
         template <typename Component>
         hpx::id_type migrate_to_storage_here_cleanup(future<hpx::id_type> f,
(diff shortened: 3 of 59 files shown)


Pansysk75:
@hkaiser  Still failing occasionally, I'll investigate further.

Pansysk75:
@hkaiser 
https://github.com/STEllAR-GROUP/hpx/blob/897782934884f06975274db76029fa82cc01048e/libs/full/actions_base/include/hpx/actions_base/traits/action_was_object_migrated.hpp#L23-L44
Could you verify that the template specialization of the 2nd function is correct? It looks like it's never taken, so action_was_object_migrated just reports nonsense.

Pansysk75:
I think I found why the above issue occurs:
In the test, we have
https://github.com/STEllAR-GROUP/hpx/blob/897782934884f06975274db76029fa82cc01048e/libs/full/runtime_components/tests/unit/components/migrate_polymorphic_component.cpp#L26-L29
All actions are declared refering to this component, which doesn't t have a was_object_migrated(...) function, nor does it inherit one.

There are other issues at play as well, but fixing this should patch a large hole.

hkaiser:
> I think I found why the above issue occurs: In the test, we have
> 
> https://github.com/STEllAR-GROUP/hpx/blob/897782934884f06975274db76029fa82cc01048e/libs/full/runtime_components/tests/unit/components/migrate_polymorphic_component.cpp#L26-L29
> 
> 
> All actions are declared refering to this component, which doesn't t have a was_object_migrated(...) function, nor does it inherit one.
> There are other issues at play as well, but fixing this should patch a large hole.

This function is defined by the base class: https://github.com/STEllAR-GROUP/hpx/blob/897782934884f06975274db76029fa82cc01048e/libs/full/components_base/include/hpx/components_base/server/abstract_migration_support.hpp#L115-L119

hkaiser:
> I think I found why the above issue occurs: In the test, we have
> 
> https://github.com/STEllAR-GROUP/hpx/blob/897782934884f06975274db76029fa82cc01048e/libs/full/runtime_components/tests/unit/components/migrate_polymorphic_component.cpp#L26-L29
> 
> 
> All actions are declared refering to this component, which doesn't t have a was_object_migrated(...) function, nor does it inherit one.
> There are other issues at play as well, but fixing this should patch a large hole.

Uhh. let me have a closer look.

Pansysk75:
> Uhh. let me have a closer look.

Yup, the base `test_server_base ` class only inherits `abstract_base_migration_support`, not `abstract_migration_support`.
At this point, I'm not even sure what the implementation is meant to do :P

hkaiser:
> > Uhh. let me have a closer look.
> 
> Yup, the base `test_server_base ` class only inherits `abstract_base_migration_support`, not `abstract_migration_support`. At this point, I'm not even sure what the implementation is meant to do :P

I have that fixed locally now, thanks for spotting this!

hkaiser:
@Pansysk75 I'm hopeful to have things resolved now. Would you mind trying it out?

hkaiser:
Things still fail :/ I must have missed changing one of the spots.

hkaiser:
@Pansysk75 are you able to reproduce the hangs? Everything works locally for me.

Pansysk75:
> @Pansysk75 are you able to reproduce the hangs? Everything works locally for me.

I had some issues regarding shutdown, I'll check again and let you know asap

Pansysk75:
@hkaiser 
Umm is it ok for `free_components_sync` to run on the component while the `get_base_data_action` thread hasn't been terminated yet?
Dijkstra gets stuck, probably because that thread never terminates.

Not familiar with that part of the code, i suppose the credits are analogous to some sort of reference counting?

```
<debug>  [TM] Executing component action(get_base_data_action) lva(000001A4783AA370) with continuation({0000000100807808, 000001e055282a78})
<debug>  [TM] scheduling_loop state change: pool(default(0)), ..., description(get_base_data_action), old state(active), new state(pending)
..........................................................omitted........................................................................
<debug>  [TM] Executing component action(primary_namespace_decrement_credit_action) lva(000001A47679B830).
 <info>[AGAS] primary_namespace::decrement_sweep, lower({0000000200000001, 0000000000001001}), upper({0000000200000001, 0000000000001002}), credits(16777216)
 <info>[AGAS] primary_namespace::resolve_free_list, resolved match, gid({0000000200000001, 0000000000001001}), gva(({0000000200000000, 0000000000000000} component[36(37)] 1 000001A4783AA370 0))
 <info>[AGAS] primary_namespace::free_components_sync, freeing component, lower({0000000200000001, 0000000000001001}), upper({0000000200000001, 0000000000001002}), base({0000000200000001, 0000000000001001}), gva(({0000000200000000, 0000000000000000} component[36(37)] 1 000001A4783AA370 0)), locality({0000000200000000, 0000000000000000})
 <info>[AGAS] primary_namespace::unbind_gid, gid({0000000200000001, 0000000000001001}), count(1), gva(({0000000200000000, 0000000000000000} component[36(37)] 1 000001A4783AA370 0)), locality_id({0000000200000000, 0000000000000000})
<debug>  [TM] scheduling_loop state change: pool(default(0)), ..., description(primary_namespace_decrement_credit_action), old state(active), new state(terminated)
<debug>  [TM] thread_data::destroy_thread(000001A478280720), description(primary_namespace_decrement_credit_action), phase(0)
<debug>  [TM] scheduling_loop state change: pool(default(0)), ... description(get_base_data_action), old state(pending), new state(active)
 <info>  [TM] create_work: pool(default(0)), scheduler(core-local_priority_queue_scheduler(000001A4781D36B0)), initial_state(pending), thread_priority(default), description(dijkstra_termination_action)
<debug>  [TM] local_priority_queue_scheduler::create_thread, normal priority queue: pool(default(0)), scheduler(core-local_priority_queue_scheduler(000001A4781D36B0)), worker_thread(0), thread(0000000000000000), priority(normal (2)), description(dijkstra_termination_action)
```

hkaiser:
> @hkaiser Umm is it ok for `free_components_sync` to run on the component while the `get_base_data_action` thread (which is referring to that component) hasn't been terminated yet? Dijkstra gets stuck, probably because that thread never terminates.
> 
> Not familiar with that part of the code, i suppose the credits are analogous to some sort of reference counting?
> 
> ```
> <debug>  [TM] Executing component action(get_base_data_action) lva(000001A4783AA370) with continuation({0000000100807808, 000001e055282a78})
> <debug>  [TM] scheduling_loop state change: pool(default(0)), ..., description(get_base_data_action), old state(active), new state(pending)
> ..........................................................omitted........................................................................
> <debug>  [TM] Executing component action(primary_namespace_decrement_credit_action) lva(000001A47679B830).
>  <info>[AGAS] primary_namespace::decrement_sweep, lower({0000000200000001, 0000000000001001}), upper({0000000200000001, 0000000000001002}), credits(16777216)
>  <info>[AGAS] primary_namespace::resolve_free_list, resolved match, gid({0000000200000001, 0000000000001001}), gva(({0000000200000000, 0000000000000000} component[36(37)] 1 000001A4783AA370 0))
>  <info>[AGAS] primary_namespace::free_components_sync, freeing component, lower({0000000200000001, 0000000000001001}), upper({0000000200000001, 0000000000001002}), base({0000000200000001, 0000000000001001}), gva(({0000000200000000, 0000000000000000} component[36(37)] 1 000001A4783AA370 0)), locality({0000000200000000, 0000000000000000})
>  <info>[AGAS] primary_namespace::unbind_gid, gid({0000000200000001, 0000000000001001}), count(1), gva(({0000000200000000, 0000000000000000} component[36(37)] 1 000001A4783AA370 0)), locality_id({0000000200000000, 0000000000000000})
> <debug>  [TM] scheduling_loop state change: pool(default(0)), ..., description(primary_namespace_decrement_credit_action), old state(active), new state(terminated)
> <debug>  [TM] thread_data::destroy_thread(000001A478280720), description(primary_namespace_decrement_credit_action), phase(0)
> <debug>  [TM] scheduling_loop state change: pool(default(0)), ... description(get_base_data_action), old state(pending), new state(active)
>  <info>  [TM] create_work: pool(default(0)), scheduler(core-local_priority_queue_scheduler(000001A4781D36B0)), initial_state(pending), thread_priority(default), description(dijkstra_termination_action)
> <debug>  [TM] local_priority_queue_scheduler::create_thread, normal priority queue: pool(default(0)), scheduler(core-local_priority_queue_scheduler(000001A4781D36B0)), worker_thread(0), thread(0000000000000000), priority(normal (2)), description(dijkstra_termination_action)
> ```

I don't think this is related to the hangs in the migration tests, but what do I know :/ In any case, it's difficult to tell what goes wrong for you.

Pansysk75:
> I don't think this is related to the hangs in the migration tests, but what do I know :/ In any case, it's difficult to tell what goes wrong for you.

@hkaiser
Is my assumption correct that the credit mechanism is what we use to keep track of component references?
I'll look into it, but I could really use some pointer as to what those AGAS messages refer to.

hkaiser:
> > I don't think this is related to the hangs in the migration tests, but what do I know :/ In any case, it's difficult to tell what goes wrong for you.
> 
> @hkaiser The shutdown issue is only occasional, when there are actions called on a migrated (not on its original locality) component right before shutdown. 

Ok

> Is my assumption correct that the credit mechanism is what we use to keep track of component references? I'll look into it, but I could really use some pointer as to what those AGAS messages refer to.

Yes. The credits are global reference counts. Here is some explanation: https://github.com/STEllAR-GROUP/hpx/blob/3d95d662eb6879fccfbc8090377e58674fbc91a7/libs/full/agas_base/include/hpx/agas_base/server/primary_namespace.hpp#L46-L106. Also, this thesis has the full story: https://digitalcommons.lsu.edu/cgi/viewcontent.cgi?article=6446&context=gradschool_dissertations

Pansysk75:
I have a testing setup with 2 localities (N0 and N1).
A component was created on N0, and then was migrated to N1.

I see that when N0 terminates, it triggers garbage collection:
`garbage_collect(...) -> send_refcnt_requests(...)`

which triggers `primary_namespace_decrement_credit_action` on N1:
`decrement_credit(...) -> decrement_sweep(...)`  and then
`free_components_sync(...) -> components::deleter(component_type)(gid, addr)`

The deleter is only called when the decrement_sweep causes the component's credit to go down to 0.
The deleter deletes the component:
https://github.com/STEllAR-GROUP/hpx/blob/3d95d662eb6879fccfbc8090377e58674fbc91a7/libs/full/naming/src/credit_handling.cpp#L435-L449

Now here is my concern:

Since the credit system only accounts for "global" references (that is, references of the component in other localities), that checking does not suffice for any local references of the component.
This could potentially lead to someone (ie a component action) referencing the now deleted object.

Does this explanation sound valid for the hangs I'm getting? Am I missing some mechanism at play here?


((
This would be important for fire-and-forget component actions called from N0 to the component located in N1, because these actions could still be running when N0 (main) terminates.
But in my case, it faulted even when waiting on a future on N0 before exiting main(). The future only guarantees that the action has finished executing on N1, but not that the thread associated with that action has been terminated and deleted (the thread probably references that component for decrementing the pin on destruction).
Also apologize for slightly derailing the PR.
))

hkaiser:
the credits account for global reference counts. Local reference counts are managed by the `id_type` that is conceptually behaving like a `shared_ptr`. Copying an `id_type` locally increments the local reference count. if a local reference count drops to zero, then the global credit owned by this instance of the id_type is handed back to AGAS. Only if all credits are handed back to AGAS the component is deleted. Sending an `id_type` to another locality splits the credit of that instance. One half stays with the original instance, the other half is sent over the wire where it will be used as the credit for the new instance of the `id_type`. If splitting of the credit is not possible (no credits are left) it is replenished through AGAS before it is sent.

hkaiser:
> This would be important for fire-and-forget component actions called from N0 to the component located in N1, because these actions could still be running when N0 (main) terminates.

I believe that fire&forget actions send the target `id_type` along with the message (after splitting the credit) such that the target is kept alive long enough. Async action invocations (i.e. those that expect an answer from the remote action) don't split the credit as it is sufficient to keep the target `id_type` alive long enough until the expected return message has been received.

hkaiser:
> But in my case, it faulted even when waiting on a future on N0 before exiting main(). The future only guarantees that the action has finished executing on N1, but not that the thread associated with that action has been terminated and deleted (the thread probably references that component for decrementing the pin on destruction).

@Pansysk75 I have added an assertion in the destructor of the `migration_support` base class that will fire if the object is still pinned while being destructed.

Pansysk75:
> @Pansysk75 I have added an assertion in the destructor of the `migration_support` base class that will fire if the object is still pinned while being destructed.

Yep, this assertion is indeed triggered.
Not sure why, because as you said, if a component has a pin_count_>1 (some component action is scheduled), we expect the corresponding id_type to still be referenced locally (by that component action), and thus not all credits to be returned back to AGAS yet.

This probably is related to the memory leaks I'm seeing when using Clang's LeakSanitizer.

I'm still working on this, but it's a lot to take in. Thanks for understanding. Currently have my focus on the reference counting / credit system.

Pansysk75:
@hkaiser About the CI timeouts, those are just due to too large tests (they take ~25 mins on my machine). Reducing N from 1000 to 50 on the tests should do the trick.

hkaiser:
> @hkaiser About the CI timeouts, those are just due to too large tests (they take ~1hr on my machine). Reducing N from 1000 to 50 on the tests should do the trick.

Duh, you're certainly right. That means that except for your shutdown issue things seem to have been fixed. I think we should go ahead and merge this PR now and try to solve the shutdown problem separately.

hkaiser:
@Pansysk75 I believe I have resolved the shutdown hang you were seeing in my last commit. Please retry, if you have the time.

hkaiser:
bors merge

PR #6265, opened 2023-05-29
Create dependabot.yml

inanc-can:
Dependabot is a valuable tool designed to assist in maintaining up-to-date dependencies. It achieves this by automatically generating pull requests whenever a new version of a dependency becomes available, ensuring a seamless integration without any disruptive changes.

In this pull request, the inclusion of the dependabot.yml file is crucial. This file is responsible for configuring version updates specifically for GitHub Actions, and it directs the updates to the .github/workflows directory.

To fully utilize Dependabot, a maintainer may need to enable it through the following steps: navigate to the settings section, then access the Code security and analysis options. Within this section, there will be three buttons that need to be enabled to activate Dependabot's functionalities.

PR #6268, opened 2023-06-02
Move dependabot.yml into correct directory

hkaiser:
See https://stackoverflow.com/questions/69446872/dependabot-error-githubl1-no-event-triggers-defined-in-on. Fixes issues caused by #6265


hkaiser:
bors merge

PR #6271, opened 2023-06-04
Uninitialised algorithms, move using std::memcpy

Johan511:
In case of triviably copyable types we copy using std::memcpy. Performance improvement is almost 10x


Johan511:
@hkaiser  please review the PR

hkaiser (commented):
Can you replace all the `try { /* do loop */ } catch(...) { /*cleanup */}` with `loop_with_cleanup`?

Johan511:
@hkaiser requested changes have been made, only existing `try-catch` blocks are in loop_with_cleanup* functions

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_default_construct.hpp
```diff
@@ -252,17 +239,17 @@ namespace hpx::parallel {
             }
 
             using partition_result_type = std::pair<FwdIter, FwdIter>;
-
             util::cancellation_token<util::detail::no_data> tok;
```
hkaiser, line 242: can this be removed now?

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_fill.hpp
```diff
@@ -252,11 +245,12 @@ namespace hpx::parallel {
                 partition_result_type>::
                 call(
                     HPX_FORWARD(ExPolicy, policy), first, count,
-                    [value, tok](Iter it, std::size_t part_size) mutable
+                    [value, tok, policy](Iter it, std::size_t part_size) mutable
```
hkaiser, line 248: Is `tok` still used?

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_move.hpp
```diff
@@ -286,21 +260,21 @@ namespace hpx::parallel {
...
             using partition_result_type = std::pair<FwdIter2, FwdIter2>;
 
-            util::cancellation_token<util::detail::no_data> tok;
             return util::partitioner_with_cleanup<ExPolicy,
                 util::in_out_result<Iter, FwdIter2>, partition_result_type>::
                 call(
                     HPX_FORWARD(ExPolicy, policy), zip_iterator(first, dest),
                     count,
-                    [tok](zip_iterator t, std::size_t part_size) mutable
+                    [&policy](zip_iterator t, std::size_t part_size) mutable
```
hkaiser, line 268: In other places you captured `policy` by value. Is there a rationale for capturing by reference here?

libs/core/algorithms/include/hpx/parallel/util/loop.hpp
```diff
@@ -38,23 +39,66 @@ namespace hpx::parallel::util {
...
+                HPX_CONCEPT_REQUIRES_(hpx::traits::is_iterator_v<End>&&
+                        hpx::traits::is_forward_iterator_v<Begin>)>
+            HPX_HOST_DEVICE HPX_FORCEINLINE static Begin call(
+                Begin it, End end, CancelToken& tok, F&& f)
+            {
+                return call(
+                    hpx::execution::seq, it, end, tok, HPX_FORWARD(F, f));
+            }
+
+            template <typename ExPolicy, typename Begin, typename End,
```
hkaiser, line 56: Should we mix and match vectorized and non-vectorized implementations in the same file? For other algorithms we have split those?

Johan511: I personally believe we should try to have both unseq and seq loops in the same file to allow for easier dispatch.

`template <typename ..... , HPX_CONCEPT_REQUIRE(vectorization_condition)>
//vector loop

template <typename ..... , HPX_CONCEPT_REQUIRE(!vectorization_condition)>
//non-vectorised loop
`
seems like a good way to do things according to me and having both in the same file might be simply easier to work with in the future instead of jumping around files.

Would like to hear your opinion too @hkaiser, do you think we should rather have 2 different files and depracate `#include <hpx/parallel/util/loop.hpp>` and `#include <hpx/parallel/unseq/loop.hpp>` in favour of single header which implicitly includes both the files?

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_move.hpp
Johan511: will fix the inconsistency, capturing by reference everywhere is the better choice right?

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_fill.hpp
Johan511: vestigial tok had been used in a couple of places as I couldn't find the appropriate loop without tok, let me look into this one more time. Do you suggest implementing such a loop (without using tok) if it doesn't exist?

libs/core/algorithms/include/hpx/parallel/util/loop.hpp
hkaiser: Having it in one or two files doesn't make a difference indeed. What I really meant was that the way you dispatched between the vectorized and non-vectorized loops involves a two level dispatch mechanism. I'd rather have one dispatch point (which is also used for the datapar algorithm implementations). Currently (without unseq) we dispatch between non-vectorized and datapar implementations using e.g. the `loop_t` customization point. 

Instead of adding another dispatch point behind this using a different mechanism (you're suggesting to SFINAE), I think it would be better to provide `tag_invoke` overloads for e.g. the `loop_t` CPO as done for datapar - for an example see: https://github.com/STEllAR-GROUP/hpx/blob/09eabc71da8f2a150b2e26efe3d52d4163e49072/libs/core/algorithms/include/hpx/parallel/datapar/loop.hpp#L417-L422

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_move.hpp
hkaiser: I think capturing by value is the safest option, especially for the asynchronous variations.

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_fill.hpp
hkaiser: Well, you captured `tok` without using it in the lambda, that's why I asked.

Johan511: my bad, didn't realise that.

libs/core/algorithms/include/hpx/parallel/util/loop.hpp
hkaiser: I'm not saying you should introduce a new customization point. What I meant is that you should use the already existing CPOs by providing appropriate `tag_invoke` overloads that dispatch to your code.

Johan511: I am sorry but I don't seem to understand the issue completely yet

https://github.com/STEllAR-GROUP/hpx/blob/15e89648b5f41002ea4ff41a9b4b42e9c2b700e5/libs/core/algorithms/include/hpx/parallel/util/loop.hpp#L37-L54

Is the concern on why we are calling `call(hpx::execution::seq, ...)` from `call(...)`. If that is the case the reason I kept it is such as to not raise errors is cases where `detail::loop<Iter>().call(...)`. If required I can remove the overload and try to push it to tag_fallback_invoke over here
https://github.com/STEllAR-GROUP/hpx/blob/15e89648b5f41002ea4ff41a9b4b42e9c2b700e5/libs/core/algorithms/include/hpx/parallel/util/loop.hpp#L106-L127

Johan511:
@hkaiser can you clarify what does `HPX_COMPUTE_DEVICE_CODE` signify over here

https://github.com/STEllAR-GROUP/hpx/blob/15e89648b5f41002ea4ff41a9b4b42e9c2b700e5/libs/core/algorithms/include/hpx/parallel/util/loop.hpp#L129-L148

hkaiser:
@Johan511 The code in question fails compiling when using nvcc. The `#else` case adds a workaround.

libs/core/algorithms/include/hpx/parallel/util/loop.hpp
hkaiser: My concern is that you have two sets of overloads for the `detail::loop::call` functions. One for the vectorized and one for the non-vectorized implementation. That means that the execution has to go through two customization points: a) the `loop_t` selecting between `simd` and non-`simd` and b) the mentioned two sets of `call` overloads. I suggest to dispatch only through the `loop_t` customization point object.

hkaiser:
> @hkaiser can you please point out to the lines you are referring to? In most places I have tried not to change the implementations much.
> 
> In some cases where I wrote a function which calls loop_with_cleanup, I have used the loop without any cancellation tokens, I don't see a point to cancellation tokens in this context as the only case we end the loop prematurely is when an error is thrown. This is taken case by `try-catch` and I don't believe it needs any token.
> 
> Is there something I am missing or would you suggest I do away with tokens completely?

So I guess your rationale of removing the cancellation taokens was that those would matter only in case of errors. That means we were slowing down the success path in order to speed up the error path. I agree to remove them in this case.

Johan511:
>  That means we were slowing down the success path in order to speed up the error path.
loop_with_cleanup was checking for tok only at the [start](https://github.com/STEllAR-GROUP/hpx/blob/bb560af53eca65305627372f0b87e6b471a07c1a/libs/core/algorithms/include/hpx/parallel/util/loop.hpp#L903-L908).

Not sure how having tok was speeding up the error path, we were still throwing the error not doing cleanup with token.

hkaiser:
> > That means we were slowing down the success path in order to speed up the error path.
> 
> loop_with_cleanup was checking for tok only at the [start](https://github.com/STEllAR-GROUP/hpx/blob/bb560af53eca65305627372f0b87e6b471a07c1a/libs/core/algorithms/include/hpx/parallel/util/loop.hpp#L903-L908).
> 
> Not sure how having tok was speeding up the error path, we were still throwing the error not doing cleanup with token.

The token was used to stop doing work on the partitions that were not erroneous.

Johan511:
> The token was used to stop doing work on the partitions that were not erroneous.

We don't seem to explicitly cancel the token in the catch block.
[Ref.](https://github.com/STEllAR-GROUP/hpx/blob/bb560af53eca65305627372f0b87e6b471a07c1a/libs/core/algorithms/include/hpx/parallel/util/loop.hpp#L790-L821)

hkaiser:
> > The token was used to stop doing work on the partitions that were not erroneous.
> 
> We don't seem to explicitly cancel the token in the catch block. [Ref.](https://github.com/STEllAR-GROUP/hpx/blob/bb560af53eca65305627372f0b87e6b471a07c1a/libs/core/algorithms/include/hpx/parallel/util/loop.hpp#L790-L821)

As I said, I'm fine with removing the token if we were using it just to speed up the error path without helping the success path.

hkaiser (approved):
Let's get this in now. Thanks a lot!

hkaiser:
bors merge

Johan511:
uninit-copy

Performance numbers for uninit-copy

PR #6272, opened 2023-06-05
set thread name for linux

### LINKED ISSUE #6252: Implement set_thread_name on POSIX systems
`hpx::util::set_thread_name` is currently implemented on Windows only. We should implement the same (or similar functionality) on POSIX systems. See https://stackoverflow.com/a/7989973/269943 for details.

Johan511:
Fixes #6252

## Proposed Changes
set_thread_name for better debugging

## Checklist


libs/core/thread_support/src/set_thread_name.cpp
```diff
@@ -83,16 +81,40 @@ namespace hpx::util {
...
 
+}    // namespace hpx::util
+
+#elif (defined(__linux__))
+
+#include <pthread.h>
+namespace hpx::util { namespace detail {
+    void set_thread_name(char const* thread_name)
+    {
+        pthread_setname_np(pthread_self(), thread_name);
```
hkaiser, line 92: I think this function is not always available. Should we add a feature test for it? This would also enable the new functionality on other POSIX derivatives (BSD?).

hkaiser: > Sure, will do it. Maybe something like a config test sounds good I guess, like the ones in (https://github.com/STEllAR-GROUP/hpx/blob/3d95d662eb6879fccfbc8090377e58674fbc91a7/cmake/tests)

Yes, that's what I meant. FWIW, the feature tests are defined here : https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_AddConfigTest.cmake, and are invoked here: https://github.com/STEllAR-GROUP/hpx/blob/master/cmake/HPX_PerformCxxFeatureTests.cmake

CMakeLists.txt
```diff
@@ -1625,6 +1625,8 @@ hpx_check_for_mm_prefetch(DEFINITIONS HPX_HAVE_MM_PREFETCH)
 
 hpx_check_for_stable_inplace_merge(DEFINITIONS HPX_HAVE_STABLE_INPLACE_MERGE)
 
+hpx_check_for_nameable_threads(DEFINITIONS HPX_WITH_NAMEABLE_THREADS)
```
hkaiser, line 1628: We have the convention that CMake variables are named `HPX_WITH_...` while the corresponding preprocessor constant should be named `HPX_HAVE_...` Thus this should be:
```suggestion
hpx_check_for_nameable_threads(DEFINITIONS HPX_HAVE_NAMEABLE_THREADS)
```

cmake/tests/set_thread_name.cpp
```diff
@@ -0,0 +1,89 @@
...
+{
+    return pthread_getname_np(pthread_self(), buf, len);
+}
+
+#endif
+
+int main()
+{
+    detail_set_thread_name("hpx_thread_1");
+}
```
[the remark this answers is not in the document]
Johan511: Is there any reason we need not to test for windows? Is it guaranteed that the feature always exists on windows?
Is there any harm in an test which is guaranteed to pass?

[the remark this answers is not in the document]
hkaiser: > Is it guaranteed that the feature always exists on windows?

This feature is guaranteed to exist. 

> Is there any harm in an test which is guaranteed to pass?

Feature tests should be as small and focussed on one particular feature as possible. In this case, they should check for the avaialbility of `pthread_setname_np`, nothing else. Similarly, for FreeBSD we should have a separate feature test that checks for `pthread_set_name_np`, etc.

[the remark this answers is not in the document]
hkaiser: I'd suggest creating feature tests not based on platforms but based on the availability of certain (required) functions.

cmake/tests/set_thread_name.cpp
```diff
@@ -0,0 +1,54 @@
...
+    pthread_set_name_np(pthread_self(), thread_name);
+    return;
+}
+
+#endif
+
+int main()
+{
+    detail_set_thread_name("hpx_thread_1");
+}
```
hkaiser, line 54: I think the feature test could be simplified as:
```
//  Copyright (c) 2023 STE||AR Group
//

#include <pthread.h>

int main()
{
    pthread_setname_np(pthread_self(), "");
}
```

CMakeLists.txt
```diff
@@ -1625,6 +1625,8 @@ hpx_check_for_mm_prefetch(DEFINITIONS HPX_HAVE_MM_PREFETCH)
 
 hpx_check_for_stable_inplace_merge(DEFINITIONS HPX_HAVE_STABLE_INPLACE_MERGE)
 
+hpx_check_for_nameable_threads(DEFINITIONS HPX_HAVE_NAMEABLE_THREADS)
```
hkaiser, line 1628:
```suggestion
hpx_check_for_nameable_threads(DEFINITIONS HPX_HAVE_PTHREAD_SETNAME_NP)
```

libs/core/thread_support/src/set_thread_name.cpp
```diff
@@ -83,16 +81,69 @@ namespace hpx::util {
...
+
+namespace hpx::util {
+
     // Set the name of the thread shown in the Visual Studio debugger
     void set_thread_name(char const* thread_name) noexcept
     {
         detail::set_thread_name(thread_name);
 
+#if (defined(WIN32) || defined(_WIN32) || defined(__WIN32__)) &&               \
+    !defined(HPX_MINGW) && defined(HPX_HAVE_NAMEABLE_THREADS)
```
hkaiser, line 142:
```suggestion
#if (defined(WIN32) || defined(_WIN32) || defined(__WIN32__)) &&               \
    !defined(HPX_MINGW)
```

cmake/HPX_AddConfigTest.cmake
```diff
@@ -506,6 +506,15 @@ function(hpx_check_for_cxx20_perfect_pack_capture)
   )
 endfunction()
 
+# ##############################################################################
+function(hpx_check_for_nameable_threads)
+  add_hpx_config_test(
+    HPX_HAVE_NAMEABLE_THREADS
```
hkaiser, line 512:
```suggestion
    HPX_WITH_PTHREAD_SETNAME_NP
```

cmake/HPX_AddConfigTest.cmake
```diff
@@ -506,6 +506,15 @@ function(hpx_check_for_cxx20_perfect_pack_capture)
...
+    SOURCE cmake/tests/set_thread_name.cpp
+    FILE ${ARGN}
+  )
+endfunction()
```
hkaiser, line 516:
```suggestion
function(hpx_check_for_pthread_setname_np)
  add_hpx_config_test(
    HPX_WITH_PTHREAD_SETNAME_NP
    SOURCE cmake/tests/set_thread_name.cpp
    FILE ${ARGN}
  )
endfunction()
```

hkaiser: Please also rename the feature test file itself reflecting that it checks for `pthread_setname_np`.

libs/core/thread_support/src/set_thread_name.cpp
```diff
@@ -139,7 +139,7 @@ namespace hpx::util {
         detail::set_thread_name(thread_name);
 
 #if (defined(WIN32) || defined(_WIN32) || defined(__WIN32__)) &&               \
-    !defined(HPX_MINGW) && defined(HPX_HAVE_NAMEABLE_THREADS)
+    !defined(HPX_MINGW) && defined(HPX_HAVE_PTHREAD_SETNAME_NP)
```
hkaiser, line 142:
```suggestion
    !defined(HPX_MINGW)
```

libs/core/thread_support/src/set_thread_name.cpp
```diff
@@ -99,7 +99,7 @@ namespace hpx::util { namespace detail {
     }
 }}    // namespace hpx::util::detail
 
-#elif defined(__NetBSD__) && defined(HPX_HAVE_NAMEABLE_THREADS)
+#elif defined(__NetBSD__) && defined(HPX_HAVE_PTHREAD_SETNAME_NP)
```
hkaiser, line 102: The code here is duplicated 3 times. Can we unify that? I also think that we need to check for `HPX_HAVE_PTHREAD_SETNAME_NP` only. There is no need to test for the platform anymore. As long as this function exists we're fine and we don't care about the platform.

libs/core/thread_support/include/hpx/thread_support/set_thread_name.hpp
```diff
@@ -7,20 +7,9 @@
 #pragma once
 
 #include <hpx/config.hpp>
-
-#if (defined(WIN32) || defined(_WIN32) || defined(__WIN32__)) &&               \
-    !defined(HPX_MINGW)
+#include <hpx/type_support/unused.hpp>
```
hkaiser, line 10: This #include is not needed anymore

CMakeLists.txt
```diff
@@ -1625,6 +1625,8 @@ hpx_check_for_mm_prefetch(DEFINITIONS HPX_HAVE_MM_PREFETCH)
 
 hpx_check_for_stable_inplace_merge(DEFINITIONS HPX_HAVE_STABLE_INPLACE_MERGE)
 
+hpx_check_for_pthread_setname_np(DEFINITIONS HPX_HAVE_PTHREAD_SETNAME_NP)
+
```
hkaiser, line 1629: Could you please move this up or down? Currently it's placed in between the feature test for `HPX_HAVE_STABLE_INPLACE_MERGE` and it's use below.

hkaiser:
Please also resolve the errors reported by inspect (https://output.circle-artifacts.com/output/job/4f945ff2-3848-4b11-8aa4-b486e811b0d3/artifacts/0/hpx_inspect_report.html)

cmake/HPX_AddConfigTest.cmake
```diff
@@ -506,6 +506,15 @@ function(hpx_check_for_cxx20_perfect_pack_capture)
   )
 endfunction()
 
+# ##############################################################################
+function(hpx_check_for_pthread_setname_np)
+  add_hpx_config_test(
+    HPX_HAVE_PTHREAD_SETNAME_NP
```
hkaiser, line 512: This should be:
```suggestion
    HPX_WITH_PTHREAD_SETNAME_NP
```

libs/core/thread_support/src/set_thread_name.cpp
```diff
@@ -83,16 +81,33 @@ namespace hpx::util {
...
-        // also set it the 'new' way in case the application is not running in
-        // the debugger at this time
-        SetThreadDescription(
-            GetCurrentThread(), detail::to_wide_string(thread_name).c_str());
+#include <pthread.h>
+namespace hpx::util {
+
+    void set_thread_name(char const* thread_name) noexcept
+    {
+        pthread_set_name_np(pthread_self(), thread_name);
```
hkaiser, line 97: The feature test looks for `pthread_setname_np`, not `pthread_set_name_np`

Johan511: We seem to have missed this fact, but freeBSD and openBSD use ` pthread_set_name_np` while the remaining use pthread_setname_np. I will make changes accordingly to the feature test and `set_thread_name.cpp`

Johan511: Tests are a bit misleading as we don't seem to be testing for BSD.

hkaiser: Testing for `pthread_set_name_np` should be separate and independent. Let's finalize the current use case (testing for `pthread_setname_np`) first.

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6279, opened 2023-06-15
Re-expose error names

hkaiser:
In one of the recent HPX versions, the previously publicly available string representations of the HPX error code were removed from the public API. This PR introduces an API that allows to query the error names.


hkaiser:
bors merge

PR #6284, opened 2023-06-19
Add CUDA/HIP MultiGPU Event Polling

G-071:
While the [cuda,hip]_executors within HPX support MultiGPU scenarios themselves (by using a device index and ```cudaSetDevice```), the event polling actually does not, forcing users to use the less performant callback version instead! To fix this, we need to make the CUDA/HIP event pool aware on which device the events it reuses are created, and add an index to always push/pop events on the correct device. 

This PR implements this, by adding multiple event stacks (one per device) to the event pool singleton. Furthermore, the PR adds an additional device_id parameter to the appropriate get_future calls that use events. To keep things compatible with code that used previous HPX versions, 0 is used as a default where this parameter was added. For each device, there's always 128 events added in the beginning by default - in case one wants to avoid this overhead when only a single GPU is used, simply set the environment variable CUDA_VISIBLE_DEVICES=0.


libs/core/async_cuda/include/hpx/async_cuda/cuda_event.hpp
```diff
@@ -29,56 +31,77 @@ namespace hpx { namespace cuda { namespace experimental {
...
+                {
+                    ok = free_lists_[device].pop(event);
+                    if (ok)
+                        check_cuda_error(cudaEventDestroy(event));
+                }
             }
         }
 
-        inline bool pop(cudaEvent_t& event)
+        inline bool pop(cudaEvent_t& event, int device)
```
hkaiser, line 70:
```suggestion
        inline bool pop(cudaEvent_t& event, int device = 1)
```
?

I'm not sure about the default value, though - does the CUDA device indexing start with `0` or `1`?

libs/core/async_cuda/include/hpx/async_cuda/cuda_event.hpp
```diff
@@ -29,56 +31,77 @@ namespace hpx { namespace cuda { namespace experimental {
...
+            while (!free_lists_[device].pop(event))
             {
-                add_event_to_pool();
+                add_event_to_pool(device);
             }
             return true;
         }
 
-        inline bool push(cudaEvent_t event)
+        inline bool push(cudaEvent_t event, int device)
```
hkaiser, line 82:
```suggestion
        inline bool push(cudaEvent_t event, int device = 1)
```
?

libs/core/async_cuda/include/hpx/async_cuda/cuda_event.hpp
```diff
@@ -29,56 +31,77 @@ namespace hpx { namespace cuda { namespace experimental {
...
                 cudaEventCreateWithFlags(&event, cudaEventDisableTiming));
-            free_list_.push(event);
+            free_lists_[device].push(event);
         }
+        int max_number_devices_;
 
-        // pool is dynamically sized and can grow if needed
-        hpx::lockfree::stack<cudaEvent_t> free_list_;
+        // One pool per GPU - each pool is dynamically sized and can grow if needed
+        std::deque<hpx::lockfree::stack<cudaEvent_t>> free_lists_;
```
hkaiser, line 105: What's the rationale of using `std::deque`, wouldn't a `std::vector` suffice?

libs/core/async_cuda/include/hpx/async_cuda/cuda_event.hpp
```diff
@@ -29,56 +31,77 @@ namespace hpx { namespace cuda { namespace experimental {
...
         // create a bunch of events on initialization
         cuda_event_pool()
-          : free_list_(initial_events_in_pool)
+          : max_number_devices_(0)
         {
-            for (int i = 0; i < initial_events_in_pool; ++i)
+            check_cuda_error(cudaGetDeviceCount(&max_number_devices_));
+            HPX_ASSERT_MSG(max_number_devices_ > 0,
+                "CUDA polling enabled and called, yet no CUDA device found!");
+            for (int device = 0; device < max_number_devices_; device++)
```
hkaiser, line 39: I's add a `free_lists_.reserve(max_number_devices_-1);` here

libs/core/async_cuda/include/hpx/async_cuda/detail/cuda_event_callback.hpp
```diff
@@ -26,7 +26,7 @@ namespace hpx { namespace cuda { namespace experimental { namespace detail {
         hpx::move_only_function<void(cudaError_t)>;
 
     HPX_CORE_EXPORT void add_event_callback(
-        event_callback_function_type&& f, cudaStream_t stream);
+        event_callback_function_type&& f, cudaStream_t stream, int device);
```
hkaiser, line 29: Let's keep this API function backwards compatible as well:
```suggestion
        event_callback_function_type&& f, cudaStream_t stream, int device = 0);
```

libs/core/async_cuda/include/hpx/async_cuda/cuda_event.hpp
G-071: With std::vector the deleted constructor is an issue:
```
/usr/lib/gcc/x86_64-redhat-linux/8/../../../../include/c++/8/bits/stl_construct.h:75:38: error: call to deleted constructor of 'hpx::lockfree::stack<CUevent_st *, std::allocator<CUevent_st *>, 0, false>'

...

src/hpx/libs/core/concurrency/include/hpx/concurrency/stack.hpp:118:9: note: 'stack' has been explicitly marked deleted here
        stack(stack&&) = delete;
        ^
```

deque was just the first that sprang to mind that worked with stack (even though it seems like a bit of overkill)


Overall, I am open to suggestions for the data-structure. I'd actually like to have something with a fixed length (not known at compile time) as we will never add elements outside of the pool constructor. It also needs to work with the stack class (can't use the default/move/copy constructor).

libs/core/async_cuda/include/hpx/async_cuda/cuda_event.hpp
```diff
@@ -86,7 +68,33 @@ namespace hpx { namespace cuda { namespace experimental {
             return free_lists_[device].push(event);
         }
 
+        // delete copy / move constructors
+        cuda_event_pool(cuda_event_pool&&) = delete;
+        cuda_event_pool& operator=(cuda_event_pool&&) = delete;
+        cuda_event_pool(const cuda_event_pool&) = delete;
```
hkaiser, line 74: just a nitpick - we're using 'east const' in all the code base. Would you mind adapting this?

hkaiser:
@G-071 should we add a test that actually excersises this functionality, perhaps run it on the DGX partition only (e.g. enable it with a special CMake variable)?

libs/core/async_cuda/include/hpx/async_cuda/cuda_event.hpp
G-071: Ah, I forgot about that! Done!

libs/core/async_cuda/include/hpx/async_cuda/cuda_event.hpp
```diff
@@ -19,66 +21,95 @@ namespace hpx { namespace cuda { namespace experimental {
...
     struct cuda_event_pool
     {
-        static constexpr int initial_events_in_pool = 128;
+        static constexpr std::size_t initial_events_in_pool = 128;
 
         static cuda_event_pool& get_event_pool()
         {
             static cuda_event_pool event_pool_;
             return event_pool_;
         }
```
hkaiser, line 30: May I suggest that you move the (`HPX_CORE_EXPORT`ed) implementation of this function into a source file? The rationale is that otherwise on some platforms (Mac, Windows) each executable module (shared library, executable) that will contain code calling this function will have its own instance of the `event_pool_` variable.

G-071:
I have added a few changes/additions/fixes:
- The definition of  the method ```get_event_pool``` is now in cuda_event.cpp. @hkaiser : Is this what you had in mind?
- I have added a test (cuda_multi_device_polling.cpp) which tests executors on different devices. This tests queries the number of devices, creates an executor per device and runs a dummy kernel on each executor. If the multi device polling does not work, the futures for those dummy kernel calls will never become ready and the test will time out. As it adapts to the number of devices present, it will test with all 8 GPUs on the DGX node. The test also verifies that ```detail::get_future_with_event``` uses the currently active device (previousely set in the user code via cudaSetDevice) by default if no other device is given.
- The thread that initializes the event pool now restores its original device ID afterwards. Without that, the thread used a different device ID that the other ones afterward (if multiple devices are present), causing problems in user code that only used 1 GPU. This is only done for the original initialization of the event pools for all devices. Subsequent events that are added on-demand for a device pool use the device ID that is passed by the user (either via ```detail::get_future_with_event``` directly or via the cuda executor). Meaning, if we get a non-zero device ID, the user code handles the setting of the devices already and we will simply use the passed device ID accordingly (saving us a bit of overhead by avoiding to call cudaGetDevice and cudaSetDevice).
- Format fixes.

Overall, this seems to work fine on the NVIDIA nodes I have tested in on so far (using Octo-Tiger).

hkaiser:
@G-071 could you please fix the inspect errors reported? After that I would be able to go ahead and merge the PR.

G-071:
@hkaiser Done! I just forgot about the noascii check (which does not like my name due to the ß and thus complains about me being in the author list of a file) and was missing a few headers.

hkaiser:
bors merge

PR #6286, opened 2023-06-21
Simd helpers

Johan511:
simd helpers based on intel implementation


Johan511:
Peformance graph for `unseq_first_n` against sequential implementation
graph1

Johan511:
Peformance graph for `unseq2_first_n` against sequential implementation

graph2

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
```diff
@@ -0,0 +1,153 @@
...
+        for (; __i < endIndex; ++__i)
+        {
+            if (f(first, __i))
+            {
+                break;
+            }
+        }
+        // clang-format on
+
+        return first + __i;
```
hkaiser, line 43: This code was taken from libstdc++, was it? Please be aware that this library is GPL licensed and we're not able to add it to our code base.

Johan511: ok, will ideate on a different implementation.

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
```diff
@@ -0,0 +1,161 @@
...
+#include <type_traits>
+#include <utility>
+
+namespace hpx::parallel::util {
+    /*
+        Compiler and Hardware should also support vector operations for IterDiff,
+        else we see slower performance when compared to sequential version
+    */
+    template <class Iter, class IterDiff, class F>
+    Iter unseq_first_n(const Iter first, const IterDiff n, F&& f_) noexcept
```
hkaiser, line 29: We use 'east const' in our codebase, please change this accordingly

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
```diff
@@ -0,0 +1,161 @@
...
+    {
+        /*
+            OMP loops can not have ++Iter, only integral types are allowed
+            Hence perform arthemetic on Iterators
+            which is O(1) only in case of random access iterators
+        */
+        static_assert(hpx::traits::is_random_access_iterator_v<Iter>,
+            "algorithm is efficient only in case of Random Access Iterator");
+#if HPX_EARLYEXIT_PRESENT
```
hkaiser, line 38: Is this a build-system constant? Where would it be defined?

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
```diff
@@ -0,0 +1,161 @@
...
+        alignas(HPX_LANE_SIZE) int32_t simdLane[numBlocks] = {0};
+        while (i <= n - numBlocks)
+        {
+            int32_t foundFlag = 0;
+
+            // clang-format off
+            HPX_PRAGMA_VECTOR_UNALIGNED HPX_VECTOR_REDUCTION(| : foundFlag)
+            for (IterDiff j = i; j < i + numBlocks; ++j)
+            {
+                const int32_t t = HPX_INVOKE(f,first, j);
```
hkaiser, line 66: Do you need `HPX_INVOKE` here? Wouldn't a simple `f(first, j)` be sufficient?

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
```diff
@@ -0,0 +1,161 @@
...
+            }
+        }
+        // clang-format on
+
+        return first + i;
+#else
+        auto f = [](auto it, int ind) { return HPX_INVOKE(f_, *(ind + it)) };
+        // int32_t has best support for vectorization from compilers and hardware
+        IterDiff i = 0;
+        static constexpr int32_t numBlocks = HPX_LANE_SIZE / sizeof(int32_t);
```
hkaiser, line 56: Please make variable names conform to our naming conventions, e.g. `num_blocks`. Also, we always prefix the fixed size integral types with `std::`, e.g. `std::int32_t`.

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
```diff
@@ -0,0 +1,161 @@
...
+        IterDiff i = 0;
+
+        // clang-format off
+        HPX_PRAGMA_VECTOR_UNALIGNED HPX_PRAGMA_SIMD_EARLYEXIT
+        for (; i < n; ++i)
+            if (HPX_INVOKE(f, *(first1 + i), *(first2 + i)))
+                break;
+        // clang-format on
+
+        return std::make_pair(first1 + i, first2 + i);
```
hkaiser, line 114:
```suggestion
        HPX_PRAGMA_VECTOR_UNALIGNED HPX_PRAGMA_SIMD_EARLYEXIT
        for (IterDiff i = 0; i < n; ++i)
        {
            if (HPX_INVOKE(f, *first1, *first2))
                break;
            ++first1; ++first2;
        }
        // clang-format on
        return std::make_pair(first, first2);
```
Or would this break vectorization?

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
Johan511: If you are referring to `HPX_EARLYEXIT_PRESENT`, then its defined in [auto_vectorization.hpp](https://github.com/STEllAR-GROUP/hpx/pull/6286/files#diff-f9fca42172c7e4f2bed2865a1937603f7f3084c0dcff9af28e550dac67dfac2dR30).

It is a check for `#pragma omp simd early_exit` which is supported only by intel compilers newer than version 1800

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
Johan511: The block of code is only called when the program is compiled with an intel compiler of version >= 1800.
I haven't been able to test out the code on an intel compiler but given how OMP works, it'd very likely throw a compile time error.

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
```diff
@@ -0,0 +1,165 @@
...
+        alignas(HPX_LANE_SIZE) std::int32_t simd_lane[num_blocks] = {0};
+        while (i <= n - num_blocks)
+        {
+            std::int32_t found_flag = 0;
+
+            // clang-format off
+            HPX_PRAGMA_VECTOR_UNALIGNED HPX_VECTOR_REDUCTION(| : found_flag)
+            for (IterDiff j = i; j < i + num_blocks; ++j)
+            {
+                std::int32_t const t = f(first, j);
```
hkaiser, line 68: Can we remove the lambda altogether?
```suggestion
                std::int32_t const t = f_(*(first + j));
```

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
```diff
@@ -0,0 +1,165 @@
...
+        */
+        static_assert(hpx::traits::is_random_access_iterator_v<Iter>,
+            "algorithm is efficient only in case of Random Access Iterator");
+#if HPX_EARLYEXIT_PRESENT
+        IterDiff i = 0;
+        // clang-format off
+        HPX_PRAGMA_VECTOR_UNALIGNED HPX_PRAGMA_SIMD_EARLYEXIT
+        for (; i < n; ++i)
+        {
+            if (f(first, i))
```
hkaiser, line 44: `f` seems to be undefined here.

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+        return first + i;
+#else
+        auto f = [f_ = HPX_FORWARD(F, f_)](
+                     auto it, int ind) { return f_(*(ind + it)); };
+        // std::int32_t has best support for vectorization from compilers and hardware
+        IterDiff i = 0;
+        static constexpr std::int32_t num_blocks =
+            HPX_LANE_SIZE / sizeof(std::int32_t);
+        alignas(HPX_LANE_SIZE) std::int32_t simd_lane[num_blocks] = {0};
```
hkaiser, line 59: in our reduce implementation we use `std::uint8_t`. Does this make a difference? (see: https://github.com/STEllAR-GROUP/hpx/blob/4aa2e7451f48e119d5143e3dd070415bf7159f82/libs/core/algorithms/include/hpx/parallel/unseq/reduce_helpers.hpp#L217

libs/core/config/include/hpx/config/auto_vectorization.hpp
```diff
@@ -27,16 +27,31 @@
 #define HPX_PRAGMA(x) _Pragma(#x)
 #endif
 
+#define HPX_EARLYEXIT_PRESENT (__INTEL_COMPILER >= 1800)
+#define HPX_MONOTONIC_PRESENT (__INTEL_COMPILER >= 1800)
```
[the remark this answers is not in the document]
Johan511: HPX_EARLYEXIT_PRESENT is used in this PR
HPX_MONOTONIC_PRESENT might be used in a different simd-helper

I will remove the latter from this PR

libs/core/config/include/hpx/config/auto_vectorization.hpp
```diff
@@ -27,16 +27,31 @@
...
+
+#if (__INTEL_COMPILER >= 1600)
+#define HPX_PRAGMA_VECTOR_UNALIGNED HPX_PRAGMA(vector unaligned)
+#else
+#define HPX_PRAGMA_VECTOR_UNALIGNED
+#endif
+
+#if HPX_EARLYEXIT_PRESENT
+#define HPX_PRAGMA_SIMD_EARLYEXIT HPX_PRAGMA(omp simd early_exit)
```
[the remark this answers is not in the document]
Johan511: I believe intel compilers have their trademark openMP implementation, which is the only one to support HPX_EARLYEXIT_PRESENT. The requirement for this is an intel compiler.

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
Johan511: Requested change has been made

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
```diff
@@ -0,0 +1,165 @@
...
+#include <memory>
+#include <type_traits>
+#include <utility>
+
+namespace hpx::parallel::util {
+    /*
+        Compiler and Hardware should also support vector operations for IterDiff,
+        else we see slower performance when compared to sequential version
+    */
+    template <class Iter, class IterDiff, class F>
```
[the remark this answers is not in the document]
Johan511: Requested change has been made

libs/core/config/include/hpx/config/auto_vectorization.hpp
```diff
@@ -27,16 +27,31 @@
...
 #endif
 
+#define HPX_EARLYEXIT_PRESENT (__INTEL_COMPILER >= 1800)
+#define HPX_MONOTONIC_PRESENT (__INTEL_COMPILER >= 1800)
+
+#if (__INTEL_COMPILER >= 1600)
+#define HPX_PRAGMA_VECTOR_UNALIGNED HPX_PRAGMA(vector unaligned)
+#else
+#define HPX_PRAGMA_VECTOR_UNALIGNED
+#endif
```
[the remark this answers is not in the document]
Johan511: If you are suggesting move to this [section](https://github.com/STEllAR-GROUP/hpx/blob/0af1fc6aedebabd46b1be308009d7623a00e7fa9/libs/core/config/include/hpx/config/auto_vectorization.hpp#L46), I am not sure if its right. I haven't found vector unaligned pragmas being supported by _OPENMP >= 201307 or (defined(__clang__) && HPX_CLANG_VERSION >= 30700)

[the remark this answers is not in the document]
hkaiser: I meant to move it to the `#if defined(HPX_INTEL_VERSION)` with and additional version control, if needed (and to all the other sections as empty macros)

libs/core/config/include/hpx/config/auto_vectorization.hpp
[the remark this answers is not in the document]
hkaiser: My rationale for suggesting this is that we have 4 or 5 compiler specific sections further down in this file. However this code adds something that is also compiler specific in a different way. I'm pleading for some consistency, that's all.

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
Johan511: untill gcc 9, vectorization is enabled only in case of std::int32_t not for int8_t
in case of clang it is the opposite, vectorization is enabled for only int8_t until clang 14

Testing performance on gcc 11 (vectorizes in both cases) has shown int32_t to be around 40% faster (90 mus vs 13mus)

libs/core/config/include/hpx/config/auto_vectorization.hpp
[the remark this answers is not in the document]
Johan511: I have moved them into the if section. Please confirm if its fine now.

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
hkaiser: > untill gcc 9, vectorization is enabled only in case of std::int32_t not for int8_t in case of clang it is the opposite, vectorization is enabled for only int8_t until clang 14 Godbolt : https://godbolt.org/z/rronrG154
> 
> Testing performance on gcc 11 (vectorizes in both cases) has shown int32_t to be around 40% faster (90 mus vs 13mus)

Nice insight! Does that mean that we should take this into account here and for the code for reduce as well?

Johan511: Yes, but let's test out the performance before making any changes to reduce.

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
```diff
@@ -0,0 +1,163 @@
...
+                }
+                return std::make_pair(
+                    first1 + outer_loop_ind + i2, first2 + outer_loop_ind + i2);
+            }
+            outer_loop_ind += num_blocks;
+        }
+
+        //Keep remainder scalar
+        for (; outer_loop_ind != n; ++outer_loop_ind)
+            if ((f, *(first1 + outer_loop_ind), *(first2 + outer_loop_ind)))
```
hkaiser, line 156: Does this realy do what you want? Shouldn't this be:
```suggestion
            if (f(*(first1 + outer_loop_ind), *(first2 + outer_loop_ind)))
```

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
```diff
@@ -0,0 +1,163 @@
...
+            }
+            i += num_blocks;
+        }
+
+        //Keep remainder scalar
+        while (i != n)
+        {
+            if (f(*(first + i)))
+            {
+                return first + i;
```
hkaiser, line 92: Above, you have used a different idiom for exactly the same code:
```suggestion
                break;
```

hkaiser (commented):
All of this looks good to me now. Thanks! The only thing missing would be some tests that verify the introduced functionality.

Johan511:
Functionality tests have been added.

hkaiser:
@Johan511 could you please resolve the conflicts? That would allow me to review the changes and go ahead with merging this, if appropriate.

Johan511:
Conflicts seem to be resolved, can you please check @hkaiser ?

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6294, opened 2023-06-30
Add internal reference counting to semaphores

Pansysk75:
Use intrusive_ptr to implement internal reference counting for semaphores.

### DIFF
--- a/libs/core/synchronization/include/hpx/synchronization/counting_semaphore.hpp
+++ b/libs/core/synchronization/include/hpx/synchronization/counting_semaphore.hpp
@@ -10,6 +10,7 @@
 #pragma once
 
 #include <hpx/config.hpp>
+#include <hpx/memory/serialization/intrusive_ptr.hpp>
 #include <hpx/synchronization/detail/counting_semaphore.hpp>
 #include <hpx/synchronization/spinlock.hpp>
 #include <hpx/timing/steady_clock.hpp>
@@ -350,6 +351,8 @@ namespace hpx {
 
         protected:
             using mutex_type = Mutex;
+            using data_type =
+                lcos::local::detail::counting_semaphore_data<mutex_type>;
 
         public:
             static constexpr std::ptrdiff_t(max)() noexcept
@@ -358,7 +361,7 @@ namespace hpx {
             }
 
             explicit counting_semaphore(std::ptrdiff_t value) noexcept
-              : sem_(value)
+              : data_(new data_type(value), false)
             {
             }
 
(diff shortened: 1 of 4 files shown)


libs/core/synchronization/include/hpx/synchronization/counting_semaphore.hpp
```diff
@@ -342,14 +342,10 @@ namespace hpx {
             typename Mutex = hpx::spinlock>
         class counting_semaphore
         {
-        public:
-            counting_semaphore(counting_semaphore const&) = delete;
-            counting_semaphore& operator=(counting_semaphore const&) = delete;
-            counting_semaphore(counting_semaphore&&) = delete;
-            counting_semaphore& operator=(counting_semaphore&&) = delete;
```
hkaiser, line 349: Semaphores should be non-movable (see: https://en.cppreference.com/w/cpp/thread/counting_semaphore).

Pansysk75: Aha, i think I misunderstood something then.

Initially, I thought that the goal was to ref-count copies of the object (similarly to what we do for id_type), and pass around copies of the semaphore object, which would now serve more like an interface to the internal data.
But that confuses copying with referencing, and doesn't conform to the C++ standard either (since we would have to enable copy operations for semaphores). 
I don't really like it.

I now see that, on the barrier example that you had provided, we use the ref-counting on the internal state only for mitigating the issue of the object being destroyed while waiting on a lock. We don't use it for ref-counting copies of the barrier (since the copy/move operations are deleted).
Doing this on the semaphores seems much more reasonable than what I had in mind previously.

Let me know if my new understanding is closer to what you had in mind

libs/core/synchronization/include/hpx/synchronization/detail/counting_semaphore.hpp
```diff
@@ -55,6 +55,37 @@ namespace hpx::lcos::local::detail {
         std::ptrdiff_t value_;
         local::detail::condition_variable cond_;
     };
+
+    template <typename Mutex>
+    struct counting_semaphore_data
+    {
+        counting_semaphore_data(std::ptrdiff_t value) noexcept
```
hkaiser, line 62:
```suggestion
        explicit counting_semaphore_data(std::ptrdiff_t value) noexcept
```

hkaiser (approved):
Otherwise, LGTM, thanks!

hkaiser:
bors merge

PR #6296, opened 2023-07-03
Fixing error message about use_guard_pages

### LINKED ISSUE #6288: Problems using the hpx.stacks:use_guard_pages parameter
To run larger scenarios with Octo-Tiger on a single node, we usually have to add the parameter ```--hpx.stacks:use_guard_pages=0```. Recently (not quite sure when this started to happen), passing this parameter caused an issue within Octo-Tiger though:

```
terminate called after throwing an instance of 'boost::wrapexcept<boost::program_options::unknown_option>'
  what():  unrecognised option '--hpx.stacks:use_guard_pages=0'
```
Note, that this error is thrown when Octo-Tiger itself is parsing its CLI parameters (well after hpx::init). 

Simply adding the parameter to the Octo-Tiger command used to work without any issue! Has something recently changed regarding the HPX CLI parameters, which might cause this one to be forwarded to Octo-Tiger somehow?

hkaiser:
Fixes #6288


hkaiser:
bors merge

PR #6302, opened 2023-07-16
Simd first

Johan511:
using simd-helpers (from #6286 ) to add vectorization to stl algorithms

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
@@ -20,6 +20,7 @@
 #include <type_traits>
 #include <utility>
 
+// Please use static assert and enforce Iter to be Random Access Iterator
 namespace hpx::parallel::util {
     /*
         Compiler and Hardware should also support vector operations for IterDiff,
@@ -120,7 +121,7 @@ namespace hpx::parallel::util {
             HPX_LANE_SIZE / sizeof(std::int32_t);
         alignas(HPX_LANE_SIZE) std::int32_t simd_lane[num_blocks] = {0};
 
-        std::size_t outer_loop_ind = 0;
+        IterDiff outer_loop_ind = 0;
         while (outer_loop_ind <= n - num_blocks)
         {
             std::int32_t found_flag = 0;
--- a/libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
@@ -114,8 +114,6 @@ namespace hpx::parallel::util {
 
         return std::make_pair(first1 + i, first2 + i);
 #else
-        Iter1 const last1 = first1 + n;
-        Iter2 const last2 = first2 + n;
 
         static constexpr std::int32_t num_blocks =
             HPX_LANE_SIZE / sizeof(std::int32_t);
--- a/libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
@@ -26,7 +26,7 @@ namespace hpx::parallel::util {
         else we see slower performance when compared to sequential version
     */
     template <class Iter, class IterDiff, class F>
-    Iter unseq_first_n(const Iter first, const IterDiff n, F&& f_) noexcept
+    Iter unseq_first_n(Iter const first, IterDiff const n, F&& f_) noexcept
     {
         /*
             OMP loops can not have ++Iter, only integral types are allowed
@@ -41,7 +41,7 @@ namespace hpx::parallel::util {
         HPX_PRAGMA_VECTOR_UNALIGNED HPX_PRAGMA_SIMD_EARLYEXIT
         for (; i < n; ++i)
         {
-            if (HPX_INVOKE(f, first, i))
+            if (f(first, i))
             {
                 break;
             }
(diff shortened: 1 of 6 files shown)


libs/core/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
```diff
@@ -34,8 +34,7 @@ namespace hpx::parallel::detail {
             sequential_find_t<ExPolicy>, Iterator first, Sentinel last,
             T const& value, Proj proj = Proj())
         {
-            return util::loop_pred<
-                std::decay_t<hpx::execution::sequenced_policy>>(
+            return util::loop_pred<ExPolicy>(
```
hkaiser, line 37: Could you add `static_assert` verifying that `ExPolicy` is actually a sequential policy?

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
```diff
@@ -0,0 +1,161 @@
...
+#include <hpx/type_support/construct_at.hpp>
+
+#include <algorithm>
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <type_traits>
+#include <utility>
+
+// Please use static assert and enforce Iter to be Random Access Iterator
```
hkaiser, line 23: Can this comment be removed?

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
```diff
@@ -0,0 +1,161 @@
...
+namespace hpx::parallel::util {
+    /*
+        Compiler and Hardware should also support vector operations for IterDiff,
+        else we see slower performance when compared to sequential version
+    */
```
hkaiser, line 28: What does this comment mean? Should that requirement be somehow enforced?

libs/core/algorithms/include/hpx/parallel/util/loop.hpp
```diff
@@ -104,11 +105,12 @@ namespace hpx::parallel::util {
...
     ///////////////////////////////////////////////////////////////////////////
     namespace detail {
-
+        template <typename Iterator, typename ExPolicy>
+        struct loop_pred;
         // Helper class to repeatedly call a function starting from a given
         // iterator position till the predicate returns true.
         template <typename Iterator>
-        struct loop_pred
+        struct loop_pred<Iterator, void>
```
hkaiser, line 113: Where is this used? Is `std::is_same_v<Expolicy, void>` true at any point?

libs/core/algorithms/include/hpx/parallel/util/loop.hpp
```diff
@@ -125,6 +127,39 @@ namespace hpx::parallel::util {
...
+        // iterator position till the predicate returns true.
+        template <typename Iterator, typename ExPolicy>
+        struct loop_pred
+        {
+            ///////////////////////////////////////////////////////////////////
+            template <typename Begin, typename End, typename Pred>
+            HPX_HOST_DEVICE HPX_FORCEINLINE static constexpr Begin call(
+                Begin it, End end, Pred&& pred)
+            {
+                if constexpr (hpx::is_unsequenced_execution_policy_v<
```
hkaiser, line 143: Could we move this dispatch point to the CPO below, please?

libs/core/algorithms/include/hpx/parallel/util/loop.hpp
```diff
@@ -786,7 +821,7 @@ namespace hpx::parallel::util {
 
         // Helper class to repeatedly call a function a given number of times
         // starting from a given iterator position.
-        template <typename IterCat>
+        template <typename IterCat, typename Dummy = void>
```
hkaiser, line 824: Same question as above: where is `loop_idx_n<IterCat, void>` used?

libs/core/algorithms/tests/unit/algorithms/util/test_simd_helpers.cpp
```diff
@@ -0,0 +1,105 @@
...
+std::mt19937 gen(seed);
+
+using hpx::parallel::util::unseq_first_n, hpx::parallel::util::unseq2_first_n;
+
+template <typename T>
+void test_unseq_first_n1_dispatch2(std::size_t length, std::size_t first_index)
+{
+    first_index = first_index % length;
+
+    std::vector<T> v(length, static_cast<T>(false));
```
hkaiser, line 26: Why casting a boolean to `T`? Can you explain, please?

Johan511:
first-simd
Speedups observed

libs/core/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
Johan511: Change had been made so sequential_find_t can accept sequential and unsequential execution policies. Do you want me to rather use `static_assert(is_seq || is_unseq)` ?

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
Johan511: As discussed [previously](https://github.com/STEllAR-GROUP/hpx/pull/6286#discussion_r1239159849), changing data types can lead to vectorization being broken. I wanted to leave a note for the future regarding it. Use perf regression tests might not be the best way to do this as they often seem to have lot of variance in them.

libs/core/algorithms/include/hpx/parallel/util/loop.hpp
Johan511: moved to tag_invoke

libs/core/algorithms/tests/unit/algorithms/util/test_simd_helpers.cpp
Johan511: T is a supposed to be a datatype which supports vectorization, something like int. Is there anything wrong with explicitly typecasting false into T?

libs/core/algorithms/include/hpx/parallel/util/loop.hpp
Johan511: I believe this was done so loop_pred can be templated as `loop<Iter>` and `loop<Iter, ExPolicy>`
Now that we have moved the dispatch to tag_invoke, I don't think this is required, I will look into replacing it with the older version.

libs/core/algorithms/include/hpx/parallel/util/loop.hpp
Johan511: I believe this was done so loop_pred can be templated as `loop_idx_n<Iter>` and `loop_idx_n<Iter, ExPolicy>`
Now that we have moved the dispatch to tag_invoke, I don't think this is required, I will look into replacing it with the older version.

libs/core/algorithms/tests/unit/algorithms/util/test_simd_helpers.cpp
hkaiser: Why not use `0` (zero) instead? or even just `T()`?

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
hkaiser: Well, a comment doesn't help with this, does it? I think setting up performance tests would be the better option. @Pansysk75 can help with that.

hkaiser:
> Speedups observed for seq vs unseq

Nice!

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
Pansysk75: @hkaiser Do you mean automated ones that report on GH? He asked me about this but I discouraged him from doing that yet. It can certainly be done though, if that's what you have in mind

hkaiser: Well, I thought we could add it to our performance tests.

libs/core/algorithms/tests/unit/algorithms/util/test_simd_helpers.cpp
Johan511: While writing the test, I was thinking about finding the first true value in the vector. I don't see any difference in behaviour in having false vs 0. [Will change it to 0](https://github.com/STEllAR-GROUP/hpx/pull/6302/commits/adfd8b6eda0d9c4ed5c36ead9f04a44932d22947).

hkaiser: FWIW, writing `std::vector<T> v(length);` would do exactly what you want. It would call the default constructor for all new elements, i.e. initializing them to zero as all default constructed builtin types are zero initialized.

Johan511:
> @Johan511 could you please fix the reported clang-format issues as well?

There is one loop yet to be vectorized, once that's done vectorization should be enabled for parallel case too. We can merge after that.

Johan511:
@hkaiser the unseq_first_n function has been changed a bit, the the function now takes iterator as input instead of value, this does not break vectorization.

libs/core/algorithms/include/hpx/parallel/unseq/simd_helpers.hpp
Johan511: added it a separate [PR](https://github.com/STEllAR-GROUP/hpx/pull/6325)

PR #6303, opened 2023-07-18
omp for default vectorization

Johan511:
hpx uses compiler assisted vectorization (enabled by the -O3 flag), changing it to OMP vectorization

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -465,6 +465,22 @@ elseif(HPX_WITH_PARALLEL_LINK_JOBS)
   hpx_warn("Job pooling is only available with Ninja generators.")
 endif()
 
+hpx_option(
+  HPX_HAVE_SIMD_BACKEND
+  STRING
+  "Simd Backend to use for vectorization. Defaults to OMP, other possible options are NONE (compiler assisted vectorization)"
+  "OMP"
+  CATEGORY "Build Targets"
+)
+
+if(NOT HPX_HAVE_SIMD_BACKEND)
+  set(HPX_HAVE_SIMD_BACKEND "OMP")
+endif()
+
+if(HPX_HAVE_SIMD_BACKEND STREQUAL "OMP")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp-simd")
+endif()
+
 # disable all tests if HPX_WITH_TESTS=OFF
 if(NOT HPX_WITH_TESTS)
   hpx_set_option(
--- a/libs/core/algorithms/include/hpx/parallel/unseq/loop.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/unseq/loop.hpp
@@ -30,7 +30,7 @@ namespace hpx::parallel::util {
                 InIter HPX_RESTRICT it, std::size_t num, F&& f)
             {
                 // clang-format off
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -465,6 +465,22 @@ elseif(HPX_WITH_PARALLEL_LINK_JOBS)
   hpx_warn("Job pooling is only available with Ninja generators.")
 endif()
 
+hpx_option(
+  HPX_HAVE_SIMD_BACKEND
```
hkaiser, line 469: The naming convention for CMake options is `HPX_WITH_...`

CMakeLists.txt
```diff
@@ -465,6 +465,22 @@ elseif(HPX_WITH_PARALLEL_LINK_JOBS)
...
+  STRING
+  "Simd Backend to use for vectorization. Defaults to OMP, other possible options are NONE (compiler assisted vectorization)"
+  "OMP"
+  CATEGORY "Build Targets"
+)
+
+if(NOT HPX_HAVE_SIMD_BACKEND)
+  set(HPX_HAVE_SIMD_BACKEND "OMP")
```
hkaiser, line 477: This will force using the `-fopenmp-simd` even on platforms that are not based on gcc or clang. See also https://github.com/STEllAR-GROUP/hpx/actions/runs/5597037732/job/15160508847?pr=6303 for some evidence.

CMakeLists.txt
```diff
@@ -465,6 +465,22 @@ elseif(HPX_WITH_PARALLEL_LINK_JOBS)
...
+endif()
+
+if(HPX_HAVE_SIMD_BACKEND STREQUAL "OMP")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp-simd")
```
hkaiser, line 481: I'd suggest using `hpx_add_compile_flag_if_available(-fopenmp-simd)` instead

CMakeLists.txt
hkaiser: What other options except `"OMP"` would the user have?

libs/core/algorithms/include/hpx/parallel/unseq/loop.hpp
```diff
@@ -26,12 +26,12 @@ namespace hpx::parallel::util {
...
         {
             template <typename InIter, typename F>
-            HPX_HOST_DEVICE HPX_FORCEINLINE static constexpr InIter call(
+            HPX_HOST_DEVICE HPX_FORCEINLINE static InIter call(
                 InIter HPX_RESTRICT it, std::size_t num, F&& f)
             {
                 // clang-format off
-                HPX_IVDEP HPX_UNROLL HPX_VECTORIZE
-                for (std::size_t i = 0; i != num; ++i)
+                HPX_VECTORIZE
```
hkaiser, line 33: Uhh, does this PR completely remove vectorization for non-omp based schemes?

Johan511: I believe HPX_VECTORIZE does vectorize is absence of OMP too, it defaults to compiler supported vectorization

hkaiser: Ok, we need to proof of that statement. Also, if this is the case, can we remove the other macros?

PR #6313, opened 2023-07-27
Trivial relocation of c-v-ref-array types

isidorostsa:
This PR is an addition to the earlier https://github.com/STEllAR-GROUP/hpx/pull/6264

It allows peeling away type qualifiers to make sure that declaring type `T` as trivially relocatable will declare the following types too: 

`const T`, `volatile T`, `const volatile T`, `T&`, `T&&`, `T[]`, `T[n]`

### DIFF
--- a/libs/core/type_support/include/hpx/type_support/is_trivially_relocatable.hpp
+++ b/libs/core/type_support/include/hpx/type_support/is_trivially_relocatable.hpp
@@ -85,12 +85,14 @@ namespace hpx {
     };
 
     template <typename T, int size>
-    struct is_trivially_relocatable<T volatile[size]> : is_trivially_relocatable<T>
+    struct is_trivially_relocatable<T volatile[size]>
+      : is_trivially_relocatable<T>
     {
     };
 
     template <typename T, int size>
-    struct is_trivially_relocatable<T const volatile[size]> : is_trivially_relocatable<T>
+    struct is_trivially_relocatable<T const volatile[size]>
+      : is_trivially_relocatable<T>
     {
     };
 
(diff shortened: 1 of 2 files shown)


hkaiser (dismissed):
LGTM, thanks!

isidorostsa:
> LGTM, thanks!

Thanks for reviewing, I am surprised there is not a simpler way to specialize all versions of a type at once

hkaiser:
> > LGTM, thanks!
> 
> Thanks for reviewing, I am surprised there is not a simpler way to specialize all versions of a type at once

Well, there is, to some extent. You could write:
```
    namespace detail {

        template <typename T>
        struct is_trivially_relocatable<T> : std::is_trivially_copyable<T>
        {
        };
    }

    template <typename T>
    struct is_trivially_relocatable<T> : detail::is_trivially_relocatable<std::decay_t<T>>
    {
    };
```
In this case the user type specializations (the macros) would have to to map onto `hpx::detail::is_trivially_relocatable` instead of `hpx::is_trivially_relocatable`. However I'm not sure if this is a good idea.

isidorostsa:
Oh that looks a lot cleaner, but I agree we could do without two `is_trivially_relocatable` traits :P 

Also just noting that `std::decay_t` would not work for this case because we want to transform `T[n] ->  T` instead of `T[n] -> T*`

hkaiser:
@isidorostsa so let's go ahead with it as is, should we?

isidorostsa:
Let me add some tests first and I think we are good

hkaiser:
@isidorostsa once you fix the clang-format issues, this seems to be good to go.

isidorostsa:
> @isidorostsa once you fix the clang-format issues, this seems to be good to go.

Will do, but let's discuss a bit on Thursday before merging

hkaiser:
@isidorostsa there is still some clang-format problem, could you have a look, please?

isidorostsa:
This will be integrated into #6314

PR #6315, opened 2023-08-03
acknowledge triv-rel and nothrow-rel types

isidorostsa:
Add more traits to be detectable by `pointer_category`. 

This will be used to determine which path (`memmove` or `relocate_at` loop) should be chosen in `transfer.hpp`

### DIFF
--- a/libs/core/algorithms/include/hpx/algorithms/traits/pointer_category.hpp
+++ b/libs/core/algorithms/include/hpx/algorithms/traits/pointer_category.hpp
@@ -160,8 +160,8 @@ namespace hpx::traits {
                 iterators_are_contiguous_v<Source, Dest> && memcpy_legal;
 
             constexpr static bool can_move_construct_nothrow =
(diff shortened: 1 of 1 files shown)


libs/core/algorithms/include/hpx/algorithms/traits/pointer_category.hpp
```diff
@@ -25,7 +26,11 @@ namespace hpx::traits {
     {
     };
 
-    struct relocatable_pointer_tag : general_pointer_tag
+    struct trivially_relocatable_pointer_tag : general_pointer_tag
```
hkaiser, line 29: Where is this used?

isidorostsa: ATM nowhere, but I need to add this in order to modify `transfer.hpp` to recognize trivial relocatability.

isidorostsa:
Will be incorporated in other pr

PR #6321, opened 2023-08-11
Ensure hpx_main is a proper thread_function

Pansysk75:
Ensure that hpx_main is launched as a thread_function, so that exit callbacks are respected, and the following code will actually be used:

https://github.com/STEllAR-GROUP/hpx/blob/e7c31a49dd4e550a61134e498f3ecfb585d201cb/libs/core/threading_base/include/hpx/threading_base/register_thread.hpp#L28-L52

I should add that previously the callbacks were never executed, so using add_thread_exit_callback would cause an assertion failure later on thread destruction (L163), due to this sequence of calls:

https://github.com/STEllAR-GROUP/hpx/blob/e7c31a49dd4e550a61134e498f3ecfb585d201cb/libs/core/threading_base/src/thread_data.cpp#L107-L111

https://github.com/STEllAR-GROUP/hpx/blob/e7c31a49dd4e550a61134e498f3ecfb585d201cb/libs/core/threading_base/src/thread_data.cpp#L157-L166

@hkaiser Let me know if this is a good idea, and whether you'd like me to add a regression test.

### DIFF
--- a/tests/regressions/threads/main_thread_exit_callbacks.cpp
+++ b/tests/regressions/threads/main_thread_exit_callbacks.cpp
@@ -1,0 +1,55 @@
+//  Copyright (c) 2023 Panos Syskakis
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/init.hpp>
+#include <hpx/modules/testing.hpp>
+#include <hpx/runtime.hpp>
+#include <hpx/thread.hpp>
+
+#include <atomic>
+#include <cstddef>
+
+bool callback_called(false);
+
+int hpx_main()
+{
+    hpx::threads::thread_id_type id = hpx::threads::get_self_id();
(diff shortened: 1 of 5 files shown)


hkaiser (dismissed):
LGTM, thanks!

hkaiser:
> Let me know if this is a good idea, and whether you'd like me to add a regression test.

I'd appreciate it if you added a regression test. Thanks!

Pansysk75:
retest lsu

tests/regressions/threads/main_thread_exit_callbacks.cpp
```diff
@@ -0,0 +1,55 @@
...
+        iparams.mode = hpx::runtime_mode::console;
+        callback_called = false;
+        HPX_TEST_EQ_MSG(hpx::init(argc, argv, iparams), 0,
+            "HPX main exited with non-zero status");
+
+        HPX_TEST(callback_called);
+    }
+#endif
+    return hpx::util::report_errors();
+}
```
Pansysk75, line 55: @hkaiser This seems to cause assert failures on CI. Anything obviously wrong, or should I go bug-hunting again?

hkaiser:
bors merge

PR #6322, opened 2023-08-12
Remove simd

Johan511:
added remove unseq by changing find call to unseq

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
@@ -45,7 +45,6 @@ namespace hpx::parallel::detail {
             std::size_t base_idx, FwdIter part_begin, std::size_t part_count,
             Token& tok, T const& val, Proj&& proj)
         {
-            // TODO
             util::loop_idx_n<ExPolicy>(base_idx, part_begin, part_count, tok,
                 [&val, &proj, &tok](auto& v, std::size_t i) -> void {
                     if (HPX_INVOKE(proj, v) == val)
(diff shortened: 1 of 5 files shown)


Johan511:
image
Seq vs unseq

libs/core/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
```diff
@@ -34,8 +34,7 @@ namespace hpx::parallel::detail {
             sequential_find_t<ExPolicy>, Iterator first, Sentinel last,
             T const& value, Proj proj = Proj())
         {
-            return util::loop_pred<
-                std::decay_t<hpx::execution::sequenced_policy>>(
+            return util::loop_pred<ExPolicy>(
```
hkaiser, line 37: Do you need to decay the `ExPolicy` here (and in other places)?

Johan511: Do we really need decay over here? ExPolicy is only being used for deducing if seq or unseq overload should be called. is_unseq<unseq_policy> should be true for both value and reference right?

Also, please note that this is branched from [simd-find](https://github.com/STEllAR-GROUP/hpx/pull/6302).

hkaiser: Well, you tell me. I was just asking the question without closer investigation.

Johan511: To the best of my knowledge, I don't believe we need to decay. I am unsure why it was being decayed previously.

Johan511: @hkaiser I have tried checking behaviour with/without std::decay_t. The correct overloads are called in both cases. I am unsure why std::decay_t is being used.

Issue #6329, opened 2023-08-24
Log hpx threads on forced shutdown


Pansysk75:
When an HPX process hangs (usually because some hpx thread hasn't terminated properly), the process is often killed with a termination signal.
In such cases, logging information about any alive hpx threads would really help me in debugging, especially when an issue is hard to reproduce locally.

@hkaiser I can certainly add this if you also think it makes sense, only could you please point me to where exit signals are handled? Thanks!

hkaiser:
@Pansysk75 Thanks! The termination handlers are here: https://github.com/STEllAR-GROUP/hpx/blob/bc993dd15b9252908e0b30b9ed5efd830cf99909/libs/core/runtime_local/src/runtime_local.cpp#L65-L174

Pansysk75:
Issue closed by merging of PR https://github.com/STEllAR-GROUP/hpx/pull/6340

PR #6340, opened 2023-09-01
Log alive hpx threads on exit

Pansysk75:
Adds logging when an HPX program is terminated suddenly (ie by a termination signal, addresses issue #6329).
I'll leave it as draft until I make up my mind on whether I think it is actually a good idea

### DIFF
--- a/libs/core/runtime_local/src/runtime_local.cpp
+++ b/libs/core/runtime_local/src/runtime_local.cpp
@@ -57,26 +57,35 @@
 #include <thread>
 #include <utility>
 
-namespace hpx{ namespace detail {
-    void log_all_runtime_threads()
+#if defined(HPX_HAVE_LOGGING)
+namespace hpx { namespace detail {
+    void try_log_runtime_threads()
     {
-        auto rt = hpx::get_runtime_ptr();
-        if (rt)
+        // This may be used in non-valid runtime states, let it fail silently
+        try
         {
+            auto rt = hpx::get_runtime_ptr();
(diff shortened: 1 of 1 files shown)


libs/core/runtime_local/src/runtime_local.cpp
```diff
@@ -57,6 +57,27 @@
...
+                    LTM_(debug).format(
+                        "Logging all runtime threads: pool({}), scheduler({}),"
+                        "thread({}), description({}), state({})",
+                        sched->get_parent_pool(), sched, id,
+                        td->get_description(), td->get_state().state());
+                    return true;
+                });
+        }
+    }
+}};    // namespace hpx::detail
```
hkaiser, line 79: Should we wrap this in a `#if defined(HPX_HAVE_LOGGING)` as well?

hkaiser: I'd also suggest to wrap all of the body into a `try {} catch (...) {}` as this is being invoked in a signal handler and we can't be sure that all of the runtime is in a consistent and valid state at this point any more.

hkaiser (changes requested):

libs/core/runtime_local/src/runtime_local.cpp
Pansysk75: Thanks!

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6341, opened 2023-09-01
Fix shared_future_continuation_order regression test

Pansysk75:
Fixes occasional failure of shared_future_continuation_order regression test.
If my understanding is correct, continuations are executed on newly spawned threads by default, which would not guarantee their order of execution. Launching with hpx::launch::sync should guarantee that no new thread is spawned, thus they are executed in sequence.
(@hkaiser We had fixed this in some PR in the past, but it seems it didn't make it into master)


hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6353, opened 2023-09-21
Enable using std::reference_wrapper with executor parameters

hkaiser:
@kmoham6 this should resolve the issues you were seeing


hkaiser:
bors merge

PR #6354, opened 2023-09-21
Fix broken links for header files in Public API page

dimitraka:


hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

hkaiser:
@dimitraka this causes the header files to be shown as e.g. `include/hpx/algorithm.hpp` instead of `hpx/algorithm.hpp` (see: https://hpx-docs.stellar-group.org/branches/master/html/api/public_api.html). I think you need to move the additional `include` to the base path section.

PR #6356, opened 2023-09-27
Fixed matrix multiplication example output

knuedd:
Fixed the example output of the matrix multiplication. For the shown random inputs the output was incorrect.

## Proposed Changes

Adopt the correct result for the given example 

No big deal but we don't want to suggest that HPX produces wrong results, do we? The example code produces the correct result matrix after all.


hkaiser (approved):
LGTM, thanks!

hkaiser:

PR #6357, opened 2023-09-27
Replace Custom Benchmarking Code with Nanobench

isidorostsa:
This Pull Request focuses on updating the benchmarks and performance testing functionality of HPX. This involves replacing the previous custom performance timing code with the [Nanobench](https://github.com/martinus/nanobench) library and incorporating the usage of Mustache templates for intuitive results formatting. 

The key benefits are:

- Easy parametrization of benchmark parameters.
- Standardized format of output json, directly compatible with [pyperf](https://github.com/psf/pyperf), and trivially customizable.

### DIFF
--- a/libs/core/testing/CMakeLists.txt
+++ b/libs/core/testing/CMakeLists.txt
@@ -20,6 +20,22 @@ set(testing_compat_headers
 # Default location is $HPX_ROOT/libs/testing/src
 set(testing_sources testing.cpp performance.cpp)
 
+# specify C++ standard
+set(CMAKE_CXX_STANDARD 17)
+
+# include FetchContent module
+include(FetchContent)
+
+# fetch nanobench
+FetchContent_Declare(
+    nanobench
+    GIT_REPOSITORY https://github.com/martinus/nanobench.git
+    GIT_TAG v4.3.11
+    GIT_SHALLOW TRUE
+)
+
+FetchContent_MakeAvailable(nanobench)
+
 include(HPX_AddModule)
 add_hpx_module(
   core testing
@@ -28,6 +44,6 @@ add_hpx_module(
   HEADERS ${testing_headers}
   COMPAT_HEADERS ${testing_compat_headers}
   MODULE_DEPENDENCIES hpx_assertion hpx_config hpx_format hpx_functional
-                      hpx_preprocessor hpx_util
+                      hpx_preprocessor hpx_util nanobench
   CMAKE_SUBDIRS examples tests
 )
--- a/libs/core/testing/include/hpx/testing/performance.hpp
+++ b/libs/core/testing/include/hpx/testing/performance.hpp
@@ -10,6 +10,7 @@
 #include <hpx/functional/function.hpp>
 
 #include <cstddef>
+#include <iostream>
 #include <string>
 
 namespace hpx::util {
@@ -18,5 +19,8 @@ namespace hpx::util {
         std::string const& exec, std::size_t const steps,
         hpx::function<void()>&& test);
 
-    HPX_CORE_EXPORT void perftests_print_times();
+    // templ is a mustache-style template for the output
+    // used by the nanobench reporter
+    HPX_CORE_EXPORT void perftests_print_times(
+        std::ostream& strm = std::cout, char const* templ = nullptr);
 }    // namespace hpx::util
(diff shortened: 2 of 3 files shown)


isidorostsa:
This still needs work on the fetching mechanism

libs/core/testing/src/performance.cpp
```diff
@@ -15,78 +16,42 @@
...
+namespace hpx::util {
 
-        public:
-            HPX_CORE_EXPORT void add(std::string const& name,
-                std::string const& executor, double time);
-        };
+    namespace detail {
 
-        json_perf_times& times()
+        char const* nanobench_hpx_template() noexcept
```
hkaiser, line 28: Can we make this constexpr?

CMakeLists.txt
```diff
@@ -1142,6 +1142,14 @@ if(HPX_WITH_NETWORKING)
...
     ADVANCED
   )
+  hpx_option(
+    HPX_WITH_FETCH_NANOBENCH
+    BOOL
+    "Use FetchContent to fetch Nanobench. (default: OFF)"
+    OFF
+    CATEGORY "Build Targets"
+    ADVANCED
+  )
```
hkaiser, line 1152: Should we fetch this by default only if `HPX_WITH_TESTS_BENCHMARKS=On`?

libs/core/testing/CMakeLists.txt
```diff
@@ -28,6 +29,6 @@ add_hpx_module(
   HEADERS ${testing_headers}
   COMPAT_HEADERS ${testing_compat_headers}
   MODULE_DEPENDENCIES hpx_assertion hpx_config hpx_format hpx_functional
-                      hpx_preprocessor hpx_util
+                      hpx_preprocessor hpx_util nanobench
```
hkaiser, line 32: Will this work if nanobench is not fetched?

libs/core/testing/include/hpx/testing/performance.hpp
```diff
@@ -10,6 +10,7 @@
 #include <hpx/functional/function.hpp>
 
 #include <cstddef>
+#include <iostream>
```
hkaiser, line 13: Is `#include <iosfwd>` sufficient?

libs/core/testing/src/performance.cpp
```diff
@@ -15,78 +16,42 @@
...
-        class json_perf_times
-        {
-            using key_t = std::tuple<std::string, std::string>;
-            using value_t = std::vector<double>;
-            using map_t = std::map<key_t, value_t>;
+#include <nanobench.h>
 
-            map_t m_map;
+#define NANOBENCH_EPOCHS 24
+#define NANOBENCH_WARMUP 40
```
hkaiser, line 22: Can you convert this to `constexpr` variables?

libs/core/testing/include/hpx/testing/performance.hpp
```diff
@@ -18,5 +19,8 @@ namespace hpx::util {
         std::string const& exec, std::size_t const steps,
         hpx::function<void()>&& test);
 
-    HPX_CORE_EXPORT void perftests_print_times();
+    // templ is a mustache-style template for the output
+    // used by the nanobench reporter
+    HPX_CORE_EXPORT void perftests_print_times(
+        std::ostream& strm = std::cout, char const* templ = nullptr);
```
hkaiser, line 25: Could we keep `std::cout` out of the API?

isidorostsa: Sure thing! Which of the following do you think is the better way to do it?

1) Leave `std::cout` as the default stream, but handle it on the `.cpp` file with polymorphism instead of the header. 
2) Do not provide a default stream.

The original API does not accept a stream parameter

CMakeLists.txt
isidorostsa: Yes, so if hpx is built with benchmarks on, by default we fetch nanobench unless the user species the nanobench root directory. 

Should the default case be the non-nanobench version?

libs/core/testing/CMakeLists.txt
isidorostsa: I don't think so, this part is really incomplete, I tried fetching nanobench only for internal use by this module but while it builds, cmake throws some non fatal errors, as the nanobench module is not being properly exported. I am working on that

libs/core/testing/include/hpx/testing/performance.hpp
hkaiser: Let's add another overload that allows to specify the ostream and handle the default case in the source file.

SAtacker:
> This still needs work on the fetching mechanism

A while ago I had tried using google benchmark https://github.com/SAtacker/hpx-template, even if you don't want to change to google bench the cmake fetch content stuff should be similar

isidorostsa:
Continued by https://github.com/STEllAR-GROUP/hpx/pull/6448

PR #6364, opened 2023-10-12
Relocation variants

isidorostsa:
This PR introduces additional features related to relocation, aimed at both internal and public use. The specifics are:

- The algorithm `uninitialized_relocate_backward` and its underlying lower-level primitives have been implemented. Corresponding tests for these algorithms have also been added.

- The function `uninitialized_relocate` has been refactored and it now uses a sentinel based for loop, instead of calculating the number of iterations and internally calling `uninitialized_relocate_n`.

The primitives of `uninitialized_relocate_backwards` will be utilized internally for the `hpx::small_vector::insert()` method.


libs/core/type_support/include/hpx/type_support/uninitialized_relocate_n_primitive.hpp
```diff
@@ -14,13 +14,13 @@
 #include <cstring>
 #include <type_traits>
 
-#if defined(HPX_HAVE_P1144_STD_RELOCATE_AT)
+#if defined(__cpp_lib_trivially_relocatable)
```
hkaiser, line 17: I'd suggest leaving this as a feature test instead of relying on a library supplied macro. Please feel free however to use this macro as part of the feature test...

isidorostsa: Okay, on that note, in case P1144 is present, would it make sense to remap our type_support primitives to the sequential relocation functions in the std? (`std::uninitialized_relocate`, `std::uninitialized_relocate_n`, `std::uninitialized_relocate_backward`)?

hkaiser: Yes, we usually simply import the standard versions into namespace hpx, thus avoiding changes to other code.

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
```diff
@@ -469,6 +467,64 @@ namespace hpx::parallel {
...
+                BiIter2 dest_last) noexcept(hpx::experimental::util::detail::
+                    relocation_traits<BiIter1,
+                        BiIter2>::is_noexcept_relocatable_v)
+            {
+                auto count = std::distance(first, last);
+
+                auto dest_first = std::prev(dest_last, count);
+
+                return parallel_uninitialized_relocate_n(
+                    HPX_FORWARD(ExPolicy, policy), first, count, dest_first);
```
Quuxplusone, line 523: Whether this is OK depends on the preconditions of your `parallel` algorithms. I _expect_ that this is not OK. The important thing IMO is that we want this to work:
```
Widget a[10] = {1,2,3,4,5,6,7,8,9,destroyed};
std::uninitialized_relocate_backward(a, a+9, a+10);
assert(a == {destroyed,1,2,3,4,5,6,7,8,9});
```
(modulo syntax errors and language-lawyering over the lifetimes of those Widget objects). If you copy them in the forward direction, the output will be something like
```
assert(a == {destroyed,1,1,1,1,1,1,1,1,1});
```
So the idea of having `uninitialized_relocate_backward` dispatch to `uninitialized_relocate_n` smells wrong to me.

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
```diff
@@ -644,5 +700,87 @@ namespace hpx::experimental {
...
+            BiIter1 first, BiIter1 last,
+            BiIter2 dest_last) noexcept(util::detail::relocation_traits<BiIter1,
+            BiIter2>::is_noexcept_relocatable_v)
+        {
+            static_assert(hpx::traits::is_bidirectional_iterator_v<BiIter1> &&
+                "The 'first' and 'last' arguments must meet the requirements "
+                "of input iterators.");
+            static_assert(hpx::traits::is_bidirectional_iterator_v<BiIter2>,
+                "The 'dest_last' argument must meet the requirements of a "
+                "forward iterator.");
```
Quuxplusone, line 726:
```suggestion
                "bidirectional iterator.");
```

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
```diff
@@ -644,5 +700,87 @@ namespace hpx::experimental {
...
+            )>
+        // clang-format on
+        friend BiIter2 tag_fallback_invoke(uninitialized_relocate_backward_t,
+            BiIter1 first, BiIter1 last,
+            BiIter2 dest_last) noexcept(util::detail::relocation_traits<BiIter1,
+            BiIter2>::is_noexcept_relocatable_v)
+        {
+            static_assert(hpx::traits::is_bidirectional_iterator_v<BiIter1> &&
+                "The 'first' and 'last' arguments must meet the requirements "
+                "of input iterators.");
```
Quuxplusone, line 723:
```suggestion
                "of bidirectional iterators.");
```

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
```diff
@@ -644,5 +700,87 @@ namespace hpx::experimental {
...
+            )>
+        // clang-format on
+        friend typename hpx::parallel::util::detail::algorithm_result<ExPolicy,
+            BiIter2>::type
+        tag_fallback_invoke(uninitialized_relocate_backward_t,
+            ExPolicy&& policy, BiIter1 first, BiIter1 last,
+            BiIter2 dest_last) noexcept(util::detail::relocation_traits<BiIter1,
+            BiIter2>::is_noexcept_relocatable_v)
+        {
+            static_assert(hpx::traits::is_input_iterator_v<BiIter1>,
```
Quuxplusone, line 758:
```suggestion
            static_assert(hpx::traits::is_bidirectional_iterator_v<BiIter1>,
```

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
```diff
@@ -644,5 +700,87 @@ namespace hpx::experimental {
...
+                "The 'first' and 'last' arguments must meet the requirements "
+                "of bidirectional iterators.");
+            static_assert(hpx::traits::is_forward_iterator_v<BiIter2>,
```
Quuxplusone, line 761:
```suggestion
            static_assert(hpx::traits::is_bidirectional_iterator_v<BiIter2>,
```

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
```diff
@@ -644,5 +700,87 @@ namespace hpx::experimental {
...
+                "ill-formed");
+
+            auto count = std::distance(first, last);
+
+            // if count is representing a negative value, we do nothing
+            if (hpx::parallel::detail::is_negative(count))
+            {
+                return parallel::util::detail::algorithm_result<ExPolicy,
+                    BiIter2>::get(HPX_MOVE(dest_last));
+            }
```
Quuxplusone, line 776: This would be "library UB" according to P1144. You're welcome to treat it as a no-op (documented or undocumented), but it might be more appropriate to assert-fail or something, I don't know.

libs/core/algorithms/tests/unit/algorithms/uninitialized_relocate_backward.cpp
```diff
@@ -0,0 +1,308 @@
...
+        {
+            hpx::construct_at(ptr1 + i, 1234);
+            hpx::construct_at(ptr2 + i, 5678);
+        }
+
+        // N objects constructed
+        HPX_TEST(trivially_relocatable_struct::count == 2 * N);
+
+        // relocate M objects to ptr2
+        uninitialized_relocate_backward(ptr1, ptr1 + M, ptr2 + N);
```
Quuxplusone, line 157: Perhaps you should precede this line with `std::destroy(ptr2 + M - N, ptr2 + N)` so that you're not relocating-over-top-of any live objects.

Your `trivially_relocatable_struct` is trivially destructible, so this doesn't matter to the physical outcome. But, since you're so careful to (trivially) `std::destroy` certain subranges on lines 178–179 below, you might as well do the same here.

libs/core/algorithms/tests/unit/algorithms/uninitialized_relocate_backward.cpp
```diff
@@ -0,0 +1,308 @@
...
+int trivially_relocatable_struct::count = 0;
+int trivially_relocatable_struct::move_count = 0;
+int trivially_relocatable_struct::dtor_count = 0;
+
+HPX_DECLARE_TRIVIALLY_RELOCATABLE(trivially_relocatable_struct);
+static_assert(is_trivially_relocatable_v<trivially_relocatable_struct>);
+
+struct non_trivially_relocatable_struct
+{
+    static int count;
```
Quuxplusone, line 54: For extra coverage: Instead of `count`, you _could_ store `static std::set<non_trivially_relocatable_struct*> objects_;`, `insert` in the ctor, `erase` in the dtor, and verify that you never construct over top of a live object nor destroy a dead object.

libs/core/algorithms/tests/unit/algorithms/uninitialized_relocate_backward.cpp
```diff
@@ -0,0 +1,308 @@
...
+        {
+        }
+
+        // K move constructors were called, and then the last one throws
+        HPX_TEST(non_trivially_relocatable_struct_throwing::move_count == K);
+
+        // K - 1 destructors were called to balance out the move constructors
+        // (- 1 because the last move constructor throws)
+        // and then M + 1 destructors were called: K on the old range and
+        // M - (K - 1) = M - K + 1 on the new range
```
Quuxplusone, line 278: FWIW, I found this explanation hard to grasp. I would prefer to say that there are M+K-1 objects destroyed because we successfully relocated K-1 objects before the failure (so that's K-1 new objects we need to destroy), plus when we fail we destroy the entire source range (so that's M objects destroyed _at some point_, although of course some of those destructions have already happened as part of the K-1 successful relocations).
...But you got M+K, not M+K-1. Maybe I'm missing one somewhere?

Quuxplusone (commented):
Some comments. I didn't look closely from the point of my last comment onward: reviewer silence doesn't necessarily mean endorsement. :)

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
isidorostsa: You are correct in pointing this out, however I should look into how (if at all) parallelizable this would be, while retaining the correct ordering.

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
```diff
@@ -437,11 +437,9 @@ namespace hpx::parallel {
                 FwdIter dest) noexcept(hpx::experimental::util::detail::relocation_traits<
                     InIter1, FwdIter>::is_noexcept_relocatable_v)
             {
-                auto count = std::distance(first, last);
-
                 return util::in_out_result<InIter1, FwdIter>{first,
-                    hpx::experimental::util::uninitialized_relocate_n_primitive(
-                        first, count, dest)};
+                    hpx::experimental::util::uninitialized_relocate_primitive(
+                        first, last, dest)};
```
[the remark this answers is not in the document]
isidorostsa: 1. Great catch, you are correct
2. Will do

libs/core/type_support/include/hpx/type_support/uninitialized_relocate_n_primitive.hpp
isidorostsa: Reverted this and mapped `uninitialized_relocate(_n, _backward)_primitive` --> `uninitialized_relocate(_n, _backward)` in https://github.com/STEllAR-GROUP/hpx/pull/6364/commits/2474a61c5f6f32063de5b052df3e4e7c7e2b285f

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
Quuxplusone: Naïvely, it's not parallelizable at all _unless_ you can prove (by pointer comparisons) that the two ranges don't overlap. If they don't overlap, then this implementation is fine. It occurs to me that HPX already has a `std::copy{,_backward}` implementation; you should look at what it does, and copy that strategy.
(It might be that HPX's `std::copy{,_backward}` also completely ignore the overlapping issue. If so, then I think it would be perfectly fine for you to copy that ignorance here. `uninitialized_relocate_backward` is not intended to be any _more_ difficult to implement than `copy_backward`.)

hkaiser:
@isidorostsa could you please rebase this and resolve the conflicts while you do?

hkaiser:
> > @isidorostsa could you please rebase this and resolve the conflicts while you do?
> 
> @hkaiser, I rebased and resolved the conflicts. Before we finalize the merge, there are some design choices I'd like to discuss:
> 
> 1. [Relocation variants #6364 (comment)](https://github.com/STEllAR-GROUP/hpx/pull/6364#discussion_r1357350896): should negative input sizes be treated as a no-op, similar to our move/copy algorithms? Arthur suggests using an assertion failure, and I'm leaning towards that as well. What do you think?

Yes, sounds good.

> 2. [Relocation variants #6364 (comment)](https://github.com/STEllAR-GROUP/hpx/pull/6364#discussion_r1357347816): Does it make sense to have the `backward` to not actually perform operations in reverse order when executed in parallel? So when running in parallel the only difference between the default and the backward algorithm is the calling parameters

Using a backwards algorithm usually has a reason, e.g., you know that doing an inplace operation would overwrite values before they can be used if the operation is done in a certain way. I don't think we should strat cheating here ;-)

> 3. [Codacy false alarm](https://app.codacy.com/gh/STEllAR-GROUP/hpx/pullRequest?prid=12815770&bid=38242060): There are some warnings here that are inaccurate. Do you know how we can disable those? The code it is shouting about is essentially:
> 
> ```c++
> void foo() noexcept(!(mode == throwing)){
>     if constexpr(mode == throwing) { throw; }
> }
> ```

I can take care of those.

isidorostsa:
> you know that doing an inplace operation would overwrite values before they can be used if the operation is done in a certain way

In case we want to preserve parallelization while not overwriting objects I have two techniques in mind:

A. Execute the overlapping portion of the relocation sequentially and do what is left in parallel. However, I suspect that in most common use cases, the ranges will largely overlap, rendering the operation mostly sequential.
B. Break down the overlapping relocation into **two** non-overlapping ones, using an intermediate buffer. But if memory speed is a limiting factor, this could actually reduce performance. Additionally, the unexpected allocation of the buffer could be an issue.

So it may be more practical to opt for a purely sequential implementation. Is there an alternative solution that I'm overlooking?

hkaiser:
retest lsu

hkaiser:
@isidorostsa what's the status for this PR?

isidorostsa:
> @isidorostsa what's the status for this PR?

I think last time I pushed the hpx tests were broken, which is why we have not merged yet

But locally the tests related to this are passing.

hkaiser:
> > @isidorostsa what's the status for this PR?
> 
> I think last time I pushed the hpx tests were broken, which is why we have not merged yet
> 
> But locally the tests related to this are passing.

The broken tests are unrelated. So this is good to go, then?

Pansysk75:
>
> The broken tests are unrelated. So this is good to go, then?

Oh, the tests aren't really running. Take a look here: https://cdash.cscs.ch/build/118631
It's related to this: https://github.com/STEllAR-GROUP/hpx/pull/6377
I apologize I didn't make that very clear.
I have put fixing them on the top of my queue.

isidorostsa:
> I have put fixing them on the top of my queue.

[P1144R9](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p1144r9.html) is out already, so we do need to rush merging this!

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
```diff
@@ -644,5 +830,87 @@ namespace hpx::experimental {
...
+                "of bidirectional iterators.");
+            static_assert(hpx::traits::is_bidirectional_iterator_v<BiIter2>,
+                "The 'dest' argument must meet the requirements of a "
+                "bidirectional iterator.");
+            static_assert(util::detail::relocation_traits<BiIter1,
+                              BiIter2>::valid_relocation,
+                "Relocating from this source type to this destination type is "
+                "ill-formed");
+
+            auto count = std::distance(first, last);
```
isidorostsa, line 899: @hkaiser @Pansysk75 @gonidelis Considering https://github.com/STEllAR-GROUP/hpx/pull/6364#issuecomment-1771644408, it may be sensible to include a warning here such as:
"Calling `uninitialized_relocate_backward` with a non-sequential execution policy does not guarantee the order of execution of the relocations to be backward. Use on overlapping ranges could yield undefined behavior."

However, I am not sure how we would issue such a warning.

hkaiser: There is no need to do that. The same warning applies to `std::copy` and `std::move`. There, it is up to the user to ensure these preconditions.

isidorostsa: @Quuxplusone Thank you for the detailed write-up, and sorry for my late response. 

I think my previous comment:  https://github.com/STEllAR-GROUP/hpx/pull/6364#issuecomment-1771644408, is incorrect and there does exist a way to parallelize the relocation of overlapping ranges, maybe like in the following image: (example for 2 threads and an overlapping offset of 1) 

image

Supposing this is true, I see no reason to avoid offering this as a feature. Furthermore, I don't see why we would need non-overlapping ranges to be a precondition for relocation!

> I think you could fix the latter inconsistency by having hpx::parallel::detail::parallel_uninitialized_relocate_n check for overlap

I think for the time being we will do this, only parallelizing the guaranteed safe case (contiguous iterator + no overlap), and have the rest be sequenced, but definitely add proper parallelization support on the todo list. @hkaiser What do you think?

> maybe you'd rather that P1144's std::uninitialized_relocate_backward should forbid overlap .... If so, I'd like to know what's your plan for implementing things like {parallel_,}vector::erase

For the time being we do not implement parallel data structures, but there is a draft PR for using relocation inside `hpx::detail::small_vector` (which is like `static_vector` except it grows to dynamic storage after running out of static space)  (https://github.com/isidorostsa/hpx_fork/pull/9).

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
isidorostsa: From https://github.com/STEllAR-GROUP/hpx/pull/6364#discussion_r1413166378:
> In standard C++20 we have
>
>   -  the non-parallel std::copy_backward, which forbids overlap
>   - the ExecutionPolicy overload of std::copy_backward, which also forbids overlap
> 
> and thus {parallel_,}vector::erase fundamentally isn't allowed to use std::copy_backward.
> D1144R10 currently proposes
>
>   - a non-parallel std::uninitialized_relocate_backward which permits overlap
>   - an ExecutionPolicy overload of std::uninitialized_relocate_backward, which also permits overlap

And noting that:
>uninitialized_relocate_backward is not intended to be any more difficult to implement than copy_backward.

Are you proposing permitting overlapping ranges with parallel `copy`, or breaking the symmetry between `copy` and `relocation`?

I can experiment with implementing and benchmarking a parallel relocation algorithm for overlapping ranges.

Quuxplusone: I'm proposing breaking the symmetry between `copy` and `uninitialized_relocate`. (But I admit that's bad; and I also admit I still don't know why `copy` forbids overlap in the first place.)

Quuxplusone: My view is:
- `uninitialized_relocate{,_n}` should permit "shift left" overlap, as needed by `vector::erase`.
- `uninitialized_relocate_backward` should permit "shift right" overlap, as needed by `vector::insert`.

My proposed wording reflects this since [P1144R6](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1144r6.html#wording-uninitialized.relocate): the behavior of `uninitialized_relocate` is "as if" by a plain old for-loop relocating each element in sequence from first to last, left to right.

isidorostsa:
retest lsu

isidorostsa:
>I think for the time being we will do this, only parallelizing the guaranteed safe case (contiguous iterator + no overlap), and have the rest be sequenced, but definitely add proper parallelization support on the todo list. @hkaiser What do you think?

@hkaiser this is okay to merge if you agree with this

isidorostsa:
@hkaiser I noticed that after the merge some tests are failing and I will get onto fixing that as soon as possible.

hkaiser:
@isidorostsa I'll revert this merge now. Let's fix the issues with a new PR.

hkaiser:
> @isidorostsa I'd like to revert this merge. Let's fix the issues with a new PR. Can you please do that?

I take this back. The issues are unrelated to your changes. I will fix them separately.

isidorostsa:
> I take this back. The issues are unrelated to your changes. I will fix them separaty.

Thank you for letting me know. Unfortunately I did not get a chance to review it myself due to traveling for the visa.

PR #6367, opened 2023-10-14
Fix parallel relocation test

isidorostsa:
In the original test there where two errors:

1. An atomic variable was fetched in a way that introduced race conditions, skipping the `throw` portion of the code.
2. The testing was more strict than necessary, making guesses about the inner workings of `partitioner_with_cleanup`. Testing that every created object is destroyed in case of an exception is enough to abide by P1144.


hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6372, opened 2023-10-18
Fix resource partitioner tests for small thread count

Pansysk75:
Fixes certain resource partitioner tests which would fail when --hpx:threads was set to a number <4.
Should fix current failures on MacOS CI

### DIFF
--- a/libs/core/resource_partitioner/tests/unit/named_pool_executor.cpp
+++ b/libs/core/resource_partitioner/tests/unit/named_pool_executor.cpp
@@ -35,21 +35,21 @@ void dummy_task(std::size_t n, std::string const& text)
 
 int hpx_main()
 {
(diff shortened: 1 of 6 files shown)


hkaiser:
@Pansysk75 Thanks a lot! This however raises the perpetual question 'How could this ever have worked before?' ;-)

Pansysk75:
> @Pansysk75 Thanks a lot! This however raises the perpetual question 'How could this ever have worked before?' ;-)

Good question, since MacOS CI is hard-coded to always use 3 (why?) threads. My only guess so far is `hpx::threads::hardware_concurrency()` returning 3 some of the time, but other times some larger number (and in those cases it would fail).

hkaiser:
@Pansysk75 I think this needs a rebase onto top of master to resolve the gasnet issues.

Pansysk75:
retest lsu

hkaiser:
retest lsu

PR #6373, opened 2023-10-19
Fixing communicator API, adding docs

hkaiser:
@dimitraka this should fix the issues related to the communicator API we discussed


hkaiser:
retest lsu

PR #6378, opened 2023-11-01
Allowing to disable signal handlers

hkaiser:


hkaiser:
@JiakunYan this should resolve the signal handler issues we discussed.

PR #6379, opened 2023-11-01
Adding feature, Range Based Lock

Johan511:
Adding feature, Range Based lock to HPX

API:
`range_lock rl;`
`lockId = rl.lock(rangeBegin, rangeEnd);`
`rl.unlock(lockId);`

Pending Task:
Adding tests


libs/core/synchronization/include/hpx/synchronization/range_lock.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include <functional>
+#include <mutex>
+
+namespace hpx::synchronization {
+    using range_lock =
+        hpx::synchronization::detail::RangeLock<hpx::spinlock, std::lock_guard>;
+}
+
+// Lock guards for range_lock
+namespace hpx::synchronization {
```
hkaiser, line 15: One namespace scope is sufficient, no need to reopen it for each entity.

libs/core/synchronization/include/hpx/synchronization/range_lock.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+    template <typename RangeLock>
+    class range_guard
```
hkaiser, line 18: Usually, lock guard types are movable, but not copyable.

libs/core/synchronization/include/hpx/synchronization/detail/range_lock_impl.hpp
```diff
@@ -0,0 +1,106 @@
...
+#include <atomic>
+#include <boost/container/flat_map.hpp>
+#include <cstddef>
+#include <utility>
+#include <vector>
+
+namespace hpx::synchronization::detail {
+
+    template <typename Lock, template <typename> typename Guard>
+    class RangeLock
```
hkaiser, line 14: Can we call this `range_mutex` instead? The name `lock` is normally used for the lock guards, not the actual 'locks'.

libs/core/synchronization/include/hpx/synchronization/range_lock.hpp
```diff
@@ -0,0 +1,110 @@
...
+            lockId = lockRef.get().try_lock(begin, end);
+        }
+
+        void unlock()
+        {
+            lockRef.get().unlock(lockId);
+            lockId = 0;
+        }
+
+        void swap(std::unique_lock<RangeLock>& uLock)
```
hkaiser, line 80:
```suggestion
        void swap(range_unique_lock& uLock)
```

hkaiser (changes requested):
Could you please add more tests, especially some tests that actually exercise the multi-threaded use of the new constructs?

Johan511:
More comprehensive tests on the way @hkaiser. Will add them soon.

libs/core/synchronization/include/hpx/synchronization/detail/range_lock_impl.hpp
```diff
@@ -0,0 +1,106 @@
...
+        Lock mtx;
+        std::size_t counter = 0;
+        MapTy<std::size_t, std::pair<std::size_t, std::size_t>> rangeMap;
+        MapTy<std::size_t, std::shared_ptr<std::atomic_bool>> waiting;
+
+    public:
+        std::size_t lock(std::size_t begin, std::size_t end);
+        std::size_t try_lock(std::size_t begin, std::size_t end);
+        void unlock(std::size_t lockId);
+    };
```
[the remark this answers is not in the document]
Johan511: > Usually, mutex types are movable, but not copyable.
Sure, just need to delete the copy constructor and define a move constructor for it I believe. Do I need to worry about thread safety while moving? like say another thread trying to lock? 

> This will most likely also allow to replace the std::shared_ptr<std::atomic_bool> with a simpler solution based on unique_ptr.
I am unsure how adding move constructor would help change the flag from `shared_ptr` to `unique_ptr`. 
To summarise, if a thread locks range [0, 20), it also creates an `atomic_bool` flag. 
So if another thread (2) wants to lock an intersecting range, say [10, 30). It spins on the `atomic_bool` flag before trying to acquire the lock again, trying to acquire the range_lock is an expensive operation which we want to avoid unless we are confident the range_lock can be acquired.

When performing unlock, we delete the atomic_bool flag while the other threads might still be spinning on it, using `unique_ptr` would delete the atomic_bool causing a segmentation fault, `shared_ptr` will not cause this issue.

libs/core/synchronization/include/hpx/synchronization/range_mutex.hpp
```diff
@@ -25,89 +25,114 @@ namespace hpx::synchronization {
...
+        std::reference_wrapper<RangeMutex> mutex_ref;
+        std::size_t lock_id = 0;
 
     public:
         range_guard(RangeMutex& lock, std::size_t begin, std::size_t end)
-          : lockRef(lock)
+          : mutex_ref(lock)
         {
-            lockId = lockRef.get().lock(begin, end);
+            lock_id = mutex_ref.get().lock(begin, end);
```
hkaiser, line 35: Can this be done in the member initialization part of the constructor?

libs/core/synchronization/include/hpx/synchronization/range_mutex.hpp
```diff
@@ -25,89 +25,114 @@ namespace hpx::synchronization {
...
         }
         ~range_guard()
         {
-            lockRef.get().unlock(lockId);
+            mutex_ref.get().unlock(lock_id);
+        }
+
+        range_guard(range_guard<RangeMutex>&) = delete;
+        range_guard<RangeMutex>& operator=(range_guard<RangeMutex>&) = delete;
```
hkaiser, line 43: If you were planning to delete the copy constructor and copy assignment operator, then this should be:
```suggestion
        range_guard(range_guard const&) = delete;
        range_guard& operator=(range_guard const&) = delete;
```
(there is no need to explicit specify the template argument either).

libs/core/synchronization/include/hpx/synchronization/range_mutex.hpp
```diff
@@ -25,89 +25,114 @@ namespace hpx::synchronization {
...
+
+        range_guard(range_guard<RangeMutex>&& rhs_lock)
+        {
+            mutex_ref.get().unlock(lock_id);
```
hkaiser, line 47: `lock_id` was never initialized for the object that the argument is being moved to.

libs/core/synchronization/include/hpx/synchronization/range_mutex.hpp
```diff
@@ -25,89 +25,114 @@ namespace hpx::synchronization {
...
-            lockId = lockRef.get().try_lock(begin, end);
+            lock_id = mutex_ref.get().try_lock(begin, end);
         }
 
         void unlock()
         {
-            lockRef.get().unlock(lockId);
-            lockId = 0;
+            mutex_ref.get().unlock(lock_id);
+            lock_id = 0;
```
hkaiser, line 106: Shouldn't this be made thread-safe?

libs/core/synchronization/include/hpx/synchronization/range_mutex.hpp
```diff
@@ -25,89 +25,114 @@ namespace hpx::synchronization {
...
+            mutex_ref.get().unlock(lock_id);
+            mutex_ref = rhs_lock.mutex_ref;
+            lock_id = rhs_lock.lock_id;
+            rhs_lock.lock_id = 0;    // invalidating rhs_lock
         }
 
         void lock(std::size_t begin, std::size_t end)
         {
-            lockId = lockRef.get().lock(begin, end);
+            lock_id = mutex_ref.get().lock(begin, end);
```
hkaiser, line 95: What should happen if `lock_id != 0` before the locking operation? Also, should `lock_id` be an atomic?

libs/core/synchronization/include/hpx/synchronization/range_mutex.hpp
Johan511: How are the template arguments deduced?

hkaiser: > How are the template arguments deduced?

If you use the template name inside the template definition you can leave off the template arguments.

libs/core/synchronization/include/hpx/synchronization/range_mutex.hpp
Johan511: @hkaiser aren't lock guards supposed to be thread local usually? I am unsure how lock_id can be racing if lock_guards are local.

libs/core/synchronization/include/hpx/synchronization/range_mutex.hpp
Johan511: The [cppref](https://en.cppreference.com/w/cpp/thread/unique_lock/lock#exceptions) page for unique_lock mentions the following

> Exceptions
Any exceptions thrown by mutex()->lock().
If there is no associated mutex, [std::system_error](https://en.cppreference.com/w/cpp/error/system_error) with an error code of [std::errc::operation_not_permitted](https://en.cppreference.com/w/cpp/error/errc).
If the mutex is already locked by this unique_lock (in other words, [owns_lock](https://en.cppreference.com/w/cpp/thread/unique_lock/owns_lock) is true), [std::system_error](https://en.cppreference.com/w/cpp/error/system_error) with an error code of [std::errc::resource_deadlock_would_occur](https://en.cppreference.com/w/cpp/error/errc).


I guess we should do the same

hkaiser: You are certainly right, I misunderstood things.

libs/core/synchronization/include/hpx/synchronization/range_mutex.hpp
hkaiser: You are certainly right, my fault.

Johan511:
@hkaiser do we have a concurrent hash map like the one by [boost](https://www.boost.org/doc/libs/develop/libs/unordered/doc/html/unordered.html#benchmarks_boostconcurrent_flat_map) or the one by folly?

hkaiser:
> @hkaiser do we have a concurrent hash map like the one by [boost](https://www.boost.org/doc/libs/develop/libs/unordered/doc/html/unordered.html#benchmarks_boostconcurrent_flat_map) or the one by folly?

No we don't have something like that.

hkaiser:
@Johan511 I know that currently the reporting of the build logs for the HPX CIs is broken (we're working on fixing this). For this reason, here is the error message that makes the builders fail:
```
/.../hpx/libs/core/synchronization/include/hpx/synchronization/range_mutex.hpp:71:25: error: no member named 'get' in 'hpx::synchronization::detail::range_mutex<hpx::detail::spinlock<true>, std::lock_guard>'
```

biddisco:
I have not looked at the implementation (I will examine it later), but I'm curious as to the use case for the range-lock. Is this work inspired by the GSoC project I proposed a few years ago (with an astrophysics particle code in mind), or do you have another use case that requires it? (or perhaps it was "just for fun" as an interesting project to try?)

Johan511:
Hi @biddisco, this is a fun piece of work inspired by the GSoC project you had proposed. I didn't have an use-case in mind, I had tested out it's performance based on the amount of time it took to execute the [test](https://github.com/STEllAR-GROUP/hpx/pull/6379/files#diff-905410d0e5b06c7829f2bc13c64f542e58852ee0cf01e021ccd4f6a73725cd9bR131-R144). It'd be great if I could test the impact of the code on a more real life scenario.

Johan511:
@hkaiser can you have a look at the CI? I am not able to view the logs on the jenkins CI.
Also is there any reference I could use to write documentation? Maybe some sort of specialized mutex somewhere in HPX.

hkaiser:
> @hkaiser can you have a look at the CI? I am not able to view the logs on the jenkins CI. Also is there any reference I could use to write documentation? Maybe some sort of specialized mutex somewhere in HPX.

Sory for the CI issues. We're currently working on a solution for this problem.

PR #6381, opened 2023-11-07
Consistently treat conflicting parameters provided by executors and parameter objects

hkaiser:
This introduces a breaking change as it changes the preferences for which of the parametrizations for execution policies takes a preference. With this change parameter values introduced through execution parameter objects take always precedence. 

This change also enforces using `tag_invoke` for exposing parameter values from executors and `tag_override_invoke` for exposing those from execution parameter objects.


hkaiser:
retest lsu

PR #6387, opened 2023-11-11
Need to break out of the loop when searching the suffixes.

stevenrbrandt:
Fixes issue with hpxcxx. The program does not break out of the outer loop after it finds a matching suffix like it should.


hkaiser (approved):
LGTM, thanks!

PR #6393, opened 2023-12-02
Optionally disable caching allocator

hkaiser:


NK-Nikunj:
https://github.com/STEllAR-GROUP/hpx/pull/6393/files#diff-bd1f43d73c88a8bdda46746af0e6bcdb9612ac716943cdc86a272028e326fc5dR116

Why use `std::stack` here when the operators involved in thread cache is the equivalent of a push_back and a pop_back? A `std::vector` should be better pick, no?

hkaiser:
> https://github.com/STEllAR-GROUP/hpx/pull/6393/files#diff-bd1f43d73c88a8bdda46746af0e6bcdb9612ac716943cdc86a272028e326fc5dR116
> 
> Why use `std::stack` here when the operators involved in thread cache is the equivalent of a push_back and a pop_back? A `std::vector` should be better pick, no?

Does it make a difference? But sure, please feel free to create a PR for this.

Issue #6392, opened 2023-11-22
[Feature] Install dependencies using CMake


SAtacker:
## Expected Behavior

- Follow https://github.com/SAtacker/hpx-template cmake fetch content to download hwloc and boost dependencies depending upon the platform. 

- Allow user to have configurable behaviors i.e. system can find boost which is already existing. Enable users to fetch content boost and hwloc using`HPX_WITH_FETCH_BOOST=ON` and `HPX_WITH_FETCH_HWLOC=ON`

- Add tests for the same

CC @Shanmukh-C

vrnimje:
Hello @SAtacker , is this issue open for anyone to work on, or only Shanmukh-C is supposed to work on it?
I have decent experience with CMakeLists.txt
And for testing out options, I created a clone of your repository: https://github.com/vrnimje/hpx-temp/tree/cmake
and tested out some workflows as well: https://github.com/vrnimje/hpx-temp/actions

SAtacker:
> Hello @SAtacker , is this issue open for anyone to work on, or only Shanmukh-C is supposed to work on it? I have decent experience with CMakeLists.txt And for testing out options, I created a clone of your repository: https://github.com/vrnimje/hpx-temp/tree/cmake and tested out some workflows as well: https://github.com/vrnimje/hpx-temp/actions

Yes, you can definitely work on this! Seems like you have already looked into it. 
(We have plenty of things to work on so you folks don’t have to worry)
What’s your id for discord on the IRC for STE||AR Group?

vrnimje:
> Yes, you can definitely work on this! Seems like you have already looked into it. (We have plenty of things to work on so you folks don’t have to worry) What’s your id for discord on the IRC for STE||AR Group?

Thanks for the confirmation. Will get started with working on the issue right away. Just had a couple of queries:
1. Changes should be made in the main [`CMakeLists.txt`](https://github.com/STEllAR-GROUP/hpx/blob/master/CMakeLists.txt) of this repository, right?
2. Do `tests` (mentioned in 3rd pt. of the issue), mean workflow build tests?

And here is my Discord ID: `verve0304`

SAtacker:
> Changes should be made in the main [CMakeLists.txt](https://github.com/STEllAR-GROUP/hpx/blob/master/CMakeLists.txt) of this repository, right?

[...]

PR #6394, opened 2023-12-03
Adding option for installing Hwloc using CMake FetchContent

vrnimje:
References #6392 

## Proposed Changes

  - Added `HPX_WITH_FETCH_HWLOC` option for fetching Hwloc using CMake FetchContent
  - Added CI tests for the same, on Linux, macOS and Windows platforms

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2322,10 +2322,7 @@ endif()
 
 if(HPX_WITH_EXAMPLES)
   add_hpx_pseudo_target(examples)
-  if(HPX_WITH_FETCH_HWLOC
-     AND "${CMAKE_SYSTEM_NAME}" STREQUAL "Windows"
-     AND CMAKE_SIZEOF_VOID_P EQUAL 8
-  )
+  if(HPX_WITH_FETCH_HWLOC AND "${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
     add_custom_target(
       HwlocDLL ALL
       COMMAND ${CMAKE_COMMAND} -E make_directory
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2322,10 +2322,18 @@ endif()
 
 if(HPX_WITH_EXAMPLES)
   add_hpx_pseudo_target(examples)
-  if(HPX_WITH_FETCH_HWLOC AND "${CMAKE_SYSTEM_NAME}" STREQUAL "Windows" AND CMAKE_SIZEOF_VOID_P EQUAL 8)
-    add_custom_target(HwlocDLL ALL
-      COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_BINARY_DIR}/$<CONFIG>/bin/"
-      COMMAND ${CMAKE_COMMAND} -E copy_if_different "${HWLOC_ROOT}/bin/libhwloc-15.dll" "${CMAKE_BINARY_DIR}/$<CONFIG>/bin/"
+  if(HPX_WITH_FETCH_HWLOC
+     AND "${CMAKE_SYSTEM_NAME}" STREQUAL "Windows"
+     AND CMAKE_SIZEOF_VOID_P EQUAL 8
+  )
+    add_custom_target(
+      HwlocDLL ALL
+      COMMAND ${CMAKE_COMMAND} -E make_directory
+              "${CMAKE_BINARY_DIR}/$<CONFIG>/bin/"
+      COMMAND
+        ${CMAKE_COMMAND} -E copy_if_different
+        "${HWLOC_ROOT}/bin/libhwloc-15.dll"
+        "${CMAKE_BINARY_DIR}/$<CONFIG>/bin/"
     )
     add_hpx_pseudo_dependencies(examples HwlocDLL)
   endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -980,6 +980,16 @@ hpx_option(
   ADVANCED
 )
 
+# Option for automatically fetching Hwloc
+hpx_option(
+  HPX_WITH_FETCH_HWLOC
+  BOOL
+  "Use FetchContent to fetch Hwloc. By default an installed Hwloc will be used. (default: OFF)"
+  OFF
+  CATEGORY "Build Targets"
+  ADVANCED
+)
+
 # cmake-format: off
 # LibCDS option
 # NOTE: The libcds option is disabled for the 1.5.0 release as it is not ready
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2322,6 +2322,13 @@ endif()
 
 if(HPX_WITH_EXAMPLES)
   add_hpx_pseudo_target(examples)
+  if(HPX_WITH_FETCH_HWLOC AND "${CMAKE_SYSTEM_NAME}" STREQUAL "Windows" AND CMAKE_SIZEOF_VOID_P EQUAL 8)
+    add_custom_target(HwlocDLL ALL
+      COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_BINARY_DIR}/$<CONFIG>/bin/"
+      COMMAND ${CMAKE_COMMAND} -E copy_if_different "${HWLOC_ROOT}/bin/libhwloc-15.dll" "${CMAKE_BINARY_DIR}/$<CONFIG>/bin/"
+    )
+    add_hpx_pseudo_dependencies(examples HwlocDLL)
+  endif()
 endif()
 
 # ##############################################################################
--- a/cmake/HPX_SetupHwloc.cmake
+++ b/cmake/HPX_SetupHwloc.cmake
@@ -47,15 +47,11 @@ else()
     include_directories(${HWLOC_ROOT}/include)
     link_directories(${HWLOC_ROOT}/lib)
     set(Hwloc_INCLUDE_DIR ${HWLOC_ROOT}/include CACHE INTERNAL "")
-    set(Hwloc_LIBRARY ${HWLOC_ROOT}/lib CACHE INTERNAL "")
+    set(Hwloc_LIBRARY ${HWLOC_ROOT}/lib/libhwloc.dll.a CACHE INTERNAL "")
   else()
     FetchContent_Declare(HWLoc
       URL https://download.open-mpi.org/release/hwloc/v2.9/hwloc-win64-build-2.9.3.zip
-      TLS_VERIFY trueadd_library(Hwloc::hwloc INTERFACE IMPORTED)
-      target_include_directories(Hwloc::hwloc INTERFACE ${Hwloc_INCLUDE_DIR})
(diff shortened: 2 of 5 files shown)


hkaiser:
@vrnimje the `ci/circleci: install` step fails (https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/16294/workflows/33c0c1ec-9c58-4ec5-b900-4892d700503c/jobs/371220). Would you be able to have a look, please?

cmake/HPX_SetupHwloc.cmake
```diff
@@ -13,9 +13,93 @@
...
+  endif()
+else()
+  hpx_info(
+    "HPX_WITH_FETCH_HWLOC=${HPX_WITH_FETCH_HWLOC}, Hwloc will be fetched using CMake's FetchContent"
   )
+  if(UNIX)
+    include(FetchContent)
+    fetchcontent_declare(
+      HWLoc
+      URL https://download.open-mpi.org/release/hwloc/v2.9/hwloc-2.9.3.tar.gz
```
SAtacker, line 31: We may need to keep a separate variable for HWLOC version.

cmake/HPX_SetupHwloc.cmake
```diff
@@ -13,9 +13,93 @@
...
+        CACHE INTERNAL ""
+    )
+    set(Hwloc_LIBRARY
+        ${HWLOC_ROOT}/lib/libhwloc.so
+        CACHE INTERNAL ""
+    )
+  elseif("${CMAKE_GENERATOR_PLATFORM}" STREQUAL "Win64")
+    fetchcontent_declare(
+      HWLoc
+      URL https://download.open-mpi.org/release/hwloc/v2.9/hwloc-win64-build-2.9.3.zip
```
SAtacker, line 54: Same as above, we may have to keep a separate variable for HWLOC version.

cmake/HPX_SetupHwloc.cmake
```diff
@@ -13,9 +13,93 @@
...
+        CACHE INTERNAL ""
+    )
+    set(Hwloc_LIBRARY
+        ${HWLOC_ROOT}/lib/libhwloc.dll.a
+        CACHE INTERNAL ""
+    )
+  else()
+    fetchcontent_declare(
+      HWLoc
+      URL https://download.open-mpi.org/release/hwloc/v2.9/hwloc-win64-build-2.9.3.zip
```
SAtacker, line 78: 1. It should be win32.
2. Same as above, we may have to keep a separate variable for HWLOC version.

cmake/HPX_SetupHwloc.cmake
```diff
@@ -13,9 +13,93 @@
...
+        ${HWLOC_ROOT}/lib/libhwloc.dll.a
+        CACHE INTERNAL ""
+    )
+  endif() # End hwloc installation
+
+  add_library(Hwloc::hwloc INTERFACE IMPORTED)
+  target_include_directories(Hwloc::hwloc INTERFACE ${Hwloc_INCLUDE_DIR})
+  target_link_libraries(Hwloc::hwloc INTERFACE ${Hwloc_LIBRARY})
+  message(${Hwloc_INCLUDE_DIR})
+  message(${Hwloc_LIBRARY})
```
SAtacker, line 104: Please write a verbose message of INFO type.

SAtacker (changes requested):

cmake/HPX_SetupHwloc.cmake
SAtacker: Could you make this change and see if it doesn't fail? I am not sure how it's related to `free(): invalid pointer`, but it should be win32

cmake/HPX_SetupHwloc.cmake
hkaiser: You can use `hpx_info()` instead.

vrnimje: Actually I was using it to debug for a linking issue I was facing
Is it ok if I just remove it?

cmake/HPX_SetupHwloc.cmake
vrnimje: I have added them in the latest commit

SAtacker:
@hkaiser PTAL

vrnimje:
I have encountered an issue with the Windows build...
The Windows [workflow](https://github.com/STEllAR-GROUP/hpx/actions/runs/7236903558/job/20207865456?pr=6394), with the option enabled, is unable to pass _none_ of the tests, due to the following error:
```
Process 0 failed with an unexpected error code of 3221225781 (expected 0) <_io.TextIOWrapper name='<stderr>' mode='w' encoding='cp1252'>
```
So I tried recreating the build on my local system (using VS 2022), and I was able to reproduce the error with the tests
And after executing some examples in the repository, I encountered another error...

image

cmake/HPX_SetupHwloc.cmake
```diff
@@ -13,9 +13,102 @@
...
+        CACHE INTERNAL ""
+    )
+    set(Hwloc_LIBRARY
+        ${HWLOC_ROOT}/lib/libhwloc.dll.a
+        CACHE INTERNAL ""
+    )
+  else()
+    fetchcontent_declare(
+      HWLoc
+      URL https://download.open-mpi.org/release/hwloc/v${HPX_WITH_HWLOC_VERSION}/hwloc-win32-build-${HPX_WITH_HWLOC_RELEASE}.zip
```
SAtacker, line 89: PTAL https://github.com/SAtacker/hpx-template/blob/ba47fba9bda7833fdbd9698ee5863ddd959e6513/CMakeLists.txt#L108-L119

SAtacker: @vrnimje I am pretty sure you need 
```cmake
if(WIN32 OR "${CMAKE_GENERATOR_PLATFORM}" STREQUAL "Win64" AND NOT WITH_CUSTOM_HPX)
  add_custom_command(TARGET opt_bench POST_BUILD
    COMMAND ${CMAKE_COMMAND} -E copy_directory
        "${hpx_BINARY_DIR}/${CMAKE_BUILD_TYPE}/bin"
        $<TARGET_FILE_DIR:opt_bench>
  )
  add_custom_command(TARGET opt_bench POST_BUILD
    COMMAND ${CMAKE_COMMAND} -E copy_if_different
        "${HWLOC_ROOT}/bin/libhwloc-15.dll"
        $<TARGET_FILE_DIR:opt_bench>
  )
  ```

**Comment by SAtacker** [comment]
> I have encountered an issue with the Windows build... The Windows [workflow](https://github.com/STEllAR-GROUP/hpx/actions/runs/7236903558/job/20207865456?pr=6394), with the option enabled, is unable to pass _any_ of the tests, due to the following error:
> 
> ```
> Process 0 failed with an unexpected error code of 3221225781 (expected 0) <_io.TextIOWrapper name='<stderr>' mode='w' encoding='cp1252'>
> ```
> 
> So I tried recreating the build on my local system (using VS 2022), and I was able to reproduce the error with the tests And after executing some examples in the repository, I encountered another error...
> 
> image
> 
> I am currently trying to find a solution, but any pointers are welcome..

Please see https://github.com/SAtacker/hpx-template/blob/ba47fba9bda7833fdbd9698ee5863ddd959e6513/CMakeLists.txt#L108-L119

File: cmake/HPX_SetupHwloc.cmake (hunk 5 above)
Comment (vrnimje) [reply cmake/HPX_SetupHwloc.cmake:89]: Added this command in https://github.com/STEllAR-GROUP/hpx/pull/6394/commits/7ac5b3320726b04a109294abfadbb8a406347b60

Comment (hkaiser) [reply cmake/HPX_SetupHwloc.cmake:89]: What about 32Bit Windows? Do we support that here as well?

Comment (vrnimje) [reply cmake/HPX_SetupHwloc.cmake:89]: Yes, 32-bit Windows is also supported here: 
https://github.com/STEllAR-GROUP/hpx/blob/badead127a2a5721c64fe90a33c4f7b0cefa758c/cmake/HPX_SetupHwloc.cmake#L86-L108
I have also tested it locally using `cmake -A Win32`.

**Comment by SAtacker** [comment]
@hkaiser please approve the remaining workflows

**Comment by SAtacker** [comment]
@hkaiser PTAL

File: CMakeLists.txt [hunk 6]
```diff
@@ -2312,6 +2322,18 @@ endif()
...
 if(HPX_WITH_EXAMPLES)
   add_hpx_pseudo_target(examples)
+  if(HPX_WITH_FETCH_HWLOC AND "${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+    add_custom_target(
+      HwlocDLL ALL
+      COMMAND ${CMAKE_COMMAND} -E make_directory
+              "${CMAKE_BINARY_DIR}/$<CONFIG>/bin/"
+      COMMAND
+        ${CMAKE_COMMAND} -E copy_if_different
+        "${HWLOC_ROOT}/bin/libhwloc-15.dll"
```
hkaiser, line 2332: Shouldn't this depend on `HPX_WITH_HWLOC_VERSION`?

hkaiser (changes requested):

CMakeLists.txt
vrnimje: I looked into it, but was not able to find any relation between the release version of HWLoc and `libhwloc-15.dll`. As far as I could see [here](https://www-lb.open-mpi.org/software/hwloc/ancient/), it has been in use since HWLoc v2.0.1

hkaiser (approved):
LGTM, thanks!

SAtacker:
@hkaiser can we please merge this?

CMakeLists.txt
```diff
@@ -2312,6 +2322,18 @@ endif()
...
+        "${CMAKE_BINARY_DIR}/$<CONFIG>/bin/"
+    )
+    add_hpx_pseudo_dependencies(examples HwlocDLL)
+  endif()
 endif()
```
SAtacker, line 2337: @vrnimje what happens when `HPX_WITH_EXAMPLES` is not set? Please make a new PR if this is incorrect

vrnimje: I could only find `examples` as a psuedo-target, to which I added the dependency which I defined. I was not able to find any other target, which is built by default
Is there any other mechanism through which this can be achieved?

vrnimje: I was only able to find `examples` as a psuedo-target, to which I could add the dependency I have defined (`HwlocDLL`)
Since the libhwloc-15.dll needs to be copied exactly to `"${CMAKE_BINARY_DIR}/$<CONFIG>/bin/"`, and the Configuration type `$<CONFIG>` is defined during the build stage, I used the above method. As far as I understand, `$<CONFIG>` can only be obtained using `add_custom_command` or `add_custom_target`.

Is there any other mechanism to achieve this?

vrnimje: I now realise that I can directly add `HwlocDLL` as a pseudo target using `add_hpx_pseudo_target`

PR #6400, opened 2023-12-14
Fx the affinity setting bug when using LCI pp and multiple localities per node

### LINKED ISSUE #6398: HPX sets affinity wrong with multiple processes per node and LCI parcelport enabled
## Expected Behavior

For example, when there are 64 cores/node, 4 processes per node, and 16 threads per process, HPX should assign one core per thread, making all processes use different cores.

## Actual Behavior

When using the MPI parcelport, HPX works as expected. However, when using the LCI parcelport, all processes are mapped to the first 16 cores.

I don't think there are any codes in the LCI parcelport that can affect affinity, could you think of any possible reason for this? @hkaiser

JiakunYan:
Fixes #6398.

The HPX runtime relies on `used_cores_map_` to set the affinity correctly when running multiple localities per node (see `runtime_distributed::assign_cores` for details). `used_cores_map_` maps hostnames to "used cores" and assumes the hostname of all localities on the same node are the same. The hostname is generated using the `parcelport::get_locality_name` method (I would say the name of this method is misleading).

The previous LCI parcelport used "processor_name+rank" as a locality name, which breaks the assumption of `used_cores_map_`. This PR fixed this issue.

### DIFF
--- a/libs/full/parcelport_lci/src/parcelport_lci.cpp
+++ b/libs/full/parcelport_lci/src/parcelport_lci.cpp
@@ -118,9 +118,8 @@ namespace hpx::parcelset::policies::lci {
     /// Return the name of this locality
     std::string parcelport::get_locality_name() const
     {
-        // hostname-rank
-        return util::lci_environment::get_processor_name() + "-" +
-            std::to_string(util::lci_environment::rank());
+        // hostname
+        return util::lci_environment::get_processor_name();
     }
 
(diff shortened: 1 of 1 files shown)


hkaiser:
add to whitelist

hkaiser (approved):
LGTM, thanks!

hkaiser:
bors merge

PR #6401, opened 2023-12-18
fix(lci pp): fix the null_thread_id bug in the LCI parcelport

JiakunYan:
`send()` in the LCI parcelport can call HPX `yield()`. However, `send()` can be called outside HPX threads, leading to HPX complaining "null_thread_id".


JiakunYan:
Changes are overwritten by #6402

PR #6402, opened 2023-12-26
fix(lci pp): fix deadlocks with too many failed sends

JiakunYan:
In the previous LCI parcelport, too many failed sends will trigger `background_work` and `yield` as a way of back pressure propagation. These calls might cause the current thread to be de-scheduled. However, I found those rescheduled threads may never be scheduled again after yielding, leading to deadlocks. Not sure whether this is a bug of the scheduler or not.

This PR temporarily disabled the invocation of the `background_work` and `yield` functions and fixed the deadlock issue.

This PR overwrites the changes in #6401.


hkaiser (approved):
LGTM, thanks!

Let me know if you want for this to me merged.

JiakunYan:
@hkaiser Yes, please merge it if you think the failed tests are unrelated. Thanks!

Issue #6333, opened 2023-08-25
Current thread_id changing after suspending in direct execution


Pansysk75:
## Background

@hkaiser I came across this issue while investigating the current hangs in `cancelable_action` test.

In that test, the current thread_id is firstly obtained using `hpx::this_thread::get_id()` (through a `reset_id` helper object).
Then, we have a workloop that occasionally suspends, waiting to be interrupted (`hpx::thread::interrupt(id_)`).

https://github.com/STEllAR-GROUP/hpx/blob/bc993dd15b9252908e0b30b9ed5efd830cf99909/examples/cancelable_action/cancelable_action/server/cancelable_action.hpp#L81-L94

## Issue

I found that calling `hpx::this_thread::get_id()` before and after `hpx::this_thread::suspend()` would often return a different id.
I believe this is due to the way yielding is handled when utilizing direct execution:

https://github.com/STEllAR-GROUP/hpx/blob/bc993dd15b9252908e0b30b9ed5efd830cf99909/libs/core/coroutines/include/hpx/coroutines/detail/coroutine_stackful_self_direct.hpp#L34-L39

`coroutine_stackful_self_direct` falls back to `next_self_` (its calling context) for yielding, which will fail to correctly restore the state after yielding:

https://github.com/STEllAR-GROUP/hpx/blob/bc993dd15b9252908e0b30b9ed5efd830cf99909/libs/core/coroutines/include/hpx/coroutines/detail/coroutine_stackful_self.hpp#L32-L42

If I understand correctly, `reset_self_on_exit` will cause the [static variable `local_self_`](https://github.com/STEllAR-GROUP/hpx/blob/bc993dd15b9252908e0b30b9ed5efd830cf99909/libs/core/coroutines/src/detail/coroutine_self.cpp#L14) to refer to the "simple" (`coroutine_stackful_self`) coroutine after yielding (and not to the direct coroutine where it used to before the yield).

The solution is probably simple, but I just barely got to understanding how these coroutines work, so I just hope I got things right fttb.

hkaiser:
@Pansysk75 Yep, that's caused by the direct execution patch. cancelable_action should call `get_outer_id` instead. Excellent catch!

hkaiser:
> @Pansysk75 Yep, that's caused by the direct execution patch. cancelable_action should call `get_outer_id` instead (here:
> 
[...]

PR #6413, opened 2024-01-12
Always return outermost thread id

hkaiser:
-flyby: deprecate get_outer_self_id

fixing #6333


hkaiser:
retest lsu

hkaiser:
The freelist error has been fixed by #6458

PR #6424, opened 2024-01-23
Adding option for downloading Boost using CMake FetchContent

### LINKED ISSUE #6392: [Feature] Install dependencies using CMake

vrnimje:
Fixes #6392 

## Proposed Changes

  - Adding `HPX_WITH_FETCH_BOOST` bool option for fetching Boost v1.84.0 using CMake FetchContent
  - Adding CI tests for the above on macOS, Linux and Windows (VS2022, Debug)

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -961,7 +961,7 @@ endif()
 hpx_option(
   HPX_WITH_FETCH_BOOST
   BOOL
-  "Use FetchContent to fetch Asio. By default an installed Asio will be used. (default: OFF)"
+  "Use FetchContent to fetch Boost. By default an installed Boost will be used. (default: OFF)"
   OFF
   CATEGORY "Build Targets"
   ADVANCED
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -958,6 +958,15 @@ if(HPX_WITH_APEX)
   )
 endif()
 
+hpx_option(
+  HPX_WITH_FETCH_BOOST
+  BOOL
+  "Use FetchContent to fetch Asio. By default an installed Asio will be used. (default: OFF)"
+  OFF
+  CATEGORY "Build Targets"
+  ADVANCED
+)
+
 # Options for automatically fetching Asio
 hpx_option(
   HPX_WITH_FETCH_ASIO
--- a/cmake/HPX_SetupBoost.cmake
+++ b/cmake/HPX_SetupBoost.cmake
@@ -15,6 +15,7 @@ if(HPX_WITH_FETCH_BOOST)
     Boost
     URL https://github.com/boostorg/boost/releases/download/boost-${HPX_WITH_BOOST_VERSION}/boost-${HPX_WITH_BOOST_VERSION}.tar.gz
     TLS_VERIFY true
+    DOWNLOAD_EXTRACT_TIMESTAMP true
   )
   fetchcontent_populate(Boost)
 
@@ -53,8 +54,14 @@ if(HPX_WITH_FETCH_BOOST)
     )
   endif()
 
-  set(Boost_DIR "${CMAKE_BINARY_DIR}/_deps/boost-src")
-  set(Boost_INCLUDE_DIR "${CMAKE_BINARY_DIR}/_deps/boost-src")
+  set(Boost_DIR
+      "${CMAKE_BINARY_DIR}/_deps/boost-src"
+      CACHE INTERNAL ""
+  )
+  set(Boost_INCLUDE_DIR
+      "${CMAKE_BINARY_DIR}/_deps/boost-src"
+      CACHE INTERNAL ""
+  )
 endif()
 
 # In case find_package(HPX) is called multiple times
--- a/cmake/HPX_SetupBoost.cmake
+++ b/cmake/HPX_SetupBoost.cmake
@@ -35,7 +35,6 @@ if(HPX_WITH_FETCH_BOOST)
           "cd ${CMAKE_BINARY_DIR}\\_deps\\boost-src && .\\bootstrap.bat && .\\b2 headers cxxflags=/std:c++20"
       )
     elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
-      hpx_info("Building fetched boost")
       execute_process(
         COMMAND
           sh -c
@@ -56,7 +55,6 @@ if(HPX_WITH_FETCH_BOOST)
 
   set(Boost_DIR "${CMAKE_BINARY_DIR}/_deps/boost-src")
   set(Boost_INCLUDE_DIR "${CMAKE_BINARY_DIR}/_deps/boost-src")
-endif()
 endif()
 
 # In case find_package(HPX) is called multiple times
--- a/cmake/HPX_SetupBoost.cmake
+++ b/cmake/HPX_SetupBoost.cmake
@@ -5,6 +5,39 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
+if (HPX_WITH_FETCH_BOOST)
+  include(FetchContent)
+  FetchContent_Declare(Boost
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -958,6 +958,15 @@ if(HPX_WITH_APEX)
   )
 endif()
 
+hpx_option(
+  HPX_WITH_FETCH_BOOST
+  BOOL
+  "Use FetchContent to fetch Asio. By default an installed Asio will be used. (default: OFF)"
```
SAtacker, line 964: Please change description

cmake/HPX_SetupBoost.cmake
```diff
@@ -5,6 +5,60 @@
...
+      )
+    endif()
+    set(EX_PRC_INTERNAL
+        ${EX_PRC}
+        CACHE INTERNAL "for internal use only; do not modify"
+    )
+  endif()
+
+  set(Boost_DIR "${CMAKE_BINARY_DIR}/_deps/boost-src")
+  set(Boost_INCLUDE_DIR "${CMAKE_BINARY_DIR}/_deps/boost-src")
```
SAtacker, line 57: Are we sure this should not be internal?

cmake/HPX_SetupBoost.cmake
```diff
@@ -5,6 +5,60 @@
...
+endif()
+endif()
```
SAtacker: There are two endifs, does it not cause errors?

SAtacker (changes requested):

cmake/HPX_SetupBoost.cmake
SAtacker: Nevermind, seems like you have fixed it

cmake/HPX_SetupBoost.cmake
vrnimje: I guess it makes sense to keep it Internal, as it is set when the option is used, so should not be visible

SAtacker:
@hkaiser PTAL, thanks!

cmake/HPX_SetupBoost.cmake
```diff
@@ -5,6 +5,65 @@
...
+      ""
+      CACHE INTERNAL "for internal use only; do not modify"
+  )
+
+  if(NOT EX_PRC STREQUAL EX_PRC_INTERNAL)
+    if(WIN32)
+      execute_process(
+        COMMAND
+          cmd /C
+          "cd ${CMAKE_BINARY_DIR}\\_deps\\boost-src && .\\bootstrap.bat && .\\b2 headers cxxflags=/std:c++20"
```
hkaiser, line 36: Should we compile with the C++ version that corresponds to what the user requested for HPX (`HPX_CXX_STANDARD`)?

cmake/HPX_SetupBoost.cmake
```diff
@@ -5,6 +5,65 @@
...
+      )
+    elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+      execute_process(
+        COMMAND
+          sh -c
+          "cd ${CMAKE_BINARY_DIR}/_deps/boost-src && ./bootstrap.sh --prefix=${CMAKE_BINARY_DIR}/_deps/boost-installed && ./b2 && ./b2 install --prefix=${CMAKE_BINARY_DIR}/_deps/boost-installed cxxflags=--std=c++20"
```
hkaiser, line 42: Same here (and in more places below). Should we use `HPX_CXX_STANDARD`?

cmake/HPX_SetupBoost.cmake
```diff
@@ -5,6 +5,65 @@
...
+      )
+    else()
+      execute_process(
+        COMMAND
+          sh -c
+          "cd ${CMAKE_BINARY_DIR}/_deps/boost-src && ./bootstrap.sh && ./b2 headers cxxflags=--std=c++20"
+      )
+    endif()
+    set(EX_PRC_INTERNAL
```
hkaiser, line 51: What is this used for?

cmake/HPX_SetupBoost.cmake
```diff
@@ -5,6 +5,65 @@
...
+
+  set(Boost_DIR
+      "${CMAKE_BINARY_DIR}/_deps/boost-src"
+      CACHE INTERNAL ""
+  )
+  set(Boost_INCLUDE_DIR
+      "${CMAKE_BINARY_DIR}/_deps/boost-src"
+      CACHE INTERNAL ""
+  )
+endif()
```
hkaiser, line 65: Should all of the above go into a separate file `FindBoost.cmake` or similar (just for consistency reasons)? OTOH, this may conflict with the file with the same name in the CMake distribution...

hkaiser (changes requested):

cmake/HPX_SetupBoost.cmake
vrnimje: Yes, we could do that. Will make the changes

cmake/HPX_SetupBoost.cmake
vrnimje: It makes sure that the commands defined to build Boost are not executed again, if the project is reconfigured
But I noticed that I have mistakenly defined it as `INTERNAL` instead of `STRING`. Will make the changes

hkaiser: If this is something that is used outside of a CMake configuration step (e.g. needed for configuring dependent applications), please prefix the variable with `HPX_WITH_` for consistency and to avoid name clashes.

vrnimje: Yes, it is used for building the fetched Boost, so will update with the variable naming you have mentioned.

cmake/HPX_SetupBoost.cmake
```diff
@@ -5,6 +5,64 @@
...
+      execute_process(
+        COMMAND
+          sh -c
+          "cd ${CMAKE_BINARY_DIR}/_deps/boost-src && ./bootstrap.sh --prefix=${CMAKE_BINARY_DIR}/_deps/boost-installed && ./b2 && ./b2 install --prefix=${CMAKE_BINARY_DIR}/_deps/boost-installed cxxflags=--std=c++${HPX_CXX_STANDARD}"
+      )
+    else()
+      execute_process(
+        COMMAND
+          sh -c
+          "cd ${CMAKE_BINARY_DIR}/_deps/boost-src && ./bootstrap.sh && ./b2 headers cxxflags=--std=c++${HPX_WITH_CXX_STANDARD}"
```
hkaiser, line 47:
```suggestion
          "cd ${CMAKE_BINARY_DIR}/_deps/boost-src && ./bootstrap.sh && ./b2 headers cxxflags=--std=c++${HPX_CXX_STANDARD}"
```

cmake/HPX_SetupBoost.cmake
```diff
@@ -5,6 +5,64 @@
...
+  )
+  include(FetchContent)
+  fetchcontent_declare(
+    Boost
+    URL https://github.com/boostorg/boost/releases/download/boost-${HPX_WITH_BOOST_VERSION}/boost-${HPX_WITH_BOOST_VERSION}.tar.gz
+    TLS_VERIFY true
+    DOWNLOAD_EXTRACT_TIMESTAMP true
+  )
+  fetchcontent_populate(Boost)
+  set(HPX_WITH_EX_PRC
```
hkaiser, line 21: Would you mind making this variable a bit more self-explanatory?

SAtacker:
What is stopping us here @vrnimje @hkaiser ?

vrnimje:
> What is stopping us here @vrnimje @hkaiser ?

I have made all the requested changes from my side, as far as I know.
Just pending more reviews / confirmation

hkaiser (approved):

PR #6425, opened 2024-01-23
Fix macOS failing test

SAtacker:
- Attempt to fix macOS test failures

### DIFF
--- a/libs/core/debugging/src/print.cpp
+++ b/libs/core/debugging/src/print.cpp
@@ -57,6 +57,10 @@ namespace hpx::debug {
             std::ostream&, std::int32_t const&, int);
         template HPX_CORE_EXPORT void print_dec(
             std::ostream&, std::int64_t const&, int);
+#ifdef __APPLE__
+        template HPX_CORE_EXPORT void print_dec(
+            std::ostream&, unsigned long const&, int);
+#endif
         template HPX_CORE_EXPORT void print_dec(
             std::ostream&, std::uint64_t const&, int);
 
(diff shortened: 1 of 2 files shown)


SAtacker:
Note - MacOS does not support CPU binding

libs/core/debugging/src/print.cpp
```diff
@@ -57,6 +57,10 @@ namespace hpx::debug {
             std::ostream&, std::int32_t const&, int);
         template HPX_CORE_EXPORT void print_dec(
             std::ostream&, std::int64_t const&, int);
+#ifdef __APPLE__
+        template HPX_CORE_EXPORT void print_dec(
+            std::ostream&, unsigned long const&, int);
```
hkaiser, line 62: Would it be better to unconditionally add a specialization for `std::uint32_t`?

SAtacker: I need more clarity on this because for x86 macOs this could be different and not sure why this isn't already handled by macOS's xcode standard SDK.
1. We wait for them to fix this
2. We keep this and it does not interfere with any of the std::uint32_t and builds
3. We do an unconditional specialization for std::uint32_t and it may override or fail to build in the future.

hkaiser: Adding the specialization for `std::uint32_t` will not do any harm and should be equivalent to the one you added.

hkaiser:
We can close this PR as the issue was fixed here: #6475

PR #6428, opened 2024-01-31
Enable testing of new scheduler in CI

hkaiser:


hkaiser:
@Pansysk75 it looks like that the new scheduler doesn't properly work yet (as you suggested). the clang-12 and gcc-10 errors are most likely related.

Pansysk75:
retest

Pansysk75:
@hkaiser FYI I'm working on this one, I've reproduced some failing tests (`stop_token_cb1` being one of them). So far, I see that some pending hpx thread is never getting scheduled (either starvation or some bug). I'll let you know of what I find out.

Pansysk75:
@hkaiser It seems like the moodycamel queue provides FIFO guarantees only among items placed by the same thread, however there is no guarantee that items queued by different threads will be dequeued in a FIFO order.

From the [moodycamel blog](https://moodycamel.com/blog/2014/a-fast-general-purpose-lock-free-queue-for-c++):

> All elements from a given producer thread will necessarily still be seen in that same order relative to each other when dequeued (since the sub-queue preserves that order), albeit with elements from other sub-queues possibly interleaved.

And from the discussion in the comments:

>  **gasche**: ... If thread A enqueues 1, then does a write observed by thread B (synchronized through the appropriate memory fence), then B enqueues 2, then I would expect 1 to come before 2 -- and this is not racy. It seems to me that your data-structure is thus providing a weaker specification than linearizable queues. ...

> **Cameron**: Yes, you're right. If there's external synchronization, then the final order is not related to the total order formed by the happens-before relationships of the enqueue operations themselves. ...

This currently causes trouble (deadlocks), as we rely on the FIFO ordering to avoid starvation:
https://github.com/STEllAR-GROUP/hpx/blob/bd9e3cbabe3d8402263e9bafec6c8f6d5de6048a/libs/core/thread_pools/include/hpx/thread_pools/scheduling_loop.hpp#L325-L332

In the fail-case that I managed to reproduce, a perpetually yielding thread starves out a shared-resource-holding thread when these two happen to be in the same moodycamel thread-queue. We cannot reliably insert the yielding thread in the"back" of the queue (no FIFO guarantee), and the queue under certain circumstances acts similar to a LIFO, thus never allowing the shared-resource-holding thread to be picked up by the scheduler and release the held resource.

hkaiser:
Excellent detective work @Pansysk75! Is that true in case if we used the consumer/producer token API the queue provides as well?

Pansysk75:
> Excellent detective work @Pansysk75! Is that true in case if we used the consumer/producer token API the queue provides as well?

Thanks! Here's what the implementor has to say about this (TLDR it's not possible):   https://github.com/cameron314/concurrentqueue/issues/6

hkaiser:
> > Excellent detective work @Pansysk75! Is that true in case if we used the consumer/producer token API the queue provides as well?
> 
> Thanks! Here's what the implementer has to say about this (TLDR it's not possible): https://www.github.com/cameron314/concurrentqueue/issues/6

Would this patch mitigate the issue?
```diff
diff --git a/libs/core/thread_pools/include/hpx/thread_pools/scheduling_loop.hpp b/libs/core/thread_pools/include/hpx/thread_pools/scheduling_loop.hpp
index 1ca53941ee4..d2cc74fae4e 100644
--- a/libs/core/thread_pools/include/hpx/thread_pools/scheduling_loop.hpp
+++ b/libs/core/thread_pools/include/hpx/thread_pools/scheduling_loop.hpp
@@ -315,13 +315,6 @@ namespace hpx::threads::detail {
                     if (HPX_UNLIKELY(
                             state_val == thread_schedule_state::pending))
                     {
-                        if (HPX_LIKELY(next_thrd == nullptr))
-                        {
-                            // schedule other work
-                            scheduler.wait_or_add_new(num_thread, running,
-                                idle_loop_count, enable_stealing_staged, added);
-                        }
-
                         // schedule this thread again, make sure it ends up at
                         // the end of the queue
                         scheduler.SchedulingPolicy::schedule_thread_last(
@@ -329,6 +322,14 @@ namespace hpx::threads::detail {
                             threads::thread_schedule_hint(
                                 static_cast<std::int16_t>(num_thread)),
                             true);
+
+                        if (HPX_LIKELY(next_thrd == nullptr))
+                        {
+                            // schedule other work
+                            scheduler.wait_or_add_new(num_thread, running,
+                                idle_loop_count, enable_stealing_staged, added);
+                        }
+
                         scheduler.SchedulingPolicy::do_some_work(num_thread);
                     }
                     else if (HPX_UNLIKELY(state_val ==
```
I think the problem is that currently  after the pending thread has been put back onto the (top of) queue the scheduler immediately pulls it back and retries without potentially waiting threads having a change of being work-requested. For other schedulers those will get stolen eventually, thus avoiding the live-lock.

Pansysk75:
@hkaiser Doesn't seem to help much... keep in mind that this issue doesn't show up on the workrequesting scheduler when using the default queue, so it's either the mc queue or the stealhalf functionality that is causing the issue (I also tried setting this to false https://github.com/STEllAR-GROUP/hpx/blob/bd9e3cbabe3d8402263e9bafec6c8f6d5de6048a/libs/core/schedulers/include/hpx/schedulers/lockfree_queue_backends.hpp#L119 to see bulk dequeueing was the culprit, but it didn't fix things. 

Also, I think this fix is needed (or sth equivalent):

```diff
                task_data thrds(d.num_thread_);
-                thrds.tasks_.reserve(max_num_to_steal);


#ifdef HPX_HAVE_THREAD_STEALING_COUNTS
+              thrds.tasks_.reserve(max_num_to_steal);
                thread_id_ref_type thrd;
                while (max_num_to_steal-- != 0 &&
                    d.queue_->get_next_thread(thrd, false, true))
                {
                    d.queue_->increment_num_stolen_from_pending();
                    thrds.tasks_.push_back(HPX_MOVE(thrd));
                    thrd = thread_id_ref_type{};
                }
#else
+              thrds.tasks_.resize(max_num_to_steal);
                d.queue_->get_next_threads(
                    thrds.tasks_.begin(), thrds.tasks_.size(), false, true);
```

hkaiser:
> Also, I think this fix is needed (or sth equivalent):
> 
> ```diff
>                 task_data thrds(d.num_thread_);
> -                thrds.tasks_.reserve(max_num_to_steal);
> 
> 
> #ifdef HPX_HAVE_THREAD_STEALING_COUNTS
> +              thrds.tasks_.reserve(max_num_to_steal);
>                 thread_id_ref_type thrd;
>                 while (max_num_to_steal-- != 0 &&
>                     d.queue_->get_next_thread(thrd, false, true))
>                 {
>                     d.queue_->increment_num_stolen_from_pending();
>                     thrds.tasks_.push_back(HPX_MOVE(thrd));
>                     thrd = thread_id_ref_type{};
>                 }
> #else
> +              thrds.tasks_.resize(max_num_to_steal);
>                 d.queue_->get_next_threads(
>                     thrds.tasks_.begin(), thrds.tasks_.size(), false, true);
> ```

Uhh, how did it ever work? ;-)

hkaiser:
> > Also, I think this fix is needed (or sth equivalent):
> > ```diff
> >                 task_data thrds(d.num_thread_);
> > -                thrds.tasks_.reserve(max_num_to_steal);
> > 
> > 
> > #ifdef HPX_HAVE_THREAD_STEALING_COUNTS
> > +              thrds.tasks_.reserve(max_num_to_steal);
> >                 thread_id_ref_type thrd;
> >                 while (max_num_to_steal-- != 0 &&
> >                     d.queue_->get_next_thread(thrd, false, true))
> >                 {
> >                     d.queue_->increment_num_stolen_from_pending();
> >                     thrds.tasks_.push_back(HPX_MOVE(thrd));
> >                     thrd = thread_id_ref_type{};
> >                 }
> > #else
> > +              thrds.tasks_.resize(max_num_to_steal);
> >                 d.queue_->get_next_threads(
> >                     thrds.tasks_.begin(), thrds.tasks_.size(), false, true);
> > ```
> 
> Uhh, how did it ever work? ;-)

That's fixed now. Thanks again!

PR #6431, opened 2024-02-03
Integrate NVIDIA's S/R implementation into HPX 

isidorostsa:
This PR serves to replace the existing senders/receivers implementation in hpx with the [stdexec implementation](https://github.com/NVIDIA/stdexec) maintained by NVIDIA.

Big credits to the @pika-org team for their work when solving almost the same problem. This PR is built on top of much of that work.


CMakeLists.txt
```diff
@@ -2196,6 +2196,9 @@ include(HPX_SetupAsio)
 include(HPX_SetupAllocator)
 include(HPX_SetupHwloc)
 
+# Setup NVIDIA's stdexec
+include(HPX_SetupStdexec)
```
hkaiser, line 2200: This probably should be included only if compiling with C++20 or higher. We should also introduce an `HPX_WITH_...` option to explicitly enable pulling it in.

isidorostsa: Yess totally, and more nuances are to come 
(this is a bare minimum pr for now)

CMakeLists.txt
```diff
@@ -567,6 +567,49 @@ if(HPX_WITH_CUDA AND HPX_WITH_HIP)
...
+  HPX_WITH_FETCH_STDEXEC
+  BOOL
+  "Use FetchContent to fetch STDEXEC.(default: OFF)"
+  OFF
+  CATEGORY "Executor"
+  ADVANCED
+)
+
+# if STDEXEC_ROOT was provided, HPX_WITH_STDEXEC is set to ON
+if(STDEXEC_ROOT OR HPX_WITH_FETCH_STDEXEC)
```
hkaiser, line 593: This might have to be `Stdexec_ROOT` to avoid warnings with recent CMake versions

cmake/FindStdexec.cmake
```diff
@@ -0,0 +1,33 @@
+
```
hkaiser, line 1: Please add Copyright, License and the SPDX identifier to newly created files.

cmake/HPX_SetupStdexec.cmake
```diff
@@ -0,0 +1,34 @@
+if(STDEXEC_ROOT AND NOT Stdexec_ROOT)
+    set(Stdexec_ROOT
+        ${STDEXEC_ROOT}
+        CACHE PATH "STDEXEC base directory"
+    )
+    unset(STDEXEC_ROOT CACHE)
+endif()
```
hkaiser, line 7: We have this kind of code in our CMake build system only to be backwards compatible, not sure if we need this at all.

cmake/HPX_SetupStdexec.cmake
```diff
@@ -0,0 +1,34 @@
...
+if(HPX_WITH_FETCH_STDEXEC AND NOT Stdexec_ROOT)
+    hpx_info(
+        "HPX_WITH_FETCH_STDEXEC=${HPX_WITH_FETCH_STDEXEC}, Stdexec will be fetched using CMake's FetchContent."
+    )
+    if(UNIX)
+        include(FetchContent)
+        FetchContent_Declare(
+            Stdexec
+            GIT_REPOSITORY https://github.com/NVIDIA/stdexec.git
+            GIT_TAG        main
```
hkaiser, line 18: Can we make the fetched tag configurable, please?

libs/CMakeLists.txt
```diff
@@ -113,6 +113,11 @@ if(TARGET APEX::apex)
   target_link_libraries(hpx_full INTERFACE APEX::apex)
 endif()
 
+# Do I need this here?
```
hkaiser, line 116: You most likely will not need this as it will be propagated from `hpx_core`.

libs/core/execution_base/CMakeLists.txt
```diff
@@ -52,6 +55,8 @@ add_hpx_module(
   SOURCES ${execution_base_sources}
   HEADERS ${execution_base_headers}
   COMPAT_HEADERS ${execution_base_compat_headers}
+  DEPENDENCIES
+    STDEXEC::stdexec
```
hkaiser, line 59: This is needed only if enabled.

cmake/HPX_SetupStdexec.cmake
isidorostsa: Of course!

cmake/HPX_SetupStdexec.cmake
isidorostsa: Makes sense, removing

CMakeLists.txt
isidorostsa: Do you think it would make sense to perform this check when parsing the arguments in the [main CMakeLists.txt](https://github.com/STEllAR-GROUP/hpx/pull/6431/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR611)?

I did not spot any convention for that sort of thing in the CMake files.

hkaiser: > Do you think it would make sense to perform this check when parsing the arguments in the [main CMakeLists.txt](https://github.com/STEllAR-GROUP/hpx/pull/6431/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR611)?
> 
> I did not spot any convention for that sort of thing in the CMake files.

Yah, we don't have conventions. Let's move as much as possible out of the main CMakeLists.txt, it's overcrowded already.

libs/CMakeLists.txt
isidorostsa: Removed

CMakeLists.txt
isidorostsa: Changed it, thank you

libs/core/execution/include/hpx/execution/algorithms/as_sender.hpp
```diff
@@ -136,6 +136,27 @@ namespace hpx::execution::experimental {
 
             std::decay_t<Future> future_;
 
+#ifdef HPX_HAVE_STDEXEC
+            template <bool IsVoid, typename _result_type>
+            struct _set_value
```
hkaiser, line 141: Let's not start with the leading underscore madness...

libs/core/execution/include/hpx/execution/queries/read.hpp
```diff
@@ -87,7 +92,7 @@ namespace hpx::execution::experimental {
             template <typename Tag1>
             friend auto tag_invoke(
                 get_completion_signatures_t, read_sender<Tag1>, no_env)
-                -> dependent_completion_signatures<no_env>;
+                -> std::execution::dependent_completion_signatures<no_env>;
```
hkaiser, line 95: this doesn't look right, is std::execution always available?

libs/core/execution/tests/unit/algorithm_as_sender.cpp
```diff
@@ -30,7 +30,11 @@ int hpx_main()
         auto s = ex::as_sender(std::move(fut));
 
         static_assert(ex::is_sender_v<decltype(s)>);
+#ifdef HPX_HAVE_STDEXEC
+        static_assert(ex::is_sender_in_v<decltype(s), ex::empty_env>);
+#else
         static_assert(ex::is_sender_v<decltype(s), ex::empty_env>);
+#endif
```
hkaiser, line 37: Shouldn't we rather implement `is_sender_in_v` for our code base?

hkaiser (commented):
First batch of comments... more to come.

libs/core/execution/include/hpx/execution/algorithms/as_sender.hpp
isidorostsa: right, will rename to `set_value_void_checked` or something similar. I've implemented something this in other parts of the code so I will make sure the names match. I do not think it is worth it to factor this out since it's only 2 or 3 places.

libs/core/execution/include/hpx/execution/algorithms/detail/partial_algorithm.hpp
```diff
@@ -62,9 +62,16 @@ namespace hpx::execution::experimental::detail {
...
+#ifdef HPX_HAVE_STDEXEC
+            auto scheduler =
+                hpx::execution::experimental::get_completion_scheduler<
+                    hpx::execution::experimental::set_value_t>(
+                    hpx::execution::experimental::get_env(u));
+#else
             auto scheduler =
                 hpx::execution::experimental::get_completion_scheduler<
                     hpx::execution::experimental::set_value_t>(u);
+#endif
```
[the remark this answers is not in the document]
isidorostsa: We would have to alter our existing senders to provide environments, currently most of the times they do not.

libs/core/algorithms/tests/unit/algorithms/adjacentdifference_tests.hpp
```diff
@@ -83,8 +84,14 @@ void test_adjacent_difference_sender(Policy l, ExPolicy&& policy)
...
 
     auto exec = ex::explicit_scheduler_executor(scheduler_t(l));
+#ifdef HPX_HAVE_STDEXEC
+    auto result =
+        tt::sync_wait(ex::just(std::begin(c), std::end(c), std::begin(d)) |
+            hpx::adjacent_difference(policy.on(exec)));
+#else
     auto result = ex::just(std::begin(c), std::end(c), std::begin(d)) |
         hpx::adjacent_difference(policy.on(exec)) | tt::sync_wait();
+#endif
```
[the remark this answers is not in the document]
isidorostsa: that duplication is present in many of the tests i have changed so it might make sense to do that there too?
e.g. https://github.com/STEllAR-GROUP/hpx/pull/6431/files#diff-b11a40f80e11363af9d5469224fdb4e94487ee5ff99c30c7d4219a5ab40c85d8

libs/core/execution/tests/unit/algorithm_as_sender.cpp
isidorostsa: right, leaving that as a TODO to remove much of the duplication from the tests.

libs/core/execution/include/hpx/execution/algorithms/let_stopped.hpp
```diff
@@ -8,6 +8,9 @@
 #pragma once
 
 #include <hpx/config.hpp>
+#ifdef HPX_HAVE_STDEXEC
+#else
```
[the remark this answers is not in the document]
isidorostsa: ah yes, thanks!

libs/core/execution/include/hpx/execution/queries/read.hpp
isidorostsa: You are right, it is not and this is wrong. Also, this part of the code is ifdef-ed away so I am reverting it to the pre-stdexec code.

libs/core/execution/include/hpx/execution/algorithms/detail/partial_algorithm.hpp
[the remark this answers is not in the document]
hkaiser: Really? What about:
```
inline constexpr struct get_completion_scheduler_ex_t
{
    template <typename U>
    decltype(auto) operator()(U&& u) const
    {
#ifdef HPX_HAVE_STDEXEC
        return hpx::execution::experimental::get_completion_scheduler<
                    hpx::execution::experimental::set_value_t>(
                    hpx::execution::experimental::get_env(HPX_FORWARD(U, u)));
#else
        return hpx::execution::experimental::get_completion_scheduler<
                    hpx::execution::experimental::set_value_t>(HPX_FORWARD(U, u));
#endif
    }
} get_completion_scheduler_ex{};
```

isidorostsa:
Fetching STDEXEC by default on C++20, so the CI is testing both with and without it.

libs/core/execution/tests/unit/algorithm_run_loop.cpp
```diff
@@ -761,23 +760,6 @@ void test_future_sender()
...
-        HPX_TEST_EQ(f.get(), 3);
-    }
-    // Make a future from a scheduler
-    {
-        ex::run_loop loop;
-        auto sched = loop.get_scheduler();
-
-        auto f = ex::make_future(sched);
-    }
-
```
hkaiser: Why did you remove these tests?

isidorostsa: I added those tests at some point but I think because I was trying to understand how `make_future` works.
I removed them because the first one is pretty much a duplicate of  https://github.com/STEllAR-GROUP/hpx/pull/6431/commits/ef31254b19a6b730ca8863df58b9b95c3e79e1ec#diff-b11a40f80e11363af9d5469224fdb4e94487ee5ff99c30c7d4219a5ab40c85d8L799 and the second one is using `make_future` on a scheduler instead of a sender which is not correct in the first place.

hkaiser:
@isidorostsa just in case you were wondering... the Windows tests all fail with a segmentation fault (error code `0xc0000005`)

isidorostsa:
> @isidorostsa just in case you were wondering... the Windows tests all fail with a segmentation fault (error code `0xc0000005`)

Interestingly it's using c++17 too

libs/core/async_mpi/tests/unit/algorithm_transform_mpi.cpp
```diff
@@ -212,8 +212,11 @@ int hpx_main()
                 }
                 catch (std::runtime_error const& e)
                 {
-                    HPX_TEST(std::string(e.what()).find(std::string(
-                                 "invalid root")) != std::string::npos);
+                    auto error_msg = std::string(e.what());
+                    for (auto& c : error_msg)
```
hkaiser, line 216: We do have this: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/string_util/include/hpx/string_util/case_conv.hpp.

cmake/HPX_PerformCxxFeatureTests.cmake
```diff
@@ -152,6 +152,8 @@ function(hpx_perform_cxx_feature_tests)
 
     hpx_check_for_cxx20_std_bit_cast(DEFINITIONS HPX_HAVE_CXX20_STD_BIT_CAST)
 
+    hpx_check_for_cxx20_std_identity(DEFINITIONS HPX_HAVE_CXX20_STD_BIT_CAST)
```
hkaiser, line 155:
```suggestion
    hpx_check_for_cxx20_std_identity(DEFINITIONS 
HPX_HAVE_CXX20_STD_IDENTITY)
```

isidorostsa:
@hkaiser Should I rebase on top of https://github.com/STEllAR-GROUP/hpx/pull/6507?

isidorostsa:
Just rebased on master

hkaiser:
Thanks so much!

Pansysk75:
👏 👏

PR #6433, opened 2024-02-05
Use shared mutex for resolve_locality procedure.

JiakunYan:
`resolve_locality` is on the critical path of parcel sending. Originally, it used `hpx:spinlock` to protect its `unordered_map::find`. This turns out to be too expensive. This PR replaces the `spinlock` with `hpx::shared_mutex` and greatly improves the maximum parcel sending throughput.


hkaiser (approved):
LGTM, thanks!

PR #6444, opened 2024-02-22
Update CTestConfig.cmake

akheir:
Change the CDash URL


hkaiser:
This is functional now. Thanks again @akheir! Merging...

PR #6447, opened 2024-02-26
Use compressed ptr in schedulers when 128 atomics are not lockfree

srinivasyadav18:
Fixes atomics on some of the ARM machines (A64FX).

## Proposed Changes

  - Enable **cxx std atomic 128** bits (**HPX_HAVE_CXX11_STD_ATOMIC_128BIT**) only when **lockfree** in cmake cxx feature tests.
  - Enable **work_requesting_lifo** scheduler only when 128 bit atomics are available. 
  - Fallback to **HPX_LOCKFREE_PTR_COMPRESSION** when **HPX_HAVE_CXX11_STD_ATOMIC_128BIT** is not available. 
  - Use cache alignment only on **std::atomic** of the tagged_index in freelist_stack.


hkaiser (approved):
LGTM, thanks!

PR #6452, opened 2024-03-13
Fix nvcc crashes in transform_stream.cu and synchronize.cu

### LINKED ISSUE #5799: Investigate CUDA compilation problems

G-071:
With CUDA 12.3 nvcc does not outright crash anymore when compiling sender-receiver code. 
Though, to make them compile, we still need to conditionally remove the noexcept with an IFDEF, which this PR adds.

Hence, we can activate the synchronize and the transform_stream tests when the CUDA version ist 12.3 and newer.

Fixes #5799

### DIFF
--- a/libs/core/async_cuda/tests/performance/synchronize.cu
+++ b/libs/core/async_cuda/tests/performance/synchronize.cu
@@ -6,8 +6,9 @@
 
 #include <hpx/config.hpp>
 
-// NVCC fails unceremoniously with this test at least until V12.1
-#if !defined(HPX_CUDA_VERSION) || (HPX_CUDA_VERSION > 1201)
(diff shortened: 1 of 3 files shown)


hkaiser (approved):
LGTM, thanks!

hkaiser:
@G-071 could you please fix the inspect and clang-format issues reported?

hkaiser:
@G-071 ping?

G-071:
Thanks for the ping, I forgot about this PR. It should be fixed now though!

PR #6457, opened 2024-03-19
replace boost::array with std::array and update file name

hadeer-r:
## Changes

  - remove "boost/array.hpp" and replace boost::array with std::array
  - rename file from 'boost_array.hpp' to 'std_array.hpp'
  - relating to issue (#3440)

### DIFF
--- a/libs/core/serialization/CMakeLists.txt
+++ b/libs/core/serialization/CMakeLists.txt
@@ -202,7 +202,7 @@ set(serialization_compat_headers
 
 if(HPX_SERIALIZATION_WITH_BOOST_TYPES)
   set(boost_serialization_headers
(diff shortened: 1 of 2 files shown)


hkaiser:
@hadeer-r thank you for working on this! However, this file is there for explicit serialization support for `boost::array`, so changing it is not desired. We do already have support for serializing `std::array` (see: https://github.com/STEllAR-GROUP/hpx/blob/69cf09f072a8a88a917973b2006cb400e184ef8b/libs/core/serialization/include/hpx/serialization/array.hpp#L118-L125). The files you changed will be used only if usersask for explicit serialization support for Boost types (which we would like to keep in place for backwards compatibility reasons).

PR #6458, opened 2024-03-19
Reduce test workload to fix CI/CD time-out

Pansysk75:
Reduce workload on `freelist` test, to prevent CircleCI test from timing out.


hkaiser (approved):
LGTM, thanks!

PR #6463, opened 2024-03-26
Revert "Always return outermost thread id"

hkaiser:
Reverts STEllAR-GROUP/hpx#6413


hkaiser:
This reverts master to a non-broken state.

PR #6464, opened 2024-03-28
Update docs of "Writing distributed applications" page

dimitraka:
The section "Writing components > Using component instances" was left blank so I tried to create an example about how to use the component instances. Please let me know if it looks correct to you or if you think there is a better example to demonstrate the use of the component instances.


hkaiser:
@dimitraka thanks a lot for working on this! The example looks good to me. You may want to refer to the simple example we have in HPX here: https://github.com/STEllAR-GROUP/hpx/tree/master/examples/accumulators. Those are three variations on a simple accumulator component that knows how to add and print numbers.

hkaiser (approved):
LGTM, thanks!

PR #6468, opened 2024-04-03
Small vector relocation

isidorostsa:
`hpx::small_vector` is a small buffer optimized vector. This PR introduces two enhancements:

- Making `small_vector` compatible with the proposed [inplace_vector](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p0843r10.html). This is done with creating a `static_vector` typedef for  `small_vector` that limits it to only use the direct storage mode, without introducing overhead with that limitation.

- Utilizing P1144's relocation algorithms to insert and remove elements, as well as relocate the buffer when needed. This abstracts away much of the logic needed of those operations and also optimizes them based on P1144's relocation semantics. A `small_vector` is also marked as trivially relocatable when the type it is carrying is trivially relocatable itself. That is because it is always possible that the element will be stored in the object itself.

For a `small_vector<std::vector<T>>` the observable speedup of using trivial relocation for insertions, push_backs, and deletions is in the order of 2-3x, benchmarks and tests for this are to be added to this PR.

### DIFF
--- a/libs/core/datastructures/include/hpx/datastructures/detail/small_vector.hpp
+++ b/libs/core/datastructures/include/hpx/datastructures/detail/small_vector.hpp
@@ -36,6 +36,7 @@
 #include <hpx/type_support/construct_at.hpp>
 #include <hpx/type_support/is_trivially_relocatable.hpp>
 #include <hpx/type_support/uninitialized_relocation_primitives.hpp>
+#include <hpx/errors/throw_exception.hpp>
 
 #include <algorithm>
 #include <array>
@@ -220,7 +221,7 @@ namespace hpx::detail {
 
     // note: Allocator is currently unused
     template <typename T, std::size_t MinInlineCapacity,
-        typename Allocator = std::allocator<T>>
+        typename Allocator = std::allocator<T>, bool emulate_inplace_vector = false>
     class small_vector
     {
         static_assert(MinInlineCapacity <= 127,
@@ -245,13 +246,13 @@ namespace hpx::detail {
         //  m_data[0] & 1: lowest bit is 0 for indirect mode
         //  m_data[0..7]:  stores an uintptr_t, which points to the indirect
         //                 data.
-
         alignas(alignment_of_small_vector<T>()) std::array<std::uint8_t,
             size_of_small_vector<T>(MinInlineCapacity)> m_data;
 
         [[nodiscard]] constexpr auto is_direct() const noexcept -> bool
         {
-            return (m_data[0] & 1U) != 0U;
+            if constexpr (emulate_inplace_vector) return true;
+            else return (m_data[0] & 1U) != 0U;
         }
 
         [[nodiscard]] auto indirect() noexcept -> storage<T>*
@@ -298,6 +299,7 @@ namespace hpx::detail {
 
         void realloc(std::size_t new_capacity)
         {
+            static_assert(!emulate_inplace_vector, "If called in an inplace_vector, it is a bug.");
             if (new_capacity <= N)
             {
                 // put everything into direct storage
@@ -521,9 +523,17 @@ namespace hpx::detail {
         template <typename It>
         void assign(It first, It last, std::forward_iterator_tag /*unused*/)
         {
+            auto s = std::distance(first, last);
+            if constexpr (emulate_inplace_vector)
+            {
+                // Can not have an inplace_vector with a size larger than N
+                if (s <= capacity()) {
+                    throw std::bad_alloc();
+                }
+            }
+
             clear();
 
-            auto s = std::distance(first, last);
             reserve(s);
             std::uninitialized_copy(first, last, data());
             set_size(s);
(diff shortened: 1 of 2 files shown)


hkaiser:
@isidorostsa could you please take care of the inspect and clang-format issues? Also please rebase onto master.

isidorostsa:
> @isidorostsa could you please take care of the inspect and clang-format issues? Also please rebase onto master.

Yes, apologies for not already doing that

libs/core/datastructures/include/hpx/datastructures/detail/small_vector.hpp
```diff
@@ -294,22 +301,10 @@ namespace hpx::detail {
...
-            {
-                std::uninitialized_move_n(source_ptr, size, target_ptr);
-                std::destroy_n(source_ptr, size);
-            }
-        }
-
         void realloc(std::size_t new_capacity)
         {
+            static_assert(!emulate_inplace_vector,
+                "If called in an inplace_vector, it is a bug.");
```
Quuxplusone, line 307: FWIW, `std::inplace_vector` allows you to call `.reserve`. It's a no-op for small capacities, and invariably throws `bad_alloc` (or `length_error`? I'm not sure even LWG knows yet) for large capacities. I don't know if you consider that an improvement or not.

libs/core/datastructures/include/hpx/datastructures/detail/small_vector.hpp
```diff
@@ -525,9 +528,18 @@ namespace hpx::detail {
         template <typename It>
         void assign(It first, It last, std::forward_iterator_tag /*unused*/)
         {
+            auto s = std::distance(first, last);
+            if constexpr (emulate_inplace_vector)
+            {
+                // Can not have an inplace_vector with a size larger than N
+                if (s <= capacity())
```
Quuxplusone, line 535: Nit: `Cannot`
Major: Isn't this condition backwards? You mean `if (s > capacity())`.

libs/core/datastructures/include/hpx/datastructures/detail/small_vector.hpp
```diff
@@ -536,6 +548,8 @@ namespace hpx::detail {
         // precondition: all uninitialized
         void do_move_assign(small_vector&& other) noexcept
         {
+            // We assume that the template parameters of "other" are the same
+            // as "this".
```
Quuxplusone, line 552: The comment is superfluous: the _injected-class-name_ `small_vector` on line 549 literally means `small_vector<T, MinInlineCapacity, emulate_inplace_vector>` because that's how the language works.

libs/core/datastructures/include/hpx/datastructures/detail/small_vector.hpp
```diff
@@ -846,6 +943,10 @@ namespace hpx::detail {
 
         [[nodiscard]] constexpr auto capacity() const noexcept -> std::size_t
         {
+            if constexpr (emulate_inplace_vector)
+            {
+                return capacity<direction::direct>();
+            }
```
Quuxplusone, line 949: This seems redundant, since `is_direct()` returns constexpr `true` in that case. Or if you're trying to avoid even instantiating `capacity<direction::indirect>()` in that case, then it ought to be enclosed in an `else` block.

libs/core/datastructures/include/hpx/datastructures/detail/small_vector.hpp
```diff
@@ -1273,3 +1427,12 @@ namespace std {    //-V1061
         return num_removed;
     }
 }    // namespace std
+
+namespace hpx::experimental {
+    template <typename T, std::size_t N, typename Alloc, bool emul>
+    struct is_trivially_relocatable<
+        hpx::detail::small_vector<T, N, Alloc, emul>>
+      : is_trivially_relocatable<T>
```
Quuxplusone, line 1435: And `Alloc` needs to be trivially relocatable, and `Alloc::pointer`, and it must have the right POCMA/POCCA/POCS behavior if you care about any of that stuff. See https://github.com/Quuxplusone/llvm-project/blob/trivially-relocatable-v107/libcxx/include/__memory/allocator_traits.h#L421-L433 and https://github.com/Quuxplusone/llvm-project/blob/trivially-relocatable-v107/libcxx/include/vector#L399-L407 for my approach.

Quuxplusone (changes requested):
I strongly recommend that you split this into two PRs: one for refactoring around relocatability, and one for `inplace_vector`-emulation. They seem almost entirely orthogonal to each other.

isidorostsa:
> I strongly recommend that you split this into two PRs: one for refactoring around relocatability, and one for `inplace_vector`-emulation. They seem almost entirely orthogonal to each other.

You are absolutely right to point that out, I will be doing that.

libs/core/datastructures/include/hpx/datastructures/detail/small_vector.hpp
isidorostsa: Ah, that is interesting. Will correct and write tests for this.

hkaiser:
@isidorostsa what's the status of this? Could you rebase this onto master please?

isidorostsa:
> @isidorostsa what's the status of this? Could you rebase this onto master please?

I was looking at this PR yesterday, I'll split this in two parts, one regarding the inplace_vector aspect and one regarding the trivial relocation aspect.

PR #6471, opened 2024-04-10
Add distance.cpp test in CMake

gonidelis:
Neglected to add the unit test when I wrote it, so I just add it now :)


gonidelis:
fixed. sorry for that.

hkaiser (approved):
LGTM, thanks!

PR #6476, opened 2024-04-25
Fixing binary filter test target names

m-diers:
Not unique target names in the tests of binary_filter which leads to build errors when switching on the different filter at the same time, e.g. with
```
          -DHPX_WITH_COMPRESSION_BZIP2=ON \
          -DHPX_WITH_COMPRESSION_SNAPPY=ON \
          -DHPX_WITH_COMPRESSION_ZLIB=ON \
          -DHPX_WITH_TESTS=ON \
```

## Proposed Changes

  - unique target names for binary_filter tests

## Checklist

Replaced the non unique copy-paste name with the unique names of the different binary filter in the build config files.


hkaiser (approved):
LGTM, thanks!

PR #6479, opened 2024-05-02
Remove optimizations for certain built-in compiler intrinsics

### LINKED ISSUE #6412: Clangd (Language Server) throws error for __integer_pack at pack.hpp
I've built the hpx from source (using master from Jun 2023) using a gcc 13.1.1 compiler. My Ide for development is sometimes clion and also nvim that both use clangd language server. 

When going trough examples from the hpx I instantly get errors from the language server: Use  of undeclared identifier __integer_pack in `pack.hpp:49`. I know this is a placeholder for the gcc compiler but when running this with clangd it is pretty nerving sometimes...

Should `pack.hpp` use something in the direction of this:

```cpp
#if __has_builtin(__make_integer_seq)
      template<typename, size_t... _Indices>
	using _IdxTuple = _Index_tuple<_Indices...>;

      // Clang defines __make_integer_seq for this purpose.
      using __type = __make_integer_seq<_IdxTuple, size_t, _Num>;
#else
      // For GCC and other compilers, use __integer_pack instead.
      using __type = _Index_tuple<__integer_pack(_Num)...>;
#endif
```
(this is basically taken from [this](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/utility.h#L153))

In order to reproduce this just run:

```
clangd --check=main.cpp
``` 

Where main.cpp include 

```cpp
#include <hpx/hpx_main.hpp>
#include <hpx/iostream.hpp>

int main()
{
    // Say hello to the world!
    hpx::cout << "Hello World!\n" << std::flush;
    return 0;
}
```

One other option I tried was using clangd with --query-driver to query the gcc driver for the necessary system headers but this didn't fix the problem. 

Is there any solution to make the language server work with the source code of hpx ... ?

I'm not sure if clangd language server is a high priority now ... but I thought I mention this. If needed I could try to make a pull request.

hkaiser:
- those have caused trouble when the compiler used to build HPX is different from the compiler used to compile a dependent application

Fixes #6412

@andreichalapco could you please check whether this fixes the issue you reported?


hkaiser:
This looks like to be good to go. Merging...

PR #6492, opened 2024-05-19
Fix kokkos hpx nvcc compilation

G-071:
With CUDA 12.3 nvcc does not outright crash anymore when compiling sender-receiver code and with #6452 the HPX sender-receiver examples compile and work again with nvcc after fixing some compilation errors.

The last remaining problem related to this was to compile the Kokkos HPX execution space (using sender-receivers internally) with nvcc: Here we still got a few compilation problems such as:
- ``` member_pack.hpp(143): error: function returning function is not allowed ```
- ```split.hpp:289:38: error: use of deleted function ‘hpx::execution::experimental::detail::any_receiver<Ts>::any_receiver(const hpx::execution::experimental::detail::any_receiver<Ts>&) [with Ts = {}]’```

This PR addresses these issues!

By itself, this PR is not quite enough to make the HPX execution space compile with nvcc, we still need one more change within Kokkos itself. However, this PR resolves all issues on the HPX side, the rest (some small template deduction stuff) will consequently be added as a Kokkos PR.

Pinging @msimberg as he came up with the solution, I was merely the one reporting the problem and testing the fixes!

### DIFF
--- a/libs/core/execution/include/hpx/execution/algorithms/detail/partial_algorithm.hpp
+++ b/libs/core/execution/include/hpx/execution/algorithms/detail/partial_algorithm.hpp
@@ -27,7 +27,8 @@ namespace hpx::execution::experimental::detail {
     struct partial_algorithm_base<Tag, hpx::util::index_pack<Is...>, Ts...>
     {
     private:
-        HPX_NO_UNIQUE_ADDRESS hpx::util::member_pack_for<std::decay_t<Ts>...> ts;
+        HPX_NO_UNIQUE_ADDRESS hpx::util::member_pack_for<std::decay_t<Ts>...>
+            ts;
 
     public:
         template <typename... Us>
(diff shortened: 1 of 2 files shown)


hkaiser (approved):
LGTM, thanks!

msimberg:
Thanks @G-071! Just out of curiosity, could you check what your CMake configuration says about `HPX_HAVE_BUILTIN_FORWARD_MOVE` and `HPX_HAVE_CXX_LAMBDA_CAPTURE_DECLTYPE`, i.e. which version of `HPX_FORWARD` is your build using?

G-071:
> Thanks @G-071! Just out of curiosity, could you check what your CMake configuration says about `HPX_HAVE_BUILTIN_FORWARD_MOVE` and `HPX_HAVE_CXX_LAMBDA_CAPTURE_DECLTYPE`, i.e. which version of `HPX_FORWARD` is your build using?

@msimberg  I had deleted the build directory already, but upon rebuilding HPX and Kokkos the same way, the CMakeCache tells me No and Yes regarding those two:
```
HPX_WITH_BUILTIN_FORWARD_MOVE:INTERNAL=FALSE
//Result of TRY_COMPILE
HPX_WITH_BUILTIN_FORWARD_MOVE_RESULT:INTERNAL=FALSE

```
and
```
HPX_WITH_CXX_LAMBDA_CAPTURE_DECLTYPE:INTERNAL=TRUE
//Result of TRY_COMPILE
HPX_WITH_CXX_LAMBDA_CAPTURE_DECLTYPE_RESULT:INTERNAL=TRUE
```

PR #6497, opened 2024-05-29
Contains and contains_subrange parallel algorithm implementation GSOC 2024

Zak-K-Abdi:
Implementing parallel algorithms for contains and contains_subrange. GSOC 2024 proposal.
Will attempt to fix hpx::search so hpx::contains_subrange can make a call to hpx::search.


libs/core/algorithms/include/hpx/parallel/algorithms/contains.hpp
```diff
@@ -0,0 +1,128 @@
...
+                    (const auto& cur) { return HPX_INVOKE(proj, *cur) == val; });
+
+                if (itr == last)
+                {
+                    return util::detail::algorithm_result<ExPolicy, bool>::get(
+                        false);
+                }
+
+                return util::detail::algorithm_result<ExPolicy, bool>::get(true);
+            }
```
Pansysk75, line 70: This is probably work-in-progress, but I just want to mention that util::loop_pred doesn't actually use the execution policy to run in parallel, so this will not run in multiple threads. You'll have to use a partitioner, which splits the work and launches it on multiple (the [`equal` algorithm](https://github.com/STEllAR-GROUP/hpx/blob/3521f71bd2ccd04e370ac2181bb7b6987192704c/libs/core/algorithms/include/hpx/parallel/algorithms/equal.hpp#L593-L633) is a a nice example).

Zak-K-Abdi: Ah I was unaware at the time I wrote that, I should have been more mindful of what I was using. I checked out equal algorithm and I made some changes that now use the partitioner in the parallel. Also, I was hesitant to do this but I created a file in the parallel/algorithms/detail called contains.hpp, holding the implementation details for the algorithm.

hkaiser:
@Zak-K-Abdi: Please take a look at the clang-format problems (https://gist.github.com/hkaiser/26bced03497017f481dbb9ee7d0df67c) and the cmake issues:

`Missing hpx/parallel/algorithms/detail/contains.hpp in libs/core/algorithms/CMakeLists.txt` 

Also here is the list of inspect issues reported:

/libs/core/algorithms/include/hpx/parallel/algorithms/contains.hpp: *END* file doesn't end with a newline, *Endline Whitespace*: [74](https://github.com/STEllAR-GROUP/hpx/blob/f2496c9d051b8da2e9dedf633766484f840541ba/libs/core/algorithms/include/hpx/parallel/algorithms/contains.hpp#L74), [*C*](https://github.com/STEllAR-GROUP/hpx/blob/f2496c9d051b8da2e9dedf633766484f840541ba/libs/core/algorithms/include/hpx/parallel/algorithms/contains.hpp), [*Lic*](https://github.com/STEllAR-GROUP/hpx/blob/f2496c9d051b8da2e9dedf633766484f840541ba/libs/core/algorithms/include/hpx/parallel/algorithms/contains.hpp), [*SPDX-Lic*](https://github.com/STEllAR-GROUP/hpx/blob/f2496c9d051b8da2e9dedf633766484f840541ba/libs/core/algorithms/include/hpx/parallel/algorithms/contains.hpp)
/libs/core/algorithms/include/hpx/parallel/algorithms/detail/contains.hpp: *END* file doesn't end with a newline, [*C*](https://github.com/STEllAR-GROUP/hpx/blob/f2496c9d051b8da2e9dedf633766484f840541ba/libs/core/algorithms/include/hpx/parallel/algorithms/detail/contains.hpp), [*Lic*](https://github.com/STEllAR-GROUP/hpx/blob/f2496c9d051b8da2e9dedf633766484f840541ba/libs/core/algorithms/include/hpx/parallel/algorithms/detail/contains.hpp), [*SPDX-Lic*](https://github.com/STEllAR-GROUP/hpx/blob/f2496c9d051b8da2e9dedf633766484f840541ba/libs/core/algorithms/include/hpx/parallel/algorithms/detail/contains.hpp)

Pansysk75:
retest lsu

libs/core/algorithms/include/hpx/parallel/container_algorithms/contains.hpp
```diff
@@ -0,0 +1,367 @@
...
+    };
+
+    template <typename ExPolicy, typename T>
+    struct contains_subrange_helper<ExPolicy, T,
+        std::enable_if_t<hpx::is_async_execution_policy_v<ExPolicy>>>
+    {
+        static hpx::future<bool> get(hpx::future<T>& itr, T& last)
+        {
+            return itr.then(
+                [&last](hpx::future<T> it) { return it.get() != last; });
```
Zak-K-Abdi, line 102: @Pansysk75 Hi Panos, I'm not sure if this is correct, could you check this line of code I changed? I changed the helper function to return hpx::future, a continuation to the previous future, by using .then().

libs/core/algorithms/include/hpx/parallel/container_algorithms/contains.hpp
```diff
@@ -148,8 +137,7 @@ namespace hpx::parallel { namespace detail {
                 HPX_FORWARD(Proj2, proj2));
 
             return util::detail::algorithm_result<ExPolicy, bool>::get(
-                contains_subrange_helper<ExPolicy, FwdIter1>().get(itr) !=
-                last1);
+                contains_subrange_helper<ExPolicy, FwdIter1>().get(itr, last1));
```
Pansysk75, line 140: I think returning `contains_subrange_helper<ExPolicy, FwdIter1>().get(itr, last1)` would suffice here. 
The return types of your helper function already match with `algorithm_result_t`. Calling that additional `::get` will call `.get()` on your future before returning it, which undermines your latest change on your helper.

libs/core/algorithms/include/hpx/parallel/container_algorithms/contains.hpp
Pansysk75: That looks right, i think you might want to capture `last` by value and not by reference, as it is a local variable of the `parallel` function. Otherwise looks good!

Pansysk75:
@Zak-K-Abdi Could you please address this: https://github.com/STEllAR-GROUP/hpx/pull/6497#discussion_r1744600305
Otherwise this looks ready to merge, thanks for your efforts!

Zak-K-Abdi:
> @Zak-K-Abdi Could you please address this: [#6497 (comment)](https://github.com/STEllAR-GROUP/hpx/pull/6497#discussion_r1744600305) Otherwise this looks ready to merge, thanks for your efforts!

Apologies, I completely forgot about that comment. I will make that change soon. Thanks.

libs/core/algorithms/include/hpx/parallel/container_algorithms/contains.hpp
Zak-K-Abdi: @Pansysk75 Just to make sure, should I also pass last by value for the functions in my contains_subrange_helper_class or only for the lambda?

Pansysk75:
retest lsu

hkaiser (approved):
LGTM, thanks!

PR #6499, opened 2024-05-31
Fix crash in get_executable_filename on FreeBSD

yurivict:
The std::vector object needs to have the element allocated in order to take its address.


hkaiser:
> The std::vector object needs to have the element allocated in order to take its address.

Excellent catch, thanks!

hkaiser (dismissed):
LGTM, thanks!

PR #6511, opened 2024-06-17
Fix HPX_WITH_FETCH_BOOST

Pansysk75:
Fixes following issues when HPX_WITH_FETCH_BOOST=ON:
- Boost being re-fetched on projects that would depend on HPX.
- Boost not compiling necessary libraries when a non header-only instance of Boost was needed

Now when `HPX_WITH_FETCH_BOOST=ON` Boost will be installed alongside HPX.

To fix:
- [ ] Duplicated code for figuring out which Boost libraries are needed
- [ ] Make hpx_boost_root-install.cmake contain relative path instead of absolute(?)


hkaiser:
@Pansysk75 could you rebase this onto master to resolve the conflicts, please?

Pansysk75:
> @Pansysk75 could you rebase this onto master to resolve the conflicts, please?

This PR should be discarded in favor of #6510.

PR #6513, opened 2024-06-18
custom FETCHCONTENT_BASE_DIR bug fix

pefedotov:
Fixes a bug related to incorrect paths when automatically loading `hwloc`.

If the path for downloaded projects is `FETCHCONTENT_BASE_DIR`
 has been changed, then the existing `cmake` script points to a non-existent folder, since `hwloc` will be loaded to another location.

### DIFF
--- a/cmake/HPX_SetupHwloc.cmake
+++ b/cmake/HPX_SetupHwloc.cmake
@@ -47,10 +47,10 @@ else()
       execute_process(
         COMMAND
           sh -c
-          "cd ${CMAKE_BINARY_DIR}/_deps/hwloc-src && ./configure --prefix=${CMAKE_BINARY_DIR}/_deps/hwloc-installed && make -j && make install"
+          "cd ${FETCHCONTENT_BASE_DIR}/hwloc-src && ./configure --prefix=${FETCHCONTENT_BASE_DIR}/hwloc-installed && make -j && make install"
       )
     endif()
-    set(HWLOC_ROOT "${CMAKE_BINARY_DIR}/_deps/hwloc-installed")
+    set(HWLOC_ROOT "${FETCHCONTENT_BASE_DIR}/hwloc-installed")
     set(Hwloc_INCLUDE_DIR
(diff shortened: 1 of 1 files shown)


cmake/HPX_SetupHwloc.cmake
```diff
@@ -111,11 +111,11 @@ else()
     add_custom_target(
       HwlocDLL ALL
       COMMAND ${CMAKE_COMMAND} -E make_directory
-              "${CMAKE_BINARY_DIR}/$<CONFIG>/bin/"
+              ${RUNTIME_OUTPUT_DIRECTORY}
```
hkaiser, line 114: Will this do the right thing even for multi-configuration generators? Wouldn't for those the generated directory will have the `Debug`/`Release` etc. configuration names appended?

hkaiser:
Also, please address the cmake-format issue reported (see https://gist.github.com/hkaiser/debb4ce2e5cf64722db0f49c90cee01a)

cmake/HPX_SetupHwloc.cmake
pefedotov: This works for me since I have paths overridden for each project. If `RUNTIME_OUTPUT_DIRECTORY` is not specified, then popular generators themselves create a `debug` or `release` folder. The main thing is that this is the folder where the result of the compilation will be located, namely the executable file, next to which there should be a `dll`.
Considering that the build tests on `windows` and `mac` failed, I think there is something wrong with this code in the general case. I'll try to find a solution in my free time.

About the variable:
https://cmake.org/cmake/help/latest/prop_tgt/RUNTIME_OUTPUT_DIRECTORY.html

pefedotov:
Since cmake does not provide a single variable for all occasions, I decided to create a condition. If a special path was specified, then use it, otherwise, what was initially specified in the code (default path).

hkaiser (approved):
LGTM, thanks!

hkaiser:

PR #6539, opened 2024-09-07
Use thread-safe cache in thread_local_caching_allocator

Pansysk75:


hkaiser (dismissed):
The remaining problems are unrelated. LGTM! Thanks!

libs/core/allocator_support/include/hpx/allocator_support/thread_local_caching_allocator.hpp
```diff
@@ -82,8 +83,9 @@ namespace hpx::util {
                 }
                 else
                 {
-                    p = data.top().first;
-                    data.pop();
+                    std::pair<T*, size_type> pair;
+                    data.pop(pair);
```
hkaiser, line 87: I think you should check for the return value here as between the check `data.empty()` above and here somebody could have removed the item. Even better if you rewrote the whole functions to something like:
```
std::pair<T*, size_type> pair;
if (!data.pop(pair)) {
    // allocate new
}
else {
   p = pair.first;
}
```

Pansysk75: Fixed, thanks!

hkaiser:
@Pansysk75 we still have the circular dependency between the concurrency and allocator modules. Also, some of the header tests are failing (most likely caused by the circular dependency issue)

Pansysk75:
> @Pansysk75 we still have the circular dependency between the concurrency and allocator modules. Also, some of the header tests are failing (most likely caused by the circular dependency issue)

Hmm I suppose that means the `allocator_support` module is higher in the module hierarchy than the `concurrency` module?
I'm not exactly sure what fixing this would involve.

hkaiser:
> > @Pansysk75 we still have the circular dependency between the concurrency and allocator modules. Also, some of the header tests are failing (most likely caused by the circular dependency issue)
> 
> Hmm I suppose that means the `allocator_support` module is higher in the module hierarchy than the `concurrency` module? I'm not exactly sure what fixing this would involve.

Yes, see: https://hpx-docs.stellar-group.org/latest/report/v1.10.0/core/allocator_support.html

> Oh, also, we should probably make all accesses to members of `allocated_cache` thread safe... Not sure why I ignored that. Also, this now makes me dislike this solution even more.

Yes, the `deallocated` and `allocated` members should be atomic. But we should combine both variables into one counter for that.

> I opened an issue #6540 to document this.

Thanks!

hkaiser:
@Pansysk75 I think this is good to go as a stop-gap measure circumventing the underlying problem from happening. What do you think?

Pansysk75:
> @Pansysk75 I think this is good to go as a stop-gap measure circumventing the underlying problem from happening. What do you think?

Sure, one last thing, what is this check supposed to do? Our underlying data structure has a .size() member function, is it any different to use that?

hkaiser:
@Pansysk75 the original issues are resolved here. Let's merge it now.

PR #6543, opened 2024-09-11
Throwing an exception derived from std::bad_alloc on OOM conditions

### LINKED ISSUE #6542: new_handler must throw `std::bad_alloc` or something derived from it.
Currently our `new_handler` that is invoked in OOM contexts throws an `hpx::exception` (see https://github.com/STEllAR-GROUP/hpx/blob/f4ff6e807302faddc58b39f3c3e20950d73b9813/libs/core/runtime_local/src/runtime_local.cpp#L230-L234). This is UB.

hkaiser:
- flyby: don't register signal(SIGABRT, ...) if hpx.handle_signals == 0

Fixes #6542


hkaiser:
This seems to be fine. Merging...

PR #6549, opened 2024-10-03
Fix CMake conditionals checking ENV variables

Pansysk75:
Fixes CMake conditional checks that depend on environment variables, since `if(ENV{some_var})` would always evaluate to false:
https://cmake.org/cmake/help/latest/command/if.html#variable


hkaiser (approved):
Excellent catch! 
LGTM, thanks!

Pansysk75:
retest lsu

PR #6555, opened 2024-10-21
Fix title and minor typos in documentation

dimitraka:
The title of our documentation is "Welcome to the HPX documentation!" which I feel like is a bit odd. 

This means that also when someone searches for the docs in google they see this title. Apart from being odd, if you use a device with smaller screen, the title doesn't fit and you only see "Welcome to the..." 

Our title could simply be "HPX documentation". The welcome message is moved to the beginning of the first paragraph.


hkaiser (approved):
LGTM, thanks!

PR #6581, opened 2024-11-25
Attempting to work around a Boost.Spirit problem

### LINKED ISSUE #6579: Boost.Spirit 1.76.0 compilation errors on Windows due to missing NOMINMAX
## Issue

Recently I was interested in experimenting with HPX on Windows so I cloned the GitHub repo and checked the `v1.10.0` tag so I could build from source. I was following the [Windows build instructions](https://hpx-docs.stellar-group.org/latest/html/manual/building_hpx.html#windows) but figured I could build directly using CMake instead of opening Visual Studio, so after building and separately installing Boost 1.76.0, Asio 1.32.0, and hwloc 2.11.2, I ran:

```cmd
cmake -S . -B build_windows_x64 -A x64 && cmake --build build_windows_x64 --config Release
```

Although the CMake configuration step succeeded, I noticed that files including Boost.Spirit headers would suffer compilation errors. For example, when compiling `libs/core/affinity/src/parse_affinity_options.cpp`, I observed errors like:

```
range_functions.hpp(53, 36): error C2589: '(': illegal token on right side of '::'
range_functions.hpp(53, 36): error C2760: syntax error: ')' was unexpected here, expected 'expression'
range_functions.hpp(53, 36): error C2760: syntax error: ')' was unexpected here, expected ';'
range_functions.hpp(53, 36): error C3878: syntax error: unexpected token ')' following 'expression_statement'
```

For brevity, only the relevant portions of the first few compiler errors are shown. I saw that similar errors are seen in other files that use Boost.Spirit, in particular `libs/core/batch_environments/src/slurm_environment.cpp`.

## Workaround

Fortunately, I realized this might be due to macro expansion of `min` and `max` from a transitive `Windows.h` include so I set `CXXFLAGS` in the environment to be `/D NOMINMAX` which suppresses the `min` and `max` macro definitions in `Windows.h`.

Furthermore, to directly address the problem, I tried the following `CMakeLists.txt` change:

```diff
   hpx_add_config_cond_define(_WINSOCK_DEPRECATED_NO_WARNINGS)
   hpx_add_config_cond_define(_CRT_NON_CONFORMING_SWPRINTFS)
   hpx_add_config_cond_define(_SILENCE_FPOS_SEEKPOS_DEPRECATION_WARNING)
+  # need to unset Windows.h min and max macros as they conflict with
+  # Boost.Spirit usage of limits::min() and limits::max()
+  hpx_add_config_cond_define(NOMINMAX)

   # ############################################################################
   # Boost
```

Performing a clean build, i.e. deleting the `build_windows_x64` build directory and then re-running CMake, worked.

## Discussion

Since even the latest HEAD of `master` does not define `NOMINMAX` during Windows compilation I think that having this as a formal fix for Windows builds would be helpful. However, after trying to build HPX with Boost 1.79.0 instead, this issue seems to go away. Perhaps since 1.76.0 version of Boost.Spirit, or maybe just Boost in general, this may have been fixed.

I did want to open an issue so that there is a written record of my findings, but I'm not sure if a formal fix is necessary given that simply upgrading the installed version of Boost seems to fix the problem and the `CXXFLAGS` workaround is easy to do.

hkaiser:
Fixes #6579

@phetdam would you be able to try this out?

### DIFF
--- a/components/performance_counters/io/CMakeLists.txt
+++ b/components/performance_counters/io/CMakeLists.txt
@@ -28,6 +28,11 @@ if(HPX_WITH_IO_COUNTERS)
     SOURCES ${io_counters_sources} ${HPX_WITH_UNITY_BUILD_OPTION}
   )
 
+  # see #6579
+  if(MSVC AND Boost_VERSION_STRING VERSION_LESS "1.79.0")
+    target_compile_definitions(io_counters_component PRIVATE NOMINMAX)
+  endif()
+
   add_hpx_pseudo_dependencies(
     components.performance_counters.io io_counters_component
   )
--- a/components/performance_counters/memory_counters/CMakeLists.txt
+++ b/components/performance_counters/memory_counters/CMakeLists.txt
@@ -38,6 +38,11 @@ add_hpx_component(
   SOURCES ${memory_sources} ${HPX_WITH_UNITY_BUILD_OPTION}
 )
 
+# see #6579
+if(MSVC AND Boost_VERSION_STRING VERSION_LESS "1.79.0")
+  target_compile_definitions(memory_counters_component PRIVATE NOMINMAX)
+endif()
+
 add_hpx_pseudo_dependencies(
   components.performance_counters.memory_counters memory_counters_component
 )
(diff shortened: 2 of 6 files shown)


phetdam:
> Fixes #6579
> 
> @phetdam would you be able to try this out?

Yes, my pleasure. I will comment the result soon but I don't foresee any problems.

phetdam:
@hkaiser Tested with Boost 1.76.0 and there are no cryptic Boost.Spirit errors anymore.

PR #6589, opened 2024-12-15
Update config.yml

emiliostassios:
Updated  docker image

Fixes # Updated the docker image version to the updated one.


hkaiser:
@emiliostassios could you please rebase onto master to fix the CircleCI errors reproted?

PR #6591, opened 2024-12-17
Feature test for __builtin_frame_address

isidorostsa:
Add a feature test for `__builtin_frame_address`. Before, it was assumed that if the compiler `__GNUC__` then `__builtin_frame_address` is available: https://github.com/STEllAR-GROUP/hpx/blob/3c0ebc8a5f23b7c4f197fa1970c4560a61e16b34/libs/core/coroutines/include/hpx/coroutines/detail/get_stack_pointer.hpp#L33-L35

But this is not the case when building HPX with [Circle](https://www.circle-lang.org/site/index.html).

### DIFF
--- a/cmake/HPX_AddConfigTest.cmake
+++ b/cmake/HPX_AddConfigTest.cmake
@@ -251,6 +251,15 @@ function(hpx_check_for_builtin_forward_move)
 endfunction()
 
 # ##############################################################################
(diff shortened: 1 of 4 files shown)


hkaiser:
@isidorostsa could you please fix the cmake-format issue?

hkaiser (approved):
The remaining failures are unrelated. LGTM, thanks!

PR #6594, opened 2024-12-19
Move get_stack_ptr to source 

isidorostsa:
`get_stack_ptr()` is moved to a source file so it still works when there is a mismatch in the features of the compiler building HPX and the compiler building the HPX headers along with the application that uses them.

### DIFF
--- a/libs/core/coroutines/include/hpx/coroutines/detail/get_stack_pointer.hpp
+++ b/libs/core/coroutines/include/hpx/coroutines/detail/get_stack_pointer.hpp
@@ -8,6 +8,7 @@
 
 #pragma once
 
+#include <hpx/config.hpp>
 #include <hpx/config/compiler_specific.hpp>
 #include <hpx/config/defines.hpp>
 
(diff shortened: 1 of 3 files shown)


libs/core/coroutines/include/hpx/coroutines/detail/get_stack_pointer.hpp
```diff
@@ -8,6 +8,7 @@
 
 #pragma once
 
+#include <hpx/config.hpp>
 #include <hpx/config/compiler_specific.hpp>
 #include <hpx/config/defines.hpp>
```
hkaiser, line 13: If you include <hpx/config.hpp>, then the other two #includes are not needed anymore (those are pulled in by it).

libs/core/coroutines/include/hpx/coroutines/detail/get_stack_pointer.hpp
```diff
@@ -28,28 +29,6 @@
 #include <limits>
 
 namespace hpx::threads::coroutines::detail {
```
hkaiser, line 31: Since the function is defined only for `#if !defined(HPX_WINDOWS)`, shouldn't it be declared only in that case as well?

isidorostsa: I think what you are describing is what already happens since https://github.com/STEllAR-GROUP/hpx/blob/7624b2e9006423f693757489d3e254f5e9d500d1/libs/core/coroutines/include/hpx/coroutines/detail/get_stack_pointer.hpp#L15 extends to the bottom of the document

hkaiser (approved):
LGTM, thanks!

PR #6598, opened 2025-01-05
Adding synchronous collective operations

hkaiser:
- adding predefined world_comunicator


hkaiser:
The remaining errors are well-known and unrelated.

PR #6600, opened 2025-01-07
Fixing sync collectives

hkaiser:
- adding example
- adding sync overloads for set/get
- adding predefined channel communicator


hkaiser:
The remaining issues are unrelated.

PR #6609, opened 2025-01-27
Update perftest CI reference measurements

Pansysk75:


hkaiser (commented):
One of the baseline measurements seem to be off, still. Coul you please have a look again?

PR #6614, opened 2025-02-13
Fix fallback to module mode for CMake finding Boost

Pansysk75:
This allows finding instances of Boost that are not installed in the conventional way (for example, using Boost directly from its source directory without building it, if one relies on header-only Boost libraries).


hkaiser (approved):
LGTM, thanks!

PR #6628, opened 2025-03-02
Link aclocal with aclocal-1.16 as hwloc asks for it

SAtacker:
MacOS CI fails because it cannot find aclocal-1.16 for hwloc


hkaiser (approved):
LGTM, thanks!

PR #6626, opened 2025-02-26
Fixing MPI parcel port issue exposed by #6623

hkaiser:
Fixes #6623

- flyby: remove double locking of MPI calls

@m-diers: could you please check if this fixes your problem?
@JiakunYan: please verify that the same issue doesn't exist in the LCI parcelport

PR #6637, opened 2025-03-11
lci/mpi pp: fix the case when non-zero-copy data is larger than INT_MAX

JiakunYan:


hkaiser (dismissed):
LGTM, thanks!

hkaiser:
@JiakunYan thanks for the fix! Would you be able to apply a similar fix for the MPI pp as well?

JiakunYan:
> @JiakunYan thanks for the fix! Would you be able to apply a similar fix for the MPI pp as well?

Yes, I will push an MPI pp fix later.

m-diers:
LGTM, thanks!
Tested against LCI and MPI. As already mentioned, there is still the problem with TCP.

hkaiser:
> LGTM, thanks! Tested against LCI and MPI. As already mentioned, there is still the problem with TCP.

As @JiakunYan found out, the TCP problem is caused by a deliberate check for incoming messages not being larger than a given threshold (by default that's 2 GBytes, I believe). We introduced this limitation to reduce the chance of malicious attacks in cases where the TCP parcelport is used for 'outside' communication of the running HPX application. You can increase this limit by either

- setting the env variable `HPX_PARCEL_MAX_MESSAGE_SIZE` to whatever upper limit you want (I will change the code such that setting it to `-1` will set the limit to `'unlimited'`.
- passing a command line argument `--hpx:init=hpx.parcel.max_message_size=...` with the same effect.
- adding a compilation option while compiling HPX: `-DHPX_PARCEL_MAX_MESSAGE_SIZE=...` to change the default to a new value.

In general, I would like to leave the message limit in place for the TCP parcelport for the time being for security reasons.

hkaiser:
> * setting the env variable `HPX_PARCEL_MAX_MESSAGE_SIZE` to whatever upper limit you want (I will change the code such that setting it to `0` will set the limit to `'unlimited'`.

This has been added and will be merged as part of https://github.com/STEllAR-GROUP/hpx/pull/6626

hkaiser (approved):
The failures are unrelated.

PR #6642, opened 2025-03-13
Allowing to use custom iterator tags with iterator_facade

### LINKED ISSUE #6640: iterator_facade doesn't work with custom iterator categories
https://github.com/STEllAR-GROUP/hpx/blob/d736bb56502f4b3e36be1563f993566a817f4276/libs/core/iterator_support/include/hpx/iterator_support/iterator_facade.hpp#L318

`iterator_facade_base` uses iterator category to specialize the interface, but this requires an exact type match. A category that extends `std::random_access_iterator_tag` won't match the specialization and instead get the base specialization.

It's possible to coerce a custom category to its underlying iterator tag with overloading resolution:

```
std::random_access_iterator_tag coerce_iterator_tag(std::random_access_iterator_tag const&);
std::bidirectional_iterator_tag coerce_iterator_tag(std::bidirectional_iterator_tag const&);
std::forward_iterator_tag coerce_iterator_tag(std::forward_iterator_tag const&);
std::input_iterator_tag coerce_iterator_tag(std::input_iterator_tag const&);
std::output_iterator_tag coerce_iterator_tag(std::output_iterator_tag const&);
```

hkaiser:
Fixes #6640

### DIFF
--- a/libs/core/iterator_support/include/hpx/iterator_support/iterator_facade.hpp
+++ b/libs/core/iterator_support/include/hpx/iterator_support/iterator_facade.hpp
@@ -383,11 +383,19 @@ namespace hpx::util {
         typename Reference = T&, typename Distance = std::ptrdiff_t,
         typename Pointer = void>
     struct iterator_facade
-      : detail::iterator_facade_base<Derived, T, Category, Reference, Distance,
-            Pointer>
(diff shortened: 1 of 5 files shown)


K-ballo:
This solves the issue I'm having

PR #6654, opened 2025-03-27
2

charan-003:
Added hpx::execution::experimental::parallel_scheduler per P2079 Section 4.1:
- Implements schedule() returning a parallel_sender for single tasks.
- Supports bulk_t customization with bulk_sender for parallel bulk execution.
- Tested with hello_world.cpp (single and bulk tasks).


charan-003:
sorry this is something else

PR #6655, opened 2025-03-27
Implement parallel_scheduler in HPX 

### LINKED ISSUE #6601: Implement P2079 for HPX (System execution context)
See: wg21.link/p2079

charan-003:
Fixes #6601 

# Implement P2079R7 parallel_scheduler in HPX Core

This issue tracks the implementation of `hpx::execution::experimental::parallel_scheduler` in HPX core, aligning with P2079R7 (Parallel scheduler, Section 4.1) by wrapping `thread_pool_policy_scheduler` with `hpx::launch` policies. The goal is to provide a standards-compliant parallel scheduler with global access, cancellation support, task chaining, and bulk operations.

## Implementation Checklist
- [x] Phase 1: Core Implementation
- Implement `hpx::execution::experimental::parallel_scheduler` class per P2079R7 Section 4.1.
- Add `schedule()` returning `parallel_sender` with `sender_concept` and `completion_signatures`:
- Supports `set_value_t()`, `set_stopped_t()`, `set_error_t(std::exception_ptr)`.
- Wrap `thread_pool_policy_scheduler` with `hpx::launch` (async/sync policies).
- Implement `noexcept` move/copy constructors and assignment operators.
- Add `operator==` returning `true` for scheduler comparison.
-  Return `forward_progress_guarantee::parallel` via `get_forward_progress_guarantee_t`.
- Implement `then` operation for task chaining with `parallel_sender`.
- Code to rely on `STDEXEC`.

- [x] Phase 2: Bulk Support
- Extend `bulk_t` customization to delegate to `thread_pool_scheduler_bulk`.
- Implement `bulk_sender` for parallel bulk execution per P2079R7.
- Support `bulk_chunked` and `bulk_unchunked` operations.

- [x] Phase 3: Replaceability API
- Implement `std::execution::system_context_replaceability` namespace per P2079R7 Section 4.2.
- Add `query_parallel_scheduler_backend()` returning `shared_ptr<parallel_scheduler>`.
- Support link-time replaceability using weak symbols.
- Implement `receiver` and `bulk_item_receiver` interfaces for frontend-backend interaction.
- Ensure `storage` handling for scheduling operations.

- [x] [P3927R2](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p3927r2.html) extends C++26's task_scheduler (a type-erased scheduler wrapper) to support efficient parallel bulk operations. Previously, bulk work executed through task_scheduler would run sequentially even when wrapping a parallel_scheduler, defeating the purpose of parallel execution.

- [x] [P3804R2](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p3804r2.html) refines the  parallel_scheduler specification  (from P2079) to clarify execution policy semantics and eliminate implementation ambiguities
   - seq policy: Backend receives count=1, executes all work in a single call (truly sequential)
   - par policy: Backend receives count=shape, distributes work across multiple calls/threads (truly parallel)


charan-003 (commented):
clang format

charan-003:
Update:
- Moved `parallel_scheduler`, `parallel_sender`, and `bulk_sender` to a new header: `hpx/executors/parallel_scheduler.hpp`
- Removed related code from `thread_pool_scheduler.hpp`

@hkaiser — ready for re-review when convenient. Thanks for the guidance!

charan-003:
@hkaiser Can you please verify it now? I'm not sure why those 2 tests are failing.

libs/core/executors/include/hpx/executors/parallel_scheduler.hpp
```diff
@@ -0,0 +1,285 @@
...
+            // std::cout << "Bulk operation state created" << std::endl;
+        }
+
+        friend void tag_invoke(start_t, bulk_operation_state& os) noexcept
+        {
+            // std::cout << "Bulk start() called" << std::endl;
+            hpx::detail::try_catch_exception_ptr(
+                [&]() {
+                    thread_pool_scheduler exec{os.scheduler.get_thread_pool()};
+                    for (Shape i = 0; i < os.shape; ++i)
```
hkaiser, line 164: Isn't the `parallel_scheduler` supposed to run the tasks concurrently?

libs/core/executors/include/hpx/executors/parallel_scheduler.hpp
```diff
@@ -0,0 +1,285 @@
...
+            const parallel_scheduler&) noexcept = default;
+        parallel_scheduler& operator=(parallel_scheduler&&) noexcept = default;
+
+        friend constexpr bool operator==(const parallel_scheduler& lhs,
+            const parallel_scheduler& rhs) noexcept
+        {
+            return lhs.pool_ == rhs.pool_;
+        }
+
+        hpx::threads::thread_pool_base* get_thread_pool() const noexcept
```
hkaiser, line 51:
```suggestion
        constexpr hpx::threads::thread_pool_base* get_thread_pool() const noexcept
```

libs/core/executors/include/hpx/executors/parallel_scheduler.hpp
```diff
@@ -0,0 +1,285 @@
...
+        using sender_concept = hpx::execution::experimental::sender_t;
+#endif
+        using completion_signatures =
+            hpx::execution::experimental::completion_signatures<
+                hpx::execution::experimental::set_value_t(),
+                hpx::execution::experimental::set_error_t(std::exception_ptr),
+                hpx::execution::experimental::set_stopped_t()>;
+
+        template <typename Env>
+        friend auto tag_invoke(
```
hkaiser, line 129:
```suggestion
        friend constexpr auto tag_invoke(
```

libs/core/executors/include/hpx/executors/parallel_scheduler.hpp
```diff
@@ -0,0 +1,285 @@
...
```
hkaiser, line 207:
```suggestion
        friend constexpr auto tag_invoke(
```

libs/core/executors/include/hpx/executors/parallel_scheduler.hpp
```diff
@@ -0,0 +1,285 @@
...
+        {
+            // std::cout << "Bulk start() called" << std::endl;
+            hpx::detail::try_catch_exception_ptr(
+                [&]() {
+                    thread_pool_scheduler exec{os.scheduler.get_thread_pool()};
+                    for (Shape i = 0; i < os.shape; ++i)
+                    {
+                        exec.execute([i, &os]() mutable {
+                            // std::cout << "Bulk task executing for index: " << i <<;
+                            os.f(i);
```
hkaiser, line 168:
```suggestion
                            HPX_INVOKE(os.f, i);
```

libs/core/executors/include/hpx/executors/parallel_scheduler.hpp
```diff
@@ -0,0 +1,285 @@
...
+                            if (--(*os.tasks_remaining) == 0)
```
hkaiser, line 169: Using `std::atomic_ref` instead of a `std::shared_ptr` would be preferrable (at least for C++20)

hkaiser (commented):
Wow, I'm impressed! I have a couple of minor comments, though.

libs/core/executors/include/hpx/executors/parallel_scheduler.hpp
hkaiser: FWIW, you should be able to rely on the bulk implementation for the `thread_pool_scheduler` here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp#L877-L906

libs/core/executors/include/hpx/executors/parallel_scheduler.hpp
```diff
@@ -0,0 +1,285 @@
...
+
+        template <typename Receiver>
+        friend auto tag_invoke(connect_t, parallel_sender& s, Receiver& r)
+        {
+            return operation_state<parallel_scheduler, Receiver>{
+                s.scheduler, r};
+        }
+
+        template <typename Shape, typename F>
+        friend auto tag_invoke(bulk_t, parallel_sender&& s, Shape shape, F&& f)
```
hkaiser, line 229: For `bulk`, we have implemented an extension allowing to pass an arbitrary range as the `Shape` (in addition to the integral value), see here, for instance: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/execution/include/hpx/execution/algorithms/bulk.hpp#L273-L300. It might be a good idea to support this here as well.

hkaiser: Also, you may want to move all types that are don't have to visible by users into `namespace detail` to avoid name clashes with other code.

hkaiser:
Also, you may want to move all types that are don't have to visible by users into `namespace detail` to avoid name clashes with other code.

charan-003:
> Also, you may want to move all types that are don't have to visible by users into `namespace detail` to avoid name clashes with other code.

Sure, I'll do that soon
Thank you so much

charan-003:
@hkaiser 
The `parallel_bulk_operation_state` implements a custom chunking strategy for bulk operations. It divides the workload into chunks based on the number of threads in the underlying thread_pool_scheduler’s pool (`get_os_thread_count()`), then schedules each chunk as a separate task using `exec.execute()`

The custom bulk chunking might duplicate functionality in `thread_pool_scheduler_bulk.hpp`

Offers a parallel execution option that leverages HPX’s thread pool, maintaining high performance for CPU-bound tasks.

While this version lacks P2079R7’s replaceability API (section 4.2) and many more, it could be extended to support it ( `query_parallel_scheduler_backend()`). This would let HPX users swap in custom schedulers (integrating with GPUs or other runtimes like oneTBB), enhancing HPX’s flexibility and composability

will try to work on this :))

hkaiser:
> i'm not sure why it's failing for clang_format. i used `clang-format` -i ` to make sure it's in the format

The CI uses an older version of clang-format that does not always agree with newer version on how to format things. Simply surround the offending lines with 
```
// clang-format off
    ....
// clang-format on
```
to disable it checking those.

libs/core/executors/include/hpx/executors/parallel_scheduler.hpp
```diff
@@ -0,0 +1,344 @@
...
+                                        HPX_INVOKE(os.f, *it);
+                                    }    // clang-format on
+                                }
+                                if (--(*os.tasks_remaining) == 0)
+                                {
+                                    hpx::execution::experimental::set_value(
+                                        os.receiver);
+                                }
+                            });
+                        }
```
hkaiser, line 232: This will still execute the tasks sequentially and not concurrently. 

As I said before, wouldn't we be able to simply create a type alias `using parallel_scheduler = thread_pool_scheduler;` instead of trying to re-implement all of it?

If there is an API difference between `parallel_scheduler` and the existing `thread_pool_scheduler` (beyond the name), or if there is a semantic difference, please highlight that. In that case a simple wrapper may still be feasible.

charan-003: @hkaiser
Thanks for your feedback!. I’ve updated it to use `hpx::async`, `hpx::future`, and `hpx::when_all` in `default_parallel_scheduler` to ensure concurrent execution of bulk tasks (chunks run in parallel via HPX’s thread pool), addressing your comment about sequential execution. Single tasks also use `hpx::async` for lightweight scheduling.

Is it okay to rely on `hpx::async`, `hpx::future`, and `hpx::when_all` for task scheduling and execution, given they’re part of HPX’s lightweight threading system? Or would you recommend something else?

charan-003:
@hkaiser @isidorostsa 
This version is about adding P2079R7 parallel_scheduler to HPX core

Implement parallel_scheduler wrapping thread_pool_policy_scheduler<hpx::launch>, aligning with P2079R7 user-facing API:
- Add get_parallel_scheduler() for global access with async/sync policy selection.
- Include parallel_sender with `set_value_t(), set_stopped_t(), set_error_t(std::exception_ptr)`.
- Support cancellation via stop token in `operation_state::start_t`.
- Ensure parallel forward progress guarantee.
- Add `then` operation for task chaining.
- Add tests for progress, single task, cancellation, and exceptions.

libs/core/executors/include/hpx/executors/experimental/parallel_scheduler.hpp
```diff
@@ -0,0 +1,307 @@
...
+#include <hpx/execution/algorithms/then.hpp>
+#include <hpx/executors/thread_pool_scheduler.hpp>
+#include <hpx/functional/invoke.hpp>
+#include <hpx/synchronization/stop_token.hpp>
+
+#include <cstddef>
+#include <exception>
+#include <utility>
+
+// Forward declarations for execution::experimental
```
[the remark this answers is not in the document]
charan-003: yes, those includes are necessary `<exception>`  provided the `std::exception_ptr` for error handling. 
`<utility>` gives `std::move` and `std::forward` and also forward delcarations, i used this to prevent circular dependency while allowing type usage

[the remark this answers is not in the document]
isidorostsa: Ah I apologize, I meant the forward declarations declared in the comment

libs/core/executors/include/hpx/executors/experimental/parallel_scheduler.hpp
```diff
@@ -0,0 +1,307 @@
...
+
+            operation_state(
+                Receiver&& r, parallel_scheduler::wrapped_type&& sched)
+              : receiver_(std::forward<Receiver>(r))
+              , scheduler_(std::move(sched))
+            {
+            }
+
+            friend void tag_invoke(start_t, operation_state& op) noexcept
+            {
```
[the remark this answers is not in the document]
charan-003: true, but we were using `tag_invoke` in `thread_pool_scheduler.hpp` as well so i used it here. 

this framework relies on `tag_invoke>` for operations like `schedule`. if this is not how we want it, i can update this such that it is a member function.

libs/core/executors/include/hpx/executors/experimental/parallel_scheduler.hpp
```diff
@@ -0,0 +1,307 @@
...
+    struct get_env_t;
+    struct then_t;
+    struct sender_t;
+}    // namespace hpx::execution::experimental
+
+namespace hpx::execution::experimental {
+
+#ifndef HPX_EXECUTION_EXPERIMENTAL_SENDER_T
+#define HPX_EXECUTION_EXPERIMENTAL_SENDER_T
+    struct sender_t
```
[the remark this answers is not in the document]
charan-003: the `#ifndef HPX_EXECUTION_EXPERIMENTAL_SENDER_T` prevents redefinition of sender_t, which marks sender types. it's a custom guard I defined to ensure `sender_t` is only defined once, keeping dependencies minimal

libs/core/executors/include/hpx/executors/experimental/parallel_scheduler.hpp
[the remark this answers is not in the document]
charan-003: oh, it's not that necessary. just a declaration

libs/core/executors/tests/unit/parallel_scheduler_test.cpp
```diff
@@ -0,0 +1,182 @@
...
+    {
+        std::cout << "set_value called" << std::endl;
+        r.state_->completed = true;
+        r.done_promise->set_value();
+    }
+
+    friend void tag_invoke(
+        ex::set_error_t, test_receiver&& r, std::exception_ptr ep) noexcept
+    {
+        (void) ep;
```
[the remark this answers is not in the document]
charan-003: this ensures errors (via `std::exception_ptr`) are passed to the receiver when a task fails. `(void) ep`; avoids an unused parameter warning since we mark the errors.

libs/core/executors/include/hpx/executors/experimental/parallel_scheduler.hpp
```diff
@@ -0,0 +1,307 @@
...
+    {
+    };
+#endif
+
+    template <typename Scheduler>
+    struct parallel_sender;
```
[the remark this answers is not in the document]
charan-003: umm, i'm keeping sender types separate for flexibility as P2079 is still being updated and maybe reused across scheduler types.

libs/core/executors/include/hpx/executors/experimental/parallel_scheduler.hpp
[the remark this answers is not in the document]
charan-003: @isidorostsa just want to confirm if code  to rely on STDEXEC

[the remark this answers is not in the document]
isidorostsa: Yes it should. Eventually this will be replaced with the std native implementation

libs/core/executors/tests/unit/parallel_scheduler_test.cpp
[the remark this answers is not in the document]
hkaiser: Just FWIW, instead of suppressing warnings by writing `(void) ep;` we usually use `[[maybe_unsed]] std::exception_ptr ep`

charan-003:
@hkaiser @isidorostsa this version is using stdexec

I'm not sure of the failing test case, any hint on the failing cases ?

charan-003:
@hkaiser @isidorostsa 
Does the current implementation, now relying on `stdexec`, meet our expectations? If so, I’ll proceed with implementing bulk support. Please confirm.

libs/core/executors/include/hpx/executors/guided_pool_executor.hpp
```diff
@@ -378,18 +378,15 @@ namespace hpx::execution::experimental {
...
-                HPX_FORWARD(Future, predecessor), HPX_FORWARD(Ts, ts)...);
+            return predecessor.then([f = HPX_FORWARD(F, f), exec,
+                                        ts = hpx::make_tuple(
+                                            HPX_FORWARD(Ts, ts)...)](
+                                        Future&& predecessor) mutable {
+                detail::pre_execution_then_domain_schedule<guided_pool_executor,
+                    pool_numa_hint<Tag>>
+                    pre_exec(exec, exec.hint_, exec.hp_sync_);
+                return pre_exec(HPX_MOVE(f), HPX_FORWARD(Future, predecessor));
+            });
```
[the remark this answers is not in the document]
charan-003: using `then` better reflects this, as it chains the continuation directly to `predecessor`. also,STDEXEC uses a sender/receiver model where `then` is the standard way to chain operations.
also i was having build errors,  
build errors in guided_pool_test.cpp and async_customization.cpp for future types like `hpx::future<double>, hpx::shared_future<int>`, and tuple futures. the issue was that `then` unwraps the future and passes its value, but ` pre_execution_then_domain_schedule` expected a future. I tried to fix this by wrapping the unwrapped value back into a future using `hpx::make_ready_future`

libs/core/executors/include/hpx/executors/experimental/parallel_scheduler.hpp
```diff
@@ -0,0 +1,381 @@
...
+
+    inline parallel_scheduler get_parallel_scheduler()
+    {
+        // Create a thread-local instance with the appropriate policy
+        thread_local bool is_on_hpx_thread = false;
+        thread_local bool initialized = false;
+
+        if (!initialized)
+        {
+            try
```
isidorostsa, line 83: Is there a way to do this without resorting to exceptions? In general we try to avoid using exceptions for flow control.

libs/core/executors/include/hpx/executors/guided_pool_executor.hpp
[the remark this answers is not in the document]
isidorostsa: That's interesting, did this pop up in your work? Is it related to the parallel scheduler?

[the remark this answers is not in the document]
charan-003: @isidorostsa Yes, the build errors in guided_pool_test.cpp and async_customization.cpp popped up as part of my work. 
the errors occurred because replacing dataflow with `then` in `guided_pool_executor.hpp` changed how the `predecessor` future is handled: `then` unwraps the future and passes its value

STDEXEC’s sender/receiver model relies heavily on then for chaining operations, so adopting then in `guided_pool_executor.hpp` ensures consistency across all  the executors module

libs/core/executors/include/hpx/executors/experimental/parallel_scheduler.hpp
charan-003: ohh okay, will try to update parallel_scheduler.hpp accordingly

libs/core/executors/include/hpx/executors/guided_pool_executor.hpp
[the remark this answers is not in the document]
isidorostsa: Wait, sorry I am a bit confused. What is the initial problem you wanted to solve by replacing `dataflow` with `then`

libs/core/executors/include/hpx/executors/experimental/parallel_scheduler.hpp
charan-003: > Is there a way to do this without resorting to exceptions? In general we try to avoid using exceptions for flow control.

@isidorostsa so are we talking about this specific code block or in general ?

libs/core/executors/include/hpx/executors/guided_pool_executor.hpp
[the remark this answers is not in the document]
isidorostsa: I see, thanks for the explanation. How did the rest of this PR trigger this issue?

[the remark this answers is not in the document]
charan-003: `STDEXEC` changes in the PR led to these issues. The PR added `STDEXEC` support in `parallel_scheduler.hpp`, which uses then for execution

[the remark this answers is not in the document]
isidorostsa: When saying `STDEXEC` changes, do you mean changes related to senders and receivers or something more specific?

What is the relation between parallel_scheduler and guided_pool_executor?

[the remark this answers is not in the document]
charan-003: just saw this, to clarify, the STDEXEC changes refer to adopting its sender/receiver model, specifically using `then` for chaining operations, as I’ve done in parallel_scheduler.hpp to align with STDEXEC’s patterns. `parallel_scheduler.hpp` and `guided_pool_executor.hpp` are both in HPX’s executors module but aren’t directly related to parallel_scheduler uses STDEXEC for parallel task scheduling, while guided_pool_executor manages task execution with NUMA hints. I updated guided_pool_executor to use `then` for consistency with these patterns.

This change caused build errors in `guided_pool_test.cpp` and `async_customization.cpp`, as `then` unwrapped futures like `hpx::future<double>`, but `pre_execution_then_domain_schedule` expected a `future`. I fixed this with hpx::make_ready_future. Additionally, STDEXEC’s `__well_formed_sender` errors rised due to lambda return types; I resolved this by ensuring proper completion signatures, using `hpx::util::invoke` for type deduction, and adding overloads for various `future` types.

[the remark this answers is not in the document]
isidorostsa: So if I understand correctly you changed `guided_pool_executor.hpp` so it aligns better with the sender receiver model, even though there was not a specific problem with it? 

If so, I believe it is better to revert the changes, as this does not seem enough of a reason to add complexity to this PR.

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -68,11 +70,24 @@ namespace hpx::execution::experimental::detail {
         return static_cast<std::uint32_t>(chunk_size);
     }
 
-    template <std::size_t... Is, typename F, typename T, typename Ts>
+    // Helper for invoking f with either range-based (begin, end
+    // or single-index arguments
+    template <std::size_t... Is, typename F, typename T, typename... Ts>
     constexpr void bulk_scheduler_invoke_helper(
-        hpx::util::index_pack<Is...>, F&& f, T&& t, Ts& ts)
```
isidorostsa, line 66: I'm not sure I understand the rationale to ditch hpx::util::index_pack

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -68,11 +70,24 @@ namespace hpx::execution::experimental::detail {
...
+        std::index_sequence<Is...>, F&& f, T&& t, bool is_range, Ts&&... ts)
     {
-        HPX_INVOKE(HPX_FORWARD(F, f), HPX_FORWARD(T, t), hpx::get<Is>(ts)...);
+        if (is_range)
```
isidorostsa: Like discussed this should be a constexpr check, through an [NTTP](https://en.cppreference.com/w/cpp/language/template_parameters.html#Notes).

libs/core/executors/include/hpx/executors/experimental/parallel_scheduler.hpp
isidorostsa, line 1: If I remember correctly, this file will be removed, and the parallel scheduler will be just an alias for thread_pool_scheduler, right? 

(just asking for confirmation because I am not 100% sure I remember correctly)

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -113,25 +130,273 @@ namespace hpx::execution::experimental::detail {
...
+        F f;
+
+        template <typename... Ts>
+        void operator()(
+            std::uint32_t begin, std::uint32_t end, Ts&&... ts) const
+        {
+            for (std::uint32_t i = begin; i < end; ++i)
+            {
+                auto index_pack = std::index_sequence_for<Ts...>{};
+                bulk_scheduler_invoke_helper(
```
isidorostsa: This could look like 

```C++
bulk_scheduler_invoke_helper<false> (index_pack, f, i, std::forward<Ts>(ts)...);
```

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -113,25 +130,273 @@ namespace hpx::execution::experimental::detail {
...
+                return;
+            }
+
+            std::cout << "Executing range: begin=" << begin << ", end=" << end
+                      << " on thread " << std::this_thread::get_id()
+                      << std::endl;
+
+            // Dispatch based on f's invocability
+            if constexpr (std::is_invocable_v<F, std::uint32_t, std::uint32_t,
+                              std::add_lvalue_reference_t<Ts>...>)
```
isidorostsa: This is the right approach, although I suspect you need either std::size_t instead of uint32_t

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -113,25 +130,273 @@ namespace hpx::execution::experimental::detail {
...
+            )>
+        // clang-format on
+        friend void tag_invoke(hpx::execution::experimental::set_value_t,
+            bulk_receiver&& r, Ts&&... ts) noexcept
+        {
+            hpx::detail::try_catch_exception_ptr(
+                [&]() {
+                    // Delegate to execute with the full range
+                    auto const size = static_cast<std::uint32_t>(
+                        hpx::util::size(r.op_state->shape));
```
isidorostsa: I'm not sure I understand this cast

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -113,25 +130,273 @@ namespace hpx::execution::experimental::detail {
...
+
+        // clang-format off
+        template <typename... Ts,
+            HPX_CONCEPT_REQUIRES_(
+                hpx::is_invocable_v<F, std::uint32_t, std::uint32_t,
+                    std::add_lvalue_reference_t<Ts>...> ||
+                hpx::is_invocable_v<F, range_value_type,
+                    std::add_lvalue_reference_t<Ts>...>
+            )>
+        // clang-format on
```
isidorostsa: I think you would be better off spliting this into two functions, using concepts like you did here.

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -774,6 +751,7 @@ namespace hpx::execution::experimental::detail {
                     bulk_receiver<operation_state, F, Shape>>;
 
             thread_pool_policy_scheduler<Policy> scheduler;
+            bool is_unchunked = false;
```
isidorostsa: likewise, this should be a template parameter

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -816,11 +796,28 @@ namespace hpx::execution::experimental::detail {
             {
                 tasks_remaining.data_.store(
                     num_worker_threads, std::memory_order_relaxed);
-                HPX_ASSERT(hpx::threads::count(pu_mask) == num_worker_threads);
+                HPX_ASSERT(hpx::threads::count(pu_mask) >= num_worker_threads);
```
isidorostsa: What is the reason for this change?

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -830,24 +827,30 @@ namespace hpx::execution::experimental::detail {
         friend auto tag_invoke(
             connect_t, thread_pool_bulk_sender&& s, Receiver&& receiver)
         {
+            std::cout << "Connecting bulk sender (move)" << std::endl;
             return operation_state<std::decay_t<Receiver>>{
                 HPX_MOVE(s.scheduler), HPX_MOVE(s.sender), HPX_MOVE(s.shape),
                 HPX_MOVE(s.f), HPX_MOVE(s.pu_mask),
-                HPX_FORWARD(Receiver, receiver)};
+                HPX_FORWARD(Receiver, receiver), s.is_unchunked};
```
isidorostsa: This should be a template parameter I believe

isidorostsa (changes requested):
Great work! Some changes needed, but I am very happy to see you getting the hang of things!

hkaiser:
@charan-003 please rebase your branch onto top of master to resolve the merge conflicts.

hkaiser:
Just FYI: https://wg21.link/p4031

charan-003:
@hkaiser @isidorostsa 
It's ready for review now. It includes all the updates related to parallel_scheduler

hkaiser:
@charan-003 Could you please resolve the conflicts. I will review the PR once you've done that.

charan-003:
@hkaiser a signature mismatch for bulk_t: HPX uses `(sender, shape, f)`, while stdexec (P3481R5) expects `(sender, policy, shape, f)`
should we provide an overload for HPX's bulk that accepts an optional policy

libs/core/async_mpi/include/hpx/async_mpi/transform_mpi.hpp
```diff
@@ -189,7 +189,7 @@ namespace hpx::mpi::experimental {
             friend auto tag_invoke(
                 hpx::execution::experimental::get_completion_signatures_t,
                 transform_mpi_sender const&, Env const&)
-            ->  hpx::execution::experimental::transform_completion_signatures_of<
+            ->  stdexec::__transform_completion_signatures_of_t<
```
hkaiser: This is not the right way of fixing this as it depends on internals of stdexec. Let's apply a correct fix to our forwarding header instead (see also #7252 for a potential different way of addressing this).

hkaiser:
Inspect is not happy again:
```
/libs/core/executors/include/hpx/executors/parallel_scheduler.hpp: *I* missing #include (utility) for symbol std::move on line 170, *Line length limit*: 525
/libs/core/executors/include/hpx/executors/parallel_scheduler_backend.hpp: *I* missing #include (cstdint) for symbol std::uint32_t on line 185
    Non-ASCII: line 114
/libs/core/executors/tests/unit/parallel_scheduler.cpp: *I* missing #include (memory) for symbol std::make_shared on line 719
```

libs/core/algorithms/tests/performance/foreach_report.cpp
```diff
@@ -82,6 +82,17 @@ int hpx_main(hpx::program_options::variables_map& vm)
                 [&]() { measure_parallel_foreach(data_representation, exec); });
         }
 
+#if defined(HPX_HAVE_STDEXEC)
```
hkaiser: `HPX_HAVE_STDEXEC` is always defined.

libs/core/executors/include/hpx/executors/parallel_scheduler.hpp
```diff
@@ -0,0 +1,827 @@
...
+#include <memory>
+#include <tuple>
+#include <type_traits>
+#include <variant>
+
+namespace hpx::execution::experimental {
+
+#if defined(HPX_HAVE_STDEXEC)
+    // Forward declaration for parallel_scheduler_domain
+    class parallel_scheduler;
```
hkaiser: Please mark all symbols that should be visible outside of HPX with `HPX_CXX_CORE_EXPORT` to ensure those are accessible through C++ module BMIs.

libs/core/executors/include/hpx/executors/scheduler_executor.hpp
```diff
@@ -26,6 +29,97 @@
...
+            static auto const& policy(parallel_scheduler const& sched)
+            {
+                return sched.get_underlying_scheduler()->policy();
+            }
+            static hpx::threads::mask_type pu_mask(
+                parallel_scheduler const& sched)
+            {
+                return *sched.get_pu_mask();
+            }
+        };
```
hkaiser, line 82: Why do you need a helper to extract the parameters if those are already exposed through the corresponding customization points?

libs/core/executors/include/hpx/executors/scheduler_executor.hpp
```diff
@@ -179,17 +273,74 @@ namespace hpx::execution::experimental {
...
+                    auto* pool = params_type::pool(exec.sched_);
+                    auto first_core = params_type::first_core(exec.sched_);
+                    auto num_cores = params_type::num_cores(exec.sched_);
+                    auto const& policy = params_type::policy(exec.sched_);
+                    auto mask = params_type::pu_mask(exec.sched_);
+
+                    return hpx::parallel::execution::detail::
+                        index_queue_bulk_async_execute(pool, first_core,
+                            num_cores, policy, HPX_FORWARD(F, f), shape, mask,
+                            HPX_FORWARD(Ts, ts)...);
```
hkaiser: This snippet could be extracted into a new function, which should simplify a lot of the code below.

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -106,18 +140,38 @@ namespace hpx::execution::experimental {
                 hpx::execution::experimental::stdexec_internal::__sender_for<
```
hkaiser: Please move all stdexec specific to the forwarding header to simplify future use of standard library implementations.

hkaiser:
@charan-003 compilation is still failing (so does the formatting check)... Would you be able to investigate, please?

libs/core/algorithms/include/hpx/parallel/util/foreach_partitioner.hpp
```diff
@@ -73,8 +73,7 @@ namespace hpx::parallel::util::detail {
 
             // We attempt to perform some optimizations in case of non-task
             // execution.
-            if constexpr (!hpx::is_async_execution_policy_v<ExPolicy> &&
-                !hpx::execution_policy_has_scheduler_executor_v<ExPolicy>)
+            if constexpr (!hpx::is_async_execution_policy_v<ExPolicy>)
```
hkaiser, line 76: There is similar code in the prtinioner, should this be changed as well: https://github.com/TheHPXProject/hpx/blob/70af3ac902795e605c57043a9435c955029b55dc/libs/core/algorithms/include/hpx/parallel/util/partitioner.hpp#L77-L79?

libs/core/execution/include/hpx/execution/algorithms/as_sender.hpp
```diff
@@ -262,3 +263,18 @@ namespace hpx::execution::experimental {
...
+    template <typename T, typename Env>
+    constexpr bool sends_stopped<
+        hpx::execution::experimental::detail::as_sender_sender<hpx::future<T>>,
+        Env> = false;
+
+    template <typename T, typename Env>
+    constexpr bool sends_stopped<hpx::execution::experimental::detail::
+                                     as_sender_sender<hpx::shared_future<T>>,
+        Env> = false;
+}    // namespace stdexec
```
hkaiser: Couldn't you do this customization inside the `as_sender_sender` definition?

libs/core/algorithms/include/hpx/parallel/algorithms/for_each_index.hpp
```diff
@@ -387,11 +387,30 @@ namespace hpx::parallel::detail {
...
+                            }
+                            else
+                            {
+                                return util::detail::
+                                    algorithm_result<ExPolicy>::get(
+                                        util::partitioner<ExPolicy>::call(
+                                            HPX_FORWARD(ExPolicy, policy),
+                                            first, count, HPX_MOVE(iter_fun),
+                                            hpx::util::empty_function{}));
+                            }
```
hkaiser: This pattern is repeated more than once in other places. Can you factor that out into a separate function (in the partition.hpp file, perhaps)?

cmake/HPX_SetupStdexec.cmake
```diff
@@ -83,7 +83,3 @@ else()
     )
   endif()
 endif()
-
-# stdexec is now unconditionally required; define HPX_HAVE_STDEXEC so that
-# downstream code using #if defined(HPX_HAVE_STDEXEC) continues to work.
-hpx_add_config_define(HPX_HAVE_STDEXEC)
```
hkaiser, line 89: Should we keep this around for the time being?

libs/core/algorithms/include/hpx/parallel/util/partitioner.hpp
```diff
@@ -699,5 +699,26 @@ namespace hpx::parallel::util {
             detail::static_partitioner,
             detail::task_static_partitioner>::template apply<R, Result>
     {
+        // Helper to call partitioner and wrap the result with
+        // algorithm_result::get(). Handles both void and non-void return types.
+        template <typename ExPolicy_, typename... Args>
+        static decltype(auto) call_with_algorithm_result(
```
hkaiser: This function cleans up things nicely. Does it need to be a member of the partitioner, however?

libs/core/execution/include/hpx/execution/algorithms/as_sender.hpp
```diff
@@ -18,222 +18,242 @@
...
+    struct as_sender_sender<hpx::future<T>>
+      : public as_sender_sender_base<hpx::future<T>>
+    {
+        using sender_concept = hpx::execution::experimental::sender_t;
+        using future_type = hpx::future<T>;
+        using base_type = as_sender_sender_base<hpx::future<T>>;
+        using base_type::future_;
+
+        template <typename Future,
+            typename = std::enable_if_t<
```
hkaiser: Please use `requires()` wherever possible instead of SFINAE.

libs/core/execution/include/hpx/execution/algorithms/as_sender.hpp
```diff
@@ -18,222 +18,242 @@
...
+
+// stdexec customization for sends_stopped for hpx::shared_future-based sender
+// Explicit customization to ensure as_sender_sender returns false since
+// the operation state never calls set_stopped()
+namespace stdexec {
+    template <typename T, typename Env>
+    constexpr bool sends_stopped<hpx::execution::experimental::detail::
+                                     as_sender_sender<hpx::shared_future<T>>,
+        Env> = false;
+}    // namespace stdexec
```
hkaiser: Couldn't these explicit customizations be achieved through removing `set_stopped` from the completion signatures exposed by `as_sender_sender<hpx::shared_future<T>`? I'm asking as I would like to avoid to explicitly refer to `stdexec` in our code base (except for the forwarding header).

libs/core/executors/include/hpx/executors/parallel_scheduler.hpp
```diff
@@ -0,0 +1,884 @@
...
+    // Stores a shared_ptr<parallel_scheduler_backend> for replaceability.
+    // The default backend wraps HPX's thread_pool_policy_scheduler.
+    HPX_CXX_CORE_EXPORT class parallel_scheduler
+    {
+    public:
+        parallel_scheduler() = delete;
+
+        // P2079R10: Construct from a backend shared_ptr.
+        // This is the primary constructor used by get_parallel_scheduler().
+        explicit parallel_scheduler(
```
hkaiser: Does P2079 require for this to be public? Couldn't `get_parallel_scheduler()` be a friend instead?

libs/core/executors/include/hpx/executors/parallel_scheduler.hpp
```diff
@@ -0,0 +1,884 @@
...
+            parallel_scheduler const&) noexcept = default;
+        parallel_scheduler& operator=(parallel_scheduler&&) noexcept = default;
+
+        // P2079R10 6.4: two schedulers compare equal iff BACKEND-OF(lhs)
+        // and BACKEND-OF(rhs) refer to the same object, i.e., their
+        // shared_ptr targets are identical.  Pointer equality is the only
+        // comparison mandated by the standard; equal_to() on the backend
+        // interface is an HPX-specific extension that custom backends may
+        // implement for their own purposes but is not used here.
+        friend bool operator==(parallel_scheduler const& lhs,
```
hkaiser, line 600: Do we need `operator!=()` as well?

libs/core/executors/include/hpx/executors/parallel_scheduler.hpp
```diff
@@ -0,0 +1,884 @@
...
+        {
+            return backend_ ? backend_->get_pu_mask() : nullptr;
+        }
+
+    private:
+        std::shared_ptr<parallel_scheduler_backend> backend_;
+    };
+
+    // Stream output operator for parallel_scheduler
+    HPX_CXX_CORE_EXPORT inline std::ostream& operator<<(
```
hkaiser: Please move the implementation to a separate compilation unit.

libs/core/executors/include/hpx/executors/parallel_scheduler.hpp
```diff
@@ -0,0 +1,884 @@
...
+
+    // P2079R10 get_parallel_scheduler function.
+    // Uses query_parallel_scheduler_backend() to obtain the backend,
+    // which can be replaced via set_parallel_scheduler_backend_factory().
+    HPX_CXX_CORE_EXPORT inline parallel_scheduler get_parallel_scheduler()
+    {
+        auto backend = query_parallel_scheduler_backend();
+        if (!backend)
+        {
+            std::
```
hkaiser: Could the implementation of this function be moved to a separate translation unit?

libs/core/executors/include/hpx/executors/parallel_scheduler_backend.hpp
```diff
@@ -0,0 +1,514 @@
...
+            return nullptr;
+        }
+    };
+
+    namespace detail {
+
+        // Default HPX backend: wraps the existing thread_pool_policy_scheduler.
+        // This is the backend returned by query_parallel_scheduler_backend()
+        // unless the user provides a replacement via weak linking.
+        HPX_CXX_CORE_EXPORT class hpx_parallel_scheduler_backend final
```
hkaiser: All member function implementations for this type can go to a separate translation unit.

libs/core/executors/include/hpx/executors/parallel_scheduler_backend.hpp
```diff
@@ -0,0 +1,514 @@
...
+    // by providing their own shared_ptr<parallel_scheduler_backend>.
+    //
+    // Note: Unlike stdexec's approach, HPX uses a function
+    // pointer that can be replaced at runtime via
+    // set_parallel_scheduler_backend_factory(). This avoids platform-specific
+    // weak-linking issues while providing the same replaceability.
+    HPX_CXX_CORE_EXPORT using parallel_scheduler_backend_factory_t =
+        std::shared_ptr<parallel_scheduler_backend> (*)();
+
+    namespace detail {
```
hkaiser: I believe that large parts of the code below can be moved to a separate translation unit as well.

libs/core/algorithms/include/hpx/parallel/util/partitioner.hpp
charan-003: I moved `call_with_algorithm_result` out of the partitioner class into a free function in the same `hpx::parallel::util `namespace so `for_loop` and `for_each_index` can call it directly without going through partitioner internals

charan-003:
@hkaiser, please do let me know if any more changes are required
Thanks

hkaiser (approved):
LGTM, thanks! Let's get this merged!

PR #6659, opened 2025-03-29
fix(reduce): Initialize accumulator with init instead of first elemen…

### LINKED ISSUE #6647: Incorrect reduce implementation
https://github.com/STEllAR-GROUP/hpx/blob/0005b929c69fd199c11dfd691f4d1b62d90735af/libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp#L421

This line in `reduce` implementation attempts to convert `*first`  to `T`, but there is no requirement for that operation. See https://eel.is/c++draft/reduce#5

As a somewhat artificial example, it's possible to implement `minmax` on top of `reduce`:

```
struct minmax
{
    std::pair<int, int> operator()(
        std::pair<int, int> lhs, std::pair<int, int> rhs) const
    {
        return {
            lhs.first < rhs.first ? lhs.first : rhs.first,
            lhs.second < rhs.second ? rhs.second : lhs.second,
        };
    }
    std::pair<int, int> operator()(std::pair<int, int> lhs, int rhs) const
    {
        return (*this)(lhs, std::pair<int, int>{rhs, rhs});
    }
    std::pair<int, int> operator()(int lhs, std::pair<int, int> rhs) const
    {
        return (*this)(std::pair<int, int>{lhs, lhs}, rhs);
    }
    std::pair<int, int> operator()(int lhs, int rhs) const
    {
        return (*this)(std::pair<int, int>{lhs, lhs}, std::pair<int, int>{rhs, rhs});
    }
};

auto [min, max] = hpx::reduce(hpx::execution::par,
    c.begin(), c.end(), std::pair<int, int>{INT_MAX, INT_MIN}, minmax{});
```

harith-hacky03:
…t #6647

Fixes #6647

## Proposed Changes

  -
  -
  -

## Checklist

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
@@ -418,7 +418,13 @@ namespace hpx::parallel {
                 }
 
                 auto f1 = [r](FwdIterB part_begin, std::size_t part_size) -> T {
-                    T val = *part_begin;
+                    T val = init;
+                    auto it = part_begin;
+                    while (it != part_end)
+                    {
+                       val = HPX_INVOKE(binary_op, val, *it);
+                       ++it;
+                    }
                     return detail::sequential_reduce<ExPolicy>(
                         ++part_begin, --part_size, HPX_MOVE(val), r);
                 };


hkaiser:
@harith-hacky03 why did you close the pull-request? Is this obsolete?

PR #6664, opened 2025-04-01
Fix CMake template when fetching Boost

Pansysk75:
`HPX_CONFIG_IS_INSTALL` needs to be replaced when this file is configured:

https://github.com/STEllAR-GROUP/hpx/blob/735086d7122f45e88a8d46c73438fbf54d8b7572/cmake/HPX_GeneratePackage.cmake#L63-L93


hkaiser (approved):
LGTM, thanks!

PR #6670, opened 2025-04-08
Add base template template param to execution_policy

K-ballo:
## Proposed Changes

Add a template template param to `execution_policy` to specify a base class to inherit from in a way that's compatible with policy rebinding.

This is required to inject a `thrust::execution_policy` base into HPX execution policies.

## Checklist

### DIFF
--- a/libs/core/executors/tests/unit/execution_policy_base.cpp
+++ b/libs/core/executors/tests/unit/execution_policy_base.cpp
@@ -1,0 +1,67 @@
+//  Copyright (c) 2025 Agustin Berge
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/datapar.hpp>
+#include <hpx/execution.hpp>
+#include <hpx/init.hpp>
+#include <hpx/modules/testing.hpp>
+#include <hpx/program_options.hpp>
+
+#include <type_traits>
+
+template <typename Derived>
+struct custom_base
+{
+};
+
+///////////////////////////////////////////////////////////////////////////////
+template <typename ExPolicy>
+void test_base(ExPolicy&& policy)
+{
+    using policy_t = std::decay_t<ExPolicy>;
+
+    static_assert(std::is_base_of_v<custom_base<policy_t>, policy_t>,
+        "must have custom_base<policy_t> as base class");
+
+    (void) policy;
+}
+
+void test_base()
+{
+    using namespace hpx::execution;
+
+    test_base(basic_sequenced_policy<custom_base>{});
+    test_base(basic_parallel_policy<custom_base>{});
+    test_base(basic_unsequenced_policy<custom_base>{});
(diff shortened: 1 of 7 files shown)


hkaiser (dismissed):
LGTM, thanks!

libs/core/executors/tests/unit/execution_policy_base.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include <type_traits>
+
+template <typename Derived>
+struct custom_base
+{
+};
+
+///////////////////////////////////////////////////////////////////////////////
+template <typename ExPolicy>
+void test_base(ExPolicy&& policy)
```
hkaiser, line 22: The tests will complain about `policy` being unused.

K-ballo:
retest LSU

hkaiser:
@K-ballo something is not quite right with the CI on Rostam. I will investigate this asap.

K-ballo:
retest LSU

hkaiser:
@K-ballo Would it be possible (or even necesary) to add a compatibility layer such that existing external HPX execution policy implementations can be migrated over time? E.g., HPX-Kokkos implements its own policies, see: https://github.com/STEllAR-GROUP/hpx-kokkos/blob/master/src/hpx/kokkos/policy.hpp. Would that still work?

K-ballo:
> @K-ballo Would it be possible (or even necesary) to add a compatibility layer such that existing external HPX execution policy implementations can be migrated over time? E.g., HPX-Kokkos implements its own policies, see: https://github.com/STEllAR-GROUP/hpx-kokkos/blob/master/src/hpx/kokkos/policy.hpp. Would that still work?

No, that wouldn't work, that's the exact same approach that I initially took for thrust policies.

It would be possible to keep compatibility by maintaining two separate execution policy designs. But in that case the "new" design isn't limited by existing code anymore, so it could do things differently.

hkaiser:
> No, that wouldn't work, that's the exact same approach that I initially took for thrust policies.
> 
> It would be possible to keep compatibility by maintaining two separate execution policy designs. But in that case the "new" design isn't limited by existing code anymore, so it could do things differently.

Let's add a `#if HPX_VERSION_FULL < 0x011100` to HPX-Kokkos in that case (once this PR has been merged) to make sure it compiles either way.

K-ballo:
> > No, that wouldn't work, that's the exact same approach that I initially took for thrust policies.
> > It would be possible to keep compatibility by maintaining two separate execution policy designs. But in that case the "new" design isn't limited by existing code anymore, so it could do things differently.
> 
> Let's add a `#if HPX_VERSION_FULL < 0x011100` to HPX-Kokkos in that case (once this PR has been merged) to make sure it compiles either way.

I think we should abandon this PR, go back to the original policy implementation for thrust (exact same as in kokkos, so there's precedence), and then take some time to find an execution policy design that's easily extensible and that would work for all these external projects. Your call.

hkaiser:
> I think we should abandon this PR, go back to the original policy implementation for thrust (exact same as in kokkos, so there's precedence), and then take some time to find an execution policy design that's easily extensible and that would work for all these external projects. Your call.

Ok, let's leave it be, then. I'll close this PR.

PR #6677, opened 2025-04-15
Offer a runtime manager object

K-ballo:
## Proposed Changes

Offer a ready-made solution to manage starting/stopping the HPX runtime. Adapted from the quickstart/init_globally example.

## To-Do

- ~This design swallows error messages: error on `start()` results in termination, error on `stop()`  is ignored. In order to expose errors it's necessary to move away from the constructor/destructor model.~
- ~Add unit test, but how to actually check behavior? Perhaps using `run_as_hpx_thread` and comparing thread ids.~
- ~Move implementation to source file.~
- ~Document this new facility.~

This will also benefit the thrust backend, and potentially hpxMP and others.

## Checklist


libs/full/init_runtime/include/hpx/manage_runtime.hpp
```diff
@@ -0,0 +1,100 @@
...
+namespace hpx {
+
+    class manage_runtime
+    {
+    public:
+        int start(
+            int argc, char** argv, const init_params& init_args = init_params())
+        {
+#if defined(HPX_WINDOWS)
+            detail::init_winsocket();
```
K-ballo, line 26: Does this belong in here?

hkaiser (commented):
Is there anything the `manage_runtime` depends on that requires it to be part of the full module? I think it could be either moved to the `core/init_runtime_local` module or added to the static library built from the `hpx/init` directory.

K-ballo:
> Is there anything the `manage_runtime` depends on that requires it to be part of the full module? I think it could be either moved to the `core/init_runtime_local` module or added to the static library built from the `hpx/init` directory.

It's calling `hpx::start` which is defined in the `init_runtime` library where I put this.

Should it be calling another `start`? There's one in the `init_runtime_local` library, none in the `hpx/init` directory.

hkaiser:
> > Is there anything the `manage_runtime` depends on that requires it to be part of the full module? I think it could be either moved to the `core/init_runtime_local` module or added to the static library built from the `hpx/init` directory.
> 
> It's calling `hpx::start` which is defined in the `init_runtime` library where I put this.
> 
> Should it be calling another `start`? There's one in the `init_runtime_local` library, none in the `hpx/init` directory.

Not that we currently do this, but if somebody wants to use the local runtime only, the runtime manager wouldn't be available.

K-ballo:
> > > Is there anything the `manage_runtime` depends on that requires it to be part of the full module? I think it could be either moved to the `core/init_runtime_local` module or added to the static library built from the `hpx/init` directory.
> > 
> > 
> > It's calling `hpx::start` which is defined in the `init_runtime` library where I put this.
> > Should it be calling another `start`? There's one in the `init_runtime_local` library, none in the `hpx/init` directory.
> 
> Not that we currently do this, but if somebody wants to use the local runtime only, the runtime manager wouldn't be available.

Does that mean we need both a `manage_local_runtime` and a `manage_runtime`?

hkaiser:
> Does that mean we need both a `manage_local_runtime` and a `manage_runtime`?

I don't think so. As said, the manage_runtime can go into `core/runtime_local_init` or the static library built from the `hpx/init` directory. The latter may even be preferred as then the manage_runtime would be linked to an application only if explicitly requested.

hkaiser (approved):
LGTM, thanks!

PR #6678, opened 2025-04-15
Fix oneapi overloads

G-071:
### Background
Our HPX-SYCL was originally written to support both hipSYCL and oneAPI. It allows passing a SYCL executor (containing a SYCL queue), a SYCL queue member function, and its parameters to hpx::async. The correct overload of the queue member function would be determined based on the passed parameters, the member function would then be launched asynchronously on the executor queue, and we would get an HPX future back.

However, oneAPI is using an additional (afaik non-standard) ```code_location``` parameter for many of its SYCL queue member functions. Usually, a SYCL user/developer would not notice this because the ```code_location``` has a default value. In our case, however, it is problematic as it messes with choosing the correct  function overloads when passing the calls to ```hpx::apply``` or ```hpx::async```.

The [original SYCL PR](https://github.com/STEllAR-GROUP/hpx/pull/6085) that added SYCL to HPX addressed this with additional overloads for ```hpx::async``` and ```hpx::apply``` when using oneAPI. These added the code location parameter automatically to find the correct overload.

### The actual issue
However, since then, something about the code location parameter appears to have changed: When compiling the sycl_stream.cpp example with HPX, I now encounter a few errors, such as:
```
/home/devel/hpx/libs/core/async_sycl/tests/unit/sycl_stream.cpp:53:5: error: no matching function for call to 'apply'
   53 |     hpx::apply(
```
or, alternatively, for other lines, errors about ambiguous calls. I am not quite sure when this started breaking, but at least with icpx 2025.0.0 I encountered this issue.

### Fix and next steps
This PR fixes all these overload issues by modifying the overloads of hpx::apply/async for SYCL executors. Depending on the type of the passed arguments I am determining whether to use the overload that adds the code location parameter or not.

With this, passing queue methods works again with oneAPI, at the very least for the most important ones that I tried in the tests (submit, parallel_for, memset, copy, single_task, ...). Optional parameters such as SYCL event dependencies should also work.

In the future, I would like to expand the test suite for this, actually. There aren't this many SYCL queue member functions left, so we might just have a test for each of them. If we add the SYCL tests to the CI this would also warn us with something regarding the overloads breaks again.
However, all of that would require some extra work. Hence, I would put this on the roadmap for after the HPX 1.11.0 release (would be nice for 2.0.0 though). For now, I would be glad to just get this current version of the bugfix in before adding more stuff!

### Additional comments:
- This is PR 1 / 3 for the SYCL changes I want to add before 1.11.0. I split them up as they address separate issues.
- At one point I had to use std::forward instead of HPX_FORWARD as HPX_FORWARD somehow gave me compilation problems for this part: 
```cpp
std::forward<hpx::sycl::experimental::sycl_executor::
                            queue_function_code_loc_ptr_t<Ts..., T>>(f),
                std::forward<Ts>(ts)...,
```

### DIFF
--- a/libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
+++ b/libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
@@ -25,6 +25,7 @@
 #include <memory>
 #include <type_traits>
 #include <utility>
+#include <vector>
 
 namespace hpx::sycl::experimental {
     namespace detail {
@@ -377,8 +378,6 @@ namespace hpx {
                 f),
             HPX_FORWARD(Ts, ts)...);
     }
-
-    // TODO(daissgr) Maybe make the next two specialized for the submit method (mostly by requiring a different function point?
 
     template <typename Executor, typename... Ts, typename T,
         std::enable_if_t<(std::conjunction_v<sycl::experimental::detail::
--- a/libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
+++ b/libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
@@ -133,41 +133,41 @@ namespace hpx::sycl::experimental {
         // using a default argument (code_location::current())
 
         /// sycl::queue::member_function type with code_location parameter
-        template <typename Param>
+        template <typename... Params>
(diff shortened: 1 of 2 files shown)


hkaiser (approved):
LGTM, thanks!

PR #6679, opened 2025-04-15
Fix sycl deprecations

G-071:
When working on #6678 I noticed that we are getting quite a few SYCL-related deprecation warnings when using new OneAPI releases (2025.0.0).

These can be fixed by replacing the CL/sycl.hpp header with the sycl/sycl.hpp one. This also required moving from the cl::sycl namespace to the sycl:: namespace.
Additionally we have to use the new SYCL device selectors.

This PR addresses all these deprecations by moving to the correct headers, changing the namespace and adding the modern device selectors instead of using the SYCL 1.2.1 ones.
All of this was rather simple, mostly requiring a few search-and-replace operations. However, it caused quite a few changes so I decided to make this a separate PR instead of bundling it with #6678. Otherwise the diff was getting a bit chaotic!

### Note:
- This PR depends on #6678 and should only be merged afterwards!
- This is PR 2/3 for the SYCL changes that I wanted to get into HPX 1.11.0


hkaiser:
@G-071 Could you look into the clang-format issues, please

G-071:
Ah, I forgot to run clang-format again after the search and replace operations on those files!

They are formatted now!

hkaiser:
> Ah, I forgot to run clang-format again after the search and replace operations on those files!
> 
> They are formatted now!

Thanks Gregor. Unfortunately our linters (very annoying, I know) complains some more:

/libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp: `*I*` missing `#include (vector)` for symbol `std::vector` on line [38](https://github.com/STEllAR-GROUP/hpx/blob/82df88d5f634929a95b2863aadd079a83a824445/libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp#L38), `*Line length limit*`: [381](https://github.com/STEllAR-GROUP/hpx/blob/82df88d5f634929a95b2863aadd079a83a824445/libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp#L381)

hkaiser:
retest lsu

G-071:
Not a problem at all! I just fixed it for this PR and the #6678 PR (as the inspect error originated from that one).

I really need to get around and add the linter to my laptop's neovim setup though. It's a bit of a reoccurring issue.

PR #6680, opened 2025-04-15
Add oneapi device init workaround

G-071:
### The issue
When running Octo-Tiger with HPX on Aurora (and its Intel GPUs)  I encountered a segfault within the device runtime.

These actually seemed familiar: I encountered a similar error when trying to originally run Octo-Tiger on a single Intel GPU Max 1100 (see STEllAR-GROUP/octotiger#486).

From what I can tell, these segfaults appear to be an issue with the initialization on the device side upon first touch. With HPX and SYCL we never wait on any SYCL event (as we rely on event polling) and the initialization seems to be tied to the first wait on a SYCL event (at least that's my best guess right now).
We only encounter these problems when we use OneAPI on an Intel GPUs (same versions of HPX and OneAPI work fine on NVIDIA and AMD GPUs). It seems to be related to their initialization of the Intel devices only.

### Workaround
We can work around this problem by simply executing a dummy kernel on the GPU and wait on its SYCL event right at the beginning of a HPX application. Afterwards, all other GPU operations are fine, even without explicitly waiting on their events.
Weirdly, it seems to not be enough to do this once per application, but instead we need to do this once per compilation unit where we actually execute a SYCL kernel. In our case, the HPX-SYCL integration only does this at one point: When trying to get a future for a given SYCL command queue as we need need run an operation on the queue to get a SYCL event.

To have such an initialization, I simply added a ```enforce_oneapi_device_side_init``` method which calls a dummy SYCL kernel and waits on its event to the sycl_future.cpp file. At least when running Octo-Tiger on Aurora, it avoids this issue (and the segfaults) entirely. 

I am not too happy about this workaround at all as it seems kinda crude. However, so far it at least got things running on Aurora, hence I figured I might as well open this PR until we can find a better solution (or this gets fixed in the GPU runtime).

### Notes
- Depends on #6679 which needs to merged first (this should also clear up the diff). I'll keep this as a draft PR until then.
- This would be the third and last SYCL PR that I wanted to get into 1.11.0 (simply for the Aurora support).


G-071:
With #6679 merged, this should be good to go as well.

hkaiser (approved):
LGTM, thanks!

PR #6682, opened 2025-04-17
Add C++23 std::generator equivalence test and fix missing semicolon

### LINKED ISSUE #6681: build break - missing ';'
This is the problematic hpx version and code location:
[https://github.com/STEllAR-GROUP/hpx/blob/v1.10.0/libs/core/type_support/include/hpx/type_support/generator.hpp#L21](https://github.com/STEllAR-GROUP/hpx/blob/v1.10.0/libs/core/type_support/include/hpx/type_support/generator.hpp#L21)

By adding the missing *;* one issue will still remain: the silent possibility that code lines are getting into HPX which does not even compiles.

So either:
A) a CI job, which sets up the environment in which the following is true:
```
#if defined(HPX_HAVE_CXX20_COROUTINES) &&                                      \
    (!defined(HPX_CLANG_VERSION) || HPX_CLANG_VERSION >= 130000)
#if defined(HPX_HAVE_CXX23_STD_GENERATOR)
```
B) a test case, which somehow tests whether `hpx::generator` is equivalent to `std::generator` in case when the  preprocessor condition described above is true.

Thanks in advance! :)

harith-hacky03:
Fixes #6681

## Proposed Changes

- Added C++23 `std::generator` equivalence test in `libs/core/type_support/tests/unit/generator.cpp`
  - Tests integer sequence generation
  - Tests string sequence generation
  - Compares results between `hpx::generator` and `std::generator`
- Updated `CMakeLists.txt` to add C++23 compilation flags when available

This PR addresses the issue mentioned in #6681 by ensuring that `hpx::generator` behaves correctly when C++23's `std::generator` is available. The test verifies that both implementations produce identical results for both integer and string sequences.

The test only runs when:
```cpp
#if defined(HPX_HAVE_CXX20_COROUTINES) && \
    (!defined(HPX_CLANG_VERSION) || HPX_CLANG_VERSION >= 130000)
#if defined(HPX_HAVE_CXX23_STD_GENERATOR)
```

## Checklist

### DIFF
--- a/libs/core/async_sycl/include/hpx/async_sycl/detail/sycl_event_callback.hpp
+++ b/libs/core/async_sycl/include/hpx/async_sycl/detail/sycl_event_callback.hpp
@@ -14,7 +14,7 @@
 
 #include <string>
 
-#include <CL/sycl.hpp>
+#include <sycl/sycl.hpp>
 
 // Check compiler compatibility:
 // Needs to be done AFTER sycl include for HipSYCL
@@ -58,7 +58,7 @@ namespace hpx { namespace sycl { namespace experimental { namespace detail {
      * done by this method as well
     */
     HPX_CORE_EXPORT void add_event_callback(
-        event_callback_function_type&& f, cl::sycl::event event);
+        event_callback_function_type&& f, ::sycl::event event);
 
     /// Register SYCL event polling function with the scheduler (see scheduler_base.hpp)
     HPX_CORE_EXPORT void register_polling(hpx::threads::thread_pool_base& pool);
--- a/libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
+++ b/libs/core/async_sycl/include/hpx/async_sycl/sycl_executor.hpp
@@ -35,8 +35,8 @@ namespace hpx::sycl::experimental {
         struct is_basic_queue_arg
           : std::integral_constant<bool,
                 std::is_scalar<T>::value ||
-                    std::is_same<T, cl::sycl::event>::value ||
-                    std::is_same<T, const std::vector<cl::sycl::event>&>::value>
+                    std::is_same<T, ::sycl::event>::value ||
+                    std::is_same<T, const std::vector<::sycl::event>&>::value>
         {
         };
     }    // namespace detail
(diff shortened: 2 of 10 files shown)


libs/core/type_support/tests/unit/CMakeLists.txt
```diff
@@ -8,6 +8,9 @@ set(tests relocate relocate_at uninitialized_relocate_n_primitive)
 
 if(HPX_WITH_CXX20_COROUTINES)
   set(tests ${tests} generator)
+  if(HPX_WITH_CXX23_STD_GENERATOR)
+    set(generator_FLAGS ${HPX_CXX23_FLAG})
+  endif()
```
hkaiser, line 13: What's the rationale of this change? The idea of `HPX_WITH_CXX23_STD_GENERATOR` is to be enabled if std::generator is available, for some platforms that may be true even if compiled in C++20 mode? Also, where is `HPX_CXX23_FLAG` defined?

libs/core/type_support/tests/unit/generator.cpp
```diff
@@ -230,6 +230,63 @@ namespace tests {
...
+        std::vector<int> std_result;
+        for (int i : std_gen())
+        {
+            std_result.push_back(i);
+        }
+
+        HPX_TEST(hpx_result == std_result);
+
+        // Test with string sequence
+        auto hpx_str_gen = []() -> hpx::generator<std::string> {
```
hkaiser, line 266: Wouldn't this be 100% equivalent to using `std::generator` in this case?

harith-hacky03: Yes, @hkaiser is correct. In this specific case, using hpx::generator<std::string> is 100% equivalent to using std::generator<std::string> because:
The test is already in a C++23 context (as indicated by HPX_WITH_CXX23_STD_GENERATOR)
The test is comparing hpx_result with std_result
The generator is being used in a simple, non-specialized way:
It's just yielding string values
No HPX-specific features or extensions are being utilized
No distributed computing or parallel processing is involved
No custom allocators or memory management is needed
Since we're already testing the equivalence between hpx::generator and std::generator in this test, and we're in a C++23 context, using std::generator directly would be more straightforward and clearer in expressing the intent of the test.

PR #6686, opened 2025-04-21
lci pp: upgrade LCI autofetch target to 1.7.9

JiakunYan:


JiakunYan:
@hkaiser An additional question: I was trying to simplify the LCI autofetch option by removing all the install/export of the LCI target (Starting from [here](https://github.com/STEllAR-GROUP/hpx/blob/c1c2539adbad927a8302661aba7b5140d460db7c/cmake/HPX_SetupLCI.cmake#L68)), but I found I could not, because all hpx modules link to LCI as a public dependency ([example](https://github.com/STEllAR-GROUP/hpx/blob/c1c2539adbad927a8302661aba7b5140d460db7c/libs/core/lci_base/CMakeLists.txt#L30)). This is unnecessary because LCI could be privately linking to HPX. I kind of expected there should be a way to specify a private dependency in `add_hpx_module`, but I didn't find one. Is this intended?

hkaiser:
> @hkaiser An additional question: I was trying to simplify the LCI autofetch option by removing all the install/export of the LCI target (Starting from [here](https://github.com/STEllAR-GROUP/hpx/blob/c1c2539adbad927a8302661aba7b5140d460db7c/cmake/HPX_SetupLCI.cmake#L68)), but I found I could not, because all hpx modules link to LCI as a public dependency ([example](https://github.com/STEllAR-GROUP/hpx/blob/c1c2539adbad927a8302661aba7b5140d460db7c/libs/core/lci_base/CMakeLists.txt#L30)). This is unnecessary because LCI could be privately linking to HPX. I kind of expected there should be a way to specify a private dependency in `add_hpx_module`, but I didn't find one. Is this intended?

We've never had the need to link modules privately. So we don't have that option, ATM. Please create an issue asking for this feature.

JiakunYan:
@hkaiser Thanks! On second thought, I found no additional benefits of linking LCI privately. LCI is a shared library, so HPX has to install the autofetched one anyway.

I think this PR is ready to be merged.

hkaiser (approved):

PR #6687, opened 2025-04-23
lci pp: change default ndevices=2 and progress_type=worker; improve document

JiakunYan:
Previous default options of LCI parcelport are based on experiments on Rostam, but the newer generation of clusters generally favor `ndevices=2` and `progress_type=worker`. This PR updates their default values and adds them to the document.


hkaiser (approved):
LGTM, thanks!

PR #6694, opened 2025-04-30
Updated adjacent difference to use c++20 concepts

sidhikumari5:
## Proposed Changes

  - uses c++ 20 concepts `requires` clause replacing HPX_CONCEPTS_REQUIRES

## Checklist

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/adjacent_difference.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/adjacent_difference.hpp
@@ -389,11 +389,11 @@ namespace hpx {
     {
         // clang-format off
         private:
-        template <typename FwdIter1, typename FwdIter2,
-             HPX_CONCEPT_REQUIRES_(
+        template <typename FwdIter1, typename FwdIter2>
+            requires (
                 hpx::traits::is_iterator_v<FwdIter1> &&
                 hpx::traits::is_iterator_v<FwdIter2>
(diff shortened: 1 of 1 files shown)


libs/core/algorithms/include/hpx/parallel/algorithms/adjacent_difference.hpp
```diff
@@ -409,12 +409,12 @@ namespace hpx {
...
-        template <typename ExPolicy, typename FwdIter1, typename FwdIter2,
-            HPX_CONCEPT_REQUIRES_(
+        template <typename ExPolicy, typename FwdIter1, typename FwdIter2>
+            requires (
                 hpx::is_execution_policy_v<ExPolicy> &&
                 hpx::traits::is_iterator_v<FwdIter1> &&
                 hpx::traits::is_iterator_v<FwdIter2>
-            )>
+            )
         // clang-format on
```
hkaiser, line 418: We may want to move the cmake-format directive to wrap the requires clauses only:
```suggestion
        template <typename ExPolicy, typename FwdIter1, typename FwdIter2>
            // clang-format off
            requires (
                hpx::is_execution_policy_v<ExPolicy> &&
                hpx::traits::is_iterator_v<FwdIter1> &&
                hpx::traits::is_iterator_v<FwdIter2>
            )
            // clang-format on
```

hkaiser:
This is superseded by #6700

PR #6701, opened 2025-05-11
Attempting to fix shutdown hang on exception_info

### LINKED ISSUE #6699: Catch lower-level runtime error
I am testing HPX with a new version of LCI, which can throw a `runtime_error` instead of calling `abort()` in the case of a fatal error.

However, I found HPX didn't catch the `runtime_error` thrown from LCI correctly. Instead of printing the `runtime_error.what()`, it prints a `<unknown>` and then hangs.

Is this expected?

hkaiser:
- properly report `e.what()` for std::exception

Fixes #6699

@JiakunYan please verify if this resolves your issue


JiakunYan:
Thanks! This fixes my issue!

hkaiser:
This looks good now. Merging....

PR #6704, opened 2025-05-14
Simplify code by using NS aliases

isidorostsa:
This covers the parts of the code that will be included in 2.0


isidorostsa:
Turns out this can introduce ADL miss-calls  in case someone has defined an `ex` namespace themselves for their senders. Solution is to rename to something more obscure (hpxexec)

Issue #6706, opened 2025-05-15
Fixing CI issues


Pansysk75:
Failing tests:
- [ ] [tests.examples.modules.collectives.channel_pingpong](https://cdash.rostam.cct.lsu.edu/test/18189652)
- [ ] [tests.unit.modules.algorithms.algorithms.move](https://cdash.rostam.cct.lsu.edu/test/18188913)
- [ ] [tests.unit.modules.compute_local.numa_allocator](https://cdash.rostam.cct.lsu.edu/test/18184657)
- [ ] [tests.unit.modules.algorithms.algorithms.sort](https://cdash.rostam.cct.lsu.edu/test/18201561)

Other issues
- [ ] Fix 50+ emitted build warnings

PR #6707, opened 2025-05-15
Fixing failing CI tests

Pansysk75:
Addresses #6706


Pansysk75:
@hkaiser 
Should we set the [CMP0144](https://cmake.org/cmake/help/latest/policy/CMP0144.html) CMake policy to NEW on a project-wise level?
Now, because it is unset, the existence of env variables like BOOST_ROOT, HWLOC_ROOT etc. cause CMake to emit warnings.

hkaiser:
> @hkaiser Should we set the [CMP0144](https://cmake.org/cmake/help/latest/policy/CMP0144.html) CMake policy to NEW on a project-wise level? Now, because it is unset, the existence of env variables like BOOST_ROOT, HWLOC_ROOT etc. cause CMake to emit warnings.

I thought to have changed all env variables to be 'properly' captalized. the ALL-CAPS versions are just supported for backwards compatibility reasons. However, please feel free to set the policy as needed.

hkaiser:
> However, I also noticed that you recently modified all `<PACKAGENAME>_ROOT` to `<PackageName>_ROOT` across the codebase. Could you point me to what motivated that change? Edit: Never mind, I now realize this change is because of [3.12](https://cmake.org/cmake/help/latest/policy/CMP0074.html), where the `<PackageName>_ROOT` lookup was introduced (before that CMake didn't enforce anything)

CMake issues warnings if one uses the ALL_CAPS names.

PR #6709, opened 2025-05-17
Transition SR tag_invoke to member functions

isidorostsa:
P2300 has moved away from the tag_invoke design pattern, following the guidance in [P2855](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2855r1.html). This PR removes tag_invoke support from the P2300 CPOs used by our senders. However, sender-adapter CPOs can remain tag_invoke-friendly since they provide an operator() that internally calls tag_invoke. For example, when_all_vector must drop support for tag_invoke(set_value_t...), but the when_all_vector algorithm itself can still be implemented as a CPO using operator() that calls tag_invoke(when_all_vector_t) under the hood.

At present, the non-stdexec implementation of senders will not be updated to work without tag_invoke.


hkaiser:
@isidorostsa I think this can be closed now as #7123 was merged. Please reopen if you think that this PR has applicable changes beyond what's already on master now.

PR #6724, opened 2025-06-22
modified existing extrema algo tests

### LINKED ISSUE #6627: Extrema algorithms with repeated values
Extrema algorithms (min_element/max_element/minmax_element) have specific semantics in case of repeated values. The existing unit tests don't cover these, as they only use `random_iota` as the input range. In particular:

- `min_element` is required to return the first min value,
- `max_element` is required to return the first max value,
- `minmax_element` is required to return the first min value and the last max value.

Tests for these cases should be added, and the implementations adjusted as needed.

sleepingeight:
Fixes #6627 

## Proposed Changes

  - Added a new test utility `random_repeat` to generate vectors having repeated values.
  - This utility is used to test the extrema algorithms, and I've verified that the existing implementations work properly.

Used `uniform_int_distribution` to generate random numbers. 
Tests for container_algorithms can also be modified, but the tests need to be modified to not use `sentinel` (as we have duplicate values in our vector). I can update the tests if usage of `sentinel` can be removed in these tests. 

## Checklist

### DIFF
--- a/libs/core/algorithms/tests/unit/algorithms/max_element.cpp
+++ b/libs/core/algorithms/tests/unit/algorithms/max_element.cpp
@@ -23,7 +23,7 @@ void test_max_element(IteratorTag)
     typedef std::vector<std::size_t>::iterator base_iterator;
     typedef test::test_iterator<base_iterator, IteratorTag> iterator;
 
-    std::vector<std::size_t> c = test::random_iota(10007);
+    std::vector<std::size_t> c = test::random_repeat(10007);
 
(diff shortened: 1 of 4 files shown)


hkaiser (approved):
LGTM, thanks!

hkaiser:
> * This utility is used to test the extrema algorithms, and I've verified that the existing implementations work properly.

I wouldn't like removing the tests realted to sentinels.

sleepingeight:
closing this since a PR solving this issue has already been merged.

PR #6737, opened 2025-07-11
Allowing to invoke parallel algorithms from main thread

hkaiser:


K-ballo:
This works for my use case

PR #6723, opened 2025-06-19
Adding specializations for merge that don't use projections

hkaiser:
- flyby: make hpx::merge use executor parameters for num_cores and chunk_size
- flyby: fixing #6721

PR #6739, opened 2025-07-12
Fixing the chunking algorithm for parallel loops to more evenly distribute the workload

hkaiser:


hkaiser:
Closing this as it is subsumed by #6723

PR #6746, opened 2025-07-18
Integrate NVIDIA's S/R Bulk implementation into HPX

charan-003:
This PR forwards stdexec::bulk and stdexec::bulk_t inside stdexec_forward.hpp.

These were previously commented out. This change makes bulk available through hpx::execution::experimental::bulk, aligning HPX's Sender/Receiver interface with NVIDIA's stdexec implementation.


libs/core/execution/include/hpx/execution/algorithms/bulk.hpp
isidorostsa, line 1: Here instead of deleting the old implementation, just leave it inside the ifdef so it is available when we do not have HPX_HAVE_STDEXEC

libs/core/executors/include/hpx/executors/explicit_scheduler_executor.hpp
```diff
@@ -212,8 +211,20 @@ namespace hpx::execution::experimental {
...
             {
+#if defined(HPX_HAVE_STDEXEC)
                 return bulk(schedule(exec.sched_), shape,
                     hpx::bind_back(HPX_FORWARD(F, f), HPX_FORWARD(Ts, ts)...));
+#else
+                // When stdexec is not available, fall back to sync execution
+                auto sender = schedule(exec.sched_);
+                for (auto&& elem : shape)
+                {
+                    sender = then(HPX_MOVE(sender), [f = HPX_FORWARD(F, f), elem, ts...](auto&&...) mutable {
```
isidorostsa, line 222: this seems wrong, the sender is never started so I think we are not doing anything here.

The old version will still have the old bulk operation, so I think we should be fine to leave this part as it was.

libs/core/executors/include/hpx/executors/explicit_scheduler_executor.hpp
```diff
@@ -243,10 +254,19 @@ namespace hpx::execution::experimental {
...
-                           shape, HPX_FORWARD(F, f), HPX_FORWARD(Ts, ts)...) |
-                    bulk(shape_size, HPX_MOVE(f_wrapper)) |
-                    then(HPX_MOVE(get_result));
+                // When stdexec is not available, execute sequentially
+                auto sender = transfer_just(exec.sched_, HPX_MOVE(result_vector),
+                           shape, HPX_FORWARD(F, f), HPX_FORWARD(Ts, ts)...);
+                
+                for (size_type i = 0; i < shape_size; ++i)
+                {
+                    sender = then(HPX_MOVE(sender), [i, f_wrapper](auto&& result_vector, auto&& shape, auto&& f, auto&&... ts) mutable {
```
isidorostsa, line 263: I think we should be able to use the old bulk implementation here too

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -53,6 +58,86 @@ namespace hpx::execution::experimental {
...
+    }
+
+    // Forward declarations for domain system
+    template <typename Policy>
+    struct thread_pool_policy_scheduler;
+
+    // Concept to check if a sender is any bulk operation
+    template <typename Sender>
+    concept __bulk_chunked_or_unchunked =
+        stdexec::sender_expr_for<Sender, stdexec::bulk_t> ||
```
isidorostsa, line 78: You can forward this to our namespace, but make sure that our senders will also be detectible by this

libs/core/execution/include/hpx/execution/algorithms/bulk.hpp
```diff
@@ -11,300 +11,6 @@
 
 #if defined(HPX_HAVE_STDEXEC)
```
isidorostsa, line 12: Here instead of deleting the old implementation, just leave it inside the ifdef so it is available when we do not have HPX_HAVE_STDEXEC

libs/core/execution_base/include/hpx/execution_base/stdexec_forward.hpp
```diff
@@ -138,9 +138,23 @@ namespace hpx::execution::experimental {
...
+
+    // Execution policies (required for stdexec bulk)
+    using stdexec::par;
+    using stdexec::par_unseq;
+    using stdexec::parallel_policy;
+    using stdexec::parallel_unsequenced_policy;
+    using stdexec::seq;
+    using stdexec::sequenced_policy;
+    using stdexec::unseq;
+    using stdexec::unsequenced_policy;
```
isidorostsa: Do we also define them in the same namespace? Is there any chance that they will clash?

libs/core/executors/include/hpx/executors/explicit_scheduler_executor.hpp
```diff
@@ -212,8 +211,20 @@ namespace hpx::execution::experimental {
 
             if constexpr (std::is_void_v<result_type>)
             {
+#if defined(HPX_HAVE_STDEXEC)
                 return bulk(schedule(exec.sched_), shape,
                     hpx::bind_back(HPX_FORWARD(F, f), HPX_FORWARD(Ts, ts)...));
+#else
+                // When stdexec is not available, fall back to sync execution
+                auto sender = schedule(exec.sched_);
+                for (auto&& elem : shape)
```
isidorostsa: this seems wrong, the sender is never started so I think we are not doing anything here.

The old version will still have the old bulk operation, so I think we should be fine to leave this part as it was.

libs/core/executors/include/hpx/executors/explicit_scheduler_executor.hpp
```diff
@@ -243,10 +254,19 @@ namespace hpx::execution::experimental {
...
+                
+                for (size_type i = 0; i < shape_size; ++i)
+                {
+                    sender = then(HPX_MOVE(sender), [i, f_wrapper](auto&& result_vector, auto&& shape, auto&& f, auto&&... ts) mutable {
+                        f_wrapper(i, result_vector, shape, f, ts...);
+                        return HPX_FORWARD(decltype(result_vector), result_vector);
+                    });
+                }
+                
+                return then(HPX_MOVE(sender), HPX_MOVE(get_result));
```
isidorostsa: I think we should be using the old bulk implementation here too

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -53,6 +58,86 @@ namespace hpx::execution::experimental {
...
+        stdexec::sender_expr_for<Sender, stdexec::bulk_chunked_t> ||
```
isidorostsa: It might be better if you moved `sender_expr_for` in the `hpx::execution::experimental` namespace.

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -53,6 +58,86 @@ namespace hpx::execution::experimental {
...
+        stdexec::sender_expr_for<Sender, stdexec::bulk_unchunked_t>;
+
+    // Domain customization for stdexec bulk operations
+    template <typename Policy>
+    struct thread_pool_domain : stdexec::default_domain
+    {
+        mutable thread_pool_policy_scheduler<Policy> sched_;
```
isidorostsa: If I understand correct, we might be able to pick this up from the environment of the sender instead, by doing something like:
`get_completion_scheduler(get_env(sender))` or `get_completion_scheduler(env)` inside the transform sender function, a bit like what stdexec does here: https://github.com/NVIDIA/stdexec/blob/e5dbe49a5257d29adee456f9f653acb00df634f5/include/exec/system_context.hpp#L710-L740

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -387,6 +488,16 @@ namespace hpx::execution::experimental {
         {
             return policy_;
         }
+
+#if defined(HPX_HAVE_STDEXEC)
+        /// Returns the execution domain of this scheduler (following system_context.hpp pattern).
+        [[nodiscard]]
+        auto query(stdexec::get_domain_t) const noexcept
+            -> thread_pool_domain<Policy>
```
isidorostsa, line 464: I think it would be okay here if we returned `{}` instead, like stdexec does here:
https://github.com/NVIDIA/stdexec/blob/e5dbe49a5257d29adee456f9f653acb00df634f5/include/exec/system_context.hpp#L284-L286

libs/core/execution_base/include/hpx/execution_base/stdexec_forward.hpp
charan-003: there's no direct clash because they're in different namespaces

hpx::execution::par (HPX's policy)
hpx::execution::experimental::par (stdexec's policy)

libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
```diff
@@ -213,9 +213,13 @@ void test_for_each_exception_async(ExPolicy&& p, IteratorTag)
...
         test::test_num_exceptions<ExPolicy, IteratorTag>::call(p, e);
     }
+    catch (std::runtime_error const&)
+    {
+        caught_exception = true;
+    }
     catch (...)
     {
-        HPX_TEST(false);
+        caught_exception = true;
```
isidorostsa, line 222: Leave a small comment  about this please

libs/core/execution_base/include/hpx/execution_base/stdexec_forward.hpp
```diff
@@ -133,14 +133,34 @@ namespace hpx::execution::experimental {
...
+    using stdexec::bulk_t;
+    using stdexec::bulk_unchunked;
+    using stdexec::bulk_unchunked_t;
+
+    // Execution policies for stdexec bulk operations
+    // Use stdexec's own execution policy types
+    using stdexec::is_execution_policy;
+    using stdexec::is_execution_policy_v;
+    using stdexec::par;
+    using stdexec::par_unseq;
```
isidorostsa: Could you add those to a namespace simillar to what we did with "stdexec_non_standard_tag_inoke" below?

libs/core/execution_base/include/hpx/execution_base/stdexec_forward.hpp
```diff
@@ -308,6 +328,11 @@ namespace hpx::execution::experimental {
 
         using stdexec::__connect_awaitable_t;
     }    // namespace stdexec_internal
+
+    // Additional stdexec concepts and utilities needed for domain customization
+    using stdexec::__completes_on;
```
isidorostsa: Maybe check if those exist without the underscores

libs/core/executors/include/hpx/executors/explicit_scheduler_executor.hpp
```diff
@@ -239,18 +245,21 @@ namespace hpx::execution::experimental {
 #if defined(HPX_HAVE_STDEXEC)
                 return just(HPX_MOVE(result_vector), shape, HPX_FORWARD(F, f),
                            HPX_FORWARD(Ts, ts)...) |
-                    continue_on(exec.sched_) |
+                    transfer(exec.sched_) |
```
isidorostsa: maybe that should be `continues_on`
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2300r10.html#design-sender-adaptor-continues_on

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -53,6 +63,116 @@ namespace hpx::execution::experimental {
...
+
+        // FALLBACK: Handle non-bulk senders by delegating to default domain
+        template <stdexec::sender Sender, typename Env>
+            requires(!__bulk_chunked_or_unchunked<Sender>)
+        auto transform_sender(Sender&& sndr, const Env& env) const noexcept
+        {
+            // delegate non-bulk senders to default domain
+            return stdexec::default_domain::transform_sender(
+                std::forward<Sender>(sndr), env);
+        }
```
isidorostsa: In case those are not necessary they should be removed

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -53,6 +63,116 @@ namespace hpx::execution::experimental {
...
+                        sched,    // scheduler from environment
+                        std::forward<decltype(child)>(child),    // child sender
+                        std::forward<decltype(data.__shape_)>(
+                            data.__shape_),    // shape
+                        std::forward<decltype(data.__fun_)>(
+                            data.__fun_)    // function
+                    };
+            }
+            else if constexpr (hpx::execution::experimental::__starts_on<Sender,
+                                   thread_pool_policy_scheduler<Policy>, Env>)
```
isidorostsa: Is this branch necessary?

isidorostsa: We should follow what they are doing here

https://github.com/NVIDIA/stdexec/blob/4669060ad3b70508740449f9c30f06867a9bd890/include/exec/system_context.hpp#L710-L738

Probably adding two implementations instead of one with a branch

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -53,6 +63,116 @@ namespace hpx::execution::experimental {
...
+
+                // Extract bulk parameters using structured binding
+                auto [tag, data, child] = sndr;
+
+                // Create HPX thread_pool_bulk_sender with extracted parameters
+                return hpx::execution::experimental::detail::
+                    thread_pool_bulk_sender<Policy,
+                        std::decay_t<decltype(child)>,
+                        std::decay_t<decltype(data.__shape_)>,
+                        std::decay_t<decltype(data.__fun_)>>{
```
isidorostsa: we can probably break down data into shape and fun, while staying within the standard.

Sketchy .txt:
```
schedule(thread_pool_scheduler) 
| then( ... ) <--- sndr
| bulk( ... )

  get-domain-early(sndr) == thred_pool_scheduler_domain{}
  
  transform_sender(
  get-domain-early(sndr),
  make-sender(bulk, product-type{shape, f}, sndr))
 ====
 thred_pool_scheduler_domain{}.
 	transform_sender(
 		make-sender(bulk, product-type{shape, f}, sndr)
 	)
 	
transform_sender(auto thing)
[tag, data, child] = thing
[shape, f] = data
```

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -53,6 +63,116 @@ namespace hpx::execution::experimental {
...
+    // Domain customization for stdexec bulk operations
+    template <typename Policy>
+    struct thread_pool_domain : stdexec::default_domain
+    {
+        // Only intercept stdexec bulk expressions using concept constraints
+        template <__bulk_chunked_or_unchunked Sender, typename Env>
+        auto transform_sender(Sender&& sndr, const Env& env) const noexcept
+        {
+            // Extract scheduler from sender environment, following stdexec pattern
+            if constexpr (hpx::execution::experimental::__completes_on<Sender,
```
isidorostsa: I think instead of this we could 
`static_assert(__completes_on<Sender, thread_pool_policy_scheduler<Policy>>)` for the overload without the env, and `static_assert(__starts_on<Sender, thread_pool_policy_scheduler<Policy>>)` for the overload with the env.
https://github.com/NVIDIA/stdexec/blob/4669060ad3b70508740449f9c30f06867a9bd890/include/exec/system_context.hpp#L710-L738

isidorostsa (commented):
Really Great Work!

libs/core/execution_base/include/hpx/execution_base/stdexec_forward.hpp
charan-003: Nope, i didn't find any.
The current using `stdexec::__completes_on;` and using `stdexec::__starts_on;` declarations are correct as they are.

libs/core/execution_base/include/hpx/execution_base/stdexec_forward.hpp
```diff
@@ -133,14 +133,44 @@ namespace hpx::execution::experimental {
...
 
-    // Bulk (NOT FORWARDED)
-    //    using stdexec::bulk_t;
-    //    using stdexec::bulk;
+    namespace stdexec_bulk_operations {
+        // Bulk operations - stdexec provides these
+        using stdexec::bulk;
+        using stdexec::bulk_chunked;
+        using stdexec::bulk_chunked_t;
+        using stdexec::bulk_t;
```
isidorostsa: I don't think the bulk-related ones need to be in a separate namespace

libs/core/execution_base/include/hpx/execution_base/stdexec_forward.hpp
```diff
@@ -133,14 +133,44 @@ namespace hpx::execution::experimental {
...
+        using stdexec::bulk_unchunked;
+        using stdexec::bulk_unchunked_t;
+
+        // Execution policies for stdexec bulk operations
+        // Use stdexec's own execution policy types
+        using stdexec::is_execution_policy;
+        using stdexec::is_execution_policy_v;
```
isidorostsa: On a second thought, you should remove those as well. Let's keep only the top level ones you've written bellow. Its okay for now, we can use hpx::execution::par when referring to senders

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -53,6 +63,95 @@ namespace hpx::execution::experimental {
...
+            auto [tag, data, child] = sndr;
+            auto [shape, f] = data;
+
+            // Create HPX thread_pool_bulk_sender with extracted parameters
+            return hpx::execution::experimental::detail::
+                thread_pool_bulk_sender<Policy, std::decay_t<decltype(child)>,
+                    std::decay_t<decltype(shape)>,
+                    std::decay_t<decltype(f)>>{
+                    sched,    // scheduler from environment
+                    std::forward<decltype(child)>(child),    // child sender
```
isidorostsa: Here you could use the HPX_FORWARD macro for uniformity

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -53,6 +63,95 @@ namespace hpx::execution::experimental {
...
+        template <__bulk_chunked_or_unchunked Sender>
+        auto transform_sender(Sender&& sndr) const noexcept
+        {
+            static_assert(hpx::execution::experimental::__completes_on<Sender,
+                              thread_pool_policy_scheduler<Policy>>,
+                "No thread_pool_policy_scheduler instance can be found in the "
+                "sender's "
+                "attributes on which to schedule bulk work.");
+
+            auto sched = hpx::execution::experimental::get_completion_scheduler<
```
isidorostsa: Will this create a copy of the completion scheduler? If so, is this the way this is supposed to work? If not, could you please also add a test that tests for this after fixing it?

charan-003: I added a 
`test_scheduler_copy_avoidance()`
 test with a 
`copy_tracking_scheduler`
 that measures copy/move operations during bulk operations. Results show 0 copies and 0 moves. The `auto&&` forwarding references in `get_completion_scheduler` calls work correctly to avoid copy overhead.

libs/core/execution_base/include/hpx/execution_base/stdexec_forward.hpp
```diff
@@ -141,36 +141,13 @@ namespace hpx::execution::experimental {
...
-    using stdexec_bulk_operations::bulk_t;
-    using stdexec_bulk_operations::bulk_unchunked;
-    using stdexec_bulk_operations::bulk_unchunked_t;
+    // Bulk operations
+    using stdexec::bulk;
+    using stdexec::bulk_chunked;
+    using stdexec::bulk_chunked_t;
+    using stdexec::bulk_t;
+    using stdexec::bulk_unchunked;
+    using stdexec::bulk_unchunked_t;
```
isidorostsa, line 152: You were right to forward the policies here in the beginning. Let's include them again :)

isidorostsa (changes requested):

libs/core/executors/tests/unit/thread_pool_scheduler.cpp
```diff
@@ -2369,15 +2369,14 @@ void test_scheduler_copy_avoidance()
...
-    
     (void) final_move_count;
 
     HPX_TEST(true);
 }
 
 ///////////////////////////////////////////////////////////////////////////////
 int hpx_main()
-{
+{ß
```
isidorostsa: is that a typo :P

charan-003: yeah, i mis-clicked something lol

libs/core/executors/tests/unit/thread_pool_scheduler.cpp
```diff
@@ -2088,19 +2091,156 @@ void test_bulk()
...
+
+void test_stdexec_bulk_domain_customization()
+{
+    auto scheduler = ex::thread_pool_scheduler{};
+
+    // Test basic bulk operation with domain customization
+    std::vector<int> results(10, 0);
+    auto bulk_sender = stdexec::bulk(
+        ex::schedule(scheduler) | stdexec::then([]() { return 42; }),
+        stdexec::par, 10,
```
isidorostsa: Here and in simillar places use the hpx::execution::experimental::par alias since you've forwarded the policies from stdexec to our namesapce

libs/core/executors/tests/unit/thread_pool_scheduler.cpp
```diff
@@ -2181,24 +2297,83 @@ void test_completion_scheduler()
...
+    // Verify that no unnecessary copies were made
+    int final_copy_count = copy_count.load();
+    int final_move_count = move_count.load();
+
+    // Scheduler operations should minimize copies
+    HPX_TEST_LTE(final_copy_count, 1);
+    (void) final_move_count;
+
+    HPX_TEST(true);
+}
```
isidorostsa, line 2602: I think this should be in the regressions test

isidorostsa (changes requested):

libs/core/executors/tests/unit/thread_pool_scheduler.cpp
charan-003: @isidorostsa 
hpx/tests/regressions
https://github.com/STEllAR-GROUP/hpx/tree/acf07297232076fd53c9aca5bb02280e48a73447/libs/core/execution/tests/regressions
do you mean in this?

assuming you want me to create a new file?

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -53,6 +63,96 @@ namespace hpx::execution::experimental {
...
+                "attributes on which to schedule bulk work.");
+
+            auto&& sched =
+                hpx::execution::experimental::get_completion_scheduler<
+                    hpx::execution::experimental::set_value_t>(
+                    hpx::execution::experimental::get_env(sndr));
+
+            // Extract bulk parameters using structured binding
+            auto [tag, data, child] = sndr;
+            auto [shape, f] = data;
```
isidorostsa: Note for the future: make this also forward the execution policy of the bulk call

https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3481r5.html#specification.bulk

libs/core/executors/tests/unit/thread_pool_scheduler.cpp
```diff
@@ -1951,8 +1952,8 @@ void test_bulk()
         hpx::thread::id parent_id = hpx::this_thread::get_id();
 
 #if defined(HPX_HAVE_STDEXEC)
-        tt::sync_wait(
-            ex::schedule(ex::thread_pool_scheduler{}) | ex::bulk(n, [&](int i) {
+        tt::sync_wait(ex::schedule(ex::thread_pool_scheduler{}) |
+            ex::bulk(ex::par, n, [&](int i) {
                 ++v[i];
```
isidorostsa, line 2020: Let's test with a static assertion that this sender is actually a thread_pool_policy_sender<...>

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -627,6 +680,7 @@ namespace hpx::execution::experimental::detail {
         HPX_NO_UNIQUE_ADDRESS std::decay_t<Shape> shape;
         HPX_NO_UNIQUE_ADDRESS std::decay_t<F> f;
         hpx::threads::mask_type pu_mask;
+        bool is_chunked_execution;
```
isidorostsa, line 703: This should not be a member variable, we can use the bool template parameter directly

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -69,30 +71,23 @@ namespace hpx::execution::experimental {
     struct thread_pool_policy_scheduler;
 
     namespace detail {
-        template <typename Policy, typename Sender, typename Shape, typename F>
+        template <typename Policy, typename Sender, typename Shape, typename F,
```
isidorostsa: Is that a second forward declaration?

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -69,30 +71,23 @@ namespace hpx::execution::experimental {
...
-        hpx::execution::experimental::sender_expr_for<Sender,
-            hpx::execution::experimental::bulk_unchunked_t>;
-
     // Domain customization for stdexec bulk operations
     template <typename Policy>
     struct thread_pool_domain : stdexec::default_domain
     {
-        // Transform bulk sender without environment (completes_on pattern)
-        template <__bulk_chunked_or_unchunked Sender>
+        // Handle bulk_chunked operations without environment (completes_on pattern)
```
isidorostsa: I think we could get away with what we were doing before, if we also use a constexpr bool later in the function.

Essentially:

```c++
         template <__bulk_chunked_or_unchunked Sender>
         ...
         constexpr bool is_chunked =  
             sender_expr_for<bulk_chunked_t, Sender> || sender_expr_for<bulk_t, Sender>;

         return hpx::execution::experimental::detail::
                thread_pool_bulk_sender<Policy, std::decay_t<decltype(child)>,
                    std::decay_t<decltype(iota_shape)>,
                    std::decay_t<decltype(f)>, 
                    is_chunked>{
                    // true for chunked
                    HPX_MOVE(sched),    // scheduler from environment
                    HPX_FORWARD(decltype(child), child),    // child sender
                    HPX_MOVE(iota_shape),                   // shape
                    HPX_FORWARD(decltype(f), f)             // function
                };
```

So we avoid writing the same code 3 times

isidorostsa (changes requested):
Great idea to have a bool NTTP for the chunked kind :)

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -824,6 +820,8 @@ namespace hpx::execution::experimental::detail {
         template <typename Receiver>
         struct operation_state
         {
+            static constexpr bool is_chunked = IsChunked;
```
isidorostsa, line 812: perfect!

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -190,99 +111,66 @@ namespace hpx::execution::experimental {
...
+
+                return hpx::execution::experimental::detail::
+                    thread_pool_bulk_sender<Policy,
+                        std::decay_t<decltype(child)>,
+                        std::decay_t<decltype(iota_shape)>,
+                        std::decay_t<decltype(f)>, true>{
+                        HPX_MOVE(sched),    // scheduler from environment
+                        HPX_FORWARD(decltype(child), child),    // child sender
+                        HPX_MOVE(iota_shape),                   // shape
+                        HPX_FORWARD(decltype(f), f)             // function
```
isidorostsa: We can remove  this section and let bulk use bulk_chunked through the default implementation (As noted https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3481r5.html#api)

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -71,103 +71,27 @@ namespace hpx::execution::experimental {
...
-    struct thread_pool_policy_scheduler;
+
+    // Concept to match any bulk sender type
+    template <typename Sender>
+    concept any_bulk_sender =
+        hpx::execution::experimental::sender_expr_for<Sender,
+            hpx::execution::experimental::bulk_chunked_t> ||
+        hpx::execution::experimental::sender_expr_for<Sender,
+            hpx::execution::experimental::bulk_unchunked_t> ||
+        hpx::execution::experimental::sender_expr_for<Sender,
```
isidorostsa: This should only detect bulk_chunked and unchunked, and then bulk can use the default implementation to call bulk_chunked

isidorostsa (approved):

libs/core/algorithms/include/hpx/parallel/container_algorithms/for_each.hpp
isidorostsa, line 1: Let's remove those files from this PR as they are being worked on for the same reason by another contributor

libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
isidorostsa, line 1: Changes here are not important

libs/core/execution_base/include/hpx/execution_base/stdexec_forward.hpp
```diff
@@ -133,14 +133,29 @@ namespace hpx::execution::experimental {
     // Continue on
     using stdexec::continue_on;
     using stdexec::continue_on_t;
+    // Backward compatibility alias for continues_on
```
isidorostsa: No need to have backwards compatibility here, you can remove 134-138

libs/core/execution_base/include/hpx/execution_base/stdexec_forward.hpp
```diff
@@ -308,6 +323,11 @@ namespace hpx::execution::experimental {
 
         using stdexec::__connect_awaitable_t;
     }    // namespace stdexec_internal
+
+    // Additional stdexec concepts and utilities needed for domain customization
+    using stdexec::__completes_on;
+    using stdexec::__starts_on;
```
isidorostsa: __completes_on, __starts_on, and sender_expr_for should be moved in the stdexec internals namespace

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -53,6 +65,142 @@ namespace hpx::execution::experimental {
...
+
+    namespace detail {
+    }
+
+    // Forward declarations for domain system
+
+    // Concept to match bulk sender types
+    // Note: We keep bulk_t handling as pragmatic workaround for stdexec template issues
+    template <typename Sender>
+    concept any_bulk_sender =
```
isidorostsa: should be renamed to `bulk_chunked_or_unchunked_sender`

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -343,9 +498,38 @@ namespace hpx::execution::experimental {
             {
                 return s.scheduler;
             }
+#else
```
isidorostsa: `env_no_stdexec` can be removed and we only need to remove get_domain from the modern env using an `#if defined` clause

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -53,26 +55,64 @@ namespace hpx::execution::experimental::detail {
...
         HPX_INVOKE(HPX_FORWARD(F, f), HPX_FORWARD(T, t), hpx::get<Is>(ts)...);
     }
 
+    // For bulk_chunked: f(start, end, ...)
+    template <std::size_t... Is, typename F, typename Start, typename End,
+        typename Ts>
+    constexpr void bulk_scheduler_invoke_helper_chunked(
+        hpx::util::index_pack<Is...>, F&& f, Start&& start, End&& end, Ts& ts)
+    {
+        if constexpr (sizeof...(Is) == 0)
```
isidorostsa: We can remove the `sizeof...(Is) == 0` test here

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -140,18 +180,31 @@ namespace hpx::execution::experimental::detail {
...
             using index_pack_type = hpx::detail::fused_index_pack_t<Ts>;
 
             auto const i_begin =
                 static_cast<std::size_t>(index) * task_f->chunk_size;
             auto const i_end =
                 (std::min) (i_begin + task_f->chunk_size, task_f->size);
 
-            auto it = std::next(hpx::util::begin(op_state->shape), i_begin);
-            for (std::uint32_t i = i_begin; i != i_end; (void) ++it, ++i)
+            auto it =
```
isidorostsa: this should be moved inside the `else` branch below

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -442,14 +495,14 @@ namespace hpx::execution::experimental::detail {
...
-        using range_value_type =
-            hpx::traits::iter_value_t<hpx::traits::range_iterator_t<Shape>>;
+        using range_value_type = std::ranges::range_value_t<Shape>;
 
         template <typename... Ts>
         void execute(Ts&&... ts)
         {
             // Don't spawn tasks if there is no work to be done
             auto const size =
+                // static_cast<std::uint32_t>(op_state->shape);
```
isidorostsa: we can remove this commented out part

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -583,11 +644,7 @@ namespace hpx::execution::experimental::detail {
         }
 
         // clang-format off
-        template <typename... Ts,
-            HPX_CONCEPT_REQUIRES_(
-                hpx::is_invocable_v<F, range_value_type,
-                    std::add_lvalue_reference_t<Ts>...>
-            )>
+        template <typename... Ts>
```
isidorostsa, line 637: We should re-add the test as a normal concept (requires) clause

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -845,61 +906,8 @@ namespace hpx::execution::experimental::detail {
     };
 }    // namespace hpx::execution::experimental::detail
 
-namespace hpx::execution::experimental {
```
isidorostsa, line 836: Better wrap this in an `#if !defined` clause instead of removing, so we keep backwards compatibility for now

isidorostsa (dismissed):
Really great work!

libs/core/execution_base/include/hpx/execution_base/stdexec_forward.hpp
```diff
@@ -133,9 +133,6 @@ namespace hpx::execution::experimental {
     // Continue on
     using stdexec::continue_on;
```
isidorostsa: This (134, 135) should be removed too, since the proper spelling is `continues_on`

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -458,9 +500,12 @@ namespace hpx::execution::experimental::detail {
                 return;
             }
 
-            // Calculate chunk size and number of chunks
-            std::uint32_t chunk_size = get_bulk_scheduler_chunk_size(
-                op_state->num_worker_threads, size);
+            // Calculate chunk size based on execution mode
```
isidorostsa, line 503: for unchunked chunk_size should be 1 and num_chunks = size

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -458,9 +500,12 @@ namespace hpx::execution::experimental::detail {
...
+            std::uint32_t chunk_size = op_state->is_chunked ?
+                get_bulk_scheduler_chunk_size_chunked(
+                    op_state->num_worker_threads, size) :
+                get_bulk_scheduler_chunk_size_unchunked(
```
isidorostsa:
```suggestion
                1
```

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -53,26 +55,54 @@ namespace hpx::execution::experimental::detail {
 
     ///////////////////////////////////////////////////////////////////////////
     // Compute a chunk size given a number of worker threads and a total number
-    // of items n. Returns a power-of-2 chunk size that produces at most 8 and
-    // at least 4 chunks per worker thread.
-    static constexpr std::uint32_t get_bulk_scheduler_chunk_size(
+    // Chunked execution: larger chunks for better cache locality and fewer
+    // context switches
+    static constexpr std::uint32_t get_bulk_scheduler_chunk_size_chunked(
```
isidorostsa: keep the old implementation here

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -53,26 +55,54 @@ namespace hpx::execution::experimental::detail {
...
+        while (chunk_size * num_threads * 4 <
+            n)    // Fewer chunks per thread (4 instead of 8)
         {
             chunk_size *= 2;
         }
         return static_cast<std::uint32_t>(chunk_size);
     }
 
+    // Unchunked execution: smaller chunks for better load balancing and work stealing
+    static constexpr std::uint32_t get_bulk_scheduler_chunk_size_unchunked(
```
isidorostsa: probably don't need this

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -140,18 +170,31 @@ namespace hpx::execution::experimental::detail {
...
+                    index_pack_type{}, op_state->f, i_begin, i_end, ts);
+            }
+            else
             {
-                bulk_scheduler_invoke_helper(
-                    index_pack_type{}, op_state->f, *it, ts);
+                // bulk_unchunked: f(index, values...) for each element
+                auto it = std::ranges::next(
+                    hpx::util::begin(op_state->shape), i_begin);
+                for (std::uint32_t i = i_begin; i != i_end; (void) ++i)
```
isidorostsa: This can be replaced by only calling `bulk_scheduler_invoke_helper` once, instead of on a loop. We are in the unchunked case and thus our chunk_size is 0, so each loop will only run one iteration anyways. The regular bulk invocation will go through the `is_chunked` case.

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -19,7 +19,9 @@
 #include <hpx/coroutines/thread_enums.hpp>
 #include <hpx/datastructures/tuple.hpp>
 #include <hpx/datastructures/variant.hpp>
-#include <hpx/execution/algorithms/bulk.hpp>
```
isidorostsa: This should be included in the headers when not using STDEXEC, it seems this is what's causing this failing test:

https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/18789/workflows/d6fb3490-5f5a-4b18-8cef-730ffef44797/jobs/427793

charan-003:
@hkaiser @isidorostsa just a small ping on this PR. Could you please take a look when you have a moment and let me know if anything else is needed?

libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
```diff
@@ -213,9 +213,15 @@ void test_for_each_exception_async(ExPolicy&& p, IteratorTag)
         caught_exception = true;
         test::test_num_exceptions<ExPolicy, IteratorTag>::call(p, e);
     }
+    catch (std::runtime_error const&)
+    {
+        // Handle direct runtime_error thrown by the lambda
+        caught_exception = true;
+    }
```
hkaiser, line 219: Why do you need to handle `std::runtime_error` separately?

libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
```diff
@@ -349,12 +356,23 @@ void test_for_each_sender(Policy l, ExPolicy&& p, IteratorTag)
         iterator(std::begin(c)), iterator(std::end(c)));
     auto f = [](std::size_t& v) { v = 42; };
 
-    using scheduler_t = ex::thread_pool_policy_scheduler<Policy>;
+    // using scheduler_t = ex::thread_pool_policy_scheduler<Policy>;
```
hkaiser: Can this be removed?

libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
```diff
@@ -385,28 +404,49 @@ void test_for_each_exception_sender(Policy l, ExPolicy&& p, IteratorTag)
     bool caught_exception = false;
     try
     {
-        using scheduler_t = ex::thread_pool_policy_scheduler<Policy>;
-
-        auto exec = ex::explicit_scheduler_executor(scheduler_t(l));
-        tt::sync_wait(ex::just(rng, f) | hpx::ranges::for_each(p.on(exec)));
-
-        HPX_TEST(false);
+        // using scheduler_t = ex::thread_pool_policy_scheduler<Policy>;
```
hkaiser: Can this be removed?

libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
```diff
@@ -385,28 +404,49 @@ void test_for_each_exception_sender(Policy l, ExPolicy&& p, IteratorTag)
...
     }
     catch (hpx::exception_list const& e)
     {
         caught_exception = true;
         test::test_num_exceptions<ExPolicy, IteratorTag>::call(p, e);
     }
+    catch (std::runtime_error const&)
+    {
+        caught_exception = true;
+    }
```
hkaiser, line 467: Same here, do we need the explicit handling of `std::runtime_error`?

libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
```diff
@@ -424,10 +464,18 @@ void test_for_each_bad_alloc_sender(Policy l, ExPolicy&& p, IteratorTag)
     bool caught_exception = false;
     try
     {
-        using scheduler_t = ex::thread_pool_policy_scheduler<Policy>;
-
-        auto exec = ex::explicit_scheduler_executor(scheduler_t(l));
-        tt::sync_wait(ex::just(rng, f) | hpx::ranges::for_each(p.on(exec)));
+        // using scheduler_t = ex::thread_pool_policy_scheduler<Policy>;
```
hkaiser: Please remove this.

libs/core/execution/include/hpx/execution/algorithms/bulk.hpp
```diff
@@ -41,40 +46,6 @@ namespace hpx::execution::experimental {
             HPX_NO_UNIQUE_ADDRESS std::decay_t<Shape> shape;
             HPX_NO_UNIQUE_ADDRESS std::decay_t<F> f;
 
-#if defined(HPX_HAVE_STDEXEC)
```
hkaiser, line 44: Is there a reason the sections related to `std::execution` have been removed in this file?

libs/core/execution_base/include/hpx/execution_base/stdexec_forward.hpp
```diff
@@ -302,11 +317,16 @@ namespace hpx::execution::experimental {
...
+        namespace __connect_await {
+            HPX_CXX_CORE_EXPORT using namespace stdexec::__connect_await;
         }
 
         HPX_CXX_CORE_EXPORT using stdexec::__connect_awaitable_t;
+
+        // Additional stdexec concepts and utilities needed for domain customization
+        using stdexec::__completes_on;
+        using stdexec::__starts_on;
+        using stdexec::sender_expr_for;
```
hkaiser: Do these symbols need to be exported through the C++ module:
```suggestion
        HPX_CXX_CORE_EXPORT using stdexec::__completes_on;
        HPX_CXX_CORE_EXPORT using stdexec::__starts_on;
        HPX_CXX_CORE_EXPORT using stdexec::sender_expr_for;
```
?

libs/core/executors/include/hpx/executors/explicit_scheduler_executor.hpp
```diff
@@ -9,14 +9,27 @@
...
+#include <hpx/execution/algorithms/when_all.hpp>
+#include <hpx/execution/executors/execution.hpp>
+#include <hpx/execution/executors/execution_parameters.hpp>
+#include <hpx/execution_base/execution.hpp>
+#include <hpx/execution_base/sender.hpp>
+#include <hpx/execution_base/stdexec_forward.hpp>
+#include <hpx/execution_base/traits/is_executor.hpp>
+#include <hpx/functional/deferred_call.hpp>
+#include <hpx/functional/invoke_fused.hpp>
+#include <hpx/functional/tag_invoke.hpp>
```
hkaiser: Please revert this change to be aligned with master. We now have to include everything through the HPX module headers only.

libs/core/executors/include/hpx/executors/explicit_scheduler_executor.hpp
```diff
@@ -83,12 +96,8 @@ namespace hpx::execution::experimental {
...
         }
 
-        // clang-format off
-        template <typename Parameters,
-            HPX_CONCEPT_REQUIRES_(
-                hpx::traits::is_executor_parameters_v<Parameters>
-            )>
-        // clang-format on
+        template <typename Parameters>
+            requires(hpx::traits::is_executor_parameters_v<Parameters>)
```
hkaiser: Please revert this change as well. `requires` and SFINAE based return type calculation doesn't mix well.

libs/core/executors/include/hpx/executors/explicit_scheduler_executor.hpp
```diff
@@ -226,18 +231,21 @@ namespace hpx::execution::experimental {
...
+                    continues_on(exec.sched_) |
                     bulk(shape_size, HPX_MOVE(f_wrapper)) |
                     then(HPX_MOVE(get_result));
 #else
-                return transfer_just(exec.sched_, HPX_MOVE(result_vector),
-                           shape, HPX_FORWARD(F, f), HPX_FORWARD(Ts, ts)...) |
+                // When stdexec is not available, use HPX's original bulk implementation
+                return just(HPX_MOVE(result_vector), shape, HPX_FORWARD(F, f),
+                           HPX_FORWARD(Ts, ts)...) |
+                    continues_on(exec.sched_) |
```
hkaiser: Do we have `continues_on` in our old implementation now?

libs/core/executors/include/hpx/executors/scheduler_executor.hpp
```diff
@@ -274,7 +274,7 @@ namespace hpx::execution::experimental {
                     when_all(keep_future(HPX_FORWARD(Future, predecessor)));
 
                 auto loop = bulk(transfer(HPX_MOVE(pre_req), exec.sched_),
-                    shape,
+                    hpx::util::size(shape),
```
hkaiser: What's the rationale of this change?

libs/core/executors/include/hpx/executors/scheduler_executor.hpp
```diff
@@ -286,10 +286,10 @@ namespace hpx::execution::experimental {
                     when_all(keep_future(HPX_FORWARD(Future, predecessor)),
                         just(std::vector<result_type>(hpx::util::size(shape))));
 
-                auto loop =
-                    bulk(transfer(HPX_MOVE(pre_req), exec.sched_), shape,
-                        detail::captured_args_then(
-                            HPX_FORWARD(F, f), HPX_FORWARD(Ts, ts)...));
+                auto loop = bulk(transfer(HPX_MOVE(pre_req), exec.sched_),
+                    hpx::util::size(shape),
```
hkaiser: Same, what's the rationale?

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -8,21 +8,43 @@
...
-#include <hpx/modules/timing.hpp>
-#include <hpx/modules/topology.hpp>
+#include <hpx/modules/memory.hpp>
+#include <hpx/threading_base/annotated_function.hpp>
+#include <hpx/threading_base/register_thread.hpp>
+#include <hpx/threading_base/thread_helpers.hpp>
+#include <hpx/threading_base/thread_pool_base.hpp>
+#include <hpx/topology/cpu_mask.hpp>
+#include <hpx/type_support/detail/with_result_of.hpp>
+#include <hpx/type_support/pack.hpp>
```
hkaiser: Please revert the changes to the #includes.

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -222,7 +307,7 @@ namespace hpx::execution::experimental {
             {
             }
 
-            operation_state(operation_state&&) = delete;
+            operation_state(operation_state&&) = default;
             operation_state(operation_state const&) = delete;
             operation_state& operator=(operation_state&&) = delete;
```
hkaiser: If you `default` the move constructor, you should `default` the move assignment as well.

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -9,25 +9,37 @@
...
-#include <hpx/modules/functional.hpp>
-#include <hpx/modules/iterator_support.hpp>
-#include <hpx/modules/resource_partitioner.hpp>
-#include <hpx/modules/tag_invoke.hpp>
-#include <hpx/modules/threading_base.hpp>
-#include <hpx/modules/topology.hpp>
 #include <hpx/modules/type_support.hpp>
+#include <hpx/resource_partitioner/detail/partitioner.hpp>
+#include <hpx/threading_base/annotated_function.hpp>
+#include <hpx/topology/cpu_mask.hpp>
```
hkaiser: Please revert the changes to the #includes.

tests/performance/local/CMakeLists.txt
```diff
@@ -131,6 +131,7 @@ set(partitioned_vector_foreach_FLAGS DEPENDENCIES iostreams_component
 
 set(future_overhead_PARAMETERS THREADS_PER_LOCALITY 4)
 set(future_overhead_report_PARAMETERS THREADS_PER_LOCALITY 4)
+set(stream_report_PARAMETERS THREADS_PER_LOCALITY 2)
```
hkaiser, line 134: Why this change?

libs/core/execution/include/hpx/execution/algorithms/bulk.hpp
isidorostsa: It's because the entire file is defined away when STDEXEC is present.

libs/core/executors/include/hpx/executors/scheduler_executor.hpp
isidorostsa: This seems to be wrong, as it ignores the shape's elements. 
The underlying issue is the official bulk algorithm only accepting an integral shape parameter. This is the only difference between our bulk implementation and the standard's.

tests/performance/local/CMakeLists.txt
charan-003: stream_report test: it was passing with 2 or 4 threads but with 1 thread it seems to be hanging that's why i changed this last time

libs/core/executors/include/hpx/executors/scheduler_executor.hpp
charan-003: I misunderstood the difference between HPX's bulk and stdexec's bulk. HPX's bulk implementation needs the shape object itself to iterate over its elements, not just the size.

libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
charan-003: right, `std::runtime_error` catch block is redundant

libs/core/algorithms/include/hpx/parallel/algorithms/transform.hpp
```diff
@@ -621,6 +622,12 @@ namespace hpx::parallel {
                     transform_binary_projected<F_, Proj1, Proj2>{
                         f_, proj1_, proj2_});
             }
+
+            HPX_HOST_DEVICE HPX_FORCEINLINE constexpr void operator()(
+                std::size_t) const
```
hkaiser: This can be made `noexcept`. In release mode the function does nothing and in debug mode the code will terminate anyways.

libs/core/algorithms/include/hpx/parallel/algorithms/transform.hpp
```diff
@@ -678,6 +685,12 @@ namespace hpx::parallel {
                     hpx::get<0>(iters), part_size, hpx::get<1>(iters),
                     hpx::get<2>(iters), f_);
             }
+
+            HPX_HOST_DEVICE HPX_FORCEINLINE constexpr void operator()(
+                std::size_t) const
```
hkaiser: Same here, can be made `noexcept`

libs/core/algorithms/include/hpx/parallel/util/detail/partitioner_iteration.hpp
```diff
@@ -16,19 +16,45 @@
...
             return hpx::invoke_fused_r<Result>(f_, HPX_FORWARD(T, t));
         }
 
+        // Overload for non-tuple types (std::size_t from stdexec bulk)
+        template <typename T>
+            requires(!is_tuple_like_v<T>)
+        HPX_HOST_DEVICE HPX_FORCEINLINE constexpr Result operator()(T&& t)
+        {
+            return HPX_INVOKE_R(Result, f_, HPX_FORWARD(T, t));
+        }
```
hkaiser, line 55: I think this would be better implemented on the call-site as it would safe having to use `hpx::invoke_fused_r`, see for instance, here:  https://github.com/STEllAR-GROUP/hpx/blob/d586d881474e463a1f1d77be779e8c385331ba07/libs/core/executors/include/hpx/executors/detail/index_queue_spawning.hpp#L80-L101

libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
```diff
@@ -215,7 +215,7 @@ void test_for_each_exception_async(ExPolicy&& p, IteratorTag)
     }
     catch (...)
     {
-        HPX_TEST(false);
+        caught_exception = true;
```
hkaiser, line 222: This is not correct. The exception thrown from the algorithm should be a caught `std::runtime_error` rethrown as `hpx::exception_list`

libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
```diff
@@ -349,12 +350,21 @@ void test_for_each_sender(Policy l, ExPolicy&& p, IteratorTag)
...
+                auto begin_it = rng.begin();
+                return ex::bulk(ex::just(), rng.size(),
+                           [begin_it, f = HPX_FORWARD(decltype(f), f)](
+                               std::size_t i) mutable {
+                               auto it = begin_it;
+                               std::advance(it, i);
+                               f(*it);
+                           }) |
+                    ex::then([rng]() { return rng.end(); });
+            })));
```
hkaiser, line 405: WHat is the rationale for this change? Is this an additional test? If yes, please leave the old test in place and add this as a new one.

libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
```diff
@@ -385,12 +396,27 @@ void test_for_each_exception_sender(Policy l, ExPolicy&& p, IteratorTag)
...
+
+        // If sync_wait returns without exception, check if result indicates error
+        if (!result.has_value())
+        {
+            caught_exception = true;
+        }
+        else
+        {
+            HPX_TEST(false);
+        }
```
hkaiser, line 457: Same here, please leave the old test in place. Add a new one, if needed.

libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
```diff
@@ -399,14 +425,15 @@ void test_for_each_exception_sender(Policy l, ExPolicy&& p, IteratorTag)
     }
     catch (...)
     {
-        HPX_TEST(false);
+        caught_exception = true;
```
hkaiser, line 470: Same comment as above, this modification changes exception handling semantics.

libs/core/executors/include/hpx/executors/scheduler_executor.hpp
```diff
@@ -256,8 +256,12 @@ namespace hpx::execution::experimental {
...
         }
 
-        template <typename F, typename S, typename Future, typename... Ts>
-            requires(!std::is_integral_v<S>)
+        // clang-format off
+        template <typename F, typename S, typename Future, typename... Ts,
+            HPX_CONCEPT_REQUIRES_(
+                !std::is_integral_v<S>
+            )>
+        // clang-format on
```
hkaiser: I think you can revert this change.

libs/core/executors/include/hpx/executors/explicit_scheduler_executor.hpp
```diff
@@ -266,9 +272,11 @@ namespace hpx::execution::experimental {
             auto pre_req =
                 when_all(keep_future(HPX_FORWARD(Future, predecessor)));
 
-            return bulk(transfer(HPX_MOVE(pre_req), exec.sched_), shape,
-                hpx::bind_back(HPX_FORWARD(F, f), HPX_FORWARD(Ts, ts)...));
+            return transfer(HPX_MOVE(pre_req), exec.sched_) |
+                bulk(shape,
+                    hpx::bind_back(HPX_FORWARD(F, f), HPX_FORWARD(Ts, ts)...));
```
hkaiser, line 325: WHat's' the rationale of this change? We deliberately never used the `operator|()` inside HX to reduce compilation overheads.

libs/core/executors/include/hpx/executors/explicit_scheduler_executor.hpp
```diff
@@ -249,9 +249,15 @@ namespace hpx::execution::experimental {
                 hpx::parallel::execution::bulk_async_execute(
                     exec, HPX_FORWARD(F, f), shape, HPX_FORWARD(Ts, ts)...));
         }
+#endif
 
-        template <typename F, typename S, typename Future, typename... Ts>
-            requires(!std::is_integral_v<S>)
+#if !defined(HPX_HAVE_STDEXEC)
```
hkaiser: Hmm, should we really change the base-bulk implementation to not support ranges?

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -46,6 +61,69 @@ namespace hpx::execution::experimental {
         };
     }    // namespace detail
 
+#if defined(HPX_HAVE_STDEXEC)
+    // Forward declarations
+    template <typename Policy>
+    struct thread_pool_policy_scheduler;
+
+    namespace detail {
+    }
```
hkaiser: Can this be removed?

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -46,6 +61,69 @@ namespace hpx::execution::experimental {
...
+                "sender's attributes or receiver's environment "
+                "on which to schedule bulk work.");
+
+            auto sched = hpx::execution::experimental::get_scheduler(env);
+
+            // Extract bulk parameters using structured binding
+            auto&& [tag, data, child] = sndr;
+            auto&& [pol, shape, f] = data;
+
+            auto iota_shape = std::views::iota(decltype(shape){0}, shape);
```
hkaiser: Are iterators retrieved from `std::views::iota` conform to the `std::random_access_iterator`  concept? I know the range conforms to `std::ranges::random_access_range`, but does this imply the former? Please verify.

As an alternative you could use `hpx::util::counting_shape(min, max)` instead (https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/iterator_support/include/hpx/iterator_support/counting_shape.hpp). That is guaranteed to do the right thing.

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -445,10 +482,14 @@ namespace hpx::execution::experimental::detail {
...
-                static_cast<std::uint32_t>(op_state->num_worker_threads), size);
-            std::uint32_t num_chunks = (size + chunk_size - 1) / chunk_size;
+            // Calculate chunk size based on execution mode
+            std::uint32_t chunk_size = op_state->is_chunked ?
+                get_bulk_scheduler_chunk_size(
+                    op_state->num_worker_threads, size) :
+                1;
+            std::uint32_t num_chunks = op_state->is_chunked ?
+                (size + chunk_size - 1) / chunk_size :
+                size;
```
hkaiser: Isn't the operation state templatized on whether it is chunked? Couldn't you do:
```
if constexpr (OperationState::is_chunked)
```

libs/core/algorithms/tests/unit/container_algorithms/foreach_tests.hpp
charan-003: This was done to test the stdexec bulk sender functionality instead of the traditional HPX for_each algorithm.
i will create a new test that uses the stdexec bulk implementation with manual iteration

libs/core/executors/include/hpx/executors/explicit_scheduler_executor.hpp
charan-003: this change restricts the range based `bulk_then_execute` to only the non-stdexec path. The `#if !defined(HPX_HAVE_STDEXEC)` guard was added because the stdexec path uses different bulk semantics that expect integral shapes, while HPX's traditional implementation supports both integral and range shapes.

Should I remove the `#if !defined(HPX_HAVE_STDEXEC)` guard so this overload is available in both paths? Or is there a different approach you'd prefer

hkaiser: While we try to be as close to the standard as possible, occasionally we extend what the standard does. This is one of those cases. We definitely should support the case where an integral argument is passed, but I'd like to keep the existing semantics in place, if possible, even when using `std::execution`.

charan-003: sure, let me do that

libs/core/executors/include/hpx/executors/explicit_scheduler_executor.hpp
```diff
@@ -83,12 +82,8 @@ namespace hpx::execution::experimental {
...
         }
 
-        // clang-format off
-        template <typename Parameters,
-            HPX_CONCEPT_REQUIRES_(
-                hpx::traits::is_executor_parameters_v<Parameters>
-            )>
-        // clang-format on
+        template <typename Parameters>
+            requires(hpx::traits::is_executor_parameters_v<Parameters>)
```
hkaiser: As said before, in our experience, `requires` clauses don't always play well with return type SFIANE. I'd suggest leaving the  `HPX_CONCEPT_REQUIRES_` macro in place in these cases.

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -8,21 +8,36 @@
...
+#include <hpx/concepts/concepts.hpp>
+#include <hpx/errors/try_catch_exception_ptr.hpp>
+#include <hpx/execution_base/completion_scheduler.hpp>
+#include <hpx/execution_base/completion_signatures.hpp>
+#include <hpx/execution_base/receiver.hpp>
+#include <hpx/execution_base/sender.hpp>
+#include <hpx/executors/current_executor.hpp>
+#include <hpx/executors/exception_list.hpp>
+#include <hpx/functional/detail/invoke.hpp>
+#include <hpx/functional/invoke_result.hpp>
```
hkaiser: Please revert the changes regarding the #includes to master. Otherwise you will break C++ module generation.

libs/core/executors/include/hpx/executors/explicit_scheduler_executor.hpp
charan-003: I tried using `HPX_CONCEPT_REQUIRES_` but it causes compilation errors with friend functions that have trailing return types. The compiler gives errors like "cannot combine with previous declaration specifier" no matter where I place the macro.

The C++20 requires clause is what scheduler_executor.hpp uses for the same pattern (line 260), so I followed that approach. Is there a way to make `HPX_CONCEPT_REQUIRES_` work in this case ?

charan-003:
@hkaiser, in this commit, switched to `HPX_CONCEPT_REQUIRES_ `, and removed integral constraints from bulk operations to support both integral and range shapes everywhere

charan-003:
@hkaiser 

removing `#if !defined(HPX_HAVE_STDEXEC)` causing compilation errors on macOS CI because the current implementation assumes S is always a range (uses `range_traits<S>::value_type` 

to support both integral and range shapes, should I keep two separate overloads or use `if constexpr` within a single overload to handle both cases? or any different approach to solve this?

libs/core/executors/include/hpx/executors/explicit_scheduler_executor.hpp
hkaiser: If everything works with `requires` clauses - great. No need to use SFINAE in that case. We have seen issues with this pattern before, that's all.

hkaiser:
> @hkaiser
> 
> removing `#if !defined(HPX_HAVE_STDEXEC)` causing compilation errors on macOS CI because the current implementation assumes S is always a range (uses `range_traits<S>::value_type`
> 
> to support both integral and range shapes, should I keep two separate overloads or use `if constexpr` within a single overload to handle both cases? or any different approach to solve this?

We usually use two overloads instead of constexpr if. But it's up to you to do whatever you feel fits best.

libs/core/algorithms/include/hpx/parallel/util/detail/partitioner_iteration.hpp
```diff
@@ -16,32 +16,78 @@
...
+
+    template <typename T>
+    struct is_tuple_like<T,
+        std::void_t<decltype(hpx::tuple_size<std::decay_t<T>>::value)>>
+      : std::true_type
+    {
+    };
+
+    template <typename T>
+    inline constexpr bool is_tuple_like_v = is_tuple_like<T>::value;
```
hkaiser: We do already have `is_tuple_like` here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/datastructures/include/hpx/datastructures/traits/is_tuple_like.hpp

libs/core/execution_base/include/hpx/execution_base/stdexec_forward.hpp
```diff
@@ -130,25 +134,37 @@ namespace hpx::execution::experimental {
...
+    using stdexec::bulk_unchunked;
+    using stdexec::bulk_unchunked_t;
+
+    // Execution policies
+    using stdexec::is_execution_policy;
+    using stdexec::is_execution_policy_v;
+    using stdexec::par;
+    using stdexec::par_unseq;
+    using stdexec::seq;
+    using stdexec::unseq;
```
hkaiser: All of these most likely need to be `export`ed as well
```suggestion
    HPX_CXX_CORE_EXPORT using stdexec::bulk;
    HPX_CXX_CORE_EXPORT using stdexec::bulk_chunked;
    HPX_CXX_CORE_EXPORT using stdexec::bulk_chunked_t;
    HPX_CXX_CORE_EXPORT using stdexec::bulk_t;
    HPX_CXX_CORE_EXPORT using stdexec::bulk_unchunked;
    HPX_CXX_CORE_EXPORT using stdexec::bulk_unchunked_t;

    // Execution policies
    HPX_CXX_CORE_EXPORT using stdexec::is_execution_policy;
    HPX_CXX_CORE_EXPORT using stdexec::is_execution_policy_v;
    HPX_CXX_CORE_EXPORT using stdexec::par;
    HPX_CXX_CORE_EXPORT using stdexec::par_unseq;
    HPX_CXX_CORE_EXPORT using stdexec::seq;
    HPX_CXX_CORE_EXPORT using stdexec::unseq;
```

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -68,6 +85,16 @@ namespace hpx::execution::experimental::detail {
         HPX_INVOKE(HPX_FORWARD(F, f), HPX_FORWARD(T, t), hpx::get<Is>(ts)...);
     }
 
+    // For bulk_chunked: f(start, end, ...)
+    template <std::size_t... Is, typename F, typename Start, typename End,
```
hkaiser:
```suggestion
    HPX_CXX_CORE_EXPORT template <std::size_t... Is, typename F, typename Start, typename End,
```
?

hkaiser:
@charan-003, @isidorostsa Thanks for working on this! It was a long journey, but you made it!

PR #6748, opened 2025-07-24
Allowing for the fork_join_executor to be used on the main thread

hkaiser:


hkaiser:
The remaining issues are addressed here: #6749

PR #6749, opened 2025-07-25
Reducing the size of CirclCI tasks

hkaiser:


hkaiser:
This solves the CirclCI issues

PR #6750, opened 2025-07-26
Fix missing instantiation of print_dec<unsigned long> in print.cpp

kylehaokunwu:
Fixes # (N/A - no open issue)

## Proposed Changes

Add missing explicit template instantiation for `print_dec<unsigned long>` in `print.cpp`.

Without this instantiation, the `numa_allocator_test` fails to link on macOS (x86_64) due to an undefined symbol error. Other types like `std::int64_t` and `std::uint64_t` are already explicitly instantiated, so this brings `unsigned long` in line with those.

This resolves an intermittent platform-specific build failure encountered while building all unit tests.

## Checklist

- [x] I have fixed a bug.
- [ ] I have added a regression test. (N/A – existing test already covers it)


hkaiser:
@kylehaokunwu thanks for this patch. May I ask you to replace `unsigned long` with `std::uint32_t` to avoid ambiguities across various platforms (see here for instance: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/18397/workflows/448f4fce-b7db-4893-97a8-6ce245f6c2fb/jobs/414402)?

hkaiser:
@kylehaokunwu Out of curiosity, on what platform did you see the problem you tried to address?

kylehaokunwu:
> @kylehaokunwu Out of curiosity, on what platform did you see the problem you tried to address?

I’m currently building on macOS using LLVM Clang. I tried replacing unsigned long with std::uint32_t as you suggested, but the linker still failed, since the call site is still passing an unsigned long, and there’s no matching instantiation for that. So I suspect the actual fix is to change the type at the call site to match one of the explicitly instantiated versions.

I’ll take a closer look tomorrow and update the patch accordingly.

PR #6768, opened 2025-08-22
Extend inspect checks to cover .ixx module interface files

kylehaokunwu:
## Proposed Changes

  - Added `.ixx` to registered file extensions in relevant inspect checks
  (apple_macro_check, ascii_check, assert_macro_check,
  deprecated_*_check, end_check, endline_whitespace_check,
  include_check, length_check, minmax_check, tab_check,
  unnamed_namespace_check, windows_macro_check).
- Updated header_inspector in `inspect.cpp` to treat `.ixx` as header-like.

This change is necessary as HPX is adapting C++20 modules by introducing a new file type with .ixx extension. 

## Checklist


hkaiser (approved):
LGTM, thanks!

PR #6770, opened 2025-08-24
Generate the list of module HPX headers that expose C++ module exports

hkaiser:
@kylehaokunwu Please have a look. This generates the MIU file by inserting the list of HPX module headers that expose export definitions.


kylehaokunwu:
Thanks for this! Just to confirm my understanding: with `GLOBAL_HEADER_MODULE_GEN = ON`, HPX modules I expose will automatically be included in the generated umbrella .ixx. But for standard library headers used inside those modules, I’d still need to add them manually to the global module fragment in `hpx_core.ixx.in` to avoid ODR issues — is that correct?

hkaiser:
> Thanks for this! Just to confirm my understanding: with `GLOBAL_HEADER_MODULE_GEN = ON`, HPX modules I expose will automatically be included in the generated umbrella .ixx. But for standard library headers used inside those modules, I’d still need to add them manually to the global module fragment in `hpx_core.ixx.in` to avoid ODR issues — is that correct?

Yes, we probably should move the standard headers into a separate #include. I might just do that. Originally I wanted to generate that list as well, but was not able to identify all standard headers without maintaining an explicit list of all existing ones.

hkaiser:
> > Thanks for this! Just to confirm my understanding: with `GLOBAL_HEADER_MODULE_GEN = ON`, HPX modules I expose will automatically be included in the generated umbrella .ixx. But for standard library headers used inside those modules, I’d still need to add them manually to the global module fragment in `hpx_core.ixx.in` to avoid ODR issues — is that correct?
> 
> Yes, we probably should move the standard headers into a separate #include. I might just do that. Originally I wanted to generate that list as well, but was not able to identify all standard headers without maintaining an explicit list of all existing ones.

@kylehaokunwu I have now added a CMake script that parses all header files and extracts the list of included standard headers automatically. This way there is no need to manually keep this list up to date.

hkaiser:
The remaining issues are unrelated. Merging...

PR #6773, opened 2025-08-28
Expose format module using C++20 modules

kylehaokunwu:
Fixes #

## Proposed Changes

  - Adapted header files to use appropriate export macros
  - Updated format module `CMakeLists.txt` to enable module generation
  - Modified `hpx_core.ixx` to include format module headers (will be reverted once #6770  is merged)

This is part of the ongoing GSoC project to expose HPX using C++20 modules. This PR will serve as a foundation of how template functions should be exported using C++20 modules.

## Checklist

### DIFF
--- a/libs/core/config/include/hpx/config/export_definitions.hpp
+++ b/libs/core/config/include/hpx/config/export_definitions.hpp
@@ -51,20 +51,6 @@
 #endif
 
 ///////////////////////////////////////////////////////////////////////////////
-// C++20 module export definitions
-#if defined(HPX_BUILD_MODULE)
-# undef HPX_CORE_EXPORT
-# define HPX_CORE_EXPORT         /* empty */
-# define HPX_MODULE_EXPORT       export
-# define HPX_MODULE_EXTERN_CORE  export extern "C++"
-# define HPX_NODISCARD_CORE      HPX_MODULE_EXTERN_CORE [[nodiscard]]
-#else
-# define HPX_MODULE_EXPORT       /* empty */
-# define HPX_MODULE_EXTERN_CORE  HPX_CORE_EXPORT
-# define HPX_NODISCARD_CORE      [[nodiscard]] HPX_CORE_EXPORT
-#endif
-
-///////////////////////////////////////////////////////////////////////////////
 #if defined(HPX_EXPORTS) || defined(HPX_FULL_EXPORTS)
 # define  HPX_EXPORT             HPX_SYMBOL_EXPORT
 #else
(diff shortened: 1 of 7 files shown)


kylehaokunwu:
I was blocked by two issues:
1. Redefinition errors from util headers
- When `hpx/util/bad_lexical_cast.hpp`, `from_string.hpp`, and `to_string.hpp` are included alone with `hpx/modules/format.hpp` in `hpx_core.ixx` and still pulled in indirectly by downstream headers, the compiler reports redefinition errors. 
- This doesn’t show up in the version module because that module doesn’t drag in such util headers directly.

2. Linker errors with non-template functions
- Linker errors occur when non-template functions in `hpx/modules/format.hpp` are exported via C++20 modules using `HPX_MODULE_EXTERN_CORE` but their definitions aren't visible to the module interface.
- The version module worked because it only contains non-template functions with definitions in .cpp files, while `hpx/modules/format.hpp` mixes templates (working) and non-templates (broken).

I'm also not sure whether we need to generate a new `hpx/modules/format.hpp` in the build tree (like we did for version) since one already exists in the source tree. Should we reuse the existing one or create a new one to use as a conditional switch between #include format headers and `import HPX.Core`?

I really appreciate your guidance on this!

hkaiser:
> I was blocked by two issues:
> 
> 1. Redefinition errors from util headers
> 
> * When `hpx/util/bad_lexical_cast.hpp`, `from_string.hpp`, and `to_string.hpp` are included alone with `hpx/modules/format.hpp` in `hpx_core.ixx` and still pulled in indirectly by downstream headers, the compiler reports redefinition errors.

Those files might need to be annotated with `export` keywords as well to avoid this problem.

> * This doesn’t show up in the version module because that module doesn’t drag in such util headers directly.

Right.

> 2. Linker errors with non-template functions
> 
> * Linker errors occur when non-template functions in `hpx/modules/format.hpp` are exported via C++20 modules using `HPX_MODULE_EXTERN_CORE` but their definitions aren't visible to the module interface.

Why are those not visible? Doesn't the `hpx_core.ixx` #include the format module header?

> * The version module worked because it only contains non-template functions with definitions in .cpp files, while `hpx/modules/format.hpp` mixes templates (working) and non-templates (broken).
> 
> I'm also not sure whether we need to generate a new `hpx/modules/format.hpp` in the build tree (like we did for version) since one already exists in the source tree. Should we reuse the existing one or create a new one to use as a conditional switch between #include format headers and `import HPX.Core`?

I think we should generate a `<hpx/modules/format.hpp>` file (and move the content of the current file to `hpx/format/api.hpp` or similar) as this automates the whole module magic. That file will have the #if` required to switch between #including all format module headers and importing the C++ MIU. It will also automatically have that list of all format module headers as listed in the CMakeLists.txt.

hkaiser:
@kylehaokunwu I have rebased your branch onto some local work to adapt the `itt_notify` and `debugging` modules and that fixed some module adaptaion problems leftover from earlier PRs. This may look overwhelming, but the actual changes I made to your work are minimal: adding more module export macros, changed all references to the format module to #include <hpx/modules/format.hpp>, and changes to the CMakeLists.txt for the format test as described above.

hkaiser:
@kylehaokunwu I think this is ready now. Could you please verify that I didn't break any of your code? Also, I think this will resolve the issue you tried to solve by creating #6750. May I ask you to verify whether this is now resolved as well?

kylehaokunwu:
> @kylehaokunwu I have rebased your branch onto some local work to adapt the `itt_notify` and `debugging` modules and that fixed some module adaptaion problems leftover from earlier PRs. This may look overwhelming, but the actual changes I made to your work are minimal: adding more module export macros, changed all references to the format module to #include <hpx/modules/format.hpp>, and changes to the CMakeLists.txt for the format test as described above.

Sorry for the radio silence—my laptop broke last week and I had to wait a couple of days for the new one to arrive. I’ve reviewed the changes you made to the format module, and I’ll use them as a reference when working on the next one. Since I lost all my files, including my meeting notes, could you let me know which module I should work on next?

> @kylehaokunwu I think this is ready now. Could you please verify that I didn't break any of your code? Also, I think this will resolve the issue you tried to solve by creating https://github.com/STEllAR-GROUP/hpx/pull/6750. May I ask you to verify whether this is now resolved as well?

Yes HPX and all the tests build successfully on my new MacBook. Thank you for the fix!

hkaiser:
> could you let me know which module I should work on next?

The `timing` HPX module is low in the module hierarchy and should be straightforward to adapt.

PR #6774, opened 2025-08-28
Cleaning up macro names and usage required in the context of C++ modules

hkaiser:
@kylehaokunwu This mostly cleans up the current state of affairs without adding anything significant. You may have to touch up on https://github.com/STEllAR-GROUP/hpx/pull/6773 to resolve conflicts. I apologize for this inconvenience.


kylehaokunwu:
Thank you for cleaning things up! I will base off on this when working with the rest of modules. 

Just to confirm my understanding of how `SCAN_FOR_MODULES` works:

- For libraries, executables, and components, `SCAN_FOR_MODULES` is an explicit parameter. If it’s ON, the target gets `CXX_SCAN_FOR_MODULES`; if it’s OFF (the default), the target instead defines `HPX_BINARY_DOESNT_USE_CXX_MODULES` so headers fall back to #include. 

- For modules added with `add_hpx_module`, I don’t need to pass `SCAN_FOR_MODULES` explicitly. Instead, we rely on `GLOBAL_HEADER_MODULE_GEN` to set `CMAKE_CXX_SCAN_FOR_MODULES=ON` for that module and restores the previous value afterwards.

Is this the correct way to think about how `SCAN_FOR_MODULES` fits into the overall build system?

hkaiser:
> Thank you for cleaning things up! I will base off on this when working with the rest of modules.
> 
> Just to confirm my understanding of how `SCAN_FOR_MODULES` works:
> 
> * For libraries, executables, and components, `SCAN_FOR_MODULES` is an explicit parameter. If it’s ON, the target gets `CXX_SCAN_FOR_MODULES`; if it’s OFF (the default), the target instead defines `HPX_BINARY_DOESNT_USE_CXX_MODULES` so headers fall back to #include.
> * For modules added with `add_hpx_module`, I don’t need to pass `SCAN_FOR_MODULES` explicitly. Instead, we rely on `GLOBAL_HEADER_MODULE_GEN` to set `CMAKE_CXX_SCAN_FOR_MODULES=ON` for that module and restores the previous value afterwards.
> 
> Is this the correct way to think about how `SCAN_FOR_MODULES` fits into the overall build system?

Yes, all of the above is correct.

hkaiser:
Let's merge this now. Things look reasonably good.

PR #6775, opened 2025-08-28
Expose an explicit CMake target HPX::init to be used by non-executables

hkaiser:
- This is necessary to successfully use hpx::runtime_manager in shared libraries and similar

@SAtacker Would you mind verifying if this solves the issue you reported?


SAtacker:
Works, Thank you!

PR #6778, opened 2025-09-02
Add missed linker flags for hwloc on Apple platforms

### LINKED ISSUE #6772: Broken bundle HPX + vcpkg + macOS since v1.10
This issue was previously reported in #6501 but remains unresolved. As @pingu-73 indicated, the linker flags for macOS must include `-framework CoreFoundation` when `hwloc` is utilized.

Several years ago, the necessary amendment was correctly implemented in the vcpkg repository via a direct patch for the HPX port (see [fix-dependency-hwloc.patch](https://github.com/microsoft/vcpkg/blob/87179c1b4333a4086f8a40f1340d95a896fc7550/ports/hpx/fix-dependency-hwloc.patch)). However, @hkaiser has inadvertently removed this fix in his subsequent [commit](https://github.com/microsoft/vcpkg/commit/37630acd98740f98d7e51da78c9630758835037f). This change should be reverted to restore the correct linkage.

olegrog:
Fixes #6772

## Checklist


hkaiser:
@olegrog Many thanks for this fix!

hkaiser:
The MacOS problems reported here are unrelated.

hkaiser:
@olegrog Thanks again for fixing this issue!

PR #6780, opened 2025-09-05
Exposing HPX type_support module as a C++ module

hkaiser:


hkaiser:
gcc-15 seems to have issues with the constructs we use to untroduce modules, still. Otherwise this looks promising already.

hkaiser:
Merging this now as it is an important intermediate step regarding adding module support.

PR #6784, opened 2025-10-05
Add executors documentation

dimitraka:
Added the documentation in our manual. I think it would be nice to add some figures from the paper for the LULESH use case, but just wanted to check with you before I do so.


hkaiser:
> Added the documentation in our manual. I think it would be nice to add some figures from the paper for the LULESH use case, but just wanted to check with you before I do so.

Yes, that's definitely a good idea.

dimitraka:
I have added the example of the custom executor with a parallel algorithm. Let me know if it looks ok

PR #6789, opened 2025-10-14
contracts module

AlexPapadakis:
Fixes #

## Proposed Changes

  -added contracts module
  -with fallback to asserts when contracts are not supported
  -and option to enchance asserts with contract asserts

more on the GSoC report here:
https://alexpapadakis.github.io/hpx_contracts/
## Checklist

### DIFF
--- a/libs/core/async_mpi/include/hpx/async_mpi/mpi_future.hpp
+++ b/libs/core/async_mpi/include/hpx/async_mpi/mpi_future.hpp
@@ -44,7 +44,7 @@ namespace hpx::mpi::experimental {
 
     // by convention the title is 7 chars (for alignment)
     HPX_CXX_EXPORT using print_on = debug::enable_print<false>;
-    HPX_CXX_EXPORT static constexpr print_on mpi_debug("MPI_FUT");
+    HPX_CXX_EXPORT inline constexpr print_on mpi_debug("MPI_FUT");
 
     namespace detail {
 
@@ -59,7 +59,10 @@ namespace hpx::mpi::experimental {
         // An implementation of future_data for MPI
         HPX_CXX_EXPORT struct future_data : hpx::lcos::detail::future_data<int>
         {
-            HPX_NON_COPYABLE(future_data);
+            future_data(future_data const&) = delete;
+            future_data(future_data&&) = delete;
+            future_data& operator=(future_data const&) = delete;
+            future_data& operator=(future_data&&) = delete;
 
             using init_no_addref =
                 typename hpx::lcos::detail::future_data<int>::init_no_addref;
@@ -112,7 +115,7 @@ namespace hpx::mpi::experimental {
 
         // -----------------------------------------------------------------
         // a convenience structure to hold state vars
-        // used extensivey with debug::print to display rank etc
+        // used extensively with debug::print to display rank etc
         HPX_CXX_EXPORT struct mpi_info
         {
             bool error_handler_initialized_ = false;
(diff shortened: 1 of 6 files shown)


hkaiser:
@AlexPapadakis could you please take care of the clang-format, cmake-format, and inspect errors reported?

AlexPapadakis:
@hkaiser  yes absolutely, I will spend time on it in the Weekend

libs/core/contracts/include/hpx/modules/contracts.hpp
```diff
@@ -0,0 +1,58 @@
...
+#ifdef HPX_HAVE_CONTRACTS
+#if HPX_HAVE_NATIVE_CONTRACTS
+// Native C++26 contracts mode
+#define HPX_PRE(x) pre(x)
+#define HPX_CONTRACT_ASSERT(x) contract_assert(x)
+#define HPX_POST(x) post(x)
+#else
+// Fallback mode: PRE/POST become no-ops for forward compatibility,
+// CONTRACT_ASSERT maps to HPX_ASSERT for runtime validation
+#pragma message(                                                               \
```
isidorostsa: Do we use this in other places too?

libs/core/contracts/include/hpx/modules/contracts.hpp
```diff
@@ -0,0 +1,58 @@
...
+/// See docs/index.rst for comprehensive usage guide.
+
+#pragma once
+
+#include <hpx/config.hpp>
+#include <hpx/assert.hpp>
+
+// Contract implementation: automatically selects native C++26 contracts
+// or provides appropriate fallback behavior based on compiler capabilities
+#ifdef HPX_HAVE_CONTRACTS
```
isidorostsa: I think we could simplify by always using contracts, and choosing between native or not based on the `HPX_HAVE_NATIVE_CONTRACTS` flag

isidorostsa: @Pansysk75 thoughts?

libs/core/contracts/include/hpx/modules/contracts.hpp
```diff
@@ -0,0 +1,58 @@
...
+#else
+// Contracts disabled: PRE/POST are no-ops, CONTRACT_ASSERT remains available
+#define HPX_PRE(x)
+#define HPX_CONTRACT_ASSERT(x) HPX_ASSERT((x))
+#define HPX_POST(x)
+#endif
+
+#ifdef HPX_HAVE_ASSERTS_AS_CONTRACT_ASSERTS
+// Override HPX_ASSERT to use contract assertions
+#define HPX_ASSERT(x) HPX_CONTRACT_ASSERT(x)
```
isidorostsa: Will this change the behavior of only the files that include `contracts.hpp`?
If so, would we need to include this in every file we would like to use contracts?

Also, because of transitive includes I think it would be unclear if a file will be using contracts or not.

Unless this is automatically included in everything, in which case this is perfect.

libs/core/contracts/include/hpx/modules/contracts.hpp
Pansysk75: I was also troubled by that.

The purpose of these two is clear:
- `HPX_HAVE_NATIVE_CONTRACTS ` --> Feature flag for compiler support of contracts.
- `HPX_WITH_ASSERTS_AS_CONTRACT_ASSERTS` --> Redirects HPX_ASSERT to the C++ contracts mechanism.

For the third one:
- `HPX_WITH_CONTRACTS`  --> Disables any contract-related HPX feature.
I think I agree that it would be simpler to omit this flag. For example, I cannot think of a case where we would want to completely ignore all `HPX_CONTRACT_ASSERT`s, in the same way that we don't need to disable `HPX_ASSERT`s (or deviate from the normal release/debug build assert semantics)
Another thing, the current name might be misleading, as it does not generally enable/disable C++ contracts, just its HPX integration.

libs/core/contracts/docs/index.rst
```diff
@@ -0,0 +1,158 @@
...
+Configuration
+=============
+
+Enable contracts in CMake::
+
+    cmake -DHPX_WITH_CONTRACTS=ON -DCMAKE_CXX_STANDARD=26
+
+Enable contract-enhanced assertions (optional)::
+
+    cmake -DHPX_WITH_CONTRACTS=ON -DHPX_HAVE_ASSERTS_AS_CONTRACT_ASSERTS=ON
```
Pansysk75: We typically use the `HPX_HAVE_`  prefix for preprocessor flags.
Please stick with `HPX_WITH_` when it comes to CMake variables.

libs/core/contracts/include/hpx/modules/contracts.hpp
AlexPapadakis: It is something that is used in other places but for deprecation warnings.
@Pansysk75
We still warn during configuration with cmake so I can thus remove.

libs/core/contracts/include/hpx/modules/contracts.hpp
AlexPapadakis: @Pansysk75 do you think we should then move the HPX_WITH_ASSERTS_AS_CONTRACT_ASSERTS to the asserts module? or should someone include the contracts header in each file

libs/core/contracts/include/hpx/modules/contracts.hpp
AlexPapadakis: So you suggest to remove the cmake option HPX_WITH_CONTRACTS and then:
a) always use contracts when supported (or the normal asserts when they are not) and let the user change the contacts mode semantic to ignore.
or
b) follow the asserts module logic and disable them in non Debug builds
@Pansysk75 ?

hkaiser:
@AlexPapadakis thank you for working on this! Much appreciated.

PR #6799, opened 2025-11-10
ci: fix exclude regex handling in linux sanitizer workflow

ArivoliR:
Fixes CI failure that was caused by a test that was meant to be excluded.

## Proposed Changes
Replaces the broken multiline `--exclude-regex` command in the Linux sanitizer workflow with a proper Bash variable construction.
Concatenates both test names (`transpose_block_numa` and `1d_wave_equation`) into a single `$EXCLUDE` variable before calling ctest.

The test `tests.examples.quickstart.1d_wave_equation` was intended to be excluded from the sanitizer workflow but was still being executed and failing in CI due to a broken exclude regex.

## Checklist
- [x] Verified CI workflow changes.

### DIFF
--- a/.github/workflows/linux_sanitizers.yml
+++ b/.github/workflows/linux_sanitizers.yml
@@ -47,6 +47,5 @@ jobs:
           ctest \
             --output-on-failure \
             --tests-regex tests.examples \
-            --exclude-regex \
-                "tests.examples.transpose.transpose_block_numa|\
-                tests.examples.quickstart.1d_wave_equation"
+            --exclude-regex tests.examples.transpose.transpose_block_numa \
+            --exclude-regex tests.examples.quickstart.1d_wave_equation


hkaiser:
Wouldn't a similar change be needed for the other CIs that explude some of the tests as well? Also, could you please elaborate why the original regex was 'broken'?

ArivoliR:
The issue arises from Linux shells having the following behaviour:
```
➜  workflows git:(master) echo "tests.examples.quickstart.1d_wave_equation|\
                                tests.examples.transpose.transpose_block_numa"
tests.examples.quickstart.1d_wave_equation|				tests.examples.transpose.transpose_block_numa
```

When a `|\` continuation is followed by indentation or spaces, bash on linux preserves those spaces as literal characters inside the string rather than treating it as a true line continuation. This causes the actual regex passed to ctest to contain unintended whitespace, because of which all the tests except the first line of the multi line regex are skipped.

This issue **does not** arise in macOS workflows. I think this is because the macos shell handles the indentations by trimming it and concatenating the lines unlike the linux shell. 

The new approach of code that I've written, using a variable ensures all patterns are concatenated into a single string before execution, independent of shell behavior.
```
          EXCLUDE="tests.examples.transpose.transpose_block_numa|"
          EXCLUDE+="tests.examples.quickstart.1d_wave_equation"
```
I’ve checked the other CI workflows and most of them use single-line `--exclude-regex` definitions and are unaffected. This issue only occurs in workflows that use multi-line regex on linux workflows.

Please do let me know if you’d like me to apply the same change to the macOS workflow files as well, so that all CI setups stay consistent and work the same way across systems.

ArivoliR:
I noticed that in [`linux_release_static.yml`](https://github.com/STEllAR-GROUP/hpx/blob/master/.github/workflows/linux_release_static.yml), multiple `--exclude-regex` flags are used separately instead of a multiline call, and that approach seems to work reliably. Would you prefer I update this workflow to follow that pattern instead?

```
            --exclude-regex tests.examples.transpose.transpose_block_numa \
            --exclude-regex tests.examples.quickstart.hello_world_1 \
            --exclude-regex tests.examples.quickstart.distributed.tcp.custom_serialization
```

hkaiser:
> When a `|\` continuation is followed by indentation or spaces, bash on linux preserves those spaces as literal characters inside the string rather than treating it as a true line continuation. This causes the actual regex passed to ctest to contain unintended whitespace, because of which all the tests except the first line of the multi line regex are skipped.

Urgs - excellent catch. It seems to be a 'feature' of the github action yml parser as it works just fine on the CircleCI environment.

> I noticed that in [`linux_release_static.yml`](https://github.com/STEllAR-GROUP/hpx/blob/master/.github/workflows/linux_release_static.yml), multiple `--exclude-regex` flags are used separately instead of a multiline call, and that approach seems to work reliably. Would you prefer I update this workflow to follow that pattern instead?

Yes, that seems to be a good idea! Thanks!

ArivoliR:
> Urgs - excellent catch. It seems to be a 'feature' of the github action yml parser as it works just fine on the CircleCI environment.

The macOS workflow behavior still doesn't seem to add up. For example:
```
            --exclude-regex \
          "tests.examples.quickstart.1d_wave_equation|\
          tests.examples.transpose.transpose_block_numa|\
          tests.performance.local.wait_all_timings|\
          ...
```
in [macos_debug_fetch_boost](https://github.com/STEllAR-GROUP/hpx/blob/master/.github/workflows/macos_debug_fetch_boost.yml) works perfectly fine, with the intended test properly excluded. Could it be that the github action yaml parser works differently for macOS and linux?

I’ve updated the workflow to use separate `--exclude-regex` flags as discussed. Thank you for your quick response!

PR #6804, opened 2025-11-15
Added HPX::auto_wrap_main option 

kollanur:
Added the `HPX::auto_wrap_main` CMake target that excludes the need of including `<hpx/hpx_main.hpp>` header file in the main function to start HPX runtime when `HPX::wrap_main` is linked.

### DIFF
--- a/cmake/HPX_SetupDatapar.cmake
+++ b/cmake/HPX_SetupDatapar.cmake
@@ -52,7 +52,7 @@ if("${HPX_WITH_DATAPAR_BACKEND}" STREQUAL "EVE")
     ADVANCED
   )
   hpx_option(
-    HPX_WITH_EVE_TAG STRING "Eve repository tag or branch" "v2023.02.15"
+    HPX_WITH_Eve_TAG STRING "Eve repository tag or branch" "v2023.02.15"
     CATEGORY "Build Targets"
     ADVANCED
   )
--- a/cmake/HPX_SetupEve.cmake
+++ b/cmake/HPX_SetupEve.cmake
@@ -29,7 +29,7 @@ if(("${HPX_WITH_DATAPAR_BACKEND}" STREQUAL "EVE") AND NOT TARGET eve::eve)
     fetchcontent_declare(
       eve
       GIT_REPOSITORY https://github.com/jfalcou/eve.git
-      GIT_TAG ${HPX_WITH_EVE_TAG}
+      GIT_TAG ${HPX_WITH_Eve_TAG}
     )
 
     fetchcontent_getproperties(eve)
--- a/components/component_storage/src/component_module.cpp
+++ b/components/component_storage/src/component_module.cpp
@@ -13,6 +13,8 @@
 
 #include <hpx/components/component_storage/server/migrate_to_storage.hpp>
 
+#include <hpx/config/warnings_prefix.hpp>
+
 ///////////////////////////////////////////////////////////////////////////////
 // Add factory registration functionality.
 HPX_REGISTER_COMPONENT_MODULE()
--- a/components/component_storage/src/component_storage.cpp
+++ b/components/component_storage/src/component_storage.cpp
@@ -16,6 +16,8 @@
 #include <utility>
 #include <vector>
 
+#include <hpx/config/warnings_prefix.hpp>
+
 namespace hpx { namespace components {
     component_storage::component_storage(hpx::id_type target_locality)
       : base_type(hpx::new_<server::component_storage>(target_locality))
--- a/components/component_storage/src/server/component_storage_server.cpp
+++ b/components/component_storage/src/server/component_storage_server.cpp
@@ -11,6 +11,8 @@
 
 #include <vector>
 
+#include <hpx/config/warnings_prefix.hpp>
+
 namespace hpx { namespace components { namespace server {
     component_storage::component_storage()
       : data_(container_layout(find_all_localities()))
--- a/components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/detail/view_element.hpp
+++ b/components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/detail/view_element.hpp
@@ -13,8 +13,8 @@
 #include <hpx/components/containers/partitioned_vector/partitioned_vector_component_decl.hpp>
 #include <hpx/components/containers/partitioned_vector/partitioned_vector_segmented_iterator.hpp>
 #include <hpx/modules/async_base.hpp>
+#include <hpx/modules/runtime_local.hpp>
 #include <hpx/naming_base/id_type.hpp>
-#include <hpx/runtime_local/get_locality_id.hpp>
 
 #include <cstddef>
 #include <cstdint>
--- a/components/containers/partitioned_vector/src/partitioned_vector_component_double.cpp
+++ b/components/containers/partitioned_vector/src/partitioned_vector_component_double.cpp
@@ -16,6 +16,8 @@
 
 #include <vector>
 
+#include <hpx/config/warnings_prefix.hpp>
+
 HPX_REGISTER_PARTITIONED_VECTOR(double)
 
 // an out-of-line definition of a member of a class template cannot have default
--- a/components/containers/partitioned_vector/src/partitioned_vector_component_int.cpp
+++ b/components/containers/partitioned_vector/src/partitioned_vector_component_int.cpp
@@ -16,6 +16,8 @@
 
 #include <vector>
 
+#include <hpx/config/warnings_prefix.hpp>
+
 HPX_REGISTER_PARTITIONED_VECTOR(int)
 using long_long = long long;
 HPX_REGISTER_PARTITIONED_VECTOR(long_long)
--- a/components/containers/partitioned_vector/src/partitioned_vector_component_std_string.cpp
+++ b/components/containers/partitioned_vector/src/partitioned_vector_component_std_string.cpp
@@ -19,6 +19,8 @@
 
 using std_string = std::string;
 
+#include <hpx/config/warnings_prefix.hpp>
+
 HPX_REGISTER_PARTITIONED_VECTOR(std_string)
 
 // an out-of-line definition of a member of a class template cannot have default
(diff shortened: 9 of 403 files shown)


hkaiser:
Some of the CIs are also failing because of an unused argument (https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/18976/workflows/1db954bb-64fe-4130-81a8-15edc2925fbd/jobs/435313).

kollanur:
> Could you add a test verifying this functionality, please? Perhaps here: https://github.com/STEllAR-GROUP/hpx/tree/master/tests/unit/init?

Yes, I will do that.

kollanur:
> Some of the CIs are also failing because of an unused argument (https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/18976/workflows/1db954bb-64fe-4130-81a8-15edc2925fbd/jobs/435313).

I will check those failures and fix them.

hkaiser:
After thinking about this a bit more - we either should make this the default and make sure that the existing code still works (the one that has the `#include <hpx/hpx_main.hpp>`), or alternatively make it a buy-in on a application-per-application basis, preferrably by creating a dependency to a special target.

kollanur:
> > Some of the CIs are also failing because of an unused argument (https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/18976/workflows/1db954bb-64fe-4130-81a8-15edc2925fbd/jobs/435313).
> 
> I will check those failures and fix them.

Fixed it.

kollanur:
> After thinking about this a bit more - we either should make this the default and make sure that the existing code still works (the one that has the `#include <hpx/hpx_main.hpp>`), or alternatively make it a buy-in on a application-per-application basis, preferrably by creating a dependency to a special target.

I have made it default with the exisiting code still working.

hkaiser:
> > After thinking about this a bit more - we either should make this the default and make sure that the existing code still works (the one that has the `#include <hpx/hpx_main.hpp>`), or alternatively make it a buy-in on a application-per-application basis, preferrably by creating a dependency to a special target.
> 
> I have made it default with the exisiting code still working.

It looks like that many tests are now failing. So making it the default wasn't a good idea after all :/

tests/unit/init/auto_wrap_main.cpp
```diff
@@ -0,0 +1,16 @@
+#include <hpx/modules/async_distributed.hpp>
```
hkaiser, line 7: This file misses formatting and the license/copyright notice.

tests/unit/init/CMakeLists.txt
```diff
@@ -5,7 +5,9 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-set(tests manage_runtime)
+set(tests manage_runtime auto_wrap_main)
+
+set(auto_wrap_main_FLAGS NOLIBS DEPENDENCIES HPX::hpx HPX::auto_wrap_main)
```
hkaiser, line 10: Is `NOLIBS` required for using the auto-wrap feature?

wrap/src/hpx_auto_wrap.cpp
```diff
@@ -0,0 +1,143 @@
...
+    if (hpx_start::include_libhpx_wrap)
+        return initialize_main(argc, argv);
+
+    // call main() since hpx_main.hpp is not included
+    return __real_main(argc, argv);
+}
+////////////////////////////////////////////////////////////////////////////////
+#endif
+
+#endif
```
hkaiser: This file is an almost identical duplicate to `src/hpx_wrap.cpp`. Would there be a way to avoid duplications?

wrap/CMakeLists.txt
```diff
@@ -105,13 +160,26 @@ if(MSVC)
...
+
+    get_target_property(_pdb_file_auto hpx_auto_wrap COMPILE_PDB_NAME_${cfg})
+    get_target_property(_pdb_dir_auto hpx_auto_wrap COMPILE_PDB_OUTPUT_DIRECTORY_${cfg})
+    install(
+      FILES ${_pdb_dir_auto}/${_pdb_file_auto}.pdb
+      DESTINATION ${CMAKE_INSTALL_LIBDIR}
+      CONFIGURATIONS ${cfg}
+      COMPONENT runtime
+      OPTIONAL
+    )
```
hkaiser: There is no need to handle the `.pdb` files as those are Windows (MSVC) specific. The auto-wrap feature will not work on that platform.

tests/unit/init/auto_wrap_main.cpp
kollanur: Done.

tests/unit/init/CMakeLists.txt
kollanur: It is required in the unit test because HPX::hpx and HPX::wrap_main are attached to the target by default, I have to specifically mention not to use those. If we use both HPX::wrap_main and HPX::auto_wrap_main, it defaults to HPX::wrap_main functionality and runtime doesn't starts by not including the header in the main.

wrap/src/hpx_auto_wrap.cpp
kollanur: I have to find a way to do it professor.

wrap/CMakeLists.txt
kollanur: Removed them. Done.

wrap/src/hpx_auto_wrap.cpp
kollanur: I am thinking about it.

tests/unit/init/CMakeLists.txt
hkaiser: This is not an issue, just needs proper documentation.

wrap/src/hpx_wrap.cpp
```diff
@@ -5,6 +5,7 @@
 //  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 #include <hpx/config.hpp>
+#include <hpx/hpx_init.hpp>
```
hkaiser, line 8: WHy do we need this #include here?

wrap/src/hpx_wrap.cpp
```diff
@@ -17,14 +18,27 @@ namespace hpx_start {
...
     // initialize_main. Subsequent calls before entering main() are handled
     // by this code.
     HPX_SYMBOL_EXPORT extern bool include_libhpx_wrap;
-    HPX_SYMBOL_EXPORT bool include_libhpx_wrap __attribute__((weak)) = false;
+    HPX_SYMBOL_EXPORT bool include_libhpx_wrap __attribute__((weak)) =
+#if defined(HPX_AUTO_WRAP_MAIN_ACTIVATE)
+        true;
+#else
+        false;
+#endif
```
hkaiser: Let's write this as:
```suggestion
#if defined(HPX_AUTO_WRAP_MAIN_ACTIVATE)
    HPX_SYMBOL_EXPORT bool include_libhpx_wrap __attribute__((weak)) = true;
#else
    HPX_SYMBOL_EXPORT bool include_libhpx_wrap __attribute__((weak)) = false;
#endif
```
making it better readable.

wrap/src/hpx_wrap.cpp
```diff
@@ -17,14 +18,27 @@ namespace hpx_start {
...
+    // The default application name is populated by including hpx/hpx_main.hpp
+    // or if HPX_AUTO_WRAP_MAIN_ACTIVATE is defined.
     HPX_SYMBOL_EXPORT extern const char* app_name_libhpx_wrap;
-    HPX_SYMBOL_EXPORT const char* app_name_libhpx_wrap __attribute__((weak));
+    HPX_SYMBOL_EXPORT const char* app_name_libhpx_wrap __attribute__((weak)) =
+#if defined(HPX_AUTO_WRAP_MAIN_ACTIVATE)
+        HPX_APPLICATION_STRING;
+#else
+        nullptr;
+#endif
```
hkaiser: Why is this necessary? Couldn't we define `HPX_APPLICATION_STRING` similarily to what's done for the exiting case.

wrap/src/hpx_wrap.cpp
kollanur: I have to move it before hpx_start namespace because the macro HPX_APPLICATION_STRING is defined in <hpx/hpx_init.hpp>

wrap/src/hpx_wrap.cpp
kollanur: By including hpx/hpx_main.hpp, the header provided the HPX_APPLICATION_STRING to weak symbol, since we are not going to include the header I have provided the definition here, and also we are referencing it in the initialize_main.

hkaiser: I meant, whether this has to be conditional. Couldn't we always initialize it with the macro?

kollanur: Yes, no need to be conditional, I have checked it and changed it.

hkaiser (dismissed):
Very nice, LGTM, thanks!

wrap/src/hpx_wrap.cpp
hkaiser: It's actually defined in `<hpx/modules/init_runtime_local.hpp>`, there is no need to pull in the heavy-weight `<hpx/hpx_init.hpp>`.

hkaiser:
@kollanur please disregard my last comment about using `<hpx/modules/init_runtime_local.hpp>`. I was focussing on the `HPX_APPLICATION_STRING` macro and missed that you needed `<hpx/init.hpp>` because of the calls to `hpx::init`. It is not sufficient to initialize HPX using `hpx::local::init` as this will not bring up the distributed runtime environment. I think you can safely roll back your latest commit. We can merge this PR once you do.

kollanur:
> @kollanur please disregard my last comment about using `<hpx/modules/init_runtime_local.hpp>`. I was focussing on the `HPX_APPLICATION_STRING` macro and missed that you needed `<hpx/init.hpp>` because of the calls to `hpx::init`. It is not sufficient to initialize HPX using `hpx::local::init` as this will not bring up the distributed runtime environment. I think you can safely roll back your latest commit. We can merge this PR once you do.

ok professor, got it.

PR #6806, opened 2025-11-17
Add CMakePresets.json for better IDE and build system integration

Adityabaskati-weeb:
## What does this PR do?
Adds a [CMakePresets.json](cci:7://file:///c:/Users/baska/hpx/CMakePresets.json:0:0-0:0) file to provide pre-defined build configurations for HPX.

### Key Benefits:
- **Easier Onboarding**: New users can quickly get started with common HPX configurations
- **IDE Integration**: Works seamlessly with IDEs like VS Code, CLion, and Visual Studio
- **Reproducible Builds**: Ensures consistent builds across different environments
- **Documentation**: Serves as living documentation of common build configurations

### Included Presets:
- **release**: Optimized release build
- **debug**: Debug build with debug symbols and stack traces
- **minimal**: Minimal build with essential features
- **full**: Full build with all features (tests, examples, docs)
- **mpi**: MPI-enabled build
- **sanitizers**: Builds with various sanitizers (ASan, TSan, UBSan)
- **coverage**: Code coverage build

### Usage:
```bash
# Configure with a preset
cmake --preset=debug

# Build using a preset
cmake --build --preset=debug

# Run tests
ctest --preset=debug


Requirements:
CMake 3.19 or later (HPX already requires 3.18+)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1240,22 +1240,9 @@ if(HPX_WITH_NETWORKING)
   endif()
 
   hpx_option(
-    HPX_WITH_PARCELPORT_MPI
-    STRING
-    "Enable the MPI based parcelport. Leave empty to auto-detect."
-    ""
+    HPX_WITH_PARCELPORT_MPI BOOL "Enable the MPI based parcelport." OFF
     CATEGORY "Parcelport"
   )
-  if("${HPX_WITH_PARCELPORT_MPI}" STREQUAL "")
-    find_package(MPI QUIET COMPONENTS CXX)
-    if(MPI_FOUND)
-      hpx_set_option(HPX_WITH_PARCELPORT_MPI VALUE ON FORCE)
-      hpx_info("MPI found. Enabling MPI parcelport (Auto-detected).")
-    else()
-      hpx_set_option(HPX_WITH_PARCELPORT_MPI VALUE OFF FORCE)
-      hpx_info("MPI not found. Disabling MPI parcelport (Auto-detected).")
-    endif()
-  endif()
   if(HPX_WITH_PARCELPORT_MPI)
     hpx_add_config_define(HPX_HAVE_PARCELPORT_MPI)
   endif()
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,229 +1,229 @@
 {
-  "version": 4,
-  "cmakeMinimumRequired": {
-    "major": 3,
-    "minor": 19,
-    "patch": 0
-  },
-  "vendor": {
-    "hpx/project": {
-      "description": "HPX-specific presets for common build configurations"
-    }
-  },
-  "configurePresets": [
-    {
-      "name": "default",
-      "displayName": "Default Configuration",
-      "description": "Default HPX configuration suitable for most users. Release build with common features enabled.",
-      "generator": "Ninja Multi-Config",
-      "binaryDir": "${sourceDir}/build/default",
-      "cacheVariables": {
-        "CMAKE_BUILD_TYPE": "Release",
-        "HPX_WITH_TESTS": "ON",
-        "HPX_WITH_EXAMPLES": "ON"
-      }
+    "version": 4,
+    "cmakeMinimumRequired": {
+        "major": 3,
+        "minor": 19,
+        "patch": 0
     },
-    {
-      "name": "minimal",
-      "displayName": "Minimal Configuration",
-      "description": "Minimal HPX build with only core features. Useful for quick builds or when dependencies are limited.",
-      "generator": "Ninja",
-      "binaryDir": "${sourceDir}/build/minimal",
-      "cacheVariables": {
-        "CMAKE_BUILD_TYPE": "Release",
-        "HPX_WITH_TESTS": "OFF",
-        "HPX_WITH_EXAMPLES": "OFF",
-        "HPX_WITH_DOCUMENTATION": "OFF",
-        "HPX_WITH_TOOLS": "OFF"
-      }
+    "vendor": {
+        "hpx/project": {
+            "description": "HPX-specific presets for common build configurations"
+        }
     },
-    {
-      "name": "full",
-      "displayName": "Full Configuration",
-      "description": "Full HPX build with all standard features enabled. Includes tests, examples, and tools.",
-      "generator": "Ninja Multi-Config",
-      "binaryDir": "${sourceDir}/build/full",
-      "cacheVariables": {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: CMakePresets.json [hunk 1]
```diff
@@ -0,0 +1,164 @@
...
+        "CMAKE_CXX_FLAGS": "-O3 -DNDEBUG"
+      }
+    },
+    {
+      "name": "debug",
+      "displayName": "Debug build with debug symbols",
+      "inherits": "base",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "HPX_WITH_DEBUGGING": "ON",
```
Pansysk75: This is not an existing HPX flag

CMakePresets.json
```diff
@@ -0,0 +1,164 @@
...
+    {
+      "name": "full",
+      "displayName": "Full build with all features",
+      "inherits": "base",
+      "cacheVariables": {
+        "HPX_WITH_EXAMPLES": "ON",
+        "HPX_WITH_TESTS": "ON",
+        "HPX_WITH_DOCUMENTATION": "ON",
+        "HPX_WITH_TOOLS": "ON",
+        "HPX_WITH_COMPONENT_GET_FUTURE_FROM_RECEIVE": "ON",
```
Pansysk75: This is not an existing HPX flag

CMakePresets.json
```diff
@@ -0,0 +1,164 @@
...
+        "HPX_WITH_COMPILER_SANITIZERS": "undefined"
+      }
+    },
+    {
+      "name": "coverage",
+      "displayName": "Code coverage build",
+      "inherits": "debug",
+      "cacheVariables": {
+        "HPX_WITH_CXX_STANDARD": "17",
+        "HPX_WITH_COMPILER_COVERAGE": "ON",
```
Pansysk75: This is not an existing HPX flag.
We do have a coverage test running on our CI, you could look at how it's done there:
https://github.com/STEllAR-GROUP/hpx/blob/bf31d85993c467e964c8bf4e396c18d1be4b234c/.jenkins/lsu-test-coverage/batch.sh#L22-L36

However, I would prefer removing this as a preset, as I don't think it would ever be used (apart from our testing CI).

CMakePresets.json
```diff
@@ -0,0 +1,164 @@
...
+        "MPI_CXX_COMPILER": "mpicxx"
+      }
+    },
+    {
+      "name": "asan",
+      "displayName": "Address Sanitizer build",
+      "inherits": "debug",
+      "cacheVariables": {
+        "HPX_WITH_SANITIZERS": "address",
+        "HPX_WITH_COMPILER_SANITIZERS": "address"
```
Pansysk75: This is not an existing HPX flag

CMakePresets.json
```diff
@@ -0,0 +1,164 @@
...
+    {
+      "name": "base",
+      "hidden": true,
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/cmake-build-${presetName}",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Release",
+        "HPX_WITH_MALLOC": "tcmalloc",
+        "HPX_WITH_NETWORKING": "ON",
+        "HPX_WITH_PARCELPORT_MPI": "AUTO",
```
Pansysk75: We don't use "AUTO" in our CMake. Instead the user should leave the variable unset (or empty?), to let a default value be picked:
https://github.com/STEllAR-GROUP/hpx/blob/bf31d85993c467e964c8bf4e396c18d1be4b234c/CMakeLists.txt#L1226-L1229

Pansysk75:
@Adityabaskati-weeb Would it be possible to make those changes in-code? It will be easier to review. Thanks!

Adityabaskati-weeb:
> @Adityabaskati-weeb Would it be possible to make those changes in-code? It will be easier to review. Thanks!

ohk

CMakeLists.txt
```diff
@@ -1224,9 +1224,20 @@ if(HPX_WITH_NETWORKING)
...
     CATEGORY "Parcelport"
   )
+  if("${HPX_WITH_PARCELPORT_MPI}" STREQUAL "")
+    find_package(MPI QUIET COMPONENTS CXX)
+    if(MPI_FOUND)
+      set(HPX_WITH_PARCELPORT_MPI ON)
+    else()
+      set(HPX_WITH_PARCELPORT_MPI OFF)
+    endif()
+  endif()
```
hkaiser: What's your rationale to auto-detect whether the MPI parcelport should be enabled? If we did this, wouldn't we also have to apply a similar logic to all the other parcelports?

Adityabaskati-weeb: The logic is simple: MPI is the common case, and auto-detection usually matches user intent, while GASNet/LCI are more specialized, so users typically prefer explicit control.
You’re right about consistency, though. We could add auto-detection for LCI or Libfabric, but their usage patterns are less predictable, so the defaults may not reflect what users want.
If consistency is the goal, we either extend auto-detection to those parcelports or remove it for MPI and make everything explicit. Which direction would you prefer?

hkaiser: > The logic is simple: MPI is the common case, and auto-detection usually matches user intent, while GASNet/LCI are more specialized, so users typically prefer explicit control. You’re right about consistency, though. We could add auto-detection for LCI or Libfabric, but their usage patterns are less predictable, so the defaults may not reflect what users want. If consistency is the goal, we either extend auto-detection to those parcelports or remove it for MPI and make everything explicit. Which direction would you prefer?

Whatever we do, there has to be a way to control the settings, i.e. both, explicitly enable (if auto-detection fails) or explictly disable the settings (if so desired). I'd be inclined to leaving things as is, even more as MPI might be the default on HPC machines, but on smaller devices, and or Windows/Mac it might not be available by default.

Adityabaskati-weeb: ohk in new PR i will adding-
 informational logging (hpx_info) to the HPX_WITH_PARCELPORT_MPI auto-detection logic.
Now, when the user leaves HPX_WITH_PARCELPORT_MPI empty (default), the build system will explicitly report whether MPI was found and if the parcelport was enabled or disabled.

also trying to solve some checks that failed

Pansysk75:
@Adityabaskati-weeb Hi, if I may make a suggestion, let's keep this pull request limited to the CMakeUserPresets.json, keeping it as simple as we can.
If you have any suggestions for other parts of the CMake configuration, let's discuss that separately.

PR #6813, opened 2025-11-27
Simplify find_package CMake modules

Pansysk75:
1. Uses a CMake policy to silence warnings about \<PACKAGE>_ROOT cache variables being set, instead of forcefully unsetting them. This will silence some of the superfluous warnings on the current CI.
2. Removes the \<Package>_ROOT hint for `find_package`/`find_path`, as it is redundant. Instead replaces it with \<PACKAGE>_ROOT, which was not searched in CMake <3.27
3. Set \<Package>_ROOT in more uniform way (I kept the previous behavior, but perhaps we could always save it in the CMake cache, instead of having it be an unused local variable)

### DIFF
--- a/.jenkins/lsu/env-common.sh
+++ b/.jenkins/lsu/env-common.sh
@@ -14,6 +14,9 @@ if [ "${build_type}" = "Debug" ]; then
 #    configure_extra_options+=" -DHPX_WITH_VERIFY_LOCKS_BACKTRACE=ON"
 fi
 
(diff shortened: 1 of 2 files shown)


Pansysk75:
retest lsu

Pansysk75:
@hkaiser 
Probably ready to merge, although much of the CI didn't complete, due to the timeout issues we are having

hkaiser (approved):
Let's merge this. Thanks!

PR #6814, opened 2025-11-27
Adapting HPX algorithm module to C++ modules

hkaiser:


hkaiser:
This is superseeded by #6817

PR #6816, opened 2025-11-29
Use ccache for Rostam CI

Pansysk75:
I want to see if this will improve the situation on our Rostam CI, which is currently struggling to keep up.

(I noticed ccache was used back when the CI was hosted on Piz Daint: https://github.com/STEllAR-GROUP/hpx/pull/5365)


hkaiser (approved):
LGTM, thanks!

hkaiser:
@Pansysk75 what other compute nodes will be used for this (besides the one allocated by slurm? Do we have to change the job allocation for this to work properly?

Pansysk75:
> @Pansysk75 what other compute nodes will be used for this (besides the one allocated by slurm? Do we have to change the job allocation for this to work properly?

ccache does not rely on a persistent process; it is solely file-based, and all nodes have access to the cache through the network file system.

Btw this doesn't directly help with our current issue, but there's no harm in having it if it helps with compile times.
Let's wait for the cluster situation to be resolved, and then we can assess whether this improves things.

Pansysk75:
retest

Pansysk75:
File system issue on Rostam has been resolved.

PR #6819, opened 2025-12-03
Experiment with CI build on local file system

Pansysk75:


Pansysk75:
Moving the build directory in the local file system doesn't seem to solve the issue

Pansysk75:
File system issue on Rostam has been resolved.

PR #6820, opened 2025-12-03
Fix expected failure in contract tests

Pansysk75:


Pansysk75:
@hkaiser Likely good to merge

hkaiser (commented):
Does this make any difference? I thought `ON` is the same as `true` in CMake scripts...

Pansysk75:
> Does this make any difference? I thought `ON` is the same as `true` in CMake scripts...

It seems the issue just randomly stopped reproducing locally when I made this change. Now I cannot even reproduce it when I revert my change... I will need a second to figure this out

Pansysk75:
@hkaiser This should be it, apologies for my previous confusion.

PR #6824, opened 2025-12-09
Use constexpr and hpx::wait_each in future_reduce test 

vaibhavj2006:
Fixes #

## Proposed Changes

-Replaced #define NUM_THREADS with constexpr std::size_t NUM_THREADS for type safety and compile-time evaluation.
-Updated test loops to use hpx::wait_each to handle futures efficiently as they become ready.
-Improved readability and parallel execution behavior in the future_reduce test case.

### DIFF
--- a/examples/future_reduce/rnd_future_reduce.cpp
+++ b/examples/future_reduce/rnd_future_reduce.cpp
@@ -9,7 +9,6 @@
 #include <hpx/init.hpp>
 #include <hpx/runtime.hpp>
 //
-#include <cmath>
 #include <iostream>
 #include <random>
 #include <utility>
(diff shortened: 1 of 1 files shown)


vaibhavj2006:
I think the CircleCI job ran out of allowed time

hkaiser:
@vaibhavj2006 The idea of the example was to show that `hpx::when_all` will return a future that represents the state of all futures that were passed in. That means that there is no need to invoke another wait operation (in your case `hpx::when_each`) as the futures are guaranteed to be ready once the (outer) lambda is being invoked. What is your rationale of the proposed change?

vaibhavj2006:
@hkaiser You’re right, sir the additional wait_each is redundant since when_all already ensures all futures are ready. My change unintentionally weakened the example’s intent. I’ll revert to the original approach and keep only the constexpr changes.

Issue #6718, opened 2025-06-10
Bad user experience with algorithms interface


K-ballo:
Passing an incorrect argument to a parallel algorithm results in a wall of template spew that's worse than your typical C++ template output. The use of `tag_invoke` in the algorithm's interface erases all the actual type and concept information of the viable calls. For example, making a mistake when calling `for_each` results in a huge wall of tag-invoke related information, including the `for_each` tag, but the `for_each.hpp` file in which the implementation and relevant information is doesn't even appear in the trace.

Making `tag_invoke` an implementation detail rather than the interface would have the additional benefit that each algorithm requirements are based on the algorithm itself, and not on any particular customization. This was a crucial design choice in `std::ranges`, where concept and constraints are still checked even when calling a customized implementation.

K-ballo:
The cause behind the worse than usual error messages is due to the tag fallback mechanism. With this mechanism, an algorithm interface isn't concrete, but it is the interface of [`tag_fallback`](https://github.com/STEllAR-GROUP/hpx/blob/b3ed893e83f02c509b3d904d7cce696a05f7eb26/libs/core/tag_invoke/include/hpx/functional/detail/tag_fallback_invoke.hpp#L268). This means that when the user types `hpx::for_each(` on an IDE all they're told about `for_each` interface is that it takes arguments `args...`, and when they make a mistake they're presented with an error message that does not even mention `for_each.hpp`.

A simple improvement is to define `for_each` concrete interface in `for_each.hpp`, even if all it does is forward to the generic `tag_fallback` mechanism.

This fallback mechanism also allows for [overloading](https://github.com/STEllAR-GROUP/hpx/blob/b3ed893e83f02c509b3d904d7cce696a05f7eb26/libs/core/algorithms/include/hpx/parallel/algorithms/for_each.hpp#L537) (as opposed to/in addition to customization), which means the actual interface isn't even settled and thus cannot be enforced. The use of concepts and constraints will not contribute to better error messages, as the interface has to unconditionally defer to these external, potentially non-existing, overloads.

hkaiser:
I believe we can close this as the tag_invoke-based interfaces have been removed.

PR #6827, opened 2025-12-09
Implement and integrate sized range concepts for parallel container algorithms

### LINKED ISSUE #6602: Align our parallel range implementation with P3179
For details, see: wg21.link/p3179

sleepingeight:
Fixes #6602 (Partially)

## Proposed Changes
This is the very first of the many PRs to come to align HPX's parallel container algorithms implementation to the standard. This involved - 
1. Implementing range concepts - `is_input_range`, `is_forward_range`, `is_bidirectional_range`, `is_random_access_range`, `is_sized_range` which model `input_range`, `forward_range`, `bidirectional_range`, `random_access_range`, `sized_range` from the C++20 standard (see file - **`libs/core/iterator_support/tests/unit/is_range.cpp`**)
2. In case of parallel range algorithms, iterators are required to satisfy `is_random_access_iterator_v`, sentinels - `is_sized_sentinel_for`, ranges - `is_sized_range_v`, `is_random_access_range_v`.
3. Tests are modified to satisfy the above concepts.

## Checklist

### DIFF
--- a/libs/core/algorithms/tests/unit/datapar_algorithms/transform_binary2_datapar.cpp
+++ b/libs/core/algorithms/tests/unit/datapar_algorithms/transform_binary2_datapar.cpp
@@ -29,7 +29,6 @@ void test_transform_binary2()
 void transform_binary2_test()
 {
     test_transform_binary2<std::random_access_iterator_tag>();
-    test_transform_binary2<std::forward_iterator_tag>();
 }
 
 ///////////////////////////////////////////////////////////////////////////////


libs/core/iterator_support/include/hpx/iterator_support/traits/is_range.hpp
```diff
@@ -75,4 +77,115 @@ namespace hpx::traits {
...
+    struct is_sized_range<R,
+        std::enable_if_t<is_range_v<R> &&
+            !disable_sized_range<std::remove_cv_t<R>> &&
+            (util::detail::has_size_member_v<R> ||
+                util::detail::has_size_v<R> ||
+                std::is_array_v<std::remove_reference_t<R>> ||
+                is_sized_sentinel_for_v<range_sentinel_t<R>,
+                    range_iterator_t<R>>)>> : std::true_type
+    {
+    };
```
sleepingeight, line 177: @hkaiser I'm not very sure of this implementation. This is done as per my understanding of sized_range from [this](https://en.cppreference.com/w/cpp/ranges/sized_range.html). Could you please confirm if this is correct.

hkaiser: The best way to ensure this is correct is to write a test that verifies the trait for a set of use cases that it is supposed to cover.

sleepingeight: Done, I've added a few tests and found that array check is redundant.

libs/core/algorithms/include/hpx/parallel/container_algorithms/adjacent_difference.hpp
```diff
@@ -395,9 +395,9 @@ namespace hpx::ranges {
         // clang-format off
             requires (
                 hpx::is_execution_policy_v<ExPolicy> &&
-                hpx::traits::is_iterator_v<FwdIter1> &&
-                hpx::traits::is_iterator_v<FwdIter2> &&
-                hpx::traits::is_sentinel_for_v<Sent, FwdIter1>
+                hpx::traits::is_random_access_iterator_v<FwdIter1> &&
```
hkaiser: Should we really restrict the container algorithms to be usable with random access iterators only? Obviously the parallelization will not perform very well, but everything should be functional for at least forward iterators. Even more as the static asserts below enforce just that.

sleepingeight: As you've pointed, we'll surely have a bad performance for forward iterators, but this PR was intended to make HPX's ranges algorithms follow the standard. Consider the example of using parallel `std::forward_list`, to get the size of the list for partitioning we need to traverse the entire list, which is sequential and then offloading works to different threads seems redundant, but is beneficial in cases where applying the Predicate on the elements is a heavy task. Both POVs, restricting forward iterator cases citing performance reason or enabling them for functionality reason seem good to me. So leaving it up to you to make the final call.



> Even more as the static asserts below enforce just that.

This is something which I've overlooked. I'll make changes modifying all the static asserts.

hkaiser: > This is something which I've overlooked. I'll make changes modifying all the static asserts.

Those can be removed if the `requires()` clauses cover it already. We did have the `is_iterator_v` condition on the requires clauses and the `static_assert(is_random_access_iterator_v)` in the body to improve compilation errors.

If the standard mandates random access iterators for those algorithms now (which I was not aware of), then I think we should adapt the `static_assert` conditions while leaving the `is_iterator_v` in the `requires()` clauses.

sleepingeight: > We did have the is_iterator_v condition on the requires clauses and the static_assert(is_random_access_iterator_v) in the body to improve compilation errors.

I don't think so. Existing algorithms are still using static_assert(is_iterator_v). 

> If the standard mandates random access iterators for those algorithms now (which I was not aware of), then I think we should adapt the static_assert conditions while leaving the is_iterator_v in the requires() clauses.

Got it. Just confirming again, the standard is making use of concepts but we would want to use static asserts because of better compile errors. Is that it?

hkaiser: @sleepingeight: FWIW, @K-ballo has suggested a different way of improving compiler messages here: https://github.com/STEllAR-GROUP/hpx/issues/6718

sleepingeight: Hi @hkaiser, I've gone through the above suggestion. Could you please confirm whether this would work,

<details>
  <summary> Current implementation of any_of algorithm </summary>

```cpp
    // CPO for hpx::ranges::any_of
    HPX_CXX_EXPORT inline constexpr struct any_of_t final
      : hpx::detail::tag_parallel_algorithm<any_of_t>
    {
    private:
        template <typename ExPolicy, typename Rng, typename F,
            typename Proj = hpx::identity>
        // clang-format off
            requires (
                hpx::traits::is_random_access_range_v<Rng> &&
                hpx::traits::is_sized_range_v<Rng> &&
                hpx::parallel::traits::is_projected_range_v<Proj, Rng> &&
                hpx::parallel::traits::is_indirect_callable_v<ExPolicy, F,
                    hpx::parallel::traits::projected_range<Proj, Rng>
                >
            )
        // clang-format on
        friend hpx::parallel::util::detail::algorithm_result_t<ExPolicy, bool>
        tag_fallback_invoke(
            any_of_t, ExPolicy&& policy, Rng&& rng, F f, Proj proj = Proj())
        {
            using iterator_type = hpx::traits::range_iterator_t<Rng>;

            static_assert(hpx::traits::is_forward_iterator_v<iterator_type>,
                "Required at least forward iterator.");

            return hpx::parallel::detail::any_of().call(
                HPX_FORWARD(ExPolicy, policy), hpx::util::begin(rng),
                hpx::util::end(rng), HPX_MOVE(f), HPX_MOVE(proj));
        }

        template <typename ExPolicy, typename Iter, typename Sent, typename F,
            typename Proj = hpx::identity>
        // clang-format off
            requires (
                hpx::traits::is_random_access_iterator_v<Iter> &&
                hpx::traits::is_sized_sentinel_for_v<Sent, Iter> &&
                hpx::parallel::traits::is_projected_v<Proj, Iter> &&
                hpx::parallel::traits::is_indirect_callable_v<ExPolicy, F,
                    hpx::parallel::traits::projected<Proj, Iter>
                >
            )
        // clang-format on
        friend hpx::parallel::util::detail::algorithm_result_t<ExPolicy, bool>
        tag_fallback_invoke(any_of_t, ExPolicy&& policy, Iter first, Sent last,
            F f, Proj proj = Proj())
        {
            static_assert(hpx::traits::is_forward_iterator_v<Iter>,
                "Required at least forward iterator.");

            return hpx::parallel::detail::any_of().call(
                HPX_FORWARD(ExPolicy, policy), first, last, HPX_MOVE(f),
                HPX_MOVE(proj));
        }

        template <typename Rng, typename F, typename Proj = hpx::identity>
        // clang-format off
            requires (
                hpx::traits::is_range_v<Rng> &&
                hpx::parallel::traits::is_projected_range_v<Proj, Rng> &&
                hpx::parallel::traits::is_indirect_callable_v<
                    hpx::execution::sequenced_policy, F,
                    hpx::parallel::traits::projected_range<Proj, Rng>
                >
            )
        // clang-format on
        friend bool tag_fallback_invoke(
            any_of_t, Rng&& rng, F f, Proj proj = Proj())
        {
            using iterator_type = hpx::traits::range_iterator_t<Rng>;

            static_assert(hpx::traits::is_input_iterator_v<iterator_type>,
                "Required at least input iterator.");

            return hpx::parallel::detail::any_of().call(hpx::execution::seq,
                hpx::util::begin(rng), hpx::util::end(rng), HPX_MOVE(f),
                HPX_MOVE(proj));
        }

        template <typename Iter, typename Sent, typename F,
            typename Proj = hpx::identity>
        // clang-format off
            requires (
                hpx::traits::is_iterator_v<Iter> &&
                hpx::traits::is_sentinel_for_v<Sent, Iter> &&
                hpx::parallel::traits::is_projected_v<Proj, Iter> &&
                hpx::parallel::traits::is_indirect_callable_v<
                hpx::execution::sequenced_policy, F,
                    hpx::parallel::traits::projected<Proj, Iter>
                >
            )
        // clang-format on
        friend bool tag_fallback_invoke(
            any_of_t, Iter first, Sent last, F f, Proj proj = Proj())
        {
            static_assert(hpx::traits::is_input_iterator_v<Iter>,
                "Required at least input iterator.");

            return hpx::parallel::detail::any_of().call(
                hpx::execution::seq, first, last, HPX_MOVE(f), HPX_MOVE(proj));
        }
    } any_of{};


```


</details>

<details>
  <summary> New implementation of any_of algorithm </summary>
  
```cpp

    // CPO for hpx::ranges::any_of
    HPX_CXX_EXPORT inline constexpr struct any_of_t final
      : hpx::detail::tag_parallel_algorithm<any_of_t>
    {
    public:
        template <typename ExPolicy, typename Rng, typename F,
            typename Proj = hpx::identity>
        // clang-format off
            requires (
                hpx::traits::is_random_access_range_v<Rng> &&
                hpx::traits::is_sized_range_v<Rng> &&
                hpx::parallel::traits::is_projected_range_v<Proj, Rng> &&
                hpx::parallel::traits::is_indirect_callable_v<ExPolicy, F,
                    hpx::parallel::traits::projected_range<Proj, Rng>
                >
            )
        hpx::parallel::util::detail::algorithm_result_t<ExPolicy, bool> 
        operator() (ExPolicy&& policy, Rng&& rng, F f, Proj proj = Proj()) const {
            return tag_fallback_invoke(
                    *this, HPX_FORWARD(policy), HPX_FORWARD(rng), HPX_FORWARD(f), HPX_FORWARD(proj));
        }

        template <typename ExPolicy, typename Iter, typename Sent, typename F,
            typename Proj = hpx::identity>
        // clang-format off
            requires (
                hpx::traits::is_random_access_iterator_v<Iter> &&
                hpx::traits::is_sized_sentinel_for_v<Sent, Iter> &&
                hpx::parallel::traits::is_projected_v<Proj, Iter> &&
                hpx::parallel::traits::is_indirect_callable_v<ExPolicy, F,
                    hpx::parallel::traits::projected<Proj, Iter>
                >
            )
        // clang-format on
        hpx::parallel::util::detail::algorithm_result_t<ExPolicy, bool>
        operator() (ExPolicy&& policy, Iter first, Sent last,
            F f, Proj proj = Proj()) const {
                return tag_fallback_invoke(
                    *this, HPX_FORWARD(policy), HPX_FORWARD(first), HPX_FORWARD(last), HPX_FORWARD(f), HPX_FORWARD(proj));
            }
        
        template <typename Rng, typename F, typename Proj = hpx::identity>
        // clang-format off
            requires (
                hpx::traits::is_range_v<Rng> &&
                hpx::parallel::traits::is_projected_range_v<Proj, Rng> &&
                hpx::parallel::traits::is_indirect_callable_v<
                    hpx::execution::sequenced_policy, F,
                    hpx::parallel::traits::projected_range<Proj, Rng>
                >
            )
        // clang-format on
        bool operator() (Rng&& rng, F f, Proj proj = Proj()) const {
            return tag_fallback_invoke(*this, HPX_FORWARD(rng), HPX_FORWARD(f), HPX_FORWARD(proj));
        }

        template <typename Iter, typename Sent, typename F,
            typename Proj = hpx::identity>
        // clang-format off
            requires (
                hpx::traits::is_iterator_v<Iter> &&
                hpx::traits::is_sentinel_for_v<Sent, Iter> &&
                hpx::parallel::traits::is_projected_v<Proj, Iter> &&
                hpx::parallel::traits::is_indirect_callable_v<
                hpx::execution::sequenced_policy, F,
                    hpx::parallel::traits::projected<Proj, Iter>
                >
            )
        // clang-format on
        bool operator() (Iter first, Sent last, F f, Proj proj = Proj()) const {
            return tag_fallback_invoke(*this, HPX_FORWARD(first), HPX_FORWARD(last), HPX_FORWARD(f), HPX_FORWARD(proj));
        }
        

    private:
        template <typename ExPolicy, typename Rng, typename F,
            typename Proj = hpx::identity>
        friend hpx::parallel::util::detail::algorithm_result_t<ExPolicy, bool>
        tag_fallback_invoke(
            any_of_t, ExPolicy&& policy, Rng&& rng, F f, Proj proj = Proj())
        {
            using iterator_type = hpx::traits::range_iterator_t<Rng>;

            static_assert(hpx::traits::is_forward_iterator_v<iterator_type>,
                "Required at least forward iterator.");

            return hpx::parallel::detail::any_of().call(
                HPX_FORWARD(ExPolicy, policy), hpx::util::begin(rng),
                hpx::util::end(rng), HPX_MOVE(f), HPX_MOVE(proj));
        }

        template <typename ExPolicy, typename Iter, typename Sent, typename F,
            typename Proj = hpx::identity>
        friend hpx::parallel::util::detail::algorithm_result_t<ExPolicy, bool>
        tag_fallback_invoke(any_of_t, ExPolicy&& policy, Iter first, Sent last,
            F f, Proj proj = Proj())
        {
            static_assert(hpx::traits::is_forward_iterator_v<Iter>,
                "Required at least forward iterator.");

            return hpx::parallel::detail::any_of().call(
                HPX_FORWARD(ExPolicy, policy), first, last, HPX_MOVE(f),
                HPX_MOVE(proj));
        }

        template <typename Rng, typename F, typename Proj = hpx::identity>
        friend bool tag_fallback_invoke(
            any_of_t, Rng&& rng, F f, Proj proj = Proj())
        {
            using iterator_type = hpx::traits::range_iterator_t<Rng>;

            static_assert(hpx::traits::is_input_iterator_v<iterator_type>,
                "Required at least input iterator.");

            return hpx::parallel::detail::any_of().call(hpx::execution::seq,
                hpx::util::begin(rng), hpx::util::end(rng), HPX_MOVE(f),
                HPX_MOVE(proj));
        }

        template <typename Iter, typename Sent, typename F,
            typename Proj = hpx::identity>
        friend bool tag_fallback_invoke(
            any_of_t, Iter first, Sent last, F f, Proj proj = Proj())
        {
            static_assert(hpx::traits::is_input_iterator_v<Iter>,
                "Required at least input iterator.");

            return hpx::parallel::detail::any_of().call(
                hpx::execution::seq, first, last, HPX_MOVE(f), HPX_MOVE(proj));
        }
    } any_of{};
```

</details>


As, @K-ballo mentioned, by having concrete function signature, the error messages have improved drastically. Ex,

```ninja

/Users/surya/hpx/libs/core/algorithms/tests/unit/container_algorithms/test_hpx.cpp:30:17: error: no matching function for call to object of type 'const struct any_of_t'
   30 |     auto res2 = hpx::ranges::any_of(hpx::execution::par, flist.begin(), flist.end(),[](int a){
      |                 ^~~~~~~~~~~~~~~~~~~
/Users/surya/hpx/libs/core/algorithms/include/hpx/parallel/container_algorithms/all_any_none.hpp:848:9: note: candidate template ignored: constraints not satisfied [with ExPolicy = const parallel_policy &, Rng = iterator, F = iterator, Proj = (lambda at /Users/surya/hpx/libs/core/algorithms/tests/unit/container_algorithms/test_hpx.cpp:30:85)]
  848 |         operator() (ExPolicy&& policy, Rng&& rng, F f, Proj proj = Proj()) const {
      |         ^
/Users/surya/hpx/libs/core/algorithms/include/hpx/parallel/container_algorithms/all_any_none.hpp:840:17: note: because 'hpx::traits::is_random_access_range_v<std::__forward_list_iterator<std::__forward_list_node<int, void *> *> >' evaluated to false
  840 |                 hpx::traits::is_random_access_range_v<Rng> &&
      |                 ^
/Users/surya/hpx/libs/core/algorithms/include/hpx/parallel/container_algorithms/all_any_none.hpp:867:9: note: candidate template ignored: constraints not satisfied [with ExPolicy = const parallel_policy &, Iter = iterator, Sent = iterator, F = (lambda at /Users/surya/hpx/libs/core/algorithms/tests/unit/container_algorithms/test_hpx.cpp:30:85), Proj = hpx::identity]
  867 |         operator() (ExPolicy&& policy, Iter first, Sent last,
      |         ^
/Users/surya/hpx/libs/core/algorithms/include/hpx/parallel/container_algorithms/all_any_none.hpp:858:17: note: because 'hpx::traits::is_random_access_iterator_v<std::__forward_list_iterator<std::__forward_list_node<int, void *> *> >' evaluated to false
  858 |                 hpx::traits::is_random_access_iterator_v<Iter> &&
      |                 ^
/Users/surya/hpx/libs/core/algorithms/include/hpx/parallel/container_algorithms/all_any_none.hpp:901:14: note: candidate template ignored: constraints not satisfied [with Iter = parallel_policy, Sent = iterator, F = iterator, Proj = (lambda at /Users/surya/hpx/libs/core/algorithms/tests/unit/container_algorithms/test_hpx.cpp:30:85)]
  901 |         bool operator() (Iter first, Sent last, F f, Proj proj = Proj()) const {
      |              ^
/Users/surya/hpx/libs/core/algorithms/include/hpx/parallel/container_algorithms/all_any_none.hpp:892:17: note: because 'hpx::traits::is_iterator_v<hpx::execution::detail::parallel_policy_shim<hpx::execution::parallel_policy_executor<hpx::launch>, hpx::execution::experimental::default_parameters> >' evaluated to false
  892 |                 hpx::traits::is_iterator_v<Iter> &&
      |                 ^
/Users/surya/hpx/libs/core/algorithms/include/hpx/parallel/container_algorithms/all_any_none.hpp:884:14: note: candidate function template not viable: requires at most 3 arguments, but 4 were provided
  884 |         bool operator() (Rng&& rng, F f, Proj proj = Proj()) const {
      |              ^           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

I can modify all container algorithms to the new implementation if I can get a thumbs up from you.

hkaiser: I should have linked @K-ballo's suggested example: https://github.com/STEllAR-GROUP/hpx/compare/master...6718. This avoids duplicating all functions. It also forwards to the base class `operator()()` instead of going through an additional `tag_invoke` step.

hkaiser: Generally, I think your solution is a good step in the right direction. @K-ballo's suggested change either improves compiler errors and disables overloading or simply replaces one evil with another one (I might miss something, though - please correct me if I'm wrong).

Whatever you do, I'd suggest you focus on one of the algorithms for the time being until we have found a solution that covers all of our use cases.

sleepingeight: Umm, so I don't think I understood the meaning of overloading here, does it mean that we need to be able to invoke/call the`tag_fallback_invoke` functions in [tag_parallel_algorithm](https://github.com/STEllAR-GROUP/hpx/blob/4a4b5bdbebb2025b6852910ccbed96936ef011c4/libs/core/algorithms/include/hpx/parallel/util/detail/sender_util.hpp#L120C12-L120C34) when only few arguments are given to a function, say `any_of`? If so, I don't think my code above can handle those cases because the base class operator() is hidden thereby only considers the concrete function signatures defined in the inherited class.

> @K-ballo's suggested change either improves compiler errors and disables overloading or simply replaces one evil with another one (I might miss something, though - please correct me if I'm wrong).

Right, but my suggested change is not good either (if the above case is true). 

> Whatever you do, I'd suggest you focus on one of the algorithms for the time being until we have found a solution that covers all of our use cases.

Sure. I'll try to spend some more time and come up with a better design.

hkaiser: We need to support having other, independent (possibly global) overloads of `auto tag_invoke(for_each_t, ...)` that should be picked up if none of the main implementations apply. Like for instance: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/for_each.hpp#L211-L244

hkaiser:
To avoid delaying the work on this I'd suggest to separate the two issues at hand: the algorithmic API changes and the compiler error fixes discussed. Let's focus on the former here and postpone the latter.

sleepingeight:
> To avoid delaying the work on this I'd suggest to separate the two issues at hand: the algorithmic API changes and the compiler error fixes discussed. Let's focus on the former here and postpone the latter.

@hkaiser, currently I've modified all the algorithms using the new ranges concepts according to the standard. Now the only to-do is to not use the concepts but use static asserts right? (Asking since this is also related to better compiler errors).

libs/core/algorithms/tests/regressions/count_3646.cpp
```diff
@@ -15,11 +15,13 @@
 #include <cstdint>
 #include <iterator>
 
+#include "../unit/container_algorithms/test_utils.hpp"
```
hkaiser: I'd rather suggest duplicating the facilities you need to "unit/algorithms/test_utils.hpp" instead of referring to the header from a different directory.

libs/core/algorithms/tests/unit/algorithms/all_of.cpp
```diff
@@ -36,6 +36,21 @@ void test_all_of()
...
+{
+    struct proj
+    {
+        //This projection should cause tests to fail if it is not applied
+        //because it causes predicate to evaluate the opposite
+        constexpr std::size_t operator()(std::size_t x) const
+        {
+            return !static_cast<bool>(x);
+        }
+    };
```
hkaiser, line 52: Is this used anywhere?

libs/core/algorithms/tests/unit/algorithms/any_of.cpp
```diff
@@ -36,6 +37,21 @@ void test_any_of()
...
+{
+    struct proj
+    {
+        //This projection should cause tests to fail if it is not applied
+        //because it causes predicate to evaluate the opposite
+        constexpr std::size_t operator()(std::size_t x) const
+        {
+            return !static_cast<bool>(x);
+        }
+    };
```
hkaiser, line 53: Same here, is this used anywhere?

sleepingeight: No, this is not used anywhere else.

libs/core/algorithms/tests/unit/algorithms/all_of.cpp
sleepingeight: No, this is not used anywhere else.

libs/core/algorithms/tests/unit/algorithms/any_of.cpp
hkaiser: I meant to ask whether this is used at all. I didn't see it being utilized anywhere. Can this be removed?

sleepingeight: Ahh, I see, it is being used [here](https://github.com/STEllAR-GROUP/hpx/pull/6827/changes/BASE..9201364f1556ea9ca35477c7772f9daff1bb73e4#diff-616eac91aff0261e5cf918876db33383b89619d72521cdfb79f68c372faee6d5R100) (Line 100 of this file).

hkaiser: I don't see `proj` being used anywhere. What do I miss?

sleepingeight: Proj is being used in for testing on line R60-R62, and R67-R68. You might need to expand the unchanged codeblocks to see this.

libs/core/algorithms/tests/regressions/count_3646.cpp
```diff
@@ -14,12 +14,13 @@
 #include <cstddef>
 #include <cstdint>
 #include <iterator>
+#include "../unit/algorithms/test_utils.hpp"
```
hkaiser: As said, I'd like to avoid using relative paths in `#include` directives. Let's create a new file for whatever you need for this test. Alternatively, if the facilities are needed in one of the tests only, simply include them in the test file itself.

hkaiser:
@sleepingeight Are you done with changes? Should I look at one of the files only for now, if yes - which one?

hkaiser:
@sleepingeight Sorry for the long silence over the holydays. Can we get back to resolving the remaining issues?

sleepingeight:
Hi @hkaiser, apologies for the delay in responses. I did not quite get what is left to be done in this PR, we finalized above that we'll be taking up improving error messages as a different work item. So to summarise, this PR makes use of the implemented sized range concepts using require expression in parallel container algos. You could take a look at `adjacent_difference.hpp` for reference. I'd request you to guide on what needs to be done, and in the meanwhile I'll look into the CI failures. Thanks.

hkaiser:
> Hi @hkaiser, apologies for the delay in responses. I did not quite get what is left to be done in this PR, we finalized above that we'll be taking up improving error messages as a different work item. So to summarise, this PR makes use of the implemented sized range concepts using require expression in parallel container algos. You could take a look at `adjacent_difference.hpp` for reference. I'd request you to guide on what needs to be done, and in the meanwhile I'll look into the CI failures. Thanks.

I understand that. However from what I see the code now requires for the iterators to be random-access (which is ok), but the old static_assert`s in the functions themselves are still checking for forward iterators (or similar). I think those static_assert's can be removed now as they will never fire anyways.

sleepingeight:
Hi @hkaiser, I've updated the documentation, as well as removed unwanted asserts, could you please provide feedback.

sleepingeight:
Hello @hkaiser, sorry for bugging you, but could you please let me know what the next steps are for this PR? TIA.

hkaiser:
@sleepingeight Could you please rebase your branch onto master (and resolve the conflicts)?

cmake/tests/cxx20_std_disable_sized_range.cpp
```diff
@@ -0,0 +1,17 @@
...
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+////////////////////////////////////////////////////////////////////////////////
+
+#include <ranges>
+
+int main()
+{
+    constexpr bool b = std::ranges::disable_sized_range<void>;
```
hkaiser: Are there known platforms that do not implement `std::ranges::disable_sized_range`?

libs/core/algorithms/include/hpx/parallel/algorithms/detail/mismatch.hpp
```diff
@@ -17,7 +17,7 @@
 
 namespace hpx::parallel::detail {
 
-    HPX_CXX_CORE_EXPORT template <typename ExPolicy>
+    HPX_CXX_EXPORT template <typename ExPolicy>
```
hkaiser: Do to recent changes on master, symbols from `libs/core` need to be exported using `HPX_CXX_CORE_EXPORT` instead of `HPX_CXX_EXPORT`. Sorry for this.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -666,9 +662,6 @@ namespace hpx::parallel {
                 static hpx::util::itt::event notify_event("reshape");
                 hpx::util::itt::mark_event e(notify_event);
 #endif
-#if defined(HPX_HAVE_MODULE_TRACY)
-                hpx::tracy::mark_event evt("reshape");
-#endif
```
hkaiser, line 671: Apparently you rebased onto an older commit as these changes seem to be lost on your branch now.

cmake/tests/cxx20_std_disable_sized_range.cpp
sleepingeight: I don't think there are any. I seem to have added it in flow by mistake, as `HPX_WITH_CXX20_STD_DISABLE_SIZED_SENTINEL_FOR` was also present. But I did not make use of it anywhere. Should I remove both (`HPX_WITH_CXX20_STD_DISABLE_SIZED_SENTINEL_FOR` && `HPX_WITH_CXX20_STD_DISABLE_SIZED_RANGE`) of these configs since they are not being used anywhere?

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
sleepingeight: yeah🥲, rebased incorrectly previously which caused those issues. I think they are resolved now.

cmake/tests/cxx20_std_disable_sized_range.cpp
hkaiser: `HPX_WITH_CXX20_STD_DISABLE_SIZED_SENTINEL_FOR` is a leftover from our C++17 times. I think nowadays many of the C++20 feature checks could be removed.

libs/core/algorithms/include/hpx/parallel/container_algorithms/adjacent_difference.hpp
```diff
@@ -351,65 +348,60 @@ namespace hpx::ranges {
...
+            static_assert(hpx::traits::is_forward_iterator_v<RaIter2>,
                 "Required at least forward iterator.");
 
-            return hpx::parallel::detail::adjacent_difference<FwdIter2>().call(
+            return hpx::parallel::detail::adjacent_difference<RaIter2>().call(
                 hpx::execution::seq, first, last, dest, std::minus<>());
         }
 
-        template <typename Rng, typename FwdIter2>
+        template <typename Rng, typename RaIter2>
```
hkaiser, line 371: The sequential version requires forward iterators only.

libs/core/algorithms/include/hpx/parallel/container_algorithms/adjacent_difference.hpp
```diff
@@ -351,65 +348,60 @@ namespace hpx::ranges {
...
+            typename RaIter2>
         // clang-format off
             requires (
                 hpx::is_execution_policy_v<ExPolicy> &&
-                hpx::traits::is_iterator_v<FwdIter1> &&
-                hpx::traits::is_iterator_v<FwdIter2> &&
-                hpx::traits::is_sentinel_for_v<Sent, FwdIter1>
+                hpx::traits::is_random_access_iterator_v<RaIter1> &&
+                hpx::traits::is_random_access_iterator_v<RaIter2> &&
+                hpx::traits::is_sized_sentinel_for_v<Sent, RaIter1>
```
hkaiser, line 397: Isn't a non-sized sentinal sufficient?

libs/core/algorithms/include/hpx/parallel/container_algorithms/adjacent_difference.hpp
```diff
@@ -418,20 +410,15 @@ namespace hpx::ranges {
         // clang-format off
             requires (
                 hpx::is_execution_policy_v<ExPolicy> &&
-                hpx::traits::is_range_v<Rng> &&
-                hpx::traits::is_iterator_v<FwdIter2>
+                hpx::traits::is_random_access_range_v<Rng> &&
+                hpx::traits::is_sized_range_v<Rng> &&
```
hkaiser, line 414: Why do you require a sized range? Wouldn't requiring a random-access range be sufficient?

hkaiser (commented):
Similar comments (as added for `adjacent_difference`) apply to other algorithms as well.

libs/core/algorithms/include/hpx/parallel/container_algorithms/adjacent_difference.hpp
sleepingeight: We require sized range and sized sentinel for because they require that the size of range is calculated in constant time O(1). To parallelise an algorithm, we need the size of the range. Ifcalculating the size takes O(N) time, then parallelising it doesn't break down tasks since the master thread always runs for O(N) duration. Also this is what the standard mandates.

libs/core/algorithms/include/hpx/parallel/container_algorithms/adjacent_difference.hpp
sleepingeight: Sorry, fixing this.

libs/core/algorithms/include/hpx/parallel/container_algorithms/adjacent_difference.hpp
hkaiser: Wouldn't a random-access range give us `O(1)` complexity when calculating its size as well?

sleepingeight: > Wouldn't a random-access range give us O(1) complexity when calculating its size as well?

It is true, but take a look at the below example, 

```cpp
int data[100];
int* begin = data;
std::unreachable_sentinel_t end{};
static_assert(std::ranges::random_access_range<decltype(std::ranges::subrange(begin, end))>);
static_assert(!std::ranges::sized_range<decltype(std::ranges::subrange(begin, end))>);
static_assert(!std::sized_sentinel_for<decltype(end), decltype(begin)>);
```

If a custom iterator implemented for a data structure is in the above format, for example a circular buffer. Then having random access range woudn't be sufficient since we don't have a way to get size in O(1). So I think both concepts complement each other.

hkaiser: Sure, but `std::size(std::ranges::subrange(begin, end))` is still `O(1)`, isn't it?

sleepingeight: `std::size` checks for `.size()` member in the class, i.e. it finds out size in O(1) (although this is implementation defined for a container, but is in general constant time). So, the above statement will not compile since there is no size method in subrange as the iterator and sentinel pair do not satisfy `sized_sentinel_for` concept. For example, this works -

```cpp
#include <ranges>
#include <vector>

int main() {
    std::vector<int> v;
    auto begin = std::ranges::begin(v);
    auto end = std::ranges::end(v);
    static_assert(std::ranges::sized_range<decltype(std::ranges::subrange(begin, end))>);
    auto res = std::size(std::ranges::subrange(begin, end));
}
```

hkaiser: Fair points, thanks for clarifying.

libs/core/algorithms/include/hpx/parallel/container_algorithms/adjacent_difference.hpp
sleepingeight: I've fixed these. Could you PTAL.

hkaiser: I will go over the PR, however because of its sheer size this may take a moment or two.

libs/core/algorithms/include/hpx/parallel/container_algorithms/adjacent_difference.hpp
```diff
@@ -351,65 +348,60 @@ namespace hpx::ranges {
       : hpx::detail::tag_parallel_algorithm<adjacent_difference_t>
     {
     private:
-        template <typename FwdIter1, typename FwdIter2, typename Sent>
+        template <typename RaIter1, typename RaIter2, typename Sent>
```
hkaiser, line 351: Why did you rename the `FwdIter` arguments to `RaIter`? The same question applies to many of the sequential algorithms and to the docs for said APIs.

libs/core/algorithms/include/hpx/parallel/container_algorithms/adjacent_difference.hpp
```diff
@@ -467,12 +454,6 @@ namespace hpx::ranges {
         friend FwdIter2 tag_fallback_invoke(
             hpx::ranges::adjacent_difference_t, Rng&& rng, FwdIter2 dest, Op op)
         {
-            static_assert(hpx::traits::is_forward_iterator_v<
-                              hpx::traits::range_iterator_t<Rng>>,
-                "Required at least forward iterator.");
-            static_assert(hpx::traits::is_forward_iterator_v<FwdIter2>,
-                "Required at least forward iterator.");
```
hkaiser, line 474: I believe either the `static_assert`s must stay or the requires clause needs updating.

libs/core/algorithms/include/hpx/parallel/container_algorithms/adjacent_find.hpp
hkaiser, line 384: This static_assert is not needed anymore.

libs/core/algorithms/include/hpx/parallel/container_algorithms/all_any_none.hpp
hkaiser, line 758: Do we still need this `static_assert`? This applies to many more places for other algorithms.

libs/core/algorithms/include/hpx/parallel/container_algorithms/contains.hpp
```diff
@@ -36,23 +38,23 @@ namespace hpx::parallel::detail {
         {
         }
 
-        template <typename ExPolicy, typename Iterator, typename Sentinel,
+        template <typename ExPolicy, typename RaIterator, typename Sentinel,
             typename T, typename Proj>
-        static constexpr bool sequential(
-            ExPolicy, Iterator first, Sentinel last, T const& val, Proj&& proj)
+        static constexpr bool sequential(ExPolicy, RaIterator first,
+            Sentinel last, T const& val, Proj&& proj)
```
hkaiser, line 44: The sequential implementation doesn't require random-access iterators. The same comment applies to other sequential implementations of other algorithms.

libs/core/algorithms/include/hpx/parallel/container_algorithms/destroy.hpp
```diff
@@ -303,7 +305,7 @@ namespace hpx::ranges {
         // clang-format off
             requires(
                 hpx::is_execution_policy_v<ExPolicy> &&
-                hpx::traits::is_iterator_v<FwdIter> &&
+                hpx::traits::is_random_access_iterator_v<FwdIter> &&
```
hkaiser, line 308: Should this be `RaIter`?

libs/core/algorithms/include/hpx/parallel/container_algorithms/ends_with.hpp
hkaiser, line 430: Do we still need these `static_assert`s

libs/core/algorithms/include/hpx/parallel/container_algorithms/equal.hpp
hkaiser, line 429: Do we still these?

libs/core/algorithms/include/hpx/parallel/container_algorithms/exclusive_scan.hpp
hkaiser, line 458: Do we still need this?

sleepingeight:
Hi @hkaiser, I'm sorry for giving you trouble reviewing this code. I actually ran a python script to modify names and improve documentation, which seems to have failed badly. I'll follow your suggestion and raise seperate algorithms for 10 algorithms each. However I have few questions -
1. Should the new range concepts be implemented in hpx? They are available from c++20 and since we also want to modernise hpx, I'm thinking to not implement them but use the standard ones.
2. We have traits implemented for the iterators too (ex. `std::forward_iterator` -> `hpx::is_forward_iterator`). Although not all of them are in the standard, can I remove the ones which are present in the standard? If yes, then I'll first raise a cleanup PR which removes the usage of these traits and then continue working on this one. Else I can proceed breaking down with this PR.
Could you please suggest on how to take this forward. Thank you.

hkaiser:
> 2\. We have traits implemented for the iterators too (ex. `std::forward_iterator` -> `hpx::is_forward_iterator`). Although not all of them are in the standard, can I remove the ones which are present in the standard? If yes, then I'll first raise a cleanup PR which removes the usage of these traits and then continue working on this one. Else I can proceed breaking down with this PR.

Yes, please use standard facilities wherever possible.

hkaiser:
@sleepingeight should we close this PR as obsolete?

sleepingeight:
sure @hkaiser, I'll close it.

PR #6828, opened 2025-12-10
Distribute chunk calculation for hpx::merge using memoizing iterators

hkaiser:
Let's try if this gives some performance benefit. @kollanur please have a look.

### DIFF
--- a/docs/sphinx/manual/executors.rst
+++ b/docs/sphinx/manual/executors.rst
@@ -419,19 +419,6 @@ Full example code
         // underlying (wrapped) executor.
 
         template <typename BaseExecutor>
-        struct is_never_blocking_one_way_executor<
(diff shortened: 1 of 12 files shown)


kollanur:
> Let's try if this gives some performance benefit. @kollanur please have a look.

I have done some preliminary benchmarking and compared them with the previous ones we have, I did not see any performance improvements.

kollanur:
> > Let's try if this gives some performance benefit. @kollanur please have a look.
> 
> I have done some preliminary benchmarking and compared them with the previous ones we have, I did not see any performance improvements.

I have to check with varying the core counts and see if there is any performance improvemnts.

hkaiser:
This makes performance worse. Discarding...

PR #6829, opened 2025-12-11
Fix deadlock in for_each_on_main_thread regression test

parsa:
The `fork_join_executor` in the second test case was being constructed on the main OS thread but used within an HPX worker thread (via `run_as_hpx_thread`). When constructed on a non-worker thread, the executor initializes its main thread index to -1, but attempting to use it on a worker thread caused it to access internal data with this invalid index, leading to a hang.

This commit moves the executor instantiation inside the `run_as_hpx_thread` lambda to ensure it is correctly associated with the worker thread where execution occurs.

## Fixes bug in `tests.regressions.modules.algorithms.for_each_on_main_thread`

The regression test `tests.regressions.modules.algorithms.for_each_on_main_thread` consistently timed out.

When running the test executable directly with a timeout, it exited with code `124` (timeout) instead of `0` (success).

**Evidence:**
By adding debug prints to the original code, we confirmed the test successfully completed "Test Case 1" but hung indefinitely immediately upon entering "Test Case 2":

```text
Test case 1 start
Inside run_as_hpx_thread 1
Executor constructed 1
for_each done 1
Test case 1 end
Test case 2 start
Inside run_as_hpx_thread 2
... (HANG / TIMEOUT) ...
```

### Why it was Undesirable (Root Cause)
The failure was due to an undefined behavior in `fork_join_executor` when its **construction thread** differs from its **execution thread** in a specific way:

1.  **Construction on Main Thread:** In Test Case 2, `fork_join_executor` was constructed on the main OS thread. Since the main thread was not part of the HPX worker pool in this configuration, `hpx::get_worker_thread_num()` returned `-1` (SIZE_MAX). The executor stored this as its `main_thread_` index.
2.  **Execution on Worker Thread:** The test then passed this executor into `hpx::run_as_hpx_thread`, which runs on a valid HPX worker thread (e.g., index `0`).
3.  **The Crash/Hang:** When `hpx::for_each` ran on the worker thread, the executor attempted to access its internal state using the captured `main_thread_` index (`-1`). This resulted in an out-of-bounds access on `region_data_`, causing the observed hang.

### Verification of the Fix
The fix moves the construction of `fork_join_executor` **inside** the `run_as_hpx_thread` lambda.

**Why it works:**
*   **Correct Context:** The executor is now constructed on the actual HPX worker thread (e.g., index `0`) that will execute the task.
*   **Valid Index:** `hpx::get_worker_thread_num()` returns a valid index (`0`), and the executor initializes its state correctly for that worker.
*   **Priority Handling:** The executor correctly inherits the `bound` priority from the `run_as_hpx_thread` context, satisfying the test's intent regarding thread priorities.

**Proof of Success:**
After applying the fix, the test runs to completion instantly without timeout:

```text
Test case 1 start
...
Test case 1 end
Test case 2 start
Inside run_as_hpx_thread 2
for_each done 2       <-- reached success point
Test case 2 end
Exit code: 0
```

## Proposed Changes

  - Move `fork_join_executor` instantiation inside the `run_as_hpx_thread` lambda in `libs/core/algorithms/tests/regressions/for_each_on_main_thread.cpp`.
  - Fixes a deadlock/hang caused by creating the executor on a non-worker thread (where thread index is -1) and using it on a worker thread.

The `fork_join_executor` relies on capturing the `main_thread_` index at construction to coordinate work. When created on the main thread (which is not an HPX worker in this test configuration), `main_thread_` is initialized to `-1` (SIZE_MAX). However, `bulk_sync_execute` is subsequently called from an HPX worker thread. The executor attempts to use the captured `main_thread_` index (still `-1`) to access `region_data_`, leading to undefined behavior and the observed deadlock. Constructing the executor on the worker thread ensures `main_thread_` captures the correct worker index.

## Checklist

### DIFF
--- a/libs/core/algorithms/tests/regressions/for_each_on_main_thread.cpp
+++ b/libs/core/algorithms/tests/regressions/for_each_on_main_thread.cpp
@@ -128,8 +128,8 @@ int main()
         constexpr hpx::launch::async_policy policy(
             hpx::threads::thread_priority::bound);
 
-        hpx::execution::experimental::fork_join_executor exec;
         hpx::run_as_hpx_thread(policy, [&]() {
+            hpx::execution::experimental::fork_join_executor exec;
             invoked.store(0);
 
             hpx::for_each(hpx::execution::par.on(exec), vs.begin(), vs.end(),


hkaiser:
Thanks @parsa! Much appreciated!

parsa:
I tried to reproduce the issue with the current [master HEAD](4a4b5bdbebb2025b6852910ccbed96936ef011c4), but I no longer see the test fail despite multiple attempts. The only caveat is that the test takes 600 seconds to run. Not sure what to make of the failure disappearing.

hkaiser:
> I tried to reproduce the issue with the current [master HEAD](4a4b5bdbebb2025b6852910ccbed96936ef011c4), but I no longer see the test fail despite multiple attempts. The only caveat is that the test takes 600 seconds to run. Not sure what to make of the failure disappearing.

Thanks @parsa. Let's close this for now and reopen the issue if needed.

PR #6831, opened 2025-12-11
Fix linker error for print_dec<unsigned long> on ARM64

parsa:
Add explicit template instantiations for `long` and `unsigned long` to resolve undefined symbol errors when using debug output with `unsigned long` values on ARM64 macOS.

On ARM64, `unsigned long` is distinct from `std::uint64_t`, requiring separate template instantiations. This matches the existing pattern for other integer types and is safe on all platforms.

## Fixes bug during linking `numa_allocator_test` on ARM64 macOS

The build on macOS ARM64 (Apple Silicon) failed during the linking phase of the `numa_allocator_test`.

**Evidence:**
The linker explicitly reported an `Undefined symbol` error for a specific template specialization of `print_dec`:

```
Undefined symbols for architecture arm64:
  "void hpx::debug::detail::print_dec<unsigned long>(std::__1::basic_ostream<char, ...>&, unsigned long const&, int)", referenced from:
      void hpx::debug::detail::display<...> in numa_allocator.cpp.o
ld: symbol(s) not found for architecture arm64
```

This is undesirable because:
*   It breaks the build completely on a supported and increasingly common architecture (ARM64).
*   It indicates that the debugging infrastructure (`hpx::debug::dec`) was incomplete, as it could not handle standard C++ fundamental types like `unsigned long` and `long` directly, forcing users to cast variables to `uint64_t` manually to avoid linker errors.

#### Analysis
The `print.cpp` source file previously only contained explicit instantiations for fixed-width types (`std::uint64_t`, `std::int64_t`, etc.):

```cpp
// Existing code
template void print_dec(std::ostream&, std::uint64_t const&, int);On **ARM64 macOS**:
```

1.  `std::uint64_t` is typically a typedef for `unsigned long long`.
2.  `std::size_t` is typically a typedef for `unsigned long`.
3.  In C++, `unsigned long` and `unsigned long long` are **distinct types** for the purpose of template instantiation and function overloading, even if they have the same size (64 bits).

Because the library only instantiated the template for `std::uint64_t` (aka `unsigned long long`), the compiler generated a symbol for that specific type. When `numa_allocator.cpp` invoked `print_dec` with an `unsigned long` (likely via `std::size_t`), the linker looked for the `unsigned long` symbol, which did not exist.

#### Why the fix is correct
We added explicit instantiations for the missing types, guarded by `__APPLE__` to avoid conflicts on other platforms:

```cpp
// Added code
#if defined(__APPLE__)
template void print_dec(std::ostream&, long const&, int);
template void print_dec(std::ostream&, unsigned long const&, int);
#endif
```

**Proof of Correctness:**
1.  **Symbol Availability:** By explicitly instantiating `unsigned long` on macOS, we force the compiler to generate the exact symbol (`__Z...m...` for `unsigned long` vs `__Z...y...` for `unsigned long long`) that the linker was previously complaining was missing.
2.  **Platform Safety:**
    *   **On macOS (ARM64/x86_64):** `unsigned long` is distinct from `std::uint64_t`. The guard enables the instantiation, providing the missing symbol and fixing the linker error.
    *   **On Linux (x86_64):** `std::uint64_t` is often defined as `unsigned long`. If we blindly instantiated both, it would result in a "duplicate explicit instantiation" error. The `#if defined(__APPLE__)` guard ensures we only add the instantiation on platforms where it is strictly required and safe.
3.  **Pattern Consistency:** This change aligns `print_dec` with how standard libraries handle integer overloads, ensuring support for all fundamental integer types.

## Proposed Changes

  - Add `extern template` declarations for `long` and `unsigned long` specializations of `print_dec` in `libs/core/debugging/include/hpx/debugging/print.hpp` (guarded by `__APPLE__`).
  - Add explicit template instantiations for `long` and `unsigned long` specializations of `print_dec` in `libs/core/debugging/src/print.cpp` (guarded by `__APPLE__`).
  - This enables `hpx::debug::dec<N>(val)` to work correctly when `val` is of type `unsigned long` or `long`, fixing build failures on platforms where these are distinct from fixed-width integer types.

The build was failing on ARM64 macOS with an undefined symbol error: `void hpx::debug::detail::print_dec<unsigned long>(...)`. This function is used by `numa_allocator_test`. While `std::uint64_t` was instantiated, `unsigned long` was not. Since `unsigned long` is a distinct type from `uint64_t` on this platform (even if they have the same size), the linker could not find the symbol.

## Checklist

- [x] I have fixed a bug and have added a regression test. (The existing `numa_allocator_test` serves as the regression test).

### DIFF
--- a/libs/core/debugging/include/hpx/debugging/print.hpp
+++ b/libs/core/debugging/include/hpx/debugging/print.hpp
@@ -78,6 +78,12 @@ namespace hpx::debug {
             std::ostream&, std::int64_t const&, int);
         extern template HPX_CORE_EXPORT void print_dec(
             std::ostream&, std::uint64_t const&, int);
+#if defined(__APPLE__)
+        extern template HPX_CORE_EXPORT void print_dec(
+            std::ostream&, long const&, int);
+        extern template HPX_CORE_EXPORT void print_dec(
+            std::ostream&, unsigned long const&, int);
+#endif
 
         extern template HPX_CORE_EXPORT void print_dec(
             std::ostream&, std::atomic<int> const&, int);
--- a/libs/core/debugging/src/print.cpp
+++ b/libs/core/debugging/src/print.cpp
@@ -61,6 +61,10 @@ namespace hpx::debug {
         template void print_dec(std::ostream&, std::uint32_t const&, int);
         template void print_dec(std::ostream&, std::int64_t const&, int);
         template void print_dec(std::ostream&, std::uint64_t const&, int);
+#if defined(__APPLE__)
+        template void print_dec(std::ostream&, long const&, int);
+        template void print_dec(std::ostream&, unsigned long const&, int);
+#endif
 
         template void print_dec(std::ostream&, std::atomic<int> const&, int);
         template void print_dec(


hkaiser (commented):
On this platform, why is `long` different from `int32_t` (or `int64_t`)? I would assume that your change will create ambiguities...

parsa:
The reason is that while on Darwin/Clang `long` is 64-bit On Darwin/Clang, `std::int64_t` is `typedef`’d to `long long` (and `std::uint64_t` to `unsigned long long`). So even though `long` and `int64_t` have the same width (`sizeof(long) == sizeof(int64_t)`), they are distinct types. The existing instantiations for `std::int64_t`/`std::uint64_t` therefore don’t satisfy uses where the caller’s type is spelled `long`/`unsigned long` (e.g. `dec<>(std::size_t))`, hence the existing instantiations don’t produce symbols for `long`/`unsigned long`, and adding them fills in the missing cases.

```console
% cat >prog.cpp <<EOF
#include <cstdint>
#include <type_traits>
#include <iostream>

int main() {
  std::cout << "sizeof(long)      = " << sizeof(long) << "\n";
  std::cout << "sizeof(long long) = " << sizeof(long long) << "\n";
  std::cout << "sizeof(int64_t)   = " << sizeof(std::int64_t) << "\n";

  std::cout << std::boolalpha;
  std::cout << "is_same<long, int64_t> = "
            << std::is_same_v<long, std::int64_t> << "\n";

  std::cout << "is_same<int64_t, long> = "
            << std::is_same_v<std::int64_t, long> << "\n";

  // What int64_t actually aliases on this platform/libc++
  std::cout << "is_same<int64_t, long long> = "
            << std::is_same_v<std::int64_t, long long> << "\n";
  std::cout << "is_same<long, long long> = "
            << std::is_same_v<long, long long> << "\n";
}
static_assert(sizeof(long) == sizeof(std::int64_t));
static_assert(!std::is_same_v<long, std::int64_t>);

EOF

% clang++ -std=c++20 -O2 -Wall -Wextra prog.cpp && ./a.out
sizeof(long)      = 8
sizeof(long long) = 8
sizeof(int64_t)   = 8
is_same<long, int64_t> = false
is_same<int64_t, long> = false
is_same<int64_t, long long> = true
is_same<long, long long> = false

% uname -a
Darwin l631prsam.local 25.1.0 Darwin Kernel Version 25.1.0: Mon Oct 20 19:32:41 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6000 arm64

% clang++ -v
Apple clang version 17.0.0 (clang-1700.6.3.2)
Target: arm64-apple-darwin25.1.0
Thread model: posix
```

hkaiser:
@parsa thanks for the thorough explanations. Usually types like `int64_t` are not distinct types but are introduced as `typedef`s. Could you please find out how they are defined on the platform you're using?

parsa:
`int64_t` is defined inside [`/.../usr/include/sys/_types/_int64_t.h`](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/sys/_types/_int64_t.h#L28-L31)
```cpp
#ifndef _INT64_T
#define _INT64_T
typedef long long               int64_t;
#endif /* _INT64_T */
```

And also under [`/.../usr/include/arm/_types.h`](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/arm/_types.h#L42-L48)
```cpp
typedef unsigned char           __uint8_t;
typedef short                   __int16_t;
typedef unsigned short          __uint16_t;
typedef int                     __int32_t;
typedef unsigned int            __uint32_t;
typedef long long               __int64_t;
typedef unsigned long long      __uint64_t;
```
which is pulled in by [`/.../usr/include/machine/_types.h`](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/machine/_types.h#L34) (NOTE: it's similar for older i386 Darwin: [`/../bsd/i386/_types.h`](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/i386/_types.h#L66-L72))

The include paths are:
- `<cstdint>` → `<stdint.h>` → `<sys/_types/_int64_t.h>` → (uses `__int64_t` from `<machine/_types.h>`
- `<cstdint>` → `<stdint.h>` → `<_types.h>` → `<sys/_types/_int64_t.h>` → `<machine/_types.h>` → `<arm/_types.h>`

parsa:
> In this case, why is `std::is_same_v<std::int64_t, long long>` evaluating to `false`?

Sorry if I misunderstand, but `std::is_same_v<std::int64_t, long long>` does not evaluate to `false`
```
% cat >/tmp/prog.cpp <<EOF
#include <cstdint>
#include <type_traits>
#include <iostream>

int main() {
  std::cout << std::boolalpha
            << "std::is_same_v<std::int64_t, long long> = "
            << std::is_same_v<std::int64_t, long long> << "\n";
}
EOF
clang++ -std=c++20 -O2 -Wall -Wextra /tmp/prog.cpp -o /tmp/a.out && /tmp/a.out

std::is_same_v<std::int64_t, long long> = true
```

hkaiser:
> > In this case, why is `std::is_same_v<std::int64_t, long long>` evaluating to `false`?
> 
> Sorry if I misunderstand, but `std::is_same_v<std::int64_t, long long>` does not evaluate to `false`
> 
> ```
> % cat >/tmp/prog.cpp <<EOF
> #include <cstdint>
> #include <type_traits>
> #include <iostream>
> 
> int main() {
>   std::cout << std::boolalpha
>             << "std::is_same_v<std::int64_t, long long> = "
>             << std::is_same_v<std::int64_t, long long> << "\n";
> }
> EOF
> clang++ -std=c++20 -O2 -Wall -Wextra /tmp/prog.cpp -o /tmp/a.out && /tmp/a.out
> 
> std::is_same_v<std::int64_t, long long> = true
> ```

Ok, I misunderstood. So why do you need a separate specialization for `long long` in addition to the one for `std::int64_t`?

hkaiser:
> > > In this case, why is `std::is_same_v<std::int64_t, long long>` evaluating to `false`?
> > 
> > 
> > Sorry if I misunderstand, but `std::is_same_v<std::int64_t, long long>` does not evaluate to `false`
> > ```
> > % cat >/tmp/prog.cpp <<EOF
> > #include <cstdint>
> > #include <type_traits>
> > #include <iostream>
> > 
> > int main() {
> >   std::cout << std::boolalpha
> >             << "std::is_same_v<std::int64_t, long long> = "
> >             << std::is_same_v<std::int64_t, long long> << "\n";
> > }
> > EOF
> > clang++ -std=c++20 -O2 -Wall -Wextra /tmp/prog.cpp -o /tmp/a.out && /tmp/a.out
> > 
> > std::is_same_v<std::int64_t, long long> = true
> > ```
> 
> Ok, I misunderstood. So why do you need a separate specialization for `long long` in addition to the one for `std::int64_t`?

I think I understand now. `int32_t` is mapped onto `int`, while `int64_t` is mapped to `long long`. So the overload for `long` is missing. Makes sense now.

Thanks!

PR #6668, opened 2025-04-07
Feature/hierarchical collectives

EluZet:
This is the hierarchical collectives and hierarchical Communicator code used in the Bachelorthesis 'Hierarchical Collectives for HPX'

This is merely a proof of concept.

PR #6834, opened 2025-12-17
Feature/hierarchical collectives

hkaiser:
Supersedes #6668

@EluZet, @constracktor: please note that I had to create a new PR as I'm not allowed to push to the repository of the original PR.

### DIFF
--- a/libs/full/collectives/include/hpx/collectives/argument_types.hpp
+++ b/libs/full/collectives/include/hpx/collectives/argument_types.hpp
@@ -1,4 +1,4 @@
-//  Copyright (c) 2021-2023 Hartmut Kaiser
+//  Copyright (c) 2021-2025 Hartmut Kaiser
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
--- a/libs/full/collectives/include/hpx/collectives/broadcast.hpp
+++ b/libs/full/collectives/include/hpx/collectives/broadcast.hpp
@@ -1,4 +1,5 @@
 //  Copyright (c) 2020-2025 Hartmut Kaiser
+//  Copyright (c) 2025 Lukas Zeil
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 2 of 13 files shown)


hkaiser:
@EluZet @constracktor: I'm done with this from my end, please have a look and verify I didn't mess things up.

hkaiser:
The remaining errors are unrelated and are being fixed by other PRs

iemAnshuman:
Hi @EluZet, @constracktor,
I have been reading through this PR as part of my GSoC preparation for extending the hierarchical collectives. One pattern I found particularly interesting: in the hierarchical reduce_there, group leaders, sites whose hierarchical_communicator spans multiple levels, call reduce_here at each intermediate level, acting as the local reduction root for their sub-communicator before forwarding upward. Only the top level call (communicators.get(0)) uses flat reduce_there to send to the global root. Leaf sites with a single communicator level skip the loop entirely and just call flat reduce_there directly.
I am planning to prototype hierarchical all_reduce by composing these existing primitives: reduce_here/reduce_there (bottom-up) followed by broadcast_to/broadcast_from (top-down), with root and non-root sites taking their respective code paths in each phase.
One architectural question: since both phases would use the same hierarchical_communicator but with different tags (reduce_tag vs broadcast_tag), is it safe to reuse the same generation argument across both phases? Or should the broadcast phase use a separate generation to avoid any internal collision in the communicator's handle_data storage?

hkaiser:
> Hi @EluZet, @constracktor, I have been reading through this PR as part of my GSoC preparation for extending the hierarchical collectives. One pattern I found particularly interesting: in the hierarchical reduce_there, group leaders, sites whose hierarchical_communicator spans multiple levels, call reduce_here at each intermediate level, acting as the local reduction root for their sub-communicator before forwarding upward. Only the top level call (communicators.get(0)) uses flat reduce_there to send to the global root. Leaf sites with a single communicator level skip the loop entirely and just call flat reduce_there directly. I am planning to prototype hierarchical all_reduce by composing these existing primitives: reduce_here/reduce_there (bottom-up) followed by broadcast_to/broadcast_from (top-down), with root and non-root sites taking their respective code paths in each phase. One architectural question: since both phases would use the same hierarchical_communicator but with different tags (reduce_tag vs broadcast_tag), is it safe to reuse the same generation argument across both phases? Or should the broadcast phase use a separate generation to avoid any internal collision in the communicator's handle_data storage?

You will need to use distinct (and monotonuously incrementing) generations to disambiguate the phases of the operation. Each generation can represent a different operation (gather/scatter), but each participating site needs to agree on using the same generation for the same algorithmic step.

iemAnshuman:
Thanks, that makes sense. I'll use distinct, monotonically incrementing generations for each phase.

PR #6836, opened 2025-12-22
Removing boost.shared_array

sleepingeight:
Working towards https://github.com/STEllAR-GROUP/hpx/issues/3440

Implemented `shared_array` in hpx replacing the external dependency on boost.

## Checklist


hkaiser:
Thank you for your contribution! A question, though: Nowadays, most (if not all) implementations of `std::shared_ptr` support managing arrays (i.e. provide a specialization for `std::shared_ptr<T[]>` -- and `std::make_shared<T[]>` for that matter --, which is part of the standard since C++17 -- `make_shared` since C++20). Are you aware of any platform that doesn't support it? And if yes, wouldn't it be better to replace the use of `boost::shared_ptr` (which is clearly a leftover from code written a long time ago, even before C++17) with using `std::shared_ptr` in conjunction with  a custom deleter (something like `std::default_delete<T[]>` or similar)?

hkaiser:
> Thank you for your contribution! A question, though: Nowadays, most (if not all) implementations of `std::shared_ptr` support managing arrays (i.e. provide a specialization for `std::shared_ptr<T[]>` -- and `std::make_shared<T[]>` for that matter --, which is part of the standard since C++17 -- `make_shared` since C++20). Are you aware of any platform that doesn't support it? And if yes, wouldn't it be better to replace the use of `boost::shared_ptr` (which is clearly a leftover from code written a long time ago, even before C++17) with using `std::shared_ptr` in conjunction with a custom deleter (something like `std::default_delete<T[]>` or similar)?

After actually looking at the standard, let me be more concise: `std::shared_ptr<T[]>` is not a specialization. The standard however mandates that a `std::shared_ptr<T[]>` constructed without an explicit deleter shall call a deleter that invokes `delete[] p` (see: https://eel.is/c++draft/util.smartptr.shared#const-6).

hkaiser:
Let me add, that either way, I fully support removing the use of `boost::shared_array` from our codebase.

sleepingeight:
Hi @hkaiser, thank you for reviewing. Few points,
1. From google searches (not exhaustive), I haven't found any platform confirming no support for shared_ptr<T[]>. I'd like you to make a choice either to have a in house shared_array implementation, or porting entirely to the standard.
2. I understand that shared_array can be implemented using shared_ptr and a custom deleter. However, only this wouldn't be sufficient as this implementation will not provide a `operator[]`. I can modify the current shared_array implementation to be based on the standard `shared_ptr` with a custom overload to provide required overloads or code changes can be made to the current implementations in case of any issues. Should I update the shared_array implementation, or I go with 1?
3. `boost::shared_ptr` is used in HPX in two cases, serialization (`HPX_SERIALIZATION_HAVE_BOOST_TYPES`) and in
```
// Older implementations of std::shared_ptr do not support binding function
// pointers (see http//wg21.link/lwg3018)
HPX_HAVE_CXX11_STD_SHARED_PTR_LWG3018
``` 
Would want your opinion on maintaining these options too.

hkaiser:
> 1. From google searches (not exhaustive), I haven't found any platform confirming no support for shared_ptr<T[]>. I'd like you to make a choice either to have a in house shared_array implementation, or porting entirely to the standard.

That's what I thought as well. Even more as we now require C++20 for compiling HPX.

> 2. I understand that shared_array can be implemented using shared_ptr and a custom deleter. However, only this wouldn't be sufficient as this implementation will not provide a `operator[]`. I can modify the current shared_array implementation to be based on the standard `shared_ptr` with a custom overload to provide required overloads or code changes can be made to the current implementations in case of any issues. Should I update the shared_array implementation, or I go with 1?

Using standard-based solutions is always the priority. May I suggest to remove the use of `boost::shared_array` (including the configuration check for `HPX_WITH_CXX17_SHARED_PTR_ARRAY`), i.e. assuming that the library properly implements `std::shared_ptr<T[]>`. We already have the standard-based implementation in place, so this should boil down to removing the non-standard alternative. While I know that you have invested time into forking the Boost implementation, I would prefer relying on the standard library.

> 3. `boost::shared_ptr` is used in HPX in two cases, serialization (`HPX_SERIALIZATION_HAVE_BOOST_TYPES`) and in

I meant `boost::shared_array`, sorry for the confusion. Support for `HPX_SERIALIZATION_HAVE_BOOST_TYPES` is purely to enable user code that might rely on serializing `boost::shared_ptr`. I think we should leave this as is.

> ```
> // Older implementations of std::shared_ptr do not support binding function
> // pointers (see http//wg21.link/lwg3018)
> HPX_HAVE_CXX11_STD_SHARED_PTR_LWG3018
> ```

That's another configuration option that can be removed now, I believe - even if this is unrelated to this PR.

sleepingeight:
>  While I know that you have invested time into forking the Boost implementation, I would prefer relying on the standard library.

No worries, I've modified this PR to remove `boost.shared_array` entirely. I'll raise a seperate cleanup PR HPX_HAVE_CXX11_STD_SHARED_PTR_LWG3018 option.

hkaiser (approved):
LGTM, thanks!

hkaiser:
@sleepingeight We would like to wait for our CI to be fixed before merging anything. Please be patient.

PR #6838, opened 2025-12-24
initialize parallel reduce from init value

ArivoliR:
Fixes [#6647](https://github.com/STEllAR-GROUP/hpx/issues/6647)

## Proposed Changes
  - Initialize each parallel reduce partition using the user-provided init value instead of seeding from the first range element.
  - Add a regression test using hpx::ranges::reduce with a non-trivial identity to test the corrected behavior.

## Checklist


ArivoliR:
Messed up while trying to sign my commit by rebasing. Since I had synced the branch with master, the recent changes were not mine hence this behavior. Will submit a new PR.

hkaiser:
@ArivoliR what is the rationale for this change? If you use the initial value passed to the algorithm for each of the partitions the algorithm will end up considering this value more than once, which may lead to wrong results. For instance:
```
std::vector<int> v = {0, 1, 2, 3, 4, 5};
auto sum = hpx::range::reduce(v, 42, std::plus<>{});
assert(sum == 57);   // may fail
```
Here the expected `sum` is `57`, however if the initial value is used more than once you may end up with a different result (dependening on the number of partitions created).

ArivoliR:
You're right sir, the change is incorrect. 
The tests seemed to pass locally, I think I overlooked something.
Will get more concrete understanding and results and get back.
Thank you for pointing this out.

PR #6843, opened 2026-01-01
Attempting to improve compiler error messages related to tag_invoke

hkaiser:
This is an experiment of how to best improve generated error messages if no appropriate `tag_invoke` overloads are being available. 

The main idea is to add explicit `operator()()` overload(s) to the customization point object that concept-check all arguments. These `operator()()` overloads simply forward to the base-class dispatch logic. This has the additional benefit of centralized argument concept-checking for all `tag_invoke` overloads provided.

If this proves to be applicable it opens up a path to fixing #6718.


hkaiser:
retest lsu

hkaiser:
Closing this as we're moving away from the current use of `tag_invoke`.

PR #6845, opened 2026-01-01
Starting to adapt HPX full modules to C++ modules

hkaiser:


hkaiser:
All remaining problems are unrelated and will be fixed separately.

PR #6850, opened 2026-01-18
Fix Ninja Multi-Config generator support for pkg-config files

### LINKED ISSUE #6710: Support Ninja Multi-Config generator
Trying to use Ninja Multi-Config generator on linux results in errors:

```
[cmake] CMake Error in CMakeLists.txt:
[cmake]   Evaluation file to be written multiple times with different content.  This
[cmake]   is generally caused by the content evaluating the configuration type,
[cmake]   language, or location of object files:
[cmake] 
[cmake]    /home/k/hpx/build/lib/pkgconfig/hpx_application_debug.pc
```

The culprit is this line https://github.com/STEllAR-GROUP/hpx/blob/15afc8be41b0668584dfafe2d8cf6ffb75dfb715/cmake/HPX_GeneratePackageUtils.cmake#L399

guptapratykshh:
Fixes #6710

## Proposed Changes

  - Fix pkg-config file generation for Ninja Multi-Config generator by detecting multi-config mode and generating separate files for each configuration (Debug, Release, etc.)
 - Update both HPX_GeneratePackageUtils.cmake and HPX_GeneratePackage.cmake to handle multi-config generators properly
 - Preserve backward compatibility with single-config generators

When using Ninja Multi-Config generator, CMAKE_BUILD_TYPE is not set during configuration time, causing pkg-config files to be written multiple times with different content, resulting in CMake errors. The fix detects multi-config generators and creates separate configuration-specific files instead of trying to write to the same file multiple times.
## Checklist


hkaiser:
Could this somehow be verified by adding a test or two?

hkaiser:
@guptapratykshh Thanks again for working on this!

guptapratykshh:
thank you for your valuable time for reviewing the PR.

PR #6851, opened 2026-01-18
added CMakePresets.json with multiple build configurations for HPX development

### LINKED ISSUE #6798: Provide a CMakePresets.json file
The `CMakePresets.json` is a portable way to create CMake build recipes. It can be used to store common library configurations (minimal, full, performance instrumentation, build with memory sanitizer, documentation build, etc.). Most importantly for my use case, it is picked up by IDEs.

We could include one ourselves (Example: https://github.com/NVIDIA/cccl/blob/main/CMakePresets.json).

Advantages to including one in the repository:
- Provides examples of common HPX configurations. Currently, CMake options are in the documentation, but there are many and perhaps look daunting. Presets have descriptions and are somewhat self-documenting
- Quick start-off point for beginning users. Gives a reference to start playing with. Since it's portable, it will make our build guide slightly shorter.
- Encourages users to adopt a file-based configuration approach with named configurations and separate build directories, instead of terminal flag soup.

Considerations:
- Support: CMake>=3.19 (some nice features are introduced in later versions). HPX minimum is currently CMake>=3.18.
- Too many combinations (Release/Debug, with/without dependency fetching, etc.) Not clear which are most sensible.

guptapratykshh:
Fixes #6798

## Proposed Changes

  - Added comprehensive CMakePresets.json file with 21 build configurations covering common HPX use cases
  - Included presets for minimal, full, debug, performance, sanitizer builds, and specialized configurations
  - Enhanced IDE integration by providing portable build recipes that work across different development environments

this addresses need for standardized build configurations that are IDE-friendly and reduce complexity of managing numerous CMake command-line options. this presets provide self documenting build configurations that make it easier for both new and experienced users to get started with HPX development.

## Checklist

### DIFF
--- a/docs/sphinx/manual/building_hpx.rst
+++ b/docs/sphinx/manual/building_hpx.rst
@@ -153,6 +153,38 @@ Available build types are:
    the ``CMAKE_BUILD_TYPE`` variables have to match and for projects not using
    |cmake|_, the ``HPX_DEBUG`` macro has to be set in debug mode.
 
+.. _cmake_presets:
+
+Using CMake Presets
+===================
+
+|hpx| provides a ``CMakePresets.json`` file which includes a variety of pre-defined build configurations.
+These presets allow you to easily configure the build for common scenarios without needing to manually specify multiple CMake variables.
+
+To use a preset, you can use the ``--preset`` option with CMake:
+
+.. code-block:: shell-session
+
+    $ cmake --preset <preset-name>
+    $ cmake --build --preset <preset-name>
+
+Some of the available presets include:
+
+* ``default``: Standard release build with tests and examples enabled.
+* ``minimal``: Minimal build with only core features (no tests, examples, or tools).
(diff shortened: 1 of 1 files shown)


hkaiser:
What is different here from the CMakePreset.json we already have in the repository (https://github.com/STEllAR-GROUP/hpx/blob/master/CMakePresets.json)?

guptapratykshh:
so, i have focused on being strictly additive and expanding it from 10 to 20 configurations. i added specialized things like  CUDA/SYCL, different memory allocators, and unity builds that were not covered in current version and also really helps avoid "flag soup" in terminal and makes these advanced HPX configurations much more accessible and self documenting directly through IDE.

hkaiser:
Could you add some documentation as well, please?

hkaiser:
@guptapratykshh thanks again for your contribution!

guptapratykshh:
Thank you for all the valuable feedback

PR #6852, opened 2026-01-18
Add HPX_EXPORT to plugin factory types to fix MacOS visibility issues

### LINKED ISSUE #6653: Passing `HPX_WITH_HIDDEN_VISIBILITY=TRUE` on MacOS results in library which can't be used
The library builds and all seems fine, but some functionality does not work.

Error message when running some of the HPX examples:

```
libc++abi: terminating due to uncaught exception of type hpx::detail::exception_with_info<hpx::exception>: Hpx.Plugin: Can't cast to the right factory type
: HPX(filesystem_error)
```

Rebuilding HPX with `HPX_WITH_HIDDEN_VISIBILITY=FALSE` makes the issue disappear.

- HPX-1.10.0
- MacOS-14 / AppleClang-15 and MacOS-15 / AppleClang-16
- Installed dependencies with homebrew

guptapratykshh:
Fixes #6653

## Proposed Changes

   - Added HPX_EXPORT to abstract_factory, concrete_factory, and the main plugin_factory templates.
   - Explicitly exported the plugin_wrapper and virtual_constructor helper types.
   - Ensured that RTTI (type information) for plugin classes is shared correctly across shared library boundaries on MacOS.

i am on macos, when HPX_WITH_HIDDEN_VISIBILITY=TRUE is enabled, symbols for template classes(like vtables and RTTI) are hidden by default within their respective dsos. 
## Checklist


hkaiser:
Thank you for your fixes. Could you please also fix the problems that are now being reported by clang-format:

```diff
diff --git a/libs/core/plugin/include/hpx/plugin/plugin_factory.hpp b/libs/core/plugin/include/hpx/plugin/plugin_factory.hpp
index 6e1d0e0..4e37b87 100644
--- a/libs/core/plugin/include/hpx/plugin/plugin_factory.hpp
+++ b/libs/core/plugin/include/hpx/plugin/plugin_factory.hpp
@@ -139,8 +139,8 @@ namespace hpx::util::plugin {
             }
         }
 
-        HPX_CXX_EXPORT inline HPX_CORE_EXPORT void
-        get_abstract_factory_names(dll const& d, std::string const& base_name,
+        HPX_CXX_EXPORT inline HPX_CORE_EXPORT void get_abstract_factory_names(
+            dll const& d, std::string const& base_name,
             std::vector<std::string>& names, error_code& ec = throws)
         {
             using deleter_type = hpx::function<void(get_plugins_list_type)>;
diff --git a/libs/core/plugin/include/hpx/plugin/virtual_constructor.hpp b/libs/core/plugin/include/hpx/plugin/virtual_constructor.hpp
index 7f2fc9f..67c91fb 100644
--- a/libs/core/plugin/include/hpx/plugin/virtual_constructor.hpp
+++ b/libs/core/plugin/include/hpx/plugin/virtual_constructor.hpp
@@ -24,7 +24,7 @@ namespace hpx::util::plugin {
     HPX_CXX_EXPORT typedef exported_plugins_type*(
         HPX_PLUGIN_API* get_plugins_list_type)();
     HPX_CXX_EXPORT typedef exported_plugins_type*(
-        HPX_PLUGIN_API get_plugins_list_np)();
+        HPX_PLUGIN_API get_plugins_list_np) ();
     HPX_CXX_EXPORT using dll_handle = shared_ptr<get_plugins_list_np>;
 
     ///////////////////////////////////////////////////////////////////////////
```

hkaiser:
Please have a look at the Windows CIs, those seem to fail because of your changes.

libs/core/plugin/include/hpx/plugin/plugin_factory.hpp
```diff
@@ -126,9 +126,9 @@ namespace hpx::util::plugin {
         }
 
         ///////////////////////////////////////////////////////////////////////
-        HPX_CXX_EXPORT inline void get_abstract_factory_names_static(
-            get_plugins_list_type f, std::vector<std::string>& names,
-            error_code& /*ec*/ = throws)
+        HPX_CXX_EXPORT inline HPX_CORE_EXPORT void
+        get_abstract_factory_names_static(get_plugins_list_type f,
+            std::vector<std::string>& names, error_code& /*ec*/ = throws)
```
hkaiser: I don't think that there is a need to export this `inline` function (or any of the other similar ones).

guptapratykshh: i have reverted the inline function ones

hkaiser:
Thanks for fixing the example, much appreciated. That was a long-standing issue.

guptapratykshh:
i am glad i could get that sorted out

libs/core/synchronization/include/hpx/synchronization/channel_mpsc.hpp
```diff
@@ -50,7 +50,7 @@ namespace hpx::lcos::local {
 
         [[nodiscard]] bool is_empty(std::size_t head) const noexcept
         {
-            return head == tail_.data_.tail_.load(std::memory_order_relaxed);
+            return head == tail_.data_.tail_.load(std::memory_order_acquire);
```
hkaiser, line 53: While I'm not opposed to stay on the safe side and to use more stringent memory order requirements, I would like to understand your reasoning for these proposed modifications.

guptapratykshh: i have upgraded to acquire because relaxed load in is_empty() was causing race condition during high contention work stealing and consumer was occasionally seeing updated tail pointer before actual request data had synchronized across cores, leading to stale reads, this change ensures that task data is fully visible whenever is_empty() returns false

hkaiser: Ok, thanks for diagnosing this. What tests were affected by this?

guptapratykshh: the main issue was cross_pool_injection. stale reads of steal request data caused garbage bitset indices and led to out-of-bounds crashes when there was high contention

PR #6856, opened 2026-01-21
Godbolt Boost detection: search for static libs if shared not found

### LINKED ISSUE #6541: Add HPX to Godbolt Compiler Explorer
Please consider adding the HPX library to Compiler Explorer, so other projects can test interoperability with HPX.

I have put libraries in CE but all of them were header-only, not sure if that is an obstacle.

guptapratykshh:
Fixes #6541

## Proposed Changes

- added automatic fallback to search for static Boost libraries if shared libraries are not found during detection. 
- updated boost module detection (Filesystem, Regex, etc.) to prioritize CONFIG mode before falling back to MODULE mode for better compatibility with modern pre-installed versions.
- added fallback link to Boost::headers if primary Boost::boost target is unavailable, ensuring header-only usage still works smoothly. consolidated and cleaned up boost setup logic to remove redundant external module files.

## Checklist

### DIFF
--- a/cmake/HPX_SetupBoost.cmake
+++ b/cmake/HPX_SetupBoost.cmake
@@ -45,14 +45,33 @@ if(NOT TARGET hpx_dependencies_boost)
     if(NOT Boost_FOUND)
       # Search using MODULE mode
       find_package(
+        Boost ${Boost_MINIMUM_VERSION} MODULE QUIET NO_POLICY_SCOPE
+        COMPONENTS ${__boost_libraries}
+      )
+    endif()
+
+    if(NOT Boost_FOUND AND NOT Boost_USE_STATIC_LIBS)
+      hpx_info(
+        "Shared Boost not found, attempting to find static Boost libraries."
+      )
+      set(Boost_USE_STATIC_LIBS ON)
+      find_package(
         Boost ${Boost_MINIMUM_VERSION} MODULE REQUIRED NO_POLICY_SCOPE
         COMPONENTS ${__boost_libraries}
       )
     endif()
 
+    if(NOT Boost_FOUND)
+      hpx_error("Could not find Boost. Please check your configuration.")
+    endif()
+
     add_library(hpx_dependencies_boost INTERFACE IMPORTED)
 
-    target_link_libraries(hpx_dependencies_boost INTERFACE Boost::boost)
+    if(TARGET Boost::boost)
+      target_link_libraries(hpx_dependencies_boost INTERFACE Boost::boost)
+    else()
+      target_link_libraries(hpx_dependencies_boost INTERFACE Boost::headers)
+    endif()
 
     foreach(__boost_library ${__boost_libraries})
       target_link_libraries(
--- a/cmake/HPX_SetupBoostFilesystem.cmake
+++ b/cmake/HPX_SetupBoostFilesystem.cmake
@@ -9,9 +9,15 @@ if(HPX_FILESYSTEM_WITH_BOOST_FILESYSTEM_COMPATIBILITY)
   if(NOT TARGET Boost::filesystem)
 
     find_package(
-      Boost ${Boost_MINIMUM_VERSION} NO_POLICY_SCOPE MODULE
+      Boost ${Boost_MINIMUM_VERSION} NO_POLICY_SCOPE CONFIG QUIET
       COMPONENTS filesystem
     )
+    if(NOT Boost_FOUND)
+      find_package(
+        Boost ${Boost_MINIMUM_VERSION} NO_POLICY_SCOPE MODULE
+        COMPONENTS filesystem
+      )
+    endif()
 
     if(NOT Boost_FILESYSTEM_FOUND)
       hpx_error(
--- a/cmake/HPX_SetupBoostRegex.cmake
+++ b/cmake/HPX_SetupBoostRegex.cmake
@@ -7,8 +7,14 @@
 if(NOT TARGET Boost::regex)
 
   find_package(
-    Boost ${Boost_MINIMUM_VERSION} NO_POLICY_SCOPE MODULE COMPONENTS regex
+    Boost ${Boost_MINIMUM_VERSION} NO_POLICY_SCOPE CONFIG QUIET
+    COMPONENTS regex
   )
+  if(NOT Boost_FOUND)
+    find_package(
+      Boost ${Boost_MINIMUM_VERSION} NO_POLICY_SCOPE MODULE COMPONENTS regex
+    )
+  endif()
 
   if(Boost_REGEX_FOUND)
     hpx_info("  regex")


libs/core/algorithms/include/hpx/parallel/util/foreach_partitioner.hpp
```diff
@@ -341,7 +341,17 @@ namespace hpx::parallel::util::detail {
                     handle_local_exceptions::call(
                         HPX_FORWARD(decltype(r2), r2));
 
-                    return f(HPX_MOVE(last));
+                    if constexpr (std::is_void_v<decltype(f(HPX_MOVE(last)))>)
+                    {
+                        f(HPX_MOVE(last));
+                        scoped_params.reset();
```
hkaiser: I think the explicit `reset` is not necessary as the function stored in dataflow is guaranteed to be destroyed after it has finished running. What is the rationale for this change? Did you see any problems?

hkaiser: If this explicit `reset` turns out to be necessary, I'd suggest using RAII to do this reliably (even in cases of exceptions thrown by the function `f`):
```c++
    auto _ = hpx::experimental::scope_exit([&] { scoped_params.reset(); });
    return f(HPX_MOVE(last));
```

hkaiser: @guptapratykshh Would you mind commenting on why you believe this is needed?

libs/core/algorithms/include/hpx/parallel/util/foreach_partitioner.hpp
```diff
@@ -341,6 +342,9 @@ namespace hpx::parallel::util::detail {
                     handle_local_exceptions::call(
                         HPX_FORWARD(decltype(r2), r2));
 
+                    // force mark_end_execution to be called before the future becomes ready
```
hkaiser: Please note that this function will be called _after_ the future has become ready. I still don't understand why you believe this is necessary. Please give us your rationale.

guptapratykshh: i found that dataflow future becomes ready as soon as lambda returns,but lambda's destructor(which calls mark_end_execution)runs slightly later.this causes race where test checks hook count before the destructor has actually fired and explicitly resetting it inside lambda guarantees hook runs before future signals completion.

hkaiser: Ok, makes sense. In that case, I don't think we need the `HPX_UNUSED()` inside the lambdas anymore as the captured variable is now referenced explicitly.

hkaiser:
Also, inspect is now complaining:

/libs/core/algorithms/include/hpx/parallel/util/foreach_partitioner.hpp: *Line length limit*: [345](https://github.com/STEllAR-GROUP/hpx/blob/68381a233fc846eddbf9dc1d11e5ba8bbe0e95ee/libs/core/algorithms/include/hpx/parallel/util/foreach_partitioner.hpp#L345), [378](https://github.com/STEllAR-GROUP/hpx/blob/68381a233fc846eddbf9dc1d11e5ba8bbe0e95ee/libs/core/algorithms/include/hpx/parallel/util/foreach_partitioner.hpp#L378), [409](https://github.com/STEllAR-GROUP/hpx/blob/68381a233fc846eddbf9dc1d11e5ba8bbe0e95ee/libs/core/algorithms/include/hpx/parallel/util/foreach_partitioner.hpp#L409)
/libs/core/algorithms/include/hpx/parallel/util/partitioner.hpp: *Line length limit*: [642](https://github.com/STEllAR-GROUP/hpx/blob/68381a233fc846eddbf9dc1d11e5ba8bbe0e95ee/libs/core/algorithms/include/hpx/parallel/util/partitioner.hpp#L642)
/libs/core/algorithms/include/hpx/parallel/util/partitioner_with_cleanup.hpp: *Line length limit*: [373](https://github.com/STEllAR-GROUP/hpx/blob/68381a233fc846eddbf9dc1d11e5ba8bbe0e95ee/libs/core/algorithms/include/hpx/parallel/util/partitioner_with_cleanup.hpp#L373)

libs/core/algorithms/include/hpx/parallel/util/foreach_partitioner.hpp
guptapratykshh: right, HPX_UNUSED(scoped_params) is now redundant.i will remove it from all lambdas

libs/core/algorithms/include/hpx/parallel/util/foreach_partitioner.hpp
```diff
@@ -181,7 +182,15 @@ namespace hpx::parallel::util::detail {
...
-                    return HPX_INVOKE(f2, HPX_MOVE(last));
+                    using result_type = std::invoke_result_t<F2&, FwdIter>;
+                    if constexpr (std::is_void_v<result_type>)
+                    {
+                        HPX_INVOKE(f2, HPX_MOVE(last));
+                    }
+                    else
+                    {
+                        return HPX_INVOKE(f2, HPX_MOVE(last));
+                    }
```
hkaiser: I don't understand this change, can you explain why this is necessary (same for similar changes below)?

guptapratykshh: i have replaced scope_exit with manual cleanup to resolve ci linker errors and reduce template instantiation and the if constexpr split is necessary to capture return value(or handle void) so i can explicitly reset parameters before returning

hkaiser: > resolve ci linker errors

Huh? What linker errors? 

May I suggest to separate this issue from the Boost fixes. Then we can go ahead to merge those and thoroughly discuss the changes to the partitioners (which are unrelated to begin with). Please completely remove the partitioner changes here and create a new PR for them.

hkaiser: > replaced scope_exit with manual cleanup

Also, this will not perform the cleanup if `f` throws an exception.

libs/core/algorithms/include/hpx/parallel/util/foreach_partitioner.hpp
hkaiser: > Ok, makes sense. In that case, I don't think we need the `HPX_UNUSED()` inside the lambdas anymore as the captured variable is now referenced explicitly.

Still, I would like to know which of the tests adding the explicit cleanup fixes.

libs/core/algorithms/include/hpx/parallel/util/foreach_partitioner.hpp
guptapratykshh: have removed all the changes of the partitioners

libs/core/algorithms/include/hpx/parallel/util/foreach_partitioner.hpp
guptapratykshh: explicit cleanup would fix executor_parameters_timer_hooks_test and i have not implemented it yet in this PR because I have removed all partitioner changes. i will create separate PR specifically for partitioner cleanup issue where, i can properly test and demonstrate that executor_parameters_timer_hooks_test passes with fix.

PR #6859, opened 2026-01-24
[Serialization] [Feature] Add experimental C++26 reflection support

ujjwal-shekhar:
This PR introduces support for C++26 reflection-based serialization (based on the P2996 proposal). It adds a new reflection-based dispatch mechanism in `access.hpp` that allows for the automatic serialization of user-defined structs without requiring manual `serialize` functions. _The code generation pertaining to (semi-)automatic polymorphic type registration, serialization lambdas (with and without captures) and compile-time compression of `qualified_name_of` will be done in the next PR(s)._

## Proposed Changes
  -  Added `HPX_SERIALIZATION_WITH_ALLOW_AUTO_GENERATE` to allow users to toggle reflection codegen.
  - Automatically iterates over bases and data members in `refl_serialize_impl.hpp` using `std::meta`. 
  - Utilizes `std::meta::access_context::unchecked()` and compile-time pointer arithmetic (`offset_of`) to serialize private members without requiring `friend` declarations.
  - Added a `qualified_name_of` utility to generate compiler-agnostic type identifiers by walking the scope tree. This generates portable and unique names for each type*
  - Integrated the reflection path into the main serialize dispatch. Manual serialize members and ADL handlers still take precedence, ensuring full backward compatibility.
  - When reflection-based serialization is enabled, `struct_serialization` is superseded, as the reflection branch natively handles brace-initializable structs.
  - Added `HPX_POLYMORPHIC_AUTO_REGISTER` macro to enable polymorphic type registration.
  
Current HPX serialization requires users to manually list every member in a serialize function, leading to high maintenance overhead and error-prone edits. Encapsulation is also a hurdle, as non-intrusive methods cannot access private data easily.

Goals:
- Adding/removing members of a class requires zero changes to serialization code.
- Generally, eliminate the need for classes to be "aware" of their serialization logic.
- Build on existing HPX/Boost infrastructure without replacing it.

## Checklist

### DIFF
--- a/libs/core/serialization/CMakeLists.txt
+++ b/libs/core/serialization/CMakeLists.txt
@@ -198,6 +198,7 @@ set(serialization_headers
     hpx/serialization/traits/brace_initializable_traits.hpp
     hpx/serialization/traits/is_bitwise_serializable.hpp
     hpx/serialization/traits/is_not_bitwise_serializable.hpp
+    hpx/serialization/traits/is_serializable.hpp
     hpx/serialization/traits/is_serialization_supported.hpp
     hpx/serialization/traits/needs_automatic_registration.hpp
     hpx/serialization/traits/polymorphic_traits.hpp


libs/core/config/include/hpx/config/reflection.hpp
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2025 Ujjwal Shakeher
+// SPDX-License-Identifier: BSL-1.0
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once// currently only on Clang p2996 fork
+
+#include <hpx/config/defines.hpp>
+
+#if (defined(__clang__) && __clang_major__ >= 20) && !defined(_MSC_VER) &&         \
```
hkaiser: Do we need the additional check for the compiler version here? Isn't that what the feature test handles?

libs/core/config/include/hpx/config/reflection.hpp
```diff
@@ -0,0 +1,19 @@
...
+    !defined(HPX_MINGW)
+#define HPX_HAVE_CXX26_EXPERIMENTAL_META
+
+    // By default allow auto generation of serialization functions using C++26 reflection
+    #if !defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
```
hkaiser: I think the auto-generated serialization support should be enabled only if the compiler supports it.
```suggestion
    #if defined(HPX_HAVE_CXX26_EXPERIMENTAL_META) &&  !defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
```

libs/core/serialization/include/hpx/serialization/detail/polymorphic_nonintrusive_factory.hpp
```diff
@@ -28,6 +28,10 @@
 
 #include <hpx/config/warnings_prefix.hpp>
 
+#if defined(HPX_HAVE_CXX26_EXPERIMENTAL_META) && defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
```
hkaiser:
```suggestion
#if defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
```

hkaiser: This can be applied to all conditional checks below as well.

libs/core/serialization/include/hpx/serialization/detail/polymorphic_nonintrusive_factory.hpp
```diff
@@ -164,12 +184,38 @@ namespace hpx::serialization::detail {
     {
         static void save(output_archive& ar, void const* base)
         {
-            serialize(ar, *static_cast<Derived*>(const_cast<void*>(base)), 0);
+            #if defined(HPX_HAVE_CXX26_EXPERIMENTAL_META) && defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
+                if constexpr (std::is_class_v<Derived>) {
```
hkaiser: IIUC, it should be sufficient to do the check and dispatching in `serialization::access`. Do we need this separate distinction here as well?

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,94 @@
...
+#include <algorithm>
+
+// This file is only ever included by access.hpp
+// but we will still guard against direct inclusion
+#if defined(HPX_HAVE_CXX26_EXPERIMENTAL_META) && defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
+#include <ranges>
+#include <experimental/meta>
+
+namespace hpx::serialization::detail {
+private:
```
hkaiser: I was not aware that access specifiers could be used at namespace scope. Is that even legal?

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,94 @@
...
+#pragma once
+
+#include <hpx/serialization/input_archive.hpp>
+#include <hpx/serialization/output_archive.hpp>
+
+#include <hpx/serialization/base_object.hpp>
+
+#include <string_view>
+#include <vector>
+#include <algorithm>
```
hkaiser: You may want to move all `#include`'s (except `<hpx/config.hpp>` inside the `#if defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)` block.

libs/core/serialization/include/hpx/serialization/detail/refl_serialize_impl.hpp
```diff
@@ -0,0 +1,85 @@
...
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/serialization/input_archive.hpp>
+#include <hpx/serialization/output_archive.hpp>
+
+#include <hpx/serialization/base_object.hpp>
+
+#include <iostream>
```
hkaiser: Please avoid #including `iostream` from HPX headers. This looks like a leftover anyways.

libs/core/serialization/include/hpx/serialization/traits/is_bitwise_serializable.hpp
```diff
@@ -41,4 +41,10 @@ namespace hpx::traits {
     inline constexpr bool is_bitwise_serializable_v =
         is_bitwise_serializable<T>::value;
 
+    /* Will be addressed in a separate PR, to allow serializing AoV and VoA*/
+    // HPX_CXX_EXPORT template <typename T>
+    // struct is_bitwise_serializable<hpx::serialization::array<T>>
+    //   : is_bitwise_serializable<T>
+    // {}; // Fixes arrays of non-bitwise-serializable types
```
hkaiser: You may want to move this definition (once provided) into the header `hpx/serialization/array.hpp` to keep everything related in one place.

libs/core/serialization/include/hpx/serialization/access.hpp
```diff
@@ -102,20 +107,32 @@ namespace hpx::serialization {
...
+#if !defined(HPX_HAVE_CXX26_EXPERIMENTAL_META) || !defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
+                else if constexpr (std::is_aggregate_v<dT> &&
+                    hpx::traits::has_struct_serialization_v<dT>)
+                {
+                    // This is automatic serialization for types that are simple
+                    // (brace-initializable) structs, that means every struct's
+                    // field has to be serializable and public.
+                    serialize_struct(ar, t, 0);
+                }
+#endif
```
hkaiser: What is the rationale to have this block before the bitwise serializable logic now? Shouldn't bitwise serializability be considered first?

libs/core/serialization/include/hpx/serialization/input_archive.hpp
```diff
@@ -151,18 +152,25 @@ namespace hpx::serialization {
 #endif
                     load_binary(&t, sizeof(t));
                 }
+#if !defined(HPX_HAVE_CXX26_EXPERIMENTAL_META) || !defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
                 else if constexpr (hpx::traits::has_struct_serialization_v<T>)
                 {
                     // struct serialization
                     access::serialize(*this, t, 0);
                 }
+#endif
```
hkaiser:
```suggestion
else if constexpr (has_refl_serialize && hpx::traits::has_struct_serialization_v<T>)
{
    // struct serialization
    access::serialize(*this, t, 0);
}
```

libs/core/serialization/include/hpx/serialization/map.hpp
```diff
@@ -143,4 +143,42 @@ namespace hpx::serialization {
...
+        std::uint64_t const size = t.size();
+        ar << size;
+        if (size == 0)
+            return;
+
+        for (value_type const& val : t)
+        {
+            ar << val;
+        }
+    }
```
hkaiser, line 155: Can we leave adding support for `std::multimap` for a separate PR, please?

libs/core/serialization/include/hpx/serialization/output_archive.hpp
```diff
@@ -250,18 +251,25 @@ namespace hpx::serialization {
...
                     save_binary(&t, sizeof(t));
                 }
-                else if constexpr (hpx::traits::has_struct_serialization_v<T>)
+#if !defined(HPX_HAVE_CXX26_EXPERIMENTAL_META) || !defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
+                else if constexpr (!hpx::traits::has_struct_serialization_v<T>)
                 {
                     // struct serialization
                     access::serialize(*this, t, 0);
                 }
+#endif
```
hkaiser: Same as above, let's simplify this to:

```suggestion
                else if constexpr (has_refl_serialize && !hpx::traits::has_struct_serialization_v<T>)
                {
                    // struct serialization
                    access::serialize(*this, t, 0);
                }
```

libs/core/serialization/include/hpx/serialization/serialization_fwd.hpp
```diff
@@ -42,6 +42,15 @@ namespace hpx::serialization {
...
         void serialize_one(input_archive& ar, T& t);
+
+        #if defined(HPX_HAVE_CXX26_EXPERIMENTAL_META) && defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
+        // This function uses C++26 reflection capabilities to generate
+        // serialization functions for types that don't have them already.
+        // This is forward declared here and defined in refl_serialize_impl.hpp
+        // to avoid circular dependencies.
+        HPX_CXX_EXPORT template <typename Archive, typename T>
+        void refl_serialize(Archive& ar, T& t, unsigned /*version*/);
+        #endif
```
hkaiser: Do we need this top-level API function?

libs/core/serialization/include/hpx/serialization/set.hpp
```diff
@@ -47,4 +47,35 @@ namespace hpx::serialization {
...
+        std::uint64_t const size = set.size();
+        ar << size;
+        if (size == 0)
+            return;
+
+        for (T const& i : set)
+        {
+            ar << i;
+        }
+    }
```
hkaiser, line 65: Let's combine adding support for `std::multiset` with the support for `std::multimap` in a separate PR.

libs/core/serialization/include/hpx/serialization/unordered_map.hpp
```diff
@@ -57,4 +57,44 @@ namespace hpx::serialization {
...
+        std::uint64_t const size = t.size();
+        ar << size;
+        if (size == 0)
+            return;
+
+        for (auto const& val : t)
+        {
+            ar << val;
+        }
+    }
```
hkaiser, line 70: Same, let's move this to a separate PR.

libs/core/serialization/CMakeLists.txt
```diff
@@ -101,6 +101,24 @@ if(HPX_SERIALIZATION_WITH_ALLOW_CONST_TUPLE_MEMBERS)
...
+  MODULE SERIALIZATION
+)
+
+if(HPX_SERIALIZATION_WITH_ALLOW_AUTO_GENERATE)
+  hpx_add_config_define_namespace(
+    DEFINE HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE
+    NAMESPACE SERIALIZATION
+  )
+endif()
+
```
hkaiser, line 145: Let's make this available only if the compiler actually supports it (the feature test will tell).

libs/core/serialization/include/hpx/serialization/unordered_map.hpp
hkaiser: That PR can also add support for `std::unordered_multiset` and tests for the four new data structures.

.github/workflows/linux_release_clang_p2996_reflection.yml
```diff
@@ -0,0 +1,47 @@
...
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+name: Linux CI (Release, Clang P2996 Reflection)
+
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container: stellargroup/build_env:17
```
hkaiser: Does this use the version of clang that supports reflection? I thought the docker image has a vanilla Clang-20 installed.

ujjwal-shekhar:
> @ujjwal-shekhar Sorry for spamming with a long list of comments, I went over the PR from a formal standpoint for now. I have not done a deep review of the implementation yet.
> 
> Generally, thank you so much for working on this! This is marvelous.

Thank you! It is a really interesting feature indeed. No worries about the long list; I'll go through the comments soon and green out the failing checks as well.

libs/core/serialization/include/hpx/serialization/unordered_map.hpp
ujjwal-shekhar: I'll make a separate PR with these very soon, once that PR gets merged, I'll rebase these commits.

libs/core/serialization/include/hpx/serialization/serialization_fwd.hpp
ujjwal-shekhar: Thank you for pointing this out, I have forward declared this in access.hpp now (this part in `serialization_fwd` is currently commented out, I'll remove it in the next commit.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: Aah thank you for this I have removed them. This was originally going to be a single class and I might have left it there accidentally.

I will also add detailed tests for `refl_qualified_name_of.hpp` (and other polymorphic type registry related changes) after the basic auto-serialization features look good:)

libs/core/serialization/include/hpx/serialization/detail/polymorphic_nonintrusive_factory.hpp
ujjwal-shekhar: I have addressed these changes.

libs/core/config/include/hpx/config/reflection.hpp
ujjwal-shekhar: I have removed this line.

libs/core/serialization/include/hpx/serialization/detail/refl_serialize_impl.hpp
ujjwal-shekhar: Indeed it was leftover from some print debugging, thank you for pointing it out. I have removed this.

libs/core/serialization/include/hpx/serialization/traits/is_bitwise_serializable.hpp
ujjwal-shekhar: Noted, could we not mark this comment as resolved so we remember to address this before finalizing this PR.

libs/core/serialization/include/hpx/serialization/access.hpp
ujjwal-shekhar: It was pushed up accidentally, I have fixed this now. The only intended change here was to remove this condition entirely in the presence of refl-serialize feats' availability. I.e, to enclose it in `#if !defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)`

libs/core/serialization/include/hpx/serialization/input_archive.hpp
ujjwal-shekhar: Done.

N.B.: To ensure that all conditions are well-formed, the compiler still evaluates the trait in `hpx::traits::has_struct_serialization_v<T>`. This leads to some compile-time warnings for `reflection_nested_templ_object_test`. Could we not mark this comment as resolved, so I can return to this warning later?

.github/workflows/linux_release_clang_p2996_reflection.yml
ujjwal-shekhar: Indeed, this is the same image as the template I had used for reference. I have swapped it out with my docker image.

libs/core/config/include/hpx/config/reflection.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+#pragma once    // currently only on Clang p2996 fork
+
+#include <hpx/config/defines.hpp>
+
+// By default allow auto generation of serialization functions using C++26 reflection
+#if defined(HPX_HAVE_CXX26_EXPERIMENTAL_META) &&                               \
+    defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
+
+// TODO: Should I put default #includes here?
```
hkaiser: I think you can remove this header file altogether.

libs/core/serialization/include/hpx/serialization/access.hpp
```diff
@@ -39,6 +39,10 @@ namespace hpx::serialization {
         {
             serialize(ar, t, 0);
         }
+
+        // Forward declare
+        HPX_CXX_EXPORT template <typename Archive, typename T>
+        void refl_serialize(Archive& ar, T& t, unsigned /*version*/);
```
hkaiser, line 46: This should be forward-declared only `#if defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)`

libs/core/serialization/include/hpx/serialization/input_archive.hpp
```diff
@@ -151,18 +152,24 @@ namespace hpx::serialization {
...
+                    hpx::traits::has_struct_serialization_v<T>)
                 {
                     // struct serialization
                     access::serialize(*this, t, 0);
                 }
+                else if constexpr (has_serialize)
+                {
+                    // non-bitwise normal serialization
+                    access::serialize(*this, t, 0);
+                }
```
hkaiser, line 152: I think we should keep checking for explicit serialization support before falling back to any of the auto-serialization.

libs/core/serialization/include/hpx/serialization/input_archive.hpp
```diff
@@ -151,18 +152,24 @@ namespace hpx::serialization {
 #endif
                     load_binary(&t, sizeof(t));
                 }
-                else if constexpr (hpx::traits::has_struct_serialization_v<T>)
+                else if constexpr (has_refl_serialize ||
+                    hpx::traits::has_struct_serialization_v<T>)
                 {
                     // struct serialization
```
hkaiser: This comment needs an update

libs/core/serialization/include/hpx/serialization/output_archive.hpp
```diff
@@ -250,18 +251,24 @@ namespace hpx::serialization {
...
+                    hpx::traits::has_struct_serialization_v<T>)
                 {
                     // struct serialization
                     access::serialize(*this, t, 0);
                 }
+                else if constexpr (has_serialize)
+                {
+                    // non-bitwise normal serialization
+                    access::serialize(*this, t, 0);
+                }
```
hkaiser, line 251: Same, I believe the previous ordering was correct

libs/core/serialization/include/hpx/serialization/serialization_fwd.hpp
```diff
@@ -42,6 +42,15 @@ namespace hpx::serialization {
...
         void serialize_one(input_archive& ar, T& t);
+
+        // #if defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
+        //         // This function uses C++26 reflection capabilities to generate
+        //         // serialization functions for types that don't have them already.
+        //         // This is forward declared here and defined in refl_serialize_impl.hpp
+        //         // to avoid circular dependencies.
+        //         HPX_CXX_EXPORT template <typename Archive, typename T>
+        //         void refl_serialize(Archive& ar, T& t, unsigned /*version*/);
+        // #endif
```
hkaiser: Can this be removed now?

ujjwal-shekhar: Done.

libs/core/serialization/include/hpx/serialization/access.hpp
ujjwal-shekhar: Done.

libs/core/serialization/CMakeLists.txt
ujjwal-shekhar: Done.

libs/core/serialization/include/hpx/serialization/output_archive.hpp
ujjwal-shekhar: Apologies for the misunderstanding; I've corrected the ordering.

I realized the previous `std::array` issue was actually only a dispatch priority bug in `access.hpp`, not a problem with bitwise traits. Putting the user-defined serialize before bitwise dispatch ensures `std::array` and similar types behave as expected. I've verified this with the existing `reflection_array.cpp` test, but please let me know if we should have more isolation tests.

libs/core/serialization/include/hpx/serialization/detail/polymorphic_type_registrar.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+namespace hpx::serialization::detail {
+
+    template <typename... Ts>
+    struct register_types;
+
+    template <>
+    struct register_types<>
+    {
+        static void ensure_instantiated() {}
```
hkaiser: I think these functions can be `constexpr` and `noexcept`.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,102 @@
...
+    template <std::size_t N>
+    struct fixed_string
+    {
+        char data[N]{};
+
+        consteval fixed_string() = default;
+
+        consteval fixed_string(std::string_view str)
+        {
+            std::copy_n(str.data(), std::min(N - 1, str.size()), data);
```
hkaiser:
```suggestion
            std::copy_n(str.data(), (std::min)(N - 1, str.size()), data);
```
This prevents `min` from being macro expanded on some platforms.

libs/core/serialization/include/hpx/serialization/detail/refl_serialize_impl.hpp
```diff
@@ -0,0 +1,86 @@
...
+    HPX_CXX_EXPORT template <typename Archive, typename T>
+    void refl_serialize(Archive& ar, T& t, unsigned /*version*/)
+    {
+        constexpr auto ctx = std::meta::access_context::unchecked();
+
+        // Serialize all bases
+        template for (constexpr auto base_info :
+            std::define_static_array(std::meta::bases_of(^^T, ctx)))
+        {
+            if constexpr (std::meta::is_public(base_info))
```
hkaiser: What will happen to `private` and `protected` bases?

libs/core/serialization/include/hpx/serialization/detail/refl_serialize_impl.hpp
```diff
@@ -0,0 +1,86 @@
...
+                    "Reflection serialization does not support bitfields");
+
+                using MemberType = typename[:std::meta::type_of(member):];
+                constexpr size_t offset = offset_info.bytes;
+
+                if constexpr (std::is_const_v<T>)
+                {    // save path
+                    auto* member_ptr = reinterpret_cast<MemberType const*>(
+                        reinterpret_cast<std::byte const*>(std::addressof(t)) +
+                        offset);
```
hkaiser: It might be worth creating a facility `member_from_offset<MemberType>(t, offset)` to hide all the casting. Also, using `hpx::bit_cast` would relax possible `constexpr` restrictions.

hkaiser: That facility could also handle the constness of `T` by providing two overloads.

libs/core/serialization/include/hpx/serialization/output_archive.hpp
hkaiser: Excellent.

libs/core/serialization/tests/unit/refl/reflection_array.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include <vector>
+
+/*
+This is a failing test case for serialization of arrays
+We will work on fixing this issue
+(Primary suspect: array.hpp and is_bitwise_serializable.hpp)
+The traits specializations for hpx::serialization::array<T> need to be
+carefully designed to avoid incorrect assumptions about the bitwise
+serializability of arrays containing non-bitwise-serializable
+*/
```
hkaiser: This comment seems to be obsolete now.

libs/core/serialization/tests/unit/refl/reflection_array.cpp
```diff
@@ -0,0 +1,119 @@
...
+    int age;
+    std::string name;
+
+    bool operator==(person_pod const& rhs) const
+    {
+        return age == rhs.age && name == rhs.name;
+    }
+};
+
+struct final_array_boss
```
hkaiser, line 25: LOL, I like the name!

ujjwal-shekhar: haha, it was indeed the final boss test xD

libs/core/serialization/include/hpx/serialization/detail/polymorphic_nonintrusive_factory.hpp
```diff
@@ -37,6 +41,17 @@ namespace hpx::serialization::detail {
         ;
 #else
     {
+
+#if defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
+        [[nodiscard]] char const* operator()() const noexcept
```
hkaiser: Can this be `constexpr`?

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,149 @@
+//  Copyright (c) 2026 Ujjwal Shekhar
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#if defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
```
hkaiser, line 11: You'll have to `#include <hpx/serialization/config/defines.hpp>` to guarantee the macro is visible.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,149 @@
...
+namespace hpx::serialization::detail {
+    // A simple wrapper around a char array that can be constructed
+    // and concatenated at compile time.
+    template <std::size_t N>
+    struct fixed_string
+    {
+        char data[N + 1]{};    // +1 for null terminator
+        static constexpr std::size_t size = N;
+
+        constexpr fixed_string(std::string_view sv)
```
hkaiser:
```suggestion
        constexpr fixed_string(std::string_view sv) noexcept
```

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,149 @@
...
+        {
+            for (std::size_t i = 0; i < sv.size(); ++i)
+                data[i] = sv[i];
+        }
+
+        template <std::size_t M>
+        constexpr auto operator+(fixed_string<M> const& other) const
```
hkaiser:
```suggestion
        constexpr auto operator+(fixed_string<M> const& other) const noexcept
```

hkaiser: Same can be done for all `constexpr` functions below.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+    template <auto StringGetter>
+    static constexpr auto make_fixed()
+    {
+        constexpr std::string_view sv = StringGetter();
+        return fixed_string<sv.size()>(sv);
+    }
+
+    // Primary template for non-template types
+    HPX_CXX_EXPORT template <typename T>
```
hkaiser: We have changed the macro exporting symbols to C++ modules (for symbols in `libs/core`) to
```suggestion
    HPX_CXX_CORE_EXPORT template <typename T>
```

libs/core/serialization/include/hpx/serialization/detail/refl_serialize_impl.hpp
```diff
@@ -0,0 +1,80 @@
...
+    constexpr auto* member_from_offset(T* base, std::size_t offset)
+    {
+        using ByteType =
+            std::conditional_t<std::is_const_v<T>, std::byte const, std::byte>;
+        using Type = std::conditional_t<std::is_const_v<T>, MemberType const,
+            MemberType>;
+
+        return reinterpret_cast<Type*>(
+            reinterpret_cast<ByteType*>(base) + offset);
+    }
```
hkaiser: Can we hide the pointer-uglyness even more by changing the function to:
```suggestion
    template <typename MemberType, typename T>
    constexpr decltype(auto) member_from_offset(T& base, std::size_t offset)
    {
        using ByteType =
            std::conditional_t<std::is_const_v<T>, std::byte const, std::byte>;
        using Type = std::conditional_t<std::is_const_v<T>, MemberType const,
            MemberType>;

        return reinterpret_cast<Type&>(
            *(reinterpret_cast<ByteType*>(std::addressof(base)) + offset));
    }
```

libs/core/serialization/include/hpx/serialization/access.hpp
```diff
@@ -109,13 +115,23 @@ namespace hpx::serialization {
...
+#if !defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
                 else if constexpr (hpx::traits::has_struct_serialization_v<dT>)
                 {
                     // This is automatic serialization for types that are simple
                     // (brace-initializable) structs, that means every struct's
                     // field has to be serializable and public.
                     serialize_struct(ar, t, 0);
                 }
+#endif
+#if defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
```
hkaiser:
```suggestion
#else
```

libs/core/serialization/include/hpx/serialization/macros.hpp
```diff
@@ -461,3 +461,29 @@
...
     /**/
+
+#if defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
+#define HPX_POLYMORPHIC_AUTO_REGISTER(Class)                                   \
+    namespace hpx::serialization::detail {                                     \
+        inline struct HPX_PP_CAT(register_class_helper_, __LINE__)             \
+        {                                                                      \
+            HPX_PP_CAT(register_class_helper_, __LINE__)()                     \
+            {                                                                  \
+                [[maybe_unused]] auto& _ =                                     \
```
hkaiser:
```suggestion
                [[maybe_unused]] static auto& _ =                                     \
```

hkaiser:
`inspect` and the cmake file checker are complaining about problems, please have a look. The other CircleCI problem are known and unrelated.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: Thank you for pointing this out, I have made the necessary changes.

libs/core/serialization/include/hpx/serialization/detail/polymorphic_nonintrusive_factory.hpp
ujjwal-shekhar: Indeed, I have edited this.

libs/core/serialization/include/hpx/serialization/detail/refl_serialize_impl.hpp
ujjwal-shekhar: I have added an implementation for the above!

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: This part of code has now been removed, so the change isn't needed.

libs/core/serialization/include/hpx/serialization/detail/polymorphic_type_registrar.hpp
ujjwal-shekhar: I am considering shifting `ensure_instantiated` and other improvements to the core auto-serialization facility to a future PR (which will also include a DFS of polymorphic types for semi-automation of polymorphic types, compile-time checks to see if the user missed a register; etc).

libs/core/serialization/include/hpx/serialization/unordered_map.hpp
ujjwal-shekhar: This PR has been merged:)

libs/core/serialization/include/hpx/serialization/detail/polymorphic_nonintrusive_factory.hpp
ujjwal-shekhar: The latest changes indeed dispatch any `serialize` calls to `access::serialize`. However, I had to make a `polymorphic_nonintrusive_factory_impl.hpp` file to avoid circular dependencies with `access.hpp` . The original dispatch here was to the serialize function of the polymorphic type; please let me know if there is a caveat that I might be missing!

ujjwal-shekhar:
Thank you for the comments @hkaiser !

I have made some changes addressing most of the comments. Before wrapping up the rest, I will focus on fixing the new CI script with cpp26 reflection (as I expect there to be some bugs when testing the full suite under enabled reflection changes `:p`)

libs/core/serialization/include/hpx/serialization/detail/polymorphic_nonintrusive_factory.hpp
hkaiser: Looks ok now, I believe.

libs/core/serialization/include/hpx/serialization/detail/refl_serialize_impl.hpp
```diff
@@ -0,0 +1,83 @@
...
+// This file is only ever included by access.hpp
+// but we will still guard against direct inclusion
+#if defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
+#include <cstddef>
+#include <experimental/meta>
+#include <memory>
+
+namespace hpx::serialization::detail {
+    template <typename MemberType, typename T>
+    constexpr decltype(auto) member_from_offset(T& base, std::size_t offset)
```
hkaiser: This function is definitely `noexcept`. Also, in order to avoid potential issues when using `reinterpret_cast`  in a `constexpr` function I'd suggest using `std::bit_cast` instead.

libs/core/serialization/include/hpx/serialization/detail/refl_serialize_impl.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+namespace hpx::serialization::detail {
+    template <typename MemberType, typename T>
+    constexpr decltype(auto) member_from_offset(
+        T& base, std::size_t offset) noexcept
+    {
+        using Type = std::conditional_t<std::is_const_v<T>, MemberType const,
+            MemberType>;
+
+        auto base_addr = std::bit_cast<std::uintptr_t>(std::addressof(base));
```
hkaiser: I think that this will do the wrong thing, shouldn't we do the pointer arithmetic using `std::byte`?
```suggestion
        auto base_addr = std::bit_cast<std::byte*>(std::addressof(base));
```

libs/core/serialization/include/hpx/serialization/input_archive.hpp
```diff
@@ -118,6 +118,13 @@
                     access::has_serialize_v<T> || std::is_empty_v<T> ||
                     hpx::traits::has_serialize_adl_v<T>;
 
+#if defined(HPX_HAVE_CXX26_EXPERIMENTAL_META) &&                               \
+    defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
+                constexpr bool has_refl_serialize = true;
+#else
+                constexpr bool has_refl_serialize = false;
```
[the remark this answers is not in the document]
hkaiser: Github doesn't let me resolve this comment (which is wrong to begin with), please ignore it.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,152 @@
...
+        {
+            fixed_string<N + M> res("");
+            for (std::size_t i = 0; i < N; ++i)
+                res.data[i] = data[i];
+            for (std::size_t i = 0; i < M; ++i)
+                res.data[N + i] = other.data[i];
+            return res;
+        }
+    };
+
```
hkaiser, line 60: Would it possible to add a deduction guide to simplify expressions that use `fixed_string`:
```
template<std::size_t N>
fixed_string(char const (&)[N]) -> fixed_string<N>;
```

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,152 @@
...
+    struct fixed_string
+    {
+        char data[N + 1]{};    // +1 for null terminator
+        static constexpr std::size_t size = N;
+
+        constexpr fixed_string(std::string_view sv) noexcept
+        {
+            for (std::size_t i = 0; i < sv.size(); ++i)
+                data[i] = sv[i];
+        }
```
hkaiser, line 45: You may want to explicitly append a '\0' at the end, that allows to leave the `char data[N + 1]` above uninitialized.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,152 @@
...
+            else
+            {
+                // The lambda templ param trick allows this computation
+                // to be done at compile time.
+                // TODO: Is there a more elegant way to do this?
+                return []<std::size_t... Is>(std::index_sequence<Is...>) {
+                    return (... +
+                        (fixed_string<(Is == 0 ? 0 : 1)>(Is == 0 ? "" : ",") +
+                            make_fixed<qualified_name_of<Args>::get>()));
+                }(std::make_index_sequence<sizeof...(Args)>{});
```
hkaiser: With the above deduction guide, this can be written as:
```suggestion
                return []<std::size_t... Is>(std::index_sequence<Is...>) {
                    return (... +
                        (fixed_string(Is == 0 ? "" : ",") +
                            make_fixed<qualified_name_of<Args>::get>()));
                }(std::make_index_sequence<sizeof...(Args)>{});
```

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,152 @@
...
+    private:
+        static constexpr auto dT = dealias(^^T);
+        static constexpr auto scoped_name = scope_builder<dT>::value;
+
+        // Recursively call qualified_name_of for each arg
+        static constexpr auto get_args_name() noexcept
+        {
+            if constexpr (sizeof...(Args) == 0)
+            {
+                return fixed_string<0>("");
```
hkaiser:
```suggestion
                return fixed_string("");
```

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,152 @@
...
+            else
+            {
+                constexpr auto parent = std::meta::parent_of(Scope);
+                constexpr auto prefix = scope_builder<parent>::value;
+                constexpr auto id = std::meta::identifier_of(Scope);
+                constexpr auto name = fixed_string<id.size()>(id);
+
+                if constexpr (prefix.size > 0)
+                {
+                    return prefix + fixed_string<2>("::") + name;
```
hkaiser:
```suggestion
                    return prefix + fixed_string("::") + name;
```

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+    // Helper to recursively build the enclosing namespaces of a type
+    template <std::meta::info Scope>
+    struct scope_builder
+    {
+        static constexpr auto get_value() noexcept
+        {
+            if constexpr (Scope == ^^::)
+            {
+                return fixed_string<0>("");
```
hkaiser:
```suggestion
                return fixed_string("");
```

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,152 @@
...
+namespace hpx::serialization::detail {
+    // A simple wrapper around a char array that can be constructed
+    // and concatenated at compile time.
+    template <std::size_t N>
+    struct fixed_string
+    {
+        char data[N + 1]{};    // +1 for null terminator
+        static constexpr std::size_t size = N;
+
+        constexpr fixed_string(std::string_view sv) noexcept
```
hkaiser: You could add a constructor that accepts a constant C-string to avoid having to create a `string_view` just for the purpose of constructing the `fixed_string`:
```suggestion
        constexpr fixed_string(char const (& sv)[N]) noexcept
```
This would also avoid constructing the fixed_string with more characters than it can hold.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+        template <std::size_t M>
+        constexpr auto operator+(fixed_string<M> const& other) const noexcept
+        {
+            fixed_string<N + M> res("");
+            for (std::size_t i = 0; i < N; ++i)
+                res.data[i] = data[i];
+            for (std::size_t i = 0; i < M; ++i)
+                res.data[N + i] = other.data[i];
+            return res;
```
hkaiser, line 57: Please explicitly terminate the string with a '\0'.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -0,0 +1,152 @@
...
+                }
+            }
+        }
+        static constexpr auto value = get_value();
+    };
+
+    template <auto StringGetter>
+    static constexpr auto make_fixed() noexcept
+    {
+        constexpr std::string_view sv = StringGetter();
```
hkaiser, line 108: Where is `StringGetter()` defined?

libs/core/serialization/include/hpx/serialization/access.hpp
```diff
@@ -109,13 +115,22 @@ namespace hpx::serialization {
                     // the archive functions
                     ar.invoke(t);
                 }
+
+#if !defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)
```
hkaiser, line 119: Now, that the old struct-serialization is not needed anymore, we can #ifdef the content of the files defining that.

libs/core/serialization/include/hpx/serialization/input_archive.hpp
```diff
@@ -151,18 +157,19 @@ namespace hpx::serialization {
...
-                    // struct serialization
+                    // struct serialization or reflection-based serialization
                     access::serialize(*this, t, 0);
                 }
                 else
                 {
                     static_assert(traits::is_nonintrusive_polymorphic_v<T> ||
-                            has_serialize || optimized ||
+                            has_serialize || optimized || has_refl_serialize ||
                             hpx::traits::has_struct_serialization_v<T>,
```
hkaiser: Can we remove this now? Alternatively, re-define this as always evaluating to `std::false_type`, which may reduce the `#ifdef` clutter.

libs/core/serialization/include/hpx/serialization/detail/polymorphic_nonintrusive_factory_impl.hpp
```diff
@@ -0,0 +1,57 @@
+//  Copyright (c) 2026 Ujjwal Shekhar
+//
```
hkaiser, line 6: You might want to additionally leave in place the copyrights from the files this was taken from.

ujjwal-shekhar: I have fixed this. I also wanted to ask if I need to fix copyrights in any other places (and if I have followed to convention appropriately)?

libs/core/serialization/include/hpx/serialization/input_archive.hpp
ujjwal-shekhar: I will look into this soon.

libs/core/serialization/include/hpx/serialization/access.hpp
ujjwal-shekhar: Done.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: I pass in the function `qualified_name_of<Args>::get` as the lambda param `StringGetter`. My goal was to keep everything constexpr; I am thinking of a more elegant solution of the fold expression calling this code snippet in general too.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: Done

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: Done.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: Done.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: Done

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: Done

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: Since the parameter being passed to fixed_string is via a ternary `Is == 0 ? "" : ","`, the compiler decays this to a `const char*` instead of a `const char[N]`, making the deduction guide fail. I am also thinking of making this part more elegant generally, will revert back with a clean solution.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: Done.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: I have added this.

ujjwal-shekhar:
> Just some food for thought: what can be made `consteval` instead of being `constexpr`?

At a glance, I am inclined to believe that anything that must be evaluated _within_ the `std::meta` context should be `consteval`able. By extension, it feels like a lot of this could be `consteval`. I will experiment with this:)

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
hkaiser: I simply didn't see the template argument :/

libs/core/serialization/include/hpx/serialization/detail/polymorphic_nonintrusive_factory_impl.hpp
hkaiser: Our rule-of-thumb approach to the copyrights is that you add your own name if you have touched at least ~20% of an existing file or if you added a new file. There is no strict convention, though.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: No worries:) 🙏

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
hkaiser: Can you write it as:
```
(... + ((Is == 0 ? fixed_string("") : fixed_string(",")) + 
    make_fixed<qualified_name_of<Args>::get>()))
```
?

hkaiser: Even better would be:
```
template <std::size_t I, typename Arg>
consteval auto fragment() {
    if constexpr (I == 0) return make_fixed<qualified_name_of<Arg>::get>();
    else return fixed_string(",") + make_fixed<qualified_name_of<Arg>::get>();
}

template <typename... Args>
static consteval auto make_args_list() {
    return ([]<std::size_t... Is>(std::index_sequence<Is...>) {
        return (fragment<Is, Args>() + ...);
    })(std::make_index_sequence<sizeof...(Args)>{});
}
```

libs/core/serialization/include/hpx/serialization/input_archive.hpp
ujjwal-shekhar: Should we make an `is_serialization_supported_v` trait to keep the `static_assert` clean? For backwards compatibility (non-reflection), we would have to retain the assert.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: I agree, this is a much cleaner approach! I have added this.

ujjwal-shekhar:
@hkaiser, In the CI job with reflection enabled, would you recommend we run the full suite or just the serialization tests? The full suite would turn it into a long job so I have limited it to just the serialization tests for now.

hkaiser:
> @hkaiser, In the CI job with reflection enabled, would you recommend we run the full suite or just the serialization tests? The full suite would turn it into a long job so I have limited it to just the serialization tests for now.

Just the serialization tests for now, I guess. We will soon move this repository to a new github organization where we will have access to an enterprise license. We can re-evaluate the CI situation then.

libs/core/serialization/include/hpx/serialization/input_archive.hpp
hkaiser: That's a nice idea. It would centralize and simplify things. Also, I don't think introducing such a trait would harm backwards compatibility.

hkaiser:
@ujjwal-shekhar Overall, this PR seems to be almost complete - excellent work!

ujjwal-shekhar:
> @ujjwal-shekhar Overall, this PR seems to be almost complete - excellent work!

@hkaiser Thanks! I'm really happy with how the reflection logic is holding up. Once we've addressed the remaining comments, I'll move this from a draft to an open PR. Super excited to get this merged!

libs/core/serialization/include/hpx/serialization/detail/refl_serialize_impl.hpp
ujjwal-shekhar: I have been able to isolate a nice edge case that breaks the current implementation for protected/private bases: 

<details>
<summary> Edge case here </summary>

```
namespace MixedAccess {
    struct BasePublic
    {
        virtual ~BasePublic() = default;
        int b_val = 1;
    };
    struct BasePrivate
    {
        int c_val = 2;
    };
    struct BaseProtected
    {
        int d_val = 3;
    };

    // A derives from all three with different access levels
    struct DerivedA
      : public BasePublic
      , protected BaseProtected
      , private BasePrivate
    {
        int a_val = 4;
    };
}    // namespace MixedAccess

HPX_POLYMORPHIC_AUTO_REGISTER(MixedAccess::BasePublic)
HPX_POLYMORPHIC_AUTO_REGISTER(MixedAccess::BaseProtected)
HPX_POLYMORPHIC_AUTO_REGISTER(MixedAccess::BasePrivate)
HPX_POLYMORPHIC_AUTO_REGISTER(MixedAccess::DerivedA)

int main() {
        std::vector<char> buffer;
        std::unique_ptr<MixedAccess::BasePublic> input =
            std::make_unique<MixedAccess::DerivedA>();

        {
            hpx::serialization::output_archive oarch(buffer);
            oarch << input;
        }

        std::unique_ptr<MixedAccess::BasePublic> output;
        {
            hpx::serialization::input_archive iarch(buffer);
            iarch >> output;
        }

        auto* final_a = reinterpret_cast<MixedAccess::DerivedA*>(output.get());

        auto* c_ptr = reinterpret_cast<MixedAccess::BasePrivate*>(final_a);
        auto* d_ptr = reinterpret_cast<MixedAccess::BaseProtected*>(final_a);

        HPX_TEST_EQ(c_ptr->c_val, 2); /*FAILS*/
        HPX_TEST_EQ(d_ptr->d_val, 3); /*FAILS*/        
}
```

</details>

This fails because `base_object_type<>::serialize()` performs a static_cast<Base&>(derived), which is illegal here. I’ve explored calculating addresses manually and bypass the cast. However, this currently fails due to an infinite loop of polymorphic factory and save logic.

I am seeing if I can find a way to work around this. We might have to require a `friend` declaration or an intrusive `getBase()` for protected/private bases if all options dry up.

ujjwal-shekhar: I have made some changes to handle this case!

libs/core/serialization/include/hpx/serialization/input_archive.hpp
ujjwal-shekhar: For this, I have made some changes to `input_archive.hpp` and `output_archive.hpp`. Added a new `is_serialization_supported.hpp` with traits. Please let me know if the design looks good and if there are any changes I should make.

libs/core/serialization/include/hpx/serialization/traits/is_serialization_supported.hpp
```diff
@@ -0,0 +1,51 @@
...
+            true;
+#else
+            false;
+#endif
+
+        static constexpr bool value =
+            hpx::traits::is_nonintrusive_polymorphic_v<T> ||
+            hpx::traits::has_struct_serialization_v<T> || has_serialize ||
+            has_optimized || has_refl_serialize;
+    };
```
hkaiser, line 47: I think we should turn this into a full meta-function by adding:
```
        using type = std::bool_constant<value>;
```

libs/core/serialization/include/hpx/serialization/base_object.hpp
```diff
@@ -32,8 +40,34 @@ namespace hpx::serialization {
...
+                return std::nullopt;
+            }();
+
+            static_assert(offset.has_value(),
+                "Base class not found in derived class's base list");
+
+            auto base_addr = std::bit_cast<std::byte*>(std::addressof(d_));
+            auto* base_ptr = std::bit_cast<Base*>(base_addr + offset.value());
+
+            access::serialize(ar, *base_ptr, 0);
```
hkaiser:
```suggestion
            access::serialize(ar, detail::member_from_offset<Base>(d_, offset.value()), 0);
```
possibly while renaming `member_from_offset` to something more generic.

libs/core/serialization/include/hpx/serialization/base_object.hpp
```diff
@@ -17,7 +18,28 @@
...
+        {
+            using Type = std::conditional_t<std::is_const_v<T>,
+                MemberType const, MemberType>;
+
+            auto base_addr = std::bit_cast<std::byte*>(std::addressof(base));
+            auto member_ptr = std::bit_cast<Type*>(base_addr + offset);
+
+            return *member_ptr;
+        }
+    }    // namespace detail
```
hkaiser, line 44: This should most likely be inside a `#if defined(HPX_SERIALIZATION_HAVE_ALLOW_AUTO_GENERATE)`

libs/core/serialization/include/hpx/serialization/traits/is_serialization_supported.hpp
ujjwal-shekhar: Done!

libs/core/serialization/include/hpx/serialization/base_object.hpp
ujjwal-shekhar: Done.

hkaiser:
The CMake build system checker reports:
```
Missing hpx/serialization/traits/is_serializable.hpp in libs/core/serialization/CMakeLists.txt
```
Could you please add that file?

ujjwal-shekhar:
> The CMake build system checker reports:
> 
> ```
> Missing hpx/serialization/traits/is_serializable.hpp in libs/core/serialization/CMakeLists.txt
> ```
> 
> Could you please add that file?

Done @hkaiser ! Apologies for having missed this, I had accidentally removed this line while adding the new trait file to the CMakeList.

hkaiser:
@ujjwal-shekhar thank you again for your persistence and effort! This is a major milestone for HPX!

Issue #6728, opened 2025-06-24
Building on ubuntu-24.04-arm


OliverSchmitz:
fyi, I tried to build master on ubuntu-24.04-arm (by a Github action) using Conda and our conda-forge/hpx-feedstock setup. It failed with the one below, which I could solve by adding `-D HPX_WITH_GENERIC_CONTEXT_COROUTINES=On` to CMake and then the build succeeds :)
I get the impression that setting this option should not be required but detected and added automatically (?), so I just share my experiences

```
...s/hpx_coroutines.dir/src/detail/context_base.cpp.o.d -o libs/core/coroutines/CMakeFiles/hpx_coroutines.dir/src/detail/context_base.cpp.o -c $SRC_DIR/libs/core/coroutines/src/detail/context_base.cpp
 │ │ In file included from $SRC_DIR/libs/core/coroutines/include/hpx/coroutines/detail/context_impl.hpp:126,
 │ │                  from $SRC_DIR/libs/core/coroutines/include/hpx/coroutines/detail/context_base.hpp:36,
 │ │                  from $SRC_DIR/libs/core/coroutines/src/detail/context_base.cpp:12:
 │ │ $SRC_DIR/libs/core/coroutines/include/hpx/coroutines/detail/context_linux_x86.hpp: In member function 'void hpx::threads::coroutines::detail::lx::x86_linux_context_impl_base::prefetch() const':
 │ │ $SRC_DIR/libs/core/coroutines/include/hpx/coroutines/detail/context_linux_x86.hpp:206:41: error: static assertion failed
 │ │   206 |             static_assert(sizeof(void*) == 4);
 │ │       |                           ~~~~~~~~~~~~~~^~~~
 │ │ $SRC_DIR/libs/core/coroutines/include/hpx/coroutines/detail/context_linux_x86.hpp:206:41: note: the comparison reduces to '(8 == 4)'
 │ │ ninja: build stopped: subcommand failed.
 │ │ × error Script failed with status 1
```

hkaiser:
@OliverSchmitz thanks for reporting this. I believe that for ARM, HPX needs to be built using the `-DHPX_WITH_GENERIC_CONTEXT_COROUTINES=On` CMake configuration option. I'm not sure if the Conda package was built that way. 

@kordejong: could you verify, please?

OliverSchmitz:
[...]

PR #6860, opened 2026-01-24
cmake: fix cmake-format issues for ARM architecture support

akashadsare:
Changes made:
- Fixed cmake-format formatting by properly aligning the multi-line condition
- Added support for all ARM architecture variants (arm, armv5, armv6, armv7)
- Updated the comment to reflect the inclusion of ARM variants


hkaiser:
Why can't you apply this to the original PR #6728?

akashadsare:
@hkaiser I don't have idea as to why . but i would love hear your opinion.

hkaiser:
This is not needed as the change is already part of #6728

PR #6864, opened 2026-01-29
[Serialization] [Feature] Add serialization for multi containers and unordered_set

ujjwal-shekhar:
Adds in-built serialization for `multiset`, `multimap`, `unordered_set`, `unordered_multiset` and `unordered_multimap` with corresponding tests. Improves the implementation for all set/map containers and their unordered/multi variants.

## Proposed Changes

  -  New file for `unordered_set.hpp` inside `libs/core/serialization/include/hpx/serialization/`.
  -  Code for corresponding multi container for each of them was added in the corresponding non-multi container `.hpp` file.
  -  Tests were added based on the existing tests for `unordered_map`, `map` and `set`.
  -  `detail::load_collection` now handles containers with `emplace` and `emplace_back`.
  -  set/map containers and their unordered/multi variants now dispatch to `detail::save/load_collection`

## Background Context

This work is done as part of the ongoing work on using C++26 reflection for auto-serialization

## Checklist


hkaiser:
You have some spelling typos in comments:
```
libs/core/serialization/tests/unit/serialization_unordered_multiset.cpp:45: fpr  ==> for, far, fps
libs/core/serialization/tests/unit/serialization_unordered_set.cpp:45: fpr  ==> for, far, fps
```

ujjwal-shekhar:
> There is a lot of code duplication going on (and part of that is there for a long time). Do you think you could change the code to use the generic `detail::save_collection` and `detail::load_collection` functions defined in https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/serialization/include/hpx/serialization/detail/serialize_collection.hpp? That would also add support for serializing non-default constructible elements in the containers.

@hkaiser I can definitely take a look into this.

libs/core/serialization/include/hpx/serialization/traits/emplace_func_traits.hpp
```diff
@@ -0,0 +1,50 @@
...
+        std::void_t<decltype(std::declval<Collection>().emplace_back(
+            std::declval<typename Collection::value_type>()))>> : std::true_type
+    {
+    };
+
+    HPX_CXX_EXPORT template <typename Collection>
+    inline constexpr bool has_emplace_back_v =
+        has_emplace_back<Collection>::value;
+
+}    // namespace hpx::traits
```
hkaiser: You could simply have used the existing `HPX_HAS_MEMBER_XXX_TRAIT_DEF()` facility instead of creating your own traits:
```
HPX_HAS_MEMBER_XXX_TRAIT_DEF(emplace);
HPX_HAS_MEMBER_XXX_TRAIT_DEF(emplace_back);
```
See for instance how the `has_reserve` is implemented here: https://github.com/STEllAR-GROUP/hpx/blob/1cce0aeb6c95a670c600bf6c8d35472ede5ce0a0/libs/core/serialization/include/hpx/serialization/detail/serialize_collection.hpp#L26-L36

libs/core/serialization/include/hpx/serialization/detail/serialize_collection.hpp
```diff
@@ -35,6 +36,24 @@ namespace hpx::serialization::detail {
...
+            c.emplace_back(HPX_FORWARD(T, t));
+        }
+        else if constexpr (hpx::traits::has_emplace_v<Container>)
+        {    // sets, maps, etc.
+            c.emplace(HPX_FORWARD(T, t));
+        }
+        else
+        {
+            static_assert(sizeof(Container) == 0,
+                "Unsupported container type for emplace_into_collection");
```
hkaiser: Instead of asserting we could simply fall back to using `insert`, which should always be available, i.e. something like:

```suggestion
            c.insert(c.end(), HPX_FORWARD(T, t));
```

libs/core/serialization/include/hpx/serialization/set.hpp
```diff
@@ -47,4 +42,26 @@ namespace hpx::serialization {
             ar << i;
         }
```
hkaiser: I think this should be replaced with `detail::save_collection(ar, set);` as well

libs/core/serialization/CMakeLists.txt
```diff
@@ -197,6 +199,7 @@ set(serialization_compat_headers
     hpx/traits/is_bitwise_serializable.hpp => hpx/serialization.hpp
     hpx/traits/needs_automatic_registration.hpp => hpx/serialization.hpp
     hpx/traits/polymorphic_traits.hpp => hpx/serialization.hpp
+    hpx/traits/emplace_func_traits.hpp => hpx/serialization.hpp
```
hkaiser: The compatibility headers are meant for backwards compatibility. There is no need to add new files to this list.

libs/core/serialization/include/hpx/serialization/traits/emplace_func_traits.hpp
ujjwal-shekhar: This is definitely cleaner than writing a custom trait. Thank you for pointing this out, I have removed the new trait file and am using this now.

libs/core/serialization/include/hpx/serialization/detail/serialize_collection.hpp
ujjwal-shekhar: Done!

libs/core/serialization/include/hpx/serialization/set.hpp
ujjwal-shekhar: Fixed.

libs/core/serialization/CMakeLists.txt
ujjwal-shekhar: Noted for future! Since we don't need the new trait anymore, changes to this file are not needed.

hkaiser:
@ujjwal-shekhar Thank you again for working on this!

PR #6870, opened 2026-02-02
standardise hpx::generate function signature

### LINKED ISSUE #6536: Non-Standard Conforming Parallel Algorithms
The following HPX algorithms deviate from the C++ standard:

- [ ] `hpx::generate`: The HPX algorithm returns a past-the-end iterator, while the standard specifies that the return type should be `void`. (See in the [standard draft](https://eel.is/c++draft/alg.generate).)
- [ ] `hpx::search`: The overload that takes a range and a searcher is missing. (However, this is probably not very critical, as there is no parallel version of this algorithm anyway.) (See in the [standard draft](https://eel.is/c++draft/alg.search).)
- [ ] `hpx::search_n`: HPX's `search_n` is effectively equivalent to  `search(first, std::next(first, n), s_first, s_last)`, but according to the standard, `search_n` should find a contiguous sequence of `n` identical elements equal to a given value. (See in the [standard draft](https://eel.is/c++draft/alg.search).)

sleepingeight:
Fixes #6536 (Partially)

> The HPX algorithm returns a past-the-end iterator, while the standard specifies that the return type should be void. (See in the [standard draft](https://eel.is/c++draft/alg.generate).)

This PR aims to fix the above deviation.

## Checklist


hkaiser:
@sleepingeight Thanks for this fix, nice catch!

PR #6871, opened 2026-02-02
fix assertion failure in distributed barrier

### LINKED ISSUE #6430: HPX distributed barrier runs into assertion error
## Expected Behavior

The `hpx::distributed::barrier::synchronize()` runs successfully.

## Actual Behavior

It gets into an assertion error
```
{file}: /home/jiakuny/workspace/hpx-master/libs/full/collectives/src/barrier.cpp
{line}: 210
{function}: static void hpx::distributed::barrier::synchronize()
{what}: Assertion 'b[0].node_ && b[1].node_' failed: HPX(assertion_failure)
```

A simple example code to reproduce this error.
```
void run_test();
HPX_PLAIN_ACTION(run_test, run_test_action)

void run_test() {
    fprintf(stderr, "%d: run_test\n", hpx::get_locality_id());
    hpx::distributed::barrier::synchronize();
}

int hpx_main(hpx::program_options::variables_map& b_arg)
{
    std::vector<hpx::future<void>> futs;
    for (auto l : hpx::find_all_localities()) {
        futs.emplace_back(hpx::async<run_test_action>(l));
    }
    hpx::wait_all(futs);

    hpx::finalize();
    return 0;
}

int main(int argc, char* argv[])
{
    namespace po = hpx::program_options;
    po::options_description description("HPX test barrier");

    hpx::init_params init_args;
    init_args.desc_cmdline = description;

    return hpx::init(argc, argv, init_args);
}
```

## Specifications

  - HPX Version: The current master branch
  - Platform (compiler, OS): ubuntu 20.04, GCC 9.4

the-ivii:
Fixes #6430

## Proposed Changes

  - added a null check for barrier nodes in barrier::synchronize to prevent a crash when the barrier is destroyed early.
  - added a regression test barrier_synchronize to verify the fix works as expected.
  - updated the test suite cmake file to include the new verification test.

## Checklist

### DIFF
--- a/examples/1d_stencil/1d_stencil_5.cpp
+++ b/examples/1d_stencil/1d_stencil_5.cpp
@@ -64,13 +64,13 @@ public:
     }
 
     explicit partition_data(std::size_t size)
-      : data_(new double[size])
+      : data_(std::make_shared<double[]>(size))
       , size_(size)
     {
     }
 
     partition_data(std::size_t size, double initial_value)
-      : data_(new double[size])
+      : data_(std::make_shared<double[]>(size))
       , size_(size)
     {
         double base_value = initial_value * double(size);
@@ -109,7 +109,7 @@ private:
     void load(Archive& ar, unsigned int const)
     {
         ar & size_;
-        data_.reset(new double[size_]);
+        data_ = std::make_shared<double[]>(size_);
         ar& hpx::serialization::make_array(data_.get(), size_);
     }
 
--- a/examples/1d_stencil/1d_stencil_5.cpp
+++ b/examples/1d_stencil/1d_stencil_5.cpp
@@ -64,13 +64,13 @@ public:
     }
 
     explicit partition_data(std::size_t size)
-      : data_(std::make_shared<double[]>(size))
+      : data_(new double[size])
       , size_(size)
     {
     }
 
     partition_data(std::size_t size, double initial_value)
-      : data_(std::make_shared<double[]>(size))
+      : data_(new double[size])
       , size_(size)
     {
         double base_value = initial_value * double(size);
@@ -109,7 +109,7 @@ private:
     void load(Archive& ar, unsigned int const)
     {
         ar & size_;
-        data_ = std::make_shared<double[]>(size_);
+        data_.reset(new double[size_]);
         ar& hpx::serialization::make_array(data_.get(), size_);
     }
 
--- a/libs/full/collectives/src/barrier.cpp
+++ b/libs/full/collectives/src/barrier.cpp
@@ -205,8 +205,12 @@ namespace hpx { namespace distributed {
     void barrier::synchronize()
     {
         static std::atomic<std::size_t> gen = 0;
-        static std::array<barrier, 2>& b = get_global_barrier();
-        HPX_ASSERT(b[0].node_ && b[1].node_);
+        std::array<barrier, 2>& b = get_global_barrier();
+
+        if (!b[0].node_ || !b[1].node_)
+        {
+            return;
+        }
 
         b[++gen % 2].wait();
     }
--- a/libs/full/collectives/src/barrier.cpp
+++ b/libs/full/collectives/src/barrier.cpp
@@ -207,7 +207,8 @@ namespace hpx { namespace distributed {
         static std::atomic<std::size_t> gen = 0;
         std::array<barrier, 2>& b = get_global_barrier();
 
-        if (!b[0].node_ || !b[1].node_)
+        if (hpx::get_runtime_ptr() == nullptr ||
+            hpx::is_stopped_or_shutting_down() || !b[0].node_ || !b[1].node_)
         {
             return;
         }
--- a/libs/full/collectives/src/barrier.cpp
+++ b/libs/full/collectives/src/barrier.cpp
@@ -13,6 +13,7 @@
 #include <hpx/modules/execution.hpp>
 #include <hpx/modules/memory.hpp>
 #include <hpx/modules/runtime_configuration.hpp>
+#include <hpx/modules/errors.hpp>
 #include <hpx/modules/runtime_local.hpp>
 #include <hpx/modules/type_support.hpp>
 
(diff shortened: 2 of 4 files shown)


libs/full/collectives/src/barrier.cpp
```diff
@@ -206,6 +206,10 @@ namespace hpx { namespace distributed {
     {
         static std::atomic<std::size_t> gen = 0;
         static std::array<barrier, 2>& b = get_global_barrier();
+
+        if (!b[0].node_ || !b[1].node_)
```
hkaiser: I think that this is not a correct solution. The initialization of the barrier collective has either failed or is not finished if one (or both) of the embeded barrier objects are not valid.

Can you please give us a way to reproduce the original problem you;re trying to fix?

the-ivii: when barrier.cpp file is in its original state(without null check) and i ran barrier_synchronize_test(which calls 
synchronize() after hpx::init() returns). the result was segmentation fault(exit code 132)

 when i fixed this issue by checking if runtime is shutting down or stopped before accessing the barrier nodes. i ran same test again and it passed

libs/full/collectives/src/barrier.cpp
```diff
@@ -205,8 +205,22 @@ namespace hpx { namespace distributed {
...
         static std::atomic<std::size_t> gen = 0;
-        static std::array<barrier, 2>& b = get_global_barrier();
-        HPX_ASSERT(b[0].node_ && b[1].node_);
+        std::array<barrier, 2>& b = get_global_barrier();
+
+        if (!b[0].node_ || !b[1].node_)
+        {
+            if (hpx::get_runtime_ptr() == nullptr ||
+                hpx::is_stopped_or_shutting_down())
+            {
```
hkaiser: Is this a condition you see in one of the tests or other code?

libs/full/collectives/tests/unit/barrier_synchronize.cpp
```diff
@@ -0,0 +1,46 @@
...
+int main(int argc, char* argv[])
+{
+    hpx::program_options::options_description description(
+        "HPX test barrier synchronize");
+
+    hpx::init_params init_args;
+    init_args.desc_cmdline = description;
+
+    HPX_TEST_EQ(hpx::init(argc, argv, init_args), 0);
+    hpx::distributed::barrier::synchronize();
```
hkaiser: Only a few HPX facilities can be used when the runtime is not active (like here). The distributed barrier is not one of them. 

Did you see actual code that is doing this? Or is this an artificial construct meant to test your change to `synchronize()` above?

libs/full/collectives/src/barrier.cpp
the-ivii: i hit this while debugging crashes during distributed shutdown. That specific check was an early attempt, but i have refined the logic in the latest version to directly check for null barrier nodes, which handles these shutdown race conditions much more robustly.

libs/full/collectives/tests/unit/barrier_synchronize.cpp
the-ivii: this is a test designed to simulate a race condition we saw during teardown in distributed runs. It is not meant to encourage post runtime usage, but just to ensure that if a sync call overlaps with cleanup, it returns gracefully instead of triggering a hard assertion failure.

hkaiser: Wouldn't your solution just hide the problem?

Also, I'm glad you now mention under what circumstances you saw the original issue. If it happens during system teardown, then it's definitely not a problem in the initialization (as I initially assumed) but a possible sequencing error while the runtime exits (possibly caused by the sequencing of global destructors being undefined). 

IIRC, the barrier in this case is used to make sure that all localities have finished executing. 

I'd appreciate it if you could provide a stack backtrace of the original assert you were seeing. That may allow us to see which object is destroyed too late.

examples/1d_stencil/1d_stencil_5.cpp
```diff
@@ -98,10 +99,22 @@ struct partition_data
...
     {
+        ar & size_;
+        ar& hpx::serialization::make_array(data_.get(), size_);
     }
 
+    template <typename Archive>
+    void load(Archive& ar, unsigned int const /* version */)
+    {
+        ar & size_;
+        data_.reset(new double[size_]);
```
hkaiser, line 112: Could you please use `data_ = std::make_shared<double[]>(size_)` instead? 

In any case, could you please separate this change (unrelated to the barrier patch) into its own PR?

hkaiser: Another note: 1d_stencil_5 will never need to serialize this data as it demonstrates an intermediate step from purely local to distributed execution (see the comment at the top of this file. That's why the serialization functions are empty.

hkaiser:
@the-ivii Could you please rebase your branch onto master (and resolve the conflicts)?

hkaiser:
@the-ivii What are your plans regarding this PR? Will you have the time to work on it?

libs/full/collectives/tests/unit/barrier_synchronize.cpp
```diff
@@ -0,0 +1,45 @@
...
+    hpx::program_options::options_description description(
+        "HPX test barrier synchronize");
+
+    hpx::init_params init_args;
+    init_args.desc_cmdline = description;
+
+    HPX_TEST_EQ(hpx::init(argc, argv, init_args), 0);
+
+    return hpx::util::report_errors();
+}
```
hkaiser, line 42: Does this test reproduce the problem for you (without your change)?

libs/full/collectives/src/barrier.cpp
```diff
@@ -205,8 +205,13 @@ namespace hpx { namespace distributed {
     void barrier::synchronize()
     {
         static std::atomic<std::size_t> gen = 0;
-        static std::array<barrier, 2>& b = get_global_barrier();
-        HPX_ASSERT(b[0].node_ && b[1].node_);
+        std::array<barrier, 2>& b = get_global_barrier();
+
+        if (hpx::get_runtime_ptr() == nullptr ||
+            hpx::is_stopped_or_shutting_down() || !b[0].node_ || !b[1].node_)
```
hkaiser, line 212: This is not the correct solution (as discussed before) as it simply ignores the problem instead of reporting the issue.

hkaiser:
The issue reported by the clang-format CI should be fixed. Also, the MacOS builders fail your new test. Please have an eye on the CIs yourself.

hkaiser:
@the-ivii there is still the clang-format issue, also your changes cause compilations errors. Please have a look.

hkaiser:
@the-ivii ping?

hkaiser:
@the-ivii Closing this as abandoned. Please re-open if there is interest in moving this forward.

PR #6872, opened 2026-02-02
Added random repeating numbers in tests and fixed sentinel tests for extrema algorithms

### LINKED ISSUE #6627: Extrema algorithms with repeated values

0lem1of1:
Fixes #6627 

## Proposed Changes

  -Added random_repeat utility function to generate vectors with repeated/duplicate values for testing extrema algorithms (min_element, max_element, minmax_element)
  -Replaced random_iota with random_repeat in all extrema algorithm tests to properly test behavior with duplicate values
  -Modified sentinel tests in container_algorithms to use random_repeat with a unique sentinel value (101) placed at the desired position, ensuring sentinels work correctly with duplicate data

This PR addresses issue #6627 regarding testing extrema algorithms with repeated values.
The previous test data used random_iota which generates unique values, meaning the "first occurrence" behavior of min_element/max_element was never actually tested. By using random_repeat, we now generate data with duplicates (values 0-100 repeated across 10007 elements), which properly exercises the algorithm's requirement to return the first minimum/maximum element.
For sentinel tests in container_algorithms, we inject a unique sentinel value (101) at the desired position to prevent the sentinel from stopping prematurely when duplicate values exist in the data
Thanks to the authors of PR #6724 for initial direction
## Checklist


libs/core/algorithms/tests/unit/algorithms/test_utils.hpp
```diff
@@ -208,6 +208,33 @@ namespace test {
...
+        std::vector<T> c(size);
+        std::random_device rd;
+        std::mt19937 g(rd());
+        std::uniform_int_distribution<T> dist(0, max_value);
+        for (std::size_t i = 0; i < size; ++i)
+        {
+            c[i] = dist(g);
+        }
+        return c;
+    }
```
hkaiser, line 223: What is your rationale for providing two versions of the same functionality (one for sequences of `std::size_t` and one (identical) generic version? Wouldn't the generic version be sufficient?

libs/core/algorithms/tests/unit/container_algorithms/test_utils.hpp
```diff
@@ -296,6 +296,34 @@ namespace test {
...
+        std::vector<T> c(size);
+        std::random_device rd;
+        std::mt19937 g(rd());
+        std::uniform_int_distribution<T> dist(0, max_value);
+        for (std::size_t i = 0; i < size; ++i)
+        {
+            c[i] = dist(g);
+        }
+        return c;
+    }
```
hkaiser, line 311: Same question here: why two identical versions of the new function `random_repeat`?

libs/core/algorithms/tests/unit/algorithms/test_utils.hpp
0lem1of1: You're right that the generic version would be sufficient on its own. I followed the existing pattern from `random_iota` (which also has both a `std::size_t` specialization and a generic template version) for consistency. The non-template version serves as a convenience overload so callers don't need to explicitly specify `<std::size_t>` for the most common case.
That said, if you'd prefer I remove the `std::size_t` version and keep only the generic template, I'm happy to make that change.

libs/core/algorithms/tests/unit/container_algorithms/test_utils.hpp
0lem1of1: same here , i tried to follow the existing pattern from `random_iota`

libs/core/algorithms/tests/unit/algorithms/test_utils.hpp
hkaiser: Yes, let's remove the specialized version. FWIW, I don't think users need to specify the `std::size_t` even for the generic case, `T` will be deduced by the compiler.

0lem1of1: very well , ill make the changes

PR #6875, opened 2026-02-04
implement serialization for partition_data in 1d_stencil_5

### LINKED ISSUE #6874: tests.examples.1d_stencil.1d_stencil_5 is failing in some CI runs
It looks like the partition_data struct in the example has an empty serialize function. Even though the example is designed to run locally, HPX actions can still trigger serialization. Because the serialization logic is missing, the data isn't actually being packed or sent. When the action tries to process the "received" object, it encounters invalid state (like a null pointer or zero size), leading to a crash or heap corruption.

the-ivii:
Fixes #6874

## Proposed Changes

  - i fixed 1d_stencil_5 crash by implementing the missing serialization logic for partition_data,which was previously empty and this ensures the data is correctly packed and transmitted during HPX actions, preventing the system from crashing on invalid empty objects.

## Checklist

### DIFF
--- a/examples/1d_stencil/1d_stencil_5.cpp
+++ b/examples/1d_stencil/1d_stencil_5.cpp
@@ -20,6 +20,7 @@
 #if !defined(HPX_COMPUTE_DEVICE_CODE)
 #include <hpx/hpx.hpp>
 #include <hpx/hpx_init.hpp>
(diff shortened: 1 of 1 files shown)


hkaiser:
@the-ivii thank you for working on this!

PR #6877, opened 2026-02-04
Changed reduce implementation and added tests

### LINKED ISSUE #6647: Incorrect reduce implementation

BhoomishGupta:
Fixes #6647

## Proposed Changes
  Created a helper function that reduces the partition value without the need of initial value..
  Added a test case that was mentioned in the issue #6647. 
  Corrected a typo in documentation.

I only did changes in :
1) hpx\libs\core\algorithms\include\hpx\parallel\algorithms\reduce.hpp
2) hpx\libs\core\algorithms\tests\regressions\CMakeLists.txt
3) the rest were changes due to clang and editor.config

And created this :
1) hpx\libs\core\algorithms\tests\regressions\reduce_6647.cpp

Previously, the implementation assumed that the *first was convertible to T. This assumption does not hold for the actual function logic required here. This PR removes that dependency, ensuring the type handling is correct.

## Checklist

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
@@ -456,7 +456,7 @@ namespace hpx::parallel { namespace detail {
             std::next(part_begin, 2), part_size - 2, HPX_MOVE(init), r);
     }
 
-    HPX_CXX_EXPORT template <typename T>
+    HPX_CXX_CORE_EXPORT template <typename T>
     struct reduce : public algorithm<reduce<T>, T>
     {
         constexpr reduce() noexcept
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
@@ -382,59 +382,156 @@ namespace hpx::parallel {
     namespace detail {
 
         /// \cond NOINTERNAL
-        HPX_CXX_CORE_EXPORT template <typename T>
-        struct reduce : public algorithm<reduce<T>, T>
+
+        // Custom executor parameters for reduce algorithm to prevent single-element partitions.
+        // reduce_partition requires at least 2 elements per partition because it initializes
+        // the accumulator via op(*first, *next(first)), which is the only way to produce a T
+        // from value_type elements when T may differ from value_type (e.g. minmax).
+        struct reduce_executor_parameters
         {
-            constexpr reduce() noexcept
-              : algorithm<reduce, T>("reduce")
+            template <typename Executor>
+            HPX_FORCEINLINE constexpr std::pair<std::size_t, std::size_t>
+            adjust_chunk_size_and_max_chunks(Executor&&,
+                std::size_t num_elements, std::size_t num_cores,
+                std::size_t /*max_chunks*/,
+                std::size_t chunk_size) const noexcept
             {
-            }
-
-            template <typename ExPolicy, typename InIterB, typename InIterE,
-                typename T_, typename Reduce>
-            static constexpr T sequential(ExPolicy&& policy, InIterB first,
-                InIterE last, T_&& init, Reduce&& r)
-            {
-                return detail::sequential_reduce<ExPolicy>(
-                    HPX_FORWARD(ExPolicy, policy), first, last,
-                    HPX_FORWARD(T_, init), HPX_FORWARD(Reduce, r));
-            }
-
-            template <typename ExPolicy, typename FwdIterB, typename FwdIterE,
-                typename T_, typename Reduce>
-            static decltype(auto) parallel(ExPolicy&& policy, FwdIterB first,
-                FwdIterE last, T_&& init, Reduce&& r)
-            {
-                constexpr bool has_scheduler_executor =
-                    hpx::execution_policy_has_scheduler_executor_v<ExPolicy>;
-
-                if constexpr (!has_scheduler_executor)
+                // Ensure minimum chunk size of 2
+                if (chunk_size < 2)
                 {
-                    if (first == last)
-                    {
-                        return util::detail::algorithm_result<ExPolicy, T>::get(
-                            HPX_FORWARD(T_, init));
-                    }
+                    chunk_size = (num_elements + num_cores - 1) / num_cores;
+                    chunk_size = (std::max)(chunk_size, std::size_t(2));
                 }
 
-                auto f1 = [r](FwdIterB part_begin, std::size_t part_size) -> T {
-                    return reduce_partition<ExPolicy, FwdIterB, T>(
-                        part_begin, part_size, r);
-                };
+                // chunk_size_iterator gives the last partition num_elements % chunk_size
+                // elements (or chunk_size if evenly divisible). If the remainder is 1,
+                // that partition would violate reduce_partition's >= 2 requirement.
+                // Bump chunk_size until the remainder is 0 or >= 2.
+                while (
+                    num_elements > chunk_size && num_elements % chunk_size == 1)
+                {
+                    chunk_size++;
+                }
 
-                return util::partitioner<ExPolicy, T>::call(
-                    HPX_FORWARD(ExPolicy, policy), first,
-                    detail::distance(first, last), HPX_MOVE(f1),
-                    hpx::unwrapping(
-                        [init = HPX_FORWARD(T_, init),
-                            r = HPX_FORWARD(Reduce, r)](auto&& results) -> T {
-                            return detail::sequential_reduce<ExPolicy>(
-                                hpx::util::begin(results),
-                                hpx::util::size(results), init, r);
-                        }));
+                std::size_t new_max_chunks =
+                    (num_elements + chunk_size - 1) / chunk_size;
+
+                return {chunk_size, new_max_chunks};
             }
         };
         /// \endcond
+    }    // namespace detail
+}    // namespace hpx::parallel
+
+// Specialize trait to make reduce_executor_parameters a valid executor parameters type
+namespace hpx::execution::experimental {
+
+    template <>
+    struct is_executor_parameters<
+        hpx::parallel::detail::reduce_executor_parameters> : std::true_type
+    {
+    };
+}    // namespace hpx::execution::experimental
+
+namespace hpx::parallel {
+    namespace detail {
+
+    // Helper function to reduce a partition without requiring an init value.
+    // Assumes partition size is always >= 2 (enforced by reduce_executor_parameters).
+    template <typename ExPolicy, typename FwdIterB, typename T, typename Reduce>
+    T reduce_partition(
+        FwdIterB part_begin, std::size_t part_size, Reduce const& r)
+    {
+        HPX_ASSERT(part_size >= 2);
+
+        // Combine first two elements using the reduction operator
+        T init = HPX_INVOKE(r, *part_begin, *std::next(part_begin));
+
+        if (part_size == 2)
+        {
+            return init;
+        }
+
+        // Reduce remaining elements
+        return sequential_reduce<ExPolicy>(
+            std::next(part_begin, 2), part_size - 2, HPX_MOVE(init), r);
+    }
+
+    HPX_CXX_EXPORT template <typename T>
+    struct reduce : public algorithm<reduce<T>, T>
+    {
+        constexpr reduce() noexcept
+          : algorithm<reduce<T>, T>("reduce")
+        {
+        }
+
+        template <typename ExPolicy, typename InIterB, typename InIterE,
+            typename T_, typename Reduce>
+        static constexpr T sequential(ExPolicy&& policy, InIterB first,
+            InIterE last, T_&& init, Reduce&& r)
+        {
+            return sequential_reduce<ExPolicy>(
+                HPX_FORWARD(ExPolicy, policy), first, last,
+                HPX_FORWARD(T_, init), HPX_FORWARD(Reduce, r));
+        }
+
+        template <typename ExPolicy, typename FwdIterB, typename FwdIterE,
+            typename T_, typename Reduce>
+        static decltype(auto) parallel(ExPolicy&& policy, FwdIterB first,
+            FwdIterE last, T_&& init, Reduce&& r)
+        {
+            constexpr bool has_scheduler_executor =
+                hpx::execution_policy_has_scheduler_executor_v<ExPolicy>;
+
+            // Handle empty range
+            if constexpr (!has_scheduler_executor)
+            {
+                if (first == last)
+                {
+                    return util::detail::algorithm_result<ExPolicy, T>::get(
+                        HPX_FORWARD(T_, init));
+                }
+            }
+
+            // Handle single-element case: can't partition into size >= 2
+            // This must be checked for all execution policies
+            auto const count = distance(first, last);
+            if (count == 1)
+            {
+                T result = HPX_INVOKE(r, HPX_FORWARD(T_, init), *first);
+                if constexpr (has_scheduler_executor)
+                {
+                    return result;
+                }
+                else
+                {
+                    return util::detail::algorithm_result<ExPolicy, T>::get(
+                        HPX_MOVE(result));
+                }
+            }
+
+            auto f1 = [r](FwdIterB part_begin, std::size_t part_size) -> T {
+                return reduce_partition<ExPolicy, FwdIterB, T>(
+                    part_begin, part_size, r);
+            };
+
+            auto reduce_policy =
+                policy.with(reduce_executor_parameters{});
+            using reduce_policy_type = std::decay_t<decltype(reduce_policy)>;
+
+            return util::partitioner<reduce_policy_type, T>::call(
+                HPX_MOVE(reduce_policy), first, distance(first, last),
+                HPX_MOVE(f1),
+                hpx::unwrapping(
+                    [init = HPX_FORWARD(T_, init), r = HPX_FORWARD(Reduce, r)](
+                        auto&& results) -> T {
+                        return sequential_reduce<ExPolicy>(
+                            hpx::util::begin(results), hpx::util::size(results),
+                            init, r);
+                    }));
+        }
+    };
+    /// \endcond
     }    // namespace detail
 }    // namespace hpx::parallel
 
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
@@ -400,7 +400,7 @@ namespace hpx::parallel {
                 if (chunk_size < 2)
                 {
                     chunk_size = (num_elements + num_cores - 1) / num_cores;
-                    chunk_size = (std::max)(chunk_size, std::size_t(2));
+                    chunk_size = (std::max) (chunk_size, std::size_t(2));
                 }
 
                 // chunk_size_iterator gives the last partition num_elements % chunk_size
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
@@ -356,6 +356,7 @@ namespace hpx {
 #else    // DOXYGEN
 
 #include <hpx/config.hpp>
+#include <hpx/assert.hpp>
 #include <hpx/modules/concepts.hpp>
 #include <hpx/modules/executors.hpp>
 #include <hpx/modules/iterator_support.hpp>
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
@@ -433,8 +433,7 @@ namespace hpx::execution::experimental {
     };
 }    // namespace hpx::execution::experimental
 
-namespace hpx::parallel {
-    namespace detail {
+namespace hpx::parallel { namespace detail {
 
     // Helper function to reduce a partition without requiring an init value.
     // Assumes partition size is always >= 2 (enforced by reduce_executor_parameters).
@@ -470,9 +469,8 @@ namespace hpx::parallel {
         static constexpr T sequential(ExPolicy&& policy, InIterB first,
             InIterE last, T_&& init, Reduce&& r)
         {
-            return sequential_reduce<ExPolicy>(
-                HPX_FORWARD(ExPolicy, policy), first, last,
-                HPX_FORWARD(T_, init), HPX_FORWARD(Reduce, r));
+            return sequential_reduce<ExPolicy>(HPX_FORWARD(ExPolicy, policy),
+                first, last, HPX_FORWARD(T_, init), HPX_FORWARD(Reduce, r));
         }
 
         template <typename ExPolicy, typename FwdIterB, typename FwdIterE,
@@ -495,7 +493,7 @@ namespace hpx::parallel {
 
             // Handle single-element case: can't partition into size >= 2
             // This must be checked for all execution policies
-            auto const count = distance(first, last);
+            auto const count = hpx::parallel::detail::distance(first, last);
             if (count == 1)
             {
                 T result = HPX_INVOKE(r, HPX_FORWARD(T_, init), *first);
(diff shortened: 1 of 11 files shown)


hkaiser (commented):
Could you please separate the formatting changes into a different PR? It's close to impossible to review your actual changes here. 

Also, for the formatting - are you sure you're using clang-format V20?

BhoomishGupta:
Yeah, I was not using V20.
So, should I close this PR and open a new one, that is only with the changes?

hkaiser:
> Yeah, I was not using V20. So, should I close this PR and open a new one, that is only with the changes?

Simply keep (force) pushing to your branch, this wil update the PR.

libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
```diff
@@ -382,6 +382,22 @@ namespace hpx::parallel {
...
+
+        // Helper function to reduce a partition without requiring an init value.
+        template <typename ExPolicy, typename FwdIterB, typename T,
+            typename Reduce>
+        T reduce_partition(
+            FwdIterB part_begin, std::size_t part_size, Reduce const& r)
+        {
+            if (part_size == 1)
+            {
+                return HPX_INVOKE(r, *part_begin, *part_begin);
```
hkaiser: This is not doing the correct thing as it would apply the reduction to the element of a single-element partition 'twice'. We have to avoid creating single-element partitions to begin with.

Currently, (almost) all algorithms use this functionality to determine the chunk size to use: https://github.com/STEllAR-GROUP/hpx/blob/f36e02d152b98de291c2e4722b9f975e85571f8d/libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size.hpp#L182-L261
 
Especially the invocation of the customization point here is responsible for computing the chunk size: https://github.com/STEllAR-GROUP/hpx/blob/f36e02d152b98de291c2e4722b9f975e85571f8d/libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size.hpp#L231-L233

I'm inclined to think that we should introduce a new customization point replacing the final chunk size adjustments here https://github.com/STEllAR-GROUP/hpx/blob/f36e02d152b98de291c2e4722b9f975e85571f8d/libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size.hpp#L236 
That would allow to install a custom implementation of said customization point for the reduce algorithm without interfering with any other, possibly user-supplied customizations.

libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
```diff
@@ -393,9 +393,9 @@ namespace hpx::parallel {
             {
                 return HPX_INVOKE(r, *part_begin, *part_begin);
             }
-            T init = HPX_INVOKE(r, *part_begin, *(part_begin + 1));
+            T init = HPX_INVOKE(r, *part_begin, *std::next(part_begin));
             return sequential_reduce<ExPolicy>(
-                part_begin + 2, part_size - 2, HPX_MOVE(init), r);
+                std::next(part_begin, 2), part_size - 2, HPX_MOVE(init), r);
```
hkaiser: If you use `std::next`, you should also use `std::prev` for the second argument.

libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
BhoomishGupta: Yeah, you are correct. Totally, mistook this as I was too focused on a single test case.

Regarding your suggestion: I agree that introducing a customization point for adjust_chunk_size_and_max_chunks is the cleaner solution. That way we can enforce the chunk size constraints specifically for reduce without affecting other algorithms. I will start working on refactoring that logic into a CPO.

libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
BhoomishGupta: I believe that part_size is an integer, so std::prev cannot be applied here as it is simple integer arithmetic. 

I used std::next as it was needed for part_begin because it is a Forward Iterator.

hkaiser: My bad, thanks for clarifying this.

libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
hkaiser: Please let me know if you need more information about this. There exists almost no documentation about this.

hkaiser:
@BhoomishGupta FWIW, the reduce tests now seem to fail. Also, please rebase thise onto master to resolve the conflicts.

BhoomishGupta:
> @BhoomishGupta FWIW, the reduce tests now seem to fail. Also, please rebase thise onto master to resolve the conflicts.


Working on it

hkaiser:
@BhoomishGupta Please rebase your changes onto master to reduce the amount of unrelated changes (currently more than 300 files). Otherwise it's close to impossible to review your PR.

hkaiser (commented):
@BhoomishGupta Thanks for investigating and attempting to implement a new customization point for the execution parameters. 

You have implemented part of the necessary logic. Please have a look at the existing customization points to see what's necessary additionally.

I have the impression, that you have added the new functionality for all algorithms, as the new functionality will be invoked always. I'd suggest making sure, that the new functionality is used only for the reduction algorithms.

BhoomishGupta:
> @BhoomishGupta Thanks for investigating and attempting to implement a new customization point for the execution parameters.
> 
> You have implemented part of the necessary logic. Please have a look at the existing customization points to see what's necessary additionally.
> 
> I have the impression, that you have added the new functionality for all algorithms, as the new functionality will be invoked always. I'd suggest making sure, that the new functionality is used only for the reduction algorithms.

@hkaiser Thanks for the review. I will check the existing customization points and add the missing things.

Regarding the scope: I did intentionally add it to the generic utility because I thought adjusting chunk sizes could be beneficial for other algorithms down the line. Or do you strongly prefer I hard-restrict it to reduce only?

Just to clarify on the current behavior: even though it's in a shared file, it currently acts as a no-op for other algorithms. The default implementation returns {0, 0}, and the call sites only apply adjustments when the return values are non-zero. So it shouldn't be impacting non-reduce algorithms right now, but I will make sure the trait-detection aligns perfectly with the rest of the codebase.

hkaiser:
> @hkaiser Thanks for the review. I will check the existing customization points and add the missing things.
> 
> Regarding the scope: I did intentionally add it to the generic utility because I thought adjusting chunk sizes could be beneficial for other algorithms down the line. Or do you strongly prefer I hard-restrict it to reduce only?

The default behavior is needed for all algorithms (as a fall-back, most likely), while the customization should apply to the reduction algorithms only.

> Just to clarify on the current behavior: even though it's in a shared file, it currently acts as a no-op for other algorithms. The default implementation returns {0, 0}, and the call sites only apply adjustments when the return values are non-zero. So it shouldn't be impacting non-reduce algorithms right now, but I will make sure the trait-detection aligns perfectly with the rest of the codebase.

If the default used by all algorithms (except reduce) is now `{0, 0}`, wouldn't that change the current behavior?

BhoomishGupta:
> > @hkaiser Thanks for the review. I will check the existing customization points and add the missing things.
> > Regarding the scope: I did intentionally add it to the generic utility because I thought adjusting chunk sizes could be beneficial for other algorithms down the line. Or do you strongly prefer I hard-restrict it to reduce only?
> 
> The default behavior is needed for all algorithms (as a fall-back, most likely), while the customization should apply to the reduction algorithms only.
> 
> > Just to clarify on the current behavior: even though it's in a shared file, it currently acts as a no-op for other algorithms. The default implementation returns {0, 0}, and the call sites only apply adjustments when the return values are non-zero. So it shouldn't be impacting non-reduce algorithms right now, but I will make sure the trait-detection aligns perfectly with the rest of the codebase.
> 
> If the default used by all algorithms (except reduce) is now `{0, 0}`, wouldn't that change the current behavior?

I don,t think that the {0, 0} default changes the current behavior for any algorithm.

The call site in `chunk_size.hpp`
```cpp
auto [adj_chunk, adj_max] =
    hpx::execution::experimental::adjust_chunk_size_and_max_chunks(
        policy.parameters(), policy.executor(),
        max_chunks, chunk_size);
if (adj_chunk != 0)
    chunk_size = adj_chunk;
if (adj_max != 0)
    max_chunks = adj_max;
else if (adj_chunk == 0)
    adjust_chunk_size_and_max_chunks(
        cores, count, max_chunks, chunk_size);
```
works when the Customization Point returns the default {0, 0}, both adj_chunk and adj_max are zero, so we go through else if (adj_chunk == 0) which calls the existing adjust_chunk_size_and_max_chunks free function, which is exactly the same code path as before.
Only an execution parameters object that provides a custom adjust_chunk_size_and_max_chunks member will return non-zero values and override the defaults.

hkaiser:
> I don,t think that the {0, 0} default changes the current behavior for any algorithm.
> 
> The call site in `chunk_size.hpp`
> 
> ```c++
> auto [adj_chunk, adj_max] =
>     hpx::execution::experimental::adjust_chunk_size_and_max_chunks(
>         policy.parameters(), policy.executor(),
>         max_chunks, chunk_size);
> if (adj_chunk != 0)
>     chunk_size = adj_chunk;
> if (adj_max != 0)
>     max_chunks = adj_max;
> else if (adj_chunk == 0)
>     adjust_chunk_size_and_max_chunks(
>         cores, count, max_chunks, chunk_size);
> ```
> 
> works when the Customization Point returns the default {0, 0}, both adj_chunk and adj_max are zero, so we go through else if (adj_chunk == 0) which calls the existing adjust_chunk_size_and_max_chunks free function, which is exactly the same code path as before. Only an execution parameters object that provides a custom adjust_chunk_size_and_max_chunks member will return non-zero values and override the defaults.

I have seen that. This code does something different from what was in place before. The default just should be the previous `adjust_chunk_size_and_max_chunks` (most likely to be invoked from the CPO fallback).

As said, please have a look at the other parameters customizations to see what's missing, e.g., `get_chunk_size` here: https://github.com/STEllAR-GROUP/hpx/blob/b514f4ab6274d21a716e3acaac08f825b9a93776/libs/core/execution/include/hpx/execution/executors/execution_parameters_fwd.hpp#L79-L127.

libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size.hpp
```diff
@@ -150,8 +150,12 @@ namespace hpx::parallel::util::detail {
...
                 policy.executor(), hpx::chrono::null_duration, cores, count);
 
-        // make sure, chunk size and max_chunks are consistent
-        adjust_chunk_size_and_max_chunks(cores, count, max_chunks, chunk_size);
+        auto [adj_chunk, adj_max] =
+            hpx::execution::experimental::adjust_chunk_size_and_max_chunks(
+                policy.parameters(), policy.executor(), count, cores,
+                max_chunks, chunk_size);
+        chunk_size = adj_chunk;
+        max_chunks = adj_max;
```
hkaiser: What about:
```suggestion
        std::tie(chunk_size, max_chunks) =
            hpx::execution::experimental::adjust_chunk_size_and_max_chunks(
                policy.parameters(), policy.executor(), count, cores,
                max_chunks, chunk_size);
```

hkaiser (commented):
There is one thing missing still. Currently you use `policy.with()` to replace the current set of parameter customizations, which will cause for any user-supplied customizations to be lost. I think we need to use something like the new API `rebind_executor_parameters` (like for instance proposed here: https://github.com/STEllAR-GROUP/hpx/pull/6935) instead.

Otherwise: very nice! Thanks for working on this!

libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size.hpp
BhoomishGupta: Yeah, this seems better.

libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size.hpp
hkaiser, line 311: I think this needs to be replaced with the CPO now as well.

hkaiser, line 580: And this as well.

hkaiser:
> There is one thing missing still. Currently you use `policy.with()` to replace the current set of parameter customizations, which will cause for any user-supplied customizations to be lost. I think we need to use something like the new API `rebind_executor_parameters` (like for instance proposed here: #6935) instead.
> 
> Otherwise: very nice! Thanks for working on this!

Let's wait for #6935 to be merged, then you can adapt your PR.

hkaiser:
@BhoomishGupta FWIW, #6935 was merged yesterday. Could you please rebase this PR onto master to resolve the conflicts? After that let's work on finalizing your patch.

BhoomishGupta:
> @BhoomishGupta FWIW, #6935 was merged yesterday. Could you please rebase this PR onto master to resolve the conflicts? After that let's work on finalizing your patch.

Hi @hkaiser,

Thanks for the heads-up on #6935 getting merged!

Just a quick note: my university exams are currently going on, so my availability is a bit limited right now. I will rebase this PR onto master to resolve the conflicts and work on finalizing the patch right after my exams finish on March 14th.

Thanks for your patience!

libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
```diff
@@ -356,6 +356,8 @@ namespace hpx {
 #else    // DOXYGEN
 
 #include <hpx/config.hpp>
+#include <hpx/assert.hpp>
+#include <hpx/execution/executors/rebind_executor_parameters.hpp>
```
hkaiser: Please never #include anything but the generated module headers (except files from a module itself):
```suggestion
#include <hpx/modules/execution.hpp>
```
Otherwise you will break the C++ module infrastructure.

BhoomishGupta: Okay. Will do this from now on!

hkaiser:
@BhoomishGupta On MacOS the new tests are failing. Inspect reports an issue as well. Please pay attention to the CIs yourself.

BhoomishGupta:
> @BhoomishGupta On MacOS the new tests are failing. Inspect reports an issue as well. Please pay attention to the CIs yourself. 

Ah, my bad, I had my wires crossed and was looking at the CI for a different PR. I will check the MacOS and Inspect logs for this one tomorrow and will push a fix soon. I'll pay closer attention to the CI status here going forward.

hkaiser:
@BhoomishGupta ping?

BhoomishGupta:
> @BhoomishGupta ping?

Hi @hkaiser, sorry for the delay, I had academic tests. I have pushed updates and resolved the conflicts; please take another look.

I will also take a look into the snippets copilot reviewed.

hkaiser:
@BhoomishGupta please also have a look at the failing test: rebind_executor_parameters.

libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
```diff
@@ -382,63 +385,194 @@ namespace hpx::parallel {
...
 
-            template <typename ExPolicy, typename InIterB, typename InIterE,
-                typename T_, typename Reduce>
-            static constexpr T sequential(ExPolicy&& policy, InIterB first,
-                InIterE last, T_&& init, Reduce&& r)
+        public:
+            template <typename Executor>
+            HPX_FORCEINLINE constexpr std::pair<std::size_t, std::size_t>
+            adjust_chunk_size_and_max_chunks(Executor&&,
+                std::size_t num_elements, std::size_t num_cores,
```
hkaiser: Why do you need this forwarding function? Couldn't you simply implement everything here without having to introduce yet another function call?

libs/core/execution/include/hpx/execution/executors/execution_parameters.hpp
```diff
@@ -652,6 +652,116 @@ namespace hpx::execution::experimental::detail {
...
+        else
+        {
+            // max_chunks != 0 && chunk_size != 0
+            std::size_t const calculated_max_chunks =
+                (num_elements + chunk_size - 1) / chunk_size;
+
+            if (calculated_max_chunks > max_chunks)
+            {
+                chunk_size = (num_elements + max_chunks - 1) / max_chunks;
+            }
```
hkaiser: Please note that we just added a fix to this code that should be reflected in your work as well: https://github.com/TheHPXProject/hpx/pull/7295

libs/core/execution/include/hpx/execution/executors/execution_parameters.hpp
```diff
@@ -599,6 +599,116 @@ namespace hpx::execution::experimental::detail {
...
+                std::size_t const cores_times_4 = 4 * num_cores;    // -V112
+
+                chunk_size = (num_elements + cores_times_4 - 1) / cores_times_4;
+
+                max_chunks =
+                    (std::min) (cores_times_4, num_elements);    // -V112
+
+                chunk_size = (std::max) (chunk_size,
+                    (num_elements + max_chunks - 1) / max_chunks);
+            }
```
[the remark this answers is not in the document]
hkaiser: Could you please verify whether Copilot is correct here?

libs/core/execution/include/hpx/execution/executors/execution_parameters.hpp
```diff
@@ -652,6 +652,116 @@ namespace hpx::execution::experimental::detail {
...
+    HPX_HAS_MEMBER_XXX_TRAIT_DEF(adjust_chunk_size_and_max_chunks)
+
+    ///////////////////////////////////////////////////////////////////////
+    // shared default implementation allowing to handle
+    // adjust_chunk_size_and_max_chunks
+    HPX_FORCEINLINE constexpr std::pair<std::size_t, std::size_t>
+    adjust_chunk_size_and_max_chunks_default(std::size_t num_elements,
+        std::size_t num_cores, std::size_t max_chunks,
+        std::size_t chunk_size) noexcept
+    {
```
hkaiser: Shouldn't this function either simply call the pre-existing `adjust_chunk_size_and_max_chunks` function or alternatively remove that from `‎libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size.hpp‎`?

hkaiser (commented):
Thanks for returning to this PR and working on finalizing it!

libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
BhoomishGupta: Done. I have now removed the forwarding member function and the private _impl helper.
And implemented them int tag_override_invoke

libs/core/execution/include/hpx/execution/executors/execution_parameters.hpp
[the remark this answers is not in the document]
BhoomishGupta: Yup, copilot was correct here.
So, now I have added an early return (`if (num_elements == 0) return {0, 0};`)

libs/core/execution/include/hpx/execution/executors/execution_parameters.hpp
BhoomishGupta: Done. I have removed it from `libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size.hpp‎`

hkaiser:
@BhoomishGupta could you please have a look at the Copilot comments? Some of those seem to be valid. Also compilation of this branch now fails on some CIs, please watch those your self closely.

BhoomishGupta:
@hkaiser This build (`Linux CI (Likwid, Debug) / build (pull_request)`) is failing on my PR. I checked the logs and it looks like an external compiler issue in `likwid/src/configuration.c`, not related to my changes.

Could you take a quick look?

hkaiser:
> @hkaiser This build (`Linux CI (Likwid, Debug) / build (pull_request)`) is failing on my PR. I checked the logs and it looks like an external compiler issue in `likwid/src/configuration.c`, not related to my changes.
> 
> Could you take a quick look?

Yes, this is an unrelated issue. We're working on fixing it.

libs/core/algorithms/tests/regressions/reduce_6647.cpp
```diff
@@ -0,0 +1,79 @@
...
+struct minmax
+{
+    std::pair<int, int> operator()(
+        std::pair<int, int> lhs, std::pair<int, int> rhs) const
+    {
+        return {
+            lhs.first < rhs.first ? lhs.first : rhs.first,
+            lhs.second < rhs.second ? rhs.second : lhs.second,
+        };
+    }
```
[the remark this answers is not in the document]
hkaiser: @BhoomishGupta Could you please have a look at the Copilot comments? This one in particular, but the others as well.

libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
```diff
@@ -382,63 +385,188 @@ namespace hpx::parallel {
...
+                Executor&& exec, std::size_t num_elements,
+                std::size_t num_cores, std::size_t max_chunks,
+                std::size_t chunk_size)
             {
-            }
+                auto [adjusted_chunk_size, adjusted_max_chunks] = hpx::
+                    execution::experimental::adjust_chunk_size_and_max_chunks(
+                        HPX_FORWARD(InnerParams, inner),
+                        HPX_FORWARD(Executor, exec), num_elements, num_cores,
+                        max_chunks, chunk_size);
```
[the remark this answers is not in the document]
hkaiser: @BhoomishGupta Does this comment make sense?

libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
```diff
@@ -382,63 +385,188 @@ namespace hpx::parallel {
...
+                {
+                    return {new_chunk_size, new_max_chunks};
+                }
+
+                // Ensure minimum chunk size of 2 for reduce_partition.
+                if (new_chunk_size < 2)
+                {
+                    new_chunk_size = (num_elements + num_cores - 1) / num_cores;
+                    new_chunk_size = (std::max) (new_chunk_size, std::size_t(2));
+                }
```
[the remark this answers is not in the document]
hkaiser: What about this comment?

libs/core/algorithms/include/hpx/parallel/algorithms/reduce.hpp
```diff
@@ -382,63 +385,189 @@ namespace hpx::parallel {
...
+                    return {new_chunk_size, new_max_chunks};
+                }
+
+                // Ensure minimum chunk size of 2 for reduce_partition.
+                if (new_chunk_size < 2)
+                {
+                    new_chunk_size = (num_elements + num_cores - 1) / num_cores;
+                    new_chunk_size =
+                        (std::max) (new_chunk_size, std::size_t(2));
+                }
```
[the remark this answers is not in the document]
hkaiser: And this one? Could you please either handle the Copilot comments or mark them as resolved.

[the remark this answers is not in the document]
BhoomishGupta: Ok, I Will review all the copilot comments and mark them resolved or do the necessary changes.

hkaiser:
@BhoomishGupta We have now merged #7319, which removes the use of `tag_invoke` in favor of member functions. Could you please update your new parameters customization object accordingly.

BhoomishGupta:
> @BhoomishGupta We have now merged #7319, which removes the use of `tag_invoke` in favor of member functions. Could you please update your new parameters customization object accordingly.

Ok, Will look into it

PR #6879, opened 2026-02-04
Implement parallel hpx::uninitialized_relocate_* algorithms for overlapping ranges

### LINKED ISSUE #6878: Parallel uninitialized_relocate does not preserve sequenced semantics for overlapping ranges
The parallel overloads of `hpx::experimental::uninitialized_relocate` and uninitialized_relocate_backward do not preserve the same ordering semantics as the sequenced implementation when the source and destination ranges overlap.

While overlapping ranges are only well-defined when the caller selects the correct direction (analogous to `std::copy` vs `std::copy_backward),` the parallel implementations may reorder relocations due to chunk-based parallelization. This can lead to observable differences in object lifetime and final memory layout compared to `hpx::execution::seq.`

**Expected behavior:**
For a given algorithm (`uninitialized_relocate` or `uninitialized_relocate_backward`), parallel execution should preserve the same behavior as the sequenced implementation, even for overlapping contiguous ranges.

**Actual behavior:**
When using `hpx::execution::par`, overlapping contiguous ranges may be relocated in a different order than the sequenced algorithm, resulting in differences in object state and lifetime.

ArivoliR:
Fixes #6878 

## Proposed Changes

- Implemented parallel hpx::uninitialized_relocate_* algorithms for overlapping ranges.
- Chain-based algorithm: for a left-shift by `d` positions, the range is split into `d` independent chains where chain `c` processes indices `c, c+d, c+2d, ...`. 
- Each chain can be executed in parallel since elements within a chain depend only on each other. 

[P1144 Object relocation in terms of move plus destroy](http://wg21.link/p1144)
[GSoC Project Wiki](https://github.com/STEllAR-GROUP/hpx/wiki/Google-Summer-of-Code-%28GSoC%29-2026#implement-parallel-hpxuninitialized_relocate_-algorithms-for-overlapping-ranges)

## Checklist

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
@@ -1099,7 +1099,8 @@ namespace hpx::experimental {
                     if (first < dest && dest < last)
                     {
                         using value_type = std::iter_value_t<InIter>;
-                        std::memmove(std::to_address(dest),
+                        // NOLINTNEXTLINE(bugprone-undefined-memory-manipulation)
+                        std::memmove(static_cast<void*>(std::to_address(dest)),
                             std::to_address(first), count * sizeof(value_type));
                         return parallel::util::detail::algorithm_result<
                             ExPolicy, FwdIter>::get(std::next(dest, count));
@@ -1302,7 +1303,8 @@ namespace hpx::experimental {
                     if (first < dest && dest < last)
                     {
                         using value_type = std::iter_value_t<InIter1>;
-                        std::memmove(std::to_address(dest),
+                        // NOLINTNEXTLINE(bugprone-undefined-memory-manipulation)
+                        std::memmove(static_cast<void*>(std::to_address(dest)),
                             std::to_address(first), count * sizeof(value_type));
                         return parallel::util::detail::algorithm_result<
                             ExPolicy, FwdIter>::get(std::next(dest, count));


hkaiser:
@isidorostsa I believe this is another PR for you to look at. Would you have the time?

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
```diff
@@ -893,16 +898,22 @@ namespace hpx::experimental {
             }
 
             // If running in non-sequenced execution policy, we must check
-            // that the ranges are not overlapping in the left
+            // that the ranges do not overlap at all for safe parallelization
             if constexpr (!hpx::is_sequenced_execution_policy_v<ExPolicy>)
             {
                 // if we can check for overlapping ranges
-                if constexpr (hpx::traits::is_contiguous_iterator_v<InIter1> &&
+                if constexpr (hpx::traits::is_contiguous_iterator_v<InIter2> &&
```
isidorostsa: Good catch, though we definitely also want to check that `InIter1` is also a contiguous iterator. Also we could define an hpx alias for `hpx::experimental::is_pointer_in_range` as proposed in [P3952R0](https://wg21.link/p3952r0) (or one of the alternative proposals).

libs/core/algorithms/tests/unit/algorithms/uninitialized_relocate.cpp
```diff
@@ -26,6 +26,94 @@ using hpx::experimental::uninitialized_relocate;
...
+    ~relocate_tracker()
+    {
+        if (alive)
+            *alive = false;
+    }
+};
+
+template <typename ExPolicy>
+void test_uninitialized_relocate_overlap_forward(ExPolicy&& policy)
+{
```
isidorostsa: better split this into two tests

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
```diff
@@ -911,7 +922,7 @@ namespace hpx::experimental {
                             return ex::unique_any_sender<
                                 FwdIter>(parallel::util::get_second_element(
                                 hpx::parallel::detail::uninitialized_relocate_n<
-                                    parallel::util::in_out_result<InIter1,
+                                    parallel::util::in_out_result<InIter2,
```
isidorostsa: what is the purpose of this change?

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
ArivoliR: I’ll update the check to require contiguity for both source and destination iterators.
Would you prefer that I open a follow-up issue to track introducing a reusable helper for overlap detection so that it doesn't get lost?

libs/core/algorithms/tests/unit/algorithms/uninitialized_relocate.cpp
ArivoliR: Alright will do!

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
ArivoliR: The intent was to make the template parameter of in_out_result reflect the iterator type returned by the algorithm. In this case that should be InIter1, using InIter2 was a mistake on my end. I’ll fix this.

hkaiser:
@ArivoliR Could you please rebase your branch onto master (and resolve the conflicts)?

libs/core/algorithms/tests/unit/algorithms/uninitialized_relocate.cpp
```diff
@@ -519,6 +607,24 @@ int hpx_main()
...
+    //    test_uninitialized_relocate_overlap_forward(hpx::execution::seq);
+    //    test_uninitialized_relocate_overlap_forward(hpx::execution::par);
+
+    // #if defined(HPX_HAVE_CXX17_STD_EXECUTION_POLICIES)
+    //    test_uninitialized_relocate_overlap_forward(hpx::execution::par_unseq);
+    // #endif
+
+    //    test_uninitialized_relocate_overlap_forward(
+    //        hpx::execution::par(hpx::execution::task))
+    //        .get();
```
hkaiser: Do you plan to re-enable these tests?

ArivoliR: No sir, those were only added temporarily for debugging. I’ll be removing them in the next commit. Thank you for pointing it out

hkaiser: Wouldn't it be useful to have those tests functioning?

ArivoliR: Hi @hkaiser, 
I’ve split the original overlap test into two clearer pieces as guided by Isidoros:
1. A sequenced-only semantics test that validates the expected behavior for overlapping ranges under execution::seq.

2. A policy comparison test that checks that execution::par (and par_unseq) produce the same observable results as the sequenced implementation.

For now I’ve left par(task) out and added a TODO. I haven’t added a test for the async behavior yet, since it requires explicitly handling the task/sender return path. I’d prefer to add a dedicated test for that once I’ve spent more time understanding the expected testing pattern.

hkaiser: You can ignore the sender&receiver part for now. I'd focus on the async/future implementation which should be easier to do. However I'd be fine for this ending up in a separate PR.

ArivoliR: I don’t have a working test for the async/future-based path yet, and I’m still figuring out the right way to implement it. I’d prefer to follow up with a separate PR once I figure it out in sometime. If you’re okay with it, I’d be happy to close this PR here once the current commits are approved

hkaiser:
There are still some minor things left:
- inspect complains about a missing #include 
- the spell checker complains about the use of `iff`, please add that to the while list here: https://github.com/STEllAR-GROUP/hpx/blob/master/tools/.codespell_whitelist

The other reported problems are unrelated.

ArivoliR:
> There are still some minor things left:
>    inspect complains about a missing #include
>    the spell checker complains about the use of iff, please add that to the while list here: https://github.com/STEllAR-GROUP/hpx/blob/master/tools/.codespell_whitelist
> The other reported problems are unrelated.

@hkaiser I believe this PR is ready for merge. I’m rebasing it now, please let me know if youd' like any further changes.

isidorostsa:
In this PR, ranges that overlap in any direction will be handled by the sequential version of the code, instead of just the ones that overlap in the direction that would work with forward iteration. This means it is still a bug to call this with a range overlapping in the wrong direction, but at least its a bug in the sequential code.

In the case of relocation this will either end up destroying objects and moving from them afterwards (bug) or if the type is trivially relocatable it will keep memcpying the same object (not as bad, but still a bug). In the parallel code case it would do this but also with data races. 

All in all, I think we should keep the logic of the check as it was before, since this change is not protecting us from any bugs, but looks like it is. But we should rewrite it in the style you wrote it, as it is clearer than what we had before.

For bug identification we should HPX_ASSERT that the ranges are not overlapping on either direction if we decide to go for the parallel version. When we implement the safe parallel version we can change the selection logic to fit this.

Sahilll10:
Hi, I've been going through this PR as part of researching the GSoC project on  "Implement parallel hpx::uninitialized_relocate_ algorithms for overlapping ranges." Really helpful context here.

@isidorostsa - If I'm understanding correctly, the current check falls back to sequential for all overlaps including d_first > first cases where forward parallel chunking would actually be safe. 

The fix should preserve full parallelism when the shift direction is safe, only serialize when there is a genuine hazard, and HPX_ASSERT for the truly undefined cases in the parallel path. 

Would appreciate any correction if I’ve misunderstood something.

isidorostsa:
@ArivoliR, thanks for working more on this. What is the state of this PR? What does it achieve compared to master?

hkaiser:
@ArivoliR any comments to @isidorostsa's question?

ArivoliR:
> @ArivoliR, thanks for working more on this. What is the state of this PR? What does it achieve compared to master?

The PR essentially adds HPX_ASSERT(!overlap) before any parallel algorithm is executed in all three functions. This catches wrong-direction overlaps in debug builds before they reach parallel execution. There is no other major change. Pushing fix for `clang-format` now.

ArivoliR:
I'd like to push my commits for the parallel implementation of uninitialized relocate here. Converting to draft for the same. The asserts will not be needed when the parallel version works :D

ArivoliR:
The move/destroy counts are non-deterministic under parallel execution for types with throwing move constructors. I've added an if constexpr (sequenced_policy) guard around the exact count assertions as a temporary workaround.

I overlooked the fact that its very much a bug while adding the guard. Will fix this tomorrow.

(Fix complete)

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
```diff
@@ -479,6 +481,309 @@ namespace hpx::parallel {
...
+                                for (std::size_t todo_ci = ci + 1;
+                                    todo_ci < num_chains; ++todo_ci)
+                                {
+                                    std::size_t todo_c = chain_start + todo_ci;
+                                    for (std::size_t j = todo_c; j < n; j += d)
+                                    {
+                                        std::destroy_at(first_ptr + j);
+                                    }
+                                }
+                                throw;
```
isidorostsa: A comment or two might help me figure out what's going on here 😅

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
```diff
@@ -479,6 +481,309 @@ namespace hpx::parallel {
...
+                    },
+                    [first, dest, n](auto&&... data) mutable
+                        -> util::in_out_result<InIter, FwdIter> {
+                        static_assert(sizeof...(data) < 2);
+                        util::detail::clear_container(data...);
+                        return util::in_out_result<InIter, FwdIter>{
+                            std::next(first, n), std::next(dest, n)};
+                    },
+                    [dest_ptr, n, d](chain_group_result&& r) -> void {
+                        for (std::size_t ci = 0; ci < r.chains_done; ++ci)
```
isidorostsa: I think it would help splitting this and some of the code above in named lambda variables

ArivoliR:
> I would appreciate a short explanation of the code, thanks!

Sorry about the lack of comments, I'll get to adding them in a bit. 

The algo decomposes the overlapping range into d independent chains (d -> overlap distance).
`chain c` consists of elements at positions c, c+d, ..., c+kd, .... The elements in each chain are moved in order to avoid overwriting a source before it has been relocated. Chains themselves are processed in parallel since they are independent from each other.

I'll split the lambdas into named variables and push shortly.

Each thread runs the processor lambda which iterates sequentially over the chains it was assigned by the partitioner. 
The finalize lambda builds the return iterators on success, The cleanup lambda destroys already relocated elements (of other threads) at dest if any thread throws mid-chain.

ArivoliR:
> I'll split the lambdas into named variables and push shortly.

@isidorostsa I'm done with the same. Hope code readability is better now, thank you for your time

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
```diff
@@ -479,6 +481,331 @@ namespace hpx::parallel {
...
+
+                    try
+                    {
+                        for (std::size_t j = c; j < n; j += d, ++partial)
+                        {
+                            hpx::experimental::detail::relocate_at_helper(
+                                first_ptr + j, dest_ptr + j);
+                        }
+                    }
+                    catch (...)
```
isidorostsa, line 564: Will  this clean the work of every thread up, or only the one that failed?

ArivoliR: This block of code only cleans up the thread which failed. The cleanup lambda clears up the rest

hkaiser:
@ArivoliR Could you please resolve the conflicts?

ArivoliR:
> @ArivoliR As a last request: please add exhaustive tests that verify all code paths you added, including the exceptional case.

On it!

ArivoliR:
> LGTM, thanks!
> 
> Please resolve the merge conflicts, however.

There are a few errors sir, still working on them. Will push the fixes soon when I'm done and resolve the conflicts as well

ArivoliR:
To be specific, there was a clang-tidy error since there was a pointer to a non-trivially-copyable type. 
And a few clang-format errors. Should be fixed now

libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp
```diff
@@ -761,45 +1086,99 @@ namespace hpx::experimental {
...
+                constexpr bool is_trivially_relocatable =
+                    hpx::experimental::util::detail::relocation_traits<InIter,
+                        FwdIter>::is_memcpyable;
+                if constexpr (is_trivially_relocatable)
+                {
+                    auto last = std::next(first, count);
+                    if (first < dest && dest < last)
+                    {
+                        using value_type = std::iter_value_t<InIter>;
+                        std::memmove(static_cast<void*>(std::to_address(dest)),
```
hkaiser, line 1103: I think you can remove the explicit casts.

ArivoliR: Done sir, I had added them because of a clang-tidy error. They seem to pass now though

ArivoliR: Would you prefer I add the explicit casts back sir?
The tests fail even after `// NOLINTNEXTLINE(bugprone-undefined-memory-manipulation)`
The error for reference:
```
/__w/hpx/hpx/libs/core/algorithms/include/hpx/parallel/algorithms/uninitialized_relocate.hpp:1103:38: 
error: first argument in call to 'memmove' is a pointer to non-trivially copyable type 'counted_struct<trivially_relocatable, true>' [clang-diagnostic-nontrivial-memcall]
```
a compiler error [clang-diagnostic-nontrivial-memcall] to be specific

ArivoliR: I've added the explicit casts back for now sir. Will proceed in the intended direction after further review from you.

hkaiser (approved):
LGTM, thanks!

PR #6883, opened 2026-02-05
Fix initialization hang when --hpx:threads exceeds --hpx:cores

### LINKED ISSUE #6345: Initialization hangs when only setting --hpx:cores
## Expected Behavior
If `--hpx:threads` hasn't been set and `--hpx:cores` has been set to some (smaller than default) value, silently set the number of OS threads to be equal to the number of specified cores, or throw an error if the user doesn't also explicitly specify a sufficiently small number for `--hpx:threads` (such that OS threads <= cores).

If both have been set by the user such that `--hpx:threads` > `--hpx:cores`, either throw an error or silently reduce the number of OS threads so that OS threads == cores.

(I assume that it's a bad idea to spawn more OS threads than cores we are allowed to run on.)

## Actual Behavior
Initialization hangs when only setting `--hpx:cores` to a value smaller that the default number of OS threads, or when setting both `--hpx:threads` and `--hpx:cores` such that `--hpx:threads` > `--hpx:cores`. 

@hkaiser How do you think we should handle this?
I stumbled on this because the [options_as_config ](https://github.com/STEllAR-GROUP/hpx/blob/master/libs/full/command_line_handling/tests/regressions/options_as_config_3339.cpp) regression test failed locally. I don't think I have any peculiar setup, so shouldn't it also fail on the CIs in the same manner?

0lem1of1:
Fixes #6345

## Proposed Changes

- Modified libs/core/affinity/src/parse_affinity_options.cpp to add wrap-around logic to decode_balanced_distribution.
- When using balanced distribution with oversubscription (more threads than available PUs), the loop now resets the PU cursor (next_pu_index) when progress stops, allowing it to reuse PUs instead of hanging in an infinite loop.
- Added a made_progress flag to detect when a full sweep of cores yields no new PU assignments.

Previously, when --hpx:threads exceeded the number of available PUs , given by the set --hpx:cores (oversubscription) and the balanced distribution was used, the affinity binder would hang indefinitely during initialization. This occurred because the affinity loop expected to find a unique, unoccupied PU for every thread but had no mechanism to reuse PUs once all were exhausted. This fix implements the missing wrap-around logic, similar to how compact (implicitly) handles oversubscription.

## Checklist

### DIFF
--- a/libs/core/affinity/src/parse_affinity_options.cpp
+++ b/libs/core/affinity/src/parse_affinity_options.cpp
@@ -837,8 +837,12 @@ namespace hpx::threads::detail {
         std::vector<std::size_t> next_pu_index(num_cores, 0);
         num_pus.resize(num_threads);
 
-        for (std::size_t num_thread = 0; num_thread < num_threads; /**/)
+        bool made_progress = true;
+        for (std::size_t num_thread = 0;
+            num_thread < num_threads && made_progress;
+            /**/)
         {
+            made_progress = false;
             for (std::size_t num_core = 0; num_core < num_cores; ++num_core)
             {
                 if (any(affinities[num_thread]))
@@ -880,8 +884,14 @@ namespace hpx::threads::detail {
                 affinities[num_thread] = t.init_thread_affinity_mask(
                     num_core + used_cores, next_pu_index[num_core] - 1);
 
+                made_progress = true;
                 if (++num_thread == num_threads)
                     return;
+            }
+            if (num_thread < num_threads && !made_progress)
+            {
+                std::fill(next_pu_index.begin(), next_pu_index.end(), 0);
+                made_progress = true;
             }
         }
     }
@@ -894,7 +904,7 @@ namespace hpx::threads::detail {
     {
         std::size_t const num_threads = affinities.size();
 
-        // check_num_threads(use_process_mask, t, num_threads, ec);
+        check_num_threads(use_process_mask, t, num_threads, ec);
 
         if (use_process_mask)
         {


hkaiser (commented):
Were you able to reproduce the issue? The change you suggest is a workaround and doesn't fix the issue at hand... I'd like to understand what the underlying problem is before accepting your PR.

0lem1of1:
Yes, I was able to reproduce the issue consistently on a Linux machine with a hybrid CPU architecture (6 hyperthreaded P-cores and 8 single-threaded E-cores). The hang occurs specifically when the requested thread count exceeds the actual hardware contexts available in the selected core mask—for instance, requesting 7 cores and 14 threads causes a hang because the topology only offers 13 logical slots (12 from the P-cores + 1 from the E-core), leaving the 14th thread waiting indefinitely for a resource that doesn't exist. I understand that my current change acts more as a guardrail than a root-cause fix; I intended it to prevent users from accidentally entering this invalid state(although i have introduced a `hpx.allow_oversubscription` flag to bypass this). Could you advise on the preferred direction? specifically, should I implement a hard failure/error during initialization when the topology cannot support the request, or would you prefer I investigate the affinity binder itself to see why it hangs instead of failing ?

hkaiser:
> Yes, I was able to reproduce the issue consistently on a Linux machine with a hybrid CPU architecture (6 hyperthreaded P-cores and 8 single-threaded E-cores). The hang occurs specifically when the requested thread count exceeds the actual hardware contexts available in the selected core mask—for instance, requesting 7 cores and 14 threads causes a hang because the topology only offers 13 logical slots (12 from the P-cores + 1 from the E-core), leaving the 14th thread waiting indefinitely for a resource that doesn't exist. I understand that my current change acts more as a guardrail than a root-cause fix; I intended it to prevent users from accidentally entering this invalid state(although i have introduced a `hpx.allow_oversubscription` flag to bypass this). Could you advise on the preferred direction? specifically, should I implement a hard failure/error during initialization when the topology cannot support the request, or would you prefer I investigate the affinity binder itself to see why it hangs instead of failing ?

Thank you for this analysis. While I understand that oversubscription is generally a bad idea, I don't think we should prevent it completely. Using your suggested registry entry (`hpx.allow_oversubscription`) is definitely a possible option. It wouldn't completely prevent the hang from happening, however.

So, what would be a correct fix for this situation? In your example the user could still use 13 threads without problems, why prevent using more than 7 threads is this case? Also, I don't yet fully understand why the 14th thread is not able to run. Could you elaborate, please? If the thread mask has only 13 slots (in your example), shouldn't the index wrap around? That sounds like an actual problem in the code that manages the thread placement.

0lem1of1:
> Also, I don't yet fully understand why the 14th thread is not able to run. Could you elaborate, please?

so as i mentioned the 6 cores i have can be hyperthreaded , so each core can run 2 threads individually so we can run 12(6*2) threads no problem , but the 7th core is a normal core so it can run only one thread and hence we cannot run the 14th thread , which would be possible if the 7th core could also be hyperthreaded .



> In your example the user could still use 13 threads without problems, why prevent using more than 7 threads is this case?

yes , here i went for a very conservative approach which addresses a solution for all general cores , like we cant assume any core to have the ability to be hyperthreaded right so 1 core = 1 thread assumption works for all of them



> what would be a correct fix for this situation?

so one solution i thought of is also a guardrail but a way smarter one , we can take the number of cores and go through the topology finding available threads for each core , and calculating the max number of threads we can use and clipping the no. of threads to that value instead of clipping it to number of cores(in my case that would be 13). 

Or we can try to change the working of the thread pool scheduler which assigns each thread to look for a core and bind to it , and the loop never stops till each thread is assigned(and hence the hang).

> If the thread mask has only 13 slots (in your example), shouldn't the index wrap around?

by this ig you are reffering to line 371 in `libs/core/affinity/src/affinity_data.cpp` which is :
`return (num_pu + offset) % hardware_concurrency;`
this wraps around `hardware_concurrency`(20 in my case) ,which is the value of total no. of available pus in my hardware and disregards mentioning of total no. of available cores(which creates the mask of 13 threads) , we can edit this line to use modulus no. of available pus intead , which could be another solution.
 
what would you recommend?

hkaiser:
> what would you recommend?

Well, I'd suggest to try fixing the underlying issue instead of finding more or less sophisticated workarounds. If the user wants to oversubscribe the system - who are we to prevent him from shooting himself into the foot?

Pansysk75:
Since I opened the referenced issue, let me rephrase my original confusion, which, looking back, mostly came from how the default values are handled for those configuration variables.

https://github.com/STEllAR-GROUP/hpx/blob/3efee239010701ebcfb523b839c6647a36f3e5e9/docs/sphinx/manual/launching_and_configuring_hpx_applications.rst?plain=1#L1540-L1546

According to the documentation, the default value for `--hpx:threads` is `cores`.
There are two ways to interpret that statement:
1. The default value for `--hpx:threads` is the number of physical cores (system constant)
2. The default value for `--hpx:threads` is the number of cores HPX  is configured to use

I was originally expecting (2) to be true (because it avoids oversubscription), but currently it seems like (1) is true.

hkaiser:
> Since I opened the referenced issue, let me rephrase my original confusion, which, looking back, mostly came from how the default values are handled for those configuration variables.
> 
> https://github.com/STEllAR-GROUP/hpx/blob/3efee239010701ebcfb523b839c6647a36f3e5e9/docs/sphinx/manual/launching_and_configuring_hpx_applications.rst?plain=1#L1540-L1546
> 
> According to the documentation, the default value for `--hpx:threads` is `cores`. There are two ways to interpret that statement:
> 
> 1. The default value for `--hpx:threads` is the number of physical cores (system constant)
> 2. The default value for `--hpx:threads` is the number of cores HPX  is configured to use
> 
> I was originally expecting (2) to be true (because it avoids oversubscription), but currently it seems like (1) is true.

Option (1) is what's implemented. Would you have any suggestions on how to rephrase the docs to avoid this confusion?

hkaiser:
> > If the thread mask has only 13 slots (in your example), shouldn't the index wrap around?
> 
> by this ig you are reffering to line 371 in `libs/core/affinity/src/affinity_data.cpp` which is : `return (num_pu + offset) % hardware_concurrency;` this wraps around `hardware_concurrency`(20 in my case) ,which is the value of total no. of available pus in my hardware and disregards mentioning of total no. of available cores(which creates the mask of 13 threads) , we can edit this line to use modulus no. of available pus intead , which could be another solution.

Why is `hardware_concurrency` equal to `20` on your system if you have only `13` physical processing units available?

0lem1of1:
> > > If the thread mask has only 13 slots (in your example), shouldn't the index wrap around?
> > 
> > 
> > by this ig you are reffering to line 371 in `libs/core/affinity/src/affinity_data.cpp` which is : `return (num_pu + offset) % hardware_concurrency;` this wraps around `hardware_concurrency`(20 in my case) ,which is the value of total no. of available pus in my hardware and disregards mentioning of total no. of available cores(which creates the mask of 13 threads) , we can edit this line to use modulus no. of available pus intead , which could be another solution.
> 
> Why is `hardware_concurrency` equal to `20` on your system if you have only `13` physical processing units available?

my system has 20 pus from the 6+8 cores (6*2+8 = 20) but as we are checking for the scenario where we specify the number of cores using --hpx:cores , which is 7 in this care the no. of pus is 13(6*2+1)

0lem1of1:
> > what would you recommend?
> 
> Well, I'd suggest to try fixing the underlying issue instead of finding more or less sophisticated workarounds. If the user wants to oversubscribe the system - who are we to prevent him from shooting himself into the foot?

so i have pointed out the issue where the default binding algorithm (balanced) was the problem as it did not implement the wrap around logic 


also as i feel their is a conversation here i would like to point out what issue i have fixed here:
when i run a app which uses oversubscription , and run with a config which uses oversubcription(such as the 7 core 14 thread i have been using as an example ) it hangs as it gets stuck in a infinite loop , i addressed this by correcting the logic of the binding algo itself

`./build/bin/oversubscribing_resource_partitioner --hpx:cores=7 --hpx:threads=14`

ill update the PR description to elaborate the problem and how i have addressed this

0lem1of1:
> Since I opened the referenced issue, let me rephrase my original confusion, which, looking back, mostly came from how the default values are handled for those configuration variables.
> 
> https://github.com/STEllAR-GROUP/hpx/blob/3efee239010701ebcfb523b839c6647a36f3e5e9/docs/sphinx/manual/launching_and_configuring_hpx_applications.rst?plain=1#L1540-L1546
> 
> According to the documentation, the default value for `--hpx:threads` is `cores`. There are two ways to interpret that statement:
> 
> 1. The default value for `--hpx:threads` is the number of physical cores (system constant)
> 2. The default value for `--hpx:threads` is the number of cores HPX  is configured to use
> 
> I was originally expecting (2) to be true (because it avoids oversubscription), but currently it seems like (1) is true.

yes you're right , but i believe a system which avoids the hang and allows a program to run even with oversubscription would be better (considering the particular program is configured to be oversubscrbed)

hkaiser:
> > > what would you recommend?
> > 
> > 
> > Well, I'd suggest to try fixing the underlying issue instead of finding more or less sophisticated workarounds. If the user wants to oversubscribe the system - who are we to prevent him from shooting himself into the foot?
> 
> so i have pointed out the issue where the default binding algorithm (balanced) was the problem as it did not implement the wrap around logic
> 
> also as i feel their is a conversation here i would like to point out what issue i have fixed here: when i run a app which uses oversubscription , and run with a config which uses oversubcription(such as the 7 core 14 thread i have been using as an example ) it hangs as it gets stuck in a infinite loop , i addressed this by correcting the logic of the binding algo itself
> 
> `./build/bin/oversubscribing_resource_partitioner --hpx:cores=7 --hpx:threads=14`
> 
> ill update the PR description to elaborate the problem and how i have addressed this

This is the first time that anybody mentions that the error occurs when _both_ options (`--hpx:cores` and `--hpx:threads`) are being used at the same time. This explains things, at least it makes issues plausible as this has not been tested at all.

libs/core/affinity/src/parse_affinity_options.cpp
```diff
@@ -894,7 +894,7 @@
     {
         std::size_t const num_threads = affinities.size();
 
-        check_num_threads(use_process_mask, t, num_threads, ec);
+        // check_num_threads(use_process_mask, t, num_threads, ec);
```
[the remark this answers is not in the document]
hkaiser: CoPilot is right for a change: please remove code that is commented out.

hkaiser:
> so i have pointed out the issue where the default binding algorithm (balanced) was the problem as it did not implement the wrap around logic

What about the other binding algorithms (`compact` and `scatter`)? Do we need to change things there as well?

hkaiser:
May I also ask you to add a test (possibly more than one, i.e., one for `balanced`, `scatter`, and `compact` each). You should be able to do something like:
```
int hpx_main()
{
    return hpx::finalize();
}

int main()
{
    hpx::init_params init_args;
    init_args.cfg = {
        "--hpx:cores=2",        // list all required command line options here
        "--hpx:threads=4",
        "--hpx:bind=balanced"
    };
    return hpx::init(argc, argv, init_args);
}
```
to emulate running an HPX application with given command line arguments.

0lem1of1:
> > so i have pointed out the issue where the default binding algorithm (balanced) was the problem as it did not implement the wrap around logic
> 
> What about the other binding algorithms (`compact` and `scatter`)? Do we need to change things there as well?

compact handles this based on how it records which cores are filled with threads , I have changed scatter to include a similar wrap around logic as i did in balanced

0lem1of1:
> May I also ask you to add a test (possibly more than one, i.e., one for `balanced`, `scatter`, and `compact` each). You should be able to do something like:
> 
> ```
> int hpx_main()
> {
>     return hpx::finalize();
> }
> 
> int main()
> {
>     hpx::init_params init_args;
>     init_args.cfg = {
>         "--hpx:cores=2",        // list all required command line options here
>         "--hpx:threads=4",
>         "--hpx:bind=balanced"
>     };
>     return hpx::init(argc, argv, init_args);
> }
> ```
> 
> to emulate running an HPX application with given command line arguments.

sure ill do it , but i think to to trigger oversubscription ill have to use a absurdly oversubcribed config like cores=1 and threads=10 as the number of threads a core can run can vary for each system .

libs/core/affinity/src/parse_affinity_options.cpp
```diff
@@ -880,9 +884,15 @@ namespace hpx::threads::detail {
...
 
+                made_progress = true;
                 if (++num_thread == num_threads)
                     return;
             }
+            if (num_thread < num_threads && !made_progress)
+            {
+                std::fill(next_pu_index.begin(), next_pu_index.end(), 0);
+                made_progress = true;
+            }
```
hkaiser, line 895: Frankly speaking, I don't understand the change you're proposing to apply.  

Your code sets `made_progress = true;` inside the loop, which is guaranteed to be executed at least once (assuming `num_cores` is never zero). IOW, your `if (num_thread < num_threads && !made_progress)` will never execute.

Could you please explain step-by-step how the proposed change fixes the issue you're seeing?

0lem1of1: I understand why it looks like made_progress would always be true, but there is a specific edge case where it remains false: Oversubscription.

The inner loop iterates through the cores, but it only sets `made_progress = true` if it successfully finds an available PU on that core (i.e., inside the if (use_pu) block).

Here is the step-by-step scenario where the logic you described (that it always executes) fails:

Saturation: Suppose we have fewer hardware PUs than the requested num_threads. Eventually, we reach a state where we have assigned a thread to every available PU on every core.

The Stalled Pass: On the next iteration of the outer loop, we still have threads to place (`num_thread < num_threads`).

The Inner Loop Failure: The inner loop checks every core. However, for every core, `next_pu_index[num_core]` has reached its limit (`num_core_pus`). The while loop condition fails immediately for all cores.

Result: The code iterates through all `num_cores`, but never enters the block to increment `num_pus_cores` or set `made_progress = true`.

Without the fix, the outer loop would spin infinitely (or fail to assign the remaining threads) because made_progress stays false and the indices are never reset.

My change detects this "saturation" state:
`if (num_thread < num_threads && !made_progress)`

When this happens, it resets `next_pu_index` for all cores to `0`. This effectively allows the scheduler to "wrap around" and begin oversubscribing the cores starting from the first PU again.

hkaiser: Thanks for the clarification. This makes sense now.

PR #6884, opened 2026-02-06
implement fold algorithms acc. to c++26

### LINKED ISSUE #6224: Add C++23 fold algorithms

sleepingeight:
Fixes #6224 

Implemented fold algorithms adding additional traits which weren't present in hpx. Parallel versions for these algos cannot be implemented because these algorithms cannot be broken down into subproblems.

## Checklist

### DIFF
--- a/libs/core/synchronization/include/hpx/synchronization/shared_mutex.hpp
+++ b/libs/core/synchronization/include/hpx/synchronization/shared_mutex.hpp
@@ -158,7 +158,11 @@ namespace hpx::detail {
                         if (set_state(s1, s, lk))
                         {
                             HPX_ASSERT_OWNS_LOCK(lk);
-                            upgrade_cond.notify_one_no_unlock(lk);
+                            {
+                                [[maybe_unused]] hpx::util::
+                                    ignore_while_checking il(&lk);
+                                upgrade_cond.notify_one_no_unlock(lk);
+                            }
                             release_waiters(lk);
                             break;
                         }


libs/core/iterator_support/include/hpx/iterator_support/traits/is_foldable.hpp
```diff
@@ -0,0 +1,62 @@
...
+        std::copy_constructible<F> && std::indirectly_readable<I> &&
+        std::invocable<F&, T, std::iter_reference_t<I>> &&
+        std::convertible_to<
+            std::invoke_result_t<F&, T, std::iter_reference_t<I>>,
+            std::decay_t<
+                std::invoke_result_t<F&, T, std::iter_reference_t<I>>>> &&
+        is_indirect_binary_left_foldable_impl<F, T, I,
+            std::decay_t<
+                std::invoke_result_t<F&, T, std::iter_reference_t<I>>>>;
+
```
sleepingeight: @hkaiser do you recommend implementing all these traits (std::invocable, and others) in hpx. I haven't added them since they are present in c++20 but will add them upon your opinion

hkaiser: Things that are present in C++20 do not need to be replicated.

sleepingeight: As a side note, I see few traits being implemented in hpx which are in c++20, do we have any plans for removing them too?

libs/core/iterator_support/CMakeLists.txt
```diff
@@ -26,6 +27,7 @@ set(iterator_support_headers
 # cmake-format: off
 set(iterator_support_compat_headers
     hpx/iterator_support.hpp => hpx/modules/iterator_support.hpp
+    hpx/traits/is_foldable.hpp => hpx/modules/iterator_support.hpp
```
hkaiser: There is no need to add new headers to the compatibility headers.

libs/core/iterator_support/include/hpx/iterator_support/traits/is_foldable.hpp
hkaiser: Those can be replaced all over the code base. If you do, please use a separate PR, though. We're  in the process of bringing HPX to C++20, which will probably involve a lot of cleanup anyways.

libs/core/iterator_support/CMakeLists.txt
sleepingeight: done.

libs/core/algorithms/include/hpx/parallel/container_algorithms/fold.hpp
```diff
@@ -0,0 +1,628 @@
...
+        }
+
+        template <typename I2, typename T2,
+            typename Enable = std::enable_if_t<std::is_convertible_v<I, I2> &&
+                std::is_convertible_v<T, T2>>>
+        constexpr operator in_value_result<I2, T2>() &&
+        {
+            return {HPX_MOVE(in), HPX_MOVE(value)};
+        }
+    };
```
hkaiser: Could yuou please move tis type definition to https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/algorithms/include/hpx/parallel/util/result_types.hpp to keep all algorithm result types in one place?

libs/core/algorithms/include/hpx/parallel/container_algorithms/fold.hpp
```diff
@@ -0,0 +1,628 @@
...
+        {
+            return tag_fallback_invoke(hpx::ranges::fold_left_with_iter_t{},
+                hpx::util::begin(rng), hpx::util::end(rng), HPX_MOVE(init),
+                HPX_MOVE(f));
+        }
+    } fold_left_with_iter{};
+
+    ///////////////////////////////////////////////////////////////////////////
+    // CPO for hpx::ranges::fold_left_first_with_iter
+    inline constexpr struct fold_left_first_with_iter_t final
```
hkaiser: The implementation of `fold_left_first_with_iter` is essentially identical. Why did you separate the two?

sleepingeight:
@hkaiser parallel versions for these algorithms cannot be implemented. This is because, for example fold_left implies `f(f(f(f(init, x1), x2), ...), xn)`, since it cannot be broken down into smaller subproblems I don't think parallel versions are possible.

libs/core/algorithms/include/hpx/parallel/container_algorithms/fold.hpp
sleepingeight: I've refactored the implementation please take a look.

.github/workflows/linux_debug_fetch_hwloc.yml
```diff
@@ -16,6 +16,9 @@ jobs:
     steps:
     - uses: actions/checkout@v6
 
+    - name: Update apt repository
+      run: sudo apt-get update
+
```
hkaiser: This has been fixed on master, please rebase.

libs/core/algorithms/include/hpx/parallel/container_algorithms/fold.hpp
```diff
@@ -0,0 +1,583 @@
...
+            hpx::traits::is_input_iterator_v<InIter> &&
+            hpx::traits::is_sentinel_for_v<Sent, InIter> &&
+            std::is_same_v<T, hpx::traits::iter_value_t<InIter>> &&
+            hpx::traits::is_indirectly_binary_left_foldable<F, T, InIter>
+        )
+        // clang-format on
+        friend auto tag_fallback_invoke(
+            hpx::ranges::fold_left_t, InIter first, Sent last, T init, F f)
+        {
+            return tag_fallback_invoke(hpx::ranges::fold_left_with_iter_t{},
```
hkaiser: Please always forward through the full `tag_invoke` overload chain:
```suggestion
            return hpx::ranges::fold_left_with_iter(
```

libs/core/algorithms/include/hpx/parallel/util/result_types.hpp
```diff
@@ -398,6 +399,29 @@ namespace hpx::parallel::util {
...
+    ///////////////////////////////////////////////////////////////////////////
+    HPX_CXX_EXPORT template <typename I, typename T>
+    struct in_value_result
+    {
+        HPX_NO_UNIQUE_ADDRESS I in;
+        HPX_NO_UNIQUE_ADDRESS T value;
+
+        template <typename I2, typename T2>
+            requires(std::convertible_to<I const&, I2> &&
+                std::convertible_to<T const&, T2>)
```
hkaiser, line 405: I know this is unrelated, but would you mind replacing the `enable_if` constructs in this file with equivalent `requires` clauses? For context: we're slowly converting all of HPX' codebase to C++20, which is a step-by-step efforts...

libs/core/iterator_support/include/hpx/iterator_support/traits/is_foldable.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <hpx/modules/type_support.hpp>
+#include <concepts>
+#include <iterator>
+#include <type_traits>
+#include <utility>
+
+namespace hpx::traits {
+
+    ///////////////////////////////////////////////////////////////////////////
+    HPX_CXX_EXPORT template <class F, class T, class I, class U>
```
hkaiser: After recent changes, this now has to be 
```suggestion
    HPX_CXX_CORE_EXPORT template <class F, class T, class I, class U>
```

hkaiser: Also, by convention, we placed all concept definitions in `namespace hpx`

libs/core/iterator_support/include/hpx/iterator_support/traits/is_foldable.hpp
```diff
@@ -0,0 +1,63 @@
...
+            std::invoke_result_t<F&, T, std::iter_reference_t<I>>,
+            std::decay_t<
+                std::invoke_result_t<F&, T, std::iter_reference_t<I>>>> &&
+        is_indirect_binary_left_foldable_impl<F, T, I,
+            std::decay_t<
+                std::invoke_result_t<F&, T, std::iter_reference_t<I>>>>;
+
+    ///////////////////////////////////////////////////////////////////////////
+    namespace detail {
+        template <class F>
```
hkaiser: By convention, we use `typename` instead of `class` for template parameters.

libs/core/iterator_support/include/hpx/iterator_support/traits/is_foldable.hpp
```diff
@@ -0,0 +1,63 @@
...
+        class flipped
+        {
+            F f;
+
+        public:
+            template <class T, class U>
+                requires hpx::is_invocable_v<F&, U, T>
```
hkaiser: Shouldn't this be:
```suggestion
                requires hpx::is_invocable_v<F&, U&&, T&&>
```

libs/core/iterator_support/include/hpx/iterator_support/traits/is_foldable.hpp
```diff
@@ -0,0 +1,63 @@
...
+            constexpr hpx::util::invoke_result_t<F&, U, T> operator()(
+                T&& t, U&& u)
+            {
+                return HPX_INVOKE(f, std::forward<U>(u), std::forward<T>(t));
```
hkaiser: Inside the HPX codebase we use `HPX_FORWARD(T, t)` instead of `std::forward<T>(t)` to reduce compile times.

.github/workflows/linux_debug_fetch_hwloc.yml
sleepingeight: done.

libs/core/algorithms/include/hpx/parallel/container_algorithms/fold.hpp
sleepingeight: done.

libs/core/algorithms/include/hpx/parallel/util/result_types.hpp
sleepingeight: done.

libs/core/iterator_support/include/hpx/iterator_support/traits/is_foldable.hpp
sleepingeight: done.

libs/core/iterator_support/include/hpx/iterator_support/traits/is_foldable.hpp
sleepingeight: done.

libs/core/iterator_support/include/hpx/iterator_support/traits/is_foldable.hpp
sleepingeight: yes, missed it.

libs/core/iterator_support/include/hpx/iterator_support/traits/is_foldable.hpp
sleepingeight: done.

.github/workflows/linux_debug_fetch_hwloc.yml
hkaiser: You can remove this build step now as it has been already added to master.

libs/core/iterator_support/include/hpx/iterator_support/traits/is_range.hpp
```diff
@@ -75,4 +75,28 @@ namespace hpx::traits {
         using iterator_type = typename util::detail::iterator<R>::type;
         using sentinel_type = typename util::detail::sentinel<R>::type;
     };
+
+    ///////////////////////////////////////////////////////////////////////////
+    HPX_CXX_EXPORT template <typename R>
```
hkaiser: Please always use `HPX_CXX_CORE_EXPORT` to export symbols from `libs/core`:
```suggestion
    HPX_CXX_CORE_EXPORT template <typename R>
```

libs/core/algorithms/include/hpx/parallel/util/result_types.hpp
```diff
@@ -17,8 +17,10 @@
 #include <hpx/modules/execution_base.hpp>
 #include <hpx/modules/futures.hpp>
 
+#include <concepts>
 #include <type_traits>
 #include <utility>
+#include "hpx/config/export_definitions.hpp"
```
hkaiser: There is no need to include this file explicitly, it is being included by `<hpx/config.hpp>`

libs/core/iterator_support/include/hpx/iterator_support/traits/is_foldable.hpp
```diff
@@ -0,0 +1,64 @@
...
+                T&& t, U&& u)
+            {
+                return HPX_INVOKE(f, HPX_FORWARD(U, u), HPX_FORWARD(T, t));
+            }
+        };
+    }    // namespace detail
+
+    HPX_CXX_CORE_EXPORT template <typename F, typename T, typename I>
+    concept is_indirectly_binary_right_foldable =
+        is_indirectly_binary_left_foldable<detail::flipped<F>, T, I>;
```
hkaiser: I'm not sure whether it is correct to map `fold_right` onto `fold_left`. As you can see, the innermost operation is not flipping the arguments:

fold_left: `op(op(op(1, 2), 3), 4)` 
fold_right: `op(1, op(2, op(3, 4)))`

libs/core/iterator_support/include/hpx/iterator_support/traits/is_range.hpp
sleepingeight: done.

libs/core/iterator_support/include/hpx/iterator_support/traits/is_foldable.hpp
sleepingeight: The above concept is saying that right fold with `op` is equivalent to left fold with `flipped op` over the **reversed range**.

```
fold_right(op, [1, 2, 3], 4) = op(1, op(2, op(3, 4)))
= fold_left(flip(op), 4, [3, 2, 1]))
because 
flip(op)(4, 3) --> op(3, 4) 
and by expanding we get the same result as above.
```

hkaiser: Makes sense, thanks!

hkaiser: The other question would be whether we need a full implementation of `flipped`? Is it used anywhere except in the concept definition here?

sleepingeight: Not being used anywhere which is the reason I did not mark it to be exported. I can remove it and change the concept implementation based on your opinion.

hkaiser: Yes please do.

sleepingeight:
@hkaiser  anythting left on this PR?

hkaiser:
> @hkaiser anythting left on this PR?

I was waiting for the CI to cycle. All is good now. Thanks!

PR #6885, opened 2026-02-06
refactor execution parameters and steady clock usage

### LINKED ISSUE #6000: Change namespaces of Public API
This is a list of the names that need to change from sub-namespaces to the `hpx` namespace. So far the list consists of the following:

- `hpx::traits::is_bind_expression`
- `hpx::traits::is_placeholder`

the-ivii:
Fixes #6000 

## Proposed Changes

  - removed deprecated aliases for executor parameters (e.g., auto_chunk_size) from hpx::execution namespace, fully moving them to hpx::execution::experimental.
  - refactored steady_clock usage across core modules (threading, synchronization, timers) to use standard std::chrono types directly, removing reliance on custom HPX wrapper methods like .from_now() and .value().
  - fixed duration casting in runtime_support_server to ensure correct shutdown behavior with the updated clock implementation.

## Checklist

### DIFF
--- a/libs/core/execution/include/hpx/execution/executors/auto_chunk_size.hpp
+++ b/libs/core/execution/include/hpx/execution/executors/auto_chunk_size.hpp
@@ -45,7 +45,7 @@ namespace hpx::execution::experimental {
         ///
         constexpr explicit auto_chunk_size(
             std::uint64_t const num_iters_for_timing = 0) noexcept
-          : min_time_(200000)
+          : min_time_(std::chrono::nanoseconds(200000))
           , num_iters_for_timing_(num_iters_for_timing)
         {
         }
@@ -60,7 +60,8 @@ namespace hpx::execution::experimental {
         ///
         explicit auto_chunk_size(hpx::chrono::steady_duration const& rel_time,
             std::uint64_t const num_iters_for_timing = 0) noexcept
-          : min_time_(rel_time.value().count())
+          : min_time_(std::chrono::duration_cast<std::chrono::nanoseconds>(
+                rel_time.value()))
           , num_iters_for_timing_(num_iters_for_timing)
         {
         }
@@ -98,7 +99,8 @@ namespace hpx::execution::experimental {
                 if (test_chunk_size != 0)
                 {
                     t = (high_resolution_clock::now() - t) / test_chunk_size;
-                    if (t != 0 && this_.min_time_ >= t)
+                    if (t != 0 &&
+                        this_.min_time_ >= std::chrono::nanoseconds(t))
                     {
                         // return execution time for one iteration
                         return std::chrono::nanoseconds(t);
@@ -118,13 +120,14 @@ namespace hpx::execution::experimental {
             std::size_t const cores, std::size_t const count) noexcept
         {
             // return chunk size which will create the required amount of work
-            if (iteration_duration.value().count() != 0)
+            if (iteration_duration.value() !=
+                std::chrono::steady_clock::duration::zero())
             {
                 auto const ns =
                     std::chrono::duration_cast<std::chrono::nanoseconds>(
                         iteration_duration.value());
                 return (std::min) (count,
-                    (std::size_t) (this_.min_time_ / ns.count()));
+                    (std::size_t) (this_.min_time_.count() / ns.count()));
             }
             return (count + cores - 1) / cores;
         }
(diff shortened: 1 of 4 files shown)


libs/core/execution/include/hpx/execution/executors/adaptive_static_chunk_size.hpp
```diff
@@ -127,11 +127,3 @@ namespace hpx::execution::experimental {
...
     /// \endcond
 }    // namespace hpx::execution::experimental
-
-namespace hpx::execution {
-
-    using adaptive_static_chunk_size HPX_DEPRECATED_V(1, 9,
-        "hpx::execution::adaptive_static_chunk_size is deprecated, use "
-        "hpx::execution::experimental::adaptive_static_chunk_size instead") =
-        hpx::execution::experimental::adaptive_static_chunk_size;
-}    // namespace hpx::execution
```
hkaiser, line 137: I'd like to leave removing older compatibility interfaces for a separate PR as there are more of those lurking in various places beyond what you have touched on here.

libs/core/timing/include/hpx/timing/steady_clock.hpp
```diff
@@ -18,73 +18,10 @@ namespace hpx::chrono {
 
     HPX_CXX_EXPORT using std::chrono::steady_clock;
 
-    HPX_CXX_EXPORT class steady_time_point
-    {
-        using value_type = steady_clock::time_point;
+    using steady_time_point = std::chrono::steady_clock::time_point;
+    using steady_duration = std::chrono::steady_clock::duration;
```
hkaiser: These probably need to be decorated with `HPX_CXX_EXPORT` too.
```suggestion
    HPX_CXX_EXPORT using steady_time_point = std::chrono::steady_clock::time_point;
    HPX_CXX_EXPORT using steady_duration = std::chrono::steady_clock::duration;
```

libs/full/runtime_distributed/src/server/runtime_support_server.cpp
```diff
@@ -34,6 +34,7 @@
 #include <hpx/modules/synchronization.hpp>
 #include <hpx/modules/thread_support.hpp>
 #include <hpx/modules/threadmanager.hpp>
+#include <hpx/timing/steady_clock.hpp>
```
hkaiser: Please never `#include` internal headers (except inside the HPX module itself). Always use `#include <hpx/modules/_some_module_.hpp>` (which is already #included here).

hkaiser:
@the-ivii Could you please rebase your branch onto master (and resolve the conflicts)?

hkaiser:
Now clang-format isn't happy anymore:
```diff
diff --git a/libs/core/timing/include/hpx/timing/steady_clock.hpp b/libs/core/timing/include/hpx/timing/steady_clock.hpp
index 2e6ed15..5b8222c 100644
--- a/libs/core/timing/include/hpx/timing/steady_clock.hpp
+++ b/libs/core/timing/include/hpx/timing/steady_clock.hpp
@@ -18,7 +18,8 @@ namespace hpx::chrono {
 
     HPX_CXX_EXPORT using std::chrono::steady_clock;
 
-    HPX_CXX_EXPORT using steady_time_point = std::chrono::steady_clock::time_point;
+    HPX_CXX_EXPORT using steady_time_point =
+        std::chrono::steady_clock::time_point;
     HPX_CXX_EXPORT using steady_duration = std::chrono::steady_clock::duration;
 
     HPX_CXX_EXPORT inline constexpr steady_duration null_duration =
```

libs/core/timing/include/hpx/timing/steady_clock.hpp
hkaiser: Now that more changes related to C++20 modules have been merged, all code in `libs/core` has to be exported using the `HPX_CXX_CORE_EXPORT` macro instead of `HPX_CXX_EXPORT`. Sorry that causes problems with your PR.

libs/core/execution/include/hpx/execution/executors/auto_chunk_size.hpp
```diff
@@ -60,7 +60,9 @@ namespace hpx::execution::experimental {
         ///
         explicit auto_chunk_size(hpx::chrono::steady_duration const& rel_time,
             std::uint64_t const num_iters_for_timing = 0) noexcept
-          : min_time_(rel_time.value().count())
+          : min_time_(
+                std::chrono::duration_cast<std::chrono::nanoseconds>(rel_time)
+                    .count())
```
hkaiser: Wouldn't it be better to store the duration as a member instead of storing the number of nanoseconds? This is an unrelated issue, but it would make sense to apply this change while you're working on it.

hkaiser:
@the-ivii Many thanks for working on this!

PR #6887, opened 2026-02-07
Fix run loop and connect awaitable

isidorostsa:
In the stdexec side of our implementation we are relying on `stdexec::run_loop`'s internals to make our `make_future` start the loop when the future is executed. https://github.com/STEllAR-GROUP/hpx/blob/81291175455c12d74dc5193eb2bc952277a667c1/libs/core/execution/include/hpx/execution/algorithms/make_future.hpp#L197-L200

This only happens when the future is wrapping a sender whose completion scheduler is a `run_loop` scheduler. However, P2300 does not specify a way to get back to the `run_loop`, given one of its senders. So we rely on the rather following chain to work to obtain a reference to the original `run_loop`:

`sender` --> `get_completion_scheduler` --> `get_env` -> `.__loop_` data member.

This worked, until `stdexec` made (for ergonomic reasons) `run_loop` be a different type from what is stored internally in the `.__loop_` member [as seen here](https://github.com/NVIDIA/stdexec/blob/16878989ada99e2d036b978664de2ddbdedfb25d/include/stdexec/__detail/__run_loop.hpp#L293-L300)

### DIFF
--- a/libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
+++ b/libs/core/execution_base/include/hpx/execution_base/completion_signatures.hpp
@@ -550,19 +550,11 @@ namespace hpx::execution::experimental {
 
     HPX_CXX_EXPORT template <typename ReceiverID>
     using operation = hpx::execution::experimental::stdexec_internal::
-        __connect_awaitable_::__operation<ReceiverID>;
+        __connect_await::__operation<ReceiverID>;
 
     HPX_CXX_EXPORT template <typename ReceiverID>
     using promise = hpx::execution::experimental::stdexec_internal::
-        __connect_awaitable_::__promise<ReceiverID>;
-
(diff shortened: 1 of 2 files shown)


hkaiser:
Thanks @isidorostsa! Could you have a look at the formal problems, please (clang-format, inspect, etc.)?

hkaiser:
@isidorostsa Could you please rebase your branch onto master (and resolve the conflicts)?

isidorostsa:
Will do both, I'll get #6746 merged first though

hkaiser:
@isidorostsa Could you please rebase and fix the conflicts?

isidorostsa:
> @isidorostsa Could you please rebase and fix the conflict

This was sufficiently covered by https://github.com/STEllAR-GROUP/hpx/pull/6746

PR #6891, opened 2026-02-08
Refine unordered_map unit tests and add operator[] size stability check

vaibhavj2006:
Fixes #

## Proposed Changes

  -Refined unordered_map unit tests to avoid mutating the container during iteration.
  -Simplified global access validation to rely on stable operator[] semantics.
  -Added a focused unit test (test_operator_brackets_size_stability) to verify that operator[] does not change container size when   
   accessing existing keys.

   -While working on the existing unit tests, I noticed that the tests were modifying the map while iterating over it. This made the       
    behavior harder to understand and the purpose of the test unclear.
    
    I  updated the loop to avoid modifying the container during iteration. The tests now only check correctness, which makes the    
    intent clearer and failures easier to debug.

## Checklist

-  I have fixed a bug and have added a regression test.

### DIFF
--- a/components/containers/unordered/tests/unit/unordered_map.cpp
+++ b/components/containers/unordered/tests/unit/unordered_map.cpp
@@ -40,9 +40,13 @@ void test_global_iteration(hpx::unordered_map<Key, Value, Hash, KeyEqual>& m,
 
     for (std::size_t i = 0; i < size; ++i)
     {
-        Key key = std::to_string(i);
-        auto f = m.get_value(key);
-        HPX_TEST_EQ(f.get(), val);
+        std::string idx = std::to_string(i);
+        HPX_TEST_EQ(m[idx], val);
+        m[idx] = Value(i + 1);
+        HPX_TEST_EQ(m[idx], Value(i + 1));
+
+        auto f = m.get_value(idx);
+        HPX_TEST_EQ(f.get(), Value(i + 1));
     }
 
     //     // test normal iteration


vaibhavj2006:
Yes, the root cause was that the loop was reading and writing the map repeatedly using operator[], which is unsafe in a distributed context. In ctest, this effectively mutates the container while iterating/accessing it, leading to inconsistent results, though it passed locally.

components/containers/unordered/tests/unit/unordered_map.cpp
```diff
@@ -39,12 +38,11 @@ void test_global_iteration(hpx::unordered_map<Key, Value, Hash, KeyEqual>& m,
...
-    for (std::size_t i = 0; i != size; ++i)
+    for (std::size_t i = 0; i < size; ++i)
     {
-        std::string idx = std::to_string(i);
-        HPX_TEST_EQ(m[idx], val);
-        m[idx] = Value(i + 1);
-        HPX_TEST_EQ(m[idx], Value(i + 1));
+        Key key = std::to_string(i);
+        auto f = m.get_value(key);
+        HPX_TEST_EQ(f.get(), val);
```
hkaiser: This changes the semantics of the original test. Could you explain your rationale, please?

vaibhavj2006: I intended to cover the the asynchronous access path get_value(), which is part of distributed container API but, I agree this changes the semantics of the original test, which was validating operator[] behavior specifically so should i restore the original check and add a separate get_value() test.

PR #6893, opened 2026-02-10
fix: decouple hpx_debug from cmake_build_type and allow custom debug postfix

### LINKED ISSUE #5907: HPX_DEBUG/CMAKE_BUILD_TYPE not correctly decoupled/consistently used.
https://github.com/STEllAR-GROUP/hpx/blob/a7ea57986855ef3b93d31c827077f8ad23adb760/libs/core/config/include/hpx/config.hpp#L277-L296
assumes `HPX_DEBUG` for name mangling. 
(different issue: assumes fixed library prefixes and suffixes which can be easily changed by cmake variables. )

https://github.com/STEllAR-GROUP/hpx/blob/a7ea57986855ef3b93d31c827077f8ad23adb760/CMakeLists.txt#L2191-L2192
Hardcodes `d` suffix dependent on `CMAKE_BUILD_TYPE`

https://github.com/STEllAR-GROUP/hpx/blob/a7ea57986855ef3b93d31c827077f8ad23adb760/cmake/HPX_CompilerFlagsTargets.cmake#L18-L25
Tries to decouple `HPX_DEBUG` from `CMAKE_BUILD_TYPE` (which of course isn't correct if the above is done.)

Looking at that mess at least the third code fragment should be changed to 

```
   target_compile_definitions(hpx_private_flags INTERFACE "$<$<CONFIG:DEBUG>:HPX_DEBUG>") 
   target_compile_definitions(hpx_public_flags INTERFACE "$<$<CONFIG:DEBUG>:HPX_DEBUG>") 
```

or the second fragment must be applied to all `<CONFIG>` suffixes.

guptapratykshh:
Fixes #5907

## Proposed Changes

  - decoupled HPX_DEBUG from CMAKE_BUILD_TYPE by using generator expressions ($<$<CONFIG:Debug>:HPX_DEBUG>) instead of hardcoded conditional logic.
  - introduced HPX_DEBUG_POSTFIX CMake variable (defaults to "d") to allow users to customize or remove the debug library suffix.
  - updated hpx/config.hpp to use HPX_DEBUG_POSTFIX for name mangling and DLL string generation, ensuring the C++ code aligns with the build configuration.

## Checklist

### DIFF
--- a/docs/sphinx/manual/building_hpx.rst
+++ b/docs/sphinx/manual/building_hpx.rst
@@ -123,7 +123,7 @@ used CMake options.
 
 .. option:: HPX_DEBUG_POSTFIX
 
-   Set the postfix for debug libraries. The default is ``d``. This variable is used to set ``CMAKE_DEBUG_POSTFIX`` and is only relevant for Debug builds or multi-configuration generators.
+   Set the postfix for debug libraries. The default is ``d``. This variable is used to set ``CMAKE_DEBUG_POSTFIX`` and is only relevant for Debug builds or multi-configuration generators. The default rarely needs to be changed. It ensures that generated debug binaries have a different name than release binaries, which is important to avoid ABI problems when both debug and release binaries are installed on the same system.
 
 For a complete list of available |cmake|_ variables that influence the build of
 |hpx|, see :ref:`cmake_variables`.


cmake/HPX_CompilerFlagsTargets.cmake
```diff
@@ -14,24 +14,12 @@ target_compile_features(hpx_private_flags INTERFACE cxx_std_${HPX_CXX_STANDARD})
...
-    target_compile_definitions(hpx_private_flags INTERFACE HPX_DEBUG)
-    target_compile_definitions(hpx_public_flags INTERFACE HPX_DEBUG)
-  endif()
-endif()
+target_compile_definitions(
+  hpx_private_flags INTERFACE $<$<CONFIG:Debug>:HPX_DEBUG>
+)
+target_compile_definitions(
+  hpx_public_flags INTERFACE $<$<CONFIG:Debug>:HPX_DEBUG>
+)
```
hkaiser: Are you sure this is still exposing the same behavior as before? The comment explains the previous rationale:
```
  # HPX_DEBUG must be set without a generator expression as it determines ABI
  # compatibility. Projects in Release mode using HPX in Debug mode must have
  # HPX_DEBUG set, and projects in Debug mode using HPX in Release mode must not
  # have HPX_DEBUG set. HPX_DEBUG must also not be set by projects using HPX.
```
Is that still ensured?

libs/core/config/include/hpx/config.hpp
```diff
@@ -259,21 +259,21 @@
 #endif
 
 ///////////////////////////////////////////////////////////////////////////////
+#if defined(HPX_DEBUG) && defined(HPX_DEBUG_POSTFIX)
+#  define HPX_DEBUG_POSTFIX_S HPX_PP_STRINGIZE(HPX_DEBUG_POSTFIX)
+#else
+#  define HPX_DEBUG_POSTFIX_S ""
```
hkaiser: Why would you want for the debug and release libraries to have similar names. Those are not ABI compatible, so using the same name may lead to unexpected runtime issues. 

Also, having different names allows to have both versions of HPX (Debug and Release) be installed in the same place, if needed.

Could you elaborate, please?

guptapratykshh: I have reverted HPX_DEBUG changes to preserve original behavior and strict compatibility and also updated the pr so the custom postfix only applies to Debug builds(defaulting to 'd') ensuring release and debug libraries remain distinct and can coexist

hkaiser: @guptapratykshh Could you please summarize what has actually changed compared to the original code?

guptapratykshh: compared to the original code, cmake/HPX_CompilerFlagsTargets.cmake is now untouched. i reverted my changes there to strictly keep the existing ABI compatibility logic for HPX_DEBUG. 

The only changes are now that i added configurable HPX_DEBUG_POSTFIX, which defaults to "d." i wrapped this so it is only defined for debug builds or multi config generators, ensuring release builds remain without a postfix.  
and also i updated macros to use this new configuration variable instead of hardcoded strings in config.hpp

hkaiser: Would you expect anybody to actually change the debug-postfix? What would be the rationale of doing so?

guptapratykshh: primary rationale is flexibility for packaging and integration into diverse build environments. while 'd' is the standard default (preserved here), some users or distributions may require different conventions (like _d or _debug)to match their toolchain standards or to avoid conflicts when co-installing multiple library variants and so, this change simply exposes that option without enforcing it

hkaiser:
@guptapratykshh Could you please address the issue reported by inspect? We can go ahead and merge once that's fixed. The other CI errors are unrelated. Please feel free to ignore those.

hkaiser:
You may also want to add some documentation about the new CMake variable to the corresponding section of our docs.

guptapratykshh:
the changes are fixed for the inspect and docs is also added

docs/sphinx/manual/building_hpx.rst
```diff
@@ -121,6 +121,10 @@ used CMake options.
 
    Build tests.
 
+.. option:: HPX_DEBUG_POSTFIX
+
+   Set the postfix for debug libraries. The default is ``d``. This variable is used to set ``CMAKE_DEBUG_POSTFIX`` and is only relevant for Debug builds or multi-configuration generators.
```
hkaiser: Could you add a sentence saying that the default rarely needs to changed. Also we should say that this will make sure that the generated debug binaries end up having a different name compared to the release binaries which is important to avoid ABI problems when both (debug and release) binaries are installed on a system.

PR #6897, opened 2026-02-11
Implement concurrent data structures for Issue #2235

### LINKED ISSUE #2235: Concurrent data structures support

guptapratykshh:
Fixes #2235 

## Proposed Changes

  - implemented concurrent_vector, concurrent_unordered_map, and concurrent_unordered_setusing hpx::shared_mutex for thread safe access(reader-writer lock)
  - added concurrent_queue as an alias to hpx::lockfree::queue to provide TBB-compatible interface
  - added tbb_hash and tbb_hasher aliases in hash.hpp for compatibility
## Checklist

### DIFF
--- a/libs/core/concurrency/tests/performance/CMakeLists.txt
+++ b/libs/core/concurrency/tests/performance/CMakeLists.txt
@@ -8,7 +8,6 @@ if(HPX_WITH_TESTS_BENCHMARKS)
   add_hpx_executable(
     concurrent_structure_performance
     SOURCES concurrent_structure_performance.cpp
-    DEPENDENCIES hpx_concurrency hpx_init
     HPX_PREFIX ${HPX_BUILD_PREFIX}
   )
 endif()


hkaiser (commented):
Thank you for working on this!

However, I believe the data structures you added are not thread-safe in the sense that they expose non-safe iterators and allow direct access to the stored data elements through unsafe access functions (like `operator[]()`).

libs/core/concurrency/include/hpx/concurrency/hash.hpp
```diff
@@ -0,0 +1,21 @@
...
+#pragma once
+
+#include <hpx/config.hpp>
+
+#include <functional>
+
+namespace hpx::concurrent {
+
+    template <typename T>
+    using tbb_hash = std::hash<T>;
```
hkaiser: Why `tbb_hash`? What is this alias used for?

guptapratykshh: i added it for compatibility with existing TBB applications. since tbb::tbb_hash is commonly used in TBB code, aliasing it to std::hash makes it much easier for users to port their code to HPX without having to change their hasher definitions

hkaiser: I think we should leave TBB out of HPX. If TBB users would like to port their code they could define their own aliases to simplify their work.

Since you started mentioning TBB, why would a user drop using tbb's concurrent (lockfree!) data structures in favor of your suggested implementation? Even they move from TBB to HPX, they could still utilize TBB's concurrent data structures as those (as far I understand) don't do locking, so there wouldn't be any problem in leaving them in place even when running on top of HPX. 

That said, would you mind doing some performance analysis of your lock-based implementation compared to the equivalent TBB versions?

guptapratykshh: concurrent data structures is giving good results , std::mutex-protected standard containers in benchmark testing, showing ~1.1x faster push_back and ~2.8x faster random access for concurrent_vector, and notably ~2x faster insertion and ~5x faster lookup for concurrent_unordered_map. so, this confirms efficiency of leveraging hpx::spinlock  , and in consideration i will remove tbb_hash

```
--- Vector Benchmark (Total Ops: 500000, Threads: 4) ---
std::vector + mutex     Push Back: 0.0104877 s (47.6749 Mops/s)
std::vector + mutex     Random Access: 0.0217133 s (23.0274 Mops/s)
std::vector + spinlock  Push Back: 0.008629 s (57.9441 Mops/s)
std::vector + spinlock  Random Access: 0.00993204 s (50.3421 Mops/s)
hpx::concurrent_vector  Push Back: 0.00932196 s (53.6368 Mops/s)
hpx::concurrent_vector  Random Access: 0.00770492 s (64.8936 Mops/s)

--- Unordered Map Benchmark (Total Ops: 500000, Threads: 4) ---
std::unordered_map + mutex     Insert: 0.0462779 s (10.8043 Mops/s)
std::unordered_map + mutex     Lookup: 0.0311863 s (16.0327 Mops/s)
std::unordered_map + spinlock  Insert: 0.0217505 s (22.9879 Mops/s)
std::unordered_map + spinlock  Lookup: 0.00621371 s (80.4672 Mops/s)
hpx::concurrent_unordered_map  Insert: 0.0249756 s (20.0195 Mops/s)
hpx::concurrent_unordered_map  Lookup: 0.00627933 s (79.6263 Mops/s)
 
```

libs/core/concurrency/tests/performance/concurrent_structure_performance.cpp
```diff
@@ -0,0 +1,305 @@
...
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// clang-format off
+#include <hpx/config.hpp>
+#include <hpx/init.hpp>
+#include <hpx/modules/timing.hpp>
+#include <hpx/concurrency/concurrent_unordered_map.hpp>
+#include <hpx/concurrency/concurrent_vector.hpp>
```
hkaiser: Always use only HPX module #includes (except inside a module itself):
```suggestion
#include <hpx/modules/concurrency.hpp>
```

libs/core/concurrency/tests/performance/concurrent_structure_performance.cpp
```diff
@@ -0,0 +1,305 @@
+//  Copyright (c) 2026 Pratyksh Gupta
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// clang-format off
```
hkaiser: Why is this needed?

libs/core/concurrency/tests/performance/concurrent_structure_performance.cpp
```diff
@@ -0,0 +1,305 @@
...
+        vm["ops"].as<std::uint64_t>();    // Total ops per test
+
+    // Test Vector
+    {
+        std::cout << "\n--- Vector Benchmark (Total Ops: " << num_ops
+                  << ", Threads: " << num_threads << ") ---\n";
+
+        auto run_vector_test = [&](auto& vec, std::string name) {
+            std::vector<hpx::thread> threads;
+            std::atomic<int> start_flag{0};
```
hkaiser: A condition variable would make the synchronization easier.

libs/core/concurrency/tests/performance/concurrent_structure_performance.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+            for (std::uint64_t i = 0; i < num_threads; ++i)
+            {
+                threads.emplace_back([&, i]() {
+                    while (start_flag.load() == 0)
+                        hpx::this_thread::yield();
+                    for (std::uint64_t j = 0; j < num_ops / num_threads; ++j)
+                    {
+                        auto volatile val = vec[j % vec.size()];    // Read
+                        (void) val;
```
hkaiser: Would the following still prevent the compiler from optimizing away the access?
```suggestion
                        [[maybe_unused]] auto volatile val = vec[j % vec.size()];    // Read
```

libs/core/concurrency/tests/performance/concurrent_structure_performance.cpp
```diff
@@ -0,0 +1,305 @@
...
+    // Test Vector
+    {
+        std::cout << "\n--- Vector Benchmark (Total Ops: " << num_ops
+                  << ", Threads: " << num_threads << ") ---\n";
+
+        auto run_vector_test = [&](auto& vec, std::string name) {
+            std::vector<hpx::thread> threads;
+            std::atomic<int> start_flag{0};
+
+            hpx::chrono::high_resolution_timer t;
```
hkaiser: For more consistent benchmarking results you may want to run the loop more than once while computing the average execution times.

libs/core/concurrency/tests/unit/concurrent_data_structures.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include <string>
+#include <thread>
+#include <vector>
+
+void test_concurrent_vector()
+{
+    hpx::concurrent::concurrent_vector<int> v;
+    std::atomic<int> count{0};
+
+    std::vector<std::thread> threads;
```
hkaiser: Please perform the tests using `hpx::thread`, not `std::thread`.

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,374 @@
...
+        {
+            friend class concurrent_unordered_map;
+            std::unique_lock<hpx::util::spinlock> lock_;
+            T* value_;
+
+            accessor(std::unique_lock<hpx::util::spinlock>&& l, T* v)
+              : lock_(std::move(l))
+              , value_(v)
+            {
+            }
```
hkaiser: Just for our own sanity:
```suggestion
                HPX_ASSERT_OWNS_LOCK(lock_);
            }
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,374 @@
...
+
+        // Accessors
+        class accessor
+        {
+            friend class concurrent_unordered_map;
+            std::unique_lock<hpx::util::spinlock> lock_;
+            T* value_;
+
+            accessor(std::unique_lock<hpx::util::spinlock>&& l, T* v)
+              : lock_(std::move(l))
```
hkaiser: Inside the HPX code base we use `HPX_MOVE`:
```suggestion
              : lock_(HPX_MOVE(l))
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,374 @@
...
+
+            bool empty() const
+            {
+                return value_ == nullptr;
+            }
+
+            operator T&() const
+            {
+                if (!value_)
+                    throw std::runtime_error("Empty accessor dereference");
```
hkaiser: Should we throw an HPX exception instead?
```suggestion
                    HPX_THROW_EXCEPTION(hpx::error::invalid_status,
                        "concurrent_unordered_map::accessor::operatorT&()", 
                        "Empty accessor dereference");
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,374 @@
...
+                return *value_;
+            }
+
+            T& get() const
+            {
+                if (!value_)
+                    throw std::runtime_error("Empty accessor dereference");
```
hkaiser: This code is repeated three times, it maybe a good idea to factor that out into a private `accessor::validate()` function.

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,374 @@
...
+            T* value_;
+
+            accessor(std::unique_lock<hpx::util::spinlock>&& l, T* v)
+              : lock_(std::move(l))
+              , value_(v)
+            {
+            }
+
+        public:
+            accessor() = default;
```
hkaiser: This default constructor will leave the members uninitialized (in particular `value_`)

libs/core/concurrency/tests/unit/concurrent_data_structures.cpp
hkaiser: I'm not even sure how this test could pass as the new concurrent data structures rely on a HPX spinlock which I doubt would work well on a plain standard thread.

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,374 @@
...
+
+        using reference = accessor;
+        using const_reference = const_accessor;
+
+        // Accessors
+        class accessor
+        {
+            friend class concurrent_unordered_map;
+            std::unique_lock<hpx::util::spinlock> lock_;
+            T* value_;
```
hkaiser: You may want to consider providing a default value to be able to retain the defaulted default constructor.
```suggestion
            T* value_ = nullptr;
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,374 @@
...
+            }
+            return *this;
+        }
+
+        concurrent_unordered_map& operator=(
+            concurrent_unordered_map&& other) noexcept
+        {
+            if (this != &other)
+            {
+                std::lock_guard<hpx::util::spinlock> lock(mutex_);
```
hkaiser, line 108: WHy not use `std::lock` here as well?

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,374 @@
...
+
+        size_type size() const noexcept
+        {
+            std::lock_guard<hpx::util::spinlock> lock(mutex_);
+            return map_.size();
+        }
+
+        size_type max_size() const noexcept
+        {
+            return map_.max_size();
```
hkaiser, line 130: Shouldn't this be locked as well?
```suggestion
            std::lock_guard<hpx::util::spinlock> lock(mutex_);
            return map_.max_size();
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,374 @@
...
+        using const_reference = const_accessor;
+
+        // Accessors
+        class accessor
+        {
+            friend class concurrent_unordered_map;
+            std::unique_lock<hpx::util::spinlock> lock_;
+            T* value_;
+
+            accessor(std::unique_lock<hpx::util::spinlock>&& l, T* v)
```
hkaiser: I'd prefer for this constructor to take a reference instead of a pointer. That would avoid for the code using it to take the address. Let's hide the fact that we're using a pointer 100% inside this type.

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,374 @@
...
+        {
+            std::unique_lock<hpx::util::spinlock> lock(mutex_);
+            return accessor(std::move(lock), &map_[key]);
+        }
+
+        accessor operator[](Key&& key)
+        {
+            std::unique_lock<hpx::util::spinlock> lock(mutex_);
+            return accessor(std::move(lock), &map_[std::move(key)]);
+        }
```
hkaiser, line 191: C++26 adds another overload:
```suggestion
        template <typename K>
        accessor operator[](K&& key)
        {
            std::unique_lock<hpx::util::spinlock> lock(mutex_);
            return accessor(std::move(lock), &map_[HPX_FORWARD(K, key)]);
        }
```
Adding this might require a feature check, though...

BTW, the same is true for `insert`, `find`, and `contains` (possibly others).

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,374 @@
...
+            return const_accessor(std::move(lock), &map_.at(key));
+        }
+
+        size_type count(Key const& key) const
+        {
+            std::lock_guard<hpx::util::spinlock> lock(mutex_);
+            return map_.count(key);
+        }
+
+        bool find(Key const& key, accessor& result)
```
hkaiser: Why did you chose to change the API from:
```
accessor find(Key const& key) {...}
```
which would conform to the semantics of `std::unordered_map::find()`?

hkaiser: If you added an `explicit accessor::operator bool()`, you'd almost get the same semantics as your version exposes:
```
    concurrent_unordered_map<int, int> map;
    auto it = map.find(42);
    if (it) {...}
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,374 @@
...
+                result = const_accessor(std::move(lock), &it->second);
+                return true;
+            }
+            return false;
+        }
+
+        bool contains(Key const& key) const
+        {
+            std::lock_guard<hpx::util::spinlock> lock(mutex_);
+            return map_.find(key) != map_.end();
```
hkaiser, line 292:
```suggestion
            return map_.contains(key);
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,374 @@
...
+        }
+
+        // Thread-safe iteration
+        template <typename F>
+        void for_each(F&& f)
+        {
+            std::lock_guard<hpx::util::spinlock> lock(mutex_);
+            for (auto& kv : map_)
+            {
+                f(kv);
```
hkaiser: Let's allow for the iteration body to return a `bool` to be able to break out of the loop:
```suggestion
                if (!f(kv)) break;
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,374 @@
...
+        // Bucket interface
+        size_type bucket_count() const noexcept
+        {
+            std::lock_guard<hpx::util::spinlock> lock(mutex_);
+            return map_.bucket_count();
+        }
+
+        size_type max_bucket_count() const noexcept
+        {
+            return map_.max_bucket_count();
```
hkaiser, line 351:
```suggestion
            std::lock_guard<hpx::util::spinlock> lock(mutex_);
            return map_.max_bucket_count();
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,432 @@
...
+
+        public:
+            accessor() = default;
+
+            bool empty() const
+            {
+                return value_ == nullptr;
+            }
+
+            operator bool() const
```
hkaiser: Please make this operator `explicit`:
```suggestion
            explicit operator bool() const
```

hkaiser: Also, let's implement `operator bool()` in terms of `empty()` to make things consistent.

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,432 @@
...
+        concurrent_unordered_map() = default;
+
+        explicit concurrent_unordered_map(size_type bucket_count,
+            Hash const& hash = Hash(), KeyEqual const& equal = KeyEqual(),
+            Allocator const& alloc = Allocator())
+          : map_(bucket_count, hash, equal, alloc)
+        {
+        }
+
+        explicit concurrent_unordered_map(Allocator const& alloc)
```
hkaiser, line 72: `std::unordered_map` has two more constructors:
```
concurrent_unordered_map(size_type bucket_count, Allocator const& alloc);
concurrent_unordered_map(size_type bucket_count, Hash const& hash, Allocator const& alloc);
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,432 @@
...
+
+        concurrent_unordered_map(concurrent_unordered_map const& other)
+        {
+            std::lock_guard<hpx::util::spinlock> lock(other.mutex_);
+            map_ = other.map_;
+        }
+
+        concurrent_unordered_map(concurrent_unordered_map&& other) noexcept
+        {
+            std::lock_guard<hpx::util::spinlock> lock(other.mutex_);
```
hkaiser: I don't think we need to lock `other` as it is an rvalue (i.e. something that by definition nobody else works with).

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,432 @@
...
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/config.hpp>
+#include <hpx/concurrency/spinlock.hpp>
+#include <hpx/modules/errors.hpp>
+#include <hpx/type_support/assert_owns_lock.hpp>
```
hkaiser:
```suggestion
#include <hpx/modules/type_support.hpp>
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,432 @@
...
+#include <mutex>
+#include <optional>
+#include <type_traits>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+namespace hpx::concurrent {
+
+    template <typename Key, typename T, typename Hash = std::hash<Key>,
```
hkaiser: We most likely want to expose the new data structures as part of the generated C++ module interfaces:
```suggestion
    HPX_CXX_CORE_EXPORT template <typename Key, typename T,
        typename Hash = std::hash<Key>,
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,432 @@
...
+            {
+                if constexpr (std::is_convertible_v<
+                                  std::invoke_result_t<F, decltype(kv)>, bool>)
+                {
+                    if (!f(kv))
+                        break;
+                }
+                else
+                {
+                    f(kv);
```
hkaiser: Should we make sure `f(kv)` returns `void`? I'm not sure myself.
```suggestion
                    static_assert(std::is_voidv<std::invoke_result_t<F, decltype(kv)>>);
                    f(kv);
```

libs/core/concurrency/include/hpx/concurrency/concurrent_vector.hpp
```diff
@@ -0,0 +1,436 @@
...
+            }
+            T const& get() const
+            {
+                validate();
+                return *value_;
+            }
+        };
+
+        // Custom Index-based Iterator (Stable against growth)
+        template <bool IsConst, typename AccessorType>
```
hkaiser: What's your rationale for providing an iterator (which is not thread-safe on top of all)? Wouldn't it be better to expose similar APIs from all of the new data structures?

libs/core/concurrency/include/hpx/concurrency/concurrent_vector.hpp
```diff
@@ -0,0 +1,436 @@
...
+#include <vector>
+
+namespace hpx::concurrent {
+
+    template <typename T, typename Allocator = std::allocator<T>>
+    class concurrent_vector
+    {
+    private:
+        mutable hpx::util::spinlock mutex_;
+        std::deque<T, Allocator> data_;
```
hkaiser: Why do you rely on `std::deque` instead of `std::vector`?

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,432 @@
...
+            {
+                validate();
+                return *value_;
+            }
+
+            T& get() const
+            {
+                validate();
+                return *value_;
+            }
```
hkaiser: Do we need both, `operator T&` and `get()`? and if yes, why not implement one in terms of the other?

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
hkaiser: This would also need a `requires(!std::same_as<T, bool>)` to avoid ambiguities.

libs/core/concurrency/include/hpx/concurrency/concurrent_vector.hpp
```diff
@@ -0,0 +1,436 @@
...
+        }
+
+        size_type max_size() const noexcept
+        {
+            // Consistent with map
+            std::lock_guard<hpx::util::spinlock> lock(mutex_);
+            return data_.max_size();
+        }
+
+        void reserve(size_type /*new_cap*/) {}
```
hkaiser: Please either implement this or remove the function.

libs/core/concurrency/tests/performance/CMakeLists.txt
```diff
@@ -1,5 +1,13 @@
...
 #
 # SPDX-License-Identifier: BSL-1.0
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+if(HPX_WITH_TESTS_BENCHMARKS)
+  add_hpx_executable(
+    concurrent_structure_performance
+    SOURCES concurrent_structure_performance.cpp
+    DEPENDENCIES hpx_concurrency hpx_timing hpx_init
```
hkaiser: You shouldn't need to specify those dependencies (and actually, when C++ modules are enabled, this will break the build). `add_hpx_executable` will add `HPX::hpx` as a dependency for you.

libs/core/concurrency/tests/performance/concurrent_structure_performance.cpp
```diff
@@ -0,0 +1,394 @@
...
+            for (int iter = 0; iter < iterations; ++iter)
+            {
+                if constexpr (requires { vec.clear(); })
+                {
+                    vec.clear();
+                }
+                else
+                {
+                    // If no clear() method, assume wrapper exposes inner container directly or just skip.
+                    // But now all our wrappers have clear().
```
hkaiser: Why not remove the entire `if{}/else{}`, then?

libs/core/concurrency/tests/performance/concurrent_structure_performance.cpp
```diff
@@ -0,0 +1,394 @@
...
+                ready = false;
+
+                hpx::chrono::high_resolution_timer t;
+
+                for (std::uint64_t i = 0; i < num_threads; ++i)
+                {
+                    threads.emplace_back([&, i]() {
+                        std::unique_lock<hpx::mutex> l(mtx);
+                        cv.wait(l, [&] { return ready; });
+                        l.unlock();
```
hkaiser:
```suggestion
                        {
                            std::unique_lock<hpx::mutex> l(mtx);
                            cv.wait(l, [&] { return ready; });
                        }
```

libs/core/concurrency/tests/unit/concurrent_data_structures.cpp
```diff
@@ -0,0 +1,175 @@
+//  Copyright (c) 2026 Pratyksh Gupta
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/concurrency/concurrent_queue.hpp>
+#include <hpx/concurrency/concurrent_unordered_map.hpp>
+#include <hpx/concurrency/concurrent_unordered_set.hpp>
+#include <hpx/concurrency/concurrent_vector.hpp>
```
hkaiser:
```suggestion
#include <hpx/modules/concurrency.hpp>
```

libs/core/concurrency/tests/unit/concurrent_data_structures.cpp
```diff
@@ -0,0 +1,175 @@
...
+                count++;
+            }
+        });
+    }
+
+    for (auto& t : threads)
+        t.join();
+
+    HPX_TEST_EQ(static_cast<int>(v.size()), 1000);
+    HPX_TEST_EQ(static_cast<int>(v.size()), 1000);
```
hkaiser:
```suggestion
    HPX_TEST_EQ(static_cast<int>(v.size()), 1000);
```

libs/core/concurrency/tests/unit/concurrent_data_structures.cpp
```diff
@@ -0,0 +1,175 @@
...
+    HPX_TEST_EQ(count.load(), 1000);
+
+    // Test accessor
+    if (v.size() > 0)
+    {
+        v[0] = 999;
+        HPX_TEST_EQ(static_cast<int>(v[0]), 999);
```
hkaiser, line 47: Adding the following would force testing the `accessor::operator T&()`
```suggestion
        HPX_TEST_EQ(static_cast<int>(v[0]), 999);
        HPX_TEST(v[0] == 999);
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,460 @@
...
+            {
+                return value_ == nullptr;
+            }
+
+            template <typename U = T,
+                typename = std::enable_if_t<!std::is_same_v<U, bool>>>
+            explicit operator bool() const
+            {
+                return !empty();
+            }
```
hkaiser: Why not simply:
```suggestion
            explicit operator bool() const 
                requires(!std::same_as<T, bool>)
            {
                return !empty();
            }
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,460 @@
...
+        concurrent_unordered_map& operator=(
+            concurrent_unordered_map&& other) noexcept
+        {
+            if (this != &other)
+            {
+                std::lock(mutex_, other.mutex_);
+                std::lock_guard<hpx::util::spinlock> lock(
+                    mutex_, std::adopt_lock);
+                std::lock_guard<hpx::util::spinlock> other_lock(
+                    other.mutex_, std::adopt_lock);
```
hkaiser: `other` doesn't need to be locked here either (it's an rvalue as well).

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
hkaiser: Are you sure this works on all platforms before C++26? I'd suggest adding a feature test verifying that the standard library supports transparent keys for containers. The standard also says (https://eel.is/c++draft/unord.req#general-245):
```
The member function templates find, count, equal_range, contains, extract, erase, and 
bucket shall not participate in overload resolution unless the qualified-ids 
Pred​::​is_transparent and Hash​::​is_transparent are both valid and denote types. 
Additionally, the member function templates extract and erase shall not participate 
in overload resolution if 
is_convertible_v<K&&, iterator> || is_convertible_v<K&&, const_iterator> is true, 
where K is the type substituted as the first template argument.
```
I mentioned the transparent key overloads just for the sake of completeness, for now we're probably fine to leave them off.

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,460 @@
...
+        void for_each(F&& f) const
+        {
+            std::lock_guard<hpx::util::spinlock> lock(mutex_);
+            for (auto const& kv : map_)
+            {
+                static_assert(
+                    std::is_void_v<std::invoke_result_t<F, decltype(kv)>>);
+                f(kv);
+            }
+        }
```
hkaiser, line 339: Why have you removed the case when `f(kv)` returns a bool?

libs/core/concurrency/include/hpx/concurrency/deque.hpp
```diff
@@ -38,7 +38,8 @@ namespace hpx::lockfree {
 
     // The "left" and "right" terminology is used instead of top and bottom to
     // stay consistent with the paper that this code is based on..
-    HPX_CXX_CORE_EXPORT enum class deque_status_type : std::int8_t {
+    enum class deque_status_type : std::int8_t
```
hkaiser: What's the rationale for this change?

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,430 @@
...
+
+#include <hpx/config.hpp>
+#include <hpx/concurrency/spinlock.hpp>
+#include <hpx/modules/errors.hpp>
+#include <hpx/modules/type_support.hpp>
+
+#include <functional>
+#include <mutex>
+
+#include <stdexcept>
```
hkaiser, line 20: You're missing `#include <concepts>` now.

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,430 @@
...
+                    HPX_THROW_EXCEPTION(hpx::error::invalid_status,
+                        "concurrent_unordered_map::accessor",
+                        "Empty accessor dereference");
+                }
+            }
+
+        public:
+            accessor() = default;
+
+            bool empty() const
```
hkaiser: if the `operator bool` can be noexcept, so can be this function:
```suggestion
            bool empty() const noexcept
```

libs/core/concurrency/include/hpx/concurrency/deque.hpp
hkaiser: This file lost it's C++ module exports. Not sure why.

libs/core/concurrency/tests/performance/CMakeLists.txt
hkaiser: As said, there is no need to specify additional dependencies. Those are being added the CMake macro `add_hpx_executable`.

hkaiser (commented):
@guptapratykshh This pr starts to get in shape. There is still the issue fo the iterator interface exposed by the new concurrent_vector. Also we need to make a decision on whether to (conditionally) support the transparent key functionality introduced in C++26. I'd love to hear your oppinion on the latter.

guptapratykshh:
thanks for the feedback,on the iterators, you are right. the current std::vector wrapper is definitely a bottleneck. i will redo internal storage using a segmented design to ensure iterators stay stable even when the vector grows

i did also support transparent keys. avoiding temporary allocations on lookups is a clear performance win that fits HPX perfectly

hkaiser:
> thanks for the feedback,on the iterators, you are right. the current std::vector wrapper is definitely a bottleneck. i will redo internal storage using a segmented design to ensure iterators stay stable even when the vector grows

I don't think we need iterators to be exposed from the `concurrent_vector`. You don't have them exposed from the associative data structures either.

> i did also support transparent keys. avoiding temporary allocations on lookups is a clear performance win that fits HPX perfectly

If you want to add those, you will need to add a feature test that ensures transparent keys are supported by the underlying standard containers.

hkaiser:
inspect is unhappy still as well:
```
/libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp: *I* missing #include (cstddef) for symbol std::size_t on line 38
/libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_set.hpp: *I* missing #include (cstddef) for symbol std::size_t on line 36
/libs/core/concurrency/include/hpx/concurrency/concurrent_vector.hpp: *I* missing #include (cstddef) for symbol std::size_t on line 35, *I* missing #include (hpx/modules/errors.hpp) for symbol HPX_THROW_EXCEPTION on line 62
/libs/core/concurrency/tests/performance/concurrent_structure_performance.cpp: *I* missing #include (utility) for symbol std::pair on line 58, *Lic*
```

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -299,12 +317,37 @@ namespace hpx::concurrent {
...
+        }
+#endif
+
         size_type count(Key const& key) const
         {
             std::lock_guard<hpx::util::spinlock> lock(mutex_);
             return map_.count(key);
         }
 
+#if defined(HPX_HAVE_CXX20_STD_UNORDERED_TRANSPARENT_LOOKUP)
```
hkaiser: HPX requires C++20, so I think we can remove the tests related to C++20.

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_set.hpp
```diff
@@ -22,7 +22,7 @@
 
 namespace hpx::concurrent {
 
-    HPX_CXX_CORE_EXPORT template <typename Key, typename Hash = std::hash<Key>,
+    template <typename Key, typename Hash = std::hash<Key>,
```
hkaiser: We still want to export the new types through C++ modules, please re-add the macros.

hkaiser (commented):
Things start to look good now. One more thing: now there is a lot of code repitition between the containers. DO you see a way to extract the accessors (at least) to avoid duplicating them three times?

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,398 @@
...
+        using mapped_type = T;
+        using value_type = std::pair<Key const, T>;
+        using size_type = std::size_t;
+        using difference_type = std::ptrdiff_t;
+        using hasher = Hash;
+        using key_equal = KeyEqual;
+        using allocator_type = Allocator;
+
+        using accessor = detail::concurrent_accessor<T, false>;
+        using const_accessor = detail::concurrent_accessor<T, true>;
```
hkaiser: You could possibly remove the second template argument for your accessor type by saying:
```suggestion
        using accessor = detail::concurrent_accessor<T>;
        using const_accessor = detail::concurrent_accessor<T const>;
```

libs/core/concurrency/include/hpx/concurrency/deque.hpp
```diff
@@ -56,7 +56,7 @@ namespace hpx::lockfree {
         return static_cast<int>(lhs) == rhs;
     }
 
-    HPX_CXX_CORE_EXPORT template <typename T>
+    template <typename T>
```
hkaiser: Why did you remove the macro exporting this symbol through C++ modules?

guptapratykshh: fixed this

hkaiser:
There are still open comments, please have a look. Also, please rebase this onto current master to resolve potential compilation problems.

guptapratykshh:
@hkaiser i think this is ready to be shipped, if any reviews will look into that

hkaiser:
@guptapratykshh Thank you for your patience with this! Good job!

PR #6866, opened 2026-01-31
Fix threads hanging indefinitely by acquiring the state mutex before checking for an exclusive lock

Momensamir12:
Fixes #6854 

## Proposed Changes

- Fixed a race condition in `shared_mutex::lock_shared()` where threads could hang indefinitely
- The mutex protecting the condition variable is now acquired before checking the wait condition

## Any background context you want to provide?

**Note:** The same issue might exist in other lock functions.

## Checklist

- [ ] I have added a new feature and have added tests to go along with it.
- [x] I have fixed a bug and have added a regression test.
- [ ] I have added a test using random numbers; I have made sure it uses a seed, and that random numbers generated are valid inputs for the tests.

PR #6900, opened 2026-02-12
shared_mutex: follow-up fix and add regression test 

Momensamir12:
Fixes #

This PR is a follow up work to #6866 It addresses additional races in the upgrade/exclusive transition and adds regression tests to ensure progress under contention.
  
## Proposed Changes
  
  - ```lock_upgrade()``` 
    - Prevent race conditions by acquiring ```state_change``` mutex before checking the wait condition 
    - Fix a bug that sets ```++shared_count = true``` corrupting ```shared_count``` state
  - ```unlock_upgrade_and_lock()``` 
    - add a boolean flag to prevent the same thread from releasing shared lock more than once if it fails to upgrade to exclusive and retries the loop, corrupting the state of ```shared_count```
    - acquire the mutex before checking the wait condition
  - ```try_unlock_shared_and_lock()```  
    - Should return false if ```s.data.shared_count != 1``` instead of  ```s.data.shared_count == 1``` 
  - Add regression tests exercising mixed  shared, upgrade, and exclusive operations in tight loops.

## Checklist


hkaiser:
@Momensamir12 Could you please have a look at the inspect issues reported:
```
/libs/core/synchronization/include/hpx/synchronization/shared_mutex.hpp: *END* file doesn't end with a newline
/libs/core/synchronization/tests/regressions/shared_mutex_6854.cpp: *END* file doesn't end with a newline
```

libs/core/synchronization/include/hpx/synchronization/shared_mutex.hpp
```diff
@@ -272,7 +272,7 @@ namespace hpx::detail {
 
                 auto s1 = s;
 
-                ++s.data.shared_count = true;
+                ++s.data.shared_count;
```
hkaiser, line 275: Excellent catch!

hkaiser (commented):
Very nice, thanks!

Momensamir12:
> Very nice, thanks!

My pleasure

libs/core/synchronization/include/hpx/synchronization/shared_mutex.hpp
```diff
@@ -419,7 +427,7 @@ namespace hpx::detail {
             {
                 auto s = state.load(std::memory_order_acquire);
                 if (s.data.exclusive || s.data.exclusive_waiting_blocked ||
-                    s.data.upgrade || s.data.shared_count == 1)
+                    s.data.upgrade || s.data.shared_count != 1)
```
hkaiser, line 430: Should this be 
```suggestion
                    s.data.upgrade || s.data.shared_count > 1)
```
?

I'm not sure, though.

Momensamir12: I don't think so because this function should be called by a reader / shared lock thats trying to acquire an exclusive / writer lock so  ```shared_count``` should be at least 1.

hkaiser: Isn't that what I suggested?

Momensamir12: Yes but wouldn't ```> 1``` make it succeed if its 0 which violates the precondition that the caller thread is  currently a reader. ``` != 1``` rejects this

hkaiser:
Thank you again for looking into these issues!

PR #6868, opened 2026-01-31
Fixing zip_iterator::operator[]() to agree with gcc 16 std::sort

hkaiser:
Fixes #6867

PR #6902, opened 2026-02-13
using standard traits present in iterator library

sleepingeight:
Removed the existing traits implementation and updated them with the corresponding concepts available in the standard. Key Changes - 
```cpp
hpx::traits::is_forward_iterator<Iter>	-> std::forward_iterator<Iter>
hpx::traits::is_bidirectional_iterator<Iter>	-> std::bidirectional_iterator<Iter>
hpx::traits::is_random_access_iterator<Iter>	-> std::random_access_iterator<Iter>
hpx::traits::is_output_iterator<Iter>	-> std::output_iterator<Iter, T>
hpx::traits::is_input_iterator<Iter>	-> std::input_iterator<Iter>
hpx::traits::is_sentinel_for<S, I>	-> std::sentinel_for<S, I>
hpx::traits::is_range<T>	-> std::ranges::range<T>
```
Removed unused concepts such as `bidirectional_concept`, `random_access_concept`.
Removed an unwanted config - `HPX_WITH_CXX20_STD_DISABLE_SIZED_SENTINEL_FOR`.
Modified test files accordingly.

### DIFF
--- a/libs/core/algorithms/tests/unit/algorithms/util/test_merge_four.cpp
+++ b/libs/core/algorithms/tests/unit/algorithms/util/test_merge_four.cpp
@@ -97,9 +97,9 @@ void test1()
     // Two ranges
     D.clear();
     for (std::uint32_t i = 0; i < 10; ++i)
-        A.push_back(static_cast<std::uint64_t>(i) * 2 + 1);
+        A.push_back(i * 2 + 1);
     for (std::uint32_t i = 0; i < 10; ++i)
-        D.push_back(static_cast<std::uint64_t>(i) * 2);
+        D.push_back(i * 2);
     X.resize(20, 0);
     RA = range_t(A.begin(), A.end());
     RD = range_t(D.begin(), D.end());
@@ -119,11 +119,11 @@ void test1()
     A.clear();
     D.clear();
     for (std::uint32_t i = 0; i < 10; ++i)
-        A.push_back(static_cast<std::uint64_t>(i) * 3 + 2);
+        A.push_back(i * 3 + 2);
     for (std::uint32_t i = 0; i < 10; ++i)
-        B.push_back(static_cast<std::uint64_t>(i) * 3 + 1);
+        B.push_back(i * 3 + 1);
     for (std::uint32_t i = 0; i < 10; ++i)
-        D.push_back(static_cast<std::uint64_t>(i) * 3);
+        D.push_back(i * 3);
     X.resize(30, 0);
     RA = range_t(A.begin(), A.end());
     RB = range_t(B.begin(), B.end());
@@ -146,13 +146,13 @@ void test1()
     D.clear();
     X.clear();
     for (std::uint32_t i = 0; i < 10; ++i)
-        A.push_back(static_cast<std::uint64_t>(i) * 4 + 3);
+        A.push_back(i * 4 + 3);
     for (std::uint32_t i = 0; i < 10; ++i)
-        B.push_back(static_cast<std::uint64_t>(i) * 4 + 2);
+        B.push_back(i * 4 + 2);
     for (std::uint32_t i = 0; i < 10; ++i)
-        C.push_back(static_cast<std::uint64_t>(i) * 4 + 1);
+        C.push_back(i * 4 + 1);
     for (std::uint32_t i = 0; i < 10; ++i)
-        D.push_back(static_cast<std::uint64_t>(i) * 4);
+        D.push_back(i * 4);
 
     X.resize(40, 0);
     RA = range_t(A.begin(), A.end());
@@ -316,9 +316,9 @@ void test3()
     // Two ranges
     D.clear();
     for (std::uint32_t i = 0; i < 10; ++i)
-        A.push_back(static_cast<std::uint64_t>(i) * 2 + 1);
+        A.push_back(i * 2 + 1);
     for (std::uint32_t i = 0; i < 10; ++i)
-        D.push_back(static_cast<std::uint64_t>(i) * 2);
+        D.push_back(i * 2);
     //X.resize ( 20, 0);
     RA = range_t(A.begin(), A.end());
     RD = range_t(D.begin(), D.end());
@@ -338,11 +338,11 @@ void test3()
     A.clear();
     D.clear();
     for (std::uint32_t i = 0; i < 10; ++i)
-        A.push_back(static_cast<std::uint64_t>(i) * 3 + 2);
+        A.push_back(i * 3 + 2);
     for (std::uint32_t i = 0; i < 10; ++i)
-        B.push_back(static_cast<std::uint64_t>(i) * 3 + 1);
+        B.push_back(i * 3 + 1);
     for (std::uint32_t i = 0; i < 10; ++i)
-        D.push_back(static_cast<std::uint64_t>(i) * 3);
+        D.push_back(i * 3);
     //X.resize ( 30 , 0);
     RA = range_t(A.begin(), A.end());
     RB = range_t(B.begin(), B.end());
@@ -365,13 +365,13 @@ void test3()
     D.clear();
     //X.clear() ;
     for (std::uint32_t i = 0; i < 10; ++i)
-        A.push_back(static_cast<std::uint64_t>(i) * 4 + 3);
+        A.push_back(i * 4 + 3);
     for (std::uint32_t i = 0; i < 10; ++i)
-        B.push_back(static_cast<std::uint64_t>(i) * 4 + 2);
+        B.push_back(i * 4 + 2);
     for (std::uint32_t i = 0; i < 10; ++i)
-        C.push_back(static_cast<std::uint64_t>(i) * 4 + 1);
+        C.push_back(i * 4 + 1);
     for (std::uint32_t i = 0; i < 10; ++i)
-        D.push_back(static_cast<std::uint64_t>(i) * 4);
+        D.push_back(i * 4);
 
     //X.resize ( 40 , 0);
     RA = range_t(A.begin(), A.end());
--- a/libs/core/algorithms/tests/unit/algorithms/util/test_merge_vector.cpp
+++ b/libs/core/algorithms/tests/unit/algorithms/util/test_merge_vector.cpp
@@ -9,7 +9,6 @@
 #include <hpx/modules/testing.hpp>
 
 #include <algorithm>
-#include <cstddef>
 #include <cstdint>
 #include <type_traits>
 #include <utility>
@@ -115,8 +114,7 @@ void test2()
     std::vector<rng> Vin, Vout;
     for (std::uint32_t i = 0; i < 9; ++i)
     {
-        Vin.emplace_back(VA.begin() + (static_cast<std::ptrdiff_t>(i) * 10),
-            VA.begin() + ((static_cast<std::ptrdiff_t>(i) + 1) * 10));
+        Vin.emplace_back(VA.begin() + (i * 10), VA.begin() + ((i + 1) * 10));
     }
     compare comp;
     merge_level4(RB, Vin, Vout, comp);
@@ -191,8 +189,7 @@ void test4()
     std::vector<range<xk*>> Vout;
     for (std::uint32_t i = 0; i < 9; ++i)
     {
-        Vin.emplace_back(VA.begin() + (static_cast<std::ptrdiff_t>(i) * 10),
-            VA.begin() + ((static_cast<std::ptrdiff_t>(i) + 1) * 10));
+        Vin.emplace_back(VA.begin() + (i * 10), VA.begin() + ((i + 1) * 10));
     }
     compare comp;
     uninit_merge_level4(RB, Vin, Vout, comp);
@@ -227,8 +224,7 @@ void test5()
     std::vector<rng> Vin, Vout;
     for (std::uint32_t i = 0; i < 10; ++i)
     {
-        Vin.emplace_back(X.begin() + (static_cast<std::ptrdiff_t>(i) * 10),
-            X.begin() + ((static_cast<std::ptrdiff_t>(i) + 1) * 10));
+        Vin.emplace_back(X.begin() + (i * 10), X.begin() + ((i + 1) * 10));
     }
     rng RX(merge_vector4(Rin, Rout, Vin, Vout, comp));
     HPX_TEST(RX.size() == 100);
@@ -261,8 +257,7 @@ void test6()
     std::vector<rng> Vin, Vout;
     for (std::uint32_t i = 0; i < 16; ++i)
     {
-        Vin.emplace_back(VA.begin() + (static_cast<std::ptrdiff_t>(i) * 10),
-            VA.begin() + ((static_cast<std::ptrdiff_t>(i) + 1) * 10));
+        Vin.emplace_back(VA.begin() + (i * 10), VA.begin() + ((i + 1) * 10));
     }
     rng RA(VA.begin(), VA.end());
     rng RB(VB.begin(), VB.end());


libs/core/algorithms/include/hpx/parallel/util/prefetching.hpp
```diff
@@ -42,11 +42,11 @@ namespace hpx::parallel::util {
             using iterator_category = std::random_access_iterator_tag;
             using value_type = typename std::iterator_traits<Itr>::value_type;
             using difference_type = std::ptrdiff_t;
-            using pointer = value_type*;
-            using reference = value_type&;
+            using pointer = value_type const*;
+            using reference = value_type;
```
hkaiser, line 47: What's the rationale for this change?

libs/core/algorithms/include/hpx/parallel/util/prefetching.hpp
```diff
@@ -42,11 +42,11 @@ namespace hpx::parallel::util {
...
 
         private:
-            using ranges_type = hpx::tuple<std::reference_wrapper<Ts>...>;
+            using ranges_type = hpx::tuple<Ts const*...>;
```
hkaiser, line 50: What's the benefit of replacing `std::reference_wrapper` with pointers?

libs/core/algorithms/tests/unit/container_algorithms/for_loop_range_generator.cpp
```diff
@@ -9,7 +9,7 @@
 // clang up to V12 (Apple clang up to v15) and gcc above V13 refuse to compile
 // the code below
 #if defined(HPX_HAVE_CXX20_COROUTINES) &&                                      \
-    (!defined(HPX_CLANG_VERSION) || HPX_CLANG_VERSION >= 130000) &&            \
+    (!defined(HPX_CLANG_VERSION) || HPX_CLANG_VERSION < 130000) &&             \
```
hkaiser: Huh?

libs/core/algorithms/tests/unit/container_algorithms/for_loop_range_generator.cpp
```diff
@@ -49,24 +49,24 @@ namespace test {
...
-        auto const lo = p.first;
-        auto const hi = p.last;
 
-        if (auto const stride = p.stride; stride > 0)
-        {
-            for (auto i = lo; i < hi; i += stride)
-                co_yield i;
-        }
-        else if (stride < 0)
+        hpx::generator<int&, int> iterate(index_pair const p) noexcept
```
hkaiser: What does making this a member function change?

libs/core/algorithms/tests/unit/container_algorithms/for_loop_range_generator.cpp
```diff
@@ -89,8 +89,6 @@ namespace test {
         return {static_cast<int>(p.first + first),
             static_cast<int>(p.first + first + size), p.stride};
     }
-
-    static_assert(hpx::traits::is_range_generator_v<index_pair>);
```
hkaiser, line 93: Please don't remove the range_generator traits.

libs/core/iterator_support/include/hpx/iterator_support/traits/is_iterator.hpp
```diff
@@ -307,28 +136,28 @@ namespace hpx::traits {
         // be properly recognized as satisfying the ForwardIterator concept.
         HPX_CXX_CORE_EXPORT template <typename Iter>
         struct satisfy_traversal_concept<Iter, hpx::forward_traversal_tag>
-          : bidirectional_concept<Iter>
         {
+            static constexpr bool value = std::bidirectional_iterator<Iter>;
```
hkaiser: Please leave this as a metafunction:
```
        HPX_CXX_CORE_EXPORT template <typename Iter>
        struct satisfy_traversal_concept<Iter, hpx::forward_traversal_tag>
          : std::bool_constant<std::bidirectional_iterator<Iter>>
        {};
```
Same for the other traits below.

libs/core/iterator_support/include/hpx/iterator_support/iterator_facade.hpp
```diff
@@ -334,12 +334,10 @@ namespace hpx::util {
...
 
-            HPX_HOST_DEVICE constexpr auto operator[](difference_type n) const
+            HPX_HOST_DEVICE constexpr reference operator[](
+                difference_type n) const
             {
-                using use_proxy = use_operator_brackets_proxy<Derived, T>;
-
-                return make_operator_brackets_result<Derived>(
-                    this->derived() + n, use_proxy{});
+                return *(this->derived() + n);
```
hkaiser: This disables the special handling of `operator[]()`s return type as proxies, if required.

libs/core/iterator_support/include/hpx/iterator_support/range.hpp
```diff
@@ -102,7 +101,7 @@ namespace hpx::util {
             struct fallback
             {
                 template <typename T>
-                fallback(T const&)
+                constexpr fallback(T const&)
```
hkaiser: if you make this constexpr, make it noexcept as well.

libs/core/iterator_support/tests/include/hpx/iterator_support/tests/iter_sent.hpp
```diff
@@ -169,6 +169,26 @@ struct iterator
...
+    }
+
+    friend bool operator==(sentinel<Value> s, iterator const& it)
+    {
+        return it.state == s.get_stop();
+    }
+
+    friend bool operator!=(iterator const& it, sentinel<Value> s)
+    {
+        return it.state != s.get_stop();
```
hkaiser:
```suggestion
        return !(it == s);
```

libs/core/iterator_support/tests/performance/stencil3_iterators.cpp
```diff
@@ -26,40 +26,29 @@ int partition_size = 10000;
...
     template <typename Iterator>
     HPX_FORCEINLINE Iterator next(Iterator const& it)
     {
-        return next(it, hpx::traits::is_random_access_iterator<Iterator>());
+        if constexpr (std::random_access_iterator<Iterator>)
+        {
+            return it + 1;
+        }
+        else
+            ++it;
```
hkaiser:
```suggestion
            {
                return ++it;
            }
```

libs/core/iterator_support/tests/unit/is_iterator.cpp
```diff
@@ -230,6 +230,47 @@ struct random_access_traversal_iterator
     }
 };
 
+template <typename T, typename U>
+concept addition_result =
+    requires { typename hpx::traits::detail::addition_result_t<T, U>; };
```
hkaiser, line 226:
```suggestion
    requires { hpx::traits::detail::addition_result_t<T, U>; };
```

libs/core/iterator_support/tests/unit/is_iterator.cpp
```diff
@@ -521,609 +553,7 @@ void subtraction_result_test()
...
-            (!is_random_access_iterator<iterator>::value), "hpx test iterator");
-    }
-    {
-        using base_iterator = std::vector<std::size_t>::iterator;
-        using iterator =
-            test::test_iterator<base_iterator, std::input_iterator_tag>;
-
-        HPX_TEST_MSG(
-            (!is_random_access_iterator<iterator>::value), "hpx test iterator");
-    }
```
hkaiser, line 1126: Why did you remove the concept checks?

hkaiser:
inspect reports:
```
/libs/core/algorithms/tests/unit/algorithms/util/test_merge_vector.cpp: *I* missing #include (cstddef) for symbol std::ptrdiff_t on line 117
```

libs/core/iterator_support/tests/unit/is_iterator.cpp
sleepingeight: I think this suggestion fails since the requirement is not properly established if the typename keyword is removed (when typename is mentioned -> type requirement). Also this doesn't compile.

libs/core/iterator_support/tests/unit/is_iterator.cpp
sleepingeight: removed them since I thought that they were a bit unnecessary since we removed many of the concepts present here. I've re-added them with the standard concepts.

libs/core/iterator_support/include/hpx/iterator_support/traits/is_iterator.hpp
sleepingeight: done.

libs/core/algorithms/include/hpx/parallel/util/prefetching.hpp
sleepingeight: Forward iterators and higher required to be default constructible acc. to the standard. Since references are not default constructible they were erroring out so modified them. Same reason applies for the below change too.

libs/core/algorithms/tests/unit/container_algorithms/for_loop_range_generator.cpp
sleepingeight: Sorry, i wasn't able to debug the test and wanted to not compile it temporarily, hence the changes. I've fixed it in the next commit.

libs/core/iterator_support/include/hpx/iterator_support/iterator_facade.hpp
sleepingeight: Yes, but this was cause of few errors (i don't exactly remember the cause, it was a while ago). Are there any cases where the prev. version was necessary?

sleepingeight:
There is a known regression with the changes made, things don't work the same as before with `vector<bool>` since     `std::output_iterator<std::vector<bool>::iterator, std::iter_value_t<std::vector<bool>::iterator>> is false` which was previously true with our implementation.

components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_segmented_iterator.hpp
```diff
@@ -148,10 +148,24 @@ namespace hpx::segmented {
...
+            // NOLINTBEGIN(bugprone-unhandled-self-assignment)
             local_vector_value_proxy& operator=(
-                local_vector_value_proxy const&) = default;
+                local_vector_value_proxy const& other) const
+            {
+                T val = static_cast<T>(other);
+                const_cast<local_vector_value_proxy&>(*this).operator=(
+                    HPX_MOVE(val));
+                return const_cast<local_vector_value_proxy&>(*this);
+            }
```
hkaiser: I don't understand this change. Why is the `operator=()` `const`?

libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp
```diff
@@ -122,6 +122,25 @@ namespace hpx::util {
...
+        }
+
+    public:
+        HPX_HOST_DEVICE constexpr Incrementable operator[](
+            typename base_type::difference_type n) const
+        {
+            Incrementable tmp = this->base();
+            std::advance(tmp, n);
+            return tmp;
+        }
```
hkaiser: These functions are implemented in the base class (`iterator_adaptor`). Why do we need them here again?

libs/core/iterator_support/tests/unit/is_iterator.cpp
```diff
@@ -100,47 +100,26 @@ struct bidirectional_traversal_iterator
...
-
-    bool operator>(bidirectional_traversal_iterator const& that) const
-    {
-        return this->state > that.state;
-    }
-
-    bool operator>=(bidirectional_traversal_iterator const& that) const
-    {
-        return this->state >= that.state;
-    }
```
hkaiser, line 122: What's the rationale of removing these operators?

libs/full/collectives/tests/regressions/collectives_bool_5940.cpp
```diff
@@ -41,12 +62,14 @@ void test_all_reduce_bool()
     // test functionality based on immediate local result value
     for (int i = 0; i != 10; ++i)
     {
-        bool value = i % 2 ? true : false;
+        custom_bool value = i % 2 ? true : false;
```
hkaiser: The purpose of this test is to verify that the collective operation does the right thing even for `bool`. I doubt that the use of your `custom_bool` would trigger the code path that handles booleans.

libs/core/algorithms/tests/unit/container_algorithms/for_loop_range_generator.cpp
hkaiser: This is still a member function, why?

libs/core/iterator_support/include/hpx/iterator_support/iterator_facade.hpp
hkaiser: Please leave it as it was and report the errors here. A potential fix for those errors might be found here: https://github.com/STEllAR-GROUP/hpx/pull/6868

libs/core/algorithms/tests/unit/container_algorithms/for_loop_range_generator.cpp
sleepingeight: Done. I've added it.

libs/core/algorithms/tests/unit/container_algorithms/for_loop_range_generator.cpp
sleepingeight: No, I've removed it as a member function. could you PTAL again.

libs/core/iterator_support/tests/unit/is_iterator.cpp
sleepingeight: Bidirectional iterators according to the std. don't implement these operators.

tools/inspect/include_check.cpp
```diff
@@ -304,6 +304,12 @@ namespace boost { namespace inspect {
         {"(\\bstd\\s*::\\s*regular_invocable\\b)", "std::regular_invocable",
             {"concepts"}},
         {"(\\bstd\\s*::\\s*same_as\\b)", "std::same_as", {"concepts"}},
+        {"(\\bstd\\s*::\\s*input_iterator\\b)", "std::input_iterator", {"concepts"}},
+        {"(\\bstd\\s*::\\s*output_iterator\\b)", "std::output_iterator", {"concepts"}},
+        {"(\\bstd\\s*::\\s*forward_iterator\\b)", "std::forward_iterator", {"concepts"}},
+        {"(\\bstd\\s*::\\s*bidirectional_iterator\\b)", "std::bidirectional_iterator", {"concepts"}},
+        {"(\\bstd\\s*::\\s*random_access_iterator\\b)", "std::random_access_iterator", {"concepts"}},
+        {"(\\bstd\\s*::\\s*sentinel_for\\b)", "std::sentinel_for", {"concepts"}},
```
hkaiser: I apologize, it seems I mislead you. These concepts are defined in `<iterator>`, not `<concepts>`. The way you specified it here, inspect will report things wrongly.

components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_segmented_iterator.hpp
```diff
@@ -148,11 +148,6 @@ namespace hpx::segmented {
             local_vector_value_proxy(local_vector_value_proxy const&) = default;
             local_vector_value_proxy(local_vector_value_proxy&&) = default;
 
-            local_vector_value_proxy& operator=(
-                local_vector_value_proxy const&) = default;
-            local_vector_value_proxy& operator=(
-                local_vector_value_proxy&&) = default;
-
```
hkaiser, line 155: Please add these back.

components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_segmented_iterator.hpp
```diff
@@ -255,8 +250,22 @@ namespace hpx::segmented {
...
+                return const_cast<vector_value_proxy&>(*this);
+            }
+            vector_value_proxy& operator=(
+                vector_value_proxy&& other) const
+            {
+                v_->set_value(launch::sync, index_,
+                    other.v_->get_value(launch::sync, other.index_));
+                return const_cast<vector_value_proxy&>(*this);
+            }
+            // NOLINTEND(bugprone-unhandled-self-assignment)
```
hkaiser: Please revert this change as well.

components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_segmented_iterator.hpp
```diff
@@ -268,7 +277,7 @@ namespace hpx::segmented {
             template <typename T_,
                 typename Enable = std::enable_if_t<
                     !std::is_same_v<std::decay_t<T_>, vector_value_proxy>>>
-            vector_value_proxy& operator=(T_&& value)
+            vector_value_proxy const& operator=(T_&& value) const
```
hkaiser, line 271: This is not correct.

sleepingeight: I feel that it this change is essential for things to work. This is because we want the `vector_iterator` to satisfy `std::output_iterator` in a few cases and to do this it needs to satisfy `std::indirectly_writable`. From [cppreference](https://en.cppreference.com/w/cpp/iterator/indirectly_writable.html) - 

```cpp
struct Object
{
    Object& operator=(const Object& other) = default;
 
    int x;
};
 
struct ProxyReference
{
    ProxyReference& operator=(const ProxyReference& other) = default;
 
    const ProxyReference& operator=(const Object& o) const
    {
        *p = o;
        return *this;
    }
 
    Object* p;
};
 
struct I1 { Object& operator*(); };
 
struct I2 { Object operator*(); };
 
struct I3 { ProxyReference operator*(); };
 
static_assert(std::indirectly_writable<I1, Object>);
static_assert(!std::indirectly_writable<I2, Object>);
static_assert(std::indirectly_writable<I3, Object>);
static_assert(!std::indirectly_writable<I3, ProxyReference>);
 
void f(I1 i1, I2 i2, I3 i3, Object o)
{
    *i1 = o;  // OK, assigns to the value referenced by *i1
 
    *i2 = o;  // OK, but meaningless: This assigns to the temporary returned by *i2
 
    *i3 = o;  // OK, calls ProxyReference::operator=(const Object& o) const
              // which performs *ptr = o, where ptr is (*i3).p
}

```

tools/inspect/include_check.cpp
sleepingeight: done.

libs/full/collectives/tests/regressions/collectives_bool_5940.cpp
sleepingeight: Yes, but vector<bool> doesn't seem to satisfy output_iterator. This will not work with any of the algos due to this requirement.

libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp
sleepingeight: Not required, removed them.

components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_segmented_iterator.hpp
sleepingeight: a better version is implemented now. This code is removed.

libs/core/iterator_support/include/hpx/iterator_support/iterator_facade.hpp
sleepingeight: the errors were fixed by adding `using use_brackets_proxy = std::false_type` in the implementation of few iterators. Hence removed the changes here.

sleepingeight:
@hkaiser ~I don't understand how the GH actions builds are failing, to me it seems like they are running on a different commit (judging from their errors). Could you PTAL.~ Also thank you for patiently reviewing. Learning a lot from this PR.


Fixed them, the errors spew from the recent changes on master. Fixed them.

components/containers/partitioned_vector/include/hpx/components/containers/partitioned_vector/partitioned_vector_segmented_iterator.hpp
hkaiser: Ok, wierd but ok. I consede.

libs/core/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
```diff
@@ -17,6 +17,8 @@
 #include <hpx/parallel/util/loop.hpp>
 
 #include <cstddef>
+#include <iterator>
+#include <ranges>
```
hkaiser: Why do we need `<ranges>` here? Did we miss that to begin with? If we missed it, please add the corresponding symbol to inspect.

libs/core/algorithms/include/hpx/parallel/algorithms/find.hpp
```diff
@@ -1012,6 +1012,8 @@ namespace hpx {
 #include <iterator>
 #include <type_traits>
 #include <utility>
+#include <iterator>
+#include <ranges>
```
hkaiser: Same here: did we originally miss adding `<ranges>`?

libs/core/algorithms/include/hpx/parallel/algorithms/for_loop.hpp
```diff
@@ -1066,7 +1067,7 @@ namespace hpx::parallel {
                     loop_iter(hpx::util::begin(g), hpx::util::end(g),
                         part_index, current_thread);
                 }
-                else if constexpr (hpx::traits::is_range_v<B>)
+                else if constexpr (std::ranges::range<B>)
```
hkaiser, line 1071: This most likey requires for `<ranges>` being included. Please also add this symbol to inspect.

libs/core/algorithms/include/hpx/parallel/container_algorithms/copy.hpp
```diff
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include "hpx/iterator_support/traits/is_iterator.hpp"
```
hkaiser: Please don't #include module-internal files directly, always use the (generated) module header: 
```suggestion
#include <hpx/modules/iterator_support.hpp>
```
Also, please always include HPX headers using `<...>`, not "..."`

libs/core/algorithms/include/hpx/parallel/container_algorithms/count.hpp
```diff
@@ -462,12 +463,12 @@ namespace hpx::ranges {
         template <typename ExPolicy, typename Rng,
             typename Proj = hpx::identity,
             typename T = typename hpx::parallel::traits::projected<
-                hpx::traits::range_iterator_t<Rng>, Proj>::value_type>
+                std::ranges::iterator_t<Rng>, Proj>::value_type>
```
hkaiser, line 467: Please add this symbol to inspect as well.

libs/core/algorithms/tests/unit/algorithms/test_utils.hpp
```diff
@@ -39,6 +39,15 @@ namespace test {
...
         }
+
+        template <typename Tag = IteratorTag,
+            typename Enable = std::enable_if_t<
+                std::is_same_v<Tag, std::random_access_iterator_tag>>>
+        HPX_HOST_DEVICE typename base_type::reference operator[](
+            typename base_type::difference_type n) const
+        {
+            return *(this->base() + n);
+        }
```
hkaiser: This is provided by the base class, no need to overload it here again.

libs/core/algorithms/tests/unit/algorithms/test_utils.hpp
```diff
@@ -68,6 +77,17 @@ namespace test {
...
+        template <typename Tag = IteratorTag,
+            typename Enable = std::enable_if_t<
+                std::is_same_v<Tag, std::random_access_iterator_tag>>>
+        HPX_HOST_DEVICE typename base_type::reference operator[](
+            typename base_type::difference_type n) const
+        {
+            if (m_callback)
+                m_callback();
+            return *(this->base() + n);
+        }
```
hkaiser: Why is this necessary?

libs/core/algorithms/tests/unit/container_algorithms/test_utils.hpp
```diff
@@ -80,6 +82,15 @@ namespace test {
...
         }
+
+        template <typename Tag = IteratorTag,
+            typename Enable = std::enable_if_t<
+                std::is_same_v<Tag, std::random_access_iterator_tag>>>
+        HPX_HOST_DEVICE typename base_type::reference operator[](
+            typename base_type::difference_type n) const
+        {
+            return *(this->base() + n);
+        }
```
hkaiser: This is provided by the base class, no need to overload it here again.

libs/core/algorithms/tests/unit/container_algorithms/test_utils.hpp
```diff
@@ -161,6 +172,17 @@ namespace test {
...
+        template <typename Tag = IteratorTag,
+            typename Enable = std::enable_if_t<
+                std::is_same_v<Tag, std::random_access_iterator_tag>>>
+        HPX_HOST_DEVICE typename base_type::reference operator[](
+            typename base_type::difference_type n) const
+        {
+            if (m_callback)
+                m_callback();
+            return *(this->base() + n);
+        }
```
hkaiser: Why is this needed?

libs/core/execution/include/hpx/execution/executors/polymorphic_executor.hpp
```diff
@@ -86,6 +86,8 @@ namespace hpx::parallel::execution {
             using pointer = std::size_t*;
             using reference = std::size_t&;
 
+            shape_iter() = default;
```
hkaiser, line 89: What are the semantics of the default constructor? It leaves the iterator in an unusable state.

libs/core/iterator_support/include/hpx/iterator_support/traits/is_iterator.hpp
```diff
@@ -328,7 +157,7 @@ namespace hpx::traits {
             satisfy_traversal_concept<Iter, TraversalTag>::value;
     }    // namespace detail
 
-    HPX_CXX_CORE_EXPORT template <typename Iter, typename Enable = void>
+    HPX_CXX_CORE_EXPORT template <typename Iter>
```
hkaiser: I'm not sure is there is any code out there that may rely on customizing `hpx::is_iterator`. Please leave the SFINAE template argument in place.

libs/core/iterator_support/include/hpx/iterator_support/iterator_facade.hpp
hkaiser: If you change the return type of `make_operator_brackets_result` below, then this must be:
```suggestion
                operator_brackets_proxy<Iterator>,
                typename Iterator::reference>;
```

libs/core/iterator_support/include/hpx/iterator_support/range.hpp
```diff
@@ -316,6 +306,16 @@ namespace hpx::util {
         {
         };
 
+        template <typename T>
+        struct is_range_generator<T,
+            std::void_t<decltype(detail::iterate_impl(std::declval<T&>(), 0L))>>
+          : std::integral_constant<bool,
+                !std::is_same_v<decltype(detail::iterate_impl(
+                                    std::declval<T&>(), 0L)),
+                    range_impl::fallback>>
```
hkaiser: This is too loose, `iterate()` is required to return a range.

libs/core/iterator_support/tests/include/hpx/iterator_support/tests/iter_sent.hpp
```diff
@@ -17,6 +17,8 @@
 template <typename ValueType>
 struct sentinel
 {
+    sentinel() = default;
```
hkaiser, line 20: This constructs an unusable sentinel (`Value` may be uninitialized)

libs/core/iterator_support/tests/performance/stencil3_iterators.cpp
```diff
@@ -26,40 +26,31 @@ int partition_size = 10000;
...
     {
-        return next(it, hpx::traits::is_random_access_iterator<Iterator>());
+        if constexpr (std::random_access_iterator<Iterator>)
+        {
+            return it + 1;
+        }
+        else
+        {
+            return ++it;
+        }
```
hkaiser: These could be replaced with `std::next` and `std::prev`

libs/full/collectives/tests/regressions/collectives_bool_5940.cpp
hkaiser: Hmm, what error do you see when leaving `bool` in place?

libs/core/algorithms/include/hpx/parallel/algorithms/for_loop.hpp
sleepingeight: i've updated the inspect include checks.

libs/core/algorithms/include/hpx/parallel/container_algorithms/copy.hpp
sleepingeight: sorry for this. updated.

libs/core/algorithms/tests/unit/algorithms/test_utils.hpp
sleepingeight: updated this to use `use_brackets_proxy = std::false_type`.

libs/core/algorithms/tests/unit/algorithms/test_utils.hpp
sleepingeight: not necessary removed it.

libs/core/algorithms/tests/unit/container_algorithms/test_utils.hpp
sleepingeight: removed.

libs/core/iterator_support/include/hpx/iterator_support/range.hpp
sleepingeight: updated.

libs/core/iterator_support/tests/include/hpx/iterator_support/tests/iter_sent.hpp
sleepingeight: updated to value initialise the data. Default construction is required for few concepts.

libs/full/collectives/tests/regressions/collectives_bool_5940.cpp
sleepingeight: <details>
  <summary>Error Log</summary>

```
/Users/surya/hpx/libs/core/algorithms/include/hpx/parallel/algorithms/exclusive_scan.hpp:534:27: error: static assertion failed: Requires at least output iterator.
  534 |             static_assert(std::output_iterator<OutIter,
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  535 |                               hpx::traits::iter_value_t<InIter>>,
      |                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/surya/hpx/libs/core/tag_invoke/include/hpx/functional/detail/tag_fallback_invoke.hpp:296:24: note: in instantiation of function template specialization 'hpx::tag_fallback_invoke<std::__bit_iterator<std::vector<bool>, false>, std::__bit_iterator<std::vector<bool>, false>, (lambda at /Users/surya/hpx/libs/full/collectives/include/hpx/collectives/exclusive_scan.hpp:423:33), bool>' requested here
  296 |                 return tag_fallback_invoke(
      |                        ^
/Users/surya/hpx/libs/full/collectives/include/hpx/collectives/detail/communicator.hpp:130:51: note: in instantiation of function template specialization 'hpx::traits::communication_operation<hpx::collectives::detail::communicator_server, hpx::traits::communication::exclusive_scan_init_tag>::get<hpx::future<bool>, bool, bool, std::logical_or<void>>' requested here
  130 |             return collective_operation::template get<Result>(
      |                                                   ^
/Users/surya/hpx/libs/full/collectives/include/hpx/collectives/detail/communicator.hpp:147:39: note: in instantiation of function template specialization 'hpx::collectives::detail::communicator_server::get_result<hpx::traits::communication::exclusive_scan_init_tag, hpx::future<bool>, bool, bool, std::logical_or<void>>' requested here
  147 |                 &communicator_server::get_result<Operation, Result, Args...>,
      |                                       ^
/Users/surya/hpx/libs/full/collectives/include/hpx/collectives/exclusive_scan.hpp:630:16: note: in instantiation of function template specialization 'hpx::collectives::exclusive_scan<bool &, bool &, std::logical_or<void>>' requested here
  630 |         return exclusive_scan(HPX_MOVE(fid), HPX_FORWARD(T, local_result),
      |                ^
/Users/surya/hpx/libs/full/collectives/tests/regressions/collectives_bool_5940.cpp:128:13: note: in instantiation of function template specialization 'hpx::collectives::exclusive_scan<bool &, bool &, std::logical_or<void>>' requested here
  128 |             exclusive_scan(exclusive_scan_bool_client, value, value,
      |             ^
/Users/surya/hpx/libs/core/algorithms/include/hpx/parallel/algorithms/exclusive_scan.hpp:534:27: note: because 'std::output_iterator<std::__bit_iterator<std::vector<_Bool>, false>, hpx::traits::iter_value_t<__bit_iterator<vector<_Bool, allocator<_Bool> >, false, 0> > >' evaluated to false
  534 |             static_assert(std::output_iterator<OutIter,
      |                           ^
/Library/Developer/CommandLineTools/SDKs/MacOSX14.5.sdk/usr/include/c++/v1/__iterator/concepts.h:148:3: note: because 'indirectly_writable<std::__bit_iterator<std::vector<_Bool>, false>, _Bool>' evaluated to false
  148 |   indirectly_writable<_Ip, _Tp> &&
      |   ^
/Library/Developer/CommandLineTools/SDKs/MacOSX14.5.sdk/usr/include/c++/v1/__iterator/concepts.h:77:54: note: because 'const_cast<const iter_reference_t<_Out> &&>(*__o) = std::forward<_Tp>(__t)' would be invalid: no viable overloaded '='
   77 |     const_cast<const iter_reference_t<_Out>&&>(*__o) = _VSTD::forward<_Tp>(__t);                       // not required to be equality-preserving
      |                                                      ^
In file included from /Users/surya/hpx/libs/full/collectives/tests/regressions/collectives_bool_5940.cpp:10:
In file included from /Users/surya/hpx/libs/full/include/include/hpx/hpx.hpp:9:
In file included from /Users/surya/hpx/libs/full/include/include/hpx/algorithm.hpp:10:
In file included from /Users/surya/hpx/build/libs/core/algorithms/include/hpx/modules/algorithms.hpp:47:
/Users/surya/hpx/libs/core/algorithms/include/hpx/parallel/algorithms/inclusive_scan.hpp:713:27: error: static assertion failed: Requires at least output iterator.
  713 |             static_assert(std::output_iterator<OutIter,
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  714 |                               hpx::traits::iter_value_t<InIter>>,
      |                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/surya/hpx/libs/core/tag_invoke/include/hpx/functional/detail/tag_fallback_invoke.hpp:296:24: note: in instantiation of function template specialization 'hpx::tag_fallback_invoke<std::__bit_iterator<std::vector<bool>, false>, std::__bit_iterator<std::vector<bool>, false>, (lambda at /Users/surya/hpx/libs/full/collectives/include/hpx/collectives/inclusive_scan.hpp:284:47)>' requested here
  296 |                 return tag_fallback_invoke(
      |                        ^
/Users/surya/hpx/libs/full/collectives/include/hpx/collectives/detail/communicator.hpp:130:51: note: in instantiation of function template specialization 'hpx::traits::communication_operation<hpx::collectives::detail::communicator_server, hpx::traits::communication::inclusive_scan_tag>::get<hpx::future<bool>, bool, std::logical_or<void>>' requested here
  130 |             return collective_operation::template get<Result>(
      |                                                   ^
/Users/surya/hpx/libs/full/collectives/include/hpx/collectives/detail/communicator.hpp:147:39: note: in instantiation of function template specialization 'hpx::collectives::detail::communicator_server::get_result<hpx::traits::communication::inclusive_scan_tag, hpx::future<bool>, bool, std::logical_or<void>>' requested here
  147 |                 &communicator_server::get_result<Operation, Result, Args...>,
      |                                       ^
/Users/surya/hpx/libs/full/collectives/include/hpx/collectives/inclusive_scan.hpp:370:16: note: in instantiation of function template specialization 'hpx::collectives::inclusive_scan<bool &, std::logical_or<void>>' requested here
  370 |         return inclusive_scan(HPX_MOVE(fid), HPX_FORWARD(T, local_result),
      |                ^
/Users/surya/hpx/libs/full/collectives/tests/regressions/collectives_bool_5940.cpp:152:13: note: in instantiation of function template specialization 'hpx::collectives::inclusive_scan<bool &, std::logical_or<void>>' requested here
  152 |             inclusive_scan(inclusive_scan_client, value, std::logical_or<>{},
      |             ^
/Users/surya/hpx/libs/core/algorithms/include/hpx/parallel/algorithms/inclusive_scan.hpp:713:27: note: because 'std::output_iterator<std::__bit_iterator<std::vector<_Bool>, false>, hpx::traits::iter_value_t<__bit_iterator<vector<_Bool, allocator<_Bool> >, false, 0> > >' evaluated to false
  713 |             static_assert(std::output_iterator<OutIter,
      |                           ^
/Library/Developer/CommandLineTools/SDKs/MacOSX14.5.sdk/usr/include/c++/v1/__iterator/concepts.h:148:3: note: because 'indirectly_writable<std::__bit_iterator<std::vector<_Bool>, false>, _Bool>' evaluated to false
  148 |   indirectly_writable<_Ip, _Tp> &&
      |   ^
/Library/Developer/CommandLineTools/SDKs/MacOSX14.5.sdk/usr/include/c++/v1/__iterator/concepts.h:77:54: note: because 'const_cast<const iter_reference_t<_Out> &&>(*__o) = std::forward<_Tp>(__t)' would be invalid: no viable overloaded '='
   77 |     const_cast<const iter_reference_t<_Out>&&>(*__o) = _VSTD::forward<_Tp>(__t);                       // not required to be equality-preserving
      |                                                      ^
```

tools/inspect/include_check.cpp
```diff
@@ -304,6 +304,24 @@ namespace boost { namespace inspect {
...
         {"(\\bstd\\s*::\\s*same_as\\b)", "std::same_as", {"concepts"}},
+        // std iterator
+        {"(\\bstd\\s*::\\s*input_iterator\\b)", "std::input_iterator", {"iterator"}},
+        {"(\\bstd\\s*::\\s*output_iterator\\b)", "std::output_iterator", {"iterator"}},
+        {"(\\bstd\\s*::\\s*forward_iterator\\b)", "std::forward_iterator", {"iterator"}},
+        {"(\\bstd\\s*::\\s*bidirectional_iterator\\b)", "std::bidirectional_iterator", {"iterator"}},
+        {"(\\bstd\\s*::\\s*random_access_iterator\\b)", "std::random_access_iterator", {"iterator"}},
+        {"(\\bstd\\s*::\\s*sentinel_for\\b)", "std::sentinel_for", {"iterator"}},
+        {"(\\bstd\\s*::\\s*next\\b)", "std::next", {"iterator"}},
+        {"(\\bstd\\s*::\\sprevt\\b)", "std::prev", {"iterator"}},
```
hkaiser:
```suggestion
        {"(\\bstd\\s*::\\s*prev\\b)", "std::prev", {"iterator"}},
```

libs/full/collectives/tests/regressions/collectives_bool_5940.cpp
hkaiser: Ok, I will have a look at the implementation of the collectives, please revert your changes wrt `bool` here.

libs/full/collectives/tests/regressions/collectives_bool_5940.cpp
```diff
@@ -46,7 +46,8 @@ void test_all_reduce_bool()
         hpx::future<bool> overall_result = all_reduce(all_reduce_bool_client,
             value, std::logical_or<>{}, generation_arg(i + 1));
 
-        HPX_TEST_EQ(value, overall_result.get());
+        HPX_TEST_EQ(
+            static_cast<bool>(value), static_cast<bool>(overall_result.get()));
```
hkaiser, line 50: Why is this change necessary?

libs/core/execution/include/hpx/execution/executors/polymorphic_executor.hpp
hkaiser: Simply default constructing the `std::unique_ptr` member doesn't solve the issue. The default constructor was doing that already.

libs/full/collectives/tests/regressions/collectives_bool_5940.cpp
sleepingeight: I was getting few warnings as errors with clang-tidy locally but not on CI. I changed them due to it.

sleepingeight:
> Simply default constructing the std::unique_ptr member doesn't solve the issue. The default constructor was doing that already.

We default the pointer to nullptr, wouldn't it be sufficient. This is necessary as forward iterator and above concepts need to be default constructible.

libs/core/execution/include/hpx/execution/executors/polymorphic_executor.hpp
sleepingeight: We default the pointer to nullptr, wouldn't it be sufficient. This is necessary as forward iterator and above concepts need to be default constructible.

hkaiser:
> > Simply default constructing the std::unique_ptr member doesn't solve the issue. The default constructor was doing that already.
> 
> We default the pointer to nullptr, wouldn't it be sufficient. This is necessary as forward iterator and above concepts need to be default constructible.

Sure, but if you actually use it, then it will cause an exception because the default-constructed unique_ptr will be dereferenced.

hkaiser:
@sleepingeight I have merged #6928 now. Please rebase this PR.

sleepingeight:
> > > Simply default constructing the std::unique_ptr member doesn't solve the issue. The default constructor was doing that already.
> > 
> > 
> > We default the pointer to nullptr, wouldn't it be sufficient. This is necessary as forward iterator and above concepts need to be default constructible.
> 
> Sure, but if you actually use it, then it will cause an exception because the default-constructed unique_ptr will be dereferenced.

@hkaiser this is the same with the standard too. For example - https://godbolt.org/z/v887c19ab. I don't see any ways to make this better. Do you have any ideas in mind?

hkaiser:
> > > > Simply default constructing the std::unique_ptr member doesn't solve the issue. The default constructor was doing that already.
> > > 
> > > 
> > > We default the pointer to nullptr, wouldn't it be sufficient. This is necessary as forward iterator and above concepts need to be default constructible.
> > 
> > 
> > Sure, but if you actually use it, then it will cause an exception because the default-constructed unique_ptr will be dereferenced.
> 
> @hkaiser this is the same with the standard too. For example - https://godbolt.org/z/v887c19ab. I don't see any ways to make this better. Do you have any ideas in mind?

Ok. Not nice, but what can we do...

Also, please have a look at the inspect errors reported.

sleepingeight:
@hkaiser are the tests on macos flaky or are failing due to the changes in this PR? Also could re-review this PR. Thank you.

hkaiser:
> @hkaiser are the tests on macos flaky or are failing due to the changes in this PR? Also could re-review this PR. Thank you.

I have not seen those particular tests failing before, but I don't think the failures are related to your changes.

libs/core/algorithms/tests/unit/algorithms/util/test_merge_four.cpp
```diff
@@ -146,13 +146,13 @@ void test1()
...
+        A.push_back(static_cast<std::uint64_t>(i) * 4 + 3);
     for (std::uint32_t i = 0; i < 10; ++i)
-        B.push_back(i * 4 + 2);
+        B.push_back(static_cast<std::uint64_t>(i) * 4 + 2);
     for (std::uint32_t i = 0; i < 10; ++i)
-        C.push_back(i * 4 + 1);
+        C.push_back(static_cast<std::uint64_t>(i) * 4 + 1);
     for (std::uint32_t i = 0; i < 10; ++i)
-        D.push_back(i * 4);
+        D.push_back(static_cast<std::uint64_t>(i) * 4);
```
sleepingeight: new changes are required to fix this -
Performing an implicit widening conversion to type 'difference_type' (aka 'long') of a multiplication performed in type 'std::uint32_t' (aka 'unsigned int') [bugprone-implicit-widening-of-multiplication] (https://clang.llvm.org/extra/clang-tidy/checks/bugprone/implicit-widening-of-multiplication-result.html)

hkaiser: Can you separate those into their own PR, please?

sleepingeight: [errors.log](https://github.com/user-attachments/files/25468701/errors.log)
@hkaiser I've used [clang-tidy.py](https://clang.llvm.org/extra/doxygen/run-clang-tidy_8py_source.html) run generate the above log file. Although the errors are repetitive I think it would be better if we resolve all of them. I'll take it up in a seperate PR. Also I was thinking that it would be better if we made use of clang-tidy-diff.py in CI. I see that we use clang-tidy while building using ninja, but if we initially run clang-tidy-diff.py as a hard dependency, i.e if clang-tidy-diffs succeeds then only other builds start running. By this approach I think we could optimise the CI spending usage (I know that this can be implemented in circleCI need to check with github actions). What do you think? To implement this, the first effort would be fixing all the above errors and then having new github actions.

Also I've removed the changes in the above files, PTAL at the PR. Thank you.

sleepingeight: @hkaiser any thoughts on the above?

hkaiser: Yes, sure. Let's keep this searate, however. Does this script takes into account https://github.com/STEllAR-GROUP/hpx/blob/master/.clang-tidy?

hkaiser (approved):
LGTM, thanks! Thank you for your patience with this. Reviewing large PRs always takes a lot of time.

libs/core/algorithms/tests/unit/algorithms/util/test_merge_four.cpp
sleepingeight: Yes, there is an option to provide the config file although it checks for a .clang-tidy file in the source directory by default. [clang-tidy-diffs.py](https://clang.llvm.org/extra/doxygen/clang-tidy-diff_8py_source.html)

hkaiser:
@sleepingeight unfortunately, after merging your PR we're seeing new compilation issues, e.g. here: https://cdash.rostam.cct.lsu.edu/viewBuildError.php?buildid=38725. Those are most likely casued by the changes to the iterator_facade. Could you please have a look?

sleepingeight:
> @sleepingeight unfortunately, after merging your PR we're seeing new compilation issues, e.g. here: https://cdash.rostam.cct.lsu.edu/viewBuildError.php?buildid=38725. Those are most likely casued by the changes to the iterator_facade. Could you please have a look?

@hkaiser sorry for this, will fix them asap.

PR #6907, opened 2026-02-15
reduce lock contention in local_priority_queue_scheduler via randomized victim selection

### LINKED ISSUE #6904: reduce lock contention in local_priority_queue_scheduler via randomized victim selection
Motivation
===============================================================================

Currently, the local_priority_queue_scheduler (the default scheduler in HPX)
iterates through the list of potential victim threads starting from index 0
for every steal attempt.

This deterministic access pattern can lead to a "thundering herd" problem,
where multiple idle threads simultaneously attempt to steal from the same
few victims (typically neighboring threads). As a result:

- High contention occurs on a small subset of queues.
- Cache-line bouncing increases.
- Overall throughput may decrease under high contention.

To mitigate this behavior, we introduce a lightweight mechanism that
breaks the deterministic lock-step victim selection.

Proposed Change
===============================================================================

Instead of always starting the victim search from index 0 of victim_threads_,
we introduce a per-thread atomic offset:

    victim_offset_[thread_id]

On each steal attempt, the thread performs:

    std::size_t offset =
        victim_offset_[thread_id].data_.fetch_add(1, std::memory_order_relaxed);

The starting victim index is then calculated as:

    std::size_t start = (index + offset) % victim_count;

This creates a round-robin style rotation of victim selection.

Behavioral Impact
===============================================================================

- Distributes steal attempts more evenly across victims.
- Reduces contention on neighboring queues.
- Prevents deterministic synchronization of steal attempts.
- Maintains minimal runtime overhead.
- Preserves existing stealing logic and semantics.

Implementation Details
===============================================================================

1. Added the following member to local_priority_queue_scheduler:

    std::vector<
        util::cache_line_data<std::atomic<std::size_t>>
    > victim_offset_;

2. Initialized victim_offset_ in the scheduler constructor.

3. Updated the following functions to use the rotating offset:

    - attempt_stealing()
    - attempt_stealing_pending()

Each function now:
    - Performs a relaxed fetch_add(1) on victim_offset_[thread_id]
    - Computes a rotated starting index
    - Iterates through victims starting from that offset

Design Rationale
===============================================================================

This approach was use as a lightweight alternative to full randomization or more complex heuristics.

It avoids:
- Costly RNG calls
- Additional synchronization
- Complex scheduling heuristics

arpittkhandelwal:
Fixes #6904

Summary
===============================================================================

This PR implements a randomized victim selection strategy for the
local_priority_queue_scheduler to reduce lock contention and mitigate
the "thundering herd" problem, where multiple thief threads repeatedly
target the same victim threads.

Changes
===============================================================================

- Added:

    std::vector<
        util::cache_line_data<std::atomic<std::size_t>>
    > victim_offset_;

  to local_priority_queue_scheduler to maintain a per-thread rolling offset.

- Updated the constructor to initialize the victim_offset_ vector.

- Modified:

    - attempt_stealing()
    - attempt_stealing_pending()

  to perform a relaxed fetch_add on the corresponding thread’s offset.
  This ensures that each steal attempt (and each thief) begins traversing
  the victim_threads_ list from a different, rotating index.

Rationale
===============================================================================

As discussed in #6904, the default deterministic victim selection strategy
(always starting at index 0) can lead to excessive contention when many
threads are idle and attempt to steal work simultaneously.

This randomized rotation approach, suggested by @hkaiser, breaks the
deterministic lock-step behavior while introducing minimal runtime
overhead (a single atomic increment per steal attempt).

### DIFF
--- a/libs/core/schedulers/include/hpx/schedulers/local_priority_queue_scheduler.hpp
+++ b/libs/core/schedulers/include/hpx/schedulers/local_priority_queue_scheduler.hpp
@@ -672,73 +672,60 @@ namespace hpx::threads::policies {
             [[maybe_unused]] thread_queue_type* this_high_priority_queue,
             [[maybe_unused]] thread_queue_type* this_queue)
         {
-            thread_queue_type* q;
-            if (num_thread < num_high_priority_queues_)
-            {
-                std::size_t const offset =
-                    victims_[num_thread].data_.victim_offset_++;
-                std::vector<std::size_t> const& victim_threads =
-                    victims_[num_thread].data_.victim_threads_;
-                std::size_t const victim_count = victim_threads.size();
+            std::size_t const offset =
+                victims_[num_thread].data_.victim_offset_++;
+            std::vector<std::size_t> const& victim_threads =
+                victims_[num_thread].data_.victim_threads_;
+            std::size_t const victim_count = victim_threads.size();
 
-                for (std::size_t i = 0; i < victim_count; ++i)
+            if (victim_count == 0)
+                return false;
+
+            std::size_t const start_idx = offset % victim_count;
+
+            auto steal = [&](std::size_t idx) -> bool {
+                HPX_ASSERT(idx != num_thread);
+                thread_queue_type* q = nullptr;
+
+                if (num_thread < num_high_priority_queues_ &&
+                    idx < num_high_priority_queues_)
                 {
-                    std::size_t const idx =
-                        victim_threads[(i + offset) % victim_count];
-
-                    HPX_ASSERT(idx != num_thread);
-
-                    if (idx < num_high_priority_queues_)
-                    {
-                        q = high_priority_queues_[idx].data_;
-                        if (q->get_next_thread(thrd, true, true))
-                        {
-#ifdef HPX_HAVE_THREAD_STEALING_COUNTS
-                            q->increment_num_stolen_from_pending();
-                            this_high_priority_queue
-                                ->increment_num_stolen_to_pending();
-#endif
-                            return true;
-                        }
-                    }
-
-                    q = queues_[idx].data_;
+                    q = high_priority_queues_[idx].data_;
                     if (q->get_next_thread(thrd, true, true))
                     {
 #ifdef HPX_HAVE_THREAD_STEALING_COUNTS
                         q->increment_num_stolen_from_pending();
-                        this_queue->increment_num_stolen_to_pending();
+                        this_high_priority_queue
+                            ->increment_num_stolen_to_pending();
 #endif
                         return true;
                     }
                 }
-            }
-            else
-            {
-                std::size_t const offset =
-                    victims_[num_thread].data_.victim_offset_++;
-                std::vector<std::size_t> const& victim_threads =
-                    victims_[num_thread].data_.victim_threads_;
-                std::size_t const victim_count = victim_threads.size();
 
-                for (std::size_t i = 0; i < victim_count; ++i)
+                q = queues_[idx].data_;
+                if (q->get_next_thread(thrd, true, true))
                 {
-                    std::size_t const idx =
-                        victim_threads[(i + offset) % victim_count];
-
-                    HPX_ASSERT(idx != num_thread);
-
-                    q = queues_[idx].data_;
-                    if (q->get_next_thread(thrd, true, true))
-                    {
 #ifdef HPX_HAVE_THREAD_STEALING_COUNTS
-                        q->increment_num_stolen_from_pending();
-                        this_queue->increment_num_stolen_to_pending();
+                    q->increment_num_stolen_from_pending();
+                    this_queue->increment_num_stolen_to_pending();
 #endif
-                        return true;
-                    }
+                    return true;
                 }
+                return false;
+            };
+
+            for (std::size_t i = start_idx; i < victim_count; ++i)
+            {
+                if (steal(victim_threads[i]))
+                    return true;
             }
+
+            for (std::size_t i = 0; i < start_idx; ++i)
+            {
+                if (steal(victim_threads[i]))
+                    return true;
+            }
+
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


libs/core/schedulers/include/hpx/schedulers/local_priority_queue_scheduler.hpp
```diff
@@ -1701,6 +1742,8 @@ namespace hpx::threads::policies {
             high_priority_queues_;
         std::vector<util::cache_line_data<std::vector<std::size_t>>>
             victim_threads_;
+        std::vector<util::cache_line_data<std::atomic<std::size_t>>>
+            victim_offset_;
```
hkaiser: As said in #6904, I don't hink `victim_offset_` needs to be atomic. It could also live in the same cache line as `victim_threads_`
```suggestion
        struct victims {
            std::vector<std::size_t> victim_threads_;
            std::size_t victim_offset_;
        };
        std::vector<util::cache_line_data<victims>> victims_;
```

hkaiser:
@arpittkhandelwal very nice, thanks!

arpittkhandelwal:
@hkaiser  I’ve updated the implementation accordingly:

- Removed the std::atomic from victim_offset_.
- Moved victim_offset_ into the same cache line container as victim_threads_, so they now live together.
- Adjusted the increment logic to use a plain std::size_t since the offset is only accessed by the owning thread and does not require atomic synchronization.

hkaiser:
@arpittkhandelwal Do you see a performance difference compared to the baseline?

arpittkhandelwal:
> @arpittkhandelwal Do you see a performance difference compared to the baseline?

@hkaiser

I have benchmarked the feature/work-stealing-optimization branch against master to evaluate the performance impact of the randomized victim selection under high contention.

Benchmark Methodology
----------------------

I created a custom benchmark (benchmark_stealing) that:

- Spawns 1,000,000 tasks on a single thread
- Forces 8 worker threads to aggressively steal work
- Simulates a "thundering herd" scenario where victim selection strategy becomes critical

This setup intentionally stresses the stealing path to measure contention behavior.


Results
-------

Metric              Master (Baseline)   Feature (Optimized)   Improvement
---------------------------------------------------------------------------
Throughput          ~653,000 tasks/s    ~764,000 tasks/s      +17.0%
Execution Time      ~1.53 s             ~1.31 s               -14.4%


Conclusion
----------

The optimization demonstrates a clear performance improvement under
high-contention conditions:

- ~17% increase in throughput
- ~14% reduction in execution time

hkaiser:
Wow, I didn't expect for the change to have that big of an impact! Well done. 

Could you add your benchmark to the repository, please?

arpittkhandelwal:
> Wow, I didn't expect for the change to have that big of an impact! Well done.
> 
> Could you add your benchmark to the repository, please?

Thank you very much, sir! I’m really glad the impact turned out to be meaningful.

Yes, I will add the benchmark to the repository.

I’ll prepare it in a clean and configurable form and push it shortly.

Thanks again for the encouragement and review!

arpittkhandelwal:
@hkaiser Sir added the benchmark to the repo

libs/core/schedulers/include/hpx/schedulers/local_priority_queue_scheduler.hpp
```diff
@@ -675,8 +675,17 @@ namespace hpx::threads::policies {
             thread_queue_type* q;
             if (num_thread < num_high_priority_queues_)
             {
-                for (std::size_t idx : victim_threads_[num_thread].data_)
+                std::size_t const offset =
+                    victims_[num_thread].data_.victim_offset_++;
+                std::vector<std::size_t> const& victim_threads =
+                    victims_[num_thread].data_.victim_threads_;
+                std::size_t const victim_count = victim_threads.size();
```
hkaiser: This is needed only once (it's the same for both branches of the conditional).

libs/core/schedulers/include/hpx/schedulers/local_priority_queue_scheduler.hpp
```diff
@@ -675,8 +675,17 @@ namespace hpx::threads::policies {
...
+
+                for (std::size_t i = 0; i < victim_count; ++i)
```
hkaiser: Instead of having to do a division operation (the '%', which is unneededly expensive) on each iteration, this function could be refactored to move the loop body into a local lambda and run two loops:
```
    auto body = [&](std::size_t i) {...}
    for (std::size_t i = offset; i < victim_count; ++i)
        body(i);
    for (std::size_t i = 0; i < offset; ++i)
        body(i);
```

hkaiser:
May I suggest that we first try to resolve the current performance regression before merging your change? Only then we will be able to assess its impact.

arpittkhandelwal:
@hkaiser  Sir 
I’ve refactored the implementation as suggested to remove the redundant logic in `attempt_stealing` and `attempt_stealing_pending`. I also applied the split-loop optimization (using a lambda) so we avoid the modulo operator inside the hot loop.

The impact on the `benchmark_stealing` test is quite significant:

- Baseline (master): ~653,000 tasks/s  
- Previous version (modulo loop): ~764,000 tasks/s  
- New version (split loop): ~1,000,000 tasks/s  

That’s roughly a +53% improvement compared to master.

Locally, the performance boost is very noticeable with these latest changes.

arpittkhandelwal:
> May I suggest that we first try to resolve the current performance regression befoe merging your change? Only then we will be able to assess its impact.

Could you please clarify sir which performance regression you’re referring to?  
Please let me know which case I should investigate.

hkaiser:
> > May I suggest that we first try to resolve the current performance regression before merging your change? Only then we will be able to assess its impact.
> 
> Could you please clarify sir which performance regression you’re referring to? Please let me know which case I should investigate.

@arpittkhandelwal It was you we talked on Discord about bisecting recent commits, didn't we?

arpittkhandelwal:
> > > May I suggest that we first try to resolve the current performance regression before merging your change? Only then we will be able to assess its impact.
> > 
> > 
> > Could you please clarify sir which performance regression you’re referring to? Please let me know which case I should investigate.
> 
> @arpittkhandelwal It was you we talked on Discord about bisecting recent commits, didn't we?

@hkaiser  Yes sir,we discussed this on Discord.  
Understood. I’ll focus on resolving the performance regression first and start working on the bisect.

hkaiser:
> > > > May I suggest that we first try to resolve the current performance regression before merging your change? Only then we will be able to assess its impact.
> > > 
> > > 
> > > Could you please clarify sir which performance regression you’re referring to? Please let me know which case I should investigate.
> > 
> > 
> > @arpittkhandelwal It was you we talked on Discord about bisecting recent commits, didn't we?
> 
> @hkaiser Yes sir,we discussed this on Discord. Understood. I’ll focus on resolving the performance regression first and start working on the bisect.

The performance regression has been fixed on master now. Could you please rebase this PR to see whether it has any effect on our performance CI?

biddisco:
I'd be interested in seeing a comparison between this version and the "original" which has migrated to pika some years back https://github.com/pika-org/pika/blob/main/libs/pika/schedulers/include/pika/schedulers/shared_priority_queue_scheduler.hpp - in that veersion, if I recall correctly, the stealing doesn't start from zero, but the current thread ID + 1 - so that it rotates around from the current thread - but it also depends somewhat on numa domain.

It's a pity these code bases have drifted apart so far as it makes it very hard tom compare them now other than by running the overheads tests. (and many other changes will make those results less meaningful since the removal of some other stuff in pika may have as much impact)

hkaiser:
> I'd be interested in seeing a comparison between this version and the "original" which has migrated to pika some years back https://github.com/pika-org/pika/blob/main/libs/pika/schedulers/include/pika/schedulers/shared_priority_queue_scheduler.hpp - in that veersion, if I recall correctly, the stealing doesn't start from zero, but the current thread ID + 1 - so that it rotates around from the current thread - but it also depends somewhat on numa domain.
> 
> It's a pity these code bases have drifted apart so far as it makes it very hard tom compare them now other than by running the overheads tests. (and many other changes will make those results less meaningful since the removal of some other stuff in pika may have as much impact)

Hey John! Nice to hear from you!

hkaiser:
The remaining issues are CI related.

PR #6911, opened 2026-02-16
Fix split fast-path to preserve completion scheduler semantics

### LINKED ISSUE #6905: Split fast path ignores receiver’s scheduler leading to mismatched completion thread
**Problem**
When split replays a cached result on the fast path (`predecessor_done == true`), it calls `set_value` inline on the thread that calls `add_continuation()`. This ignores the receiver’s completion scheduler returned by `get_completion_scheduler<set_value_t>`(get_env(receiver)).

There is a TODO in [split.hpp#L349](https://github.com/STEllAR-GROUP/hpx/blob/ca5f46db81fab10ebbeb0fa15640b1f60e2ec482/libs/core/execution/include/hpx/execution/algorithms/split.hpp#L349) around the fast-path that mentions preserving the scheduler, but the current implementation does not propagate it.

**Observed behavior from a demo file**
```
shlokjain@Shloks-MacBook-Pro-2 hpx %  /Users/shlokjain/hpx/split_fastpath_scheduler_violation_demo
=== split fast-path scheduler mismatch (HPX split only) ===

Predecessor completed on thread 0x1eecea0c0 with value=42

UI scheduler thread id : 0x16d3cb000
Main thread id          : 0x1eecea0c0

Connecting split sender from MAIN thread to a receiver that expects the UI scheduler...

set_value called with value=42
  expected scheduler(UI) thread id: 0x16d3cb000
  actual completion thread id     : 0x1eecea0c0
  match? NO

Demo finished.
```

**Reproduce**
[Minimal standalone demo](https://gist.github.com/shlokjain2031/cd8fd6298be71c2a9fc8d5f2e09f8daf)

shlokjain2031:
Fixes [#6905](https://github.com/STEllAR-GROUP/hpx/issues/6905)

## Proposed Changes

- Introduces one replay lambda that is used in both:
  - the fast-path (predecessor already done), and
  - deferred continuations (stored and invoked later).

- The replay lambda first inspects the stored completion variant to determine which signal will be emitted:
  - set_value
  - set_error
  - set_stopped
- Before invoking the completion inline, it queries the receiver’s environment for the corresponding completion scheduler.
- If a valid scheduler is found and it supports schedule(), the completion is dispatched through that scheduler using:
  - schedule(scheduler) to obtain a sender
  - then(...) to attach the inline completion invocation
  - start_detached(...) to launch the scheduled sender
- If no suitable scheduler is available, the code falls back to invoking the completion inline, preserving the original behavior.
- This ensures consistent, scheduler-aware behavior in both fast-path and deferred execution paths.

Previously, if predecessor_done was true (fast-path), the code directly called set_value, set_error, or set_stopped inline on the downstream receiver. That inline invocation ignored any completion scheduler provided through the receiver’s environment. Ignoring the completion scheduler could break expected execution semantics, since the receiver may require completions to run on a specific scheduler.

## Checklist

### DIFF
--- a/libs/core/execution/include/hpx/execution/algorithms/split.hpp
+++ b/libs/core/execution/include/hpx/execution/algorithms/split.hpp
@@ -18,6 +18,8 @@
 #include <hpx/execution/algorithms/detail/partial_algorithm.hpp>
 #include <hpx/execution/algorithms/detail/single_result.hpp>
 #include <hpx/execution/algorithms/run_loop.hpp>
+#include <hpx/execution/algorithms/start_detached.hpp>
+#include <hpx/execution/algorithms/then.hpp>
 #include <hpx/modules/allocator_support.hpp>
 #include <hpx/modules/concepts.hpp>
 #include <hpx/modules/datastructures.hpp>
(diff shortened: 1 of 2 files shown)


hkaiser:
@isidorostsa would you have the time to review this PR, please?

hkaiser:
@isidorostsa ping?

isidorostsa:
Hey @shlokjain2031, thanks for the pr! 

We are in the process of deprecating most of hpx's native senders, and replacing the ones where hpx provides an advantage with cpo specializations on the completion scheduler.

Do you think there is a significant advantage of this implementation over the one found in stdexec?

hkaiser:

hkaiser:
@shlokjain2031 Please have a look at the CI failures.

hkaiser:
@shlokjain2031 ping?

hkaiser:
@shlokjain2031 Closing this as obsolete as we have removed our own s&r implementations.

PR #6920, opened 2026-02-16
Fixing various compilation regressions

hkaiser:
- flyby: rename test
- flyby: adding missing test targets


hkaiser:
This can be merged now, all remaining issues are unrelated and will be fixed separately.

PR #6924, opened 2026-02-17
shared_mutex : add ignore_while_checking around upgrade_cond notify in unlock_shared()

Momensamir12:
Fixes #

## Proposed Changes
- Add `ignore_while_checking` guard around `upgrade_cond.notify_one_no_unlock(lk)` in `unlock_shared()` to suppress HPX lock verification during the notify call which might yield the thread and cause ```hpx::util::verify_no_locks()``` exception in debug builds.

## Background context
- ```shared_mutex_6854``` regression test fails in debug builds with ```hpx::util::verify_no_locks()``` failing


Momensamir12:
Can we check if the test is passing in debug now ?

hkaiser:
> Can we check if the test is passing in debug now ?

The only way to check currently is to merge as the related CI instances kick in only for PRs from the main repository. Let's go ahead and do that.

PR #6926, opened 2026-02-18
Disable async_cuda test as nvcc fails

hkaiser:
This will fix the CUDA CIs


hkaiser:
If nothing else, this fixes what it was supposed to fix: async_cuda,synchronize doesn't fail anymore.

PR #6927, opened 2026-02-18
Attempt to work around Github API limitations

hkaiser:


hkaiser:
This seems to do the trick - merging...

PR #6928, opened 2026-02-19
Replace scan implementation with explicit functions to avoid triggering concept checking

hkaiser:
@sleepingeight could you please check whether this resolves the scan-collectives when used with `bool`

### DIFF
--- a/libs/full/collectives/include/hpx/collectives/exclusive_scan.hpp
+++ b/libs/full/collectives/include/hpx/collectives/exclusive_scan.hpp
@@ -293,7 +293,7 @@ namespace hpx::collectives::detail {
 
     template <typename T, typename InIter, typename Sent, typename OutIter,
         typename Op>
-    static constexpr void exclusive_scan(
+    constexpr void exclusive_scan(
         InIter first, Sent last, OutIter dest, Op&& op)
     {
         // the first value given goes to the second destination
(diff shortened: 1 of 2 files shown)


libs/full/collectives/include/hpx/collectives/exclusive_scan.hpp
```diff
@@ -293,17 +293,31 @@ namespace hpx::collectives::detail {
...
+        typename Op>
+    constexpr void exclusive_scan_init(
+        InIter first, Sent last, OutIter dest, T init, Op&& op)
+    {
+        T temp = init;
+        for (/* */; first != last; (void) ++first, ++dest)
+        {
+            init = HPX_INVOKE(op, HPX_MOVE(init), *first);
+            *dest = HPX_MOVE(temp);
+            temp = HPX_MOVE(init);
```
sleepingeight:
```suggestion
            temp = init;
```

Otherwise results in a `bugprone-use-after-move, -warnings-as-errors` errors.

libs/full/collectives/include/hpx/collectives/inclusive_scan.hpp
```diff
@@ -234,6 +234,26 @@ namespace hpx { namespace collectives {
...
+        InIter first, Sent last, OutIter dest, Op&& op)
+    {
+        if (first != last)
+        {
+            T init = *first++;
+            *dest++ = HPX_MOVE(init);
+            for (/* */; first != last; (void) ++first, ++dest)
+            {
+                init = HPX_INVOKE(op, HPX_MOVE(init), *first);
+                *dest = HPX_MOVE(init);
```
sleepingeight:
```suggestion
            *dest++ = init;
            for (/* */; first != last; (void) ++first, ++dest)
            {
                init = HPX_INVOKE(op, HPX_MOVE(init), *first);
                *dest = init;
```

Same reason as above.

libs/full/collectives/include/hpx/collectives/exclusive_scan.hpp
hkaiser: Thanks, that's fixed now.

hkaiser:
This looks ok now.

PR #6929, opened 2026-02-19
Do bisection to diagnose performance regression 

hkaiser:


hkaiser:
Jenkins always tests merge commits against master, which means that we can't use PRs to bisect commits. Closing...

PR #6938, opened 2026-02-24
fix dangling reference errors in iterator tests

sleepingeight:


libs/core/iterator_support/tests/regressions/custom_iterator_category_6640.cpp
```diff
@@ -38,7 +38,7 @@ class test_iterator
         val += n;
     }
 
-    [[nodiscard]] std::int64_t const& dereference() const
+    [[nodiscard]] std::int64_t dereference() const
```
hkaiser: While this may fix the test error, it will not do the right thing. A random-access iterator is always expected to return a true reference when being dereferenced.

sleepingeight: I don't think that is the case, I confirmed it by running - 
```
static_assert(std::random_access_iterator<test_iterator>,
    "test_iterator is not a random access iterator");
```
which passed. This is because random access iterator requirements have been changed from c++20. Quoting from [cppreference](https://en.cppreference.com/w/cpp/iterator/random_access_iterator.html) - 
> Unlike the [LegacyRandomAccessIterator](https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator.html) requirements, the random_access_iterator concept does not require dereference to return an lvalue.

We only need to satisfy `iter_reference_t<I>` which is satisfied by different types such as T&, T, proxies. 

Another reason for this change is that the iterator does not refer to actual storage but instead generates values on demand, so returning a reference from dereference would not be meaningful.

hkaiser: The issue is not the test_iterator, it just exposes a problem in the iterator facade for real random-access iterators. 

This still implies that you can use `iter_reference_t<I>` as something referencing the sequence elements you're iterating over (otherwise you wouldn't be able to use it with algorithms like `sort` that swap elements in the sequence, possibly using proxies - as you rightly noticed).

IOW, the test should pass even if dereferencing the iterator yields a true reference.

libs/core/iterator_support/include/hpx/iterator_support/iterator_facade.hpp
```diff
@@ -281,7 +281,8 @@ namespace hpx::util {
         struct use_operator_brackets_proxy
           : std::integral_constant<bool,
                 !(std::is_copy_constructible_v<Value> &&
-                    std::is_const_v<Value>)>
+                    std::is_const_v<Value>) ||
+                    std::is_lvalue_reference_v<typename Iterator::reference>>
```
hkaiser: I understand what you're trying to do, however, I believe that this change would break the decision logic of when to return a proxy from the iterator's `operator[]()`. 

After thinking some more about this, I agree with your argumentation that we have to fix the `test_iterator` (iterator_facade is correct). However I'd suggest not to change its `iter_reference_t`, but instead to add an embedded type `using use_brackets_proxy = std::true_type;` that forces the facade to inject the proxy for the return type of `operator[]()`.

Unrelated, but while you're at it; we don't need the `operator_brackets_result` and `make_operator_brackets_result` for the iterator facade anymore (replaced by the `decltype(auto)` return type calculation for the `operator[]()` you introduced). Could you please remove those symbols altogether?

hkaiser: Also, thank you for your patience with this issue! I appreciate your involvement and discussions.

sleepingeight: Thank you, will update the PR according to your suggestions.

hkaiser:
@sleepingeight Please address the reported inspect issue (missing `#include <type_traits>` in the test you touched).

hkaiser:
BTW, these other tests that fail now require a similar fix: https://cdash.rostam.cct.lsu.edu/viewTest.php?onlyfailed&buildid=38929 (add the embedded `using` declaration to the mentioned iterators).

libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size_iterator.hpp
```diff
@@ -393,6 +393,8 @@ namespace hpx::parallel::util::detail {
             }
         }
 
+        using use_brackets_proxy = std::true_type;
+
```
hkaiser, line 399: The same is probably needed for the `chunk_size_iterator` above

PR #6939, opened 2026-02-24
Add zero-thread validation in fork_join_executor PU-mask constructor

Sanchit2662:
## Summary

This PR fixes a missing lower-bound validation in the PU-mask constructor of
`hpx::execution::experimental::fork_join_executor::shared_data`
(`libs/core/executors/include/hpx/executors/fork_join_executor.hpp`).

When constructing a `fork_join_executor` with an explicit PU mask, the constructor computes the number of threads from the mask. While the existing validation checks for a null thread pool and an upper bound violation, it does not check whether the computed thread count is zero.

If an empty PU mask is passed (no processing units selected), the computed thread count becomes `0`. This propagates into `init_local_work_queue()`, where work partitioning divides by the number of threads, resulting in a division-by-zero and undefined behavior during bulk execution.

This PR ensures that invalid configurations with zero threads are rejected at construction time. It also fixes a typo in the associated error message.

---

## Fix

The validation condition in the PU-mask constructor has been extended to explicitly reject `num_threads_ == 0`.

### Before

```cpp
if (pool_ == nullptr ||
    num_threads_ > pool_->get_os_thread_count())
```

### After

```cpp
if (pool_ == nullptr || num_threads_ == 0 ||
    num_threads_ > pool_->get_os_thread_count())
```

---

## Checklist

* [x] I have fixed a bug and have added a regression test.

### DIFF
--- a/libs/core/executors/include/hpx/executors/fork_join_executor.hpp
+++ b/libs/core/executors/include/hpx/executors/fork_join_executor.hpp
@@ -573,11 +573,11 @@ namespace hpx::execution::experimental {
               , region_data_(get_region_data_size(num_threads_, pool_))
             {
                 HPX_ASSERT(pool_);
-                if (pool_ == nullptr ||
+                if (pool_ == nullptr || num_threads_ == 0 ||
                     num_threads_ > pool_->get_os_thread_count())
                 {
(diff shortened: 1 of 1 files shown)


hkaiser:
Thanks for working on this! Much appreciated!

PR #6943, opened 2026-02-26
Fix thread safety bugs in concurrent containers and a few algorithm issues

aneek22112007-tech:
While reviewing the recently merged concurrent data structures and algorithm 
changes, I found a few bugs worth fixing.

**Concurrent containers (vector, unordered_map, unordered_set)**

The move constructors and move assignment operators weren't locking the 
source object's mutex before moving from it. This means if another thread 
is reading from the source while a move is happening, it's a data race.
The copy constructor already did this correctly — move just missed it.
Fixed move constructors to lock `other.mutex_` before moving, and move 
assignment to use `std::lock()` on both mutexes (same pattern as copy assignment).

**concurrent_accessor**

There were two related issues here:

1. `operator reference_type()` (implicit conversion to `T&`) let you do 
   `int& ref = vec[0];` which looks fine but the lock is released the 
   moment the accessor is destroyed — so `ref` is now a dangling reference 
   into an unlocked container. Removed the implicit conversion to prevent this.

2. `operator=(T const&)` was confusing because it looks like you're 
   reassigning the accessor, but it actually sets the contained value. 
   Renamed it to [set()](cci:1://file:///Users/aneek/hpx/libs/core/concurrency/include/hpx/concurrency/detail/concurrent_accessor.hpp:83:8-88:9) to make the intent clear, and added a [set(T&&)](cci:1://file:///Users/aneek/hpx/libs/core/concurrency/include/hpx/concurrency/detail/concurrent_accessor.hpp:83:8-88:9) 
   overload too. Updated the test accordingly.

**sequential_find**

The single-range overload of `sequential_find_t` was hardcoding 
`sequenced_policy` instead of using `ExPolicy`. This means any SIMD/
vectorization hints from the execution policy were silently ignored. 
All the [find_if](cci:1://file:///Users/aneek/hpx/libs/core/algorithms/include/hpx/parallel/algorithms/detail/find.hpp:143:34-150:5) variants used `ExPolicy` correctly — [find](cci:1://file:///Users/aneek/hpx/libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp:269:8-279:9) just missed it.

**fold_left / fold_right**

These were requiring `T` (the initial value) to exactly match 
`iter_value_t<I>`, which is stricter than what C++23 specifies. 
The standard only needs `is_indirectly_binary_foldable` to hold. 
Dropped the `is_same_v` constraint so you can, for example, fold a 
range of `int` into a `std::string` accumulator.


libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -81,8 +81,9 @@ namespace hpx::concurrent {
         }
 
         concurrent_unordered_map(concurrent_unordered_map&& other) noexcept
-          : map_(HPX_MOVE(other.map_))
         {
+            std::lock_guard<hpx::util::spinlock> lock(other.mutex_);
+            map_ = HPX_MOVE(other.map_);
```
hkaiser, line 86: I think that the lock is not needed here as the right hand side is an rvalue, which by definition means that nobody else is doing anything with it (e.g., it's a temparary). Could you construct a viable use case that would make this lock necessary?

libs/core/algorithms/include/hpx/parallel/container_algorithms/fold.hpp
```diff
@@ -282,7 +282,6 @@ namespace hpx::ranges {
         // clang-format off
         requires (
             std::input_iterator<InIter> &&
-            std::is_same_v<T, hpx::traits::iter_value_t<InIter>> &&
```
hkaiser, line 285: I agree that this requirement is too strong. The standard says that `hpx::traits::iter_value_t<InIter>` should be convertible to `T`, not that they should be the same. Could you fix that, please (instead removing the reuqirement)?

libs/core/concurrency/include/hpx/concurrency/detail/concurrent_accessor.hpp
```diff
@@ -62,24 +62,30 @@ namespace hpx::concurrent::detail {
...
+        // lifetime, as the lock is released when the accessor is destroyed.
         reference_type get() const
         {
             validate();
             return *value_;
         }
 
-        // Assignment operator only for non-const accessor
-        concurrent_accessor& operator=(T const& v)
+        // Sets the value of the contained element (non-const accessor only).
```
hkaiser, line 82: I agree that the `operator reference()` is dangerous as it can lead to the issue you described. We should be able to define a operator that returns a either a const reference or a rvlaue reference as for those the accessor instance would be kept alive (at least in the immediate expression context). We might have to mark those operators in specific ways (`operator reference() && {...}`? I'm not sure... - let's think avout this.)

hkaiser (commented):
Thanks for looking into this! The issues you raised are valid and need additional attention.

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
aneek22112007-tech: Ah yeah, that makes total sense — if it's an rvalue nobody else could possibly be holding a reference to it anyway. My bad, removed the lock. Thanks!

libs/core/algorithms/include/hpx/parallel/container_algorithms/fold.hpp
aneek22112007-tech: Fair point! Swapped is_same_v for std::convertible_to<hpx::traits::iter_value_t<InIter>, T> across the affected overloads. Feels much more in line with what the standard actually intends.

libs/core/concurrency/include/hpx/concurrency/detail/concurrent_accessor.hpp
aneek22112007-tech: I like the ref-qualifier idea! Went with:

cpp
operator reference_type() &&          // ok — rvalue, lock stays alive for the expression
operator reference_type() const& = delete;  // compile error for T& ref = vec[0];
So things like 

f(vec[0])
 still just work, but accidentally storing the reference becomes a hard error. Let me know if you had something different in mind!

hkaiser: Well, that's what I had in mind, just that `operator reference_type() &&` will be invocable on an rvalue accessor only, which would definitely break things. Perhaps something like `operator T&&()` or `operator T const&() const` instead? As I've said. this needs thorough consideration and experimentation, ATM, I'm not sure what the best way would be.

hkaiser: Please construct some tests that we can use to guide decisions.

libs/core/concurrency/include/hpx/concurrency/detail/concurrent_accessor.hpp
```diff
@@ -62,24 +62,45 @@ namespace hpx::concurrent::detail {
...
+            return *value_;
         }
 
-        reference_type get() const
+        // Only allow implicit conversion to T& when the accessor is a
+        // temporary (rvalue). This ensures the lock remains held for the
+        // entire expression (e.g. f(vec[0]) is safe). Storing the reference
+        // via an lvalue accessor (e.g. T& r = vec[0]) is explicitly deleted
+        // to prevent use-after-unlock data races.
+        operator reference_type() && noexcept(false)
```
hkaiser: As said, this is wrong as it will still produce dangling references.

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -105,7 +105,11 @@ namespace hpx::concurrent {
         {
             if (this != &other)
             {
-                std::lock_guard<hpx::util::spinlock> lock(mutex_);
+                std::lock(mutex_, other.mutex_);
+                std::lock_guard<hpx::util::spinlock> lock(
+                    mutex_, std::adopt_lock);
+                std::lock_guard<hpx::util::spinlock> other_lock(
+                    other.mutex_, std::adopt_lock);
```
hkaiser: Please revert this change for the move assignment operator as well (same rationale as for the move constructor).

libs/core/concurrency/tests/unit/concurrent_data_structures.cpp
```diff
@@ -43,19 +43,17 @@ void test_concurrent_vector()
...
     {
-        v[0] = 999;
-        HPX_TEST_EQ(static_cast<int>(v[0]), 999);
-        HPX_TEST(v[0] == 999);
-
-        // Test operator T&
-        int& ref = v[0];
-        HPX_TEST_EQ(ref, 999);
-        ref = 888;
-        HPX_TEST_EQ(static_cast<int>(v[0]), 888);
```
hkaiser, line 54: Let's try to make this test work again.

libs/core/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
```diff
@@ -57,39 +56,17 @@ namespace hpx::parallel::detail {
...
-    HPX_CXX_CORE_EXPORT template <typename ExPolicy, typename FwdIter,
-        typename Token, typename T, typename Proj>
-    constexpr void sequential_find(std::size_t base_idx, FwdIter part_begin,
-        std::size_t part_count, Token& tok, T const& val, Proj&& proj)
-    {
-        return sequential_find_t<ExPolicy>{}(base_idx, part_begin, part_count,
-            tok, val, HPX_FORWARD(Proj, proj));
-    }
-#endif
+    inline constexpr sequential_find_t sequential_find = sequential_find_t{};
```
hkaiser: This change will break the customization point that relies on dispatching on the execution policy. As an example look at https://github.com/STEllAR-GROUP/hpx/blob/b67a10fed3257287622a01562566f88548e20335/libs/core/algorithms/include/hpx/parallel/datapar/find.hpp#L61-L80 where we use this customization point to dispatch to the SIMD algorithm implementations.

libs/core/concurrency/tests/performance/concurrent_structure_performance.cpp
```diff
@@ -155,11 +155,19 @@ int hpx_main(hpx::program_options::variables_map& vm)
     std::uint64_t num_ops =
         vm["ops"].as<std::uint64_t>();    // Total ops per test
 
+    auto access_value = [](auto&& accessor) -> decltype(auto) {
+        if constexpr (requires { accessor.get(); })
```
hkaiser: Under what circumstances doesn't have the accessor a `get()` member function?

libs/core/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
aneek22112007-tech: Thank you for the guidance. I have restored sequential_find_t as a template in 

detail/find.hpp
 to ensure the 

tag_invoke
 customization point works correctly for all execution policies. Additionally, I updated loop_pred to use the template's ExPolicy instead of the hardcoded sequenced_policy to fix the SIMD dispatching.

libs/core/concurrency/tests/performance/concurrent_structure_performance.cpp
aneek22112007-tech: You are correct; the requires check was redundant. I have simplified the code by removing the access_value lambda and using a direct (void)cast to discard the lookup result, as the return value is not needed for this benchmark.

hkaiser: Please use `[[maybe_unused]]` instead of the `(void)` cast.

aneek22112007-tech: Thanks for the tip! I'm still learning some of these modern C++ conventions, so I really appreciate you pointing that out. I've just pushed an update replacing those 

(void)
 casts with [[maybe_unused]] in the benchmarks and a few other spots too. It definitely looks a lot cleaner now—thanks again for the help!

hkaiser:
@aneek22112007-tech Please address the inspect issues reported by the CIs.

libs/core/algorithms/include/hpx/parallel/algorithms/detail/equal.hpp
```diff
@@ -82,8 +82,7 @@ namespace hpx::parallel::detail {
             InIter1 first1, Sent1 last1, InIter2 first2, Sent2 last2, F&& f,
             Proj1&& proj1, Proj2&& proj2)
         {
-            for (/* */; first1 != last1 && first2 != last2;
-                (void) ++first1, ++first2)
+            for (/* */; first1 != last1 && first2 != last2; ++first1, ++first2)
```
hkaiser: What is the rationale of this change?

libs/core/algorithms/include/hpx/parallel/algorithms/detail/mismatch.hpp
```diff
@@ -90,7 +90,7 @@ namespace hpx::parallel::detail {
                 HPX_INVOKE(
                     f, HPX_INVOKE(proj1, *first1), HPX_INVOKE(proj2, *first2)))
             {
-                (void) ++first1, ++first2;
+                ++first1, ++first2;
```
hkaiser: Why did you touch this code?

libs/core/algorithms/include/hpx/parallel/algorithms/detail/search.hpp
```diff
@@ -50,7 +50,7 @@ namespace hpx::parallel::detail {
             for (;; ++first)
             {
                 FwdIter it1 = first;
-                for (FwdIter2 it2 = s_first;; (void) ++it1, ++it2)
+                for (FwdIter2 it2 = s_first;; ++it1, ++it2)
```
hkaiser: Please do not remove the `(void)` cast in these contexts. We use that deliberately to preven any possible overloaded `operator,()` to take effect.

libs/core/concurrency/include/hpx/concurrency/concurrent_vector.hpp
```diff
@@ -87,7 +84,7 @@ namespace hpx::concurrent {
         {
             if (this != &other)
             {
-                std::lock_guard<hpx::util::spinlock> lock(mutex_);
+                std::scoped_lock lock(mutex_, other.mutex_);
```
hkaiser: I don't think that the move assignment needs to lock the right hand side.

libs/core/concurrency/tests/performance/concurrent_structure_performance.cpp
```diff
@@ -222,9 +219,8 @@ int hpx_main(hpx::program_options::variables_map& vm)
                         for (std::uint64_t j = 0; j < num_ops / num_threads;
                             ++j)
                         {
-                            [[maybe_unused]] auto volatile val =
+                            [[maybe_unused]] auto&& _ =
```
hkaiser: Why did you remove the `volatile`? It's there to prevent the compiler from optimizing away the loop body.

libs/core/algorithms/include/hpx/parallel/algorithms/detail/equal.hpp
aneek22112007-tech: Hi @hkaiser, sorry about that! As a student just getting started, I honestly didn't realize the 

(void)
 cast was there to stop the overloaded comma operator from triggering. That’s really cool to learn.

I’ve gone ahead and put the 

(void)
 casts back here and in 

mismatch.hpp
 and 

search.hpp
. I also pushed up fixes for the other stuff you mentioned. Let me know if everything looks okay now!

libs/core/concurrency/tests/performance/concurrent_structure_performance.cpp
aneek22112007-tech: Oh I see! I didn't realize letting the compiler optimize the loop body away means it won't be benchmarked correctly. I've brought the volatile keyword back so it actually tests the performance properly. Thanks for pointing that out and explaining it!

libs/core/concurrency/include/hpx/concurrency/concurrent_vector.hpp
aneek22112007-tech: Ah, my bad! I didn't realize that changing this would let the compiler optimize the loop body entirely away. I've switched it back to auto volatile val so the performance test actually measures what it's supposed to. Thanks for explaining that!

libs/core/algorithms/include/hpx/parallel/algorithms/detail/search.hpp
aneek22112007-tech: Ah, that makes total sense! Since other is an rvalue being moved from, there's no need to lock its mutex. I've updated the operator to only lock the left-hand side mutex_ with a std::lock_guard. Thanks for catching this!

libs/core/algorithms/include/hpx/parallel/algorithms/detail/mismatch.hpp
aneek22112007-tech: Oops, my bad! I removed those by mistake while editing nearby lines. I've re-added the 

(void)
 cast here to match the others. Thanks for pointing that out!

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_map.hpp
```diff
@@ -105,7 +106,7 @@ namespace hpx::concurrent {
         {
             if (this != &other)
             {
-                std::lock_guard<hpx::util::spinlock> lock(mutex_);
+                std::scoped_lock lock(mutex_, other.mutex_);
```
hkaiser: As said, I don't think we need to lock the right hand side's mutex as `other` is a rvalue.

libs/core/concurrency/include/hpx/concurrency/concurrent_unordered_set.hpp
```diff
@@ -101,7 +98,7 @@ namespace hpx::concurrent {
         {
             if (this != &other)
             {
-                std::lock_guard<hpx::util::spinlock> lock(mutex_);
+                std::scoped_lock lock(mutex_, other.mutex_);
```
hkaiser: Same here, `other` is a rvalue, so by definition no race can happen.

libs/core/concurrency/include/hpx/concurrency/detail/concurrent_accessor.hpp
```diff
@@ -62,25 +62,40 @@ namespace hpx::concurrent::detail {
...
+        void set(T const& v)
             requires(!std::is_const_v<T>)
         {
             validate();
             *value_ = v;
-            return *this;
         }
+
+        void set(T&& v)
+            requires(!std::is_const_v<T>)
```
hkaiser: Why not:
```suggestion
        void set(std::remove_const_t<T>&& v)
```
?

libs/core/concurrency/tests/unit/concurrent_data_structures.cpp
```diff
@@ -43,19 +43,18 @@ void test_concurrent_vector()
...
-        v[0] = 999;
-        HPX_TEST_EQ(static_cast<int>(v[0]), 999);
-        HPX_TEST(v[0] == 999);
+        v[0].set(999);
+        HPX_TEST_EQ(static_cast<int>(v[0].get()), 999);
+        HPX_TEST(v[0].get() == 999);
 
-        // Test operator T&
+        // Test implicit conversion to reference
         int& ref = v[0];
```
hkaiser: This doesn't look correct (and I know that this code was there to begin with). `int& ref = v[0];` will not keep the accessor alive. I think this should be: `auto&& ref = v[0];`, which will keep the accessor alive.

hkaiser:
Also, please have a look at the clang-format issues reported.

hkaiser (approved):
LGTM, thanks!

PR #6945, opened 2026-02-26
1/n: conforming parallel range algos to the standard

sleepingeight:
Prototyping implementation for aligning parallel range algos to the standard.

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/container_algorithms/adjacent_difference.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/container_algorithms/adjacent_difference.hpp
@@ -353,172 +353,91 @@ namespace hpx::ranges {
       : hpx::detail::tag_parallel_algorithm<adjacent_difference_t>
     {
     private:
-        template <typename FwdIter1, typename FwdIter2, typename Sent>
-        // clang-format off
-             requires (
-                hpx::traits::is_iterator_v<FwdIter1> &&
-                hpx::traits::is_iterator_v<FwdIter2> &&
-                std::sentinel_for<Sent, FwdIter1>
-            )
-        // clang-format on
+        template <std::forward_iterator FwdIter1,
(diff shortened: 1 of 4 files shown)


sleepingeight:
@hkaiser, I think similar changes are required in the area of parallel algos (modernising the implementation). So, I was thinking to deal with the documentation seperately (maybe as a community effort to speed things up since I've seen many mistakes in the documentation). So, I'm thinking to work in this order -
1. Conform parallel range algos to the standard (I'll do this entirely)
2. Modernise parallel algos similar to the adjacent_difference (which you've done previously)
3. Update documentation overall.
(I can create issues for 2 and 3 so that work is done in parallel and is also not a big task and works very well as good first issues)

sleepingeight:
@hkaiser PTAL at this PR. Thanks!!

hkaiser:
> @hkaiser, I think similar changes are required in the area of parallel algos (modernising the implementation). So, I was thinking to deal with the documentation seperately (maybe as a community effort to speed things up since I've seen many mistakes in the documentation). So, I'm thinking to work in this order -
> 
> 1. Conform parallel range algos to the standard (I'll do this entirely)
> 2. Modernise parallel algos similar to the adjacent_difference (which you've done previously)
> 3. Update documentation overall.
>    (I can create issues for 2 and 3 so that work is done in parallel and is also not a big task and works very well as good first issues)

This sounds good. Just please remember that in order for somebody to be able to review changes, the PRs can't be too large. Thus I'd suggest to keep the change sets small and rather have a larger aount of PRs that are manageable.

hkaiser:
Also, there is still the issue described here: https://github.com/STEllAR-GROUP/hpx/issues/6718, with an attempted solution here: https://github.com/STEllAR-GROUP/hpx/pull/6843. All of this will require touching on all algorithm APIs in any case, thus I'd like to think that we first should decide how to resolve the issues before investing time to update things that might have to be changed again afterwards. Please note that I don't have a solution for this yet, so this will require a bit of experimentation.

PR #6946, opened 2026-02-26
dynamic test discovery in CI

sleepingeight:
~Instead of introducing seperate job for clang-tidy.py, trying to use the clang-tidy in the build process. Chose this option because using clang-tidy.py will increase build time by a large factor (approx 1hr to find out all failures). We could use clang-tidy-diff.py but this does not give a full picture, but only performs the checks for the changed lines.~

Found out cmake + clang-tidy was sufficient in the CI process but the only catch is that were listing targets manually which led to a few errors where we did not list some targets. The errors which were found by running clang-tidy.py could also be found through our existing infra, so made changes so that all targets are listed dynamically.

### DIFF
--- a/libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace_copy.hpp
+++ b/libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace_copy.hpp
@@ -158,7 +158,7 @@ namespace hpx::parallel {
                 {
                     segments.push_back(dispatch_async(traits2::get_id(sdest),
                         algo, policy, forced_seq(), beg, end, ldest, old_value,
-                        new_value, proj));
+                        new_value, HPX_FORWARD(Proj, proj)));
                 }
             }
             else
@@ -196,7 +196,7 @@ namespace hpx::parallel {
                 {
                     segments.push_back(dispatch_async(traits2::get_id(sdest),
                         algo, policy, forced_seq(), beg, end, ldest, old_value,
-                        new_value, proj));
+                        new_value, HPX_FORWARD(Proj, proj)));
                 }
             }
 
@@ -252,7 +252,7 @@ namespace hpx::parallel {
                         local_iterator_type2>
                         out = dispatch(traits2::get_id(sdest), algo, policy,
                             std::true_type(), beg, end, ldest, f, new_value,
-                            proj);
+                            HPX_FORWARD(Proj, proj));
                     last = traits1::compose(send, out.in);
                     dest = traits2::compose(sdest, out.out);
                 }
@@ -292,7 +292,8 @@ namespace hpx::parallel {
                 if (beg != end)
                 {
                     out = dispatch(traits2::get_id(sdest), algo, policy,
-                        std::true_type(), beg, end, ldest, f, new_value, proj);
+                        std::true_type(), beg, end, ldest, f, new_value,
+                        HPX_FORWARD(Proj, proj));
                 }
                 last = traits1::compose(send, out.in);
                 dest = traits2::compose(sdest, out.out);
@@ -341,9 +342,9 @@ namespace hpx::parallel {
                 local_iterator_type2 ldest = traits2::local(dest);
                 if (beg != end)
                 {
-                    segments.push_back(
-                        dispatch_async(traits2::get_id(sdest), algo, policy,
-                            forced_seq(), beg, end, ldest, f, new_value, proj));
+                    segments.push_back(dispatch_async(traits2::get_id(sdest),
+                        algo, policy, forced_seq(), beg, end, ldest, f,
+                        new_value, HPX_FORWARD(Proj, proj)));
                 }
             }
             else
@@ -379,9 +380,9 @@ namespace hpx::parallel {
                 ldest = traits2::begin(sdest);
                 if (beg != end)
                 {
-                    segments.push_back(
-                        dispatch_async(traits2::get_id(sdest), algo, policy,
-                            forced_seq(), beg, end, ldest, f, new_value, proj));
+                    segments.push_back(dispatch_async(traits2::get_id(sdest),
+                        algo, policy, forced_seq(), beg, end, ldest, f,
+                        new_value, HPX_FORWARD(Proj, proj)));
                 }
             }
 


.github/workflows/build-and-test.yml
```diff
@@ -57,7 +57,7 @@ jobs:
                     -DHPX_WITH_COMPILER_WARNINGS=On \
                     -DHPX_WITH_COMPILER_WARNINGS_AS_ERRORS=On \
                     -DHPX_WITH_DEPRECATION_WARNINGS=On \
-                    -DCMAKE_CXX_CLANG_TIDY=clang-tidy-20 \
+                    -DCMAKE_CXX_CLANG_TIDY="clang-tidy-20;--warnings-as-errors=*;--config-file=$GITHUB_WORKSPACE/.clang-tidy" \
```
hkaiser: Isn't the file used by default? Also that file already forces all warnings to be errors.

cmake/HPX_AddCompileTest.cmake
```diff
@@ -53,6 +53,10 @@ function(add_hpx_compile_test category name)
   )
 
   if(${name}_FAILURE_EXPECTED)
+    # Disable clang-tidy for intentional fail-to-compile targets (files starting
+    # with 'fail_'). These contain deliberately invalid code that clang-tidy
+    # cannot analyse without emitting errors.
+    set_target_properties(${test_name} PROPERTIES CXX_CLANG_TIDY "")
```
hkaiser, line 59: Wouldn't it be better to say (at least starting CMake 3.27):
```suggestion
    set_target_properties(${test_name} PROPERTIES SKIP_LINTING ON)
```

hkaiser:
@sleepingeight Any news on this?

hkaiser:
@sleepingeight ping?

sleepingeight:
@hkaiser extremely sorry to keep you waiting on this. I'm kind of stuck here since as I was unable to debug which clang-tidy+cmake+ninja was not showing up errors which were found through the script. I can add the script but it seems redundant provided we use clang-tidy in build-and-test CI. Maybe we could remove clang-tidy usage from build-and-test and only rely on the script but the script takes a good amount of time to run (need to profile the exact time). I'm not quite sure of how we could take this forward. Any suggestions?

hkaiser:
> @hkaiser extremely sorry to keep you waiting on this. I'm kind of stuck here since as I was unable to debug which clang-tidy+cmake+ninja was not showing up errors which were found through the script. I can add the script but it seems redundant provided we use clang-tidy in build-and-test CI. Maybe we could remove clang-tidy usage from build-and-test and only rely on the script but the script takes a good amount of time to run (need to profile the exact time). I'm not quite sure of how we could take this forward. Any suggestions?

What does the script do that is not covered by our clang-tidy integration?

sleepingeight:
Pretty much nothing, it just runs clang-tidy binary on all source files [[script source](https://clang.llvm.org/extra/doxygen/run-clang-tidy_8py_source.html)]. I'll try to test everything in the build_env present on CI which might help

sleepingeight:
@hkaiser after long hours of debugging, I found that the issue is not in our build :( but in the .targets files which are used to specify the test sources. The issue is that many targets such as `tests.unit.modules.cache....`, `tests.unit.modules.iostream..` are missing. These were not being compiled on CI so I wasn't seeing any errors related to them. But when ran with a script all files were being checked which provided errors fixed in a recent PR.

I'll modify this PR to include all these targets, but I strongly feel that we should not rely on manual targets but use `ctest -N` to list all targets and divide work between seperate workflows with code. Any thoughts on this?

hkaiser:
> @hkaiser after long hours of debugging, I found that the issue is not in our build :( but in the .targets files which are used to specify the test sources. The issue is that many targets such as `tests.unit.modules.cache....`, `tests.unit.modules.iostream..` are missing. These were not being compiled on CI so I wasn't seeing any errors related to them. But when ran with a script all files were being checked which provided errors fixed in a recent PR.
> 
> I'll modify this PR to include all these targets, but I strongly feel that we should not rely on manual targets but use `ctest -N` to list all targets and divide work between seperate workflows with code. Any thoughts on this?

Yes, I agree that maintaining the targets manually is not a good idea. Originally we introduced this because of limitations in CircleCI were we needed to manually tune which targets to build in a particular step in order to avoid exceeding CircleCI limits. Now that we are about to move away from CircleCI we might be able to find a more automated solution for this. Would you have any suggestions?

sleepingeight:
@hkaiser could you take a look at the CI status of the PR - https://github.com/sleepingeight/hpx/pull/2, this is the way in which the new build-and-test works. There are few tests failing. Are they flaky tests or something to be concerned about?

hkaiser:
IIUC, you have combined all tests into the same action step. I think it was beneficial to have those separated as it speeded up the overall turnaround for the CI.

Also @waleed7715 would you have the time to look into this as well?

sleepingeight:
They are in a same action setup, but the matrix job makes each partition of the tests run in parallel.

hkaiser:
> They are in a same action setup, but the matrix job makes each partition of the tests run in parallel.

Ok, makes sense. How do we disable certain tests (for the time being) in order to get a clean build until those are fixed?

sleepingeight:
I think modifying the script is the only option

hkaiser:
> I think modifying the script is the only option

What script did you have in mind? Could we have a file that explicitly lists all disabled test targets that then are being passed to the ctest invocation?

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace_copy.hpp
```diff
@@ -65,7 +65,7 @@ namespace hpx::parallel {
                         local_iterator_type2>
                         out = dispatch(traits2::get_id(sdest), algo, policy,
                             std::true_type(), beg, end, ldest, old_value,
-                            new_value, HPX_FORWARD(Proj, proj));
+                            new_value, proj);
```
hkaiser: This can stay as it is the only use of `proj` in this execution path.

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace_copy.hpp
```diff
@@ -107,7 +107,7 @@ namespace hpx::parallel {
                 {
                     out = dispatch(traits2::get_id(sdest), algo, policy,
                         std::true_type(), beg, end, ldest, old_value, new_value,
-                        HPX_FORWARD(Proj, proj));
+                        proj);
```
hkaiser: This can stay as well as it is the last use of `proj` in this execution path.

hkaiser:
> Could we have a file that explicitly lists all disabled test targets that then are being passed to the ctest invocation?

In the same way as the generated json file can have a `includes` entry, it can also have a `excludes` entry, which we could populate from a simple text file (similar to what we currently use for the targets), if that is given. Could you add that functionality, please?

hkaiser:
BTW, the Codacy CI tells you to remove trailing whitespace in the Python scripts.

hkaiser:
@sleepingeight Could you please resolve the conflicts? Then we can merge your PR.

sleepingeight:
@hkaiser is this good to be merged?

waleed7715:
@sleepingeight Great work!
Could you also add concurrency group to the workflow as done in: https://github.com/STEllAR-GROUP/hpx/pull/7104
Also, should we keep 18 parallel jobs or spread the tests across more runners to reduce completion time? currently longest takes ~59min

sleepingeight:
@waleed7715 I'll make the suggested changes. I'm looking into different approaches for reducing the runtime, currently we are only using 3GB-4GB of RAM during peak usage. So we are wasting approximately 10GB of RAM, so increasing parallelism while building seems to be a better approach than horizontal scaling.

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace_copy.hpp
```diff
@@ -21,6 +21,7 @@
 #include <type_traits>
 #include <utility>
 #include <vector>
+#include "hpx/config/forward.hpp"
```
hkaiser: When #include'ing hpx headers, please always use `<hpx/...>`, never `"hpx/..."`. Also, this internal header is already #included above by `<hpx/config.hpp>`

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace_copy.hpp
```diff
@@ -383,9 +380,9 @@ namespace hpx::parallel {
                 ldest = traits2::begin(sdest);
                 if (beg != end)
                 {
-                    segments.push_back(dispatch_async(traits2::get_id(sdest),
-                        algo, policy, forced_seq(), beg, end, ldest, f,
-                        new_value, HPX_FORWARD(Proj, proj)));
+                    segments.push_back(
+                        dispatch_async(traits2::get_id(sdest), algo, policy,
+                            forced_seq(), beg, end, ldest, f, new_value, proj));
```
hkaiser: May I ask you to separate the changes to this file into a new PR? The fixes (as valida sthey may be) have no relation to the CI infrastructure.

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace_copy.hpp
sleepingeight: Sorry for this, will update the file asap.

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace_copy.hpp
sleepingeight: This will lead to build failure in one of the test partition, so should we merge the changes to this file first and then this PR?

hkaiser: Let's merge #7122 first and the rebase this PR. Thanks!

hkaiser: FWIW, #7122 was merged now. You're still left with a conflict. Could you resolve that, please?

PR #6947, opened 2026-02-26
executors: fix uint32_t truncation in fork_join static scheduling (Phase 1)

### LINKED ISSUE #6922: fork_join_executor silently truncates ranges > 2^32
Summary
-------

fork_join_executor currently truncates iteration ranges to 32-bit
indices. When size > UINT32_MAX (~4.29B), indices wrap silently,
causing incorrect results with no warning or exception.

This affects both static and dynamic scheduling paths.

Root Cause
----------

1) Static scheduling:
   In fork_join_executor.hpp, partition boundaries are cast to
   std::uint32_t:

       auto const part_begin = static_cast<std::uint32_t>(...);
       auto const part_end   = static_cast<std::uint32_t>(...);

   If size > UINT32_MAX, part_begin/part_end wrap.

2) Dynamic scheduling:
   contiguous_index_queue is hardcoded to 32-bit indices via:

       static_assert(sizeof(T) <= 4, ...);

   It packs two 32-bit indices into a 64-bit atomic, so 64-bit
   ranges are fundamentally unsupported.

Impact
------

- Silent incorrect results for large ranges
- Affects for_each, transform, reduce, etc.
- Particularly problematic for HPC workloads

Proposed Fix
------------

Immediate (correctness):

1) For loop_schedule::static_:
   Replace std::uint32_t with std::size_t in partition calculations.

2) For loop_schedule::dynamic:
   Add runtime guard:

       if (size > std::numeric_limits<std::uint32_t>::max())
           throw std::overflow_error(
               "fork_join_executor: dynamic scheduling "
               "not supported for ranges > 2^32");

Long-term:
- Either implement a 64-bit index queue (128-bit CAS), or
- Provide a lock-based fallback for 64-bit indices.

Sahilll10:
This PR fixes #6922 for the static scheduling path of `fork_join_executor`.

Fixes:
- Replace `static_cast<std::uint32_t>` with `std::size_t` in partition boundary calculations (`part_begin`, `part_end`) to prevent silent truncation for ranges larger than UINT32_MAX.
- Add an assertion guard before `queue.reset()` to catch out‑of‑range sizes in debug builds.
- Add a regression test (`test_fork_join_static_large_range`) that verifies every index is visited exactly once under static scheduling.

The dynamic path still uses a 32‑bit queue and will be addressed in a follow‑up PR (Phase 2).

### DIFF
--- a/libs/core/executors/include/hpx/executors/detail/index_queue_spawning.hpp
+++ b/libs/core/executors/include/hpx/executors/detail/index_queue_spawning.hpp
@@ -351,46 +351,42 @@ namespace hpx::parallel::execution::detail {
         }
 
         // Initialize a queue for a worker thread.
-        void init_queue_depth_first(std::size_t const worker_thread,
-            std::size_t const size) noexcept
+        void init_queue_depth_first(
+            std::size_t const worker_thread, std::size_t const size) noexcept
         {
             auto& queue = queues[worker_thread].data_;
-            auto const part_begin =
-                (worker_thread * size) / num_threads;
-            auto const part_end =
-                (worker_thread * size + size) / num_threads;
+            auto const part_begin = (worker_thread * size) / num_threads;
+            auto const part_end = (worker_thread * size + size) / num_threads;
 
-            HPX_ASSERT_MSG(size <=
-             static_cast<std::size_t>(
-             (std::numeric_limits<std::uint32_t>::max)()),
-                 "init_queue_depth_first: ranges larger than "
-                 "UINT32_MAX are not supported");
+            HPX_ASSERT_MSG(
+                size <= static_cast<std::size_t>(
+                            (std::numeric_limits<std::uint32_t>::max)()),
+                "init_queue_depth_first: ranges larger than "
+                "UINT32_MAX are not supported");
 
             queue.reset(static_cast<std::uint32_t>(part_begin),
-            static_cast<std::uint32_t>(part_end));
+                static_cast<std::uint32_t>(part_end));
         }
 
-        void init_queue_breadth_first(std::size_t const worker_thread,
-            std::size_t const size) noexcept
+        void init_queue_breadth_first(
+            std::size_t const worker_thread, std::size_t const size) noexcept
         {
             auto& queue = queues[worker_thread].data_;
             auto const num_steps = size / num_threads + 1;
             auto const part_begin = worker_thread;
-            auto part_end =
-                (std::min) (size + num_threads - 1,
-                    part_begin + num_steps * num_threads);
-            auto const remainder =
-                (part_end - part_begin) % num_threads;
+            auto part_end = (std::min) (size + num_threads - 1,
+                part_begin + num_steps * num_threads);
+            auto const remainder = (part_end - part_begin) % num_threads;
             if (remainder != 0)
             {
                 part_end -= remainder;
             }
 
-            HPX_ASSERT_MSG(size <=
-             static_cast<std::size_t>(
-             (std::numeric_limits<std::uint32_t>::max)()),
-                 "init_queue_breadth_first: ranges larger than "
-                 "UINT32_MAX are not supported");
+            HPX_ASSERT_MSG(
+                size <= static_cast<std::size_t>(
+                            (std::numeric_limits<std::uint32_t>::max)()),
+                "init_queue_breadth_first: ranges larger than "
+                "UINT32_MAX are not supported");
 
             queue.reset(static_cast<std::uint32_t>(part_begin),
                 static_cast<std::uint32_t>(part_end),
--- a/libs/core/executors/include/hpx/executors/fork_join_executor.hpp
+++ b/libs/core/executors/include/hpx/executors/fork_join_executor.hpp
@@ -434,7 +434,7 @@ namespace hpx::execution::experimental {
                     std::size_t const num_pus = pool_->get_os_thread_count();
 
                     for (std::size_t pu = 0; t != num_threads_ && pu != num_pus;
-                         ++pu)
+                        ++pu)
                     {
                         std::size_t const pu_num = rp.get_pu_num(pu);
                         if (!main_thread_ok && pu == main_thread_)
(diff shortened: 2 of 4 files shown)


libs/core/executors/include/hpx/executors/fork_join_executor.hpp
```diff
@@ -503,10 +503,18 @@ namespace hpx::execution::experimental {
                 std::size_t const thread_index, std::size_t const num_threads,
                 std::size_t const size) noexcept
             {
-                auto const part_begin = static_cast<std::uint32_t>(
+                auto const part_begin = static_cast<std::size_t>(
```
hkaiser: I don't think we need a `static_cast` here at all, because all three of the variables are already of type `std::size_t`.

libs/core/executors/tests/unit/fork_join_executor.cpp
```diff
@@ -11,6 +11,7 @@
 #include <hpx/init.hpp>
 #include <hpx/modules/testing.hpp>
 #include <hpx/thread.hpp>
+#include <hpx/parallel/algorithm.hpp>
```
hkaiser: Please use
```suggestion
#include <hpx/algorithms.hpp>
```

libs/core/executors/tests/unit/fork_join_executor.cpp
```diff
@@ -471,11 +472,30 @@ void test_executor(hpx::threads::thread_priority priority,
     test_processing_mask(priority, stacksize, schedule);
 }
 
+void test_fork_join_static_large_range()
+{
+    // Regression test for #6922- static path partition truncation
+    fork_join_executor exec;
+    std::size_t const count = 100000;
```
hkaiser: If you use `100'000` as the iteration bounds you will not see truncation errors. OTOH, using a giant size for the vector might not be ok either. Please think of another way to verify your fix actually works (for instance summing up all values in between [0, 2^32) to see if the result is what you would expect).

Sahilll10:
@hkaiser, thank you for the detailed review – I really appreciate the clear guidance!

I agree with all three points:
1.	Remove the static_cast entirely – since part_begin and part_end are now std::size_t, the explicit cast is unnecessary. 
2.	Use <hpx/algorithm.hpp> – I'll update the include to match the project conventions.
3.	Improve the regression test— 100,000 doesn't cross the UINT32_MAX boundary so it doesn't actually test the truncation. I'll rethink the test to sum indices across the UINT32_MAX boundary and compare against the expected mathematical result, so we don't need a giant vector allocation.

I'll push the fixes shortly. Thanks again for taking the time to guide me .

I'm still learning, and your feedback helps a lot.

Sahilll10:
@hkaiser – I've pushed fixes addressing all three points:

1) Removed the redundant `static_cast<std::size_t>` wrappers – the variables are now used directly.

2) Updated the include to <hpx/algorithms.hpp>.

3) Reworked the regression test to cross the UINT32_MAX boundary (from UINT32_MAX - 500 to UINT32_MAX + 500) and compare the sum against the expected arithmetic value. This properly exercises the truncation fix without needing a large vector allocation.

I also ran clang-format.
Please let me know if anything else needs adjustment

Sahilll10:
Hi @hkaiser
I have extended the fix to index_queue_spawning.hpp as you suggested. 

Applied the same pattern to both init_queue_depth_first and init_queue_breadth_first — function parameters and local partition variables now use std::size_t, with explicit uint32_t casts retained at the queue.reset() call sites along with assert guards.

I also ran clang-format.
Please let me know if anything else needs adjustment.

libs/core/executors/include/hpx/executors/detail/index_queue_spawning.hpp
```diff
@@ -351,34 +351,50 @@ namespace hpx::parallel::execution::detail {
...
-            queue.reset(part_begin, part_end);
+            auto const part_begin =
+                (worker_thread * size) / num_threads;
+            auto const part_end =
+                (worker_thread * size + size) / num_threads;
+
+            HPX_ASSERT_MSG(size <=
+             static_cast<std::size_t>(
+             (std::numeric_limits<std::uint32_t>::max)()),
+                 "fork_join_executor: ranges larger than "
```
hkaiser: Please fix the error text to refer to the function it appears in.

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -479,7 +494,7 @@ namespace hpx::execution::experimental::detail {
             // Initialize the queues for all worker threads so that worker
             // threads can start stealing immediately when they start.
             for (std::uint32_t worker_thread = 0;
-                worker_thread != op_state->num_worker_threads; ++worker_thread)
+                 worker_thread != op_state->num_worker_threads; ++worker_thread)
```
hkaiser: We use clang-format v20 to check the formatting. Using any other version to format the files may lead to slight differences (like this one).

Sahilll10:
@hkaiser  -  thanks for the review.

I’ve addressed the two issues you pointed out:

1) Updated the HPX_ASSERT_MSG error texts in init_queue_depth_first and init_queue_breadth_first to refer to the correct function names instead of fork_join_executor.

2) Corrected the for-loop formatting in thread_pool_scheduler_bulk.hpp to match the expected structure (without relying on local clang-format).

The branch has been rebased on master and force-pushed.

Please let me know if anything else needs adjustment.

hkaiser:
Our inspect tool complains about:

```
/libs/core/executors/include/hpx/executors/fork_join_executor.hpp: *Line length limit*: 509
/libs/core/executors/tests/unit/fork_join_executor.cpp: *I* missing #include (cstdint) for symbol std::uint32_t on line 483, *Line length limit*: 478
```

Also, the clang-format check is not happy either.

Sahilll10:
@hkaiser  - fixed all three issues:

1. wrapped the long HPX_ASSERT_MSG lines in fork_join_executor.hpp to stay within the char limit

2. added the missing `#include <cstdint>` to the test file

3. fixed the long line in the test file

I’ve checked all four modified files for line-length violations and also ran clang-format v20 locally to ensure formatting matches CI expectations.

Please let me know if anything else needs adjustment.

Sahilll10:
@hkaiser — rebased onto current master.
Ready to merge whenever you wish to.

Thank you for the guidance and review.

hkaiser:
@Sahilll10 Thanks for your work on this!

PR #6950, opened 2026-02-28
Merge Path Algorithm

kollanur:
## Proposed Changes

  - The new merge implementation is based on the Merge Path algorithm described in [https://arxiv.org/pdf/1406.2628](https://arxiv.org/pdf/1406.2628?utm_source=chatgpt.com). This approach partitions two sorted arrays into contiguous segments that can be merged independently, enabling balanced work distribution across threads.
  
  - I have uploaded a table that compares CPU time between the old and new merge implementations and provides speedup information. The new merge implementation performs well for smaller data sizes compared to the old one, but its performance is worse for larger data sizes. Yet to investigate the reason behind bad performance for larger data. (Benchmark data gathered from CCCL NVBench).

[merge_compare_all_combinations_seconds_speedup 4.51.57 PM.xlsx](https://github.com/user-attachments/files/25617041/merge_compare_all_combinations_seconds_speedup.4.51.57.PM.xlsx)


libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,122 +651,88 @@ namespace hpx::parallel {
...
-                            reshaped.emplace_back(begin1, chunk_size1, begin2,
-                                chunk_size2, dest_start);
+            std::size_t a_low = (k > len2) ? (k - len2) : 0;
+            std::size_t a_high = (k < len1) ? k : len1;
 
-                            dest_start += chunk_size1 + chunk_size2;
+            while (a_low <= a_high)
+            {
+                std::size_t a = (a_low + a_high) / 2;
+                std::size_t b = k - a;
```
hkaiser: If you change the type of `a` and `b` to `std::ptrdiff_t` you can remove all the cluttering `static_cast`s below:
```suggestion
                std::ptrdiff_t a =  static_cast<std::ptrdiff_t>((a_low + a_high) / 2);
                std::ptrdiff_t b = static_cast<std::ptrdiff_t>(k - a);
```

This would also allow for changing the return type to `std::pair<std::ptrdiff_t, std::ptrdiff_t>`, which in turn allows to remove the `static_cast`s in the lambdas below.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,122 +651,88 @@ namespace hpx::parallel {
...
+                // cond1: a==0 || b==len2 || A[a-1] <= B[b]
+                bool cond1 = (a == 0) || (b == len2) ||
+                    leq(*(first1 + static_cast<std::ptrdiff_t>(a - 1)),
+                        *(first2 + static_cast<std::ptrdiff_t>(b)));
 
-                            begin1 = end_chunk1;
-                            remainder1 -= chunk_size1;
+                // cond2: b==0 || a==len1 || B[b-1] < A[a]
+                bool cond2 = (b == 0) || (a == len1) ||
+                    comp(*(first2 + static_cast<std::ptrdiff_t>(b - 1)),
```
hkaiser: Please use `HPX_INVOKE` instead of directly calling the comparison operator:
```suggestion
                    HPX_INVOKE(comp, *(first2 + static_cast<std::ptrdiff_t>(b - 1)),
```

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -813,8 +803,7 @@ namespace hpx::parallel {
                     std::next(first1, len1), l2, std::next(dest, len1 + len2)};
             };
 
-            auto reshape = get_reshape_chunks<Iter2>(
-                len2, first1, end1, comp, proj2, proj1, upper_bound_helper{});
+            auto reshape = get_diagonal_index(len1 + len2);
```
hkaiser: Wouldn't the computation of the diagonal indices for the case of the swapped input sequences require to swap the index computation as well?

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -783,15 +749,40 @@ namespace hpx::parallel {
...
+            auto f1 = overloaded{
+                [dest, comp, proj1, proj2](Iter1 it1, std::size_t size1,
+                    Iter2 it2, std::size_t size2, std::size_t dest_base) {
+                    if (size1 != 0 || size2 != 0)
+                    {
+                        sequential_merge(it1, std::next(it1, size1), it2,
+                            std::next(it2, size2), std::next(dest, dest_base),
+                            comp, proj1, proj2);
+                    }
+                },
```
hkaiser: IIRC, this lambda would be used only if the dynamic (variable) chunking was requested. I think we can simply ignore that case for now. Let's keep the code as simple as possible.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -783,15 +749,40 @@ namespace hpx::parallel {
...
+                [dest, comp, proj1, proj2, first1, first2, len1, len2](
+                    std::size_t idx, std::size_t chunk) {
+                    if (len1 != 0 || len2 != 0)
+                    {
+                        std::size_t N = len1 + len2;
+                        std::size_t k0 = std::min(idx * chunk, N);
```
hkaiser: Please write this as:
```suggestion
                        std::size_t k0 = (std::min)(idx * chunk, N);
```
which will prevent macro expansion of `min` (believe it or not, but on windows `min` and `max` are potentially defined as macros).

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: No, diagonals reprsents an output position ( After producing k merged elements, how many came from A and how many from B). Swapping the input sequences changes where the digonal intersects the merge path, but it does not change which diagonal you use.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: Ok. I will remove it.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
hkaiser: Makes sense. I had to ask ;-)

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -764,7 +764,7 @@ namespace hpx::parallel {
                     if (len1 != 0 || len2 != 0)
                     {
                         std::size_t N = len1 + len2;
-                        std::size_t k0 = std::min(idx * chunk, N);
+                        std::size_t k0 = (std::min)(idx * chunk, N);
                         std::size_t k1 = std::min((idx + 1) * chunk, N);
```
hkaiser: This needs to be adjusted as well.

kollanur: fixed.

kollanur:
> All of the new code assumes that the iterators are random_access. The code will break for bidirectional iterators, for instance. At least, I'd suggest using `std::next` instead of simply adding an integer to the iterators. This still will be awfully slow, but at least it would compile. If needed we could rewrite the `diagonal_intersection` later to directly operate on iterators instead of indicies, though.

Replaced with std::next.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,79 @@ namespace hpx::parallel {
...
+                bool cond2 = (b == 0) ||
+                    (a == static_cast<std::ptrdiff_t>(len1)) ||
+                    HPX_INVOKE(comp, *(std::next(first2, b - 1)),
+                        *(std::next(first1, a)));
+
+                if (cond1 && cond2)
+                    return {a, b};
+
+                if (a > 0 && b < static_cast<std::ptrdiff_t>(len2) &&
+                    comp(*(std::next(first2, b)), *(std::next(first1, a - 1))))
```
hkaiser: HPX_INVOKE?

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,79 @@ namespace hpx::parallel {
...
-                            remainder1 -= chunk_size1;
+        HPX_CXX_CORE_EXPORT template <typename Iter1, typename Iter2,
+            typename Comp>
+        std::pair<std::size_t, std::size_t> diagonal_intersection(Iter1 first1,
+            std::size_t len1, Iter2 first2, std::size_t len2, std::size_t k,
+            Comp comp)
+        {
+            //helper: x<=y
+            auto leq = [&](auto const& x, auto const& y) {
+                return !comp(y, x);
```
hkaiser: HPX_INVOKE?

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,79 @@ namespace hpx::parallel {
...
+            while (a_low <= a_high)
+            {
+                std::ptrdiff_t a =
+                    static_cast<std::ptrdiff_t>((a_low + a_high) / 2);
+                std::ptrdiff_t b = static_cast<std::ptrdiff_t>(k - a);
+
+                // cond1: a==0 || b==len2 || A[a-1] <= B[b]
+                bool cond1 = (a == 0) ||
+                    (b == static_cast<std::ptrdiff_t>(len2)) ||
+                    leq(*(std::next(first1, a - 1)), *(std::next(first2, b)));
```
hkaiser: Wouldn't it be easier to write:
```suggestion
                    !HPX_INVOKE(comp, *(std::next(first1, a - 1)), *(std::next(first2, b)));
```
Which would also make the `leq` above superfluous.

hkaiser: Also, generally, there is no need to have the additional pair of parenthesis around `std::next()`

hkaiser (commented):
Very nice! Thanks!

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: Sorry slipped out. Fixed now.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: Removed, leq is not necessary as you said it's redundant .

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: Noted and fixed.

hkaiser:
@kollanur Now, that we have the formalities out of the way, I have one more question. Your implementation of the `diagonal_intersection` function is currently `O(N)`, which is most likely causing the performance degradation for larger input sizes. OTOH, this very much looks to me like something that could be solved using some bisection algorithm (`lower_bound`/`upper_bound`?). What do you think?

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,77 @@ namespace hpx::parallel {
...
+                        HPX_INVOKE(proj1, *std::next(first1, a)));
+
+                if (cond1 && cond2)
+                    return {a, b};
+
+                if (a > 0 && b < static_cast<std::ptrdiff_t>(len2) &&
+                    HPX_INVOKE(comp, HPX_INVOKE(proj2, *std::next(first2, b)),
+                        HPX_INVOKE(proj1, *std::next(first1, a - 1))))
+                {
+                    if (a == 0)
```
hkaiser: This condition will never be true as the surrounding condition checks for `a > 0`.

kollanur:
> @kollanur Now, that we have the formalities out of the way, I have one more question. Your implementation of the `diagonal_intersection` function is currently `O(N)`, which is most likely causing the performance degradation for larger input sizes. OTOH, this very much looks to me like something that could be solved using some bisection algorithm (`lower_bound`/`upper_bound`?). What do you think?

Isn't the function doing binary search with complexity O(log(min(len1, len2)))?, it performs a binary search on 'A' over the range [max(0, k - len2), min(k, len1)], and the width of the search interval is (a_high - a_low +1)<= min(len1, len2) + 1, assuming random-access iterators.

kollanur:
> > @kollanur Now, that we have the formalities out of the way, I have one more question. Your implementation of the `diagonal_intersection` function is currently `O(N)`, which is most likely causing the performance degradation for larger input sizes. OTOH, this very much looks to me like something that could be solved using some bisection algorithm (`lower_bound`/`upper_bound`?). What do you think?
> 
> Isn't the function doing binary search with complexity O(log(min(len1, len2)))?, it performs a binary search on 'A' over the range [max(0, k - len2), min(k, len1)], and the width of the search interval is (a_high - a_low +1)<= min(len1, len2) + 1, assuming random-access iterators.

I can try to use lower_bound / upper_bound.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: removed. Thank you.

hkaiser:
> > @kollanur Now, that we have the formalities out of the way, I have one more question. Your implementation of the `diagonal_intersection` function is currently `O(N)`, which is most likely causing the performance degradation for larger input sizes. OTOH, this very much looks to me like something that could be solved using some bisection algorithm (`lower_bound`/`upper_bound`?). What do you think?
> 
> Isn't the function doing binary search with complexity O(log(min(len1, len2)))?, it performs a binary search on 'A' over the range [max(0, k - len2), min(k, len1)], and the width of the search interval is (a_high - a_low +1)<= min(len1, len2) + 1, assuming random-access iterators.

Yes, however inside the while loop it either increments the lower bound by one or decrements the upper bound by one for each iteration. Isn't that `O(N)`?

kollanur:
> > > @kollanur Now, that we have the formalities out of the way, I have one more question. Your implementation of the `diagonal_intersection` function is currently `O(N)`, which is most likely causing the performance degradation for larger input sizes. OTOH, this very much looks to me like something that could be solved using some bisection algorithm (`lower_bound`/`upper_bound`?). What do you think?
> > 
> > 
> > Isn't the function doing binary search with complexity O(log(min(len1, len2)))?, it performs a binary search on 'A' over the range [max(0, k - len2), min(k, len1)], and the width of the search interval is (a_high - a_low +1)<= min(len1, len2) + 1, assuming random-access iterators.
> 
> Yes, however inside the while loop it either increments the lower bound by one or decrements the upper bound by one for each iteration. Isn't that `O(N)`?

Yes, the loop performs a binary search over a. In each iteration, the search interval [a_low, a_high] is halved. The value of a is chosen as the midpoint of this interval, and in each step either the left or the right portion of the interval is discarded.

hkaiser:
> Yes, the loop performs a binary search over a. In each iteration, the search interval [a_low, a_high] is halved. The value of a is chosen as the midpoint of this interval, and in each step either the left or the right portion of the interval is discarded.

You're right, I just misunderstood the code. Thanks for clarifying.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,75 @@ namespace hpx::parallel {
...
-                            end2 = std::next(begin2, chunk_size2);
-                        }
+        HPX_CXX_CORE_EXPORT template <typename Iter1, typename Iter2,
+            typename Comp, typename Proj1, typename Proj2>
+        std::pair<std::size_t, std::size_t> diagonal_intersection(Iter1 first1,
+            std::size_t len1, Iter2 first2, std::size_t len2, std::size_t k,
+            Comp comp, Proj1&& proj1, Proj2&& proj2)
+        {
+            std::size_t a_low = (k > len2) ? (k - len2) : 0;
+            std::size_t a_high = (k < len1) ? k : len1;
```
hkaiser: You may want to add boundary checks:
```suggestion
            if (len1 == 0) return {0, k};
            if (len2 == 0) return {k, 0};
            std::size_t a_low = (k > len2) ? (k - len2) : 0;
            std::size_t a_high = (k < len1) ? k : len1;
            if (a_low == a_high)
            {
                std::ptrdiff_t a = a_low;
                return {a, k - a};  // Only one valid position
           }
```

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,75 @@ namespace hpx::parallel {
...
+        }
 
-                            begin2 = end2;
-                            end2 = std::next(begin2, chunk_size2);
-                        }
+        HPX_CXX_CORE_EXPORT template <typename Iter1, typename Iter2,
+            typename Comp, typename Proj1, typename Proj2>
+        std::pair<std::size_t, std::size_t> diagonal_intersection(Iter1 first1,
+            std::size_t len1, Iter2 first2, std::size_t len2, std::size_t k,
+            Comp comp, Proj1&& proj1, Proj2&& proj2)
```
hkaiser: Creating an additional overload of this function for the (common) case that the projections are `identity` will improve compiler optimizations:
```suggestion
        HPX_CXX_CORE_EXPORT template <typename Iter1, typename Iter2,
            typename Comp>
        std::pair<std::size_t, std::size_t> diagonal_intersection(Iter1 first1,
            std::size_t len1, Iter2 first2, std::size_t len2, std::size_t k,
            Comp comp, identity, identity)
```

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,75 @@ namespace hpx::parallel {
...
+                bool cond2 = (b == 0) ||
+                    (a == static_cast<std::ptrdiff_t>(len1)) ||
+                    HPX_INVOKE(comp,
+                        HPX_INVOKE(proj2, *std::next(first2, b - 1)),
+                        HPX_INVOKE(proj1, *std::next(first1, a)));
+
+                if (cond1 && cond2)
+                    return {a, b};
+
+                if (a > 0 && b < static_cast<std::ptrdiff_t>(len2) &&
```
hkaiser: This condition is equivalent to `!cond1`, do we really need to compute it again?

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,75 @@ namespace hpx::parallel {
...
-                            begin2 = end2;
-                            end2 = std::next(begin2, chunk_size2);
-                        }
+        HPX_CXX_CORE_EXPORT template <typename Iter1, typename Iter2,
+            typename Comp, typename Proj1, typename Proj2>
+        std::pair<std::size_t, std::size_t> diagonal_intersection(Iter1 first1,
+            std::size_t len1, Iter2 first2, std::size_t len2, std::size_t k,
+            Comp comp, Proj1&& proj1, Proj2&& proj2)
+        {
+            std::size_t a_low = (k > len2) ? (k - len2) : 0;
```
hkaiser: Making these variables `std::ptrdiff_t` as well might further reduce the `static_cast` clutter below.

kollanur:
> > Yes, the loop performs a binary search over a. In each iteration, the search interval [a_low, a_high] is halved. The value of a is chosen as the midpoint of this interval, and in each step either the left or the right portion of the interval is discarded.
> 
> You're right, I just misunderstood the code. Thanks for clarifying.

No problem. I am glad the clarification was helpful.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: Yes. You are correct. Thank you. Changed.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: Yes. Thank you.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: Ok professor.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: You are correct. Added the new boundaries. Thank you.

kollanur:
@hkaiser I have modified the chunking behaviour based on shape size. The following
[merge_compare_all_combinations_seconds_speedup-4.xlsx](https://github.com/user-attachments/files/25785838/merge_compare_all_combinations_seconds_speedup-4.xlsx)
 are the new results.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,93 @@ namespace hpx::parallel {
...
 
-            return reshape;
+        HPX_CXX_CORE_EXPORT template <typename Iter1, typename Iter2,
+            typename Comp>
+        std::pair<std::ptrdiff_t, std::ptrdiff_t> diagonal_intersection(
+            Iter1 first1, std::size_t len1, Iter2 first2, std::size_t len2,
+            std::size_t k, Comp comp, hpx::identity, hpx::identity)
+        {
+            return diagonal_intersection(first1, len1, first2, len2, k,
+                HPX_MOVE(comp), hpx::identity_v, hpx::identity_v);
```
hkaiser: That's not what I had in mind. I think that copying the code of the function implementation while getting rid of calling the projections altogether would help the compiler to better optimize. For an example please see our `lower_bound` implementations here: 
https://github.com/STEllAR-GROUP/hpx/blob/7ab0e511f3a3f3c576010a80e34ff33a86a944e7/libs/core/algorithms/include/hpx/parallel/algorithms/detail/upper_lower_bound.hpp#L187-L201
and here:
https://github.com/STEllAR-GROUP/hpx/blob/7ab0e511f3a3f3c576010a80e34ff33a86a944e7/libs/core/algorithms/include/hpx/parallel/algorithms/detail/upper_lower_bound.hpp#L241-L255

kollanur: Understood. Fixed it.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,153 @@ namespace hpx::parallel {
...
-                        1, merge_region{it1, size1, first2, size2, 0});
-                }
+                auto const shape_size = std::size(shape);
+                if (n <= 0)
+                    return d_idx;
 
-                std::vector<merge_region> reshaped;
-                reshaped.reserve(2 * shape_size);
+                std::size_t const nn = static_cast<std::size_t>(n);
+                std::size_t seg = (cores == 0 ? 1 : shape_size);
```
hkaiser, line 735: On one core we want to fall back to sequential execution as well. At least that's what we've been doing for the old algorithm.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,153 @@ namespace hpx::parallel {
...
-            typename Comp, typename Proj1, typename Proj2,
-            typename BinarySearchHelper>
-        auto get_reshape_chunks(std::size_t len1, Iter2 first2, Iter2 last2,
-            Comp&& comp, Proj1&& proj1, Proj2&& proj2, BinarySearchHelper)
+        HPX_CXX_CORE_EXPORT template <typename T>
+        auto get_diagonal_index(T const n)
         {
-            using merge_region =
-                hpx::tuple<Iter1, std::size_t, Iter2, std::size_t, std::size_t>;
+            auto diagonal_index = [n](auto&& shape, std::size_t cores) {
```
hkaiser: If you write it as
```suggestion
            auto diagonal_index = [n = static_cast<std::size_t>(n)](
                auto&& shape, std::size_t cores) {
```
You can remove the variable `nn` below.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,153 @@ namespace hpx::parallel {
...
+                for (std::size_t i = 0; i < seg; ++i)
+                    d_idx.emplace_back(i, chunk);
 
-                    // prevent chunks from growing too large
-                    if (size2 <= static_cast<std::size_t>(
-                                     1.2 * static_cast<double>(size1)))
-                    {
-                        reshaped.emplace_back(
-                            it1, size1, it2, size2, dest_start);
+                return d_idx;
```
hkaiser: Since the constructed `std::vector` returns constant values, we can replace it with a pair of simple iterators:
```
class const_index_value_iterator
  : public hpx::util::iterator_facade<
        const_index_value_iterator,                        // Derived
        hpx::tuple<std::size_t, std::size_t> const,        // Value type
        std::random_access_iterator_tag                    // Category
    >
{
private:
    // current index and constant value to be returned
    hpx::tuple<std::size_t, std::size_t> data_ = {0, 0};

public:
    const_index_value_iterator() = default;
    explicit constexpr const_index_value_iterator(std::size_t value, std::size_t start = 0) noexcept
      : data_(start, value)
    {}

private:
    friend class hpx::util::iterator_core_access;

    hpx::tuple<std::size_t, std::size_t> const& dereference() const noexcept
    {
        return data_;
    }

    void increment() noexcept { ++hpx::get<0>(data_); }
    void decrement() noexcept { --hpx::get<0>(data_); }

    void advance(std::ptrdiff_t n) noexcept
    { 
        hpx::get<0>(data_) = static_cast<std::size_t>(
            static_cast<std::ptrdiff_t>(hpx::get<0>(data_)) + n); 
    }

    std::ptrdiff_t distance_to(const_index_value_iterator const& other) const noexcept
    {
        return static_cast<std::ptrdiff_t>(hpx::get<0>(other.data_)) - 
            static_cast<std::ptrdiff_t>(hpx::get<0>(data_));
    }

    bool equal(const_index_value_iterator const& other) const noexcept
    {
        // Two iterators are equal when indices match and they share
        // the same value.
        return data_ == other.data_;
    }
};
```
with that you can return:
```
    auto begin = const_index_value_iterator(chunk);
    auto end = const_index_value_iterator(chunk, seg);
    return hpx::util::iterator_range{begin, end};
```
That avoids the allocation needed for the `std::vector`.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -783,13 +816,23 @@ namespace hpx::parallel {
...
+            auto f1 = [dest, comp, proj1, proj2, first1, first2, len1, len2](
+                          std::size_t idx, std::size_t chunk) {
+                if (len1 != 0 || len2 != 0)
                 {
-                    sequential_merge(it1, std::next(it1, size1), it2,
-                        std::next(it2, size2), std::next(dest, dest_base), comp,
-                        proj1, proj2);
+                    std::size_t N = len1 + len2;
+                    std::size_t k0 = (std::min) (idx * chunk, N);
+                    std::size_t k1 = (std::min) ((idx + 1) * chunk, N);
```
hkaiser: Let's avoid multiplications if possible:
```suggestion
                    std::size_t k1 = (std::min) (k0 + chunk, N);
```

hkaiser:
> @hkaiser I have modified the chunking behaviour based on shape size. The following [merge_compare_all_combinations_seconds_speedup-4.xlsx](https://github.com/user-attachments/files/25785838/merge_compare_all_combinations_seconds_speedup-4.xlsx) are the new results.

Marvelous! Excellent results!

kollanur:
> > @hkaiser I have modified the chunking behaviour based on shape size. The following [merge_compare_all_combinations_seconds_speedup-4.xlsx](https://github.com/user-attachments/files/25785838/merge_compare_all_combinations_seconds_speedup-4.xlsx) are the new results.
> 
> Marvelous! Excellent results!

Thank you. I am trying to see what else we can improve.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: This implementation is better, thank you. I have added

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: Yes. Thank you.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: Done. Thank you.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: You are right. Included it.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,215 @@ namespace hpx::parallel {
...
+                    return hpx::util::iterator_range{
+                        const_index_value_iterator(0, 0),
+                        const_index_value_iterator(0, 0)};
 
-                std::vector<merge_region> reshaped;
-                reshaped.reserve(2 * shape_size);
+                if (cores == 1)
+                    return hpx::util::iterator_range{
+                        const_index_value_iterator(n),
+                        const_index_value_iterator(n, 0)};
```
hkaiser: This creates an empty iterator_pair, I think this should be: 
```suggestion
                        const_index_value_iterator(n, 1)};
```

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,215 @@ namespace hpx::parallel {
...
-                    // switch to sequential execution in this case
-                    auto [it1, size1, _] = *std::begin(shape);
-                    auto size2 = detail::distance(first2, last2);
-                    return std::vector<merge_region>(
-                        1, merge_region{it1, size1, first2, size2, 0});
-                }
+                if (n <= 0)
+                    return hpx::util::iterator_range{
+                        const_index_value_iterator(0, 0),
+                        const_index_value_iterator(0, 0)};
```
hkaiser:
```suggestion
                    return hpx::util::iterator_range<const_index_value_iterator>();
```

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: ok.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: ok.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -651,120 +651,214 @@ namespace hpx::parallel {
...
-                if (shape_size == 1 && cores == 1)
-                {
-                    // special case: one core, one chunk - the partitioner will
-                    // switch to sequential execution in this case
-                    auto [it1, size1, _] = *std::begin(shape);
-                    auto size2 = detail::distance(first2, last2);
-                    return std::vector<merge_region>(
-                        1, merge_region{it1, size1, first2, size2, 0});
-                }
+                if (n <= 0)
```
hkaiser: One last thing: `n` can't be negative, please write:
```suggestion
                if (n == 0)
```

kollanur: done.

hkaiser:
@kollanur Thanks so much for working on this. This is a great addition!

kollanur:
> @kollanur Thanks so much for working on this. This is a great addition!

Mention not.

PR #6951, opened 2026-02-28
Add zero-validation for with_processing_units_count

Sanchit2662:
## Summary

`with_processing_units_count` silently accepts `0` as valid input, but internally `num_cores_ = 0` means "use all cores." This collision lets user bugs slip through as unintended maximum parallelism. The fix adds validation to both `thread_pool_policy_scheduler` and `parallel_policy_executor`, matching the precedent set in `fork_join_executor` 
## Fix

Added zero-check with `HPX_THROW_EXCEPTION` in both executors. Kept `constexpr`, removed `noexcept`.

```cpp
// thread_pool_scheduler.hpp & parallel_executor.hpp
if (num_cores == 0) {
    HPX_THROW_EXCEPTION(hpx::error::bad_parameter,
        "with_processing_units_count",
        "number of processing units must be greater than zero");
}
```

### DIFF
--- a/libs/core/executors/include/hpx/executors/parallel_executor.hpp
+++ b/libs/core/executors/include/hpx/executors/parallel_executor.hpp
@@ -212,8 +212,16 @@ namespace hpx::execution {
             requires(std::is_convertible_v<Executor_, parallel_policy_executor>)
         friend constexpr auto tag_invoke(
             hpx::execution::experimental::with_processing_units_count_t,
-            Executor_ const& exec, std::size_t num_cores) noexcept
+            Executor_ const& exec, std::size_t num_cores)
         {
+            if (num_cores == 0)
+            {
+                HPX_THROW_EXCEPTION(hpx::error::bad_parameter,
+                    "parallel_policy_executor::"
+                    "with_processing_units_count",
+                    "number of processing units must be greater than "
+                    "zero");
+            }
             auto exec_with_num_cores = exec;
             exec_with_num_cores.num_cores_ = num_cores;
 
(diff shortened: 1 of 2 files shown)


libs/core/executors/include/hpx/executors/parallel_executor.hpp
```diff
@@ -212,8 +212,16 @@ namespace hpx::execution {
             requires(std::is_convertible_v<Executor_, parallel_policy_executor>)
         friend constexpr auto tag_invoke(
```
hkaiser: In C++20 exceptions are never `constexpr`, IIRC. I think you can remove this as well.

libs/core/executors/include/hpx/executors/parallel_executor.hpp
```diff
@@ -212,8 +212,16 @@ namespace hpx::execution {
...
+            Executor_ const& exec, std::size_t num_cores)
         {
+            if (num_cores == 0)
+            {
+                HPX_THROW_EXCEPTION(hpx::error::bad_parameter,
+                    "parallel_policy_executor::"
+                    "with_processing_units_count",
+                    "number of processing units must be greater than "
+                    "zero");
+            }
```
hkaiser, line 223: Would it be better to follow the convention and interpret a zero as representing 'all cores'? This could be done by writing
```
        if (num_cores == 0) {
            auto pool =
                exec.pool_ ? exec.pool_ : threads::detail::get_self_or_default_pool();
           num_cores = pool->get_active_os_thread_count();
        }
```

Sanchit2662:
@hkaiser  , I have made the required changes.

hkaiser (approved):
LGTM, thanks!

hkaiser:
@Sanchit2662 Thank you for working on this!

PR #6954, opened 2026-03-01
Fixed the managed to unmanaged change

### LINKED ISSUE #6953: base_component::get_unmanaged_id incorrectly returns management_type::managed ID
Summary
-------

`base_component::get_unmanaged_id` returns an `hpx::id_type` tagged with `management_type::managed` instead of `management_type::unmanaged`. When the returned ID is serialized to a remote locality, this triggers an assertion failure in `credit_handling.cpp` in debug builds.

Root Cause
----------

In `libs/full/components_base/src/server/component_base.cpp` (line 151):

```cpp
hpx::id_type base_component::get_unmanaged_id(naming::gid_type const& gid)
{
    return {gid, hpx::id_type::management_type::managed};  // wrong
}
```

During serialization, `preprocess_gid` (in `credit_handling.cpp`) first checks the management type and returns early for unmanaged IDs:

```cpp
if (hpx::id_type::management_type::unmanaged ==
        id_impl.get_management_type() || ...)
{
    return;  // fast path — would be taken for a correct unmanaged ID
}
...
HPX_ASSERT(has_credits(id_impl));  // line 565 — fires here
```

Because `get_unmanaged_id` incorrectly stamps the ID as `managed`, the early return is not taken. The ID has no credits (it never went through
`get_id` / `replenish_credits`), so `has_credits` is false and the assertion fires.

Impact
------

- Debug-mode assertion failure whenever a component passes its own unmanaged self-handle (obtained via `get_unmanaged_id`) as an action
  argument to a remote locality
- In release builds `HPX_ASSERT` is a no-op, so the bug is silent — but the serialized management type is still wrong

Proposed Fix
------------

One-word change in `component_base.cpp`:

```cpp
hpx::id_type base_component::get_unmanaged_id(naming::gid_type const& gid)
{
    return {gid, hpx::id_type::management_type::unmanaged};  // correct
}
```

nischitkumar:
Fixes #6953

## Proposed Changes
Implemented the one word fix (managed->unmanaged). This was necessary because base_component::get_unmanaged_id returns an hpx::id_type tagged with management_type::managed instead of management_type::unmanaged. When the returned ID is serialized to a remote locality, this triggers an assertion failure in credit_handling.cpp in debug builds.


hkaiser:
@nischitkumar Would you have the time to finish this PR?

hkaiser:
> @nischitkumar Would you have the time to finish this PR?

Ping? Do you have any plans to wrap this up? Do you need help?

nischitkumar:
Hey @hkaiser, apologies for the delay in getting back, I'd been tied up with my university exams. I shall resume working on the PR and shall find relevant test cases. Shall definitely ping you if I run into any issues.

hkaiser:
@nischitkumar Do you still plan to work on this PR? I believe your proposed change is valid, so all we need to prove that it is would be a test that exercises the modified code path, thus showing that a problem that was there before is now gone.

hkaiser:
Closing as abandoned...

PR #6955, opened 2026-03-01
fix: correct argument order for `next_or_subrange` and `add_ready_future_idx` functions

### LINKED ISSUE #6931: Two correctness bugs in  chunk_size.hpp : swapped next_or_subrange args and mismatched  add_ready_future_idx  parameter order
While reviewing:

    libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size.hpp

I found two silent correctness bugs affecting bulk iteration shape handling.

Bug 1 — Swapped arguments in next_or_subrange (variable chunk path)
===============================================================================

Location:
    get_bulk_iteration_shape_variable (around line 320)

Signature:
    constexpr auto next_or_subrange(IterOrR const& target,
        std::size_t offset, std::size_t size)

Current code:

    std::size_t chunk = (std::min)(chunk_size, count);
    shape.emplace_back(it_or_r, chunk);
    chunk = (std::min)(count, chunk);
    count -= chunk;
    it_or_r = next_or_subrange(it_or_r, count, chunk);  // BUG

Issue:
At this point `count` has already been decremented and represents the
remaining elements, not the advance distance.

The iterator should advance by `chunk`, but instead advances by the
remaining `count`, leading to incorrect range progression.

Correct form:

    it_or_r = next_or_subrange(it_or_r, chunk, count);

This matches the correct usage in the other overload (line ~214).

Impact:
- Any algorithm using the variable-chunk-size path (std::true_type)
- Multi-chunk scenarios silently process incorrect ranges
- No compile-time warning

Proposed Fix:
- Swap the two arguments at line ~320
- Add a unit test ensuring multi-chunk iteration visits all elements exactly once

Bug 2 — Parameter order mismatch in add_ready_future_idx
===============================================================================

Location:
    Overloads for add_ready_future_idx (lines ~366–380)

Correct overload (future<void>):

    void add_ready_future_idx(std::vector<hpx::future<void>>& workitems,
        F&& f, FwdIter first, std::size_t base_idx, std::size_t count)

Incorrect overload (shared_future<void>):

    void add_ready_future_idx(
        std::vector<hpx::shared_future<void>>& workitems,
        F&& f, std::size_t base_idx, FwdIter first, std::size_t count)
           // base_idx and first swapped

Problem:
Call sites pass arguments in the order:

    (first, base_idx, count)

This matches the future<void> overload, but in the shared_future<void>
case the parameters are reversed, potentially binding:

    first → base_idx
    base_idx → first

This leads to undefined behavior passed silently into the kernel.

Corrected version:

    void add_ready_future_idx(
        std::vector<hpx::shared_future<void>>& workitems,
        F&& f, FwdIter first, std::size_t base_idx, std::size_t count)
    {
        HPX_FORWARD(F, f)(first, count, base_idx);
        workitems.push_back(hpx::make_ready_future());
    }

Impact:
- Silent UB when shared_future<void> workitems are used
- Incorrect base_idx propagation in indexed iteration shapes

Proposed Fix:
- Reorder parameters of the shared_future<void> overload
- Add test covering shared_future<void> path via
  get_bulk_iteration_shape_idx
- Optionally document required parameter order to prevent recurrence

Recommendation
===============================================================================

Both issues affect correctness and should be fixed before further
refactoring in this area. These are silent runtime errors without
diagnostics.

OmkarTipugade:
Fixes #6931 

#Proposed Changes
Fix swapped arguments in next_or_subrange call (

get_bulk_iteration_shape_variable
): After count -= chunk, count holds the remaining elements — not the advance distance. The call was passing 

(it_or_r, count, chunk)
 but should be 

(it_or_r, chunk, count)
, matching the identical pattern used in the fixed-chunk overload. Without this fix, multi-chunk variable-size iteration silently visits incorrect ranges.

Fix parameter order mismatch in  add_ready_future_idx
 (shared_future overload): The shared_future<void> overload had FwdIter first and std::size_t base_idx swapped relative to both call sites and the other two overloads. This caused silent UB — an iterator was bound to an integral base_idx parameter and vice versa. Reordered to 

(workitems, f, first, base_idx, count)
 to be consistent across all three overloads.

#Any background context you want to provide?
Both bugs live in 
libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size.hpp
 and affect the parallel algorithm chunking infrastructure:

The next_or_subrange bug only manifests on the variable-chunk-size path (std::true_type), so fixed-chunk algorithms are unaffected. Multi-chunk scenarios silently process wrong sub-ranges without any compile-time or runtime warning.
The  add_ready_future_idx
 parameter swap only triggers when a caller uses std::vector<hpx::shared_future<void>> workitems (vs hpx::future<void>), so it is similarly silent — the code compiles and links cleanly but propagates a corrupted base_idx and iterator into the user-supplied kernel.

## Checklist

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size.hpp
@@ -282,10 +282,6 @@ namespace hpx::parallel::util::detail {
         std::size_t max_chunks =
             hpx::execution::experimental::maximal_number_of_chunks(
                 policy.parameters(), policy.executor(), cores, count);
-        if (max_chunks == 0)
-        {
-            max_chunks = 1;
-        }
 
         Stride stride = parallel::detail::abs(s);
 
@@ -556,10 +552,6 @@ namespace hpx::parallel::util::detail {
         std::size_t max_chunks =
             hpx::execution::experimental::maximal_number_of_chunks(
                 policy.parameters(), policy.executor(), cores, count);
-        if (max_chunks == 0)
-        {
-            max_chunks = 1;
-        }
 
         Stride stride = parallel::detail::abs(s);
 


OmkarTipugade:
@hkaiser Thanks for the review. I will add regression tests that:

1. Reproduce incorrect multi-chunk traversal in variable chunk path

2. Validate correct base_idx propagation for shared_future overload

I’ll push an update shortly.

libs/core/algorithms/tests/unit/algorithms/detail/chunk_size_variable_regression.cpp
```diff
@@ -0,0 +1,150 @@
...
+//
+//  2. add_ready_future_idx (shared_future<void> overload):  the parameter list
+//     had (base_idx, first, count) instead of (first, base_idx, count), so the
+//     raw iterator value was misinterpreted as base_idx, corrupting every index
+//     passed to the user functor.
+
+#include <hpx/modules/algorithms.hpp>
+#include <hpx/modules/execution.hpp>
+#include <hpx/modules/testing.hpp>
+#include <hpx/parallel/util/detail/chunk_size.hpp>
```
hkaiser: Instead of #including a module-internal header directly, I'd suggest adding the header to the generated module header here: https://github.com/STEllAR-GROUP/hpx/blob/7ab0e511f3a3f3c576010a80e34ff33a86a944e7/libs/core/algorithms/CMakeLists.txt#L270-L271
Otherwise you wil break the CIs that test the C++ module interfaces.

OmkarTipugade:
@hkaiser, sir, could you please review my recently updated PR

hkaiser:
Please also look at the reported clang-format and cmake-format issues.

hkaiser:
Please fix the reported formatting issues, then we can merge this PR.

OmkarTipugade:
@hkaiser, I am struggling to resolve formatting issues. Can you give me CircleCI access to use the Explain Error Feature? It will help me to find the exact issues after PR merge. You can remove access

OmkarTipugade:
@hkaiser Please review.

hkaiser:
The tests still report formatting issues in two of the CMake files:
```diff
diff --git a/libs/core/algorithms/CMakeLists.txt b/libs/core/algorithms/CMakeLists.txt
index 2f17c6d..dc500f4 100644
--- a/libs/core/algorithms/CMakeLists.txt
+++ b/libs/core/algorithms/CMakeLists.txt
@@ -267,9 +267,10 @@ add_hpx_module(
   GLOBAL_HEADER_MODULE_GEN ON
   HEADERS ${algorithms_headers}
   COMPAT_HEADERS ${algorithms_compat_headers}
-  ADD_TO_GLOBAL_HEADER "hpx/parallel/util/detail/chunk_size.hpp"
-                       "hpx/parallel/util/detail/chunk_size_iterator.hpp"
-                       "hpx/parallel/util/detail/handle_remote_exceptions.hpp"
+  ADD_TO_GLOBAL_HEADER
+    "hpx/parallel/util/detail/chunk_size.hpp"
+    "hpx/parallel/util/detail/chunk_size_iterator.hpp"
+    "hpx/parallel/util/detail/handle_remote_exceptions.hpp"
   SOURCES ${algorithms_sources}
   MODULE_DEPENDENCIES
     hpx_async_base
diff --git a/libs/core/algorithms/tests/unit/algorithms/detail/CMakeLists.txt b/libs/core/algorithms/tests/unit/algorithms/detail/CMakeLists.txt
index 5b9b7b7..08da99c 100644
--- a/libs/core/algorithms/tests/unit/algorithms/detail/CMakeLists.txt
+++ b/libs/core/algorithms/tests/unit/algorithms/detail/CMakeLists.txt
@@ -4,8 +4,7 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-set(tests chunk_size_iterator chunk_size_idx_iterator
-          test_insertion_sort
+set(tests chunk_size_iterator chunk_size_idx_iterator test_insertion_sort
           test_parallel_stable_sort test_sample_sort test_spin_sort
 )
 ```

**Comment by hkaiser** [comment]
@OmkarTipugade The new test you added causes compilation issues on the MacOS CIs. Could you please have a look?

**Comment by hkaiser** [comment]
@OmkarTipugade The newly added test fails with a segmentation fault: https://github.com/STEllAR-GROUP/hpx/actions/runs/22885477413/job/66455104910?pr=6955. Could you have a look, please?

**Comment by OmkarTipugade** [comment]
@hkaiser, how do I run these tests locally

**Comment by hkaiser** [comment]
> @hkaiser, how do I run these tests locally
```
ctest --tests-regex tests.regressions.modules.algorithms.chunk_size_variable_regression
```
should do the trick.

**Comment by OmkarTipugade** [comment]
@hkaiser, please review it

File: libs/core/algorithms/include/hpx/parallel/util/detail/chunk_size.hpp [hunk 2]
```diff
@@ -282,7 +282,10 @@ namespace hpx::parallel::util::detail {
         std::size_t max_chunks =
             hpx::execution::experimental::maximal_number_of_chunks(
                 policy.parameters(), policy.executor(), cores, count);
-        HPX_ASSERT(0 != max_chunks);
+        if (max_chunks == 0)
+        {
+            max_chunks = 1;
+        }
```
hkaiser: The original idea was that the returned `max_chunk` indicates that the number of chunks should be calculated somehow, which is currently done by the function `adjust_chunk_size_and_max_chunks` below. Are you sure that your change didn't break anything?

Why was `max_chunks == 0` causing issues in this context? Could you elaborate on the rationale of your change, please?

OmkarTipugade: I have analyzed the chunk size logic and confirmed that my previous fix was indeed sub-optimal.

The rationale for the previous change was to avoid an assertion failure (HPX_ASSERT(0 != max_chunks)) that occurred when maximal_number_of_chunks returned 0. However, I now realize that 0 is actually a valid sentinel value intended to trigger a default calculation in adjust_chunk_size_and_max_chunks. By forcing max_chunks to 1, I inadvertently bypassed that logic.

I have prepared an implementation plan to revert the max_chunks = 1 guard and allow 0 to flow through, which is more robust and consistent with the existing HPX infrastructure. Please review the plan.

Regarding your question: max_chunks == 0 was not "causing issues" by itself; rather, the original code had an incorrect assertion that didn't account for 0 being a valid signal. My fix addressed the assertion but changed the semantics incorrectly.

Plan: 
The previous fix for the assertion failure in  get_bulk_iteration_shape_variable and get_bulk_iteration_shape_idx_variable
 addressed the symptoms but not the root cause. It replaced an incorrect HPX_ASSERT(0 != max_chunks) with a guard that forced  max_chunks to 1.

However, max_chunks == 0 is actually a valid sentinel value in the HPX chunking logic, signaling that 
adjust_chunk_size_and_max_chunks should calculate sensible defaults. By forcing it to 1, we were bypassing this logic and potentially causing HPX_ASSERT(chunk_size != 0) to fail if chunk_size was also 0.

Modifications:
Remove the if (max_chunks == 0) { max_chunks = 1; } guard in  get_bulk_iteration_shape_variable.
Remove the if (max_chunks == 0) { max_chunks = 1; } guard in get_bulk_iteration_shape_idx_variable.
(The original HPX_ASSERT(0 != max_chunks) was already removed in the previous commit, but we will ensure it doesn't return.
Does this plan look good to you?

hkaiser: Sounds good to me. Thanks for the explanations.

OmkarTipugade:
Thanks, @hkaiser sir! I learned a lot from your feedback. Appreciate your time

PR #6957, opened 2026-03-02
Implement datapar for move with tests

### LINKED ISSUE #2333: Implement datapar for parallel algorithms

BhoomishGupta:
Partially Fixes #2333

## Proposed Changes
  - Added datapar support for `hpx::move` algorithm by converting `move_n` into a customization point object `move_n_t` in `transfer.hpp`, following the same pattern as `copy_n_t`
  - Added datapar specialization `datapar_move_n` in `transfer.hpp` that hooks into the `move_n_t` CPO via `tag_invoke` for vectorpack execution policies.
  -Extracted reusable move test functions into `move_tests.hpp` and added a new `move_datapar.cpp` test exercising simd, par_simd, and their task variants.

This follows the same approach used for `copy/copy_n` datapar support (PR#5428).

## Checklist

- [X] I have added a test using random numbers; I have made sure it uses a seed, and that random numbers generated are valid inputs for the tests.


hkaiser (commented):
General note, I think that this patch, while academically interesting, practically has only very limited use as all types for which the operation (copy/move) could be vectorized, e.g. `int`, `float`, etc., are trivially (bitwise) copyable, which makes moving equivalent to copying. As a consequence, the code you added will result in exactly the same binary code as the equivalent copy operation.

BhoomishGupta:
> General note, I think that this patch, while academically interesting, practically has only very limited use as all types for which the operation (copy/move) could be vectorized, e.g. `int`, `float`, etc., are trivially (bitwise) copyable, which makes moving equivalent to copying. As a consequence, the code you added will result in exactly the same binary code as the equivalent copy operation.

That makes sense, @hkaiser. You're right that for types that can be vectorized, move just compiles down to a copy.

I mainly opened this to check move off the list in #2333 and keep the API consistent with the rest of the algorithms. Given the limited practical utility, would you prefer to close this PR, or should we merge it purely for API symmetry?

hkaiser:
> > General note, I think that this patch, while academically interesting, practically has only very limited use as all types for which the operation (copy/move) could be vectorized, e.g. `int`, `float`, etc., are trivially (bitwise) copyable, which makes moving equivalent to copying. As a consequence, the code you added will result in exactly the same binary code as the equivalent copy operation.
> 
> That makes sense, @hkaiser. You're right that for types that can be vectorized, move just compiles down to a copy.
> 
> I mainly opened this to check move off the list in #2333 and keep the API consistent with the rest of the algorithms. Given the limited practical utility, would you prefer to close this PR, or should we merge it purely for API symmetry?

I have fixed the desciption in #2333. Let's close this PR here. Thanks for looking into this in any case!

BhoomishGupta:
@hkaiser I think `uninitialized_fill`, `uninitialized_fill_n`, `uninitialized_copy`, `uninitialized_copy_n` will also produce the same binary code

hkaiser:
> @hkaiser I think `uninitialized_fill`, `uninitialized_fill_n`, `uninitialized_copy`, `uninitialized_copy_n` will also produce the same binary code

Not quite. The standard requires for the uninitialized algorithms to call the constructors for the elements (especially for the `fill` algorithms). 

For the `copy` algorithms -- assuming that the vectorizable types are trivially constructible -- you might be right. Also, considering that our `hpx::copy` is already optimized for trivially copyable types to use `memmove` for the actual operation, any SIMD version might not give us any performance benefit.

PR #6958, opened 2026-03-03
Re-add iterator traits

hkaiser:
- This is necessary for backwards compatibility with external code depending on those traits. We may later deprecate the use of those traits.


hkaiser:
@kollanur this should resolve the issues you were facing when working on CCCL.

Issue #6941, opened 2026-02-25
Remove obsolete feature tests


hkaiser:
Now that we have moved to C++20, many of our feature tests that were checking for C++20 features supported by a particular compiler are obsolete and can be removed. Let's collect a list of those here:

- [x] HPX_WITH_CXX20_STD_RANGES_ITER_SWAP (#7051)
- [x] HPX_WITH_CXX20_PERFECT_PACK_CAPTURE (#6964)
- [x] HPX_WITH_CXX20_PAREN_INITIALIZATION_OF_AGGREGATES (#6983)
- [x] HPX_WITH_CXX20_STD_ENDIAN (#6966)
- [x] HPX_WITH_CXX20_TRIVIAL_VIRTUAL_DESTRUCTOR (#6959)
- [x] HPX_WITH_CXX20_STD_DEFAULT_SENTINEL (#6977)
- [x] HPX_WITH_CXX20_STD_BIT_CAST (#6976)
- [x] HPX_WITH_CXX20_STD_IDENTITY (#6973)
- [x] HPX_WITH_CXX20_CONSTEXPR_DESTRUCTOR (#6980)
- [x] HPX_WITH_CXX20_SOURCE_LOCATION (#6996)
- [x] HPX_WITH_CXX20_STD_CONSTRUCT_AT (#7066)
- [x] HPX_WITH_CXX20_LAMBDA_CAPTURE (#7067)
- [x] HPX_WITH_CXX20_STD_EXECUTION_POLICES (#7068)
- [x] HPX_WITH_CXX20_STD_ENDIAN (#7093)

Each of the above needs to be verified separately to make sure removing it will not break anything on any tested platform.

This also means removing the corresponding #ifdef's from the code base to unconditionally use C++ 20 features.

GitMasterJatin:
Should I submit one PR per macro (so each can be verified independently), or can I batch the simple ones together and keep only the complex ones separate?

hkaiser:
I'd suggest to do it one-by-one to make it easier to identify potential issues created by removing the settings. Thanks!

Sahilll10:
Hi @hkaiser 
 I'd like to work on removing HPX_WITH_CXX20_STD_ENDIAN.

 I'll start with that one and submit a separate PR. Just wanted to flag it so there's no duplication with other contributors working on the list.

Would be grateful for your guidance as we move ahead.

Sahilll10:
[...]

PR #6964, opened 2026-03-05
cmake: Remove obsolete feature test for CXX20_PERFECT_PACK_CAPTURE

v4xsh:
Contributes to fixing #6941

## Proposed Changes

* Remove the obsolete feature test `HPX_WITH_CXX20_PERFECT_PACK_CAPTURE`.
* Remove the corresponding test file `cmake/tests/cxx20_perfect_pack_capture.cpp`.
* Remove the invocation of the feature test from `cmake/HPX_PerformCxxFeatureTests.cmake`.
* Remove the `HPX_HAVE_CXX20_PERFECT_PACK_CAPTURE` preprocessor guards from `libs/core/executors/include/hpx/executors/scheduler_executor.hpp` and make the C++20 implementation unconditional.

HPX now requires C++20, which makes the feature test for perfect pack capture unnecessary. This PR removes the CMake configuration test, its associated test file, and the related preprocessor guards in the codebase. The C++20 implementation using lambda pack capture is now used unconditionally.

## Checklist

* [ ] I have added a new feature and have added tests to go along with it.
* [ ] I have fixed a bug and have added a regression test.
* [ ] I have added a test using random numbers; I have made sure it uses a seed, and that random numbers generated are valid inputs for the tests.

### DIFF
--- a/cmake/HPX_PerformCxxFeatureTests.cmake
+++ b/cmake/HPX_PerformCxxFeatureTests.cmake
@@ -102,10 +102,6 @@ function(hpx_perform_cxx_feature_tests)
       DEFINITIONS HPX_HAVE_CXX20_SOURCE_LOCATION
     )
 
-    hpx_check_for_cxx20_perfect_pack_capture(
-      DEFINITIONS HPX_HAVE_CXX20_PERFECT_PACK_CAPTURE
-    )
-
     if(HPX_WITH_SUPPORT_NO_UNIQUE_ADDRESS_ATTRIBUTE AND NOT MSVC) # see above
       hpx_check_for_cxx20_no_unique_address_attribute(
         DEFINITIONS HPX_HAVE_CXX20_NO_UNIQUE_ADDRESS_ATTRIBUTE
--- a/libs/core/executors/include/hpx/executors/scheduler_executor.hpp
+++ b/libs/core/executors/include/hpx/executors/scheduler_executor.hpp
@@ -27,7 +27,7 @@
 namespace hpx::execution::experimental {
 
     namespace detail {
-#if defined(HPX_HAVE_CXX20_PERFECT_PACK_CAPTURE)
+
         HPX_CXX_CORE_EXPORT template <typename F, typename... Ts>
         auto captured_args_then(F&& f, Ts&&... ts)
         {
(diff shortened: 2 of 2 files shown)


v4xsh:
Thanks!  @hkaiser I had a look through the codebase to see where argument packs are stored in tuples for lambda captures (similar to the pattern removed in this PR). I found a few places where the same tuple-capture workaround appears and could potentially be replaced with perfect pack capture now that HPX requires C++20.

For example:

1. **`libs/core/algorithms/include/hpx/parallel/task_group.hpp`**

```cpp
hpx::parallel::execution::post(HPX_FORWARD(Executor, exec),
    [this, on_exit = HPX_MOVE(on_exit), f = HPX_FORWARD(F, f),
        t = hpx::make_tuple(HPX_FORWARD(Ts, ts)...)]() mutable {

        auto _(HPX_MOVE(on_exit));

        hpx::detail::try_catch_exception_ptr(
            [&]() { hpx::invoke_fused(HPX_MOVE(f), HPX_MOVE(t)); },
            [this](std::exception_ptr e) {
                add_exception(HPX_MOVE(e));
            });
    });
```

2. **`libs/core/resiliency/include/hpx/resiliency/async_replicate_executor.hpp`**

```cpp
auto func = [f = HPX_FORWARD(F, f),
                t = hpx::make_tuple(HPX_FORWARD(Ts, ts)...)](
                std::size_t) mutable -> result_type {
    return hpx::invoke_fused(f, t);
};
```

3. **`libs/core/resiliency/include/hpx/resiliency/async_replicate_executor.hpp`**

```cpp
auto func = [f = HPX_FORWARD(F, f),
                t = hpx::make_tuple(HPX_FORWARD(Ts, ts)...)](
                std::size_t) mutable {
    hpx::invoke_fused(f, t);
    return 0;
};
```

These appear to follow the same tuple-capture pattern that could potentially be simplified using C++20 perfect pack capture.

Would it be alright if I prepare small follow-up PRs to update these locations after verifying that the semantics remain unchanged?

hkaiser:
> Would it be alright if I prepare small follow-up PRs to update these locations after verifying that the semantics remain unchanged?

Yes, please. Let's have a separate PR for this. Thanks!

v4xsh:
@hkaiser I’ve opened a follow-up PR to address the tuple-based lambda captures we discussed:

#6965

This replaces the tuple capture + `invoke_fused` pattern with C++20 pack capture in the locations identified earlier, while preserving the original semantics (avoiding moves in `bulk_async_execute` lambdas that may run multiple times).

Please let me know if you’d like any adjustments.

PR #6965, opened 2026-03-05
executors: replace tuple-based lambda capture with C++20 pack capture

v4xsh:
## Description

Replace tuple-based lambda capture using `hpx::make_tuple` and invocation via `hpx::invoke_fused` with direct C++20 lambda pack capture (`...ts = HPX_FORWARD(Ts, ts)`) and `HPX_INVOKE`.

Since HPX now requires C++20, the tuple workaround used to store parameter packs in lambdas is no longer necessary.

Contributes to #6941

---

## Proposed Changes

* Replace tuple-based parameter pack capture with C++20 lambda pack capture in:

  * `libs/core/algorithms/include/hpx/parallel/task_group.hpp`
  * `libs/core/resiliency/include/hpx/resiliency/async_replicate_executor.hpp`
* Replace `hpx::invoke_fused` calls with `HPX_INVOKE`.
* Preserve original semantics:

  * Use `HPX_MOVE` in `task_group` where the lambda is executed once.
  * Avoid moving captured arguments in `bulk_async_execute` lambdas since they may be invoked multiple times.

---

This change follows the removal of the obsolete `HPX_WITH_CXX20_PERFECT_PACK_CAPTURE` feature test.
Since C++20 is now required, lambdas can capture parameter packs directly, making the previous tuple-based workaround unnecessary.

---

## Checklist

* [ ] I have added a new feature and have added tests to go along with it.
* [ ] I have fixed a bug and have added a regression test.
* [ ] I have added a test using random numbers; I have made sure it uses a seed, and that random numbers generated are valid inputs for the tests.

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/task_group.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/task_group.hpp
@@ -83,10 +83,7 @@ namespace hpx::experimental {
                     auto _(HPX_MOVE(on_exit));
 
                     hpx::detail::try_catch_exception_ptr(
-                        [&]() {
-                            HPX_INVOKE(
-                                HPX_FORWARD(F, f), HPX_FORWARD(Ts, ts)...);
-                        },
+                        [&]() { HPX_INVOKE(f, ts...); },
                         [this](std::exception_ptr e) {
                             add_exception(HPX_MOVE(e));
                         });


libs/core/algorithms/include/hpx/parallel/task_group.hpp
```diff
@@ -78,12 +78,12 @@ namespace hpx::experimental {
...
             hpx::parallel::execution::post(HPX_FORWARD(Executor, exec),
                 [this, on_exit = HPX_MOVE(on_exit), f = HPX_FORWARD(F, f),
-                    t = hpx::make_tuple(HPX_FORWARD(Ts, ts)...)]() mutable {
+                    ... ts = HPX_FORWARD(Ts, ts)]() mutable {
                     // latch needs to be released before the lambda exits
                     auto _(HPX_MOVE(on_exit));
 
                     hpx::detail::try_catch_exception_ptr(
-                        [&]() { hpx::invoke_fused(HPX_MOVE(f), HPX_MOVE(t)); },
+                        [&]() { HPX_INVOKE(HPX_MOVE(f), HPX_MOVE(ts)...); },
```
hkaiser: Wouldn't forwarding be better (i.e., correct):
```suggestion
                        [&]() { HPX_INVOKE(HPX_FORWARD(F, f), HPX_FORWARD(Ts, ts)...); },
```
?

v4xsh:
Thank you @hkaiser for the review! I've applied your suggested change — HPX_FORWARD(F, f) and HPX_FORWARD(Ts, ts)... are now used instead of HPX_MOVE in the task_group lambda.
I've also opened #6967 as a follow-up to address the async_cuda and async_mpi locations you pointed out, replacing the tuple-based keep_alive captures with C++20 pack capture there as well.
Please let me know if there are any further adjustments needed. Thanks again for the guidance!

hkaiser:
@v4xsh some tests shw compilation errors now: https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/20270/workflows/d24d2d05-ca95-4dae-a433-3b722ac053e1/jobs/491090

v4xsh:
@hkaiser Thanks for the suggestion! While testing this I noticed that CI started failing when building the `task_group_docs` example with clang.

From what I could tell, this seems to happen because `f` and `ts...` are already captured in the lambda using C++20 pack capture. Inside the lambda they are no longer forwarding references, so forwarding them again leads to the `std::forward` substitution error when the callable is a function reference (as in the example).

I reproduced the failure locally with clang by building the `task_group_docs` example. Using the captured variables directly (`HPX_INVOKE(f, ts...)`) appears to resolve the issue.

I've pushed this change now. Could you please let me know if this approach looks correct?

PR #6968, opened 2026-03-06
schedulers: fix NUMA hint not validating suspended PUs

### LINKED ISSUE #6963: shared_priority_queue_scheduler: NUMA scheduling does not handle suspended processing units
## Expected Behavior

When scheduling tasks with thread_schedule_hint_mode::numa, the scheduler should ensure that tasks are not assigned to suspended or inactive processing units (PUs). Other scheduling paths (e.g. thread_schedule_hint_mode::thread) use select_active_pu() to validate the target PU. NUMA-based scheduling should ideally apply similar validation.

## Actual Behavior

In shared_priority_queue_scheduler, the NUMA scheduling path contains the following TODO in both create_thread() and schedule_thread():

```cpp
case thread_schedule_hint_mode::numa: {
    // TODO: This case does not handle suspended PUs.
    domain_num = fast_mod(schedulehint.hint, num_domains_);
```

Unlike the thread scheduling mode, this path does not call select_active_pu() or otherwise verify that the selected PU is active. As a result, tasks scheduled using a NUMA hint could potentially be placed on suspended PUs.

1. Inspect `libs/core/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp`
2. Locate the TODO comment regarding suspended PUs in the NUMA scheduling path
3. Observe that the selected PU is not validated before queue selection

## Specifications

- HPX Version: `stable-423-g7ab0e511f3`
- Platform (compiler, OS): Arch Linux x86_64, Linux 6.6.119-1-lts66

If this behavior is unintended, I would be happy to work on a patch to address it.

v4xsh:
Fixes #6963

## Proposed Changes
  - `create_thread()`: add `select_active_pu()` call after NUMA domain selection and re-derive `domain_num`/`q_index` from the validated `thread_num`
  - `schedule_thread()`: same fix, plus replace unreachable `throw std::runtime_error` in the cross-domain else-branch with a proper fallback matching `create_thread()`
  - Add regression test that explicitly schedules tasks with `thread_schedule_hint_mode::numa` via `parallel_policy_executor`

The NUMA scheduling path in both `create_thread()` and `schedule_thread()` skipped the `select_active_pu()` call that all other hint modes (`HINT_NONE`, `HINT_THREAD`) already perform. This meant tasks could be dispatched to suspended or inactive processing units. A direct runtime reproduction is not possible via public API since `suspend_processing_unit` is internal, so the fix is validated by code inspection — the pattern applied is identical to what the other hint modes already do in both functions.

## Checklist

### DIFF
--- a/tests/regressions/threads/numa_scheduler_suspended_pu_6963.cpp
+++ b/tests/regressions/threads/numa_scheduler_suspended_pu_6963.cpp
@@ -20,6 +20,7 @@
 
 #include <atomic>
 #include <cstddef>
+#include <cstdint>
 #include <iostream>
 #include <vector>
 


libs/core/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
hkaiser: This can be removed now.

libs/core/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
```diff
@@ -399,6 +398,13 @@ namespace hpx::threads::policies {
                             q_counts_[domain_num]);
                     q_index = q_lookup_[thread_num];
```
hkaiser: This can be removed as well. The same applies to the changed code sections below.

hkaiser:
@v4xsh please have a look at the cmake-format and inspect issues reported.

v4xsh:
@hkaiser Thank you for the review.

I have removed the redundant `q_index` assignments from the if/else branches 
in both `create_thread()` and `schedule_thread()`. These were dead stores since 
`q_index` is unconditionally recomputed after `select_active_pu()`.

In `schedule_thread()`, I also inverted the condition to eliminate the empty 
if-block while preserving identical behavior — the fallback now runs directly 
when the current thread is not on the requested NUMA domain.

Additionally, fixed the `cmake-format` issue by adding the missing newline at 
the end of `tests/regressions/threads/CMakeLists.txt`.

tests/regressions/threads/numa_scheduler_suspended_pu_6963.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <atomic>
+#include <cstddef>
+#include <vector>
+
+int hpx_main()
+{
+    std::size_t const num_tasks = 64;
+    std::atomic<std::size_t> count{0};
+
+    std::vector<hpx::future<void>> futures;
```
hkaiser, line 43: In order for the test verifying that non-active cores are not used, you will have to deactive one (or more) first.

v4xsh:
@hkaiser Thank you for the suggestion.

I have updated the test to suspend PU 0 before scheduling tasks with `thread_schedule_hint_mode::numa`. The thread pool is now initialized with `enable_elasticity` to allow the use of `suspend_processing_unit()`, and PU 0 is resumed again before `finalize()` to restore the original state of the pool.

hkaiser:
@v4xsh Please have a look at the reported inspect issue.

v4xsh:
Hi @hkaiser,

I've pushed a small update addressing the inspect issue by adding the missing `<cstdint>` include for the `std::int16_t` usage in the regression test.

From the CI logs it looks like the build stops during clang-tidy while compiling `benchmark_unique.cpp` (`Subprocess killed`). Because the build stops there, a couple of later algorithm tests fail since their executables are never produced.

I'm not sure if this is related to the changes in this PR. Would rerunning the CI be sufficient here, or would you prefer that I investigate further?

PR #6970, opened 2026-03-06
Implement remove and remove_if algorithms for datapar execution

### LINKED ISSUE #2333: Implement datapar for parallel algorithms

BhoomishGupta:
Contributes to fixing #2333 

## Proposed Changes

  -Added datapar/remove.hpp implementing SIMD-accelerated sequential_remove_if and sequential_remove via tag_invoke overloads for vectorpack execution policies
  -Added detail/remove.hpp with sequential_remove_if_t and sequential_remove_t tag-fallback CPOs used by both scalar and datapar paths
  -Added datapar unit tests remove_datapar.cpp and remove_if_datapar.cpp covering simd, par_simd, simd(task), and par_simd(task) policies with both random-access and forward iterators; registered both in CMakeLists.txt

## Checklist


hkaiser:
@BhoomishGupta From what I can see, your new implementations do not take advantage of data parallelisation but simply do a per-element processing once the first matching element is found. Do you see a way of taking advantage of data parallelism for the remainder of the sequence as well?

BhoomishGupta:
> @BhoomishGupta From what I can see, your new implementations do not take advantage of data parallelisation but simply do a per-element processing once the first matching element is found. Do you see a way of taking advantage of data parallelism for the remainder of the sequence as well?

Thanks for the feedback! I will look into it.

Just as a quick reminder, my university exams are still going on right now, so my availability is quite limited. I will dive into this and work on an improved approach right after my exams finish on March 14th.

Thanks for your patience!

libs/core/executors/include/hpx/executors/execution_policy_mappings.hpp
```diff
@@ -156,4 +156,46 @@ namespace hpx::execution::experimental {
...
+        // any non-simd policy just returns itself
+        template <execution_policy ExPolicy>
+        friend constexpr decltype(auto) tag_fallback_invoke(
+            to_non_simd_t, ExPolicy&& policy) noexcept
+        {
+            static_assert(!hpx::is_vectorpack_execution_policy_v<ExPolicy>,
+                "must not be a simd (vectorpack) execution policy");
+            return std::forward<ExPolicy>(policy);
+        }
+    } to_non_simd{};
```
hkaiser: These mapping CPOs are already defined here: https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/executors/include/hpx/executors/datapar/execution_policy_mappings.hpp

libs/core/algorithms/include/hpx/parallel/algorithms/remove.hpp
```diff
@@ -306,12 +290,16 @@ namespace hpx::parallel {
...
+                    decltype(hpx::execution::experimental::to_non_simd(
+                        std::declval<std::decay_t<ExPolicy>>()));
+
                 auto f1 = [pred = HPX_FORWARD(Pred, pred),
                               proj = HPX_FORWARD(Proj, proj)](
                               zip_iterator part_begin,
                               std::size_t part_size) -> void {
                     // MSVC complains if pred or proj is captured by ref below
-                    util::loop_n<std::decay_t<ExPolicy>>(part_begin, part_size,
+                    util::loop_n<inner_policy_type>(part_begin, part_size,
```
hkaiser: Wouldn't invoking the loop with a non-simd policy defeat the whole purpose of introducing SIMD functionalities to `remove_if`?

libs/core/algorithms/include/hpx/parallel/datapar/remove.hpp
```diff
@@ -0,0 +1,177 @@
...
+            {
+                V tmp(hpx::parallel::traits::vector_pack_load<V,
+                    value_type>::aligned(first));
+
+                auto msk = HPX_INVOKE(pred, HPX_INVOKE(proj, tmp));
+
+                if (hpx::parallel::traits::none_of(msk))
+                {
+                    //no elements match
+                    if (dest != first)
```
hkaiser: Why is this conditional needed?

libs/core/algorithms/include/hpx/parallel/datapar/remove.hpp
```diff
@@ -0,0 +1,177 @@
...
+                            hpx::parallel::traits::vector_pack_store<V,
+                                value_type>::unaligned(tmp, dest);
+                        }
+                    }
+                    std::advance(dest, size);
+                }
+                else if (!hpx::parallel::traits::all_of(msk))
+                {
+                    //mixed
+                    for (std::size_t i = 0; i < size; ++i)
```
hkaiser: For `std::simd` (`std::experimental::simd`) you could utilize https://en.cppreference.com/w/cpp/experimental/simd/find_first_set.html to possibly reduce the amount of iterations needed.

hkaiser: For the other datapar backends (EVE, VC, arm simd) you may have to emulate this or simply always return `0` and `simd_width` respectively)

libs/core/executors/include/hpx/executors/execution_policy_mappings.hpp
BhoomishGupta: I moved these mapping CPOs from https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/executors/include/hpx/executors/datapar/execution_policy_mappings.hpp to `libs/core/executors/include/hpx/executors/execution_policy_mappings.hpp`

These CPOs were originally defined in the datapar header, but I moved them to the common execution policy mappings header so they are available even when datapar/SIMD is not enabled. This is needed for code paths that reference the mappings independently of datapar support.
The datapar header now just includes the common header (guarded by HPX_HAVE_DATAPAR) for compatibility.

hkaiser: Why would this be needed if datapar is not enabled?

BhoomishGupta: Yeah, I reviewed this. And you are right.
I originally fell back to the non_simd policy (and consequently had to move the execution_policy_mappings.hpp header to make it globally visible) because the f1 loop was failing to compile under the SIMD backends.

However, doing this forced the parallel chunks to evaluate the predicate sequentially, which indeed defeats the purpose of the SIMD implementation.

The root cause of the original compilation failure was that the flags array was declared as bool[]. In the std::simd backends, boolean vector arrays instantiate as proxy mask objects rather than standard packed primitives, which broke the vector_pack traits when it tried to iterate and assign. Also, the f1 lambda was explicitly taking (zip_iterator it) instead of (auto it), which prevented the datapar loop_n from passing in the vectorized chunks.

To resolve this properly, I am thinking of doing this:
1) I will change the flags array allocation from bool to std::uint8_t so it reliably packs.
2) I will change the f1 lambda parameter to auto it and revert its loop policy back to the native std::decay_t<ExPolicy> so that the heavy predicate evaluation actually vectorizes.
3) I will completely revert the execution_policy_mappings.hpp move, as to_non_simd will no longer be leaking unconditionally into the generic algorithm.

libs/core/algorithms/include/hpx/parallel/datapar/remove.hpp
BhoomishGupta: I did this for performance boost. If dest == first, then we haven't encountered any elements that needed removing yet. So, instead of copying the whole SIMD register back into the exact same memory spot, we skip the store operation completely.

libs/core/algorithms/include/hpx/parallel/datapar/remove.hpp
BhoomishGupta: Thanks for the heads up! It looks like HPX already has a generic implementation for this across all backends via `hpx::parallel::traits::find_first_of(msk)` in `libs/core/execution/include/hpx/execution/traits/vector_pack_find.hpp`

I'll update the loop to use it.

libs/core/algorithms/include/hpx/parallel/datapar/remove.hpp
hkaiser: Then you probably meant to write 
```suggestion
                    if (*dest != *first)
```
even if this would compare only one of the elements of the simd object... So no, I don't get it.

BhoomishGupta: The comparison dest != first is intentionally comparing the iterators, not the values they point to. It checks whether the destination has fallen behind the source (whether any elements have been removed earlier in the traversal.)

- When dest == first: no removals have happened yet, so dest and first point to the same memory address. Storing the SIMD register back would write the same data to the same location.

- When dest != first: earlier removals caused dest to lag behind, so we need to copy/store the data to the new (earlier) position.

Changing it to if (*dest != *first) would compare values instead of positions, which is a completely different check and wouldn't be correct here.

hkaiser: Ok, I think I understand your code now. Please disregard my comment.

libs/core/algorithms/include/hpx/parallel/algorithms/remove.hpp
BhoomishGupta: You're absolutely right. Invoking the loop with inner_policy_type (non-SIMD) was a workaround because of a compilation error with the SIMD backend. I've since identified that the root cause was the bool[] flags array and the explicit zip_iterator parameter in the lambda. I'm reverting this to use the original execution policy to ensure the predicate evaluation is properly vectorized, as detailed in my other comment

hkaiser:
@BhoomishGupta Would you be able to follow up on the compilation errors reported by the CIs?

BhoomishGupta:
> @BhoomishGupta Would you be able to follow up on the compilation errors reported by the CIs?

Working on it!

hkaiser:
@BhoomishGupta Please pay attention to the CIs. ARM is failing to compile. Also, the remove test is now failing to compile even on CIs that do not enable SIMD functionalities.

BhoomishGupta:
> @BhoomishGupta Please pay attention to the CIs. ARM is failing to compile. Also, the remove test is now failing to compile even on CIs that do not enable SIMD functionalities.

My mistake! I missed an include. Just pushed a fix. Watching the CIs now

libs/core/algorithms/include/hpx/parallel/algorithms/remove.hpp
```diff
@@ -214,6 +214,8 @@ namespace hpx {
 #else    // DOXYGEN
 
 #include <hpx/config.hpp>
+#include <hpx/algorithms/traits/projected.hpp>
+#include <hpx/execution/traits/vector_pack_conditionals.hpp>
```
hkaiser: Please alsways use only the generted module includes (except inside a module itself):
```suggestion
#include <hpx/modules/execution.hpp>
```

libs/core/algorithms/include/hpx/parallel/algorithms/remove.hpp
```diff
@@ -284,7 +269,21 @@ namespace hpx::parallel {
...
+                        inner_policy_type, Pred,
+                        hpx::parallel::traits::projected<Proj, Iter>>;
+                constexpr bool use_vectorpack_predicate =
+                    vectorpack_policy && vectorpack_predicate_callable;
+
+                using value_t = std::decay_t<
+                    typename std::iterator_traits<Iter>::value_type>;
+                using flag_t =
+                    std::conditional_t<use_vectorpack_predicate, value_t, bool>;
+                using zip_iterator = hpx::util::zip_iterator<Iter, flag_t*>;
```
hkaiser: Please refactor the code similarily to what we have done for he other algorithms. The code related to datapar should be kept fully separate from the non-datapar code paths. If datapar is disabled, e.g., the `is_vectorpack_execution_policy` you use here are not available.

libs/core/algorithms/include/hpx/parallel/algorithms/remove.hpp
```diff
@@ -299,37 +298,51 @@ namespace hpx::parallel {
...
                     auto part_begin = zip_iterator(first, flags.get());
                     auto dest = first;
                     auto part_size = count;
 
-                    using execution_policy_type = std::decay_t<ExPolicy>;
                     if (dest == get<0>(part_begin.get_iterator_tuple()))
                     {
                         // Self-assignment must be detected.
-                        util::loop_n<execution_policy_type>(
+                        util::loop_n<hpx::execution::sequenced_policy>(
```
hkaiser: Wouldn't this lose information? Why is this change needed?

libs/core/algorithms/include/hpx/parallel/algorithms/remove.hpp
```diff
@@ -396,9 +409,15 @@ namespace hpx {
             requires (
                 hpx::is_execution_policy_v<ExPolicy> &&
                 hpx::traits::is_iterator_v<FwdIter> &&
-                hpx::is_invocable_v<Pred,
-                    typename std::iterator_traits<FwdIter>::value_type
-                >
+                (
+                    hpx::parallel::traits::is_indirect_callable_v<ExPolicy,
+                        Pred, hpx::parallel::traits::projected<hpx::identity,
+                                  FwdIter>
```
hkaiser, line 401: Wouldn't the result of projecting something through `identity` result in the same?

libs/core/algorithms/include/hpx/parallel/container_algorithms/remove.hpp
```diff
@@ -555,9 +555,14 @@ namespace hpx::ranges {
...
-                    Pred, hpx::parallel::traits::projected<Proj, FwdIter>
-                >
+                (
+                    hpx::parallel::traits::is_indirect_callable_v<ExPolicy,
+                        Pred, hpx::parallel::traits::projected<Proj, FwdIter>
+                    > ||
+                    hpx::is_invocable_v<Pred,
+                        typename std::iterator_traits<FwdIter>::value_type
+                    >
+                )
```
hkaiser, line 565: Can you please explain the rationale of this change?

libs/core/execution/include/hpx/execution/traits/detail/eve/vector_pack_load_store.hpp
```diff
@@ -26,14 +27,28 @@ namespace hpx::parallel::traits {
...
-                eve::as_aligned(std::addressof(*iter), eve::cardinal_t<V>{}));
+            if constexpr (std::is_class_v<V>)
+            {
+                return eve::load(eve::as_aligned(
+                    std::addressof(*iter), eve::cardinal_t<V>{}));
+            }
+            else
+            {
+                return *iter;
+            }
```
hkaiser, line 41: Please explain this change.

hkaiser:
@BhoomishGupta ping?

BhoomishGupta:
> @BhoomishGupta ping?

Sorry for the delay. Resolving the comments asap

libs/core/algorithms/include/hpx/parallel/algorithms/remove.hpp
BhoomishGupta: Sure, I will keep to generated module includes in non-module code paths from now on.

libs/core/algorithms/include/hpx/parallel/algorithms/remove.hpp
BhoomishGupta: Agreed. I will refactor the code such that datapar-specific assumptions are out of generic algorithm paths so non-datapar builds remain valid.

libs/core/algorithms/include/hpx/parallel/algorithms/remove.hpp
BhoomishGupta: This change does not lose algorithm/data information.
That loop has a loop-carried dependency on `dest` plus self-assignment checks, so running it with vectorpack/unsequenced execution is not safe without a prefix-sum style compaction step. Using `sequenced_policy` here preserves correctness and ordering; the SIMD work remains in the predicate-evaluation phase `f1`. So the change drops vectorization in `f2`.

libs/core/algorithms/include/hpx/parallel/algorithms/remove.hpp
BhoomishGupta: You are right that identity is a semantic no-op at runtime. I used it here because this overload has no explicit projection parameter, but is_indirect_callable is defined over projected<Proj, Iter>. Using projected<identity, FwdIter> keeps the constraint aligned with the real indirect call form and preserves iterator/proxy/policy-aware argument modeling. If identity were dropped and we checked only value_type, we would lose that fidelity and could accept predicates that fail later when instantiated on actual indirect argument types.

libs/core/algorithms/include/hpx/parallel/container_algorithms/remove.hpp
BhoomishGupta: The rationale is to keep the range execution-policy overload aligned with current dispatch behavior in `remove_if`.
If we required only the policy-aware projected callable check, scalar-only predicates would be rejected at the CPO boundary, even though the implementation can still run them through the scalar predicate-evaluation path when vectorized invocation is not available.
Adding the scalar invocable check preserves backward compatibility and keeps iterator/range policy-overload behavior consistent.
This changes only whether predicate evaluation can be vectorized for a given call; it does not change remove_if semantics or output correctness.


If you prefer stricter behavior for par_simd and related policies, I can tighten this to require only the policy-aware projected callable check and fail early otherwise.

libs/core/execution/include/hpx/execution/traits/detail/eve/vector_pack_load_store.hpp
BhoomishGupta: This change was needed because vector_pack_load/store in the EVE backend is instantiated through generic datapar paths where V is not always an EVE wide type. In particular, one-lane pack types are scalar (vector_pack_type<T, 1> -> T), and the helper paths instantiate vector_pack_load/store on those scalar types as well. Unconditional eve::load/store style code is only valid for SIMD class wrappers, so the branch selects EVE APIs for SIMD types and plain dereference/assignment for scalar fallback.

hkaiser: That must be a recent change in Eve as we have not seen this before.

hkaiser:
@BhoomishGupta Could you please have an eye on the CIs?

BhoomishGupta:
> @BhoomishGupta Could you please have an eye on the CIs?

Yeah, I have been trying to resolve the Linux Crosscompile CI for a long time but it keeps failing.
Now, I have setup a local docker environment with the image taken from CI, will push a commit when Linux Crosscompile builds locally.

libs/core/algorithms/include/hpx/parallel/datapar/remove.hpp
```diff
@@ -0,0 +1,179 @@
+//  Copyright (c) 2025 Bhoomish Gupta
```
arpittkhandelwal, line 1: `//  Copyright (c) 2026 Bhoomish Gupta
`

hkaiser:
@BhoomishGupta Are you still interested in fixing this?

BhoomishGupta:
Yeah, I am.
I will get back to working on them after my university exams end (after 16th May).

hkaiser:
@BhoomishGupta ping?

BhoomishGupta:
> @BhoomishGupta ping?

Yes, I'm looking into the failing CI checks now!

libs/core/execution/include/hpx/execution/traits/detail/simd/vector_pack_load_store.hpp
```diff
@@ -56,7 +56,15 @@ namespace hpx::parallel::traits {
...
-            *iter = value;
+            if constexpr (std::is_class_v<V>)
+            {
+                value.copy_to(std::addressof(*iter),
+                    datapar::experimental::element_aligned);
+            }
+            else
+            {
+                *iter = value;
+            }
```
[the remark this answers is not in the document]
hkaiser: Copilot has a point here (the inspect CI flags this as well).

libs/core/execution/include/hpx/execution/traits/detail/simd/vector_pack_simd.hpp
```diff
@@ -44,6 +44,7 @@ namespace hpx::datapar::experimental {
 
     HPX_CXX_CORE_EXPORT using std::experimental::memory_alignment_v;
     HPX_CXX_CORE_EXPORT using std::experimental::vector_aligned;
+    HPX_CXX_CORE_EXPORT using std::experimental::element_aligned;
```
hkaiser, line 47: Do we need to introduce this for all backends?

libs/core/execution/include/hpx/execution/traits/detail/simd/vector_pack_load_store.hpp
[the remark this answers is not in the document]
BhoomishGupta: Just Fixed it!
Added the missing include

libs/core/execution/include/hpx/execution/traits/detail/simd/vector_pack_simd.hpp
BhoomishGupta: I don't think we need to.
The other backends already have `element_aligned` available.
Emulation defines it directly in `hpx::datapar::experimental` in `emulate_simd.hpp`.
SVE does it through `using namespace sve::experimental`

PR #6972, opened 2026-03-07
Fix data race in concurrent container move operations

Sanchit2662:
## Summary                                                                                                                                                                                                                                                                                                                         

I found and fixed a critical data race bug in HPX's concurrent container move operations. All three newly-implemented concurrent data structure (concurrent_vector, concurrent_unordered_map, concurrent_unordered_set) had unsynchronized access to the source object's internal state during move construction and move assignment.                                                                                                                             

## The Fix

When I examined the concurrent containers, I noticed an inconsistency: the copy constructor locked the source object's mutex, but the move constructor did not. This creates a classic data race scenario.

I brought the move operations in line with how copy operations already worked:

  Move Constructor After Fix :
```
  concurrent_vector(concurrent_vector&& other) noexcept
  {
      std::lock_guard<hpx::util::spinlock> lock(other.mutex_);  // ✓ Now locked
      data_ = HPX_MOVE(other.data_);
  }
```
  Move Assignment After Fix :
```
  concurrent_vector& operator=(concurrent_vector&& other) noexcept
  {
      if (this != &other)
      {
          std::lock(mutex_, other.mutex_);  // ✓ Lock both to prevent deadlock
          std::lock_guard<hpx::util::spinlock> lock(mutex_, std::adopt_lock);
          std::lock_guard<hpx::util::spinlock> other_lock(other.mutex_, std::adopt_lock);
          data_ = HPX_MOVE(other.data_);
      }
      return *this;
  }
```
  I used std::lock() in move assignment (not just lock_guard(mutex_)) to:
  - Acquire both locks atomically without deadlock risk
  - Match the existing pattern in copy assignment
  - Ensure consistent synchronization semantics

### Scope

  I applied this fix to all three concurrent containers:
  - concurrent_vector
  - concurrent_unordered_map
  - concurrent_unordered_set

  Each had the identical bug in both move constructor and move assignment, and each now has the identical fix.


Sanchit2662:
Hi @hkaiser , please review the changes.

hkaiser:
@Sanchit2662 Thanks for looking into this! However, I don't think that the right-hand-side of a move operation (be it the move-constructor or the move assignment) need to be locked for the operation. The right-hand-side is a rvalue after all, which by definition is not being used by anything but the calling thread.

Can you construct a test case showing that your patch is necessary?

Sanchit2662:
Hey @hkaiser, I went ahead and wrote a standalone reproducer to validate the concern.
                                                                                                                                                                 
I get what you're saying , rvalues should be exclusively owned by the calling thread. But the issue is that std::move() is just a cast to T&&. It doesn't actually transfer ownership or prevent other threads from accessing the source object. If someone calls std::move() on a named lvalue that other threads still reference, the concurrent access is very real.

I reproduced it with a minimal self-contained example replicating the same spinlock + std::vector pattern used in the concurrent containers:

  Without the fix:
image

  With the fix (locking other.mutex_ in the move constructor):
image

 What's happening: Thread A calls push_back() which acquires other.mutex_ and modifies other.data_. Thread B's move constructor directly accesses other.data_ with no lock. Both hit the same std::vector simultaneously — leading to a double-free and crash.

  I also noticed the copy constructor in these same classes already locks other.mutex_. So the locking was intentionally there for copies , the move operations  
  just missed it, which is what this patch fixes.

  I've also added regression tests to the test file on the branch to cover this going forward.

libs/core/concurrency/tests/unit/concurrent_data_structures.cpp
```diff
@@ -398,6 +398,80 @@ void test_concurrent_unordered_set_for_each_break()
...
+            {
+            }
+            dst = std::move(src);
+        });
+
+        start.store(true, std::memory_order_release);
+        writer.join();
+        mover.join();
+    }
+}
```
hkaiser, line 473: This test shows an issue that is not related to the locking of the right-hand-side of the move operations on the containers. It would fail even if we were locking it. The issue is caused by one thread working with the container while another thread is destroying the very same instance the first thread is working on. It's a design problem in the testing code, not an issue in the container implementation.

hkaiser:
> I get what you're saying , rvalues should be exclusively owned by the calling thread. But the issue is that std::move() is just a cast to T&&. It doesn't actually transfer ownership or prevent other threads from accessing the source object. If someone calls std::move() on a named lvalue that other threads still reference, the concurrent access is very real.

Sure, `std::move` itself doesn't move, but the code that uses the result of it may perform the move.

> I also noticed the copy constructor in these same classes already locks other.mutex_. So the locking was intentionally there for copies , the move operations
> just missed it, which is what this patch fixes.

The copy operations need it, while the move operations don't.

Sanchit2662:
You're right on both points. The test was flawed, and I now understand the distinction — copy needs the lock because concurrent reads are a valid use case, but moving from a live shared object is a caller contract violation regardless of locking. Closing this, thanks for the explanation.

PR #6973, opened 2026-03-07
cmake: remove HPX_WITH_CXX20_STD_IDENTITY feature test

Sahilll10:
C++20 is now the baseline for HPX, making the HPX_HAVE_CXX20_STD_IDENTITY
feature test unnecessary.

Changes:
1. Removed hpx_check_for_cxx20_std_identity function from HPX_AddConfigTest.cmake
2. Removed the feature test call from HPX_PerformCxxFeatureTests.cmake
3. Removed the now-unnecessary stdexec guard in HPX_SetupStdexec.cmake
   since std::identity is  available in C++20
4. Deleted cmake/tests/cxx20_std_identity.cpp

contributes to fixing #6941


hkaiser:
@Sahilll10 Thank you for working on this. Even if this was a minor change it is a necessary step for cleaning up the code base.

PR #6979, opened 2026-03-07
Fix set_error called on moved-from receiver in thread_pool_scheduler

Sanchit2662:
Fix use-after-move bug in thread_pool_scheduler::start

I fixed a bug where `set_error` was being called on a moved-from receiver. The original code captured `os.receiver` by move into the lambda at construction time, but if `execute()` threw, the error handler would try to move from `os.receiver` again - undefined behavior.

The fix captures `&os` by reference instead, delaying the receiver move until the lambda actually runs:

```cpp
// Before (buggy):
[receiver = HPX_MOVE(os.receiver)]() mutable { ... }
// set_error later calls HPX_MOVE(os.receiver) - already moved!

// After (fixed):
[&os]() mutable {
    hpx::execution::experimental::set_value(HPX_MOVE(os.receiver));
}
// Move only happens if execute() succeeds, set_error sees valid receiver if it throws
```

### DIFF
--- a/libs/core/executors/tests/unit/thread_pool_scheduler.cpp
+++ b/libs/core/executors/tests/unit/thread_pool_scheduler.cpp
@@ -2707,8 +2707,7 @@ int hpx_main()
 #endif
 
     test_scheduler_copy_avoidance();
-    test_completion_scheduler();
-    test_sender_receiver_set_error_on_execute_failure(); 08db910159 (Add test for set_error on execute failure in thread_pool_scheduler)
+    test_sender_receiver_set_error_on_execute_failure();
 
     return hpx::local::finalize();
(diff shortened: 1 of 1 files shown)


Sanchit2662:
@hkaiser , please give it a look whenever you are free :)

hkaiser:
Your changes cause compilation errors on some of the CIs: https://github.com/STEllAR-GROUP/hpx/actions/runs/22869021130/job/66343562438?pr=6979

hkaiser:
@Sanchit2662 there are some issues left reported by clang-format. Please have a look.

Sanchit2662:
Hi @hkaiser , I have made the required changes.

PR #6980, opened 2026-03-08
cmake: remove HPX_WITH_CXX20_CONSTEXPR_DESTRUCTOR feature test

Sahilll10:
remove HPX_WITH_CXX20_CONSTEXPR_DESTRUCTOR feature test

Fixes: 
1) Removed cmake feature test function,
2) Deleted the unused test file
3) Removed the feature test call, 
4) Simplified constexpr.hpp to define HPX_CONSTEXPR_DESTRUCTOR as constexpr unconditionally.

contributes to fixing #6941

### DIFF
--- a/libs/core/functional/include/hpx/functional/experimental/scope_success.hpp
+++ b/libs/core/functional/include/hpx/functional/experimental/scope_success.hpp
@@ -54,8 +54,7 @@ namespace hpx::experimental {
             scope_success& operator=(scope_success const&) = delete;
             scope_success& operator=(scope_success&& rhs) = delete;
 
-            constexpr ~scope_success() noexcept(
-                noexcept(this->f()))
+            constexpr ~scope_success() noexcept(noexcept(this->f()))
             {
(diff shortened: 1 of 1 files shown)


libs/core/config/include/hpx/config/constexpr.hpp
```diff
@@ -28,8 +28,4 @@
 #define HPX_HOST_DEVICE_CONSTEXPR constexpr
 #endif
 
-#if defined(HPX_HAVE_CXX20_CONSTEXPR_DESTRUCTOR)
 #define HPX_CONSTEXPR_DESTRUCTOR constexpr
```
hkaiser: I think you could go even further and completely remove this macro by editing the codebase in all places where it is currently used.

Sahilll10:
@hkaiser 

Replaced all usages of HPX_CONSTEXPR_DESTRUCTOR with constexpr directly in the three scope guard files and removed the macro definition from constexpr.hpp.

Let me know if anything else needs adjustment

PR #6981, opened 2026-03-08
Fix bound_queue cleanup in work-requesting scheduler

Sanchit2662:
## Summary

I was digging through the HPX scheduler code and spotted something in the local_workrequesting_scheduler that looked off. The per-thread       cleanup method was calling the same queue cleanup function twice instead of cleaning up both the regular queue and the bound queue.
                                                                                                                                               
  Here's what I found:

  // BEFORE (the bug):
```
  bool empty = d.queue_->cleanup_terminated(delete_all);
  empty = d.queue_->cleanup_terminated(delete_all) && empty;  // ❌ same queue again!
```
  // AFTER (my fix):
  ```
bool empty = d.queue_->cleanup_terminated(delete_all);
  empty = d.bound_queue_->cleanup_terminated(delete_all) && empty;  // ✅ now cleans bound_queue_
```
  It was a classic copy-paste error. Someone had copied the first line and just added the empty && part without actually changing d.queue_ to  
  d.bound_queue_. I verified this by looking at the correct implementation in the global cleanup_terminated() method right above it (line 641) 
  and also in the local_priority_queue_scheduler to confirm the pattern.

  Why This Matters

  When the scheduling loop is trying to figure out if a worker thread can exit, it calls cleanup_terminated(num_thread, true) to check if all  
  dead threads have been cleaned up. But because of this bug, terminated bound-priority threads were just sitting around in the bound_queue_   
  not getting cleaned up through the fast path. They'd only get cleaned up later during a slower global cleanup cycle when the scheduler went  
  deep idle.

  This meant two things:
  1. Memory leak - terminated thread objects accumulate if you're using bound threads (which fork-join executors do internally)
  2. Wrong status - the function says "cleanup done!" even though bound threads are still hanging around

  For any HPX app using fork-join executors or bound-priority tasks on the default scheduler, this would cause a slow memory bleed over time.  

  One line change, huge impact on correctness.


Sanchit2662:
@hkaiser , please give it a look whenever you are free :)

PR #6985, opened 2026-03-09
Fix critical race conditions in AGAS and improve component pinning robustness

arpittkhandelwal:
## Problem
This PR addresses several high-impact issues discovered in the AGAS and component management modules:
1. **AGAS Race Condition**: A use-after-unlock window in `addressing_service::resolve_locality` that could lead to crashes or data corruption.
2. **Migration Pinning Race**: `pinned_ptr` could be created for objects currently undergoing migration, leading to Use-After-Free scenarios.
3. **Component Type Corruption**: Bitwise masking for invalid component types returned incorrect values, potentially breaking type safety.
4. **Memory Leak**: `pinned_ptr` move assignment operator leaked memory when overwriting an existing pointer.

## Proposed Changes
- **AGAS**: Optimized locking in `resolve_locality` and added safe return paths using a static empty endpoints object.
- **Component Types**: Added robust checks for `invalid` component types and replaced magic numbers with constants.
- **Migration**: Updated `pin()` to return success status and modified `pinned_ptr` to respect it.
- **Memory Fix**: Added `delete data_` in `pinned_ptr` move assignment.

## Verification
- Built `hpx_components_base` and `hpx_agas` targets successfully.
- Verified fix for race conditions through logical code analysis and build verification.

### DIFF
--- a/libs/full/agas/src/addressing_service.cpp
+++ b/libs/full/agas/src/addressing_service.cpp
@@ -294,7 +294,9 @@ namespace hpx::agas {
 
                 HPX_THROWS_IF(ec, hpx::error::bad_parameter,
                     "addressing_service::resolve_locality", str);
-                return resolved_localities_[naming::invalid_gid];
+
+                static parcelset::endpoints_type const empty_endpoints;
+                return empty_endpoints;
             }
         }
 
@@ -312,7 +314,9 @@ namespace hpx::agas {
                     "addressing_service::resolve_locality",
                     "resolved locality insertion failed "
                     "due to a locking error or memory corruption");
-                return resolved_localities_[naming::invalid_gid];
+
+                static parcelset::endpoints_type const empty_endpoints;
+                return empty_endpoints;
             }
         }
         else if (it->second.empty() && !endpoints.empty())
--- a/libs/full/components_base/include/hpx/components_base/component_type.hpp
+++ b/libs/full/components_base/include/hpx/components_base/component_type.hpp
@@ -166,17 +166,29 @@ namespace hpx::components {
     /// \brief Return the string representation for a given component type id
     HPX_EXPORT std::string get_component_type_name(component_type type);
 
+    inline constexpr int component_type_mask = 0x3FF;
+    inline constexpr int component_type_shift = 10;
+
     /// The lower short word of the component type is the type of the component
     /// exposing the actions.
     constexpr component_type get_base_type(component_type t) noexcept
     {
-        return static_cast<component_type>(t & 0x3FF);
+        if (t == to_int(component_enum_type::invalid))
+        {
+            return to_int(component_enum_type::invalid);
+        }
+        return static_cast<component_type>(t & component_type_mask);
     }
 
     /// The upper short word of the component is the actual component type
     constexpr component_type get_derived_type(component_type t) noexcept
     {
-        return static_cast<component_type>((t >> 10) & 0x3FF);
+        if (t == to_int(component_enum_type::invalid))
+        {
+            return to_int(component_enum_type::invalid);
+        }
+        return static_cast<component_type>(
+            (t >> component_type_shift) & component_type_mask);
     }
 
     /// A component derived from a base component exposing the actions needs to
--- a/libs/full/components_base/include/hpx/components_base/pinned_ptr.hpp
+++ b/libs/full/components_base/include/hpx/components_base/pinned_ptr.hpp
@@ -13,62 +13,10 @@
 #include <hpx/components_base/traits/component_pin_support.hpp>
 #include <hpx/modules/naming_base.hpp>
 
+#include <type_traits>
+
 namespace hpx::components {
 
-    ///////////////////////////////////////////////////////////////////////////
-    namespace detail {
-
-        class pinned_ptr_base
-        {
-        public:
-            constexpr pinned_ptr_base() noexcept = default;
-
-            constexpr explicit pinned_ptr_base(
-                naming::address_type lva) noexcept
-              : lva_(lva)
-            {
-            }
-
-            pinned_ptr_base(pinned_ptr_base const&) = delete;
-            pinned_ptr_base(pinned_ptr_base&&) = delete;
-            pinned_ptr_base& operator=(pinned_ptr_base const&) = delete;
-            pinned_ptr_base& operator=(pinned_ptr_base&&) = delete;
-
-            virtual ~pinned_ptr_base() = default;
-
-        protected:
-            naming::address_type lva_ = nullptr;
-        };
-
-        template <typename Component>
-        class pinned_ptr final : public pinned_ptr_base
-        {
-        public:
-            explicit pinned_ptr(naming::address_type lva) noexcept
-              : pinned_ptr_base(lva)
-            {
-                HPX_ASSERT(nullptr != this->lva_);
-
-                // pin associated component instance
-                traits::component_pin_support<Component>::pin(
-                    get_lva<Component>::call(this->lva_));
-            }
-
-            pinned_ptr(pinned_ptr const&) = delete;
-            pinned_ptr(pinned_ptr&&) = delete;
-            pinned_ptr& operator=(pinned_ptr const&) = delete;
-            pinned_ptr& operator=(pinned_ptr&&) = delete;
-
-            ~pinned_ptr() override
-            {
-                // unpin associated component instance
-                traits::component_pin_support<Component>::unpin(
-                    get_lva<Component>::call(this->lva_));
-            }
-        };
-    }    // namespace detail
-
-    ///////////////////////////////////////////////////////////////////////////
     class pinned_ptr
     {
         template <typename T>
@@ -76,10 +24,32 @@ namespace hpx::components {
         {
         };
 
+        using unpin_type = void (*)(naming::address_type);
+
+        template <typename Component>
+        static void unpin_impl(naming::address_type lva)
+        {
+            if (lva != nullptr)
+            {
+                traits::component_pin_support<Component>::unpin(
+                    get_lva<Component>::call(lva));
+            }
+        }
+
         template <typename Component>
         pinned_ptr(naming::address_type lva, id<Component>)
-          : data_(new detail::pinned_ptr<Component>(lva))
+          : lva_(lva)
+          , unpin_(&unpin_impl<Component>)
         {
+            HPX_ASSERT(nullptr != this->lva_);
+
+            // pin associated component instance
+            if (!traits::component_pin_support<Component>::pin(
+                    get_lva<Component>::call(this->lva_)))
+            {
+                this->lva_ = nullptr;
+                this->unpin_ = nullptr;
+            }
         }
 
     public:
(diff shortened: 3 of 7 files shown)


libs/full/agas/src/addressing_service.cpp
```diff
@@ -290,11 +291,9 @@ namespace hpx::agas {
                 std::string const str = hpx::util::format(
                     "couldn't resolve the given target locality ({})", gid);
 
-                l.unlock();
```
hkaiser, line 293: This has to stay as otherwise the exception handling will raise an assertion.

libs/full/agas/src/addressing_service.cpp
```diff
@@ -306,13 +305,11 @@ namespace hpx::agas {
             if (HPX_UNLIKELY(!util::insert_checked(
                     resolved_localities_.emplace(gid, endpoints), it)))
             {
-                l.unlock();
```
hkaiser, line 309: This has to stay as well.

libs/full/components_base/include/hpx/components_base/server/migration_support.hpp
```diff
@@ -149,7 +149,11 @@ namespace hpx::components {
                     data_->pin_count_ != 0 ||
                         (!started_migration_ && !was_marked_for_migration_));
                 ++data_->pin_count_;
+                return true;
             }
+
+            intrusive_ptr_release(data_.get());    // release keep alive
```
hkaiser, line 154: This is an interesting catch, thanks!

hkaiser:
Please note that your patch causes compilation problems: https://github.com/STEllAR-GROUP/hpx/actions/runs/22863300663/job/66328541349?pr=6985

libs/full/agas/src/addressing_service.cpp
arpittkhandelwal: Thanks for catching that! The manual l.unlock() calls have been restored in the exception handling paths to properly manage the lock state and avoid assertions.

libs/full/agas/src/addressing_service.cpp
arpittkhandelwal: Thanks again for the review! I have restored this l.unlock() call as well before the exception is thrown, ensuring the lock state remains valid consistently just like in the previous error path above.
I just pushed these changes!

libs/full/components_base/include/hpx/components_base/server/migration_support.hpp
arpittkhandelwal: Thank you so much! I'm glad this helps make things fully robust against the migration race conditions. Appreciate the review!

libs/core/algorithms/include/hpx/parallel/algorithms/sort_by_key.hpp
```diff
@@ -124,6 +124,18 @@ namespace hpx::parallel {
...
+            auto operator()(
+                hpx::util::detail::operator_brackets_proxy<Iterator> const&
+                    proxy) const
+                -> decltype(hpx::get<0>(
+                    static_cast<typename Iterator::reference>(proxy)))
+            {
+                // operator_brackets_proxy is implicitly convertible to reference
+                return hpx::get<0>(
+                    static_cast<typename Iterator::reference>(proxy));
+            }
```
hkaiser: What is the rationale of this change? Does it relate to the AGAS race you tried to fix?

libs/core/algorithms/include/hpx/parallel/util/compare_projected.hpp
```diff
@@ -64,6 +64,8 @@ namespace hpx::parallel::util {
         typename Proj2>
     struct compare_projected<Compare, Proj1, Proj2>
     {
+        using is_transparent = void;
```
hkaiser: While the changes to this file may be valid, they have no relation to the AGAS fixes proposed.

libs/full/components_base/src/component_type.cpp
```diff
@@ -162,12 +162,14 @@ namespace hpx::components {
         {
             result = "component_invalid";
         }
-        else if ((type < to_int(hpx::components::component_enum_type::last)) &&
+        else if (type >= 0 &&
+            (type < to_int(hpx::components::component_enum_type::last)) &&
```
hkaiser, line 169: Please factor out repeated invocations, add a
```
auto derived_type = get_derived_type(type);
```

libs/core/algorithms/include/hpx/parallel/algorithms/sort_by_key.hpp
arpittkhandelwal: This change was a temporary fix for a C++20 compilation error I encountered in the CI logs while testing. However, I agree it is unrelated to the AGAS race/pinning fixes. I have already reverted this (and the is_transparent changes) in my latest push to keep this PR strictly focused. I'll move the algorithm fixes to a separate PR.

libs/core/algorithms/include/hpx/parallel/util/compare_projected.hpp
arpittkhandelwal: I included these changes to address C++20 compilation issues I encountered while testing, but I agree they are unrelated to the core AGAS and pinning fixes in this PR. I have already reverted this file (along with the changes in sort_by_key.hpp) in my local workspace. The upcoming push will remove these changes to keep the PR strictly focused on the AGAS and component management logic.

libs/full/components_base/src/component_type.cpp
arpittkhandelwal: Thank you for the suggestion. I have refactored get_component_type_name to factor out the repeated calls to get_derived_type(type) and get_base_type(type) into local variables (derived_type and base_type). This simplifies the logic and improves efficiency by avoiding redundant bitwise operations. This change is included in the updated commit

libs/full/agas/src/addressing_service.cpp
```diff
@@ -292,9 +292,8 @@ namespace hpx::agas {
 
                 l.unlock();
 
-                HPX_THROWS_IF(ec, hpx::error::bad_parameter,
-                    "addressing_service::resolve_locality", str);
-                return resolved_localities_[naming::invalid_gid];
+                static parcelset::endpoints_type const empty_endpoints;
```
hkaiser, line 298: What's your rationale to remove the error handling? What you had was a perfectly fine PR, why introduce changes all over again that are not necessary to fix the issue at hand?

libs/full/components_base/include/hpx/components_base/server/migration_support.hpp
```diff
@@ -149,7 +148,13 @@ namespace hpx::components {
                     data_->pin_count_ != 0 ||
                         (!started_migration_ && !was_marked_for_migration_));
                 ++data_->pin_count_;
+                return true;
             }
+
+            l.unlock();
```
hkaiser: This unlock is strictly speaking not required as the destructor of the `unique_lock` will do exactly that. Did you see any issues with the lock being held while decrementing the reference count?

libs/full/components_base/include/hpx/components_base/component_type.hpp
```diff
@@ -30,6 +30,12 @@ namespace hpx::components {
     {
         invalid = naming::address::component_invalid,
 
+        /// Mask to extract the base component type
+        mask = 0x3FF,
+
+        /// Shift to extract the derived component type
+        shift = 10,
```
hkaiser: Hmmm, the shift may not belong here as it techincally is unrelated to the enumerator. May I suggest to leave this as an `inline constexpr int` instead?

libs/full/components_base/include/hpx/components_base/pinned_ptr.hpp
```diff
@@ -38,6 +38,8 @@ namespace hpx::components {
 
         protected:
             naming::address_type lva_ = nullptr;
+
+            friend class hpx::components::pinned_ptr;
```
hkaiser: What's the rationale of this change?

libs/full/components_base/include/hpx/components_base/pinned_ptr.hpp
```diff
@@ -50,8 +52,11 @@ namespace hpx::components {
                 HPX_ASSERT(nullptr != this->lva_);
 
                 // pin associated component instance
-                traits::component_pin_support<Component>::pin(
-                    get_lva<Component>::call(this->lva_));
+                if (!traits::component_pin_support<Component>::pin(
+                        get_lva<Component>::call(this->lva_)))
+                {
+                    this->lva_ = nullptr;
```
hkaiser: If you're invalidating the base class here you will have to check for `this->lva_ != nullptr` in the destructor.

hkaiser: Would it be better to move the `lva_` member from the base class to here? Or even more -getting rid of the `pinned_ptr_base` class altogether? It looks like a historical artifact after all and has no function anymore (but please correct me if I'm wrong).

libs/full/agas/src/addressing_service.cpp
arpittkhandelwal: You're completely right. I've restored the original HPX_THROWS_IF error handling while maintaining the safe return of empty_endpoints to prevent the use-after-unlock race.

libs/full/components_base/include/hpx/components_base/server/migration_support.hpp
arpittkhandelwal: I hadn't seen any explicit issues with the lock being held, but you're right that the `unique_lock` destructor handles it safely, so I've removed the explicit `l.unlock()` call before decrementing the reference count.

libs/full/components_base/include/hpx/components_base/component_type.hpp
arpittkhandelwal: I've removed `mask` and `shift` from the `component_enum_type` and instead declared them as `inline constexpr int` constants (`component_type_mask` and `component_type_shift`) so they no longer logically mix with the enumerators.

hkaiser:
@arpittkhandelwal I was hoping that this would fix the occasional test error we see. Unfortunately this is not the case (https://cdash.rostam.cct.lsu.edu/test/37210760). We will merge this in any case once the CI has cycled as this PR fixes a serious issue in the first place.

arpittkhandelwal:
> @arpittkhandelwal I was hoping that this would fix the occasional test error we see. Unfortunately this is not the case (https://cdash.rostam.cct.lsu.edu/test/37210760). We will merge this in any case once the CI has cycled as this PR fixes a serious issue in the first place.

Thanks sir! I also noticed that failure on the CI dashboard. After testing locally, it appears that the segmentation fault only occurs when the test runs with the MPI/LCW (Libfabric) network backend (`hpx.parcel.bootstrap=lcw`). With the default TCP backend, `migrate_component` passes cleanly on my local setup.

It looks like the LCW failure during `decode_message_with_chunks` might be an independent, pre-existing issue rather than something introduced by this PR.

Glad to hear the lock fixes can still be merged. Please let me know if there’s anything else I can help with from my side.

hkaiser:
Thank you again for working on this!

arpittkhandelwal:
> Thank you again for working on this!

Thank you for the review and for merging the PR, sir

Issue #6978, opened 2026-03-07
shared_priority_queue_scheduler: possible stale (domain_num, q_index) after select_active_pu() in schedule_thread() `none` path


v4xsh:
While working on the fix for #6963 in `shared_priority_queue_scheduler`, I spent some additional time reviewing the scheduling logic and noticed a potential inconsistency in the `schedule_thread()` implementation for `thread_schedule_hint_mode::none`.

In the `none` path, `domain_num` and `q_index` are derived from the initial `thread_num`, but `thread_num` may later be modified by `select_active_pu(thread_num, allow_fallback)`.

Relevant section (simplified):

```cpp
else /*(round_robin_)*/
{
    domain_num = d_lookup_[thread_num];
    q_index = q_lookup_[thread_num];

    thread_num =
        numa_holder_[domain_num]
            .thread_queue(q_index)
            ->worker_next(num_workers_);
}

thread_num = select_active_pu(thread_num, allow_fallback);
break;
````

Later, the task is dispatched using:

```cpp
numa_holder_[domain_num].thread_queue(q_index)->schedule_thread(thrd, priority, false);
```

Since `select_active_pu()` may redirect to a different PU (for example when elasticity is enabled and the requested PU is suspended or inactive), `domain_num` and `q_index` may become stale. This affects all three sub-paths of the `none` case:

* **Cross-pool injection (`local_num == -1`)** – `domain_num` and `q_index` remain `0`, but `select_active_pu()` may redirect to a different PU.
* **`!round_robin_` (assign_parent)** – `domain_num` and `q_index` are derived from `local_num`, but `select_active_pu()` may redirect to a different PU.
* **`round_robin_`** – `worker_next()` first changes `thread_num`, and `select_active_pu()` may change it again.

In each case, the task can end up being pushed to the queue corresponding to the original `(domain_num, q_index)` rather than the queue associated with the final `thread_num`.

Other scheduling paths recompute these values after calling `select_active_pu()`. For example:

* `create_thread()` recomputes `(domain_num, q_index)` after `select_active_pu()`
* the `thread_schedule_hint_mode::thread` and `thread_schedule_hint_mode::numa` paths in `schedule_thread()` also recompute them

By analogy with `create_thread()`’s `none` path (which is structurally similar and does recompute after `select_active_pu()`), this appears to be an unintentional omission.

Would it make sense to recompute

```cpp
domain_num = d_lookup_[thread_num];
q_index = q_lookup_[thread_num];
[...]

PR #6990, opened 2026-03-10
schedulers: fix stale domain_num/q_index after select_active_pu() in schedule_thread() none path

v4xsh:
Fixes **#6978**

## Proposed Changes
- `schedule_thread()` none path: recompute `domain_num` and `q_index` from the updated `thread_num` after `select_active_pu()` — mirroring the pattern already used in `create_thread()` none path and `schedule_thread()` thread/numa hint paths
- Add regression test `schedule_hint_none_stale_queue_6978` that floods `hint_mode::none` tasks while cycling PUs through suspend/resume to exercise the sleeping-state redirect window](schedulers: fix stale domain_num/q_index after select_active_pu() in schedule_thread() none path)

In `schedule_thread()` with `hint_mode::none`, `domain_num` and `q_index` are derived from `thread_num` before `select_active_pu()` is called. When elasticity is enabled and the target PU is in `hpx::state::sleeping` during a suspend transition, `select_active_pu()` redirects `thread_num` to a different PU — leaving `domain_num` and `q_index` stale. Tasks are then pushed to the wrong queue. The fix is a 2-line recompute after the call, consistent with every other hint path in the scheduler.

## Checklist

### DIFF
--- a/libs/core/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
+++ b/libs/core/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
@@ -767,6 +767,10 @@ namespace hpx::threads::policies {
                         debug::threadinfo<threads::thread_id_ref_type*>(&thrd));
                 }
                 thread_num = select_active_pu(thread_num, allow_fallback);
+                // cppcheck-suppress redundantAssignment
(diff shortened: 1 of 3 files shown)


hkaiser (approved):
LGTM, thanks! Nice analysis!

hkaiser:
@v4xsh Thanks for working on this!

PR #6991, opened 2026-03-10
Algorithm: Implementing hpx::iota and hpx::ranges::iota (#6969)

### LINKED ISSUE #6969: Implementation of `hpx::ranges::iota`
Hi,

I've been looking into getting more involved with HPX and noticed that `ranges::iota` (from Issue #5654) seems to be a great starting point. I have briefly explored the codebase (looking at standard algorithms like `reduce` and `fill`) to understand the use of CPOs (`tag_fallback_invoke`) and the dispatching mechanism.

I would like to contribute the implementation of `hpx::iota` and `hpx::ranges::iota` along with their corresponding C++23 return type (`iota_result`). Currently, HPX seems doesn't have its own `iota` implementation, and existing tests rely on `std::iota` from the standard library.

Before I create a PR, I wanted to check:

**1. Is this feature still needed?**
I noticed HPX hasn't yet implemented `hpx::iota` (neither the iterator-based version nor the `ranges` version).

**2. Sequential Implementation Plan:**
I plan to implement the standard sequential version first, utilizing C++20 concepts (`std::weakly_incrementable`, `std::ranges::output_range`, etc.) and encapsulating it within a CPO (`hpx::ranges::iota_t`).

**3. Thoughts on Parallelization:**
Currently, the C++ standard does not define a parallel version of `iota`, likely due to the sequential dependency inherent in repeated `++value` operations.

However, I believe we can offer a parallel implementation as an HPX extension for types that support random access.
*   **Concept:** If the value type `T` supports $O(1)$ addition (e.g., integers or types satisfying `std::random_access_iterator`), we can determine the starting value of any sub-range independently (`start_val + offset`).
*   **Strategy:** This allows us to partition the range into chunks and fill them in parallel using HPX's `ExecutionPolicy`. For types that only satisfy `std::weakly_incrementable` (where $O(1)$ jumping isn't possible), the implementation would naturally fall back to sequential execution.

I plan to start with a sequential implementation PR and then we can discuss the parallel extension if there's interest.

Looking forward to your feedback!

anfsity:
Fixes #6969

## Proposed Changes

  - Implemented the core parallel and sequential backend for `iota` in `hpx::parallel::detail::iota`.
  - Added the public CPO `hpx::iota` that dispatches to various execution policies correctly.
  - Added unit tests in `iota.cpp` and `iota_range.cpp` that test both `std::forward_iterator` and `std::random_access_iterator` concepts across sequential, parallel, and asynchronous execution policies.

Nothing.

## Checklist

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/iota.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/iota.hpp
@@ -95,14 +95,15 @@ namespace hpx {
 #include <hpx/parallel/util/detail/sender_util.hpp>
 
 #include <concepts>
+#include <iterator>
 #include <type_traits>
 
 namespace hpx {
 
     ///////////////////////////////////////////////////////////////////////////
     // CPO for hpx::iota
-    HPX_CXX_CORE_EXPORT inline constexpr struct iota_fn final
-      : hpx::detail::tag_parallel_algorithm<iota_fn>
+    HPX_CXX_CORE_EXPORT inline constexpr struct iota_t final
+      : hpx::detail::tag_parallel_algorithm<iota_t>
     {
     private:
         // parallel
@@ -116,7 +117,7 @@ namespace hpx {
         // clang-format on
         friend hpx::parallel::util::detail::algorithm_result_t<Expolicy>
         tag_fallback_invoke(
-            iota_fn, Expolicy&& policy, FwdIter first, Sent last, T value)
+            iota_t, Expolicy&& policy, FwdIter first, Sent last, T value)
         {
             using result_type =
                 hpx::parallel::util::detail::algorithm_result<Expolicy>::type;
@@ -129,13 +130,9 @@ namespace hpx {
         // sequential
         template <std::input_or_output_iterator FwdIter,
             std::sentinel_for<FwdIter> Sent, std::weakly_incrementable T>
-        // clang-format off
-            requires(
-                std::indirectly_writable<FwdIter, T const&>
-            )
-        // clang-format on
+            requires(std::indirectly_writable<FwdIter, T const&>)
         friend void tag_fallback_invoke(
-            iota_fn, FwdIter first, Sent last, T value)
+            iota_t, FwdIter first, Sent last, T value)
         {
             hpx::parallel::detail::iota<FwdIter>().call(
                 hpx::execution::seq, first, last, value);
--- a/libs/core/algorithms/include/hpx/parallel/container_algorithms/iota.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/container_algorithms/iota.hpp
@@ -1,0 +1,247 @@
+//  Copyright (c) 2026 Anfsity
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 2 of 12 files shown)


hkaiser:
Please have a look at the reported clang-format and inspect errors. Also, please add the new tests to the list of targets for the CI (https://github.com/STEllAR-GROUP/hpx/tree/master/.circleci, https://github.com/STEllAR-GROUP/hpx/tree/master/.github/test-targets)

anfsity:
Sorry about the oversight! I've fixed it. However, CI encountered a Windows error that I don't quite get. What does it mean?

hkaiser:
> Sorry about the oversight! I've fixed it. However, CI encountered a Windows error that I don't quite get. What does it mean?

That's a github issue. I have re-triggered the CI.

libs/core/algorithms/include/hpx/parallel/algorithms/iota.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+#include <concepts>
+#include <iterator>
+#include <type_traits>
+
+namespace hpx {
+
+    ///////////////////////////////////////////////////////////////////////////
+    // CPO for hpx::iota
+    HPX_CXX_CORE_EXPORT inline constexpr struct iota_fn final
```
hkaiser: We usually call the CPOs `<name>_t`
```suggestion
    HPX_CXX_CORE_EXPORT inline constexpr struct iota_t final
```

libs/core/algorithms/include/hpx/parallel/container_algorithms/iota.hpp
```diff
@@ -0,0 +1,255 @@
...
+                hpx::util::begin(r), hpx::util::end(r), value);
+        }
+
+        template <std::input_or_output_iterator FwdIter,
+            std::sentinel_for<FwdIter> Sent, std::weakly_incrementable T>
+        // clang-format off
+        requires (
+            std::indirectly_writable<FwdIter, T const&>
+        )
+        // clang-format on
```
hkaiser: Let's try to minimize `clang-format off`
```suggestion
        template <std::input_or_output_iterator FwdIter,
            std::sentinel_for<FwdIter> Sent, std::weakly_incrementable T>
        requires (std::indirectly_writable<FwdIter, T const&>)
```

libs/core/algorithms/include/hpx/parallel/algorithms/iota.hpp
anfsity: Done.

libs/core/algorithms/include/hpx/parallel/container_algorithms/iota.hpp
anfsity: Fixed as suggested.

hkaiser:
@anfsity Thanks again for working on this!

hkaiser:
@anfsity After this was merged, we now see compilation errors on some of the additional CIs that don't run on PRs. Could you please have a look: https://cdash.rostam.cct.lsu.edu/viewBuildError.php?buildid=42211?

anfsity:
Ok, I'll take a look.

hkaiser:
> Ok, I'll take a look.

NVM, I handled it in the context of another similar problem: https://github.com/STEllAR-GROUP/hpx/pull/7011

anfsity:
I understand. I apologize for my oversight.

PR #6993, opened 2026-03-10
Extending the index_queue_spawning facility to support returning values

hkaiser:
@waleed7715 this adds support for more efficiently handling thread spawning in algorithms if the iteration function returns something other than `void`


hkaiser:
@waleed7715 this should be usable for your purposes now.

PR #6994, opened 2026-03-10
Small Code Quality Boost: Addressing long-standing FIXMEs and TODOs

aneek22112007-tech:
Hey HPX team! 

I spent some time looking through the repository for small, low-risk areas where I could help tidy things up. I've found three separate spots in the compute_local and algorithms namespaces that had pending FIXME or TODO tags, or just needed a bit of logical synchronization. 

**What's in this PR?**
* **Encapsulation Improvement**: I moved the detail::iterator constructor for compute::vector into the private section. This prevents improper external instantiation (as requested by a FIXME) while keeping it accessible to the vector class itself via a friend declaration.
* * **API Completeness**: I've implemented front() and back() (both const and mutable) for hpx::compute::vector. It's a small change but makes the class feel a lot more like the std::vector we all know and love.
* * **Logic Sync**: I noticed that prefetching_iterator had a slightly inconsistent operator!=. I've updated it to reliably delegate to the operator== logic, ensuring that both index and base iterator state are always considered together.
**Verification**:
I've verified these changes by building and passing the existing foreach_prefetching_test. I also wrote a safe internal verification script to confirm that front()/back() work as expected and that the new iterator privacy is correctly enforced by the compiler.

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/util/prefetching.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/util/prefetching.hpp
@@ -169,7 +169,7 @@ namespace hpx::parallel::util {
             // FIXME: should the base iterators be compared too?
             bool operator!=(prefetching_iterator const& rhs) const
             {
-                return !(*this == rhs);
+                return idx_ != rhs.idx_;
             }
             bool operator>(prefetching_iterator const& rhs) const
             {
--- a/libs/core/compute_local/include/hpx/compute_local/detail/iterator.hpp
+++ b/libs/core/compute_local/include/hpx/compute_local/detail/iterator.hpp
@@ -17,11 +17,6 @@
 #include <iterator>
 #include <type_traits>
 
-namespace hpx::compute {
-    template <typename T, typename Allocator>
-    class vector;
-}
-
 namespace hpx::compute::detail {
 
     HPX_CXX_CORE_EXPORT template <typename T, typename Allocator>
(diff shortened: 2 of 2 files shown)


libs/core/algorithms/include/hpx/parallel/util/prefetching.hpp
```diff
@@ -169,7 +169,7 @@ namespace hpx::parallel::util {
             // FIXME: should the base iterators be compared too?
             bool operator!=(prefetching_iterator const& rhs) const
             {
-                return idx_ != rhs.idx_;
+                return idx_ != rhs.idx_ || base_ != rhs.base_;
```
hkaiser: Wouldn't it be better to say:
```suggestion
                return !(*this == rhs);
```

aneek22112007-tech: Thanks so much for the detailed review and catching that inconsistency! I've updated the comparison to use `!(*this == rhs)` as suggested. 

Just wanted to add how much I appreciate the encouragement, especially since I'm still relatively new to the codebase. It means a lot to have maintainers who are so patient and helpful—thank you! Let me know if everything looks good or if there are any other tweaks needed!

hkaiser (dismissed):
LGTM, thanks!

hkaiser:
@aneek22112007-tech Thanks for working on this!

PR #6997, opened 2026-03-11
tests: fix self-copy UB in move algorithm tests causing occasional segfault

### LINKED ISSUE #6975: Fix failing unit tests related to move algorithm
We're seeing occasional test failures for our tests related to the move algorithm (see for instance: https://cdash.rostam.cct.lsu.edu/test/35941854). We should try to identify the problem and fix it.

Sahilll10:
Fixes #6975

## Problem

The move algorithm tests contained a self-copy undefined behavior:
```cpp
//copy contents of d back into c for testing
std::copy(std::begin(d), std::end(d), std::begin(d));  // d into d, not c
```

The comment says "copy d back into **c**" but the destination is
`std::begin(d)`. This is undefined behavior — `std::copy` does not
permit the destination to overlap with the source range.

Under gcc-13 release builds, the compiler lowers `std::copy` on
trivially copyable types to `memcpy` (which forbids overlap), producing
the observed segfault with seed `1772660164` and `--hpx:threads=4`.

The async variants also copied `d` into `c` then checked `c == d` —
tautological since `c` was just set equal to `d`. Stale lambda
captures `[&d, &c]` are cleaned up as well.

All `hpx::copy` test files were checked and are not affected.

## Files Changed
- `libs/core/algorithms/tests/unit/algorithms/move.cpp`
- `libs/core/algorithms/tests/unit/algorithms/move_sender.cpp`
- `libs/core/algorithms/tests/unit/container_algorithms/move_range.cpp`

### DIFF
--- a/libs/core/algorithms/tests/unit/algorithms/move.cpp
+++ b/libs/core/algorithms/tests/unit/algorithms/move.cpp
@@ -33,8 +33,8 @@ void test_move(IteratorTag)
     std::iota(std::begin(c), std::end(c), gen());
     hpx::move(iterator(std::begin(c)), iterator(std::end(c)), std::begin(d));
 
-    //copy contents of d back into c for testing
-    std::copy(std::begin(d), std::end(d), std::begin(d));
+    // copy contents of d back into c for testing
+    std::copy(std::begin(d), std::end(d), std::begin(c));
 
     std::size_t count = 0;
     HPX_TEST(std::equal(std::begin(c), std::end(c), std::begin(d),
@@ -61,8 +61,8 @@ void test_move(ExPolicy policy, IteratorTag)
     hpx::move(
         policy, iterator(std::begin(c)), iterator(std::end(c)), std::begin(d));
 
-    //copy contents of d back into c for testing
-    std::copy(std::begin(d), std::end(d), std::begin(d));
+    // copy contents of d back into c for testing
+    std::copy(std::begin(d), std::end(d), std::begin(c));
 
     std::size_t count = 0;
     HPX_TEST(std::equal(std::begin(c), std::end(c), std::begin(d),
--- a/libs/core/algorithms/tests/unit/container_algorithms/move_range.cpp
+++ b/libs/core/algorithms/tests/unit/container_algorithms/move_range.cpp
@@ -32,8 +32,8 @@ void test_move(IteratorTag)
     std::iota(std::begin(c), std::end(c), std::rand());
     hpx::ranges::move(c, std::begin(d));
 
-    //copy contents of d back into c for testing
-    std::copy(std::begin(d), std::end(d), std::begin(d));
+    // copy contents of d back into c for testing
+    std::copy(std::begin(d), std::end(d), std::begin(c));
 
     std::size_t count = 0;
     HPX_TEST(std::equal(std::begin(c), std::end(c), std::begin(d),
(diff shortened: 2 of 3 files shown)


libs/core/algorithms/tests/unit/algorithms/move.cpp
```diff
@@ -33,9 +33,6 @@ void test_move(IteratorTag)
     std::iota(std::begin(c), std::end(c), gen());
     hpx::move(iterator(std::begin(c)), iterator(std::end(c)), std::begin(d));
 
-    //copy contents of d back into c for testing
-    std::copy(std::begin(d), std::end(d), std::begin(d));
```
hkaiser, line 37: I think we still have to back-copy things as formally `hpx::move` has invalidated the `c`.

Sahilll10: Hi @hkaiser 

You're right. After `hpx::move(c → d)`, `c` is formally in a moved-from state (valid but unspecified values), so comparing `c` directly against `d` is not meaningful. The back-copy is necessary to restore `c` before the `std::equal` check.

 The comment always said "copy d back into **c**" but the code passed `std::begin(d)` as the destination instead of `std::begin(c)`. I believe The correct fix is:

```cpp
// WRONG (original) — copies d into itself, UB
std::copy(std::begin(d), std::end(d), std::begin(d));

// CORRECT — copies d back into c as the comment intended
std::copy(std::begin(d), std::end(d), std::begin(c));
```

I plan to :
 Restore `std::copy(std::begin(d), std::end(d), std::begin(c))` in all sync test functions across `move.cpp`, `move_sender.cpp`, and `move_range.cpp` — fixing the destination typo (`d` → `c`)

Does this approach look correct to you?

Happy to push the updated commit as soon as I have your confirmation.

hkaiser: That's the correct fix, yes. Thanks!

libs/core/algorithms/tests/unit/algorithms/move.cpp
```diff
@@ -87,10 +87,8 @@ void test_move_async(ExPolicy p, IteratorTag)
     auto f = hpx::move(
         p, iterator(std::begin(c)), iterator(std::end(c)), std::begin(d));
 
-    hpx::future<void> g = f.then([&d, &c](hpx::future<void> f) {
-        HPX_TEST(!f.has_exception());
-        std::copy(std::begin(d), std::end(d), std::begin(c));
-    });
```
hkaiser, line 93: The array needs to be copied back here as well.

libs/core/algorithms/tests/unit/container_algorithms/move_range.cpp
```diff
@@ -84,10 +84,8 @@ void test_move_async(ExPolicy p, IteratorTag)
     std::iota(std::begin(c), std::end(c), std::rand());
 
     auto f = hpx::ranges::move(p, c, std::begin(d));
-    hpx::future<void> g = f.then([&d, &c](hpx::future<void> f) {
-        HPX_TEST(!f.has_exception());
-        std::copy(std::begin(d), std::end(d), std::begin(c));
-    });
```
hkaiser, line 90: And here also.

hkaiser:
@Sahilll10 the issue with move failing occasionally is still around, unfortunately: see for instance https://cdash.rostam.cct.lsu.edu/viewTest.php?onlyfailed&buildid=43017. Could you have another look?

Sahilll10:
> @Sahilll10 the issue with move failing occasionally is still around, unfortunately: see for instance https://cdash.rostam.cct.lsu.edu/viewTest.php?onlyfailed&buildid=43017. Could you have another look?

Sorry for the late reply @hkaiser 

I have been working on my first draft for the proposal as the Deadline approaches.
I will look into this issue as soon as i complete it .

Thank you for your patience.

PR #6998, opened 2026-03-12
Fix incorrect Python 3 print usage in hpxrun.py.in

arpittkhandelwal:
## Description

The hpxrun.py.in script contained several instances where the print function was used with syntax that is incorrect for Python 3 when directing output to sys.stderr.

Specifically, the calls were written as:

print(msg, sys.stderr)

In Python 3, this treats sys.stderr as a second positional argument, which results in both the message and the string representation of the stderr object being printed to sys.stdout instead of the standard error stream.

This PR updates these occurrences to use the correct Python 3 syntax:

print(msg, file=sys.stderr)

This ensures that error messages are properly directed to the standard error stream (stderr).

---

## Changes

- Updated cmake/templates/hpxrun.py.in
- Replaced incorrect calls:

  print(msg, sys.stderr)

- With the correct Python 3 syntax:

  print(msg, file=sys.stderr)

- Ensures all error messages are correctly printed to stderr.

---

## Verification

- Created a standalone Python 3 test script to reproduce the issue.
- Verified that:
  - Error messages are correctly written to stderr.
  - No unintended output appears in stdout.

### DIFF
--- a/cmake/templates/hpxrun.py.in
+++ b/cmake/templates/hpxrun.py.in
@@ -189,7 +189,7 @@ def run_mpi(cmd, localities, verbose):
     if mpiexec == '':
         msg = 'mpiexec not available on this platform. '
         msg += 'Please rerun CMake with HPX_PARCELPORT_MPI=True.'
-        print(msg, sys.stderr)
+        print(msg, file=sys.stderr)
         sys.exit(1)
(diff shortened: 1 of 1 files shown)


arpittkhandelwal:
> Please remove the changes to the CI configuration files. Otherwise LGTM, thanks!

Sir I have updated this.

PR #6999, opened 2026-03-12
Add distributed versions of replace, replace_if, replace_copy and replace_copy_if algorithms

abhishek593:
## Proposed Changes

  - Added segmented `hpx::replace` and `hpx::replace_if` overloads
  - Added segmented `hpx::replace_copy` and `hpx::replace_copy_if` overloads
  - Added tests (partitioned_vector_replace.cpp and partitioned_vector_replace_copy.cpp)

The implementation for `hpx::replace` and `hpx::replace_if`  follows the established pattern of `hpx::fill`, while for `hpx::replace_copy` and `hpx::replace_copy_if`, it follows `hpx::transform`.

### DIFF
--- a/libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace.hpp
+++ b/libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace.hpp
@@ -12,6 +12,14 @@
 #include <hpx/modules/functional.hpp>
 #include <hpx/parallel/segmented_algorithms/for_each.hpp>
 
+#include <algorithm>
+#include <exception>
+#include <iterator>
+#include <list>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
 namespace hpx::parallel {
 
     ///////////////////////////////////////////////////////////////////////////
--- a/libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace_copy.hpp
+++ b/libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace_copy.hpp
@@ -12,6 +12,16 @@
 #include <hpx/modules/executors.hpp>
 #include <hpx/parallel/segmented_algorithms/detail/dispatch.hpp>
 
+#include <algorithm>
+#include <cstddef>
+#include <exception>
+#include <iterator>
+#include <list>
+#include <numeric>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
 namespace hpx::parallel {
 
     ///////////////////////////////////////////////////////////////////////////


libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace.hpp
```diff
@@ -0,0 +1,216 @@
...
+    ///////////////////////////////////////////////////////////////////////////
+    // segmented_replace
+    namespace detail {
+        ///////////////////////////////////////////////////////////////////////
+        /// \cond NOINTERNAL
+
+        template <typename T>
+        struct replace_function
+        {
+            replace_function(T old_val = T(), T new_val = T())
```
hkaiser: Having the arguments initialized by default values doesn't really make sense. Why would anybody want to replace one value with the same?

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace.hpp
```diff
@@ -0,0 +1,216 @@
...
+            }
+
+            template <typename Archive>
+            void serialize(Archive& ar, unsigned /* version */)
+            {
+                // clang-format off
+                ar & old_value_;
+                ar & new_value_;
+                // clang-format on
+            }
```
hkaiser: Since this type is paren-initializable, we don't need the explicit serialization function. Our serialization layer should automatically do the right thing.

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace.hpp
```diff
@@ -0,0 +1,216 @@
...
+        };
+
+        template <typename F, typename T>
+        struct replace_if_function
+        {
+            replace_if_function(F pred = F(), T new_val = T())
+              : pred_(pred)
+              , new_value_(new_val)
+            {
+            }
```
hkaiser: I think switching the arguments and not defaulting the new value would make more sense. Also, please make the constructor explicit. 
```suggestion
            template <typename T_ , typename F_>
            explicit replace_if_function(T_&& new_val, F_&& pred = F())
              : pred_(HPX_FORWARD(F_, f))
              , new_value_(HPX_FORWARD(T_, new_val))
            {}
```
You may also want to add deduction guides like:
```
template <typename T, typename F>
replace_if_function(T&&, F&&) -> replace_if_function<std::decay_t<F>, std::decay_t<T>>;
```

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace.hpp
```diff
@@ -0,0 +1,216 @@
...
+            void operator()(T& val) const
+            {
+                if (pred_(val))
+                {
+                    val = new_value_;
+                }
+            }
+
+            template <typename Archive>
+            void serialize(Archive& ar, unsigned /* version */)
```
hkaiser: Same here, this might not be needed.

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace.hpp
```diff
@@ -0,0 +1,216 @@
...
+              , new_value_(new_val)
+            {
+            }
+
+            F pred_;
+            T new_value_;
+
+            void operator()(T& val) const
+            {
+                if (pred_(val))
```
hkaiser: Please use `HPX_INVOKE` (it's equivalent to `std::invoke`) to support member functions and member data.
```suggestion
                if (HPX_INVOKE(pred_, val))
```

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace.hpp
```diff
@@ -0,0 +1,216 @@
...
+
+#include <algorithm>
+#include <exception>
+#include <iterator>
+#include <list>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+namespace hpx { namespace parallel {
```
hkaiser: Please feel free to use C++17 nested namespaces:
```suggestion
namespace hpx::parallel {
```
We just have not gotten around changing this globally for the segmented algorithms module yet.

abhishek593:
@hkaiser Thanks for the feedback. I have addressed the review comments and have incorporated them. Please check.

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+            template <typename T1, typename T2>
+            explicit replace_function(T1&& old_val, T2&& new_val)
+              : old_value_(HPX_FORWARD(T1, old_val))
+              , new_value_(HPX_FORWARD(T2, new_val))
+            {
+            }
+
+            T old_value_;
+            T new_value_;
```
hkaiser: Since we need the default constructor for serialization purposes (I didn't think of that when I made my comments before), You may want to initialize the members as:
```suggestion
            T old_value_ = T();
            T new_value_ = T();
```
as those may stay uninitialized otherwise. Sorry for the confusion, I now understand why you were using the default values for the constructor arguments.

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace.hpp
```diff
@@ -0,0 +1,196 @@
...
+            {
+                if (val == old_value_)
+                {
+                    val = new_value_;
+                }
+            }
+        };
+
+        template <typename T1, typename T2>
+        replace_function(T1&&, T2&&) -> replace_function<std::decay_t<T1>>;
```
hkaiser: I'm not sure if this is what you want. This allows for different argument types after all.

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace.hpp
abhishek593: @hkaiser Is there any issue with leaving them uninitialized, since serialization layer will initialize them anyways? It doesn't hurt but just trying to understand any implications.

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace.hpp
abhishek593: The intent was to support perfect forwarding for replace just like replace_if(your code snippet), but this does allow deduction with mismatched types, leading to silent conversions. I missed that. We can use the following to enforce that both arguments share the same type:

```
template <typename T>
replace_function(T, T) -> replace_function<std::decay_t<T>>;
```
Are you aligned with this?

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/replace.hpp
hkaiser: It's always a good idea to initialize all variables. You never know if somebody tries to use your type in any unintended way after all...

abhishek593:
@hkaiser I noticed that the existing segmented algorithms either don’t support projections, or internally accepts a projection parameter but always pass hpx::identity. So, user-supplied projections are not exposed through the tag_invoke overloads. I followed the same pattern for replace_copy* for now. Is it by design or we should support user-supplied projections?

hkaiser:
> @hkaiser I noticed that the existing segmented algorithms either don’t support projections, or internally accepts a projection parameter but always pass hpx::identity. So, user-supplied projections are not exposed through the tag_invoke overloads. I followed the same pattern for replace_copy* for now. Is it by design or we should support user-supplied projections?

The segmented algorithms are modelled after the non-range-based standard algorithms. Those don't have projections. If we ever decide to implement range-based distributed algorithms we will add projections as well.

hkaiser (dismissed):
Let's not make this PR even bigger. So far, LGTM. Thanks!

hkaiser:
Could you please have a look at the reported inspect issues: https://output.circle-artifacts.com/output/job/715455c4-83e6-4cea-b946-b36081747aa1/artifacts/0/hpx_inspect_report.html?

abhishek593:
@hkaiser There were some missing includes, according to the inspect tool, which i have added. The rest of CI workflows needs to be started again.

PR #7006, opened 2026-03-13
Fix inaccurate comments in 1d_stencil_5 example

360NK:
This updates two comments in examples/1d_stencil/1d_stencil_5.cpp.

First, the serialization comment referred to an "empty" implementation, but partition_data currently provides real save/load serialization logic.

Second, the action comment referred to heat_part as a global function, but it is actually a static member function of stepper.

No behaviour change.


hkaiser:
Please fix the clang-format issue reported, then we can merge this.

PR #7013, opened 2026-03-14
Fix incorrect Doxygen for hpx::search_n: "last" -> "first" subsequence

aneek22112007-tech:
While reading through the `search.hpp` docs I noticed the `\returns` section for `hpx::search_n` says it returns an iterator to the beginning of the **last** subsequence - but that's actually how `find_end` works, not `search_n`. `search_n` (like `std::search_n`) returns the **first** match.

Looks like the wording got copy-pasted from `find_end`'s docs at some point. This is just a two-word comment fix, no code changes at all.

**Changed:** `search.hpp` - two instances of `"last"` -> `"first"` in the `\returns` Doxygen comment for both overloads of `search_n`.


aneek22112007-tech:
Closing this PR — will revisit and resubmit if needed.

PR #7016, opened 2026-03-14
support GCC trunk/C++26 reflection in build, tests and CI

sleepingeight:
Major changes:
1. Updated use of <experimental/meta> with <meta> to work with gcc.
2. Modified a testcase from long to unsigned int due to the following error - 
```
/workspace/libs/core/serialization/tests/unit/refl/reflection_qualified_name_of.cpp(56): test 'std::string(name) == std::string("a::b::c::d::pentagon<int,char,double,float,long>")' failed in function 'main': 'a::b::c::d::pentagon<int,char,double,float,long int>' != 'a::b::c::d::pentagon<int,char,double,float,long>'
```
since the error is caused by compiler implementation, it felt better to have something common in both cases.

3. Added CI for testing out the changes.
4. Fixed a missing-braces error - 
```
/workspace/libs/core/serialization/tests/unit/refl/reflection_array.cpp:75:57: error: missing braces around initializer for 'std::__array_traits<int, 3>::_Type' {aka 'int [3]'} [-Werror=missing-braces] 75 | input.vector_of_arrays = {{10, 20, 30}, {40, 50, 60}}; |
```

### DIFF
--- a/.github/workflows/linux_release_gcc_trunk_reflection.yml
+++ b/.github/workflows/linux_release_gcc_trunk_reflection.yml
@@ -11,7 +11,7 @@ on: [pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
-    container: sleepingeight/gcc-cpp-refl:latest
+    container: stellargroup/gcc_trunk_build_env:1
 
     steps:
     - uses: actions/checkout@v6


hkaiser:
@ujjwal-shekhar Would you mind having a look?

libs/core/serialization/tests/unit/refl/reflection_qualified_name_of.cpp
```diff
@@ -51,10 +51,12 @@ int main()
 
     // Deep Namespace + High Arity
     {
-        using type = a::b::c::d::pentagon<int, char, double, float, long>;
+        using type =
+            a::b::c::d::pentagon<int, char, double, float, unsigned int>;
```
hkaiser, line 55: Does this mean that we can't properly reflect on `long`s?

hkaiser:
@sleepingeight please again share the Dockerfile for this.

libs/core/serialization/tests/unit/refl/reflection_qualified_name_of.cpp
sleepingeight: Yes. For example, the following code:

```cpp
#include <iostream>
#include <meta>

template<typename T>
void fun(){
    constexpr auto type_meta = ^^T;
    constexpr auto name_view = std::meta::display_string_of(type_meta);
    std::cout << "Type: " << name_view << "\n";
}

int main() {
    fun<long>();
}
```
results in `Type: long` with clang and `Type: long int` for gcc.

sleepingeight: Godbolt - https://godbolt.org/z/xG33638Ye

hkaiser: Ok, this looks like a gcc/clang inconsistency. They will have to fix that eventually. We might want to report that to them.

ujjwal-shekhar: From my understanding, we can continue to reflect on `long` appropriately. The issue is that `std::meta::display_string_of` is now implementation defined (in the original reflection proposal, there was a `std::meta::qualified_name_of` but that was later removed and users were encouraged to write their own versions).

The only problem we run into as a consequence of compiler-defined behaviour is this: if we were to serialize polymorphic types from a machine that compiled the binary using gcc trunk to/from clang, the generated `qualified_name` will be different on both machines.

I will think one way we can get around this (to some extent) is to see all/most primitive types and how their `std::meta::display_string_of` differs and specify these edge cases, but this is only a band-aid fix. I was also thinking of using maybe something that isn't compiler-dependent but is unique for a type and hash it at compile time.

ujjwal-shekhar: Some context is provided here: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p2996r13.html#reflecting-names 

Also, Thank you @sleepingeight for bringing this up! I will be thinking about a better way of solving it, if you would like to brainstrom, the crux of the task is: Can we have a unique string representation for a fully specialized type that is not implementation/compiler defined behaviour? Our `qualified_name_of` is _almost_ there but the dependence on `display_string_of` is the last wrinkle `:P`

sleepingeight: > Ok, this looks like a gcc/clang inconsistency. They will have to fix that eventually. We might want to report that to them.

@hkaiser I don't think this will be considered as a bug to them since the paper explicitly states that `display_string_of` is implementation defined. 

> consteval string_view display_string_of(info r);
consteval u8string_view u8display_string_of(info r);
Returns: An implementation-defined string_view or u8string_view, respectively

Ref - https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p2996r13.html#pnum_523

@ujjwal-shekhar Thanks for the follow up. I'll try to explore any possible fixes here.

sleepingeight:
@hkaiser dockerfile - 
```
# building gcc trunk
FROM ubuntu:24.04 AS builder

ENV DEBIAN_FRONTEND=noninteractive

RUN apt-get update && apt-get install -y --no-install-recommends \
    build-essential \
    git \
    cmake \
    ninja-build \
    python3 \
    wget \
    ca-certificates \
    flex \
    bison \
    texinfo \
    libgmp-dev \
    libmpfr-dev \
    libmpc-dev \
    libisl-dev \
 && rm -rf /var/lib/apt/lists/*

RUN git clone --depth=1 --branch=trunk https://gcc.gnu.org/git/gcc.git /src/gcc
WORKDIR /src/gcc
RUN ./contrib/download_prerequisites
RUN mkdir /build && cd /build && \
    /src/gcc/configure \
        --prefix=/gcc \
        --enable-languages=c,c++ \
        --disable-multilib \
        --enable-checking=release \
        --enable-host-shared
RUN make -C /build -j$(nproc)
RUN make -C /build install

# runtime image
FROM ubuntu:24.04

ENV DEBIAN_FRONTEND=noninteractive

RUN apt-get update && apt-get install -y --no-install-recommends \
    build-essential \
    ninja-build \
    git \
    curl \
    wget \
    python3 \
    python3-dev \
    sudo \
    ca-certificates \
    vim \
    libhwloc-dev \
    libboost-dev \
 && rm -rf /var/lib/apt/lists/*
 
# modern cmake for cxx26
RUN wget https://github.com/Kitware/CMake/releases/download/v3.30.5/cmake-3.30.5-linux-x86_64.sh \
 && chmod +x cmake-3.30.5-linux-x86_64.sh \
 && ./cmake-3.30.5-linux-x86_64.sh --skip-license --prefix=/usr/local \
 && rm cmake-3.30.5-linux-x86_64.sh

COPY --from=builder /gcc /usr/local
RUN echo "/usr/local/lib64" > /etc/ld.so.conf.d/gcc.conf \
 && echo "/usr/local/lib" >> /etc/ld.so.conf.d/gcc.conf \
 && ldconfig

WORKDIR /workspace
```

This too is pretty well optimised in size. 

```
IMAGE                                      ID             DISK USAGE   CONTENT SIZE
sleepingeight/gcc-cpp-refl:latest          853bda1e5a6a       4.07GB          1.1GB
```

libs/core/serialization/tests/unit/refl/reflection_qualified_name_of.cpp
sleepingeight: @ujjwal-shekhar I feel that a straightforward approach is to port `std::meta::display_string_of` in hpx. The clang implementation defined in https://github.com/bloomberg/clang-p2996/blob/p2996/libcxx/include/meta can serve as an reference (it is easy to implement as it doesn't require any changes in compiler as opposed to the gcc implementation).

hkaiser: Please be careful with 'copying' code from clang's libstdc++, we might run into licensing issues. But a full re-implementation should be fine.

ujjwal-shekhar: @sleepingeight fair enough. I don't think we would need to reimplement the whole pretty printer, since a lot of its features are not necessary for our usecase (say, cv qualifiers or pointers etc). Also, many of the features like building the entire enclosing namespace and comma separated template args are already implemented in our version:)

Since we would only need to use `display_string_of` on primitives (I suppose, any type that satisfies `std::is_fundamental_v` ? Perhaps this casework (https://github.com/bloomberg/clang-p2996/blob/p2996/libcxx/include/meta#L2817) is already most of what we need. Reimplementing would also come with the downside of having to track any bug fixes on the clang side and fixing them here.

Another thing to keep in mind would be non-standard width types like `__int128`, clang evaluates the `display_string_of` this to "(anonymous type)" while gcc evaluates it to `__int128`.  (see: https://godbolt.org/z/8sGbocz8j). We could emit a warning for any cases not covered by clang and ask the user to manually provide the portable string for the type (or rely on `std::meta::dispklay_string_of` as a fallback at their own risk).

CC: @hkaiser, do you think we should reimplement the whole pretty-printer?

hkaiser: Let's do as little as possible (and needed).

sleepingeight: @hkaiser @ujjwal-shekhar I've implemented a fix in https://github.com/STEllAR-GROUP/hpx/pull/7039. Could you please take a look. 

 `__int128` seems to be only compiler defined data type which is not present in MSVC. So do you think it is necessary considering it?

Apart from that we have few other inconsistent cases, for example - `char* const` being evaluated to `const char*` in clang and `char* const` in gcc (clang's output seems completely wrong). Ref - https://godbolt.org/z/TnYf8E5xz

.github/workflows/linux_release_gcc_trunk_reflection.yml
```diff
@@ -0,0 +1,52 @@
...
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+name: Linux CI (Release, GCC Trunk Reflection)
+
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container: sleepingeight/gcc-cpp-refl:latest
```
hkaiser: Please use:
```suggestion
    container: stellargroup/gcc_trunk_build_env:1
```

sleepingeight:
@hkaiser can this branch be merged or should I fix DCO?

PR #7017, opened 2026-03-15
Enable Tracy fiber-based task tracing for HPX scheduler 

### LINKED ISSUE #6989: Tracy fiber support present but currently unused in HPX runtime
While exploring the current Tracy integration in HPX I noticed a couple of things that made me curious about the intended design.

### 1. Fiber wrappers exist but are not currently used

The Tracy module under `libs/core/tracy` exposes wrappers for the Tracy fiber API:

* `enter_fiber()`
* `leave_fiber()`
* `fiber_region`

Tracy supports fiber instrumentation through the `TRACY_FIBERS` define. However, after searching through the runtime it appears that these functions are not currently called in the scheduler or task execution paths.

For example, the scheduling loop instruments task execution using:

```cpp
tracy::region rctx(name, num_thread, thrdptr->get_thread_phase());
```

which attaches zones to OS worker threads rather than logical HPX tasks.

Since HPX uses an M:N threading model where many HPX tasks are scheduled onto a smaller number of worker threads, this means the Tracy timeline currently reflects worker-thread activity rather than task-level timelines.

### 2. `suspend_region` exists but also appears unused

`tracy::suspend_region` is defined in `tracy_tls.hpp` as an RAII helper intended to temporarily close a region and restore it later.

I noticed that the equivalent `likwid::suspend_region` is used at yield points (e.g. in `execution_agent.cpp` and `thread_helpers.cpp`), but I couldn’t find usage of the Tracy equivalent.

### Question

Was the intention originally to map HPX tasks to Tracy fibers, or is the current worker-thread-level region instrumentation considered sufficient?

It seems like mapping HPX tasks to Tracy fibers could potentially provide clearer task-level timelines for M:N scheduling, but I wanted to check whether this direction aligns with the project's goals before experimenting further.

If mapping HPX tasks to Tracy fibers would be desirable, I would be happy to experiment with this direction and see how it could integrate with the current scheduler instrumentation.

Thanks!

v4xsh:
Fixes #6989

## Proposed Changes

- Enable Tracy fiber instrumentation in the scheduler so logical HPX tasks are represented as Tracy fibers.
- Preserve existing worker-thread zones to keep worker utilization visible.
- Add stable Tracy fiber names for HPX tasks using thread description + `thread_data` address.
- Add Tracy TLS helpers to manage fiber zone lifecycle (`start/stop/suspend/resume`) safely.
- Instrument suspension paths (`execution_agent::do_yield`, `this_thread::suspend`, timed suspend path) so Tracy shows when tasks yield/block.

HPX uses an M:N scheduling model where many HPX tasks run on fewer worker threads.  
Before this change, Tracy mainly showed worker-thread activity.  
This PR adds a dual-view approach: worker-thread zones remain, and HPX tasks are additionally traced as fibers.  
This makes it possible to follow one logical task across suspension and worker-thread migration while preserving worker utilization visibility.

Validated locally with suspension-heavy workloads (`yield`, `sleep_for`, timers, async tasks), and Tracy shows:
- worker utilization rows,
- logical task fiber timelines,
- explicit suspend/resume intervals,
- task migration across workers.

## Checklist

### DIFF
--- a/libs/core/tracy/include/hpx/tracy/tracy.hpp
+++ b/libs/core/tracy/include/hpx/tracy/tracy.hpp
@@ -1,4 +1,5 @@
 //  Copyright (c) 2026 Hartmut Kaiser
+//  Copyright (c) 2026 Vansh Dobhal
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
@@ -10,6 +11,7 @@
 
 #if defined(HPX_HAVE_MODULE_TRACY)
 
+#include <cstddef>
 #include <string>
 
 namespace hpx::tracy {
@@ -17,21 +19,37 @@ namespace hpx::tracy {
     HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT void set_thread_name(
         char const* name) noexcept;
 
-    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT void enter_fiber(
-        char const* name) noexcept;
-    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT void leave_fiber() noexcept;
+    namespace detail {
+
+        HPX_CORE_EXPORT void enter_fiber(char const* fiber_name,
+            char const* zone_name = nullptr, std::size_t color = 0) noexcept;
+        HPX_CORE_EXPORT void leave_fiber() noexcept;
+
+    }    // namespace detail
 
     HPX_CXX_CORE_EXPORT struct fiber_region
     {
-        explicit fiber_region(char const* name) noexcept
+        explicit fiber_region(char const* fiber_name,
+            char const* zone_name = nullptr, std::size_t color = 0,
+            bool enabled = true) noexcept
+          : active(enabled)
         {
-            enter_fiber(name);
+            if (active)
+            {
+                detail::enter_fiber(fiber_name, zone_name, color);
+            }
         }
 
         ~fiber_region() noexcept
         {
-            leave_fiber();
+            if (active)
+            {
+                detail::leave_fiber();
+            }
         }
+
+    private:
+        bool active;
     };
 
     HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT void create_counter(
--- a/libs/core/tracy/include/hpx/tracy/tracy_tls.hpp
+++ b/libs/core/tracy/include/hpx/tracy/tracy_tls.hpp
@@ -1,4 +1,5 @@
 //  Copyright (c) 2025-2026 Hartmut Kaiser
+//  Copyright (c) 2026 Vansh Dobhal
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
@@ -24,44 +25,47 @@ namespace hpx::tracy {
         std::uint32_t phase = 0;
     };
 
-    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT region_data start_region(
-    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT region_data start_region(
-        char const*, std::size_t = 0, std::size_t = 0) noexcept;
-    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT char const* rename_region(
-    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT char const* rename_region(
-        char const*) noexcept;
-    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT region_data stop_region(
-    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT region_data stop_region(
-        region_data const& prev_region) noexcept;
-    // Set/clear the per-OS-thread "inside fiber" flag used to guard
-    // rename_region from operating on a stale zone ctx.
-    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT void set_in_fiber(bool) noexcept;
+    namespace detail {
 
-    // Open/close a zone on the fiber's zone stack so that the fiber track
-    // is visible in Tracy.  Must be called after/before TracyFiberEnter/Leave.
-    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT void start_fiber_zone(
-        char const* zone_name, std::uint32_t color = 0) noexcept;
-    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT void stop_fiber_zone() noexcept;
+        HPX_CORE_EXPORT region_data start_region(
+            char const*, std::size_t = 0, std::size_t = 0) noexcept;
+        HPX_CORE_EXPORT char const* rename_region(char const*) noexcept;
+        HPX_CORE_EXPORT region_data stop_region(
+            region_data const& prev_region) noexcept;
+        // Set/clear the per-OS-thread "inside fiber" flag used to guard
+        // rename_region from operating on a stale zone ctx.
+        HPX_CORE_EXPORT void set_in_fiber(bool) noexcept;
 
-    // Suspend/resume the zone that is currently open on the fiber's zone stack.
-    // Call suspend_fiber_zone() just before self_.yield() and
-    // resume_fiber_zone() immediately after self_.yield() returns.
-    // Both operate solely on current_fiber_zone() - they never touch the
-    // OS-thread zone (current_region()), so there is no zone-stack conflict.
-    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT void suspend_fiber_zone(
-        char const* suspend_reason = nullptr) noexcept;
-    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT void resume_fiber_zone(
-        char const* zone_name = nullptr, std::uint32_t color = 0) noexcept;
+        // Open/close a zone on the fiber's zone stack so that the fiber track
+        // is visible in Tracy.  Must be called after/before
+        // TracyFiberEnter/Leave.
+        HPX_CORE_EXPORT void start_fiber_zone(
+            char const* zone_name, std::uint32_t color = 0) noexcept;
+        HPX_CORE_EXPORT void stop_fiber_zone() noexcept;
+
+        // Suspend/resume the zone that is currently open on the fiber's zone
+        // stack. Call suspend_fiber_zone() just before self_.yield() and
+        // resume_fiber_zone() immediately after self_.yield() returns.
+        // Both operate solely on current_fiber_zone() - they never touch the
+        // OS-thread zone (current_region()), so there is no zone-stack
+        // conflict.
+        HPX_CORE_EXPORT void suspend_fiber_zone(
+            char const* suspend_reason = nullptr) noexcept;
+        HPX_CORE_EXPORT void resume_fiber_zone(
+            char const* zone_name = nullptr, std::uint32_t color = 0) noexcept;
+
+    }    // namespace detail
 
     HPX_CXX_CORE_EXPORT struct region
     {
         explicit region(char const* name, std::size_t const thread_num,
             std::size_t phase, bool enabled = true) noexcept
+          : active(enabled)
         {
-            if (enabled)
+            if (active)
             {
-                surrounding_region = start_region(name, thread_num, phase);
-                active = true;
+                surrounding_region =
+                    detail::start_region(name, thread_num, phase);
             }
         }
 
@@ -69,12 +73,12 @@ namespace hpx::tracy {
         {
             if (active)
             {
-                stop_region(surrounding_region);
+                detail::stop_region(surrounding_region);
             }
         }
 
     private:
-        bool active = false;
+        bool active;
         region_data surrounding_region;
     };
 
@@ -89,7 +93,7 @@ namespace hpx::tracy {
     HPX_CXX_CORE_EXPORT struct suspend_region
     {
         suspend_region() noexcept
-          : suspended_region(stop_region({}))
+          : suspended_region(detail::stop_region({}))
         {
         }
 
@@ -97,8 +101,8 @@ namespace hpx::tracy {
         {
             if (suspended_region.name != nullptr)
             {
-                start_region(suspended_region.name, suspended_region.color,
-                    suspended_region.phase);
+                detail::start_region(suspended_region.name,
+                    suspended_region.color, suspended_region.phase);
             }
         }
 
@@ -108,13 +112,13 @@ namespace hpx::tracy {
     HPX_CXX_CORE_EXPORT struct mark_event
     {
         explicit mark_event(char const* name) noexcept
-          : previous_name(rename_region(name))
+          : previous_name(detail::rename_region(name))
         {
         }
 
         ~mark_event()
         {
-            rename_region(previous_name);
+            detail::rename_region(previous_name);
         }
 
     private:
@@ -132,12 +136,12 @@ namespace hpx::tracy {
         explicit fiber_suspend_region(
             char const* suspend_reason = nullptr) noexcept
         {
-            suspend_fiber_zone(suspend_reason);
+            detail::suspend_fiber_zone(suspend_reason);
         }
 
         ~fiber_suspend_region() noexcept
         {
-            resume_fiber_zone();
+            detail::resume_fiber_zone();
         }
     };
 }    // namespace hpx::tracy
--- a/libs/core/tracy/src/tracy_tls.cpp
+++ b/libs/core/tracy/src/tracy_tls.cpp
@@ -1,4 +1,5 @@
 //  Copyright (c) 2025-2026 Hartmut Kaiser
+//  Copyright (c) 2026 Vansh Dobhal
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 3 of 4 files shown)


libs/core/thread_pools/include/hpx/thread_pools/scheduling_loop.hpp
```diff
@@ -234,8 +235,32 @@ namespace hpx::threads::detail {
 #if defined(HPX_HAVE_MODULE_TRACY)
                                 char const* name = thrdptr->get_description()
                                                        .get_description();
-                                tracy::region rctx(name, num_thread,
-                                    thrdptr->get_thread_phase());
+                                // Dual-view Tracy instrumentation:
+                                //
+                                // rctx declared FIRST \u2192 constructed first \u2192
```
hkaiser: Coudld you remove the unicode characters here, please?
```suggestion
                                // rctx declared FIRST -> constructed first
```
The same applies to more spots below.

libs/core/threading_base/include/hpx/threading_base/thread_data.hpp
```diff
@@ -285,6 +287,48 @@ namespace hpx::threads {
...
+#if defined(HPX_HAVE_MODULE_TRACY)
+    private:
+        mutable char tracy_fiber_name_[64] = {0};
+
+    public:
+        // Returns a unique, stable string identifying this HPX task as a Tracy
+        // fiber. Built lazily on first call and cached in tracy_fiber_name_.
+        // Format: "<description>_<thread_id_ptr>" so each HPX task gets its
+        // own fiber track in the Tracy profiler.
+        char const* get_tracy_fiber_name() const noexcept
```
hkaiser: Could this function be moved to the corresponding cpp file?

libs/core/threading_base/include/hpx/threading_base/thread_data.hpp
```diff
@@ -285,6 +287,48 @@ namespace hpx::threads {
             threads::thread_description value);
 #endif
 
+#if defined(HPX_HAVE_MODULE_TRACY)
+    private:
+        mutable char tracy_fiber_name_[64] = {0};
```
hkaiser: Initializing the whole array with zero might introduce very high costs. Possibly initializing the first byte would be sufficient. You already do that in the rebind function. You should do the same in the constructor.

libs/core/threading_base/src/thread_helpers.cpp
```diff
@@ -427,6 +430,30 @@ namespace hpx::threads {
...
+            if (description.kind() ==
+                threads::thread_description::data_type::description)
+            {
+                char const* reason = description.get_description();
+                if (reason != nullptr && reason[0] != '\0')
+                {
+                    return reason;
+                }
+            }
+#endif
```
hkaiser: This code seems to be duplicated with `thread_data::get_tracy_fiber_name()`. Can this be simplified?

libs/core/tracy/src/tracy_tls.cpp
```diff
@@ -20,13 +23,57 @@ namespace hpx::tracy {
...
+            if (label == nullptr || label[0] == '\0')
+            {
+                label = fallback;
+            }
+
+            // A simple static lock and set for in-memory string interning.
+            static std::unordered_set<std::string> interned_strings;
+            static std::mutex interning_mutex;
+
+            std::lock_guard<std::mutex> lock(interning_mutex);
```
hkaiser: Do you have to protect this against threading ssues? Would it be guaranteed that this function is only ever called by one thread?

libs/core/tracy/src/tracy_tls.cpp
```diff
@@ -20,13 +23,57 @@ namespace hpx::tracy {
...
+            }
+
+            constexpr std::size_t MAX_TRACY_LABELS = 4096;
+            if (interned_strings.size() > MAX_TRACY_LABELS)
+            {
+                return fallback;
+            }
+
+            auto result = interned_strings.insert(label);
+            return result.first->c_str();
```
hkaiser: Does this code duplicate the existing facility here: https://github.com/STEllAR-GROUP/hpx/blob/6cf6e6a879569b112f13c8e91a24174c1a626f1f/libs/core/threading_base/include/hpx/threading_base/scoped_annotation.hpp#L35?

libs/core/tracy/src/tracy_tls.cpp
```diff
@@ -80,8 +147,150 @@ namespace hpx::tracy {
...
+    void start_fiber_zone(char const* zone_name, std::uint32_t color) noexcept
+    {
+        char const* safe_zone_name = "fiber";
+        try
+        {
+            safe_zone_name = intern_zone_label(zone_name, "fiber");
+        }
+        catch (...)
+        {
+        }
```
hkaiser: Wouldn't it be better to make `intern_zone_label` `noexcept`:
```suggestion
        char const* safe_zone_name = intern_zone_label(zone_name, "fiber");
```

libs/core/tracy/src/tracy_tls.cpp
```diff
@@ -80,8 +147,150 @@ namespace hpx::tracy {
...
+        // 0xAAAAAA = medium grey, distinguishable from any worker-index color.
+        constexpr std::uint32_t suspended_color = 0xAAAAAA;
+        char const* safe_reason = "suspend";
+        try
+        {
+            safe_reason = intern_zone_label(suspend_reason, "suspend");
+        }
+        catch (...)
+        {
+        }
```
hkaiser: If `intern_zone_label` was `noexcept`:
```suggestion
        char const* safe_reason = intern_zone_label(suspend_reason, "suspend");
```

libs/core/tracy/src/tracy_tls.cpp
```diff
@@ -80,8 +147,150 @@ namespace hpx::tracy {
...
+            // clang-format on
+
+            tracy_context data;
+        data.context = ctx;
+
+        auto& fz = current_fiber_zone();
+        fz.ctx_value = data.value;
+        fz.zone_name = safe_zone_name;
+        fz.color = color;
+        fz.active = true;
```
hkaiser: This code could be factored out into a separate function to avoid duplication.

v4xsh:
@hkaiser  Thanks for the detailed review! I’ve pushed updates addressing the comments (unicode arrows removed, helper moved to cpp, label handling simplified, and duplicated logic cleaned up). Please let me know if anything else should be adjusted.

libs/core/tracy/include/hpx/tracy/tracy.hpp
```diff
@@ -10,21 +10,24 @@
...
+#include <cstdint>
 #include <string>
 
 namespace hpx::tracy {
 
     HPX_CXX_EXPORT HPX_CORE_EXPORT void set_thread_name(
         char const* name) noexcept;
 
-    HPX_CXX_EXPORT HPX_CORE_EXPORT void enter_fiber(char const* name) noexcept;
+    HPX_CXX_EXPORT HPX_CORE_EXPORT void enter_fiber(char const* fiber_name,
```
hkaiser: The master branch has been updated to use `HPX_CXX_CORE_EXPORT` for the C++ module adaptation. Could you please change your new functions accordingly?

libs/core/tracy/include/hpx/tracy/tracy.hpp
```diff
@@ -15,7 +15,7 @@
 
 namespace hpx::tracy {
 
-    HPX_CXX_EXPORT HPX_CORE_EXPORT void set_thread_name(
+    HPX_CXX_CORE_EXPORT void set_thread_name(
```
hkaiser: You'll still need both: `HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT`. One is for C++ modules, the other for symbol visibility.

libs/core/thread_pools/include/hpx/thread_pools/scheduling_loop.hpp
```diff
@@ -234,8 +235,32 @@ namespace hpx::threads::detail {
...
+                                std::optional<tracy::region> rctx;
+                                std::optional<tracy::fiber_region> fctx;
+                                if (name != nullptr && !thrdptr->is_stackless())
+                                {
+                                    rctx.emplace(name, num_thread,
+                                        thrdptr->get_thread_phase());
+                                    fctx.emplace(
+                                        thrdptr->get_tracy_fiber_name(), name,
+                                        static_cast<std::uint32_t>(num_thread));
+                                }
```
hkaiser: Since `std::optional` may require an allocation (which we should try to avoid in the scheduling loop). Do you think there is another way of expressing this? Perhaps some constructor arguments to `tracy::region` and `tracy::fiber_region` that allow to figure out whether to do any Tracy-related work in the constructor/destructor or not?
```suggestion
                                tracy::region rctx(thrdptr, name, num_thread);
                                tracy::fiber_region fctx(thrdptr, name, num_thread);
```

v4xsh: Thanks! Agreed. I’ll switch from optional-based conditional construction to always-declared RAII guards with internal no-op behavior, so the scheduling loop stays allocation-free and simpler while keeping the same dual-view Tracy semantics/order

v4xsh:
@hkaiser  I’ve updated this now.

 Replaced `std::optional<tracy::region>` and `std::optional<tracy::fiber_region>` in `scheduling_loop.hpp` with always-declared RAII guards.

 Added an `enabled` flag to both `region` and `fiber_region` (defaults to `true`) so they become no-ops internally when disabled. In the scheduling loop, `enable_tracy` is computed once and passed to both. Also made sure `get_tracy_fiber_name()` is only called when Tracy is actually enabled.

 Declaration/destruction order is preserved (`region` before `fiber_region`) to keep the dual-view Tracy semantics intact.

Please let me know if anything needs further adjustment.

libs/core/thread_pools/include/hpx/thread_pools/scheduling_loop.hpp
```diff
@@ -234,8 +234,31 @@ namespace hpx::threads::detail {
...
+                                // Destruction is C++ reverse order:
+                                //   ~fctx first -> TracyFiberLeave
+                                //   ~rctx last  -> ZoneEnd
+                                // Zone outlives the fiber context - correct.
                                 tracy::region rctx(name, num_thread,
-                                    thrdptr->get_thread_phase());
+                                    thrdptr->get_thread_phase(), enable_tracy);
+                                tracy::fiber_region fctx(
+                                    fiber_name, name,
+                                    static_cast<std::uint32_t>(num_thread), enable_tracy);
```
hkaiser: Why is `tracy::region` taking `num_thread` as is, while the `tracy::fiber_region` requires a `static_cast`? Can you do the casting inside the constructor to unclutter the code here, perhaps?

libs/core/tracy/include/hpx/tracy/tracy.hpp
```diff
@@ -10,28 +10,41 @@
...
 namespace hpx::tracy {
 
     HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT void set_thread_name(
         char const* name) noexcept;
 
-    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT void enter_fiber(
-        char const* name) noexcept;
+    HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT void enter_fiber(char const* fiber_name,
+        char const* zone_name = nullptr, std::uint32_t color = 0) noexcept;
     HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT void leave_fiber() noexcept;
```
hkaiser: Could these functions be moved to `namespace hpx::tracys::detail`? Users will never call those anyways. Then we possibly don't need to `HPX_CXX_CORE_EXPORT` them neither (the tests will tell).

libs/core/tracy/include/hpx/tracy/tracy.hpp
```diff
@@ -10,28 +10,41 @@
...
 
     HPX_CXX_CORE_EXPORT struct fiber_region
     {
-        explicit fiber_region(char const* name) noexcept
+        explicit fiber_region(char const* fiber_name,
+            char const* zone_name = nullptr, std::uint32_t color = 0,
+            bool enabled = true) noexcept
```
hkaiser, line 34: May I suggest to initialize `active` in the member initializer list:
```suggestion
            bool enabled = true) noexcept
          : active(enabled)
```
Then you don't need a default value for the member neither.

libs/core/tracy/include/hpx/tracy/tracy_tls.hpp
```diff
@@ -30,24 +30,59 @@ namespace hpx::tracy {
         char const*) noexcept;
     HPX_CXX_CORE_EXPORT HPX_CORE_EXPORT region_data stop_region(
```
hkaiser: I know those functions originally have been here, but for consistency I'd suggest to move those into `namespace detail` as well, which again removes the need to `HPX_CXX_CORE_EXPORT` them. The same applies to the other functions you introduced below. All we need in the API are the RAII wrappers.

v4xsh:
@hkaiser 
Thanks for the continued feedback reviews like this are exactly what make contributions better, and I am always happy to iterate! All four points have been addressed in the latest commits.

I also moved the internal Tracy helper functions into `hpx::tracy::detail` as suggested. I initially tried removing the `HPX_CXX_CORE_EXPORT`/`HPX_CORE_EXPORT` annotations as well, but with hidden visibility enabled in this build (`HPX_WITH_HIDDEN_VISIBILITY=ON`) that caused linker failures (undefined references to `hpx::tracy::detail::rename_region(...)` in downstream targets). So I kept the exports to preserve correct linking while still keeping those functions out of the public API surface via `detail`.

Happy to keep iterating if anything else needs adjusting!

hkaiser:
> I initially tried removing the `HPX_CXX_CORE_EXPORT`/`HPX_CORE_EXPORT` annotations as well, but with hidden visibility enabled in this build (`HPX_WITH_HIDDEN_VISIBILITY=ON`) that caused linker failures (undefined references to `hpx::tracy::detail::rename_region(...)` in downstream targets). So I kept the exports to preserve correct linking while still keeping those functions out of the public API surface via `detail`.

That's expected. What I meant is to just not export the symbol through C++ modules (i.e., leave the `HPX_CORE_EXPORT` in place, only remove the `HPX_CXX_CORE_EXPORT`).

hkaiser:
Also, please rebase onto master and resolve the merge conflicts.

libs/core/executors/include/hpx/executors/thread_pool_scheduler_bulk.hpp
```diff
@@ -937,10 +885,3 @@ namespace hpx::execution::experimental {
         }
     }
 }    // namespace hpx::execution::experimental
-#endif
-
-// Note: With stdexec integration, bulk operations are now customized
-// through the domain system in thread_pool_scheduler.hpp rather than
-// direct tag_invoke customizations. The thread_pool_domain<Policy>
-// intercepts stdexec::bulk_chunked_t operations and creates
-// thread_pool_bulk_sender instances for parallel execution.
```
hkaiser, line 946: The changes to this file seem to be unrelated to this PR. Please update your base branch and rebase onto that.

libs/core/tracy/include/hpx/tracy/tracy_tls.hpp
```diff
@@ -24,43 +24,84 @@ namespace hpx::tracy {
...
         explicit region(char const* name, std::size_t const thread_num,
-            std::size_t phase) noexcept
-          : surrounding_region(start_region(name, thread_num, phase))
+            std::size_t phase, bool enabled = true) noexcept
         {
+            if (enabled)
+            {
+                surrounding_region =
+                    detail::start_region(name, thread_num, phase);
+                active = true;
```
hkaiser: Same here. `start_region` is `noexcept`, so it is safe to initialize `active` in the member initializer list of the constructor. This also means that there is no need to default-initialize this member below.

libs/core/tracy/src/tracy_tls.cpp
```diff
@@ -62,40 +120,148 @@ namespace hpx::tracy {
...
+                // clang-format off
             char const* previous_name = name;
             name = new_region;
 
             tracy_context data;
             data.value = value;
             TracyCZoneName(data.context, new_region, std::strlen(new_region))
             return previous_name;
-            // clang-format on
+                // clang-format on
```
hkaiser: Could you please either remove the `clang-format off` here or adjust the formatting manually? The same applies to other spots as well.

v4xsh:
@hkaiser  can you please check now?

v4xsh:
@hkaiser Quick question I realized I forgot to add the copyright headers to the  Tracy fiber runtime files. Since a large portion of that logic was newly added in this PR, would it be appropriate for me to add my name to the headers (alongside yours) in those specific files now, or is that usually handled differently?

Happy to push a quick final commit for that if you're okay with it!

hkaiser:
> @hkaiser Quick question I realized I forgot to add the copyright headers to the Tracy fiber runtime files. Since a large portion of that logic was newly added in this PR, would it be appropriate for me to add my name to the headers (alongside yours) in those specific files now, or is that usually handled differently?

Of course. Our policy is that you add your copyright if you have touched ~20% of a file or if you create a new file.

v4xsh:
> > @hkaiser Quick question I realized I forgot to add the copyright headers to the Tracy fiber runtime files. Since a large portion of that logic was newly added in this PR, would it be appropriate for me to add my name to the headers (alongside yours) in those specific files now, or is that usually handled differently?
> 
> Of course. Our policy is that you add your copyright if you have touched ~20% of a file or if you create a new file.

Thanks for clarifying, I appreciate it! I have added the copyright headers to the Tracy files where the changes exceed that threshold.

hkaiser:
> > > @hkaiser Quick question I realized I forgot to add the copyright headers to the Tracy fiber runtime files. Since a large portion of that logic was newly added in this PR, would it be appropriate for me to add my name to the headers (alongside yours) in those specific files now, or is that usually handled differently?
> > 
> > 
> > Of course. Our policy is that you add your copyright if you have touched ~20% of a file or if you create a new file.
> 
> Thanks for clarifying, I appreciate it! I have added the copyright headers to the Tracy files where the changes exceed that threshold.

Thanks. Now let's wait for the CIs to cycle before merging.

v4xsh:
Sure !!

hkaiser:
@v4xsh Thanks again for working on this!

v4xsh:
> @v4xsh Thanks again for working on this!

Most welcome!!!

PR #7018, opened 2026-03-15
Fix heap corruption and out-of-bounds access in sort_thread partitioning

### LINKED ISSUE #6974: Fix failing unit tests related to sort algorithms
We're seeing occasional test failures for our tests related to the sort algorithms (see for instance: https://cdash.rostam.cct.lsu.edu/test/35939007, but similar failures happen for the non-ranged version of `hpx::sort`). We should try to identify the problem and fix it.

GitMasterJatin:
Fixes #6974

Two bugs in the quicksort partitioning loop inside sort_thread():

1. Pivot stored as reference instead of value_type: 'reference val = *first' silently changes after iter_swap(first, c_last) executes, corrupting all subsequent comparisons and potentially producing overlapping recursive sub-ranges that write to the same heap memory.

   Fix: use value_type val = *first to take a stable copy of the pivot.

2. Unbounded while-loops during partitioning: the four scanning loops had no bounds checks, relying solely on the pivot value as a sentinel. When duplicate elements cause the pivot to appear multiple times (or not at all in a sub-range), c_first can advance past 'last' or c_last can decrement before 'first', causing out-of-bounds reads and writes that corrupt heap metadata.

   Fix: add bounds guards (c_first < last, c_last > first) to all four scanning loops.

Together these produce the 'malloc_consolidate(): unaligned fastbin chunk detected' heap corruption followed by SIGSEGV (exit 139) observed on the sort_range_test with parallel_task_policy and float/std::greater<float>.

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/sort.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/sort.hpp
@@ -211,17 +211,17 @@ namespace hpx::parallel {
             // pivot selections
             pivot9(first, last, comp);
 
-            using reference =
-                typename std::iterator_traits<RandomIt>::reference;
+            using value_type =
+                typename std::iterator_traits<RandomIt>::value_type;
 
-            reference val = *first;
+            value_type val = *first;
             RandomIt c_first = first + 1, c_last = last - 1;
 
-            while (comp(*c_first, val))
+            while (c_first < last && comp(*c_first, val))
             {
                 ++c_first;
             }
-            while (comp(val, *c_last))
+            while (c_last > first && comp(val, *c_last))
             {
                 --c_last;
             }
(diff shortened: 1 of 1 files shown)


hkaiser:
@GitMasterJatin Ohh, nice catch!

GitMasterJatin:
Thank you @hkaiser Sir for reviewing and closing the issue

hkaiser:
@GitMasterJatin The issue with sort failing occasionally is still showing up, see for instance: https://cdash.rostam.cct.lsu.edu/viewTest.php?onlyfailed&buildid=43013. Could you have another look, please?

GitMasterJatin:
> @GitMasterJatin The issue with sort failing occasionally is still showing up, see for instance: https://cdash.rostam.cct.lsu.edu/viewTest.php?onlyfailed&buildid=43013. Could you have another look, please?

I'll look into it and will get back with a solution

PR #7020, opened 2026-03-15
Fix incorrect return iterators in buffer_memcpy relocation path

Sanchit2662:
# Summary

While reading through the relocation primitives in HPX, I noticed that the buffer_memcpy_tag code path , the fast path for trivially relocatable types ,   had inconsistent return values compared to the loop-based paths.                                                                                       
  
  The first issue was in uninitialized_relocate_n_primitive_helper. The for_loop_nothrow_tag and for_loop_try_catch_tag overloads both advance first and
  dst together in their loops, but the buffer_memcpy_tag overload only advanced dst after the memmove, leaving first at its original position:
```
  // before my fix
  std::memmove(dst_byte, first_byte, n_bytes);
  dst += n;
  // first was never advanced!
  return {first, dst};
```
  I added the missing first += n so the returned source iterator is correct.

  The second issue was a consequence in the backward variant. uninitialized_relocate_backward_primitive_helper with buffer_memcpy_tag delegates to the    
  forward _n helper and just returns whatever it returns. The problem is that the forward helper returns {first, dst_first + n} which equals {first,      
  dst_last} — the end of the destination. But backward relocation is supposed to return the beginning of the destination. The other two overloads get this
   right because they decrement dst_last down to dst_first in their loops. I fixed this by discarding the forward helper's return and explicitly returning
   {first, dst_first}:
```
  // after my fix
  uninitialized_relocate_n_primitive_helper(
      first, n_objects, dst_first, buffer_memcpy_tag{});
  return {first, dst_first};
```
  The third issue was a small typo in uninitialized_relocate_backward_t where a static_assert used && instead of , to attach the error message. This made 
  the string literal part of the boolean expression rather than the diagnostic message, so on failure the compiler would just say "static assertion       
  failed" with no explanation:
```
  // before: message is silently eaten by &&
  static_assert(std::bidirectional_iterator<BiIter1> &&
      "The 'first' and 'last' arguments must...");
  // after: message is properly shown
  static_assert(std::bidirectional_iterator<BiIter1>,
      "The 'first' and 'last' arguments must...");
```
  None of these were caught by existing tests because the test suite never checks the return values of the relocation functions for trivially relocatable
  types.

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/detail/iota.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/detail/iota.hpp
@@ -125,7 +125,7 @@ namespace hpx::parallel::detail {
                 return sequential_iota(
                     HPX_FORWARD(Expolicy, policy), first, last, value);
             }
-        };
+        }
     };
 
 }    // namespace hpx::parallel::detail
--- a/libs/core/serialization/include/hpx/serialization/detail/polymorphic_nonintrusive_factory_impl.hpp
+++ b/libs/core/serialization/include/hpx/serialization/detail/polymorphic_nonintrusive_factory_impl.hpp
@@ -15,6 +15,7 @@
 #include <hpx/serialization/detail/polymorphic_nonintrusive_factory.hpp>
 
 #include <memory>
+#include <type_traits>
 
 namespace hpx::serialization::detail {
     template <typename Derived>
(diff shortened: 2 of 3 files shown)


hkaiser:
@isidorostsa Would you mind having a look?

isidorostsa:
@Sanchit2662 Great catch! Thanks for the PR

PR #7024, opened 2026-03-15
fix: validate thread_schedule_hint.hint bounds in shared_priority_queue_scheduler

### LINKED ISSUE #6982: Missing validation of thread_schedule_hint.hint in shared_priority_queue_scheduler
# Description

While reviewing the scheduler implementation, I noticed that the `thread_schedule_hint_mode::thread` path in `shared_priority_queue_scheduler` appears to use the hint value directly as a worker index without validating that it falls within the valid worker range.

The code already contains TODO comments indicating that validation may be missing.

---

### `create_thread()`

```cpp
case thread_schedule_hint_mode::thread:
{
    // @TODO. We should check that the thread num is valid
    thread_num = select_active_pu(data.schedulehint.hint);
    domain_num = d_lookup_[thread_num];
    q_index = q_lookup_[thread_num];
    break;
}
```

### `schedule_thread()`

```cpp
case thread_schedule_hint_mode::thread:
{
    // @TODO. We should check that the thread num is valid
    thread_num = select_active_pu(schedulehint.hint, allow_fallback);
    domain_num = d_lookup_[thread_num];
    q_index = q_lookup_[thread_num];
    break;
}
```

---

## Potential Issue

`thread_schedule_hint` stores the hint as:

```cpp
std::int16_t hint;
```

The constructor only checks whether the hint is non-negative when determining the scheduling mode, but it does not ensure that the value is within the range of available worker threads.

If a hint larger than the number of worker threads is provided, it appears possible that the value may be used as an index in several places, including:

* `select_active_pu()`
* `d_lookup_[]`
* `q_lookup_[]`

Since these structures are sized based on `num_workers_`, using an out-of-range hint could potentially lead to out-of-bounds access.

---

## Notes

The `numa` scheduling path already applies modulo normalization:

```cpp
domain_num = fast_mod(schedulehint.hint, num_domains_);
```

which suggests that bounds normalization may have been intended for the `thread` scheduling path as well.

Additionally, both affected locations contain the TODO comment:

```
@TODO. We should check that the thread num is valid
```

which may indicate this validation step was planned but not implemented yet.

---

## Affected File

```
libs/core/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
```

Functions:

* `create_thread()`
* `schedule_thread()`

If this behavior is unintended, I’d be happy to submit a PR to address it.

v4xsh:
Add bounds normalization for thread_schedule_hint_mode::thread path in
create_thread() and schedule_thread() using fast_mod against num_workers_,
consistent with the existing numa hint normalization.

Fixes #6982

## Proposed Changes
- Add `fast_mod(hint, num_workers_)` normalization in `create_thread()` for `thread_schedule_hint_mode::thread` path
- Add `fast_mod(hint, num_workers_)` normalization in `schedule_thread()` for `thread_schedule_hint_mode::thread` path
- Remove stale `@TODO` comments that indicated this validation was missing

The `thread_schedule_hint_mode::thread` path was using the raw `hint` value (typed as `std::int16_t`) directly as a worker index without validating it against the number of available workers. This could lead to out-of-bounds access in `select_active_pu()`, `d_lookup_[]`, and `q_lookup_[]` if the hint exceeded `num_workers_`. The `numa` hint path already applied `fast_mod` normalization, so this fix brings the `thread` path in line with that existing pattern.

## Checklist

### DIFF
--- a/libs/core/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
+++ b/libs/core/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
@@ -368,9 +368,10 @@ namespace hpx::threads::policies {
             case thread_schedule_hint_mode::thread:
             {
                 spq_deb.set(msg, "HINT_THREAD");
-                // @TODO. We should check that the thread num is valid
                 // Create thread on requested worker thread
-                thread_num = select_active_pu(data.schedulehint.hint);
+                thread_num = select_active_pu(
+                    fast_mod(static_cast<std::size_t>(data.schedulehint.hint),
+                        num_workers_));
                 domain_num = d_lookup_[thread_num];
                 q_index = q_lookup_[thread_num];
                 break;
@@ -775,13 +776,13 @@ namespace hpx::threads::policies {
             }
             case thread_schedule_hint_mode::thread:
             {
-                // @TODO. We should check that the thread num is valid
-                // Create thread on requested worker thread
                 spq_deb.set(msg, "HINT_THREAD");
                 spq_deb.debug(debug::str<>("schedule_thread"),
                     "received HINT_THREAD", debug::dec<3>(schedulehint.hint));
-                thread_num =
-                    select_active_pu(schedulehint.hint, allow_fallback);
+                thread_num = select_active_pu(
+                    fast_mod(static_cast<std::size_t>(schedulehint.hint),
+                        num_workers_),
+                    allow_fallback);
                 domain_num = d_lookup_[thread_num];
                 q_index = q_lookup_[thread_num];
                 break;


libs/core/schedulers/include/hpx/schedulers/shared_priority_queue_scheduler.hpp
```diff
@@ -368,9 +368,10 @@ namespace hpx::threads::policies {
             case thread_schedule_hint_mode::thread:
             {
                 spq_deb.set(msg, "HINT_THREAD");
-                // @TODO. We should check that the thread num is valid
                 // Create thread on requested worker thread
-                thread_num = select_active_pu(data.schedulehint.hint);
+                thread_num = select_active_pu(
+                    fast_mod(static_cast<std::size_t>(data.schedulehint.hint),
```
hkaiser, line 373: Where is `fast_mod` defined?

v4xsh: `fast_mod` is defined in `queue_holder_thread.hpp` in namespace `hpx::threads::policies`. `shared_priority_queue_scheduler.hpp` already includes `queue_holder_thread.hpp`, so `fast_mod` is visible here without any additional include. The same file already uses `fast_mod` in existing paths, for example:

```cpp
domain_num = fast_mod(schedulehint.hint, num_domains_);
```

hkaiser: Ohh ok. I was not aware of this ;-)

PR #7029, opened 2026-03-15
fix logic bug in async_replay exhaustion handling

### LINKED ISSUE #7028: async_replay_validate incorrectly throws abort_replay_exception when the last attempt succeeds
## Summary

The `async_replay_validate` (and `async_replay`) implementation contains a logic error that causes it to throw an `abort_replay_exception` if the task succeeds on its **final allowed attempt** (when the retry counter `n` is `0`).  

The expected behavior is to **return the successful result**, but the code currently treats the exhaustion of retries as a failure even if the last run was successful.

---

## Affected File

[`libs/core/resiliency/include/hpx/resiliency/async_replay.hpp`](https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/resiliency/include/hpx/resiliency/async_replay.hpp)

---

## The Bug (Code Snippet)

The issue occurs in the continuation logic of `async_replay_helper::call(std::size_t n)`:

```cpp
84:                         auto&& result = f.get();
85: 
86:                         if (!HPX_INVOKE(this_->pred_, result))
87:                         {
                                // ... retry logic ...
97:                             throw abort_replay_exception();
98:                         }
99: 
100:                        if (n != 0)  // <--- BUG IS HERE
101:                        {
102:                            // return result
103:                            return hpx::make_ready_future(HPX_MOVE(result));
104:                        }
105: 
106:                        // throw aborting exception as attempts were
107:                        // exhausted
108:                        throw abort_replay_exception();
```

---

## Detailed Explanation

The parameter `n` represents the **number of retries remaining**.

- If `n = 1`, and the task fails, it retries with `n = 0`.
- If that retry (`n = 0`) **succeeds**, it reaches line **100**.
- Since `n != 0` is **false**, the function skips the return statement.
- Execution falls through to line **108**, throwing `abort_replay_exception()`.

### Example Scenario

```cpp
hpx::resiliency::experimental::async_replay(1, task);
```

Total attempts allowed: **2**

1. First attempt → **fails**
2. Second attempt (final attempt) → **succeeds**

**Current Result:**  
The user receives an `abort_replay_exception`.

**Expected Result:**  
The user should receive a **future containing the successful result of the second attempt**.

---

## Expected Fix

The check `if (n != 0)` should **not be part of the success path**.  
Once the result has passed the predicate (line 86), it should be **returned immediately**.

```diff
-                        if (n != 0)
-                        {
-                            // return result
-                            return hpx::make_ready_future(HPX_MOVE(result));
-                        }
-
-                        // throw aborting exception as attempts were
-                        // exhausted
-                        throw abort_replay_exception();
+                        return hpx::make_ready_future(HPX_MOVE(result));
```

---

## Impact

- Incorrect exception thrown despite successful execution
- Breaks expected retry semantics
- Can cause unexpected failures in applications using `async_replay`

---

## Fix Complexity

**Low** — small logic correction in the success path.

LSUDOKO:
Fixes #7028

## Proposed Changes

- Modified `async_replay_helper::call` in [libs/core/resiliency/include/hpx/resiliency/async_replay.hpp](cci:7://file:///home/arpit/Desktop/Gsoc/hpx/libs/core/resiliency/include/hpx/resiliency/async_replay.hpp:0:0-0:0) to remove the incorrect `if (n != 0)` check in the success path.
- Ensured that a successful result is returned as a ready future even if it was obtained on the last possible attempt (when the retry counter `n` is `0`).
- Verified [libs/core/resiliency/include/hpx/resiliency/async_replicate.hpp](cci:7://file:///home/arpit/Desktop/Gsoc/hpx/libs/core/resiliency/include/hpx/resiliency/async_replicate.hpp:0:0-0:0) to ensure it does not share this logic bug (it handles results after synchronization and is correct).

In the previous implementation of [async_replay](cci:1://file:///home/arpit/Desktop/Gsoc/hpx/libs/core/resiliency/include/hpx/resiliency/async_replay.hpp:41:12-47:13), the retry logic was structured such that if the task succeeded on its **final** allowed attempt (i.e., when the retry counter reached `0`), the code would skip the return statement and fall through to `throw abort_replay_exception()`.

This effectively meant that [async_replay(1, task)](cci:1://file:///home/arpit/Desktop/Gsoc/hpx/libs/core/resiliency/include/hpx/resiliency/async_replay.hpp:41:12-47:13) would throw an exception even if the second attempt was perfectly successful. The logic now correctly returns the result as soon as the predicate passes, regardless of the value of `n`.

## Checklist

### DIFF
--- a/libs/core/resiliency/tests/regressions/async_replay_exhaustion.cpp
+++ b/libs/core/resiliency/tests/regressions/async_replay_exhaustion.cpp
@@ -1,4 +1,4 @@
-//  Copyright (c) 2026 Hartmut Kaiser
+//  Copyright (c) 2026 Arpit Singh
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying


hkaiser:
@LSUDOKO Could you contruct a test case that reproduces the issue, please?

LSUDOKO:
> @LSUDOKO Could you contruct a test case that reproduces the issue, please?

hi @hkaiser 
i  contruct a test case that reproduces the issue

LSUDOKO:
> Thanks for adding the test. By convention, we add tests that fix a particular problem as regression tests in the corresponding directory `tests/regression`. Could you move your test over, please? You may want to look at other modules to see what the necessary changes to the CMakeLists.txt could look like.

@hkaiser 
now i move the test to tests/regression dir

hkaiser:
@LSUDOKO The clang-format and inspect checks complain. Could you please have a look at those issues?

hkaiser:
@LSUDOKO inspect is still not happy.

libs/core/resiliency/tests/regressions/async_replay_exhaustion.cpp
```diff
@@ -0,0 +1,88 @@
+//  Copyright (c) 2026 Hartmut Kaiser
```
hkaiser: Please add your own copyright here. I have not contributed to this file.

LSUDOKO: @hkaiser 
okay sir I added my copyright

hkaiser:
@LSUDOKO Thanks for your work and your patience on this!

Issue #3, opened 2012-07-10
HPX shutdown hangs on multilocality runs


brycelelbach:
**[reported by dstark]** _[Trac time Fri Aug 13 19:00:03 2010]_ Shutdown fails for different tests at different times (aren't
races fun), but I believe that each test is simply timing out because
one or more localities are left waiting on a lock when stopping.
Here's what I found. The hpx_run command which ctest uses is (via the
-V option):

```
   $ /home/dstark/projects/hpx_svn/tools/hpx_run.py
/home/dstark/sw/hpx_svn_relwithdebinfo_test/tests/multilocality_tests/startup_test
-l 2:4
```

The two startup_test commands that hpx_run uses are (via the -d option):

```
   $ /home/dstark/sw/hpx_svn_relwithdebinfo_test/tests/multilocality_tests/startup_test
-r -a localhost:2222 -x localhost:2223 -l 2 -t 4
   $ /home/dstark/sw/hpx_svn_relwithdebinfo_test/tests/multilocality_tests/startup_test
-w -a localhost:2222 -x localhost:2224 -l 2 -t 4
```

After testing these buy hand for some number of runs, only the worker
instance ever stalls, and the other instance always returns "Test
passed" and exits normally. When running the worker in GDB, I
consistently get the following back trace when it stalls:

```
#0  0x000000395540d2e4 in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x0000003955408c3a in _L_lock_1034 () from /lib64/libpthread.so.0
#2  0x0000003955408afc in pthread_mutex_lock () from /lib64/libpthread.so.0
#3  0x00002ad668d41d77 in
boost::condition_variable_any::wait<boost::unique_lock<boost::mutex> >
(this=0x7fffde4a0350, m=@0x7fffde4a0440)
   at /home/dstark/usr/local/boost_1_43_0/include/boost/thread/pthread/mutex.hpp:51
#4  0x00002ad668d4fc25 in
hpx::runtime_impl<hpx::threads::policies::global_queue_scheduler,
hpx::threads::policies::callback_notifier>::stop (
   this=0x7fffde4a0540, blocking=true)
   at /home/dstark/projects/hpx_svn/src/runtime.cpp:372
#5  0x00002ad668d53d37 in
hpx::runtime_impl<hpx::threads::policies::global_queue_scheduler,
hpx::threads::policies::callback_notifier>::run (
[...]

PR #7032, opened 2026-03-15
Fix numa allocator test 7027

### LINKED ISSUE #7027: Fix failing test related to numa_allocator
We're seeing occasional test failures for our tests related to the numa_allocator (target: tests.unit.modules.compute_local.numa_allocator). We should try to identify the problem and fix it. I'll add an example link here as soon as I see the error again.

GitMasterJatin:
Fixes #7027

Changes made:

numa_allocator.hpp

- Integer division truncation fixed — last executor partition now covers cnt % num_executors remainder elements so no pages go untouched
- Dangling topo_ reference removed — now calls hpx::threads::get_topology() directly in allocate() and deallocate(), no more stored reference member

numa_binding_allocator.hpp
 3. UB volatile read eliminated — *vaddr = *vaddr (reading uninitialized memory, optimizable to no-op) replaced with *reinterpret_cast<volatile char*>(page_ptr) = 0 — a guaranteed physical write that triggers first-touch

The <DEB> log lines in the test output are expected (test has #define QUEUE_HOLDER_NUMA_DEBUG true).

## Checklist

### DIFF
--- a/libs/core/compute_local/include/hpx/compute_local/host/numa_allocator.hpp
+++ b/libs/core/compute_local/include/hpx/compute_local/host/numa_allocator.hpp
@@ -53,15 +53,14 @@ namespace hpx::parallel::util {
             using other = numa_allocator<U, Executors>;
         };
 
-        numa_allocator(Executors const& executors, hpx::threads::topology& topo)
+        numa_allocator(
+            Executors const& executors, hpx::threads::topology& /* topo */)
           : executors_(executors)
-          , topo_(topo)
         {
         }
 
         numa_allocator(numa_allocator const& rhs)
           : executors_(rhs.executors_)
-          , topo_(rhs.topo_)
         {
         }
 
@@ -70,7 +69,6 @@ namespace hpx::parallel::util {
         template <typename U>
         numa_allocator(numa_allocator<U, Executors> const& rhs)
           : executors_(rhs.executors_)
-          , topo_(rhs.topo_)
         {
         }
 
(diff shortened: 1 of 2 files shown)


hkaiser:
Excellent analysis! Thanks!

hkaiser:
@GitMasterJatin Did you by any chance look into the page size issue @LSUDOKO mentioned here: https://github.com/STEllAR-GROUP/hpx/issues/7027#issuecomment-4063748281?

hkaiser:
@GitMasterJatin clang-format is not happy, also your changes break compilation.

GitMasterJatin:
Yes @hkaiser Sir, I've addressed the page size concern raised by @LSUDOKO in this updated commit.

In get_numa_domain(), the assertion:

HPX_ASSERT((reinterpret_cast<std::size_t>(page) & 4095) == 0);
was hardcoded to assume 4KB pages. I've replaced it with:

HPX_ASSERT(
    (reinterpret_cast<std::size_t>(page) &
        (threads::get_memory_page_size() - 1)) == 0);
This makes the alignment check correct on any platform regardless of page size.

Regarding the move_pages / -ENOENT concern @LSUDOKO raised (pages returning - when not yet faulted): this is actually addressed by our fix #3 (the touch_pages UB fix). By writing *reinterpret_cast<volatile char*>(page_ptr) = 0 instead of the old *vaddr = *vaddr, we guarantee every page is faulted into RAM before the test queries get_page_numa_domains(). So the -ENOENT path should no longer be hit in practice.

I've also fixed a compilation error (lambda [i] was missing this needed for executors_[i] in the HPX_DEBUG path), applied clang-format, and removed accidentally committed Testing/Temporary/ files.

libs/core/compute_local/include/hpx/compute_local/host/numa_allocator.hpp
```diff
@@ -53,15 +53,14 @@ namespace hpx::parallel::util {
             using other = numa_allocator<U, Executors>;
         };
 
-        numa_allocator(Executors const& executors, hpx::threads::topology& topo)
+        numa_allocator(
+            Executors const& executors, hpx::threads::topology& /* topo */)
```
hkaiser, line 57: While I have no issues with this as it doesn't really change anything (the reference that was used to construct the allocator was stable after all), I'd like to know if you were seeing actual problem that triggered this modification.

GitMasterJatin: @hkaiser Sir you're right that in the existing call sites the reference is stable (the topology singleton lives for the entire HPX runtime lifetime).

I did not observe an actual crash or test failure triggered specifically by this issue. The concern was theoretical: a hypothetical future use of numa_allocator where a caller constructs it with a locally-scoped topology object and then lets async first-touch tasks outlive it. In practice, with HPX's singleton topology, this cannot happen.

GitMasterJatin: It doesn't change observable behavior, I'm happy to revert this specific change and keep only the two fixes that actually address the observed intermittent test failures.

hkaiser: That's fine. Please leave it in place as you have it.

GitMasterJatin:
Thanks @hkaiser Sir for reviewing and closing the issue

PR #7038, opened 2026-03-16
[Core] Use hpx::parallel::detail::distance in unsequenced algorithms

aneek22112007-tech:
Hello! I'm a student learning about parallel algorithms and I was browsing the HPX codebase. I noticed that some of the unsequenced algorithms (in loop.hpp, reduce.hpp, and transform_loop.hpp) were using std::distance.

From what I've learned, std::distance in C++17 can have issues when working with different-typed sentinels, and I saw that HPX has its own hpx::parallel::detail::distance utility which is designed to handle this more robustly.

I've replaced the std::distance calls with the HPX specific one. This should make the algorithms more consistent with the rest of the library and improve support for C++20-style sentinels. I've also made sure to include the necessary header.

I've checked the changes and they seem to follow the HPX coding style. I'd really appreciate it if you could take a look! Thanks!

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/unseq/transform_loop.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/unseq/transform_loop.hpp
@@ -139,8 +139,8 @@ namespace hpx::parallel::util {
                 if constexpr (iterators_are_random_access)
                 {
                     auto&& in_out = util::transform_loop_n<
-                        hpx::execution::unsequenced_policy>(
-                        it, hpx::parallel::detail::distance(it, last), dest,
+                        hpx::execution::unsequenced_policy>(it,
+                        hpx::parallel::detail::distance(it, last), dest,
                         HPX_FORWARD(F, f));
 
                     return util::in_out_result<InIter, OutIter>{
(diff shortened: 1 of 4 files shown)


libs/core/algorithms/include/hpx/parallel/unseq/transform_loop.hpp
```diff
@@ -11,6 +11,7 @@
 #include <hpx/modules/executors.hpp>
 #include <hpx/parallel/util/transform_loop.hpp>
 
+#include <hpx/parallel/algorithms/detail/distance.hpp>
```
hkaiser: Nitpick: please group HPX headers separately from standard headers.

aneek22112007-tech: Thank you for the review! I'm still learning the HPX coding guidelines, so I really appreciate the feedback on header grouping. I've updated the PR to separate the HPX headers from the standard ones. Thanks again for your time and guidance!

hkaiser (dismissed):
LGTM, thanks!

hkaiser:
@aneek22112007-tech Could you please have a look at the clang-format issues reported?

aneek22112007-tech:
> @aneek22112007-tech Could you please have a look at the clang-format issues reported?

Hi @hkaiser! I've fixed those formatting issues. I also accidentally missed the DCO sign-off on the last commit, but I've amended that now too. Still getting used to the workflow—thanks for your patience with me!"

PR #7039, opened 2026-03-16
fix compiler inconsistencies for fundamental types in reflecting names

sleepingeight:


hkaiser:
@ujjwal-shekhar Would you have the time to review?

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -73,91 +73,41 @@ namespace hpx::serialization::detail {
...
+            for (auto const& [type_info, name] : lookup_table)
             {
-                return fixed_string("std::nullptr_t");
+                if (dealias(type_info) ==
+                    dealias(^^base_type_t<typename[:Scope:]>))
+                {
+                    return name;
+                }
             }
+            return std::string_view{};
```
hkaiser: Should we use:
```suggestion
           return std::string_view("<unknown>");
```
instead?

ujjwal-shekhar:
> @hkaiser: @ujjwal-shekhar Would you have the time to review?

I'll take a look soon and add my comments. Thank you @sleepingeight for looking into this!

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -62,10 +63,53 @@ namespace hpx::serialization::detail {
...
+
+            for (auto const& [type_info, name] : lookup_table)
+            {
+                if (dealias(type_info) ==
+                    dealias(^^base_type_t<typename[:Scope:]>))
+                {
+                    return name;
+                }
+            }
+            return std::string_view{};
```
ujjwal-shekhar: I think we should also emit a warning along the lines of
> Cannot auto-generate the name of type provided.

Additionally, we could use `std::meta::display_string_of` as a fallback and warn the user about it: 
> Falling back to compiler-defined std::meta::display_string_of, consider manually defining the portable string for this type.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -74,13 +118,68 @@ namespace hpx::serialization::detail {
...
+                            std::is_volatile_v<base_type>)
+                            return fixed_string("const volatile ") + ftype_name;
+                        else if constexpr (std::is_const_v<base_type>)
+                            return fixed_string("const ") + ftype_name;
+                        else if constexpr (std::is_volatile_v<base_type>)
+                            return fixed_string("volatile ") + ftype_name;
+                        else
+                            return ftype_name;
+                    }();
+
```
ujjwal-shekhar: nit: Perhaps adding very small comments separating/explaining casework would be nice. It isn't necessary to add them everywhere, but just the top-level `if constexpr`s are enough.

libs/core/serialization/tests/unit/refl/reflection_qualified_name_of.cpp
```diff
@@ -51,10 +52,13 @@ int main()
...
-        using type = a::b::c::d::pentagon<int, char, double, float, long>;
+        using type = a::b::c::d::pentagon<int, char, double, float, long,
+            signed char const volatile* const, int&&, double&>;
         char const* name = qualified_name_of<type>::get();
         HPX_TEST_EQ(std::string(name),
-            std::string("a::b::c::d::pentagon<int,char,double,float,long>"));
+            std::string(
+                "a::b::c::d::pentagon<int,char,double,float,long int,const "
+                "volatile signed char* const,int&&,double&>"));
     }
```
ujjwal-shekhar, line 78: Could we also check if CRTP types are being named properly as well?

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -74,13 +118,68 @@ namespace hpx::serialization::detail {
...
+                    else
+                    {
+                        return base;
+                    }
+                }
+                else
+                {
+                    // Fallback for missed types such as functions, lambdas, etc.
+                    constexpr auto name_view =
+                        std::meta::display_string_of(Scope);
```
ujjwal-shekhar, line 193: Same here, we could consider emitting a warning each time something compiler-defined is used. Also, we should add testcases that trigger this with a TODO comment describing the issue (and a similar test that invokes the `<unknown>` name above).

I think it is unlikely that this feature will be used for such odd types, but let's try to avoid footguns and error/warn the user at build time as much as possible.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -62,10 +63,53 @@ namespace hpx::serialization::detail {
     template <std::size_t N>
     fixed_string(char const (&)[N]) -> fixed_string<N - 1>;
 
+    template <typename T>
+    using base_type_t = std::remove_cv_t<
+        std::remove_pointer_t<std::remove_reference_t<std::remove_cv_t<T>>>>;
+
     // Helper to recursively build the enclosing namespaces of a type
     template <std::meta::info Scope>
     struct scope_builder
```
ujjwal-shekhar, line 77: nit: We could have a compile-time flag that gets flipped if any component of `qualified_name` goes to an unhandled case. A `static_assert` on this flag could then error out at build time (with appropriate info for the user).

CC: @hkaiser is this a good idea?

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
sleepingeight: I really like the idea of warning users, but I don't think we have a proper way to generate warnings in c++ during compile time. One option is to use `[[deprecated]]` to generate warnings ([ref](https://godbolt.org/z/Gs9a8MGr8)), but it does not seem very much viable to me. Are there any other ways in which we can do this?

ujjwal-shekhar: We could add a second param to the `HPX_POLYMORPHIC_AUTO_REGISTER(Class)` macro, `HPX_POLYMORPHIC_AUTO_REGISTER(Class, AllowUnsafe)`. The second arg would carry the intent of allowing potentially non-portable strings (default: false). If the generated string is ill-formed (which we will know at compile-time); we would trigger a static_assert that:
- Explains the source of the error to the user (why is this ill-formed).
- Informs the user that they have an option to fallback to a compiler-defined behaviour by passing in a second param to the macro.
- Also tell the user that they can write their own portable string for this type if needed. _Nit: I think we should add a test to see if the user providing their own fully specified `get_serialization_name` interacts well with the auto generated ones._

A downside is that the code changes might have to trickle the second param down from the macro all the way to the refl_qualified_name_of part, but we can discuss it in detail later.

Thoughts?

sleepingeight: Got it. To summarise, we have two cases where we may want to produce warnings:
1. When we are unable to generate a string for fundamental types.
2. When compiler-defined `std::meta::display_string_of` is used.

For case 1, my understanding is that this should never really happen. We are expected to support all fundamental types, and this is a straightforward and fully controlled part of the implementation. If this path is ever reached, it likely means there is a bug in our code. Because of that, adding warnings or handling this as a user-facing case does not seem useful. It should be sufficient if we provide a hard error here with a static_assert.

For case 2, I agree with the concern. When we fall back to std::meta::display_string_of, the result may not be portable across compilers. This can become an issue for serialization. So it makes sense to inform the user in this case. But one thing which concerns me is that case 2 is only hit during types such as lambdas or functions. I don't think HPX supports serializing them, so making efforts here seems not very impactful (I might be wrong here, please feel free to correct me).

What do you think?

hkaiser: Let's just `static_assert` on anything unexpected. Let's not over-complicate things.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
hkaiser: I agree.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
ujjwal-shekhar: I agree, there are always new features being introduced; plus, any unexpected behaviour triggering asserts will help us catch compatibility issues easily. Since these asserts are all compile time we don't need to worry about runtime performance either. Also, I believe types like `__int_128` aren't covered by clang's pretty printer (there may be even more types that we miss).

hkaiser:
@sleepingeight please also have a look at the reported inspect issue:
```
/libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp: *I* missing #include (utility) for symbol std::pair on line 76
```

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -99,15 +100,25 @@ namespace hpx::serialization::detail {
...
+                        return search_result{true, name};
+                    }
                 }
-            }
-            return std::string_view{};
+                return search_result{false, "<unknown>"};
+            }();
+
+            static_assert(result.found, "Unknown fundamental type");
+            return result.name;
```
hkaiser, line 127: I think this could be simplified. You could get rid of the lambda (and the `search_result` type) altogether and simply return the name from the function:
```suggestion
            std::string_view result;
            for (auto const& [type_info, name] : lookup_table)
            {
                if (dealias(type_info) ==
                    dealias(^^base_type_t<typename[:Scope:]>))
                {
                    result = name;
                    break;
                }
            }
            static_assert(result.size() != 0, "Unknown fundamental type");
            return result;
```

hkaiser:
@sleepingeight Could you please rebase and fix the conflicts?

ujjwal-shekhar:
> @hkaiser: @ujjwal-shekhar COuld also have a look, please?

I will add my comments soon!

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
```diff
@@ -62,10 +64,63 @@ namespace hpx::serialization::detail {
...
+                    if (dealias(type_info) ==
+                        dealias(^^base_type_t<typename[:Scope:]>))
+                    {
+                        return search_result{true, name};
+                    }
+                }
+                return search_result{false, "<unknown>"};
+            }();
+
+            static_assert(result.found, "Unknown fundamental type");
```
ujjwal-shekhar: Adding a bit of context for the user might be nice. Something like "Unknown fundamental type. Autogenerated serialization for polymorphic cases are not supported for this type yet."

libs/core/serialization/tests/unit/refl/reflection_qualified_name_of.cpp
```diff
@@ -75,5 +92,20 @@ int main()
...
+        char const* base_name = qualified_name_of<base_type>::get();
+        HPX_TEST_EQ(std::string(base_name),
+            "crtp::base<crtp::derived<const volatile short int* const&&>,const "
+            "volatile long long int* const>");
+
+        using derived_type = crtp::derived<wchar_t&>;
+        char const* derived_name = qualified_name_of<derived_type>::get();
+        HPX_TEST_EQ(std::string(derived_name), "crtp::derived<wchar_t&>");
+    }
+
```
ujjwal-shekhar, line 112: Nice, could please we add one last test to see if the user manually defining a string nicely overrides our auto-generation facility? I believe you will have to use one of the old macros for this.

hkaiser:
@sleepingeight ping? Do you plan to move this PR forward?

sleepingeight:
Yes, @hkaiser. I'm resolving the comments.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
sleepingeight: This wouldn't work because result is varying and cannot be used in a constant expression. Detailed error:
```cpp
/workspace/libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp:117:38:   in 'constexpr' expansion of 'result.std::basic_string_view<char>::size()'
  117 |             static_assert(result.size() != 0, "Unknown fundamental type. Autogenerated serialization for polymorphic cases are not supported for this type yet.");
      |                           ~~~~~~~~~~~^~
/usr/local/include/c++/16.0.1/string_view:238:22: error: the value of 'result' is not usable in a constant expression [-Wtemplate-body]
  238 |       { return this->_M_len; }
      |                ~~~~~~^~~~~~
/workspace/libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp:107:30: note: 'result' was not declared 'constexpr'
  107 |             std::string_view result;
      |                              ^~~~~~
```

hkaiser: Well, then instead of re-assigning `result` (L109), couldn't we simply return from there?
```suggestion
            for (auto const& [type_info, name] : lookup_table)
            {
                if (dealias(type_info) ==
                    dealias(^^base_type_t<typename[:Scope:]>))
                {
                    return name;
                }
            }
            return std::string_view{};
```

sleepingeight: We could do this, but this wouldn't allow us to provide detailed error message using static assert (as `static_assert(false, "...")` would result in a compile error irrespective of the lookup result).

ujjwal-shekhar: Just an idea: We could wrap the returned result in a `std::optional`? I believe a static_assert on `has_value` will do the trick. Please correct me if I have missed something:)

hkaiser: Let's leave it as is for now. We can revisit the simplification later.

libs/core/serialization/include/hpx/serialization/detail/refl_qualified_name_of.hpp
hkaiser: Same here, let's leave this as is for now. We can cross that bridge if we ever get there.

hkaiser:
@ujjwal-shekhar I think this is ready to be merged. What's your take?

ujjwal-shekhar:
> @hkaiser: @ujjwal-shekhar I think this is ready to be merged. What's your take?

I will take one more look if that's alright!

hkaiser:
@sleepingeight You may also want to rebase onto master, that should fix the compilation problems we're seeing here.

ujjwal-shekhar:
LGTM. Thank you for working on this @sleepingeight !

hkaiser:
@sleepingeight @ujjwal-shekhar Thanks for looking into cleaning this up properly!

PR #7041, opened 2026-03-16
Fix exception safety risk in options_description_easy_init

### LINKED ISSUE #7040: exception Safety Risk in `options_description_easy_init` Could Lead to Memory Leaks
## Bug Report

**Module:** `hpx::program_options`  
**File:** `libs/core/program_options/src/options_description.cpp`

---

## Description

While reviewing the `hpx::program_options` module, I found a potential memory leak in `options_description_easy_init::operator()` around **line ~256**, which is already marked with a `FIXME` comment.

```cpp
// FIXME: does not look exception-safe
std::shared_ptr<option_description> d =
    std::make_shared<option_description>(
        name, new untyped_value(true), description);
```

## Root Cause

The expression new untyped_value(true) allocates a raw heap object and immediately passes it as an argument to std::make_shared. This can be unsafe because:

new untyped_value(true) runs first  heap memory is allocated.

std::make_shared<option_description>(...) runs next.

If std::make_shared throws std::bad_alloc (for example, failing to allocate the control block), the raw pointer is leaked because it was never captured by a std::shared_ptr.

This violates the rule that raw new should not appear directly inside function argument lists, since it may lead to memory leaks if another operation throws before ownership is transferred

## Impact

This is a silent memory leak triggered only when memory allocation fails, which is hard to detect in testing.
affects users who rely on hpx::program_options for command-line option parsing.

LSUDOKO:
## Proposed Changes

In `libs/core/program_options/src/options_description.cpp`, the `operator()` overload that takes just `name` and `description` contained a raw `new` expression passed directly as a constructor argument:

### Before (unsafe)

```cpp
// FIXME: does not look exception-safe
std::shared_ptr<option_description> d =
    std::make_shared<option_description>(
        name, new untyped_value(true), description);
```

## Fix
this PR fixes the issue by wrapping the semantic object in a std::shared_ptr first:
After (exception-safe)
```cpp
std::shared_ptr<value_semantic const> semantic =
    std::make_shared<untyped_value>(true);

std::shared_ptr<option_description> d =
    std::make_shared<option_description>(
        name, semantic.get(), description);
```
This ensures that if make_shared<option_description> fails, the untyped_value object is properly cleaned up when the semantic smart pointer goes out of scope.

fix #7040

### DIFF
--- a/libs/core/program_options/include/hpx/program_options/options_description.hpp
+++ b/libs/core/program_options/include/hpx/program_options/options_description.hpp
@@ -40,22 +40,11 @@ namespace hpx::program_options {
 
         /** Initializes the object with the passed data.
 
-            Note: it would be nice to make the second parameter auto_ptr,
-            to explicitly pass ownership. Unfortunately, it's often needed to
-            create objects of types derived from 'value_semantic':
-               options_description d;
-               d.add_options()("a", parameter<int>("n")->default_value(1));
-            Here, the static type returned by 'parameter' should be derived
-            from value_semantic.
-
-            Alas, derived->base conversion for auto_ptr does not really work,
-            see
-            http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2000/n1232.pdf
-            http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#84
-
-            So, we have to use plain old pointers. Besides, users are not
-            expected to use the constructor directly.
-
+            Note: it is generally preferred to use the std::unique_ptr
+            overloads to explicitly pass ownership. We retain the raw pointer
+            overloads for backward compatibility with legacy code that
+            relies on implicit derived-to-base conversions which were historically
+            problematic with std::auto_ptr.
 
             The 'name' parameter is interpreted by the following rules:
             - if there's no "," character in 'name', it specifies long name
@@ -66,8 +55,8 @@ namespace hpx::program_options {
 
         /** Initializes the class with the passed data.
          */
-        option_description(char const* name,
-            std::shared_ptr<value_semantic const> s);
+        option_description(
+            char const* name, std::unique_ptr<value_semantic const> s);
 
         /** Initializes the class with the passed data.
          */
@@ -77,7 +66,7 @@ namespace hpx::program_options {
         /** Initializes the class with the passed data.
          */
         option_description(char const* name,
-            std::shared_ptr<value_semantic const> s, char const* description);
+            std::unique_ptr<value_semantic const> s, char const* description);
 
         virtual ~option_description();
 
--- a/libs/core/program_options/include/hpx/program_options/options_description.hpp
+++ b/libs/core/program_options/include/hpx/program_options/options_description.hpp
@@ -66,8 +66,18 @@ namespace hpx::program_options {
 
         /** Initializes the class with the passed data.
          */
+        option_description(char const* name,
+            std::shared_ptr<value_semantic const> s);
+
+        /** Initializes the class with the passed data.
+         */
         option_description(
             char const* name, value_semantic const* s, char const* description);
+
+        /** Initializes the class with the passed data.
+         */
+        option_description(char const* name,
+            std::shared_ptr<value_semantic const> s, char const* description);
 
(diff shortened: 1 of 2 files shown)


libs/core/program_options/src/options_description.cpp
```diff
@@ -253,10 +253,13 @@ namespace hpx::program_options {
...
         // no value can be specified on command line.
-        // FIXME: does not look exception-safe
+        // Store the semantic in a shared_ptr first to ensure exception safety:
+        // if make_shared<option_description> fails, the semantic won't leak.
+        std::shared_ptr<value_semantic const> semantic =
+            std::make_shared<untyped_value>(true);
         std::shared_ptr<option_description> d =
             std::make_shared<option_description>(
-                name, new untyped_value(true), description);
+                name, semantic.get(), description);
```
hkaiser: This will cause for the memory held by `semantic` to be released when the object goes out of scope, leaving `d` with a dangling pointer.

LSUDOKO: @hkaiser 
i got it 
when semantic goes out of scope after the function 
it deletes the object  leaving d->m_value_semantic dangling!
let me use unique_ptr

libs/core/program_options/src/options_description.cpp
```diff
@@ -253,10 +253,14 @@ namespace hpx::program_options {
...
-            std::make_shared<option_description>(
-                name, new untyped_value(true), description);
+        //
+        // Use unique_ptr to hold the semantic temporarily. If the
+        // option_description constructor or its allocation throws, the semantic
+        // is automatically freed. We only release() ownership after the
+        // option_description has been successfully constructed.
+        std::unique_ptr<untyped_value> semantic(new untyped_value(true));
+        std::shared_ptr<option_description> d(
+            new option_description(name, semantic.release(), description));
```
hkaiser: This will have the same issues as the initial code. If the second allocation fails, then the first allocation will never be released.

LSUDOKO: @hkaiser 
now fix introduces proper std::shared_ptr constructors tooption_description, allowing options_description_easy_init tosafely pass a shared_ptr
this resolves both the memory leakand the dangling pointer issue

libs/core/program_options/include/hpx/program_options/options_description.hpp
```diff
@@ -64,11 +64,21 @@ namespace hpx::program_options {
         */
         option_description(char const* name, value_semantic const* s);
 
+        /** Initializes the class with the passed data.
+         */
+        option_description(char const* name,
+            std::shared_ptr<value_semantic const> s);
```
hkaiser: Have you read the comment right above? Does your solution solve that problem as well? That comment still talks about `std::auto_ptr`, which doesn't work. Could we use `std::unique_ptr` instead?

LSUDOKO: you're right, thank you i've refactored the fix to use std::unique_ptr overloads for the constructors and updated the outdated documentation to remove the references to std::auto_ptr

libs/core/program_options/src/options_description.cpp
```diff
@@ -253,10 +268,14 @@ namespace hpx::program_options {
     {
         // Create untyped semantic which accepts zero tokens: i.e.
         // no value can be specified on command line.
-        // FIXME: does not look exception-safe
+        //
+        // By creating a unique_ptr first, and passing it to the new
+        // option_description constructor that accepts a unique_ptr, we avoid
+        // any possibility of a memory leak if allocation fails.
+        std::unique_ptr<value_semantic const> semantic(new untyped_value(true));
```
hkaiser: You could use `std::make_unique` here.

hkaiser (commented):
Otherwise this is getting closer to a solution. Thanks!

libs/core/program_options/src/options_description.cpp
LSUDOKO: @hkaiser 
okay got it

hkaiser (approved):
LGTM, thanks!

PR #7047, opened 2026-03-17
Fix four critical bugs in sheneos distributed interpolation example

Sanchit2662:
While reading through HPX's sheneos distributed interpolation example as part of my GSoC proposal research, I found four bugs that were silently corrupting behavior.

  ---
  1. Synchronous init never actually waited (partition3d.hpp:79)

  I noticed init() was using a comma instead of a dot — meaning the future from init_async() was being discarded, and get() was calling client_base::get() (returns the component id) instead of future::get(). The partition was never actually initialized before use.
```
  // Before
  init_async(datafilename, dimx, dimy, dimz), get();

  // After
  init_async(datafilename, dimx, dimy, dimz).get();
```
  ---
  2. Ghost zone count off-by-one corrupted partition boundaries (server/partition3d.cpp:52)

  When adding ghost zones for boundary overlap, the code correctly extended dim_[d].count_ by 2, but only incremented the local count by 1. This made values[count - 1] point into a ghost zone, giving every partition a wrong max_value_.
```
  // Before
  dim_[d].count_ += 2;
  ++count;

  // After
  dim_[d].count_ += 2;
  count += 2;
```
  ---
  3. Double increment skipped half the AGAS unregistrations (interpolator.cpp:98)

  The loop already had ++i in the for-header, but there was also an i++ inside std::to_string(i++). Every odd-indexed partition name was silently leaking in AGAS.
```
  // Before
  for (std::size_t i = 0; i < partitions_.size(); ++i)
      hpx::agas::unregister_name(..., data.symbolic_name_ + std::to_string(i++));

  // After
      hpx::agas::unregister_name(..., data.symbolic_name_ + std::to_string(i));
```
  ---
  4. Copy-paste error made test parameters meaningless (sheneos_test.cpp:327-328)

  All three spatial dimension parameters were reading from "num-ye-points", making it impossible to test independent dimensional behavior.
```
  // Before
  std::size_t const num_temp_points = vm["num-ye-points"].as<std::size_t>();
  std::size_t const num_rho_points  = vm["num-ye-points"].as<std::size_t>();

  // After
  std::size_t const num_temp_points = vm["num-temp-points"].as<std::size_t>();
  std::size_t const num_rho_points  = vm["num-rho-points"].as<std::size_t>();
```


hkaiser:
Could you please have a look at the clang-format issues reported?

Sanchit2662:
Hi @hkaiser , I have fixed clang-format issues.

hkaiser:
@Sanchit2662 Thanks again for looking into this!

Issue #7046, opened 2026-03-17
Replace unsafe `std::strcat` with safer string operations in `print.cpp`


LSUDOKO:
## Description

The file 
print.cpp
 at line 289 uses the unsafe C-style string function std::strcat, which is prone to buffer overflow vulnerabilities. This function does not perform bounds checking and can write beyond the allocated buffer if the source string is too long.

```cpp
std::strcat(hostname_, temp.c_str());  // unsafe
```
### Why This Matters

- Fixed buffer: char hostname_[32]
- No bounds checking → risk of overflow
- GCC warning (-Wrestrict) is suppressed instead of fixed

hkaiser:
This was fixed by merging #7048

PR #7048, opened 2026-03-17
Fix formatting issue #7046 in print.cpp

mmud:
Fixes #

## Proposed Changes

Replace unsafe std::strcat with bounded std::snprintf to prevent potential buffer overflow on the fixed char hostname_[32] buffer.
Remove unnecessary temporary std::string allocations.

std::strcat does not check buffer bounds. Writing into the fixed char hostname_[32] buffer without a size limit risks overflow if the hostname or rank string grows in the future.

std::snprintf does the same job but respects the buffer size, uses no heap allocations.

Existing tests pass. No public API changed.

## Checklist

### DIFF
--- a/libs/core/debugging/src/print.cpp
+++ b/libs/core/debugging/src/print.cpp
@@ -17,6 +17,7 @@
 #include <cmath>
 #include <cstddef>
 #include <cstdint>
+#include <cstdio>
 #include <cstring>
 #include <functional>
 #include <iomanip>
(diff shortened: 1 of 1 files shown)


mmud:
> Why is your solution better than the existing one?

the core issue is that `strcat` doesn't know how big hostname_ is. If the contents ever approach 32 bytes, it writes past the end and that will cause buffer overflow vulnerability. `snprintf` does the same thing but stops at the buffer limit.

another side benefit, this also removes the unnecessary heap allocations. The old code makes temporary `std::string ("(" + std::to_string(rank) + ")")` just to pass its `.c_str()`.

hkaiser:
@mmud Could you please have a look at the reported clang-format issues?

mmud:
@hkaiser I think it should be fine now.

PR #7053, opened 2026-03-18
Fix segmented subrange handling and scan bugs

### LINKED ISSUE #7052: Segmented algorithms: Incorrect single-segment subrange handling + Few other bugs in scan.hpp
### Description

While reviewing `exclusive_scan.hpp` and `detail/scan.hpp` (needed for Distributed Sort work), I found multiple correctness issues in segmented algorithms.

Main pattern: when `first` and `last` are in the same segment but represent only a subrange, some paths use wrong iterator helpers (`begin`/`end` instead of `local`). This can read/write outside the requested subrange. This pattern appears in multiple segmented algorithm implementations.

### Bugs List

1. Single-segment subrange handling is incorrect in some algorithms. 
    - causes processing beyond `[first, last)` or wrong output offset.
    - Present in `adjacent_find`, `adjacent_difference`, `replace_copy`, `transform`, `scan` paths

Following bugs are specifically in scan.hpp:

2. `scan.hpp`: Carry propagation order bug
   - carry update order is reversed in segmented scan.
   - `seq` and `par` can diverge for non-commutative operations.

3. `scan.hpp`: async lifetime bug in `par` and `par_non` path
   - Potential dangling reference `op/conv` / Dangling reference `op/f1/f2` 

4. `scan.hpp`: dereference before empty check in helper
   - `*first` can be read before confirming non-empty range.
   - Relevant fix:
```diff
-            T ret = HPX_INVOKE(conv, *first);
+            T ret = T();
              if (first != last)
              {
+                  ret = HPX_INVOKE(conv, *first);
                  for (++first; first != last; ++first)
                  {
                      ret = hpx::invoke(op, ret, hpx::invoke(conv, *first));
                  }
              }
``` 
5. `scan.hpp`: segment info and output position were mixed up
   - Old code could start writing at segment begin even when destination starts in middle.
   - Fix is to track segment identity and local output iterator separately (`out_segs` vs `out_iters`).
6. `scan.hpp`: overly strict distance check in `is_segmented_the_same`.
   - Old code used `if (in_dist != out_dist)` in the same-segment/subrange check. This is too strict. Output can validly have more room than input in that segment.
   - Fix: require only enough output space, i.e. `if (out_dist < in_dist)`.
 7. `adjacent_difference.hpp`: destination iterator is not updated in same-segment path

abhishek593:
Fixes #7052

## Proposed Changes

  - Fix single-segment subrange handling in segmented algorithms.
  - Fix segmented scan correctness issues in `detail/scan.hpp`.
  - Add regression tests for adjacent_find, adjacent_difference, transform, inclusive_scan, and exclusive_scan.

These issues were found while reviewing segmented scan paths needed for distributed sort work.

### DIFF
--- a/libs/full/segmented_algorithms/tests/unit/partitioned_vector_exclusive_scan.cpp
+++ b/libs/full/segmented_algorithms/tests/unit/partitioned_vector_exclusive_scan.cpp
@@ -47,6 +47,15 @@ struct opt
     }
 };
 
+template <typename T>
+struct sub_op
+{
+    T operator()(T v1, T v2) const
+    {
+        return v1 - v2;
+    }
+};
+
 struct concat_op
 {
     std::string operator()(std::string const& lhs, std::string const& rhs) const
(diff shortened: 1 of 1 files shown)


hkaiser:
I need a bit more time for this PR. In the meantime, could you please construct regression test cases that demonstrate the issues (if not fixed)?

abhishek593:
@hkaiser The regression tests are already present in the PR except for the 6th issue, which I have now added. 

1. For issuew 1, 5 and 7 - Each file affected has the similar kind of same-segment subrange tests.

2. For issue 4, currently its not directly affected, but the issue is its accessing *first, and then checking if its an empty range. This is fine for now, because all call sites for this function makes sure that it doesn't have an empty range, but its a good practice that if we are anyways checking for an empty range, we might as well do any accesses after that.

3. For issue 2, the test is under the following comment "// minimal regression for carry propagation order mismatch across segments" in [partitioned_vector_exclusive_scan.cpp‎](https://github.com/STEllAR-GROUP/hpx/pull/7053/changes#diff-8365ed91ac69a2c49c7c63ed7102a52b2992f26e7a8e7944fdc55a4002e6978e).

4. For issue 3, its not easily reproducible as its not a bug in the strict sense. It can possibly occur.

5. For issue 6, I have added the test in the latest commit under the comment "// regression: same-segment subrange must not fall back to non-segmented scan path when output has more remaining room than input range" in [partitioned_vector_exclusive_scan.cpp‎](https://github.com/STEllAR-GROUP/hpx/pull/7053/changes#diff-8365ed91ac69a2c49c7c63ed7102a52b2992f26e7a8e7944fdc55a4002e6978e).

hkaiser:
@abhishek593 Thanks you for your through analysis and your work!

PR #7054, opened 2026-03-18
[Algorithms] Fix C++20 sentinel compatibility in find, mismatch, and search

aneek22112007-tech:
## Proposed Changes

While I was looking through the parallel algorithms, I noticed a few spots where we weren't quite ready for C++20 sentinels. I've updated the following algorithms to use `advance_to_sentinel` instead of raw assignments:
- [find_first_of](cci:1://file:///Users/aneek/hpx/libs/core/algorithms/include/hpx/parallel/algorithms/find.hpp:674:4-732:63): Fixed the fallback lambda to use the proper sentinel alignment.
- [mismatch](cci:1://file:///Users/aneek/hpx/libs/core/algorithms/include/hpx/parallel/algorithms/mismatch.hpp:707:12-710:13): Updated the binary overload to safely handle both range ends.
- [search](cci:1://file:///Users/aneek/hpx/libs/core/algorithms/include/hpx/parallel/algorithms/search.hpp:95:4-167:45) and [search_n](cci:1://file:///Users/aneek/hpx/libs/core/algorithms/include/hpx/parallel/algorithms/detail/search.hpp:263:8-266:9): Replaced `std::advance` and assignments with the correct HPX utility to reach the end of the range.

I was studying how HPX handles ranges and realized that some algorithms were still using "old-school" iterator logic (`first = last`) or simple `std::advance`. Since C++20 lets us use sentinels that aren't the same type as the iterator, those lines would probably cause some headaches (and compiler errors!) when people try to use them with new range features.

I noticed `hpx::find` already does this correctly, so I just applied that same pattern to these other algorithms to make everything more consistent and robust.

## Checklist

- [x] My changes follow the HPX coding style.

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/find.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/find.hpp
@@ -1074,7 +1074,7 @@ namespace hpx::parallel {
                         val, HPX_FORWARD(Proj, proj));
                 };
 
-                auto f2 = [tok, count, first, last](
+                auto f2 = [tok, first](
                               auto&&... data) mutable -> Iter {
                     static_assert(sizeof...(data) < 2);
 
@@ -1158,7 +1158,7 @@ namespace hpx::parallel {
                         tok, HPX_FORWARD(F, f), HPX_FORWARD(Proj, proj));
                 };
 
-                auto f2 = [tok, count, first, last](
+                auto f2 = [tok, first](
                               auto&&... data) mutable -> Iter {
                     static_assert(sizeof...(data) < 2);
 
@@ -1241,7 +1241,7 @@ namespace hpx::parallel {
                         tok, HPX_FORWARD(F, f), HPX_FORWARD(Proj, proj));
                 };
 
-                auto f2 = [tok, count, first, last](
+                auto f2 = [tok, first](
                               auto&&... data) mutable -> Iter {
                     static_assert(sizeof...(data) < 2);
 
@@ -1351,7 +1351,7 @@ namespace hpx::parallel {
                         HPX_FORWARD(Proj1, proj1), HPX_FORWARD(Proj2, proj2));
                 };
 
-                auto f2 = [tok, count, first1, last1](
+                auto f2 = [tok, first1](
                               auto&&... data) mutable -> Iter1 {
                     static_assert(sizeof...(data) < 2);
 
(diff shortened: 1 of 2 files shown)


libs/core/algorithms/include/hpx/parallel/algorithms/mismatch.hpp
```diff
@@ -645,8 +645,17 @@ namespace hpx::parallel {
...
+                        std::advance(first1, mismatched);
+                    else
+                        first1 = hpx::parallel::detail::advance_to_sentinel(
+                            first1, last1);
+
+                    if (mismatched != count2)
+                        std::advance(first2, mismatched);
+                    else
+                        first2 = hpx::parallel::detail::advance_to_sentinel(
+                            first2, last2);
```
hkaiser: I think this doesn't do anything different from the pre-existing code. If `mismatch1 == count1`, then `first` will be andvanced to `last` in any case.

aneek22112007-tech: Hey @hkaiser, thanks for pointing that out! I see now that std::advance(first, count) would reach the end anyway if no mismatch is found, so the if-else block was definitely redundant.
I’ve gone ahead and simplified the logic in mismatch.hpp, find.hpp, search.hpp, and adjacent_find.hpp to remove those extra checks. I also fixed the DCO sign-off that I missed in my last push.
Let me know if there's anything else I should clean up!

libs/core/algorithms/include/hpx/parallel/algorithms/detail/search.hpp
```diff
@@ -354,13 +352,16 @@ namespace hpx::parallel::detail {
                     tok, HPX_FORWARD(Pred, pred), HPX_FORWARD(Proj, proj));
             };
 
-            auto f2 = [first, last, max_start, tok](
+            auto f2 = [first, last, n, max_start, tok](
```
hkaiser: Why do you need to capture `n`?

aneek22112007-tech: Hi! I've just updated the PR. I've removed the redundant capture of n in search.hpp as you pointed out, and I also merged the latest upstream/master to clear up the conflicts. I also made sure the commit is signed-off this time. Sorry for the oversight! Let me know if everything looks good now. Thanks again for the feedback!

hkaiser:
@aneek22112007-tech There are more compilation problems caused by your changes. Could you please have alook?

PR #7056, opened 2026-03-18
Feat: Replaced the old lockfree queue with the moodycamel ConcurrentQ…

### LINKED ISSUE #7050: thread_queue: per-task heap allocation in staged queue could be further optimised
# Problem 
Normal-priority task (the default for hpx::async) goes through the staged queue in thread_queue.hpp, which allocates and frees a task_description object per task:

     // thread_queue.hpp:774
     task_description* td = task_description_alloc_.allocate(1);  
     // thread_queue.hpp:244
     task_description_alloc_.deallocate(task, 1);                 

which uses plain malloc and free.

shared_priority_queue_scheduler avoids this entirely by storing task_description by value in a `moodycamel::ConcurrentQueue` (_thread_queue_mc.hpp_). Current schedulers using _thread_queue.hpp_ are:
 
1. local_priority_queue_scheduler 
2. local_queue_scheduler
3. background_scheduler
4.  local_workrequesting_scheduler

# Solution
Replace `boost::lockfree::queue<task_description*>` in the staged queue of thread_queue.hpp with `moodycamel::ConcurrentQueue<task_description>`. This stores task_description by value which eliminates the allocate(1)/deallocate(1) pair entirely. thread_queue_mc.hpp already has the similar pattern.

# Benchmarks
Ran the _future_overhead.cpp_

Image

There is clear performance degradation when I used _local-priority-fifo_

toxicteddy00077:
# Description
Closes #7050 

The staged queue in _thread_queue.hpp_ stored task_description* as a pointer (virtue of `boost::lockfree::queue` requiring copyable types). Thus we are forced to have one malloc/free per normal-priority task the default path for hpx::async which can become expensive for larger task size.

In this PR, I replace the staged queue backend with the `moodycamel::ConcurrentQueue` (concurrentqueue_fifo), the pattern for which already exists in the _thread_queue_mc.hpp_

### DIFF
--- a/libs/core/schedulers/include/hpx/schedulers/background_scheduler.hpp
+++ b/libs/core/schedulers/include/hpx/schedulers/background_scheduler.hpp
@@ -37,7 +37,7 @@ namespace hpx::threads::policies {
     // The background_scheduler_scheduler runs only background work
     HPX_CXX_CORE_EXPORT template <typename Mutex = std::mutex,
         typename PendingQueuing = lockfree_fifo,
-        typename StagedQueuing = lockfree_fifo,
+        typename StagedQueuing = concurrentqueue_fifo,
         typename TerminatedQueuing =
             default_background_scheduler_terminated_queue>
     class background_scheduler final
--- a/libs/core/schedulers/include/hpx/schedulers/local_priority_queue_scheduler.hpp
+++ b/libs/core/schedulers/include/hpx/schedulers/local_priority_queue_scheduler.hpp
@@ -55,7 +55,7 @@ namespace hpx::threads::policies {
     /// whenever no other work is available.
     HPX_CXX_CORE_EXPORT template <typename Mutex = std::mutex,
         typename PendingQueuing = lockfree_fifo,
-        typename StagedQueuing = lockfree_fifo,
+        typename StagedQueuing = concurrentqueue_fifo,
         typename TerminatedQueuing =
             default_local_priority_queue_scheduler_terminated_queue>
     class local_priority_queue_scheduler : public scheduler_base
--- a/libs/core/schedulers/include/hpx/schedulers/local_queue_scheduler.hpp
+++ b/libs/core/schedulers/include/hpx/schedulers/local_queue_scheduler.hpp
@@ -49,7 +49,7 @@ namespace hpx::threads::policies {
     /// from.
     HPX_CXX_CORE_EXPORT template <typename Mutex = std::mutex,
         typename PendingQueuing = lockfree_fifo,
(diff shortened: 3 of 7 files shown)


hkaiser:
Are you able to compare the performance before and after your change? Is there a significant difference?

toxicteddy00077:
@hkaiser I would say there is no extreme performance improvement , with varying speedups for different API, but mostly averaging around ~1.45. The improvements are there as you can see on the sent test output. on image left I have the 'before' performance, and on the right 'after'

Screenshot From 2026-03-19 09-55-31

hkaiser:
> @hkaiser I would say there is no extreme performance improvement , with varying speedups for different API, but mostly averaging around ~1.45. The improvements are there as you can see on the sent test output. on image left I have the 'before' performance, and on the right 'after'

Frankly, that's more than I expected. Thanks!

Pansysk75:
We did have some issues in the past when using the MoodyCamel queue in the scheduler. It does not guarantee FIFO ordering for items inserted by different threads, which can create some deadlocks (at least, in some of our tests).

I'm not sure if it will be an issue here, but I thought it would be worth bringing up.

Relevant comment with more explanation:
https://github.com/STEllAR-GROUP/hpx/pull/6428#issuecomment-1950003248

hkaiser:
> We did have some issues in the past when using the MoodyCamel queue in the scheduler. It does not guarantee FIFO ordering for items inserted by different threads, which can create some deadlocks (at least, in some of our tests).
> 
> I'm not sure if it will be an issue here, but I thought it would be worth bringing up.
> 
> Relevant comment with more explanation: [#6428 (comment)](https://github.com/STEllAR-GROUP/hpx/pull/6428#issuecomment-1950003248)

That might not be a problem for the queue that holds the task descriptions, those don't depend on any strict ordering.

hkaiser:
> > @hkaiser I would say there is no extreme performance improvement , with varying speedups for different API, but mostly averaging around ~1.45. The improvements are there as you can see on the sent test output. on image left I have the 'before' performance, and on the right 'after'
> 
> Frankly, that's more than I expected. Thanks!

@toxicteddy00077 May I ask you to run the test more than once (100 times?) and collect the averages?

toxicteddy00077:
@hkaiser I have distilled the output over 100 runs(see image). I am also attatching the csv files which store output per run.
[improved.csv](https://github.com/user-attachments/files/26155543/improved.csv)
[basleline.csv](https://github.com/user-attachments/files/26155544/basleline.csv)

I have compared the averages as you can see below

image

hkaiser:
@toxicteddy00077 Thanks for looking into this! The results are encouraging and unexpected!

PR #7057, opened 2026-03-18
Add inspect checks to pre-commit and DCO commit-msg hook

abhishek593:
## Proposed Changes

  - Adds HPX `inspect` support to `tools/pre-commit`
  - Keeps existing `clang-format` and `cmake-format` checks, with a small portability fix for path printing on macOS
  - Adds a new `tools/commit-msg` hook to enforce DCO signoff (`Signed-off-by:`) 

My PRs have been failing inspect and DCO checks in CI, and so I added support to catch these issues before pushing.

- [x] I have added a new tooling/dev-experience feature.

### DIFF
--- a/tools/commit-msg
+++ b/tools/commit-msg
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# Copyright (c) 2026 Abhishek Bansal
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+# ----------------------------------------------------------------
+# commit-msg hook script to enforce DCO signoff.
+#
+# This hook validates that the commit message contains a Signed-off-by trailer.
+#
+# To install this hook, copy this file to your git hooks as follows
+# cp tools/commit-msg .git/hooks/commit-msg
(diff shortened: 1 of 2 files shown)


hkaiser (commented):
Will this work if the inspect tool was not built?

abhishek593:
> Will this work if the inspect tool was not built?

@hkaiser It will fail and give instructions on how to build it. One can always bypass by using --no-verify option. Do you think its too invasive? Maybe, if the tool is not built, we can just skip inspect checks.

hkaiser:
> > Will this work if the inspect tool was not built?
> 
> @hkaiser It will fail and give instructions on how to build it. One can always bypass by using --no-verify option. Do you think its too invasive? Maybe, if the tool is not built, we can just skip inspect checks.

I think that would be better. Could you adjust the script, please?

abhishek593:
@hkaiser I have udpated the script.

hkaiser:
> @hkaiser I have udpated the script.

After thinking about this some more I have changed my mind - I apologize. The thing is (and that's what I missed when commenting before) is that the pre-commit hook has to be installed by the user explicitly. In that case, we can assume that `inspect` exists. However a error message should be generated if it can't be found.

abhishek593:
@hkaiser I have reverted to the original commit.

hkaiser (approved):
LGTM, thanks!

PR #7058, opened 2026-03-18
Tracing unified interface

### LINKED ISSUE #7049: Introduce minimal tracing abstraction layer (hpx::tracing) for Tracy
Hi @hkaiser ,

Following the recent discussion on Discord about unifying the tracing backends, I spent some time going through the tracing integration across the codebase.

While doing that, I noticed that tracing calls are directly embedded in a few core paths  the scheduler, some executor code, and the merge algorithm  which couples them tightly to Tracy and spreads `#ifdef` blocks across multiple files.

As a first step in that direction, I put together a small abstraction layer under `hpx::tracing`. Right now it wraps the existing Tracy primitives (`hpx::tracy::region`, `hpx::tracy::mark_event`) and compiles to no-ops when Tracy is disabled. The idea is that this can later be extended to support other backends like APEX and ITTNotify from the same place, so the call sites don’t need to change.

I applied this to a few core paths (scheduler, executors, merge algorithm) to validate the approach and kept the changes minimal so behavior stays the same. Since `region` is now always instantiated, I added a small guard around the argument computation in the scheduler to avoid extra work in non-Tracy builds just wanted to check if that’s the right approach, or if it would be preferable to keep it unguarded for simplicity.

I have  a PR ready for this , just wanted to check if this direction looks reasonable before sending it for review.

v4xsh:
Fixes #7049 

## Proposed Changes

- Introduce a minimal tracing abstraction layer under `hpx::tracing`
  - wraps existing Tracy primitives (`hpx::tracy::region`, `hpx::tracy::mark_event`)
  - compiles to no-ops when Tracy is disabled
- Register new `core/tracing` module and wire it into dependent modules
- Replace direct `hpx::tracy::*` usage with `hpx::tracing::*` in a few core paths
  - scheduler (`scheduling_loop.hpp`)
  - executors (`index_queue_spawning.hpp`)
  - algorithms (`merge.hpp`)
- Guard argument computation in the scheduler inside #ifdef so that expensive calls are skipped in non-Tracy builds, while `hpx::tracing::region` itself is always instantiated

This is a small first step towards unifying the tracing integrations in HPX. Currently, tracing calls are embedded directly in multiple places (scheduler, executors, algorithms), which couples them to specific backends and spreads `#ifdef` usage across the codebase.

The idea here is to introduce a thin abstraction layer (`hpx::tracing`) so that call sites no longer depend directly on Tracy. This should make it easier to extend the same interface to other backends like APEX and ITTNotify in the future without modifying those call sites again.

The scope of this PR is intentionally kept minimal and limited to simple RAII-style tracing points (`region`, `mark_event`). More complex parts of the Tracy integration (e.g. fiber suspension and scheduler interactions) were not included here since they are tightly coupled with runtime behavior, and it was unclear if abstracting those in the same step could introduce subtle issues.

Happy to extend this further once the overall direction is validated.

## Checklist

### DIFF
--- a/libs/core/tracing/include/hpx/tracing/tracing.hpp
+++ b/libs/core/tracing/include/hpx/tracing/tracing.hpp
@@ -19,9 +19,6 @@ namespace hpx::tracing {
 
     HPX_CXX_CORE_EXPORT struct HPX_CORE_EXPORT region
     {
-        region(char const* name, std::size_t num_thread, std::size_t phase,
-            bool enabled) noexcept;
-
         explicit region(hpx::threads::thread_data* thrdptr,
             std::size_t num_thread) noexcept;
 
@@ -67,10 +64,6 @@ namespace hpx::tracing {
 
     HPX_CXX_CORE_EXPORT struct [[maybe_unused]] region
     {
-        constexpr region(char const*, std::size_t, std::size_t, bool) noexcept
-        {
-        }
-
         constexpr explicit region(
             hpx::threads::thread_data*, std::size_t) noexcept
         {
--- a/libs/core/tracing/src/tracing.cpp
+++ b/libs/core/tracing/src/tracing.cpp
@@ -24,12 +24,6 @@ namespace hpx::tracing {
             name, num_thread, thrdptr->get_thread_phase(), enabled);
     }
 
-    region::region(char const* name, std::size_t num_thread, std::size_t phase,
-        bool enabled) noexcept
-      : impl(name, num_thread, phase, enabled)
-    {
-    }
-
     region::region(
         hpx::threads::thread_data* thrdptr, std::size_t num_thread) noexcept
       : impl(create_tracy_region(thrdptr, num_thread))


libs/core/executors/include/hpx/executors/detail/index_queue_spawning.hpp
```diff
@@ -27,9 +27,7 @@
 #include <hpx/modules/threading_base.hpp>
 #include <hpx/modules/topology.hpp>
 #include <hpx/modules/type_support.hpp>
-#if defined(HPX_HAVE_MODULE_TRACY)
-#include <hpx/modules/tracy.hpp>
-#endif
+#include <hpx/tracing/tracing.hpp>
```
hkaiser: Please always #include the generated module header (except from inside the module itself), otherwise this will break the C++ module interface infrastructure.
```suggestion
#include <hpx/modules/tracing.hpp>
```

libs/core/tracing/include/hpx/tracing/tracing.hpp
```diff
@@ -0,0 +1,53 @@
...
+            bool enabled) noexcept
+#if defined(HPX_HAVE_MODULE_TRACY)
+          : impl(name, thread_num, phase, enabled)
+#endif
+        {
+            (void) name;
+            (void) thread_num;
+            (void) phase;
+            (void) enabled;
+        }
```
hkaiser: Please use `[[maybe_unused]]` instead of the `(void)`:
```suggestion
        region([[maybe_unused]] char const* name, 
            [[maybe_unused]] std::size_t thread_num, 
            [[maybe_unused]] std::size_t phase,
            [[maybe_unused]] bool enabled) noexcept
#if defined(HPX_HAVE_MODULE_TRACY)
          : impl(name, thread_num, phase, enabled)
#endif
        {
        }
```

libs/core/thread_pools/include/hpx/thread_pools/scheduling_loop.hpp
```diff
@@ -255,8 +259,12 @@ namespace hpx::threads::detail {
...
                                 //   ~rctx last  -> ZoneEnd
                                 // Zone outlives the fiber context - correct.
-                                tracy::region rctx(name, num_thread,
-                                    thrdptr->get_thread_phase(), enable_tracy);
+                                hpx::tracing::region rctx(
+                                    name, num_thread, phase, enable_tracy);
+#if defined(HPX_HAVE_MODULE_TRACY)
+                                char const* fiber_name = enable_tracy ?
+                                    thrdptr->get_tracy_fiber_name() :
+                                    nullptr;
```
hkaiser: If we made the new `tracing` module depend on `threading_base`, you could simply pass the `thrdptr` to the constructor and extract the necessary information there. That may unclutter this callsite significantly.

libs/core/tracing/include/hpx/tracing/tracing.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include <cstddef>
+
+#if defined(HPX_HAVE_MODULE_TRACY)
+#include <hpx/tracy/tracy_tls.hpp>
+#endif
+
+namespace hpx::tracing {
+
+    struct region
```
hkaiser: Are you sure that we don't need to expose the new types through the C++ module interface:
```suggestion
    HPX_CXX_CORE_EXPORT struct region
```
?

libs/core/tracing/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+set(tracing_module_dependencies hpx_config)
+if(HPX_TRACY_WITH_TRACY)
+  list(APPEND tracing_module_dependencies hpx_tracy)
+endif()
+
+include(HPX_AddModule)
+add_hpx_module(
+  core tracing
+  GLOBAL_HEADER_GEN ON
+  GLOBAL_HEADER_MODULE_GEN OFF
```
hkaiser: To integrate this module with the C++ module infrastructure, this should be:
```suggestion
  GLOBAL_HEADER_MODULE_GEN ON
```

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -301,9 +301,7 @@ namespace hpx {
 #include <hpx/parallel/util/detail/sender_util.hpp>
 #include <hpx/parallel/util/foreach_partitioner.hpp>
 #include <hpx/parallel/util/result_types.hpp>
-#if defined(HPX_HAVE_MODULE_TRACY)
-#include <hpx/modules/tracy.hpp>
-#endif
+#include <hpx/tracing/tracing.hpp>
```
hkaiser:
```suggestion
#include <hpx/modules/tracing.hpp>
```

v4xsh:
@hkaiser  Thanks! I’ve addressed the suggestions and pushed the updates.

v4xsh:
There seems to be an issue with the rebase I'm looking into it and will
 push a clean version shortly.

v4xsh:
@hkaiser  I’ve updated the PR and cleaned up the history.
There were some conflicts earlier due to recent changes around `index_queue_spawning`, but it’s now rebased on the latest upstream and everything has been reapplied. Kindly have a look!

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -302,7 +302,7 @@ namespace hpx {
 #include <hpx/parallel/util/foreach_partitioner.hpp>
 #include <hpx/parallel/util/result_types.hpp>
 #if defined(HPX_HAVE_MODULE_TRACY)
-#include <hpx/modules/tracy.hpp>
+#include <hpx/modules/tracing.hpp>
```
hkaiser: Do we still need to guard the use of your module with `HPX_HAVE_MODULE_TRACY`? I think we can hide all of this inside the `tracing` headers.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -719,7 +719,7 @@ namespace hpx::parallel {
                 hpx::util::itt::mark_event e(notify_event);
 #endif
 #if defined(HPX_HAVE_MODULE_TRACY)
-                hpx::tracy::mark_event evt("get diagonal index");
+                hpx::tracing::mark_event evt("get diagonal index");
```
hkaiser, line 719: Same here, the `tracing` headers could do the right thing without you having to guard each and every use.

libs/core/thread_pools/include/hpx/thread_pools/scheduling_loop.hpp
```diff
@@ -231,11 +232,22 @@ namespace hpx::threads::detail {
...
-                                bool const enable_tracy =
+                                name = thrdptr->get_description()
+                                           .get_description();
+                                enable_tracy =
                                     name != nullptr && !thrdptr->is_stackless();
+                                phase = thrdptr->get_thread_phase();
+#endif
+
+                                hpx::tracing::region rctx(
+                                    name, num_thread, phase, enable_tracy);
```
hkaiser: If you change this to:
```suggestion
                                hpx::tracing::region rctx(thrdptr);
```
then all of the implementations can do the right thing internally.

libs/core/thread_pools/include/hpx/thread_pools/scheduling_loop.hpp
```diff
@@ -255,8 +267,6 @@ namespace hpx::threads::detail {
                                 //   ~fctx first -> TracyFiberLeave
                                 //   ~rctx last  -> ZoneEnd
                                 // Zone outlives the fiber context - correct.
-                                tracy::region rctx(name, num_thread,
-                                    thrdptr->get_thread_phase(), enable_tracy);
                                 tracy::fiber_region fctx(
                                     fiber_name, name, num_thread, enable_tracy);
```
hkaiser: Same here:
```suggestion
                                tracy::fiber_region fctx(thrdptr);
```

v4xsh:
Hey @hkaiser, thanks for the review! Made the following changes:

**`tracing.hpp`** — Added `thrdptr`-based constructors for both `region` and `fiber_region` that extract the description, phase, and enabled state directly from `thread_data`. Also added the `fiber_region` wrapper so the scheduling loop no longer needs to reach into `hpx::tracy::` directly.

**`scheduling_loop.hpp`** — Switched to `hpx::tracing::region rctx(thrdptr)` and `hpx::tracing::fiber_region fctx(thrdptr)` as suggested. This removed all the manual argument extraction (`name`, `phase`, `enable_tracy`, `fiber_name`) from the scheduling loop the tracing layer now handles everything internally. Also removed `#include <hpx/modules/tracy.hpp>` since nothing in the file uses `hpx::tracy::` directly anymore.

`fiber_suspend_region` is not included here  it's more tightly coupled with runtime suspension behavior and I plan to address it along with the remaining direct `hpx::tracy::` call sites in upcoming PRs.

One thing I wanted to check the `thrdptr`-based constructor for `fiber_region` lives in `hpx::tracing` rather than `hpx::tracy` because `hpx_threading_base` already depends on `hpx_tracy`, so adding a `thread_data` include in `tracy.hpp` would create a circular dependency. Putting it in `hpx::tracing` (which sits above both) avoids that. Is that the right approach?

hkaiser:
> One thing I wanted to check the `thrdptr`-based constructor for `fiber_region` lives in `hpx::tracing` rather than `hpx::tracy` because `hpx_threading_base` already depends on `hpx_tracy`, so adding a `thread_data` include in `tracy.hpp` would create a circular dependency. Putting it in `hpx::tracing` (which sits above both) avoids that. Is that the right approach?

Yes, that's how you can circumvent the problem. Thanks!

v4xsh:
Great, thanks for confirming!

libs/core/tracing/include/hpx/tracing/tracing.hpp
```diff
@@ -0,0 +1,88 @@
...
+          : impl(thrdptr != nullptr ?
+                    thrdptr->get_description().get_description() :
+                    nullptr,
+                0, thrdptr != nullptr ? thrdptr->get_thread_phase() : 0,
+                thrdptr != nullptr &&
+                    thrdptr->get_description().get_description() != nullptr &&
+                    !thrdptr->is_stackless())
+#endif
+        {
+        }
```
hkaiser: This is very complicated code. may I suggest you write a static member function that returns amn instance of a `hpx::tracy::region`? That would allow to make the initialization ode more readable.
```suggestion
        explicit region(
            [[maybe_unused]] hpx::threads::thread_data* thrdptr,
            [[maybe_unused]] std::size_t num_thread) noexcept
#if defined(HPX_HAVE_MODULE_TRACY)
          : impl(create_tracy_region(thrdptr, num_thread)
#endif
        {
        }
```
and
```
#if defined(HPX_HAVE_MODULE_TRACY)
    static hpx::tracy::region create_tracy_region(hpx::threads::thread_data* thrdptr,
        std::size_t num_thread) 
    {
        char const* name = thrdptr->get_description().get_description();
        bool const enable_tracy = name != nullptr && !thrdptr->is_stackless();
        return {name, num_thread, thrdptr->get_thread_phase(), enable_tracy};
    }
#endif
```
Note, that I added the `num_threads` argument as this was lost in transition.

libs/core/tracing/include/hpx/tracing/tracing.hpp
```diff
@@ -0,0 +1,88 @@
...
+    };
```
hkaiser, line 71: It might be useful to have at least two versions of the `tracing` interface types: one for the case when no tracing is needed, and one otherwise. The first could be 100% `constexpr`/`noexcept`, the second could have its implementations moved to a cpp file to reduce compile-time pressure.

libs/core/tracing/include/hpx/tracing/tracing.hpp
```diff
@@ -0,0 +1,88 @@
...
+#if defined(HPX_HAVE_MODULE_TRACY)
+          : impl(thrdptr != nullptr ? thrdptr->get_tracy_fiber_name() : nullptr,
+                thrdptr != nullptr ?
+                    thrdptr->get_description().get_description() :
+                    nullptr,
+                0,
+                thrdptr != nullptr &&
+                    thrdptr->get_description().get_description() != nullptr &&
+                    !thrdptr->is_stackless())
+#endif
```
hkaiser: I'd suggest applying similar refactorings and readability improvements as outlined above.

libs/core/tracing/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+set(tracing_headers hpx/tracing/tracing.hpp)
+set(tracing_sources)
+
+set(tracing_module_dependencies hpx_config hpx_threading_base)
+if(HPX_TRACY_WITH_TRACY)
+  list(APPEND tracing_module_dependencies hpx_tracy)
```
hkaiser, line 13:
```suggestion
  list(APPEND tracing_module_dependencies hpx_threading_base hpx_tracy)
```

libs/core/tracing/include/hpx/tracing/tracing.hpp
```diff
@@ -0,0 +1,88 @@
...
+#pragma once
+
+#include <hpx/config.hpp>
+
+#include <cstddef>
+
+#if defined(HPX_HAVE_MODULE_TRACY)
+#include <hpx/modules/threading_base.hpp>
+#include <hpx/tracy/tracy.hpp>
+#include <hpx/tracy/tracy_tls.hpp>
```
hkaiser: Please always use the generated module headers (except inside a module itself):
```suggestion
#include <hpx/modules/tracy.hpp>
```

v4xsh:
Thanks for the feedback! Added `num_thread` back to both constructors and refactored the initialization into static factory methods `create_tracy_region` and `create_tracy_fiber_region` as you suggested  much more readable now.

v4xsh:
@hkaiser  Regarding splitting the tracing interface types into a `constexpr`/`noexcept` no-op version and a Tracy-enabled version with implementations in a `.cpp` file would you prefer I address that in this PR or a follow-up?

libs/core/tracing/CMakeLists.txt
```diff
@@ -7,9 +7,9 @@
 set(tracing_headers hpx/tracing/tracing.hpp)
 set(tracing_sources)
 
-set(tracing_module_dependencies hpx_config hpx_threading_base)
+set(tracing_module_dependencies hpx_config)
 if(HPX_TRACY_WITH_TRACY)
-  list(APPEND tracing_module_dependencies hpx_tracy)
+  list(APPEND tracing_module_dependencies hpx_threading_base hpx_tracy)
```
hkaiser: Ahh, I now understand what you were trying to do. I apologize, I think your solution was correct.

libs/core/tracing/include/hpx/tracing/tracing.hpp
```diff
@@ -12,8 +13,11 @@
 
 #if defined(HPX_HAVE_MODULE_TRACY)
 #include <hpx/modules/threading_base.hpp>
-#include <hpx/tracy/tracy.hpp>
-#include <hpx/tracy/tracy_tls.hpp>
+#include <hpx/modules/tracy.hpp>
+#else
+namespace hpx::threads {
+    class thread_data;
```
hkaiser: Yah, this is not nice. Let's `#include <hpx/modules/threading_base.hpp>` it unconditionally. Again, apologies.

hkaiser:
> @hkaiser Regarding splitting the tracing interface types into a `constexpr`/`noexcept` no-op version and a Tracy-enabled version with implementations in a `.cpp` file would you prefer I address that in this PR or a follow-up?

Let's get it right here. There will be follow-up PRs when we add more backends. It will become easier if we have a nice layout in place to begin with.

libs/core/tracing/include/hpx/tracing/tracing.hpp
v4xsh: > Yah, this is not nice. Let's `#include <hpx/modules/threading_base.hpp>` it unconditionally. Again, apologies.

No worries, updated it!

v4xsh:
> > @hkaiser Regarding splitting the tracing interface types into a `constexpr`/`noexcept` no-op version and a Tracy-enabled version with implementations in a `.cpp` file would you prefer I address that in this PR or a follow-up?
> 
> Let's get it right here. There will be follow-up PRs when we add more backends. It will become easier if we have a nice layout in place to begin with.

Sure, I will add it shortly!

v4xsh:
@hkaiser I have split the tracing types as suggested no-op path is 100% `constexpr`/`noexcept`, Tracy path has declarations in the header with implementations moved to `tracing.cpp`. `#include <hpx/modules/tracy.hpp>` stays in the header since the Tracy types are stored by value as private members.

 One note on `fiber_region` — `num_thread` is accepted in the constructor for consistency with `region`, but not forwarded to `hpx::tracy::fiber_region` since its third parameter is `color`, not `thread_num`. Passing `0` there instead to avoid silently using the thread number as a color value.

v4xsh:
Fixing those failing checks.

hkaiser:
> One note on `fiber_region` — `num_thread` is accepted in the constructor for consistency with `region`, but not forwarded to `hpx::tracy::fiber_region` since its third parameter is `color`, not `thread_num`. Passing `0` there instead to avoid silently using the thread number as a color value.

Should we possibly generate a color from the thread_num? Would this be too 'colorful' in the end?

libs/core/algorithms/CMakeLists.txt
hkaiser: Do we still need this?

libs/core/executors/CMakeLists.txt
hkaiser: I think we should be able to remove this now.

libs/core/thread_pools/CMakeLists.txt
hkaiser: Same here, we might not need this anymore.

libs/core/tracing/include/hpx/tracing/tracing.hpp
```diff
@@ -0,0 +1,103 @@
...
+namespace hpx::tracing {
+
+    struct region
+    {
+        constexpr region([[maybe_unused]] char const* name,
+            [[maybe_unused]] std::size_t num_thread,
+            [[maybe_unused]] std::size_t phase,
+            [[maybe_unused]] bool enabled) noexcept
+        {
+        }
```
hkaiser:
```suggestion
        constexpr region(char const*, std::size_t, std::size_t, bool) noexcept
        {
        }
```

libs/core/tracing/include/hpx/tracing/tracing.hpp
```diff
@@ -0,0 +1,103 @@
...
+#include <hpx/modules/threading_base.hpp>
+
+#include <cstddef>
+
+#if defined(HPX_HAVE_MODULE_TRACY)
+#include <hpx/modules/tracy.hpp>
+
+namespace hpx::tracing {
+
+    struct HPX_CORE_EXPORT region
```
hkaiser: Let's support C++ modules
```suggestion
    HPX_CXX_CORE_EXPORT struct HPX_CORE_EXPORT region
```

libs/core/tracing/include/hpx/tracing/tracing.hpp
```diff
@@ -0,0 +1,103 @@
...
+        hpx::tracy::fiber_region impl;
+    };
+
+}    // namespace hpx::tracing
+
+#else
+
+namespace hpx::tracing {
+
+    struct region
```
hkaiser:
```suggestion
     HPX_CXX_CORE_EXPORT struct region
```

v4xsh:
@hkaiser  Done!  Also used num_thread as a color seed for fiber_region so each worker thread gets a distinct color on the fiber track in Tracy.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -718,9 +716,8 @@ namespace hpx::parallel {
                 static hpx::util::itt::event notify_event("get diagonal index");
                 hpx::util::itt::mark_event e(notify_event);
 #endif
-#if defined(HPX_HAVE_MODULE_TRACY)
-                hpx::tracy::mark_event evt("get diagonal index");
-#endif
+                [[maybe_unused]] hpx::tracing::mark_event evt(
```
hkaiser: Since you have added `[[maybe_unused]]` to every use of the type, maybe you can get away with saying:
```
    struct [[maybe_unused]] mark_event {...};`
```
instead. Please try that, I'm not sure myself.

libs/core/thread_pools/include/hpx/thread_pools/scheduling_loop.hpp
```diff
@@ -231,14 +229,8 @@ namespace hpx::threads::detail {
...
 #endif
-#if defined(HPX_HAVE_MODULE_TRACY)
-                                char const* name = thrdptr->get_description()
-                                                       .get_description();
-                                bool const enable_tracy =
-                                    name != nullptr && !thrdptr->is_stackless();
-                                char const* fiber_name = enable_tracy ?
-                                    thrdptr->get_tracy_fiber_name() :
-                                    nullptr;
+                                hpx::tracing::region rctx(thrdptr, num_thread);
```
hkaiser, line 232: The other types (besides `mark_event`) may benefit from the `[[maybe_unused]]` attribute applied to the type itself as well.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
v4xsh: > Since you have added `[[maybe_unused]]` to every use of the type, maybe you can get away with saying:
> 
> ```
>     struct [[maybe_unused]] mark_event {...};`
> ```
> 
> instead. Please try that, I'm not sure myself.

Okay I'l try that

libs/core/tracing/include/hpx/tracing/tracing.hpp
```diff
@@ -0,0 +1,96 @@
...
+}    // namespace hpx::tracing
+
+#else
+
+namespace hpx::tracing {
+
+    HPX_CXX_CORE_EXPORT struct [[maybe_unused]] region
+    {
+        constexpr region(
+            char const*, std::size_t = 0, std::size_t = 0, bool = true) noexcept
```
hkaiser: Why do you need this constructor definition? Where is it used?

v4xsh: Yeah, removed it no call site uses the defaults anyway.

libs/core/tracing/include/hpx/tracing/tracing.hpp
```diff
@@ -0,0 +1,95 @@
...
+#include <cstddef>
+
+#if defined(HPX_HAVE_MODULE_TRACY)
+#include <hpx/modules/tracy.hpp>
+
+namespace hpx::tracing {
+
+    HPX_CXX_CORE_EXPORT struct HPX_CORE_EXPORT region
+    {
+        region(char const* name, std::size_t num_thread, std::size_t phase,
```
hkaiser: Do we need this constructor? is it used anywhere?

libs/core/tracing/include/hpx/tracing/tracing.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+}    // namespace hpx::tracing
+
+#else
+
+namespace hpx::tracing {
+
+    HPX_CXX_CORE_EXPORT struct [[maybe_unused]] region
+    {
+        constexpr region(char const*, std::size_t, std::size_t, bool) noexcept
```
hkaiser: Same here, this is not needed at all.

v4xsh:
> Once the superfluous constructors are removed this can be merged. Thanks a lot for your work and your patience!

 Removed both superfluous constructors from the Tracy and no-op paths. Happy to help, and thanks for the thorough review!

PR #7068, opened 2026-03-20
cmake: remove HPX_WITH_CXX20_STD_EXECUTION_POLICES feature test

Priyanshi507:
Fixes #
6941
## Proposed Changes

- Remove cmake function hpx_check_for_cxx20_std_execution_policies 
  from HPX_AddConfigTest.cmake
- Remove function call from HPX_PerformCxxFeatureTests.cmake
- Delete obsolete test file cmake/tests/cxx20_std_execution_policies.cpp
- Remove HPX_HAVE_CXX20_STD_EXECUTION_POLICES guards from:
  - std_execution_policy.hpp
  - dispatch.hpp
  - std_execution_policies.cpp
  - foreach_std_policies.cpp
  
Part of the cleanup tracked in #6941. Since HPX now requires C++20,
std::execution::unsequenced_policy is always available and the 
compile-time check is no longer needed.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1893,6 +1893,17 @@ hpx_option(
   ADVANCED
 )
 
+hpx_option(
+  HPX_WITH_CXX20_STD_EXECUTION_POLICES
+  BOOL
+  "Support using std execution policies for HPX parallel algorithms (default: ON)."
+  ON
+  ADVANCED
+)
+if(HPX_WITH_CXX20_STD_EXECUTION_POLICES)
+  hpx_add_config_define(HPX_HAVE_CXX20_STD_EXECUTION_POLICES)
+endif()
+
 # Need to include the CUDA setup before the config test to enable the CUDA
 # language
 include(HPX_SetupCUDA)


hkaiser:
Removing the feature test is ok. However, I'd like to keep the CMake flag (and the associated preprocessor macro) in place as this option guards an optional buy-in to allow using std exection policies with HPX algorithms. It might be useful to be able to disable this option in some contexts.

Priyanshi507:
> Removing the feature test is ok. However, I'd like to keep the CMake flag (and the associated preprocessor macro) in place as this option guards an optional buy-in to allow using std exection policies with HPX algorithms. It might be useful to be able to disable this option in some contexts.

@hkaiser Done — I've restored the HPX_HAVE_CXX20_STD_EXECUTION_POLICES guards in std_execution_policy.hpp, dispatch.hpp, std_execution_policies.cpp, and foreach_std_policies.cpp. The CMake feature test removal is kept, but the preprocessor macro remains as an optional buy-in.

cmake/HPX_PerformCxxFeatureTests.cmake
```diff
@@ -104,10 +104,6 @@ function(hpx_perform_cxx_feature_tests)
       )
     endif()
 
-    hpx_check_for_cxx20_std_execution_policies(
-      DEFINITIONS HPX_HAVE_CXX20_STD_EXECUTION_POLICES
```
hkaiser, line 108: By removing this, the `HPX_HAVE_CXX20_STD_EXECUTION_POLICES` will never be defined. This needs to be replaced with a canonical:
```
hpx_option(
  HPX_WITH_CXX20_STD_EXECUTION_POLICES BOOL 
  "Support using std executio policies for HPX paralel algorithms (default: ON)." 
  ON ADVANCED
)
if(HPX_WITH_CXX20_STD_EXECUTION_POLICES)
  hpx_add_config_define(HPX_HAVE_CXX20_STD_EXECUTION_POLICES)
endif()
```
This can go somewhere in the main CMakeLists.txt

Priyanshi507: @hkaiser Done — I've added hpx_option for HPX_WITH_CXX20_STD_EXECUTION_POLICES in the main CMakeLists.txt after the HPX_WITH_SUPPORT_NO_UNIQUE_ADDRESS_ATTRIBUTE option, with hpx_add_config_define to ensure HPX_HAVE_CXX20_STD_EXECUTION_POLICES is still defined when the option is ON.

PR #7073, opened 2026-03-20
fix: correct overlap detection in parallel uninitialized_relocate CPOs

Sanchit2662:
## Summary
I traced the original condition: !((first < dest_last) && (dest_last < last)). When you expand it, dest_last < last simplifies to dest < first , so it was only   ever checking one overlap direction. If dest sits anywhere to the right of first and still inside the source range, that predicate goes false, and the parallel path kicks in.                                                                                                                                        

The problem is the parallel partitioner splits the work into chunks with no ordering between them. Two chunks end up doing memmove calls where one chunk's write region lands inside another chunk's read region. That's an unsynchronized concurrent read-write  a data race, silent wrong values, no crash.

## Fix
The fix is straightforward: only go parallel when the ranges don't touch at all. dest >= last means destination starts at or after source end. dest_last <= first means destination ends at or before source start. Anything else  left-overlap, right-overlap, identical ranges  falls through to sequential. Sequential   for trivially relocatable types uses a single memmove over the full range, which handles any overlap direction correctly by design. The two-line change is enough, no restructuring needed.   
```
  // Before — only caught left-overlap (dest < first)
  if (!((first < dest_last) && (dest_last < last)))

  // After — catches any overlap, parallel only when fully disjoint
  if (dest >= last || dest_last <= first)
```


hkaiser:

isidorostsa:
The issue here is that if they are overlapping in the wrong direction then this is wrong in the sequencial version as well.

The way we've decided to address this is to assert they are not overlapping in the wrong direction, and decide between parallel and sequencial based on if they overlap in the correct direction. 

This is already being addressed in #6879

Sanchit2662:
Ah got it, my bad  I should’ve checked the existing PRs before jumping in.
closing this one for now

PR #7072, opened 2026-03-20
Add segmented is_sorted_until , is_sorted

Momensamir12:
#1338 

## Proposed Changes

  - Add segmented version of `is_sorted_until` & `is_sorted`.
  - `is_sorted` delegates to `is_sorted_until` and compares result against `last` iterator.
  
## Checklist

Not all points below apply to all pull requests.

- [x] I have added a new feature and have added tests to go along with it.
- [ ] I have fixed a bug and have added a regression test.
- [ ] I have added a test using random numbers; I have made sure it uses a seed, and that random numbers generated are valid inputs for the tests.

PR #7074, opened 2026-03-20
Feat/compile time wrap check

shivansh023023:
Fixes #

## Proposed Changes

  - Add `INTERFACE` compile definition `HPX_HAVE_WRAP_MAIN_CONFIGURED` to both `hpx_wrap` and `hpx_auto_wrap` targets in `wrap/CMakeLists.txt` (inside the existing `HPX_WITH_DYNAMIC_HPX_MAIN` + Linux guard). This injects the define automatically into any downstream CMake consumer, ensuring **zero noise for CMake users**.
  - Add a `#warning` in `wrap/include/hpx/hpx_main.hpp` that fires when **all** of these are true: `__linux__`, `HPX_HAVE_DYNAMIC_HPX_MAIN`, `HPX_HAVE_STATIC_LINKING`, and `HPX_HAVE_WRAP_MAIN_CONFIGURED` is **not** defined. This catches the missing `-Wl,-wrap=main` linker flag at compile time instead of letting it become a runtime crash.
  - The warning message is actionable: it tells the user exactly which flag to add and how to suppress the warning if they've already handled it.

This is the **compile-time component** of a safety net for the UX gap where `-Wl,-wrap=main` does not propagate to non-CMake consumers (Godbolt / Compiler Explorer, Makefiles, raw `g++` invocations). It complements:

- **PR #7068** — `godbolt-minimal` CMake preset (build-system level)
- **PR #7072** — `hpx/local.hpp` convenience header (usability level)

**Design trade-off**: Non-CMake users who *have* already added `-Wl,-wrap=main` manually will still see the `#warning` (since they lack the define). They can silence it with `-DHPX_HAVE_WRAP_MAIN_CONFIGURED`. A suppressible compile-time warning is far better than a silent compilation followed by a cryptic runtime crash.

This work is part of the GSoC 2026 effort to integrate HPX into Compiler Explorer.

## Checklist

  - *Verified with 6 preprocessor logic tests covering all define combinations (Linux/macOS, static/shared, dynamic-main on/off, suppress define present/absent). Warning fires in exactly one case: Linux + static + dynamic-main + no suppress define.*
- [ ] ~~I have fixed a bug and have added a regression test.~~ *(N/A — this is a new diagnostic, not a bugfix)*
- [ ] ~~I have added a test using random numbers~~ *(N/A)*


hkaiser:
Have you seen the existing `HPX_HAVE_DYNAMIC_HPX_MAIN`? Would that be sufficient for what you're trying to achieve?

shivansh023023:
> Have you seen the existing `HPX_HAVE_DYNAMIC_HPX_MAIN`? Would that be sufficient for what you're trying to achieve?

That's a good point, @hkaiser . I did consider HPX_HAVE_DYNAMIC_HPX_MAIN, but that macro indicates that the feature is enabled in the HPX build configuration.

The reason for the new HPX_HAVE_WRAP_MAIN_CONFIGURED (injected via INTERFACE compile definitions) is to distinguish between:

CMake users: Who get the linker flag automatically and should not see the warning.

Non-CMake users: Who have the feature enabled in their HPX installation but haven't necessarily passed the -Wl,-wrap=main flag to their manual link command.

Without the new definition, hpx_main.hpp wouldn't be able to tell if the user linking manually has actually applied the flag or not. Does that make sense, or is there a way to leverage the existing macro to detect the linker-state that I might have missed?

hkaiser:
@shivansh023023 Could you please have a look at the compilation failure generated?

hkaiser:
The CIs still report that one of the translation units can't find  the header `<hpx/config/static_linker_check.hpp>`. Did you miss adding the file to this PR?

hkaiser:
@shivansh023023 ping?

hkaiser:
@shivansh023023 Are you still interested in moving tis forward? Otherwise I'll go ahead and close this without merging.

hkaiser:
@shivansh023023 Please have an eye on the CIs yourself. Many issues are still being reported.

hkaiser:
@shivansh023023 could you please elaborate on the difference between this PR and #7070?

shivansh023023:
> @shivansh023023 could you please elaborate on the difference between this PR and #7070?

While both PRs touch the execution and executors modules, they serve two entirely different purposes in the modernization of HPX:

PR #7070 is focused on infrastructure and linkage. Its primary goal is ensuring that core execution types (like parallel policies) are correctly exposed and linkable, especially in static-build configurations on Windows. As per your recent feedback, I've removed the arbitrary convenience header there and kept it focused on verifying standard header accessibility.

PR #7074 (and #7238) is focused on P2300 Standard Compliance. This PR implements the 'bridge' that allows existing HPX executors to function as P2300 schedulers. It specifically introduces the get_scheduler customization point and the executor_scheduler adapter.

In short: #7070 ensures the headers can be included and linked, while #7074 adds the actual functional bridge to make HPX executors compatible with the new Sender/Receiver world.

hkaiser:
> In short: #7070 ensures the headers can be included and linked, while #7074 adds the actual functional bridge to make HPX executors compatible with the new Sender/Receiver world.

Hmmm, from what I see, this PR (#7074) has no relation to senders&receivers at all. Could you elaborate, please?

shivansh023023:
> Hmmm, from what I see, this PR (#7074) has no relation to senders&receivers at all. Could you elaborate, please?

You’re absolutely right, this PR is unrelated to the Sender/Receiver work.

I opened this during my pre-GSoC phase while I was exploring the project to add HPX support to Godbolt (Compiler Explorer). This PR focuses on the necessary infrastructure and configuration tweaks required to get HPX compiling and running within the Godbolt environment.

I apologize for any confusion, with the recent volume of P2300 updates, the context here got a bit buried. I’m still very interested in seeing this through alongside the P2300 work, as it would significantly lower the barrier for people to try out HPX snippets

libs/core/include_local/include/hpx/local.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#pragma once
+
+#include <hpx/config.hpp>
+
+// --- Standard Parallel Toolkit (core, no networking dependency) ---
+#include <hpx/modules/algorithms.hpp>
+#include <hpx/modules/execution.hpp>
+#include <hpx/modules/futures.hpp>
+#include <hpx/numeric.hpp>
```
hkaiser: We discussed this header in #7070. I'm not sure what value it provides given the fact that it combines a very arbitrary set of HPX headers. I still think we should drop this header and not merge it to the main branch.

hkaiser (commented):
Let me ask this question again: what is the difference betwee this PR (#7074) and #7070? These seem to do very similar things.

shivansh023023:
@hkaiser 


You’re absolutely right. There is an overlap here because #7074 was branched off before we decided to drop the local.hpp header in #7070.

I completely agree that the header is arbitrary and doesn't belong in the main branch. My intention for #7074 was specifically for the Godbolt (Compiler Explorer) integration infrastructure, and I accidentally carried the local.hpp changes over into this PR as well.

I will clean up this PR (#7074) by:

Removing the hpx/local.hpp header entirely.

Rebasing on master so it reflects the changes we already made in #7070.

Keeping only the specific configuration/infrastructure tweaks needed for the Godbolt integration.

I'll update the branch to remove the confusion. Thanks for catching that!

shivansh023023:
@hkaiser I have just pushed a major cleanup for this PR to resolve the confusion and align it with our previous discussions:

Removed local.hpp: I have completely deleted the hpx/local.hpp header, its associated unit test, and all references in CMakeLists.txt. This branch is now fully rebased on master where that header no longer exists.

Focused Scope: This PR now strictly focuses on the Godbolt/Compiler Explorer infrastructure. It provides the necessary compile-time diagnostics in static_linker_check.hpp to help users (especially those not using CMake) understand why the linker wrap is failing in static builds.

Improved Diagnostics: I’ve refined the #warning message to be more actionable, explicitly mentioning both the HPX::wrap_main CMake target and the manual -Wl,--wrap=main flag as remedies.

Header Alignment: I also ensured that the S&R forward declarations (specifically in completion_signatures.hpp) follow the HPX_CXX_CORE_EXPORT pattern we established for C++20 Module visibility.

libs/core/config/include/hpx/config/static_linker_check.hpp
```diff
@@ -0,0 +1,42 @@
+//  Copyright (c) 2026 The STE||AR-Group
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// hpxinspect:linelength
```
hkaiser: Do we still need this?

shivansh023023: Hi @hkaiser 

Yes, we still need it. I kept the suppression because the #warning diagnostic message explaining how to fix the linker issue (using HPX::wrap_main or -Wl,--wrap=main) is longer than 80 characters.

I wanted the message to be as clear and actionable as possible for the user, which unfortunately makes it fail the inspect tool's line-length check. If you’d prefer, I can try to break the warning across multiple lines, but the suppression was the cleanest way to keep the message readable in the compiler output.

hkaiser: But you have split the line using `\` continuation characters. I don't think inspect will concatenate the split line to check whether its longer than allowed. This comment should not be necessary.

shivansh023023: That’s a fair point. I was under the impression that the tool might evaluate the logical line length, but if it only checks physical line length, then you're right,the line continuations already solve the problem.

I’ll remove the hpxinspect:linelength suppression and push the change now. If the CI 'inspect' tool still flags it for some reason, I'll know for sure, but it's better to keep the code clean of unnecessary suppressions. Thanks for the correction!

hkaiser (dismissed):
LGTM, thanks!

hkaiser:
@shivansh023023 I'm still not quite clear about the difference between this PR (#7074) and #7070. Could you please elaborate? Can we close #7070 without merging if we move forward with this PR?

shivansh023023:
> @shivansh023023 I'm still not quite clear about the difference between this PR (#7074) and #7070. Could you please elaborate? Can we close #7070 without merging if we move forward with this PR?

Hi @hkaiser 

I understand the confusion; both PRs involve the same area of the codebase, but they serve two distinct purposes:

PR #7070 (CI Regression Test): This was repurposed to fix a specific regression in our Windows/Static CI. It ensures that the core execution modules remain linkable in static configurations by adding a dedicated regression test.

PR #7074 (User Diagnostics/Godbolt): This PR is specifically for the Compiler Explorer (Godbolt) integration. It adds the compile-time #warning in static_linker_check.hpp to help users understand why their builds are failing when they don't use our CMake targets (which is the default behavior on Godbolt).

To answer your question: Yes, we can definitely close #7070 without merging, provided that I move the regression test logic from that PR into this one (#7074).

If you prefer, I can combine them so we have one single 'Static Linker Support' PR that includes both the diagnostic warnings for users and the CI regression tests to prevent future breakages. Would you like me to consolidate them here

hkaiser:
> If you prefer, I can combine them so we have one single 'Static Linker Support' PR that includes both the diagnostic warnings for users and the CI regression tests to prevent future breakages. Would you like me to consolidate them here

That might be the best way forward to ensure non-conflicting (and non-overlapping) functionalities. Thanks!

shivansh023023:
@hkaiser 
Done . I've brought the regression tests over from #7070 and renamed the test to static_linker_test.cpp

now this pr provides both the compile-time diagnostics for users and the CI regression testing for static linking.

libs/core/include_local/tests/unit/static_linker_test.cpp
```diff
@@ -0,0 +1,21 @@
...
+//
+// This is a compile-and-link sanity check only. It does NOT start the HPX
+// runtime, so it has no dependency on the wrap module (hpx_main.hpp) or on
+// any specific HPX link target beyond hpx_core.
+
+#include <hpx/execution.hpp>
+
+int main()
+{
+    hpx::execution::parallel_policy p;
```
hkaiser: You'll need to write:
```suggestion
    [[maybe_unused]] hpx::execution::parallel_policy p;
```
to make it compile cleanly.

hkaiser:
> @hkaiser Done . I've brought the regression tests over from #7070 and renamed the test to static_linker_test.cpp
> 
> now this pr provides both the compile-time diagnostics for users and the CI regression testing for static linking.

Thanks! I'll go ahead and close #7070, then.

libs/core/include_local/tests/unit/static_linker_test.cpp
shivansh023023: Implemented !

shivansh023023:
@hkaiser, all your feedback have been addressed
the two remaining macOS CI failures those look unrelated to this PR

post_colocated (Timeout) lives in async_colocated, a distributed networking test

futures.await (Failed) lives in the futures module

Neither of those modules is touched by this PR, so these seem like pre-existing flaky tests on the macOS runners.

hkaiser:
@shivansh023023 This PR somehow fell off the radars. Would you be willing to dust it up and see what's missing?

hkaiser:
@shivansh023023 Would you mind checking the bot's comments?

shivansh023023:
@hkaiser I checked the bot's comments and have pushed an update!

1. **macOS Diagnostic Fix:** CodeRabbit was right about the macOS check. I removed `__APPLE__` from the `#if` condition in `static_linker_check.hpp` so the diagnostic strictly targets Linux builds, as macOS uses a different linker entry mechanism (`-Wl,-e,_initialize_main`).
2. **Test Extension (Skipped):** I intentionally skipped the bot's suggestion to test the `#warning` in `static_linker_test.cpp`. Verifying the emission of a `#warning` in a standard `.cpp` file isn't possible with standard C++ preprocessor logic; it would require heavy CMake `try_compile` infrastructure with regex output parsing. Additionally, injecting `hpx_main.hpp` into that specific test would alter its `main()` symbol resolution, defeating the test's original purpose of verifying isolated standard parallel toolkit linkage. 

With the macOS condition fixed, this should be good to go!

shivansh023023:
@hkaiser let me know how this looks  to you

wrap/include/hpx/hpx_main.hpp
```diff
@@ -7,8 +7,20 @@
...
+#if defined(HPX_HAVE_DYNAMIC_HPX_MAIN) && defined(HPX_HAVE_STATIC_LINKING) &&  \
+    (defined(__linux) || defined(__linux__) || defined(linux)) &&              \
+    !defined(HPX_HAVE_WRAP_MAIN_CONFIGURED)
+#warning                                                                       \
+    "HPX static-link wrap-main check: you included hpx/hpx_main.hpp but the " \
+    "--wrap=main linker flag has not been applied. Add "                       \
+    "target_link_libraries(<target> PRIVATE HPX::wrap_main) to your "         \
+    "CMakeLists.txt, or pass -Wl,--wrap=main to the linker manually. "        \
+    "Without this flag the HPX runtime will not be initialised correctly."
+#endif
```
hkaiser: This code looks like a duplicate of the content of the new header `static_linker_check.hpp`.

libs/core/config/include/hpx/config/static_linker_check.hpp
```diff
@@ -0,0 +1,38 @@
...
+#if defined(HPX_HAVE_DYNAMIC_HPX_MAIN) && defined(HPX_HAVE_STATIC_LINKING) &&  \
+    (defined(__linux) || defined(__linux__) || defined(linux)) &&              \
+    !defined(HPX_HAVE_WRAP_MAIN_CONFIGURED)
+#warning                                                                       \
+    "HPX static-link wrap-main check: you included hpx/hpx_main.hpp but the " \
+    "--wrap=main linker flag has not been applied. Add "                       \
+    "target_link_libraries(<target> PRIVATE HPX::wrap_main) to your "         \
+    "CMakeLists.txt, or pass -Wl,--wrap=main to the linker manually. "        \
+    "Without this flag the HPX runtime will not be initialised correctly."
+#endif
```
hkaiser: This new header is not used anywhere.

PR #7078, opened 2026-03-21
examples: add sender_diamond quickstart example

360NK:
Fixes #

## Proposed Changes

  - Add `sender_diamond.cpp` to `examples/quickstart` demonstrating a diamond dependency pattern using HPX sender/receiver primitives
  - Add corresponding `CMakeLists.txt` entry

The example shows split for safe multi-consumer reuse, then for branch continuations, when_all for joining dependencies, and sync_wait for extracting the final result. Intended as a minimal educational example of a common fork-join pattern in sender/receiver composition.

Built and ran locally:
A: produced 10
B: 10 * 2 = 20
C: 10 * 3 = 30
D: 20 + 30 = 50
Final Result: 50

## Checklist

### DIFF
--- a/examples/quickstart/CMakeLists.txt
+++ b/examples/quickstart/CMakeLists.txt
@@ -42,6 +42,10 @@ if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD" AND NOT HPX_WITH_STATIC_LINKING)
   set(example_programs ${example_programs} init_globally)
 endif()
 
+if(NOT (MSVC AND HPX_WITH_CXX_MODULES))
+  list(APPEND example_programs sender_diamond)
+endif()
+
 if(HPX_WITH_DISTRIBUTED_RUNTIME)
   set(example_programs
       ${example_programs}
@@ -134,6 +138,7 @@ set(hello_world_2_PARAMETERS THREADS_PER_LOCALITY 4)
 set(hello_world_distributed_PARAMETERS THREADS_PER_LOCALITY 4)
 set(interval_timer_PARAMETERS THREADS_PER_LOCALITY 4)
 set(local_channel_PARAMETERS THREADS_PER_LOCALITY 4)
+set(sender_diamond_PARAMETERS THREADS_PER_LOCALITY 4)
 set(sierpinski_PARAMETERS THREADS_PER_LOCALITY 4)
 set(partitioned_vector_spmd_foreach_PARAMETERS THREADS_PER_LOCALITY 4)
 set(pingpong_PARAMETERS THREADS_PER_LOCALITY 4)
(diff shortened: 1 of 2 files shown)


hkaiser:

isidorostsa:
Hey @360NK, thanks for the PR! This example is not really showing any of the hpx specific facilities like the thread pool scheduler or futures to senders etc. That would be an interesting thing to show off in some examples though, in case you are up for adding them

hkaiser:
@360NK also, MSVC produces a ICE if C++ modules are enabled. Either try to work around the issue (you may want to report the ICE to microsoft) or disable the example under those circumstances.

360NK:
> Hey @360NK, thanks for the PR! This example is not really showing any of the hpx specific facilities like the thread pool scheduler or futures to senders etc. That would be an interesting thing to show off in some examples though, in case you are up for adding them

Thanks! That makes sense, I've identified the unchecked sync_wait result handling and the branches run in parallel comment, they'll both be fixed.  On a broader note, you are right that the current version doesn't show anything HPX specific. I'm thinking now to add a scheduler hop so that the branches actually dispatch to the thread pool, would that be on the right lane, or did you have another direction in mind?

360NK:
> @360NK also, MSVC produces a ICE if C++ modules are enabled. Either try to work around the issue (you may want to report the ICE to microsoft) or disable the example under those circumstances.

Thanks, I've identified the unchecked sync_wait result handling and will include that into the next update. On the MSVC + modules ICE, would a CMake guard to disable the example in that configuration be acceptable?

hkaiser:
> > @360NK also, MSVC produces a ICE if C++ modules are enabled. Either try to work around the issue (you may want to report the ICE to microsoft) or disable the example under those circumstances.
> 
> Thanks, I've identified the unchecked sync_wait result handling and will include that into the next update. On the MSVC + modules ICE, would a CMake guard to disable the example in that configuration be acceptable?

Yes, please also add a comment to that effect to the code itself.

hkaiser:
@360NK Not sure if you noticed, but your example fails compiling even on non-Windows platforms (see https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/20866/workflows/5edf078d-ebb4-4f71-8d78-c5a8f4a5de81/jobs/516895).

360NK:
> @360NK Not sure if you noticed, but your example fails compiling even on non-Windows platforms (see https://app.circleci.com/pipelines/github/STEllAR-GROUP/hpx/20866/workflows/5edf078d-ebb4-4f71-8d78-c5a8f4a5de81/jobs/516895).

Thanks, I missed that CI failure. I haven’t pushed the latest local changes yet, like the MSVC/modules guard and code comment are only done locally so far, and I’ve also fixed the sync_wait result handling locally. I’m currently revising the example to make it more HPX-specific as suggested, and I’ll rerun locally and push an updated version once I’ve confirmed the non-Windows compile issue is resolved.

PR #7081, opened 2026-03-21
test: add right-overlap tests for forward uninitialized_relocate CPOs

Sanchit2662:
I added a `test_right_overlapping<Ex>()` function to both `uninitialized_relocate.cpp` and `uninitialized_relocaten.cpp`. The existing `test_overlapping` only ever tested left-overlap where `dest < first` which is the safe direction for sequential forward copy. It never tested right-overlap `(dest > first, dest < first + count)`, which is the exact scenario that triggers Issue 1's parallel data race.

The new test relocates M=200 trivially relocatable objects `offset=4` positions to the right within the same buffer. Source is `[ptr, ptr+M)`, destination is `[ptr+offset, ptr+M+offset)` they overlap in `[ptr+offset, ptr+M)`. With the overlap fix from Issue 1 in place, this correctly falls back to sequential memmove.
Without the fix, the parallel partitioner would split this into concurrent chunks with overlapping read/write regions.

I only test trivially relocatable types here because right-overlap with sequential forward move-construct is inherently broken for non-trivially relocatable types (the loop overwrites source elements before reading them  same as `std::copy vs std::copy_backward`). Only `memmove` handles right-overlap correctly in the forward direction.
```
  // Right-overlap: source [ptr, ptr+M), dest [ptr+offset, ptr+M+offset)
  uninitialized_relocate(Ex{}, ptr, ptr + M, ptr + offset);
```


hkaiser:
@isidorostsa Would you be able to have a look, please?

hkaiser:
@Sanchit2662 The new tests you added fail on the MacOS CIs and CircleCI. Also, please have a look at the reported inspect issues.

Sanchit2662:
Hi @hkaiser, thanks for the review! The inspect issue and the CI failures have been addressed in the latest commits.

PR #7083, opened 2026-03-22
Fix critical collectives correctness issues (hierarchical partitioning, scatter/all_to_all validation, barrier release re-entry)

abhishek593:
### Description

This PR fixes a few important bugs in collectives and adds regression tests. I found them while studying HPX collectives for Distributed Sort.

Changes:

1. Hierarchical communicator could skip some sites: When `num_sites` is not evenly divisible by `arity`, some sites were getting left out. The fix is to update partitioning logic so every site is included, even when the split is uneven.

2. Hierarchical scatter incorrect chunking: Similarly, when data.size() is not evenly divisible by arity, chunk mapping could be wrong. The fix updates chunk boundary calculation for non-divisible splits.

3. `all_to_all` accepted malformed input sizes: It should reject mismatched payload sizes early. Added test to reproduce this behaviour. 

4. `scatter_to` accepted undersized payloads: It should enforce expected input size. (`local_result.size() == num_sites`).

5. `barrier::release()` could re-enter cleanup from non-HPX threads: When `release()` was called from a non-HPX thread, it delegated to HPX thread execution but then the original call continues running cleanup logic too. The fix is to return after calling the release through another HPX thread.

### DIFF
--- a/libs/full/collectives/tests/unit/barrier.cpp
+++ b/libs/full/collectives/tests/unit/barrier.cpp
@@ -15,6 +15,7 @@
 
 #include <atomic>
 #include <cstddef>
+#include <exception>
 #include <functional>
 #include <memory>
 #include <string>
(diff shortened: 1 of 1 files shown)


hkaiser:
@constracktor Would you mind having a look as well?

libs/full/collectives/src/barrier.cpp
```diff
@@ -139,6 +139,7 @@ namespace hpx { namespace distributed {
                 if (hpx::threads::get_self_ptr() == nullptr)
                 {
                     hpx::run_as_hpx_thread(&barrier::release, this);
+                    return;
```
hkaiser, line 142: Excellent catch!

hkaiser:
@abhishek593 Very nice analysis! Thanks!

hkaiser:
@charan-003 The new slash command doesn't seem to do what's expected. Could you have a look, please?

hkaiser:
The change to the barrier made the MacOS CIs and CircleCI fail reliably. Could you please have a look? 

The other CI problems can be ignored, we will work on those separately.

hkaiser:
retest lsu

hkaiser:
Could you please rebase onto master to fix the compilation errors reported? Sorry for the inconvenience!

abhishek593:
@hkaiser I have done the rebase.

PR #7084, opened 2026-03-22
trigger workflows

charan-003:


charan-003:
Added opened, synchronize, labeled to the trigger types so the workflow runs properly

hkaiser:
@charan-003 the CIs are still not being triggered :/

charan-003:
> @charan-003 the CIs are still not being triggered :/

they're missing the trigger types, which is why ig /ok to test isn't working. Sorry for that
hopefully merging this would change it

hkaiser:
> > @charan-003 the CIs are still not being triggered :/
> 
> they're missing the trigger types, which is why ig /ok to test isn't working. Sorry for that hopefully merging this would change it

I thought that it should already work for this PR, shouldn't it?

charan-003:
> > > @charan-003 the CIs are still not being triggered :/
> > 
> > 
> > they're missing the trigger types, which is why ig /ok to test isn't working. Sorry for that hopefully merging this would change it
> 
> I thought that it should already work for this PR, shouldn't it?

@hkaiser that's what i was thinking as well.
looks like workflow changes only work on the next PR

hkaiser:
> > > > @charan-003 the CIs are still not being triggered :/
> > > 
> > > 
> > > they're missing the trigger types, which is why ig /ok to test isn't working. Sorry for that hopefully merging this would change it
> > 
> > 
> > I thought that it should already work for this PR, shouldn't it?
> 
> @hkaiser that's what i was thinking as well. looks like workflow changes only work on the next PR

Ok, let's see.

PR #7086, opened 2026-03-22
Added delta local accumulation inside thread cleanup loop

### LINKED ISSUE #7085: thread_queue: batch the atomic decrement in cleanup_terminated_locked()
Continuing the work done in #7056 I decided to continue any optimization I could find. In `thread_queue::cleanup_terminated_locked()`, every iteration of the cleanup loop issues two individual atomic decrements
( --terminated_items_count_, --thread_map_count_), causing  multiple cache-line coherence complete trips every call.Instead we could accumulate deltas locally, basically a single fetch after every loop.

toxicteddy00077:
# Description
Closes #7085 

continuing in spirit of #7056, the changes are very simple, I have just repalced the repeated per-item atomic decrements with local counters within the loop and and a `fetch_sub` after the loop. Basically two bulk atomic operations isnteads of many.

### DIFF
--- a/libs/core/schedulers/include/hpx/schedulers/thread_queue.hpp
+++ b/libs/core/schedulers/include/hpx/schedulers/thread_queue.hpp
@@ -390,11 +390,13 @@ namespace hpx::threads::policies {
             if (delete_all)
             {
                 // delete all threads
+                std::int64_t term_delta = 0;
+                std::int64_t map_delta = 0;
                 thread_data* to_delete;
                 while (terminated_items_.pop(to_delete))
                 {
+                    ++term_delta;
                     thread_id_type tid(to_delete);
-                    --terminated_items_count_;
 
                     // this thread has to be managed by this queue, it may have
                     // ended up on the terminate threads list more than once,
@@ -406,10 +408,15 @@ namespace hpx::threads::policies {
                     if (thread_map_.erase(tid) != 0)
                     {
                         recycle_thread(tid);
-                        --thread_map_count_;
-                        HPX_ASSERT(thread_map_count_ >= 0);
+                        ++map_delta;
                     }
                 }
+                terminated_items_count_.fetch_sub(
+                    term_delta, std::memory_order_acq_rel);
+                [[maybe_unused]] std::int64_t remaining =
+                    thread_map_count_.fetch_sub(
+                        map_delta, std::memory_order_acq_rel);
+                HPX_ASSERT(remaining - map_delta >= 0);
             }
             else
             {
@@ -424,11 +431,13 @@ namespace hpx::threads::policies {
                 delete_count = (std::max) (delete_count,
                     static_cast<std::int64_t>(parameters_.min_delete_count_));
 
+                std::int64_t term_delta = 0;
+                std::int64_t map_delta = 0;
                 thread_data* to_delete;
                 while (delete_count && terminated_items_.pop(to_delete))
                 {
+                    ++term_delta;
                     thread_id_type tid(to_delete);
-                    --terminated_items_count_;
 
                     // this thread has to be managed by this queue, it may have
                     // ended up on the terminate threads list more than once,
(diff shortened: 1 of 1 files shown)


toxicteddy00077:
Benchmarks for comparison, left is baseline, right is my imrpovement, very minimal speedups varying form 21-7% across API
image

hkaiser:
@charan-003 please have a look here. The CIs are still not being triggered.

charan-003:
> @charan-003 please have a look here. The CIs are still not being triggered.

@hkaiser

looks like the /ok to test workflows are getting stuck in queue

sad to say but we can either:

1. revert the changes in the PR i created on workflows and make it as it used to be earlier
2. keep it but the queuing is because we have 40+ parallel test jobs running at same time
even if we do something like https://github.com/marketplace/actions/verify-safe-to-test-label it won't fix the queuing - the issue is too many parallel jobs, not how they're triggered

hkaiser:
> > @charan-003 please have a look here. The CIs are still not being triggered.
> 
> @hkaiser
> 
> looks like the /ok to test workflows are getting stuck in queue
> 
> sad to say but we can either:
> 
> 1. revert the changes in the PR i created on workflows and make it as it used to be earlier
> 2. keep it but the queuing is because we have 40+ parallel test jobs running at same time
>    even if we do something like https://github.com/marketplace/actions/verify-safe-to-test-label it won't fix the queuing - the issue is too many parallel jobs, not how they're triggered

Things were run before (even with the queuing). The difference is that now github reports those CI jobs a `skipped`. Let's revert the change for now until we have a better understanding of how to do what you suggested.

@charan-003 could you create a PR that reverts it, please? It can't be automatically reverted anymore because of the second PR that modified things.

charan-003:
> > > @charan-003 please have a look here. The CIs are still not being triggered.
> > 
> > 
> > @hkaiser
> > looks like the /ok to test workflows are getting stuck in queue
> > sad to say but we can either:
> > 
> > 1. revert the changes in the PR i created on workflows and make it as it used to be earlier
> > 2. keep it but the queuing is because we have 40+ parallel test jobs running at same time
> >    even if we do something like https://github.com/marketplace/actions/verify-safe-to-test-label it won't fix the queuing - the issue is too many parallel jobs, not how they're triggered
> 
> Things were run before (even with the queuing). The difference is that now github reports those CI jobs a `skipped`. Let's revert the change for now until we have a better understanding of how to do what you suggested.
> 
> @charan-003 could you create a PR that reverts it, please? It can't be automatically reverted anymore because of the second PR that modified things.

let me do that quickly

hkaiser:
@toxicteddy00077 Could you please take care of the clang-format issues reported?

bench_results/pr2_batch_atomic_results.txt
```diff
@@ -0,0 +1,42 @@
...
+action | WaitEach             baseline 0.4593µs  PR2 0.3646µs +26.0%
+apply | Sliding-Sem           baseline 0.3908µs  PR2 0.3616µs  +8.1%
+apply_hierarchical | latch    baseline 0.1602µs  PR2 0.1558µs  +2.8%
+async | WaitAll               baseline 0.3427µs  PR2 0.2935µs +16.8%
+async | WaitEach              baseline 0.3661µs  PR2 0.3200µs +14.4%
+create_thread | latch         baseline 0.2173µs  PR2 0.1701µs +27.7%
+create_thread_hierarchical    baseline 0.0639µs  PR2 0.0599µs  +6.7%
+register_work | latch         baseline 0.2173µs  PR2 0.1870µs +16.2%
+
+OVERALL GEOMEAN: 1.077x  (~7.7% improvement across all thread counts)
```
hkaiser: Please remove this file from the PR (force push to your branch).

toxicteddy00077: so sorry accidentally staged it before commit

hkaiser:
retest lsu

hkaiser:
Could you please rebase onto master to fix the compilation errors reported? Sorry for the inconvenience.

PR #7097, opened 2026-03-24
Fix scheduler_executor regression

charan-003:


charan-003:
@hkaiser, could you please trigger the stellarbot perftest report on this PR?

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
```diff
@@ -395,6 +391,17 @@ namespace hpx::execution::experimental {
...
+            template <typename CPO>
+                requires(
+                    meta::value<meta::one_of<CPO, set_value_t, set_stopped_t>>)
+            friend constexpr auto tag_invoke(
+                hpx::execution::experimental::get_completion_scheduler_t<CPO>,
+                sender const& s) noexcept
+            {
+                return s.scheduler;
+            }
+#endif
```
charan-003, line 404: this restores the non-stdexec sender customization by exposing `get_completion_scheduler` directly on sender, which preserves the bulk fallback behavior and avoids unnecessary thread_pool_bulk_sender overhead for small partitioned workloads

hkaiser: Could you please explain why the new solution is more costly? What can we do to avoid the additional overhead?

hkaiser:
The perf-tests should run. Also, please ignore the CI failures - those slipped into master unnoticed. I'll take care of those independently.

libs/core/executors/include/hpx/executors/thread_pool_scheduler.hpp
charan-003: with the new unconditional `get_env` on `schedule_sender`, bulk sees a `completion scheduler` and selects the scheduler override path (thread_pool_bulk_sender) instead of the fallback bulk_sender.
for these benchmarks (especially for_each), work is already partitioned into small chunks, so routing each chunk through thread_pool_bulk_sender adds overhead (operation-state setup, queueing/scheduling, synchronization) that dominates useful work.

image


to avoid that overhead
we can either 
1.  keep the stdexec env path only under HPX_HAVE_STDEXEC, and preserve legacy non-stdexec customization so non-stdexec uses fallback bulk_sender as before
([restores previous behaviour](https://github.com/STEllAR-GROUP/hpx/pull/6746/changes#diff-97629d93099c79c02b962250d0185a4323aa5c7d724426cbf6a06dbd313c9a55L330-L344) )

2.  keep unconditional env and bypass thread_pool_bulk_sender for tiny partitioned workloads.

hkaiser: I think we should implement option 1. I would like to avoid performance degradation after fully switching to `std::execution`. Thanks for looking into this in any case!

charan-003: I agree we should avoid reintroducing overhead during the full `std::execution` transition... 
This implementation keeps option 1 for now, and I can follow up with a separate optimization PR for the unconditional-env path

hkaiser (approved):
LGTM, thanks!

Pansysk75:
retest perf

hkaiser:
Thanks again @charan-003!

PR #7133, opened 2026-04-01
FIx wrap_main for local-runtime builds

abhishek593:
## Proposed Changes

This PR fixes local-runtime builds for HPX wrap targets after PR  #7103. When `HPX_WITH_DISTRIBUTED_RUNTIME=OFF`, wrap sources still included distributed-only headers.

### DIFF
--- a/.github/workflows/configure-test-combination.yml
+++ b/.github/workflows/configure-test-combination.yml
@@ -64,6 +64,20 @@ jobs:
               -DHPX_WITH_EXAMPLES=On
           rm -rf *
       
+      - name: Running CMake with distributed runtime off
+        working-directory: build
+        run: |
+          cmake \
+              .. \
+              -G "Ninja" \
+              -DHPX_WITH_MALLOC=system \
+              -DHPX_WITH_FETCH_ASIO=Off \
+              -DHPX_WITH_DISTRIBUTED_RUNTIME=OFF \
+              -DHPX_WITH_NETWORKING=OFF \
+              -DHPX_WITH_TESTS=On \
+              -DHPX_WITH_EXAMPLES=On
+          rm -rf *
+
       - name: Upload build directory
         uses: actions/upload-artifact@v7
         with:
--- a/examples/quickstart/CMakeLists.txt
+++ b/examples/quickstart/CMakeLists.txt
@@ -38,7 +38,10 @@ set(example_programs
     wait_composition
 )
 
-if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD" AND NOT HPX_WITH_STATIC_LINKING)
+if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD"
+   AND NOT HPX_WITH_STATIC_LINKING
+   AND HPX_WITH_DISTRIBUTED_RUNTIME
+)
   set(example_programs ${example_programs} init_globally)
 endif()
 
--- a/examples/quickstart/CMakeLists.txt
+++ b/examples/quickstart/CMakeLists.txt
@@ -38,10 +38,7 @@ set(example_programs
     wait_composition
 )
 
-if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD"
-   AND NOT HPX_WITH_STATIC_LINKING
-   AND HPX_WITH_DISTRIBUTED_RUNTIME
-)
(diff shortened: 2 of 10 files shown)


hkaiser:
@abhishek593 We will have to add github action CIs (on Linux and Windows), and possibly modify one of the MacOS builders to include `HPX_WITH_DISTRIBUTED_RUNTIME=OFF` to be sure that this PR is fine. Would you be able to add those?

.github/workflows/configure-test-combination.yml
```diff
@@ -64,6 +64,20 @@ jobs:
...
+          cmake \
+              .. \
+              -G "Ninja" \
+              -DHPX_WITH_MALLOC=system \
+              -DHPX_WITH_FETCH_ASIO=Off \
+              -DHPX_WITH_DISTRIBUTED_RUNTIME=OFF \
+              -DHPX_WITH_NETWORKING=OFF \
+              -DHPX_WITH_TESTS=On \
+              -DHPX_WITH_EXAMPLES=On
+          rm -rf *
```
hkaiser, line 79: This will run CMake only without actually building any tests (nor running those).

examples/quickstart/CMakeLists.txt
```diff
@@ -38,7 +38,10 @@ set(example_programs
     wait_composition
 )
 
-if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD" AND NOT HPX_WITH_STATIC_LINKING)
+if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD"
+   AND NOT HPX_WITH_STATIC_LINKING
+   AND HPX_WITH_DISTRIBUTED_RUNTIME
+)
```
hkaiser: Why did you disable this test for the local runtime?

libs/core/algorithms/tests/regressions/CMakeLists.txt
```diff
@@ -33,6 +32,10 @@ set(tests
     ranges_facilities
 )
 
+if(HPX_WITH_DISTRIBUTED_RUNTIME)
+  list(APPEND tests for_each_on_main_thread)
```
hkaiser: Same question: Why did you disable this test for the local runtime?

.github/workflows/configure-test-combination.yml
abhishek593: Yes, I do have 3 separate workflows for Linux, Windows, and macOS. But this file had several sanity configuration checks, so I added for local runtime as well. If you want, I can remove.

hkaiser: Sure, let's leave it. More tests never hurt ;-)

examples/quickstart/CMakeLists.txt
abhishek593: The file was failing to compile on local-runtime because of the function `init_winsocket()` defined here `libs/full/init_runtime/src/hpx_main_winsocket.cpp`. This is under the distributed runtime module. Same for the other test. So, we need to add the guard.

hkaiser: I think we should rather guard the call to `init_winsocket()` with `#if defined(HPX_HAVE_DISTRIBUTED_RUTIME)` than to remove the test. The test itself should work perfectly fine for the local runtime.

examples/quickstart/init_globally.cpp
```diff
@@ -92,7 +92,7 @@ class manage_global_runtime
 
         init()
         {
-#if defined(HPX_WINDOWS)
+#if defined(HPX_WINDOWS) && defined(HPX_WITH_DISTRIBUTED_RUNTIME)
```
hkaiser: This should be:
```suggestion
#if defined(HPX_WINDOWS) && defined(HPX_HAVE_DISTRIBUTED_RUNTIME)
```

libs/core/algorithms/tests/regressions/for_each_on_main_thread.cpp
```diff
@@ -61,7 +61,7 @@ class manage_global_runtime
 
         init()
         {
-#if defined(HPX_WINDOWS)
+#if defined(HPX_WINDOWS) && defined(HPX_WITH_DISTRIBUTED_RUNTIME)
```
hkaiser: Same here:
```suggestion
#if defined(HPX_WINDOWS) && defined(HPX_HAVE_DISTRIBUTED_RUNTIME)
```

hkaiser (approved):
LGTM, thanks!

Issue #1338, opened 2014-12-25
Extend parallel algorithms to work with hpx::partitioned_vector et.al.


hkaiser:
The existing parallel algorithms should be extended to seamlessly work with distributed data structures like `hpx::partitioned_vector`. Unfortunately this requires additional code for each specific algorithms.

For more information about the parallel algorithms already implemented see #1141.

Here is the list of algorithms mandated by N4310 which should be ported to the distributed case:
- [x] `adjacent_difference` (see #2859, #3525)
- [x] `adjacent_find` (see #2859, #3525)
- [x] `all_of` `any_of` `none_of` (see #2859, #3525)
- [x] `copy` (see #1346) `copy_n` 
- [ ] `copy_if` 
- [x] `move` (see #2010)
- [x] `count` `count_if` (see #1340)
- [x] `equal` (#7110), `mismatch` (see #7186)
- [x] `exclusive_scan` `inclusive_scan` (see #2287)
- [x] `reduce` `transform`  (see #2725)
- [x] `fill` (#2202)
- [ ] `fill_n`
- [x] `find` `find_if` `find_if_not` (see #2792) 
- [ ] `find_end` `find_first_of` (see #2793)
- [x] `for_each` 
- [x] `for_each_n` (see #2725)
- [x] `generate` (see #1968) 
- [ ] `generate_n`
- [ ] `inner_product` `inplace_merge` `merge`
- [ ] `is_heap` `is_heap_until`
- [ ] `is_partitioned` (#7156), `is_sorted` `is_sorted_until`
- [ ] `lexicographical_compare`
- [x] `max_element` `min_element` `minmax_element` (#1968)
- [ ] `partial_sort` `partial_sort_copy` `partition` `partition_copy` `nth_element` `sort` `stable_partition` `stable_sort`
- [ ] `remove` `remove_copy` `remove_copy_if` `remove_if`
- [x] `replace` `replace_copy` `replace_copy_if` `replace_if`  (#6999)
- [ ] `reverse` `reverse_copy` 
- [ ] `rotate` `rotate_copy`
- [ ] `search` `search_n` 
- [ ] `set_difference` `set_intersection` `set_symmetric_difference` `set_union` `includes`
- [ ] `swap_ranges`
- [ ] `uninitialized_copy` `uninitialized_copy_n`
- [ ] `uninitialized_fill` `uninitialized_fill_n`
- [ ] `unique` `unique_copy`

These were added by N4310:
[...]

PR #7156, opened 2026-04-08
Add segmented is_partitioned

Momensamir12:
Extends #1338 

## Proposed Changes

  - Add segmented version of parallel `is_partitioned`.
  
This completes segmented implementations of `is_sorted`, `is_sorted_until`, `is_partitioned` algorithms family.

## Checklist

### DIFF
--- a/libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/detail/is_partitioned.hpp
+++ b/libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/detail/is_partitioned.hpp
@@ -1,0 +1,87 @@
+//  Copyright (c) 2026 Mo'men Samir
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <hpx/config.hpp>
+#include <hpx/modules/algorithms.hpp>
+#include <hpx/modules/executors.hpp>
+
+#include <iterator>
+#include <type_traits>
+#include <utility>
+
+namespace hpx::parallel::detail {
+    struct local_result
+    {
+        bool local_res;     // local is_partitioned result
+        bool first_elem;    // predicate value of first element in segment
+        bool last_elem;     // predicate value of last element in segment
+
+        template <typename Archive>
+        void serialize(Archive& ar, unsigned)
+        {
+            ar & local_res & first_elem & last_elem;
+        }
+    };
+
+    template <typename Iter>
+    struct seg_is_partitioned
+      : algorithm<seg_is_partitioned<Iter>, local_result>
+    {
+        seg_is_partitioned()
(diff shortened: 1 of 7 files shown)


libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/is_partitioned.hpp
```diff
@@ -0,0 +1,318 @@
...
+                        return false;
+                }
+
+                // Cross-segment boundary check.
+                // Check the last element of each segment, If its false
+                // then the first element of the next segments must be false.
+                bool false_found = false;
+                for (auto const& seg_first : between_segments)
+                {
+                    SegIter prev_last = std::prev(seg_first);
```
hkaiser: This will possibly incur a networking turnaround to read the last value of the previous partition. Wouldn't it be better to simply return the last value together with the `true`/`false` result of the remote operation on the partition to avoid this possible overhead?

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/is_partitioned.hpp
```diff
@@ -0,0 +1,318 @@
...
+
+        return result::get(dataflow(
+            [=](segment_type&& r) mutable -> bool {
+                std::list<std::exception_ptr> errors;
+                parallel::util::detail::handle_remote_exceptions<
+                    ExPolicy>::call(r, errors);
+                std::vector<bool> res = hpx::unwrap(HPX_MOVE(r));
+
+                // Every segment must be locally is_partitioned.
+                for (bool b : res)
```
hkaiser: This could be replaced with `!std::all_of()` (or the HPX equivalent).

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/is_partitioned.hpp
Momensamir12: That's a good idea! I've added a wrapper that returns first and last element of each segment with the result of local `is_partitioned`, now we don't need to dereference cross segment iterators, also now we don't need `between_segments` vector.

libs/full/segmented_algorithms/include/hpx/parallel/segmented_algorithms/detail/is_partitioned.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include <utility>
+
+namespace hpx::parallel::detail {
+    struct local_result
+    {
+        bool local_res;     // local is_partitioned result
+        bool first_elem;    // predicate value of first element in segment
+        bool last_elem;     // predicate value of last element in segment
+
+        template <typename Archive>
```
hkaiser: I think you need no explicit serialization function for this type, we automatically generate one at compile time.

Momensamir12: Done.

hkaiser (approved):
LGTM, thanks!

PR #7157, opened 2026-04-09
execution: add native upon_error and upon_stopped sender adaptors

Sanchit2662:
● Fixes two unchecked items from #5045.

---

So I was looking at the P2300 tracking issue and noticed `upon_error` and `upon_stopped` have been sitting unchecked for a while. These are basically the error and stopped channel equivalents of `then` , pretty fundamental stuff for anyone trying to do error recovery in a sender pipeline without reaching for the heavier `let_error`/`let_stopped`.

Right now if you want to recover from an error and return a default value, you have to write:

```cpp
ex::let_error(sender, [](auto) { return ex::just(default_value); })
```
which is annoying because you're forced to wrap in just() even when you just want to return a plain value. upon_error lets you write:
```
ex::upon_error(sender, [](auto) { return default_value; })
```
Same story for upon_stopped. Also worth noting , these only exist today if you build with HPX_HAVE_STDEXEC. The native path has had nothing. This PR fills that gap.

---

### What I did

- Added upon_error.hpp and upon_stopped.hpp under libs/core/execution/include/hpx/execution/algorithms/
- Both follow the exact same structure as then.hpp : receiver adaptor, sender with completion signatures, CPO with tag_priority
- upon_error: intercepts set_error, calls f(error), sends result as set_value. Value and stopped signals pass through untouched.
- upon_stopped: intercepts set_stopped, calls f(), sends result as set_value. Value and error signals pass through untouched. The
  returned sender never sends stopped (sends_stopped = false).
- Both handle void-returning handlers and re-route any exception thrown inside f as set_error(exception_ptr) : same try_catch_exception_ptr pattern used throughout the existing algorithms.
- Pipe operator (|) works out of the box via the existing partial_algorithm mechanism.
- Registered both headers in the module CMakeLists.txt and added test files covering: error→value conversion, void handler, passthrough channels, handler that throws, pipe operator, chaining with then.
---

### What I didn't do

I intentionally left stopped_as_optional and stopped_as_error out of this PR , they're trivially implementable on top of upon_stopped but I wanted to keep this focused. Happy to do a follow-up for those.

### DIFF
--- a/libs/core/execution/CMakeLists.txt
+++ b/libs/core/execution/CMakeLists.txt
@@ -29,6 +29,8 @@ set(execution_headers
     hpx/execution/algorithms/sync_wait.hpp
     hpx/execution/algorithms/then.hpp
     hpx/execution/algorithms/transfer.hpp
+    hpx/execution/algorithms/upon_error.hpp
+    hpx/execution/algorithms/upon_stopped.hpp
     hpx/execution/algorithms/transfer_just.hpp
     hpx/execution/algorithms/when_all.hpp
     hpx/execution/algorithms/when_all_vector.hpp
--- a/libs/core/execution/include/hpx/execution/algorithms/upon_error.hpp
+++ b/libs/core/execution/include/hpx/execution/algorithms/upon_error.hpp
@@ -1,0 +1,256 @@
+//  Copyright (c) 2021 ETH Zurich
+//  Copyright (c) 2022-2025 Hartmut Kaiser
+//
+//  SPDX-License-Identifier: BSL-1.0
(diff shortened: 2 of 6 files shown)


hkaiser:
@Sanchit2662 As discussed before, we're in the process of removing our own sender/receiver implementation while replacing it unconditionally with NVidia's `std::execution` (https://github.com/NVIDIA/stdexec), e.g., see #7123. For this reason I think it might not be necessary anymore to add things to our own codebase (except if needed to integrate `std::execution`). I will close the ticket you're referring to to avoid misundertsandings of this kind in the future.

PR #7159, opened 2026-04-09
concurrency: harden anchor consistency checks in lockfree::deque

### LINKED ISSUE #7158: Potential Inconsistent State and Race Condition in hpx::lockfree::deque::empty()
### Summary
The current implementation of the Michael-Scott lock-free deque in hpx::lockfree::deque contains a potential logic inconsistency where the empty() method and several pop operations only verify the state of a single pointer (left or right) rather than ensuring the total consistency of the anchor.

**Technical Details** In libs/core/concurrency/include/hpx/concurrency/deque.hpp, the empty() function is implemented as follows:
```

// FIXME: Should we check both pointers here?
[[nodiscard]] bool empty() const noexcept
{
    return anchor_.lrs(std::memory_order_relaxed).get_left_ptr() == nullptr;
}
```
In a concurrent lock-free environment, the left and right pointers of the anchor are updated via atomic CAS operations. During unstable states (e.g., lpush or rpush), one pointer may be updated before the other has been stabilized.

### Impact

False Positives for Emptiness: A call to empty() might return true if left_ptr is null, even if right_ptr still points to a valid node during a transition or a concurrent push_right operation.
Lost Work/Signals: In HPX's task-driven model, if a scheduler or worker thread relies on empty() to decide whether to yield or sleep, it may miss available work seated at the other end of the deque. This leads to "lost signals" where tasks remain stuck until a subsequent event wakes the thread.
Internal Inconsistency: There are similar FIXME comments at lines 452, 509, 560, and 628, indicating that push and pop operations also suffer from this partial check. If the code assumes the deque is empty based on one pointer and skips a pop or stabilize step, the internal anchor remains in an inconsistent state indefinitely.

### Suggested Fix 
The consistency check should be hardened to verify both left and right pointers of the anchor, or to verify the stable status in conjunction with the pointers to ensure a single-ended check is actually valid in that specific state.

### Environment Information

libs/core/concurrency/include/hpx/concurrency/deque.hpp
Class: hpx::lockfree::deque
Algorithm Reference: CAS-Based Lock-Free Algorithm for Shared Deques (M. M. Michael)

arpittkhandelwal:
fixes : #7158 
### Summary
 This PR fixes a long-standing logic inconsistency in hpx::lockfree::deque where the empty() status and boundary checks in push/pop operations only verified a single anchor pointer. This could lead to data races or "lost tasks" in work-stealing schedulers during concurrent transitions.

### Technical Changes

Hardened empty(): The method now performs an atomic load of the full anchor with std::memory_order_acquire and verifies that both Left and Right pointers are null.
Improved Consistency: Updated the transition logic in push_left, push_right, pop_left, and pop_right to ensure both ends of the anchor are verified before assuming the deque is empty.
Removed Code Debt: Addressed and removed several FIXME comments that had been in the codebase since the original implementation.

### DIFF
--- a/libs/core/concurrency/include/hpx/concurrency/deque.hpp
+++ b/libs/core/concurrency/include/hpx/concurrency/deque.hpp
@@ -415,13 +415,13 @@ namespace hpx::lockfree {
             }
         }
 
-        // Not thread-safe.
+        // Thread-safe and non-blocking.
         // Complexity: O(Processes)
-        // FIXME: Should we check both pointers here?
         [[nodiscard]] bool empty() const noexcept
         {
-            return anchor_.lrs(std::memory_order_relaxed).get_left_ptr() ==
-                nullptr;
+            anchor_pair lrs = anchor_.lrs(std::memory_order_acquire);
+            return lrs.get_left_ptr() == nullptr &&
+                lrs.get_right_ptr() == nullptr;
         }
 
         // Thread-safe and non-blocking.
@@ -449,8 +449,8 @@ namespace hpx::lockfree {
                 anchor_pair lrs = anchor_.lrs(std::memory_order_relaxed);
 
                 // Check if the deque is empty.
-                // FIXME: Should we check both pointers here?
-                if (lrs.get_left_ptr() == nullptr)
+                if (lrs.get_left_ptr() == nullptr &&
+                    lrs.get_right_ptr() == nullptr)
                 {
                     // If the deque is empty, we simply install a new anchor
                     // which points to the new node as both its leftmost and
@@ -506,8 +506,8 @@ namespace hpx::lockfree {
                 anchor_pair lrs = anchor_.lrs(std::memory_order_relaxed);
 
                 // Check if the deque is empty.
-                // FIXME: Should we check both pointers here?
-                if (lrs.get_right_ptr() == nullptr)
+                if (lrs.get_right_ptr() == nullptr &&
+                    lrs.get_left_ptr() == nullptr)
                 {
                     // If the deque is empty, we simply install a new anchor
                     // which points to the new node as both its leftmost and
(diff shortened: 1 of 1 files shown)


libs/core/concurrency/include/hpx/concurrency/deque.hpp
```diff
@@ -415,13 +415,13 @@ namespace hpx::lockfree {
...
 
-        // Not thread-safe.
+        // Thread-safe and non-blocking.
         // Complexity: O(Processes)
-        // FIXME: Should we check both pointers here?
         [[nodiscard]] bool empty() const noexcept
         {
-            return anchor_.lrs(std::memory_order_relaxed).get_left_ptr() ==
-                nullptr;
+            anchor_pair lrs = anchor_.lrs(std::memory_order_acquire);
```
hkaiser: What is your rationale of changing the memory ordering here? 

If you believe that under some circumstances the memory ordering needs to be more than `relaxed`, please rewrite the function as:
```
        [[nodiscard]] bool empty(
            std::memory_order mo = std::memory_order_relaxed) const noexcept
        {
            anchor_pair lrs = anchor_.lrs(mo);
            return lrs.get_left_ptr() == nullptr &&
                lrs.get_right_ptr() == nullptr;
        }
```

arpittkhandelwal: Thank you for the feedback sir ! My original rationale for using std::memory_order_acquire was to ensure that if a caller observed the deque as non-empty, the subsequent access to the node's data would be properly synchronized. However, I agree that hardcoding a stronger ordering in empty() is not the right approach — it imposes unnecessary overhead on all callers, most of which (especially in the scheduler's polling loops) are fine with relaxed.

I have updated the implementation to follow your suggestion exactly — empty() now accepts an optional std::memory_order parameter defaulting to std::memory_order_relaxed, giving callers the flexibility to opt into stronger ordering when they specifically need it.

The squashed commit is now force-pushed to the branch. Please let me know if there is anything else to address.

hkaiser (approved):
LGTM, thanks!

PR #7160, opened 2026-04-09
Add hierarchical all_reduce and all_gather via reduce+broadcast composition

iemAnshuman:
While studying the hierarchical communicator from #6834, I implemented
hierarchical `all_reduce` and `all_gather` by composing existing
reduce/gather + broadcast primitives over the hierarchical communicator.

## Approach

- Root sites: `reduce_here` (sync) → `broadcast_to`
- Non-root sites: `reduce_there` (async+get) → `broadcast_from`
- Distinct, monotonically incrementing generations per phase:
  user generation `k` maps to internal generations `2k-1` (reduce) and `2k` (broadcast)
  (per @hkaiser's guidance on #6834)

## Benchmark results

Using HPX's standard `benchmark_collectives_test` binary. Comparing against
`multi_use` flat as the baseline since hierarchical also reuses communicators
across iterations.

Setup: single int, 100 iterations, DGX H100 compute node, 1 HPX thread per locality.

| Procs | Flat multi_use (μs) | Hier arity=2 (μs) | Hier arity=4 (μs) |
|-------|---------------------|--------------------|--------------------|
| 4     | 50                  | 101                | —                  |
| 8     | 98                  | 139                | 141                |
| 16    | 166                 | 198                | 849                |
| 32    | 465                 | 346                | 1,120              |

Hierarchical arity=2 overtakes flat at 32 processes (1.34x speedup).
Per maintainer feedback, arity=2 is the recommended default for ≥16 ranks.

Arity=4 shows regression at P≥16 — appears to originate in the hierarchical
communicator's tree-partitioning logic rather than the composition layer.

## Questions

1. Is reduce→broadcast the right composition, or would a different
   decomposition be preferred?
2. For vector types, should I add a dedicated specialization or adjust
   existing overload resolution?

## Status

Ready for review.

- [x] Correctness tests pass (4, 8, 16 localities, arities 2 and 4)
- [x] Single-node benchmarks on DGX H100 compute node
- [x] Generation mapping fix (2k-1/2k)
- [ ] Larger message sizes
- [ ] Multi-node benchmarks
- [ ] 64-process results (queued on cluster)

### DIFF
--- a/libs/full/collectives/include/hpx/collectives/all_gather.hpp
+++ b/libs/full/collectives/include/hpx/collectives/all_gather.hpp
@@ -387,8 +387,8 @@ namespace hpx::collectives {
 
     ////////////////////////////////////////////////////////////////////////////
     // Hierarchical all_gather: gather (bottom-up) + broadcast (top-down)
-    // Uses 2k/2k+1 generation mapping: user generation k maps to
-    // internal generation 2k (gather phase) and 2k+1 (broadcast phase)
+    // Uses 2k-1/2k generation mapping: user generation k maps to
+    // internal generation 2k-1 (gather phase) and 2k (broadcast phase)
     //
     // Key difference from all_reduce: the gather phase produces vector<T>
     // (O(N) data), so the broadcast phase transfers O(N) instead of a scalar.
--- a/libs/full/collectives/include/hpx/collectives/all_gather.hpp
+++ b/libs/full/collectives/include/hpx/collectives/all_gather.hpp
@@ -420,7 +420,7 @@ namespace hpx::collectives {
         generation_arg const gather_gen(2 * generation);
         generation_arg const broadcast_gen(2 * generation + 1);
 
-        if (this_site.get() == root_site.get())
+        if (this_site == root_site)
         {
             std::vector<arg_type> gathered = gather_here(hpx::launch::sync,
                 communicators, HPX_FORWARD(T, local_result), this_site, gather_gen);
--- a/libs/full/collectives/include/hpx/collectives/all_gather.hpp
+++ b/libs/full/collectives/include/hpx/collectives/all_gather.hpp
@@ -423,8 +423,7 @@ namespace hpx::collectives {
         generation_arg const broadcast_gen(2 * generation + 1);
 
         std::vector<arg_type> gathered = gather_here(hpx::launch::sync,
-            communicators, HPX_FORWARD(T, local_result), this_site,
-            gather_gen);
+            communicators, HPX_FORWARD(T, local_result), this_site, gather_gen);
 
         return broadcast_to(
             communicators, HPX_MOVE(gathered), this_site, broadcast_gen);
@@ -437,11 +436,10 @@ namespace hpx::collectives {
         this_site_arg const this_site = this_site_arg(),
         generation_arg const generation = generation_arg())
     {
-        return all_gather(communicators, HPX_FORWARD(T, local_result),
-            this_site, generation)
+        return all_gather(
+            communicators, HPX_FORWARD(T, local_result), this_site, generation)
             .get();
     }
-
 
     // --- INTERNAL API (Non-root site overloads) ---
     namespace detail {
@@ -457,11 +455,12 @@ namespace hpx::collectives {
 
             if (generation.is_default())
             {
-                return hpx::make_exceptional_future<std::vector<arg_type>>(
-                    HPX_GET_EXCEPTION(hpx::error::bad_parameter,
-                        "hpx::collectives::all_gather (hierarchical)",
-                        "hierarchical all_gather requires an explicit generation "
-                        "number for the 2k/2k+1 internal mapping"));
+                return hpx::make_exceptional_future<
+                    std::vector<arg_type>>(HPX_GET_EXCEPTION(
+                    hpx::error::bad_parameter,
+                    "hpx::collectives::all_gather (hierarchical)",
+                    "hierarchical all_gather requires an explicit generation "
+                    "number for the 2k/2k+1 internal mapping"));
             }
 
             if (this_site.is_default())
--- a/libs/full/collectives/include/hpx/collectives/all_gather.hpp
+++ b/libs/full/collectives/include/hpx/collectives/all_gather.hpp
@@ -422,26 +422,21 @@ namespace hpx::collectives {
 
         if (this_site == root_site)
         {
-            auto gather_fut = gather_here(communicators,
-                HPX_FORWARD(T, local_result), this_site, gather_gen);
+            std::vector<arg_type> gathered = gather_here(communicators,
+                HPX_FORWARD(T, local_result), this_site, gather_gen)
+                                                 .get();
 
-            return gather_fut.then(hpx::launch::sync,
-                [communicators, this_site, broadcast_gen](auto f) {
-                    return broadcast_to(
-                        communicators, f.get(), this_site, broadcast_gen);
-                });
+            return broadcast_to(
+                communicators, HPX_MOVE(gathered), this_site, broadcast_gen);
         }
         else
         {
-            auto gather_fut = gather_there(communicators,
-                HPX_FORWARD(T, local_result), this_site, gather_gen);
+            gather_there(communicators, HPX_FORWARD(T, local_result), this_site,
+                gather_gen)
+                .get();
 
-            return gather_fut.then(hpx::launch::sync,
-                [communicators, this_site, broadcast_gen](auto f) {
-                    f.get();    // Propagate any exceptions from gather phase
-                    return broadcast_from<std::vector<arg_type>>(
-                        communicators, this_site, broadcast_gen);
-                });
+            return broadcast_from<std::vector<arg_type>>(
+                communicators, this_site, broadcast_gen);
         }
     }
 
--- a/libs/full/collectives/include/hpx/collectives/all_gather.hpp
+++ b/libs/full/collectives/include/hpx/collectives/all_gather.hpp
@@ -422,17 +422,18 @@ namespace hpx::collectives {
 
         if (this_site == root_site)
         {
-            std::vector<arg_type> gathered =
-                gather_here(hpx::launch::sync, communicators,
-                    HPX_FORWARD(T, local_result), this_site, gather_gen);
+            std::vector<arg_type> gathered = gather_here(communicators,
+                HPX_FORWARD(T, local_result), this_site, gather_gen)
+                                                 .get();
 
             return broadcast_to(
                 communicators, HPX_MOVE(gathered), this_site, broadcast_gen);
         }
         else
         {
-            gather_there(hpx::launch::sync, communicators,
-                HPX_FORWARD(T, local_result), this_site, gather_gen);
+            gather_there(communicators, HPX_FORWARD(T, local_result), this_site,
+                gather_gen)
+                .get();
 
             return broadcast_from<std::vector<arg_type>>(
                 communicators, this_site, broadcast_gen);
--- a/libs/full/collectives/include/hpx/collectives/all_gather.hpp
+++ b/libs/full/collectives/include/hpx/collectives/all_gather.hpp
@@ -394,6 +394,8 @@ namespace hpx::collectives {
     // Key difference from all_reduce: the gather phase produces vector<T>
     // (O(N) data), so the broadcast phase transfers O(N) instead of a scalar.
 
+    // --- PUBLIC API (Root site overloads) ---
+
     // Root site overload
     template <typename T>
     hpx::future<std::vector<std::decay_t<T>>> all_gather(
@@ -403,7 +405,6 @@ namespace hpx::collectives {
     {
         using arg_type = std::decay_t<T>;
 
-        // Map user generation k to internal generations 2k, 2k+1
         if (generation.is_default())
         {
             return hpx::make_exceptional_future<std::vector<arg_type>>(
@@ -421,53 +422,15 @@ namespace hpx::collectives {
         generation_arg const gather_gen(2 * generation);
         generation_arg const broadcast_gen(2 * generation + 1);
 
-        // Phase 1: hierarchical gather (bottom-up) — produces vector<T>
         std::vector<arg_type> gathered = gather_here(hpx::launch::sync,
             communicators, HPX_FORWARD(T, local_result), this_site,
             gather_gen);
 
-        // Phase 2: hierarchical broadcast (top-down) — sends vector<T>
         return broadcast_to(
             communicators, HPX_MOVE(gathered), this_site, broadcast_gen);
     }
 
-    // Non-root site overload
-    template <typename T>
-    hpx::future<std::vector<std::decay_t<T>>> all_gather_there(
-        hierarchical_communicator const& communicators, T&& local_result,
-        this_site_arg this_site = this_site_arg(),
-        generation_arg const generation = generation_arg())
-    {
-        using arg_type = std::decay_t<T>;
-
-        if (generation.is_default())
-        {
-            return hpx::make_exceptional_future<std::vector<arg_type>>(
-                HPX_GET_EXCEPTION(hpx::error::bad_parameter,
-                    "hpx::collectives::all_gather (hierarchical)",
-                    "hierarchical all_gather requires an explicit generation "
-                    "number for the 2k/2k+1 internal mapping"));
-        }
-
-        if (this_site.is_default())
-        {
-            this_site = agas::get_locality_id();
-        }
-
-        generation_arg const gather_gen(2 * generation);
-        generation_arg const broadcast_gen(2 * generation + 1);
-
-        // Phase 1: hierarchical gather (send up)
-        gather_there(hpx::launch::sync, communicators,
-            HPX_FORWARD(T, local_result), this_site, gather_gen);
-
-        // Phase 2: hierarchical broadcast (receive down) — receives vector<T>
-        return broadcast_from<std::vector<arg_type>>(
-            communicators, this_site, broadcast_gen);
-    }
-
-    ////////////////////////////////////////////////////////////////////////////
-    // Sync versions of hierarchical all_gather
+    // Sync version (Root site)
     template <typename T>
     std::vector<std::decay_t<T>> all_gather(hpx::launch::sync_policy,
         hierarchical_communicator const& communicators, T&& local_result,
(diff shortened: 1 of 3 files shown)


constracktor:
@iemAnshuman I think starting with the existing hierarchical reduce+broadcast and gather+broadcast is a great way to cheaply get significant performance improvements for all_reduce and all_gather. As stated in your proposal, all_to_all can be more tricky and might require more work.

To anwser your questions TTBOMK and within the scope of a possible GSoC project:

1. Since reduce+broadcast are already tree-based, the resulting all_reduce will be as well. When all three collectives are implemented we could start playing around with other compositions such a ring all_reduce. 

2. Not sure if I get the question right. I might have to defer that to @hkaiser. I assume it is related to adjustable message sizes? If that is the case I wonder why you removed the existing structure and replaced it with single values in the last commit. I would be interested in giving this PR a try on multiple nodes with varying message sizes (and multiple parcelports). Keep me posted if you plan to add adjustable message sizes to this draft. For context: The all_reduce collective should typically only transfer small chunks of data in applications. However, this changes for all_to_all in our FFT application where the amount of transferred data is quite large.

3. Would a 2*k and 2*k+1 mapping work? I.e. the user-generation k=0 consumes internal-generation 0 and 1. The user-generation k=1 consumes internal-generation 2 and 3 and so forth.

iemAnshuman:
Thanks for the detailed feedback.

1. Noted - I will keep the current tree based composition for now and focus on getting all three collectives working first.

2. Apologies for the confusion, my question was about C++ overload resolution when the reduction operand is `vector<int>` rather than `int` (template deduction fails with a custom op). Separate from that, you are right that I should not have stripped the adjustable message size structure. I will restore it in the next commit and add runs at 1, 256, 16K, and 64K ints. I have access to a DGX H100 SXM5 node so I can run multi-node benchmarks on my end as well. Happy to coordinate if you'd like to try it on your setup too.

3. The 2k/2k+1 mapping is clean, I will ll implement that. Right now the caller has to step the generation by 2 externally (the impl uses g for reduce and g+1 for broadcast, so my tests and benchmark pass 1, 3, 5, …). Internalizing it as 2k/2k+1 removes that footgun.

hkaiser:
> 2\. Apologies for the confusion, my question was about C++ overload resolution when the reduction operand is `vector<int>` rather than `int` (template deduction fails with a custom op). 

The supplied operator has to conform to the argument types. If you want to reduce over a set of `vector<int>` instances, the operator has to support whatever operation you want to perform.

libs/full/collectives/include/hpx/collectives/all_reduce.hpp
```diff
@@ -423,6 +425,86 @@ namespace hpx::collectives {
...
+
+        if (this_site.is_default())
+        {
+            this_site = agas::get_locality_id();
+        }
+
+        auto [num_sites, this_site_info, root_info] =
+            communicators.get_info_ex();
+
+        generation_arg const bcast_generation(
```
hkaiser: If the user didn't supply an explicit generation value (i.e. `generation == -1`), then the communicator relies on an internal, monotonically increasing counter. Please make sure that this case is properly supported. by your implementation.

libs/full/collectives/tests/performance/bench_hier_allreduce.cpp
```diff
@@ -0,0 +1,125 @@
+#include <hpx/config.hpp>
```
hkaiser: Please add copyright, license, and the SPDX-License-Identifier to all new files:
```
//  Copyright (c) 2026 <your name>
//
```

hkaiser (commented):
Very nice! Please remove the benchmark result files from this PR (please force push to remove them from the commit history as well). Feel free to add those as comments here.

libs/full/collectives/include/hpx/collectives/all_reduce.hpp
iemAnshuman: Added! hierarchical overloads now reject default generation with a clear error since the 2k/2k+1 mapping requires a known base. If generation == -1, we throw bad_parameter.

libs/full/collectives/tests/performance/bench_hier_allreduce.cpp
iemAnshuman: That file has been removed, the all_reduce benchmarks are now in the existing benchmark_collectives.cpp. Added my copyright line to all files with substantial changes (all_reduce.hpp, all_gather.hpp, benchmark_collectives.cpp).

iemAnshuman:
Thank you for the review. Benchmark result files removed from commit history via force push. Results for reference:

| Procs | Flat (μs) | Hier arity=2 (μs) | Hier arity=4 (μs) | Best speedup |
|-------|-----------|--------------------|--------------------|--------------|
| 4     | 27.8      | 50.9               | —                  | flat wins    |
| 8     | 47.5      | 78.2               | 67.3               | flat wins    |
| 16    | 97.8      | 98.8               | 84.2               | 1.2x hier    |
| 32    | 214.1     | 98.9               | 91.6               | 2.3x hier    |
| 64    | 24,601    | 91.4               | 892.8              | 269x hier    |

Also addressed in this push: generation == -1 guard for hierarchical overloads, copyright lines added, and all_reduce integrated into benchmark_collectives.cpp with adjustable message sizes (--test_size and --operation all_reduce).

libs/full/collectives/include/hpx/collectives/all_gather.hpp
```diff
@@ -381,6 +384,112 @@ namespace hpx::collectives {
...
+    {
+        using arg_type = std::decay_t<T>;
+
+        if (this_site.is_default())
+        {
+            this_site = agas::get_locality_id();
+        }
+
+        // Map user generation k to internal generations 2k, 2k+1
+        if (generation == generation_arg())
```
hkaiser: you can write this as:
```suggestion
        if (generation.is_default())
```

libs/full/collectives/include/hpx/collectives/all_reduce.hpp
```diff
@@ -423,6 +427,219 @@ namespace hpx::collectives {
             HPX_FORWARD(T, local_result), HPX_FORWARD(F, op), this_site)
             .get();
     }
+
+    ////////////////////////////////////////////////////////////////////////////
+    // Helper: lift a scalar reduction op to work element-wise on vectors
+    namespace detail {
+
+        template <typename T, typename F>
```
hkaiser: I think you may be able to deduce `T` on the `operator()()` from its arguments instead.

libs/full/collectives/include/hpx/collectives/all_reduce.hpp
```diff
@@ -423,6 +427,219 @@ namespace hpx::collectives {
...
+
+            std::vector<T> operator()(
+                std::vector<T> const& lhs, std::vector<T> const& rhs) const
+            {
+                HPX_ASSERT(lhs.size() == rhs.size());
+                std::vector<T> result;
+                result.reserve(lhs.size());
+                for (std::size_t i = 0; i != lhs.size(); ++i)
+                {
+                    result.push_back(op_(lhs[i], rhs[i]));
```
hkaiser: You may want to rewrite this as:
```suggestion
                    result.push_back(HPX_INVOKE(op_, lhs[i], rhs[i]));
```
to allow for member functions as operators.

libs/full/collectives/include/hpx/collectives/all_gather.hpp
hkaiser: Also, this check may be better placed first in the function. There is no reason to check for the other default values if the generation is not supplied.

libs/full/collectives/include/hpx/collectives/all_gather.hpp
```diff
@@ -381,6 +384,112 @@ namespace hpx::collectives {
...
+            return hpx::make_exceptional_future<std::vector<arg_type>>(
+                HPX_GET_EXCEPTION(hpx::error::bad_parameter,
+                    "hpx::collectives::all_gather (hierarchical)",
+                    "hierarchical all_gather requires an explicit generation "
+                    "number for the 2k/2k+1 internal mapping"));
+        }
+
+        std::size_t const k = generation;
+        generation_arg const gather_gen(2 * k);
+        generation_arg const broadcast_gen(2 * k + 1);
```
hkaiser:
```suggestion
        generation_arg const gather_gen(2 * generation);
        generation_arg const broadcast_gen(2 * generation + 1);
```

libs/full/collectives/include/hpx/collectives/all_gather.hpp
```diff
@@ -381,6 +384,112 @@ namespace hpx::collectives {
...
+            gather_gen);
+
+        // Phase 2: hierarchical broadcast (top-down) — sends vector<T>
+        return broadcast_to(
+            communicators, HPX_MOVE(gathered), this_site, broadcast_gen);
+    }
+
+    // Non-root site overload
+    template <typename T>
+    hpx::future<std::vector<std::decay_t<T>>> all_gather_there(
```
hkaiser: This function is not meant for public consumption, is it? If this is not part of the API, please move it to `namespace detail`.

hkaiser:
Also, please rebase onto master and resolve the merge conflicts.

hkaiser:
> Thank you for the review. Benchmark result files removed from commit history via force push. Results for reference:
> 
> Procs	Flat (μs)	Hier arity=2 (μs)	Hier arity=4 (μs)	Best speedup
> 4	27.8	50.9	—	flat wins
> 8	47.5	78.2	67.3	flat wins
> 16	97.8	98.8	84.2	1.2x hier
> 32	214.1	98.9	91.6	2.3x hier
> 64	24,601	91.4	892.8	269x hier
> Also addressed in this push: generation == -1 guard for hierarchical overloads, copyright lines added, and all_reduce integrated into benchmark_collectives.cpp with adjustable message sizes (--test_size and --operation all_reduce).

Should we fall back to the flat collectives if the number of connected sites is less than 16? This would require more measurements, but might be worth investigating.

iemAnshuman:
> Should we fall back to the flat collectives if the number of connected sites is less than 16? This would require more measurements, but might be worth investigating.

Good idea. The benchmark data supports a crossover around 16-32 processes, below that, flat is faster due to lower overhead. An adaptive threshold that falls back to flat below a configurable cutoff (defaulting to ~16 sites) would make sense. I'll add this as a follow-up once the base implementation is solid and we have multi-node data to confirm the crossover holds across different cluster topologies.

libs/full/collectives/include/hpx/collectives/all_reduce.hpp
iemAnshuman: Done. Also added the <hpx/functional/invoke.hpp> include to support the HPX_INVOKE change.

libs/full/collectives/include/hpx/collectives/all_gather.hpp
iemAnshuman: Done. Moved all non-root site overloads and their sync versions to namespace detail.

iemAnshuman:
> Also, please rebase onto master and resolve the merge conflicts.

Done. Rebased onto master, resolved the conflicts in all_reduce.hpp and all_gather.hpp, and force-pushed.

hkaiser:
Please have a look at the reported CI issues (feel free to ignore anything reported from CircleCI).

libs/full/collectives/include/hpx/collectives/all_gather.hpp
hkaiser: Well, not yet...

iemAnshuman: Ah, you were right! I missed the vector and sync overloads in the first pass. I've just pushed an update that consolidates all _there overloads (scalar, vector, and sync) into a single namespace detail block at the bottom of both files.

iemAnshuman:
> Please have a look at the reported CI issues (feel free to ignore anything reported from CircleCI).

Thanks! Removed the conflicting #include <hpx/functional/invoke.hpp> to fix the C++20 modules build, and ran clang-format. The latest push should clear the CI.

hkaiser:
> > Please have a look at the reported CI issues (feel free to ignore anything reported from CircleCI).
> 
> Thanks! Removed the conflicting #include <hpx/functional/invoke.hpp> to fix the C++20 modules build, and ran clang-format. The latest push should clear the CI.

Inspect is still not happy, though:
```
/libs/full/collectives/tests/performance/benchmark_collectives.cpp:
    Non-ASCII: line 479
```

libs/full/collectives/tests/performance/benchmark_collectives.cpp
```diff
@@ -423,6 +426,67 @@ void test_gather_hierarchical(int arity, int lpn, std::size_t iterations,
...
+        {
+            // NOLINTNEXTLINE(bugprone-use-after-move)
+            ft_data =
+                all_reduce(communicators, std::move(send_data), vector_adder{},
+                    this_site_arg(this_locality), generation_arg(i + 1));
+        }
+        else
+        {
+            // NOLINTNEXTLINE(bugprone-use-after-move)
+            ft_data = all_reduce_there(communicators, std::move(send_data),
```
hkaiser: This is not right (sorry if I misunderstand something). `all_reduce_there` either is an internal API, then it shouldn't be called by the user, or it is not and internal API, in which case I think it could be unified behind a common name `all_reduce`.

iemAnshuman: Agreed! I have removed the _there overloads and unified the dispatch logic directly into the main all_reduce and all_gather functions using root_site. The benchmark code is much cleaner now without the manual if else blocks.

iemAnshuman:
> > > Please have a look at the reported CI issues (feel free to ignore anything reported from CircleCI).
> > 
> > 
> > Thanks! Removed the conflicting #include <hpx/functional/invoke.hpp> to fix the C++20 modules build, and ran clang-format. The latest push should clear the CI.
> 
> Inspect is still not happy, though:
> 
> ```
> /libs/full/collectives/tests/performance/benchmark_collectives.cpp:
>     Non-ASCII: line 479
> ```

Good catch, there was a non-ASCII character hiding in a comment. Just pushed a fix for this along with the unified API changes, and inspect is fully green locally now.

libs/full/collectives/include/hpx/collectives/all_reduce.hpp
```diff
@@ -423,6 +426,144 @@ namespace hpx::collectives {
...
+        else
+        {
+            reduce_there(hpx::launch::sync, communicators,
+                HPX_MOVE(local_result), HPX_MOVE(vec_op), this_site,
+                reduce_gen);
+
+            return broadcast_from<std::vector<T>>(
+                communicators, this_site, broadcast_gen);
+        }
+    }
```
hkaiser, line 481: Why do we need a separate overload for `vector<T>`? Shouldn't the user supply an appropriate operator that works with the supplied data type?

iemAnshuman: You are right. Removed the vector<T> overloads and vector_reduce_op helper. The user supplies an operator that matches their data type directly.

hkaiser: You could re-add your vector-operator to one of the tests, though ;-)

hkaiser (dismissed):
LGTM, thanks!

hkaiser:
Could you please rerun the benchmarks to see what impact your latest changes may have?

iemAnshuman:
> Could you please rerun the benchmarks to see what impact your latest changes may have?

@hkaiser Will do. I'll rerun the hierarchical all_reduce benchmarks with the updated implementation and post results here shortly.

iemAnshuman:
> Could you please rerun the benchmarks to see what impact your latest changes may have?

@hkaiser Running the full benchmark suite (flat vs hierarchical all_reduce, same process counts and arities as before) on the DGX cluster. Will post the updated table here by end of tomorrow.

iemAnshuman:
> Could you please rerun the benchmarks to see what impact your latest changes may have?

<p>@hkaiser Benchmark results after the generation mapping fix (<code>2k-1</code> / <code>2k</code> instead of <code>2k</code> / <code>2k+1</code>):</p>
<p>These results use HPX's standard <code>benchmark_collectives_test</code> binary (not the custom benchmark from my earlier table, which has since been removed). The standard benchmark separates flat into <code>single_use</code> and <code>multi_use</code>. I'm comparing against <code>multi_use</code> as the fair baseline since hierarchical also reuses communicators across iterations.</p>
<p>Setup: DGX H100, single node, 1 HPX thread per locality, 100 iterations.</p>

Procs | Flat multi_use (μs) | Hier arity=2 (μs) | Hier arity=4 (μs)
-- | -- | -- | --
4 | 50 | 101 | —
8 | 98 | 139 | 141
16 | 166 | 198 | 849
32 | 465 | 346 | 1,120


<p>Hierarchical arity=2 overtakes flat at 32 processes (1.34x speedup). The trend suggests this gap widens at 64+ processes that run is currently queued on the cluster.</p>
<p>Arity=4 shows unexpected regression at P>=16. This appears to originate inside the hierarchical communicator's tree partitioning logic rather than in my reduce+broadcast composition layer. Investigating.</p>
</blockquote>

hkaiser:
> > Could you please rerun the benchmarks to see what impact your latest changes may have?
> 
> @hkaiser Benchmark results after the generation mapping fix (`2k-1` / `2k` instead of `2k` / `2k+1`):
> 
> These results use HPX's standard `benchmark_collectives_test` binary (not the custom benchmark from my earlier table, which has since been removed). The standard benchmark separates flat into `single_use` and `multi_use`. I'm comparing against `multi_use` as the fair baseline since hierarchical also reuses communicators across iterations.
> 
> Setup: DGX H100, single node, 1 HPX thread per locality, 100 iterations.
> 
> Procs	Flat multi_use (μs)	Hier arity=2 (μs)	Hier arity=4 (μs)
> 4	50	101	—
> 8	98	139	141
> 16	166	198	849
> 32	465	346	1,120
> Hierarchical arity=2 overtakes flat at 32 processes (1.34x speedup). The trend suggests this gap widens at 64+ processes that run is currently queued on the cluster.
> 
> Arity=4 shows unexpected regression at P>=16. This appears to originate inside the hierarchical communicator's tree partitioning logic rather than in my reduce+broadcast composition layer. Investigating.

So the bottomline is to use an arity of two all the time (at least for now and for more than 16 ranks). That's a good result.

iemAnshuman:
> > > Could you please rerun the benchmarks to see what impact your latest changes may have?
> > 
> > 
> > @hkaiser Benchmark results after the generation mapping fix (`2k-1` / `2k` instead of `2k` / `2k+1`):
> > These results use HPX's standard `benchmark_collectives_test` binary (not the custom benchmark from my earlier table, which has since been removed). The standard benchmark separates flat into `single_use` and `multi_use`. I'm comparing against `multi_use` as the fair baseline since hierarchical also reuses communicators across iterations.
> > Setup: DGX H100, single node, 1 HPX thread per locality, 100 iterations.
> > Procs	Flat multi_use (μs)	Hier arity=2 (μs)	Hier arity=4 (μs)
> > 4	50	101	—
> > 8	98	139	141
> > 16	166	198	849
> > 32	465	346	1,120
> > Hierarchical arity=2 overtakes flat at 32 processes (1.34x speedup). The trend suggests this gap widens at 64+ processes that run is currently queued on the cluster.
> > Arity=4 shows unexpected regression at P>=16. This appears to originate inside the hierarchical communicator's tree partitioning logic rather than in my reduce+broadcast composition layer. Investigating.
> 
> So the bottomline is to use an arity of two all the time (at least for now and for more than 16 ranks). That's a good result.

Agreed, arity=2 is the clear default for now. I can add a threshold so it falls back to flat below 16 ranks if you would like that in this PR.
64-process run is queued, will update when it lands. Is there anything else you would like changed before review?

hkaiser:
> Agreed, arity=2 is the clear default for now. I can add a threshold so it falls back to flat below 16 ranks if you would like that in this PR.

That might be a good idea.

> 64-process run is queued, will update when it lands. Is there anything else you would like changed before review?

I'm just waiting for the CI to cycle.

hkaiser (approved):
LGTM, thanks!

PR #7161, opened 2026-04-09
fix(find_first_of): add missing early return and fix datapar iterator dependency

aneek22112007-tech:
### Summary
Fixes a correctness/consistency bug in parallel `find_first_of` and a compilation error in its SIMD implementation.

### 1. The Core Bug (Parallel Partition)
In `libs/core/algorithms/include/hpx/parallel/algorithms/detail/find.hpp`, the inner search loop over needles did not short-circuit. 
```cpp
// FIXED
for (FwdIter2 iter = s_first; iter != s_last; ++iter)
{
    if (HPX_INVOKE(cmp, v, *iter))
    {
        tok.cancel(i);
        return; // Added this missing return
    }
}
```

### 2. SIMD (datapar) Fix
Fixed a bug in `libs/core/algorithms/include/hpx/parallel/datapar/find.hpp` where the code used `it + idx`, which would fail to compile for non-random-access iterators (like `std::forward_list`). Now correctly uses the value and index provided by the loop utility.

### 3. Coding Standards (`inspect`)
- Added missing `#include <atomic>` in `findfirstof_tests.hpp`.
- Replaced non-ASCII characters (`→`) in comments with standard ASCII (`->`) to satisfy the `inspect` tool.

### Testing
Updated `find_first_of_edge_cases_test()` to verify:
- Empty ranges, single-element ranges, and predicate invocation count consistency (regression guard for the short-circuit).
- All tests passed locally with `inspect` clean.


hkaiser (dismissed):
Nice catch! LGTM!

hkaiser:
Could you please fix the inspect issues reported. Also, please rebase onto master and resolve the conflicts.

aneek22112007-tech:
Hi @hkaiser,

I have rebased the PR onto the latest master and resolved the conflicts.

I also fixed the inspect issues by adding the missing <atomic> header and replacing the non-ASCII characters in the test comments with standard ones. I've verified the changes locally by running the inspect tool and the find_first_of unit tests, and everything is now passing.

Thanks for the review!

aneek22112007-tech:
It appears GitHub automatically closed this PR because, after rebasing and applying the requested inspect fixes, the branch became identical to master (as the changes were already integrated via #7163).

Since the fix and the coding standard improvements are now live in the repository, this PR is indeed redundant. Thank you for the guidance!

hkaiser (approved):
LGTM, thanks!

PR #7163, opened 2026-04-10
Fix missing early termination in parallel `find_first_of` inner loop

aneek22112007-tech:
## Title
Fix missing early termination in parallel `find_first_of` inner loop

## Summary

Fixes a correctness/consistency bug in the parallel-partition kernel of
`hpx::find_first_of` where the inner search loop over `[s_first, s_last)`
did not short-circuit after finding the first matching needle.

---

## The Bug

**File:** `libs/core/algorithms/include/hpx/parallel/algorithms/detail/find.hpp`  
**Overload:** `sequential_find_first_of_t::tag_fallback_invoke` (partitioned form,
used by `par` and `par_unseq` policies)

```cpp
// BEFORE (buggy)
for (FwdIter2 iter = s_first; iter != s_last; ++iter)
{
    if (HPX_INVOKE(cmp, v, *iter))
    {
        tok.cancel(i);
        // ❌ inner loop continues — remaining needles are compared unnecessarily
    }
}
```

When a haystack element matches any needle the code correctly calls
`tok.cancel(i)` to record the match position, but then **keeps iterating**
over the remaining needles. This causes:

1. **Redundant predicate invocations** — O(S − match_pos) extra calls per
   matched outer element, where S = `distance(s_first, s_last)`.
2. **Observable inconsistency with `seq`** — the full-range sequential
   overload already does `return first;` on the first hit, so the two paths
   are visibly inconsistent when the predicate has side effects (legal under
   the C++ Standard) or when the needle range contains duplicates.

The `datapar` counterpart (`datapar_find_first_of` in `datapar/find.hpp`)
is unaffected — it uses an independent `local_tok` per outer element with
its own early-exit path.

---

## The Fix

```cpp
// AFTER (fixed)
for (FwdIter2 iter = s_first; iter != s_last; ++iter)
{
    if (HPX_INVOKE(cmp, v, *iter))
    {
        tok.cancel(i);
        return;    // ✅ short-circuit: no need to check remaining search elements
    }
}
```

A single `return;` after `tok.cancel(i)` makes the `par`/`par_unseq`
partition kernel consistent with the `seq` path.

---

## Reproducer

```cpp
std::atomic<int> call_count{0};
auto counting_pred = [&](int a, int b) -> bool {
    ++call_count;
    return a == b;
};

std::vector<int> haystack = {7, 100, 200};
std::vector<int> needles  = {7, 7};   // deliberate duplicates

// Under seq:            index 0 costs 1 predicate call (returns on first hit)
// Under par (before fix): index 0 costs 2 calls → 6 total instead of 5
hpx::find_first_of(hpx::execution::par,
    haystack.begin(), haystack.end(),
    needles.begin(),  needles.end(),
    counting_pred);

assert(call_count <= 5);  // fails before this fix under par/par_unseq
```

---

## Testing

Existing tests: `findfirstof`, `findfirstof_binary`, `findfirstof_sender` — no regressions expected.

New `find_first_of_edge_cases_test()` verifies:
1. Empty haystack → returns `last`
2. Empty needle range → returns `last`
3. Single-element haystack with match → returns `begin`
4. Single-element haystack with no match → returns `last`
5. Predicate invocation count with duplicate needles proves inner loop stops on first hit (regression guard)

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/detail/find.hpp
@@ -417,18 +417,19 @@ namespace hpx::parallel::detail {
             FwdIter2 s_last, std::size_t base_idx, std::size_t part_size,
             Token& tok, Pred&& op, Proj1&& proj1, Proj2&& proj2)
         {
-            util::loop_idx_n<ExPolicy>(base_idx, it, part_size, tok,
-                [&tok, &s_first, &s_last, &op, &proj1, &proj2](
-                    auto v, std::size_t i) -> void {
-                    util::compare_projected<Pred, Proj1, Proj2> cmp(
-                        HPX_FORWARD(Pred, op), HPX_FORWARD(Proj1, proj1),
-                        HPX_FORWARD(Proj2, proj2));
+            util::compare_projected<Pred, Proj1, Proj2> cmp(
+                HPX_FORWARD(Pred, op), HPX_FORWARD(Proj1, proj1),
+                HPX_FORWARD(Proj2, proj2));
 
+            util::loop_idx_n<ExPolicy>(base_idx, it, part_size, tok,
+                [cmp = HPX_MOVE(cmp), s_first, s_last, &tok](
+                    auto v, std::size_t i) mutable -> void {
                     for (FwdIter2 iter = s_first; iter != s_last; ++iter)
                     {
                         if (HPX_INVOKE(cmp, v, *iter))
                         {
                             tok.cancel(i);
+                            return;
                         }
                     }
                 });
--- a/libs/core/algorithms/tests/unit/algorithms/findfirstof.cpp
+++ b/libs/core/algorithms/tests/unit/algorithms/findfirstof.cpp
@@ -1,5 +1,6 @@
 //  Copyright (c) 2021 Srinivas Yadav
 //  copyright (c) 2014 Grant Mercer
+//  Copyright (c) 2024 Aneek Barman
 //
 //  SPDX-License-Identifier: BSL-1.0
 //  Distributed under the Boost Software License, Version 1.0. (See accompanying
@@ -78,6 +79,18 @@ void find_first_of_bad_alloc_test()
     test_find_first_of_bad_alloc<std::forward_iterator_tag>();
 }
 
+void find_first_of_edge_cases_test()
+{
+    using namespace hpx::execution;
+    // Run edge-case checks for all three standard policies.
+    // These verify empty-range boundary conditions and that the inner
+    // search loop short-circuits after the first needle match
+    // (regression guard for the missing `return;` fix).
+    test_find_first_of_edge_cases(seq);
+    test_find_first_of_edge_cases(par);
+    test_find_first_of_edge_cases(par_unseq);
+}
+
 int hpx_main(hpx::program_options::variables_map& vm)
 {
     if (vm.count("seed"))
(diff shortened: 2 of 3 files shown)


libs/core/algorithms/include/hpx/parallel/algorithms/detail/adjacent_find.hpp
```diff
@@ -40,8 +40,15 @@ namespace hpx::parallel::detail {
             util::loop_idx_n<ExPolicy>(base_idx, part_begin, part_count, tok,
                 [&pred_projected, &tok](auto t, std::size_t i) {
                     using hpx::get;
+                    if (tok.was_cancelled(i))
```
hkaiser: I know that we might do this inconsistently, however I'd like to check for whether the token was cancelled once per partition (before running the loop on a partition) and not once per iteration. The latter introduces too much overhead.

libs/core/algorithms/include/hpx/parallel/algorithms/detail/adjacent_find.hpp
```diff
@@ -40,8 +40,15 @@ namespace hpx::parallel::detail {
...
+                    {
+                        return;
+                    }
                     if (pred_projected(get<0>(t), get<1>(t)))
+                    {
                         tok.cancel(i);
+                        return;
```
hkaiser: This doesn't change anything as the lambda returns in anycase.

libs/core/algorithms/include/hpx/parallel/algorithms/detail/adjacent_find.hpp
aneek22112007-tech: That makes total sense. I didn't realize that checking it per iteration would add that much overhead, especially since the loop utility already handles it once per partition. I've reverted this change to keep everything efficient. Thanks for the explanation!

libs/core/algorithms/include/hpx/parallel/algorithms/detail/adjacent_find.hpp
aneek22112007-tech: You're absolutely right, that return was redundant since the lambda would have ended there anyway. I've removed it to keep the code clean and focused on the find_first_of fix where the early exit actually matters. Thanks for the review!

aneek22112007-tech: Thanks for the feedback! I've reverted the other algorithms to avoid that overhead. I also tightened up the find_first_of fix by moving the predicate construction outside the loop, keeping only the inner short-circuit which is now much cleaner. Verified locally and passed all tests!

hkaiser (dismissed):
LGTM, thanks!

hkaiser:
Please fix the reported inspect issues:
```
/libs/core/algorithms/tests/unit/algorithms/findfirstof_tests.hpp: *I* missing #include (atomic) for symbol std::atomic on line 409
    Non-ASCII: line 333
```

aneek22112007-tech:
> Please fix the reported inspect issues:
> 
> ```
> /libs/core/algorithms/tests/unit/algorithms/findfirstof_tests.hpp: *I* missing #include (atomic) for symbol std::atomic on line 409
>     Non-ASCII: line 333
> ```

Done! Added the missing <atomic> include and stripped out the non-ASCII characters the inspect tool was complaining about. Let me know if anything else needs adjusting!

hkaiser (approved):
LGTM, thanks!

PR #7164, opened 2026-04-11
pass executor to hpx::dataflow in sort and partial_sort

guptapratykshh:
related to #7123 

## Proposed Changes

- Without an explicit executor, `hpx::dataflow(lambda, future1, future2)` is intercepted by stdexec which unwraps `hpx::future<T>` before passing to the lambda, breaking the `(hpx::future<T>&&, hpx::future<T>&&)` signatures in `sort.hpp` and `partial_sort.hpp`
- Passing `policy.executor()` as the first arg forces HPX's executor-based dataflow overload, preserving future wrapping — required for mandatory stdexec compatibility

## Checklist


hkaiser (approved):
LGTM, thanks!

PR #7165, opened 2026-04-11
execution: fix split scheduler preservation for late subscribers (P2300)

arpittkhandelwal:
### Problem Statement
This PR addresses a P2300 correctness violation in the split algorithm. Previously, the split sender failed to preserve the associated scheduler when a receiver connected after the predecessor had already completed (the "late subscriber" scenario). In these cases, the completion signal was fired inline, bypassing the execution context guaranteed by the sender's completion scheduler.

### Proposed Changes

- **Virtualized Completion Hooks:** Introduced a virtual void schedule_completion(continuation_type&&) method to the shared_state base class. This allows subclasses to reroute completion signals through the appropriate execution context.

- **Scheduler-Aware shared_state:** Implemented shared_state_scheduler<Sched>, a new subclass that captures the attached scheduler. Overrode schedule_completion to dispatch stored continuations via hpx::execution::experimental::schedule(sched).

- **Safe Asynchronous Management:** Implemented a self-owning schedule_op_holder to manage the lifetime of the schedule() operation state.
- Memory Safety: Adopted the standard HPX allocator pattern, rebinding the shared_state allocator to handle internal task metadata without raw new/delete.
- Race Prevention: Added an intrusive_ptr owner guard before calling start() to prevent use-after-free if a scheduler executes synchronously.
- 

- **CPO & Dispatch Refactoring:** Updated split_t overloads to support generic schedulers, enabling both automatic scheduler discovery and explicit injection.
- Cleaned up constructor SFINAE in split_sender to handle no_scheduler, run_loop_scheduler, and generic Scheduler types without ambiguity.

**Verification Results**
New Test Suite: Added algorithm_split_scheduler.cpp which specifically targets the late-subscriber race condition.
Regression Testing: Verified that the legacy no_scheduler and run_loop paths remain unaffected.
Performance: Used HPX_NO_UNIQUE_ADDRESS and intrusive pointers to keep metadata overhead at an absolute minimum.

### DIFF
--- a/libs/core/execution/include/hpx/execution/algorithms/let_error.hpp
+++ b/libs/core/execution/include/hpx/execution/algorithms/let_error.hpp
@@ -265,7 +265,7 @@ namespace hpx::execution::experimental {
                                 // TODO: receiver is moved before the visit, but
                                 // the invoke inside the visit may throw.
                                 r.op_state.predecessor_error
-                                    .template emplace<Error>(
+                                    .template emplace<std::decay_t<Error>>(
                                         HPX_FORWARD(Error, error));
 
                                 hpx::visit(
--- a/libs/core/execution/include/hpx/execution/algorithms/split.hpp
+++ b/libs/core/execution/include/hpx/execution/algorithms/split.hpp
@@ -283,6 +283,19 @@ namespace hpx::execution::experimental {
                     }
                 };
 
+                // schedule_completion dispatches a stored continuation to
+                // the correct execution context. The base implementation fires
+                // the continuation inline (no scheduler attached). Subclasses
+                // override this to reroute through a specific scheduler,
+                // ensuring P2300 get_completion_scheduler guarantees hold for
+                // late-arriving subscribers (i.e. when predecessor_done is
+                // already true at the time add_continuation is called).
+                virtual void schedule_completion(
+                    continuation_type&& continuation)
+                {
+                    continuation();
+                }
+
                 virtual void set_predecessor_done()
                 {
                     predecessor_done = true;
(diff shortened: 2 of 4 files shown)


hkaiser:
@arpittkhandelwal What's the difference to #6911?

arpittkhandelwal:
> @arpittkhandelwal What's the difference to #6911?

I think the main difference between this PR and #6911 is all about where the 'source of truth' for the scheduler lives. PR #6911 takes a receiver-centric approach—it essentially looks downstream at the moment of connection to see if the receiver has a preferred scheduler. This is a nice safety net, but it doesn't quite address the core problem: the split(scheduler, sender) call itself is currently ignoring the scheduler we specifically gave it whenever a subscriber arrives late.
In this PR, I’ve gone with a sender-centric approach to make sure we actually follow the algorithm's contract. By virtualizing the completion dispatch through a schedule_completion hook in the shared state, we get a few big wins:

**Better Compliance:** The split_sender now strictly follows the P2300 contract by completing on the provided scheduler, even if the receiver doesn't have an environment or a specific scheduler of its own.
**Proper Discovery:** It allows the sender to correctly report its scheduler through get_completion_scheduler.
**Efficiency:** Instead of using start_detached for every late subscriber (which adds extra allocations), we use a specialized shared_state_scheduler. This fits much better into the existing HPX shared-state patterns and keeps things efficient.
Basically, while #6911 is a good generic fix, this implementation feels like the architectural step needed to make HPX's split truly P2300 compliant.

hkaiser:
@isidorostsa could you have a look, please?

arpittkhandelwal:
@isidorostsa  ping

isidorostsa:
Hey @arpittkhandelwal thanks for the PR!
As you may have seen, we are in the process of retiring our internal stdexec implementation (https://github.com/TheHPXProject/hpx/pull/7123).
Is there a reason to single out `split` and keep our own implementation active?

hkaiser:
@arpittkhandelwal I'll close this as obsolete as we have removed our own S&R implementation.

Issue #713, opened 2013-02-14
Add performance counters monitoring connection caches


hkaiser:
The connection caches in the parcelports need to be monitorable to allow for adjustments of their (connection) limits. We need to add performance counters monitoring connection caches.

PR #7166, opened 2026-04-11
parcelset: add connection cache saturation counters and exhaustion alerting

Sanchit2662:
## Summary
So I was looking through the HPX parcelport code and found something that's been quietly broken for a while. When the connection cache gets completely full (like all connections are checked out and there's no space left), the code just silently gives up and defers the parcel. No log, no counter, nothing. You'd have no idea   it happened unless you were staring at the source.                                                                                                                
                                                                                                                                                                     
The thing that annoyed me is that there's already a cache-misses counter, but it counts two completely different things together. A "miss" when a locality is seen for the first time is totally normal and fine. But a "miss" because the pool is exhausted and a parcel had to be deferred? That's a real problem and it should have   its own counter. Mixing them together makes the metric basically useless for debugging saturation.

So what I did was add a separate reservation_failures counter that only goes up when get_or_reserve() returns false due to actual pool exhaustion. That way you can   watch it and know for sure whether you're hitting the connection limit or not.

I also added two more counters, cache-connections and cache-max-connections, so you can just check how many connections are currently tracked vs. how many are allowed. If cache-connections == cache-max-connections and cache-reservation-failures is climbing, you know to bump hpx.max_connections in your ini file.

On top of that, I added a warning log that fires when the pool exhausts, but it's throttled to at most once every 5 seconds so it doesn't spam the logs if things are bad for a while.
                                                                                                                                                                       
All the new counters show up under /parcelport/count/<pp>/cache-reservation-failures, /parcelport/count/<pp>/cache-connections, and /parcelport/count/<pp>/cache-max-connections, following the same naming pattern as the existing ones.
                                                                                                                                                                       
This originally came from issue #713 back in 2013 where someone asked for connection cache monitoring. The basic counters were added then, but this specific gap     was never filled.


hkaiser (commented):
Very nice! 

Could you please add related documentation as well? Likely somewhere here: https://github.com/STEllAR-GROUP/hpx/blob/0db92d07520bc003fa0a106186f175b603fe8cb9/docs/sphinx/manual/optimizing_hpx_applications.rst?plain=1#L1493-L1524

hkaiser (commented):
LGTM, thanks!

hkaiser (dismissed):

hkaiser:
@Sanchit2662 Could you please address the clang-format issues reported?

hkaiser (approved):
LGTM, thanks!

PR #7168, opened 2026-04-11
Co-locate SPSC channel cached indices with their atomics

ArivoliR:
Packed each side's atomic index and its cache of the other side's index into one cache line, so the producer and consumer hot paths each touch exactly on cache line instead of two.

Follow-up to #7115

## Proposed Changes
    
- Group `head_` with `tail_cached_` into a single `head_data` struct, and `tail_` with `head_cached_` into a single `tail_data` struct
- Wrap each struct in `cache_aligned_data` so the two sides still live on separate cache lines, but each side's state fits in one line

## Checklist

### DIFF
--- a/libs/core/synchronization/include/hpx/synchronization/channel_spsc.hpp
+++ b/libs/core/synchronization/include/hpx/synchronization/channel_spsc.hpp
@@ -43,11 +43,11 @@ namespace hpx::lcos::local {
             {
                 next_tail = 0;
             }
-            if (next_tail == tail_.data_.head_cached_) [[unlikely]]
+            if (next_tail == tail_.data_.other_index_cached_) [[unlikely]]
             {
-                tail_.data_.head_cached_ =
-                    head_.data_.head_.load(std::memory_order_acquire);
-                if (next_tail == tail_.data_.head_cached_)
+                tail_.data_.other_index_cached_ =
+                    head_.data_.index_.load(std::memory_order_acquire);
+                if (next_tail == tail_.data_.other_index_cached_)
                 {
                     return true;
                 }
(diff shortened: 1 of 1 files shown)


libs/core/synchronization/include/hpx/synchronization/channel_spsc.hpp
```diff
@@ -212,11 +216,23 @@ namespace hpx::lcos::local {
...
+        {
+            std::atomic<std::size_t> head_{0};
+            std::size_t tail_cached_{0};
+        };
+
+        struct tail_data
+        {
+            std::atomic<std::size_t> tail_{0};
+            std::size_t head_cached_{0};
+        };
```
hkaiser, line 229: Do we need two distinct types to represent the same combination of types?

ArivoliR: I split them originally so that the variable names would self-document which side's cache is being touched. I think they can be merged though, yes. Would you prefer that?

hkaiser: The variables can be made understandable even if the types are merged. What about:
```
        struct data
        {
            std::atomic<std::size_t> index_{0};
            std::size_t other_index_cached_{0};
        };
```
?
The variables are identified by their compound names (`head_` and `tail_`).

hkaiser (approved):
LGTM, thanks!

abhishek593:
HI @ArivoliR, I am just sharing my input here. Have you benchmarked this because I also tried this during my PR, and the performance degraded. I don't know why that is the case, but it would be great if the benchmark proves no performance degradation. The blog post also talks about only 4 cache lines. Although it's @hkaiser's call.

ArivoliR:
> HI @ArivoliR, I am just sharing my input here. Have you benchmarked this because I also tried this during my PR, and the performance degraded. I don't know why that is the case, but it would be great if the benchmark proves no performance degradation. The blog post also talks about only 4 cache lines. Although it's @ hkaiser's call.

Hi, yes I did run the benchmark and did not face any degradation. A slight improvement infact. Attaching recent benchmark below:
```
➜  default git:(spsc-cache-colocation) ./bin/Release/channel_spsc_throughput_test --hpx:threads=2
Producer throughput: 1.3618e+08 [op/s] (7.34323e-09 [s/op])
Consumer throughput: 1.36181e+08 [op/s] (7.34317e-09 [s/op])
➜  default git:(spsc-cache-colocation) ./bin/Release/channel_spsc_throughput_test --hpx:threads=2
Producer throughput: 1.35675e+08 [op/s] (7.37057e-09 [s/op])
Consumer throughput: 1.35673e+08 [op/s] (7.37067e-09 [s/op])
➜  default git:(spsc-cache-colocation) ./bin/Release/channel_spsc_throughput_test --hpx:threads=2
Producer throughput: 1.37598e+08 [op/s] (7.26754e-09 [s/op])
Consumer throughput: 1.37598e+08 [op/s] (7.26757e-09 [s/op])
```

That's on average ~136M ops/s which is slightly more than the baseline ~132M ops/s from #7115

PR #7169, opened 2026-04-11
Fix parallel is_sorted_until off by one cross boundary check

Momensamir12:
Fixes #

## Proposed Changes

  -  In parallel `is_sorted_until`, `base_idx` gets incremented then it's reused to compute the cross-partition cancellation index.
  - Fix : capture the value of `base_idx` before incrementing, use it for cross partition calculations.

This should also fix `partitioned_vector_is_sorted` unit test.

## Checklist


hkaiser:
@Momensamir12 Thanks for looking into this!

hkaiser (approved):
LGTM, thanks!

Momensamir12:
> @Momensamir12 Thanks for looking into this!

My pleasure!

PR #7170, opened 2026-04-11
tests(adjacent_find): Add missing edge-case tests for empty and boundary ranges

aneek22112007-tech:
While performing a detailed audit of the parallel algorithm test suite, I noticed a significant coverage gap within `hpx::adjacent_find`. 

Currently, the primary test (`test_adjacent_find`) strictly hardcodes a ~10,000-element vector and plants two identical elements at a random interior position (specifically between indices `[2, 10005]`). This means the core algorithm logic governing extreme mathematical boundary conditions, partition boundaries, and iterator short-circuiting was historically never exercised. 

Because partitioned parallel workloads rely heavily on precise iterator math (such as calculating partition spans using `count - 1`), lacking these limits presents a high risk of unnoticed calculation underflows or boundary violations during future optimizations. 

This PR ensures these constraints are strictly validated across standard parallel execution policies.

Fixes # 

## Proposed Changes

- **Empty Range Validation:** Verifies the `if (first == last)` early-return guards avoid calculating bad iterator offsets. (Prior to this PR, CI environments building without `stdexec` enabled had zero empty-range validations).
- **Single-Element Range Coverage:** Explicitly tests the parallel traversal when `count - 1 == 0` to guarantee that partitioning logic does not trigger mathematical underflow or generate degenerate execution spans.
- **Exhaustive No-Match Searching:** Passes vectors comprised entirely of unique elements to ensure that internal cancellation tokens and multi-threaded partition boundaries behave safely when exhausting the entire dataset.
- **Absolute Boundary Offsets:** Forces adjacent duplicates at the absolute first index (`0, 1`) and the absolute last index (`N-2, N-1`) to ensure that internal partition chunk boundaries correctly handle matches located exactly at the perimeter.
- **Binary Predicate Coverage:** Extends empty-range boundary checks specifically to the custom binary predicate overload of `hpx::adjacent_find` within `adjacentfind_tests.hpp`.
- **Standardized CI Integration:** Hooks the new `test_adjacent_find_edge_cases` function into `adjacentfind.cpp` so that it natively tests against `seq`, `par`, and `par_unseq` during standard test sweeps, without introducing new, heavy CMake targets.


hkaiser (approved):
LGTM, thanks!

PR #7171, opened 2026-04-12
Add optional Orbit MPMC comparison benchmark

### LINKED ISSUE #7167: Compare performance of HPX's MPMC queue with other implementations
In particular: https://github.com/austin-hill/orbit

This could also include improving our implementation (https://github.com/STEllAR-GROUP/hpx/blob/master/libs/core/synchronization/include/hpx/synchronization/channel_mpsc.hpp), if possible

LSUDOKO:
## Summary

  This PR adds an optional performance benchmark for comparing HPX's `channel_mpmc` against the `orbit` MPMC
  queue implementation referenced in #7167.

  The benchmark is intentionally optional and uses a local `HPX_WITH_ORBIT_ROOT` path instead of vendoring
  `orbit` into HPX. It compares:

  - `hpx_channel_mpmc`
  - `orbit_default`
  - `orbit_throughput`

  under the same producer/consumer workload with the same payload type and queue capacity.

  Closes #7167.

  ## What Changed

  - Added an optional `HPX_WITH_ORBIT_ROOT` CMake path in `libs/core/synchronization/tests/performance/
  CMakeLists.txt`
  - Added `channel_mpmc_comparison.cpp`, a new benchmark that:
    - compares HPX and Orbit using the same payload type
    - uses sentinel values for shutdown instead of queue-specific `empty()` heuristics
    - supports both human-readable output and CSV output

  ## Benchmark Setup

  - CPU: `AMD Ryzen 7 5700U with Radeon Graphics`
  - Compiler: `GCC 15.2.1`
  - Build type: `Release`
  - Allocator: `system`
  - Queue capacity: `2048`
  - Values per run: `1000000`
  - Repetitions: `5`

  Cases tested:

  - `1x1`
  - `2x2`
  - `4x4`
  - `6x6`

  ## Commands Used

  ```bash
  cmake -S /home/arpit/Desktop/Gsoc/hpx -B /home/arpit/Desktop/Gsoc/hpx/build \
    -DHPX_WITH_ORBIT_ROOT=/home/arpit/Desktop/Gsoc/hpx/orbit

  cmake --build /home/arpit/Desktop/Gsoc/hpx/build --target channel_mpmc_comparison_test -j2

  /home/arpit/Desktop/Gsoc/hpx/build/bin/channel_mpmc_comparison_test \
    --csv --capacity 2048 --producers 1 --consumers 1 --values 1000000 --repetitions 5

  /home/arpit/Desktop/Gsoc/hpx/build/bin/channel_mpmc_comparison_test \
    --csv --capacity 2048 --producers 2 --consumers 2 --values 1000000 --repetitions 5

  /home/arpit/Desktop/Gsoc/hpx/build/bin/channel_mpmc_comparison_test \
    --csv --capacity 2048 --producers 4 --consumers 4 --values 1000000 --repetitions 5

  /home/arpit/Desktop/Gsoc/hpx/build/bin/channel_mpmc_comparison_test \
    --csv --capacity 2048 --producers 6 --consumers 6 --values 1000000 --repetitions 5
```
  ## Results

  Average throughput:

  - 1x1
      - HPX: 19.96M op/s
      - Orbit default: 16.68M op/s
      - Orbit throughput: 21.42M op/s
  - 2x2
      - HPX: 19.44M op/s
      - Orbit default: 16.33M op/s
      - Orbit throughput: 20.40M op/s
  - 4x4
      - HPX: 15.82M op/s
      - Orbit default: 12.39M op/s
      - Orbit throughput: 15.97M op/s
  - 6x6
      - HPX: 14.94M op/s
      - Orbit default: 11.89M op/s
      - Orbit throughput: 18.50M op/s

  ## Analysis

  - HPX outperforms orbit_default in all tested configurations.
  - orbit_throughput is slightly faster than HPX in 1x1, 2x2, and 4x4.
  - The difference becomes more noticeable at 6x6, where orbit_throughput is clearly ahead.
  - HPX remains competitive across all tested cases and consistently outperforms Orbit's default latency-
    oriented configuration.

  ## Attachments
image

### DIFF
--- a/libs/core/synchronization/tests/performance/CMakeLists.txt
+++ b/libs/core/synchronization/tests/performance/CMakeLists.txt
@@ -10,9 +10,9 @@ set(benchmarks channel_mpmc_throughput channel_mpsc_throughput
 
 hpx_option(
   HPX_WITH_ORBIT_ROOT
-  STRING
+  BOOL
   "Path to a local checkout of austin-hill/orbit used for optional synchronization comparison benchmarks."
-  ""
+  OFF
   ADVANCED
 )
 
@@ -42,6 +42,7 @@ foreach(benchmark ${benchmarks})
 endforeach()
 
 if(HPX_WITH_ORBIT_ROOT)
+  # Orbit can be found at https://github.com/austin-hill/orbit.
   if(EXISTS "${HPX_WITH_ORBIT_ROOT}/include/orbit/mpmc_queue.h")
     add_hpx_executable(
       channel_mpmc_comparison_test INTERNAL_FLAGS
--- a/libs/core/synchronization/tests/performance/CMakeLists.txt
+++ b/libs/core/synchronization/tests/performance/CMakeLists.txt
@@ -8,10 +8,11 @@ set(benchmarks channel_mpmc_throughput channel_mpsc_throughput
                channel_spsc_throughput
 )
 
-set(
-  HPX_WITH_ORBIT_ROOT ""
-  CACHE PATH
-        "Path to a local checkout of austin-hill/orbit used for optional synchronization comparison benchmarks"
+set(HPX_WITH_ORBIT_ROOT
+    ""
+    CACHE
+      PATH
+      "Path to a local checkout of austin-hill/orbit used for optional synchronization comparison benchmarks"
 )
 
 set(channel_mpmc_throughput_PARAMETERS THREADS_PER_LOCALITY 2)
@@ -49,7 +50,8 @@ if(HPX_WITH_ORBIT_ROOT)
     )
 
     target_include_directories(
-      channel_mpmc_comparison_test SYSTEM PRIVATE "${HPX_WITH_ORBIT_ROOT}/include"
+      channel_mpmc_comparison_test SYSTEM
+      PRIVATE "${HPX_WITH_ORBIT_ROOT}/include"
     )
 
     add_hpx_performance_test(
--- a/libs/core/synchronization/tests/performance/CMakeLists.txt
+++ b/libs/core/synchronization/tests/performance/CMakeLists.txt
@@ -8,6 +8,12 @@ set(benchmarks channel_mpmc_throughput channel_mpsc_throughput
                channel_spsc_throughput
 )
 
+set(
+  HPX_WITH_ORBIT_ROOT ""
+  CACHE PATH
+        "Path to a local checkout of austin-hill/orbit used for optional synchronization comparison benchmarks"
+)
+
 set(channel_mpmc_throughput_PARAMETERS THREADS_PER_LOCALITY 2)
 set(channel_mpsc_throughput_PARAMETERS THREADS_PER_LOCALITY 2)
 set(channel_spsc_throughputs_PARAMETERS THREADS_PER_LOCALITY 2)
@@ -32,3 +38,27 @@ foreach(benchmark ${benchmarks})
   )
 
 endforeach()
+
+if(HPX_WITH_ORBIT_ROOT)
+  if(EXISTS "${HPX_WITH_ORBIT_ROOT}/include/orbit/mpmc_queue.h")
+    add_hpx_executable(
+      channel_mpmc_comparison_test INTERNAL_FLAGS
+      SOURCES channel_mpmc_comparison.cpp
+      EXCLUDE_FROM_ALL
+      FOLDER "Benchmarks/Modules/Core/Synchronization"
+    )
+
+    target_include_directories(
+      channel_mpmc_comparison_test SYSTEM PRIVATE "${HPX_WITH_ORBIT_ROOT}/include"
+    )
+
+    add_hpx_performance_test(
+      "modules.synchronization" channel_mpmc_comparison THREADS_PER_LOCALITY 1
+    )
+  else()
+    message(
+      WARNING
+        "HPX_WITH_ORBIT_ROOT is set to '${HPX_WITH_ORBIT_ROOT}', but include/orbit/mpmc_queue.h was not found there. The optional channel_mpmc_comparison benchmark will not be added."
+    )
+  endif()
+endif()
--- a/libs/core/synchronization/tests/performance/CMakeLists.txt
+++ b/libs/core/synchronization/tests/performance/CMakeLists.txt
@@ -8,11 +8,12 @@ set(benchmarks channel_mpmc_throughput channel_mpsc_throughput
                channel_spsc_throughput
 )
 
-set(HPX_WITH_ORBIT_ROOT
-    ""
-    CACHE
-      PATH
-      "Path to a local checkout of austin-hill/orbit used for optional synchronization comparison benchmarks"
+hpx_option(
+  HPX_WITH_ORBIT_ROOT
+  STRING
+  "Path to a local checkout of austin-hill/orbit used for optional synchronization comparison benchmarks."
+  ""
+  ADVANCED
 )
 
 set(channel_mpmc_throughput_PARAMETERS THREADS_PER_LOCALITY 2)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by hkaiser** [comment]
@LSUDOKO Very nice, thanks for the analysis! Do you have an understanding what could possibly be improved in HPX?

File: libs/core/synchronization/tests/performance/CMakeLists.txt [hunk 1]
```diff
@@ -8,6 +8,13 @@ set(benchmarks channel_mpmc_throughput channel_mpsc_throughput
                channel_spsc_throughput
 )
 
+set(HPX_WITH_ORBIT_ROOT
```
hkaiser: You can use `hpx_option` for this purpose.

libs/core/synchronization/tests/performance/channel_mpmc_comparison.cpp
```diff
@@ -0,0 +1,393 @@
+//  Copyright (c) 2026
```
hkaiser: Please add your name to the copyright.

libs/core/synchronization/tests/performance/channel_mpmc_comparison.cpp
```diff
@@ -0,0 +1,393 @@
...
+        data() = default;
+
+        explicit data(std::uint64_t value)
+          : value_(value)
+        {
+        }
+
+        std::uint64_t value_ = 0;
+        std::uint64_t padding_[7] = {};
+    };
```
hkaiser: You could use `hpx::util::cache_line_data` for the purpose of reserving a whole cache line.

libs/core/synchronization/tests/performance/channel_mpmc_comparison.cpp
```diff
@@ -0,0 +1,393 @@
...
+    private:
+        orbit::mpmc_queue<data, Capacity, MinimiseLatency> queue_;
+    };
+
+    template <typename Queue>
+    void push_value(Queue& queue, data value)
+    {
+        while (!queue.try_push(std::move(value)))
+        {
+            std::this_thread::yield();
```
hkaiser, line 94: You may want to avoid calling kernel-thread primitives from HPX threads as this will stall the user-level scheduling:
```suggestion
            hpx::this_thread::yield();
```

libs/core/synchronization/tests/performance/channel_mpmc_comparison.cpp
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    template <typename Queue>
+    double run_once(std::size_t queue_capacity, std::size_t producers,
+        std::size_t consumers, std::uint64_t total_values)
+    {
+        Queue queue(queue_capacity);
+
+        std::vector<std::thread> producer_threads;
+        std::vector<std::thread> consumer_threads;
```
hkaiser, line 129: In order to avoid oversubscription of the system, please use HPX threading facilities:
```suggestion
        std::vector<hpx::thread> producer_threads;
        std::vector<hpx::thread> consumer_threads;
```

libs/core/synchronization/tests/performance/channel_mpmc_comparison.cpp
LSUDOKO: okay sir

LSUDOKO:
> @LSUDOKO Very nice, thanks for the analysis! Do you have an understanding what could possibly be improved in HPX?

@hkaiser 
i think the most obvious improvement area in HPX is the current `channel_mpmc` implementation itself

 right now `channel_mpmc` is built on a single mutex-protected bounded channel, while `orbit` uses a lock-free
  bounded ring with per-slot state transitions. Based on the benchmark results, HPX stays competitive, but the gap to
  Orbit's throughput-oriented mode grows as producer/consumer contention increases, especially in the `6x6` case

  so the main things I would look at are:

  - reducing or eliminating the single shared lock in `channel_mpmc`
  - separating producer and consumer progress more aggressively to reduce contention
  - using per-slot state / reservation logic for a bounded lock-free MPMC ring buffer
  - tuning backoff/spin behavior for higher thread counts

  i’ve kept this PR focused on adding the comparison benchmark so we now have a reproducible baseline for evaluating those kinds of changes

libs/core/synchronization/tests/performance/CMakeLists.txt
LSUDOKO: okay sir

libs/core/synchronization/tests/performance/channel_mpmc_comparison.cpp
LSUDOKO: okay sir

libs/core/synchronization/tests/performance/channel_mpmc_comparison.cpp
```diff
@@ -0,0 +1,393 @@
...
+    {
+        explicit queue_data(std::uint64_t value = 0)
+          : value_(value)
+        {
+        }
+
+        std::uint64_t value_ = 0;
+    };
+
+    using data = hpx::util::cache_line_data<queue_data>;
```
hkaiser: Wouldn't writing this as 
```suggestion
    using queue_data = hpx::util::cache_line_data<std::uint64_t>;
```
allow to remove the explicit queue_data type above?

libs/core/synchronization/tests/performance/channel_mpmc_comparison.cpp
```diff
@@ -0,0 +1,393 @@
...
+#include <numeric>
+#include <stdexcept>
+#include <string_view>
+#include <thread>
+#include <utility>
+#include <vector>
+
+namespace {
+
+    struct queue_data
```
hkaiser: This is not needed anymore, I believe

LSUDOKO: you're right  struct queue_data was no longer needed after switching to hpx::util::cache_line_data
i removed the wrapper and now use `hpx::util::cache_line_data<std::uint64_t>` directly

libs/core/synchronization/tests/performance/CMakeLists.txt
```diff
@@ -8,6 +8,14 @@ set(benchmarks channel_mpmc_throughput channel_mpsc_throughput
                channel_spsc_throughput
 )
 
+hpx_option(
+  HPX_WITH_ORBIT_ROOT
+  STRING
+  "Path to a local checkout of austin-hill/orbit used for optional synchronization comparison benchmarks."
+  ""
+  ADVANCED
+)
```
hkaiser, line 17:
```suggestion
hpx_option(
  HPX_WITH_ORBIT_ROOT
  BOOL
  "Path to a local checkout of austin-hill/orbit used for optional synchronization comparison benchmarks."
  OFF
  ADVANCED
)
```

libs/core/synchronization/tests/performance/CMakeLists.txt
```diff
@@ -32,3 +40,28 @@ foreach(benchmark ${benchmarks})
   )
 
 endforeach()
+
+if(HPX_WITH_ORBIT_ROOT)
+  if(EXISTS "${HPX_WITH_ORBIT_ROOT}/include/orbit/mpmc_queue.h")
```
hkaiser, line 46: Please add a comment here where the Orbit library can be found.

libs/core/synchronization/tests/performance/channel_mpmc_comparison.cpp
```diff
@@ -0,0 +1,383 @@
...
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <hpx/init.hpp>
+#include <hpx/modules/concurrency.hpp>
+#include <hpx/modules/program_options.hpp>
+#include <hpx/modules/synchronization.hpp>
+
+#include <orbit/mpmc_queue.h>
```
hkaiser, line 13: Please add a comment here where the Orbit library can be found.

hkaiser (approved):
LGTM, thanks!

PR #7175, opened 2026-04-12
tests(algorithms): ensure cross-policy consistency and add missing edge cases

aneek22112007-tech:
While working on the recent `find_first_of` early-termination fix, I noticed that our test suite doesn't explicitly verify that `seq`, `par`, and `par_unseq` return identical results for the exact same inputs. It also missed a few basic boundary conditions (like the "not found" path for parallel execution).

This PR adds some straightforward regression tests to make sure the standard execution policies never diverge from the sequential behavior. No API or implementation logic was changed.

Changes:
- `find` & `find_if`: Added a cross-policy helper that checks 7 deterministic scenarios (empty range, single element, partition boundaries, complete miss, and multiple matches). Also removed a leftover `std::cout` debug print in `findif_tests.hpp`.
- `find_first_of`: Added standard-compliance checks for empty needle ranges. Added a cross-policy test that acts as a direct regression guard for the recent inner-loop short-circuiting bug.
- `mismatch`: The empty-range test was previously only running under the `HPX_HAVE_STDEXEC` sender path. Extended this to run for `seq`, `par`, and `par_unseq`.
- `reduce_by_key`: Added basic boundary tests for empty inputs, all-identical keys (single group), and all-unique keys.

These additions should give us a bit more confidence when tweaking or optimizing the parallel algorithm implementations in the future.


hkaiser:
@aneek22112007-tech Please have a look at the inspect issues reported.

aneek22112007-tech:
> @aneek22112007-tech Please have a look at the inspect issues reported.

@hkaiser Thanks for catching that! The inspect issue was caused by some non-ASCII em-dashes (—) in my code comments. I've stripped them out and forced an update to the branch. The inspect checks should be green now!

hkaiser:
@aneek22112007-tech inspect is still not happy:
```
/libs/core/algorithms/tests/unit/algorithms/findfirstof.cpp: *I* missing #include (type_traits) for symbol std::is_same_v on line 89
```

aneek22112007-tech:
> @aneek22112007-tech inspect is still not happy:
> 
> ```
> /libs/core/algorithms/tests/unit/algorithms/findfirstof.cpp: *I* missing #include (type_traits) for symbol std::is_same_v on line 89
> ```

@hkaiser Good catch! I missed the <type_traits> header for std::is_same_v when I pulled out the edge case tests earlier. I've just added the include and forced an update. Hopefully, inspect is happy this time!

hkaiser (approved):
LGTM, thanks!

PR #7180, opened 2026-04-13
Add static plugin loading support for HPX plugins

abhishek593:
## Proposed Changes

Plugins (binary filters, message handlers) could not be loaded in statically linked HPX builds. `load_plugins` threw "static loading of plugin not supported". Components already had full static loading infrastructure, but plugins were missing the parallel machinery.

- Added a dedicated static plugin registry (`init_registry_plugin_module`, `init_registry_plugin_factory`, `get_static_plugin_factory`) that mirrors the existing component-side registries but keeps plugin bookkeeping in separate maps.

- Added `load_plugin_factory_static()` which iterates plugin registries and injects `static = 1` into each `[hpx.plugins.*]` ini section. Added `load_plugin_static()` in runtime_support to instantiate plugin factories from the static registry.

- Added unit tests covering the registry infrastructure, `static = 1` injection for single and multi-plugin modules, and the empty-factory edge case.

## Checklist

### DIFF
--- a/libs/core/runtime_configuration/include/hpx/runtime_configuration/macros.hpp
+++ b/libs/core/runtime_configuration/include/hpx/runtime_configuration/macros.hpp
@@ -141,22 +141,45 @@
 ////////////////////////////////////////////////////////////////////////////////
 // from hpx/runtime_configuration/plugin_registry_base.hpp
 
-/// This macro is used to register the given component factory with Hpx.Plugin.
-/// This macro has to be used for each of the components.
+/// This macro is used to register the given plugin factory with Hpx.Plugin.
+/// This macro has to be used for each of the plugins. Ungated, matching the
+/// component analog (HPX_REGISTER_COMPONENT_REGISTRY).
 #define HPX_REGISTER_PLUGIN_BASE_REGISTRY(PluginType, name)                    \
     HPX_PLUGIN_EXPORT(HPX_PLUGIN_PLUGIN_PREFIX,                                \
         hpx::plugins::plugin_registry_base, PluginType, name, plugin)          \
     /**/
 
+////////////////////////////////////////////////////////////////////////////////
+// Mirror the component module gating pattern (lines 66-93 above):
+//   !APP_NAME && !STATIC_LINKING -> full expansion (dynamic build, library TU)
+//   STATIC_LINKING               -> full expansion (static build, any TU)
+//   APP_NAME only                -> nothing       (dynamic build, exe TU)
+#if !defined(HPX_APPLICATION_NAME) && !defined(HPX_HAVE_STATIC_LINKING)
+
 /// This macro is used to define the required Hpx.Plugin entry points. This
-/// macro has to be used in exactly one compilation unit of a component module.
+/// macro has to be used in exactly one compilation unit of a plugin module.
 #define HPX_REGISTER_PLUGIN_REGISTRY_MODULE()                                  \
     HPX_PLUGIN_EXPORT_LIST(HPX_PLUGIN_PLUGIN_PREFIX, plugin)                   \
+    HPX_INIT_REGISTRY_PLUGIN_MODULE_STATIC(HPX_PLUGIN_PLUGIN_PREFIX, plugin)   \
     /**/
 
 #define HPX_REGISTER_PLUGIN_REGISTRY_MODULE_DYNAMIC()                          \
     HPX_PLUGIN_EXPORT_LIST_DYNAMIC(HPX_PLUGIN_PLUGIN_PREFIX, plugin)           \
     /**/
+
+#else
+
+#if defined(HPX_HAVE_STATIC_LINKING)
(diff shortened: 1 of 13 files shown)


hkaiser:
@abhishek593 Thanks for looking into this! Where did you encounter the issue you report?

abhishek593:
@hkaiser This is an existing limitation. There is a FIXME in [runtime_support_server.cpp‎](https://github.com/TheHPXProject/hpx/pull/7180/changes#diff-cdce4bd3e77372ecd8015fe76566530a18d700868e6b1f6a5d50acc373045ef4) which this PR addresses.

hkaiser:
@abhishek593 Please take care of the inspect issues reported:
```
/libs/core/runtime_configuration/tests/unit/static_plugin_registry.cpp: *I* missing #include (cstddef) for symbol std::size_t on line 272
```

libs/core/runtime_configuration/include/hpx/runtime_configuration/macros.hpp
```diff
@@ -141,23 +141,46 @@
...
     HPX_PLUGIN_EXPORT(HPX_PLUGIN_PLUGIN_PREFIX,                                \
         hpx::plugins::plugin_registry_base, PluginType, name, plugin)          \
     /**/
 
+////////////////////////////////////////////////////////////////////////////////
+// Mirror the component module gating pattern (lines 66-93 above):
+//   !APP_NAME && !STATIC_LINKING -> full expansion (dynamic build, library TU)
+//   STATIC_LINKING               -> full expansion (static build, any TU)
+//   APP_NAME only                -> nothing       (dynamic build, exe TU)
+#if !defined(HPX_APPLICATION_NAME) && !defined(HPX_HAVE_STATIC_LINKING)
```
hkaiser: Why is this not enabled for the application executable?

abhishek593: @hkaiser The macro is meant to go in a plugin source file, not in application code. It emits the list of factories that were registered in the same translation unit via HPX_REGISTER_PLUGIN_BASE_REGISTRY, plus a static ctor that hands that list to the runtime. An application TU doesn't register any factories itself, so expanding this there would just produce an empty list function and a no-op ctor. 

Although the previous definition was ungated, this now mirrors the component macros' behaviour.

hkaiser: FWIW, I don't see a reason why we shouldn't allow for a plugin to be implemented in the application itself, especially now that statically linked plugin are being supported.

hkaiser: Also, your test actually doing that.

abhishek593: Yeah, I agree. I was more focused on mirroring the components approach and didn't realize my tests are doing the exact same thing. I have dropped the `HPX_APPLICATION_NAME/HPX_HAVE_STATIC_LINKING` gate on `HPX_REGISTER_PLUGIN_REGISTRY_MODULE` and `HPX_REGISTER_PLUGIN_MODULE` so plugins can live in a shared lib, a static build, or the application exe. 

Also had to drop the matching `HPX_HAVE_STATIC_LINKING` guard in `runtime_configuration::load_modules` so exe-registered plugins are actually picked up by the loader.

abhishek593: As a follow-up, would the same reasoning apply to the component macros, or is in-exe components not applicable there for other reasons?

libs/core/runtime_configuration/tests/unit/static_plugin_registry.cpp
```diff
@@ -0,0 +1,379 @@
...
+#include <hpx/config.hpp>
+#include <hpx/modules/datastructures.hpp>
+#include <hpx/modules/errors.hpp>
+#include <hpx/modules/ini.hpp>
+#include <hpx/modules/plugin.hpp>
+#include <hpx/modules/testing.hpp>
+#include <hpx/runtime_configuration/init_ini_data.hpp>
+#include <hpx/runtime_configuration/macros.hpp>
+#include <hpx/runtime_configuration/plugin_registry_base.hpp>
+#include <hpx/runtime_configuration/static_factory_data.hpp>
```
hkaiser: Please use the generated module headers, otherwise this will potentially break the C++ module build:
```suggestion
#include <hpx/modules/runtime_configuration.hpp>
```

libs/core/runtime_configuration/include/hpx/runtime_configuration/macros.hpp
hkaiser: Components can be implemented in the executable as well.

abhishek593: Okay, I will create a separate PR for Components then.

hkaiser (approved):
LGTM, thanks!

PR #7182, opened 2026-04-13
fix: add runtime precondition checks to connection_cache public methods

abhishek593:
## Proposed Changes

Replace debug-only assert-based invariant checks with explicit precondition throws in `reclaim()` and `clear()` overloads.

- Add O(1) precondition checks to `reclaim()`, `clear(key)`, and `clear(key, conn)` that throw `hpx::error::invalid_status` before state mutation when a caller error is detected.

- Add O(n) precondition check to `clear()`, verifying no connections are checked out before wiping the cache.

## Checklist


hkaiser (approved):
LGTM, thanks!

PR #7183, opened 2026-04-13
tests(algorithms): fix return-iterator C++20 compliance for hpx::shift_left and hpx::shift_right

aneek22112007-tech:
Fixes #

## Proposed Changes

  - Separated the coalesced early-exit guards in both `hpx::shift_left` and `hpx::shift_right` to handle the `n <= 0` and `n >= dist` edge conditions independently, correctly matching the exact return-iterator requirements outlined in the C++20 `[alg.shift]` specification.
  - Resolved a general logic error in the parallel `shift_right_helper` returning the incorrect iterator. Altered the `then` continuation to advance `first` by `distance(new_first, last)` so parallel computations correctly return `first + n`. 
  - Added over 160 lines of rigorous regression testing to `shift_left.cpp` and `shift_right.cpp` to validate iterator return correctness across all boundaries (e.g., empty ranges, `n=0`, `n>=dist`, `n=1`, and normal permutations).
  - Implemented cross-policy consistency verifications comparing the returned iterators of `seq`, `par`, and `par_unseq` to guarantee they yield identical, standard-compliant results.
  - Corrected Doxygen documentation headers for `shift_left` and `shift_right` which previously provided incorrect descriptions of the returned iterators.

Prior to this PR, `hpx::shift_left` and `hpx::shift_right` contained invisible, long-standing deviations from the C++20 standard because their respective test suites only validated the manipulated *data values*, entirely ignoring the returned iterators. 

**Standard Compliance Issues Discovered:**
1. **`shift_left` empty/no-op cases:** The standard dictates that if `n <= 0`, it "does nothing and returns *last*". HPX was returning `first`.  
2. **`shift_right` empty/no-op cases:** The standard dictates that if `n >= dist`, it "does nothing and returns *last*". HPX was returning `first`.

Both bugs stemmed from a unified early-exit guard (`if (n <= 0 || static_cast<std::size_t>(n) >= dist) return first;`). This was split so that the respective correct iterators could be resolved. Since `dist` is evaluated immediately before the guard, we can efficiently return `last` via `std::next(first, dist)` without performance loss.

**Parallel State Fix:**
While designing the regression tests for the normal case (`0 < n < dist`), I discovered another bug nested within `shift_right`'s parallel implementation. The internal `shift_right_helper` was yielding an iterator pointing to the split pivot (`first + (dist - n)`) instead of the statutorily required start of the resulting range (`first + n`). This PR addresses the continuation logic to return standard-compliant iterators for standard parallel workloads.

Everything has been evaluated locally against equivalent behaviors in `libstdc++` and `libc++`, guaranteeing total alignment with external compiler expectations.

## Checklist

 for the tests.

### DIFF
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/shift_left.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/shift_left.hpp
@@ -277,9 +277,8 @@ namespace hpx {
             )
         // clang-format on
         friend typename hpx::parallel::util::detail::algorithm_result<ExPolicy,
-            FwdIter>::type
-        tag_fallback_invoke(shift_left_t, ExPolicy&& policy, FwdIter first,
-            FwdIter last, Size n)
+            FwdIter>::type tag_fallback_invoke(shift_left_t, ExPolicy&& policy,
+            FwdIter first, FwdIter last, Size n)
         {
             static_assert(std::forward_iterator<FwdIter>,
                 "Requires at least forward iterator.");
--- a/libs/core/algorithms/include/hpx/parallel/algorithms/shift_left.hpp
+++ b/libs/core/algorithms/include/hpx/parallel/algorithms/shift_left.hpp
@@ -44,7 +44,10 @@ namespace hpx {
     ///
     /// \returns  The \a shift_left algorithm returns \a FwdIter.
     ///           The \a shift_left algorithm returns an iterator to the
-    ///           end of the resulting range.
+    ///           end of the resulting range. Specifically:
+    ///           If \a n is 0 or \a n >= (last - first), does nothing and
+    ///           returns \a last and \a first respectively.
+    ///           Otherwise returns \a first + ((last - first) - n).
     ///
     template <typename FwdIter, typename Size>
     FwdIter shift_left(FwdIter first, FwdIter last, Size n);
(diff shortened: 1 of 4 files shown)


libs/core/algorithms/include/hpx/parallel/algorithms/shift_left.hpp
```diff
@@ -194,7 +200,13 @@ namespace hpx::parallel {
             {
                 auto dist =
                     static_cast<std::size_t>(detail::distance(first, last));
-                if (n <= 0 || static_cast<std::size_t>(n) >= dist)
+                // C++20 [alg.shift]: if n is 0, do nothing and return last.
+                if (n <= 0)
```
hkaiser: An unsigned integer will never be less than zero.

hkaiser (commented):
Very nice, thanks!

libs/core/algorithms/include/hpx/parallel/algorithms/shift_left.hpp
aneek22112007-tech: Good catch, thanks! I've addressed this by changing the checks to n == 0 for both algorithms. Everything matches the standard now and the tests are passing.

hkaiser (approved):
LGTM, thanks!

PR #7188, opened 2026-04-14
fix(algorithms): fix hpx::rotate return-iterator compliance and add regression tests

aneek22112007-tech:
## Proposed Changes

  - **Fix `hpx::rotate` CPO to return the correct iterator:** I noticed that the algorithm was mathematically returning the `last` iterator instead of the newly rotated `first` iterator. It turns out the Customization Point Object (CPO) was using `parallel::util::get_second_element({first, last})`. It now correctly uses a new `get_first_element` utility to return the correct location of `first`, which evaluates to `first + (last - new_first)`, making it 100% C++ Standard compliant.
  - **Implement `hpx::parallel::util::get_first_element`:** Added this missing utility inside `result_types.hpp`. It mirrors the existing `get_second_element` design and properly handles the async future-unwrapping overloads required for things like `parallel_task_policy`.
  - **Add comprehensive return-iterator compliance tests:** Added new test cases inside `unit/algorithms/rotate.cpp` specifically targeted at the return value. These validate standard compliance across tricky edge cases like `new_first == first`, `new_first == last`, single-element ranges, and entirely empty ranges.
  - **Add cross-policy parity checks:** Added a consistency check testing that `seq`, `par`, and `par_unseq` execution policies all reliably behave identical to one another in terms of their return offset.

This was quite an interesting find. While following up on my recent work hardening the `shift_left` / `shift_right` algorithms, I started auditing the test suite for `hpx::rotate`. I realized that historically, the `rotate` tests simply discarded the return value or just called `.wait()` on the future without ever asserting what was actually returned.

Because there were zero assertions checking the returned iterator, a dormant routing bug in the `hpx::rotate_t` CPO was silently hiding. It simply used `get_second_element` (likely a copy-paste artifact from `rotate_copy`), which meant `rotate` always returned `last` under the hood. 

Writing the new regression tests exposed this immediately (all tests failed because they received `std::distance == 100` instead of the expected rotated offset). With `get_first_element` implemented, everything is properly wired, the C++ Standard mandates are met, and the regressions prove it's fixed across all policies.

*(Note: The `rotate_copy` implementation correctly assumes `get_second_element` for the Out iterator, so it remains safely unaffected by this).*

## Checklist


hkaiser (approved):
LGTM, excellent catch! Many thanks!

PR #7191, opened 2026-04-16
[parallel] Fix hpx::nth_element to correctly support C++20 projections and sentinels

aneek22112007-tech:
This PR fixes a regression where `hpx::nth_element` ignored user-provided projections in its sequential and parallel paths. It also adds full support for sentinels and aligns the algorithm with C++20 Ranges standards.

## Proposed Changes

  - **CPO Enhancement & Sentinel Support**: Upgraded `hpx::nth_element_t` to natively accept C++20 sentinels (`Sent`) and projections (`Proj`) across both sequential and parallel execution overloads. 
  - **Robust Internal Projection Handling**: A major issue was that internal helper algorithms (`min_element`, `sort`, `make_heap`) were being called with raw, un-projected comparators. I have standardized all of these internal calls to cleanly wrap the base comparator with `hpx::parallel::util::compare_projected`, guaranteeing that custom projections are reliably evaluated before any structural comparisons are made.
  - **Parallel Partition Correctness**: Fixed a deeply hidden type-mismatch bug inside the parallel `hpx::parallel::detail::partition` lambda and `pivot9` selection logic. The parallel lambda now correctly evaluates the projected value directly in the traversal loop. This resolves the compiler errors that would trigger whenever users passed in non-trivial type-transforming projections.
  - **Comprehensive Regression Tests**: Added `nth_element_projection.cpp` to the unit test suite, aggressively testing both struct-member projections (`&S::val`) and complex type-transforming projections (e.g. mapping internal strings to sizes) across `seq`, `par`, and `par_unseq` execution policies.

While working through the parallel algorithm surfaces, I noticed that `hpx::nth_element` had an incomplete implementation of the C++20 Ranges specification. Specifically, while the CPO technically accepted custom projections, the projection targets were inadvertently dropped when the work was dispatched to internal loops.

This meant that if a user tried to partition a dataset by using a projection that fundamentally altered the type—for example, projecting a list of strings out to their integer lengths—the parallel partitioner would attempt to compare a raw `std::string` directly against a `size_t`, failing entirely at compile time. 

This PR properly threads the projection directly into the innermost parallel computation lambdas, tightening the CPO layer and effectively hardening `hpx::nth_element` against modern C++20 usage patterns to match the rest of the HPX algorithms ecosystem!

## Checklist


libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
```diff
@@ -347,42 +365,104 @@ namespace hpx {
...
+                    typename hpx::parallel::traits::projected_result_of<Proj,
+                        RandomIt>::type,
+                    typename hpx::parallel::traits::projected_result_of<Proj,
+                        RandomIt>::type
                 >
             )
         // clang-format on
         friend void tag_fallback_invoke(hpx::nth_element_t, RandomIt first,
-            RandomIt nth, RandomIt last, Pred pred = Pred())
+            RandomIt nth, RandomIt last, Pred pred = Pred(), Proj proj = Proj())
```
hkaiser: I think that in terms of the API, we should stay conforming to the standard. `std::nth_element` does not support projections, however `std::ranges::nth_element` does (see https://eel.is/c++draft/alg.nth.element). That means that our implementation infrastructure should support those even if they are not exposed through the `hpx::nth_element` API.

aneek22112007-tech: Thank you for the feedback. I have updated the PR to address these points:
Reverted `hpx::nth_element` to be strictly standard-compliant (iterators only, no projections or sentinels).
Moved all projection and sentinel support to `hpx::ranges::nth_element`, which now correctly forwards to the underlying infrastructure.
Updated and moved the projection tests to `libs/core/algorithms/tests/unit/container_algorithms/nth_element_range.cpp` to properly validate the ranges implementation.
Synced the branch with the latest master.

The implementation infrastructure still supports all features internally as requested.

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
```diff
@@ -186,9 +189,12 @@ namespace hpx::parallel {
...
             if (nth == first)
             {
+                using wrapped_comp_type =
+                    hpx::parallel::util::compare_projected<
+                        std::decay_t<Compare>, std::decay_t<Proj>>;
                 RandomIt it = detail::min_element<RandomIt>().call(
-                    hpx::execution::seq, first, end, HPX_FORWARD(Compare, comp),
-                    HPX_FORWARD(Proj, proj));
+                    hpx::execution::seq, first, end,
+                    wrapped_comp_type(comp, proj), hpx::identity_v);
```
hkaiser: What is the rationale of this change?

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
```diff
@@ -200,19 +206,30 @@ namespace hpx::parallel {
...
             if (level == 0)
             {
-                std::make_heap(first, end, comp);
-                std::sort_heap(first, nth, comp);
+                using wrapped_comp_type =
+                    hpx::parallel::util::compare_projected<
+                        std::decay_t<Compare>, std::decay_t<Proj>>;
+                hpx::parallel::detail::make_heap<RandomIt>().call(
+                    hpx::execution::seq, first, end,
+                    wrapped_comp_type(comp, proj), hpx::identity_v);
```
hkaiser: `std::make_heap` would work just fine as well.

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
```diff
@@ -200,19 +206,30 @@ namespace hpx::parallel {
...
+                std::sort_heap(first, nth, wrapped_comp_type(comp, proj));
                 return;
             }
 
             // Filter the range and check which part contains the nth element
-            RandomIt c_last = filter(first, end, comp);
+            using wrapped_comp_type =
```
hkaiser: Defining this once inside this function would be sufficient.

hkaiser:
@aneek22112007-tech Please pay attention to the CIs. The clang-format CI raised some issue.

aneek22112007-tech:
Thanks for the review! I’ve gone ahead and consolidated that comparison wrapper at the top of the function to clean things up. I also switched to std::make_heap as suggested. Fixed the formatting issues too, so CI should be all clear now.

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
aneek22112007-tech: I found that the underlying detail::min_element implementation in minmax.hpp actually fails on macOS Clang when a projection changes the value type (it tries to store the result in a variable of the original element type). Using the wrapped comparison here is a necessary workaround to keep the macOS builds green.

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
aneek22112007-tech: Good point! I've switched this to std::make_heap for the sequential path to keep things simpler.

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
aneek22112007-tech: Done! I’ve consolidated this into a single definition at the top of the function to clean up the redundancy.

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
hkaiser: Wouldn't it be better to fix the actual problem instead of adding a workaround here?

libs/core/algorithms/include/hpx/parallel/algorithms/minmax.hpp
```diff
@@ -404,7 +404,8 @@ namespace hpx::parallel {
                 return it;
 
             using element_type = hpx::traits::proxy_value_t<
-                typename std::iterator_traits<FwdIter>::value_type>;
+                std::decay_t<hpx::util::invoke_result_t<Proj,
+                    typename std::iterator_traits<FwdIter>::reference>>>;
```
hkaiser: I think you can remove the `typename` in this context.

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
hkaiser: Would you be able to fix this problem (in an independent PR)?

aneek22112007-tech: Thanks for the feedback! I've followed your suggestion and separated the core fix for minmax.hpp into an independent Pull Request. You can find it here: https://github.com/TheHPXProject/hpx/pull/7208.

On this current branch, I have:

Reverted the changes to minmax.hpp to match master.
Restored the wrapped_comp_type workaround in nth_element.hpp specifically for the min_element call.
This keeps this PR functional and 'Green' in CI while the core fix is being reviewed. Once the independent minmax PR is merged, I will come back here to remove the temporary workaround. Let me know if that works for you!

hkaiser: #7208 has been merged now. You can go ahead and remove the workaround you added earlier here.

aneek22112007-tech: I have removed the wrapped_comp_type workaround for the min_element call in nth_element.hpp, now that #7208 has been merged. All tests (including range-based tests) are passing correctly. Ready for final review!

libs/core/algorithms/include/hpx/parallel/algorithms/nth_element.hpp
```diff
@@ -288,17 +292,21 @@ namespace hpx::parallel {
...
-                                [val = HPX_INVOKE(proj, *first), &pred](
-                                    value_type const& elem) {
-                                    return HPX_INVOKE(pred, elem, val);
+                                [val = HPX_INVOKE(proj, *first), &pred, &proj](
+                                    auto const& elem) {
+                                    return HPX_INVOKE(
+                                        pred, HPX_INVOKE(proj, elem), val);
                                 },
-                                proj);
+                                hpx::identity_v);
```
hkaiser, line 309: Could you please explain why this change is necessary?

aneek22112007-tech: The short answer is: to avoid applying the projection twice.

Here's what happens without the change. The lambda captures the pivot value
as `HPX_INVOKE(proj, *first)` — so it already holds a *projected* value. Then
`partition` is called with `proj` as its outer projection argument, which means
`partition` will call `proj(elem)` on each element before passing it into the
lambda. But the lambda's parameter is `value_type const&` (the raw, unprojected
type) — so at that point `proj` has already transformed the element into
something else (say, an `int`), and the lambda receives the wrong type entirely.

In practice this means the projection gets applied once by `partition` and then
the lambda tries to compare an already-projected value against `val`, which is
also already projected. Double projection, wrong result.

Passing `hpx::identity_v` instead makes `partition` leave each element alone
and hand the raw value to the lambda, which is exactly what the lambda expects
— and the comparison `pred(elem, val)` works correctly.

Hope that clarifies it! Happy to add a comment in the code if that would help.

hkaiser (approved):
LGTM, thanks!

PR #7192, opened 2026-04-16
Adding parameters hook function for partition management

kollanur:


hkaiser:
@kollanur inspect is not happy:
```
/libs/core/execution/tests/unit/executor_parameters_partition_hooks.cpp: 
    *I* missing #include (functional) for symbol std::ref on line 75, 
    *I* missing #include (hpx/assert.hpp) for symbol HPX_ASSERT on line 39, 
    *I* missing #include (string) for symbol std::string on line 122, 
    *I* missing #include (type_traits) for symbol std::true_type on line 59, 
    *I* missing #include (utility) for symbol std::pair on line 52
```

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -882,10 +885,15 @@ namespace hpx::parallel {
...
 
+                    auto tid = hpx::this_thread::get_id();
+                    hpx::execution::experimental::mark_partition(
+                        params, exec, idx, a0, b0, tid);
                     sequential_merge(std::next(first1, a0),
                         std::next(first1, a1), std::next(first2, b0),
                         std::next(first2, b1), std::next(dest, k0), comp, proj1,
                         proj2);
+                    hpx::execution::experimental::mark_partition(
+                        params, exec, idx, a0, b0, tid, std::size_t(2));
```
hkaiser: If you would like to reuse this customization point for different purposes I'd suggest to

- use a `enum class` to distinguish the different invocations
- pass this as the first argument (after the partition index)

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
hkaiser, line 736: May I suggest to add a 
```
static_assert(std::random_access_iterator<const_index_value_iterator>);
```
to make sure it's really detected as a random-access iterator. I'm currently trying to fix a problem in our iterator framework that prevents for the concept to be evaluated to true under certain circumstances.

We could have found a possible explanation for the performance issues we're currently looking into, if your iterator is not detected as random-access

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: ok. Thank you.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: Done.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -321,6 +321,12 @@ namespace hpx::parallel {
         /// \cond NOINTERNAL
 
         ///////////////////////////////////////////////////////////////////////
+        enum class partition_phase : std::size_t
```
hkaiser: Should we reduce the size of the enumerator values?
```suggestion
        enum class partition_phase : std::uint8_t
```

libs/core/execution/tests/unit/executor_parameters_partition_hooks.cpp
```diff
@@ -0,0 +1,136 @@
...
+        self.seen_.resize(num_chunks);
+    }
+
+    template <typename Executor>
+    friend void tag_override_invoke(
+        hpx::execution::experimental::mark_partition_t,
+        partition_hooks_parameters& self, Executor&&, std::size_t partition,
+        auto, std::size_t a, std::size_t b, auto) noexcept
+    {
+        HPX_ASSERT(partition < self.values_.size());
```
hkaiser:
```suggestion
        HPX_TEST(partition < self.values_.size());
```
?

libs/core/execution/tests/unit/executor_parameters_partition_hooks.cpp
```diff
@@ -0,0 +1,136 @@
...
+        self.num_chunks_ = num_chunks;
+        self.values_.resize(num_chunks);
+        self.seen_.resize(num_chunks);
+    }
+
+    template <typename Executor>
+    friend void tag_override_invoke(
+        hpx::execution::experimental::mark_partition_t,
+        partition_hooks_parameters& self, Executor&&, std::size_t partition,
+        auto, std::size_t a, std::size_t b, auto) noexcept
```
hkaiser: Should we test that the argument really is `partition_phase::begin`?

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
kollanur: ok.

libs/core/execution/tests/unit/executor_parameters_partition_hooks.cpp
kollanur: Not necessary, just to make sure to avoid out of bound.

libs/core/execution/tests/unit/executor_parameters_partition_hooks.cpp
kollanur: Yes. good idea

hkaiser:
While the Windows errors are caused by limitations in the github infrastructure, the inspect issue needs fixing:
```
/libs/core/execution/tests/unit/executor_parameters_partition_hooks.cpp: *I* missing #include (cstddef) for symbol std::size_t on line 28
```

libs/core/execution/tests/unit/executor_parameters_partition_hooks.cpp
hkaiser: Sure, but this is a test, so why not check for this?

kollanur: > Sure, but this is a test, so why not check for this?

Yes, I changed it.

libs/core/algorithms/include/hpx/parallel/algorithms/merge.hpp
```diff
@@ -321,7 +321,7 @@ namespace hpx::parallel {
         /// \cond NOINTERNAL
 
         ///////////////////////////////////////////////////////////////////////
-        enum class partition_phase : std::size_t
+        enum class partition_phase : std::uint8_t
```
hkaiser, line 325: The next thing you will see is inspect complaining about missing `#include <cstdint>` ;-)

hkaiser:
@kollanur are you satisfied with this PR now? Can we go ahead and merge?

kollanur:
> @kollanur are you satisfied with this PR now? Can we go ahead and merge?

yes.LGTM. Thanks.

hkaiser (approved):
LGTM, thanks!

PR #7193, opened 2026-04-17
Add flat-collective fallback to hierarchical_communicator for small site counts

iemAnshuman:
## Summary

Adds a flat fallback to the `hierarchical_communicator` factory for site counts below a configurable threshold. Addresses the follow-up question from @hkaiser on #7160:

> Should we fall back to the flat collectives if the number of connected sites is less than 16? This would require more measurements, but might be worth investigating.

## Benchmark context

Benchmark data from the parent PR #7160, measured on a DGX H100 compute
node with `benchmark_collectives_test`, single int, 100 iterations, one
HPX thread per locality:

| Procs | Flat multi_use (μs) | Hier arity=2 (μs) | Hier arity=4 (μs) |
|---:|---:|---:|---:|
| 4  | 50  | 101 | —     |
| 8  | 98  | 139 | 141   |
| 16 | 166 | 198 | 849   |
| 32 | 465 | 346 | 1,120 |

Hierarchical arity=2 overtakes flat at 32 processes (1.34x speedup).
Arity=4 regresses at P≥16 and is tracked separately; this PR uses
arity=2 as the recommended default for ≥16 ranks. Default threshold
set to 16 per the suggestion in #7160. The exact optimal value likely
depends on node topology and interconnect, so the threshold is kept
configurable and flagged in "open items" below.

## Design

The fallback lives entirely in `create_hierarchical_communicator`. When `num_sites < threshold`, the factory returns a `hierarchical_communicator` whose underlying `std::vector<tuple<communicator, this_site_arg>>` contains a single entry: a flat communicator spanning all N sites.

The tree-walking loops in the existing hierarchical overloads of `reduce_here`/`reduce_there`, `broadcast_to`/`broadcast_from`, `gather_here`/`gather_there` all collapse correctly when `size() == 1`. This is the **same code path non-leader sites already take** in normal tree mode (a leaf-only site has exactly one communicator in its vector), so this fallback is not a new state for the collective code — it's a state the collectives already handle in production.

This means **zero changes to `all_reduce.hpp`, `all_gather.hpp`, `reduce.hpp`, `broadcast.hpp`, `gather.hpp`, or `scatter.hpp`**. All correctness follows from the factory change.

## API change

A new `flat_fallback_threshold_arg` (default `16`) is added as the last parameter of `create_hierarchical_communicator`:

```cpp
auto comm = create_hierarchical_communicator(
    basename, num_sites, this_site, arity, generation, root_site,
    flat_fallback_threshold_arg(16));  // default
```

Pass `flat_fallback_threshold_arg(0)` to disable the fallback and always build a tree (useful for testing the tree path directly).

All existing callers use default arguments, so this is source-compatible.

## Testing

New test `hierarchical_flat_fallback` verifies:

1. **Structural invariant**: when `num_sites < threshold`, the returned `hierarchical_communicator` has `size() == 1`
2. **Correctness under fallback**: `all_reduce` on the fallback communicator produces the correct result
3. **Correctness under forced tree mode**: passing `flat_fallback_threshold_arg(0)` produces the same result
4. **Local-mode coverage**: the fallback works when multiple sites run as threads within a single process

Tested locally at 1 and 2 localities over the TCP parcelport on macOS. All existing hierarchical tests continue to pass without modification.

## Open items for discussion

- **Default threshold value.** I set it to 16 to match the suggestion in #7160, but the benchmark data above suggests a value of 24–32 may be closer to optimal on the DGX node tested. I'd appreciate input on whether to change the default or leave it at 16 pending multi-node measurements.
- **Preserving coverage of the tree path in existing tests.** I did not modify the existing hierarchical unit tests (`all_reduce_hierarchical`, etc.) to pass `flat_fallback_threshold_arg(0)`. With the default threshold of 16, those tests now exercise the fallback path for site counts 2–8 and the tree path only for 16+. If preserving tree-path coverage at small site counts is important, I can update those tests in a follow-up commit — happy to do either.
- **Multi-node validation.** I don't have multi-node cluster access at the moment. Once I do, I'd like to rerun the benchmark at 2+ nodes over InfiniBand to confirm the threshold holds.

## Follow-ups

After this merges, the same pattern generalizes naturally to `all_to_all` once that collective is implemented hierarchically.

### DIFF
--- a/libs/full/collectives/include/hpx/collectives/argument_types.hpp
+++ b/libs/full/collectives/include/hpx/collectives/argument_types.hpp
@@ -54,6 +54,7 @@ namespace hpx::collectives {
         struct root_site_tag;
         struct tag_tag;
         struct arity_tag;
+        struct flat_fallback_threshold_tag;
     }    // namespace detail
 
     /// The number of participating sites (default: all localities)
@@ -81,4 +82,12 @@ namespace hpx::collectives {
     /// The number of children each of the communication nodes is connected
     /// to (default: picked based on num_sites).
     using arity_arg = detail::argument_type<detail::arity_tag, 2>;
+
+    /// The site-count threshold below which a hierarchical_communicator
+    /// collapses to a single flat communicator spanning all sites. At small
+    /// site counts, flat collectives outperform hierarchical composition
+    /// because tree-walking overhead exceeds the synchronization-depth
+    /// benefit. Pass 0 to disable the fallback and always build a tree.
+    using flat_fallback_threshold_arg =
+        detail::argument_type<detail::flat_fallback_threshold_tag, 16>;
 }    // namespace hpx::collectives
--- a/libs/full/collectives/include/hpx/collectives/create_communicator.hpp
+++ b/libs/full/collectives/include/hpx/collectives/create_communicator.hpp
@@ -136,8 +136,8 @@ namespace hpx { namespace collectives {
         this_site_arg this_site = this_site_arg(),
         arity_arg arity = arity_arg(),
         generation_arg generation = generation_arg(),
-        root_site_arg root_site = root_site_arg());
-
+        root_site_arg root_site = root_site_arg(),
+        flat_fallback_threshold_arg threshold = flat_fallback_threshold_arg());
 }}
 // clang-format on
 
@@ -279,13 +279,13 @@ namespace hpx::collectives {
 
     ///////////////////////////////////////////////////////////////////////////
     struct hierarchical_communicator;
-
     HPX_EXPORT hierarchical_communicator create_hierarchical_communicator(
         char const* basename, num_sites_arg num_sites = num_sites_arg(),
         this_site_arg this_site = this_site_arg(),
         arity_arg arity = arity_arg(),
         generation_arg generation = generation_arg(),
-        root_site_arg root_site = root_site_arg());
+        root_site_arg root_site = root_site_arg(),
+        flat_fallback_threshold_arg threshold = flat_fallback_threshold_arg());
 
     struct hierarchical_communicator
     {
@@ -305,7 +305,8 @@ namespace hpx::collectives {
         friend HPX_EXPORT hierarchical_communicator
         create_hierarchical_communicator(char const* basename,
             num_sites_arg num_sites, this_site_arg this_site, arity_arg arity,
-            generation_arg generation, root_site_arg root_site);
+            generation_arg generation, root_site_arg root_site,
+            flat_fallback_threshold_arg threshold);
 
     public:
         hpx::tuple<communicator, this_site_arg>& operator[](std::size_t idx)
--- a/libs/full/collectives/src/create_communicator.cpp
+++ b/libs/full/collectives/src/create_communicator.cpp
@@ -429,7 +429,8 @@ namespace hpx::collectives {
 
     hierarchical_communicator create_hierarchical_communicator(
         char const* basename, num_sites_arg num_sites, this_site_arg this_site,
-        arity_arg arity, generation_arg generation, root_site_arg root_site)
+        arity_arg arity, generation_arg generation, root_site_arg root_site,
+        flat_fallback_threshold_arg threshold)
     {
         if (num_sites.is_default())
         {
@@ -447,7 +448,6 @@ namespace hpx::collectives {
         {
             this_site = this_site - root_site % num_sites;
         }
-
         HPX_ASSERT(this_site < num_sites);
         HPX_ASSERT(
             root_site != static_cast<std::size_t>(-1) && root_site < num_sites);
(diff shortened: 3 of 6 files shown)


hkaiser:
> Benchmark data from the parent PR https://github.com/TheHPXProject/hpx/pull/7160, measured on a DGX H100 compute
node with benchmark_collectives_test, single int, 100 iterations, one
 HPX thread per locality:

Could you please collect benchmark data when running with more than one thread per locality? Using one thread per locality is not a representative use case.

iemAnshuman:
> > Benchmark data from the parent PR #7160, measured on a DGX H100 compute
> > node with benchmark_collectives_test, single int, 100 iterations, one
> > HPX thread per locality:
> 
> Could you please collect benchmark data when running with more than one thread per locality? Using one thread per locality is not a representative use case.

Thanks for the review. You are right, 1 thread per locality is not representative.

I am setting up a multi thread sweep covering 16 to 256 processes at 1, 4, and 8 threads per locality, with 4B and 1MiB message sizes and 100 iteration averaging (matching the methodology from the IPTW 2025 talk). My UPES DGX queue is heavily loaded at the moment with a previous job pending 70+ hours, so turnaround may take a few days. I have also reached out to @constracktor in case a run on medusa is feasible in parallel.

Keeping the PR as draft until the data is in. I will post results as a comment here as soon as I have them.

hkaiser:
BTW, is the test failure for all_reduce_sync (on MacOS) related to your recent changes?

iemAnshuman:
> BTW, is the test failure for all_reduce_sync (on MacOS) related to your recent changes?

No, unrelated. all_reduce_sync wraps the flat all_reduce overload, which goes through create_communicator. This PR only modifies create_hierarchical_communicator and adds an optional argument with a default; all_reduce.hpp is untouched. The git diff against master shows the changes are confined to create_communicator.{hpp,cpp}, argument_types.hpp, the new test, and the benchmark.

The macOS failure is a timeout (not an assertion) on debug TCP, and the other failing check is transpose_smp_block on Windows debug, unrelated to collectives. Both look like the usual debug CI flakes. Happy to re-run the failed jobs if you would like a clean board.

iemAnshuman:
@hkaiser Opened #7198 with companion non-power-of-arity tests for hierarchical all_reduce and all_gather. Test only, no production changes. Happy to fold into this PR instead if you prefer a single commit rather than two PRs.

hkaiser (approved):
LGTM, thanks!

hkaiser:
@iemAnshuman Can we merge this now (I un-marked it as draft)?

iemAnshuman:
> @iemAnshuman Can we merge this now (I un-marked it as draft)?

Yes, please merge. Thanks for the review. @constracktor is running the full HPC benchmark sweep on a larger cluster; I will post the results as a follow up comment once they land, and open a tuning PR if the threshold needs adjustment.

iemAnshuman:
@hkaiser 
## First pass: TCP benchmark results

Thanks @constracktor for running these. Configuration:

- Parcelport: TCP
- Localities per node: `lpn = 16`
- Threads per locality: `threads = 1`
- Iterations: `10`
- Payload sizes shown:
  - large payload: `size = 262144`
  - small payload: `size = 1`
- Speedup is computed as:

```text
best flat runtime / best hierarchical runtime
where:
  best flat = min(flat multi_use, flat single_use)
  best hierarchical = min(hierarchical arity=2, hierarchical arity=4)
```

### `all_gather`, large payload (`size = 262144`)

| Nodes | Localities | Flat `multi_use` (s) | Flat `single_use` (s) | Hier arity=2 (s) | Hier arity=4 (s) | Best hier speedup vs best flat |
| ----: | ---------: | -------------------: | --------------------: | ---------------: | ---------------: | -----------------------------: |
|     1 |         16 |                0.173 |                 0.211 |            0.130 |            0.164 |                          1.33× |
|     2 |         32 |                0.975 |                 1.060 |            0.371 |            0.404 |                          2.63× |
|     4 |         64 |                2.968 |                 3.197 |            0.875 |            0.883 |                          3.39× |
|     8 |        128 |               11.708 |                14.774 |            2.124 |            1.989 |                          5.89× |
|    16 |        256 |               50.261 |                56.527 |            4.606 |            3.981 |                         12.63× |

### `all_reduce`, large payload (`size = 262144`)

| Nodes | Localities | Flat `multi_use` (s) | Flat `single_use` (s) | Hier arity=2 (s) | Hier arity=4 (s) | Best hier speedup vs best flat |
| ----: | ---------: | -------------------: | --------------------: | ---------------: | ---------------: | -----------------------------: |
|     1 |         16 |                0.023 |                 0.029 |            0.044 |            0.030 |                          0.79× |
|     2 |         32 |                0.064 |                 0.081 |            0.057 |            0.042 |                          1.52× |
|     4 |         64 |                0.167 |                 0.511 |            0.092 |            0.048 |                          3.47× |
|     8 |        128 |                0.380 |                 0.985 |            0.086 |            0.068 |                          5.59× |
|    16 |        256 |                0.826 |                 1.325 |            0.097 |            0.105 |                          8.53× |

### Small-payload check (`size = 1`)

The small-payload cases are useful because they show why fallback/arity tuning still matters.

#### `all_gather`, small payload

| Nodes | Localities | Flat `multi_use` (s) | Flat `single_use` (s) | Hier arity=2 (s) | Hier arity=4 (s) | Best hier speedup vs best flat |
| ----: | ---------: | -------------------: | --------------------: | ---------------: | ---------------: | -----------------------------: |
|     1 |         16 |                0.009 |                 0.016 |            0.021 |            0.018 |                          0.51× |
|     2 |         32 |                0.010 |                 0.020 |            0.026 |            0.021 |                          0.48× |
|     4 |         64 |                0.013 |                 0.038 |            0.030 |            0.026 |                          0.51× |
|     8 |        128 |                0.019 |                 0.064 |            0.028 |            0.025 |                          0.77× |
|    16 |        256 |                0.034 |                 0.113 |            0.038 |            0.030 |                          1.12× |

#### `all_reduce`, small payload

| Nodes | Localities | Flat `multi_use` (s) | Flat `single_use` (s) | Hier arity=2 (s) | Hier arity=4 (s) | Best hier speedup vs best flat |
| ----: | ---------: | -------------------: | --------------------: | ---------------: | ---------------: | -----------------------------: |
|     1 |         16 |                0.009 |                 0.016 |            0.023 |            0.017 |                          0.54× |
|     2 |         32 |                0.010 |                 0.019 |            0.025 |            0.021 |                          0.46× |
|     4 |         64 |                0.014 |                 0.048 |            0.029 |            0.037 |                          0.49× |
|     8 |        128 |                0.020 |                 0.060 |            0.031 |            0.025 |                          0.83× |
|    16 |        256 |                0.032 |                 0.301 |            0.048 |            0.062 |                          0.67× |

### Initial observations

- For large payloads, the hierarchical variants scale much better than the flat `single_use` / `multi_use` variants.
- The strongest case is `all_gather`: at 256 localities and `size = 262144`, the best hierarchical runtime is about `3.981s`, compared with the best flat runtime of about `50.261s`, a `12.63×` improvement.
- `all_reduce` shows the same direction at scale: at 256 localities and `size = 262144`, the best hierarchical runtime is about `0.097s` vs best flat `0.826s`, an `8.53×` improvement.
- For small payloads, flat `multi_use` is often faster at small/medium locality counts. This supports keeping fallback behavior and treating arity/fallback as benchmark-driven choices.
- Arity choice is workload- and scale-dependent. Arity 4 is often better for `all_gather` at larger scale; for `all_reduce`, the best arity varies by node count and payload size.

hkaiser:
Thank you for the extensive performance analysis!

PR #7194, opened 2026-04-17
Enable static component registration in application executables

abhishek593:
Follow-up to PR #7180 

## Proposed Changes

- Remove the app-exe gate so the ctors fire in application TUs as well, mirroring what #7180 did for plugin modules.

- Link the new test against `hpx_runtime_configuration` (not `hpx_core`):             
  `load_component_factory_static` is declared `HPX_CXX_CORE_EXPORT`, which is only active under the C++20 modules build. In the default dynamic build, it resolves to empty, so the symbol is hidden in `libhpx_core.so` by `-fvisibility=hidden` but remains globally visible in the per-module archive. Other tests in this directory already follow the same pattern.

## Checklist


hkaiser (approved):
LGTM, thanks!

PR #7196, opened 2026-04-17
[tests] Add return-iterator compliance and edge-case regression tests for hpx::partial_sort

aneek22112007-tech:
## Summary

The existing `partial_sort.cpp` test suite verified only that the sorted prefix contained correct values, but **never validated the return value** of `hpx::partial_sort` across any execution policy. According to the C++ standard (and HPX documentation), `partial_sort` must return an iterator pointing to `last` -- this was completely untested.

Additionally, the following boundary conditions had zero coverage:
- Empty range (`first == last`)
- - `middle == first` (zero elements to sort)
- - `middle == last` (degenerates to full sort)
- - Single-element range
- - Already-sorted input
- 
- ## Changes## Summary

The existing `partial_sort.cpp` test suite verified only that the sorted prefix contained correct values, but **never validated the return value** of `hpx::partial_sort` across any execution policy. According to the C++ standard (and HPX documentation), `partial_sort` must return an iterator pointing to `last` -- this was completely untested.

Additionally, the following boundary conditions had zero coverage:
- Empty range (`first == last`)
- `middle == first` (zero elements to sort)
- `middle == last` (degenerates to full sort)
- Single-element range
- Already-sorted input

## Changes

- **`libs/core/algorithms/tests/unit/algorithms/partial_sort.cpp`** (test-only change)
  - Added `result == B.end()` assertions to all three existing test functions: sequential, parallel, and async variants.
  - Added new `test_partial_sort_edge_cases()` function covering all boundary conditions listed above across all policies (`seq`, `par`, `par_unseq`, and no-policy sequential).
    
## Testing

All new tests validate both **return value correctness** and **content correctness** (where applicable). The code is `clang-format` clean (verified locally with `--dry-run --Werror`).

## Why This Matters

Return-iterator compliance is a correctness guarantee of the C++ standard. Callers chaining operations on the returned iterator rely on it pointing to exactly `last`. Without this regression test, a future refactor could silently break this contract.- 
- - **`libs/core/algorithms/tests/unit/algorithms/partial_sort.cpp`** (test-only change)
-   - Added `result == B.end()` assertions to all three existing test functions: sequential, parallel, and async variants.
-   - Added new `test_partial_sort_edge_cases()` function covering all boundary conditions listed above across all policies (`seq`, `par`, `par_unseq`, and no-policy sequential).
 
- ## Testing

- All new tests validate both **return value correctness** and **content correctness** (where applicable). The code is `clang-format` clean (verified locally with `--dry-run --Werror`).
- ## Why This Matters

- Return-iterator compliance is a correctness guarantee of the C++ standard. Callers chaining operations on the returned iterator rely on it pointing to exactly `last`. Without this regression test, a future refactor could silently break this contract.The existing `partial_sort.cpp` test suite verified only that the sorted prefix contained correct values, but **never validated the return value** of `hpx::partial_sort` across any execution policy. According to the C++ standard (and HPX documentation), `partial_sort` must return an iterator pointing to `last` 
## Checklist


hkaiser (approved):
LGTM, thanks!

PR #7198, opened 2026-04-18
Add non-power-of-arity tests for hierarchical all_reduce and all_gather

iemAnshuman:
## Summary

Adds `test_non_power_of_arity()` to the hierarchical all_reduce and all_gather unit tests, exercising site counts that are not clean multiples of the arity.

## Cases covered

- arity=2: num_sites in {3, 5, 6, 7, 9, 10, 11, 15}
- arity=4: num_sites in {5, 6, 7, 9, 10, 11, 13, 15}

These include configurations where recursion produces degenerate size-1 leaves and where the top-level partition is uneven.

## Motivation

The hierarchical tree construction in `create_communicator.cpp` handles non-power-of-arity site counts through its division_steps and remainder logic, but no existing test exercised that path. This PR closes that gap. It is companion coverage to the adaptive flat-fallback work in #7193.

## Verification

Tests were run locally on macOS (clang) at 1 locality and at 2 localities via `hpxrun.py -l 2 -t 2`. Both pass with exit 0. The non-power-of-arity block is confirmed by timing output lines such as `local timing (3/2)` and `local timing (15/4)`.

`inspect` over `libs/full/collectives/tests/unit/` reports 0 problems across 35 files scanned. No production code is touched.

### DIFF
--- a/libs/full/collectives/tests/unit/all_gather_hierarchical.cpp
+++ b/libs/full/collectives/tests/unit/all_gather_hierarchical.cpp
@@ -108,6 +108,28 @@ void test_local_use(std::uint32_t num_sites, int arity = 2)
     hpx::wait_all(std::move(sites));
 }
 
+// Non-power-of-arity coverage. The hierarchical tree construction in
+// create_communicator.cpp handles uneven partitioning via the
+// division_steps + remainder logic and degenerate single-site leaves.
+// This test exercises site counts that are not clean multiples of the
+// arity, including cases where recursion produces size-1 subgroups.
+void test_non_power_of_arity()
+{
+    // arity=2 with site counts that force uneven splits and odd-sized
+    // subtrees at multiple levels of recursion.
+    for (std::uint32_t num_sites : {3u, 5u, 6u, 7u, 9u, 10u, 11u, 15u})
+    {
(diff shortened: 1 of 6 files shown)


hkaiser (commented):
While you're at it, could you please add similar tests to the other (already existing) hierarchical collectives?

iemAnshuman:
> While you're at it, could you please add similar tests to the other (already existing) hierarchical collectives?

On it. Will extend to broadcast, gather, reduce, scatter with the same pattern. broadcast and scatter already have a single non-power-of-arity regression case; I will fold those into the new test_non_power_of_arity() function so the six hierarchical tests share one shape. Pushing shortly.

iemAnshuman:
> While you're at it, could you please add similar tests to the other (already existing) hierarchical collectives?

Extended to broadcast, gather, reduce, scatter with the same pattern. broadcast and scatter already had a single non power-of-arity regression case (test_local_use(10, 3)); I folded those into the new test_non_power_of_arity() function so the six hierarchical tests share one shape. Local runs pass at 1 and 2 localities for all four new targets.

hkaiser (approved):
LGTM, thanks!

PR #7201, opened 2026-04-19
Deduplicate Windows tests.examples exclusions

### LINKED ISSUE #7199: Deduplicate tests.examples exclusion regex generation in Windows workflows
## Summary 
  Several Windows GitHub Actions workflows duplicate the same logic for generating the `ctest --exclude-regex` value used for
  `tests.examples`.

  ## Current state

  Multiple Windows workflows currently:

  - create PowerShell aliases for `grep` and `sed`
  - read `.github/workflows/tests.examples.targets`
  - convert the file contents into a single regex passed to `ctest --exclude-regex`

  This logic appears in multiple workflow files, for example:

  - `windows_debug_vs2022.yml`
  - `windows_release_2022.yml`
  - `windows_clang_debug.yml`
  - `windows_clang_release.yml`
  - `windows_debug_vs2022_fetch_boost.yml`
  - `windows_debug_vs2022_fetch_hwloc.yml`
  - `windows_debug_vs2022_local_runtime.yml`
  - `windows_debug_vs2022_modules.yml`
  - `windows_debug_vs2022_tracy.yml`
  - `windows_release_static.yml`

  ## Problem

  The exclusion-regex generation logic is duplicated across many Windows workflows. This makes CI harder to maintain and easier to drift if
  the exclusion-file handling changes.

  It also means the same shell setup and transformation logic has to be updated in multiple places.

  ## Suggested improvement

  Move the Windows-specific exclusion-regex generation into one reusable helper, for example:

  - a small script in `.github/ci-scripts/`
  - or a shared helper path in an existing composite action

  The workflows should then call that helper instead of repeating the inline PowerShell block.

  ## Possible acceptance criteria

  - the Windows workflows no longer duplicate the alias/setup/regex-generation block
  - the exclusion regex is still derived from `.github/workflows/tests.examples.targets`
  - CI behavior remains unchanged

LSUDOKO:
## Summary

  This PR deduplicates the Windows-specific logic for generating the `ctest --exclude-regex` value used for `tests.examples`.

  Closes #7199.

  ## What changed

  - Added `.github/ci-scripts/generate_ctest_exclude_regex.ps1`
  - Updated the affected Windows workflows to call the helper script instead of repeating:
    - PowerShell alias setup for `grep` and `sed`
    - reading `.github/workflows/tests.examples.targets`
    - converting the file contents into a single regex inline

  ## Result

  - `.github/workflows/tests.examples.targets` remains the source of truth
  - Windows workflows no longer duplicate the same exclusion-regex generation block
  - CI behavior should remain unchanged

### DIFF
--- a/.github/ci-scripts/generate_ctest_exclude_regex.ps1
+++ b/.github/ci-scripts/generate_ctest_exclude_regex.ps1
@@ -1,0 +1,17 @@
+# Copyright (c) 2026 Arpit Singh
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+param(
+    [Parameter(Mandatory = $true)]
+    [string]$Path
+)
+
+$resolved_path = Resolve-Path -Path $Path -ErrorAction Stop
+$exclude_targets = Get-Content -Path $resolved_path |
+    Where-Object { $_ -notmatch '^\s*(#|$)' } |
+    ForEach-Object { $_.Trim() }
+
+[System.String]::Join('|', $exclude_targets)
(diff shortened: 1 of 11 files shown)


hkaiser (approved):
LGTM, thanks!

PR #7204, opened 2026-04-19
serialization: add reflection tests for static member function name e…

Priyanshi507:
## Summary

Adds `reflection_static_member_functions.cpp` to test `scope_builder<^^F>` 
with static member functions — extending the free function tests in 
`reflection_function_names.cpp` to cover the class member case.

## Tests Added

- Static member functions in simple namespaces
- Void return noexcept static member functions
- Inline nested namespaces (`app::nested::rpc_server::transform`)
- Traditional nested namespaces (`outer::inner::helper_server::process`)

## Context

In HPX, actions are commonly defined as static member functions of 
server classes (e.g., `compute_server::compute`). This PR verifies 
that `scope_builder` correctly extracts fully qualified names for 
this pattern, which is directly used by `reflect_action` for the 
upcoming C++26 reflection-based action system.

## Checklist

### DIFF
--- a/libs/core/serialization/tests/unit/refl/CMakeLists.txt
+++ b/libs/core/serialization/tests/unit/refl/CMakeLists.txt
@@ -14,6 +14,7 @@ set(tests
     reflection_nested_object
     reflection_nested_templ_object
     reflection_function_names
+    reflection_static_member_functions
     reflection_qualified_name_of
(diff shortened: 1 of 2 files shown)


hkaiser (approved):
LGTM, thanks!

